PR #2766, opened 2022-07-18
Hkobew/wizard submenu

Hweinstock:
## Problem
UI for searching log group was clunky and required many steps. 
## Solution
We can combine the steps in a submenu approach. This allows us to default the region with an escape hatch, so that the user can continue to search for log groups quicker, rather than this being two separate steps. 

## License


src/test/shared/ui/prompters/regionSubmenu.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        menuTester = createQuickPickTester(SubmenuPrompter.menuPrompter)
+        regionTester = createQuickPickTester(SubmenuPrompter.regionPrompter)
+    })
+
+    it('Prompts with log groups and escape hatch', function () {
+        const expectedMenuItems1 = ['testRegion1'].concat(region1Groups)
+        // console.log(expectedMenuItems1, menuTester.quickPick.items)
+        // The menuTester.assertItems passes regardless of input???
+        menuTester.assertItems(['none'])
+    })
```
Hweinstock, line 49: Why does this pass?

src/shared/ui/common/regionSubmenu.ts
```diff
@@ -0,0 +1,99 @@
...
+            {
+                label: 'Switch region',
+                data: switchRegion,
+                detail: `Showing groups for ${this.currentRegion}`,
+            },
+            ...prompter.quickPick.items,
+        ]
+
+        return prompter
+    }
```
JadenSimon, line 47: This is pretty much equivalent to adding a new method called `createMenuPrompter` because it creates a new one every single time it is called. Which is problematic for testing because we want to use the same UI element.

What you really want to do is keep reference pointing to either prompts while they're active, changing whenever the state changes. Something like:
```ts
public activePrompter?: QuickPickPrompter<T | Region | typeof switchRegion>

// Somewhere inside `promptUser`, do the same for the other state
const prompter = this.createMenuPrompter()
this.activePrompter = prompter
```

There are many, many other ways to test this code, but I'm just going based off what you've already written.

Hweinstock: I think I am close to figuring this out, but I am stuck on two remaining linting errors on [here](https://github.com/aws/aws-toolkit-vscode/pull/2766/commits/2b14a13eea6584e69ae819307252f586a19e91b6#:~:text=.currentRegion%2C-,data%3A%20resp,-%7D) and [here](https://github.com/aws/aws-toolkit-vscode/pull/2766/commits/2b14a13eea6584e69ae819307252f586a19e91b6#:~:text=.currentRegion%20%3D-,resp.id,-%7D). It seems to be angry on both of these lines since the `activePrompter` type is now variable. I tried asserting that the type was what we needed, but that didn't seem to work and feels like the wrong approach. Any hints for what I am missing here?

JadenSimon: Yeah I _very_ intentionally kept `const prompter` as a separate variable exactly for this reason.
You can also make it a one-liner:
```ts
const prompter = this.activePrompter = this.createMenuPrompter()
```

Hweinstock: I see. I just made the change without type assertions. So when we set `prompter` in this way, is the logic flow that it checks if `this.activePrompter` is set before running `this.createMenuPrompter`? I know this goal of this change is to avoid creating a new menu prompter each time so is that how this accomplishes that? Let me know if I misunderstood what you are saying in my recent commit/change.

JadenSimon: No this will re-assign `activePrompter` every time this logic is ran. We actually do want to create a new prompt on every state change, but we want to make it so the active prompt is externally visible for testing purposes. Having the assignment within the relevant logic block is just to make it easier for Typescript to understand what's going on. 

This logic can now be tested in a way that accurately reflects how this prompt would be used in practice. For example, you'd start the prompt (without `await`), then create a new quick pick tester on `activePrompter` after asserting it is not `undefined`. Use the tester to select an item. If you picked 'Switch region', you can create another quick pick tester to manipulate the new prompt. Repeat this process as needed. You should `await` the initial prompt promise at the end to see if the UI element terminated with the expected result.

Hweinstock: Ok I think I understand. So the purpose of `activePrompter` is much more geared towards testing purposes since it allows us to track externally which prompter we want to be using. Then, throughout tests, we can reference the `activePrompter` as a constant object and we no longer have to worry about it re-creating a new prompter between tests unless we change the state. Am I understanding that correctly?

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -73,35 +81,26 @@ async function logGroupsToArray(logGroups: AsyncIterableIterator<CloudWatchLogs.
...
     filterPattern: string
 }
 
 export class SearchLogGroupWizard extends Wizard<SearchLogGroupWizardResponse> {
-    public constructor(logGroups: string[]) {
+    public constructor() {
         super()
-
-        this.form.logGroup.bindPrompter(() => createLogGroupPrompter(logGroups))
+        this.form.submenuResponse.bindPrompter(createRegionSubmenu)
```
JadenSimon, line 103: nit: 
```suggestion
        this.form.submenuResponse.bindPrompter(() => new RegionSubmenu(getLogGroupsFromRegion))
```

src/shared/ui/common/regionSubmenu.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    public constructor(
+        private readonly itemsProvider: (region: string) => ItemLoadTypes<T>,
+        private readonly options?: ExtendedQuickPickOptions<T>,
+        private currentRegion = globals.awsContext.guessDefaultRegion()
+    ) {
+        super()
+    }
+
+    public createMenuPrompter() {
```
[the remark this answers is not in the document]
Hweinstock: Not really. The case I was testing is covered by the other test, so this can be private.

src/test/shared/ui/prompters/regionSubmenu.test.ts
```diff
@@ -0,0 +1,66 @@
...
+            }
+            return groupNames.map<DataQuickPickItem<string>>(groupName => ({
+                label: groupName,
+                data: groupName,
+            }))
+        }
+        submenuPrompter = new RegionSubmenu(fakeGroupProvider, {}, 'us-west-1')
+    })
+
+    it('Able to swap regions via escape hatch', async function () {
```
JadenSimon, line 38: nit:
```suggestion
    it('allows users to swap regions via escape hatch', async function () {
```

JadenSimon (approved):

PR #2767, opened 2022-07-19
Hkobew/new entrypoints

Hweinstock:
## Problem
Need to use command palette to access the log searching functionality. 

(merged wizard so that all changes don't have to be redone)
## Solution
- hover-icon to perform log search.
- option on right click of log group (pre-select log group and region here)

## License


src/cloudWatchLogs/activation.ts
```diff
@@ -66,6 +66,6 @@ export async function activate(context: vscode.ExtensionContext, configuration:
         // Here instead of in ../awsexplorer/activation due to dependence on the registry.
         Commands.register('aws.cwl.viewLogStream', async (node: LogGroupNode) => await viewLogStream(node, registry)),
 
-        Commands.register('aws.cwl.searchLogGroup', async () => await searchLogGroup(registry))
+        Commands.register('aws.cwl.searchLogGroup', async (node: LogGroupNode) => await searchLogGroup(node, registry))
```
JadenSimon, line 69: This change is lying to you:
```suggestion
        Commands.register('aws.cwl.searchLogGroup', async (node?: LogGroupNode) => await searchLogGroup(node, registry))
```
Lying types is a dangerous game. See `null` in older languages for a perfect example.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -18,25 +18,43 @@ import { createInputBox } from '../../shared/ui/inputPrompter'
...
 import { CloudWatchLogs } from 'aws-sdk'
+import { LogGroupNode } from '../explorer/logGroupNode'
 import { highlightDocument } from '../document/logStreamDocumentProvider'
 
-export async function searchLogGroup(registry: LogStreamRegistry): Promise<void> {
+export async function searchLogGroup(node: LogGroupNode, registry: LogStreamRegistry): Promise<void> {
     let result: telemetry.Result = 'Succeeded'
+    let logGroupInfo: CloudWatchLogsGroupInfo
+    let parameters: CloudWatchLogsParameters
+    let response: SearchLogGroupWizardResponse | undefined
```
JadenSimon, line 28: We're doing a lot in this function now. We should split it up to reduce the number of variables.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -98,10 +116,13 @@ export interface SearchLogGroupWizardResponse {
...
-    public constructor(logGroups: string[]) {
+    public constructor(logGroups: string[], logGroupInfo?: CloudWatchLogsGroupInfo) {
         super()
-
-        this.form.logGroup.bindPrompter(() => createLogGroupPrompter(logGroups))
+        if (!logGroupInfo) {
+            this.form.logGroup.bindPrompter(() => createLogGroupPrompter(logGroups))
+        } else {
+            this.form.logGroup.setDefault(logGroupInfo.groupName)
+        }
```
JadenSimon, line 125: This is a good approach, just want to add an alternative:
```ts
super({ initState: { logGroup: logGroupInfo?.groupName } })
```

Then do this:
```suggestion
        this.form.logGroup.bindPrompter(() => createLogGroupPrompter(logGroups))
```

This is because the wizard won't prompt if it starts with what it needs. Less branches are _usually_ a good thing although it comes at the cost of flexibility. 

We should also add a test case for the wizard about the expected flow when the wizard is given different `logGroupInfo`.

src/cloudWatchLogs/activation.ts
Hweinstock: I see, so with this change the type of node within `searchLogGroup` is `LogGroupNode | undefined` as well right?

src/cloudWatchLogs/commands/searchLogGroup.ts
Hweinstock: I just refactored it and in order to do I merged in the `wizardSubmenu` branch.

src/cloudWatchLogs/commands/searchLogGroup.ts
Hweinstock: I tried to adapt this to the new wizard I merged in. Let me know if I missed something.

src/cloudWatchLogs/activation.ts
JadenSimon: >  I see, so with this change the type of node within searchLogGroup is LogGroupNode | undefined as well right?

Yes. The interesting thing about Typescript is that it's self contained. Just because we say that a certain type is expected, that doesn't mean VS Code will give it to us. If we wanted to be _very_ strict  then we should do:
```ts
        Commands.register('aws.cwl.searchLogGroup', async (arg?: unknown) => {
            if (arg === undefined || arg instanceof LogGroupNode) {
                return searchLogGroup(arg, registry)
            } else {
                throw new TypeError('Expected argument to be `undefined` or `LogGroupNode`')
            }
        })
```

Anything that doesn't come from within the codebase should be treated as `unknown`.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -73,35 +106,36 @@ async function logGroupsToArray(logGroups: AsyncIterableIterator<CloudWatchLogs.
...
+        logGroupInfo
+            ? super({
+                  initState: {
+                      submenuResponse: {
+                          data: logGroupInfo.groupName,
+                          region: logGroupInfo.regionName,
+                      },
+                  },
+              })
+            : super()
```
JadenSimon, line 136: Huh I'm actually kind of surprised Typescript let's you do this. Just shows that the control-flow analysis is pretty good. But anyway this is probably more readable:
```suggestion
        super({
            initState: {
                submenuResponse: logGroupInfo ? {
                    data: logGroupInfo.groupName,
                    region: logGroupInfo.regionName,
                } : undefined,
            },
        })
```

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -8,63 +8,96 @@ import * as telemetry from '../../shared/telemetry/telemetry'
...
-    const regionCode = 'us-west-2'
-    const client = new DefaultCloudWatchLogsClient(regionCode)
-    const logGroups = await logGroupsToArray(client.describeLogGroups())
-    const response = await new SearchLogGroupWizard(logGroups).run()
-    if (response) {
-        const logGroupInfo: CloudWatchLogsGroupInfo = {
-            groupName: response.logGroup,
-            regionName: regionCode,
-        }
+function handleWizardResponse(response: SearchLogGroupWizardResponse, registry: LogStreamRegistry) {
```
jpinkney-aws, line 24: It's kind of implicit but
```suggestion
function handleWizardResponse(response: SearchLogGroupWizardResponse, registry: LogStreamRegistry): CloudWatchLogsData {
```

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -8,63 +8,96 @@ import * as telemetry from '../../shared/telemetry/telemetry'
...
+        logGroupInfo: logGroupInfo,
+        retrieveLogsFunction: filterLogEventsFromUriComponents,
+    }
+    return initialStreamData
+}
 
-        const parameters: CloudWatchLogsParameters = {
-            limit: registry.configuration.get('limit', 10000),
-            filterPattern: response.filterPattern,
+async function prepareDocument(uri: vscode.Uri, registry: LogStreamRegistry) {
```
jpinkney-aws, line 45:
```suggestion
async function prepareDocument(uri: vscode.Uri, registry: LogStreamRegistry): Promise<void> {
```

src/test/shared/ui/prompters/regionSubmenu.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        menuTester.assertItems(expectedMenuItems1)
+        menuTester.acceptItem('group1a')
+        await menuTester.result()
+    })
+
+    it('Able to swap regions via escape hatch', async function () {
+        type Inner<T> = T extends QuickPickPrompter<infer U> ? U : never
+        type Combine<T> = QuickPickPrompter<Inner<T>>
+
+        // const resp = SubmenuPrompter.prompt()
```
jpinkney-aws, line 53: Should this commented stuff be removed or was it only for testing something else?

src/cloudWatchLogs/activation.ts
Hweinstock: How is what you proposed stricter than having `searchLogGroup` expect a `LogGroupNode | undefined`? Wouldn't they throw the same error if the wrong type was passed in? What is the advantage of treating it as `unknown`? Is it that since its not from the codebase we don't wan't to make any assumptions?

src/test/shared/ui/prompters/regionSubmenu.test.ts
Hweinstock: Yea good catch. I think this is from when I merged in another unfinished branch to make a change.

src/cloudWatchLogs/activation.ts
JadenSimon: > How is what you proposed stricter than having searchLogGroup expect a LogGroupNode | undefined?

Typescript performs zero runtime validation of these types. VS Code is the one calling this function, not us. Anything can be passed into this function and so `LogGroupNode | undefined` is not accurate. The fact that Typescript lets you do this is because the whole point of Typescript is to give developers complete control over the types. 

You can try it for yourself:
```ts
vscode.commands.executeCommand('aws.cwl.searchLogGroup', 'hi')
```

What error do you see? Is it obvious what went wrong?

Hweinstock: It looks like it complains that `'hi` is not a function, which makes sense it's expecting one.

JadenSimon: No you should see something like this:
```
TypeError: Cannot read properties of undefined (reading 'logGroupName')
```

Hweinstock: Oh I did that wrong the first time. I see what you mean now. Yeah, it doesn't tell you that you broke the type contract, which isn't very helpful. It lets you pass in a type its not expecting, but then still tries to run the code with it, it seems.

src/test/cloudWatchLogs/commands/searchLogGroup.test.ts
```diff
@@ -45,4 +45,23 @@ describe('searchLogGroup', async function () {
         accept(testInput)
         assert.strictEqual(await result, testInput)
     })
+
+    it('wizaard prioritizes logGroupInfo if passed in', async function () {
```
JadenSimon, line 49:
```suggestion
    it('wizard prioritizes logGroupInfo if passed in', async function () {
```
May want to consider a different title. What does this mean in terms of what the user sees?

src/test/cloudWatchLogs/commands/searchLogGroup.test.ts
```diff
@@ -45,4 +45,23 @@ describe('searchLogGroup', async function () {
...
+        const fakeLogGroupNode = {
+            regionCode: 'test-region',
+            logGroup: {
+                logGroupName: 'test',
+            },
+        } as LogGroupNode
+
+        assert(fakeLogGroupNode.logGroup.logGroupName) // to avoid linting error
```
JadenSimon, line 57: Why bother with the fake node? Just pass the params directly.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -8,69 +8,99 @@ import * as telemetry from '../../shared/telemetry/telemetry'
...
 
+        response = await new SearchLogGroupWizard({
+            groupName: node.logGroup.logGroupName,
+            regionName: node.regionCode,
+        }).run()
+    } else {
+        response = await new SearchLogGroupWizard().run()
+    }
+
+    if (response) {
```
JadenSimon, line 83: Better to return early to ease the nesting:
```suggestion
    if (!response) {
        telemetry.recordCloudwatchlogsOpenStream({ result: 'Cancelled' })
        return
    }
```

src/test/cloudWatchLogs/commands/searchLogGroup.test.ts
```diff
@@ -45,4 +45,23 @@ describe('searchLogGroup', async function () {
...
+
+        const nodeTestWizard = createWizardTester(
+            new SearchLogGroupWizard({
+                groupName: fakeLogGroupNode.logGroup.logGroupName,
+                regionName: fakeLogGroupNode.regionCode,
+            })
+        )
+        nodeTestWizard.filterPattern.assertShowFirst()
```
JadenSimon, line 65:
```suggestion
        nodeTestWizard.filterPattern.assertShowFirst()
        nodeTestWizard.submenuResponse.assertDoesNotShow()
```

JadenSimon (approved):

src/test/cloudWatchLogs/commands/searchLogGroup.test.ts
Hweinstock: I was originally having it take the logGroupNode itself, but realized that it was overcomplicated and only needed this info.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -8,76 +8,107 @@ import * as telemetry from '../../shared/telemetry/telemetry'
...
+    const doc = await vscode.workspace.openTextDocument(uri) // calls back into the provider
+    vscode.languages.setTextDocumentLanguage(doc, 'log')
+
+    const textEditor = await vscode.window.showTextDocument(doc, { preview: false })
+    registry.setTextEditor(uri, textEditor)
+
+    // Initial highlighting of the document and then for any addLogEvent calls.
+    highlightDocument(registry, uri)
+    vscode.workspace.onDidChangeTextDocument((event: vscode.TextDocumentChangeEvent) => {
+        if (event.document.uri.toString() === doc.uri.toString()) {
```
justinmk3, line 59: Is this a stable way to compare vscode.Uris? do we need something similar to `pathUtils.areEqual()` ?

related: https://github.com/microsoft/vscode/issues/93368

PR #2768, opened 2022-07-19
cwl: Cancel log group search 

Hweinstock:
## Problem
If a search of log groups is taking too long or was accidental, the user has no way to cancel it. 
## Solution
Add a timeout that can be cancelled via user input. 

Note: This change involves using deprecated code from `waitTimeout` in `src/shared/utilities/timeoutUtils.ts`. 
Note: The `Timeout` object is currently set to auto cancel the search after 5 seconds. 

## License

### DIFF
--- a/src/cloudWatchLogs/commands/addLogEvents.ts
+++ b/src/cloudWatchLogs/commands/addLogEvents.ts
@@ -7,6 +7,7 @@ import * as vscode from 'vscode'
 import * as AsyncLock from 'async-lock'
 import { getLogger } from '../../shared/logger/logger'
 import { LogStreamRegistry } from '../registry/logStreamRegistry'
+import { CancellationError } from '../../shared/utilities/timeoutUtils'
 
 // TODO: Cut a PR to the async-lock package?...as of now, maxPending = 0 is theoretically ideal, but also falsy (which sets maxPending = 1000):
 // https://github.com/rogierschouten/async-lock/blob/78cb0c2441650d7bdc148548f99542ccc9c93fd7/lib/index.js#L19
@@ -38,9 +39,14 @@ export async function addLogEvents(
             getLogger().debug('Update done, releasing lock...')
         })
     } catch (e) {
-        // contingency in case lock isn't busy but still locked out. Don't want to accidentally trigger making codelens not busy
-        getLogger().debug(`addLogEvents already locked for lock: ${lockName}`)
-        return
+        if (CancellationError.isUserCancelled(e)) {
+            getLogger().debug('cwl: User Cancelled Search')
+        } else {
+            // TODO: We assume here that non-cancellation errors are always busty errors. This may not be true.
+            // contingency in case lock isn't busy but still locked out. Don't want to accidentally trigger making codelens not busy
+            getLogger().debug(`addLogEvents already locked for lock: ${lockName}`)
+            return
+        }
     } finally {
         registry.setBusyStatus(uri, false)
         if (onDidChangeCodeLensEvent) {
--- a/src/cloudWatchLogs/commands/addLogEvents.ts
+++ b/src/cloudWatchLogs/commands/addLogEvents.ts
@@ -8,6 +8,7 @@ import * as AsyncLock from 'async-lock'
 import { getLogger } from '../../shared/logger/logger'
 import { LogStreamRegistry } from '../registry/logStreamRegistry'
 import { CancellationError } from '../../shared/utilities/timeoutUtils'
+import { localize } from 'vscode-nls'
 
 // TODO: Cut a PR to the async-lock package?...as of now, maxPending = 0 is theoretically ideal, but also falsy (which sets maxPending = 1000):
 // https://github.com/rogierschouten/async-lock/blob/78cb0c2441650d7bdc148548f99542ccc9c93fd7/lib/index.js#L19
@@ -42,10 +43,17 @@ export async function addLogEvents(
         if (CancellationError.isUserCancelled(e)) {
             getLogger().debug('cwl: User Cancelled Search')
         } else {
-            // TODO: We assume here that non-cancellation errors are always busty errors. This may not be true.
+            // What about busy errors here?
             // contingency in case lock isn't busy but still locked out. Don't want to accidentally trigger making codelens not busy
-            getLogger().debug(`addLogEvents already locked for lock: ${lockName}`)
-            return
+            const error = e as Error
+            vscode.window.showErrorMessage(
+                localize(
+                    'AWS.cwl.searchLogGroup.errorRetrievingLogs2',
+                    'Error retrieving logs for {0}: {1}',
+                    uri.path,
+                    error.message
+                )
+            )
         }
     } finally {
         registry.setBusyStatus(uri, false)
--- a/src/cloudWatchLogs/commands/addLogEvents.ts
+++ b/src/cloudWatchLogs/commands/addLogEvents.ts
@@ -43,7 +43,6 @@ export async function addLogEvents(
         if (CancellationError.isUserCancelled(e)) {
             getLogger().debug('cwl: User Cancelled Search')
         } else {
-            // What about busy errors here?
             // contingency in case lock isn't busy but still locked out. Don't want to accidentally trigger making codelens not busy
             const error = e as Error
             vscode.window.showErrorMessage(
--- a/src/cloudWatchLogs/commands/addLogEvents.ts
+++ b/src/cloudWatchLogs/commands/addLogEvents.ts
@@ -41,10 +41,10 @@ export async function addLogEvents(
         // contingency in case lock isn't busy but still locked out. Don't want to accidentally trigger making codelens not busy
         getLogger().debug(`addLogEvents already locked for lock: ${lockName}`)
         return
-    }
-
-    registry.setBusyStatus(uri, false)
-    if (onDidChangeCodeLensEvent) {
-        onDidChangeCodeLensEvent.fire()
+    } finally {
+        registry.setBusyStatus(uri, false)
+        if (onDidChangeCodeLensEvent) {
+            onDidChangeCodeLensEvent.fire()
+        }
     }
 }
--- a/src/cloudWatchLogs/commands/searchLogGroup.ts
+++ b/src/cloudWatchLogs/commands/searchLogGroup.ts
@@ -65,7 +65,7 @@ export async function searchLogGroup(registry: LogStreamRegistry): Promise<void>
         } catch (err) {
             if (CancellationError.isUserCancelled(err)) {
                 getLogger().debug('cwl: User Cancelled Search')
-                result = 'Cancelled'
+                result = 'Failed'
             } else {
                 const error = err as Error
                 vscode.window.showErrorMessage(
--- a/src/cloudWatchLogs/commands/searchLogGroup.ts
+++ b/src/cloudWatchLogs/commands/searchLogGroup.ts
@@ -46,9 +46,13 @@ export async function searchLogGroup(registry: LogStreamRegistry): Promise<void>
         }
 
         await registry.registerLog(uri, initialStreamData)
-        const doc = await vscode.workspace.openTextDocument(uri) // calls back into the provider
-        vscode.languages.setTextDocumentLanguage(doc, 'log')
-        await vscode.window.showTextDocument(doc, { preview: false })
+
+        if (registry.hasLog(uri)) {
+            // If the log recieved data i.e. wasn't cancelled.
+            const doc = await vscode.workspace.openTextDocument(uri) // calls back into the provider
+            vscode.languages.setTextDocumentLanguage(doc, 'log')
+            await vscode.window.showTextDocument(doc, { preview: false })
+        }
     } else {
         result = 'Cancelled'
     }
--- a/src/cloudWatchLogs/commands/searchLogGroup.ts
+++ b/src/cloudWatchLogs/commands/searchLogGroup.ts
@@ -19,6 +19,9 @@ import { createURIFromArgs } from '../cloudWatchLogsUtils'
 import { DefaultCloudWatchLogsClient } from '../../shared/clients/cloudWatchLogsClient'
 import { CloudWatchLogs } from 'aws-sdk'
 import { highlightDocument } from '../document/logStreamDocumentProvider'
+import { CancellationError } from '../../shared/utilities/timeoutUtils'
+import { localize } from 'vscode-nls'
+import { getLogger } from '../../shared/logger'
 
 export async function searchLogGroup(registry: LogStreamRegistry): Promise<void> {
     let result: telemetry.Result = 'Succeeded'
@@ -47,10 +50,8 @@ export async function searchLogGroup(registry: LogStreamRegistry): Promise<void>
         }
         // Currently displays nothing if update log fails in non-cancellationError. (don't want this)
 
-        // Catch error out here
-        await registry.registerLog(uri, initialStreamData)
-        const isLogCancelled = registry.getLogCancelled(uri)
-        if (!isLogCancelled) {
+        try {
+            await registry.registerLog(uri, initialStreamData)
             const doc = await vscode.workspace.openTextDocument(uri) // calls back into the provider
             vscode.languages.setTextDocumentLanguage(doc, 'log')
             const textEditor = await vscode.window.showTextDocument(doc, { preview: false })
@@ -61,8 +62,21 @@ export async function searchLogGroup(registry: LogStreamRegistry): Promise<void>
                     highlightDocument(registry, uri)
                 }
             })
-        } else {
-            result = 'Cancelled'
+        } catch (err) {
+            if (CancellationError.isUserCancelled(err)) {
+                getLogger().debug('cwl: User Cancelled Search')
+                result = 'Cancelled'
+            } else {
+                const error = err as Error
+                vscode.window.showErrorMessage(
+                    localize(
+                        'AWS.cwl.searchLogGroup.errorRetrievingLogs',
+                        'Error retrieving logs for Log Group {0} : {1}',
+                        logGroupInfo.groupName,
+                        error.message
+                    )
+                )
+            }
         }
     } else {
         result = 'Cancelled'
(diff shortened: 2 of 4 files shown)


src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -190,23 +195,32 @@ export async function filterLogEventsFromUriComponents(
     logGroupInfo: CloudWatchLogsGroupInfo,
     parameters: CloudWatchLogsParameters,
     nextToken?: string
-): Promise<CloudWatchLogsResponse> {
+): Promise<CloudWatchLogsResponse | void> {
     const client = new DefaultCloudWatchLogsClient(logGroupInfo.regionName)
 
-    const response = await client.filterLogEvents({
+    const timeout = new Timeout(5000) // How should I set this? This waits up to 5 seconds now
```
justinmk3, line 201: Should be as long as a user might conceivably want to wait, perhaps minutes, or infinite.

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -190,23 +195,32 @@ export async function filterLogEventsFromUriComponents(
...
+        // However, this fail wouldn't really impact customers.
+        if (response) {
+            return {
+                events: response.events ? response.events : [],
+                nextForwardToken: response.nextToken,
+                nextBackwardToken: nextToken,
+            }
+        }
+    } catch (err) {
+        return
```
justinmk3, line 223: can check `CancellationError.isUserCancelled(err)` , [example](https://github.com/aws/aws-toolkit-vscode/blob/5304b7329a3289b81d2ae6f87b6a9c000f84a504/src/credentials/loginManager.ts#L92)

```
            if (!CancellationError.isUserCancelled(err)) {
                throw err
            } else {
                getLogger().info('cloudwatchlogs: user cancelled search')
            }

**Review by justinmk3 — APPROVED** [review]
Looks like the right direction!

File: src/cloudWatchLogs/registry/logStreamRegistry.ts (hunk 1 above)
Comment (Hweinstock) [reply src/cloudWatchLogs/registry/logStreamRegistry.ts]: So I can simply set this in the code? Should it ideally be user-configurable?

File: src/cloudWatchLogs/registry/logStreamRegistry.ts [hunk 3]
```diff
@@ -190,23 +195,37 @@ export async function filterLogEventsFromUriComponents(
...
-    // Use heuristic of last token as backward token and next token as forward to generalize token form.
-    // Note that this may become inconsistent if the contents of the calls are changing as they are being made.
-    // However, this fail wouldn't really impact customers.
-    return {
-        events: response.events ? response.events : [],
-        nextForwardToken: response.nextToken,
-        nextBackwardToken: nextToken,
+    let response
+    try {
+        response = await waitTimeout(responsePromise, timeout)
```
JadenSimon, line 211: This is not exactly the "right" approach but considering we kept it simple and isolated it's fine.
Cancellation is a pretty advanced topic and it's easy to make your logic a lot buggier/messier than it should be.

We should do this as well:
```ts
const response = await waitTimeout(responsePromise, timeout, { allowUndefined: false })
```

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -190,23 +195,37 @@ export async function filterLogEventsFromUriComponents(
...
+                nextBackwardToken: nextToken,
+            }
+        }
+    } catch (err) {
+        if (!CancellationError.isUserCancelled(err)) {
+            getLogger().error(`cloudwatchlogs: failed to recieve log data for ${logGroupInfo}`)
+            throw err
+        } else {
+            getLogger().info('cloudwatchlogs: user cancelled search')
+        }
```
JadenSimon, line 228: Don't catch anything. We've made the signature of `filterLogEventsFromUriComponents` more complicated by doing this. Catch cancellation errors at your entry-point instead, or whenever you need to clean-up state.

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -114,7 +115,11 @@ export class LogStreamRegistry {
         try {
             // TODO: Consider getPaginatedAwsCallIter? Would need a way to differentiate between head/tail...
             const logEvents = await stream.retrieveLogsFunction(stream.logGroupInfo, stream.parameters, nextToken)
-
+            if (!logEvents) {
+                // This means the user cancelled the search
+                this.deregisterLog(uri)
```
JadenSimon, line 120: This doesn't seem right. We dump all the state if the user cancels a search?

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -46,9 +46,13 @@ export async function searchLogGroup(registry: LogStreamRegistry): Promise<void>
         }
 
         await registry.registerLog(uri, initialStreamData)
-        const doc = await vscode.workspace.openTextDocument(uri) // calls back into the provider
-        vscode.languages.setTextDocumentLanguage(doc, 'log')
-        await vscode.window.showTextDocument(doc, { preview: false })
+
+        if (registry.hasLog(uri)) {
+            // If the log recieved data i.e. wasn't cancelled.
```
JadenSimon, line 51: Notice how much more complicated our code got by adding very basic cancellation logic. And we haven't even gotten to testing it yet.

These comments are sign of a bigger problem. Our command can now silently fail without the user having any clue what just happened because we've added a lot of assumptions. Why even let this code execute if `registerLog` was cancelled?

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -46,9 +46,13 @@ export async function searchLogGroup(registry: LogStreamRegistry): Promise<void>
...
+            const doc = await vscode.workspace.openTextDocument(uri) // calls back into the provider
+            vscode.languages.setTextDocumentLanguage(doc, 'log')
+            await vscode.window.showTextDocument(doc, { preview: false })
+        }
     } else {
         result = 'Cancelled'
```
JadenSimon, line 57: Our telemetry will now erroneously report `Succeeded` instead of `Cancelled` if the user cancels the search.

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -258,7 +277,7 @@ export type CloudWatchLogsAction = (
     logGroupInfo: CloudWatchLogsGroupInfo,
     apiParameters: CloudWatchLogsParameters,
     nextToken?: string
-) => Promise<CloudWatchLogsResponse>
+) => Promise<CloudWatchLogsResponse | void>
```
JadenSimon, line 280: This seems like a small change, but we basically just doubled the complexity by doing this. The lower level your logic is, the more costly this becomes.

src/cloudWatchLogs/registry/logStreamRegistry.ts
Hweinstock: What would be the better way? Is it a completely different approach or some small changes? What are the disadvantages of this approach?

src/cloudWatchLogs/registry/logStreamRegistry.ts
Hweinstock: If I understand correctly, we want to only catch `CancellationError` here and not any other type of error? How do I implement that? I tried `catch(err: CancellationError)` but that doesn't seem to be correct. Also, what exactly does the signature of `filterLogEventsFromUriComponents` refer to?

src/cloudWatchLogs/commands/searchLogGroup.ts
Hweinstock: I see what you mean. I am assuming that the log will be `deRegistered` only if the user cancels. The only way around this I could think of was the give the response a `cancelled` attribute that we can directly check before running this code. That way we make no assumptions. However, my implementation feels a little clunky, so not sure if this is best approach or if I just implemented it wrong.

Hweinstock:
> A catch block catches all errors. You can selectively handle errors by re-throwing the ones you can't handle.
The signature is the function argument types + return type and is effectively an interface.

So is there an approach without a catch block or are you suggesting something similar to what there was originally where it caught everything then threw errors that weren't `cancellationError`?

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -217,16 +220,18 @@ export async function filterLogEventsFromUriComponents(
...
     } catch (err) {
-        if (!CancellationError.isUserCancelled(err)) {
-            getLogger().error(`cloudwatchlogs: failed to recieve log data for ${logGroupInfo}`)
-            throw err
-        } else {
+        if (CancellationError.isUserCancelled(err)) {
             getLogger().info('cloudwatchlogs: user cancelled search')
         }
     }
+    return {
```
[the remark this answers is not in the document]
Hweinstock: So should we still run through all the logic that can be run when an error occurs? What is the alternative to returning a `CloudWatchLogsResponse` with no events? I could also return `void` but that seemed to complicate things.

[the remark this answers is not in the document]
JadenSimon: > So should we still run through all the logic that can be run when an error occurs?

Should we? We didn't before, so we need to justify why we should do it now. It may help to start from the top of your program, maybe from `searchLogGroup`, and work your way down into where the AWS API call is made. Then assume that every single line can throw. Which they could. How does your program behave in these edge cases?

The entire purpose of an error is to short-circuit conventional control flow during _exceptional_ scenarios. Whether or not something is exceptional is a matter of perspective. From the user's perspective, something like a cancellation is obviously not an error. But we should not automatically assume the same is true for each component of our implementation.

JadenSimon:
> So is there an approach without a catch block or are you suggesting something similar to what there was originally where it caught everything then threw errors that weren't cancellationError?

There are many way to do this, and I'm not saying which approach I would use because I want you to understand what's happening on a fundamental level. These concepts apply to all software.

Hweinstock:
I see, so what would be another approach where I could avoid complicating the signature of `filterLogEventsFromUriComponents`? I think now I've implemented it this way, I am having a hard way of imagining how to avoid this.

JadenSimon:
> I see, so what would be another approach where I could avoid complicating the signature of filterLogEventsFromUriComponents? I think now I've implemented it this way, I am having a hard way of imagining how to avoid this.

Start back from the beginning: what logic did you add to implement a simple 'cancel'? If we ignore the UI, all you did was use `waitTimeout`. I'm assuming you added the `catch` block because otherwise the user would see an error message. But where does _that_ error message come from?

Hweinstock:
It would come from `waitTimeout` itself right? Is there a way I could avoid using `waitTimeout` to not have to deal with `cancellationError`?

JadenSimon:
> It would come from waitTimeout itself right?

Yes that is where the error is thrown, but that is not where the UI is shown to the user. 

> Is there a way I could avoid using waitTimeout to not have to deal with cancellationError?

Sure but how does that help you? We've already seen what happens when you remove `CancellationError`: you must return something else to represent cancellation. Fundamentally, we really only have two choices:
1. Use the return value to signal cancellation (conventional control flow)
2. Use errors to signal cancellation (exceptional control flow)

Neither are necessarily better than the other. But we need to understand what that second option entails in order to make a decision.

Hweinstock:
So right now, I have implemented option 2 with the exception. When you say to figure out what it entails, I assume that principally involves investigating the consequences of this choice and how to handle those consequences? It sounds like one of those consequences is that it complicates the signature of `filterLogEventsFromUriComponents`. Is this the right track of thinking?

src/cloudWatchLogs/commands/viewLogStream.ts
```diff
@@ -55,12 +55,24 @@ export async function viewLogStream(node: LogGroupNode, registry: LogStreamRegis
...
-        const textEditor = await vscode.window.showTextDocument(doc, { preview: false })
-        registry.setTextEditor(uri, textEditor) // For consistency, even though this isn't necessary (for now)
+        try {
+            await registry.registerLog(uri, initialStreamData)
+            const doc = await vscode.workspace.openTextDocument(uri) // calls back into the provider
+            vscode.languages.setTextDocumentLanguage(doc, 'log')
+            const textEditor = await vscode.window.showTextDocument(doc, { preview: false })
+            registry.setTextEditor(uri, textEditor) // For consistency, even though this isn't necessary (for now)
+        } catch (err) {
+            result = 'Cancelled'
```
JadenSimon, line 65:
```suggestion
            result = 'Failed'
```

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -212,20 +197,29 @@ export async function filterLogEventsFromUriComponents(
...
+    const timeout = new Timeout(10000)
+    showMessageWithCancel(`Loading log data from group ${logGroupInfo.groupName}`, timeout)
+    const responsePromise = client.filterLogEvents({
         logGroupName: logGroupInfo.groupName,
         filterPattern: parameters.filterPattern,
         nextToken,
         limit: parameters.limit,
     })
 
+    const response = await waitTimeout(responsePromise, timeout, { allowUndefined: false })
```
JadenSimon, line 206: You can throw if `response` is `void`
The types for `waitTimeout` are wrong (my bad):
```suggestion
    const response = await waitTimeout(responsePromise, timeout, { allowUndefined: false })
    if (!response) {
        throw new Error('`filterLogEvents` did not return anything')
    }
```

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -276,6 +267,7 @@ export type CloudWatchLogsResponse = {
     events: CloudWatchLogs.FilteredLogEvents
     nextForwardToken?: CloudWatchLogs.NextToken
     nextBackwardToken?: CloudWatchLogs.NextToken
+    cancelled?: boolean
```
JadenSimon, line 270: Not needed?
```suggestion
```

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -45,20 +48,36 @@ export async function searchLogGroup(registry: LogStreamRegistry): Promise<void>
...
+                if (event.document.uri.toString() === doc.uri.toString()) {
+                    highlightDocument(registry, uri)
+                }
+            })
+        } catch (err) {
+            if (CancellationError.isUserCancelled(err)) {
+                getLogger().debug('cwl: User Cancelled Search')
+                result = 'Cancelled'
+            } else {
+                const error = err as Error
```
JadenSimon, line 70:
```suggestion
                const error = err as Error
                result = 'Failed'
```

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -212,20 +194,29 @@ export async function filterLogEventsFromUriComponents(
 ): Promise<CloudWatchLogsResponse> {
     const client = new DefaultCloudWatchLogsClient(logGroupInfo.regionName)
 
-    const response = await client.filterLogEvents({
+    const timeout = new Timeout(10000)
```
JadenSimon, line 197: Let's just set it to something arbitrary like 5 minutes:
```suggestion
    const timeout = new Timeout(300000)
```

src/cloudWatchLogs/commands/addLogEvents.ts
```diff
@@ -38,13 +39,18 @@ export async function addLogEvents(
...
-    registry.setBusyStatus(uri, false)
-    if (onDidChangeCodeLensEvent) {
-        onDidChangeCodeLensEvent.fire()
+        if (CancellationError.isUserCancelled(e)) {
+            getLogger().debug('cwl: User Cancelled Search')
+        } else {
+            // TODO: We assume here that non-cancellation errors are always busty errors. This may not be true.
+            // contingency in case lock isn't busy but still locked out. Don't want to accidentally trigger making codelens not busy
+            getLogger().debug(`addLogEvents already locked for lock: ${lockName}`)
+            return
```
JadenSimon, line 48: We should show an error message to maintain the original behavior, something like this maybe:
```suggestion
            const error = e as Error
            vscode.window.showErrorMessage(
                localize(
                    'AWS.cwl.searchLogGroup.errorRetrievingLogs2',
                    'Error retrieving logs for {0}: {1}',
                    uri.path,
                    error.message
                )
            )
```

JadenSimon (approved):

src/cloudWatchLogs/commands/viewLogStream.ts
Hweinstock: I did the same change in `searchLogGroup.ts`.

src/cloudWatchLogs/registry/logStreamRegistry.ts
Hweinstock: Wait sorry, what about the types are wrong for `waitTimeout`?

src/cloudWatchLogs/commands/addLogEvents.ts
Hweinstock: I thought the original behavior was simply the early return so that it essentially ignored errors here. What about the busy/locked error? Do I want to display that to the user?

src/cloudWatchLogs/registry/logStreamRegistry.ts
JadenSimon: `waitTimeout` is incorrectly saying that it can return `void` when it never can.

src/cloudWatchLogs/commands/addLogEvents.ts
JadenSimon: > I thought the original behavior was simply the early return so that it essentially ignored errors here

`updateLog` used to catch everything, now it doesn't. That's why the original catch block is so odd.

> What about the busy/locked error?

The comment about this seems really vague. If this could even happen then something is wrong with the implementation.

JadenSimon (approved):

justinmk3:
@Hweinstock remember to squash-merge PRsmerged unless the PR commits were intentionally crafted. The 13 commits in this PR are not meaningful, the PR should be squash-merged.

PR #2771, opened 2022-07-20
Hkobew/time selection

Hweinstock:
## Problem
No way to filter the time range of the logs events. 
## Solution
Implement a similar submenu to allow for recent defaults, or custom range via an `inputBox`. 

- I reused much of the code from the `regionSubmenu` to build this submenu suggesting there could be a way to generalize both. 
- I am still looking into adding more tests, however it is unclear how to best test these submenus. Looking for advice/help here. Currently, testing it piece by piece rather than the submenu as a whole. 

## License


JadenSimon:
> I reused much of the code from the regionSubmenu to build this submenu suggesting there could be a way to generalize both.

Yup there's definitely a way. I wrote it only for regions to keep it simple. You can give it a try if you want a challenge, though I will say that there is a another solution for your requirements. The region submenu was written in a particular way because we wanted to re-render the items based off the region. If we didn't need that, we could do something like this:

```ts
const nestedPromptItem = {
    label: 'Select me!',
    data: async () => {
        const resp = await showInputBox({ title: 'Enter stuff' })
        return resp === undefined ? WIZARD_RETRY : resp
    },
}

const prompter = createQuickPick([nestedPromptItem])
```

This would only work inside of a wizard because `WIZARD_RETRY` tells the state machine to re-run the same step. Which from the user's perspective it looks like they went back a step. It's not a great solution but it is one of them.

Hweinstock:
I like that this way requires way less code. However, I do think it is nice to have an object, `TimeFilterSubmenu` to organize all of the methods for this `quickPick`. It seems like in this case it is worth the sacrifice?

Hweinstock:
Also, how would I bind this prompter to the wizard since the return type of this quickPick would not necessarily follow the format of `TimeFilterResponse`? I just pushed what I currently have, but theres an error.

JadenSimon:
> It seems like in this case it is worth the sacrifice?

Up to you, though I will say that classes (and modules) are far more useful as tools for encapsulation rather than organization.  UI changes frequently and so it's nice to have implementations that are easy to test/change/delete as the requirements change. 

> Also, how would I bind this prompter to the wizard since the return type of this quickPick would not necessarily follow the format of TimeFilterResponse? I just pushed what I currently have, but theres an error.

What's the error? And why wouldn't it always follow `TimeFilterResponse`? If there's no 'type' error then there's a logic bug.

Hweinstock:
I think I had something wrong, but after making all the changes, I think the object approach leads to more straight forward testing. I was going to adjust the testing, but I think the flow of when we have an object to work with is much nicer/easier.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
```diff
@@ -64,7 +62,7 @@ export function parseCloudWatchLogsUri(uri: vscode.Uri): {
  */
 export function canShowPreviousLogs(uri: vscode.Uri): boolean {
     const params = parseCloudWatchLogsUri(uri).parameters
-    return params.filterPattern ? false : true
+    return params.filterPattern === undefined ? true : false
```
jpinkney-aws, line 65: Can't we just do something like:
```suggestion
    return params.filterPattern === undefined
```

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -12,29 +12,47 @@ import {
...
         }
+        let parameters: CloudWatchLogsParameters
 
-        const parameters: CloudWatchLogsParameters = {
-            limit: registry.configuration.get('limit', 10000),
-            filterPattern: response.filterPattern,
+        if (response.timeRange.start === response.timeRange.end) {
+            // this means no time filter.
+            parameters = {
+                limit: registry.configuration.get('limit', 10000),
```
jpinkney-aws, line 46: Does it make sense to extract:
```javascript
registry.configuration.get('limit', 10000)
```
outside the if statement since it's being used in both (just to make it easier if the 10000 limit changes and its always going to be the same in both)

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -73,35 +117,184 @@ async function logGroupsToArray(logGroups: AsyncIterableIterator<CloudWatchLogs.
...
+            return 'enter date in format YYYY/MM/DD-YYYY/MM/DD'
+        }
+        if (startTime === endTime) {
+            return 'must enter two different dates for valid range'
+        }
+        if (Date.parse(startTime) > Date.parse(endTime)) {
+            return 'first date must occur before second date'
+        }
+
+        if (Date.parse(endTime) > today.valueOf()) {
```
jpinkney-aws, line 279: Should we also validate if startTime  is > today.valueOf() as well? E.g. if a user enters 2025/01/01 as the start time and 2026/01/01 as the end time

src/codewhisperer/vue/root.vue
```diff
@@ -107,9 +107,7 @@
                                 you provide to the Beta Service, and that multiple users may receive the same or similar
                                 output. For clarity, although these computational results are Your Content, the same or
                                 similar outputs are independently created, and AWS will not be restricted from providing
-                                the same or similar output to other customers. In addition, the CodeWhisperer Beta
-                                Service may produce computational results that are attributed to AWS or our affiliates,
-                                and we retain the underlying intellectual property rights in that Content;
+                                the same or similar output to other customers.
```
jpinkney-aws, line 110: I think some other unrelated changes might have got pulled into this PR (a few codewhisperer files)

JadenSimon: I think it's just a Github bug because the relevant commit https://github.com/aws/aws-toolkit-vscode/commit/6f979456eccf8c0be81616f80ffd602bcfb6293b is in the target branch. Maybe if there's a merge conflict it shows a different diff?

It should go away once the merge conflict is resolved

src/cloudWatchLogs/cloudWatchLogsUtils.ts
Hweinstock: Yep, I think I overthought that one haha.

src/cloudWatchLogs/commands/searchLogGroup.ts
Hweinstock: I think that case is covered since I also check that the end time occurs after the start date. If `startTime > today.valueOf()` but it passes the previous check so we know that  `startTime < endTime` then it must also be true that `endTime > today.valueOf()` so it would fail this test. Let me know if I am missing something, but I believe it's covered.

src/test/cloudWatchLogs/commands/searchLogGroup.test.ts
```diff
@@ -5,44 +5,105 @@
...
+    >
+    let filterPatternPrompter: InputBoxPrompter
+
     let testWizard: WizardTester<SearchLogGroupWizard>
+    let testTimeRangeMenu: TimeFilterSubmenu
+
+    let customTimeRangePrompter: InputBoxPrompter
+    let defaultTimeRangePrompter: QuickPickTester<any>
+
+    let testDateBox: ExposeEmitters<vscode.InputBox, 'onDidAccept' | 'onDidChangeValue' | 'onDidTriggerButton'>
```
JadenSimon, line 32: We're placing a lot of variables here even though not every test uses them. You can nest `describe` blocks to avoid this:

```ts
describe('searchLogGroup', function () {
// ...
    describe('timerange submenu', function () {
        let testTimeRangeMenu: TimeFilterSubmenu
        // ...
    })
```

It's a small thing but test organization is important. Even if it does come at the cost of a little bit more boilerplate.

Oh also I'm not sure if I pointed this out but your test cases should read like `it [behavior goes here]`
Example:
```ts
    describe('Timerange Submenu', function () {
        let testTimeRangeMenu: TimeFilterSubmenu
         // test setup somewhere

         it('gives option for custom input', async function () {
             // ...
         })
    })
```
Which reads as 'Timerange Submenu gives option for custom input'

Hweinstock: I'll try to update some of my other PRs with this in mind. Thanks for pointing this out.

JadenSimon (approved):

jpinkney-aws (approved):

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -14,26 +14,44 @@ import {
...
-        limit: registry.configuration.get('limit', 10000),
-        filterPattern: response.filterPattern,
+    if (response.timeRange.start === response.timeRange.end) {
+        // this means no time filter.
+        parameters = {
+            limit: limitParam,
+            filterPattern: response.filterPattern,
+        }
+    } else {
+        parameters = {
```
justinmk3, line 49: could use a ternary, then `parameters` could be `const`. That also improves readability by avoiding nesting.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -162,5 +181,163 @@ export class SearchLogGroupWizard extends Wizard<SearchLogGroupWizardResponse> {
...
+            }
+        }
+    }
+
+    public validateDate(input: string) {
+        const parts = input.split('-')
+        const today = new Date()
+
+        if (parts.length !== 2) {
+            return 'String must include two dates seperated by `-`'
```
justinmk3, line 301: spelling

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -162,5 +181,163 @@ export class SearchLogGroupWizard extends Wizard<SearchLogGroupWizardResponse> {
...
+        }
+        const [startTime, endTime] = parts
+
+        if (!Date.parse(startTime)) {
+            return 'starting time format is invalid, use YYYY/MM/DD'
+        }
+        if (!Date.parse(endTime)) {
+            return 'ending time format is valid, use YYYY/MM/DD'
```
justinmk3, line 309: all of this logic could live in a shared common module, if we don't already have one.

src/test/cloudWatchLogs/commands/searchLogGroup.test.ts
```diff
@@ -5,44 +5,122 @@
...
+            })
+
+            it("validator won't accept invalid date formats", function () {
+                assert(testTimeRangeMenu.validateDate('2000/11/07,2000/12/08')) // No seperator
+                assert(testTimeRangeMenu.validateDate('2002/13/06-2001/11/05')) // Invalid start date
+                assert(testTimeRangeMenu.validateDate('2002/09/06-2001/00/05')) // Invalid end date
+                assert(testTimeRangeMenu.validateDate('10/02/2001-08/05/2001')) // Invalid format of dates
+                assert(testTimeRangeMenu.validateDate('2000/10/01-2000/10/01')) // Same date
+                assert(testTimeRangeMenu.validateDate('2000/10/02-2000/10/01')) // second date occuring earlier
+                assert(testTimeRangeMenu.validateDate('2000/12/01-2099/10/03')) // second date can't be in the future
```
justinmk3, line 105: why not assert the returned string?

src/test/cloudWatchLogs/commands/searchLogGroup.test.ts
```diff
@@ -5,44 +5,122 @@
...
+                customTimeRangePrompter = testTimeRangeMenu.customPrompter
+
+                const validInput = '2000/10/06-2001/11/08'
+                const result = customTimeRangePrompter.prompt()
+                accept(validInput)
+                assert.strictEqual(await result, validInput)
+            })
+
+            it("validator won't accept invalid date formats", function () {
+                assert(testTimeRangeMenu.validateDate('2000/11/07,2000/12/08')) // No seperator
```
justinmk3, line 99: what about undefined, empty string, other random input?

PR #2776, opened 2022-07-26
Hkobew/escape hatches

Hweinstock:
## Problem
There is currently no way to modify the search parameters without performing a completely new search. 
## Solution
(WIP) We can add two buttons on the navigation bar. One to change the `filterPattern` of the current `searchLogGroup` and another to change the `timeFilterRange`. 

In order to implement this, we create two new commands `aws.cwl.changeTimeFilter` and `aws.cwl.changeFilterPattern`. 

The buttons feature depends on the context. If the currently open document is a log stream, it will filter the log stream. However, if it is a log group search it will modify the parameters of the log group search. 

Example view (the search icon and filter icon on navigation bar):
image

## License


src/test/cloudWatchLogs/changeLogSearch.test.ts
```diff
@@ -0,0 +1,84 @@
...
+                message: 'Here is some original text that we want to overwrite.',
+            },
+        ],
+        parameters: oldComponenents.parameters,
+        logGroupInfo: oldComponenents.logGroupInfo,
+        retrieveLogsFunction: fakeGetLogEvents,
+        busy: false,
+    }
+
+    const newData: CloudWatchLogsData = {
```
JadenSimon, line 58: Can this test data be re-used across tests?
Reducing code duplication in tests is desirable too, just not to the same degree as other logic.

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    return newData
+}
+
+export async function changeLogSearchParams(registry: LogStreamRegistry, param: ChangeableParam): Promise<void> {
+    let result: telemetry.Result = 'Succeeded'
+
+    const oldUri = getActiveUri(registry)
+    const newData = await getNewData(registry, param, oldUri)
```
JadenSimon, line 67: Every single `await` is a place where race conditions can occur. JS is a single-threaded language, but multiple 'tasks' can still be executing concurrently.

I don't think we need to worry about this here though. Looks like every async call is pretty clean.

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
```diff
@@ -54,3 +54,17 @@ export async function highlightDocument(registry: LogStreamRegistry, uri: vscode
         textEditor.setDecorations(HIGHLIGHTER, ranges)
     }
 }
+
+export function getActiveUri(registry: LogStreamRegistry) {
```
JadenSimon, line 58: I think a more verbose name works here because it seems like a pretty important behavior:
```suggestion
export function getActiveDocumentUri(registry: LogStreamRegistry) {
```

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -0,0 +1,79 @@
...
+import { prepareDocument } from './commands/searchLogGroup'
+import { getActiveUri } from './document/logStreamDocumentProvider'
+import { CloudWatchLogsData, filterLogEventsFromUriComponents, LogStreamRegistry } from './registry/logStreamRegistry'
+import { isViewAllEvents, TimeFilterResponse, TimeFilterSubmenu } from './timeFilterSubmenu'
+
+type ChangeableParam = 'filterPattern' | 'timeFilter'
+
+export async function getNewData(
+    registry: LogStreamRegistry,
+    param: ChangeableParam,
```
JadenSimon, line 18: very small nit:
```suggestion
    param: 'filterPattern' | 'timeFilter',
```

Not everything needs a name

JadenSimon (approved):
Should be good to merge after addressing/acknowledging the feedback.

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -0,0 +1,79 @@
...
+    oldUri: vscode.Uri
+): Promise<CloudWatchLogsData | undefined> {
+    const oldData = registry.getLogData(oldUri) as CloudWatchLogsData
```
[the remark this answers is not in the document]
Hweinstock: Is there every a time where as is appropriate? Other than maybe in some testing scenarios where I've seen it used.

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -0,0 +1,79 @@
...
+    let newTimeRange: TimeFilterResponse | undefined
+
+    switch (param) {
+        case 'filterPattern':
+            newPattern = await showInputBox({
+                title: isLogStream(oldUri) ? 'Filter Log Stream Results' : 'Log Group Keyword Search',
+                placeholder: oldData.parameters.filterPattern ?? 'Enter Text Here',
+            })
+            if (newPattern === undefined) {
+                return
```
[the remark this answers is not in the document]
Hweinstock: Are there any general 'rules' of when to provide a docstring or is it more a judgement call on whether or not the function name provides all the context necessary?

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -271,6 +287,7 @@ export type CloudWatchLogsParameters = {
     endTime?: number
     limit?: number
     streamName?: string
+    streamNameOptions?: string[]
```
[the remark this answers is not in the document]
Hweinstock: In almost every way they are the same thing. The `streamNameOptions` is for limiting a log group to certain log streams whereas the `streamName` is for describing what stream is being viewed. I decided to keep them separate because this would allow for the functionality of limiting a log group search to include contents from multiple different streams (which the API supports). It also allows me reserve `streamName` to imply that we are viewing a single stream in `isLogStream`.

src/cloudWatchLogs/changeLogSearch.ts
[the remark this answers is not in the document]
JadenSimon: Can't say there's really any rules, it's mostly a judgement call. You can add docstrings whenever you want but they should say something that is not obvious just from looking at the code. Sometimes it's better to rename the function instead. For example, if we called the function `promptForNewData` then it's a lot more obvious that this is asking the user for something, not just 'getting' the data from wherever. Maybe a good 'rule' is that if the function causes "side-effects" (e.g. changes the UI) then more descriptive names are preferred.

Keep in mind that the more 'visible' something is, the more important the name. Things like private methods/variables/etc. don't matter as much because they won't be seen as much. Terse names are a good thing as long as they won't cause confusion.

src/test/cloudWatchLogs/changeLogSearch.test.ts
Hweinstock: I tried moving most of the functions and data that were used to `utils.test.ts`. Let me know if there is a better place or more that I could move.

src/cloudWatchLogs/changeLogSearch.ts
Hweinstock: When should I worry about this? If the `await` calls begin to depend on each other more?

src/cloudWatchLogs/changeLogSearch.ts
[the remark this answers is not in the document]
justinmk3: +1, but also, if I find myself thinking about quirks or special knowledge, I add that to the docstring.

src/cloudWatchLogs/changeLogSearch.ts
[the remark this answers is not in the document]
JadenSimon: > Is there ever a time where as is appropriate

Yeah there's plenty of valid uses. It's a shortcut for when a high-level of precision is not needed. But it's still a shortcut. And lots of shortcuts cause problems.

Using `as` is fine as long as we _know_ that the type assertion is valid in whatever context we're working in. By that I mean the assertion will not cause errors later on. Still best to avoid it though because, by design, it causes a "break" in the type system. Type assertions are a powerful tool if used wisely.

src/cloudWatchLogs/changeLogSearch.ts
JadenSimon: > If the await calls begin to depend on each other more?

Yes. To recognize it, we have to think about all those assumptions we're making that might be violated in a concurrent scenario. Changing state is generally something you want to look out for.

justinmk3: Always worry :) but in practice it depends on what the stakes are.

Similar to any filesystem or network call: the data could change just before or just after you read it. In a callback/promise-based system, there is an additional risk of the call getting deferred.

There is a "lock" concept for when you want some sort of atomicity: https://github.com/aws/aws-toolkit-vscode/blob/420cd0158f8a1b9b5e58c531aae1e4fe6e9f0baa/src/cloudWatchLogs/commands/addLogEvents.ts#L32

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
```diff
@@ -54,3 +54,17 @@ export async function highlightDocument(registry: LogStreamRegistry, uri: vscode
...
+
+export function getActiveDocumentUri(registry: LogStreamRegistry) {
+    const currentEditor = vscode.window.activeTextEditor
+    if (!currentEditor) {
+        throw new Error('cwl: Failed to identify active editor.')
+    }
+
+    const activeUri = currentEditor.document.uri
+    if (!registry.hasLog(activeUri)) {
+        throw new Error('cwl: Document open has unregistered uri.')
```
justinmk3, line 66: a getter is usually not the place for this kind of error. `getActiveDocumentUri` looks pretty specialized, maybe it should be a private function in the module that needs it.

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -206,6 +218,10 @@ export async function filterLogEventsFromUriComponents(
         cwlParameters.endTime = parameters.endTime
     }
 
+    if (parameters.streamNameOptions) {
+        cwlParameters.logStreamNames = parameters.streamNameOptions
```
justinmk3, line 222: hmm..

src/cloudWatchLogs/timeFilterSubmenu.ts
```diff
@@ -0,0 +1,171 @@
...
+    public customPrompter: InputBoxPrompter = this.createDateBox()
+
+    public constructor() {
+        super()
+    }
+
+    private get recentTimeOptions(): ItemLoadTypes<number> {
+        const options: DataQuickPickItem<number>[] = []
+        options.push({
+            label: 'View all events',
```
justinmk3, line 38: The name of this item defeats [parallel form](http://guidetogrammar.org/grammar/parallelism.htm), which makes the list less intuitive to scan for the user. We could preserve parallel form by naming this "Forever" or something like that, then it matches the form of the other relative-time items.

src/cloudWatchLogs/timeFilterSubmenu.ts
```diff
@@ -0,0 +1,171 @@
...
+        return createInputBox({
+            title: 'Enter custom date range',
+            placeholder: 'YYYY/MM/DD-YYYY/MM/DD',
+            validateInput: input => this.validateDate(input),
+        })
+    }
+
+    protected async promptUser(): Promise<PromptResult<TimeFilterResponse>> {
+        while (true) {
+            switch (this.currentState) {
```
justinmk3, line 89: Looks like `currentState` has known expects fields, so the `default` case could throw an error to guard against future changes (and an infinite loop here).

Also I wonder if changing the `while(true)` to something like `for(int i = 0; i < 5; i++)` makes sense, to guard against infinite loop?

src/cloudWatchLogs/timeFilterSubmenu.ts
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }
+    }
+
+    public validateDate(input: string) {
+        const parts = input.split('-')
+        const today = new Date()
+
+        if (parts.length !== 2) {
+            return 'String must include two dates seperated by `-`'
```
justinmk3, line 129: spelling

PR #2777, opened 2022-07-27
ux(auth): improve 'Create Credentials Profile' flow

JadenSimon:
## Problem
Old flow is clunky. If the user provides bad credentials they need to start all over. 

## Solution
Use updated flow that goes back a step on failure.

This flow was pulled from a separate branch with more in-depth UX improvements which is why the implementation is split into a 'template' rather than a single wizard.

The validation step can be made nicer but that's for another time.

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
     "packages": {
         "": {
             "name": "aws-toolkit-vscode",
-            "version": "1.46.0-SNAPSHOT",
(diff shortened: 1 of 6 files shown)


justinmk3:
I think this is worth a changelog item, something like `"Foo" command validates credentials and lets you edit previous steps`

JadenSimon:
Yeah I couldn't decide if it was significant enough to mention. It's an improvement for sure though.

PR #2778, opened 2022-07-28
feat(auth): "Connect to AWS" does not handle missing credentials file

justinmk3:
## Problem:
In AWS Explorer, the "Connect to AWS" node does nothing if there is no
~/.aws/credentials file.

## Solution:
- If ~/.aws/credentials is not found, run the "Create Credentials" command.
- Skip the "Enter Profile Name" step, to reduce friction always use
  "default" if there are no existing profiles.
- Don't prompt to ask, just do it. User already indicated that they want
  to connect.

## License


src/credentials/wizards/createProfile.ts
```diff
@@ -43,7 +44,13 @@ export class CreateProfileWizard extends Wizard<CreateProfileState> {
     public constructor(profiles: { readonly [name: string]: Profile }, template: ProfileTemplateProvider) {
         super()
 
-        this.form.name.bindPrompter(() => createProfileNamePrompter(profiles))
+        if (Object.keys(profiles).length === 0) {
+            // Skip the "Enter Profile" step and use "default" profile name.
+            // This reduces friction for the common case.
+            this.form.name.bindPrompter(() => new PassthroughPrompter('default'))
```
justinmk3, line 50: @JadenSimon is there another way to force `name = 'default'`?

JadenSimon: Yeah there's a declarative and imperative way:

Declarative:
```ts
    super({ initState: { name: Object.keys(profiles).length === 0 ? 'default' : undefined } })
    this.form.name.bindPrompter(() => createProfileNamePrompter(profiles))
```

Imperative:
```ts
     if (Object.keys(profiles).length === 0) {
         this.form.name.setDefault('default')
     } else {
         this.form.name.bindPrompter(() => createProfileNamePrompter(profiles))
     }
```

Haven't really decided on what the 'best' way is. So far devs seem to prefer the imperative approach so probably that one for simple scenarios.

PR #2783, opened 2022-07-28
feat(ui): decouple tree nodes from the UI representation

JadenSimon:
## Problem
While prototyping features, I kept running into the situation where I wanted to update how a node was rendered independently of the node's children. The current implementation does not easily support this use-case, forcing us to propagate events up to a parent node to get the desired behavior.

## Solution
Fully generalize the `TreeNode` interface. It was already very close to the general solution, it just needed a few tweaks to get it there. Of course, implementing this correctly with a tree data provider required more substantial changes which tests were added for.

I wouldn't recommend that _all_ tree node implementations use this pattern because it can quickly spiral out of control from a maintainability point of view. Still, I'd rather have the proper solution readily available than a bunch of one-off bespoke solutions. 

## License

### DIFF
--- a/src/awsexplorer/localExplorer.ts
+++ b/src/awsexplorer/localExplorer.ts
@@ -5,10 +5,9 @@
 
 import * as vscode from 'vscode'
 import * as telemetry from '../shared/telemetry/telemetry'
-import { cdkNode } from '../cdk/explorer/rootNode'
+import { cdkNode, CdkRootNode } from '../cdk/explorer/rootNode'
 import { ResourceTreeDataProvider, TreeNode } from '../shared/treeview/resourceTreeDataProvider'
 import { once } from '../shared/utilities/functionUtils'
-import { AppNode } from '../cdk/explorer/nodes/appNode'
 import { isCloud9 } from '../shared/extensionUtilities'
 import { codewhispererNode } from '../codewhisperer/explorer/codewhispererNode'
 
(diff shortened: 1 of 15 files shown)


src/shared/treeview/utils.ts
```diff
@@ -96,16 +96,33 @@ export function unboxTreeNode<T>(node: TreeNode, predicate: (resource: unknown)
  * would be passed in as-is.
  */
 export class TreeShim extends AWSTreeNodeBase {
```
JadenSimon, line 98: The shim was updated to support this new functionality. It's not a perfect recreation because `TestNode` can handle async rendering of the node itself while the shim would show `Loading...` on initialization.

src/shared/treeview/utils.ts
```diff
@@ -77,7 +77,7 @@ export function createPlaceholderItem(message: string): TreeNode {
     return {
         id: 'placeholder',
         resource: message,
-        treeItem: new vscode.TreeItem(message, vscode.TreeItemCollapsibleState.None),
+        getTreeItem: () => new vscode.TreeItem(message, vscode.TreeItemCollapsibleState.None),
```
JadenSimon, line 80: I don't like turning otherwise constant things into functions but it was kind of necessary for this functionality.
It makes things harder to test and reason about :/

src/awsexplorer/localExplorer.ts
```diff
@@ -49,7 +48,7 @@ export function createLocalExplorerView(): vscode.TreeView<TreeNode> {
     // Legacy CDK metric, remove this when we add something generic
     const recordExpandCdkOnce = once(telemetry.recordCdkAppExpanded)
     view.onDidExpandElement(e => {
-        if (e.element.resource instanceof AppNode) {
+        if (e.element.resource instanceof CdkRootNode) {
```
JadenSimon, line 51: Metric was being recorded when the child node expanded rather than the parent (my bad).

justinmk3 (approved):
great!

PR #2786, opened 2022-07-29
Hkobew/jumpto

Hweinstock:
## Problem
Want a method to be able to track down 'source' of logs. Meaning, given an event in a log group search result, we want to `jumpTo` the contents of the stream that contains that log event. 
## Solution
We can solve this issue by implementing the `DefinitionProvider` from vscode on our logSearch files. This will allow us to have a jump to functionality. However, currently struggling with the fact that `provideDefinition` function is not asynchronous, but loading the log content is. In order to solve this, I am currently investigating a solution involving the `VirtualFileSystem` class. 

Some crucial steps are as follows: 
- remove stream ID from the document and store them in a map. 
- remove `\` from URI and reformat Uri to be of format `{region}/{logGroup}/{streamName}` for individual streams and `{region}/{logGroup}` for log group searches. (This step should allow us to only apply the `definitionProvider` to virtual files that are from log group searches and not individual streams. 
- In order to have `streamName` in the Uri, we can move it from the `CloudWatchLogsParameters` type to the `CloudWatchLogsGroupInfo`  type. 
- We then implement a definition provider that allows us to jump to the stream containing that event. 

## License


src/cloudWatchLogs/cloudWatchLogsFiles.ts
```diff
@@ -0,0 +1,15 @@
+import { Uri, Disposable } from 'vscode'
+import { FileProvider, VirualFileSystem } from '../shared/virtualFilesystem'
+import { LogStreamRegistry } from './registry/logStreamRegistry'
+
+export class CloudWatchLogsFileSystem extends VirualFileSystem {
```
JadenSimon, line 5: A VFS is _significantly_ more complex than document providers. Would avoid for now.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -178,4 +178,32 @@ export class SearchLogGroupWizard extends Wizard<SearchLogGroupWizardResponse> {
...
     }
+}
+
+export class JumpToStream implements vscode.DefinitionProvider {
+    private registry: LogStreamRegistry
+
+    public constructor(registry: LogStreamRegistry) {
+        this.registry = registry
+    }
+    provideDefinition(
```
JadenSimon, line 189: This can be `async`:
```suggestion
    async provideDefinition(
```

You'll need to change the return type: 
```ts
Promise<vscode.Definition | vscode.LocationLink[] | undefined>
```

Hweinstock: This is what I originally tried and it gives me a linting error. It still runs and actually gives me the exact behavior I want however. It says `Property 'provideDefinition' in type 'JumpToStream' is not assignable to the same property in base type 'DefinitionProvider'.`

JadenSimon: > It still runs and actually gives me the exact behavior I want however.

This is expected. **Typescript types do not affect runtime behavior**

> Why do I remove the providerResult layer when it becomes a promise?

Because `ProviderResult<T>` is equivalent to:
```ts
T | undefined | null | Thenable<T | undefined | null>
```

We don't need to care about `Thenable` or `null` so we throw it away. The VS Code API is just saying it can support all these different types, but we only need to return a subset of these types.

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -83,16 +85,20 @@ export class LogStreamRegistry {
...
             }
+            if (datum.logStreamName) {
+                newLines = (line.match(/\n/g) || []).length
+                let curLine = lineNumber
+                while (curLine < lineNumber + newLines) {
+                    this.setStreamIdMap(uri, curLine, datum.logStreamName)
+                    curLine += 1
+                }
+            }
+            lineNumber += newLines
```
Hweinstock, line 99: There must be a better way to do this than counting `\n`.

src/cloudWatchLogs/activation.ts
```diff
@@ -18,11 +18,18 @@ import { LogStreamRegistry } from './registry/logStreamRegistry'
...
 import { changeLogSearchParams } from './changeLogSearch'
+import { JumpToStream } from './commands/searchLogGroup'
 
 export async function activate(context: vscode.ExtensionContext, configuration: Settings): Promise<void> {
     const settings = new CloudWatchLogsSettings(configuration)
     const registry = new LogStreamRegistry(settings)
 
+    const definitionProvider = new JumpToStream(registry)
+    vscode.languages.registerDefinitionProvider(
+        { language: 'log', scheme: CLOUDWATCH_LOGS_SCHEME, pattern: `**` },
```
Hweinstock, line 29: Struggling to find what the correct pattern here is. I thought it would be something like `[!:]?[!:]**:[!:]?[!:]**`, but that doesn't seem to work. Looking to only match strings that contain a single `:` with one or more characters on either side.

JadenSimon: Why do we need this pattern?
You'd use this: `.*:.+`

Hweinstock: This makes it so the jumpto `definitionProvider` only runs on the log search files rather than the single stream files. Right now, it just throws an error when it can't find the streamID causing the `definitionProvider` to fail on log stream files. Uri for log search takes form '{region}:{group}` whereas for stream its '{region}:{group}:{stream}`.

Hweinstock: I'm not getting that this pattern matches either document. Is there a step I am missing beyond replacing the pattern field with this string?

JadenSimon: Oh I think `pattern` is a glob pattern. So you'd use `**/.*:.+`

But I think the better question is why there's no stream ID for a log search. Shouldn't both searches have it?

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -185,6 +199,31 @@ export class LogStreamRegistry {
...
+    private setRangeForStreamIdMap(uri: vscode.Uri, lowerBound: number, upperBound: number, streamID: string): void {
+        // Note: Inclusive of both bounds.
+        let currentLine = lowerBound
+        while (currentLine <= upperBound) {
+            this.setStreamIdMap(uri, currentLine, streamID)
+            currentLine += 1
+        }
+    }
+
+    public setStreamIdMap(uri: vscode.Uri, lineNum: number, streamID: string): void {
```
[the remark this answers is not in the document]
Hweinstock: Could I move the logic from `getLogContent` into here or call the function here? That would be iterating through all the log data twice however. Would the best option be the combine these functions into one? That is a single function that returns both the content and the streamID maps.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -179,7 +180,53 @@ export class SearchLogGroupWizard extends Wizard<SearchLogGroupWizardResponse> {
...
+    ): Promise<vscode.Definition | vscode.LocationLink[] | undefined> {
+        const activeUri = document.uri
+        const logGroupInfo = parseCloudWatchLogsUri(activeUri).logGroupInfo
+        const curLine = document.lineAt(position.line)
+        let streamUri: vscode.Uri
+        try {
+            const streamIDMap = this.registry.getStreamIdMap(activeUri)
+            const streamID = streamIDMap?.get(curLine.lineNumber)
+            if (!streamID) {
+                throw new Error(`cwl: No streamID found for stream with uri ${activeUri}`)
```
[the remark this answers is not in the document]
Hweinstock: Presumably this should only happen when looking at a single log stream, or if the streamIDMap hadn't been set before this function was called. Should I fail if this is a stream or if the streamIDMap is empty earlier rather than combining those two cases here?

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -179,7 +180,53 @@ export class SearchLogGroupWizard extends Wizard<SearchLogGroupWizardResponse> {
...
         this.form.submenuResponse.bindPrompter(createRegionSubmenu)
-        this.form.filterPattern.bindPrompter(({ submenuResponse }) => createFilterpatternPrompter(submenuResponse!.data))
+        this.form.filterPattern.bindPrompter(({ submenuResponse }) =>
+            createFilterpatternPrompter(submenuResponse!.data)
+        )
         this.form.timeRange.bindPrompter(() => new TimeFilterSubmenu())
     }
 }
+
+export class JumpToStream implements vscode.DefinitionProvider {
```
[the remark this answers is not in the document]
Hweinstock: I did not know this was possible, but I really like the idea. I agree it doesn't need its own class. I moved it over.

src/cloudWatchLogs/activation.ts
Hweinstock: There is no streamID because the log search is combining the results from all the streams in a log group so it's not limited to a single stream.

Hweinstock: Also, I am still having trouble with the `pattern`. I feel like I must have something wrong. I just committed what I currently have.

JadenSimon: > There is no streamID because the log search is combining the results from all the streams in a log group so it's not limited to a single stream.

What's the value of this line tracking logic + definition provider if it always jumps to the same stream?

src/cloudWatchLogs/registry/logStreamRegistry.ts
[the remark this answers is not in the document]
JadenSimon: > That is a single function that returns both the content and the streamID maps.

Yeah this generally the starting point: make your function return more information.
Don't worry about this for now though. We can come back to it if we have time.

src/cloudWatchLogs/commands/searchLogGroup.ts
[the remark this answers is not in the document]
JadenSimon: > Should I generally try to prioritize throwing very specific errors rather than trying to minimize the amount of code?

There's not really a good answer to this question as it is based off application requirements. Informative error messages is a very good thing for maintainability but it does increase development time. I wouldn't spend a bunch of time crafting error messages if you're not confident about what the future application requirements might be.

I think this specific assertion is good but I'm just pointing out that failing early tends to create code that is easier to reason about.

src/shared/ui/prompter.ts
```diff
@@ -79,3 +79,23 @@ export abstract class Prompter<T> {
...
+        return this.result
+    }
+
+    public setStepEstimator(estimator: StepEstimator<string>): void {}
+    public setSteps(current: number, total: number): void {}
+    public set recentItem(response: any) {}
+    public get recentItem(): any {
+        return
+    }
+}
```
JadenSimon, line 101: This logic no longer exists on mainline
```suggestion
```

src/cloudWatchLogs/activation.ts
Hweinstock: Yes exactly.

src/cloudWatchLogs/commands/searchLogGroup.ts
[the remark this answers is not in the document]
Hweinstock: That makes sense. Yeah I was trying to not have to have two different `if` statements to catch possible errors, but this meant that an error went unchecked for longer than it needed to. I adjusted it to catch the error earlier on.

src/cloudWatchLogs/activation.ts
JadenSimon: You won't be able to do what you're trying to do w/ glob patterns. I'd omit the pattern for now and make a backlog item. We'd probably use a different language feature to avoid this problem, like rendering ghost text somewhere by relevant events.

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -73,19 +75,21 @@ export async function changeLogSearchParams(
...
         result = 'Cancelled'
         return
     }
 
-    registry.deregisterLog(oldUri)
     const newUri = createURIFromArgs(newData.logGroupInfo, newData.parameters)
 
     result = await prepareDocument(newUri, newData, registry)
-
     console.log(result)
```
JadenSimon, line 94: Make sure to delete these random `console.log` calls. It's ok for development but otherwise it's noise.

Hweinstock: Yeah, I have it in as a placeholder for telemetry, but I can delete them now?

JadenSimon: Oh I see. Yeah just delete them and leave comments instead:

```ts
// TODO: add telemetry ...
```

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
```diff
@@ -35,6 +40,54 @@ export class LogStreamDocumentProvider implements vscode.TextDocumentContentProv
...
+                data: [],
+                parameters: parameters,
+                logGroupInfo: logGroupInfo,
+                retrieveLogsFunction: getLogEventsFromUriComponents,
+                busy: false,
+            }
+            await this.registry.registerLog(streamUri, initialStreamData)
+            // Set the document language
+            const doc = await vscode.workspace.openTextDocument(streamUri)
+            vscode.languages.setTextDocumentLanguage(doc, 'log')
```
JadenSimon, line 85: This looks to be duplicated in several places. We'll want to factor this out.

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -185,6 +199,31 @@ export class LogStreamRegistry {
...
+        return this.activeLogs.get(uriToKey(uri))?.streamIds
+    }
+
+    private setRangeForStreamIdMap(uri: vscode.Uri, lowerBound: number, upperBound: number, streamID: string): void {
+        // Note: Inclusive of both bounds.
+        let currentLine = lowerBound
+        while (currentLine <= upperBound) {
+            this.setStreamIdMap(uri, currentLine, streamID)
+            currentLine += 1
+        }
```
JadenSimon, line 213: Minor change for readability:
```suggestion
        for (let currentLine = lowerBound; currentLine <= upperBound; currentLine++) {
            this.setStreamIdMap(uri, currentLine, streamID)
        }
```

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
```diff
@@ -35,6 +40,54 @@ export class LogStreamDocumentProvider implements vscode.TextDocumentContentProv
...
+    ): Promise<vscode.Definition | vscode.LocationLink[] | undefined> {
+        console.log('provideDefinition')
+        const activeUri = document.uri
+        const logGroupInfo = parseCloudWatchLogsUri(activeUri).logGroupInfo
+        if (logGroupInfo.streamName) {
+            // This means we have a stream file not a log search.
+            return
+        }
+        const curLine = document.lineAt(position.line)
+        let streamUri: vscode.Uri
```
JadenSimon, line 57: This is declared pretty far from assignment. Hurts readability.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
```diff
@@ -76,7 +82,8 @@ export function createURIFromArgs(
     logGroupInfo: CloudWatchLogsGroupInfo,
     parameters: CloudWatchLogsParameters
 ): vscode.Uri {
-    let uriStr = `${CLOUDWATCH_LOGS_SCHEME}:${logGroupInfo.groupName}:${logGroupInfo.regionName}`
+    let uriStr = `${CLOUDWATCH_LOGS_SCHEME}:${logGroupInfo.regionName}:${logGroupInfo.groupName}`
+    uriStr += logGroupInfo.streamName ? `:${logGroupInfo.streamName}` : ''
```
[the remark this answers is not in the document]
Hweinstock: Justin recommended the same, but with this change I am unsure how to parse the `logGroupInfo` from the Uri. The `groupName` and `streamName` have slashes in them so I can't simply split on the slashes to differentiate them. ANy ideas on how to solve this so I can swap to slashes?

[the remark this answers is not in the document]
JadenSimon: Ah right I forgot about that. Stream name would probably need to be apart of the query and then rendered some other way. Let's not worry about it.

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -18,28 +16,27 @@ import { isViewAllEvents, TimeFilterResponse, TimeFilterSubmenu } from './timeFi
...
+    oldData: CloudWatchLogsData
 ): Promise<CloudWatchLogsData | undefined> {
-    const oldData = registry.getLogData(oldUri)
-    if (!oldData) {
-        throw new Error(`cwl: LogStreamRegistry did not contain ${String(oldUri)}`)
-    }
+    // We must deepcopy the parameters so that we don't change their original value in oldData
     const newData: CloudWatchLogsData = {
         ...oldData,
+        parameters: { ...oldData.parameters },
```
JadenSimon, line 27: Oops I forgot to include `{ }` in my suggestion to copy objects. But you were able to figure it out.

JadenSimon (approved):

src/cloudWatchLogs/cloudWatchLogsUtils.ts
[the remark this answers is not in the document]
justinmk3: Do we need a disambiguator though? Perhaps...

- the first path segment should be "logstream" or "search".
    - `cwl://logstream/{region}/…`
    - `cwl://search/{region}/…`
- Or the presence of a querystring could be inferred to be a search.

package.json
```diff
@@ -3046,7 +3046,7 @@
                 "scheme": "awsCloudWatchLogs",
                 "formatting": {
                     "label": "${path}",
-                    "separator": "\\"
+                    "separator": "/"
```
justinmk3, line 3049: :D

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
```diff
@@ -35,6 +40,52 @@ export class LogStreamDocumentProvider implements vscode.TextDocumentContentProv
...
+        const logGroupInfo = parseCloudWatchLogsUri(activeUri).logGroupInfo
+        if (logGroupInfo.streamName) {
+            // This means we have a stream file not a log search.
+            return
+        }
+        const curLine = document.lineAt(position.line)
+        try {
+            const streamIDMap = this.registry.getStreamIdMap(activeUri)
+            if (!streamIDMap || streamIDMap.size === 0) {
+                throw new Error(`cwl: No streamIDMap found for stream with uri ${activeUri}`)
```
justinmk3, line 59: does the string form of `activeUri` potentially contain a big JSON blob? don't want that in an error message. can we get the URI without the querystring part?

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
```diff
@@ -35,6 +40,52 @@ export class LogStreamDocumentProvider implements vscode.TextDocumentContentProv
...
+                data: [],
+                parameters: parameters,
+                logGroupInfo: logGroupInfo,
+                retrieveLogsFunction: getLogEventsFromUriComponents,
+                busy: false,
+            }
+            await this.registry.registerLog(streamUri, initialStreamData)
+            // Set the document language
+            const doc = await vscode.workspace.openTextDocument(streamUri)
+            vscode.languages.setTextDocumentLanguage(doc, 'log')
```
justinmk3, line 83: do we need a `onDidClose` handler to dispose the `streamIDMap` in the registry?

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -185,6 +199,29 @@ export class LogStreamRegistry {
...
+
+    private setRangeForStreamIdMap(uri: vscode.Uri, lowerBound: number, upperBound: number, streamID: string): void {
+        // Note: Inclusive of both bounds.
+        for (let currentLine = lowerBound; currentLine <= upperBound; currentLine++) {
+            this.setStreamIdMap(uri, currentLine, streamID)
+        }
+    }
+
+    private setStreamIdMap(uri: vscode.Uri, lineNum: number, streamID: string): void {
+        const ActiveTab = this.getActiveTab(uri)
```
justinmk3, line 215: nit: style
```suggestion
        const activeTab = this.getActiveTab(uri)
```

src/test/cloudWatchLogs/registry/logStreamRegistry.test.ts
```diff
@@ -150,21 +150,62 @@ describe('LogStreamRegistry', async function () {
...
-            assert(searchLogGroupStreamName)
-            assert(viewLogStreamName)
+        it('registers stream ids to map', function () {
+            registry.getLogContent(searchLogGroupUri) // We run this to create the mappings
+            const streamIDMap = registry.getStreamIdMap(searchLogGroupUri)
+            const expectedMap = new Map<number, string>([
+                [0, testStreamNames[0]],
+                [1, testStreamNames[1]],
+            ])
+            assert.deepStrictEqual(streamIDMap, expectedMap)
```
justinmk3, line 160: This test should also assert that the streamIdMap is disposed / deleted after the document is closed.

justinmk3 (approved):
looks good enough to merge and continue in a followup PR!

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
Hweinstock: Yea, I tried to factor out the parts that were easy to. A lot of this code is in `prepareDocument`, but theres two key differences that don't allow me to reuse that. First, I don't want to show the document here, just load it. And secondly, I don't need to worry about cancellation here (can't cancel a jumpto action).

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
Hweinstock: yeah, I can change it to `uri.path`.

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
Hweinstock: That sounds like a good idea. Just went ahead and implemented that.

src/test/cloudWatchLogs/registry/logStreamRegistry.test.ts
Hweinstock: In order to do so I would have to programmatically close the `vscode.TextDocument` which I am not sure if it is supported. (https://github.com/microsoft/vscode/issues/49740 asks about this a few years ago and don't see anything more recent in linked to issue) Is there a workaround/another way to test this?

src/cloudWatchLogs/cloudWatchLogsUtils.ts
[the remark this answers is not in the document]
JadenSimon: Yeah a path segment prefix would work for making the filter pattern work.
And you could use the same strategy for separating the group from stream. It won't look very nice though.

src/test/cloudWatchLogs/registry/logStreamRegistry.test.ts
JadenSimon: It's not supported. When you run into these kinds of situations you can write your logic in such a way to make it testable by abstracting the untestable parts. For example, expose a way to deregister a URI and then test that. The document close event would call this deregister function but your test code is agnostic to that fact.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -70,6 +70,14 @@ function handleWizardResponse(response: SearchLogGroupWizardResponse, registry:
     return initialStreamData
 }
 
+export function clearStreamIdMapOnDocumentClose(registry: LogStreamRegistry, doc: vscode.TextDocument): void {
+    vscode.workspace.onDidCloseTextDocument((document: vscode.TextDocument) => {
```
justinmk3, line 60: this creates a handler. If there's no way to force vscode to trigger the handler, you may need to create a function from the `(document: vscode.TextDocument) => { ... }` closure here.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -81,6 +84,7 @@ export async function prepareDocument(
                 highlightDocument(registry, uri)
             }
         })
+        clearStreamIdMapOnDocumentClose(registry, doc)
```
justinmk3, line 87: btw, this and the `onDidChangeTextDocument` result above it, should probably be stored with the registration in the `registry`. And then when the `registry` entry is destroyed, these handlers should be disposed.

This is the risk with "registries" of all kinds: have to manage the lifecycle. If we don't dispose these things, we will accumulate memory leaks.

justinmk3: onDidCloseTextDocument() should be called exactly once for the lifetime of the Toolkit, rather than once per prepareDocument() call.

Then just add it to subscriptions.push(...).

Hweinstock: I am assuming the second comment here is out of scope of this PR? Should I make a ticket for this?

src/cloudWatchLogs/cloudWatchLogsUtils.ts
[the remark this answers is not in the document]
Hweinstock: With this approach, I see how I can determine the difference between logSeach and stream, but how could I distinguish the boundary between `streamName` and `groupName` in the parsing of the Uri? They would have slashes among them and separating them.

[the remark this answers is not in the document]
JadenSimon: You can always have the path be whatever you want and encode everything else in the query params.
It's not like anything else is parsing the path. Just be internally consistent with how the path is constructed.

src/cloudWatchLogs/commands/searchLogGroup.ts
JadenSimon: I'd do it in a separate PR IMO

src/cloudWatchLogs/cloudWatchLogsUtils.ts
```diff
@@ -120,4 +120,26 @@ export function findOccurencesOf(document: vscode.TextDocument, keyword: string)
     return ranges
 }
 
+const HIGHLIGHTER = vscode.window.createTextEditorDecorationType({
```
justinmk3, line 123:
```suggestion
const searchHighlight = vscode.window.createTextEditorDecorationType({
```

src/cloudWatchLogs/cloudWatchLogsUtils.ts
```diff
@@ -120,4 +120,26 @@ export function findOccurencesOf(document: vscode.TextDocument, keyword: string)
...
+    backgroundColor: new vscode.ThemeColor('list.focusHighlightForeground'),
+})
+
+export async function highlightDocument(registry: LogStreamRegistry, uri: vscode.Uri): Promise<void> {
+    const textEditor = registry.getTextEditor(uri)
+    const logData = registry.getLogData(uri)
```
justinmk3, line 129: Seems like highlightDocument() should just live on the `registry` object. IIRC, we decided that it's not really a "registry" but more like a manager or something? It's a singleton where CWL streams/searches are executed and presented.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
```diff
@@ -120,4 +120,26 @@ export function findOccurencesOf(document: vscode.TextDocument, keyword: string)
...
+
+    if (!logData) {
+        throw new Error(`Missing log data in registry for uri key: ${uriToKey(uri)}. Unable to highlight`)
+    }
+
+    if (!textEditor) {
+        throw new Error(`Missing textEditor in registry for uri key: ${uriToKey(uri)}. Unable to highlight`)
```
justinmk3, line 136: variable part, especially potentially long strings like a URI, should be at the very end of messages if possible. Otherwise when troubleshooting, the user sees a very long message and must visit the end to see "Unable to highlight". If "Unable to highlight" is worth mentioning, put it at the front.

```suggestion
        throw new Error(`Highlighting failed. Missing textEditor in registry for uri key: ${uriToKey(uri)}`)
```

src/cloudWatchLogs/cloudWatchLogsUtils.ts
```diff
@@ -120,4 +120,26 @@ export function findOccurencesOf(document: vscode.TextDocument, keyword: string)
...
+const HIGHLIGHTER = vscode.window.createTextEditorDecorationType({
+    backgroundColor: new vscode.ThemeColor('list.focusHighlightForeground'),
+})
+
+export async function highlightDocument(registry: LogStreamRegistry, uri: vscode.Uri): Promise<void> {
+    const textEditor = registry.getTextEditor(uri)
+    const logData = registry.getLogData(uri)
+
+    if (!logData) {
+        throw new Error(`Missing log data in registry for uri key: ${uriToKey(uri)}. Unable to highlight`)
```
justinmk3, line 132: todo/future pr: should this use ToolkitError?

src/cloudWatchLogs/cloudWatchLogsUtils.ts
```diff
@@ -120,4 +120,26 @@ export function findOccurencesOf(document: vscode.TextDocument, keyword: string)
...
+    }
+
+    if (!textEditor) {
+        throw new Error(`Missing textEditor in registry for uri key: ${uriToKey(uri)}. Unable to highlight`)
+    }
+
+    if (logData.parameters.filterPattern) {
+        const ranges = findOccurencesOf(textEditor.document, logData.parameters.filterPattern)
```
justinmk3, line 140: doesn't look like `findOccurencesOf()` is specific to CWL, so it should live in `src/shared/utilities/textDocumentUtilities.ts`

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -35,17 +35,36 @@ export class LogStreamRegistry {
...
+    public getOnCloseFuncForUri(uri: vscode.Uri): (document: vscode.TextDocument) => void {
+        return (document: vscode.TextDocument) => {
+            if (document.uri.toString() === uri.toString()) {
+                this.clearStreamIdMap(uri)
+            }
+        }
+    }
+
+    private registerLogHandlers(uri: vscode.Uri): void {
+        vscode.workspace.onDidChangeTextDocument((event: vscode.TextDocumentChangeEvent) => {
```
justinmk3, line 59: Just to be clear: this should be done once, in total. Not once per `registerLog()` call.

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -35,17 +35,36 @@ export class LogStreamRegistry {
...
+            if (event.document.uri.toString() === uri.toString()) {
+                highlightDocument(this, uri)
+            }
```
justinmk3, line 62: With a condition like the following, only one global handler is needed:

```suggestion
            if (<is uri in cwl registry?>) {
                highlightDocument(this, uri)
            }
```

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -35,17 +35,36 @@ export class LogStreamRegistry {
...
+        })
+
+        vscode.workspace.onDidCloseTextDocument(this.getOnCloseFuncForUri(uri))
```
justinmk3, line 65: Same for this.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
Hweinstock: Yeah I was trying not to overload the registry, but I agree that it seems like the most natural place for it. Best not to force it elsewhere.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
Hweinstock: I didn't know this existed but after looking into it, is there any reason not to use ToolkitError for errors within the toolkit?

src/cloudWatchLogs/cloudWatchLogsUtils.ts
JadenSimon: `const` is an interesting keyword is Javascript because all it means is that the variable won't be re-assigned. The original intent of UPPER_CASE naming is to express true universal constants used as configuation that would be in-lined into the executable. The string 'abcde' will never change regardless of execution environment whereas something like `vscode.window.createTextEditorDecorationType` is most definitely not a universal constant.

I think there's at least some value in UPPER_CASE variable names in that they stand out a lot. Still, most things developers normally make UPPER_CASE really shouldn't be. Either because the value is only used in one place or because it's obvious from context what the value represents. Might as well enforce camelCase.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
JadenSimon: If we merged the codelens provider w/ the document provider then `highlightDocument` would fit there.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
JadenSimon: It's on-going work but ideally most contributors would just use `ToolkitError`. 

The message you use might be shown directly to the user so it needs a little more thought. A normal `Error` won't be treated the same as we don't know where the error came from. 

Right now, your current message would be confusing to a user but potentially helpful for a Toolkit developer. We'd want to phrase the message differently and put useful data somewhere else in a structured way:
```ts
throw new ToolkitError(`Failed to highlight document "${uri.path}". No log data found.`, { details: { key: uriToKey(uri) } })
```

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -42,27 +52,26 @@ export class LogStreamRegistry {
...
             await this.updateLog(uri, 'tail')
         }
     }
 
-    public getOnCloseFuncForUri(uri: vscode.Uri): (document: vscode.TextDocument) => void {
-        return (document: vscode.TextDocument) => {
-            if (document.uri.toString() === uri.toString()) {
-                this.clearStreamIdMap(uri)
-            }
+    public cleanUpDocument(document: vscode.TextDocument): void {
```
justinmk3, line 63: "dispose" is the common term, let's use that for discoverability.

```suggestion
    public disposeDocument(document: vscode.TextDocument): void {
```

src/cloudWatchLogs/registry/logStreamRegistry.ts
```diff
@@ -42,27 +52,26 @@ export class LogStreamRegistry {
...
+        console.log(this.hasLog(document.uri), !isLogStreamUri(document.uri))
+        if (this.hasLog(document.uri) && !isLogStreamUri(document.uri)) {
+            this.clearStreamIdMap(document.uri)
```
justinmk3, line 65: `document.uri` is used everywhere here, so the parameter should just be a Uri.

justinmk3 (approved):

justinmk3 (approved):

PR #2787, opened 2022-07-30
fix(cwl): various fixes

justinmk3:
- Avoid "CWL" in UI.
- Avoid "keyword" in UI. Prefer "Search Log Group", "Search pattern", etc.
- Show (truncated) log group name in search wizard.

## License


src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -144,10 +145,10 @@ async function logGroupsToArray(logGroups: AsyncIterableIterator<CloudWatchLogs.
     return logGroupsArray
 }
 
-export function createFilterpatternPrompter() {
+export function createFilterpatternPrompter(logGroupName: string) {
     return createInputBox({
-        title: 'Keyword Search',
-        placeholder: 'Enter text here',
+        title: `Search Log Group ${truncate(logGroupName, -50)}`,
```
justinmk3, line 150: example:

image

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -175,7 +176,7 @@ export class SearchLogGroupWizard extends Wizard<SearchLogGroupWizardResponse> {
         })
 
         this.form.submenuResponse.bindPrompter(createRegionSubmenu)
-        this.form.filterPattern.bindPrompter(createFilterpatternPrompter)
+        this.form.filterPattern.bindPrompter(() => createFilterpatternPrompter(logGroupInfo?.groupName ?? ''))
```
JadenSimon, line 179:
```suggestion
        this.form.filterPattern.bindPrompter(({ submenuResponse }) => createFilterpatternPrompter(submenuResponse!.data))
```

You'd want to cover the non-null assertion with a test which I think it already is. I gotta make a PR to explicitly define dependencies to remove these type assertions.

JadenSimon (approved):

src/cloudWatchLogs/commands/searchLogGroup.ts
Hweinstock: I like the idea of showing only the end. That makes it much easier to see what log group we are dealing without overcrowding the user.

Hweinstock (approved):

PR #2789, opened 2022-08-01
Hkobew/fix cancel search

Hweinstock:
## Problem
BUG: Cancelling a log group search would result in an error. 
## Solution
The `prepareDocument` function in `searchLogGroup.ts` did not contain that function(`registry.registerLog`) that it needed to contain in order to catch cancellation errors. 

## License

### DIFF
--- a/src/cloudWatchLogs/commands/searchLogGroup.ts
+++ b/src/cloudWatchLogs/commands/searchLogGroup.ts
@@ -60,8 +60,13 @@ function handleWizardResponse(response: SearchLogGroupWizardResponse, registry: 
     return initialStreamData
 }
 
-export async function prepareDocument(uri: vscode.Uri, registry: LogStreamRegistry): Promise<telemetry.Result> {
+export async function prepareDocument(
+    uri: vscode.Uri,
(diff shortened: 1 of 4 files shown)


justinmk3 (approved):

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -60,8 +60,13 @@ function handleWizardResponse(response: SearchLogGroupWizardResponse, registry:
...
 }
 
-export async function prepareDocument(uri: vscode.Uri, registry: LogStreamRegistry): Promise<telemetry.Result> {
+export async function prepareDocument(
+    uri: vscode.Uri,
+    initialLogData: CloudWatchLogsData,
+    registry: LogStreamRegistry
+): Promise<telemetry.Result> {
     try {
+        await registry.registerLog(uri, initialLogData)
```
JadenSimon, line 69: This works but hopefully you can see why restricting this 'try/catch' block to `prepareDocument` is problematic. Don't change anything of course, just understand that this fix was for _one_ bug.

PR #2790, opened 2022-08-01
CWL: don't show commands on commandPalette

Hweinstock:
## Problem
There are three cloudWatchLogs commands available when viewing a log group search or log stream. `aws.cwl.changeFilterPattern` and `aws.cwl.changeTimeFilter` don't offer anything `aws.cwl.searchLogGroup` doesn't. 
## Solution
We eliminate `aws.cwl.changeFilterPattern` and `aws.cwl.changeTimeFilter` from command Palette. 

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -846,6 +846,14 @@
                     "when": "false"
                 },
                 {
+                    "command": "aws.cwl.changeFilterPattern",
+                    "when": "false"
+                },
+                {
+                    "command": "aws.cwl.changeTimeFilter",
+                    "when": "false"
+                },
+                {
                     "command": "aws.ecr.deleteRepository",
                     "when": "false"
                 },


JadenSimon (approved):

package.json
```diff
@@ -845,6 +845,14 @@
                     "command": "aws.cwl.viewLogStream",
                     "when": "false"
                 },
+                {
+                    "command": "aws.cwl.changeFilterPattern",
```
justinmk3, line 849: Is `changeFilterPattern` actually needed at all? Can the search icon just point to `aws.cwl.changeLogSearch` ?

Hweinstock: I think `aws.cwl.changeFilterPattern` and `aws.cwl. changeTimeFilter` have to be distinct commands in order to allow two distinct icons. There currently is no `aws.cwl.changeLogSearch` command, its just that both `aws.cwl.changeFilterPattern` and `aws.cwl. changeTimeFilter` are registered to the `changeLogSearchParams` function with different params. There may be another way to keep them with distinct icons and function, but this is how I thought to get around that issue.

JadenSimon: > have to be distinct commands in order to allow two distinct icons.

Yup. This is the best solution I've found so far.

justinmk3: > There currently is no aws.cwl.changeLogSearch command

sorry, see updated comment. I copy-pasted the wrong name.

JadenSimon: Both commands pretty much do the same thing as `aws.cwl.searchLogGroup`, you're just starting from existing parameters. This would be more obvious once you realize that `changeLogSearch.ts` is mostly redundant; we can re-use the wizard implementation.

justinmk3 (approved):

justinmk3:
> `changeLogSearch.ts` is mostly redundant; we can re-use the wizard implementation

Do we have a backlog item for that or is it WIP already?

JadenSimon:
> Do we have a backlog item for that or is it WIP already?

I don't think there's an item. It's a straightforward refactor though once it 'clicks' .
The same idea applies for the legacy 'viewLogStream' command.

Hweinstock:
I created a ticket a while ago to refactor `viewLogStream` and `searchLogGroup` to generalize the two. I went ahead and added `changeLogSearch` to that list. Seems likes it would be easy to hit all three in the same process.

PR #2792, opened 2022-08-02
fix(codewhisperer): Add userDecision telemetry for empty recommendation

leigaol:
## Problem
1. Recommendations of empty string or empty list are not reported in userDecision telemetry.
2. The type prefix matching for determining Discard event should only be performed when client gets the suggestion from server. 

## Solution
1. Report useDecision events for empty recommendations.
2. Do telemetry type prefix matching when suggestion first reaches client.
3. If a suggestion is marked as Discard but later it was showed to user, it won't be reported as Discard in userDecision telemetry.

Note:

This fix is not user facing so no changelog item is added.

This PR has been tested with below test cases:

1. With recommendation = [] at line 259 of recommendationHandler.ts that simulate a Empty List recommendations. userDecision event with Empty was sent.
2. With r.content = '' at line 266 of that simulate a Empty string recommendations.  userDecision event with Empty was sent. 
4. Trigger at `def ` in a python file and type `test_i` before first reco returns, the recos not starting with `test_i` was marked as Discard.

## License

### DIFF
--- a/src/codewhisperer/util/telemetryHelper.ts
+++ b/src/codewhisperer/util/telemetryHelper.ts
@@ -109,13 +109,9 @@ export class TelemetryHelper {
         acceptIndex: number,
         recommendationSuggestionState?: Map<number, string>
     ): telemetry.CodewhispererSuggestionState {
-        if (recommendationSuggestionState?.get(i) === 'Empty') {
-            return 'Empty'
-        }
-        if (recommendationSuggestionState?.get(i) === 'Filter') {
-            return 'Filter'
-        } else if (recommendationSuggestionState?.get(i) === 'Discard') {
-            return 'Discard'
+        const state = recommendationSuggestionState?.get(i)
+        if (state && ['Empty', 'Filter', 'Discard'].includes(state)) {
+            return state as telemetry.CodewhispererSuggestionState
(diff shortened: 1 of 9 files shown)


justinmk3 (approved):

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -147,22 +107,21 @@ export class TelemetryHelper {
...
-        if (unseen) return 'Unseen'
-        if (acceptIndex == -1) {
-            return this.isPrefixMatched[i] ? 'Reject' : 'Discard'
+        if (recommendationSuggestionState?.get(i) === 'Empty') {
+            return 'Empty'
         }
-        if (!this.isPrefixMatched[i]) {
+        if (recommendationSuggestionState?.get(i) === 'Filter') {
+            return 'Filter'
+        } else if (recommendationSuggestionState?.get(i) === 'Discard') {
```
zixlin7, line 117: nit: those can be combined?
```
if(['Empty', 'Filter', 'Discard'].includes(recommendationSuggestionState?.get(i))) {
   return recommendationSuggestionState?.get(i)
}
```

zixlin7 (approved):

src/codewhisperer/util/telemetryHelper.ts
leigaol: Yes they can

Issue #2797, opened 2022-08-03
Golang Debugger for Lambdas throws "Failed to install Delve for lambda container"


AndrewKovalenko:
## Problem
  Debugging of lambda written with Golang doesn't start. 
As I hit "Run" button - toolkit terminates with a message:

![image](https://user-images.githubusercontent.com/1983017/182505787-b3394e1e-2f9a-4410-b9bc-ca84eae72f8c.png)

Logs show:
```
2022-08-02 17:08:04 [ERROR]: log level: info
2022-08-02 17:08:05 [INFO]: Retrieving AWS endpoint data
2022-08-02 17:08:05 [INFO]: OS: Darwin x64 20.6.0
2022-08-02 17:08:05 [INFO]: Visual Studio Code extension host:  1.69.2
2022-08-02 17:08:05 [INFO]: AWS Toolkit:  1.46.0
2022-08-02 17:08:05 [INFO]: node: 16.13.2
2022-08-02 17:08:05 [INFO]: electron: 18.3.5
2022-08-02 17:08:05 [INFO]: telemetry cache not found: '/Users/andrewkovalenko/Library/Application Support/Code/User/globalStorage/amazonwebservices.aws-toolkit-vscode/telemetryCache'
2022-08-02 17:08:19 [INFO]: SAM CLI location: /usr/local/bin/sam
2022-08-02 17:08:19 [INFO]: SAM CLI location: /usr/local/bin/sam
2022-08-02 17:08:19 [INFO]: SAM CLI location: /usr/local/bin/sam
2022-08-02 17:08:20 [WARN]: SAM debug: missing AWS credentials (Toolkit is not connected)
2022-08-02 17:08:20 [INFO]: autoconnect: connected: 'profile:default'
2022-08-02 17:08:20 [INFO]: The Delve repo was not found in your GOPATH. Downloading in a temporary directory...
2022-08-02 17:08:21 [INFO]: Command: (not started) [/Users/andrewkovalenko/Library/Application Support/Code/User/globalStorage/amazonwebservices.aws-toolkit-vscode/debuggers/delve/install-1.9.0.sh ]
2022-08-02 17:08:21 [ERROR]: Failed to cross-compile Delve debugger: Error: spawn Unknown system error -8
``` 

Configuration generated by CodeLense :

```
{
  "configurations": [
    {
      "type": "aws-sam",
      "request": "direct-invoke",
      "name": "test-lambda:HelloWorldFunction (go1.x)",
[...]

PR #2666, opened 2022-05-29
fix(tests): use shebang line for test scripts 

JadenSimon:
## Problem
Tests on macOS/Insiders are failing due to changes in node 16 

## Solution
Add shebang line to shell scripts

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2800, opened 2022-08-03
fix(SAM/Go): update Delve install script

JadenSimon:
## Problem
See #2797

This affects MacOS running Node 16+, presumably due to a change in libuv.
Same root cause as https://github.com/aws/aws-toolkit-vscode/pull/2666

## Solution
Add shebang line. I don't think we have any other places with shell scripts we create but could be wrong.

## License

### DIFF
--- a/.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
+++ b/.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Delve debugger installation failures for Go Lambdas on MacOS"
+	"description": "Delve debugger installation fails for Go Lambdas on MacOS"
 }
--- a/.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
+++ b/.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "fixed Delve debugger installation for Go Lambdas on MacOS"
+	"description": "Delve debugger installation failures for Go Lambdas on MacOS"
 }
--- a/.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
+++ b/.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "fixed Delve debugger installation for Go Lambdas on MacOS"
+}
--- a/src/shared/sam/debugger/goSamDebug.ts
+++ b/src/shared/sam/debugger/goSamDebug.ts
@@ -163,7 +163,7 @@ interface InstallScript {
  * @returns Path for the debugger install script, undefined if we already built the binary
  */
 async function makeInstallScript(debuggerPath: string, isWindows: boolean): Promise<InstallScript | undefined> {
-    let script: string = ''
+    let script: string = isWindows ? '' : '#!/bin/sh\n'
     const DELVE_REPO: string = 'github.com/go-delve/delve'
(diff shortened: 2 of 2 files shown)


src/shared/sam/debugger/goSamDebug.ts
```diff
@@ -163,7 +163,7 @@ interface InstallScript {
  * @returns Path for the debugger install script, undefined if we already built the binary
  */
 async function makeInstallScript(debuggerPath: string, isWindows: boolean): Promise<InstallScript | undefined> {
-    let script: string = ''
+    let script: string = isWindows ? '' : '#!/bin/sh\n'
```
justinmk3, line 166: `#!/usr/bin/env sh` ?

.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "fixed Delve debugger installation for Go Lambdas on MacOS"
```
justinmk3, line 3:
```suggestion
	"description": "Delve debugger installation fails for Go Lambdas on MacOS"
```

justinmk3 (approved):

.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
JadenSimon: Wouldn't it be "no longer fails"?

src/shared/sam/debugger/goSamDebug.ts
JadenSimon: I did some reading and it seems like `#!/bin/sh` is preferred over `#!/usr/bin/env sh` for `sh`. Plus this is the same behavior as POSIX spawn which is what we want.

For other shells you'd want to use `env`

.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
justinmk3: The changelog lists items as "Fix" or "Feature", etc. There is no need to repeat "fixed" in the description. And it's clear from the "Fix" label that the thing described, is being fixed.

This usually leads to clearer, less-awkward wording for these changelog items.

JadenSimon: Yeah that makes sense. Grammatically we should state _what_ was fixed then, so "failures" instead of "fails".

jpinkney-aws (approved):

.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
justinmk3: Generally active voice ("fails") leads to less-awkward, more-concise sentence structure.

.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Delve debugger installation failures for Go Lambdas on MacOS"
```
JadenSimon, line 3:
```suggestion
	"description": "Delve debugger installation fails for Go Lambdas on MacOS"
```

.changes/next-release/Bug Fix-1859b589-4651-4fa5-9545-317fceaff3f1.json
JadenSimon: For me the active voice is kind of awkward though because the actor is implied. Toolkits is the one fixing the bug. So the following sentence is a fragment defining what the bug is.

I changed it to `fails` though for consistency. Full sentences are probably more intuitive for people to write.

justinmk3 (approved):

PR #2803, opened 2022-08-03
feat(codewhisperer): add repo url to reference tracker

zixlin7:
## Problem
to provide link to the referenced repo to reference tracker log
## Solution
- if `url` is present in the reference, we will show link to the repo instead of link to license:
Screen Shot 2022-08-03 at 2 50 32 PM
- if `url` is missing, we will keep the current behavior with link to the license and repo name:
Screen Shot 2022-08-03 at 3 28 21 PM

## License


leigaol:
I think the service-2.json needs to be updated with this change.

zixlin7:
> I think the service-2.json needs to be updated with this change.

@leigaol it's already updated with #2795, not sure why it's still failing, anything else I miss?

justinmk3:
src/codewhisperer/service/referenceLogViewProvider.ts(87,27): error TS2339: Property 'url' does not exist on type 'Reference'.

> > I think the service-2.json needs to be updated with this change.
> 
> @leigaol it's already updated with #2795, not sure why it's still failing, anything else I miss?

Hmm. CI log shows that the client + d.ts file generated. 

/retrybuilds

JadenSimon:
@zixlin7

Try merging `master` into this branch. I think the Linux tests don't do this automatically.

zixlin7:
> @zixlin7
> 
> Try merging `master` into this branch. I think the Linux tests don't do this automatically.

will try that, thanks!

JadenSimon (approved):
Unit tests look bloated but there's only two so it's ok for now.

fangwade007 (approved):

justinmk3:
Doesn't this deserve a changelog entry? Something like:

    CodeWhisperer: References include a link to the source repository if possible.

PR #2804, opened 2022-08-03
fix(tests): detectCdkProjects sometimes fails

justinmk3:
## Problem:
on Windows CI, sometimes the detectCdkProjects() tests fail:

      1) detectCdkProjects
           detects deep projects:
    + actual - expected

    + undefined
    - 'c:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\aws-toolkit-vscode\\vsctkL53wNx\\directory1\\directory2\\cdk.json'
          at Context.<anonymous> (d:\a\aws-toolkit-vscode\aws-toolkit-vscode\src\test\cdk\detectCdkProjects.test.ts:77:16)

## Solution:
Check that the test file was written before trying to detect it.

## License


JadenSimon (approved):
> Check that the test file was written before trying to detect it.

Node will have closed the file handle before `writeJSON` resolves. The flakiness probably has something to do w/ however VS Code handles either `findFiles` or `openTextDocument`. 

If either use events for optimization then we'd have a race condition.

PR #2805, opened 2022-08-04
fix(tests): update `visualizeStateMachine` 

JadenSimon:
## Problem
These tests execute extension commands, and one test specifically checks for an error. Commands now show an error message instead of bubbling it up so this test ended up stalling on the error message.

## Solution
Fix the test so it checks for an error message.

Need to figure out why Mocha timeouts don't work so these stalled tests are easier to find.

## License


justinmk3:
> Need to figure out why Mocha timeouts don't work so these stalled tests are easier to find.

Meanwhile there is `setTestTimeout` which exists to workaround that issue: https://github.com/aws/aws-toolkit-vscode/blob/61d9292d4cb2c2d705756afd7ebf46276947315d/src/integrationTest/globalSetup.test.ts#L26-L33

justinmk3 (approved):

PR #2808, opened 2022-08-05
cwl: telemetry

Hweinstock:
## Problem
CloudWatchLogs feature has not updated any telemetry metrics.
## Solution
We use `cloudwatchlogs_open` and `cloudwatchlogs_filter` and update metric `cloudwatchlogs_filter` to include the source. Both of which are fully implemented https://github.com/aws/aws-toolkit-common/blob/master/telemetry/definitions/commonDefinitions.json. 

## License

### DIFF
--- a/src/cloudWatchLogs/changeLogSearch.ts
+++ b/src/cloudWatchLogs/changeLogSearch.ts
@@ -4,7 +4,7 @@
  */
 import * as telemetry from '../shared/telemetry/telemetry'
 import { showInputBox } from '../shared/ui/inputPrompter'
-import { createURIFromArgs, telemetryFilter } from './cloudWatchLogsUtils'
+import { createURIFromArgs, isLogStreamUri, telemetryFilter } from './cloudWatchLogsUtils'
 import { prepareDocument } from './commands/searchLogGroup'
 import { getActiveDocumentUri } from './document/logStreamDocumentProvider'
 import { CloudWatchLogsData, filterLogEventsFromUriComponents, LogStreamRegistry } from './registry/logStreamRegistry'
@@ -64,7 +64,7 @@ export async function getNewData(
         resourceType = 'logStream'
     }
 
-    telemetryFilter(newData, resourceType)
+    telemetryFilter(newData, resourceType, 'escapeHatch')
 
     return newData
 }
@@ -78,12 +78,27 @@ export async function changeLogSearchParams(
 
     const oldData = registry.getLogData(oldUri)
     if (!oldData) {
+        telemetry.recordCloudwatchlogsFilter({
+            result: 'Failed',
+            source: 'escapeHatch',
+            cloudWatchResourceType: isLogStreamUri(oldUri) ? 'logStream' : 'logGroup',
+        })
         throw new Error(`cwl: Unable to find data for active URI ${oldUri}`)
     }
     const newData = await getNewData(param, oldData)
 
     if (!newData) {
-        //result = 'Cancelled'
+        telemetry.recordCloudwatchlogsFilter({
+            result: 'Cancelled',
+            source: 'escapeHatch',
+            cloudWatchResourceType: isLogStreamUri(oldUri) ? 'logStream' : 'logGroup',
+            hasTimeFilter: oldData.parameters.startTime || param === 'timeFilter' ? true : false,
+            hasTextFilter:
+                (oldData.parameters.filterPattern && oldData.parameters.filterPattern !== '') ||
+                param === 'filterPattern'
+                    ? true
+                    : false,
+        })
         return
     }
 
--- a/src/cloudWatchLogs/changeLogSearch.ts
+++ b/src/cloudWatchLogs/changeLogSearch.ts
@@ -100,6 +100,7 @@ export async function changeLogSearchParams(
                     ? true
                     : false,
         })
+
         return
     }
 
--- a/src/cloudWatchLogs/changeLogSearch.ts
+++ b/src/cloudWatchLogs/changeLogSearch.ts
@@ -57,14 +57,16 @@ export async function getNewData(
             newData.parameters.endTime = isViewAllEvents(newTimeRange) ? undefined : newTimeRange.end
             break
     }
-
-    telemetryFilterSuccess(newData)
+    let resourceType: telemetry.CloudWatchResourceType = 'logGroup'
 
     if (newData.parameters.streamName) {
         newData.retrieveLogsFunction = filterLogEventsFromUriComponents
         newData.parameters.streamNameOptions = [newData.parameters.streamName]
         newData.parameters.streamName = undefined
+        resourceType = 'logStream'
     }
+
+    telemetryFilterSuccess(newData, resourceType)
 
     return newData
 }
--- a/src/cloudWatchLogs/changeLogSearch.ts
+++ b/src/cloudWatchLogs/changeLogSearch.ts
@@ -64,7 +64,7 @@ export async function getNewData(
         resourceType = 'logStream'
     }
 
-    recordTelemetryFilter(newData, resourceType, 'Editor')
+    recordTelemetryFilter(newData, resourceType, 'EditorButton')
 
     return newData
 }
--- a/src/cloudWatchLogs/changeLogSearch.ts
+++ b/src/cloudWatchLogs/changeLogSearch.ts
@@ -64,7 +64,7 @@ export async function getNewData(
         resourceType = 'logStream'
     }
 
-    recordTelemetryFilter(newData, resourceType, 'escapeHatch')
+    recordTelemetryFilter(newData, resourceType, 'Editor')
 
     return newData
 }
--- a/src/cloudWatchLogs/changeLogSearch.ts
+++ b/src/cloudWatchLogs/changeLogSearch.ts
@@ -3,9 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import * as telemetry from '../shared/telemetry/telemetry'
-import * as vscode from 'vscode'
 import { showInputBox } from '../shared/ui/inputPrompter'
-import { createURIFromArgs, isLogStreamUri, telemetryFilter } from './cloudWatchLogsUtils'
+import { createURIFromArgs, telemetryFilter } from './cloudWatchLogsUtils'
 import { prepareDocument } from './commands/searchLogGroup'
 import { getActiveDocumentUri } from './document/logStreamDocumentProvider'
 import { CloudWatchLogsData, filterLogEventsFromUriComponents, LogStreamRegistry } from './registry/logStreamRegistry'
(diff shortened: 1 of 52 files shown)


src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -46,7 +46,11 @@ export async function getNewData(
                 return
             }
             newData.parameters.filterPattern = newPattern
-
+            telemetry.recordCloudwatchlogsFilter({
+                result: 'Succeeded',
+                cloudWatchResourceType: 'logGroup',
+                filterBy: 'prefix',
```
JadenSimon, line 52: Run `npm i`
You're local telemetry definitions are probably outdated.

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
```diff
@@ -45,10 +45,12 @@ export class LogStreamDocumentProvider implements vscode.TextDocumentContentProv
         position: vscode.Position,
         token: vscode.CancellationToken
     ): Promise<vscode.Definition | vscode.LocationLink[] | undefined> {
+        let result: telemetry.Result
         const activeUri = document.uri
         const logGroupInfo = parseCloudWatchLogsUri(activeUri).logGroupInfo
         if (logGroupInfo.streamName) {
             // This means we have a stream file not a log search.
+            // TODO: Should there be any telemetry output here?
             return
```
justinmk3, line 50: Don't think telemetry is needed here. 

Different question: is this an expected situation? If it signals a logic bug or bad input, that should be logged, if not raise an error.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
```diff
@@ -2,17 +2,38 @@
...
 // URIs are the only vehicle for delivering information to a TextDocumentContentProvider.
 // The following functions are used to structure and destructure relevant information to/from a URI.
 // Colons are not valid characters in either the group name or stream name and will be used as separators.
 
+/**
+ * records a metric that the filter was successful IF a filter was actually applied i.e one of time or filterPattern were set.
+ * @param logData
+ * @param resourceType
+ */
+export function telemetryFilter(
```
JadenSimon, line 21:
```suggestion
export function recordFilterTelemetry(
```

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
Hweinstock: Right now we are still missing a `glob` pattern to distinguish the logStream uri from a logGroup uri when we register the `defintionProvider`. So this case is hit anytime someone attempts to look for a definition in a single logStream document.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
```diff
@@ -2,17 +2,38 @@
...
+    logData: CloudWatchLogsData,
+    resourceType: telemetry.CloudWatchResourceType,
+    source: 'escapeHatch' | 'originalSearch'
```
JadenSimon, line 24: These two terms are kind of specific to the implementation. Can we come up with names closer to the UI?
Isn't `escapeHatch` the same as 'Editor' (or 'EditorButton')? And isn't `originalSearch` from the command palette?

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -53,39 +53,64 @@ export async function getNewData(
...
     }
     const newData = await getNewData(param, oldData)
 
     if (!newData) {
-        //result = 'Cancelled'
+        telemetry.recordCloudwatchlogsFilter({
+            result: 'Cancelled',
+            source: 'escapeHatch',
+            cloudWatchResourceType: isLogStreamUri(oldUri) ? 'logStream' : 'logGroup',
+            hasTimeFilter: oldData.parameters.startTime || param === 'timeFilter' ? true : false,
```
JadenSimon, line 96:
```suggestion
            hasTimeFilter: oldData.parameters.startTime || param === 'timeFilter',
```

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -53,39 +53,64 @@ export async function getNewData(
...
+            hasTextFilter:
+                (oldData.parameters.filterPattern && oldData.parameters.filterPattern !== '') ||
+                param === 'filterPattern'
+                    ? true
+                    : false,
```
JadenSimon, line 101:
```suggestion
                (oldData.parameters.filterPattern && oldData.parameters.filterPattern !== '') ||
                param === 'filterPattern',
```

src/cloudWatchLogs/cloudWatchLogsUtils.ts
```diff
@@ -2,17 +2,38 @@
...
+): void {
+    if (logData.parameters.startTime || logData.parameters.filterPattern) {
```
JadenSimon, line 26: Why do we skip if both are unset? Does some other metric get emitted?

src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -29,6 +29,22 @@
         }
     ],
     "metrics": [
+        {
+            "name": "cloudwatchlogs_filter",
+            "description": "Filters a CloudWatch Logs entity.",
+            "metadata": [
+                { "type": "result" },
+                { "type": "cloudWatchResourceType" },
+                { "type": "source" },
```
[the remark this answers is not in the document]
Hweinstock: https://github.com/aws/aws-toolkit-common/pull/351

src/cloudWatchLogs/cloudWatchLogsUtils.ts
Hweinstock: I agree that sounds like a good idea. The `originalSearch` isn't always from the palette, it could be from the explorer or anywhere the search is accessible. I don't believe I have the source of the search at this point, but I didn't think it was incredibly relevant. Should I distinguish this or label it `Search` or `OriginalSearch`?

src/cloudWatchLogs/changeLogSearch.ts
Hweinstock: had to add a `!== undefined` to avoid linting error, but believe it has the same function.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
Hweinstock: If neither is set then we aren't filtering the cloudwatchlogs, so I figured we shouldn't emit a metric that we are. The alternative would be to emit a metric with both of them as false, but that seemed weird to me. Is it still useful to see a filter metric if nothing is being filtered but a lot of the same code is running?

src/cloudWatchLogs/changeLogSearch.ts
JadenSimon: Usually if you see these kinds of errors you can do this:
```ts
!!(oldData.parameters.startTime || param === 'timeFilter')
```

This forces the expression into a boolean.

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -53,39 +53,62 @@ export async function getNewData(
...
+            hasTimeFilter: param === 'timeFilter',
+            hasTextFilter: param === 'filterPattern',
+        })
         throw new Error(`cwl: Unable to find data for active URI ${oldUri}`)
     }
     const newData = await getNewData(param, oldData)
 
     if (!newData) {
-        //result = 'Cancelled'
+        telemetry.recordCloudwatchlogsFilter({
```
justinmk3, line 92: If `recordCloudwatchlogsOpen` is for "opening a log stream without filtering" then we shouldn't record it for the filter case.

Alternatively, we don't need `recordCloudwatchlogsFilter` at all. We could add filter/search meatadata to `recordCloudwatchlogsOpen`. I don't see why we need both.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
JadenSimon: I guess you can put "Command" as I consider that to be the most generic source.

src/cloudWatchLogs/changeLogSearch.ts
JadenSimon: I believe VS uses "open" to refer to the editor being opened up while "filter" refers to any user gestures that change filters on the editor. 

This distinction is less relevant for us and can be captured via "source". I wouldn't try to overload `recordCloudwatchlogsOpen` too much though. I'd suggest using `recordCloudwatchlogsFilter` for when the editor is already opened and `recordCloudwatchlogsOpen` otherwise.

src/cloudWatchLogs/cloudWatchLogsUtils.ts
JadenSimon: Our telemetry metric names should reflect the action not necessarily the outcome. Someone removing existing filters is different from someone opening a fresh log group/stream without any filters despite having the same end result.

Hweinstock: I see. So a filter metric should still be emitted when they are both unset. Also, you highlight an important case that I missed: Undoing existing filters should definitely emit a metric.

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -53,39 +53,62 @@ export async function getNewData(
...
 
-    //result = await prepareDocument(newUri, newData, registry)
-    await prepareDocument(newUri, newData, registry)
-    // TODO: add telemetry
+    const result = await prepareDocument(newUri, newData, registry)
+    const typeOfResource = newData.parameters.streamNameOptions ? 'logStream' : 'logGroup'
+    telemetry.recordCloudwatchlogsOpen({
+        result: result,
+        cloudWatchResourceType: typeOfResource,
+        source: 'Editor',
```
justinmk3, line 112:
```suggestion
        source: 'EditorButton',
```

src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -29,6 +29,22 @@
...
+                { "type": "hasTextFilter", "required": false },
+                { "type": "hasTimeFilter", "required": false }
+            ]
+        },
+        {
+            "name": "cloudwatchlogs_jumpToStream",
```
justinmk3, line 44: If it's not hard to plumb through a source such as "JumpTo" or "ClickedSearchResults" , does it make sense to add this as a `source` on `recordCloudwatchlogsOpen` instead of adding a dedicated metric?

justinmk3: using existing definitions ,  `lambda_goToHandler` is a precedent . Could name this `cloudwatchlogs_goToStream`. 

But it's hard to argue that this is different than `cloudwatchlogs_open`, semantically. Only the source is different.

Hweinstock: I decided to swap to `cloudwatchlogs_open` since I don't see a strong argument for creating a new metric. If we don't need it, might as well not add it.

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
```diff
@@ -46,6 +46,7 @@ export class LogStreamDocumentProvider implements vscode.TextDocumentContentProv
         const logGroupInfo = parseCloudWatchLogsUri(activeUri).logGroupInfo
         if (logGroupInfo.streamName) {
             // This means we have a stream file not a log search.
+            // TODO: Should there be any telemetry output here?
```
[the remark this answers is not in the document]
justinmk3: the `// TODO: Should there be any telemetry output here?` comment should be removed since we have resolved it.

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
```diff
@@ -67,12 +68,24 @@ export class LogStreamDocumentProvider implements vscode.TextDocumentContentProv
             logGroupInfo.streamName = streamID
             const initialStreamData = getInitialLogData(logGroupInfo, parameters, getLogEventsFromUriComponents)
             const streamUri = createURIFromArgs(logGroupInfo, parameters)
+
             await this.registry.registerLog(streamUri, initialStreamData)
-            // Set the document language
             const doc = await vscode.workspace.openTextDocument(streamUri)
             vscode.languages.setTextDocumentLanguage(doc, 'log')
+
+            telemetry.recordCloudwatchlogsOpen({
```
justinmk3, line 76: Is this redundant with the `recordCloudwatchlogsOpen` metric that will be emitted when the logstream is loaded? Can the logstream viewer set `source=goto` based on the URI (maybe we could add a `?source=goto` querystring or something like that?

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
```diff
@@ -67,12 +68,24 @@ export class LogStreamDocumentProvider implements vscode.TextDocumentContentProv
...
+                result: 'Succeeded',
+                cloudWatchResourceType: 'logStream',
+                source: 'GoTo',
+            })
             return new vscode.Location(streamUri, new vscode.Position(0, 0))
         } catch (err) {
+            telemetry.recordCloudwatchlogsOpen({
```
justinmk3, line 83: this one should stay

src/cloudWatchLogs/document/logStreamDocumentProvider.ts
Hweinstock: The `recordCloudwatchlogsOpen` code is only run in the `viewLogStream` function which isn't called in the jumpTo code. There is one issue where `recordCloudwatchlogsOpen` is recorded twice (once on the hover of the text, and once on the loading of the new file). It seems like both actions run the `provideDefinition` code, and I haven't found a way to distinguish them.

justinmk3 (approved):

PR #2813, opened 2022-08-08
cwl: onDidCloseTextDocument() should be called exactly once for the lifetime of the Toolkit

Hweinstock:
## Problem
`onDidCloseTextDocument` was being called anytime registry was initiated and was not being made a disposable.
## Solution
Move the event handlers out of the registry class and into subscriptions so that they are disposable and avoid any potential memory issues. 

## License

### DIFF
--- a/src/cloudWatchLogs/activation.ts
+++ b/src/cloudWatchLogs/activation.ts
@@ -42,6 +42,7 @@ export async function activate(context: vscode.ExtensionContext, configuration: 
                 registry.disposeRegistryData(doc.uri)
             }
         }),
+        // Do highlight on text-changed event because we don't control when text is populated by vscode, also for "load more" codelens.
         vscode.workspace.onDidChangeTextDocument((event: vscode.TextDocumentChangeEvent) => {
             const eventUri = event.document.uri
             if (registry.hasLog(eventUri) && !isLogStreamUri(eventUri)) {
--- a/src/cloudWatchLogs/activation.ts
+++ b/src/cloudWatchLogs/activation.ts
@@ -6,7 +6,7 @@
 import * as vscode from 'vscode'
 import { CLOUDWATCH_LOGS_SCHEME } from '../shared/constants'
 import { Settings } from '../shared/settings'
-import { CloudWatchLogsSettings } from './cloudWatchLogsUtils'
+import { CloudWatchLogsSettings, isLogStreamUri } from './cloudWatchLogsUtils'
 import { addLogEvents } from './commands/addLogEvents'
 import { copyLogStreamName } from './commands/copyLogStreamName'
 import { saveCurrentLogStreamContent } from './commands/saveCurrentLogStreamContent'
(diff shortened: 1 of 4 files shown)


src/cloudWatchLogs/activation.ts
```diff
@@ -39,11 +39,21 @@ export async function activate(context: vscode.ExtensionContext, configuration:
...
                 registry.deregisterLog(doc.uri)
             }
         })
     )
 
+    context.subscriptions.push(
+        vscode.workspace.onDidChangeTextDocument((event: vscode.TextDocumentChangeEvent) => {
+            const eventUri = event.document.uri
+            if (registry.hasLog(eventUri) && !isLogStreamUri(eventUri)) {
+                registry.highlightDocument(eventUri)
```
justinmk3, line 52: onDidChangeTextDocument is potentially very frequent. Doesn't matter here in practice because the doc is readonly, but it's a hidden spec, which is debt (a potential future bug). What's the reason for calling highlightDocument() on this event instead of just doing it explicitly after we populate the text?

src/cloudWatchLogs/activation.ts
```diff
@@ -39,11 +39,21 @@ export async function activate(context: vscode.ExtensionContext, configuration:
...
         vscode.workspace.onDidCloseTextDocument(doc => {
             if (doc.isClosed && doc.uri.scheme === CLOUDWATCH_LOGS_SCHEME) {
+                registry.disposeRegistryData(doc.uri)
                 registry.deregisterLog(doc.uri)
             }
         })
     )
 
+    context.subscriptions.push(
+        vscode.workspace.onDidChangeTextDocument((event: vscode.TextDocumentChangeEvent) => {
```
justinmk3, line 49: nit: push() is variadic (takes unlimited args) so we can save a few lines by putting all of this disposables into one push() call.

src/cloudWatchLogs/activation.ts
```diff
@@ -39,11 +39,21 @@ export async function activate(context: vscode.ExtensionContext, configuration:
     context.subscriptions.push(
         vscode.workspace.onDidCloseTextDocument(doc => {
             if (doc.isClosed && doc.uri.scheme === CLOUDWATCH_LOGS_SCHEME) {
+                registry.disposeRegistryData(doc.uri)
                 registry.deregisterLog(doc.uri)
```
justinmk3, line 43: off-topic (future PR?): should deregisterLog be renamed? Does "log" mean "logstream" or "logdata" or "loggroup" ...?

Do we even need `deregisterLog()` anymore? It is only called where `disposeRegistryData()` is called. So its job should be merged into `disposeRegistryData()`.

justinmk3 (approved):

src/cloudWatchLogs/activation.ts
Hweinstock: I think you're right. These two function have a very similar function and can realistically always be called together. I went ahead and combined them into `disposeRegistryData()`. 

This seemed like an opportunity to reduce code volume which I assume is always a good goal to have in mind?

src/cloudWatchLogs/activation.ts
Hweinstock: Before having this be an event handler I tried having `highlightDocument` in both the `updateLog` function and the `documentProvider.provideTextDocumentContent` function. It seems that both fail to highlight newly loaded data from the `load more` codelense before it is rendered to the screen. I couldn't figure out where else to have to other than on a document change so that this behavior worked as expected.

JadenSimon: > just doing it explicitly after we populate the text?

Because we don't populate the text document. VS Code does.

justinmk3: Ok. That kind of hard-gained knowledge is worth a 1-line comment here. Something like

    // Do highlight on text-changed because we don't control when text is populated by vscode, also for "load more" codelens.

PR #2815, opened 2022-08-08
cwl: cancel view log stream + loading message. 

Hweinstock:
## Problem
`viewLogStream` could also take a long time to load and we already have to the code to improve this. 
## Solution
Implement the same logic as we have in `searchLogGroup`. 
TODO: Generalize this logic as a function `getResponseOrCancel`. Note: I tried to do this and it ended up much nastier than keeping them separate so didn't include in this PR. 

## License


JadenSimon:
> TODO: Generalize this logic as a function getResponseOrCancel. Note: I tried to do this and it ended up much nastier than keeping them separate so didn't include in this PR.

This is getting into generic programming which takes a lot of practice :)
Better to keep things simple unless the generalization is obvious and straightforward.

JadenSimon (approved):

justinmk3 (approved):
very cool!

PR #2816, opened 2022-08-09
cwl: Update custom date range prompter

Hweinstock:
## Problem
The custom time range box did not have the same buttons/functionality as all other Wizard prompts. Additionally, there was an error when the custom time range box was cancelled and then reopened. 
## Solution
We can add the common buttons to custom time range box and not reuse a previous prompter to avoid the error. 

image

## License

### DIFF
--- a/src/cloudWatchLogs/timeFilterSubmenu.ts
+++ b/src/cloudWatchLogs/timeFilterSubmenu.ts
@@ -95,9 +95,10 @@ export class TimeFilterSubmenu extends Prompter<TimeFilterResponse> {
         while (true) {
             switch (this.currentState) {
                 case 'recent-range': {
-                    this.steps && this.defaultPrompter.setSteps(this.steps[0], this.steps[1])
+                    const prompter = this.createMenuPrompter()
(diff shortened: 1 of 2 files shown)


src/cloudWatchLogs/timeFilterSubmenu.ts
```diff
@@ -112,7 +112,7 @@ export class TimeFilterSubmenu extends Prompter<TimeFilterResponse> {
                     break
                 }
                 case 'custom-range': {
-                    const resp = await this.customPrompter.prompt()
+                    const resp = await this.createDateBox().prompt()
```
JadenSimon, line 115: Looking at the rest of this code, line 122 is another potential source of bugs
The whole point of a state machine is to keep the states **isolated**

Each state should not need to be doing any sort of setup for the next. Let `createMenuPrompter` be called when the relevant prompt is used, not at the state transition. 

Looks like we don't need `defaultPrompter` either. The original 'RegionSubmenu' only exposed the active UI component as a mechanism for testing the control flow. But we're already exposing the factory functions for creating each prompt as public methods.

JadenSimon (approved):

PR #2823, opened 2022-08-10
performance: optimize `Developer Tools` view

JadenSimon:
## Problem
View is re-rendered when a node becomes "invisible".
This is a consequence of visibility being functionally different from a node's children changing.

## Solution
Externally refresh a specific node's child **without** changing the children array itself. 
This can also be implemented by increasing the scope of `onDidChangeChildren` and exposing it on the "root" node although I'll save that for another time.

## License

### DIFF
--- a/src/test/shared/treeview/resourceTreeDataProvider.test.ts
+++ b/src/test/shared/treeview/resourceTreeDataProvider.test.ts
@@ -50,7 +50,7 @@ function getLabels(nodes: TreeNode[]): Promise<string[]> {
     return Promise.all(nodes.map(async n => (await n.getTreeItem()).label ?? 'not set'))
 }
 
-function intoTestNode(node: TreeNode): TestNode | never {
+function toTestNode(node: TreeNode): TestNode | never {
     if (!(node.resource instanceof TestNode)) {
         throw new TypeError(`Node "${node.id}" had an incorrect resource: ${node.resource}`)
     }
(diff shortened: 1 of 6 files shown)


src/test/shared/treeview/resourceTreeDataProvider.test.ts
```diff
@@ -100,6 +100,20 @@ describe('ResourceTreeDataProvider', function () {
...
+        const [node0, node1] = await provider.getChildren()
+        const item0 = await provider.getTreeItem(node0)
+        const item1 = await provider.getTreeItem(node1)
+        provider.refresh(node1)
+
+        const newItem0 = await provider.getTreeItem(node0)
+        const newItem1 = await provider.getTreeItem(node1)
+        assert.strictEqual(newItem0, item0)
+        assert.notStrictEqual(newItem1, item1)
+        assert.deepStrictEqual(newItem1, item1)
```
justinmk3, line 114: The same args are "not strict equal", but _are_ "deep strict equal"? This may be worth a 1-line explaining the subtlety.

JadenSimon: They have the same data but are not the same references.
It's the most straightforward way to explicitly test caching behavior albeit somewhat abstract.

justinmk3 (approved):
LGTM, but if there is risk of regression, may want to wait to merge this until Friday.

justinmk3 (approved):

PR #2824, opened 2022-08-10
feat(codewhisperer): handle extra brackets onAcceptance of recommendation

zixlin7:
## Problem
When customer invoke CodeWhisperer, they might already have matching brackets in the right context but our current implementation is not taking it into account.

## Solution
When customer accept a recommendation, we match the brackets that are in the accepted recommendation with user's right context, and remove the extra brackets if there is any:

in VSC:

https://user-images.githubusercontent.com/60411978/184037665-0c68da1a-d13c-4947-adb5-929435664a8e.mov

in C9:

https://user-images.githubusercontent.com/60411978/184037688-1615cad5-06f9-45ee-8dfc-4207d7634423.mov

with whitespace:

https://user-images.githubusercontent.com/60411978/184675900-ea83c202-53a5-4660-9831-2a57a234b52d.mov

## License

### DIFF
--- a/src/codewhisperer/util/closingBracketUtil.ts
+++ b/src/codewhisperer/util/closingBracketUtil.ts
@@ -4,71 +4,123 @@
  */
 
 import * as vscode from 'vscode'
-import { TextEdit, workspace, WorkspaceEdit } from 'vscode'
+import { workspace, WorkspaceEdit } from 'vscode'
 import { isCloud9 } from '../../shared/extensionUtilities'
-import * as telemetry from '../../shared/telemetry/telemetry'
+import { CodeWhispererConstants } from '../models/constants'
 
-export async function handleAutoClosingBrackets(
-    triggerType: telemetry.CodewhispererTriggerType,
+interface bracketMapType {
+    [k: string]: string
+}
+
+const bracketMap: bracketMapType = {
+    ')': '(',
+    ']': '[',
+    '}': '{',
+}
+
+export const calculateBracketsLevel = (code: string) => {
+    const bracketCounts: Map<string, number> = new Map([
+        ['(', 0],
+        ['[', 0],
+        ['{', 0],
+    ])
+    const bracketsIndexLevel = []
+
+    for (let i = 0; i < code.length; i++) {
+        const char = code[i]
+        if (bracketCounts.has(char)) {
+            const count = bracketCounts.get(char) || 0
+            const newCount = count + 1
+            bracketCounts.set(char, newCount)
+            bracketsIndexLevel.push({
+                char,
+                count: newCount,
+                idx: i,
+            })
+        } else if (char in bracketMap) {
+            const correspondingBracket = bracketMap[char as keyof bracketMapType]
+            const count = bracketCounts.get(correspondingBracket) || 0
+            const newCount = count - 1
+            bracketCounts.set(bracketMap[char], newCount)
+            bracketsIndexLevel.push({
+                char: correspondingBracket,
+                count: newCount,
+                idx: i,
+            })
+        }
+    }
+    return bracketsIndexLevel
+}
+
+export const getBracketsToRemove = (recommendation: string, rightContext: string) => {
+    const recommendationBrackets = calculateBracketsLevel(recommendation)
+    const rightContextBrackets = calculateBracketsLevel(rightContext)
+    let i = 0
+    let j = 0
+    const toBeRemoved = []
+
+    while (i < recommendationBrackets.length && j < rightContextBrackets.length) {
+        const { char: char1, count: level1 } = recommendationBrackets[i]
+        const { char: char2, count: level2, idx: idx2 } = rightContextBrackets[j]
+        if (char1 !== char2 || level1 !== level2) {
+            i++
+            continue
+        }
+        toBeRemoved.push(idx2)
+        i++
+        j++
+    }
+    return toBeRemoved
+}
+
+export const removeBracketsFromRightContext = async (
     editor: vscode.TextEditor,
-    recommendation: string,
-    line: number,
-    specialChar: string
-) {
-    const openingBrackets = new Map<string, string>()
-    openingBrackets.set('{', '}')
-    openingBrackets.set('(', ')')
-    openingBrackets.set('[', ']')
-    const closingBracket = openingBrackets.get(specialChar)
-    if (
-        triggerType === 'AutoTrigger' &&
-        closingBracket !== undefined &&
-        hasExtraClosingBracket(recommendation, specialChar, closingBracket)
-    ) {
-        let curLine = line
-        let textLine = editor.document.lineAt(curLine).text
-        while (curLine > 0 && !textLine.includes(closingBracket)) {
-            curLine--
-            textLine = editor.document.lineAt(curLine).text
-        }
-        if (curLine >= 0) {
-            const pos = textLine.lastIndexOf(closingBracket)
-            const range = new vscode.Range(curLine, pos, curLine, pos + 1)
-            if (isCloud9()) {
-                const edit: TextEdit = {
-                    range: range,
-                    newText: '',
-                }
-                const wEdit = new WorkspaceEdit()
-                wEdit.set(editor.document.uri, [edit])
-                await workspace.applyEdit(wEdit)
-            } else {
-                await editor.edit(
-                    editBuilder => {
-                        editBuilder.delete(range)
-                    },
-                    { undoStopAfter: false, undoStopBefore: false }
-                )
-            }
-        }
+    idxToRemove: number[],
+    endPosition: vscode.Position
+) => {
+    const offset = editor.document.offsetAt(endPosition)
+
+    if (isCloud9()) {
+        const edits = idxToRemove.map(idx => ({
+            range: new vscode.Range(
+                editor.document.positionAt(offset + idx),
+                editor.document.positionAt(offset + idx + 1)
+            ),
+            newText: '',
+        }))
+        const wEdit = new WorkspaceEdit()
+        wEdit.set(editor.document.uri, [...edits])
+        await workspace.applyEdit(wEdit)
+    } else {
+        await editor.edit(
+            editBuilder => {
+                idxToRemove.forEach(idx => {
+                    const range = new vscode.Range(
+                        editor.document.positionAt(offset + idx),
+                        editor.document.positionAt(offset + idx + 1)
+                    )
+                    editBuilder.delete(range)
+                })
+            },
+            { undoStopAfter: false, undoStopBefore: false }
+        )
     }
 }
 
-export function hasExtraClosingBracket(
+export async function handleExtraBrackets(
+    editor: vscode.TextEditor,
     recommendation: string,
-    openingBracket: string,
-    closingBracket: string
-): boolean {
-    let count = 0
-    let pos = recommendation.indexOf(closingBracket)
-    while (pos > 0) {
-        count++
-        pos = recommendation.indexOf(closingBracket, pos + 1)
+    endPosition: vscode.Position
+) {
+    const end = editor.document.offsetAt(endPosition)
+    const rightContext = editor.document.getText(
+        new vscode.Range(
+            editor.document.positionAt(end),
+            editor.document.positionAt(end + CodeWhispererConstants.charactersLimit)
+        )
+    )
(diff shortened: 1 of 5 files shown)


andrewyuq (commented):
From the video it doens't seem to match the right context for the hints but only match the right context when on acceptance, do we also want to do that?

zixlin7:
> From the video it doens't seem to match the right context for the hints but only match the right context when on acceptance, do we also want to do that?

@andrewyuq yes I discussed this with @leigaol , there is no good way for us to do that when showing the recommendation: if we remove the brackets in right context and user rejects/discards the recommendation we will need to add them back. But we cannot capture all the cases of rejecting/discard, for example: user navigate away or close the ws. There will be edge cases might confuse the customers even more. So we decided to keep it only to when user accept the recommendation we fix the brackets.

src/codewhisperer/util/closingBracketUtil.ts
```diff
@@ -4,71 +4,123 @@
...
+
+export const calculateBracketsLevel = (code: string) => {
+    const bracketCounts: Map<string, number> = new Map([
+        ['(', 0],
+        ['[', 0],
+        ['{', 0],
+    ])
+    const bracketsIndexLevel = []
+
+    for (let i = 0; i < code.length; i++) {
```
leigaol, line 29: This will go through at most 10k characters.

andrewyuq:
> > From the video it doens't seem to match the right context for the hints but only match the right context when on acceptance, do we also want to do that?
> 
> @andrewyuq yes I discussed this with @leigaol , there is no good way for us to do that when showing the recommendation: if we remove the brackets in right context and user rejects/discards the recommendation we will need to add them back. But we cannot capture all the cases of rejecting/discard, for example: user navigate away or close the ws. There will be edge cases and might confuse the customers even more. So we decided to keep it only to when user accept the recommendation that we fix the brackets.

We don't remove the right contexts if it's just showing hints, instead we break the recommendations into different chunks and insert those after different matching characteres for the hints.

andrewyuq:
> > > From the video it doens't seem to match the right context for the hints but only match the right context when on acceptance, do we also want to do that?
> > 
> > 
> > @andrewyuq yes I discussed this with @leigaol , there is no good way for us to do that when showing the recommendation: if we remove the brackets in right context and user rejects/discards the recommendation we will need to add them back. But we cannot capture all the cases of rejecting/discard, for example: user navigate away or close the ws. There will be edge cases and might confuse the customers even more. So we decided to keep it only to when user accept the recommendation that we fix the brackets.
> 
> We don't remove the right contexts if it's just showing hints, instead we break the recommendations into different chunks and insert those after different matching characteres for the hints.

Let's say if recommendation is `a, b): return a`, and right context is a single `)` then we break recommendation into two parts `a, b` and `: return a` and insert these hints into different offsets. You want to keep track of the original copy that's unmodified for the final insertion, and also keep track of these chunks both in the beginning of receiving recommendations, and also when they are typing, these chunks need be updated accordingly and the offsets to insert them also need to be updated according to the current document.

zixlin7:
> > > > From the video it doens't seem to match the right context for the hints but only match the right context when on acceptance, do we also want to do that?
> > > 
> > > 
> > > @andrewyuq yes I discussed this with @leigaol , there is no good way for us to do that when showing the recommendation: if we remove the brackets in right context and user rejects/discards the recommendation we will need to add them back. But we cannot capture all the cases of rejecting/discard, for example: user navigate away or close the ws. There will be edge cases and might confuse the customers even more. So we decided to keep it only to when user accept the recommendation that we fix the brackets.
> > 
> > 
> > We don't remove the right contexts if it's just showing hints, instead we break the recommendations into different chunks and insert those after different matching characteres for the hints.
> 
> Let's say if recommendation is `a, b): return a`, and right context is a single `)` then we break recommendation into two parts `a, b` and `: return a` and insert these hints into different offsets. You want to keep track of the original copy that's unmodified for the final insertion, and also keep track of these chunks both in the beginning of receiving recommendations, and also when users are typing, these chunks need be updated accordingly and the offsets to insert them also need to be updated according to the current document.

@andrewyuq yes I understand your approach here, that makes sense. However it will require a major refactor of our show and accept recommendation logic that will involve existing/new inline, reference tracker and type ahead which I'm trying to avoid. I'm wondering if it's a bit overkill for this specific issue, thoughts? @leigaol

andrewyuq:
> > > > > From the video it doens't seem to match the right context for the hints but only match the right context when on acceptance, do we also want to do that?
> > > > 
> > > > 
> > > > @andrewyuq yes I discussed this with @leigaol , there is no good way for us to do that when showing the recommendation: if we remove the brackets in right context and user rejects/discards the recommendation we will need to add them back. But we cannot capture all the cases of rejecting/discard, for example: user navigate away or close the ws. There will be edge cases and might confuse the customers even more. So we decided to keep it only to when user accept the recommendation that we fix the brackets.
> > > 
> > > 
> > > We don't remove the right contexts if it's just showing hints, instead we break the recommendations into different chunks and insert those after different matching characteres for the hints.
> > 
> > 
> > Let's say if recommendation is `a, b): return a`, and right context is a single `)` then we break recommendation into two parts `a, b` and `: return a` and insert these hints into different offsets. You want to keep track of the original copy that's unmodified for the final insertion, and also keep track of these chunks both in the beginning of receiving recommendations, and also when users are typing, these chunks need be updated accordingly and the offsets to insert them also need to be updated according to the current document.
> 
> @andrewyuq yes I understand your approach here, that makes sense. However it will require a major refactor of our show recommendation logic that will involve existing/new inline, reference tracker and type ahead which I'm trying to avoid. I'm wondering if it's a bit overkill for this specific issue, thoughts? @leigaol

I would recommend to have this fix since model can't really do anything about it, so if its there it's always there. I would spend extra time working on this if there's no other higher priority items.

andrewyuq:
> > > > > > From the video it doens't seem to match the right context for the hints but only match the right context when on acceptance, do we also want to do that?
> > > > > 
> > > > > 
> > > > > @andrewyuq yes I discussed this with @leigaol , there is no good way for us to do that when showing the recommendation: if we remove the brackets in right context and user rejects/discards the recommendation we will need to add them back. But we cannot capture all the cases of rejecting/discard, for example: user navigate away or close the ws. There will be edge cases and might confuse the customers even more. So we decided to keep it only to when user accept the recommendation that we fix the brackets.
> > > > 
> > > > 
> > > > We don't remove the right contexts if it's just showing hints, instead we break the recommendations into different chunks and insert those after different matching characteres for the hints.
> > > 
> > > 
> > > Let's say if recommendation is `a, b): return a`, and right context is a single `)` then we break recommendation into two parts `a, b` and `: return a` and insert these hints into different offsets. You want to keep track of the original copy that's unmodified for the final insertion, and also keep track of these chunks both in the beginning of receiving recommendations, and also when users are typing, these chunks need be updated accordingly and the offsets to insert them also need to be updated according to the current document.
> > 
> > 
> > @andrewyuq yes I understand your approach here, that makes sense. However it will require a major refactor of our show recommendation logic that will involve existing/new inline, reference tracker and type ahead which I'm trying to avoid. I'm wondering if it's a bit overkill for this specific issue, thoughts? @leigaol
> 
> I would recommend to have this fix since model can't really do anything about it, so if its there it's always there. I would spend extra time working on this if there's no other higher priority items.

Sync'ed offline, fixing this would require the use of another proposed inline API(not the same as the current released inline-completion API) and will be creating an extra dependency to that, which is not something we want to do at the moment. The current solution will at least fix the right context when the user accepts the recommendation so can be used as a partial solution. Will need to be picked up whenever we have new offical API support.

src/codewhisperer/util/closingBracketUtil.ts
zixlin7: I tested with a 1000+ line right context and didn't see any performance issue, but I can probably lower this number

src/codewhisperer/util/closingBracketUtil.ts
```diff
@@ -4,71 +4,123 @@
...
+    recommendation: string,
+    endPosition: vscode.Position
+) {
+    const end = editor.document.offsetAt(endPosition)
+    const rightContext = editor.document.getText(
+        new vscode.Range(
+            editor.document.positionAt(end),
+            editor.document.positionAt(end + CodeWhispererConstants.charactersLimit)
+        )
+    )
```
andrewyuq, line 121: for getting the right context here, only the right context with a mixed string of consecutive brackets and empty characters(space, \t, \n) are needed. Once we detect a non-bracket or a non-empty character we can stop.

src/codewhisperer/util/closingBracketUtil.ts
andrewyuq: An optimization of this could be taking another string into account. Here we are calculating the bracket levels of the recommendation and the right context independently; One way to achieve this is to match the brackets on the fly when we iterate through recommendations, so the algorithm could be:
```
iterate through each char of recommendation:
    c = the current character in right context that needs to be bracket or space-matched
    if (c is not bracket and is not empty character):
        break
    // we want to match c in recommendation
    current = current char of recommendation
    if (current matches c both character-wise and level-wise):
        c++
        record this index to be processed further
```

andrewyuq: This is non-blocking for this PR, we can get this merged first and have a sim to work on the optimization later.

src/codewhisperer/util/closingBracketUtil.ts
andrewyuq: To this end I don't think we have suppor to match whitespaces in this PR. It doesn't hurt to get that support. And it will be simpler than bracket matching, there will be no `bracketCounts` logic involved, whenever you find a space you just add it to the returned list.

src/codewhisperer/util/closingBracketUtil.ts
zixlin7: updated to return early when processing right context and we encounter a non-bracket, non-white-space character

src/codewhisperer/util/closingBracketUtil.ts
zixlin7: @andrewyuq what do you mean by matching whitespaces? could you pls provide an example? If it's formatting we do format the doc after accepting the recommendation

andrewyuq: yea, so let's say the right context is `) {`. If we just match brackets, the algorithm will stop at the first closing paramthese, if we can also match spaces it will match the the whole right context.

zixlin7: @andrewyuq got it, yes the revised version does that, as you said we only stop if we encounter a non-bracket and non-white space character

andrewyuq: I get your point, I mean that we also need to **match** the whitespaces, let's say the recommendaiton is `){` and right context is `) {`. We can only match to the first `)` but not further, this means we also need to add the tripet entry for the whitespace character in `bracketsIndexLevel`(and probably should rename it as symbolsIndexLevel since it no longer contains brackets only). If we don't do this, the space will be ignored and they will be matched to the end by mistake.

zixlin7: @andrewyuq I don't think we need to do that. if the recommendation is `){` and right context `) {` we will match them correctly. please see the description I will add a demo video for this case.

andrewyuq: >if the recommendation is `){` and right context `) {`
The thing is we actually don't want the recommendation to match the right context in this case. Notice that the recommendation does not have a whitespace and in the right context there's a whitespace. So the expected behavior should be that the matching should end after the first character `)`.

andrewyuq: Discussed offline, 1) The extra whitespace matching could cover more cases 2) whitespaces changing in Java and Javascript would change the syntatically correctness of the code, and in vscode we always do a pretty-print sort of formatting(vscode built-in) at the end, which should adapt user's local settings on the tab indent size in python. So there should be no problems matching the whitespaces in the insertion logic on acceptance.

zixlin7: synced offline, we will match the whitespace in VSC because we are only modifying the right context after user accepting the recommendation and it will help to match more cases

PR #2831, opened 2022-08-15
testing: add `testCommand` util

JadenSimon:
## Problem
Not obvious how to test commands without splitting up the logic. 
This is only applicable when the command itself requires fakes.

## Solution
Add util function to create testable versions of commands. It's not perfect because VS Code commands are meant to be global and unique whereas testing often requires test versions of things. The implementation is also fragile but acceptable for simple tests.

## License

### DIFF
--- a/docs/CODE_GUIDELINES.md
+++ b/docs/CODE_GUIDELINES.md
@@ -143,6 +143,9 @@ that is a net cost.
 
     -   [Testing Refresh Button](https://github.com/aws/aws-toolkit-vscode/blob/b34c8f7650c862c388992781844695b014b5d974/src/test/shared/ui/prompters/rolePrompter.test.ts#L58-L65)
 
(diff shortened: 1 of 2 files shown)


justinmk3:
mention in https://github.com/aws/aws-toolkit-vscode/blob/master/docs/CODE_GUIDELINES.md#test-guidelines ?

JadenSimon:
> mention in https://github.com/aws/aws-toolkit-vscode/blob/master/docs/CODE_GUIDELINES.md#test-guidelines ?

Yeah I guess I should add a brief mention. I think ideally this utility function shouldn't be used but in order to that we first need to initialize extension sub-modules in a 'test mode'

justinmk3 (approved):

jpinkney-aws (approved):

PR #2832, opened 2022-08-15
cwl: rename logStream -> logData

Hweinstock:
## Problem
The codebase still uses the language of `logStream` to refer to log data that could be from a log group search. This only affects the CloudWatchLogs code. 
## Solution
Change any instance of `logStream` that could refer to a log group search result to a more general `logData` tag. 

## License


src/cloudWatchLogs/registry/logDataRegistry.ts
```diff
@@ -16,9 +16,9 @@ import { findOccurencesOf } from '../../shared/utilities/textDocumentUtilities'
 // TODO: Add debug logging statements
 
 /**
- * Class which contains CRUD operations and persistence for CloudWatch Logs streams.
+ * Class which contains CRUD operations and persistence for CloudWatch Log Data.
```
justinmk3, line 19: somewhere like here is where we need to define what is "log data". Is it any of:

- cw log stream
- cw filter results (which are from log streams in a group)

```suggestion
 * Operations and persistence for CloudWatch Logs Data (a single logstream, or event data from many logstreams)
```

"class which contains" is redundant in the docstring for a class.

src/cloudWatchLogs/commands/saveCurrentLogDataContent.ts
```diff
@@ -72,13 +74,14 @@ export async function saveCurrentLogStreamContent(
...
                 'AWS.cwl.invalidEditor',
-                'Not a Cloudwatch Log stream: {0}',
+                'Not a Cloudwatch Log data source: {0}',
                 vscode.window.activeTextEditor?.document.fileName
             )
         )
     }
 
-    recordCloudwatchlogsDownloadStreamToFile({
+    recordCloudwatchlogsDownload({
```
justinmk3, line 83: did the telemetry definition change? or are we using a different one?

Hweinstock: I changed it to a more general one because I don't believe it was correct in all cases. This code can be run on the result of a log group so we would be emitting the wrong telemetry in that case.

justinmk3: Ok. That sounds like a functional change, which should be in a separate PR or commit. Don't want functional changes in "rename" commits.

Can leave it this time, but good practice to mention it in the commit message then.

Hweinstock: That makes sense. I'll try to keep this PR focused on renaming.

justinmk3 (approved):

src/cloudWatchLogs/commands/saveCurrentLogDataContent.ts
```diff
@@ -9,17 +9,18 @@ const localize = nls.loadMessageBundle()
...
 
-export async function saveCurrentLogStreamContent(
+export async function saveCurrentLogDataContent(
     uri: vscode.Uri | undefined,
-    registry: LogStreamRegistry,
+    registry: LogDataRegistry,
     window = Window.vscode()
 ): Promise<void> {
     let result: Result = 'Succeeded'
+    let resourceType: CloudWatchResourceType = 'logStream' // Default to stream if it fails to find URI
```
JadenSimon, line 23: Kind of misleading on failures but this is a fault with how we do telemetry

JadenSimon (approved):

src/cloudWatchLogs/commands/saveCurrentLogDataContent.ts
Hweinstock: Yeah, I couldn't figure out another way to do this. Is there a way to change/improve the metric itself to avoid this?

PR #2838, opened 2022-08-18
fix(webviews): change default view column for webviews

JadenSimon:
## Problem
All webviews open in a new column even when it may not be the best choice.

## Solution
Open in the active column by default. Webviews that might be apart of an iterative loop are opened in a separate column.

## License

### DIFF
--- a/.changes/next-release/Feature-e0af2ed8-0de5-43b4-9184-e790005eb701.json
+++ b/.changes/next-release/Feature-e0af2ed8-0de5-43b4-9184-e790005eb701.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Most webviews now open in the active editor column by default. APIG, Lambda, SAM, and Execute State Machine webviews retain the previous behavior, opening beside the active editor."
+}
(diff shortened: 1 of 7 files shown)


justinmk3:
> APIG, Lambda, SAM, and Execute State Machine webviews retain the previous behavior, opening beside the active editor.

Is there a reason those should open in a split? I think we should not split unless there is a clear reason. The one case where we have a fairly clear reason is the "Edit SAM Debug Configuration" webview, which opens beside the launch.json file.

JadenSimon:
> Is there a reason those should open in a split?

I consider these webviews to be for iterative development and so side-by-side can be somewhat justified. In reality, the choice of whether or not to open as a split is dependent on where on the webview is opened from. The config editor only makes sense when opened from source code, and it doesn't look like the other webviews have alternate entry points. So I guess we can make it so only the SAM webview is placed beside.

justinmk3:
Yes, it seems opening in a split is not a clear win, it's kind of an arbitrary choice that then must be "fixed" by the user (or more likely tolerated--because dragging tabs is a chore). When there is not a clear win, it's best to stick to defaults.

justinmk3 (approved):

PR #2839, opened 2022-08-18
cwl: Fix bug where empty results are returned with valid nextToken

Hweinstock:
## Problem
Sometimes the results from cwl API return empty, with a `nextToken`. We don't account for this case. 
## Solution
Implement `getPaginatedAwsCallIter` in order to handle this case. 

**Note**: DONT MERGE. Codelenses are currently not working so something is wrong. It passes tests, but doesn't work on realdata. 

## License

### DIFF
--- a/src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
+++ b/src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
@@ -11,12 +11,16 @@ import { INSIGHTS_TIMESTAMP_FORMAT } from '../../../shared/constants'
 import { Settings } from '../../../shared/settings'
 import { CloudWatchLogsSettings, createURIFromArgs } from '../../../cloudWatchLogs/cloudWatchLogsUtils'
 import {
+    addedEvent,
+    backwardToken,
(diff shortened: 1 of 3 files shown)


src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
```diff
@@ -50,6 +75,16 @@ describe('LogDataRegistry', async function () {
...
     })
 
+    describe('updateLog', async function () {
+        it("properply paginates the results with 'head'", async () => {
+            testUpdateLog('head')
+        })
+
+        it("properply paginates the results with 'tail'", async () => {
+            testUpdateLog('tail')
+        })
```
JadenSimon, line 85: These tests may not actually be passing
```suggestion
        it("properply paginates the results with 'head'", async () => {
            await testUpdateLog('head')
        })

        it("properply paginates the results with 'tail'", async () => {
            await testUpdateLog('tail')
        })
```

JadenSimon:
Replaced by https://github.com/aws/aws-toolkit-vscode/pull/3042

PR #2842, opened 2022-08-19
fix(codewhisperer): Mark suggestions as Discard in some cases

leigaol:
## Problem

By the time when plugin gets first paginated suggestion(s), 

1. If the cursor is behind invocation position, it is Discarded.
2. If the typeahead does not match, it is Discarded.
3. If the editor is changed or the invocation editor lost focus, it is Discarded.

## Solution

Mark suggestions as Discard in the above 3 cases. 

https://user-images.githubusercontent.com/97199248/186044661-5dd36cf3-3ac7-452a-92d8-07eb50cd643e.mov

https://user-images.githubusercontent.com/97199248/186044683-6f3a1aeb-1ffc-46ac-a4e7-f757cc1cc5bc.mov

https://user-images.githubusercontent.com/97199248/186044693-f2454786-c7c8-4e4c-9003-00916f5189bb.mov

## License

### DIFF
--- a/src/codewhisperer/service/inlineCompletion.ts
+++ b/src/codewhisperer/service/inlineCompletion.ts
@@ -291,6 +291,7 @@ export class InlineCompletion {
                         })
                 }
             })
+
         vsCodeState.isCodeWhispererEditing = false
     }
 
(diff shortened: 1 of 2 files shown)


andrewyuq (commented):
Are there tests for this case?

leigaol:
> Are there tests for this case?

This is a good call out. Thanks!
It is very hard to create unit tests for those test cases. Those should be covered in integration tests if possible. 
I attached my manual tests in the description for you to review.

src/codewhisperer/service/inlineCompletion.ts
```diff
@@ -378,20 +379,36 @@ export class InlineCompletion {
...
+                // do not show recommendation if cursor is before invocation position or user opened another document
+                // mark suggestions as Discard and cancel paginated request
+                if (
+                    editor.selection.active.isBefore(RecommendationHandler.instance.startPos) ||
+                    editor.document.uri.fsPath !== this.documentUri?.fsPath
+                ) {
+                    RecommendationHandler.instance.cancelPaginatedRequest()
+                    RecommendationHandler.instance.recommendations.forEach((r, i) => {
+                        RecommendationHandler.instance.setSuggestionState(i, 'Discard')
+                    })
```
JadenSimon, line 391: When would you want to set all the recommendation states to discard but not report and clear them?

andrewyuq (approved):
Behavior looks good to me

src/codewhisperer/service/inlineCompletion.ts
leigaol: Hi Jaden, the state is marked as discard in the timer. But the function `getPaginatedRecommendation` is still running and when it sees the request is cancelled it will report telemetry. Thanks!

JadenSimon (approved):

PR #2843, opened 2022-08-19
feat(telemetry): add execution spans

JadenSimon:
## Problem
Large amounts of code duplication for tracking `result`/`duration` with telemetry.

## Solution
Add execution spans that can instrument arbitrary functions. 

This PR is dependent on https://github.com/aws/aws-toolkit-common/pull/354

### TODO
* More tests
* Documentation/examples for contributors
* Rearrange things
* Remove decorator?

## License

### DIFF
--- a/src/apigateway/commands/copyUrl.ts
+++ b/src/apigateway/commands/copyUrl.ts
@@ -16,7 +16,7 @@ import { Stage } from 'aws-sdk/clients/apigateway'
 import { DefaultApiGatewayClient } from '../../shared/clients/apiGatewayClient'
 import { DEFAULT_DNS_SUFFIX, RegionProvider } from '../../shared/regions/regionProvider'
 import { getLogger } from '../../shared/logger'
-import { recordApigatewayCopyUrl } from '../../shared/telemetry/telemetry'
+import { telemetry } from '../../shared/telemetry/spans'
 
 interface StageInvokeUrlQuickPick extends vscode.QuickPickItem {
     // override declaration so this can't be undefined
@@ -48,7 +48,7 @@ export async function copyUrlCommand(
         )
     } catch (e) {
         getLogger().error(`Failed to load stages: %O`, e)
-        recordApigatewayCopyUrl({ result: 'Failed' })
+        telemetry.apigateway_copyUrl.emit({ result: 'Failed' })
         return
     }
 
@@ -61,12 +61,12 @@ export async function copyUrlCommand(
         window.showInformationMessage(
             localize('AWS.apig.copyUrlNoStages', "Failed to copy URL because '{0}' has no stages", node.name)
         )
-        recordApigatewayCopyUrl({ result: 'Failed' })
+        telemetry.apigateway_copyUrl.emit({ result: 'Failed' })
         return
     } else if (quickPickItems.length === 1) {
         const url = quickPickItems[0].detail
         await copyToClipboard(url, 'URL')
-        recordApigatewayCopyUrl({ result: 'Succeeded' })
+        telemetry.apigateway_copyUrl.emit({ result: 'Succeeded' })
         return
     }
 
@@ -89,13 +89,13 @@ export async function copyUrlCommand(
     const pickerResponse = picker.verifySinglePickerOutput<StageInvokeUrlQuickPick>(choices)
 
     if (!pickerResponse) {
-        recordApigatewayCopyUrl({ result: 'Cancelled' })
+        telemetry.apigateway_copyUrl.emit({ result: 'Cancelled' })
         return
     }
 
     const url = pickerResponse.detail
     await copyToClipboard(url, 'URL')
-    recordApigatewayCopyUrl({ result: 'Succeeded' })
+    telemetry.apigateway_copyUrl.emit({ result: 'Succeeded' })
 }
 
 export function buildDefaultApiInvokeUrl(apiId: string, region: string, dnsSuffix: string, stage: string): string {
--- a/src/awsexplorer/commands/copyArn.ts
+++ b/src/awsexplorer/commands/copyArn.ts
@@ -5,7 +5,6 @@
 
 import * as nls from 'vscode-nls'
 import { showLogOutputChannel } from '../../shared/logger'
-import * as telemetry from '../../shared/telemetry/telemetry'
 const localize = nls.loadMessageBundle()
 
 import { AWSResourceNode } from '../../shared/treeview/nodes/awsResourceNode'
@@ -26,7 +25,6 @@ export async function copyArnCommand(
 ): Promise<void> {
     try {
         copyToClipboard(node.arn, 'ARN', window, env)
-        recordCopyArn({ result: 'Succeeded' })
     } catch (e) {
         const logsItem = localize('AWS.generic.message.viewLogs', 'View Logs...')
         window
@@ -43,9 +41,5 @@ export async function copyArnCommand(
                     showLogOutputChannel()
                 }
             })
-        recordCopyArn({ result: 'Failed' })
     }
 }
-
-// TODO add telemetry for copy arn
-function recordCopyArn({ result }: { result: telemetry.Result }): void {}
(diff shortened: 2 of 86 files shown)


src/apigateway/commands/copyUrl.ts
```diff
@@ -48,7 +48,7 @@ export async function copyUrlCommand(
         )
     } catch (e) {
         getLogger().error(`Failed to load stages: %O`, e)
-        recordApigatewayCopyUrl({ result: 'Failed' })
+        telemetry.apigateway_copyUrl.emit({ result: 'Failed' })
```
JadenSimon, line 51: The `emit` method is equivalent to the old behavior. I chose the verb `emit` because `record` now refers to placing data into the span but not necessarily sending it.

src/awsexplorer/commands/copyArn.ts
```diff
@@ -43,9 +41,5 @@ export async function copyArnCommand(
                     showLogOutputChannel()
                 }
             })
-        recordCopyArn({ result: 'Failed' })
     }
 }
-
-// TODO add telemetry for copy arn
-function recordCopyArn({ result }: { result: telemetry.Result }): void {}
```
JadenSimon, line 51: :/

src/codewhisperer/tracker/codewhispererTracker.ts
```diff
@@ -5,10 +5,10 @@
 
 import * as vscode from 'vscode'
 import globals from '../../shared/extensionGlobals'
-import * as telemetry from '../../shared/telemetry/telemetry'
 import { distance } from 'fastest-levenshtein'
 import { AcceptedSuggestionEntry } from '../models/model'
 import { getLogger } from '../../shared/logger/logger'
+import { telemetry } from '../../shared/telemetry/spans'
```
JadenSimon, line 11: Might be able to merge the telemetry object with the types to keep the `import * as telemetry` pattern. Potentially use a namespace?

src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -164,6 +164,75 @@
         {
             "name": "stepfunctions_previewstatemachine",
             "description": ""
+        },
```
JadenSimon, line 167: Our telemetry has two override files??
I'm deleting the one in the common repo.

src/shared/telemetry/spans.ts
```diff
@@ -0,0 +1,254 @@
...
+    return { ...state }
+}
+
+export class TelemetrySpan {
+    #startTime: Date | undefined = undefined
+    private readonly state: Partial<MetricBase> = {}
+    private readonly definition = definitions[this.name] ?? {
+        unit: 'None',
+        passive: true,
+        requiredMetadata: [],
```
justinmk3, line 97: nit: could just name this `required`.

side-note: "metadata" has seemed like a misleading name because it's not really meta--it's actually the data.

src/eventSchemas/vue/searchSchemas.ts
```diff
@@ -138,8 +139,7 @@ export async function createSearchSchemasWebView(context: ExtContext, node: Regi
         const error = err as Error
         logger.error('Error searching schemas: %O', error)
     } finally {
-        // TODO make this telemetry actually record failures
-        recordSchemasSearch({ result: webviewResult })
+        telemetry.schemas_search.emit({ result: webviewResult })
```
justinmk3, line 142: the TODO was removed; is the failure recorded in the span now?

src/codewhisperer/tracker/codewhispererTracker.ts
justinmk3: would be nice ergonomics, but perhaps not worth it if it requires a namespace

src/awsexplorer/commands/copyArn.ts
justinmk3: haha. we get this with vscode_executeCommand at least now? But we should do a proper one.

src/shared/telemetry/spans.ts
```diff
@@ -0,0 +1,254 @@
...
+    MetricBase,
+    MetricDefinition,
+    MetricName,
+    MetricShapes,
+    TelemetryBase,
+} from './telemetry'
+import { getTelemetryReason, getTelemetryResult } from '../errors'
+
+const AsyncLocalStorage: typeof AsyncLocalStorageClass =
+    require('async_hooks').AsyncLocalStorage ??
```
justinmk3, line 22: techdebt.test.ts ?

src/shared/telemetry/spans.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    public record(data: Partial<MetricBase>): this {
+        Object.assign(this.state, data)
+        return this
+    }
+
+    public emit(data?: MetricBase): void {
+        const state = Object.assign(getValidatedState(this.state, this.definition), data)
+        const metadata = Object.entries(state)
+            .filter(([k, v]) => v !== undefined && !['passive', 'value'].includes(k))
```
justinmk3, line 114: should the excluded fields be named in a private `excluded` field or something?

tsconfig.json
```diff
@@ -17,7 +17,8 @@
         "noUnusedLocals": true,
         "lib": ["dom", "es6", "esnext.asynciterable"],
         "skipLibCheck": true,
-        "jsx": "preserve"
+        "jsx": "preserve",
+        "experimentalDecorators": true
```
justinmk3, line 21: 👀

justinmk3 (approved):

src/apigateway/commands/copyUrl.ts
justinmk3: Does `record`  need to ever be explicitly called for the metric to be posted? What happens if `emit()` is never followed bye `record()` ?

src/shared/telemetry/spans.ts
JadenSimon: Originally the data was 'value' but that's hardly ever used. So yeah it should be called just 'data'. Less confusing too.

src/awsexplorer/commands/copyArn.ts
JadenSimon: Yup it's covered by commands now. Need to remove all those try/catches though to get the result/reason.

src/eventSchemas/vue/searchSchemas.ts
JadenSimon: The comment was stale as apparently I added the correct result a long time ago
To get the failure in the span we need to start using `run` everywhere

tsconfig.json
JadenSimon: Yeah I need to remove some of the test/experimental stuff from this PR still
Can always come back to decorators later (if ever)

src/apigateway/commands/copyUrl.ts
JadenSimon: > Does record need to ever be explicitly called for the metric to be posted? 

No. The closest analogy here is comparing `record` to a log statement. `getLogger().error` doesn't get to choose exactly when (or what) is written to the log file, it only outputs some basic information that may eventually be written to the logs. `emit` is kind of like a 'flush' in that it immediately sends the contents of the span to telemetry.

But just like _most_ developers don't care when the logs flush, _most_ developers won't need to worry about `emit`. 

> What happens if emit() is never followed bye record() ?

If the span was creating outside of `run`, nothing would be sent. A span would exist until the execution frame is dropped, then it gets garbage collected. There's several ways to handle this scenario but I haven't decided about what makes the most sense.

scripts/build/package.ts
```diff
@@ -97,7 +97,7 @@ function main() {
             }
         }
 
-        child_process.execSync(`vsce package`)
+        child_process.execSync(`vsce package`, { stdio: 'inherit' })
```
JadenSimon, line 100: `npm run package` now shows the output :)

justinmk3 (approved):

PR #2845, opened 2022-08-22
fix(codewhisperer): update code coverage tracker to keep tracker of removed characters

leigaol:
## Problem
Current code percentage written metric in CodeCoverageTracker does not count tokens removed by user, which makes this metric inaccurate.

## Solution

Similar to what was implemented in AWS Toolkit for JetBrains by @Will-ShaoHua , we make best effort to:

1. Estimate total user token by counting user inputs and deletions
2. Estimate total accepted tokens from CodeWhisperer suggestion by counting unmodified suggestions

Total user token is updated when there is a TextDocumentChange Event,
1. If  the event is not from human input (i.e. from plugins, formatters), this event is ignored.
2. If the event is when user copy large chunk of code, this event is ignored.

We only count the user manual keyboard input, intelliSense code acceptance and small chunk code copy as inserted total user token. For deletion events, we take the range from them and reduce the total tokens until total tokens reaches 0.

Total accepted tokens is updated when user accept a code whisperer suggestion. At that time, the formatted suggestion is counted as accepted tokens. At the time of emitting telemetry data, a edit distance comparision will be done over the remaining code in accepted suggestion range and the accepted suggesetion code to estimate how many characters user has deleted in the accepted suggestion.  This edit distance update will also happen when user changes active editor ( go to another file in VS Code).

## License

### DIFF
--- a/src/codewhisperer/commands/onAcceptance.ts
+++ b/src/codewhisperer/commands/onAcceptance.ts
@@ -94,8 +94,11 @@ export async function onAcceptance(acceptanceEntry: OnRecommendationAcceptanceEn
             completionType: acceptanceEntry.completionType,
             language: languageContext.language,
         })
-        CodeWhispererCodeCoverageTracker.getTracker(languageContext.language, globalStorage).setAcceptedTokens(
-            acceptanceEntry.recommendation
+        const codeRangeAfterFormat = new vscode.Range(start, acceptanceEntry.editor.selection.active)
+        CodeWhispererCodeCoverageTracker.getTracker(languageContext.language, globalStorage)?.countAcceptedTokens(
+            codeRangeAfterFormat,
+            acceptanceEntry.editor.document.getText(codeRangeAfterFormat),
+            acceptanceEntry.editor.document.fileName
         )
     }
 
--- a/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
+++ b/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
@@ -103,37 +103,36 @@ export class CodeWhispererCodeCoverageTracker {
     }
 
     private tryStartTimer() {
-        if (this._timer === undefined) {
-            const currentDate = new globals.clock.Date()
-            this._startTime = currentDate.getTime()
-            this._timer = setTimeout(() => {
-                try {
-                    const currentTime = new globals.clock.Date().getTime()
-                    const delay: number = CodeWhispererConstants.defaultCheckPeriodMillis
-                    const diffTime: number = this._startTime + delay
-                    if (diffTime <= currentTime) {
-                        let totalTokens = 0
-                        for (const filename in this._totalTokens) {
-                            totalTokens += this._totalTokens[filename]
-                        }
-                        if (totalTokens > 0) {
-                            this.flush()
-                        } else {
-                            getLogger().debug(
-                                `CodeWhispererCodeCoverageTracker: skipped telemetry due to empty tokens array`
-                            )
-                        }
+        if (this._timer !== undefined) return
+        const currentDate = new globals.clock.Date()
+        this._startTime = currentDate.getTime()
+        this._timer = setTimeout(() => {
+            try {
+                const currentTime = new globals.clock.Date().getTime()
+                const delay: number = CodeWhispererConstants.defaultCheckPeriodMillis
+                const diffTime: number = this._startTime + delay
+                if (diffTime <= currentTime) {
+                    let totalTokens = 0
+                    for (const filename in this._totalTokens) {
+                        totalTokens += this._totalTokens[filename]
                     }
-                } catch (e) {
-                    getLogger().verbose(`Exception Thrown from CodeWhispererCodeCoverageTracker: ${e}`)
-                } finally {
-                    this._totalTokens = {}
-                    this._acceptedTokens = {}
-                    this._startTime = 0
-                    this.closeTimer()
+                    if (totalTokens > 0) {
+                        this.flush()
+                    } else {
+                        getLogger().debug(
+                            `CodeWhispererCodeCoverageTracker: skipped telemetry due to empty tokens array`
+                        )
+                    }
                 }
-            }, CodeWhispererConstants.defaultCheckPeriodMillis)
-        }
+            } catch (e) {
+                getLogger().verbose(`Exception Thrown from CodeWhispererCodeCoverageTracker: ${e}`)
+            } finally {
+                this._totalTokens = {}
+                this._acceptedTokens = {}
+                this._startTime = 0
+                this.closeTimer()
+            }
+        }, CodeWhispererConstants.defaultCheckPeriodMillis)
     }
 
     private closeTimer() {
--- a/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
+++ b/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
@@ -8,120 +8,186 @@ import * as telemetry from '../../shared/telemetry/telemetry'
 import { getLogger } from '../../shared/logger/logger'
 import { CodeWhispererConstants } from '../models/constants'
 import globals from '../../shared/extensionGlobals'
+import { vsCodeState } from '../models/model'
+import { distance } from 'fastest-levenshtein'
+
+interface CodeWhispererToken {
+    range: vscode.Range
+    text: string
+    accepted: number
+}
+
 /**
- * This singleton class is mainly used for calculating the percentage of user modification.
- * The current calculation method is (Levenshtein edit distance / acceptedSuggestion.length).
+ * This singleton class is mainly used for calculating the code written by codeWhisperer
  */
 export class CodeWhispererCodeCoverageTracker {
-    private _acceptedTokens: string[]
-    private _totalTokens: string[]
+    private _acceptedTokens: { [key: string]: CodeWhispererToken[] }
+    private _totalTokens: { [key: string]: number }
     private _timer?: NodeJS.Timer
     private _startTime: number
     private _language: telemetry.CodewhispererLanguage
 
     private constructor(language: telemetry.CodewhispererLanguage, private readonly _globals: vscode.Memento) {
-        this._acceptedTokens = []
-        this._totalTokens = []
+        this._acceptedTokens = {}
+        this._totalTokens = {}
         this._startTime = 0
         this._language = language
     }
 
-    public setAcceptedTokens(recommendation: string) {
+    public get acceptedTokens(): { [key: string]: CodeWhispererToken[] } {
+        return this._acceptedTokens
+    }
+    public get totalTokens(): { [key: string]: number } {
+        return this._totalTokens
+    }
+
+    public countAcceptedTokens(range: vscode.Range, text: string, filename: string) {
         const terms = this._globals.get<boolean>(CodeWhispererConstants.termsAcceptedKey) || false
         if (!terms) return
-
-        // generate accepted recoomendation token and stored in collection
-        this._acceptedTokens.push(...recommendation)
-        this._totalTokens.push(...recommendation)
-    }
-
-    public get AcceptedTokensLength(): number {
-        return this._acceptedTokens.length
-    }
-
-    public setTotalTokens(content: string) {
-        if (this._totalTokens.length === 0 && this._timer == undefined) {
-            const currentDate = new globals.clock.Date()
-            this._startTime = currentDate.getTime()
-            this.startTimer()
-        }
-
-        if (content.length <= 2) {
-            this._totalTokens.push(content)
-        } else if (content.length > 2) {
-            this._totalTokens.push(...content)
-        }
+        // generate accepted recommendation token and stored in collection
+        this.addAcceptedTokens(filename, { range: range, text: text, accepted: text.length })
+        this.addTotalTokens(filename, text.length)
     }
 
     public flush() {
         const terms = this._globals.get<boolean>(CodeWhispererConstants.termsAcceptedKey) || false
         if (!terms) {
-            this._totalTokens = []
-            this._acceptedTokens = []
+            this._totalTokens = {}
+            this._acceptedTokens = {}
             this.closeTimer()
             return
         }
-        this.emitCodeWhispererCodeContribution()
+        try {
+            this.emitCodeWhispererCodeContribution()
+        } catch (error) {
+            getLogger().error(`Encountered ${error} when emitting code contribution metric`)
+        }
+    }
+
+    public updateAcceptedTokensCount(editor: vscode.TextEditor) {
+        const filename = editor.document.fileName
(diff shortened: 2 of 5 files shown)


src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
```diff
@@ -62,18 +191,20 @@ describe('codewhispererCodecoverageTracker', function () {
...
+        afterEach(function () {
+            sinon.restore()
+        })
         it(' emits codecoverage telemetry ', function () {
-            const sample = "print('Hello World!)"
             const tracker = CodeWhispererCodeCoverageTracker.getTracker(language, mockGlobalStorage)
             const assertTelemetry = assertTelemetryCurried('codewhisperer_codePercentage')
-            tracker.setAcceptedTokens(sample)
-            tracker.emitCodeWhispererCodeContribution()
+            tracker?.addAcceptedTokens(`test.py`, { range: new vscode.Range(0, 0, 0, 7), text: `print()`, accepted: 7 })
```
JadenSimon, line 202: Exposing `addAcceptedTokens` feels clunky. Isn't `accepted` always a computed field of `text`? And isn't `text` always the text document text provided by `range`? So a 'token' can be computed using a text document + a range.

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -8,120 +8,186 @@ import * as telemetry from '../../shared/telemetry/telemetry'
...
-                    const totalTokens = this._totalTokens
-                    const acceptedTokens = this._acceptedTokens
-                    if (totalTokens.length > 0 && acceptedTokens.length > 0) {
-                        this.flush()
-                    } else {
-                        getLogger().debug(
-                            `CodeWhispererCodeCoverageTracker: skipped telemetry due to empty tokens array`
-                        )
+    private tryStartTimer() {
+        if (this._timer === undefined) {
```
Will-ShaoHua, line 106: nit: if (this._timer !== undefined) return;

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
```diff
@@ -62,18 +191,20 @@ describe('codewhispererCodecoverageTracker', function () {
...
+            tracker?.addTotalTokens(`test.py`, 100)
+            tracker?.emitCodeWhispererCodeContribution()
             assertTelemetry({
```
JadenSimon, line 205: Having more tests using `assertTelemetry` would be nice if possible. Like I want to see how different combinations of user actions produce different results.

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
```diff
@@ -7,28 +7,157 @@ import * as assert from 'assert'
...
-            tracker.setAcceptedTokens(suggestion)
-            assert.strictEqual(tracker.AcceptedTokensLength, 21)
+            tracker?.addAcceptedTokens(editor.document.fileName, {
+                range: new vscode.Range(0, 0, 0, 39),
+                text: `import math\ndef two_sum(nums, target):\n`,
+                accepted: 39,
+            })
+            tracker?.addTotalTokens(editor.document.fileName, 100)
+            tracker?.updateAcceptedTokensCount(editor)
+            assert.strictEqual(tracker?.acceptedTokens[editor.document.fileName][0].accepted, 12)
```
JadenSimon, line 38: Maybe I'm misunderstanding things but I'm confused as to why this is 12?
Distance is 17 so shouldn't it be 22?

Will-ShaoHua: +1

leigaol:
> @leigaol could you add a brief walkthrough on the PR description regarding the mechanism how or when `acceptedTokens & totalTokens` will be updated

Thanks. I updated the PR description

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
Will-ShaoHua: since the test case is regarding edit-distance, could you add a code comment on it what's the user modification here

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -100,28 +127,66 @@ export class CodeWhispererCodeCoverageTracker {
...
+            vsCodeState.isCodeWhispererEditing ||
+            e.contentChanges.length !== 1
+        )
+            return
+        const content = e.contentChanges[0]
+        // do not count user tokens if user copies large chunk of code
+        if (content.text.length > 20) return
+        this.tryStartTimer()
+        if (content.text.length === 0) {
+            this.addTotalTokens(e.document.fileName, -content.rangeLength)
```
zixlin7, line 176: nit: It took me a while to figure out this is the delete case, could you pls add a comment here?

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -8,88 +8,115 @@ import * as telemetry from '../../shared/telemetry/telemetry'
...
+    }
+
+    public updateAcceptedTokensCount(editor: vscode.TextEditor) {
+        const filename = editor.document.fileName
+        if (filename in this._acceptedTokens) {
+            for (let i = 0; i < this._acceptedTokens[filename].length; i++) {
+                const oldText = this._acceptedTokens[filename][i].text
+                const newText = editor.document.getText(this._acceptedTokens[filename][i].range)
+                this._acceptedTokens[filename][i].accepted =
+                    Math.max(oldText.length, newText.length) - distance(oldText, newText)
```
Will-ShaoHua, line 74: Lei i would suggest we can extract this block of logic to be another function and and we can test against easily. And we can add this test case  https://github.com/aws/aws-toolkit-jetbrains/blob/main/jetbrains-core/tst/software/aws/toolkits/jetbrains/services/codewhisperer/CodeWhispererCodeCoverageTrackerTest.kt#L345 for readers to more easily understand what's we wanna achieve here

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
leigaol: The accepted code is `import math\ndef two_sum(nums, target):\n`

User changed it to 'import math\ndef addTwoNumbers(a, b):\n'

In this case, user accepted 'import math\ndef ', so the accepted is 16. @Will-ShaoHua we can go over this

zixlin7: +1, some explaining here could help

zixlin7 (approved):

src/codewhisperer/commands/onAcceptance.ts
```diff
@@ -94,8 +94,11 @@ export async function onAcceptance(acceptanceEntry: OnRecommendationAcceptanceEn
             completionType: acceptanceEntry.completionType,
             language: languageContext.language,
         })
-        CodeWhispererCodeCoverageTracker.getTracker(languageContext.language, globalStorage).setAcceptedTokens(
-            acceptanceEntry.recommendation
+        const codeRangeAfterFormat = new vscode.Range(start, acceptanceEntry.editor.selection.active)
+        CodeWhispererCodeCoverageTracker.getTracker(languageContext.language, globalStorage)?.countAcceptedTokens(
```
Will-ShaoHua, line 98: since `countAcceptedTokens` will also update totalToken, and documentChanged event also will update totalToken, will the calculation be performed twice

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
Will-ShaoHua: oops sorry didn't see that

Will-ShaoHua: just a nit, maybe we can add variables like`originalRecommendation` `modifiedRecommendation`

Will-ShaoHua: as for the number, i obtained 24.. lol

Will-ShaoHua: so given this formula https://github.com/aws/aws-toolkit-vscode/pull/2845/files#diff-03e0a75d43072c8dd392c13a8ce7549b075fd4caf70c2539f06fcaee2cc7424dR74, editDistance = 17 and maxOf(originalRecomm.length, modifiedRecomm.length) = maxOf(41, 39)
so the delta should be 41 - 17 = 24

src/codewhisperer/commands/onAcceptance.ts
leigaol: It won't. We are using `vsCodeState.isCodeWhispererEditing` to filter documentChange event from CWSPR

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
Will-ShaoHua: I would suggest we can use a simpler originalRecomm & modifiedRecomm and add complicate ones to another test case as mentioned in this thread https://github.com/aws/aws-toolkit-vscode/pull/2845/files#r952992751

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
Will-ShaoHua: here is ref to the `getAcceptredTokenDelta` for JB https://github.com/aws/aws-toolkit-jetbrains/blob/main/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/telemetry/CodeWhispererCodeCoverageTracker.kt#L106

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
leigaol: Field `accepted`  is used to store the number of unmodified accepted tokens

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
leigaol: More test cases have been added

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
leigaol: This test case has been updated. Thanks @Will-ShaoHua

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
leigaol: New test cases have been added.

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
```diff
@@ -7,28 +7,182 @@ import * as assert from 'assert'
...
+    describe('updateAcceptedTokensCount', function () {
         beforeEach(function () {
             resetCodeWhispererGlobalVariables()
             CodeWhispererCodeCoverageTracker.instances.delete(language)
         })
 
-        it('Should set Accepted Tokens when CodeWhisperer terms accepted', function () {
-            const suggestion = "print('Hello World!')"
+        it('Should compute edit distance to update the accepted tokens', function () {
+            const editor = createMockTextEditor('def addTwoNumbers(a, b):\n')
```
Will-ShaoHua, line 29: nit: `const originalRecommendation = 'def addTwoNumbers(a, b):\n'`

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -8,88 +8,122 @@ import * as telemetry from '../../shared/telemetry/telemetry'
...
+        }
+    }
+    // With edit distance, complicate usermodification can be considered as simple edit(add, delete, replace),
+    // and thus the unmodified part of recommendation length can be deducted/approximated
+    // ex. (modified > original): originalRecom: foo -> modifiedRecom: fobarbarbaro, distance = 9, delta = 12 - 9 = 3
+    // ex. (modified == original): originalRecom: helloworld -> modifiedRecom: HelloWorld, distance = 2, delta = 10 - 2 = 8
+    // ex. (modified < original): originalRecom: CodeWhisperer -> modifiedRecom: CODE, distance = 12, delta = 13 - 12 = 1
+    public getUnmodifiedAcceptedTokens(origin: string, after: string) {
+        return Math.max(origin.length, after.length) - distance(origin, after)
     }
```
JadenSimon, line 84: nit: you can move this outside the class now

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
JadenSimon: Right I think the clunky-ness comes from lack of explicit initialization of the tracking behavior. Not a blocker.

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -8,88 +8,122 @@ import * as telemetry from '../../shared/telemetry/telemetry'
...
 /**
- * This singleton class is mainly used for calculating the percentage of user modification.
- * The current calculation method is (Levenshtein edit distance / acceptedSuggestion.length).
+ * This singleton class is mainly used for calculating the code written by codeWhisperer
  */
 export class CodeWhispererCodeCoverageTracker {
-    private _acceptedTokens: string[]
-    private _totalTokens: string[]
+    private _acceptedTokens: { [key: string]: CodeWhispererToken[] }
+    private _totalTokens: { [key: string]: number }
```
JadenSimon, line 25: Using objects as maps can be convenient but it'll cause problems for completely arbitrary strings. For example, if a file is named 'toString' you'll see some rather odd behavior. I'd just use a `Map` as we aren't using any narrowed string types to make using an object worth it.

JadenSimon (approved):

src/test/codewhisperer/tracker/codewhispererCodeCoverageTracker.test.ts
Will-ShaoHua: Or we can just have some comments on this test case, it works too!

Will-ShaoHua (approved):

justinmk3:
looks like there are some comments still to followup on, so waiting to merge

leigaol:
> looks like there are some comments still to followup on, so waiting to merge

I talked to comment owners and these are not blockers.

PR #2847, opened 2022-08-23
feat(codewhisperer): Update security scan telemetry to track overall project size

fangwade007:
## Problem
- Update security scan telemetry to track overall project size

## Solution
- Add util method to get working directory size;
- Update security scan command to emit project size field telemetry;
- Verification is as follow images

Size from Finder view:
Screen Shot 2022-08-23 at 5 04 07 PM

Telemetry log indicates successfully emit essential field:
Screen Shot 2022-08-23 at 5 03 55 PM

## License

### DIFF
--- a/src/codewhisperer/commands/startSecurityScan.ts
+++ b/src/codewhisperer/commands/startSecurityScan.ts
@@ -24,6 +24,8 @@ import { codeScanState, CodeScanTelemetryEntry } from '../models/model'
 import { openSettings } from '../../shared/settings'
 import { ok, viewSettings } from '../../shared/localizedText'
 import { statSync } from 'fs'
+import { getFileExt } from '../util/commonUtil'
+import { getDirSize } from '../../shared/filesystemUtilities'
 
 const performance = globalThis.performance ?? require('perf_hooks').performance
 
@@ -168,8 +170,24 @@ export async function startSecurityScan(
         await vscode.commands.executeCommand('aws.codeWhisperer.refresh')
         codeScanTelemetryEntry.duration = performance.now() - codeScanStartTime
         codeScanTelemetryEntry.codeScanServiceInvocationsDuration = performance.now() - serviceInvocationStartTime
-        telemetry.recordCodewhispererSecurityScan(codeScanTelemetryEntry)
+        emitCodeScanTelemetry(editor, codeScanTelemetryEntry)
     }
+}
+
+export async function emitCodeScanTelemetry(editor: vscode.TextEditor, codeScanTelemetryEntry: CodeScanTelemetryEntry) {
+    const uri = editor.document.uri
+    const workspaceFolder = vscode.workspace.getWorkspaceFolder(uri)
+    const fileExt = getFileExt(editor.document.languageId)
+    if (workspaceFolder !== undefined && fileExt !== undefined) {
+        const projectSize = await getDirSize(
+            workspaceFolder.uri.fsPath,
+            performance.now(),
+            CodeWhispererConstants.projectSizeCalculateTimeoutSeconds * 1000,
+            fileExt
+        )
+        codeScanTelemetryEntry.codewhispererCodeScanProjectBytes = projectSize
+    }
+    telemetry.recordCodewhispererSecurityScan(codeScanTelemetryEntry)
 }
 
 export function errorPromptHelper(error: Error) {
--- a/src/codewhisperer/models/constants.ts
+++ b/src/codewhisperer/models/constants.ts
@@ -127,6 +127,7 @@ export const CodeWhispererConstants = {
     codeScanZipExt: '.zip',
     contextTruncationTimeoutSeconds: 10,
     codeScanJobTimeoutSeconds: 50,
+    projectSizeCalculateTimeoutSeconds: 10,
     codeScanJobPollingIntervalSeconds: 5,
     artifactTypeSource: 'SourceCode',
     artifactTypeBuild: 'BuiltJars',
--- a/src/codewhisperer/util/commonUtil.ts
+++ b/src/codewhisperer/util/commonUtil.ts
@@ -2,7 +2,6 @@
  * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
-
 export function getLocalDatetime() {
     const timezone = Intl.DateTimeFormat().resolvedOptions().timeZone
     return new Date().toLocaleString([], { timeZone: timezone })
@@ -26,4 +25,16 @@ export function isAwsError(error: any): boolean {
         typeof error.code === 'string' &&
         error.time instanceof Date
     )
+}
+
+export function getFileExt(languageId: string) {
+    switch (languageId) {
+        case 'java':
+            return '.java'
+        case 'python':
+            return '.py'
+        default:
+            break
+    }
+    return undefined
 }
--- a/src/shared/filesystemUtilities.ts
+++ b/src/shared/filesystemUtilities.ts
@@ -6,6 +6,7 @@
 import { access, mkdtemp, mkdirp, readFile, remove, existsSync, readdir, stat } from 'fs-extra'
 import * as crypto from 'crypto'
 import * as fs from 'fs'
+import * as fsExtra from 'fs-extra'
 import * as os from 'os'
 import * as path from 'path'
 import * as vscode from 'vscode'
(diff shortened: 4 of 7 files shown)


src/codewhisperer/commands/startSecurityScan.ts
```diff
@@ -78,6 +79,9 @@ export async function startSecurityScan(
             throw new Error(`"${editor.document.languageId}" is not supported for security scan.`)
         }
         const uri = dependencyGraph.getRootFile(editor)
+        const projectSize = await dependencyGraph.getProjectSize(uri)
```
leigaol, line 82: If this `projectSize` is only used in telemetry reporting, is it possible to let `await dependencyGraph.getProjectSize(uri)` not block the execution of `startSecurityScan` logic?

src/codewhisperer/util/dependencyGraph/dependencyGraph.ts
```diff
@@ -27,6 +27,21 @@ export interface TruncPaths {
     lines: number
 }
 
+export async function dirSize(dirPath: string): Promise<number> {
```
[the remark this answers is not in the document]
fangwade007: Updated

src/codewhisperer/commands/startSecurityScan.ts
fangwade007: Updated

leigaol (approved):

JadenSimon (approved):

src/codewhisperer/util/commonUtil.ts
```diff
@@ -27,3 +32,27 @@ export function isAwsError(error: any): boolean {
         error.time instanceof Date
     )
 }
+
+export async function getDirSize(dirPath: string, startTime: number, duration: number): Promise<number> {
```
abhikuhikar, line 36: I was wondering if it would make sense to exclude the build artifacts (or binary files) in the project from the project size as build artifacts can be very large compared to the source files and I think we are interested in knowing what percentage of the source files were actually scanned. Including the build artifacts might skew this data.

JadenSimon: In that case we probably want to re-use the dependency graph logic for source file detection here. I'd also update [the description](https://github.com/aws/aws-toolkit-common/commit/e7cfb5668f0e01431e643b5155f25bf6eea4f40e#diff-f9ad596fbedec6dc08aa26f65b8a0f9358ae2be13476b00a82557f7d7125c568R408) to clarify that only source files should be included.

src/codewhisperer/util/commonUtil.ts
```diff
@@ -27,3 +32,43 @@ export function isAwsError(error: any): boolean {
...
+            return '.java'
+        case 'python':
+            return '.py'
+        default:
+            break
+    }
+    return undefined
+}
+
+export async function getDirSize(
```
justinmk3, line 48: I see that this is not part of fs-extra because it is complex: https://github.com/jprichardson/node-fs-extra/issues/656

- That issue mentions symlinks (which can be circular, causing an infinite loop).
    - Perhaps when recursing, we should pass a "depth" parameter and cancel recursion if depth exceeds e.g. 50. That avoids infinite loops. 
    - We could return a result that indicates if the depth limit is reached.
- This function is not specific to CW. So it should live in `src/shared/filesystemUtilities.ts`

src/codewhisperer/util/commonUtil.ts
```diff
@@ -27,3 +32,43 @@ export function isAwsError(error: any): boolean {
...
+        }
+        return 0
+    })
+    return (await Promise.all(fileSizes)).reduce((accumulator, size) => accumulator + size, 0)
+}
+
+export function getWorkspaceFolder(uri: vscode.Uri) {
+    const workspaceFolder = vscode.workspace.getWorkspaceFolder(uri)
+    if (workspaceFolder === undefined) {
+        return path.dirname(uri.fsPath)
```
justinmk3, line 71: If the `getWorkspaceFolder` call fails, this returns a non-workspace function. So the `getWorkspaceFolder` name is misleading. 

Also the purpose of this function isn't obvious. Needs a short (less than 10 words) docstring.

abhikuhikar (approved):

src/codewhisperer/util/commonUtil.ts
fangwade007: Thanks for the callout Justin.

- For symbolicLink, please check https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/fs.d.ts#L50 there's a `isSymbolicLink()` and we're already using it for Java dependency graph https://github.com/aws/aws-toolkit-vscode/blob/master/src/codewhisperer/util/dependencyGraph/javaDependencyGraph.ts#L307; in this util function, we only check `isDirectory ` and `isFile`, which I should've avoided circular concern;
- sure, I can move it to `filesystemUtilities `

src/codewhisperer/util/commonUtil.ts
fangwade007: This is a good point, I will need to change the logic little bit; for no workspaceFolder use case, it usually means customer is working on a single file, in this case, I should prevent doing directory traverse since the directory contains the file is not working project.

src/codewhisperer/util/commonUtil.ts
fangwade007: Also to add to the "depth", this function has time based stop condition, which the limit now is 10 seconds.

zixlin7 (approved):

src/codewhisperer/commands/startSecurityScan.ts
```diff
@@ -168,10 +170,26 @@ export async function startSecurityScan(
...
         codeScanTelemetryEntry.duration = performance.now() - codeScanStartTime
         codeScanTelemetryEntry.codeScanServiceInvocationsDuration = performance.now() - serviceInvocationStartTime
-        telemetry.recordCodewhispererSecurityScan(codeScanTelemetryEntry)
+        emitCodeScanTelemetry(editor, codeScanTelemetryEntry)
     }
 }
 
+export async function emitCodeScanTelemetry(editor: vscode.TextEditor, codeScanTelemetryEntry: CodeScanTelemetryEntry) {
+    const uri = editor.document.uri
+    const workspaceFolder = getWorkspaceFolder(uri)
```
justinmk3, line 179: can just do:
```suggestion
    const workspaceFolder = vscode.workspace.getWorkspaceFolder(uri)?.uri
```

src/codewhisperer/util/commonUtil.ts
```diff
@@ -27,3 +28,20 @@ export function isAwsError(error: any): boolean {
...
+            return '.java'
+        case 'python':
+            return '.py'
+        default:
+            break
+    }
+    return undefined
+}
+
+export function getWorkspaceFolder(uri: vscode.Uri) {
```
justinmk3, line 44: not needed now

src/shared/filesystemUtilities.ts
```diff
@@ -20,6 +21,27 @@ export const tempDirPath = path.join(
     'aws-toolkit-vscode'
 )
 
+export async function getDirSize(
+    dirPath: string,
+    startTime: number,
+    duration: number,
+    fileExt: string
+): Promise<number> {
+    if (performance.now() - startTime > duration) return 0
```
justinmk3, line 30: should log a warn() if it exceeds the limit

src/codewhisperer/models/constants.ts
```diff
@@ -127,6 +127,7 @@ export const CodeWhispererConstants = {
     codeScanZipExt: '.zip',
     contextTruncationTimeoutSeconds: 10,
     codeScanJobTimeoutSeconds: 50,
+    projectSizeCalculateTimeoutSeconds: 10,
```
justinmk3, line 130: 10s seems high. Especially just for telemetry. Does this risk a bad experience for customers?

src/codewhisperer/commands/startSecurityScan.ts
fangwade007: right

src/codewhisperer/models/constants.ts
fangwade007: CodeScan telemetry is now async and decoupled from the feature running progress, set to 10s in case edge case which customer's project is very big; Previous job's telemetry will be guaranteed to emit before coming job finish as the SLA is 60s, average cost is ~30s now

abhikuhikar: TBH, I don't think we really need a timeout for this. It's just a telemetry and the UX is not blocked on this (or at least it should not be).

fangwade007: No, we need, have done some benchmark, even system command can take minutes to finish if it's too large; timeout is def better experience and prevents eating up customer's system resources

justinmk3: > and the UX is not blocked on this (or at least it should not be).

Couldn't it block other Toolkit features, or greatly slow them down? This is only one case also; there are many such recursive workspace functions in codewhisperer, each adding risk. A timeout or some sort of guard is something we should have in each case as a general habit. The minimum would be to log a warning, if not cancel the operation.

justinmk3 (approved):

src/codewhisperer/models/constants.ts
fangwade007: Totally agree, that's why for all those recursion in CWSPR, we have wrap with timeout either on time or size

PR #2865, opened 2022-09-02
feat(sam): improve run/debug error handling and telemetry

JadenSimon:
## Problem
Our SAM run/debug implementation has several problems with error handling:
* Inconsistent error message UX (some are placed in the logs, some are shown directly to the user)
* Telemetry only covers a portion of the relevant code paths and does not emit reasons for failure
* Some errors are swallowed while others are bubbled up to the VSC API

## Solution
* Bubble up all errors but use a single error handler to prevent the VSC error modal from showing
    * Added a button to open the launch config to preserve current UX 
* Use telemetry spans to cover almost the entire code path
* Remove `sam_attachDebugger` metric as it is redundant. Failures to attach counts as a failure for the whole execution.
* Add some basic error codes to better determine why a run/debug execution failed

Screen Shot 2022-09-02 at 9 37 24 AM

## License

### DIFF
--- a/src/shared/sam/debugger/awsSamDebugger.ts
+++ b/src/shared/sam/debugger/awsSamDebugger.ts
@@ -7,7 +7,6 @@ import * as semver from 'semver'
 import * as vscode from 'vscode'
 import * as fs from 'fs-extra'
 import * as nls from 'vscode-nls'
-import { Runtime } from 'aws-sdk/clients/lambda'
 import {
     getCodeRoot,
     getHandlerName,
@@ -17,6 +16,7 @@ import {
     GoDebugConfiguration,
     getTemplate,
     getArchitecture,
+    isImageLambdaConfig,
 } from '../../../lambda/local/debugConfiguration'
 import {
     Architecture,
@@ -53,10 +53,9 @@ import { makeInputTemplate, makeJsonFiles } from '../localLambdaRunner'
 import { SamLocalInvokeCommand } from '../cli/samCliLocalInvoke'
 import { getCredentialsFromStore } from '../../../credentials/credentialsStore'
 import { fromString } from '../../../credentials/providers/credentials'
-import { notifyUserInvalidCredentials } from '../../../credentials/credentialsUtilities'
 import { Credentials } from '@aws-sdk/types'
 import { CloudFormation } from '../../cloudformation/cloudformation'
-import { getSamCliVersion } from '../cli/samCliContext'
+import { getSamCliContext, getSamCliVersion } from '../cli/samCliContext'
 import {
     MINIMUM_SAM_CLI_VERSION_INCLUSIVE_FOR_IMAGE_SUPPORT,
     MINIMUM_SAM_CLI_VERSION_INCLUSIVE_FOR_GO_SUPPORT,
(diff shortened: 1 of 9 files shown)


justinmk3:
> Remove sam_attachDebugger metric as it is redundant

Are VS/JB still using that? Maybe keep it for now, but mark as deprecated so that the dashboards can be compared?

src/shared/sam/debugger/awsSamDebugger.ts
```diff
@@ -642,7 +711,7 @@ export class SamDebugConfigProvider implements vscode.DebugConfigurationProvider
                 return await javaDebug.invokeJavaLambda(this.ctx, config)
             }
             default: {
-                throw Error(`unknown runtimeFamily: ${config.runtimeFamily}`)
+                throw new Error(`unknown runtimeFamily: ${config.runtimeFamily}`)
```
justinmk3, line 714: is there a technical difference here? should we have a lint rule for it?

justinmk3 (approved):

src/shared/sam/debugger/awsSamDebugger.ts
JadenSimon: For this class implementation in particular you will get the same result but calling a constructor function without using `new` is functionally different. The `Error` constructor is likely using `new` internally if you don't call it with the `new` keyword so you get the same thing anyway. 

Wouldn't be a bad idea to lint for `new` preceding any Error ctor call if only for consistency.

JadenSimon:
> Are VS/JB still using that? Maybe keep it for now, but mark as deprecated so that the dashboards can be compared?

VSC only. The only metric name we share here is `lambda_invokeLocal`.

src/shared/sam/debugger/awsSamDebugger.ts
```diff
@@ -310,12 +353,31 @@ export class SamDebugConfigProvider implements vscode.DebugConfigurationProvider
...
+                    return this.invokeConfig(resolved)
+                })
+            }
+        } catch (err) {
+            if (err instanceof SamLaunchRequestError) {
+                err.showNotification()
+            } else if (err instanceof ToolkitError) {
+                new SamLaunchRequestError(err.message, { ...err }).showNotification()
+            } else {
+                SamLaunchRequestError.chain(err, 'Failed to run launch configuration').showNotification()
```
justinmk3, line 378: since the error no longer reaches vscode, do we need to call `vscode.debug.stopDebugging()` explicitly? https://github.com/microsoft/vscode/issues/102037

JadenSimon: We have to regardless as our whole integration is effectively a hack. We are launching debugger sessions during config resolution so we are responsible for terminating them appropriately: https://github.com/aws/aws-toolkit-vscode/pull/2865/files/5dca9e313f57ae439d7fc35c0e52e465174b6279#diff-b36cc56a594606340ec0320f08168751338535f63e948fdcaab177666baffb79R329-R332

justinmk3: > while the debugger session is delegated to the appropriate extension per-language.

👍  So `vscode.debug.stopDebugging()` must be called by the debugger provider, not us.

Issue #1442, opened 2020-12-15
cloud9: SAM debug fails with Image resource + python 3.8


justinmk3:
**To Reproduce**

1. create a cloud9 environment with Toolkit 1.18 9efb496
1. Create sam application with runtime `python 3.8 (Image)`
    - issue **does NOT affect python 3.7**
1. try to run/debug the Image resource
1. output log shows: `unable to execute 'gcc': No such file or directory`

---

-   OS: AL2
-   Visual Studio Code Version: cloud9/1.42
-   AWS Toolkit Version: 1.18 https://github.com/aws/aws-toolkit-vscode/pull/1438/commits/9efb496151b5e49ebc295514dfc6454250a90f2d

## Full log

```
Preparing to debug locally: API "/hello", Lambda "HelloWorldFunction"
Building SAM application...
Running command: (not started) [/home/linuxbrew/.linuxbrew/bin/sam build --build-dir /tmp/aws-toolkit-vscode/vsctkt5Ceuk/output --template /home/ec2-user/environment/py38image/app___vsctk___template.yaml --use-container --manifest /home/ec2-user/environment/py38image/hello_world/requirements.txt]
Starting Build inside a container
Building codeuri: . runtime: None metadata: {'Dockerfile': 'Dockerfile', 'DockerContext': './hello_world', 'DockerTag': 'python3.8-v1'} functions: ['HelloWorldFunction']
Building image for HelloWorldFunction function
Setting DockerBuildArgs: {'SAM_BUILD_MODE': 'debug'} for HelloWorldFunction function
Step 1/4 : FROM public.ecr.aws/lambda/python:3.8
 ---> 80e51cc8a434
Step 2/4 : COPY app.py requirements.txt ./
 ---> ebb3cdbe7aa8
Step 3/4 : RUN python3.8 -m pip install -r requirements.txt -t .
 ---> Running in 756c7c342261
Collecting requests
  Downloading requests-2.25.0-py2.py3-none-any.whl (61 kB)
Collecting ikp3db
  Downloading ikp3db-1.4.1.tar.gz (24 kB)
Collecting chardet<4,>=3.0.2
  Downloading chardet-3.0.4-py2.py3-none-any.whl (133 kB)
Collecting certifi>=2017.4.17
  Downloading certifi-2020.12.5-py2.py3-none-any.whl (147 kB)
Collecting idna<3,>=2.5
  Downloading idna-2.10-py2.py3-none-any.whl (58 kB)
[...]

PR #2870, opened 2022-09-06
fix(sam): use different debugger args for Python image lambdas on C9 

JadenSimon:
## Problem
Debugging Python image lambdas does not work on C9

## Solution
Use the same pattern that we use for debugpy

## License

### DIFF
--- a/src/shared/sam/debugger/pythonSamDebug.ts
+++ b/src/shared/sam/debugger/pythonSamDebug.ts
@@ -138,9 +138,14 @@ export async function makePythonDebugConfig(
             // --ikpdb-protocol=vscode:
             //           For https://github.com/cmorisse/vscode-ikp3db
             //           Requires ikp3db 1.5 (unreleased): https://github.com/cmorisse/ikp3db/pull/12
-            config.debugArgs = [
-                `-m ikp3db --ikpdb-address=0.0.0.0 --ikpdb-port=${config.debugPort} -ik_ccwd=${ccwd} -ik_cwd=/var/task ${logArg}`,
-            ]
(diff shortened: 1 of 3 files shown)


src/shared/sam/debugger/pythonSamDebug.ts
```diff
@@ -138,9 +138,14 @@ export async function makePythonDebugConfig(
...
-            config.debugArgs = [
-                `-m ikp3db --ikpdb-address=0.0.0.0 --ikpdb-port=${config.debugPort} -ik_ccwd=${ccwd} -ik_cwd=/var/task ${logArg}`,
-            ]
+            const debugArgs = `-m ikp3db --ikpdb-address=0.0.0.0 --ikpdb-port=${config.debugPort} -ik_ccwd=${ccwd} -ik_cwd=/var/task ${logArg}`
+
+            if (isImageLambda) {
+                const params = getPythonExeAndBootstrap(config.runtime)
+                config.debugArgs = [`${params.python} ${debugArgs} ${params.bootstrap}`]
+            } else {
+                config.debugArgs = [debugArgs]
```
justinmk3, line 147: Did the sam expectations change? Or does this fix https://github.com/aws/aws-toolkit-vscode/issues/1442

JadenSimon: Fixes #1442

jpinkney-aws (approved):

src/shared/sam/debugger/pythonSamDebug.ts
JadenSimon: I should add that the original error shown in the related issue is due `ikp3db` being built from source in the image. But it can be fixed by adding this to the Dockerfile:
```
RUN yum groupinstall -y "Development Tools"
```

This PR fixes the problem you would run into _after_ updating the Dockerfile.

justinmk3:
Looks like the integ tests are failing since this change.

PR #2877, opened 2022-09-09
[CodeWhisperer] Refactor CodeWhispererConstants

Will-ShaoHua:
## Problem
Restructure CodeWhispererConstants file for easier usage and maintainability

## License


src/codewhisperer/models/constants.ts
```diff
@@ -3,143 +3,193 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-export const CodeWhispererConstants = {
+export namespace CodeWhispererConstants {
```
JadenSimon, line 6: Good refactor but namespaces are a somewhat outdated Typescript feature

I'd suggest removing the namespace and exporting each constant by itself. Then you can refactor all the imports with a search and replace (regex):
`import \{ CodeWhispererConstants \} from '(.*)'` -> `import * as CodeWhispererConstants from '$1'`

Will-ShaoHua: ackk

JadenSimon (approved):

zixlin7 (approved):

leigaol (approved):

PR #2880, opened 2022-09-12
feat(credentials): show "Edit Credentials" in error message

justinmk3:
## Problem

When onboarding "from zero" (no ~/.aws, need credentials) or when running the "Create Credentials Profile" command, if the provided credentials are invalid, or if the user needs to edit ~/.aws/credentials to provide non-static credentials, there is no obvious way to open the ~/.aws/credentials file for new users.

## Solution

Show "Edit Credentials" in the error message if "Create Credentials Profile" fails.

image

image

## License

### DIFF
--- a/.changes/next-release/Feature-9b4c339a-7cdf-4340-b911-f33bde49b461.json
+++ b/.changes/next-release/Feature-9b4c339a-7cdf-4340-b911-f33bde49b461.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "if \"Create Credentials Profile\" fails, error message now shows \"Edit Credentials\" button"
+}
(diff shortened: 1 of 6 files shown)


jpinkney-aws (approved):
Just tested and it's definitely a useful change. Just wondering if it makes sense to/if its possible to close the InputBox when we click the edit credentials button?

JadenSimon:
> Just wondering if it makes sense to/if its possible to close the InputBox when we click the edit credentials button?

Yeah right now this message is shown while the user is still in the 'create profile' wizard so if they want to change their inputs they should go back in the wizard.

This button would be helpful if added to `notifyUserInvalidCredentials`

justinmk3:
> This button would be helpful if added to `notifyUserInvalidCredentials` too, maybe de-dupe?

nice catch! updated.

Issue #2677, opened 2022-06-06
Prompt users to install VSCode-YAML when a non template.yaml CloudFormation/SAM file is opened


jpinkney-aws:
## Problem
Currently VSCode only prompts users to install VSCode-YAML when a `template.yaml` file is opened in their workspace. Now that we have more dynamic detection of whether or not a file is CloudFormation/SAM it would be cool to prompt users if we detect that the current file is CloudFormation/SAM, regardless of the name of the file.

## Expected behavior
Prompt the user to install VSCode-YAML when a CloudFormation/SAM file regardless of whether or not the extension is named `template.yaml`/`template.yml`

Related Code: https://github.com/aws/aws-toolkit-vscode/blob/8ff0737efae7f6ad2767689943335f0c2f98283d/src/shared/sam/activation.ts#L258

PR #2883, opened 2022-09-13
feat(YAML): Prompt users to install YAML plugin when AWSTemplateFormatVersion becomes available in their YAML document

jpinkney-aws:
## Problem
Right now, we are only prompting users to install the YAML plugin when AWSTemplateFormatVersion is in a document when it's opened/focused or the file is template.yaml/template.yml but it doesn't take into account when users type into the document

## Solution
If the YAML plugin is not installed then listen to onDidChangeEvents until AWSTemplateFormatVersion becomes available. This feature is currently configured to have a higher debounce rate since the prompt doesn't need to show particularly fast. Also not every yaml file they open will be an aws related file.

Tested against 40k line yaml files

Related issue: https://github.com/aws/aws-toolkit-vscode/issues/2677

## License

### DIFF
--- a/src/shared/sam/activation.ts
+++ b/src/shared/sam/activation.ts
@@ -377,7 +377,7 @@ async function promptInstallYamlPlugin(disposables: vscode.Disposable[]) {
     const response = await vscode.window.showInformationMessage(
         localize(
             'AWS.message.info.yaml.prompt',
-            'Install YAML extension for more {0} features in SAM template.yaml files.',
+            'Install YAML extension for more {0} features in CloudFormation templates',
             getIdeProperties().company
         ),
         installBtn,
--- a/src/shared/sam/activation.ts
+++ b/src/shared/sam/activation.ts
@@ -272,7 +272,7 @@ async function createYamlExtensionPrompt(): Promise<void> {
         // user opens a template file
         vscode.workspace.onDidOpenTextDocument(
             async (doc: vscode.TextDocument) => {
-                promptInstallYamlPlugin(doc.fileName, yamlPromptDisposables)
+                promptInstallYamlPluginFromFilename(doc.fileName, yamlPromptDisposables)
             },
             undefined,
             yamlPromptDisposables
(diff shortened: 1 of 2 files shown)


.changes/next-release/Bug Fix-098a61d6-b631-487b-82c3-e884075b25e7.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
```
JadenSimon, line 2: Feels more like a feature rather than a bug fix

src/shared/sam/activation.ts
```diff
@@ -287,6 +288,33 @@ async function createYamlExtensionPrompt(): Promise<void> {
...
+        function promptOnAWSTemplateFormatVersion(event: vscode.TextDocumentChangeEvent): void {
+            for (let line = 0; line < event.document.lineCount; line++) {
+                const textLine = event.document.lineAt(line)
+                if (!textLine.text.startsWith(' ') && textLine.text.includes('AWSTemplateFormatVersion')) {
+                    promptInstallYamlPlugin(yamlPromptDisposables)
+                    return
+                }
+            }
+        }
+        const debouncedUpdate = debounce(promptOnAWSTemplateFormatVersion, 1000)
```
JadenSimon, line 306: `debounce` wouldn't be needed if we limited one task per document, something like:
```ts
const notifications = new Map<string, Promise<unknown>>()

// in the event handler
const key = document.uri.toString()
if (!notifications.has(key)) { // and whatever other checks
    notifications.set(key, promptOnAWSTemplateFormatVersion.finally(() => notifications.delete(key)))
}
```

I need to make a generic form of this pattern though as I see it a lot. 

This is overkill but you could also not delete the promise to have it only show at most 1 notification per document but it'd be a memory leak. A weak ref would be needed against the document uri.

src/shared/sam/activation.ts
```diff
@@ -287,6 +288,33 @@ async function createYamlExtensionPrompt(): Promise<void> {
...
         )
 
+        /**
+         * Prompt the user to install the YAML plugin when AWSTemplateFormatVersion becomes available as a top level key
+         * in the document
+         * @param event An vscode text document change event
+         * @returns nothing
+         */
+        function promptOnAWSTemplateFormatVersion(event: vscode.TextDocumentChangeEvent): void {
+            for (let line = 0; line < event.document.lineCount; line++) {
```
JadenSimon, line 298: Is there any perf impact for big YAML files? This is being ran on every doc change so I'd at least use `contentChanges` instead of iterating over the whole document every time (does Cloud9 give us `contentChanges` ??)

jpinkney-aws: It worked upto 40k lines fine but I switched it to just look at the changed line which should work even faster

src/shared/sam/activation.ts
```diff
@@ -305,54 +345,62 @@ async function promptInstallYamlPluginFromEditor(
...
-                    getLogger().error(`Extension ${VSCODE_EXTENSION_ID.yaml} could not be opened: `, err.message)
-                }
-                break
-            case permanentlySuppress:
-                settings.disablePrompt('yamlExtPrompt')
-        }
+    const response = await vscode.window.showInformationMessage(
+        localize(
+            'AWS.message.info.yaml.prompt',
+            'Install YAML extension for more {0} features in SAM template.yaml files.',
```
JadenSimon, line 380: The logic implemented is not specific to SAM templates
```suggestion
            'Install YAML extension for more {0} features in CloudFormation templates',
```

JadenSimon (approved):

src/shared/sam/activation.ts
jpinkney-aws: I think the original prompt message was just added as SAM because it's in the SAM folder. I'll change it

src/shared/sam/activation.ts
```diff
@@ -287,6 +287,46 @@ async function createYamlExtensionPrompt(): Promise<void> {
...
+        /**
+         * Prompt the user to install the YAML plugin when AWSTemplateFormatVersion becomes available as a top level key
+         * in the document
+         * @param event An vscode text document change event
+         * @returns nothing
+         */
+        async function promptOnAWSTemplateFormatVersion(event: vscode.TextDocumentChangeEvent): Promise<void> {
+            for (const change of event.contentChanges) {
+                const changedLine = event.document.lineAt(change.range.start.line)
+                if (changedLine.text.includes('AWSTemplateFormatVersion')) {
```
justinmk3, line 299: Does this risk false-positives? E.g. a comment or string could contain "AWSTemplateFormatVersion".

Also we know the line with be in the top 3-4 lines of the doc, right? So we can skip lines that are below that?

jpinkney-aws: It has the small potential to have a false-positive but I think if the user is editing yaml, has the AWS toolkit installed, and has `AWSTemplateFormatVersion` in the document then theres a big chance they are looking for support

Alternatively, I could change `includes` to `startsWith` and then it's guaranteed to be key in the yaml document. I think most people probably put AWSTemplateFormatVersion in the top 4 lines but I don't think it's guaranteed by the spec so we might end up missing the rare few people that don't put it there

PR #2884, opened 2022-09-13
feat(Cloud9): Use yaml-language-service for cfn/sam schema support on Cloud 9

jpinkney-aws:
## Problem
Currently Cloud 9 does not support VSCode-YAML out of the box so we do not have schema support for cloudformation and sam

## Solution
Until Cloud 9 supports VSCode-YAML out of the box, start the yaml-language-service inside of the toolkit so that users can still have cfn/sam support

## License


JadenSimon (commented):
LGTM if it works on C9

justinmk3:
Which parts of this are temporary, i.e. can be removed later? Can they live in a single module? E.g. anywhere there is a `isCloud9()` should look like this:

```
if (isCloud9()) {
    cloud9Yaml.foo()
}
```

rather than this:

```
if (isCloud9()) {
    ...lots of logic...
}
```

Then the temporary parts are isolated and identifiable.

jpinkney-aws:
Moved the files around and tested everything on C9 and it's all seems to be working. Though the tests seem to be failing with a memory issue 🤔 

Now if we wanted to remove this we would need to:
1. Remove `yamlService.ts`
2. Remove the if `if (isCloud9())` from `yaml.ts`
3. Remove the yaml-language-server dependency from `package.json`
4. Remove the yaml-language-server reference in the webpack config `umd-compat-loader` section

webpack.config.js
```diff
@@ -66,7 +66,7 @@ const baseConfig = {
                 ],
             },
             {
-                test: /node_modules[\\|/](amazon-states-language-service|vscode-json-languageservice)/,
+                test: /node_modules[\\|/](amazon-states-language-service|vscode-json-languageservice|yaml-language-server)/,
                 use: { loader: 'umd-compat-loader' },
             },
```
JadenSimon, line 71: We can delete all of this. This is probably causing the v8 heap out-of-memory fault

jpinkney-aws: Thats what I was thinking, I'm just a bit confused why it didn't have the issue with vscode-json-languageservice but had it with the yaml one 🤔

JadenSimon: Loader is fine, I was looking at the wrong build script. 
We need to change `.github/workflows/node.js.yml` and move `NODE_OPTIONS: '--max-old-space-size=8192'` so it covers all commands for Mac and Windows.

webpack.config.js
```diff
@@ -66,7 +66,7 @@ const baseConfig = {
                 ],
             },
             {
-                test: /node_modules[\\|/](amazon-states-language-service|vscode-json-languageservice)/,
+                test: /node_modules[\\|/](amazon-states-language-service|vscode-json-languageservice|yaml-language-server)/,
```
JadenSimon, line 69:
```suggestion
                // Only needed for transitive `jsonc-parser` dependency
                // Can be removed when using esm
                test: /node_modules[\\|/](amazon-states-language-service|vscode-json-languageservice|yaml-language-server)/,
```

JadenSimon (approved):

src/shared/languageServer/utils/validator.ts
```diff
@@ -0,0 +1,38 @@
+/*!
```
justinmk3, line 1: `src/shared/languageServer/utils/validator.ts` is a deep, verbose path. Can be renamed to `src/shared/lsp/validator.ts`. We aren't likely to have tens of modules in there, so "utils/" is unnecessary nesting.

justinmk3 (approved):

PR #2888, opened 2022-09-16
feat(codewhisperer): Add inline Completion

leigaol:
## Problem

The current editor.edit based inline suggestion has many problems & bugs that sometimes glitches, removes user's code.

### Summary
With this feature, the official inlineCompletion API will be enabled for users who are using VS Code >= 1.68.0 and has `editor.suggest.enabled` = true. Otherwise the inline completion fallbacks to previous solution. AWS Cloud9 uses will still use intelliSense as its suggestion API.  

### Files changed

In the activation.ts file, we added extra if else branch to enabled switching between inlineCompletionAPI vs previous inline completion. The plugin will reload if user changes `editor.suggest.enabled` to re-register disposables for different inline completion API. Similar if-else control were also added to invokeRecommendation.ts and keystrokeHandler.ts

The onInlineAcceptance.ts file implements a callback function when an inlineCompletionItem is accepted for the new inlineCompletion code path. The inlineCompletionService.ts file implements a class that integrate all inline completion related logic for the new inlineCompletion code path. To avoid the extra cost of maintaining two code paths, in the long run, when most users have migrated to VSC >= 1.68.0, we can drop the previous inline experience by removing onAcceptance.ts and inlineCompletion.ts.

VS Code inline completion API is copied into src/shared/vscode/inlineCompletion.ts

### "Hacks" implemented in this PR 

1. To support pagination, the *inlineCompletionProvider* is disposed and re-registered each time the plugin receives a new paginated response from server. 

2. To avoid the “*Click Next/Previous button making suggestion disappear*” bugs in inline suggestion popup & to avoid using the suggestion popup that is not customizable, the configuration *editor.hover.enabled* is temporarily disabled(this also disables other hover widget from other plugins) when an inline suggestion is active and restored when inline suggestion is rejected or accepted or after 10 seconds.

3. To know the current active inline suggestion index in order to show relevant code reference text:
    Show Previous, Show Next Suggestion commands are temporarily overridden to dispose and re-register a new *inlineCompletionProvider* that has only one suggestion item so that we know which item is shown
The current active inline suggestion index is maintained in our plugin instead of inside VS Code. By maintaining the index in our plugin, we implemented suggestion navigation that stops at left/rightmost index instead of looping over all suggestions.

4. To support multi-line typeahead (native API only support one line typeahead), the ghost text and its range are split by line. The ghost text will be updated whenever typeahead reaches a new line.

5. User's keybinding of showNext, showPrevious will be overwritten to left, right arrow if they are using system default.

When inlineCompletionAdditions API(https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.inlineCompletionsAdditions.d.ts#L10)  is released, we can remove 2 and 3.

## License


src/codewhisperer/commands/onInlineAcceptance.ts
```diff
@@ -0,0 +1,113 @@
...
+                    insertSpaces: true,
+                }
+            )
+            if (edits && acceptanceEntry.editor) {
+                const wEdit = new WorkspaceEdit()
+                wEdit.set(acceptanceEntry.editor.document.uri, edits)
+                await workspace.applyEdit(wEdit)
+            }
+        }
+        vsCodeState.isCodeWhispererEditing = false
```
JadenSimon, line 71: Wouldn't hurt to handle failures as `isCodeWhispererEditing` is being used as a lock:
```ts
try {
    // ...
} finally {
    vsCodeState.isCodeWhispererEditing = false
}
```

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -0,0 +1,452 @@
...
+        await HoverConfigUtil.instance.restoreHoverConfig()
+        vsCodeState.isCodeWhispererEditing = false
+        this.referenceProvider?.removeInlineReference()
+        RecommendationHandler.instance.cancelPaginatedRequest()
+        RecommendationHandler.instance.reportUserDecisionOfCurrentRecommendation(editor, -1)
+        RecommendationHandler.instance.clearRecommendations()
+        this.disposeInlineCompletion()
+        this.setCodeWhispererStatusBarOk()
+        this.disposeCommandOverrides()
+        this.clearRejectionTimer()
```
JadenSimon, line 243: This is _a lot_ of state to keep track of for a single class. Any ideas for removing some of it?

src/codewhisperer/service/keyStrokeHandler.ts
```diff
@@ -190,6 +192,17 @@ export class KeyStrokeHandler {
                 } finally {
                     RecommendationHandler.instance.isGenerateRecommendationInProgress = false
                 }
+            } else if (isInlineCompletionEnabled()) {
+                if (!vsCodeState.isCodeWhispererEditing && !InlineCompletionService.instance.isPaginationRunning()) {
```
JadenSimon, line 196: Usually it's better to hide these types of guards within the implementation rather than making it the caller's responsibility.

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -0,0 +1,452 @@
...
+                )
+
+                if (RecommendationHandler.instance.checkAndResetCancellationTokens()) {
+                    RecommendationHandler.instance.reportUserDecisionOfCurrentRecommendation(editor, -1)
+                    RecommendationHandler.instance.clearRecommendations()
+                    this.setCodeWhispererStatusBarOk()
+                    return
+                }
+
+                if (!this.isCurrentCompletionInProgress()) {
```
JadenSimon, line 287: What's the point of this? Are we silently skipping an update if the completion provider is providing results?
The current impl. is not async so `isCurrentCompletionInProgress` should always be false

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -0,0 +1,452 @@
...
+
+    setReferenceInlineProvider(provider: ReferenceInlineProvider) {
+        this.referenceProvider = provider
+    }
+
+    setExtContext(context: ExtContext) {
+        this.extContext = context
+    }
+
+    async getPaginatedRecommendation(
```
JadenSimon, line 254: This is not specific to this PR but the way CodeWhisperer currently does pagination feels overly complicated. We should try to simplify it in a separate PR.

src/codewhisperer/service/inlineCompletionService.ts
leigaol: I should just rename this function to `isCurrentSuggestionVisible`. It avoids an update if a suggestion is already visible.

It won't always be false.

isCurrentCompletionInProgress(): boolean {
        return this.inlineCompletionProvider?.getActiveItemIndex !== undefined
    }

src/codewhisperer/service/keyStrokeHandler.ts
leigaol: Agree. Will move it inside.

src/codewhisperer/service/inlineCompletionService.ts
leigaol: Agree

src/codewhisperer/activation.ts
```diff
@@ -269,10 +291,72 @@ export async function activate(context: ExtContext): Promise<void> {
...
+                 * 2. It is not a backspace
+                 * 3. It is not caused by CodeWhisperer editing
+                 * 4. It is not from undo/redo.
+                 */
+                if (
+                    e.document === vscode.window.activeTextEditor?.document &&
+                    runtimeLanguageContext.isLanguageSupported(e.document.languageId) &&
+                    e.contentChanges.length != 0 &&
+                    !vsCodeState.isCodeWhispererEditing &&
+                    !JSON.stringify(e).includes('reason')
```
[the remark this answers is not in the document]
leigaol: `reason` field is removed.

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -0,0 +1,452 @@
...
+    )
+    private nextCommand = Commands.declare(
+        'editor.action.inlineSuggest.showNext',
+        (context: ExtContext) => async () => {
+            await this.showRecommendation(1)
+        }
+    )
+    private hideCommand = Commands.declare('editor.action.inlineSuggest.hide', (context: ExtContext) => async () => {
+        await this.clearInlineCompletionStates(vscode.window.activeTextEditor)
+    })
```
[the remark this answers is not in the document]
leigaol: Great suggestion. Those changes have been committed.

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -0,0 +1,452 @@
...
+    }
+}
+
+export class InlineCompletionService {
+    private referenceProvider?: ReferenceInlineProvider
+    private inlineCompletionProvider?: CodeWhispererInlineCompletionItemProvider
+    private inlineCompletionProviderDisposable?: vscode.Disposable
+    private maxPage = 100
+    private statusBar: vscode.StatusBarItem
+    private extContext?: ExtContext
```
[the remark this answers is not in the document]
leigaol: `ExtContext` is now removed. 

I added comments to explain that this command override only happens during a suggestion lifecycle. It is only effective starting from  "user starts invocation" to "suggestion is cleared/accepted". This is to minimize impact on other plugins that uses this inlineCompletion API.

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    isPaginationRunning() {
+        return this.statusBar.text === ` $(loading~spin)CodeWhisperer`
+    }
+
+    /**
+     * Override keybindings for next, previous suggestion
+     * only when user did not set those keybindings
+     */
```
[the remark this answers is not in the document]
leigaol: This method is now moved to a separate util  function and test cases are added.

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -0,0 +1,452 @@
...
+            if (overrides.length > 0) {
+                let newKeybindings = ''
+                if (origin.includes('[') && origin.includes(']')) {
+                    newKeybindings = overrides + origin.substring(origin.indexOf('[') + 1, origin.lastIndexOf(']'))
+                } else {
+                    newKeybindings = overrides
+                }
+                fs.writeFileSync(
+                    keybindingsPath,
+                    `// Place your key bindings in this file to override the defaults\n[\n${newKeybindings}]`
```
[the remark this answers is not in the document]
leigaol: The comment has been removed.

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -0,0 +1,452 @@
...
+                            new vscode.Range(start, end),
+                            i,
+                            r.content,
+                            RecommendationHandler.instance.requestId,
+                            RecommendationHandler.instance.sessionId,
+                            TelemetryHelper.instance.triggerType,
+                            TelemetryHelper.instance.completionType,
+                            runtimeLanguageContext.getLanguageContext(document.languageId).language,
+                            r.references,
+                        ],
```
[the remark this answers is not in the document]
leigaol: Good suggestion. Can I fix this in a different refactor PR? Reasons are 
1. The acceptance command is referencing some variables like referenceLogViewProvider, referenceHoverProvider which were created in activation.ts. I need to first decouple the reference instances from activation.ts.
2. To do 1, I need to also update the previous onAcceptance code path, which is not within the scope of this new inline Code path PR. Changing onAcceptance code path will also make this PR hard to review.

src/codewhisperer/service/inlineCompletionService.ts
leigaol: As discussed in the description, when inlineCompletionAdditions API(https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.inlineCompletionsAdditions.d.ts#L10) is published, we can remove some states.

src/codewhisperer/service/inlineCompletionService.ts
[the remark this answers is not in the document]
JadenSimon: >  Can I fix this in a different refactor PR? 

Yeah separate PR makes sense to me.

src/codewhisperer/util/keybindingUtil.ts
```diff
@@ -0,0 +1,58 @@
...
+        getLogger().error(`Failed to update user keybindings, error ${error}`)
+    }
+}
+
+export function getKeybindingsPath(): string {
+    const home = os.homedir()
+    if (process.platform === 'win32') {
+        return path.join(home, CodeWhispererConstants.keyBindingPathWin)
+    } else if (process.platform === 'darwin') {
+        return path.join(home, CodeWhispererConstants.keyBindingPathMac)
```
JadenSimon, line 54: These paths might be outdated. The VS Code source uses [this](https://github.com/microsoft/vscode/blob/84c46b71a5d292a0a9c6cae60e850088270db37d/src/vs/platform/environment/node/userDataPath.js#L49) and [this](https://github.com/microsoft/vscode/blob/84c46b71a5d292a0a9c6cae60e850088270db37d/src/vs/platform/environment/common/environmentService.ts#L54)

Of course, we should think of alternative solutions at this point. Do you know if adding keybindings via `package.json` + custom `when` clause works? That would be more robust and user-friendly instead of writing to the `keybindings.json` file. We can do that in a separate PR, assuming it works.

src/codewhisperer/activation.ts
```diff
@@ -98,14 +102,17 @@ export async function activate(context: ExtContext): Promise<void> {
...
                     }
                 }
                 vscode.commands.executeCommand('aws.codeWhisperer.refresh')
             }
             if (configurationChangeEvent.affectsConfiguration('aws.codeWhisperer')) {
                 referenceLogViewProvider.update()
             }
+            if (configurationChangeEvent.affectsConfiguration('editor.inlineSuggest.enabled')) {
+                await vscode.commands.executeCommand('workbench.action.reloadWindow')
+            }
```
JadenSimon, line 115: We have to ask the user before reloading the window automatically. Could show a notification, something like: "Inline suggestion settings changed. The current window needs to be reloaded for this to take affect." with a button "Reload Now".

src/codewhisperer/activation.ts
```diff
@@ -232,6 +239,22 @@ export async function activate(context: ExtContext): Promise<void> {
         )
     }
 
+    function hideStatusBar() {
+        if (isInlineCompletionEnabled()) {
+            InlineCompletionService.instance.hideCodeWhispererStatusBar()
+        } else {
+            InlineCompletion.instance.hideCodeWhispererStatusBar()
+        }
```
JadenSimon, line 247: Are there two status bars now?

leigaol: Only one status bar.

JadenSimon: Okay, I see now that the goal is to keep the legacy implementation as isolated as possible so it can be cut in its entirety. 

> To avoid the extra cost of maintaining two code paths, in the long run, when most users have migrated to VSC >= 1.68.0, we can drop the previous inline experience by removing onAcceptance.ts and inlineCompletion.ts.

What's our threshold for this? Other than Cloud9, >95% of users are already on 1.68+.

leigaol: >keep the legacy implementation as isolated as possible so it can be cut in its entirety.

Yes.

>What's our threshold for this?

Not sure. This threshold is defined by Product. In my opinion, 95% is close but not good enough. 98% sounds good.

src/codewhisperer/activation.ts
leigaol: Added this button in https://github.com/aws/aws-toolkit-vscode/pull/2888/commits/bd730768b2a7e60147a88ecfb7974f103b0e8c32

src/codewhisperer/models/constants.ts
```diff
@@ -196,3 +196,8 @@ export const cloud9AccessStateKey = 'cloud9AccessStateKey'
 export const cloud9AccessSent = 'Access requested!'
 
 export const cloud9AccessAlreadySent = 'Access has already been requested, we are still processing it.'
+
+export const reloadWindow = 'Reload Now'
+
+export const reloadWindowPrompt =
+    'Inline suggestion settings changed. The current window needs to be reloaded for this to take affect.'
```
JadenSimon, line 203: Thinking we want to clarify that is this for CodeWhisperer  but the wording feels clunky
```suggestion
    'Inline suggestion settings changed. The current window needs to be reloaded for CodeWhisperer to use these changes.'
```

src/codewhisperer/util/keybindingUtil.ts
leigaol: >Do you know if adding keybindings via package.json + custom when clause works?

It does work. I tested https://github.com/aws/aws-toolkit-vscode/pull/2888/commits/b90b8cf44f2ed708db0cdef1c5902803e552239c

src/test/codewhisperer/commands/onInlineAcceptance.test.ts
```diff
@@ -0,0 +1,188 @@
...
+                    sessionId: '',
+                    triggerType: 'OnDemand',
+                    completionType: 'Line',
+                    language: 'python',
+                    references: fakeReferences,
+                },
+                extensionContext.globalState
+            )
+            const actualArg = trackerSpy.getCall(0).args[0]
+            assert.ok(trackerSpy.calledOnce)
```
JadenSimon, line 136: Would it be possible to use `CodeWhispererTracker.flush()` + `assertTelemetry` instead of a spy?

src/codewhisperer/service/inlineCompletionService.ts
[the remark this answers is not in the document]
leigaol: Changes are in https://github.com/leigaol/aws-toolkit-vscode/tree/refactorAcceptance. Will submit once this PR is merged.

JadenSimon (approved):

src/codewhisperer/models/constants.ts
```diff
@@ -196,3 +196,8 @@ export const cloud9AccessStateKey = 'cloud9AccessStateKey'
 export const cloud9AccessSent = 'Access requested!'
 
 export const cloud9AccessAlreadySent = 'Access has already been requested, we are still processing it.'
+
+export const reloadWindow = 'Reload Now'
```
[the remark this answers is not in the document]
leigaol: >And the way we do localization now is somewhat tedious

Agree.

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -0,0 +1,395 @@
...
+    hideCodeWhispererStatusBar() {
+        this.statusBar.hide()
+    }
+
+    isSuggestionVisible(): boolean {
+        return this.inlineCompletionProvider?.getActiveItemIndex !== undefined
+    }
+
+    isPaginationRunning() {
+        return this.statusBar.text === ` $(loading~spin)CodeWhisperer`
```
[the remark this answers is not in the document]
leigaol: I can fix it in a pagination refactor PR.

package.json
```diff
@@ -2958,6 +2958,16 @@
                 "key": "tab",
                 "mac": "tab",
                 "when": "editorTextFocus && CODEWHISPERER_SERVICE_ACTIVE"
+            },
+            {
+                "key": "right",
+                "command": "editor.action.inlineSuggest.showNext",
+                "when": "inlineSuggestionVisible && !editorReadonly"
```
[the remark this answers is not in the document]
leigaol: `CODEWHISPERER_SERVICE_ACTIVE` is specific to the legacy implementation

zixlin7 (approved):

src/codewhisperer/commands/onInlineAcceptance.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+export async function onInlineAcceptance(
+    acceptanceEntry: OnRecommendationAcceptanceEntry,
+    globalStorage: vscode.Memento
+) {
+    RecommendationHandler.instance.cancelPaginatedRequest()
+    InlineCompletionService.instance.disposeInlineCompletion()
+
+    if (acceptanceEntry.editor) {
+        sleep(CodeWhispererConstants.vsCodeCursorUpdateDelay)
```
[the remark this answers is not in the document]
leigaol: I added await in the latest commit.

src/test/codewhisperer/commands/onInlineAcceptance.test.ts
leigaol: This test case checks whether an object is enqueued into the tracker, however, what is does is almost duplicating with test case  `it('Should report telemetry that records this user decision event'` . So I removed this enqueue check test case.

Test case `it('Should report telemetry that records this user decision event'` already covers `assertTelemetry`.

src/codewhisperer/util/commonUtil.ts
```diff
@@ -27,6 +32,14 @@ export function isAwsError(error: any): boolean {
     )
 }
 
+export function isInlineCompletionEnabled() {
+    return (
+        semver.gte(vscode.version, '1.68.0') &&
+        vscode.workspace.getConfiguration('editor').get('inlineSuggest.enabled') &&
```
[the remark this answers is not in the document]
leigaol: The suggested change has been committed.

leigaol:
The CI failure is not relevant with this PR

> 1 failing
>   1) detectCdkProjects
>        detects deep projects:
>      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

JadenSimon (approved):

PR #2889, opened 2022-09-16
fix(codewhisperer): bug in handling closing brackets

zixlin7:
## Problem
the level of bracket needs to be greater than or equal to 0. When auto-triggering at a position before the opening `(`, it should be matched up with the closing `)` from right context.
## Solution
before:
Kapture 2022-09-16 at 11 01 24

after:

https://user-images.githubusercontent.com/60411978/190811164-f66720c4-0ca2-460a-b612-c075d03ebb9f.mov

## License


leigaol (approved):

JadenSimon (approved):
Would a bugfix changelog like this make sense? "CodeWhisperer duplicate closing bracket when using recommendation inside brackets"

PR #2893, opened 2022-09-19
feat(YAML): Watch untitled yaml files for cloudformation/sam contents

jpinkney-aws:
## Problem
- Currently, untitled yaml files for cloudformation/sam do not get a schema associated until the file is saved

## Solution
- Watch untitled yaml files for cloudformation/sam contents and apply the schema accordingly
- Allow the file watcher to be able to work with untitled yaml files

To test you can just create a new untitled file and make sure the language is yaml. Then add `AWSTemplateFormatVersion` and it should automatically start detecting the schema

## License

### DIFF
--- a/.changes/next-release/Feature-a4f4b3f0-0449-4e29-a8d1-f7d2c452ebc1.json
+++ b/.changes/next-release/Feature-a4f4b3f0-0449-4e29-a8d1-f7d2c452ebc1.json
@@ -1,4 +1,4 @@
 {
 	"type": "Feature",
-	"description": "Watch untitled yaml files for cloudformation/sam contents"
+	"description": "Enable syntax features for untitled (unsaved) cloudformation/sam yaml files"
 }
--- a/.changes/next-release/Feature-a4f4b3f0-0449-4e29-a8d1-f7d2c452ebc1.json
+++ b/.changes/next-release/Feature-a4f4b3f0-0449-4e29-a8d1-f7d2c452ebc1.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Watch untitled yaml files for cloudformation/sam contents"
+}
--- a/src/shared/schemas.ts
+++ b/src/shared/schemas.ts
@@ -12,11 +12,11 @@ import { FileResourceFetcher } from './resourcefetcher/fileResourceFetcher'
 import { getPropertyFromJsonUrl, HttpResourceFetcher } from './resourcefetcher/httpResourceFetcher'
 import { Settings } from './settings'
 import { once } from './utilities/functionUtils'
-import { normalizeSeparator } from './utilities/pathUtils'
 import { Any, ArrayConstructor } from './utilities/typeConstructors'
 import { AWS_SCHEME } from './constants'
 import { writeFile } from 'fs-extra'
 import { SystemUtilities } from './systemUtilities'
+import { normalizeVSCodeUri } from './utilities/vsCodeUtils'
 
 const GOFORMATION_MANIFEST_URL = 'https://api.github.com/repos/awslabs/goformation/releases/latest'
 const SCHEMA_PREFIX = `${AWS_SCHEME}://`
@@ -25,7 +25,7 @@ export type Schemas = { [key: string]: vscode.Uri }
 export type SchemaType = 'yaml' | 'json'
 
 export interface SchemaMapping {
-    path: string
+    uri: vscode.Uri
     type: SchemaType
     schema?: string | vscode.Uri
 }
@@ -103,7 +103,7 @@ export class SchemaService {
 
         const batch = this.updateQueue.splice(0, this.updateQueue.length)
         for (const mapping of batch) {
-            const { type, schema, path } = mapping
+            const { type, schema, uri } = mapping
             const handler = this.handlers.get(type)
             if (!handler) {
                 throw new Error(`no registered handler for type ${type}`)
@@ -112,7 +112,7 @@ export class SchemaService {
                 'schema service: handle %s mapping: %s -> %s',
                 type,
                 schema?.toString() ?? '[removed]',
-                path
+                uri
             )
             await handler.handleUpdate(mapping, this.schemas)
         }
@@ -309,11 +309,10 @@ export class YamlSchemaHandler implements SchemaHandler {
             this.yamlExtension = ext
         }
 
-        const path = vscode.Uri.file(normalizeSeparator(mapping.path))
         if (mapping.schema) {
-            this.yamlExtension.assignSchema(path, resolveSchema(mapping.schema, schemas))
+            this.yamlExtension.assignSchema(mapping.uri, resolveSchema(mapping.schema, schemas))
         } else {
-            this.yamlExtension.removeSchema(path)
+            this.yamlExtension.removeSchema(mapping.uri)
         }
     }
 }
@@ -356,6 +355,7 @@ export class JsonSchemaHandler implements SchemaHandler {
 
         let settings = this.getJsonSettings()
 
+        const path = normalizeVSCodeUri(mapping.uri)
         if (mapping.schema) {
             const uri = resolveSchema(mapping.schema, schemas).toString()
             const existing = this.getSettingBy({ schemaPath: uri })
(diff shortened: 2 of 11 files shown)


src/shared/schemas.ts
```diff
@@ -28,6 +28,7 @@ export interface SchemaMapping {
     path: string
```
JadenSimon, line 28: How much work would it be to make this URI instead of a file system path? Then we don't need `virtual`.

src/shared/utilities/vsCodeUtils.ts
```diff
@@ -71,3 +71,10 @@ export async function activateExtension<T>(
 export function promisifyThenable<T>(thenable: Thenable<T>): Promise<T> {
     return new Promise((resolve, reject) => thenable.then(resolve, reject))
 }
+
+export function isUntitledScheme(path: string | vscode.Uri): boolean {
+    if (typeof path === 'string') {
+        return path.startsWith('untitled:')
+    }
+    return path.scheme === 'untitled'
+}
```
JadenSimon, line 80: Personally I wouldn't accept a string here as it would encourage opaque URIs:
```suggestion
export function isUntitledDocument(document: Pick<vscode.TextDocument, 'uri'>): boolean {
    return document.uri.scheme === 'untitled'
}
```

src/shared/watchedFiles.ts
```diff
@@ -7,10 +7,11 @@ import * as vscode from 'vscode'
...
 import * as pathutils from './utilities/pathUtils'
 import * as path from 'path'
+import { isUntitledScheme } from './utilities/vsCodeUtils'
 
 export interface WatchedItem<T> {
     /**
-     * The absolute path to the file
+     * The path to the file
      */
     path: string
```
JadenSimon, line 16: Use URI instead? We shouldn't be using string paths anyway.

src/shared/watchedFiles.ts
```diff
@@ -133,6 +162,24 @@ export abstract class WatchedFiles<T> implements vscode.Disposable {
         }
     }
 
+    /**
+     * Load a virtual document (e.g. untitled documents) into the registry by their content
+     * @param uri the uri of the virtual document
+     * @param contents the contents of the virtual document
+     */
+    public async addVirtualItemToRegistry(uri: vscode.Uri, contents: string): Promise<void> {
```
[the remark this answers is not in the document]
jpinkney-aws: I mostly did it this way to avoid having to pass through the contents with something like:
```
addItemToRegistry(uri: vscode.Uri, quiet?: boolean, contents?: string)
```

I'll play around with it and see if I can simplify some things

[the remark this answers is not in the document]
JadenSimon: Two methods is ok. The bigger problem is that we are distinguishing 'virtual' and 'non-virtual' things. Though this is probably because `load` is an abstract method accepting a path when it should be a method called `parse` accepting the file contents.

JadenSimon (approved):

src/shared/utilities/vsCodeUtils.ts
```diff
@@ -71,3 +71,12 @@ export async function activateExtension<T>(
...
     return new Promise((resolve, reject) => thenable.then(resolve, reject))
 }
+
+export function isUntitledScheme(uri: vscode.Uri): boolean {
+    return uri.scheme === 'untitled'
+}
+
+// If the VSCode URI is untitled then return the string representation, otherwise normalize the filesystem path
+export function normalizeVSCodeUri(uri: vscode.Uri): string {
+    return isUntitledScheme(uri) ? uri.toString() : pathutils.normalize(uri.fsPath)
```
justinmk3, line 81: should we make a stronger check in the normalize() case, to verify that this is indeed a filesystem Uri?

src/shared/watchedFiles.ts
```diff
@@ -22,7 +23,7 @@ export interface WatchedItem<T> {
 /**
  * WatchedFiles lets us index files in the current registry. It is used
  * for CFN templates among other things. WatchedFiles holds a list of pairs of
- * the absolute path to the file along with a transform of it that is useful for
+ * the absolute path to the file or the path to an untitled file along with a transform of it that is useful for
```
justinmk3, line 26:
```suggestion
 * the absolute path to the file or "untitled:" URI along with a transform of it that is useful for
```

src/shared/watchedFiles.ts
```diff
@@ -93,6 +97,26 @@ export abstract class WatchedFiles<T> implements vscode.Disposable {
         await this.rebuild()
     }
 
+    /**
+     * Create a special watcher that operates only on untitled files.
+     * VSCode does not make it easy to watch untitled files and globs won't work without knowing exactly
```
justinmk3, line 102: "Watching" an in-memory file is just listening for the vscode `didChange` event (as done here). I wouldn't expect filesystem watchers to be involved. So the comment is kind of confusing.

```suggestion
     * To "watch" the in-memory contents of an untitled:/ file we just subscribe to `onDidChangeTextDocument`.
```

src/test/dynamicResources/awsResourceManager.test.ts
```diff
@@ -192,7 +192,7 @@ describe('ResourceManager', function () {
         const expectedSchemaLocation = path.join(tempFolder, 'sometype.schema.json')
         assert.ok(existsSync(expectedSchemaLocation))
         assert.strictEqual(mapping.type, 'json')
-        assert.strictEqual(mapping.path.toLowerCase(), editor.document.uri.fsPath.toLowerCase())
+        assert.strictEqual(mapping.uri.fsPath.toLowerCase(), editor.document.uri.fsPath.toLowerCase())
```
justinmk3, line 195: - while we're here would you mind changing this to use `pathUtils.areEqual()`
- same for the other case below

src/shared/utilities/vsCodeUtils.ts
```diff
@@ -76,7 +76,10 @@ export function isUntitledScheme(uri: vscode.Uri): boolean {
     return uri.scheme === 'untitled'
 }
 
-// If the VSCode URI is untitled then return the string representation, otherwise normalize the filesystem path
+// If the VSCode URI is not a file then return the string representation, otherwise normalize the filesystem path
 export function normalizeVSCodeUri(uri: vscode.Uri): string {
-    return isUntitledScheme(uri) ? uri.toString() : pathutils.normalize(uri.fsPath)
+    if (uri.scheme !== 'file') {
+        return uri.toString()
```
justinmk3, line 82: I don't think this will work for `untitled:/` scheme. What I meant to suggest, is something like this:

```
if isUntitledScheme(uri) {
    return uri.toString()
}
if <is-a-file> {
    return pathutils.normalize(uri.fsPath)
} else {
    throw ?
}
}
```

src/test/dynamicResources/awsResourceManager.test.ts
```diff
@@ -217,7 +220,9 @@ describe('ResourceManager', function () {
         // eslint-disable-next-line @typescript-eslint/unbound-method
         const [mapping] = capture(schemaService.registerMapping).last()
         assert.strictEqual(mapping.type, 'json')
-        assert.strictEqual(mapping.uri.fsPath.toString().toLowerCase(), editor.document.uri.fsPath.toLowerCase())
+        assert.ok(
+            pathutils.areEqual(undefined, mapping.uri.fsPath.toLowerCase(), editor.document.uri.fsPath.toLowerCase())
```
justinmk3, line 224: sorry, I meant `testUtil.assertEqualPaths()`

justinmk3 (approved):

src/test/dynamicResources/awsResourceManager.test.ts
```diff
@@ -192,7 +193,7 @@ describe('ResourceManager', function () {
         const expectedSchemaLocation = path.join(tempFolder, 'sometype.schema.json')
         assert.ok(existsSync(expectedSchemaLocation))
         assert.strictEqual(mapping.type, 'json')
-        assert.strictEqual(mapping.path.toLowerCase(), editor.document.uri.fsPath.toLowerCase())
+        testutil.assertEqualPaths(mapping.uri.fsPath.toLowerCase(), editor.document.uri.fsPath.toLowerCase())
```
justinmk3, line 196: toLowerCase() calls should no longer be needed

justinmk3 (approved):

.changes/next-release/Feature-a4f4b3f0-0449-4e29-a8d1-f7d2c452ebc1.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Watch untitled yaml files for cloudformation/sam contents"
```
justinmk3, line 3: As a user, it is hard to see the meaning of this feature description. Suggestion:

```suggestion
	"description": "Enable syntax features for untitled (unsaved) cloudformation/sam yaml files"
```

PR #2895, opened 2022-09-20
feat(CodeWhisperer) add sending serviceInvocationCount field in codeCoverageTracker

Will-ShaoHua:
### Requirement:
Tracker will increment `tracker.serviceInvocationCount` by 1 if there is a successful service invocation with specific programming language, i.e. if cwspr get invoked successfully with `python`, only python tracker will increment serviceInvocationCount.

## License


src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -188,10 +210,18 @@ export class CodeWhispererCodeCoverageTracker {
         }
     }
 
+    public activateTrackerIfNotActive(globals: vscode.Memento) {
+        const isTermsAccepted = globals.get<boolean>(CodeWhispererConstants.termsAcceptedKey) || false
+        if (this._isTrackerActive) return
+        if (!TelemetryHelper.instance.isTelemetryEnabled() || !isTermsAccepted) return
+        this._isTrackerActive = true
+    }
```
JadenSimon, line 218: I think this refactor had the right direction but could use some refinement in terms of usage. I'd put the `Memento` parameter back into the constructor and replace `_isTrackerActive` with this:
```suggestion
    public get isActive() {
        const isTermsAccepted = this.globals.get<boolean>(CodeWhispererConstants.termsAcceptedKey) || false
        
        return TelemetryHelper.instance.isTelemetryEnabled() && isTermsAccepted
    }
```

Then for ergonomics make `getTracker` use a default:
```ts
    public static getTracker(language: string, memento = globals.context.globalState): CodeWhispererCodeCoverageTracker | undefined {
```

The reasoning is that it's too easy to forget to call `activateTrackerIfNotActive` after calling `getTracker`.

Will-ShaoHua: noted, yea let me update this

leigaol:
Can you rename title to feat(codewisperer):...?

src/codewhisperer/activation.ts
```diff
@@ -421,10 +420,7 @@ export async function activate(context: ExtContext): Promise<void> {
                     }
                 }
 
-                CodeWhispererCodeCoverageTracker.getTracker(
-                    e.document.languageId,
-                    context.extensionContext.globalState
-                )?.countTotalTokens(e)
+                CodeWhispererCodeCoverageTracker.getTracker(e.document.languageId)?.countTotalTokens(e)
```
leigaol, line 423: `e.document.languageId` -> `runtimeLanguageContext.mapToRuntimeLanguage()` ?

e.document.languageId may contain jsx, javascript react

Will-ShaoHua: yep, sounds good let me fix it real quick

src/dynamicResources/model/supported_resources.json
```diff
@@ -1796,10 +1796,10 @@
         "documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lookoutvision-project.html"
     },
```
Will-ShaoHua, line 1797: no idea why I include changes on this file too

Will-ShaoHua: seems I already merged from mainline here https://github.com/aws/aws-toolkit-vscode/blob/master/src/dynamicResources/model/supported_resources.json#L389, but diff here doesn't work

JadenSimon: I think it's just `prettier`

JadenSimon:
I'm guessing the failures are because of differing behavior with the new inline suggestions class.

JadenSimon (approved):

leigaol (approved):

PR #2902, opened 2022-09-21
fix(codewhisperer): impove 'hover' override robustness for inline suggestions

JadenSimon:
## Problem
CodeWhisperer won't restore user settings if they reload/exit at certain times

## Solution
Use `globalState` + restore the setting on extension activation

## License


leigaol (approved):
Looks good!

justinmk3 (approved):

PR #2903, opened 2022-09-21
feat(codewhisperer): auto-trigger improvements

zixlin7:
## Problem
to improve our auto-trigger heuristic to be more aggressive at showing recommendations
## Solution
- remove key stroke based auto-trigger
- add new triggering based on user's idle time, we will trigger if there has been 2s since the last invocation and the previous request has finished

Inline api not enabled:

https://user-images.githubusercontent.com/60411978/192033831-71d269f3-6bba-4118-b22f-d31a830a7e3c.mov

inline api enabled:

https://user-images.githubusercontent.com/60411978/192033900-73fb747a-b315-4a84-984b-88f6ad0ce9ca.mov

cloud9:

https://user-images.githubusercontent.com/60411978/192033930-53e6d8d0-2def-48a9-9144-f86690acb9c6.mov

## License

### DIFF
--- a/src/codewhisperer/service/keyStrokeHandler.ts
+++ b/src/codewhisperer/service/keyStrokeHandler.ts
@@ -42,7 +42,7 @@ export class KeyStrokeHandler {
         return (this.#instance ??= new this())
     }
 
-    private startIdleTimeTriggerTimer(
+    public startIdleTimeTriggerTimer(
         event: vscode.TextDocumentChangeEvent,
         editor: vscode.TextEditor,
         client: DefaultCodeWhispererClient,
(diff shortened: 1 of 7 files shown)


src/codewhisperer/service/keyStrokeHandler.ts
```diff
@@ -193,6 +201,7 @@ export class KeyStrokeHandler {
                     RecommendationHandler.instance.isGenerateRecommendationInProgress = false
                 }
             } else if (isInlineCompletionEnabled()) {
+                if (InlineCompletionService.instance.isPaginationRunning()) return
```
leigaol, line 204: This if is not necessary because .getPaginatedRecommendation contains this if

src/codewhisperer/service/keyStrokeHandler.ts
```diff
@@ -42,6 +42,27 @@ export class KeyStrokeHandler {
...
+        editor: vscode.TextEditor,
+        client: DefaultCodeWhispererClient,
+        config: ConfigurationEntry
+    ) {
+        if (this.idleTriggerTimer) {
+            clearInterval(this.idleTriggerTimer)
+            this.idleTriggerTimer = undefined
+        }
+        this.idleTriggerTimer = setInterval(() => {
+            const duration = Math.floor((performance.now() - RecommendationHandler.instance.lastInvocationTime) / 1000)
```
leigaol, line 56: is `RecommendationHandler.instance.lastInvocationTime ` accurate to describe the last user input time?

zixlin7: this is not for input time, this is to ensure 2s since the lastInvocation, user input is handed by key stroke, so if there is more key stoke we cancel previous interval.

src/codewhisperer/service/keyStrokeHandler.ts
zixlin7: done, thank!

src/codewhisperer/service/keyStrokeHandler.ts
```diff
@@ -42,6 +42,28 @@ export class KeyStrokeHandler {
...
+            if (duration < CodeWhispererConstants.invocationTimeIntervalThreshold) return
+            if (isCloud9() && RecommendationHandler.instance.isGenerateRecommendationInProgress) return
+            if (isInlineCompletionEnabled() && InlineCompletionService.instance.isPaginationRunning()) return
+            if (InlineCompletion.instance.getIsActive || InlineCompletion.instance.isPaginationRunning()) return
+            this.invokeAutomatedTrigger('IdleTime', editor, client, config)
+            if (this.idleTriggerTimer) clearInterval(this.idleTriggerTimer)
+            this.idleTriggerTimer = undefined
```
JadenSimon, line 63: If you want to be extra defensive:
```suggestion
            try {
                this.invokeAutomatedTrigger('IdleTime', editor, client, config)
            } finally {
                 if (this.idleTriggerTimer) {
                     clearInterval(this.idleTriggerTimer)
                     this.idleTriggerTimer = undefined
                 }
            }
            })
```

src/test/codewhisperer/service/keyStrokeHandler.test.ts
```diff
@@ -100,19 +98,6 @@ describe('keyStrokeHandler', function () {
             assert.ok(!invokeSpy.called)
         })
 
-        it('Should not call invokeAutomatedTrigger if previous text input is within 2 seconds and it is not a specialcharacter trigger \n', async function () {
```
JadenSimon, line 103: We're removing a decent number of tests without any new tests to cover for them. Are these deleted tests covered by something else? If not, is there any way to extract out some of the logic in the `idleTriggerTimer` callback and test the same/similar behavior?

zixlin7: added tests

zixlin7:
[Please hold on merging this PR, thanks!]

PR #2904, opened 2022-09-21
refactor(codewhisperer): Simplify pagination 

leigaol:
## Problem
Pagination logic in CodeWhisperer is unnecessarily complex. See https://github.com/aws/aws-toolkit-vscode/pull/2888#discussion_r973236349

## Solution

Simplify pagination logic.

## License

### DIFF
--- a/src/codewhisperer/service/inlineCompletionService.ts
+++ b/src/codewhisperer/service/inlineCompletionService.ts
@@ -174,11 +174,15 @@ export class InlineCompletionService {
     private hide: vscode.Disposable
     private next: vscode.Disposable
     private prev: vscode.Disposable
+    private _isPaginationRunning = false
 
     constructor() {
         this.prev = prevCommand.register(this)
         this.next = nextCommand.register(this)
         this.hide = hideCommand.register(this)
+        RecommendationHandler.instance.onDidReceiveRecommendation(e => {
+            this.tryShowRecommendation()
+        })
     }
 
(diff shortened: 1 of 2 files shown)


src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -174,11 +174,15 @@ export class InlineCompletionService {
...
     private prev: vscode.Disposable
+    private _isPaginationRunning = false
 
     constructor() {
         this.prev = prevCommand.register(this)
         this.next = nextCommand.register(this)
         this.hide = hideCommand.register(this)
+        RecommendationHandler.instance.onDidReceiveRecommendation(e => {
+            this.tryShowRecommendation()
+        })
```
JadenSimon, line 185: Making this event-based is a decent strategy _if_ multiple things could potentially be mutating `RecommendationHandler` _or_ if `tryShowRecommendation` is only responsible for UI. Right now, this change doesn't simplify things too much other than creating smaller functions, but maybe if we start moving more responsibility to `RecommendationHandler` it might?

This refactor might actually be quite a bit of work because most of the problem is with all the cross-talk, not necessarily the pagination. We'd likely need to change code in _many_ places. So I'm fine with merging as-is.

JadenSimon (approved):

leigaol:
The CI failure is irrelevant to this PR.

Issue #2818, opened 2022-08-10
ECS exec: user can cancel long-running commands


justinmk3:
## Problem

If ecs exec is used to run a command that takes too long or hangs, Toolkit does not provide a way to cancel it.

https://github.com/aws/aws-toolkit-vscode/issues/2607#issuecomment-1210098083 : 

> if you accidentally run a continuous command there's no way to stop it other than stopping the task. 

ref #2607

## Proposal

- Surface ecs exec tasks via a command and/or in AWS Explorer. 
- User can cancel tasks that were started from the Toolkit ecs exec feature.

justinmk3:
Fixed in 1.51

Issue #2607, opened 2022-05-02
ECS exec: support interactive (terminal) commands


pedrovanzella:
**Is your feature request related to a problem? Please describe.**

ECS Execute-Command already uses the `--interactive` flag. However, the command execution goes to an Output (read-only) pane. We often want to interact with a REPL inside the container (such as the Rails console), so starting a terminal process would be best.

**Describe the solution you'd like**

When running a command on an ECS task, start a terminal process instead.

ckonefke:
This feature is definitely needed.  I found myself very jealous when I saw my co-worker with pycharm/JetBrains already has this ability.  Single, non-interactive commands are very limited, and if you accidentally run a continuous command there's no way to stop it other than stopping the task.  You could run another command to find the pid, and another to kill the pid, but can't we just have a CTRL-C?

justinmk3:
> if you accidentally run a continuous command there's no way to stop it other than stopping the task.

@ckonefke that's a good point, I created https://github.com/aws/aws-toolkit-vscode/issues/2818 to track this.

justinmk3:
This feature was included in AWS Toolkit 1.51. 

## `Open Terminal` on ECS tasks

Use the `Open Terminal...` command on ECS tasks to start an interactive terminal connected to the task container.

<img width="237" alt="image" src="https://user-images.githubusercontent.com/55561878/194167386-83931388-69da-4f63-b7a7-90d296fb56a2.png">

## Configure ECS settings

Look for "ecs" in VSCode settings to change the default command (shell) that is run by the `Open Terminal...` command.

<img width="546" alt="image" src="https://user-images.githubusercontent.com/55561878/194167674-a8eb57bc-b7eb-436e-8f29-a4f59b57264c.png">

PR #2905, opened 2022-09-21
feat(ecs): open task session in interactive terminal

JadenSimon:
## Problem
* https://github.com/aws/aws-toolkit-vscode/issues/2607
* https://github.com/aws/aws-toolkit-vscode/issues/2818

## Solution
* Fix ECS icons so they render correctly as a font (VS Code changed things recently)
* Refactor a few things to remove coupling to `AWSTreeNodeBase`
* Add new command that starts a terminal using SSM (available as an option on a container node)
* Add cancellable progress message when running a command 
* Terminates the SSM session when the local process exits 

Open questions:
* The terminal codepath uses `/usr/bin/env bash` as a hard-coded command, should this be a setting? (Probably...)
* Is there a way to not have the terminal close when a one-off command terminates? If not, we need to continue using the output channel

## License


package.nls.json
```diff
@@ -101,6 +102,7 @@
     "AWS.ecs.enableEcsExec": "Enable Command Execution",
     "AWS.ecs.disableEcsExec": "Disable Command Execution",
     "AWS.ecs.runCommandInContainer": "Run command in container",
+    "AWS.ecs.openTaskInTerminal": "Open interactive session in terminal",
```
justinmk3, line 105: - can we just name this "Open Terminal"? That also somewhat matches vscode's own wording.
- this is an action, right? So I guess it needs to be Title Case (same for the one above it)


```suggestion
    "AWS.ecs.openTaskInTerminal": "Open Terminal",
```

src/ecs/commands.ts
```diff
@@ -0,0 +1,182 @@
...
+async function withoutShellIntegration<T>(cb: () => T | Promise<T>): Promise<T> {
+    const userValue = Settings.instance.get('terminal.integrated.shellIntegration.enabled', Boolean)
+
+    try {
+        await Settings.instance.update('terminal.integrated.shellIntegration.enabled', false)
+        return await cb()
+    } finally {
+        Settings.instance.update('terminal.integrated.shellIntegration.enabled', userValue)
+    }
+}
```
JadenSimon, line 149: This hack is my biggest concern right now as there's a slight risk of leaving the setting `false`. Without it, the session token gets logged by VS Code. Need to open an issue with them as they shouldn't be logging arguments from extensions.

src/shared/clients/iamClient.ts
```diff
@@ -52,28 +51,14 @@ export class DefaultIamClient {
...
-        } catch (error) {
-            getLogger().error('iam: Error during policy simulation. Skipping permissions check. Error: %O', error)
-            return undefined
+    public async getDeniedActions(request: IAM.SimulatePrincipalPolicyRequest): Promise<IAM.EvaluationResult[]> {
+        const permissionResponse = await this.simulatePrincipalPolicy(request)
+        if (!permissionResponse.EvaluationResults) {
+            throw new Error('No evaluation results found')
         }
+
+        return permissionResponse.EvaluationResults.filter(r => r.EvalDecision !== 'allowed')
```
justinmk3, line 61: 👍  always a good idea to return more info instead of just a boolean, if it is cheap to do so. Allows callers to make better decisions and also surface better info to the user.

src/shared/treeview/resource.ts
```diff
@@ -17,37 +23,234 @@ export interface Resource {
...
+    private async loadFrom(loader: PageLoader<TreeNode<U>>): Promise<TreeNode<U>[]> {
+        const pages = await loader.get()
+        const merged = pages.reduce((a, b) => a.concat(b), [])
+
+        return merged
+    }
+
+    private withLoadMoreNode(loader: PageLoader<TreeNode<U>>, result: TreeNode<U>[]): [...typeof result, TreeNode] {
+        // TODO: optimize this by making `loadMore` apart of the `ResourceTreeNode` prototype?
+        // but then we can't make the method private :/
```
justinmk3, line 212: why is private important? if we think there is a risk of caller mistakes, we could add a lint rule in the future.

src/ecs/commands.ts
```diff
@@ -0,0 +1,182 @@
...
+                globals.outputChannel
+            )
+
+            const proc = new ChildProcess(path, args, { logging: 'noparams' })
+            await proc
+                .run({
+                    timeout,
+                    rejectOnError: true,
+                    rejectOnErrorCode: true,
+                    // TODO: `showOutputMessage` should not be writing to the logs...
```
justinmk3, line 114: I assume the intention there was to get visibility in CI. But it's probably better to explicitly call getLogger() instead of trying to be clever.

src/shared/treeview/resource.ts
JadenSimon: Was mainly because I had the data source decoupled from the consumer, but I might as well combine them.

package.nls.json
```diff
@@ -100,7 +101,8 @@
     "AWS.command.ecr.deleteTag": "Delete Tag...",
     "AWS.ecs.enableEcsExec": "Enable Command Execution",
     "AWS.ecs.disableEcsExec": "Disable Command Execution",
-    "AWS.ecs.runCommandInContainer": "Run command in container",
+    "AWS.ecs.runCommandInContainer": "Run Command in Container",
+    "AWS.ecs.openTaskInTerminal": "Open Task in Terminal",
```
justinmk3, line 105: I strongly suspect that "Open Terminal" will be more intuitive than "Open Task in Terminal". As a developer, the latter sounds like a new/unfamiliar feature, unrelated to the intuitive idea of "I want a terminal here". People viewing this list aren't necessarily going to remember which node is a "container" or a "task" or "...". But they do know intuitively that right-clicking and choosing "Open Terminal" will probably put a terminal on whatever they right-clicked.

justinmk3 (approved):

package.nls.json
justinmk3: Isn't `Run Command in Container` operating on the same kind of object? There also I'd suggest we rename it to `Run Command...`

PR #2908, opened 2022-09-26
feat(dev): "Show Global State"

justinmk3:
## Problem

Toolkit developers, and customers who are troubleshooting, need a way to see the various global state that Toolkit depends on.

## Solution

Introduce `Show Global State` item in the "AWS (Developer)" menu. Currently only shows env vars, but can be expanded later.

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -1060,7 +1060,7 @@
                 },
                 {
                     "command": "aws.dev.openMenu",
-                    "when": "aws.isDevMode"
+                    "when": "aws.isDevMode || isCloud9"
                 }
             ],
             "editor/title": [
--- a/src/dev/activation.ts
+++ b/src/dev/activation.ts
@@ -21,10 +21,10 @@ interface MenuOption {
 }
 
 /**
- * Currently contains all known developer tools.
+ * Defines AWS Toolkit developer tools.
  *
  * Options are displayed as quick-pick items. The {@link MenuOption.executor} callback is ran
- * if the user selects an option. There is no support for name-spacing. Just add the relevant
+ * on selection. There is no support for name-spacing. Just add the relevant
  * feature/module as a description so it can be moved around easier.
  */
 const menuOptions: Record<string, MenuOption> = {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1060,7 +1060,7 @@
                 },
                 {
                     "command": "aws.dev.openMenu",
-                    "when": "aws.isDevMode"
+                    "when": "aws.isDevMode || isCloud9"
```
jpinkney-aws, line 1063: Is this just to make customers lives easier?

src/dev/activation.ts
```diff
@@ -228,4 +247,13 @@ async function openStorageFromInput() {
     }
 }
 
+async function showGlobalState() {
+    // const document = await vscode.workspace.openTextDocument('untitled:')
```
jpinkney-aws, line 251: Do we still need these commented things?

jpinkney-aws (approved):
This will be super nice to help diagnose more issues. I wonder if we should also print things like ssh version, aws cli version (and any other cli versions we have) in here as well?

src/dev/activation.ts
```diff
@@ -34,6 +34,24 @@ const menuOptions: Record<string, MenuOption> = {
...
+    },
+}
+
+export class GlobalStateDocumentProvider implements vscode.TextDocumentContentProvider {
+    provideTextDocumentContent(uri: vscode.Uri): string {
+        let s = 'Environment variables known to AWS Toolkit:\n'
+        for (const k of Object.keys(process.env)) {
+            const v = process.env[k]
+            s += `${k}=${v}\n`
+        }
```
JadenSimon, line 51:
```suggestion
        for (const [k, v] of Object.entries(process.env)) {
            s += `${k}=${v}\n`
        }
```

src/dev/activation.ts
```diff
@@ -228,4 +247,13 @@ async function openStorageFromInput() {
     }
 }
 
+async function showGlobalState() {
+    // const document = await vscode.workspace.openTextDocument('untitled:')
+    const uri = vscode.Uri.parse('aws-dev2:global-state')
+    const doc = await vscode.workspace.openTextDocument(uri)
+    // const withLanguage = await vscode.languages.setTextDocumentLanguage(document, 'json')
+    // const editor = await vscode.window.showTextDocument(withLanguage)
```
JadenSimon, line 255:
```suggestion
```

src/dev/activation.ts
JadenSimon, line 251:
```suggestion
```

JadenSimon (approved):

package.json
justinmk3: yes

justinmk3:
>  I wonder if we should also print things like ssh version, aws cli version (and any other cli versions we have) in here as well?

Sure. Those can/should also go in the "About" dialog.

PR #2912, opened 2022-09-28
build(scripts): move client generation to its own npm script

JadenSimon:
## Problem
No standalone script to generate service client type definitions

## Solution
Split the command out into `generateClients` and remove it from `postinstall`

## License


jpinkney-aws (commented):
So now that generating service clients is out of the postinstall, is there a chance that we/any contributor might forget to do it? Would that break the build and have people wondering what they might have to do to correct the build?

JadenSimon:
> So now that generating service clients is out of the postinstall, is there a chance that we/any contributor might forget to do it? Would that break the build and have people wondering what they might have to do to correct the build?

It's still ran before any build so no. Code generation for building the package should really not be in `postinstall`.

justinmk3 (approved):

PR #2913, opened 2022-09-29
refactor(credentials): add assertion utils and clean-up SSO implementation

JadenSimon:
## Problem
* Duplicate 'missing props' implementations
* SSO caching hard to track; no logging
* Cache(s) not evicted on client faults

## Solution
* Add several utils for checking or asserting that an object has certain properties
* Add a basic factory for keyed caches on disk  
* Refactor SSO implementation to use the new utils

## License


src/credentials/sso/model.ts
```diff
@@ -0,0 +1,77 @@
...
+    readonly scopes?: string[]
+    readonly identifier?: string
+}
+
+export async function openSsoPortalLink(authorization: { readonly verificationUriComplete: string }): Promise<boolean> {
+    return vscode.env.openExternal(vscode.Uri.parse(authorization.verificationUriComplete))
+}
+
+// Most SSO 'expirables' are fairly long lived, so a one minute buffer is plenty.
+const EXPIRATION_BUFFER_MS = 60000
```
JadenSimon, line 74: This constant was in the old impl. 
Though maybe we don't need it

src/credentials/providers/credentials.ts
```diff
@@ -113,7 +113,7 @@ export interface CredentialsProvider {
      * first use (in particular, credentials that may prompt, such as SSO/MFA,
      * should _not_ attempt to auto-connect).
      */
-    canAutoConnect(): boolean
+    canAutoConnect(): boolean | Promise<boolean>
```
justinmk3, line 116: hmm.. does this mean the caller has to add `await` depending on the implementation?

src/credentials/sso/clients.ts
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        return requester(request as T)
+    }
+
+    private async handleError(error: unknown): Promise<never> {
+        if (
+            error instanceof SSOServiceException &&
+            error.$fault === 'client' &&
+            !(isThrottlingError(error) || isTransientError(error))
```
justinmk3, line 180: interesting... should we mention `@aws-sdk/service-error-classification` somewhere in our guidelines, or maybe in ToolkitError class description?

justinmk3 (approved):

src/credentials/sso/clients.ts
JadenSimon: I'll put a util function in `errors.ts`, maybe `isClientFault`.

src/credentials/providers/credentials.ts
JadenSimon: No it's ok to `await` a non-promise. Only downside is that tests/logic can be written towards specific instances of the interface which may have the narrowed synchronous type. And if everything is inferred and/or non-strict equality is used then you can introduce bugs if instance type is widened.

I'm gonna change it to only be `Promise<boolean>` as it's too easy to break things when doing this with `boolean`.

jpinkney-aws (approved):

PR #2914, opened 2022-09-30
fix(codewhisperer): more inline suggestion "hover config" updates

JadenSimon:
## Problem
Restoration of the setting is unreliable

## Solution
* Place callsites in `finally` blocks
* Use an interval instead of a timer
    * Now checks every 5 seconds instead of 10  

## License


src/codewhisperer/commands/onInlineAcceptance.ts
```diff
@@ -59,7 +59,12 @@ export const acceptSuggestion = Commands.declare(
                     references,
                 },
                 context.extensionContext.globalState
-            )
+            ).finally(async () => {
+                if (isInlineCompletionEnabled()) {
```
leigaol, line 63: We probably don't need to check isInlineCompletionEnabled() here because this entire code path is under `isInlineCompletionEnabled`'s logical branch.

leigaol (approved):

leigaol:
Thanks @JadenSimon for making this change.

The CI failure is not relevant to changes in this PR.

justinmk3 (approved):

Issue #2915, opened 2022-09-30
ECS error: "InvalidParameterException: Invalid identifier: Identifier is for cluster "


justinmk3:
steps to reproduce:

1. try to expand ECS cluster to show running services.
2. error:
    - <img width="325" alt="image" src="https://user-images.githubusercontent.com/55561878/193358133-8921f6bd-6657-4c73-91e0-2602db642941.png">

seems to fail here:

https://github.com/aws/aws-toolkit-vscode/blob/742a19c81b4d95e98e3e82308541b6be9d28ef0c/src/shared/treeview/resource.ts#L46-L49

```
2022-09-30 14:19:50 [ERROR]: [InvalidParameterException: Invalid identifier: Identifier is for cluster jan7. Your cluster is default
	at Request.extractError (/Volumes/workplace/aws-toolkit-vscode/node_modules/aws-sdk/lib/protocol/json.js:52:27)
	at Request.callListeners (/Volumes/workplace/aws-toolkit-vscode/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
	at Request.emit (/Volumes/workplace/aws-toolkit-vscode/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
	at Request.emit (/Volumes/workplace/aws-toolkit-vscode/node_modules/aws-sdk/lib/request.js:686:14)
	at Request.transition (/Volumes/workplace/aws-toolkit-vscode/node_modules/aws-sdk/lib/request.js:22:10)
	at AcceptorStateMachine.runTo (/Volumes/workplace/aws-toolkit-vscode/node_modules/aws-sdk/lib/state_machine.js:14:12)
	at /Volumes/workplace/aws-toolkit-vscode/node_modules/aws-sdk/lib/state_machine.js:26:10
	at Request.<anonymous> (/Volumes/workplace/aws-toolkit-vscode/node_modules/aws-sdk/lib/request.js:38:9)
	at Request.<anonymous> (/Volumes/workplace/aws-toolkit-vscode/node_modules/aws-sdk/lib/request.js:688:12)
```

justinmk3:
confirmed that before https://github.com/aws/aws-toolkit-vscode/pull/2905 , I'm able to expand services/tasks:

<img width="341" alt="image" src="https://user-images.githubusercontent.com/55561878/193358612-7d224610-8e01-4231-834e-c14e34ba7d07.png">

justinmk3:
[...]

PR #2916, opened 2022-09-30
fix(ecs): update `Copy ARN` command and region node

JadenSimon:
## Problem
* The "Copy ARN" command doesn't work with `TreeShim`
* `RegionNode` caches all service nodes
* Can't list services for non-default ECS cluster (#2915)

## Solution
* Fix the copy name/ARN commands
    * Added tests for this
* Change `RegionNode` to lazily instantiate service nodes
* Add cluster ARN to `listServices` call

Also added more documentation for `ResourceTreeNode`

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
                 },
                 "aws.ecs.openTerminalCommand": {
                     "type": "string",
-                    "default": "/usr/bin/env bash",
+                    "default": "/bin/sh",
                     "markdownDescription": "%AWS.configuration.description.ecs.openTerminalCommand%"
                 },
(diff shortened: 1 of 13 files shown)


justinmk3 (approved):

justinmk3:
For reference, this does not seem to fix https://github.com/aws/aws-toolkit-vscode/issues/2915

JadenSimon:
> For reference, this does not seem to fix https://github.com/aws/aws-toolkit-vscode/issues/2915

Yeah I hadn't noticed that one. Optional parameters :/

PR #2921, opened 2022-10-05
fix(codewhisperer): only send one empty user decision per pagination session

leigaol:
## Problem
userDecision events with Empty recommendation state in CodeWhisperer is sent more than expected. This event should only be sent when all paginated responses are Empty list of recommendation

## Solution
Only send one empty user decision per pagination session

No changelog item is needed for this PR.

## License

### DIFF
--- a/src/codewhisperer/service/recommendationHandler.ts
+++ b/src/codewhisperer/service/recommendationHandler.ts
@@ -283,7 +283,7 @@ export class RecommendationHandler {
             this.recommendations = isCloud9() ? recommendation : this.recommendations.concat(recommendation)
             this._onDidReceiveRecommendation.fire()
         }
-
+        // send Empty userDecision event only when all paginated request finishes
         if (this.recommendations.length === 0 && nextToken === '') {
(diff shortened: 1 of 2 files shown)


src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -282,7 +282,9 @@ export class RecommendationHandler {
             })
             this.recommendations = isCloud9() ? recommendation : this.recommendations.concat(recommendation)
             this._onDidReceiveRecommendation.fire()
-        } else {
+        }
+        // send Empty userDecision event only when all paginated request finishes
```
andrewyuq, line 286: The description should actually be: 
send Empty userDecision event if user receives no recommendations in this session at all.

andrewyuq (commented):
Also it would be good if we can have tests

src/codewhisperer/service/recommendationHandler.ts
leigaol: Updated with new description

andrewyuq (approved):

JadenSimon (approved):

PR #2923, opened 2022-10-07
tests(explorer): add utils and examples using ECS

JadenSimon:
## Problem
* Current explorer tree tests are verbose and redundant
* No tests for ECS explorer tree

## Solution
* Add some utils
* Add simple tests

The tedious part for writing these tests is collecting example data. A follow-up PR will address that.

## License

### DIFF
--- a/src/test/shared/treeview/testUtil.ts
+++ b/src/test/shared/treeview/testUtil.ts
@@ -1,0 +1,43 @@
+/*!
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+import * as assert from 'assert'
+import { selectFrom, keys } from '../../../shared/utilities/tsUtils'
+import { TreeNode } from '../../../shared/treeview/resourceTreeDataProvider'
+
+type Nodeable = { toTreeNode: () => TreeNode }
+type Node = TreeNode | Nodeable
+type Matcher = string | { [P in keyof vscode.TreeItem]: vscode.TreeItem[P] }
+
+const resolveNode = (node: Node) =>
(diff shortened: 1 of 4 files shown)


src/test/utilities/collectionUtils.ts
```diff
@@ -3,7 +3,19 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { AsyncCollection, toCollection } from '../../shared/utilities/asyncCollection'
+
 // TODO: what is the point of this? should it live in src/shared/utilities/collectionUtils.ts ?
```
JadenSimon, line 8: I guess it's just a cleaner way to write out this:
```ts
(async function* () { yield* items })()
```

Which is not particularly useful for anything but testing

src/test/shared/treeview/testUtil.ts
```diff
@@ -0,0 +1,43 @@
+/*!
```
justinmk3, line 1: Maybe all of the "test util" file should live in once place, could be `src/test/shared/` for now. Else it's hard to discover what's available, unless one guesses that a project-wide search for "testUtil" will get most of them.

justinmk3 (approved):

src/test/shared/treeview/testUtil.ts
JadenSimon: Yeah combining all the important functions into a single file could work out ok if given more verbose names. Might look a bit messy though. Could also re-export from a single file?

justinmk3: Not a single file, just a single folder :) Right now, we have testUtils in various folders.

jpinkney-aws (approved):

src/test/shared/treeview/testUtil.ts
JadenSimon: I'll follow up on this in a separate PR.

PR #2930, opened 2022-10-11
fix(codewhisperer): Fix duplicate userDecision telemetry accept

leigaol:
## Problem
When accept a reco, userDecision telemetry was sent twice. In release 1.51.0, telemetry acceptance userDecision events will be sent twice. 

## Solution
Remove recommendations before they are sent twice.

## License

### DIFF
--- a/src/codewhisperer/commands/onInlineAcceptance.ts
+++ b/src/codewhisperer/commands/onInlineAcceptance.ts
@@ -62,6 +62,7 @@ export const acceptSuggestion = Commands.declare(
             ).finally(async () => {
                 if (isInlineCompletionEnabled()) {
                     // at the end of recommendation acceptance, clear recommendations.
+                    RecommendationHandler.instance.clearRecommendations()
                     await InlineCompletionService.instance.clearInlineCompletionStates(editor)
                 }
             })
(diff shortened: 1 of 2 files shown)


src/codewhisperer/commands/onInlineAcceptance.ts
```diff
@@ -154,4 +154,7 @@ export async function onInlineAcceptance(
             )
         }
     }
+
+    // at the end of recommendation acceptance, clear recommendations.
+    RecommendationHandler.instance.clearRecommendations()
```
JadenSimon, line 159: Can this be placed on L63: https://github.com/aws/aws-toolkit-vscode/blob/1d781ae4c130c159f28cfb63f9cea7684edfc59b/src/codewhisperer/commands/onInlineAcceptance.ts#L63-L66 ?

That way we handle failures

JadenSimon (approved):

zixlin7 (approved):

src/codewhisperer/commands/onInlineAcceptance.ts
leigaol: Yes

JadenSimon (approved):

PR #2941, opened 2022-10-19
feat(telemetry): add awsRegion in commonMetadata

justinmk3:
This is unblocked since we have `guessDefaultRegion()` now.

related: a50b83a833dacf95d29216c7e446c67556cf7925

## License

### DIFF
--- a/src/shared/telemetry/telemetryLogger.ts
+++ b/src/shared/telemetry/telemetryLogger.ts
@@ -14,9 +14,9 @@ export interface MetricQuery {
     readonly metricName: string
 
     /**
-     * Attributes to filter out of the metadata
+     * Exclude metadata items matching these keys.
      */
-    readonly filters?: string[]
+    readonly excludeKeys?: string[]
 }
 
 interface Metadata {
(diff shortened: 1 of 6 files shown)


JadenSimon (approved):
This will probably break tests. You'll need to update `assertTelemetry` in `src/test/testUtil.ts` to exclude this field.

src/shared/telemetry/telemetryLogger.ts
```diff
@@ -14,9 +14,9 @@ export interface MetricQuery {
     readonly metricName: string
 
     /**
-     * Attributes to filter out of the metadata
+     * Exclude metadata items matching these keys.
      */
-    readonly filters?: string[]
+    readonly excludeKeys?: string[]
```
justinmk3, line 19: Renamed because "filters" is ambiguous: it can mean include or exclude. In `window.showSaveDialog()` it means "include" :)

JadenSimon (approved):

src/shared/telemetry/telemetryLogger.ts
JadenSimon: Yeah I think I originally had the field as a callback or something to align with `Array.filter`. Tbh I'm still confused by `filter` even after using it so much...

PR #2944, opened 2022-10-20
fix(codewhisperer): latency caused by credentials

zixlin7:
## Problem
fix the latency introduced by not cacheing the credentials
## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2946, opened 2022-10-21
fix(codewhisperer): latency caused by credentials

zixlin7:
## Problem
same as #2944 , accidentally deleted the fork
## License

### DIFF
--- a/src/codewhisperer/client/codewhisperer.ts
+++ b/src/codewhisperer/client/codewhisperer.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { AWSError, Service } from 'aws-sdk'
+import { AWSError, CognitoIdentityCredentials, Service } from 'aws-sdk'
 import apiConfig = require('./service-2.json')
 import globals from '../../shared/extensionGlobals'
 import * as CodeWhispererClient from './codewhispererclient'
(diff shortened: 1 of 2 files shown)


andrewyuq (approved):

leigaol (approved):

src/codewhisperer/client/codewhisperer.ts
```diff
@@ -39,8 +39,23 @@ export type ListCodeScanFindingsRequest = Readonly<CodeWhispererClient.ListCodeS
...
 export class DefaultCodeWhispererClient {
+    private credentials?: CognitoIdentityCredentials
+
     private async createSdkClient(): Promise<CodeWhispererClient> {
-        const credentials = !isCloud9() ? await getCognitoCredentials() : undefined
+        try {
+            if (!this.credentials && !isCloud9()) {
+                this.credentials = await getCognitoCredentials()
+            }
+            if (this.credentials && this.credentials?.needsRefresh()) {
```
justinmk3, line 49: `this.credentials` is always true now?

justinmk3 (approved):

src/codewhisperer/client/codewhisperer.ts
zixlin7: if it's Cloud9 it should still be undefined

leigaol (approved):
This is a nice catch. Should not report serviceInvocation events if the API call failed at getCognition API

JadenSimon (approved):

PR #2899, opened 2022-09-20
fix(codewhisperer) update autotrigger filtering logic and fix enter key autotrigger blocked

Will-ShaoHua:
## Problem
- Refactor keyStrokeHandler 
- fix bug: enter key auto-trigger was blocked by PR https://github.com/aws/aws-toolkit-vscode/pull/2886

## Observation and Definition
- DeleteKey: empty string. e.g. `''`
- EnterKey: 1 new line char \n possibly followed by space(s) char. e.g. `'\n'`, `'\n    '`
- TabKey: the length of string should have space only and be a multiple of tabSize defined in the IDE or 4 by default e.g. `'    '`
- SpecialChar: char(s) like (, [, { etc.
- isSingleLine: if the number of new line char in the string  == 0, the only exception is EnterKey which has exact 1 new line char followed by spaces
- isMultiLine: if the number of new line char in the string > 0, we will block cwspr service invocation

1. Parenthesis (+ closing paren generated by IDE) will generate ONLY 1 event 

https://user-images.githubusercontent.com/96078566/191853918-652b9c46-a624-4cb9-add0-4b843588d783.mov

2. Most cases, reformatting will be generate > 1 events, but 1 is still possible if there is only one line reformat

MultiLine-reformat

https://user-images.githubusercontent.com/96078566/191854249-2466f901-7813-4321-abd3-fd89ddae1b67.mov

SingleLine-reformat

https://user-images.githubusercontent.com/96078566/191854721-a0e34157-de61-414f-b5f2-635b2bd202bb.mov

### Enter key auto-trigger
https://user-images.githubusercontent.com/96078566/191380433-681aa213-3414-400a-b811-071748e8ada7.mov

https://user-images.githubusercontent.com/96078566/191380849-c762164a-77e5-4f2a-8993-313f8b0235f7.mov

### cwspr should not be triggered by code snippets generated by other extensions

https://user-images.githubusercontent.com/96078566/191380818-91e10e75-b827-4b5e-a30b-937605bac307.mov

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2947, opened 2022-10-22
fix(codewhisperer): fix auto-trigger regressions

leigaol:
## Problem
1. The 15 character keystroke auto-trigger was accidentally removed in https://github.com/aws/aws-toolkit-vscode/pull/2899. 
2. Enter auto trigger by \r\n was removed in https://github.com/aws/aws-toolkit-vscode/pull/2899. 
3. Reverted the 2 second auto trigger limitation for Special characters | Enter |intelliSense auto triggers which was removed in https://github.com/aws/aws-toolkit-vscode/pull/2899. 

## Solution
Trigger auto trigger when there are more than 15 keystroke inputs.

## License

### DIFF
--- a/.changes/next-release/Bug Fix-24b2f685-2010-4504-ace1-6bad1ca4b660.json
+++ b/.changes/next-release/Bug Fix-24b2f685-2010-4504-ace1-6bad1ca4b660.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Fix bug when CodeWhisperer auto trigger did not work as expected"
+}
--- a/.changes/next-release/Bug Fix-24b2f685-2010-4504-ace1-6bad1ca4b660.json
+++ b/.changes/next-release/Bug Fix-24b2f685-2010-4504-ace1-6bad1ca4b660.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Fix bug when CodeWhisperer auto trigger did not work as expected"
+	"description": "CodeWhisperer auto trigger not working correctly in certain circumstances"
 }
--- a/src/codewhisperer/service/keyStrokeHandler.ts
+++ b/src/codewhisperer/service/keyStrokeHandler.ts
@@ -99,6 +99,10 @@ export class KeyStrokeHandler {
                     break
                 }
             }
+            if (this.keyStrokeCount >= 15) {
+                triggerType = 'KeyStrokeCount'
+                this.keyStrokeCount = 0
+            }
             if (triggerType) {
                 this.invokeAutomatedTrigger(triggerType, editor, client, config)
             }
--- a/src/codewhisperer/service/keyStrokeHandler.ts
+++ b/src/codewhisperer/service/keyStrokeHandler.ts
@@ -57,11 +57,6 @@ export class KeyStrokeHandler {
             // Pause automated trigger when typed input matches recommendation prefix for inline suggestion
             if (InlineCompletion.instance.isTypeaheadInProgress) return
 
-            // Time duration between 2 invocations should be greater than the threshold
-            // This threshold does not applies to Enter | SpecialCharacters type auto trigger.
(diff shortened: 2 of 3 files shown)


src/codewhisperer/service/keyStrokeHandler.ts
```diff
@@ -99,6 +99,10 @@ export class KeyStrokeHandler {
                     break
                 }
             }
+            if (this.keyStrokeCount >= 15) {
+                triggerType = 'KeyStrokeCount'
+                this.keyStrokeCount = 0
+            }
```
andrewyuq, line 105: Should this logic be within the last case `DocumentChangedSource.RegularKey:`? Otherwise the triggerType will always be overridden to `KeyStrokeCount` when the count is 15 for any `DocumentChangedSource`. And this will impact telemetry and anything that uses this value.

leigaol: Makes sense.

zixlin7 (approved):

zixlin7 (commented):
I think we also need to move https://github.com/aws/aws-toolkit-vscode/pull/2947/files#diff-4f3d5ea842ea18b514c9a3ffd2536a7767fceb956f05c2c57967a454069b5bacL62-L63, the 2s cooldown should not be applied to special chars

leigaol:
> 



> I think we also need to move https://github.com/aws/aws-toolkit-vscode/pull/2947/files#diff-4f3d5ea842ea18b514c9a3ffd2536a7767fceb956f05c2c57967a454069b5bacL62-L63, the 2s cooldown should only be applied to the regular key

Agree.

andrewyuq (approved):

src/codewhisperer/service/keyStrokeHandler.ts
```diff
@@ -88,17 +86,27 @@ export class KeyStrokeHandler {
                 }
                 case DocumentChangedSource.IntelliSense: {
                     this.keyStrokeCount += 1
-                    triggerType = 'IntelliSenseAcceptance'
+                    if (duration >= CodeWhispererConstants.invocationTimeIntervalThreshold) {
+                        triggerType = 'IntelliSenseAcceptance'
+                    }
```
andrewyuq, line 91: actually the time threhold should not be applied to IntelliSense trigger I believe? At least this is the case for JB IIRC.

andrewyuq (changes requested):

src/codewhisperer/service/keyStrokeHandler.ts
leigaol: Done

andrewyuq (approved):

src/codewhisperer/service/keyStrokeHandler.ts
```diff
@@ -75,6 +70,9 @@ export class KeyStrokeHandler {
 
             let triggerType: CodewhispererAutomatedTriggerType | undefined
             const changedSource = new DefaultDocumentChangedType(event.contentChanges).checkChangeSource()
+            // Time duration between 2 invocations should be greater than the threshold
```
JadenSimon, line 73: Could we write a simple test for `processKeyStroke` to catch regressions?

JadenSimon (approved):

src/codewhisperer/service/keyStrokeHandler.ts
leigaol: No problem. I am working on it

.changes/next-release/Bug Fix-24b2f685-2010-4504-ace1-6bad1ca4b660.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Fix bug when CodeWhisperer auto trigger did not work as expected"
```
JadenSimon, line 3:
```suggestion
	"description": "CodeWhisperer auto trigger did not work as expected"
```

JadenSimon (approved):

leigaol:
The CI failure in Windows is not relevant to this PR

PR #2952, opened 2022-10-24
test(buildspec/cfn): Add basic intellisense tests for cloudformation and buildspec

jpinkney-aws:
## Problem:
- We don't have any intellisense level tests (hover, completion) for cloudformation or buildspec
    
## Solution:
- Add completion and hover tests for both cloudformation and buildspec
- The test schemas must be downloaded through the vscode git extension otherwise we risk getting rate limited on the tests. We can either do this upfront (like in this commit), which forces us to re-start the schema service to pick up the new schemas during the tests. Alternatively, we can use the git extension to download github related schemas everytime, but then we have to download an entire repo every single time we want to update the schema

## License

### DIFF
--- a/src/integrationTest/integrationTestsUtilities.ts
+++ b/src/integrationTest/integrationTestsUtilities.ts
@@ -8,7 +8,7 @@ import { waitUntil } from '../shared/utilities/timeoutUtils'
 import * as vscode from 'vscode'
 
 // java8.al2 image does a while to pull
-const LAMBDA_SESSION_TIMEOUT = 60000
+const lambdaSessionTimeout = 60000
 
 // Retrieves CodeLenses from VS Code
 export async function getCodeLenses(uri: vscode.Uri): Promise<vscode.CodeLens[] | undefined> {
@@ -29,7 +29,7 @@ export function getTestWorkspaceFolder(): string {
 export async function configureAwsToolkitExtension(): Promise<void> {
     const configAws = vscode.workspace.getConfiguration('aws')
     // How long the Toolkit will wait for SAM CLI output before ending a session.
-    await configAws.update('samcli.lambdaTimeout', LAMBDA_SESSION_TIMEOUT, false)
+    await configAws.update('samcli.lambdaTimeout', lambdaSessionTimeout, false)
     // Enable codelenses.
     await configAws.update('samcli.enableCodeLenses', true, false)
 }
--- a/src/integrationTest/integrationTestsUtilities.ts
+++ b/src/integrationTest/integrationTestsUtilities.ts
@@ -6,14 +6,10 @@
 import * as assert from 'assert'
 import { waitUntil } from '../shared/utilities/timeoutUtils'
 import * as vscode from 'vscode'
+import { getCodeLenses } from '../shared/utilities/vsCodeUtils'
 
 // java8.al2 image does a while to pull
-const lambdaSessionTimeout = 60000
-
-// Retrieves CodeLenses from VS Code
-export async function getCodeLenses(uri: vscode.Uri): Promise<vscode.CodeLens[] | undefined> {
-    return vscode.commands.executeCommand('vscode.executeCodeLensProvider', uri)
-}
+export const slowTestTimeout = 60000
 
 export function getTestWorkspaceFolder(): string {
     assert.ok(vscode.workspace.workspaceFolders, 'Integration Tests expect a workspace folder to be loaded')
@@ -29,7 +25,7 @@ export function getTestWorkspaceFolder(): string {
 export async function configureAwsToolkitExtension(): Promise<void> {
     const configAws = vscode.workspace.getConfiguration('aws')
     // How long the Toolkit will wait for SAM CLI output before ending a session.
-    await configAws.update('samcli.lambdaTimeout', lambdaSessionTimeout, false)
+    await configAws.update('samcli.lambdaTimeout', slowTestTimeout, false)
     // Enable codelenses.
     await configAws.update('samcli.enableCodeLenses', true, false)
 }
--- a/src/shared/schemas.ts
+++ b/src/shared/schemas.ts
@@ -48,7 +48,7 @@ export interface SchemaHandler {
  * Processes the update of schema mappings for files in the workspace
  */
 export class SchemaService {
-    private static readonly DEFAULT_UPDATE_PERIOD_MILLIS = 1000
+    private static readonly defaultUpdatePeriodMillis = 1000
 
     private updatePeriod: number
     private timer?: NodeJS.Timer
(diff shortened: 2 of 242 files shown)


src/integrationTest/cloudformation/intellisense.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    getTestWorkspaceFolder,
+    MAPPING_SCHEMA_INTERVAL,
+    MAPPING_SCHEMA_TIMEOUT,
+} from '../integrationTestsUtilities'
+import { activateExtension } from '../../shared/utilities/vsCodeUtils'
+import { waitUntil } from '../../shared/utilities/timeoutUtils'
+import globals from '../../shared/extensionGlobals'
+import { VSCODE_EXTENSION_ID } from '../../shared/extensions'
+
+describe('cloudformation yaml language features', async function () {
```
justinmk3, line 23: There is a lot of duplication in compared to the other test file. Can we put the common logic in one test file that iterates through different inputs?

src/integrationTest/integrationTestsUtilities.ts
```diff
@@ -10,11 +10,25 @@ import * as vscode from 'vscode'
 // java8.al2 image does a while to pull
 const LAMBDA_SESSION_TIMEOUT = 60000
 
+export const MAPPING_SCHEMA_TIMEOUT = 50000
```
justinmk3, line 13: Do we really need separate timeouts? This is roughly the same as LAMBDA_SESSION_TIMEOUT , they are both big. Just rename LAMBDA_SESSION_TIMEOUT to `SLOW_TEST_TIMEOUT` or something like that.

src/integrationTest/cloudformation/intellisense.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        assert.notStrictEqual(filterTags.length, 0, 'Expected more than 0 completion items')
+    })
+
+    it('hovers a sam file', async function () {
+        await vscode.commands.executeCommand('aws.sam.newTemplate')
+        await waitUntil(
+            () => {
+                return Promise.resolve(globals.schemaService.isMapped(testFileUri))
+            },
+            { timeout: MAPPING_SCHEMA_TIMEOUT, interval: MAPPING_SCHEMA_INTERVAL, truthy: true }
```
justinmk3, line 107: Unless `isMapped` is an expensive operation, the default of 500ms is fine. So we don't need this MAPPING_SCHEMA_INTERVAL global. Avoiding unnecessary symbols avoids questions for future developers about the intentions of the codebase.
```suggestion
            { timeout: MAPPING_SCHEMA_TIMEOUT }
```

src/integrationTest/integrationTestsUtilities.ts
```diff
@@ -10,11 +10,25 @@ import * as vscode from 'vscode'
...
+export async function getCompletionItems(
+    uri: vscode.Uri,
+    position: vscode.Position
+): Promise<vscode.CompletionList | undefined> {
+    return vscode.commands.executeCommand('vscode.executeCompletionItemProvider', uri, position)
+}
+
+export async function getHoverItems(uri: vscode.Uri, position: vscode.Position): Promise<vscode.Hover[] | undefined> {
+    return vscode.commands.executeCommand('vscode.executeHoverProvider', uri, position)
+}
```
justinmk3, line 30: these and getCodeLenses() can live in `src/shared/utilities/vsCodeUtils.ts`. They are not test-specific.

jpinkney-aws:
Putting in draft now until I get the integ tests working remotely (they work locally right now)

src/shared/schemas.ts
```diff
@@ -23,9 +23,9 @@ import { getStringHash } from './utilities/textUtilities'
 const goformationManifestURL = 'https://api.github.com/repos/awslabs/goformation/releases/latest'
 const devfileManifestURL = 'https://api.github.com/repos/devfile/api/releases/latest'
 const schemaPrefix = `${AWS_SCHEME}://`
-const hostedFilesCloudFrontDistribution = 'https://d3rrggjwfhwld2.cloudfront.net'
-const hostedFilesS3Fallback = 'https://aws-vs-toolkit.s3.amazonaws.com'
-const buildSpecHostedFilesPath = '/CodeBuild/buildspec/buildspec-standalone.schema.json'
+const buildspecHostedFilesPath = '/CodeBuild/buildspec/buildspec-standalone.schema.json'
+export const buildspecCloudfrontURL = 'https://d3rrggjwfhwld2.cloudfront.net' + buildspecHostedFilesPath
+export const buildspecS3FallbackURL = 'https://aws-vs-toolkit.s3.amazonaws.com' + buildspecHostedFilesPath
```
JadenSimon, line 28: For my own curiosity: why does an S3 fallback URL exist in addition to CF? Seems excessive, I'd at least hide this behind an abstraction if it's a pattern.

src/shared/schemas.ts
```diff
@@ -44,6 +44,10 @@ export interface SchemaHandler {
     isMapped(f: vscode.Uri | string): boolean
 }
 
+export const cfnSchemaUri = (path: vscode.Uri) => vscode.Uri.joinPath(path, 'cloudformation.schema.json')
+export const samSchemaUri = (path: vscode.Uri) => vscode.Uri.joinPath(path, 'sam.schema.json')
+export const buildSpecSchemaUri = (path: vscode.Uri) => vscode.Uri.joinPath(path, 'buildspec.schema.json')
```
JadenSimon, line 49: How likely do you think additional schemas will be added in the near-future? If the chances are pretty high then further generalizations might be justified.

src/shared/schemas.ts
justinmk3: good callout. This seems more confusing than useful.

-  adds a direct dependency to multiple names for the same thing. 
-  doubles the size of the config that we need to maintain.

however this decision was made before this PR, so probably should not change it now.

src/shared/schemas.ts
```diff
@@ -344,7 +348,14 @@ export class YamlSchemaHandler implements SchemaHandler {
         }
 
         if (mapping.schema) {
-            this.yamlExtension.assignSchema(mapping.uri, mapping.registry, resolveSchema(mapping.schema, schemas))
+            const schema = resolveSchema(mapping.schema, schemas)
+            if (!schema) {
+                getLogger().debug(
+                    `Could not assign schema ${mapping.schema} to ${mapping.uri}. Unable to find schema ${mapping.schema} locally`
```
justinmk3, line 354: - avoid mentioning duplicate values in log messages, it makes them noisy
- consider prefix the message with a topic (as done in other parts of this file), it gives context

```suggestion
                    `YamlSchemaHandler: failed to assign schema to ${mapping.uri}. Failed to find schema locally: ${mapping.schema}`
```

src/shared/schemas.ts
jpinkney-aws: I just removed the S3 fallback, I was just checking some documentation and the s3 bucket will be phased out over time. The original reason was just for extra availability just in case cloudfront went down. If I'm not mistaken, the visual studio toolkit will crash if it can't find certain files, that's why they were using that approach (and I followed that with our implementation). After double checking on VSCode though, we don't have this problem since we handle schema download failures

src/shared/schemas.ts
jpinkney-aws: I don't think we'll have any additional schemas here anymore. They should all come through the language server and we can eventually get rid of most of the schema related code

PR #2954, opened 2022-10-25
fix(schemas): Allow template registries to "own" a file path to fix race conditions in JSON Schema assignment

jpinkney-aws:
## Problem
- Currently buildspec and cloudformation are watching all yaml files and trying to race to see evaluate whether the current yaml file is buildspec or cloudformation. This means that sometimes schemas won't be set for valid buildspec or cloudformation files. This is due to an issue in the watchedfile registry, where having multiple template registries watch one file type wasn't really considered.

## Solution
- Introduce the concept of "owned" files in the template registries, where only owners of a file can update the schema for a file unless the type of the schema changes. E.g. cloudformation <-> buildspec. Since each registry is considered seperate and the implementation for the buildspec/cloudformation registries in other parts of the code require only their file types to be available, this choice may be easier than having to refactor everything. This is a temporary solution until we are able to remove the template registries completely.

## License

### DIFF
--- a/src/shared/schemas.ts
+++ b/src/shared/schemas.ts
@@ -28,6 +28,7 @@ export type SchemaType = 'yaml' | 'json'
 export interface SchemaMapping {
     uri: vscode.Uri
     type: SchemaType
+    owner?: string
     schema?: string | vscode.Uri
 }
 
@@ -50,6 +51,7 @@ export class SchemaService {
     private updateQueue: SchemaMapping[] = []
     private schemas?: Schemas
     private handlers: Map<SchemaType, SchemaHandler>
+    private owned: Map<vscode.Uri, SchemaMapping>
 
     public constructor(
         private readonly extensionContext: vscode.ExtensionContext,
@@ -68,6 +70,7 @@ export class SchemaService {
                 ['json', new JsonSchemaHandler()],
                 ['yaml', new YamlSchemaHandler()],
             ])
+        this.owned = new Map()
     }
 
     public isMapped(uri: vscode.Uri): boolean {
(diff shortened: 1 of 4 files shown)


JadenSimon (approved):

src/shared/schemas.ts
```diff
@@ -85,15 +88,33 @@ export class SchemaService {
...
      * @param mapping
      * @param flush Flush immediately instead of waiting for timer.
      */
     public registerMapping(mapping: SchemaMapping, flush?: boolean): void {
-        this.updateQueue.push(mapping)
-        if (flush === true) {
-            this.processUpdates()
+        // The owner is undefined and needs to be set
+        const isOwnerUndefined =
+            !this.owned.has(mapping.uri) && mapping.owner !== undefined && mapping.schema !== undefined
```
justinmk3, line 101: nit: "positive logic" is usually easier to read and leads to better names. `isOwned` would be the positive logic form of this, instead of `isOwnerUndefined` (which is equivalent to negative logic "is not owned").

src/shared/schemas.ts
```diff
@@ -85,15 +88,33 @@ export class SchemaService {
...
+
+        // The owner and the incoming owner are defined but the schema changed
```
justinmk3, line 103: This comment restates the code. If the comment is needed, it should explain the semantics, not the mechanics.

src/shared/schemas.ts
```diff
@@ -85,15 +88,33 @@ export class SchemaService {
...
+            this.owned.get(mapping.uri)?.schema !== mapping.schema
+
+        if (isOwnerUndefined || isSchemaChanged) {
+            this.owned.set(mapping.uri, mapping)
+        }
+
+        if (mapping.owner === this.owned.get(mapping.uri)?.owner) {
+            this.updateQueue.push(mapping)
+            if (flush === true) {
+                this.processUpdates()
```
justinmk3, line 116: Why does `this.processUpdates()` depend on the owner condition?

Issue #2953, opened 2022-10-25
Copy full path in explore


b-y-f:
## Problem
I want to copy full path instead of folder/filename
but now I got `folder/` when I click copy Path

## Expected behavior
`s3://bucket/folder`

justinmk3:
The `Copy ARN` action gives the full ARN. Are you using a tool that accepts `s3://` paths but not ARNs?

One problem I noticed with AWS CLI: `aws s3` command does not accept ARNs: https://docs.aws.amazon.com/cli/latest/reference/s3/

That is a quirk of AWS CLI and/or S3 ARNs. For reference, it works with Lambda:

```
aws lambda get-function --function-name 'arn:aws:lambda:us-west-2:111111111111:function:zzz'
```

# Proposal

- For S3 resources `Copy Path` can produce a `s3://` URI. That is likely more useful than the current behavior.
- We could also special-case `Copy ARN` to produce a `s3://` URI.

b-y-f:
> The `Copy ARN` action gives the full ARN. Are you using a tool that accepts `s3://` paths but not ARNs?
> 
> One problem I noticed with AWS CLI: `aws s3` command does not accept ARNs: https://docs.aws.amazon.com/cli/latest/reference/s3/
> 
> That is a quirk of AWS CLI and/or S3 ARNs. For reference, it works with Lambda:
> 
> ```
> aws lambda get-function --function-name 'arn:aws:lambda:us-west-2:111111111111:function:zzz'
> ```
> 
> # Proposal
> * For S3 resources `Copy Path` can produce a `s3://` URI. That is likely more useful than the current behavior.
>   
>   * Done: [feat(s3): "Copy Path" gets the full "s3://…" path #2960](https://github.com/aws/aws-toolkit-vscode/pull/2960)
> * We could also special-case `Copy ARN` to produce a `s3://` URI.

We normally need something like `s3://bucket/folder` to be pasted in aws cli or boto3 or awswrangler to read the data.

PR #2960, opened 2022-10-26
feat(s3): "Copy Path" now gets the full "s3://…" path

justinmk3:
## Problem
- `Copy Path` isn't very useful. #2953
- S3 ARNs don't work with AWS CLI `aws s3` command: https://docs.aws.amazon.com/cli/latest/reference/s3/

## Solution
Change the behavior of `Copy Path` to return `s3://` URLs instead of just the "key".

## License


JadenSimon:
I'd rather have a separate command `Copy URI` than change `Change Path`. The S3 URI is not a path.

calumblair:
Is there any progress on this? This would be a useful feature to have.

PR #2965, opened 2022-11-02
fix(credentials): "Connect" and "Choose Profile" open create credentials wizard/profile selector when credentials aren't present in aws configs

jpinkney-aws:
## Problem
- If ~/.aws/config exists but doesn't have a usable credentials profile, the Connect to AWS and Choose AWS Profile actions open ~/.aws/config immediately instead of starting the Create Credentials Profile wizard.

## Solution
- When using "Connect" and "Choose profile", open the create credentials wizard or the profile selector when credentials aren't actually present in the ~/.aws/config and ~/.aws/credentials files

## License

### DIFF
--- a/src/shared/credentials/userCredentialsUtils.ts
+++ b/src/shared/credentials/userCredentialsUtils.ts
@@ -57,11 +57,18 @@ export class UserCredentialsUtils {
      *
      * @returns array of filenames for files found.
      */
-    public static async findExistingCredentialsFilenames(): Promise<string[]> {
+    public static async findExistingCredentialsFilenames(isCredentialsRequired = false): Promise<string[]> {
         const candidateFiles: string[] = [getCredentialsFilename(), getConfigFilename()]
 
         const existsResults: boolean[] = await Promise.all(
-            candidateFiles.map(async filename => await SystemUtilities.fileExists(filename))
+            candidateFiles.map(async filename => {
+                const exists = await SystemUtilities.fileExists(filename)
+                if (!exists || !isCredentialsRequired) {
+                    return exists
+                }
+                const contents = await SystemUtilities.readFile(filename)
+                return /aws_access_key_id|aws_secret_access_key|aws_session_token/i.test(contents)
+            })
         )
 
         return candidateFiles.filter((filename, index) => existsResults[index])
--- a/src/test/shared/credentials/userCredentialsUtils.test.ts
+++ b/src/test/shared/credentials/userCredentialsUtils.test.ts
@@ -6,6 +6,7 @@
 import * as assert from 'assert'
 import * as fs from 'fs-extra'
 import * as path from 'path'
+import * as sinon from 'sinon'
 
 import { Uri } from 'vscode'
 import { loadSharedConfigFiles } from '../../../shared/credentials/credentialsFile'
@@ -23,9 +24,7 @@ describe('UserCredentialsUtils', function () {
     })
 
     afterEach(async function () {
-        const env = process.env as EnvironmentVariables
-        delete env.AWS_SHARED_CREDENTIALS_FILE
-        delete env.AWS_CONFIG_FILE
+        sinon.restore()
     })
 
     after(async function () {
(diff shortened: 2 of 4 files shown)


src/test/shared/credentials/userCredentialsUtils.test.ts
```diff
@@ -93,6 +93,22 @@ describe('UserCredentialsUtils', function () {
...
             assert.strictEqual(foundFiles.length, 0)
         })
+
+        it('returns empty result if files dont contain credentials', async function () {
+            const credentialsFilename = path.join(tempFolder, 'credentials-both-exist-but-no-credentials-test')
+            const configFilename = path.join(tempFolder, 'config-both-exist-but-no-credentials-test')
+
+            const env = process.env as EnvironmentVariables
+            env.AWS_SHARED_CREDENTIALS_FILE = credentialsFilename
+            env.AWS_CONFIG_FILE = configFilename
```
justinmk3, line 103: does `after()` need to restore the env?

src/test/shared/credentials/userCredentialsUtils.test.ts
```diff
@@ -93,6 +93,22 @@ describe('UserCredentialsUtils', function () {
...
+
+            createCredentialsFile(credentialsFilename, [])
+            createCredentialsFile(configFilename, [])
+
+            const foundFiles: string[] = await UserCredentialsUtils.findExistingCredentialsFilenames(true)
+            assert(foundFiles)
```
justinmk3, line 109: doesn't the `: string[]` type declaration on line 108 make this unnecessary? could use `foundFiles?.length` below

src/test/shared/credentials/userCredentialsUtils.test.ts
JadenSimon: Yeah I'd use `sinon.replace` here with `sinon.restore` in `afterEach`

jpinkney-aws: I went with `sinon.stub` for the latest version and changed all the other tests as well. They used to remove delete the environment variables after each test

src/test/shared/credentials/userCredentialsUtils.test.ts
jpinkney-aws: Yeah, this was a weird test convention that was used in this test file. I've removed this from all the other tests as well

src/test/shared/credentials/userCredentialsUtils.test.ts
```diff
@@ -93,6 +93,22 @@ describe('UserCredentialsUtils', function () {
             assert(foundFiles)
             assert.strictEqual(foundFiles.length, 0)
         })
+
+        it('returns empty result if files dont contain credentials', async function () {
+            const credentialsFilename = path.join(tempFolder, 'credentials-both-exist-but-no-credentials-test')
+            const configFilename = path.join(tempFolder, 'config-both-exist-but-no-credentials-test')
```
justinmk3, line 98: These files don't exist, which is worth testing. But what about the case where the files exist but

1. they contain nonsense
2. they are empty
3. they contain valid profiles but the profiles are not listable (e.g. `[sso-profile foo]`?

src/shared/credentials/userCredentialsUtils.ts
```diff
@@ -57,11 +57,18 @@ export class UserCredentialsUtils {
...
 
         const existsResults: boolean[] = await Promise.all(
-            candidateFiles.map(async filename => await SystemUtilities.fileExists(filename))
+            candidateFiles.map(async filename => {
+                const exists = await SystemUtilities.fileExists(filename)
+                if (!exists || !isCredentialsRequired) {
+                    return exists
+                }
+                const contents = await SystemUtilities.readFile(filename)
+                return /aws_access_key_id|aws_secret_access_key|aws_session_token/i.test(contents)
```
justinmk3, line 70: This is rather fragile. We already have code that logs things like this:

```
2022-11-02 22:01:54 [ERROR]: Profile foo is not a valid Credential Profile: not supported by the Toolkit
2022-11-02 22:01:54 [WARN]: Shared Credentials Profile foo is not valid. It will not be used by the toolkit.
```

can we use that logic as the source of truth?

PR #2967, opened 2022-11-03
feat(auth): persist active connection and add more docs

JadenSimon:
## Problem
The Auth service doesn't remember the active connection across sessions

## Solution
Persist the connection id and restore it later.

Originally I was going to implement something similar to `canAutoConnect` but just persisting the current connection is simpler and more user-friendly for the case of expired connections

## License

### DIFF
--- a/src/credentials/auth.ts
+++ b/src/credentials/auth.ts
@@ -275,8 +275,8 @@ export class Auth implements AuthService, ConnectionManager {
             throw new Error(`Connection does not exist: ${id}`)
         }
 
-        await this.validateConnection(id, profile)
-        const conn = (this.#activeConnection = this.getSsoConnection(id, profile))
+        const validated = await this.validateConnection(id, profile)
+        const conn = (this.#activeConnection = this.getSsoConnection(id, validated))
         this.onDidChangeActiveConnectionEmitter.fire(conn)
         await this.store.setCurrentProfileId(id)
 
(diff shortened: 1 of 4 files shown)


src/credentials/auth.ts
```diff
@@ -276,7 +356,24 @@ export class Auth implements AuthService, ConnectionManager {
         return connections.find(c => c.id === connection.id)
     }
 
-    private async updateState(id: Connection['id'], connectionState: ProfileMetadata['connectionState']) {
+    private async invalidateConnection(id: Connection['id']) {
```
justinmk3, line 359: A docstring explaining what this does would be helpful, because the details matter and can be tricky. E.g. does it make a backend request to invalidate, which things does it clear locally, etc.

src/credentials/sso/clients.ts
```diff
@@ -161,6 +162,11 @@ export class SsoClient {
         }
     }
 
+    public async logout(request: Omit<LogoutRequest, OmittedProps> = {}) {
+        const method = this.client.logout.bind(this.client)
+        await this.call(method as ExtractOverload<typeof method>, request)
```
justinmk3, line 167: this bind() + ExtractOverload pattern is hard to follow

justinmk3 (approved):

src/credentials/sso/clients.ts
JadenSimon: It's kind of a TS hack specific to JS SDK v3. Can probably make it nicer.

JadenSimon:
I'll address the feedback in the next PR

PR #2970, opened 2022-11-04
feat(auth): add minimal signout button

JadenSimon:
## Problem
No UI to signout

## Solution
Add a button on the node. This is a placeholder to make development easier and is not the final UI.

## License

### DIFF
--- a/src/credentials/auth.ts
+++ b/src/credentials/auth.ts
@@ -482,6 +482,7 @@ export async function promptLogin(auth: Auth) {
 }
 
 const loginCommand = Commands.register('aws.auth.login', promptLogin)
+Commands.register('aws.auth.logout', () => Auth.instance.logout())
 
 function mapEventType<T, U = void>(event: vscode.Event<T>, fn?: (val: T) => U): vscode.Event<U> {
     const emitter = new vscode.EventEmitter<U>()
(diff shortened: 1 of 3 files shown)


src/credentials/auth.ts
```diff
@@ -356,6 +356,12 @@ export class Auth implements AuthService, ConnectionManager {
         return connections.find(c => c.id === connection.id)
     }
 
+    /**
+     * Attempts to remove all auth state related to the connection.
+     *
+     * For SSO, this involves an API call to clear server-side state. The call happens
+     * before the local token(s) are cleared as they are needed in the request.
```
justinmk3, line 363: [off-topic for this PR] it looks like provider.invalidate() is called even if client.logout() fails. Would it make sense to skip provider.invalidate() in that case?

src/credentials/sso/clients.ts
```diff
@@ -148,8 +152,7 @@ export class SsoClient {
     }
 
     public async getRoleCredentials(request: Omit<GetRoleCredentialsRequest, OmittedProps>) {
-        const method = this.client.getRoleCredentials.bind(this.client)
-        const response = await this.call(method as ExtractOverload<typeof method>, request)
+        const response = await this.call(this.client.getRoleCredentials, request)
```
justinmk3, line 155: nice!

justinmk3 (approved):

src/credentials/auth.ts
JadenSimon: Maybe. It depends on user expectations + how the call failed. If we skip the local invalidation then users might not be directed to the browser the next time they select the profile. At least right now the call fails consistently for scoped tokens which is one reason to always clear.

Issue #2855, opened 2022-08-29
Remember target directory when downloading files from S3


pawel-kuras:
## Problem
When the aws toolkit extension is connected to S3, I can right-click a file, then choose "Download..." to download it to my local machine. Sadly, the "Save as" dialog doesn't remember the directory I select. So when I perform a second download, I have to select the same directory all over again.

Note: Essentially copied the description from the completed issue https://github.com/microsoft/vscode/issues/140358

## Expected behavior
It would be convenient if the previously-selected directory was remembered.

Alternatively, it would be convenient if you could set a default downloads directory, but this doesnt match the functionality vs-code went with, as they closed https://github.com/microsoft/vscode-remote-release/issues/5998 in favor of https://github.com/microsoft/vscode/issues/140358

justinmk3:
> Alternatively, it would be convenient if you could set a default downloads directory, but this doesnt match the functionality vs-code went with, as they closed [microsoft/vscode-remote-release#5998](https://github.com/microsoft/vscode-remote-release/issues/5998) in favor of [microsoft/vscode#140358](https://github.com/microsoft/vscode/issues/140358)

That reference is helpful, thanks! It probably makes sense for AWS Toolkit to use the same globally-stored path as VSCode? 

From https://github.com/microsoft/vscode/commit/1b8e78461329d43d79a47a1c38e081e63d7f75ae#diff-d8122abf1f3d6b4464ce28fe7e0358f9a0f22b9cc8634804f509f04eda753788R799 that's `workbench.explorer.downloadPath` in globalStorage.

justinmk3:
In AWS Toolkit 1.60, the S3 "Download" action now defaults to the last-used download location.

PR #2971, opened 2022-11-07
S3: Remember target directory when downloading files

gsweetwood:
## Problem
When downloading S3 files, the prompt for the save location always defaults to the Downloads directory.

## Solution
Save the last used download path and use it for next save location prompt.

ref https://github.com/aws/aws-toolkit-vscode/issues/2855

## License


gsweetwood:
I'm currently looking into refactoring this for whenever the toolkit prompts for download location.

src/s3/commands/downloadFileAs.ts
```diff
@@ -119,9 +120,17 @@ export async function downloadFileAsCommand(
     const sourcePath = readablePath(node)
 
     await telemetry.s3_downloadObject.run(async () => {
-        const saveLocation = await promptForSaveLocation(file.name, window)
+        let lastUsedDownloadPath: string | undefined
+        const storedPath = globals.context.globalState.get('workbench.explorer.downloadPath')
```
jpinkney-aws, line 124: Can we extract `'workbench.explorer.downloadPath'` to a top level variable?

jpinkney-aws, line 124: As far as I can tell, we don't really have conventions for the key in globals.context.globalState, but would it make sense to go with something like `S3_LATEST_DOWNLOAD_PATH` or something like that?

justinmk3: +1 for conventions
-1 for UPPER_CASE, idk why we are doing that anywhere for anything

justinmk3: rather, let's put all of this logic in a `getDefaultDownloadPath()` function

src/s3/commands/downloadFileAs.ts
```diff
@@ -119,9 +120,17 @@ export async function downloadFileAsCommand(
...
+        if (typeof storedPath === 'object' && isNonNullable(storedPath) && hasStringProps(storedPath, 'fsPath')) {
+            lastUsedDownloadPath = path.dirname(storedPath.fsPath)
+        }
+
+        const saveLocation = await promptForSaveLocation(file.name, window, lastUsedDownloadPath)
         if (!saveLocation) {
             throw new CancellationError('user')
+        } else {
+            globals.context.globalState.update('workbench.explorer.downloadPath', saveLocation)
```
justinmk3, line 133: need at least a comment here explaining the significance of this. This is not a casual line of code.

src/s3/commands/downloadFileAs.ts
jpinkney-aws: It's just codewhisperer doing the uppercase, the codebase is pretty inconsistent on it

justinmk3: Going forward we should not use uppercase. That's an old habit from the original codebase that doesn't really have a purpose.

src/s3/commands/downloadFileAs.ts
gsweetwood: This is saved in the Toolkit's globalStorage and has been renamed to `aws.latestDowloadPath`

src/shared/filesystemUtilities.ts
```diff
@@ -12,8 +12,10 @@ import * as path from 'path'
 import * as vscode from 'vscode'
 import { getLogger } from './logger'
 import * as pathutils from './utilities/pathUtils'
+import globals from '../shared/extensionGlobals'
 
 const DEFAULT_ENCODING: BufferEncoding = 'utf8'
+export const latestDownloadPathSetting = 'aws.latestDowloadPath'
```
justinmk3, line 18: not a setting, just a key.

Don't need this constant anyway, instead make a set function similar to the get function (getDefaultDownloadPath)

src/shared/filesystemUtilities.ts
```diff
@@ -220,3 +222,13 @@ export async function cloud9Findfile(dir: string, fileName: string): Promise<vsc
     }
     return []
 }
+/**
+ * @returns  A string path to the last locally stored download location. If none, returns the users 'Downloads' directory path.
+ */
+export function getDefaultDownloadPath() {
+    const lastUsedPath = globals.context.globalState.get(latestDownloadPathSetting)
+    if (typeof lastUsedPath === 'string') {
```
justinmk3, line 230: Always log unexpected circumstances.

justinmk3:
Should be possible to add a test for this, because these functions are already tested and globalStorage should "just work" in the test harness.

src/shared/filesystemUtilities.ts
JadenSimon: "recent" seems to be the preferred word for these things although `aws.downloadPath` (or `aws.downloadLocation`) would be good too. The adjective feels unnecessary IMO.

src/s3/commands/downloadFileAs.ts
```diff
@@ -149,7 +157,7 @@ async function promptForSaveLocation(fileName: string, window: Window): Promise<
         ? { [`*${extension}`]: [extension.slice(1)], 'All Files': ['*'] }
         : { 'All Files': ['*'] }
 
-    const downloadPath = path.join(downloadsDir(), fileName)
+    const downloadPath = path.join(saveLocation, fileName)
     return window.showSaveDialog({
         defaultUri: vscode.Uri.file(downloadPath),
         saveLabel: localize('AWS.s3.downloadFile.saveButton', 'Download'),
```
JadenSimon, line 163: What happens if we do this?
```suggestion
    return window.showSaveDialog({
        saveLabel: localize('AWS.s3.downloadFile.saveButton', 'Download'),
```

gsweetwood: This will pull VSC's 'last used download path', but there's two problems.

1. This doesn't add the file name of what's being downloaded to the save window. Meaning the user would need to manually enter a name every time.
2. This will NOT update VSC's 'last used download path'.

src/shared/filesystemUtilities.ts
```diff
@@ -220,3 +221,21 @@ export async function cloud9Findfile(dir: string, fileName: string): Promise<vsc
...
+            return lastUsedPath
+        } else {
+            getLogger().verbose("Expected stored 'aws.downloadPath' to return a string")
+        }
+    }
+    return downloadsDir()
+}
+
+export function setDefaultDownloadPath(downloadPath: string) {
+    globals.context.globalState.update('aws.downloadPath', path.dirname(downloadPath))
```
justinmk3, line 240: This assumes that `downloadPath` points to a file. That isn't clear from the function interface or docstring. Worth adding a isdirectory() check.

src/s3/commands/downloadFileAs.ts
```diff
@@ -119,9 +119,13 @@ export async function downloadFileAsCommand(
...
 
     await telemetry.s3_downloadObject.run(async () => {
-        const saveLocation = await promptForSaveLocation(file.name, window)
+        const downloadPath = getDefaultDownloadPath()
+
+        const saveLocation = await promptForSaveLocation(file.name, window, downloadPath)
         if (!saveLocation) {
             throw new CancellationError('user')
+        } else {
+            setDefaultDownloadPath(saveLocation.path)
```
justinmk3, line 128: Don't need `else` when the other branch exits the scope (`throw`, `return`, etc.). Please use this pattern for all future code.

```suggestion
        setDefaultDownloadPath(saveLocation.path)
```

src/shared/filesystemUtilities.ts
```diff
@@ -220,3 +221,21 @@ export async function cloud9Findfile(dir: string, fileName: string): Promise<vsc
...
+ * @returns  A string path to the last locally stored download location. If none, returns the users 'Downloads' directory path.
+ */
+export function getDefaultDownloadPath(): string {
+    const lastUsedPath = globals.context.globalState.get('aws.downloadPath')
+    if (lastUsedPath) {
+        if (typeof lastUsedPath === 'string') {
+            return lastUsedPath
+        } else {
+            getLogger().verbose("Expected stored 'aws.downloadPath' to return a string")
+        }
```
justinmk3, line 234: Don't need else when the other branch exits the scope (throw, return, etc.). Please use this pattern for all future code.

```suggestion
        getLogger().verbose("Expected stored 'aws.downloadPath' to return a string")
```

src/shared/filesystemUtilities.ts
```diff
@@ -220,3 +221,24 @@ export async function cloud9Findfile(dir: string, fileName: string): Promise<vsc
...
+/**
+ * @returns  A string path to the last locally stored download location. If none, returns the users 'Downloads' directory path.
+ */
+export function getDefaultDownloadPath(): string {
+    const lastUsedPath = globals.context.globalState.get('aws.downloadPath')
+    if (lastUsedPath) {
+        if (typeof lastUsedPath === 'string') {
+            return lastUsedPath
+        }
+        getLogger().warn("Expected stored 'aws.downloadPath' to return a string")
```
justinmk3, line 233: This is an unusual and suspicious state so error is the correct log-level.

```suggestion
        getLogger().error("Expected stored 'aws.downloadPath' to return a string")
```

src/test/s3/commands/downloadFileAs.test.ts
```diff
@@ -52,12 +53,18 @@ describe('downloadFileAsCommand', function () {
...
 
         when(s3.downloadFileStream(anything(), anything())).thenResolve(bufferToStream(Buffer.alloc(16)))
 
         await downloadFileAsCommand(node, window, outputChannel)
 
         assert.ok(window.dialog.saveOptions?.defaultUri?.path?.endsWith(fileName))
+        // VS Code nomalizes drive letters to lower case
+        assert.strictEqual(
+            window.dialog.saveOptions?.defaultUri?.fsPath.toLowerCase(),
+            path.join(temp, 'file.jpg').toLowerCase()
```
justinmk3, line 66: paths can be compared using `src/test/testUtil.ts:assertEqualPaths()`

src/s3/commands/downloadFileAs.ts
```diff
@@ -136,11 +138,17 @@ export async function downloadFileAsCommand(
             }
         )
 
+        setDefaultDownloadPath(saveLocation.fsPath)
```
justinmk3, line 141: This should happen immediately after `promptForSaveLocation()`. Otherwise, while a big file is downloading the path still isn't saved. Testing the feature confirms that this doesn't feel like the expected behavior.

PR #2976, opened 2022-11-09
Use hosted files for buildspec schema

jpinkney-aws:
## Problem
- Currently you need to configure the buildspec schema location through a setting

## Solution
- Use the hosted files cloudfront distribution for getting the buildspec schema

## License

### DIFF
--- a/src/shared/resourcefetcher/httpResourceFetcher.ts
+++ b/src/shared/resourcefetcher/httpResourceFetcher.ts
@@ -154,15 +154,10 @@ export class HttpResourceFetcher implements ResourceFetcher {
  * Retrieves JSON property value from a remote resource
  * @param property property to retrieve
  * @param url url of JSON resource
- * @param fetcher optional HTTP resource fetcher to use
  * @returns property value if available or undefined
  */
-export async function getPropertyFromJsonUrl(
-    url: string,
-    property: string,
-    fetcher?: HttpResourceFetcher
-): Promise<any | undefined> {
-    const resourceFetcher = fetcher ?? new HttpResourceFetcher(url, { showUrl: true })
+export async function getPropertyFromJsonUrl(url: string, property: string): Promise<any | undefined> {
+    const resourceFetcher = new HttpResourceFetcher(url, { showUrl: true })
     const result = await resourceFetcher.get()
     if (result) {
         try {
@@ -174,4 +169,14 @@ export async function getPropertyFromJsonUrl(
             getLogger().error(`JSON parsing failed for "${url}": ${(err as Error).message}`)
         }
     }
+}
+
+/**
+ * Gets contents from a remote resource
+ * @param url url of the resource
+ * @returns contents from the url if available or undefined
+ */
+export async function getFromUrl(url: string): Promise<string | undefined> {
+    const resourceFetcher = new HttpResourceFetcher(url, { showUrl: true })
+    return resourceFetcher.get()
 }
--- a/src/shared/schemas.ts
+++ b/src/shared/schemas.ts
@@ -10,7 +10,7 @@ import * as pathutil from '../shared/utilities/pathUtils'
 import { getLogger } from './logger'
 import { FileResourceFetcher } from './resourcefetcher/fileResourceFetcher'
 import { getPropertyFromJsonUrl, HttpResourceFetcher } from './resourcefetcher/httpResourceFetcher'
-import { DevSettings, Settings } from './settings'
+import { Settings } from './settings'
 import { once } from './utilities/functionUtils'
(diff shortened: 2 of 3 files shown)


justinmk3:
Please keep commit subjects under 72 chars, see https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#commit-messages

Contex such as "now that the build spec..." can live in the commit message detail area.

src/shared/resourcefetcher/httpResourceFetcher.ts
```diff
@@ -175,3 +176,17 @@ export async function getPropertyFromJsonUrl(
...
     }
 }
+
+/**
+ * Retrieves sha-256 hash of the contents at a remote resource
+ * @param url url of the resource
+ * @param fetcher optional HTTP resource fetcher to use
+ * @returns hashed contents from the url if available or undefined
+ */
+export async function getContentHashFromUrl(url: string, fetcher?: HttpResourceFetcher): Promise<string | undefined> {
```
justinmk3, line 186: Many things can be done with a `resourceFetcher.get()` result. But we wouldn't want to add `httpResourceFetcher.ts` functions for every operation that can be done on text. 

Instead, this is a signal that `HttpResourceFetcher` is too cumbersome to set up and use. So let's fix that instead of adding a special-case function just to hash the contents of a downloaded file. I.e. we could add a function here like:

```
export async function getFromurl(url: string, fetcher): ... {
    const resourceFetcher = fetcher ?? new HttpResourceFetcher(url, { showUrl: true })
    const result = await resourceFetcher.get()
    return result ? result : undefined
}
```

Then the call site looks like:

```
getStringHash(getFromurl(...))
```

This is a general principle that can/should be applied to all future code.

src/shared/schemas.ts
```diff
@@ -21,6 +25,9 @@ import { CloudFormation } from './cloudformation/cloudformation'
 
 const GOFORMATION_MANIFEST_URL = 'https://api.github.com/repos/awslabs/goformation/releases/latest'
 const SCHEMA_PREFIX = `${AWS_SCHEME}://`
+const BUILD_SPEC_HOSTED_FILES_PATH = '/CodeBuild/buildspec/buildspec-standalone.schema.json'
+const BUILDSPEC_CLOUDFRONT_URL = 'https://d3rrggjwfhwld2.cloudfront.net' + BUILD_SPEC_HOSTED_FILES_PATH
+const BUILDSPEC_S3_FALLBACK_URL = 'https://aws-vs-toolkit.s3.amazonaws.com' + BUILD_SPEC_HOSTED_FILES_PATH
```
justinmk3, line 30: There are not many uppercase names here, let's take the opportunity to discontinue that practice here. Just use normal names. There is nothing so special about constants that they need to be uppercase.

```suggestion
const buildspecS3FallbackUrl = 'https://aws-vs-toolkit.s3.amazonaws.com' + BUILD_SPEC_HOSTED_FILES_PATH
```

src/shared/resourcefetcher/httpResourceFetcher.ts
justinmk3: should `url` at least be `string | vscode.Uri` (or even just `vscode.Uri` ) ?

jpinkney-aws: What advantage would there to be passing in a `vscode.URI`? I think we're probably fine with just a url `string` unless we eventually expand the scope of the resource fetcher

justinmk3: There might be no immediate advantage. But that's usually the case when we start with string for Uris--then the pattern tends to bleed into the codebase over time. Not a blocker though.

src/shared/resourcefetcher/httpResourceFetcher.ts
```diff
@@ -175,3 +175,14 @@ export async function getPropertyFromJsonUrl(
...
+
+/**
+ * Gets contents from a remote resource
+ * @param url url of the resource
+ * @param fetcher optional HTTP resource fetcher to use
+ * @returns contents from the url if available or undefined
+ */
+export async function getFromUrl(url: string, fetcher?: HttpResourceFetcher): Promise<string | undefined> {
+    const resourceFetcher = fetcher ?? new HttpResourceFetcher(url, { showUrl: true })
+    return resourceFetcher.get()
```
justinmk3, line 187: Looks like the `fetcher` param doesn't make sense, because it doesn't use `url`.

justinmk3 (approved):

justinmk3:
mac/win CI failures are unrelated

JadenSimon (approved):

justinmk3:
do we need a changelog ?

PR #2979, opened 2022-11-11
fix(auth): update status bar color; use SSO connection as a fallback

JadenSimon:
## Problem
* Statusbar should be in an error state for expired credentials 
* Signout out of IAM connection puts users in a fully logged-out state

## Solution
* Update statusbar color depending on connection state
    * Use different color in dev mode 
* Fallback to an SSO connection when signing out of an IAM connection

## License

### DIFF
--- a/src/credentials/awsCredentialsStatusBarItem.ts
+++ b/src/credentials/awsCredentialsStatusBarItem.ts
@@ -30,7 +30,8 @@ export async function initializeAwsCredentialsStatusBarItem(
 
     context.subscriptions.push(
         Auth.instance.onDidChangeActiveConnection(conn => {
-            updateCredentialsStatusBarItem(statusBarItem, conn?.label)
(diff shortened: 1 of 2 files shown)


src/credentials/awsCredentialsStatusBarItem.ts
```diff
@@ -41,19 +42,18 @@ function handleDevSettings(devSettings: DevSettings, statusBarItem: vscode.Statu
     const developerMode = Object.keys(devSettings.activeSettings)
 
     if (developerMode.length > 0) {
-        ;(statusBarItem as any).backgroundColor = new vscode.ThemeColor('statusBarItem.errorBackground')
+        ;(statusBarItem as any).backgroundColor ??= new vscode.ThemeColor('statusBarItem.warningBackground')
```
JadenSimon, line 45: It's an ugly yellow now :/
I tried `statusBar.debuggingBackground` but it didn't work

PR #2984, opened 2022-11-15
fix(auth): ensure token exists before persisting connection

JadenSimon:
## Problem
Connections are created _before_ the user finishes the auth flow. This can be confusing because the connections aren't marked as 'pending'. 

## Solution
Don't persist the connection until we get a token.

## License

### DIFF
--- a/src/credentials/auth.ts
+++ b/src/credentials/auth.ts
@@ -380,17 +380,21 @@ export class Auth implements AuthService, ConnectionManager {
 
         // XXX: `id` should be based off the resolved `idToken`, _not_ the source profile
         const id = getSsoProfileKey(profile)
-        const storedProfile = await this.store.addProfile(id, profile)
-        const conn = this.getSsoConnection(id, storedProfile)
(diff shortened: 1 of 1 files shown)


src/credentials/auth.ts
```diff
@@ -380,17 +380,21 @@ export class Auth implements AuthService, ConnectionManager {
...
-        const storedProfile = await this.store.addProfile(id, profile)
-        const conn = this.getSsoConnection(id, storedProfile)
+        const tokenProvider = this.getTokenProvider(id, {
+            ...profile,
+            metadata: { connectionState: 'unauthenticated' },
+        })
 
         try {
-            await conn.getToken()
+            ;(await tokenProvider.getToken()) ?? (await tokenProvider.createToken())
```
leigaol, line 389: Is  this ; required?

JadenSimon: Yes. When we have `idToken` it won't look as ugly because the tokens will be immediately used whereas right now we're just causing side-effects.

leigaol (approved):

PR #2991, opened 2022-11-17
feat(auth): add `SecondaryAuth`

JadenSimon:
## Problem
* No "Edit Credentials" button when switching connections
* Persisting connections is not cohesive

## Solution
* Add "Edit Credentials" button
* Add `SecondaryAuth` class (dead code)
 

## License

### DIFF
--- a/src/credentials/secondaryAuth.ts
+++ b/src/credentials/secondaryAuth.ts
@@ -16,7 +16,9 @@ import { UnknownError } from '../shared/errors'
 async function promptUseNewConnection(newConn: Connection, oldConn: Connection, tools: string[]) {
     // Multi-select picker would be better ?
     const saveConnectionItem = {
-        label: `Yes, use ${newConn.label} with ${tools.join(', ')} while using ${oldConn.label} with other services.`,
+        label: `Yes, keep using ${newConn.label} with ${tools.join(', ')} while using ${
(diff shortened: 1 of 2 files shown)


src/credentials/secondaryAuth.ts
```diff
@@ -0,0 +1,173 @@
...
+})
+
+export function getSecondaryAuth<T extends Connection>(
+    toolId: string,
+    toolLabel: string,
+    isValid: (conn: Connection) => conn is T
+): SecondaryAuth<T> {
+    const auth = new SecondaryAuth(toolId, toolLabel, isValid)
+    auths.set(toolId, auth)
+    registerAuthListener()
```
justinmk3, line 64: off-topic: could once() work at the call-site instead of declaration? Would be much clearer I think.

```suggestion
    once(registerAuthListener)
```

justinmk3 (approved):

src/credentials/secondaryAuth.ts
JadenSimon: > off-topic: could once() work at the call-site instead of declaration? Would be much clearer I think.

Yeah it can be done although the behavior would still be associated with the function rather than the callsite. So like `once(foo)` would only invoke `foo` once regardless of where you called `once`.

PR #2994, opened 2022-11-18
fix(auth): adjust UI on C9

JadenSimon:
## Problem
* Missing icons 
* Tree nodes don't refresh in all cases
* Tree nodes do not render `description`
* Auth node gets sorted to the bottom which looks bad

## Solution
* Update `generateIcons.ts` to dump `@vscode/codicons`
    * All codicons retrieved with `getIcon` will now work with anything that uses `iconPath`  
* Refresh on more things when in C9
* Omit dev tools auth node in C9
* Add `description` to tree node tooltip

## License


JadenSimon:
This added ~300kb to the VSIX due to the generated C9 icons.
Also icons in the tree view look wonky if not every node has an icon :(

src/codewhisperer/views/securityPanelViewProvider.ts
```diff
@@ -49,13 +49,7 @@ export class SecurityPanelViewProvider implements vscode.WebviewViewProvider {
         }
         if (this.extensionContext) {
             this.codiconUri = webviewView.webview.asWebviewUri(
-                vscode.Uri.joinPath(
-                    this.extensionContext.extensionUri,
-                    'node_modules',
```
JadenSimon, line 54: `node_modules` don't get bundled so this was broken

scripts/build/generateIcons.ts
```diff
@@ -180,6 +194,22 @@ class GeneratedFilesManifest {
     }
 }
 
-generate().catch(error => {
-    console.error('Failed to generate icons: %s', (error as Error).message)
-})
+async function loadCodiconMappings(): Promise<Record<string, number | undefined>> {
```
JadenSimon, line 197: We can delete this if `src/codewhisperer/views/css/securityPanel.css` is updated to not refer to `content` directly

src/awsexplorer/localExplorer.ts
```diff
@@ -36,7 +56,11 @@ export function createLocalExplorerView(rootNodes: RootNode[]): vscode.TreeView<
 
     // Cloud9 will only refresh when refreshing the entire tree
     if (isCloud9()) {
-        rootNodes.forEach(node => node.onDidChangeChildren?.(() => treeDataProvider.refresh(node)))
+        rootNodes.forEach(node => {
+            const refresh = throttle(() => treeDataProvider.refresh(node), 10)
```
justinmk3, line 60: One-line comment explaining the intent would be helpful. 

    // Delay(?) each node refresh by 10ms because ...

justinmk3 (approved):

leigaol (approved):

PR #2999, opened 2022-11-21
fix(logs): improve error logging

JadenSimon:
## Problem
* Inconsistent pattern for error logging
* Stack traces aren't emitted even when source maps are available
* `aws.viewLogsAtMessage` would show an extra error in certain situations

## Solution
* Use `%s` for error substitution so the entire object isn't printed
* Add logic to map errors in the logs depending on context
    * Source maps available and we log at error level -> output stack trace
    * Otherwise format the error in a nicer way
* Handle more edge cases in `aws.viewLogsAtMessage`

I have a follow-up to this PR but it can wait.

## License

### DIFF
--- a/src/shared/logger/commands.ts
+++ b/src/shared/logger/commands.ts
@@ -47,7 +47,7 @@ export class Logging {
         const text = editor.document.getText()
         const textStart = text.indexOf(msg)
         if (textStart === -1) {
-            this.logger.warn(`logging: unable to find message with id "${logId}"`)
+            this.logger.debug(`logging: unable to find message with id "${logId}"`)
             return
         }
 
--- a/src/shared/logger/commands.ts
+++ b/src/shared/logger/commands.ts
@@ -39,20 +39,25 @@ export class Logging {
     public async openLogId(logId: number, logUri = this.defaultLogUri) {
         const msg = this.logger.getLogById(logId, logUri)
         const editor = await this.openLogUri(logUri)
-
         if (!msg || !editor) {
             return
         }
 
         // Retrieve where the message starts by counting number of newlines
         const text = editor.document.getText()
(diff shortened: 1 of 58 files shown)


src/shared/vscode/env.ts
```diff
@@ -58,6 +58,13 @@ export function isReleaseVersion(prereleaseOk: boolean = false): boolean {
     return (prereleaseOk || !semver.prerelease(extensionVersion)) && extensionVersion !== TEST_VERSION
 }
 
+/**
+ * Returns true when source mapping is available
+ */
+export function isSourceMappingAvailable(): boolean {
+    return extensionVersion === TEST_VERSION
+}
```
justinmk3, line 66: For `npm run package -- --debug` builds we might want to scrape the stacktrace to see if it looks like it's non-minified. Eventually we should produce those debug builds as CI artifacts.

justinmk3 (approved):

src/shared/vscode/env.ts
JadenSimon: Oh we don't even need to scrape anything. Webpack can inject a variable, I just didn't add it yet.

src/shared/logger/commands.ts
```diff
@@ -39,20 +39,25 @@ export class Logging {
...
-
         if (!msg || !editor) {
             return
         }
 
         // Retrieve where the message starts by counting number of newlines
         const text = editor.document.getText()
+        const textStart = text.indexOf(msg)
+        if (textStart === -1) {
+            this.logger.warn(`logging: unable to find message with id "${logId}"`)
```
JadenSimon, line 50:
```suggestion
            this.logger.debug(`logging: unable to find message with id "${logId}"`)
```

jpinkney-aws (approved):

PR #3001, opened 2022-11-21
fix(tests): sso token cache race condition

nkomonen-amazon:
## Problem
Unit tests would sporadically fail due
to cache not being cleared when expected.
Certain methods called async clear() without
waiting, and the checks in the test executed before the clear() could complete.

Example test is `"drops expired tokens if failure was a client-fault"`
where `sut.getToken()` > `await this.refreshToken(data.token, data.registration)` > `this.cache.token.clear(this.tokenCacheKey)`

## Solution
Add 'await' to the impacted clear() calls.

## License

### DIFF
--- a/src/credentials/sso/ssoAccessTokenProvider.ts
+++ b/src/credentials/sso/ssoAccessTokenProvider.ts
@@ -104,7 +104,7 @@ export class SsoAccessTokenProvider {
             return await this.authorize(registration)
         } catch (err) {
             if (err instanceof SSOOIDCServiceException && isClientFault(err)) {
-                this.cache.registration.clear(cacheKey)
+                await this.cache.registration.clear(cacheKey)
             }
 
             throw err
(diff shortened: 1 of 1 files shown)


JadenSimon (approved):
The lack of `await` was actually intentional as evicting the cache is not important for subsequent calls, however, this could definitely cause flaky tests as the cache makes no guarantees regarding execution order. So let's use `await`.

JadenSimon:
>    1) detectCdkProjects
>       at done (d:\a\aws-toolkit-vscode\aws-toolkit-vscode\node_modules\mocha\lib\mocha.js:1007:16)
>         takes into account `output` from cdk.json to build tree.json path:
>            AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

This is another flaky test unrelated to this PR. I restarted the job.

justinmk3 (approved):

jpinkney-aws (approved):

jpinkney-aws (approved):

PR #3002, opened 2022-11-22
fix(tests): decrease occurences of unreliable CDK tests

JadenSimon:
## Problem
CDK tests sometimes fail

## Solution
Increase timeout time from 3000 -> 10000 ms

## License

### DIFF
--- a/src/test/cdk/detectCdkProjects.test.ts
+++ b/src/test/cdk/detectCdkProjects.test.ts
@@ -26,7 +26,7 @@ describe('detectCdkProjects', function () {
                     return await detectCdkProjects(dirs)
                 },
                 {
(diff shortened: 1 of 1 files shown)


JadenSimon:
> detects CDK project when cdk.json exists (903ms)
> UNRESPONSIVE extension host: starting to profile NOW
> UNRESPONSIVE extension host: received responsive event and cancelling profiling session

Seems like some sort of performance issue specific to Windows. My guess is that we're seeing poor performance due to not using the workspace folder that VS Code is initialized with during tests.

justinmk3 (approved):

jpinkney-aws (approved):

PR #3006, opened 2022-11-23
fix(auth): stop race condition when auto-connecting with IAM credentials

JadenSimon:
## Problem
Some commands may fail on the first execution if the following conditions are true:
* The user has not connected yet, either automatically or manually
* The command or its dependencies use `AwsContext` early in its execution

The easiest way to reproduce this bug is by running `Sync SAM Application` on a template after reloading the window **without** opening the AWS view.

## Solution
Ensure `LoginManager` is logged-in before returning from `tryAutoConnect`

## License


JadenSimon:
Just realized that the error I saw is actually something else. This race condition still exists but it's a lot less likely than I thought.

JadenSimon:
No longer relevant

Issue #3012, opened 2022-11-25
SAM json schema: cannot change schema in template.yaml file


justinmk3:
## Problem

Cannot manually select a  different json schema in SAM template.yaml. 

## Steps to reproduce the issue

1. Choose a different schema from the status bar:
    - ![image](https://user-images.githubusercontent.com/55561878/204043965-8563ba13-081e-44cd-b8b0-4641194dbaeb.png)
2. the schema is ignored and the selection remains `aws://sam.schema.json`

## System details (run the `AWS: About Toolkit` command)

-   OS:
-   Visual Studio Code version:
-   AWS Toolkit version:

jpinkney-aws:
I just looked through a bit of the code, unfortunately it’s technically not possible to do this because the VSCode YAML extension API takes precedence over schemas that are provided through the modelline/the vscode settings

hoffa:
> I just looked through a bit of the code, unfortunately it’s technically not possible to do this because the VSCode YAML extension API takes precedence over schemas that are provided through the modelline/the vscode settings

Can you clarify what the technical challenge is making it impossible?

jpinkney-aws:
The way the VSCode-YAML API works is, given an intellisense action (hover, validation, completion, etc), it will query all the extensions that are currently registered to use it's API and then ask them if this file is supported by their extension.

E.g. In our case, we see a SAM file in VSCode so we tell VSCode-YAML that we are the ones who should be providing the schema for this file.

Once VSCode-YAML see's that an extension is ready and available to send a schema for the given file, it disables all other modes of settings the schema (modeline, `yaml.schemas`, etc) for that file.

This is by design, since theoretically an extension should be able to recognize their files and then provide support for them, that way users don't have to manually specify schemas using the modeline or the `yaml.schemas` setting.

The problem here is we aren't able to easily override the way the VSCode-YAML API works. So right now, the VSCode-YAML API sees that our extension will send back the schema for the given SAM file and then disables all other ways to set the schema

jpinkney-aws:
It's not necessarily impossible, but the approach I had to take in https://github.com/aws/aws-toolkit-vscode/pull/3013 is definitely not an ideal one, since on every SAM intellisense request you need to read in the file, check if it has a modelline and then if it does, error out so that VSCode-YAML won't continue to use the registerContributor API for that request

hoffa:
For context, this is affecting https://github.com/aws/serverless-application-model/discussions/2645

[...]

PR #3013, opened 2022-11-25
bug(yaml): Allow modelines to override default vscode-yaml registerContributor behaviour

jpinkney-aws:
## Problem
Currently users have no way of overriding what schema gets applied to their SAM document

## Solution
This PR makes it so that SAM documents can be overriden by applying the modeline from https://github.com/aws/serverless-application-model/discussions/2645

This isn't an ideal implementation, but the registerContributor API from VSCode-YAML essentially enforces that when an extension "owns" a filetype, they are responsible for telling VSCode-YAML which schema should be used at all times. In order to circumvent this API we basically have to error out so it uses the best available backup schema it can find. In this case, the schema from the modeline. The consequence of this PR is that if the user has a SAM file that's decently large, then the intellisense will be slower because it has to read the user files on every single request.

fixes: https://github.com/aws/aws-toolkit-vscode/issues/3012

## License

### DIFF
--- a/.changes/next-release/Bug Fix-25801377-bfca-4c93-8707-6fe95c0f440c.json
+++ b/.changes/next-release/Bug Fix-25801377-bfca-4c93-8707-6fe95c0f440c.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Allow modeline to override yaml schemas"
+}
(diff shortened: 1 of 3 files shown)


justinmk3:
Is there a feature request we can make to vscode-yaml to support this use-case instead?

jpinkney-aws:
I've created https://github.com/redhat-developer/yaml-language-server/pull/806. If it gets merged then I'll close this

jpinkney-aws:
Closing this as the PR on the yaml-language-server side has been approved, so this code should live there

PR #3015, opened 2022-11-28
test(GH-2090): Optimize SAM integration test

### LINKED ISSUE #2090: fix(CI): "Clone from aws-sam-cli-app-templates" step wastes time
This step is done 10+ times in `src/integrationTest/sam.test.ts`, and it can take 10+ seconds (or more, depending on network, GitHub throttling/performance):

    2021-09-28 08:58:31 [VERBOSE]: stderr: Cloning from https://github.com/aws/aws-sam-cli-app-templates

We could avoid this by using the `sam init --location` arg.

nkomonen-amazon:
The `sam init` integration tests took an
average of 7 minutes to complete. This commit
reduces them to an average of 2 minutes 30 seconds.

## Problem

The current sam cli implementation clones the entire template repo each call of sam.test.ts#createSamApplication(). This function is called more than it needs to be for the same test coverage.

## Solution
Remove redundant calls to the createSamApplication() method. Test a certain scenario only once because it does not change for each iteration of SAM project type.

Additional:
I looked in to the `sam init --location` flag and it doesn't assist in trying to test the current logic this test aims to validate. Use of `--location` is a different way to do do a `sam init`, but we are only testing Scenario 1
Initializes a new SAM project with required parameters passed as parameters:
```
//Scenario 1
$ sam init --runtime python3.7 --dependency-manager pip --app-template hello-world --name sam-app
// Scenario 2
$ sam init --location /path/to/template/folder
```
from: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-init.html

Resolves #2090

## License

### DIFF
--- a/src/integrationTest/sam.test.ts
+++ b/src/integrationTest/sam.test.ts
@@ -647,5 +647,9 @@ describe('SAM Integration Tests', async function () {
         }
         const samCliContext = getSamCliContext()
         await runSamCliInit(initArguments, samCliContext)
+        // XXX: Fixes flakiness. Ensures the files from creation of sam
+        // app are processed by code lens file watcher. Otherwise, potential
+        // issues of file not in registry before it is found.
+        await globals.codelensRootRegistry.rebuild()
     }
 })
--- a/src/integrationTest/sam.test.ts
+++ b/src/integrationTest/sam.test.ts
@@ -421,33 +421,6 @@ describe('SAM Integration Tests', async function () {
             }
 
             /**
-             * This suite cleans up at the end of each test.
-             */
-            describe('Starting from scratch', async function () {
-                let testDir: string
-
-                beforeEach(async function () {
-                    testDir = await mkdtemp(path.join(runtimeTestRoot, 'test-'))
-                    log(`testDir: ${testDir}`)
-                })
-
-                afterEach(async function () {
-                    // don't clean up after java tests so the java language server doesn't freak out
-                    if (scenario.language !== 'java') {
-                        await tryRemoveFolder(testDir)
-                    }
-                })
-
-                it('creates a new SAM Application (happy path)', async function () {
-                    await createSamApplication(testDir)
-
-                    // Check for readme file
-                    const readmePath = path.join(testDir, samApplicationName, 'README.md')
-                    assert.ok(await fileExists(readmePath), `Expected SAM App readme to exist at ${readmePath}`)
-                })
-            })
-
-            /**
              * This suite makes a sam app that all tests operate on.
              * Cleanup happens at the end of the suite.
              */
@@ -465,8 +438,11 @@ describe('SAM Integration Tests', async function () {
 
                     await createSamApplication(testDir)
                     appPath = path.join(testDir, samApplicationName, scenario.path)
+
                     cfnTemplatePath = path.join(testDir, samApplicationName, 'template.yaml')
+                    const readmePath = path.join(testDir, samApplicationName, 'README.md')
                     assert.ok(await fileExists(cfnTemplatePath), `Expected SAM template to exist at ${cfnTemplatePath}`)
+                    assert.ok(await fileExists(readmePath), `Expected SAM App readme to exist at ${readmePath}`)
 
                     samAppCodeUri = await openSamAppFile(appPath)
                 })
--- a/src/integrationTest/sam.test.ts
+++ b/src/integrationTest/sam.test.ts
@@ -397,6 +397,30 @@ describe('SAM Integration Tests', async function () {
         console.log(`DebugSessions seen in this run:\n${sessionReport}`)
     })
 
+    describe('SAM install test', async () => {
+        let runtimeTestRoot: string
+        let randomTestScenario: TestScenario
+
+        before(async function () {
+            if (scenarios.length == 0) {
+                throw new Error('There are no scenarios available.')
+            }
+            randomTestScenario = scenarios[0]
+
+            runtimeTestRoot = path.join(testSuiteRoot, 'randomScenario')
+            mkdirpSync(runtimeTestRoot)
+        })
+
+        after(async function () {
+            await tryRemoveFolder(runtimeTestRoot)
+        })
+
+        it('produces an error when creating a SAM Application to the same location', async function () {
+            await createSamApplication(runtimeTestRoot, randomTestScenario)
+            await assert.rejects(createSamApplication(runtimeTestRoot, randomTestScenario), 'Promise was not rejected')
+        })
+    })
+
     for (let scenarioIndex = 0; scenarioIndex < scenarios.length; scenarioIndex++) {
         const scenario = scenarios[scenarioIndex]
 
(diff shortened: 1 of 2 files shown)


src/integrationTest/sam.test.ts
```diff
@@ -489,7 +466,15 @@ describe('SAM Integration Tests', async function () {
                 })
 
                 it('produces an error when creating a SAM Application to the same location', async function () {
-                    await assert.rejects(createSamApplication(testDir), 'Promise was not rejected')
+                    if (alreadyTested == false) {
```
jpinkney-aws, line 469: If we only need this to run once we should probably move it out of the scenario testing

src/integrationTest/sam.test.ts
```diff
@@ -465,8 +439,11 @@ describe('SAM Integration Tests', async function () {
 
                     await createSamApplication(testDir)
                     appPath = path.join(testDir, samApplicationName, scenario.path)
+
                     cfnTemplatePath = path.join(testDir, samApplicationName, 'template.yaml')
+                    const readmePath = path.join(testDir, samApplicationName, 'README.md')
                     assert.ok(await fileExists(cfnTemplatePath), `Expected SAM template to exist at ${cfnTemplatePath}`)
+                    assert.ok(await fileExists(readmePath), `Expected SAM App readme to exist at ${readmePath}`)
```
justinmk3, line 446: Good idea. It is often more maintainable to group related asserts in one medium-sized--or bigger--test, than to have many granular tests. And often has performance benefits too.

src/integrationTest/sam.test.ts
```diff
@@ -28,6 +28,7 @@ import { insertTextIntoFile } from '../shared/utilities/textUtilities'
 import { sleep } from '../shared/utilities/timeoutUtils'
 import globals from '../shared/extensionGlobals'
 import { closeAllEditors } from '../test/testUtil'
+import { randomUUID } from 'crypto'
```
justinmk3, line 31: nit: other places where a uuid is needed use this package: https://github.com/aws/aws-toolkit-vscode-staging/blob/c617c3180f0dd8c6313f8f121e46176fd7752a83/src/shared/telemetry/util.ts#L11

Maybe we don't actually need that package? Else it's good to be consistent

src/integrationTest/sam.test.ts
```diff
@@ -397,246 +397,241 @@ describe('SAM Integration Tests', async function () {
...
-                console.log(`sam.test.ts: scenario ${scenarioIndex} (${scenario.displayName}): ${o}`)
-            }
-
-            /**
-             * This suite cleans up at the end of each test.
-             */
-            describe('Starting from scratch', async function () {
-                let testDir: string
+    function uniqueTestDirPath(rootPath: string): string {
+        return path.join(rootPath, `scenario-${randomUUID()}`)
```
justinmk3, line 401: these tests are not parallelized, but even if they were, it should be possible to increment an integer to get a unique name (which is shorter / more human-readable in logs). (The before() step should of course delete the dir first, if it exists).

justinmk3:
The changes look higher-risk now :) Is it possible to start with the smaller change first (with the perf gain), then continue this refactor in another PR?

nkomonen-amazon:
> The changes look higher-risk now :) Is it possible to start with the smaller change first (with the perf gain), then continue this refactor in another PR?

@justinmk3 I agree looking back at the change it is harder to read. I've separated changes in to more discrete commits, each with an explanation regarding what is being changed.

I'm not exactly sure what this team prefers (multiple discrete commits vs. multiple PRs), but let me know how you do things and I can change it if needed.

src/integrationTest/sam.test.ts
```diff
@@ -397,246 +397,206 @@ describe('SAM Integration Tests', async function () {
...
+
+        before(async () => {
+            runtimeTestRoot = path.join(testSuiteRoot, 'mySamApplication')
+            mkdirpSync(runtimeTestRoot)
+        })
+
+        after(async () => {
+            await tryRemoveFolder(runtimeTestRoot)
+        })
+        it('produces an error when creating a SAM Application to the same location', async function () {
```
jpinkney-aws, line 421: nitpick: spacing here

src/integrationTest/sam.test.ts
```diff
@@ -397,246 +397,206 @@ describe('SAM Integration Tests', async function () {
...
-             */
-            describe(`Starting with a newly created ${scenario.displayName} SAM Application...`, async function () {
-                let testDisposables: vscode.Disposable[]
-
-                let testDir: string
-                let samAppCodeUri: vscode.Uri
-                let appPath: string
-                let cfnTemplatePath: string
+            after(async function () {
+                await tryRemoveFolder(runtimeTestRoot)
```
jpinkney-aws, line 464: Are we able to safely remove the java check:
```typescript
if (scenario.language !== 'java') {
    await tryRemoveFolder(testDir)
}
```
I'm guessing the problem only appeared because we were creating multiple sam apps and now we're only creating one?

jpinkney-aws (commented):
Just for my own double checking, since the diff view is hard to read 😆. Besides the changes that fix the performance, you mostly just got rid of unnecessary describe blocks and were able to bring everything up a level with minor changes to the tests? (Looks like just verifying the template.yaml and readme exists) but other then that the core of the tests seem to look the same

src/integrationTest/sam.test.ts
```diff
@@ -397,246 +397,206 @@ describe('SAM Integration Tests', async function () {
...
+
+        const randomTestScenario: TestScenario = {
+            runtime: 'nodejs14.x',
+            displayName: 'nodejs14.x (ZIP)',
+            path: 'hello-world/app.js',
+            debugSessionType: 'pwa-node',
+            language: 'javascript',
+            dependencyManager: 'npm',
+            vscodeMinimum: '1.50.0',
+        }
```
JadenSimon, line 411: We need to use a known valid scenario in an invalid situation, otherwise the test may pass for the wrong reasons. Choosing a scenario we're already testing is a decent way to achieve that.
```suggestion
        const randomTestScenario = scenarios[0]
```

JadenSimon (commented):
This is a reasonable optimization considering that this better reflects how we would manually test these things. Of course, always be mindful of caching. Bad states can cause a lot of frustration.

src/integrationTest/sam.test.ts
JadenSimon: I think this logic was a hack to make the Java codelens tests work. Preserving the directory gave the Java extension time to provide document symbols. Try adding a long sleep before running the failing tests to verify this. If it works then we can figure out a better solution.

justinmk3:
Does this PR still have the perf gains? CI times look similar to current master.

nkomonen-amazon:
@justinmk3 Yeah this still has the changes. From 1 example I compared it to it saves a few minutes. This change has the tests go from ~60 git clone calls to ~20.

It should be good to go now if things look fine.

jpinkney-aws (approved):

src/shared/fs/watchedFiles.ts
```diff
@@ -221,7 +221,7 @@ export abstract class WatchedFiles<T> implements vscode.Disposable {
      * Rebuilds registry using current glob and exclusion patterns.
      * All functionality is currently internal to class, but can be made public if we want a manual "refresh" button
      */
-    private async rebuild(): Promise<void> {
+    public async rebuild(): Promise<void> {
```
justinmk3, line 224: looks like this can stay private, for now
```suggestion
    private async rebuild(): Promise<void> {
```

justinmk3:
oops I broke this, will push a fix.

justinmk3:
merged: https://github.com/aws/aws-toolkit-vscode/commit/c146b6e8edb92396d21746efdd03351c2c47475c

(couldn't push to this PR)

PR #3016, opened 2022-11-28
fix(codewhisperer): remove hover config hack

leigaol:
## Problem
Previously, CodeWhisperer disabled the hover in user config to disable the hover popup of inline suggestions when an inline completion is visible and restores it when inline completion is accepted/rejected. Reason behind this is: The hover popup of inline suggestion comes with the VS Code inline completion API, it only shows when user hovers their cursor over the suggestion which was not a very good experience for customers. 

However, the above hover workaround has problems: Restoring the hover configuration from false to true sometimes does not work. 

## Solution
Remove the hover configuration manipulation and let user see the inline suggestion hover.

https://user-images.githubusercontent.com/97199248/204353395-1b6a6cfc-1311-4da8-92c1-3689eb264e62.mov

## License


package.json
```diff
@@ -1906,6 +1906,10 @@
             ]
         },
         "commands": [
+            {
+                "command": "github.copilot.generate",
+                "title": "Open CodeWhisperer Settings"
```
justinmk3, line 1911: Does this override `github.copilot.generate` completely, or only "hook" into it (without overriding their functionality)?

leigaol: It does not override github.copilot.generate completely, but it will change the text of it even in Copilot, so I am removing it in the latest commit

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -168,6 +171,9 @@ class CodeWhispererInlineCompletionItemProvider implements vscode.InlineCompleti
             )
             this.nextMove = 0
             this._onDidShow.fire()
+            if (matchedCount >= 2 || RecommendationHandler.instance.hasNextToken()) {
```
zixlin7, line 174: just for my understanding: what is this for?

leigaol: To highlight the Left right Arrow if there are >=2 suggestions

zixlin7 (approved):

justinmk3 (approved):

PR #3022, opened 2022-11-30
fix(codewhisperer): add do not show again in access token migration popup

leigaol:
## License

### DIFF
--- a/src/codewhisperer/activation.ts
+++ b/src/codewhisperer/activation.ts
@@ -246,17 +246,30 @@ export async function activate(context: ExtContext): Promise<void> {
             }
 
             await vscode.commands.executeCommand('aws.codeWhisperer.refreshRootNode')
+            const doNoShowAgain =
+                context.extensionContext.globalState.get<boolean>(
+                    CodeWhispererConstants.accessTokenMigrationDoNotShowAgainKey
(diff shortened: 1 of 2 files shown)


src/codewhisperer/activation.ts
```diff
@@ -247,16 +247,44 @@ export async function activate(context: ExtContext): Promise<void> {
 
             await vscode.commands.executeCommand('aws.codeWhisperer.refreshRootNode')
             const t = new Date()
+            const doNotShowAgain =
+                context.extensionContext.globalState.get<boolean>(
+                    CodeWhispererConstants.accessTokenMigrationDoNotShowAgainKey
```
justinmk3, line 252: Normally I suggest `aws.suppressPrompts` for this. https://github.com/aws/aws-toolkit-vscode/blob/c146b6e8edb92396d21746efdd03351c2c47475c/package.json#L220

then you can use `isPromptEnabled()` https://github.com/aws/aws-toolkit-vscode/blob/c146b6e8edb92396d21746efdd03351c2c47475c/src/shared/settings.ts#L460

However, since this is a temporary situation, maybe it shouldn't be part of settings.

justinmk3 (approved):

PR #3023, opened 2022-11-30
feat(auth): discover accounts/roles linked to IAM Identity Center

JadenSimon:
## Problem
Users can add connections to IAM Identity Center but there's no easy way to use them for the AWS explorer.

## Solution
Automatically discover linked accounts/roles and list them in the connections picker. 

Screenshot is slightly different from the current implementation. The most recent change does not have the `sso:` prefix and uses a generic label for SSO connections instead of the exact name.

Screen Shot 2022-11-30 at 10 57 02 AM

## License


jacobrillema:
Having similar issues, any ETA on when this will be merged into the master branch?

twu13:
Also experiencing this, would love to see this fix!

scottburton11:
Would love to see this released, it's a blocker for any team with non-`us-east-1` Identity Centers

justinmk3 (approved):

Issue #2572, opened 2022-04-08
Lambda: user can "Copy URL" on Lambda functions


justinmk3:
## Description

Lambda introduced a "Function URL" feature which allows HTTP calls to the function without API Gateway.

https://aws.amazon.com/blogs/aws/announcing-aws-lambda-function-urls-built-in-https-endpoints-for-single-function-microservices/

<img width="580" alt="image" src="https://user-images.githubusercontent.com/55561878/162527626-d9e12a87-377f-430a-ae93-bd856d024924.png">

## Expected behavior

On APIGateway in AWS Explorer, we have a `Copy URL` feature (see below). Perhaps Lambda nodes should have a `Copy URL` (with `enablement=false` if the Lambda doesn't have a URL).

<img width="289" alt="image" src="https://user-images.githubusercontent.com/55561878/162528115-cd373c70-9200-406b-a8ca-019e1bb3e155.png">

justinmk3:
Available in AWS Toolkit 1.63

PR #3036, opened 2022-12-06
feat(#2572): Lambda copy url

nkomonen-amazon:
## Problem
We want to be able to copy a Lambda function URL
from within the Explorer

## Solution
Allow users to copy the lambda function URL to their
clipboard.

### Context Menu
Screen Shot 2022-12-05 at 7 00 19 PM

### If no URL was created for the Lambda
Screen Shot 2022-12-05 at 7 00 46 PM

### A quick pick appears when there are aliases of the function that also have their own URLs
Screen Shot 2022-12-07 at 9 15 03 PM

## License


src/shared/clients/lambdaClient.ts
```diff
@@ -74,6 +74,33 @@ export class DefaultLambdaClient {
...
+            const response = await request.promise()
+            // prune `Code` from logs so we don't reveal a signed link to customer resources.
+            getLogger().debug('GetFunctionUrlConfig returned response (code section pruned): %O', {
+                ...response,
+                Code: 'Pruned',
+            })
+            return response.FunctionUrlConfigs
+        } catch (e) {
+            getLogger().error('Failed to get function url config: %s', e)
+            throw e
```
JadenSimon, line 92:
```suggestion
            throw ToolkitError.chain(e, 'Failed to get Lambda function URLs')
```

src/shared/clients/lambdaClient.ts
```diff
@@ -74,6 +74,33 @@ export class DefaultLambdaClient {
...
     }
 
+    public async getFunctionUrlConfigs(name: string): Promise<Lambda.FunctionUrlConfigList> {
+        getLogger().debug(`GetFunctionUrlConfig called for function: ${name}`)
+        const client = await this.createSdkClient()
+
+        try {
+            const request = client.listFunctionUrlConfigs({ FunctionName: name })
+            const response = await request.promise()
+            // prune `Code` from logs so we don't reveal a signed link to customer resources.
```
JadenSimon, line 84: Nice.

src/lambda/commands/copyLambdaUrl.ts
```diff
@@ -0,0 +1,26 @@
...
+import globals from '../../shared/extensionGlobals'
+import { copyToClipboard } from '../../shared/utilities/messages'
+import { addCodiconToString } from '../../shared/utilities/textUtilities'
+
+export const NO_URL_INFO_MSG = 'No URL for Lambda function.'
+
+export async function copyLambdaUrl(
+    node: LambdaFunctionNode,
+    client: LambdaClient = new DefaultLambdaClient(node.regionCode),
+    copyToClipboardFunc: (data: string, label?: string) => Promise<void> = copyToClipboard
```
JadenSimon, line 17: Avoid explicit type annotations unless necessary
```suggestion
    copyToClipboardFunc = copyToClipboard
```

src/test/lambda/commands/copyLambdaUrl.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+                assert.deepStrictEqual(client.getFunctionUrl.args, [[nodeName]])
+                assert.deepStrictEqual(clipboardFunc.args, [])
+                assert.deepStrictEqual(spiedInformationMessage.args, [[NO_URL_INFO_MSG]])
+                assert.deepStrictEqual(spiedStatusBarMessage.args, [
+                    [addCodiconToString('circle-slash', NO_URL_INFO_MSG), 5000],
+                ])
+            })
+        })
+    })
```
JadenSimon, line 68: If we change `copyLambdaUrl` to throw, then these tests can be simplified:
```suggestion
    describe("URL doesn't exist", async () => {
        const emptyValues = [undefined, '']
        emptyValues.forEach(val => {
            it(`URL: '${val}'`, async () => {
                client.getFunctionUrl.resolves(val)
                await assert.rejects(() => copyLambdaUrl(node, client, clipboardFunc), ToolkitError)
            })
        })
    })
```
If you wanted to test the error in more detail then you could do this:
```ts
const result = copyLambdaUrl(node, client, clipboardFunc).catch(e => e)
assert.ok(result instanceof ToolkitError)
assert.strictEqual(result.code, 'NoURLFound')
// etc.
```

src/test/shared/clients/defaultLambdaClient.test.ts
```diff
@@ -0,0 +1,98 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { DefaultLambdaClient, LambdaClient } from '../../../shared/clients/lambdaClient'
+import { stub, createStubInstance, SinonStubbedInstance } from 'sinon'
+import { Lambda, Request } from 'aws-sdk'
+import * as assert from 'assert'
+import { FunctionUrlConfigList, FunctionUrlConfig, ListFunctionUrlConfigsResponse } from 'aws-sdk/clients/lambda'
+
+describe('Test the default lambda client', async () => {
```
JadenSimon, line 12: This is a lot of setup and tear-down to test only a few lines of logic. Can we simplify somehow?

src/test/shared/clients/defaultLambdaClient.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    })
+})
+
+/**
+ * Builds an instance of {@link FunctionUrlConfig} without the
+ * need to define all values.
+ *
+ * @param options key + value of {@link FunctionUrlConfig}
+ */
+function buildFunctionUrlConfig(options: any): FunctionUrlConfig {
```
JadenSimon, line 90:
```suggestion
function buildFunctionUrlConfig(options: Partial<FunctionUrlConfig>): FunctionUrlConfig {
```

src/shared/clients/lambdaClient.ts
```diff
@@ -74,6 +74,33 @@ export class DefaultLambdaClient {
...
+                Code: 'Pruned',
+            })
+            return response.FunctionUrlConfigs
+        } catch (e) {
+            getLogger().error('Failed to get function url config: %s', e)
+            throw e
+        }
+    }
+
+    public async getFunctionUrl(name: string): Promise<string | void> {
```
JadenSimon, line 96: We could remove this method and instead put the logic in `copyLambdaUrl.ts` directly. The less public methods the client wrapper has, the better.

src/lambda/commands/copyLambdaUrl.ts
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { DefaultLambdaClient, LambdaClient } from '../../shared/clients/lambdaClient'
+import { LambdaFunctionNode } from '../explorer/lambdaFunctionNode'
+import globals from '../../shared/extensionGlobals'
+import { copyToClipboard } from '../../shared/utilities/messages'
+import { addCodiconToString } from '../../shared/utilities/textUtilities'
+
+export const NO_URL_INFO_MSG = 'No URL for Lambda function.'
```
JadenSimon, line 12: We've been avoiding all caps for identifiers:
```suggestion
export const noUrlMessage = 'No URL for Lambda function.'
```

Also, this doesn't need an identifier if we don't export it. It's ok to inline strings directly if they're only used once. This makes things easier to read.

src/test/shared/clients/defaultLambdaClient.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    const regionCode: string = 'my-region-code'
+    const lambdaFuncName = 'lambda-func-name'
+    let lambdaApiMock: SinonStubbedInstance<Lambda>
+
+    beforeEach(async () => {
+        client = new DefaultLambdaClient(regionCode)
+        lambdaApiMock = stub(new Lambda())
+
+        // '<any>' stops linter warning since method is private
+        stub(client, <any>'createSdkClient').resolves(lambdaApiMock)
```
JadenSimon, line 23: Stubbing a private method is usually a bad idea. We should think of ways to test our implementation without needing to know every little detail of the implementation.

src/shared/clients/lambdaClient.ts
justinmk3: > The less public methods the client wrapper has, the better.

why?

.changes/next-release/Feature-9adfbb52-6012-4ddd-b1aa-d18a5e870961.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Copy Lambda Function URL in Explorer"
```
justinmk3, line 3:
```suggestion
	"description": "Copy Lambda Function URL in AWS Explorer"
```

src/lambda/commands/copyLambdaUrl.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export async function copyLambdaUrl(
+    node: LambdaFunctionNode,
+    client: LambdaClient = new DefaultLambdaClient(node.regionCode),
+    copyToClipboardFunc: (data: string, label?: string) => Promise<void> = copyToClipboard
+): Promise<void> {
+    const url = await client.getFunctionUrl(node.name)
+    if (!url) {
```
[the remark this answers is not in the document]
justinmk3: Alternatively this could be part of the lambda node info, and the `Copy URL` action could be grayed-out (`enablement=false`) if there is no URL for the given lambda.

src/test/shared/extensionManifest.test.ts
```diff
@@ -0,0 +1,65 @@
...
+        assert(packageNlsJson['AWS.generic.copyUrl'] == 'Copy URL')
+    })
+})
+
+function assertContains(arr: object[], o: object): void {
+    const oString = JSON.stringify(o)
+    assert(
+        Boolean(
+            arr.find(e => {
+                return JSON.stringify(e) == oString
```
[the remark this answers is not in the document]
justinmk3: maybe we need something like haskell's "lens" library (e.g. https://github.com/hatashiro/lens.ts ) , but until then just a generic `get(nestedKey)` would be useful.

src/lambda/commands/copyLambdaUrl.ts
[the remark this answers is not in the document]
nkomonen-amazon: This was one of the discussions I was having with Josh and this is what we concluded.

### To implement Justins approach
- [Update this function](https://github.com/nkomonen-amazon/aws-toolkit-vscode/blob/b346e3b1e3520b9f0bb291127faebf6700649835/src/lambda/explorer/lambdaNodes.ts#L74) and the method above `updateInPlace()`. This will be to allow us to pass in an object that has the Function URL info (`FunctionUrlConfig`).
- Update how [this property works](https://github.com/nkomonen-amazon/aws-toolkit-vscode/blob/b346e3b1e3520b9f0bb291127faebf6700649835/src/lambda/explorer/lambdaNodes.ts#L74) by creating an `updateContextValue()` method which would take a list of values and then join them. This will allow us to not affect the existing features but we can add additional values such as a string that looks like `"awsRegionFunctionNode/withFunctionUrl"`
- Then in the [`ViewItem`](https://github.com/nkomonen-amazon/aws-toolkit-vscode/blob/b346e3b1e3520b9f0bb291127faebf6700649835/package.json#L1487) pattern we'd need to update all the existing patterns from `/^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable)$/"` to `/(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable)/"` for the existing ones.

There is some extra work to implement this. I can more commits to this PR or open up a new one following this.

---

For Jaden's I agree to adding in a link to set up the Lambda URL if it doesn't exist.

src/test/shared/extensionManifest.test.ts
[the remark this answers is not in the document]
nkomonen-amazon: I've updated to use lodash.isEqual to do a deep comparison instead.

src/shared/clients/lambdaClient.ts
nkomonen-amazon: I agree Jaden, I've removed this function and instead find and grab the function URL in the business logic.

src/lambda/commands/copyLambdaUrl.ts
nkomonen-amazon: Removed this const in favor of putting it inline

src/shared/clients/lambdaClient.ts
```diff
@@ -74,6 +74,33 @@ export class DefaultLambdaClient {
...
+        const urls = await this.getFunctionUrlConfigs(name)
+        if (urls.length === 0) {
+            return undefined
+        }
+        return urls[0].FunctionUrl
```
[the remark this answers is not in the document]
nkomonen-amazon: moot due to removing this function

package.json
```diff
@@ -1478,6 +1482,11 @@
                     "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable)$/",
                     "group": "4@1"
                 },
+                {
+                    "command": "aws.copyLambdaUrl",
+                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable)$/",
+                    "group": "2@0"
```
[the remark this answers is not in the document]
nkomonen-amazon: This is a good idea. Though it will be a bit more work to implement this since the URL is not part of the Node instance, and instead each Node type will have their own implementation to get a URL.

I think we could do some function overloading, or something, once we start to have more copyUrl implementations.

src/shared/clients/lambdaClient.ts
```diff
@@ -74,6 +74,33 @@ export class DefaultLambdaClient {
...
+            })
+            return response.FunctionUrlConfigs
+        } catch (e) {
+            getLogger().error('Failed to get function url config: %s', e)
+            throw e
+        }
+    }
+
+    public async getFunctionUrl(name: string): Promise<string | void> {
+        const urls = await this.getFunctionUrlConfigs(name)
```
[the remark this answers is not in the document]
nkomonen-amazon: I've looked deeper in to the reason for multiple URL configs:

- The existence of Function URLs is opt-in (as you mentioned)
- If no function urls were created then `getFunctionUrlConfigs` would be an empty list
- I can get multiple `functionUrlConfig` instances if I:
  - Create an alias of a function
  - Create a url for that as well
- I've updated the code to create a quick pick in the scenario there are multiple possible URLs to copy.

src/lambda/commands/copyLambdaUrl.ts
```diff
@@ -0,0 +1,26 @@
...
+        globals.window.showWarningMessage(NO_URL_INFO_MSG)
+        globals.window.setStatusBarMessage(addCodiconToString('circle-slash', NO_URL_INFO_MSG), 5000)
+        return
```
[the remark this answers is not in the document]
nkomonen-amazon: Not sure if my [comment here](https://github.com/aws/aws-toolkit-vscode/pull/3036#discussion_r1042467640) changes the rationale to throw an error?

If no URL exists then we can direct them to the page on how to create the URL (I've updated the message to include the link to the docs as you mentioned before)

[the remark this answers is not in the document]
jpinkney-aws: Given that the lambda URL is opt-in, I don't think we can necessarily say that if there isn't a function URL it's an error (with the current code). I would have expected an info message saying something like "Function URL is not set-up: heres some instructions" (obviously in better wording).

I think in part, the issue is that it's not easy to ahead of time check enablement of the `Copy URL` right click option using the lambda client. We can technically list every lambda in one call, but checking if the lambda has a function URL requires another API call for every single lambda in the list (technically feasible, but I don't know enough about the AWS API to know if its costly or not). If we had a way to ahead of time check enablement of the node (if the node has a function URL), then, if we clicked a node and it didn't end up having a function url it would end up being an error. Without that implementation we can't really say this current state is an error.

src/test/shared/clients/defaultLambdaClient.test.ts
nkomonen-amazon: I've update the `createSdkClient()` in a separate commit. It now returns a singleton instance. Also you can pass in your own implementation to the class constructor, this will allow us to inject the mock instead of stubbing it as I did here.

src/lambda/commands/copyLambdaUrl.ts
```diff
@@ -0,0 +1,51 @@
...
+import { LambdaFunctionNode } from '../explorer/lambdaFunctionNode'
+import globals from '../../shared/extensionGlobals'
+import { copyToClipboard as _copyToClipboard } from '../../shared/utilities/messages'
+import { addCodiconToString } from '../../shared/utilities/textUtilities'
+import { createQuickPick, QuickPickPrompter } from '../../shared/ui/pickerPrompter'
+import { isValidResponse } from '../../shared/wizards/wizard'
+import { FunctionUrlConfigList } from 'aws-sdk/clients/lambda'
+
+export async function copyLambdaUrl(
+    node: LambdaFunctionNode,
```
JadenSimon, line 16: nit:

Do we have to pass in a `LambdaFunctionNode`, or could we use something simpler? Looks like the only thing that's really needed is the function name + region:
```suggestion
    node: Pick<LambdaFunctionNode, 'name' | 'regionCode'>,
```
Now for tests you can use `{ name: 'myFunction', regionCode: 'myRegion' }` instead of a stub

nkomonen-amazon: Nice recommendation, I've updated to use this.

src/lambda/commands/copyLambdaUrl.ts
```diff
@@ -0,0 +1,51 @@
...
+
+async function _quickPickUrl(configList: FunctionUrlConfigList): Promise<string | undefined> {
+    const items = configList.map(c => ({
+        label: c.FunctionArn,
+        data: c.FunctionUrl,
+    }))
+    const picker: QuickPickPrompter<string> = createQuickPick(items, { title: 'Select function to copy url from.' })
+    const res = await picker.prompt()
+    return isValidResponse(res) ? res : undefined
+}
```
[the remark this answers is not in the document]
nkomonen-amazon: Updated with your change. I'll open a sim to look in to unifying the quick pick process.

src/test/lambda/extensionManifest.test.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as assert from 'assert'
+import * as packageJson from '../../../package.json'
+import * as packageNlsJson from '../../../package.nls.json'
+import * as lodash from 'lodash'
+
+describe('Test Copy Lambda Func URL in extension manifest (package.json)', async () => {
```
jpinkney-aws, line 11: I wonder if we should just test the very minimum amount of functionality here that we can, that way we aren't tying the package.json for lambdaURL directly to this test. And when one changes we have to change this and vice versa.

E.g. testing if the command title exists, it's disabled in the command palette (this could be useful for creating tests in C9, where we could open a test file and see the list of things disabled), testing that the command exists in the menu without testing the entirety of what is contained

It technically is a bit looser of a test but i'm not sure if its worth it for us to eventually maintain strict tests for everything that goes into the package.json

src/test/shared/clients/defaultLambdaClient.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+describe('Test the default lambda client', async () => {
+    let client: LambdaClient
+    const regionCode: string = 'my-region-code'
+    const lambdaFuncName = 'lambda-func-name'
+    let lambdaApiMock: SinonStubbedInstance<Lambda>
+
+    beforeEach(async () => {
+        lambdaApiMock = stub(new Lambda())
+        client = new DefaultLambdaClient(regionCode, lambdaApiMock as any)
```
jpinkney-aws, line 20: Instead of the "any" cast, is it possible to cast to Lambda? Or does that cause an error?

package.json
```diff
@@ -2454,6 +2463,16 @@
                     }
                 }
             },
+            {
+                "command": "aws.copyLambdaUrl",
+                "title": "%AWS.generic.copyUrl%",
```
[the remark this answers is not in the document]
nkomonen-amazon: My rationale for this was that 'Copy URL' is a generic command. API gateway defines its own title for 'Copy URL' as well under `AWS.command.apig.copyUrl`. So eventually as we add more copy url commands they can all use the same one since it is just the display name in the context menu.

src/lambda/commands/copyLambdaUrl.ts
```diff
@@ -0,0 +1,55 @@
...
+    const configs = await client.getFunctionUrlConfigs(node.name)
+
+    if (configs.length == 0) {
+        globals.window.showWarningMessage(
+            'No URL for Lambda function. [How to create URL.](https://docs.aws.amazon.com/lambda/latest/dg/urls-configuration.html)'
+        )
+        globals.window.setStatusBarMessage(addCodiconToString('circle-slash', 'No URL for Lambda function.'), 5000)
+    } else {
+        let url: string | undefined = undefined
+        const doQuickPick = configs.length > 1
```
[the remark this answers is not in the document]
nkomonen-amazon: I'd like to hear your thoughts on this. I tried this way because I think it better explains 'what' the condition is trying to represent. The condition is now in English and I see it as easier to read since you don't need to guess what happens if the if statement is True.

 You don't know what happens when `if (configs.length > 1 )` is true.
But when you see, `if (doQuickPick)`, you know it will perform a quickpick if the statement is true.

I guess I could also do a comment, but let me know if you think this still isn't necessary. I have no strong feelings about this.

src/test/shared/clients/defaultLambdaClient.test.ts
nkomonen-amazon: It gave an error, but I found another way that I've updated it to `<Lambda><unknown> lambdaApiMock` which converts it to an 'unknown' then the type we want. 

Source: https://github.com/microsoft/TypeScript/issues/28067

src/lambda/commands/copyLambdaUrl.ts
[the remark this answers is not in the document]
JadenSimon: > The condition is now in English and I see it as easier to read since you don't need to guess what happens if the if statement is True.

I think your overall reasoning about this makes perfect sense: we definitely want to make logic easy for others to read. A lot of this comes down to personal preference though I think there are some general guidelines that are applicable to most developers. In this case I'd argue that `doQuickPick` is _harder_ to read because it doesn't tell me anything more than what the logic already says:
```ts
        if (configs.length > 1) {
            url = await quickPickUrl(configs)
        }
```
Which reads as "if there's more than one config, prompt the user". This snippet stands perfectly well on its own IMO. Also keep in mind that code is not always read from top to bottom. Sometimes we want to know how things work starting at the end result.

src/test/lambda/extensionManifest.test.ts
JadenSimon: > It technically is a bit looser of a test but i'm not sure if its worth it for us to eventually maintain strict tests for everything that goes into the package.json

Yeah we generally want tests to not necessarily be lenient but rather precise in what they're testing. A test for command declaration shouldn't fail if I change the command title as that's not what I'm testing.

src/shared/clients/lambdaClient.ts
```diff
@@ -96,7 +115,10 @@ export class DefaultLambdaClient {
         }
     }
 
-    private async createSdkClient(): Promise<Lambda> {
-        return await globals.sdkClientBuilder.createAwsService(Lambda, undefined, this.regionCode)
+    private async getSdkClient(): Promise<Lambda> {
+        if (!this.client) {
+            this.client = await globals.sdkClientBuilder.createAwsService(Lambda, undefined, this.regionCode)
+        }
+        return this.client
```
justinmk3, line 122: is this for performance? I would skip this for now. The SDKs cached some things so it's not always clear when a new client is needed or not. 

Eventually, we will solve this with an abstraction that generalizes this. Meanwhile, it may avoid some pain if we use the same pattern in all of the `fooClient.ts` modules.

src/test/lambda/commands/copyLambdaUrl.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    it('Single URL exists', async () => {
+        const urlConfig = buildFunctionUrlConfig({ FunctionUrl: 'url1', FunctionArn: 'arn1' })
+        client.getFunctionUrlConfigs.resolves([urlConfig])
+
+        await copyLambdaUrl(node, client, clipboardFunc)
+
+        assert(client.getFunctionUrlConfigs.calledOnceWith(nodeName))
+        assert(clipboardFunc.calledOnceWith(urlConfig.FunctionUrl, 'URL'))
```
justinmk3, line 38: should be able to actually check the clipboard here. Then we don't need these 2 spies.

src/test/lambda/commands/copyLambdaUrl.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        const expectedConfig = buildFunctionUrlConfig({ FunctionUrl: 'url1', FunctionArn: 'arn1' })
+        const urlConfigs = [expectedConfig, buildFunctionUrlConfig({ FunctionUrl: 'url2', FunctionArn: 'arn2' })]
+        client.getFunctionUrlConfigs.resolves(urlConfigs)
+        quickPickUrlFunc.resolves(expectedConfig.FunctionUrl)
+
+        await copyLambdaUrl(node, client, clipboardFunc, quickPickUrlFunc)
+
+        assert(client.getFunctionUrlConfigs.calledOnceWith(nodeName))
+        assert.deepStrictEqual(quickPickUrlFunc.args, [[urlConfigs]])
+        assert.deepStrictEqual(clipboardFunc.args, [[expectedConfig.FunctionUrl, 'URL']])
```
justinmk3, line 51: check the actual clipboard, don't need the stubs

src/test/lambda/commands/copyLambdaUrl.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        it(`URL doesn't exist`, async () => {
+            client.getFunctionUrlConfigs.resolves([])
+
+            await copyLambdaUrl(node, client, clipboardFunc)
+
+            assert.deepStrictEqual(client.getFunctionUrlConfigs.args, [[nodeName]])
+            assert.deepStrictEqual(clipboardFunc.args, [])
+            assert.deepStrictEqual(spiedInformationMessage.args, [[noLambdaFuncMessage]])
+            assert.deepStrictEqual(spiedStatusBarMessage.args, [
```
justinmk3, line 76: check actual clipboard. then we can eliminate these 4 spies and possibly some/all of the stubs.

src/test/lambda/extensionManifest.test.ts
```diff
@@ -0,0 +1,33 @@
...
```
justinmk3, line 11: - avoid verbose names.
- tests don't need to mention "test" in the name.

```suggestion
describe('package.json', async () => {
```

Here the describe() block is just "package.json", because the it() blocks below add specific context.

src/test/lambda/extensionManifest.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    it('title is in package.nls.json', async () => {
+        assertPackageNlsContains('AWS.generic.copyUrl')
+    })
+
+    it('is disabled in Command Palette', async () => {
+        assertContains(packageJson.contributes.menus.commandPalette, {
+            command: 'aws.copyLambdaUrl',
+            when: 'false',
```
justinmk3, line 19: Don't think this test is needed. It is asserting static data.

Sometimes it's useful to assert static things if the test gives a useful "view" of complicated interlocking objects--but that is not the case here.

src/lambda/commands/copyLambdaUrl.ts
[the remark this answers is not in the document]
justinmk3: > Update how [this property works](https://github.com/nkomonen-amazon/aws-toolkit-vscode/blob/b346e3b1e3520b9f0bb291127faebf6700649835/src/lambda/explorer/lambdaNodes.ts#L74) by creating an updateContextValue()

Hopefully we can avoid `updateContextValue()` and just change `updateChildren()` to get the lambda info and set a `hasUrl` property.

But yes, let's do that in a followup PR.

src/shared/clients/lambdaClient.ts
nkomonen-amazon: This change was done due to a previous PR comment recommending to not stub the `getSdkClient()` method since it was private. The goal was to allow me to pass in a mock client for testing, otherwise it created its own. 
But now I see this has a side effect due to the SDK caching you mentioned.

Do you recommend I revert this and stub the private method, or do you see another way to do this? I need to stub the client either way since it makes a network call.

nkomonen-amazon: I didn't realize we needed to worry about cache clearance. I've removed this change and instead provide the option to provide a client to the method itself, this is so I can still test without needing to make an actual network call.

src/test/lambda/commands/copyLambdaUrl.test.ts
nkomonen-amazon: I've removed the spy for the clipboard func.
I don't see a way to test the bar+info message without spying, did you have something in mind?

src/test/shared/clients/defaultLambdaClient.test.ts
```diff
@@ -0,0 +1,72 @@
...
+            [buildFunctionUrlConfig({ FunctionUrl: 'url1' })],
+            [buildFunctionUrlConfig({ FunctionUrl: 'url1' }), buildFunctionUrlConfig({ FunctionUrl: 'url2' })],
+        ]
+
+        testCases.forEach(functionUrlConfigs => {
+            it('gets the expected url config list', async () => {
+                stubListFunctionUrlConfigs(functionUrlConfigs)
+
+                // Call to test
+                const result = await client.getFunctionUrlConfigs(lambdaFuncName, <Lambda>(<unknown>lambdaApiMock))
```
justinmk3, line 49: similar to Jaden's comment above, this isn't testing enough to justify the cost of the test. It's good to exercise codepaths however we can, but in this case we should wait until we have a general strategy for testing SDKs.

The code in `getFunctionUrlConfigs` is quite small. It's not sustainable to maintain so much test code without a proportionate gain. If the code in `getFunctionUrlConfigs` were non-trivial, typically I would suggest changing its interface to take SDK result and pass that in from test, so that we don't need all this mocking in the test.

tl;dr: let's drop this test file for now. We will eventually take time to find a way to maintainably test SDK client code.

src/test/lambda/commands/copyLambdaUrl.test.ts
justinmk3: no. looks better now!

src/test/shared/clients/defaultLambdaClient.test.ts
nkomonen-amazon: Makes sense to me, I've dropped the file.

PR #3038, opened 2022-12-06
feat(dev): add logic to download beta artifacts from a URL

JadenSimon:
## Problem
We have no easy way to share beta builds with users

## Solution
Add codepath to download artifact from a URL. The user is prompted to install the new artifact if it is different from the currently installed extension.

**This codepath is never accessible in a release version**
~The dev setting `aws.dev.betaUrl` can be used to test this logic assuming the currently installed VSIX is a pre-release~
Only `src/dev/config.ts` is used to specify the beta URL to ensure that this behavior is dependent on compilation.

### Packaging
A 'config' file has been added to specify development constants. Keeping this as a separate file makes it easier to catch potential leaks and clean the git history of possibly sensitive information. If `betaUrl` is specified then the VSIX will be built with an artificially high version to stop auto-updates e.g. `1.999.0-xxxxxxxx`.

## License

### DIFF
--- a/scripts/build/package.ts
+++ b/scripts/build/package.ts
@@ -4,9 +4,11 @@
  */
 
 //
-// Creates an artifact that can be given to users for testing alpha builds:
+// Creates an artifact that can be given to users for testing alpha/beta builds:
 //
-//     aws-toolkit-vscode-1.99.0-SNAPSHOT.vsix
+//     aws-toolkit-vscode-1.999.0-xxxxxxxxxxxx.vsix
+//
+// Where `xxxxxxxxxxxx` is the first 12 characters of the commit hash that produced the artifact
 //
 // The script works like this:
 // 1. temporarily change `version` in package.json
@@ -101,13 +103,12 @@ function main() {
             const packageJson: typeof manifest = JSON.parse(fs.readFileSync(packageJsonFile, { encoding: 'utf-8' }))
             const versionSuffix = getVersionSuffix()
             const version = packageJson.version
-            if (isBeta()) {
-                // Setting the version to an arbitrarily high number stops VSC from auto-updating the beta extension
-                packageJson.version = `1.999.0${versionSuffix}`
+            // Setting the version to an arbitrarily high number stops VSC from auto-updating the beta extension
+            const betaOrDebugVersion = `1.999.0${versionSuffix}`
+            if (isBeta() || args.debug) {
+                packageJson.version = betaOrDebugVersion
             } else {
-                packageJson.version = args.debug
-                    ? `1.99.0${versionSuffix}`
-                    : version.replace('-SNAPSHOT', versionSuffix)
+                packageJson.version = version.replace('-SNAPSHOT', versionSuffix)
             }
 
             if (args.skipClean) {
--- a/scripts/build/package.ts
+++ b/scripts/build/package.ts
@@ -18,6 +18,9 @@ import type * as manifest from '../../package.json'
 import * as child_process from 'child_process'
 import * as fs from 'fs-extra'
 
+// Importing from `src` isn't great but it does make things simple
+import { betaUrl } from '../../src/dev/config'
+
 const packageJsonFile = './package.json'
 const webpackConfigJsFile = './webpack.config.js'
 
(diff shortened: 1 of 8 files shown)


src/shared/settings.ts
```diff
@@ -244,6 +244,15 @@ function createSettingsClass<T extends TypeDescriptor>(section: string, descript
             return cast<Inner[K]>(value, descriptor[key])
         }
 
+        protected getOrUndefined<K extends keyof Inner>(key: K & string) {
```
JadenSimon, line 247: This change is to fix some noisy logging I noticed

src/dev/config.ts
```diff
@@ -0,0 +1,9 @@
+/*!
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// This file is strictly used for private development
+// Nothing in this file should have a truthy value on mainline
+
+export const betaUrl = ''
```
JadenSimon, line 9: Future work: replace this file with a top-level 'config.json' that is embedded into the build, setting the default values used by `DevSettings`.

justinmk3: could it be config.ts ? That way we get autocomplete etc.

src/dev/beta.ts
```diff
@@ -0,0 +1,169 @@
...
+        }
+    } else {
+        await updateBetaToolkitData(vsixUrl, {
+            lastCheck: Date.now(),
+            needUpdate: false,
+        })
+    }
+}
+
+interface ExtensionInfo {
```
jpinkney-aws, line 98: I'm kind of surprised that this doesn't seem to exist in the vscode api :thinking:

src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -49,6 +49,11 @@
                 }
             ]
         },
+        {
+            "name": "vscode_autoUpdateBeta",
```
justinmk3, line 53: it's not really vscode-specific, right? Could potentially be useful for e.g. VS, so `aws_` prefix might make sense?

src/dev/beta.ts
```diff
@@ -0,0 +1,169 @@
+/*!
```
justinmk3, line 1: Let's mention this in CONTRIBUTING.md : https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#tooling

src/dev/beta.ts
```diff
@@ -0,0 +1,169 @@
...
+        ...globals.context.globalState.get<Record<string, BetaToolkit>>(betaToolkitKey, {}),
+        [vsixUrl]: data,
+    })
+}
+
+/**
+ * Watch the beta VSIX daily for changes.
+ * If this is the first time we are watching the beta version or if its been 24 hours since it was last checked then try to prompt for update
+ */
+export function watchBetaVSIX(vsixUrl: string): vscode.Disposable {
```
justinmk3, line 47: should we add a `once()` log an INFO or WARN message here so that the logs have positive confirmation that this codepath was entered?

scripts/build/package.ts
```diff
@@ -87,7 +101,14 @@ function main() {
             const packageJson: typeof manifest = JSON.parse(fs.readFileSync(packageJsonFile, { encoding: 'utf-8' }))
             const versionSuffix = getVersionSuffix()
             const version = packageJson.version
-            packageJson.version = args.debug ? `1.99.0${versionSuffix}` : version.replace('-SNAPSHOT', versionSuffix)
+            if (isBeta()) {
+                // Setting the version to an arbitrarily high number stops VSC from auto-updating the beta extension
+                packageJson.version = `1.999.0${versionSuffix}`
+            } else {
+                packageJson.version = args.debug
+                    ? `1.99.0${versionSuffix}`
```
justinmk3, line 109: should this also be 1.999 ? The inconsistency could raise questions for future readers.

src/dev/config.ts
JadenSimon: Oh I meant `config.ts`

src/dev/beta.ts
JadenSimon: `packageJSON` exists on the extension interface but it's typed as `any`

src/dev/beta.ts
JadenSimon: Added a log statement. `once` isn't needed as this function is only executed once.

justinmk3 (approved):

PR #3039, opened 2022-12-06
fix(codecatalyst): prompt the user to onboard if they haven't already

JadenSimon:
## Problem
Users can sign into an AWS Builder ID connection with access to CodeCatalyst but they may not have onboarded yet

## Solution
Direct the user to the onboarding page if not onboarded

## License


src/codecatalyst/auth.ts
```diff
@@ -94,6 +94,23 @@ export class CodeCatalystAuthenticationProvider {
         await this.secondaryAuth.restoreConnection()
     }
 
+    public async promptOnboarding(): Promise<void> {
+        const okItem: vscode.MessageItem = { title: localizedText.ok }
+        const cancelItem: vscode.MessageItem = { title: localizedText.cancel, isCloseAffordance: true }
+        const resp = await vscode.window.showInformationMessage(
+            'The current connection is not onboarded with CodeCatalyst. This must be done in the browser.\n\n Onboard now?',
```
justinmk3, line 101: "onboard" is somewhat jargon, should we say "sign up" instead?

justinmk3, line 101: - "onboard" is somewhat jargon, should we say "sign up" instead? Or is CC using "onboard" in their materials?
- "This must be done in the browser" is a lot of words for a rather common task, perhaps we can find tighter verbiage:

```suggestion
            'The current connection is not onboarded with CodeCatalyst. Visit sign-up page now?',
```

vscode also supports hyperlinks now (and this dialog isn't needed in Cloud9), so we could do something like:

    [Sign up to CodeCatalyst](https://...), to get started.

justinmk3 (approved):

src/codecatalyst/auth.ts
JadenSimon: This works although the notification doesn't show fully expanded which is kind of annoying.

jpinkney-aws (approved):

JadenSimon:
The UX in this PR could be improved but I'd rather wait for better APIs from CodeCatalyst to do anything more.

justinmk3:
:shipit:

justinmk3 (approved):

src/codecatalyst/auth.ts
```diff
@@ -37,6 +38,8 @@ export class CodeCatalystAuthStorage {
     }
 }
 
+export const onboardingUrl = vscode.Uri.parse('https://codecatalyst.aws/onboarding/view')
```
justinmk3, line 41: Would https://codecatalyst.aws/explore be more generally applicable "landing page" for this purpose?

Issue #1351, opened 2020-10-14
VSCode source control always uses DEFAULT user profile 


jvargh:
When using VSCODE Source control to push code changes, it always goes to system associated with Default user profile, even thought AWS Toolkit has been changed to use a custom profile i.e C1

1. change from command pallete > AWS: Connect to AWS to use C1 profile, which has working keys in credentials file
2. from VSCODE Source control commit changes and push and it always goes to system associated with default profile not the currently set C1 profile

-   OS: Win10
-   Visual Studio Code Version: v1.50
-   AWS Toolkit Version: 1.15.0 2020-10-06

justinmk3:
> 2\. from VSCODE Source control commit

Can you expand on what this means exactly? 

- Are you commiting via right-click in the VCS tree?
- Are you running a command (what is the exact name of the command)? 
  - I do not see a `Commit changes` command in VSCode. I have a `Git: Commit` command, but that does not involve AWS credentials.

Are you connected to AWS CodeCommit? What component or extension are you using for that? Can you link to a doc or other resource.

jvargh:
> > 2. from VSCODE Source control commit
> 
> Can you expand on what this means exactly?
you get to it by clicking the 'Source Control' icon in the palette 

> * Are you commiting via right-click in the VCS tree?
No. from '...', drop down and 'Push' is selected. Prior to this changes are committed by adding Message and clicking the check arrow icon. On selecting Push it pushes to the profile 'default' even though using ' AWS: Connect to AWS' was used to select custom C1 profile 

> * Are you running a command (what is the exact name of the command)?
No commands. Just using UI causes the failure as per above. However on CLI setting 'set AWS_PROFILE=c1' and 'git push' is the only way it pushes changes to the right repo   

>   * I do not see a `Commit changes` command in VSCode. I have a `Git: Commit` command, but that does not involve AWS credentials.
Clicking the check arrow icon after entering commit message in text box is the Commit method used. 

> Are you connected to AWS CodeCommit? What component or extension are you using for that? Can you link to a doc or other resource.
Yes, using CodeCommit and after a git clone using SSH. As mentioned on CLI using git push works but not from the 'Source Control' toolbar picking Push. Hope this helps

justinmk3:
> on CLI setting 'set AWS_PROFILE=c1' and 'git push' is the only way it pushes changes to the right repo
> ...
> using CodeCommit and after a git clone using SSH

[...]

PR #3040, opened 2022-12-07
feat(auth): support terminal profiles and credential injection

JadenSimon:
## Problem
Users can't easily use Toolkit connections with other tools

## Solution
This PR adds two features:
* Terminals can be created with AWS credentials injected as environment variables
* Terminals can optionally have their environment variables automatically updated when the connection changes
    * This is controlled by `aws.injectCredentials` which is disabled by default

#### Running `Terminal: Create New Terminal (With Profile)`

Screen Shot 2022-12-06 at 7 19 59 PM

#### Adding a new terminal from the bottom panel
Screen Shot 2022-12-06 at 7 18 23 PM

#### A newly created terminal
Screen Shot 2022-12-06 at 7 15 50 PM

#### The credentials injection setting (disabled by default)
Screen Shot 2022-12-06 at 7 20 48 PM

#### User sees a warning when the terminal environment needs to change
Screen Shot 2022-12-06 at 7 28 03 PM

This is only applicable with `aws.injectCredentials`

## License


justinmk3:
Really nice. Probably should add this to the root menu of AWS Explorer / Developer Tools.

justinmk3:
>  `aws.injectCredentials` which is disabled by default

Let's enable it by default, it seems to be expected behavior.

We can try to help discoverability by updating the "Switch connection" menu (and perhaps other messages): 
image

src/credentials/environment.ts
```diff
@@ -0,0 +1,119 @@
...
+    env = process.env
+) {
+    return telemetry.aws_injectCredentials.run(async () => {
+        telemetry.record({ source, passive: true })
+
+        const credentials = await connection.getCredentials()
+
+        return {
+            ...env,
+            AWS_REGION: connection.defaultRegion,
```
justinmk3, line 100: to address https://github.com/aws/aws-toolkit-vscode/issues/1351 we also need to set `AWS_PROFILE` , presumably using `ExtensionContext.environmentVariableCollection` as in https://github.com/aws/aws-toolkit-vscode/pull/3139

JadenSimon: Yeah I left out `AWS_PROFILE` because the behavior may be less well-defined across tools. Not all tools support every kind of profile even if they even support `AWS_PROFILE` at all. Should be ok to add in some case though.

.changes/next-release/Feature-f61a7284-1590-467d-9ac4-88df923b4fe8.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "The `aws.injectCredentials` setting can be used to automatically inject the current AWS connection into all terminals. This is enabled by default."
```
justinmk3, line 3:
```suggestion
	"description": "The `aws.injectCredentials` setting (enabled by default) automatically injects the current AWS connection into all [terminals](https://code.visualstudio.com/docs/terminal/basics)."
```

.changes/next-release/Feature-f61a7284-1590-467d-9ac4-88df923b4fe8.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "The `aws.injectCredentials` setting (enabled by default) automatically injects the current AWS connection into all [terminals](https://code.visualstudio.com/docs/terminal/basics)."
```
justinmk3, line 3:
```suggestion
	"description": "The `aws.injectCredentials` setting (enabled by default) automatically injects the current AWS connection into all [terminals](https://code.visualstudio.com/docs/terminal/basics) as AWS [environment variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html)."
```

package.nls.json
```diff
@@ -5,6 +5,7 @@
     "AWS.productName.cn": "Amazon Toolkit",
     "AWS.codecatalyst.submenu.title": "Manage CodeCatalyst",
     "AWS.configuration.profileDescription": "The name of the credential profile to obtain credentials from.",
+    "AWS.configuration.injectCredentialsDescription": "Automatically add AWS credentials to terminals.",
```
justinmk3, line 8:
```suggestion
    "AWS.configuration.injectCredentialsDescription": "Automatically set AWS credentials [environment variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) in terminals.",
```

src/auth/environment.ts
```diff
@@ -0,0 +1,130 @@
...
+) {
+    return telemetry.aws_injectCredentials.run(async () => {
+        telemetry.record({ source, passive: true })
+
+        const credentials = await connection.getCredentials()
+
+        return {
+            env: {
+                ...env,
+                AWS_REGION: await auth.getDefaultRegion(connection),
```
justinmk3, line 122: I wonder if we should use guessDefaultRegion here. https://github.com/aws/aws-toolkit-vscode/blob/be37644158cc5b9f0eb9f33dc01eed49d02ca5aa/src/shared/regions/regionProvider.ts#L97

justinmk3 (changes requested):
Add test coverage for:

- credentials are NOT injected if the feature is disabled
- credentials are injected if the feature is enabled

PR #3042, opened 2022-12-08
cwl: Fix bug where empty results are returned with valid nextToken

nkomonen-amazon:
## Problem
Sometimes the results from cwl API return empty, with a `nextToken`. We don't account for this case. 
## Solution
Implement `getPaginatedAwsCallIter` in order to handle this case. 

**Note**: This is an addition to #2839 since it was not completed due to a failing unit test. This unit test is now resolved along with some other minor changes.

Kapture 2022-12-13 at 14 33 37

## License

### DIFF
--- a/src/cloudWatchLogs/registry/logDataRegistry.ts
+++ b/src/cloudWatchLogs/registry/logDataRegistry.ts
@@ -129,15 +129,16 @@ export class LogDataRegistry {
         const logDataGetter: AsyncIterator<CloudWatchLogsResponse> = getPaginatedAwsCallIter({
             awsCall: async request =>
                 await logData.retrieveLogsFunction(logData.logGroupInfo, logData.parameters, request.nextForwardToken),
-            nextTokenNames: headOrTail
-                ? {
-                      request: 'nextForwardToken',
-                      response: 'nextForwardToken',
-                  }
-                : {
-                      request: 'nextBackwardToken',
-                      response: 'nextForwardToken',
-                  },
+            nextTokenNames:
+                headOrTail === 'head'
+                    ? {
+                          request: 'nextForwardToken',
+                          response: 'nextForwardToken',
+                      }
+                    : {
+                          request: 'nextBackwardToken',
+                          response: 'nextBackwardToken',
+                      },
             request,
         })
         let newLogEvents: CloudWatchLogs.FilteredLogEvents = []
@@ -398,7 +399,6 @@ export type CloudWatchLogsParameters = {
     endTime?: number
     limit?: number
     streamNameOptions?: string[]
-    nextForwardToken?: CloudWatchLogs.NextToken
 }
 
 export type CloudWatchLogsResponse = {
--- a/src/cloudWatchLogs/registry/logDataRegistry.ts
+++ b/src/cloudWatchLogs/registry/logDataRegistry.ts
@@ -125,10 +125,15 @@ export class LogDataRegistry {
         const logDataGetter: AsyncIterator<CloudWatchLogsResponse> = getPaginatedAwsCallIter({
             awsCall: async request =>
                 await logData.retrieveLogsFunction(logData.logGroupInfo, logData.parameters, request.nextForwardToken),
-            nextTokenNames: {
-                request: 'nextForwardToken',
-                response: 'nextForwardToken',
-            },
+            nextTokenNames: headOrTail
+                ? {
+                      request: 'nextForwardToken',
+                      response: 'nextForwardToken',
+                  }
+                : {
+                      request: 'nextBackwardToken',
+                      response: 'nextForwardToken',
+                  },
             request,
         })
         // TODO: Consider getPaginatedAwsCallIter? Would need a way to differentiate between head/tail...
--- a/src/cloudWatchLogs/registry/logDataRegistry.ts
+++ b/src/cloudWatchLogs/registry/logDataRegistry.ts
@@ -166,8 +166,9 @@ export class LogDataRegistry {
         const tokens: Pick<CloudWatchLogsData, 'next' | 'previous'> = {}
         // update if no token exists or if the token is updated in the correct direction.
         if (!logData.previous || headOrTail === 'head') {
-            tokens.previous = {
-                token: responseData.nextBackwardToken ?? '',
+            const token = responseData.nextBackwardToken
+            if (token) {
+                tokens.previous = { token }
             }
         }
         if (!logData.next || headOrTail === 'tail') {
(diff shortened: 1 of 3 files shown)


src/cloudWatchLogs/registry/logDataRegistry.ts
```diff
@@ -120,26 +121,61 @@ export class LogDataRegistry {
...
-                token: logEvents.nextBackwardToken ?? '',
+                token: responseData.nextBackwardToken ?? '',
             }
         }
         if (!logData.next || headOrTail === 'tail') {
             tokens.next = {
-                token: logEvents.nextForwardToken ?? '',
+                token: responseData.nextForwardToken ?? '',
             }
         }
```
JadenSimon, line 180: I think this should fix a bug:
```ts
if (!logData.next || headOrTail === 'tail') {
    const token =  responseData.nextForwardToken ?? request.nextForwardToken
    if (token) {
        tokens.next = { token }
    }
}
```

src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
```diff
@@ -50,6 +86,119 @@ describe('LogDataRegistry', async function () {
...
+            return async function (
+                logGroupInfo: CloudWatchLogsGroupInfo,
+                parameters: CloudWatchLogsParameters,
+                nextToken?: CloudWatchLogs.NextToken
+            ) {
+                return getSimulatedCwlResponse(nextToken, isPage1Empty)
+            }
+        }
+
+        const pageToken0 = undefined // Absence of token implies inital page
```
jpinkney-aws, line 116: Should we put this near the top of the describe block?

src/cloudWatchLogs/registry/logDataRegistry.ts
```diff
@@ -120,26 +121,60 @@ export class LogDataRegistry {
...
-                : logData.data.concat(logEvents.events ?? [])
+                ? (responseData.events ?? []).concat(logData.data)
+                : logData.data.concat(responseData.events ?? [])
 
         const tokens: Pick<CloudWatchLogsData, 'next' | 'previous'> = {}
         // update if no token exists or if the token is updated in the correct direction.
         if (!logData.previous || headOrTail === 'head') {
-            tokens.previous = {
-                token: logEvents.nextBackwardToken ?? '',
+            const token = responseData.nextBackwardToken
```
jpinkney-aws, line 169: I'm not that familiar with the cloudwatch logs code, but is there any reason we don't have: 
```suggestion
            const token = responseData.nextBackwardToken ?? request.nextBackwardToken
```
similar to what we have on line 175?

src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
```diff
@@ -50,6 +86,119 @@ describe('LogDataRegistry', async function () {
...
+    describe('updateLog', async function () {
+        it("properply paginates the results with 'head'", async () => {
+            await testUpdateLog('head')
+        })
+
+        it("properply paginates the results with 'tail'", async () => {
+            await testUpdateLog('tail')
+        })
+
+        function createCwlEvents(id: string, count: number): FilteredLogEvents {
```
jpinkney-aws, line 98: I don't think we have a standard on this, but I think we should pick where we define helper functions in describe blocks

src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
```diff
@@ -50,6 +86,119 @@ describe('LogDataRegistry', async function () {
         })
     })
 
+    describe('updateLog', async function () {
+        it("properply paginates the results with 'head'", async () => {
```
jpinkney-aws, line 90:
```suggestion
        it("properly paginates the results with 'head'", async () => {
```

src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
```diff
@@ -50,6 +86,119 @@ describe('LogDataRegistry', async function () {
...
+            await testUpdateLog('head')
+        })
+
+        it("properply paginates the results with 'tail'", async () => {
```
jpinkney-aws, line 94:
```suggestion
        it("properly paginates the results with 'tail'", async () => {
```

src/cloudWatchLogs/registry/logDataRegistry.ts
nkomonen-amazon: When we go "backward" we are going back in time (each page we visit is guaranteed to exist), so if a response has no backward token it implies we are at the beginning and are done.

This is different from going "forward" since a new page can eventually exist, which is why we store `request.nextForwardToken` if there is no new token from the response. When a response has no `nextForwardToken` it implies it is the tail and at some point it can have new data.

src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
nkomonen-amazon: lol

src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
nkomonen-amazon: My thought process is that these functions are currently only used by the one test. So it makes it easier if they are all grouped together. But in the event another test can use them we can change the scope of the functions.

JadenSimon: Function declaration order doesn't matter as far as scoping is concerned. All functions declared in this block are usable by all tests. I usually put functions at the top of a block below any variables. In this situation I'd probably make a new nested describe block to put all the pagination stuff in.

src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
```diff
@@ -50,6 +86,121 @@ describe('LogDataRegistry', async function () {
...
+            const firstUpdatedData = registry.getLogData(paginatedUri)
+            assert(firstUpdatedData)
+
+            const firstActual = {
+                events: firstUpdatedData.data,
+                nextForwardToken: firstUpdatedData.next?.token,
+                nextBackwardToken: firstUpdatedData.previous?.token,
+            }
+            const firstExpected = getSimulatedCwlResponse(pageToken0, page1IsEmpty)
+            assert.deepStrictEqual(firstActual, firstExpected)
```
JadenSimon, line 167: Seems like a pattern that could be de-duped. Maybe merged with `testUpdateLog` somehow?

src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
nkomonen-amazon: I've updated this test to its own describe block. Thanks!

JadenSimon (approved):

JadenSimon:
Might need to pull `feature/cwl` into your branch

jpinkney-aws (approved):

JadenSimon:
I think tests are failing due to a change with SAM CLI unrelated to this PR.

PR #3043, opened 2022-12-10
Gsweet/upload lambda skip prompt

gsweetwood:
## Problem
Invoking the context menu 'Upload Lambda...' on a `template.yaml` file prompts for directory when it isn't necessary.
## Solution
Auto select the parent directory of the `template.yaml` file as the target directory.

## License

### DIFF
--- a/src/lambda/commands/uploadLambda.ts
+++ b/src/lambda/commands/uploadLambda.ts
@@ -212,9 +212,16 @@ export class UploadLambdaWizard extends Wizard<UploadLambdaWizardState> {
         super({ initState: { lambda } })
         this.form.lambda.region.bindPrompter(() => createRegionPrompter().transform(region => region.id))
 
-        if (invokePath && fs.statSync(invokePath.fsPath).isDirectory()) {
+        if (
+            (invokePath && fs.statSync(invokePath.fsPath).isDirectory()) ||
+            invokePath?.fsPath.endsWith('template.yaml')
+        ) {
             this.form.uploadType.setDefault('directory')
-            this.form.targetUri.setDefault(invokePath)
+            if (invokePath?.fsPath.endsWith('template.yaml')) {
+                this.form.targetUri.setDefault(vscode.Uri.file(path.dirname(invokePath.fsPath)))
+            } else {
+                this.form.targetUri.setDefault(invokePath)
+            }
         } else {
             this.form.uploadType.bindPrompter(() => createUploadTypePrompter())
             this.form.targetUri.bindPrompter(({ uploadType }) => {
--- a/src/lambda/commands/uploadLambda.ts
+++ b/src/lambda/commands/uploadLambda.ts
@@ -212,12 +212,9 @@ export class UploadLambdaWizard extends Wizard<UploadLambdaWizardState> {
         super({ initState: { lambda } })
         this.form.lambda.region.bindPrompter(() => createRegionPrompter().transform(region => region.id))
 
-        if (
-            (invokePath && fs.statSync(invokePath.fsPath).isDirectory()) ||
-            invokePath?.fsPath.endsWith('template.yaml')
-        ) {
+        if (invokePath) {
             this.form.uploadType.setDefault('directory')
-            if (invokePath?.fsPath.endsWith('template.yaml')) {
+            if (fs.statSync(invokePath.fsPath).isFile()) {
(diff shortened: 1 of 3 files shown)


src/test/lambda/wizards/uploadLambdaWizard.test.ts
```diff
@@ -65,4 +67,27 @@ describe('UploadLambdaWizard', function () {
             tester.confirmedDeploy.assertShow(3)
         })
     })
+
+    describe('invoked from template', function () {
+        let tempDir: string
+        let tempDirUri: vscode.Uri
+        let invokePath: vscode.Uri
+        beforeEach(async function () {
+            tempDir = await makeTemporaryToolkitFolder()
```
jpinkney-aws, line 76: I think we should probably cleanup this folder once we are done with it. It also looks like a couple other tests in this file aren't cleaning up after themselves either 🤔

src/test/lambda/wizards/uploadLambdaWizard.test.ts
```diff
@@ -4,9 +4,11 @@
  */
 
 import * as vscode from 'vscode'
+import { writeFileSync } from 'fs-extra'
 import { createWizardTester, WizardTester } from '../../shared/wizards/wizardTestUtils'
 import { UploadLambdaWizard, UploadLambdaWizardState, LambdaFunction } from '../../../lambda/commands/uploadLambda'
 import { makeTemporaryToolkitFolder } from '../../../shared/filesystemUtilities'
+import path = require('path')
```
jpinkney-aws, line 11: I think we generally use `*` imports instead of require. E.g. `import * as path from 'path'`

jpinkney-aws (commented):
I think we probably need an enhanced git commit message and probably a changelog item as well

src/test/lambda/wizards/uploadLambdaWizard.test.ts
gsweetwood: Good call, it was auto imported and I missed that.

src/test/lambda/wizards/uploadLambdaWizard.test.ts
gsweetwood: Cleaned up test folders

src/test/lambda/wizards/uploadLambdaWizard.test.ts
```diff
@@ -65,4 +72,30 @@ describe('UploadLambdaWizard', function () {
             tester.confirmedDeploy.assertShow(3)
         })
     })
+
+    describe('invoked from template', function () {
+        let tempDir: string
+        let tempDirUri: vscode.Uri
+        let invokePath: vscode.Uri
+        beforeEach(async function () {
```
justinmk3, line 80: There's a bit of duplicate code in this beforeEach(). Please try to avoid that in the future.

src/lambda/commands/uploadLambda.ts
```diff
@@ -212,9 +212,16 @@ export class UploadLambdaWizard extends Wizard<UploadLambdaWizardState> {
         super({ initState: { lambda } })
         this.form.lambda.region.bindPrompter(() => createRegionPrompter().transform(region => region.id))
 
-        if (invokePath && fs.statSync(invokePath.fsPath).isDirectory()) {
+        if (
+            (invokePath && fs.statSync(invokePath.fsPath).isDirectory()) ||
+            invokePath?.fsPath.endsWith('template.yaml')
```
justinmk3, line 217: 1. This assumes that we won't support other filenames besides template.yaml in the future
2. the endsWith(...) check is duplicated below. Could be stored in a `const isFile = ...` above.
3. Why not check "file exists" instead of checking for "template.yaml" ?

src/test/lambda/wizards/uploadLambdaWizard.test.ts
gsweetwood: @justinmk3 These are in different `describe` blocks so they have different beforeEach() methods. Are you suggesting to place them in the same scope so they can both use that code? Maybe a 'invoked from explorer' `describe` block that contains both of these tests?

justinmk3: - there is no real reason for different describe() blocks in this file. all the it() cases could live in one shared describe()
- or, code duplication can be avoided by extracting to a common function

PR #3056, opened 2022-12-15
feat(filetypes): Prompt users to install YAML extension for Buildspec and Devfiles

jpinkney-aws:
## Problem
- Currently our yaml activation doesn't check for buildspec or devfiles so users may not be exposed to those features if they don't already have VSCode-YAML installed

## Solution
- Prompt for YAML extension activation on buildspec and devfiles. The only differences in this implementation is that we now watch on multiple different filenames and we also use the cfn/buildspec registry to determine if the prompt should activate

## License

### DIFF
--- a/src/shared/extensions/yamlActivation.ts
+++ b/src/shared/extensions/yamlActivation.ts
@@ -1,0 +1,207 @@
+/*!
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import * as vscode from 'vscode'
+import * as nls from 'vscode-nls'
+import * as path from 'path'
+import globals from '../extensionGlobals'
+
+import { VSCODE_EXTENSION_ID } from '../extensions'
+import { getIdeProperties, getIdeType, IDE } from '../extensionUtilities'
+import { getLogger } from '../logger'
+import { PromptSettings } from '../settings'
+import { awsFiletypeFileNames, AWSFiletypeNames } from '../awsFiletypes'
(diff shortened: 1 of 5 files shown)


src/shared/extensions/yamlActivation.ts
```diff
@@ -0,0 +1,207 @@
...
+ */
+async function promptInstallYamlPlugin(disposables: vscode.Disposable[], templateName?: AWSFiletypeNames) {
+    // immediately dispose other triggers so it doesn't flash again
+    for (const prompt of disposables) {
+        prompt.dispose()
+    }
+    const settings = PromptSettings.instance
+
+    const installBtn = localize('AWS.missingExtension.install', 'Install...')
+    const permanentlySuppress = localize('AWS.message.info.yaml.suppressPrompt', "Don't show again")
```
justinmk3, line 174: `localizedText.dontShow`

src/shared/extensions/yamlActivation.ts
```diff
@@ -0,0 +1,207 @@
...
+
+    const template = templateName ? templateName : 'YAML'
+    const response = await vscode.window.showInformationMessage(
+        localize(
+            'AWS.message.info.yaml.prompt',
+            'Install YAML extension for more {0} features in {1} templates',
```
justinmk3, line 180: This would read "devfile templates" and "buildspec templates". Maybe we should drop "templates" from the message or change it to "files"?

```suggestion
            'Install YAML extension for more {0} features in {1} files',
```

justinmk3 (approved):

JadenSimon (approved):

justinmk3 (approved):

PR #3058, opened 2022-12-15
fix(schemas): Improve template registry ownership model 

jpinkney-aws:
## Problem
- We currently have a concept of ownership in the template registry where a registry can declare that they "own" a uri. This is too fragile to work in practice and comes with too many edge cases (what happens if the owner isnt defined, etc, etc). This can lead to issues where the schema isn't getting correctly define for valid aws filetypes

## Solution
- Allow any template registry that defines a registry name to add a schema to a file uri. Then, when potential schemas are being evaluated use the first one that is defined. Since cfn/sam/buildspec are all distinct, we won't have any issues of them accidentally recognizing files that they shouldn't and it simplies the problem a lot

## License

### DIFF
--- a/src/shared/extensions/yaml.ts
+++ b/src/shared/extensions/yaml.ts
@@ -32,28 +32,67 @@ function evaluate(schema: vscode.Uri | (() => vscode.Uri)): vscode.Uri {
 }
 
 export interface YamlExtension {
-    assignSchema(path: vscode.Uri, schema: vscode.Uri | (() => vscode.Uri)): void
-    removeSchema(path: vscode.Uri): void
-    getSchema(path: vscode.Uri): vscode.Uri | undefined
+    assignSchema(uri: vscode.Uri, registry: string, schema: vscode.Uri | (() => vscode.Uri)): void
+    removeSchema(uri: vscode.Uri, registry: string): void
+    getSchema(uri: vscode.Uri): string | undefined
+}
+
+// helper for declaring support for a schema for both cloud 9 and vscode
+function registerSchema(
+    path: string,
+    registry: string,
+    schema: vscode.Uri | (() => vscode.Uri),
+    schemaMap: Map<string, Map<string, vscode.Uri>>
+) {
+    const schemas = schemaMap.get(path)
+    if (!schemas) {
+        schemaMap.set(path, new Map())
+    }
+    schemaMap.set(path, schemaMap.get(path)!.set(registry, applyScheme(AWS_SCHEME, evaluate(schema))))
+}
+
+// helper for removing support for a schema for both cloud 9 and vscode
+function unregisterSchema(path: string, registry: string, schemaMap: Map<string, Map<string, vscode.Uri>>) {
+    const schemas = schemaMap.get(path)
+    if (!schemas) {
+        return
+    }
+    const removed = schemas.delete(registry)
+    if (removed) {
+        schemaMap.set(path, schemas)
+    }
+}
+
+// Get the schema that should be associated with the given path, if there are multiple then choose the first
+function getSchemas(path: string, schemaMap: Map<string, Map<string, vscode.Uri>>): string | undefined {
+    const schema = schemaMap.get(path)
+    if (!schema) {
+        return undefined
+    }
+    const possibleSchemas = Array.from(schema.values()).filter(schema => schema !== undefined)
+    if (possibleSchemas.length > 0) {
+        return possibleSchemas[0].toString()
+    }
+    return undefined
 }
 
 export async function activateYamlExtension(): Promise<YamlExtension | undefined> {
-    const schemaMap = new Map<string, vscode.Uri>()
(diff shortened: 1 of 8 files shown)


src/shared/schemas.ts
```diff
@@ -109,7 +109,8 @@ export class SchemaService {
         const isSchemaChanged =
             this.owned.has(mapping.uri) &&
             mapping.owner !== undefined &&
-            this.owned.get(mapping.uri)?.schema !== mapping.schema
```
justinmk3, line 112: The PR description mentions something about not having an owner, but these conditions are already checking that.

There are subtleties here. So, a 1-line comment is important to explain the subtlety.

JadenSimon:
This bug is a sign of other problems with the schema service. I don't think having the idea of an "owner" is a bad solution but introducing this concept reveals a multi-tenancy problem. We have multiple, independent systems trying to use the same resource i.e. applying a schema to a file. 

One solution to this is to allow owners to assign schemas independently and perform some sort of conflict resolution like picking the first mapping. So for example `owners` should map one URI to potentially multiple mappings.

justinmk3:
> multi-tenancy problem.

"owner" is a risky concept for this reason; multi-tenancy is out of scope. Can the "registry" have composite keys, so that requesting an item requires both the filename and the requestor (schema-kind), then the answer returned depends on the requestor.

jpinkney-aws:
With the newest changes everything will be a lot more reliable now. Now any named registry can contribute a schema for a URI and the first one that isn't undefined is used and sent to vscode-yaml

src/shared/schemas.ts
```diff
@@ -33,7 +33,7 @@ export type SchemaType = 'yaml' | 'json'
 export interface SchemaMapping {
     uri: vscode.Uri
     type: SchemaType
-    owner?: string
+    registry?: string
```
JadenSimon, line 36: I see a lot of early returns due to this being nullable. This field is required according to https://github.com/aws/aws-toolkit-vscode/pull/3058/files#diff-6d14e8ae512380b5cec4abff41a0948289247dd6ebb53126cea22406bacbe7b0R94
```suggestion
    registry: string
```

src/shared/extensions/yaml.ts
```diff
@@ -32,28 +32,67 @@ function evaluate(schema: vscode.Uri | (() => vscode.Uri)): vscode.Uri {
 }
 
 export interface YamlExtension {
-    assignSchema(path: vscode.Uri, schema: vscode.Uri | (() => vscode.Uri)): void
-    removeSchema(path: vscode.Uri): void
-    getSchema(path: vscode.Uri): vscode.Uri | undefined
+    assignSchema(path: string, registry: string | undefined, schema: vscode.Uri | (() => vscode.Uri)): void
+    removeSchema(path: string, registry: string | undefined): void
+    getSchema(path: string): string | undefined
```
JadenSimon, line 37: Any particular reason for changing the type of 'path' to `string`?

src/shared/extensions/yaml.ts
```diff
@@ -32,28 +32,67 @@ function evaluate(schema: vscode.Uri | (() => vscode.Uri)): vscode.Uri {
...
+        return
+    }
+    const schemas = schemaMap.get(path)
+    if (!schemas) {
+        schemaMap.set(path, new Map())
+    }
+    schemaMap.set(path, schemaMap.get(path)!.set(registry, applyScheme(AWS_SCHEME, evaluate(schema))))
+}
+
+function deRegisterSchema(path: string, registry: string | undefined, schemaMap: Map<string, Map<string, vscode.Uri>>) {
```
justinmk3, line 56: "unregister" is more conventional/common. This is a nit, but also a general pattern: use common, simple names unless there is an explicit reason to use an uncommon name.

```suggestion
function unregisterSchema(path: string, registry: string | undefined, schemaMap: Map<string, Map<string, vscode.Uri>>) {
```

src/shared/extensions/yaml.ts
```diff
@@ -76,8 +115,10 @@ export async function activateYamlExtension(): Promise<YamlExtension | undefined
         }
     )
     return {
-        assignSchema: (path, schema) => schemaMap.set(path.toString(), applyScheme(AWS_SCHEME, evaluate(schema))),
-        removeSchema: path => schemaMap.delete(path.toString()),
-        getSchema: path => schemaMap.get(path.toString()),
+        assignSchema: (path, registry, schema) => {
+            registerSchema(path, registry, schema, schemaMap)
```
justinmk3, line 119: "assign" and "register" are approximately synonyms. This is a signal that the API needs more work. However, to unblock this PR, could for now add one-line comments explaining the difference.

justinmk3:
> Now any named registry can contribute a schema for a URI and the first one that isn't undefined is used and sent to vscode-yaml

This kind of practical, concise insight is a great thing to mention in the docstring on the related module or class :)

JadenSimon (approved):

PR #3060, opened 2022-12-15
codewhisperer: fix var assign error and move timestamp save

crossrac:
Fixes variable assignment error and moves warning message last seen timestamp so that timestamp is saved regardless of user action on popup.

## License

### DIFF
--- a/src/codewhisperer/activation.ts
+++ b/src/codewhisperer/activation.ts
@@ -244,14 +244,13 @@ export async function activate(context: ExtContext): Promise<void> {
                 context.extensionContext.globalState.get<boolean>(
                     CodeWhispererConstants.accessTokenMigrationDoNotShowAgainKey
                 ) || false
-            const notificationLastShown =
-                context.extensionContext.globalState.get<Date>(
+            const notificationLastShown: number =
+                context.extensionContext.globalState.get<number | undefined>(
                     CodeWhispererConstants.accessTokenMigrationDoNotShowAgainLastShown
-                ) || new Date()
+                ) || 1
 
             //Add 7 days to notificationLastShown to determine whether warn message should show
-            notificationLastShown.setDate(notificationLastShown.getDate() + 7)
-            if (doNotShowAgain || notificationLastShown <= t) {
+            if (doNotShowAgain || notificationLastShown + 1000 * 60 * 60 * 24 * 7 >= Date.now()) {
                 return
             } else if (t <= CodeWhispererConstants.accessTokenCutOffDate) {
                 vscode.window
@@ -277,7 +276,7 @@ export async function activate(context: ExtContext): Promise<void> {
                     })
                 context.extensionContext.globalState.update(
                     CodeWhispererConstants.accessTokenMigrationDoNotShowAgainLastShown,
-                    t
+                    Date.now()
                 )
             } else {
                 await globals.context.globalState.update(CodeWhispererConstants.accessToken, undefined)
--- a/src/codewhisperer/activation.ts
+++ b/src/codewhisperer/activation.ts
@@ -247,14 +247,13 @@ export async function activate(context: ExtContext): Promise<void> {
             const notificationLastShown =
                 context.extensionContext.globalState.get<Date>(
                     CodeWhispererConstants.accessTokenMigrationDoNotShowAgainLastShown
-                ) || t
+                ) || new Date()
 
             //Add 7 days to notificationLastShown to determine whether warn message should show
             notificationLastShown.setDate(notificationLastShown.getDate() + 7)
             if (doNotShowAgain || notificationLastShown <= t) {
                 return
-            }
-            if (t <= CodeWhispererConstants.accessTokenCutOffDate) {
+            } else if (t <= CodeWhispererConstants.accessTokenCutOffDate) {
                 vscode.window
                     .showWarningMessage(
                         CodeWhispererConstants.accessTokenMigrationWarningMessage,
@@ -274,12 +273,12 @@ export async function activate(context: ExtContext): Promise<void> {
                                 CodeWhispererConstants.accessTokenMigrationDoNotShowAgainKey,
                                 true
                             )
-                            await context.extensionContext.globalState.update(
-                                CodeWhispererConstants.accessTokenMigrationDoNotShowAgainLastShown,
-                                t
-                            )
                         }
                     })
+                context.extensionContext.globalState.update(
+                    CodeWhispererConstants.accessTokenMigrationDoNotShowAgainLastShown,
+                    t
+                )
             } else {
                 await globals.context.globalState.update(CodeWhispererConstants.accessToken, undefined)
                 await vscode.commands.executeCommand('aws.codeWhisperer.refreshRootNode')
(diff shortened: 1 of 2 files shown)


src/codewhisperer/activation.ts
```diff
@@ -247,14 +247,13 @@ export async function activate(context: ExtContext): Promise<void> {
             const notificationLastShown =
                 context.extensionContext.globalState.get<Date>(
                     CodeWhispererConstants.accessTokenMigrationDoNotShowAgainLastShown
-                ) || t
+                ) || new Date()
```
JadenSimon, line 250: `Date` is serialized into a string by default. I would use the epoch time in milliseconds instead:
```ts
const currentTime = Date.now()
const lastShownTime = context.extensionContext.globalState.get<number>(
    CodeWhispererConstants.accessTokenMigrationDoNotShowAgainLastShown,
    currentTime
)
``` 
Then all math should be changed to work in milliseconds.

justinmk3 (approved):

src/codewhisperer/activation.ts
justinmk3: > Then all math should be changed to work in milliseconds.

Since this is a hotfix should we wait on that as a followup?

JadenSimon: > Since this is a hotfix should we wait on that as a followup? (Great callout though)

No this is the root cause of the problem, not variable assignment order. I updated my comment to make it more clear.

crossrac: Thanks for this callout, will update

JadenSimon: Change the value of `accessTokenMigrationDoNotShowAgainLastShown` as well. Otherwise you'll have to handle state from 1.59.0

leigaol: const current_time = new Date()
const next_week =new Date(current_time.getTime() + 7 * 24 * 60 * 60 * 1000)

Then you can just compare the Date object

leigaol (approved):

src/codewhisperer/activation.ts
leigaol: Agree with what @JadenSimon said. Epoch time is better.

crossrac: Updated in latest commit

src/codewhisperer/activation.ts
```diff
@@ -244,17 +244,15 @@ export async function activate(context: ExtContext): Promise<void> {
                 context.extensionContext.globalState.get<boolean>(
                     CodeWhispererConstants.accessTokenMigrationDoNotShowAgainKey
                 ) || false
-            const notificationLastShown =
-                context.extensionContext.globalState.get<Date>(
+            const notificationLastShown: number =
```
justinmk3, line 247: I'd leave the type as `any`. Because existing state will be string, the logic after this needs to check `typeof(notificationLastShown)` and if it's a string, then default to `1` or something.

src/codewhisperer/activation.ts
```diff
@@ -244,17 +244,15 @@ export async function activate(context: ExtContext): Promise<void> {
...
+                context.extensionContext.globalState.get<number | undefined>(
                     CodeWhispererConstants.accessTokenMigrationDoNotShowAgainLastShown
-                ) || t
+                ) || 1
 
             //Add 7 days to notificationLastShown to determine whether warn message should show
-            notificationLastShown.setDate(notificationLastShown.getDate() + 7)
-            if (doNotShowAgain || notificationLastShown <= t) {
+            if (doNotShowAgain || notificationLastShown + 1000 * 60 * 60 * 24 * 7 >= Date.now()) {
```
justinmk3, line 253: unless prettier forced this, I'd use parens to be extra clear about order of operations.
```suggestion
            if (doNotShowAgain || notificationLastShown + (1000 * 60 * 60 * 24 * 7) >= Date.now()) {
```

src/codewhisperer/activation.ts
crossrac: We updated the key so it will save under that new key and will either be `undefined` or a number

src/codewhisperer/models/constants.ts
```diff
@@ -241,6 +241,6 @@ export const accessTokenMigrationDoNotShowAgain = `Don\'t Show Again`
 export const accessTokenMigrationDoNotShowAgainKey = 'CODEWHISPERER_ACCESS_TOKEN_MIGRATION_DO_NOT_SHOW_AGAIN'
 
 export const accessTokenMigrationDoNotShowAgainLastShown =
-    'CODEWHISPERER_ACCESS_TOKEN_MIGRATION_DO_NOT_SHOW_AGAIN_LAST_SHOWN'
+    'CODEWHISPERER_ACCESS_TOKEN_MIGRATION_DO_NOT_SHOW_AGAIN_LAST_SHOWN_TIME'
```
justinmk3, line 244: Future note: please avoid verbose names like this. This is 2x longer than any name should be.

crossrac: noted will be sure to keep shorter in the future

src/codewhisperer/activation.ts
JadenSimon: > globalState.get<number> makes no guarantees about the value at runtime.

Yup. The correct way to do this is actually `globalState.get<unknown>`. I usually don't mention `unknown` though because it requires even more explanation on how to convert 'unknown' types without using `as`

justinmk3: The type is still incorrect though. If you remove the explicit type annotation, typescript inference will infer it as `number|undefined` since that is what you gave in the `get<>()` call.

It looks like `notificationLastShown +` addition operation is being done on a possibly-undefined value below.

```suggestion
            const notificationLastShown =
```

src/codewhisperer/activation.ts
```diff
@@ -244,17 +244,15 @@ export async function activate(context: ExtContext): Promise<void> {
...
+            const notificationLastShown: number =
+                context.extensionContext.globalState.get<number | undefined>(
                     CodeWhispererConstants.accessTokenMigrationDoNotShowAgainLastShown
-                ) || t
+                ) || 1
 
             //Add 7 days to notificationLastShown to determine whether warn message should show
-            notificationLastShown.setDate(notificationLastShown.getDate() + 7)
-            if (doNotShowAgain || notificationLastShown <= t) {
+            if (doNotShowAgain || notificationLastShown + (1000 * 60 * 60 * 24 * 7) >= Date.now()) {
```
justinmk3, line 253: what happens if `notificationLastShown` is undefined here?

src/codewhisperer/activation.ts
justinmk3: oops I see the `|| 1` ensures `number`.

justinmk3 (approved):

src/codewhisperer/activation.ts
JadenSimon: It's because the code is still using `||` which is bad practice IMO. The correct way is:
```ts
               context.extensionContext.globalState.get<number>(
                    CodeWhispererConstants.accessTokenMigrationDoNotShowAgainLastShown,
                    1
                )
```
Explicit type annotations on variable declarations is not the same as a type assertion unlike `as` or type parameters so having `const notificationLastShown: number` is fine albeit verbose.

JadenSimon (approved):

Issue #3055, opened 2022-12-15
Editing files without extension on S3 changes the Metadata of the object


esaday:
## Problem
When editing files without extension, Toolkit mutates the Metadata of the edited object. For instance, I have a file named `some_config` on S3 and its current Metdata for `Content-Type` is `text/plain`. 

When you change the contents of the file and save on VS Code (trigger upload to bucket), it changes the `Content-Type` of the object to `octet-stream`. This made me think the Toolkit is not preserving the current Metadata for the object and fallback to default values.

## Steps to reproduce the issue

1. Go to the `S3` tab in `EXPLORER` in Toolkit.
2. Double Click on any file containing text without the file extension.
3. Make changes and save
4. Check Metadata values for the object through cli or AWS console. 
5. See the `Content-Type` is changed to `octet-stream`

## Expected behavior
Toolkit should preserve the current/available metada of the object being edited.

## System details (run the `AWS: About Toolkit` command)
- OS: Darwin x64 21.6.0
- Visual Studio Code extension host:  1.73.1
- AWS Toolkit:  1.58.0
- node: 16.14.2
- electron: 19.0.17

JadenSimon:
Thanks for the report! This likely affects more than just the file editor feature.

> This made me think the Toolkit is not preserving the current Metadata for the object and fallback to default values.

Yup I just looked at the code and this is exactly what's happening.

JadenSimon:
Should be fixed in 1.61.0. Let us know if you have any more issues!

PR #3062, opened 2022-12-16
fix(s3): handle content types more gracefully

JadenSimon:
## Problem
* Uploading files from the file editor overwrites the existing content type
    * #3055
* Non-text files without a file extension throw an error

## Solution
* Shift the `ContentType` parameter up so callers have to be explicit
* Treat files with no mime type as binary files

I also refactored `fileViewerManager.ts` a bit so it could be tested with significantly less mocks

## License

### DIFF
--- a/src/s3/fileViewerManager.ts
+++ b/src/s3/fileViewerManager.ts
@@ -215,7 +215,7 @@ export class S3FileViewerManager {
         const isTextDocument = contentType && mime.charset(contentType) == 'UTF-8'
 
         const uri = this.fileToUri(file, TabMode.Read)
-        if (await this.tryFocusTab(uri, uri.with({ scheme: this.schemes.read }))) {
+        if (await this.tryFocusTab(uri, uri.with({ scheme: this.schemes.edit }))) {
             return
         }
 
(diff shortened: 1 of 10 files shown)


src/s3/fileViewerManager.ts
```diff
@@ -88,12 +90,13 @@ export class S3FileProvider implements FileProvider {
...
             span.record({ component: 'viewer' })
 
+            const mimeType = mime.contentType(path.extname(this._file.name)) || undefined
             const result = await this.client
                 .uploadFile({
                     content,
                     key: this._file.key,
                     bucketName: this._file.bucket.name,
-                    contentType: mime.contentType(path.extname(this._file.name)) || undefined,
+                    contentType: this._file.ContentType ?? mimeType,
```
justinmk3, line 99: out of scope for this PR, but as a future todo: can/should `downloadFile` get the current mimetype stored in S3 and re-use that in write()?

src/s3/fileViewerManager.ts
```diff
@@ -54,8 +55,9 @@ export class S3FileProvider implements FileProvider {
         const stats = await this.client.headObject({ bucketName: bucket.name, key })
 
         this.updateETag(stats.ETag)
-        this._file.sizeBytes = stats.ContentLength
+        this._file.sizeBytes = stats.ContentLength ?? this._file.sizeBytes
         this._file.lastModified = stats.LastModified
+        this._file.ContentType = stats.ContentType
```
justinmk3, line 60: now we are using the contenttype given by S3, nice

justinmk3 (approved):
> also refactored fileViewerManager.ts a bit so it could be tested with significantly less mocks

👎

src/s3/fileViewerManager.ts
JadenSimon: Yeah `downloadObject` returns it too but it's a little harder to get when using the SDKs streams.

justinmk3 (approved):

PR #3063, opened 2022-12-16
feat(sam): collect sam cli errors

nkomonen-amazon:
## Problem

The sam cli may fail when executing operations
internally. An example is trying to debug a
lambda using python, but a specific python
version does not exist.

## Solution

This commit now has the sam cli invoker collect
accepted errors (based off certain prefixes)
and throws them instead of indicating
'no message available'

You can test this by:
- Running the 'Extenstion' debugger
- Creating a python3.7 SAM lambda application
- Run and debug the lambda function
- This will induce an error telling you it could not find the python3.7 runtime (assuming you don't have it locally)

## License

### DIFF
--- a/src/shared/sam/cli/samCliInvokerUtils.ts
+++ b/src/shared/sam/cli/samCliInvokerUtils.ts
@@ -55,6 +55,42 @@ export function logAndThrowIfUnexpectedExitCode(processResult: ChildProcessResul
             message = processResult.error.message
         }
     }
+    const usefulErrors = collectAcceptedErrorMessages(processResult.stderr).join('\n')
+    throw makeUnexpectedExitCodeError(message ?? usefulErrors)
+}
 
-    throw makeUnexpectedExitCodeError(message ?? 'no message available')
+/**
+ * Collect known errors messages from a sam error message
+ * that may have multiple errors in one message.
+ * @param errors A string that can have multiple error messages
+ */
+export function collectAcceptedErrorMessages(errorMessage: string): string[] {
+    const errors = errorMessage.split('\n')
+    const shouldCollectFuncs = [_startsWithEscapeSequence, _startsWithError]
+    return errors.filter(error => {
+        return shouldCollectFuncs.some(shouldCollect => {
+            return shouldCollect(error)
+        })
+    })
+}
+
+/**
+ * Returns true if text starts with an escape
+ * sequence with one of the accepted sequences.
+ */
+function _startsWithEscapeSequence(text: string): boolean {
+    const escapeInDecimal = 27
+    if (text.charCodeAt(0) !== escapeInDecimal) {
+        return false
+    }
+
+    const remainingText = text.substring(1)
+    const acceptedSequences = ['[33m']
+    return acceptedSequences.some(code => {
+        return remainingText.startsWith(code)
+    })
+}
+
+function _startsWithError(text: string): boolean {
+    return text.startsWith('Error:')
 }
--- a/src/shared/sam/cli/samCliInvokerUtils.ts
+++ b/src/shared/sam/cli/samCliInvokerUtils.ts
@@ -66,7 +66,7 @@ export function logAndThrowIfUnexpectedExitCode(processResult: ChildProcessResul
  */
 export function collectAcceptedErrorMessages(errorMessage: string): string[] {
     const errors = errorMessage.split('\n')
-    const shouldCollectFuncs = [_startsWithEscapeSequence, _startsWithError]
+    const shouldCollectFuncs = [startsWithEscapeSequence, startsWithError]
     return errors.filter(error => {
         return shouldCollectFuncs.some(shouldCollect => {
             return shouldCollect(error)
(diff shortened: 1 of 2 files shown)


src/shared/sam/cli/samCliInvokerUtils.ts
```diff
@@ -55,6 +57,25 @@ export function logAndThrowIfUnexpectedExitCode(processResult: ChildProcessResul
             message = processResult.error.message
         }
     }
+    const shortenedErrorsText = filterSamErrorMessage(processResult.stderr).join('&&')
```
justinmk3, line 60: Rather than "filter", it's more like "collect all lines that look like errors". 

Did you decide against looking for `Error:`?

src/shared/sam/cli/samCliErrors.ts
```diff
@@ -0,0 +1,81 @@
...
+            return shortenedMessage
+        }
+    }
+}
+
+/**
+ * Shortens the missing python runtime error message from SAM Cli
+ */
+export function missingPythonRuntimeShortener(message: string): string | undefined {
+    const pattern = /Do you have python for runtime: (python\S+) on your PATH\??/
```
justinmk3, line 63: If sam cli has localized messages, this won't work. That's why something more generic like looking for `Error:` would be preferrable.

nkomonen-amazon: I'm just wondering how we are supposed to tackle task #2 (providing a more user friendly message) if we don't search for this specific message in IDE-8580.

Specifically for this case, the full error message is `Error: PythonPipBuilder:Validation - Binary validation failed for python, searched for python in following locations  : ['/usr/bin/python3'] which did not satisfy constraints for runtime: python3.7. Do you have python for runtime: python3.7 on your PATH?`

justinmk3: > wondering how we are supposed to tackle task 2 of the SIM (providing a more user friendly message) if we don't search for this specific message

Good question! Let's change the SIM :)

src/test/shared/sam/cli/samCliInvokerUtils.test.ts
```diff
@@ -42,3 +43,32 @@ describe('logAndThrowIfUnexpectedExitCode', async function () {
         await assertLogContainsBadExitInformation(getTestLogger(), childProcessResult, 456)
     })
 })
+
+describe('filterSamErrorMessage()', async () => {
+    it('Python runtime message shortened', async () => {
+        const original = `Bla Bla Bla. Do you have python for runtime: python3.7 on your PATH?`
+        const result = filterSamErrorMessage(original)
+        assert(result.length == 1)
```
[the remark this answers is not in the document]
nkomonen-amazon: Thanks for the info, I'll make sure to start doing this way.

src/shared/sam/cli/samCliInvokerUtils.ts
```diff
@@ -55,6 +55,42 @@ export function logAndThrowIfUnexpectedExitCode(processResult: ChildProcessResul
...
+            return shouldCollect(error)
+        })
+    })
+}
+
+/**
+ * Returns true if text starts with an escape
+ * sequence with one of the accepted sequences.
+ */
+function _startsWithEscapeSequence(text: string): boolean {
```
JadenSimon, line 81: nit:
```suggestion
function startsWithEscapeSequence(text: string): boolean {
```

Prefixing things with an underscore is a somewhat outdated style that should be limited to private fields in classes. And even then I'd recommended not using it.

src/shared/sam/cli/samCliInvokerUtils.ts
```diff
@@ -55,6 +55,42 @@ export function logAndThrowIfUnexpectedExitCode(processResult: ChildProcessResul
...
+    return errors.filter(error => {
+        return shouldCollectFuncs.some(shouldCollect => {
+            return shouldCollect(error)
+        })
+    })
+}
+
+/**
+ * Returns true if text starts with an escape
+ * sequence with one of the accepted sequences.
```
JadenSimon, line 79: nit: We mention accepted sequences but don't say what they are or how they are derived. Perhaps move `acceptedSequences` outside of the function body.

src/shared/sam/cli/samCliInvokerUtils.ts
```diff
@@ -55,6 +55,42 @@ export function logAndThrowIfUnexpectedExitCode(processResult: ChildProcessResul
...
+ */
+function _startsWithEscapeSequence(text: string): boolean {
+    const escapeInDecimal = 27
+    if (text.charCodeAt(0) !== escapeInDecimal) {
+        return false
+    }
+
+    const remainingText = text.substring(1)
+    const acceptedSequences = ['[33m']
```
JadenSimon, line 88: ANSI escape codes are somewhat obscure so a comment or variable is helpful:
```suggestion
    // ANSI Escape Sequences
    // ref: https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797#8-16-colors
    const acceptedSequences = [
        '[33m' // Yellow Foreground
     ]
```
Comments are fine here IMO but you can always extract into a variable for 'self-documenting' code:
```ts
const yellowForeground =  '[33m' 
const acceptedSequences = [yellowForeground]
```

JadenSimon (commented):
Looks good! Just a few minor comments.

jpinkney-aws (approved):
LGTM other than Jadens comments

src/shared/sam/cli/samCliInvokerUtils.ts
nkomonen-amazon: I like the last example, updated to this.

justinmk3 (approved):

justinmk3:
@nkomonen-amazon please squash-merge PRs unless the commits are intentionally organized. `chore: Fix PR comments` doesn't look like a useful commit :)

PR #3066, opened 2022-12-20
upload lambda: disable for image based lambdas

gsweetwood:
## Problem
 The toolkit only supports 'Upload Lambda' when lambdas have a package type of 'zip'. Attempting to upload to an image based lambda results in an error.

## Solution
 Continue to show but disable the command on lambda nodes that have a package type of 'image'. The lambdas listed in the UploadLambdaWizard with this package type will be invalid (the user will not be able to select it) and a description has been added.

## Screenshot

upload disabled

invalid seletion upload lambda

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -1460,22 +1460,22 @@
                 },
                 {
                     "command": "aws.invokeLambda",
-                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable|awsCloudFormationFunctionNode)$/",
+                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable|awsCloudFormationFunctionNode|awsRegionFunctionNodeUploadable|awsRegionFunctionNodeDownloadableUploadable)$/",
                     "group": "0@1"
                 },
                 {
                     "command": "aws.downloadLambda",
-                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable)$/",
+                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable|awsRegionFunctionNodeUploadable|awsRegionFunctionNodeDownloadableUploadable)$/",
                     "group": "0@2"
                 },
                 {
                     "command": "aws.uploadLambda",
-                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable)$/",
+                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable|awsRegionFunctionNodeUploadable|awsRegionFunctionNodeDownloadableUploadable)$/",
                     "group": "1@1"
                 },
                 {
                     "command": "aws.deleteLambda",
-                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable)$/",
+                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable|awsRegionFunctionNodeUploadable|awsRegionFunctionNodeDownloadableUploadable)$/",
                     "group": "4@1"
                 },
                 {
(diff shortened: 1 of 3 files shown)


gsweetwood:
When adding tests I realized we were setting the context value of the `lambdaFunctionNodes` in the `lambdaNodes` file. This seems weird and doesn't follow the pattern of other resources. Going to move that logic and then testing should be easier.

src/test/lambda/explorer/lambdaFunctionNode.test.ts
```diff
@@ -55,4 +62,29 @@ describe('LambdaFunctionNode', function () {
...
+
+        testNode = new LambdaFunctionNode(parentNode, 'someregioncode', onlyDownloadable)
+        assert.strictEqual(testNode.contextValue, CONTEXT_VALUE_LAMBDA_FUNCTION_IMPORTABLE)
+
+        testNode = new LambdaFunctionNode(parentNode, 'someregioncode', onlyUploadable)
+        assert.strictEqual(testNode.contextValue, CONTEXT_VALUE_LAMBDA_FUNCTION_UPLOADABLE)
+
+        testNode = new LambdaFunctionNode(parentNode, 'someregioncode', neither)
+        assert.strictEqual(testNode.contextValue, CONTEXT_VALUE_LAMBDA_FUNCTION)
+    })
```
JadenSimon, line 89: This test could be split into 4 though it wouldn't improve readability that much. I'd update the test name to be more specific though. Something like "sets appropriate contextValue for importing and uploading". Being specific helps to set expectations for the body of test.

JadenSimon (approved):

package.json
```diff
@@ -3464,7 +3465,7 @@
             "aws-stepfunctions-preview": {
                 "description": "AWS Contributed Icon",
                 "default": {
-                    "fontPath": "./resources/fonts/aws-toolkit-icons.woff",
+                    "fontPath": "./resources\\fonts\\aws-toolkit-icons.woff",
```
justinmk3, line 3468: why was this changed? if `\` slashes are being used somewhere, that should be fixed.

src/lambda/explorer/lambdaFunctionNode.ts
```diff
@@ -9,6 +9,12 @@ import { getIcon } from '../../shared/icons'
 
 import { AWSResourceNode } from '../../shared/treeview/nodes/awsResourceNode'
 import { AWSTreeNodeBase } from '../../shared/treeview/nodes/awsTreeNodeBase'
+import { samLambdaImportableRuntimes } from '../models/samLambdaRuntime'
+
+export const CONTEXT_VALUE_LAMBDA_FUNCTION = 'awsRegionFunctionNode'
+export const CONTEXT_VALUE_LAMBDA_FUNCTION_IMPORTABLE = 'awsRegionFunctionNodeDownloadable'
+export const CONTEXT_VALUE_LAMBDA_FUNCTION_UPLOADABLE = 'awsRegionFunctionNodeUploadable'
+export const CONTEXT_VALUE_LAMBDA_FUNCTION_IMPORTABLE_UPLOADABLE = 'awsRegionFunctionNodeDownloadableUploadable'
```
justinmk3, line 17: we've discussed in other PRs that we want to avoid uppercase names going forward. Since this is being changed in this PR, the names should be updated.

also in general, names this long should never exist. it is not necessary to have such long names. 

```suggestion
export const contextLambdaImportableUploadable = 'awsRegionFunctionNodeDownloadableUploadable'
```

package.json
JadenSimon: Oh interesting, was this a manual change? Might need to update `scripts/build/generateIcons.ts` to only use posix paths

gsweetwood: No, this was not manual. They all were changed and I had to revert them back. I missed this one. Fixed in next commit.

justinmk3:
- missing changelog (fixed)
- updated commit message to be shorter yet clearer

justinmk3:
Tested locally, doesn't seem to work. 


```
{
    "FunctionName": "jan2023-HelloWorldFunction-…",
    "PackageType": "Image",
    ...
}
```


image


However this part works:

image

package.json
```diff
@@ -1464,22 +1464,22 @@
...
                 {
                     "command": "aws.invokeLambda",
-                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable|awsCloudFormationFunctionNode)$/",
+                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable|awsCloudFormationFunctionNode|awsRegionFunctionNodeUploadable|awsRegionFunctionNodeDownloadableUploadable)$/",
                     "group": "0@1"
                 },
                 {
                     "command": "aws.downloadLambda",
-                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable)$/",
+                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable|awsRegionFunctionNodeUploadable|awsRegionFunctionNodeDownloadableUploadable)$/",
```
justinmk3, line 1472: instead of adding all these items manually, can we use a regex like:


```suggestion
                    "when": "view == aws.explorer && viewItem =~ /^aws\\w*\\.(download|upload)$/",
```

Then to "turn on" download/upload functionality, when constructing the node we just append `.download` and/or `.upload`.

JadenSimon: I think `/^aws\\w*\\.download/` is what we want right?

PR #3083, opened 2023-01-03
fix(tests): Increase git buffer size so cfn/sam schema can be downloaded

jpinkney-aws:
## Problem
Currently the integration tests in https://github.com/aws/aws-toolkit-vscode/pull/3042 are failing because the SAM schema fails to download with ```stdout maxBuffer length exceeded```

## Solution
Increase git buffer size

## License

### DIFF
--- a/src/shared/extensions/git.ts
+++ b/src/shared/extensions/git.ts
@@ -369,7 +369,7 @@ export class GitExtension {
 
                         return execFileAsync(api.git.path, ['cat-file', type, hash], {
                             cwd: tmpDir,
-                            maxBuffer: 1024 * 1024 * 5,
+                            maxBuffer: 1024 * 1024 * 6,
                         }).then(({ stdout }) => stdout)
                     },
                 }))


nkomonen-amazon (approved):

src/shared/extensions/git.ts
```diff
@@ -369,7 +369,7 @@ export class GitExtension {
 
                         return execFileAsync(api.git.path, ['cat-file', type, hash], {
                             cwd: tmpDir,
-                            maxBuffer: 1024 * 1024 * 5,
+                            maxBuffer: 1024 * 1024 * 6,
```
JadenSimon, line 372: Let's just set it to 50MB:
```suggestion
                            maxBuffer: 1024 * 1024 * 50,
```
[This thread](https://github.com/nodejs/node/issues/9829) mentions dynamic sizing a potential solution. Not sure if Node dynamically sizes the buffer already or if the buffer is allocated based off `maxBuffer`. If the former then `Infinity` could work although it would make it easier to run out of memory on the heap, crashing V8.

JadenSimon (approved):

PR #3090, opened 2023-01-06
feat(codewhisperer): Add Client component latency telemetry

leigaol:
## Problem

Adding a new telemetry : Client component latency for codewhisperer.

This telemetry's definition in https://github.com/aws/aws-toolkit-common/commit/618a9fb1585d0e7131901a9c4f3ac60ff4a3111f. 

A client component latency metric is emitted only once for each complete GenerationRecommendation pagination session if the suggestion is shown to the user.

```

{
            "name": "codewhispererFirstCompletionLatency",
            "type": "double",
            "description": "The time it takes for the response to be received after the plugin makes a first GenerateCompletions API call."
        },
        {
            "name": "codewhispererEndToEndLatency",
            "type": "double",
            "description": "The time it takes for the first completion to be shown in the IDE after the user performs the CW trigger action."
        },
        {
            "name": "codewhispererAllCompletionsLatency",
            "type": "double",
            "description": "The time it takes for the last GenerateCompletions response to be received after plugin makes a first call to GenerateCompletions API."
        },
        {
            "name": "codewhispererPostprocessingLatency",
            "type": "double",
            "description": "The time it takes for the first completions to be displayed in the IDE after the plugin receives the initial Completions object."
        },
        {
            "name": "codewhispererPreprocessingLatency",
            "type": "double",
            "description": "The time it takes for the plugin to make the first GenerateCompletions API call after the user performs the CW trigger action."
        },
        {
            "name": "codewhispererCredentialFetchingLatency",
            "type": "double",
            "description": "The time it takes to get the Sono/SSO credential for the invocation."
        },

```

This change is not customer facing, no change log is needed. 

## Solution

Record timestamp and compare the timestamp to build the Client component latency metric and report it to service.

## License

### DIFF
--- a/src/codewhisperer/util/telemetryHelper.ts
+++ b/src/codewhisperer/util/telemetryHelper.ts
@@ -160,10 +160,10 @@ export class TelemetryHelper {
         this.sessionId = ''
     }
 
+    /** This method is assumed to be invoked first at the start of execution **/
     public setInvokeSuggestionStartTime() {
-        if (this.invokeSuggestionStartTime === 0) {
-            this.invokeSuggestionStartTime = performance.now()
-        }
+         this.resetClientComponentLatencyTime()
+         this.invokeSuggestionStartTime = performance.now()
     }
 
     public setFetchCredentialStartTime() {
--- a/src/codewhisperer/util/telemetryHelper.ts
+++ b/src/codewhisperer/util/telemetryHelper.ts
@@ -35,7 +35,7 @@ export class TelemetryHelper {
 
     public startUrl: string | undefined
 
-    // timestamps to compute client component latency
+    // variables for client component latency
     private invokeSuggestionStartTime = 0
     private fetchCredentialStartTime = 0
     private sdkApiCallStartTime = 0
@@ -43,6 +43,7 @@ export class TelemetryHelper {
     private firstSuggestionShowTime = 0
     private allPaginationEndTime = 0
     private firstResponseRequestId = ''
+    private sessionId = ''
 
     constructor() {
         this.triggerType = 'OnDemand'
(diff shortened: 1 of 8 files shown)


src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -35,12 +35,23 @@ export class TelemetryHelper {
 
     public startUrl: string | undefined
 
+    // variables for client component latency
+    private invokeSuggestionStartTime = 0
+    private fetchCredentialStartTime = 0
+    private sdkApiCallStartTime = 0
```
andrewyuq, line 41: why there's still sdk latency measured? We said we are not going to measure this in the doc.

leigaol: Hi Andrew, sdk api call start time is only used to compute the other relevant latencies, it is not used to compute SDK latency.

andrewyuq (approved):

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -137,4 +148,84 @@ export class TelemetryHelper {
...
+        this.allPaginationEndTime = 0
+        this.firstResponseRequestId = ''
+        this.sessionId = ''
+    }
+
+    public setInvokeSuggestionStartTime() {
+        if (this.invokeSuggestionStartTime === 0) {
+            this.invokeSuggestionStartTime = performance.now()
+        }
+    }
```
JadenSimon, line 167: Setting values once can cause data quality issues if the state isn't reset after a bad execution. For example, what if `getPaginatedRecommendation` failed for whatever reason? The next successful execution would use an inaccurate timestamp.

If we don't need to handle concurrency then it's safe to always reset state at the start of the execution:
```suggestion
    /** This method is assumed to be invoked first at the start of execution **/
    public setInvokeSuggestionStartTime() {
         this.resetClientComponentLatencyTime()
         this.invokeSuggestionStartTime = performance.now()
    }
```

zixlin7 (approved):

JadenSimon (approved):

mitroviv (approved):

PR #3093, opened 2023-01-10
feat(codewhisperer): auto trigger improvements

zixlin7:
## Problem
current auto-trigger heuristics is not very responsive
## Solution
replacing the character number based triggering rule with an idle time based triggering rule to make the triggering more aggressive

## License

### DIFF
--- a/src/codewhisperer/service/keyStrokeHandler.ts
+++ b/src/codewhisperer/service/keyStrokeHandler.ts
@@ -29,17 +29,60 @@ export class KeyStrokeHandler {
     /**
      * Key stroke count for automated trigger
      */
-    public keyStrokeCount: number
+
+    private idleTriggerTimer?: NodeJS.Timer
 
     constructor() {
         this.specialChar = ''
-        this.keyStrokeCount = 0
     }
 
     static #instance: KeyStrokeHandler
 
     public static get instance() {
         return (this.#instance ??= new this())
+    }
+
+    public startIdleTimeTriggerTimer(
+        event: vscode.TextDocumentChangeEvent,
+        editor: vscode.TextEditor,
+        client: DefaultCodeWhispererClient,
+        config: ConfigurationEntry
+    ) {
+        if (this.idleTriggerTimer) {
+            clearInterval(this.idleTriggerTimer)
+            this.idleTriggerTimer = undefined
+        }
+        if (!this.shouldTriggerIdleTime()) {
+            return
+        }
+        this.idleTriggerTimer = setInterval(() => {
+            const duration = (performance.now() - RecommendationHandler.instance.lastInvocationTime) / 1000
+            if (duration < CodeWhispererConstants.invocationTimeIntervalThreshold) {
+                return
+            }
+
+            try {
+                this.invokeAutomatedTrigger('IdleTime', editor, client, config)
+            } finally {
+                if (this.idleTriggerTimer) {
+                    clearInterval(this.idleTriggerTimer)
+                    this.idleTriggerTimer = undefined
+                }
+            }
+        }, CodeWhispererConstants.idleTimerPollPeriod)
+    }
+
+    public shouldTriggerIdleTime(): boolean {
+        if (isCloud9() && RecommendationHandler.instance.isGenerateRecommendationInProgress) {
+            return false
+        }
+        if (isInlineCompletionEnabled() && InlineCompletionService.instance.isPaginationRunning()) {
+            return false
+        }
+        if (InlineCompletion.instance.getIsActive || InlineCompletion.instance.isPaginationRunning()) {
+            return false
+        }
+        return true
     }
 
     async processKeyStroke(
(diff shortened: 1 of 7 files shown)


src/codewhisperer/service/keyStrokeHandler.ts
```diff
@@ -42,6 +42,49 @@ export class KeyStrokeHandler {
...
+            }
+
+            try {
+                this.invokeAutomatedTrigger('IdleTime', editor, client, config)
+            } finally {
+                if (this.idleTriggerTimer) {
+                    clearInterval(this.idleTriggerTimer)
+                    this.idleTriggerTimer = undefined
+                }
+            }
```
JadenSimon, line 71: Nice, good use of `try/finally`

src/codewhisperer/service/keyStrokeHandler.ts
```diff
@@ -42,6 +42,49 @@ export class KeyStrokeHandler {
...
+    public startIdleTimeTriggerTimer(
+        event: vscode.TextDocumentChangeEvent,
+        editor: vscode.TextEditor,
+        client: DefaultCodeWhispererClient,
+        config: ConfigurationEntry
+    ) {
+        if (this.idleTriggerTimer) {
+            clearInterval(this.idleTriggerTimer)
+            this.idleTriggerTimer = undefined
+        }
```
JadenSimon, line 54: nit: does `clearInterval` work if `idleTriggerTimer` is `undefined`? If so we can remove the `if` statement

src/codewhisperer/service/keyStrokeHandler.ts
```diff
@@ -42,6 +42,49 @@ export class KeyStrokeHandler {
...
+    public shouldTriggerIdleTime(): boolean {
+        if (isCloud9() && RecommendationHandler.instance.isGenerateRecommendationInProgress) {
+            return false
+        }
+        if (isInlineCompletionEnabled() && InlineCompletionService.instance.isPaginationRunning()) {
+            return false
+        }
+        if (InlineCompletion.instance.getIsActive || InlineCompletion.instance.isPaginationRunning()) {
+            return false
+        }
```
JadenSimon, line 84: nit: another way to do this is to assign each condition to a more descriptive name and check them all at the end

src/test/codewhisperer/service/keyStrokeHandler.test.ts
```diff
@@ -250,15 +235,15 @@ describe('keyStrokeHandler', function () {
...
-                .resolves([{ content: 'import math' }, { content: 'def two_sum(nums, target):' }])
-            await KeyStrokeHandler.instance.invokeAutomatedTrigger('Enter', mockEditor, mockClient, config)
-            assert.strictEqual(KeyStrokeHandler.instance.keyStrokeCount, 0)
+    describe('shouldTriggerIdleTime', function () {
+        it('should return false when inline is enabled and inline completion is in progress ', function () {
+            const keyStrokeHandler = new KeyStrokeHandler()
+            sinon.stub(InlineCompletionService.instance, 'isPaginationRunning').returns(true)
+            sinon.stub(InlineCompletion.instance, 'isPaginationRunning').returns(true)
+            const result = keyStrokeHandler.shouldTriggerIdleTime()
+            assert.strictEqual(result, false)
```
JadenSimon, line 246: `shouldTriggerIdleTime` could be changed to accept instances of `InlineCompletion` and `InlineCompletionService` as params to make things more explicit.

src/test/codewhisperer/service/keyStrokeHandler.test.ts
```diff
@@ -200,31 +190,26 @@ describe('keyStrokeHandler', function () {
...
                 mockEditor.document,
                 new vscode.Range(new vscode.Position(0, 0), new vscode.Position(0, 1)),
-                'a'
+                '('
             )
-            KeyStrokeHandler.instance.keyStrokeCount = 15
             await KeyStrokeHandler.instance.processKeyStroke(mockEvent, mockEditor, mockClient, config)
-            invokeSpy('KeyStrokeCount', mockEditor, mockClient)
-            assert.ok(invokeSpy.called)
+            assert.ok(!startTimerSpy.called)
```
JadenSimon, line 201: This is just feedback for future implementations:

One way to avoid spies is to make the thing your testing more "visible". For example, the goal here is to test whether or not the timer is running right? So we could make that a field on `KeyStrokeHandler` and implement a bit more logic to make it work correctly. Usually the problem (benefit?) with doing this is that more code will be executed which means more things need to be written in a testable way. But this is generally my thought-process when writing tests.

JadenSimon (approved):

src/codewhisperer/service/keyStrokeHandler.ts
```diff
@@ -42,6 +42,49 @@ export class KeyStrokeHandler {
...
+                if (this.idleTriggerTimer) {
+                    clearInterval(this.idleTriggerTimer)
+                    this.idleTriggerTimer = undefined
+                }
+            }
+        }, CodeWhispererConstants.idleTimerPollPeriod)
+    }
+
+    public shouldTriggerIdleTime(): boolean {
+        if (isCloud9() && RecommendationHandler.instance.isGenerateRecommendationInProgress) {
```
leigaol, line 76: Does this mean in C9 we are not using idle time trigger but we removed keystroke trigger?

src/test/codewhisperer/service/keyStrokeHandler.test.ts
leigaol: We are going to remove **InlineCompletion** soon since most people are on 1.68+ and they should be only using **InlineCompletionService**

src/codewhisperer/service/keyStrokeHandler.ts
zixlin7: in C9 we are also using idle time trigger, this condition is just to prevent concurrent triggering

zixlin7: we set this flag here: https://github.com/aws/aws-toolkit-vscode/blob/e6eb99db8d7e4e8b2f825f7c89660d633053d306/src/codewhisperer/service/keyStrokeHandler.ts#L167

leigaol (approved):
Don't forget to add change log item

JadenSimon (approved):

Issue #2922, opened 2022-10-07
Despite instructions to run toolkit in c9 without internet access, visualize step function graph requires internet connectivity that is easily patched.


simis2626:
## Problem
Based on the instructions [here](https://docs.aws.amazon.com/cloud9/latest/user-guide/toolkit-welcome.html#access-toolkit), aws toolkit should be largely functional without an internet connection if the user provides appropriate S3 endpoint access.

The functionality to render a Amazon Step Function State Machine Graph requires connectivity to:
- 'https://do0of8uwbahzz.cloudfront.net/sfn-0.1.5.js'
- 'https://do0of8uwbahzz.cloudfront.net/graph-0.1.5.css'
As in file [util.ts](https://github.com/aws/aws-toolkit-vscode/blob/master/src/stepFunctions/utils.ts#L27)

Based on this [codepath](https://github.com/aws/aws-toolkit-vscode/blob/481f71c019903696e72bee64898caaa1ddffbba9/src/stepFunctions/commands/visualizeStateMachine/aslVisualizationManager.ts#L36) this download process can't be skipped or otherwise configured.

However, as a user if you populate graph.js and graph.css under .c9-project

## Steps to reproduce the issue

** Prereq: set AWS toolkit to debug level logging.

1) Launch an EC2 based Cloud9 instance in a private VPC without routing to the public internet, but with the S3 VPC Endpoints as described in this doco: https://docs.aws.amazon.com/cloud9/latest/user-guide/ec2-ssm.html#configure-s3-endpoint
1) Create and save a basic test step function: in a file called `sfn.asl.yaml`
```
Comment: test
StartAt: first
States:
    first:
      Type: Pass
      End: true
```
1) Click the `render graph` text at the top of the file. 
1) Observe in the logs that this [codepath](https://github.com/aws/aws-toolkit-vscode/blob/481f71c019903696e72bee64898caaa1ddffbba9/src/stepFunctions/commands/visualizeStateMachine/aslVisualizationManager.ts#L36) fires, and fails as there's no internet access.

## Expected behavior

Expose configuration options to allow skipping/bypassing [this specific line](https://github.com/aws/aws-toolkit-vscode/blob/master/src/stepFunctions/commands/visualizeStateMachine/aslVisualizationManager.ts#L36), or check for the presence of the file `graph.js` & `graph.css` under `/.c9/amazonwebservices.aws-toolkit-vscode/visualization/` and if present progress without attempting to download.

OR

[...]

PR #3098, opened 2023-01-12
Fix/asl visualizer no internet

simis2626:
PR for #2922 - sfn render fails if host has no internet connectivity, even with cached files present locally.

## Problem
Step Function Visualizer fails to render graph due to trying to source the graph.js and graph.css files from the internet on an disconnected c9/code instance - even when those files are available locally.
## Solution
On failure of network connectivity, falls back to local files where they both exist.

- [x] Tests added
- [x] Changelog updated
- [ ] Open to futher feedback on implementation from maintainers.

## License

### DIFF
--- a/src/stepFunctions/commands/visualizeStateMachine/aslVisualizationManager.ts
+++ b/src/stepFunctions/commands/visualizeStateMachine/aslVisualizationManager.ts
@@ -33,7 +33,20 @@ export class AslVisualizationManager extends AbstractAslVisualizationManager {
 
         // Existing visualization does not exist, construct new visualization
         try {
-            await this.cache.updateCache(globalStorage)
+            try {
+                await this.cache.updateCache(globalStorage)
+            } catch (err) {
+                // So we can't update the cache, but can we use an existing on disk version.
+                try {
+                    logger.warn(
+                        'Updating State Machine Graph Visualisation assets failed, checking for fallback local cache.'
+                    )
+                    await this.cache.confirmCacheExists()
+                } catch (err) {
+                    logger.error('No local cached State Machine Graph Visualization assets found.')
+                    throw err
+                }
+            }
 
             const newVisualization = new AslVisualization(document)
             this.handleNewVisualization(document.uri.fsPath, newVisualization)
--- a/src/stepFunctions/utils.ts
+++ b/src/stepFunctions/utils.ts
@@ -37,6 +37,10 @@ export interface UpdateCachedScriptOptions {
     filePath: string
 }
 
+export interface ConfirmCachedFileExistsOptions {
+    filePath: string
+}
+
 export interface StateMachineGraphCacheOptions {
     cssFilePath?: string
     jsFilePath?: string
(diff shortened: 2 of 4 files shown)


src/stepFunctions/commands/visualizeStateMachine/aslVisualizationManager.ts
```diff
@@ -33,7 +33,20 @@ export class AslVisualizationManager extends AbstractAslVisualizationManager {
...
+                try {
+                    logger.warn(
+                        'Updating State Machine Graph Visualisation assets failed, checking for fallback local cache.'
+                    )
+                    await this.cache.confirmCacheExists()
+                } catch (err) {
+                    logger.error('No local cached State Machine Graph Visualization assets found.')
+                    throw err
+                }
+            }
```
JadenSimon, line 49: Should we move this logic to `StateMachineGraphCache.updateCache` and tweak the log messages a bit? That way `AslVisualizationCDKManager` would benefit from this fix too. Or we could add a new method to `AbstractAslVisualizationManager` and call that instead of `this.cache.updateCache`

src/stepFunctions/utils.ts
```diff
@@ -117,6 +121,38 @@ export class StateMachineGraphCache {
...
+                // This log message is intended to help users setup on disconnected C9/VSCode instances.
+                this.logger.error(`Expected to find: "${VISUALIZATION_CSS_URL}" at "${this.cssFilePath}"`)
+            }
+            if (!jsExists) {
+                this.logger.error('Failed to locate cached State Machine Graph css assets')
+                // This log message is intended to help users setup on disconnected C9/VSCode instances.
+                this.logger.error(`Expected to find: "${VISUALIZATION_SCRIPT_URL}" at "${this.jsFilePath}"`)
+            }
+            const err = new Error('Failed to located cached State Machine Graph assets')
+            throw err
```
JadenSimon, line 146: nit:
```suggestion
            throw new Error('Failed to located cached State Machine Graph assets')
```

src/stepFunctions/utils.ts
```diff
@@ -117,6 +121,38 @@ export class StateMachineGraphCache {
...
+    public async confirmCacheExists(): Promise<boolean> {
+        const cssExists = await this.confirmCachedFileExists({
+            filePath: this.cssFilePath,
+        })
+        const jsExists = await this.confirmCachedFileExists({
+            filePath: this.jsFilePath,
+        })
+        if (cssExists && jsExists) {
+            return true
+        } else {
```
JadenSimon, line 134: nit: no need to use `else` if the other branch returns

src/stepFunctions/utils.ts
```diff
@@ -117,6 +121,38 @@ export class StateMachineGraphCache {
...
     }
 
+    // Coordinates check for multiple cached files.
+    public async confirmCacheExists(): Promise<boolean> {
+        const cssExists = await this.confirmCachedFileExists({
+            filePath: this.cssFilePath,
+        })
+        const jsExists = await this.confirmCachedFileExists({
+            filePath: this.jsFilePath,
+        })
```
JadenSimon, line 131: `confirmCachedFileExists` seems unnecessary
```suggestion
        const cssExists = await this.fileExists(this.cssFilePath)
        const jsExists = await this.fileExists(this.jsFilePath)
```

src/stepFunctions/commands/visualizeStateMachine/aslVisualizationManager.ts
simis2626: Hi @JadenSimon , thanks for the review.
I'm very happy to move the logic, my concern was messing up the return meaning of `
updateCache` i.e. it would return positive/no errors even though the cache wasn't updated.

I'll have a look at `AbstractAslVisualizationManager` as hopefully a better pathway.

JadenSimon: Makes sense to me. `AbstractAslVisualizationManager` is a better place if we don't want to change semantics of existing code.

src/stepFunctions/utils.ts
```diff
@@ -117,6 +121,38 @@ export class StateMachineGraphCache {
...
+        if (cssExists && jsExists) {
+            return true
+        } else {
+            if (!cssExists) {
+                this.logger.error('Failed to locate cached State Machine Graph css assets')
+                // This log message is intended to help users setup on disconnected C9/VSCode instances.
+                this.logger.error(`Expected to find: "${VISUALIZATION_CSS_URL}" at "${this.cssFilePath}"`)
```
justinmk3, line 138: log messages should try to stay "atomic" if possible

```suggestion
                // This log message is intended to help users setup on disconnected C9/VSCode instances.
                this.logger.error(`Failed to locate cached State Machine Graph css assets. Expected to find: "${VISUALIZATION_CSS_URL}" at "${this.cssFilePath}"`)
```

yaythomas:
Thank you @simis2626 for your great work here!

I've picked up your original code, incorporated the excellent review comments & suggestions from @justinmk3 and @JadenSimon  and opened a new PR #3485 containing all the changes.

I squashed all these commits for a clean commit log, but you're still credited as the author - great first contribution, thank you so much!

Once that PR is merged, the maintainers can close this PR.

PR #3099, opened 2023-01-13
feat(sam): improved TypeScript support

sseidel16:
## Problem
TypeScript debugger has a race condition that breaks debugging much of the time. Hard-coded tsconfig also is used which breaks anything more than a very basic typescript lambda.

## Solution
Updated toolkit to generate tsconfig that attempts to match project tsconfig as close as possible, code is compiled into a ts build directory, sam deploys from there on an updated handler path (pointing to build dir). This eliminates the debugger race condition, and much more closely resembles a standard project build eliminating all of our errors. It also matches jetbrains approach [here](https://github.com/aws/aws-toolkit-jetbrains/blob/main/jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/lambda/nodejs/NodeJsLambdaBuilder.kt).

Code updated:

- src/shared/sam/activation.ts: allow typescript code to be recognized
- src/shared/sam/debugger/awsSamDebugger.ts: code template generation after individual runtime configs are made so ts can update handler path
- src/shared/sam/debugger/typescriptSamDebug.ts: updated build directory process matching jetbrains (most of the updates)
- src/shared/sam/localLambdaRunner.ts: split getting template path and creating the template to facilitate awsSamDebugger.ts changes
- src/test/shared/sam/debugger/samDebugConfigProvider.test.ts: update ts tests

Tested with unit tests, and with several typescript projects including large and small lambdas. New code lines are covered with code coverage.

## License

### DIFF
--- a/.changes/next-release/Feature-46a93a97-f979-4c7a-9a8d-f241a30987b6.json
+++ b/.changes/next-release/Feature-46a93a97-f979-4c7a-9a8d-f241a30987b6.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Improved typescript support to match jetbrains plugin"
+}
--- a/.changes/next-release/Feature-46a93a97-f979-4c7a-9a8d-f241a30987b6.json
+++ b/.changes/next-release/Feature-46a93a97-f979-4c7a-9a8d-f241a30987b6.json
@@ -1,4 +1,4 @@
 {
 	"type": "Feature",
-	"description": "Improved typescript support to match jetbrains plugin"
+	"description": "SAM: run/debug now works for a larger variety of TypeScript projects"
 }
--- a/src/shared/sam/debugger/typescriptSamDebug.ts
+++ b/src/shared/sam/debugger/typescriptSamDebug.ts
@@ -19,7 +19,6 @@ import { DefaultSamLocalInvokeCommand, WAIT_FOR_DEBUGGER_MESSAGES } from '../cli
 import { runLambdaFunction, waitForPort } from '../localLambdaRunner'
 import { SamLaunchRequestArgs } from './awsSamDebugger'
 
-const tsBuildDir = 'aws-toolkit-ts-output'
 const tsConfigFile = 'aws-toolkit-tsconfig.json'
 
 // use project tsconfig.json as initial base - if unable to parse existing config
@@ -167,7 +166,6 @@ async function compileTypeScript(config: SamLaunchRequestArgs): Promise<void> {
     }
 
     const tsConfigPath = path.join(config.codeRoot, tsConfigFile)
-    const tsOutputPath = path.join(config.codeRoot, tsBuildDir)
 
     const tsConfig =
         loadBaseConfig(tsConfigPath) ?? loadBaseConfig(path.join(config.codeRoot, tsConfigInitialBaseFile)) ?? {}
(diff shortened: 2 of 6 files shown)


src/shared/sam/debugger/typescriptSamDebug.ts
```diff
@@ -18,6 +19,12 @@ import { DefaultSamLocalInvokeCommand, WAIT_FOR_DEBUGGER_MESSAGES } from '../cli
 import { runLambdaFunction, waitForPort } from '../localLambdaRunner'
 import { SamLaunchRequestArgs } from './awsSamDebugger'
 
+const TS_BUILD_DIR = 'aws-toolkit-ts-output'
```
JadenSimon, line 22: Doing this is ok but now there's going to be inconsistencies between the runtimes. I can make a PR to simplify some of the run/debug codepaths to make things consistent while still accommodating your improvements.

src/shared/sam/localLambdaRunner.ts
```diff
@@ -90,7 +96,7 @@ export async function makeInputTemplate(
         .withRuntime(config.lambda!.runtime!)
         .withCodeUri(pathutil.normalize(config.invokeTarget.projectRoot))
 
-    if (config.lambda?.environmentVariables) {
+    if (config.lambda?.environmentVariables && Object.entries(config.lambda?.environmentVariables).length) {
```
JadenSimon, line 99: Nice improvement.

.changes/next-release/Feature-46a93a97-f979-4c7a-9a8d-f241a30987b6.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Improved typescript support to match jetbrains plugin"
```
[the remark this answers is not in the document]
JadenSimon: Oh I guess the other important change is the CodeLens. We could make that another entry or just ignore my suggestion.

src/shared/sam/debugger/typescriptSamDebug.ts
sseidel16: Good call. Updating to use outDir in tsconfig by default (or buildDir under sam in launch configuration if set).

src/shared/sam/debugger/typescriptSamDebug.ts
```diff
@@ -134,46 +142,76 @@ async function compileTypeScript(config: NodejsDebugConfiguration): Promise<void
...
+    compilerOptions.typeRoots = [...new Set(typeRoots)]
+
+    const types: string[] = Array.isArray(compilerOptions.types) ? compilerOptions.types : []
+    types.push('node')
+    compilerOptions.types = [...new Set(types)]
+
+    writeFileSync(tsConfigPath, JSON.stringify(tsConfig, undefined, 4))
+
+    // resolve ts lambda handler to point into build directory relative to codeRoot
+    const tsLambdaHandler = path.relative(config.codeRoot, `${tsBuildDir}/${config.invokeTarget.lambdaHandler}`)
```
JadenSimon, line 203:
```suggestion
    const tsLambdaHandler = path.relative(config.codeRoot, path.join(tsBuildDir, config.invokeTarget.lambdaHandler))
```

src/shared/sam/debugger/typescriptSamDebug.ts
```diff
@@ -134,46 +142,76 @@ async function compileTypeScript(config: NodejsDebugConfiguration): Promise<void
...
+        }
+
+        return undefined
     }
 
-    // Default config.
-    // Adapted from: https://github.com/aws/aws-toolkit-jetbrains/blob/911c54252d6a4271ee6cacf0ea1023506c4b504a/jetbrains-ultimate/src/software/aws/toolkits/jetbrains/services/lambda/nodejs/NodeJsLambdaBuilder.kt#L60
-    const defaultTsconfig = {
-        compilerOptions: {
+    const tsConfigPath = path.join(config.codeRoot, tsConfigFile)
```
JadenSimon, line 168: It'd be really nice if we could write this to the build directory if possible. Maybe we can use [extends](https://www.typescriptlang.org/tsconfig#extends) instead of directly merging config files?

JadenSimon (approved):

PR #3113, opened 2023-01-23
fix(codecatalyst): cache results of `verifySession`

JadenSimon:
## Problem
`verifySession` is called every time a new client is created

## Solution
Cache `verifySession` based off the access token

### Testing
The way the CodeCatalyst client is written makes it difficult to test in a sustainable way. Something similar to https://github.com/aws/aws-toolkit-vscode/pull/2917 needs to be done to reduce the amount of custom logic in each service client.

## License

### DIFF
--- a/src/shared/clients/codecatalystClient.ts
+++ b/src/shared/clients/codecatalystClient.ts
@@ -320,6 +320,7 @@ class CodeCatalystClientInternal {
         const resp = await this.call(this.sdkClient.verifySession(), false)
         assertHasProps(resp, 'identity')
 
+        CodeCatalystClientInternal.identityCache.set(accessToken, resp.identity)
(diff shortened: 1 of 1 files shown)


src/shared/clients/codecatalystClient.ts
```diff
@@ -320,6 +320,7 @@ class CodeCatalystClientInternal {
         const resp = await this.call(this.sdkClient.verifySession(), false)
         assertHasProps(resp, 'identity')
 
+        CodeCatalystClientInternal.identityCache.set(accessToken, resp.identity)
```
justinmk3, line 323: do we need a source comment here, this seems kind of special and could get lost or at least the purpose isn't obvious

justinmk3 (approved):

src/shared/clients/codecatalystClient.ts
JadenSimon: Yeah I'll add some notes

PR #3114, opened 2023-01-24
cwl: validate search input

justinmk3:
## Problem

The "filter pattern" step is not validated until after the wizard closes, so customer must start the wizard again if CloudWatch logs API rejects the filter (search) pattern.

Because of current limitations of vscode and/or our Wizard framework, we can't dynamically make a service call because _async_ `onDidAccept` event handler is not await'd.

## Solution

image

- Change "Search Pattern" step to be the last step.
- Use a workaround similar to:
https://github.com/aws/aws-toolkit-vscode/blob/183dd3ae4bf14196962c242e9236092ae509e3b0/src/credentials/wizards/createProfile.ts#L69-L104
    - Subclass `InputBoxPrompter`
    - Override `promptUser()`. If validation fails,
        - return `WIZARD_RETRY`
        - restore state in a hacky way (the `retryState` object)

## License

### DIFF
--- a/src/cloudWatchLogs/commands/searchLogGroup.ts
+++ b/src/cloudWatchLogs/commands/searchLogGroup.ts
@@ -10,7 +10,7 @@ import {
     CloudWatchLogsData,
     CloudWatchLogsGroupInfo,
     LogDataRegistry,
-    filterLogEventsFromUriComponents,
+    filterLogEventsFromUri,
     CloudWatchLogsParameters,
     initLogData,
 } from '../registry/logDataRegistry'
@@ -53,7 +53,7 @@ function handleWizardResponse(response: SearchLogGroupWizardResponse, registry: 
         }
     }
 
-    const logData = initLogData(logGroupInfo, parameters, filterLogEventsFromUriComponents)
+    const logData = initLogData(logGroupInfo, parameters, filterLogEventsFromUri)
 
     if (logData.parameters.startTime || logData.parameters.filterPattern) {
         recordTelemetryFilter(logData, 'logGroup', 'Command')
@@ -172,7 +172,7 @@ export function createFilterpatternPrompter(
             endTime: 9999,
         }
         try {
-            await filterLogEventsFromUriComponents(logGroupInfo, parameters)
+            await filterLogEventsFromUri(logGroupInfo, parameters)
         } catch (e) {
             return (e as Error).message
         }
--- a/src/cloudWatchLogs/commands/searchLogGroup.ts
+++ b/src/cloudWatchLogs/commands/searchLogGroup.ts
@@ -15,7 +15,7 @@ import {
     initLogData,
 } from '../registry/logDataRegistry'
 import { DataQuickPickItem } from '../../shared/ui/pickerPrompter'
-import { Wizard } from '../../shared/wizards/wizard'
+import { isValidResponse, isWizardControl, Wizard, WIZARD_RETRY } from '../../shared/wizards/wizard'
 import { createURIFromArgs, parseCloudWatchLogsUri, recordTelemetryFilter } from '../cloudWatchLogsUtils'
 import { DefaultCloudWatchLogsClient } from '../../shared/clients/cloudWatchLogsClient'
 import { CancellationError } from '../../shared/utilities/timeoutUtils'
@@ -23,10 +23,11 @@ import { getLogger } from '../../shared/logger'
 import { TimeFilterResponse, TimeFilterSubmenu } from '../timeFilterSubmenu'
 import { LogGroupNode } from '../explorer/logGroupNode'
 import { CloudWatchLogs } from 'aws-sdk'
-import { createInputBox, InputBoxPrompter } from '../../shared/ui/inputPrompter'
+import { ExtendedInputBoxOptions, InputBox, InputBoxPrompter } from '../../shared/ui/inputPrompter'
 import { RegionSubmenu, RegionSubmenuResponse } from '../../shared/ui/common/regionSubmenu'
 import { truncate } from '../../shared/utilities/textUtilities'
 import { createBackButton, createExitButton, createHelpButton } from '../../shared/ui/buttons'
+import { PromptResult } from '../../shared/ui/prompter'
 
 const localize = nls.loadMessageBundle()
 
@@ -68,8 +69,8 @@ export async function prepareDocument(
     registry: LogDataRegistry
 ): Promise<Result> {
     try {
+        // Gets the data: calls filterLogEventsFromUri().
         await registry.fetchNextLogEvents(uri)
-        // Gets the data: calls filterLogEventsFromUriComponents().
         const doc = await vscode.workspace.openTextDocument(uri)
         await vscode.window.showTextDocument(doc, { preview: false })
         vscode.languages.setTextDocumentLanguage(doc, 'log')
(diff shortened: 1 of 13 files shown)


src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -202,9 +198,9 @@ export class SearchLogGroupWizard extends Wizard<SearchLogGroupWizardResponse> {
         })
 
         this.form.submenuResponse.bindPrompter(createRegionSubmenu)
+        this.form.timeRange.bindPrompter(() => new TimeFilterSubmenu())
```
JadenSimon, line 201: This changes the prompt order which the tests check.

JadenSimon (approved):

src/cloudWatchLogs/commands/searchLogGroup.ts
justinmk3: that's intentional. i have some more changes to push, but the idea was that if the service rejects the search input, the wizard should not be hidden, the user should be able to edit the search input.

justinmk3:
i'm stuck. looking for away to "validate" the search query by making a service call. current blocker(s):

- seems like async validation functions are not await'd, so can't make service calls in the wizard validator
- haven't found a way for wizard steps to share information, e.g. how can Step 2 use the input from Step 1?

JadenSimon:
> seems like async validation functions are not await'd, so can't make service calls in the wizard validator

Yup this was something I never got the chance to implement but it is possible. I believe there's an older PR that does this.

> haven't found a way for wizard steps to share information, e.g. how can Step 2 use the input from Step 1?

This should be done using the `state` arg passed into the prompter callback:
```ts
```

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -97,37 +100,30 @@ export async function prepareDocument(
...
-        response = await new SearchLogGroupWizard().run()
+    const source = node ? 'Explorer' : 'Command'
+    if (node && !node.logGroup.logGroupName) {
+        throw new Error('CWL: Log Group node does not have a name.')
     }
 
+    const wizard = node?.logGroup.logGroupName
+        ? new SearchLogGroupWizard(registry, { groupName: node.logGroup.logGroupName, regionName: node.regionCode })
+        : new SearchLogGroupWizard(registry)
+    const response = await wizard.run()
```
justinmk3, line 114: no functional change, just flattening the logic

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -159,12 +160,36 @@ export function createFilterpatternPrompter(logGroupName: string, isFirst: boole
...
+        'search pattern (case sensitive; empty matches all)'
     )
+
+    async function validateSearchPattern(searchPattern: string, isFinalInput: boolean): Promise<string | undefined> {
+        if (!isFinalInput) {
+            return undefined // Skip validation while user is typing.
+        }
+        const logGroupInfo: CloudWatchLogsGroupInfo = {
+            groupName: '/aws/apprunner/aaaaaaa/5d2870895ec34e7d810fb316cd9c9bf0',
+            regionName: 'us-west-2',
```
justinmk3, line 172: need to get results of previous steps here

src/shared/ui/inputPrompter.ts
```diff
@@ -9,16 +9,17 @@ import { isValidResponse, StepEstimator, WIZARD_BACK, WIZARD_EXIT } from '../wiz
...
-}
+export type ExtendedInputBoxOptions =
+    // TODO: allow `validateInput` to return Thenable so we don't need to omit it here.
+    Omit<vscode.InputBoxOptions, 'validateInput' | 'placeHolder'> & {
+        title?: string
+        step?: number
+        placeholder?: string
+        totalSteps?: number
+        buttons?: PrompterButtons<string>
+        validateInput?(value: string, isFinalInput?: boolean): string | undefined | Thenable<string | undefined>
```
justinmk3, line 21: changed signature to allow Thenable

src/shared/ui/inputPrompter.ts
```diff
@@ -85,18 +86,21 @@ export class InputBoxPrompter extends Prompter<string> {
...
 
         this.inputBox.onDidChangeValue(
-            value => (this.inputBox.validationMessage = validate(value)),
+            async value => (this.inputBox.validationMessage = await validate(value, false)),
             undefined,
             this.validateEvents
         )
         this.inputBox.onDidAccept(
-            () => (this.inputBox.validationMessage = validate(this.inputBox.value)),
+            async () => (this.inputBox.validationMessage = await validate(this.inputBox.value, true)),
```
justinmk3, line 103: seems like vscode is not awaiting this.

JadenSimon: Ahh ok so for input boxes VSC doesn't handle async validation callbacks. We could make a contribution to their codebase for this, otherwise the solution on our end is messy.

justinmk3: ok i'll try https://github.com/aws/aws-toolkit-vscode/blob/b4695d472e7d9cc5285f27d9a1bd9c10f5e88a2e/src/credentials/wizards/createProfile.ts#L69-L104 then.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -149,18 +146,95 @@ async function logGroupsToArray(logGroups: AsyncIterableIterator<CloudWatchLogs.
...
+ *    - https://github.com/aws/aws-toolkit-vscode/pull/3114#discussion_r1085484630
+ *    - https://github.com/microsoft/vscode/blob/78947444843f4ebb094e5ab4288360010a293463/extensions/git-base/src/remoteSource.ts#L13
+ *    - https://github.com/microsoft/vscode/blob/78947444843f4ebb094e5ab4288360010a293463/src/vs/base/browser/ui/inputbox/inputBox.ts#L511
+ */
+export class SearchPatternPrompter extends InputBoxPrompter {
+    constructor(
+        public logGroup: CloudWatchLogsGroupInfo,
+        public logParams: CloudWatchLogsParameters,
+        /** HACK: also maintain ad-hoc state because `wizardState` is not mutable. */
+        public readonly retryState: any,
```
justinmk3, line 163: 😬  I tried attaching stuff to `WizardState` but it's immutable (that's probably a good thing :)

I need some way to set `this.recentItem` and `inputBox.validationMessage` (see below).

JadenSimon: Yeah it's immutable. Overriding `InputBoxPrompter` is a good strategy if you want a lot of customization.

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -149,18 +146,95 @@ async function logGroupsToArray(logGroups: AsyncIterableIterator<CloudWatchLogs.
...
+            this.retryState.searchPattern = rv?.toString()
+            if (validationResult !== undefined) {
+                // Doesn't work because state is recreated on WIZARD_RETRY?
+                // this.inputBox.validationMessage = validationResult
+                // HACK: maintain our own state and restore it.
+                this.retryState.validationMessage = validationResult
+
+                if (rv) {
+                    // Doesn't work because state is recreated on WIZARD_RETRY?
+                    // this.recentItem = rv.toString()
```
JadenSimon, line 192: Probably need to remove `recentItem` from the interface as it can be confusing. The wizard control flow logic will override your assignment.

src/shared/ui/inputPrompter.ts
```diff
@@ -84,19 +102,19 @@ export class InputBoxPrompter extends Prompter<string> {
...
+     * @param validate Validator function.
      */
-    public setValidation(validate: (value: string) => string | undefined): void {
+    public setValidation(validate: ValidateFn): void {
         this.validateEvents.forEach(d => d.dispose())
         this.validateEvents = []
 
         this.inputBox.onDidChangeValue(
-            value => (this.inputBox.validationMessage = validate(value)),
+            value => (this.inputBox.validationMessage = validate(value, false)),
```
justinmk3, line 112: Reverted the Thenable change https://github.com/aws/aws-toolkit-vscode/pull/3114/commits/7d1c81c8203ea2dfbedddf5e85f38a95e5f08f76 for now since it doesn't work with this `inputBox.onDidChangeValue` and `inputBox.onDidAccept`.

PR #3115, opened 2023-01-24
fix(inputPrompter): misplaced disposables argument

justinmk3:
Problem:
`this.validateEvents` is passed as the `thisArgs` parameter.

Solution:
pass `undefined` to `thisArgs` parameter, pass `this.validateEvents` to `disposables` parameter.

## License


JadenSimon (approved):
We should avoid passing the disposables array as a param for this reason.

PR #3116, opened 2023-01-24
tests(codewhisperer): fixed unit test case failing on windows for truncation lines

vreshu-AWS:
## Problem
Unit test cases of truncation lines for codeWhisperer were failing on windows OS.

## Solution
We have to use vscode.uri.file() rather than vscode.Uri.parse().
The *difference* between `Uri#parse` and `Uri#file` is that the latter treats the argument
 as path, not as stringified-uri. E.g. `Uri.file(path)` is *not* the same as `Uri.parse(‘file://’ + path)` because the path might contain characters that are interpreted (# and ?).

## License


fangwade007:
> I referred to this [link](https://docs.npmjs.com/cli/v6/configuring-npm/package-locks#:~:text=It%20is%20highly%20recommended%20you%20commit%20the%20generated%20package%20lock%20to%20source%20control%3A%20this%20will%20allow%20anyone%20else%20on%20your%20team%2C%20your%20deployments%2C%20your%20CI/continuous%20integration%2C%20and%20anyone%20else%20who%20runs%20npm%20install%20in%20your%20package%20source%20to%20get%20the%20exact%20same%20dependency%20tree%20that%20you%20were%20developing%20on.)
> 
> Is there any reason to exclude package.lock file ?

I was suggesting to exclude the influence of the local versions, as your commit doesn't contain any dependency changes. But I mark it as `nit`, I'm will give approval as long as toolkit team are fine w/ `package-lock.json` changes.

src/test/codewhisperer/util/dependencyGraph/javascriptDependencyGraph.test.ts
```diff
@@ -59,12 +59,12 @@ describe('javascriptDependencyGraph', function () {
     describe('generateTruncation', function () {
         it('Should generate and return expected truncation', async function () {
             const javascriptDependencyGraph = new JavascriptDependencyGraph(CodeWhispererConstants.javascript)
-            const truncation = await javascriptDependencyGraph.generateTruncation(vscode.Uri.parse(appCodePath))
+            const truncation = await javascriptDependencyGraph.generateTruncation(vscode.Uri.file(appCodePath))
```
fangwade007, line 62: nit: what about `java` dependency graph test ?

JadenSimon:
I would revert all changes to `package-lock.json` because:
* There are no dependency updates
* The diff look excessively noisy, possibly due to a difference in `npm` versions

In the future, we will hopefully be able to automatically check and fix these problems. `package-lock.json` is only relevant for build tooling, nothing more.

justinmk3:
for future refence: please include the Problem/Solution in the commit message.

fangwade007 (approved):

PR #3123, opened 2023-01-25
fix(codewhisperer): update error popup for access token migration

zixlin7:
## Problem
to update the error popup shown to users with access code after 1/31 cutoff date based on product decisions

- update the message shown on popup
- add "Don't show again" button
- add "Learn more" button with link to documentation

## Solution
before:

Screen Shot 2023-01-24 at 1 49 42 PM

after:

Screen Shot 2023-01-24 at 1 57 25 PM

## License

### DIFF
--- a/src/codewhisperer/activation.ts
+++ b/src/codewhisperer/activation.ts
@@ -251,8 +251,9 @@ export async function activate(context: ExtContext): Promise<void> {
             if (t <= CodeWhispererConstants.accessTokenCutOffDate) {
                 maybeShowTokenMigrationWarning()
             } else {
-                globals.context.globalState.update(CodeWhispererConstants.accessToken, undefined)
-                globals.context.globalState.update(CodeWhispererConstants.accessTokenExpriedKey, true)
+                await globals.context.globalState.update(CodeWhispererConstants.accessToken, undefined)
+                await globals.context.globalState.update(CodeWhispererConstants.accessTokenExpriedKey, true)
+                await vscode.commands.executeCommand('aws.codeWhisperer.refreshRootNode')
                 maybeShowTokenMigrationError()
             }
         } else if (accessTokenExpired) {
--- a/src/codewhisperer/activation.ts
+++ b/src/codewhisperer/activation.ts
@@ -51,6 +51,7 @@ import { AuthUtil, isUpgradeableConnection } from './util/authUtil'
 import { Auth } from '../credentials/auth'
 import { isUserCancelledError } from '../shared/errors'
 import { showViewLogsMessage } from '../shared/utilities/messages'
+import globals from '../shared/extensionGlobals'
 
 const performance = globalThis.performance ?? require('perf_hooks').performance
 
@@ -223,6 +224,9 @@ export async function activate(context: ExtContext): Promise<void> {
 
     async function showAccessTokenMigrationDialogue() {
         // TODO: Change the color of the buttons
+        const accessTokenExpired =
+            context.extensionContext.globalState.get<boolean>(CodeWhispererConstants.accessTokenExpriedKey) || false
+
         if (AuthUtil.instance.hasAccessToken()) {
             await Auth.instance.tryAutoConnect()
             const conn = Auth.instance.activeConnection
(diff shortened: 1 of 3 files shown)


src/codewhisperer/activation.ts
```diff
@@ -295,8 +300,14 @@ export async function activate(context: ExtContext): Promise<void> {
...
                                 true
                             )
+                        } else if (resp === CodeWhispererConstants.accessTokenMigrationLearnMore) {
+                            await vscode.commands.executeCommand('aws.codeWhisperer.accessTokenErrorLearnMore')
                         }
                     })
+                context.extensionContext.globalState.update(
+                    CodeWhispererConstants.accessTokenMigrationDoNotShowLastShown,
+                    Date.now()
+                )
```
andrewyuq, line 310: Here we are reusing these `accessTokenMigrationDoNotShowLastShown` and `accessTokenMigrationDoNotShowAgainKey` which are currently used for the warningMessage (before 1/31). So one issue it can happen is that the user closes the message on 1/30 and the next error message cannot show until 2/6(7 days after), which is not ideal(and similar for the `do not show again" again) Ideally they should have different "do not show again" status and different 7 days trackers.

andrewyuq: Another thing is all of these logic are currently under `if (AuthUtil.instance.hasAccessToken()) {` on line 226. But this boolean is no longer true after the first time it upgrades the connection. So next time(7 days after) it will not enter this if block. Try to come up with a condition that you know this user previously has accessToken and didn't upgrade the connection and simply closed the error notification(without checking "do not show again").

leigaol:
If this PR is merged after 1/31, you can remove code that only runs before 1/31.

andrewyuq:
> If this PR is merged after 1/31, you can remove code that only runs before 1/31.

It doesn't make any sense for this PR to be merged post the 1/31 date.

src/codewhisperer/activation.ts
zixlin7: makes sense, will update

zixlin7: @andrewyuq updated to use different states. For the second issue you mentioned. `AuthUtil.instance.hasAccessToken()` checks if there is a token string stored in the state, this state only gets cleared when user successfully connect to Sono/SSO. So if user had access token and closes the error popup without clicking "Don't show again", they will see the popup after 7 days. If user had access token but choose to connect with Sono/SSO, they won't see this popup ever again. Let me know if this is the expected behavior. Thanks!

andrewyuq: So we are saying, the accessToken is still there on-disk for the user, but we clear the accessToken connection after 1/31 so they will remain logged out but the accessToken will still be there until user connects to AWS Builder ID?

zixlin7: actually the accessToken will also get cleared if the user try to invoke the service using the token and got back "AccessDenied" error, and then they will not see the popup and maybe that's not ideal. Will sync with you offline.

src/codewhisperer/activation.ts
```diff
@@ -240,67 +243,98 @@ export async function activate(context: ExtContext): Promise<void> {
...
-                                true
-                            )
-                        }
-                    })
+                maybeShowTokenMigrationError()
             }
         }
     }
 
+    function maybeShowTokenMigrationWarning() {
```
justinmk3, line 255: This function and `maybeShowTokenMigrationError` are almost identical. Please avoid copy-pasting code that is clearly amenable to generalization.

If this code will be removed entirely in the near future, then we don't need to worry about it. But, this feature did have some bugs in previous iterations, and verbose, redundant code can be harder to understand....

zixlin7: yes this function will be removed after 1/31

src/codewhisperer/activation.ts
zixlin7: discussed offline, we will first nullify accessToken on 1/31 and set another flag to track if previously had an accessToken to cover all cases

andrewyuq (approved):
Logic looks good

justinmk3 (approved):

leigaol (approved):
Remember to clean up code after 1/31. Thanks!

PR #3128, opened 2023-01-27
feat(cwl): Allow cwl ctrl + click on whole line in an event stream

nkomonen-amazon:
## Problem
In cloudwatch logs when we search logs and ctrl + click we previously were only able to do this on the timestamp.

## Solution
Remove the limitation that only allows you to click on the timestamp and instead be able to click anywhere on the line.

Additionally, the whole line gets highlighted upon hover.

Kapture 2023-01-30 at 18 58 19

## License

### DIFF
--- a/src/cloudWatchLogs/document/logDataDocumentProvider.ts
+++ b/src/cloudWatchLogs/document/logDataDocumentProvider.ts
@@ -44,11 +44,6 @@ export class LogDataDocumentProvider implements vscode.TextDocumentContentProvid
     ): Promise<vscode.Definition | vscode.LocationLink[] | undefined> {
         const activeUri = document.uri
         const logGroupInfo = parseCloudWatchLogsUri(activeUri).logGroupInfo
-        const cutOffChar = this.registry.timestampSpaceEquivalent.length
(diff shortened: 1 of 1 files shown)


JadenSimon (approved):

justinmk3 (approved):

justinmk3:
Does this _highlight_ (underline) the whole line also? The main issue was that only small pieces of the timestamp were highlighted (underlined) on hover.

nkomonen-amazon:
> Does this _highlight_ (underline) the whole line also? The main issue was that only small pieces of the timestamp were highlighted (underlined) on hover.
@justinmk3 
Yup, I updated that after. Just updated the PR message.

justinmk3:
Tested locally, this is an improvement. Remaining questions:

- holding ctrl (or cmd on macOS) and hovering the mouse over a line does a service call (which then triggers the cancelable dialog). Is there any way to prevent that?
- hover shows an empty box. if we don't have anything useful to show there, can we prevent this empty thing from showing? see screenshot

Screen Shot 2023-01-30 at 4 03 22 PM

PR #3129, opened 2023-01-27
fix(codewhisperer): add additional delay time before showing recommen…

zixlin7:
…dation

## Problem
the default delay time before showing recommendation in VSC inline api is around 30-50ms, which is too short. This often result in recommendation being shown to users when they are still typing.
## Solution
Add additional 200ms to get the number close to 250ms which matches the JB behavior

## License

### DIFF
--- a/src/codewhisperer/models/constants.ts
+++ b/src/codewhisperer/models/constants.ts
@@ -22,6 +22,8 @@ export const invocationKeyThreshold = 15
 
 export const idleTimerPollPeriod = 25 // milliseconds
 
+export const showRecommendationTimerPollPeriod = 25
+
 export const specialCharactersList = ['{', '[', '(', ':', '\t', '\n']
 
 export const normalTextChangeRegex = /[A-Za-z0-9]/g
@@ -152,6 +154,9 @@ export const defaultCheckPeriodMillis = 1000 * 60 * 5
 
 // suggestion show delay, in milliseconds
 export const suggestionShowDelay = 250
+
+// add 200ms more delay on top of inline default 30-50ms
+export const inlineSuggestionShowDelay = 200
 
 export const referenceLog = 'CodeWhisperer Reference Log'
 
--- a/src/codewhisperer/service/inlineCompletionService.ts
+++ b/src/codewhisperer/service/inlineCompletionService.ts
@@ -228,7 +228,7 @@ export class InlineCompletionService {
         this.next = nextCommand.register(this)
         this.hide = hideCommand.register(this)
         RecommendationHandler.instance.onDidReceiveRecommendation(e => {
-            this.sharedStartShowRecommendationTimer()
+            this.startShowRecommendationTimer()
         })
     }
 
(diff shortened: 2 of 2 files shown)


leigaol:
Use toolkit provided lock wrapper function to lock the tryShowRecommendation function:
`
const cb = shared(this.tryShowRecommendation.bind(this))`

zixlin7:
> Use toolkit provided lock wrapper function to lock the tryShowRecommendation function: ` const cb = shared(this.tryShowRecommendation.bind(this))` use this cb to replace tryShowRecommendation
> 
> You can follow the similar implementation in JB. Do a while loop Before call tryShowRecommendation, poll each 25ms and see if delay is met. Be aware of concurrency issue.

updated to use `shared`, thanks

leigaol (approved):

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -240,6 +242,29 @@ export class InlineCompletionService {
         return this.documentUri?.fsPath
     }
 
+    private sharedStartShowRecommendationTimer = shared(this.startShowRecommendationTimer.bind(this))
+
+    private async startShowRecommendationTimer() {
```
JadenSimon, line 247: This implementation is already atomic due to the use of `_showRecommendationTimer`, no need to use `shared`:
```suggestion
    private startShowRecommendationTimer() {
```

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -240,6 +242,29 @@ export class InlineCompletionService {
...
+            clearInterval(this._showRecommendationTimer)
+            this._showRecommendationTimer = undefined
+        }
+        this._showRecommendationTimer = setInterval(() => {
+            const delay = performance.now() - vsCodeState.lastUserModificationTime
+            if (delay < CodeWhispererConstants.inlineSuggestionShowDelay) {
+                return
+            }
+            try {
+                this.tryShowRecommendation()
```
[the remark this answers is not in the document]
zixlin7: updated to wrap `tryShowRecommendation` instead

JadenSimon (approved):

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -240,6 +242,29 @@ export class InlineCompletionService {
         return this.documentUri?.fsPath
     }
 
+    private sharedTryShowRecommendation = shared(this.tryShowRecommendation.bind(this))
+
+    private startShowRecommendationTimer() {
```
justinmk3, line 247: This is a non-trivial pattern that is copy-pasted from other parts of the code. It might make sense to have a single shared implementation instead. (Could be done in a separate PR.)

src/codewhisperer/models/constants.ts
```diff
@@ -153,6 +155,9 @@ export const defaultCheckPeriodMillis = 1000 * 60 * 5
 // suggestion show delay, in milliseconds
 export const suggestionShowDelay = 250
 
+// add 200ms more delay on top of inline default 30-50ms
+export const inlineSuggestionShowDelay = 200
```
justinmk3, line 159: Why are the new constants needed? They are similar to existing constants. Over-granular configuration like this can be hard to work with.

justinmk3 (approved):

src/codewhisperer/service/inlineCompletionService.ts
zixlin7: yes it make sense for this to be extract into a shared implementation. Will do that in another PR

src/codewhisperer/models/constants.ts
zixlin7: the value is slightly different from the existing one used in the old code path without VSC inline api. This is trying to match the actual delay time to the old path, as well as JB as close as possible.

PR #3130, opened 2023-01-27
fix(cwl): loading new log stream stuck in loop

nkomonen-amazon:
# Problem

When using the load new events codelens on a non-filtered
log stream, we would get stuck in an endless loading loop.

When loading a non-filtered logstream, explicitly using
the GetLogEvents api, we assumed that there were
no more results if the next token was undefined.

The problem with this is that the GetLogEvents
api returns the SAME token if there are no more
results. This mistake was made since the
FilterLogEvents api does return undefined if there are no more results.

# Solution

Handle the scenario where the same token is returned and convert the result of the function that uses it in to undefined. We need to do this since the rest of the code expects the token to be undefined if
there are no more results.

# Steps

On a Log Group in the explorer, choose 'View Log Stream' and select a random one. Then click the load new log stream codelens. It will now be able to complete, while before it was stuck in an endless loop.

# Additional

A CancellableMessageManager was added, this allows for easy management of
the cancellable window for scenarios where duplicate windows may be called.
See the class documenation for more information.

## License

### DIFF
--- a/src/cloudWatchLogs/registry/logDataRegistry.ts
+++ b/src/cloudWatchLogs/registry/logDataRegistry.ts
@@ -74,6 +74,21 @@ export class LogDataRegistry {
         }
 
         const isHead = headOrTail === 'head'
+
+        // We are at the earliest data and trying to go back in time, there is nothing to see.
+        // If we don't return now, redundant data will be retrieved.
+        if (isHead && logData.previous?.token === undefined) {
+            const msgId = uriToKey(uri)
+            // show something so the user doesn't think nothing happened.
+            await Message.putMessage(
+                msgId,
+                `Loading data from log group: '${logData.logGroupInfo.groupName}'`,
+                undefined,
+                500
+            )
+            return []
+        }
+
         const stream = pageableToCollection(
             (r: typeof request) =>
                 logData.retrieveLogsFunction(
@@ -230,7 +245,7 @@ export async function filterLogEventsFromUri(
     })
 
     const responsePromise = client.filterLogEvents(cwlParameters)
-    const response = await waitTimeout(responsePromise, msgTimeout, { allowUndefined: false, completeTimeout: false })
+    const response = await waitTimeout(responsePromise, msgTimeout, { allowUndefined: false })
 
     // Use heuristic of last token as backward token and next token as forward to generalize token form.
     // Note that this may become inconsistent if the contents of the calls are changing as they are being made.
--- a/src/cloudWatchLogs/registry/logDataRegistry.ts
+++ b/src/cloudWatchLogs/registry/logDataRegistry.ts
@@ -199,12 +199,13 @@ export class LogDataRegistry {
 }
 
 /**
- * @see getLogEventsFromUriComponents
+ * @param completeTimeout True to close the vscode cancel window when request is completed.
  */
 export async function filterLogEventsFromUri(
     logGroupInfo: CloudWatchLogsGroupInfo,
     parameters: CloudWatchLogsParameters,
-    nextToken?: string
+    nextToken?: string,
+    completeTimeout = false
 ): Promise<CloudWatchLogsResponse> {
     const client = new DefaultCloudWatchLogsClient(logGroupInfo.regionName)
 
@@ -245,7 +246,7 @@ export async function filterLogEventsFromUri(
     })
 
     const responsePromise = client.filterLogEvents(cwlParameters)
-    const response = await waitTimeout(responsePromise, msgTimeout, { allowUndefined: false })
+    const response = await waitTimeout(responsePromise, msgTimeout, { allowUndefined: false, completeTimeout })
 
     // Use heuristic of last token as backward token and next token as forward to generalize token form.
     // Note that this may become inconsistent if the contents of the calls are changing as they are being made.
--- a/src/cloudWatchLogs/registry/logDataRegistry.ts
+++ b/src/cloudWatchLogs/registry/logDataRegistry.ts
@@ -193,6 +193,10 @@ export async function filterLogEventsFromUri(
         limit: parameters.limit,
     }
 
+    if (logGroupInfo.streamName !== undefined) {
+        cwlParameters.logStreamNames = [logGroupInfo.streamName]
+    }
+
     if (parameters.startTime && parameters.endTime) {
         cwlParameters.startTime = parameters.startTime
         cwlParameters.endTime = parameters.endTime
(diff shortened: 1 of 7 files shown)


src/cloudWatchLogs/registry/logDataRegistry.ts
```diff
@@ -364,10 +364,30 @@ export async function getLogEventsFromUriComponents(
...
+    /**
+     * The getLogEvents() api returns the same token from the request
+     * if there are no more results, but the rest of the code expects
+     * it to be undefined.
+     *
+     * The following conditional sets the token to undefined
+     * if the response token equals the requested.
+     */
+    if (nextToken) {
+        const isForward = nextToken.startsWith('f')
```
JadenSimon, line 379: Seems a little fragile. Pagination tokens should be treated as opaque strings. We might need to pass in another param to know which direction the token is associated with, or possible put this logic somewhere else that does know the direction.

src/cloudWatchLogs/registry/logDataRegistry.ts
```diff
@@ -336,41 +364,6 @@ export async function filterLogEventsFromUriComponents(
     }
 }
 
-export async function getLogEventsFromUriComponents(
```
nkomonen-amazon, line 339: We can remove this whole function in favor of `filterLogEventsFromUriComponents` since they do the same thing, except one filters. So in the scenario we want everything we just dont pass in a filter.

src/cloudWatchLogs/registry/logDataRegistry.ts
nkomonen-amazon: We were able to drop this whole function completely, so this code has been removed. I explain why we were able to drop it in another comment. Thanks!

src/shared/utilities/messages.ts
```diff
@@ -182,11 +183,114 @@ export async function showMessageWithCancel(
...
+
+        if (message === undefined) {
+            return
+        }
+
+        message.timeout.cancel()
+        delete this.messageMap[msgId]
+    }
+
+    static getTimeout(msgId: CancellableMessageId): Timeout {
```
JadenSimon, line 285: This is a leaky abstraction. A caller can cancel the timer (and therefore the message) without calling `cancelMessage`.

src/cloudWatchLogs/registry/logDataRegistry.ts
```diff
@@ -152,7 +166,13 @@ export class LogDataRegistry {
             return last
         }
 
+        const msgId = CancellableMessageManager.idFromCwlInfo(logData.logGroupInfo, logData.parameters)
+        await CancellableMessageManager.startMessage(
+            msgId,
+            `Loading data from log group: '${logData.logGroupInfo.groupName}'`
+        )
         const responseData = await firstOrLast(stream, resp => resp.events.length > 0)
+        CancellableMessageManager.cancelMessage(msgId)
```
JadenSimon, line 175: Need to clean
```suggestion
        const responseData = await firstOrLast(stream, resp => resp.events.length > 0).finally(() => {
            CancellableMessageManager.cancelMessage(msgId)
        })
```

src/shared/utilities/messages.ts
```diff
@@ -182,11 +183,114 @@ export async function showMessageWithCancel(
...
+): Promise<vscode.Progress<ProgressEntry>> {
     const progressOptions = { location: vscode.ProgressLocation.Notification, title: message, cancellable: true }
     return showProgressWithTimeout(progressOptions, timeout, window)
 }
 
+/**
+ * The message ID that {@link CancellableMessageManager} uses.
+ * It uniquely identifies a specific message.
+ */
+export type CancellableMessageId = string
```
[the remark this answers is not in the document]
nkomonen-amazon: The goal of this was to improve readability so that using this type in function args or return types was obvious to the user if it was the message id vs. another string.

src/shared/utilities/messages.ts
nkomonen-amazon: Nice catch. I've created a new class `CancellableMessageTimeout` that extends `Timeout`. It will cancel the underlying `Timeout`, but then perform the necessary cleanup that `cancelMessage()` does.

src/shared/utilities/messages.ts
```diff
@@ -182,11 +183,114 @@ export async function showMessageWithCancel(
...
+
+    /**
+     * Helper to generate unique ID from the given info
+     */
+    static idFromCwlInfo(
+        cwlGroup: CloudWatchLogsGroupInfo,
+        cwlParameters: CloudWatchLogsParameters
+    ): CancellableMessageId {
+        return `${getStringHash(JSON.stringify({ cwlGroup, cwlParameters }))}`
+    }
```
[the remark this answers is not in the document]
nkomonen-amazon: I think in terms of discoverability it will be difficult to find. What about a separate class in `messages.ts` named `CancellableMessageIdBuilder` that holds separate implementations for different modules.

```typescript
class CancellableMessageBuilder {
  static fromCwlInfo(
          cwlGroup: CloudWatchLogsGroupInfo,
          cwlParameters: CloudWatchLogsParameters
    ): CancellableMessageId {
        return `${getStringHash(JSON.stringify({ cwlGroup, cwlParameters }))}`
    }
}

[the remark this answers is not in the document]
Comment (JadenSimon) [reply src/shared/utilities/messages.ts:241]: But why does the id have to be specific to cancellable messages? The identifier is provided by the _caller_ and is completely unconstrained. Let's stay focused on the purpose of the string and not how it's created. In this case, isn't our goal to limit cancel notifications to a single URI? So why not use `uriToKey` in `src/cloudWatchLogs/cloudWatchLogsUtils.ts`? Needing a unique key for something is not an uncommon problem, and we try to re-use existing formats when it makes sense to do so.

File: src/shared/utilities/messages.ts (hunk 5 above)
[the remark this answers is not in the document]
Comment (JadenSimon) [reply src/shared/utilities/messages.ts:195]: > The goal of this was to improve readability so that using this type in function args or return types was obvious to the user if it was the message id vs. another string.

I understand the intent, I just personally don't find it useful because the string is crafted by the caller. So the intent is misleading if that makes sense? In any case this is a nitpick so if it makes sense to you it's fine with me.

File: src/shared/utilities/messages.ts (hunk 6 above)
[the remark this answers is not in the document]
Comment (nkomonen-amazon) [reply src/shared/utilities/messages.ts:241]: Thanks for pointing out the utils. I've updated to use `uriToKey()` when we have the uri and `uriToKey(createURIFromArgs(logGroupInfo, parameters))` for another section.

File: src/shared/utilities/messages.ts (hunk 5 above)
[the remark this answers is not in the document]
Comment (nkomonen-amazon) [reply src/shared/utilities/messages.ts:195]: After dropping the `idFromCwlInfo()` method in favor of using the util methods, I've dropped this since I don't see it as necessary anymore.

File: src/cloudWatchLogs/registry/logDataRegistry.ts (hunk 2 above)
Comment (justinmk3) [reply src/cloudWatchLogs/registry/logDataRegistry.ts:232]: nice! I wondered about that...

File: src/shared/utilities/messages.ts [hunk 7]
```diff
@@ -182,11 +182,134 @@ export async function showMessageWithCancel(
...
+ *
+ * To get started, you create a new message with a unique ID.
+ * Following that, you can easily update the message or cancel it.
+ *
+ * ---
+ *
+ * This class solves an issue of needing to manage duplicated
+ * messages, the interface simplifies the process.
+ */
+export class CancellableMessageManager {
```
justinmk3, line 222: Can avoid "premature explanation" here.

```suggestion
export class MessageManager {
```

src/shared/utilities/messages.ts
```diff
@@ -182,11 +182,134 @@ export async function showMessageWithCancel(
...
+        message.progress.report(progressUpdate)
+    }
+
+    /** Cancels the message and clears it from the manager. */
+    static cancelMessage(msgId: string): void {
+        this.stop(msgId, 'cancel')
+    }
+
+    /** Disposes the message and clears it from the manager. */
+    static disposeMessage(msgId: string): void {
```
justinmk3, line 290: could eliminate this, cancelMessage, and getTimeout by instead exposing `getMessage`.

src/shared/utilities/messages.ts
```diff
@@ -182,11 +182,134 @@ export async function showMessageWithCancel(
...
+        if (message === undefined) {
+            throw Error(`Cannot get timeout for '${msgId}', it does not exist.`)
+        }
+        return message.timeout
+    }
+
+    /**
+     * Updates an EXISTING message
+     */
+    static updateMessage(msgId: string, progressUpdate: ProgressEntry): void {
```
justinmk3, line 274: not used. could be eliminated by adding getMessage().

justinmk3 (approved):

justinmk3 (approved):

src/shared/utilities/messages.ts
```diff
@@ -213,6 +214,70 @@ export async function showMessageWithCancel(
...
+/**
+ * Start or Update VSCode message windows with a 'Cancel' button.
+ *
+ * ---
+ *
+ * This class helps in the scenario of ensuring that a previous
+ * identical message is completed before it attempts to create
+ * a new one.
+ */
+export class Message {
```
justinmk3, line 228: `Messages`

PR #3131, opened 2023-01-28
add telemetry: sam cli not installed

gsweetwood:
## Problem
 We have a number of commands that require the SAM CLI. We are not tracking how often users fail these commands because they do not have SAM CLI installed.

## Solution
 Add telemetry to track when this error is invoked.

## License


JadenSimon (approved):
Note that this PR doesn't cover the upload Lambda + SAM build use-case due to implementation discrepancies.

PR #3140, opened 2023-02-01
samcli: add local autoinstall for linux/windows

gsweetwood:
## Problem
 Many users don't have SAM CLI and/or don't want to follow the several steps required to install it.

## Solution
If the SAM CLI is not found, install it locally. Currently scoped to just Linux users.

## License


src/shared/utilities/cliUtils.ts
```diff
@@ -314,6 +335,53 @@ async function installSsmCli(
...
+        }
+        const dirname = path.dirname(samInstaller)
+        new admZip(samInstaller).extractAllTo(dirname, true)
+
+        const result = await new ChildProcess('sh', [
+            path.join(dirname, 'install'),
+            '-u',
+            '-i',
+            outDir,
+            '-b',
```
justinmk3, line 372: short args like `-i` and `-b` are convenience aliaes for interactive usage. For non-interactive usage, it's usually helpful to use the long names (except perhaps for some very well-known commands).

src/test/shared/sam/cli/samCliConfiguration.test.ts
```diff
@@ -74,6 +75,9 @@ describe('samCliConfiguration', function () {
             settingsConfiguration
         )
 
+        if (os.platform() === 'linux') {
+            this.skip()
```
justinmk3, line 79: this test worked before this PR. why is it skipped always on linux now?

justinmk3 (commented):
Next step is to look for a way to get this working on Windows, perhaps by installing to a higher-level directory (e.g. ~/.aws/toolkit/)

gsweetwood:
Added a small refactor to the `samCliLocator` to search locally if not found globally. (reversed for C9)

src/test/shared/sam/cli/samCliConfiguration.test.ts
gsweetwood: Now that I added the option to install or not if the CLI is not detected, we can run this test on all platforms. This test was assuring no cli path was found, but it was autoinstalling on Linux, so it failed.

gsweetwood:
Added the installation logic for Windows. Needed to use `~/.aws/toolkits` to avoid the path length error. I did try the shorter `/sam` instead of `AWSSAMCLI/bin/sam` but it wasn't enough. Symlinks DO NOT work for executables in Windows, so the cli path needs to be the same location.

PR #3144, opened 2023-02-02
s3: add presigned URL for uploading

gsweetwood:
## Problem
Users can generate a GET presigned URL but not for uploading.
## Solution
Refactor the PresignedUrlWizard and add the ability to generate a presigned PUT URL.

## License

### DIFF
--- a/src/s3/commands/presignedURL.ts
+++ b/src/s3/commands/presignedURL.ts
@@ -3,7 +3,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { SignedUrlRequest } from '../../shared/clients/s3Client'
+import * as vscode from 'vscode'
+
+import { DefaultS3Client, SignedUrlRequest } from '../../shared/clients/s3Client'
 import { Env } from '../../shared/vscode/env'
 import { copyToClipboard } from '../../shared/utilities/messages'
 import { S3FileNode } from '../explorer/s3FileNode'
@@ -12,23 +14,29 @@ import { localize } from '../../shared/utilities/vsCodeUtils'
 import { invalidNumberWarning } from '../../shared/localizedText'
 import { telemetry } from '../../shared/telemetry/telemetry'
 import { ToolkitError } from '../../shared/errors'
+import { Wizard } from '../../shared/wizards/wizard'
+import { createQuickPick, DataQuickPickItem } from '../../shared/ui/pickerPrompter'
+import { createCommonButtons } from '../../shared/ui/buttons'
+import { createInputBox } from '../../shared/ui/inputPrompter'
+import { createRegionPrompter } from '../../shared/ui/common/region'
+import { getLogger } from '../../shared/logger'
 
 export async function presignedURLCommand(
-    node: S3FileNode,
+    node?: S3FileNode,
     window = Window.vscode(),
     env = Env.vscode()
 ): Promise<void> {
     await telemetry.s3_copyUrl.run(async span => {
         span.record({ presigned: true })
 
-        const validTime = await promptTime(node.file.key, window)
-        const s3Client = node.s3
-        const request: SignedUrlRequest = {
-            bucketName: node.bucket.name,
-            key: node.file.key,
-            time: validTime * 60,
-            operation: 'getObject',
+        const response = await new PresignedUrlWizard(node).run()
+        if (!response) {
+            getLogger().debug('s3: PresignedUrlWizard returned undefined. User cancelled.')
+            return
         }
+
+        const request = response.signedUrlParams
+        const s3Client = node ? node.s3 : new DefaultS3Client(response.region)
 
         const url = await s3Client.getSignedUrl(request).catch(e => {
             throw ToolkitError.chain(
@@ -38,6 +46,7 @@ export async function presignedURLCommand(
         })
 
         await copyToClipboard(url, 'URL', window, env)
+        vscode.window.showInformationMessage(`A presigned URL has been copied to your clipboard`)
     })
 }
 
@@ -70,4 +79,105 @@ function validateTime(time: string): string | undefined {
     }
 
     return undefined
+}
+
+export interface PresignedUrlWizardState {
+    region: string
+    signedUrlParams: SignedUrlRequest
+}
+
+export class PresignedUrlWizard extends Wizard<PresignedUrlWizardState> {
+    constructor(node?: S3FileNode) {
+        super()
+        // region, for CP
+        if (node) {
+            this.form.region.setDefault(node.bucket.region)
+        } else {
+            this.form.region.bindPrompter(() => createRegionPrompter().transform(region => region.id))
+        }
+        // GET or PUT
+        this.form.signedUrlParams.operation.bindPrompter(() => createOperationPrompter())
+        // CP path, bucket next
+        if (node) {
+            this.form.signedUrlParams.bucketName.setDefault(node.bucket.name)
+        } else {
+            // ????????????? How should I do this ??  region is string | undefined for some reason
+            this.form.signedUrlParams.bucketName.bindPrompter(({ region }) => createBucketPrompter(region))
+        }
+
+        // CP path, GET needs a file
+        if (node) {
+            this.form.signedUrlParams.key.setDefault(node.file.key)
+        } else {
+            this.form.signedUrlParams.key.setDefault(Math.ceil(Math.random() * 10 ** 10).toString())
+            this.form.signedUrlParams.key.bindPrompter(
+                ({ region, signedUrlParams }) => createS3FilePrompter(region, signedUrlParams?.bucketName),
+                { showWhen: state => state.signedUrlParams?.operation === 'getObject' }
+            )
+        }
+
+        this.form.signedUrlParams.time.bindPrompter(() => createExpiryPrompter().transform(s => Number(s) * 60))
+    }
+}
+
+function createOperationPrompter() {
+    const items: DataQuickPickItem<string>[] = [
+        { label: 'Download (GET)', data: 'getObject' },
+        { label: 'Upload (PUT)', data: 'putObject' },
+    ]
+
+    return createQuickPick(items, { title: 'Presigned URL: Choose an operation', buttons: createCommonButtons() })
+}
+
+function createExpiryPrompter() {
+    return createInputBox({
+        value: '15',
+        prompt: 'Specify the expiry time (minutes) for the URL',
+        placeholder: 'Defaults to 15 minutes',
+        validateInput: validateTime,
+        buttons: createCommonButtons(),
+    })
+}
+
+function createBucketPrompter(region: string | undefined) {
+    // THIS IS JUST TO AVOID ERRORS, FIND OUT HOW TO GET REGION FROM PREVIOUS RESPONSE
+    if (!region) {
+        region = 'us-west-2'
+    }
+    const client = new DefaultS3Client(region)
+    const items = client.listBucketsIterable().map(b => [
+        {
+            label: b.Name,
+            data: b.Name,
+        },
+    ])
+
+    return createQuickPick(items, { title: 'Select an S3 Bucket', buttons: createCommonButtons() })
+}
+
+function createS3FilePrompter(region: string | undefined, bucket: string | undefined) {
+    // hack until I can guarantee arguments
+    if (!region) {
+        region = 'us-west-2'
+    }
+    if (!bucket) {
+        bucket = 'gsweet-temp'
+    }
+    const items = getS3Files(region, bucket)
+    return createQuickPick(items, {
+        title: 'Choose the file for the presigned URL',
+        buttons: createCommonButtons(),
+    })
+}
+
+async function getS3Files(region: string, bucket: string) {
+    const client = new DefaultS3Client(region)
+    const files = (await client.listFiles({ bucketName: bucket })).files
+    const items: DataQuickPickItem<string>[] = files.map(f => {
+        return {
+            label: f.name,
+            data: f.key,
+        }
+    })
+    return items
 }


src/test/lambda/wizards/presignedUrlWizard.test.ts
```diff
@@ -0,0 +1,55 @@
+/*!
```
justinmk3, line 1: why does this file live in `src/test/lambda/wizards/` but it's testing `src/s3/` ? should this test file live in `src/test/s3` ?

src/s3/commands/presignedURL.ts
```diff
@@ -3,33 +3,40 @@
...
-        const validTime = await promptTime(node.file.key, window)
-        const s3Client = node.s3
-        const request: SignedUrlRequest = {
-            bucketName: node.bucket.name,
-            key: node.file.key,
-            time: validTime * 60,
-            operation: 'getObject',
+        const response = await new PresignedUrlWizard(node).run()
+        if (!response) {
+            getLogger().debug('s3: PresignedUrlWizard returned undefined. User cancelled.')
```
justinmk3, line 33: it's good to log stuff like this, but I think the wizard framework already do so. Can you confirm?

src/s3/commands/presignedURL.ts
```diff
@@ -38,36 +45,124 @@ export async function presignedURLCommand(
         })
 
         await copyToClipboard(url, 'URL', window, env)
+        vscode.window.showInformationMessage(`A presigned URL has been copied to your clipboard`)
```
justinmk3, line 48: If we are concerned that the statusbar message isn't visible enough, let's address that in `copyToClipboard()` rather than special-case the Preseigned URL feature. And then we would probably want a "Don't show again" button if we go in that direction.

To keep this PR in scope, consider creating a backlog item for that and drop this info message for now.

src/s3/commands/presignedURL.ts
JadenSimon: The wizard won't log but the top-level command will if it catches certain exceptions. `CancellationError` didn't exist when the wizard code was written but it should probably be updated to throw instead of returning `undefined`.

src/s3/commands/presignedURL.ts
```diff
@@ -3,33 +3,40 @@
...
-        const s3Client = node.s3
-        const request: SignedUrlRequest = {
-            bucketName: node.bucket.name,
-            key: node.file.key,
-            time: validTime * 60,
-            operation: 'getObject',
+        const response = await new PresignedUrlWizard(node).run()
+        if (!response) {
+            getLogger().debug('s3: PresignedUrlWizard returned undefined. User cancelled.')
+            return
```
JadenSimon, line 34:
```suggestion
            throw new CancellationError('user')
```

src/test/lambda/wizards/presignedUrlWizard.test.ts
```diff
@@ -0,0 +1,55 @@
...
+        beforeEach(function () {
+            tester = createWizardTester(
+                new PresignedUrlWizard({
+                    bucket: { name: 'mr bucket', region: 'outer space' },
+                    file: { key: 'key' },
+                } as S3FileNode)
+            )
+        })
+
+        it('GET url - show operation and time', function () {
```
JadenSimon, line 37: nit: test titles should be written with the expected behavior/result at the start. Either place qualifiers at the end of the title or create a separate describe block.

src/s3/commands/presignedURL.ts
```diff
@@ -38,36 +45,124 @@ export async function presignedURLCommand(
...
+        })
+        this.form.region.bindPrompter(() => createRegionPrompter().transform(region => region.id))
+        this.form.signedUrlParams.operation.bindPrompter(() => createOperationPrompter())
+        this.form.signedUrlParams.bucketName.bindPrompter(({ region }) =>
+            createBucketPrompter(assertDefined(region, 'region'))
+        )
+
+        if (node) {
+            this.form.signedUrlParams.key.setDefault(node.file.key)
+            this.form.signedUrlParams.key.bindPrompter(
```
justinmk3, line 83: almost identical to the `else` case below. can we avoid the redundant code?

src/test/lambda/wizards/presignedUrlWizard.test.ts
```diff
@@ -0,0 +1,55 @@
...
+            tester.region.assertDoesNotShow()
+            tester.signedUrlParams.bucketName.assertDoesNotShow()
+            tester.signedUrlParams.key.assertDoesNotShow()
+
```
JadenSimon, line 41: This could be its own test, something like "skips key prompt and uses node's file key if available":
```suggestion
            tester.signedUrlParams.key.assertDoesNotShow()
            tester.signedUrlParams.key.assertValue('key')
```

src/s3/commands/presignedURL.ts
```diff
@@ -38,36 +45,124 @@ export async function presignedURLCommand(
...
+                    assertDefined(signedUrlParams?.operation, 'operation')
+                )
+            )
+        }
 
-    const time = Number(timeStr)
+        this.form.signedUrlParams.time.bindPrompter(() => createExpiryPrompter().transform(s => Number(s) * 60))
 
-    if (isNaN(time) || time < 0) {
-        throw new ToolkitError('The provided expiration time is not a positive number', { code: 'InvalidInput' })
```
justinmk3, line 59: old code raised `ToolkitError... code:InvalidInput`. What does the new code raise? Can this affect telemetry and possibly control flow ?

src/s3/commands/presignedURL.ts
```diff
@@ -38,36 +45,124 @@ export async function presignedURLCommand(
...
+            this.form.signedUrlParams.key.setDefault(node.file.key)
+            this.form.signedUrlParams.key.bindPrompter(
+                ({ region, signedUrlParams }) =>
+                    createS3FilePrompter(
+                        assertDefined(region, 'region'),
+                        assertDefined(signedUrlParams?.bucketName, 'bucketName'),
+                        assertDefined(signedUrlParams?.operation, 'operation')
+                    ),
+                { showWhen: state => state.signedUrlParams?.operation === 'putObject' }
+            )
```
JadenSimon, line 91: Not sure if this is the behavior we want. If the user explicitly selects a file node I would assume that they want a pre-signed URL for _that_ destination right?

src/s3/commands/presignedURL.ts
```diff
@@ -38,36 +45,124 @@ export async function presignedURLCommand(
...
+    region: string
+    signedUrlParams: SignedUrlRequest
+}
+
+export class PresignedUrlWizard extends Wizard<PresignedUrlWizardState> {
+    constructor(node?: S3FileNode) {
+        super({
+            initState: {
+                region: node?.bucket.region,
+                signedUrlParams: node ? ({ bucketName: node.bucket.name } as SignedUrlRequest) : undefined,
```
justinmk3, line 72: somewhere, perhaps in a `/** ... */` comment on this param, we need to explain what's going on. It's not obvious from any of the function names or anything AFAICT, that this wizard has two different behaviors that are decided by  (what?)

In the future, try to anticipate the perspective of other developers, this kind of question they will ask as they read the code.

src/s3/commands/presignedURL.ts
```diff
@@ -38,36 +45,124 @@ export async function presignedURLCommand(
...
+function createBucketPrompter(region: string) {
+    const client = new DefaultS3Client(region)
+    const items = client.listBucketsIterable().map(b => [
+        {
+            label: b.Name,
+            data: b.Name,
+        },
+    ])
+
+    return createQuickPick(items, { title: 'Select an S3 Bucket', buttons: createCommonButtons() })
```
justinmk3, line 141: should default to recently-used bucket. some sample code is here: https://github.com/aws/aws-toolkit-vscode/blob/e5f813fb84c7f0de5706d72acce8f1e982767588/src/shared/sam/sync.ts#L57

src/test/lambda/wizards/presignedUrlWizard.test.ts
justinmk3: Would like to encourage flat list of it() blocks where we can, rather than many nested `describe..describe..it` blocks.

src/s3/commands/presignedURL.ts
gsweetwood: Backlog item created, notification removed.

src/s3/commands/presignedURL.ts
gsweetwood: Nope, there is a validator on this prompt called `validateTime`. The user cannot progress past this prompt unless they put in a valid positive number.

src/s3/commands/presignedURL.ts
gsweetwood: For GET that is the case, but for PUT I don't know if that's what we want. It will replace that object in their bucket. They may just want to insert a new object. Asking for the key is safer in my opinion.

src/s3/commands/presignedURL.ts
gsweetwood: Finally figured out a good way to do it, so yes de-duped.

src/s3/commands/presignedURL.ts
gsweetwood: added a comment

src/test/s3/presignedUrlWizard.test.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+import { createWizardTester, WizardTester } from '../shared/wizards/wizardTestUtils'
+import { PresignedUrlWizard, PresignedUrlWizardState } from '../../s3/commands/presignedURL'
+import { S3FileNode } from '../../s3/explorer/s3FileNode'
+
+describe('PresignedUrlWizard', function () {
+    let tester: WizardTester<PresignedUrlWizardState>
+
+    describe('invoked from command palette', function () {
```
justinmk3, line 13: is this describe() needed (or accurate)? some of the it() tests are "from node" instead of "from command palette". Looks like we don't need this extra describe() scope

src/s3/commands/presignedURL.ts
```diff
@@ -41,33 +47,115 @@ export async function presignedURLCommand(
     })
 }
 
-export async function promptTime(fileName: string, window = Window.vscode()): Promise<number> {
-    const timeStr = await window.showInputBox({
-        value: '15',
-        prompt: localize(
-            'AWS.s3.promptTime.prompt',
-            'Specify the time (minutes) until URL will expire for path: {0}',
```
justinmk3, line 49: was this message dropped intentionally? It shows the path in the prompt:

image

src/s3/commands/presignedURL.ts
```diff
@@ -41,33 +47,115 @@ export async function presignedURLCommand(
...
-    return time
+    return createQuickPick(items, { title: 'Presigned URL: Choose an operation', buttons: createCommonButtons() })
 }
 
-function validateTime(time: string): string | undefined {
-    const number = Number(time)
+function createExpiryPrompter() {
+    return createInputBox({
+        value: '15',
+        prompt: 'Specify the expiry time (minutes) for the URL',
```
justinmk3, line 118: Specify the time (minutes) until URL will expire for path: {0}

src/s3/commands/presignedURL.ts
```diff
@@ -41,33 +47,115 @@ export async function presignedURLCommand(
...
+    if (operation === 'getObject') {
+        const items = getS3Files(region, bucket)
+        return createQuickPick(items, {
+            title: 'Choose the file for the presigned URL',
+            buttons: createCommonButtons(),
+        })
     }
+    return createInputBox({
+        title: 'Create a key',
+        buttons: createCommonButtons(),
```
justinmk3, line 147: "Create a key" is a bit obscure. This might be more intuitive (also uses the same "Specify" verb that was used for the "Specify the time..." message; using a verb like "Create" implies different semantics):

    Specify the key (S3 file path) where the upload will be saved

Also let's link to https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html from the help button in this step.

src/s3/commands/presignedURL.ts
```diff
@@ -41,33 +47,115 @@ export async function presignedURLCommand(
...
+
+    return createQuickPick(items, { title: 'Select an S3 Bucket', buttons: createCommonButtons() })
+}
+
+function createS3FilePrompter(region: string, bucket: string, operation: string) {
+    if (operation === 'getObject') {
+        const items = getS3Files(region, bucket)
+        return createQuickPick(items, {
+            title: 'Choose the file for the presigned URL',
+            buttons: createCommonButtons(),
```
justinmk3, line 142: help button should link to https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html 

later we can update to Toolkit-specific docs.

src/s3/commands/presignedURL.ts
```diff
@@ -41,33 +47,115 @@ export async function presignedURLCommand(
...
 
-function validateTime(time: string): string | undefined {
-    const number = Number(time)
+function createExpiryPrompter() {
+    return createInputBox({
+        value: '15',
+        prompt: 'Specify the expiry time (minutes) for the URL',
+        placeholder: 'Defaults to 15 minutes',
+        validateInput: validateTime,
+        buttons: createCommonButtons(),
```
justinmk3, line 121: help button should link to https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html 

later we can update to Toolkit-specific docs.

justinmk3:
What's an example of a `curl` (or sdk call) that successfully uploads to the presigned PUT url? I tried this:

    curl -d "@README" -X POST 'https://xxx.s3.us-west-2.amazonaws.com/...'

but it returned an error:

    <Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>

src/test/s3/presignedUrlWizard.test.ts
gsweetwood: Unintentionally left this in, it's been removed.

justinmk3:
re https://github.com/aws/aws-toolkit-vscode/pull/3144#discussion_r1100602147

> For GET that is the case, but for PUT I don't know if that's what we want. It will replace that object in their bucket. They may just want to insert a new object into that bucket. Asking for the key is safer in my opinion

- If they select "Generate PUT" on a folder then ask for a key.
- If they select "Geerate PUT" on an object then this clearly indicates to replace that object and the wizard shouldn't ask for a key.

gsweetwood:
Originally this command could only be initiated from an `S3FileNode` or the command palette. For best user experience, I added the ability to use the command from `S3BucketNode` and `S3FolderNode`.  The wizard will only show files that are direct children of the bucket/folder. The `key` is automatically assigned if called from a `S3FileNode`

src/s3/commands/presignedURL.ts
gsweetwood: This will be implemented in an upcoming change.

src/shared/clients/s3Client.ts
```diff
@@ -287,6 +287,7 @@ export class DefaultS3Client {
             Key: request.key,
             Body: request.body,
             Expires: time,
+            ContentType: 'application/octet-stream',
```
justinmk3, line 290: Are you sure this is wanted for the "GET" case? 
Should also call this out in the docstring.

gsweetwood: Updated to only set for the upload case.

src/s3/commands/presignedURL.ts
```diff
@@ -37,33 +64,135 @@ export async function presignedURLCommand(node: S3FileNode, env = Env.vscode()):
...
+        buttons: createCommonButtons(s3PresigndUrlHelpUrl),
+    })
+}
+
+function createExpiryPrompter(path: string) {
+    return createInputBox({
         value: '15',
         prompt: localize(
-            'AWS.s3.promptTime.prompt',
+            'AWS.s3.presignedURL.epiryPrompt',
```
rli, line 152:
```suggestion
            'AWS.s3.presignedURL.expiryPrompt',
```

PR #3149, opened 2023-02-06
CWL Log Registry Refactor

nkomonen-amazon:
## Problem

The CWL Log Registry coupled the management of the log events, generation of the text content, the stream ID mapping to the text content, and TextDocument Editor management all in to one class.

## Solution

Separate the components which will allow us to more easily work with these functionalities.

## License


justinmk3:
looks like ci is working now

```
  1) saveCurrentLogDataContent
       saves log content to a file:

+ actual - expected

+ '1970-01-20T09:24:11.113+00:00\tThe first log event\n' +
+   '1970-01-20T09:24:11.114+00:00\tThe second log event\n'
- '1970-01-20T04:24:11.113-05:00\tThe first log event\n' +
-   '1970-01-20T04:24:11.114-05:00\tThe second log event\n'

File: src/cloudWatchLogs/document/logDataDocumentProvider.ts [hunk 1]
```diff
@@ -29,12 +26,13 @@ export class LogDataDocumentProvider implements vscode.TextDocumentContentProvid
     }
 
     public provideTextDocumentContent(uri: vscode.Uri): string {
-        // get latest content and return
-        const content = this.registry.getLogContent(uri, { timestamps: true })
-        if (!content) {
-            getLogger().error(`No content found for URI: ${uri.path}`)
+        if (!isCwlUri(uri)) {
+            return ''
```
JadenSimon, line 30: This is a silent failure. Odds are this document provider would only be called with a non-CWL URI if there's a bug somewhere else. I would at least log something here if we don't want to raise an exception.

src/cloudWatchLogs/activation.ts
```diff
@@ -41,13 +41,6 @@ export async function activate(context: vscode.ExtensionContext, configuration:
             if (doc.isClosed && doc.uri.scheme === CLOUDWATCH_LOGS_SCHEME) {
                 registry.disposeRegistryData(doc.uri)
             }
-        }),
-        // Do highlight on text-changed event because we don't control when text is populated by vscode, also for "load more" codelens.
-        vscode.workspace.onDidChangeTextDocument((event: vscode.TextDocumentChangeEvent) => {
-            const eventUri = event.document.uri
-            if (registry.hasLog(eventUri) && !isLogStreamUri(eventUri)) {
-                registry.highlightDocument(eventUri)
```
JadenSimon, line 49: Looks like highlighting is being removed. Will this be added back?

nkomonen-amazon: When I disabled this functionality I was still receiving highlights on the new text change when I did the 'Load More' codelens. I'm not sure if something changed with VSCode when this was initially implemented??

src/cloudWatchLogs/document/logDataDocumentProvider.ts
nkomonen-amazon: An existing test required this, but I think it is better to throw an error. I've updated the test and this code to throw an error instead.

src/cloudWatchLogs/activation.ts
JadenSimon: I guess the logic was redundant.

JadenSimon (approved):

PR #3150, opened 2023-02-06
cwl: single-click on log group shows stream select

gsweetwood:
Problem: There is only a single action associated with CW log groups and it is only available through the context menu.

Solution: Execute the view log stream selector when the log group node is selected.

## Problem
Currently the toolkit requires the user to right-click and then choose `View Log Stream` to start the selection prompt. 
## Solution
Start the prompt immediately when the node is left-clicked.

## License

### DIFF
--- a/.changes/next-release/Feature-3082756c-79f7-4341-88be-d2f583a93686.json
+++ b/.changes/next-release/Feature-3082756c-79f7-4341-88be-d2f583a93686.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Reveal the log stream prompt when selecting a log group from the explorer."
+}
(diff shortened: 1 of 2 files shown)


.changes/next-release/Feature-3082756c-79f7-4341-88be-d2f583a93686.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Reveal the log stream prompt when selecting a log group from the explorer."
```
justinmk3, line 3:
```suggestion
	"description": "CloudWatch Logs: one click to list log streams for a log group"
```

PR #3151, opened 2023-02-06
CI: temporarily skip broken tests

justinmk3:
Problem:
Broken tests are blocking the pipeline.

Solution:
Skip the tests so we can work on them after upgrading the pipeline.

## License


src/integrationTest/sam.test.ts
```diff
@@ -85,15 +85,15 @@ const scenarios: TestScenario[] = [
...
-        vscodeMinimum: '1.50.0',
-    },
+    // {
+    //     runtime: 'python3.7',
+    //     displayName: 'python3.7 (ZIP)',
+    //     path: 'hello_world/app.py',
+    //     debugSessionType: 'python',
+    //     language: 'python',
+    //     dependencyManager: 'pip',
+    //     vscodeMinimum: '1.50.0',
```
justinmk3, line 95:
```
1) SAM Integration Tests
--
1308 | SAM runtime: python3.7 (ZIP)
1309 | Starting with a newly created python3.7 (ZIP) SAM Application...
1310 | produces an Add Debug Configuration codelens:
1311 | TypeError: call.getFileName is not a function
1312 | at getErrMessage (assert.js:270:25)
1313 | at innerOk (assert.js:361:17)
1314 | at Function.ok (assert.js:381:3)
1315 | at Context.<anonymous> (src/integrationTest/sam.test.ts:501:28)
```

PR #3163, opened 2023-02-13
Cfn sam schema new

nkomonen-amazon:
# Problem

For our doctype suport we use different schemas
for cfn and sam. But now there is an all-in-one
sam schema that works for cfn as well.

# Solution

Have cfn and sam use the new sam schema. This
requires us to update our url content fetching
\+ caching mechanisms to use e-tags to determine
stale content instead of using versions as we
were previously doing.

# Steps to test

CFN and SAM doctype support should remain exactly the same,
except it now uses the newest [schema from here](https://raw.githubusercontent.com/aws/serverless-application-model/main/samtranslator/schema/schema.json)

# How caching mechanism works

Right now we use an E-Tag in url requests, if it matches the remote's E-Tag then a
304 status code will be returned with no content at all. This will tell us that our
current cache is the latest.

The E-Tag is part of [this header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match)

## License


src/shared/resourcefetcher/httpResourceFetcher.ts
```diff
@@ -79,6 +82,49 @@ export class HttpResourceFetcher implements ResourceFetcher {
...
+            }
+
+            return { content: contents, eTag: eTagResponse }
+        } catch (err) {
+            const error = err as CancelError | RequestError
+            this.logger.verbose(
+                `Error downloading ${this.logText()}: %s`,
+                error.message ?? error.code ?? error.response?.statusMessage ?? error.response?.statusCode
+            )
+            return undefined
```
JadenSimon, line 124: Use `ToolkitError.chain` instead of swallowing. Let the caller decide if the exception needs to be handled.

src/shared/schemas.ts
```diff
@@ -251,6 +245,93 @@ export async function updateSchemaFromRemote(params: {
...
+        }
+
+        const { content, eTag: latestETag } = response
+        if (content === undefined) {
+            // Our cached content is the latest
+            return
+        }
+
+        params.eTag = latestETag
+        doCacheContent(content, params)
```
JadenSimon, line 296: I would log something because this isn't being awaited:
```suggestion
        doCacheContent(content, params).catch(err => getLogger().warn('schemas: failed to cache result: %s', err))
```

src/shared/schemas.ts
```diff
@@ -251,6 +245,93 @@ export async function updateSchemaFromRemote(params: {
...
+    } catch (err) {
+        if (cachedContent) {
+            getLogger().warn(
+                `schemas: failed to fetch the latest version for "${params.title}": ${
+                    (err as Error).message
+                }. Using cached schema instead.`
```
JadenSimon, line 302: nit: don't access the error message directly
```suggestion
                `schemas: using cached schema, failed to fetch the latest version for "${params.title}": %s`, e
```

src/shared/schemas.ts
nkomonen-amazon: I didn't realize I missed the await, thanks for catching that. Have we thought about enabling a lint warning when we don't use an `await`?

jpinkney-aws:
Would any of the open sam/cfn schema issues be fixed by this PR: https://github.com/aws/aws-toolkit-vscode/issues?q=is%3Aopen+is%3Aissue+label%3Ajson-schema-sam-cfn ?

src/shared/schemas.ts
JadenSimon: There's a PR out there somewhere that turns on the rule. Not using `await` is _sometimes_ intentional so adding the rule is not as straightforward as one might hope.

src/shared/schemas.ts
```diff
@@ -140,41 +139,42 @@ export class SchemaService {
...
-    const samSchemaUri = vscode.Uri.joinPath(extensionContext.globalStorageUri, 'sam.schema.json')
     const devfileSchemaUri = vscode.Uri.joinPath(extensionContext.globalStorageUri, 'devfile.schema.json')
-
-    const goformationSchemaVersion = await getPropertyFromJsonUrl(goformationManifestUrl, 'tag_name')
     const devfileSchemaVersion = await getPropertyFromJsonUrl(devfileManifestUrl, 'tag_name')
 
+    // Sam schema is a superset of Cfn schema, so we can is it for both
+    const samAndCfnSchemaDestinationUri = vscode.Uri.joinPath(extensionContext.globalStorageUri, 'sam.schema.json')
+    const samAndCfnCacheKey = 'samAndCfnSchemaVersion'
+    const samAndCfnSchemaUrl = `https://raw.githubusercontent.com/aws/serverless-application-model/main/samtranslator/schema/schema.json`
```
justinmk3, line 148: should this live next to `devfileManifestUrl` ?

src/shared/resourcefetcher/httpResourceFetcher.ts
```diff
@@ -148,6 +183,22 @@ export class HttpResourceFetcher implements ResourceFetcher {
...
+        const headers: Headers = {}
+
+        headers['User-Agent'] = VSCODE_EXTENSION_ID.awstoolkit
+
+        if (requestHeaders?.eTag !== undefined) {
+            headers['If-None-Match'] = requestHeaders.eTag
+        }
+
+        if (requestHeaders?.gZip) {
+            headers['Accept-Encoding'] = 'gzip'
```
hoffa, line 197: Where does decompression happen?

nkomonen-amazon: I was assuming I had to decompress the result myself at some point, but I think the http request module is doing it for me.

I verified and the Result object has the header `Content-Encoding: gzip`, but the text given to me is uncompressed.
---
I tested with `wget` + `Accept-Encoding: gzip` and that result is compressed, so the compression is working as expected.

justinmk3: Let's at least leave a comment somewhere here, mentioning that info

src/shared/schemas.ts
```diff
@@ -140,41 +139,42 @@ export class SchemaService {
  * @param extensionContext VSCode extension context
  */
 export async function getDefaultSchemas(extensionContext: vscode.ExtensionContext): Promise<Schemas | undefined> {
```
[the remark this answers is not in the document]
nkomonen-amazon: Since we don't have a specific telemetry event type for this yet, is it expected I create a new definition on toolkit-common?

src/shared/schemas.ts
hoffa, line 148: Note for testing purposes only: the schema in `main` doesn't yet include any docs while the one in `develop` (https://github.com/aws/serverless-application-model/blob/develop/samtranslator/schema/schema.json; about 3x larger) does. So documentation should show up upon hover for that one. `main` will be updated over coming weeks.

src/shared/schemas.ts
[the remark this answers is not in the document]
justinmk3: oops, I didn't realize we lack telemetry here. hmm...

src/shared/schemas.ts
nkomonen-amazon: Thanks for pointing this out, I wasn't sure at first if documentation just wasn't included when I was playing around with it. I'll use `develop` for now.

src/shared/resourcefetcher/httpResourceFetcher.ts
nkomonen-amazon: Added a comment to the line where we read the data, indicating we don't need to decompress the data.

src/shared/schemas.ts
```diff
@@ -17,9 +17,15 @@ import { AWS_SCHEME } from './constants'
...
 import { SystemUtilities } from './systemUtilities'
 import { normalizeVSCodeUri } from './utilities/vsCodeUtils'
+import { telemetry } from './telemetry/telemetry'
 
-const goformationManifestUrl = 'https://api.github.com/repos/awslabs/goformation/releases/latest'
 const devfileManifestUrl = 'https://api.github.com/repos/devfile/api/releases/latest'
+
+// The `develop` branch currently has documentation in the schema, eventually `main` will get that
+// and we can look to switch to using the URL below:
+// export const samAndCfnSchemaUrl = `https://raw.githubusercontent.com/aws/serverless-application-model/main/samtranslator/schema/schema.json`
```
hoffa, line 26: It now includes docs.

nkomonen-amazon: Thanks, updated.

src/shared/schemas.ts
```diff
@@ -17,8 +17,9 @@ import { AWS_SCHEME } from './constants'
 import { writeFile } from 'fs-extra'
 import { SystemUtilities } from './systemUtilities'
 import { normalizeVSCodeUri } from './utilities/vsCodeUtils'
+import { telemetry } from './telemetry/telemetry'
 
-const goformationManifestUrl = 'https://api.github.com/repos/awslabs/goformation/releases/latest'
+export const samAndCfnSchemaUrl = `https://raw.githubusercontent.com/aws/serverless-application-model/main/samtranslator/schema/schema.json`
```
justinmk3, line 22:
```suggestion
// Note: this file is currently 12+ MB. When requesting it, specify compression/gzip. 
export const samAndCfnSchemaUrl = 'https://raw.githubusercontent.com/aws/serverless-application-model/main/samtranslator/schema/schema.json'
```

justinmk3:
integ tests are failing:

```
  1) getDefaultSchemas()
       uses cache for all requests on second function call:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  {
    awsRegion: 'us-west-2',
+   result: 'Succeeded',
-   reason: 'Cache hit',
-   result: 'Cancelled',
    url: 'https://raw.githubusercontent.com/aws/serverless-application-model/main/samtranslator/schema/schema.json'
  }
      + expected - actual

       {
         "awsRegion": "us-west-2"
      -  "result": "Succeeded"
      +  "reason": "Cache hit"
      +  "result": "Cancelled"
         "url": "https://raw.githubusercontent.com/aws/serverless-application-model/main/samtranslator/schema/schema.json"
       }

PR #2094, opened 2021-09-28
fix(telemetry): SAM CLI should inherit Toolkit telemetry setting

justinmk3:
# Problem:

- SAM CLI is an implicit dependency of AWS Toolkit, but if the customer  disables Toolkit telemetry, that choice isn't applied to SAM CLI's own telemetry.
- SAM CLI telemetry messages are very noisy in the Toolkit logs.
    - Note: when the Toolkit log-level is "debug", we invoke SAM CLI      with `sam --debug`.

# Solution:
- When the customer has disabled AWS Toolkit telemetry, reflect that in SAM CLI invocations by setting an [env var](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-telemetry.html):
  ```
  SAM_CLI_TELEMETRY=0
  ```
- Reduces CI logs by 400 lines.
- before:
  ```
  2023-09-29 22:09:58 [VERBOSE]: stderr: 2023-09-29 22:09:58,967 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
  2023-09-29 22:09:58 [VERBOSE]: stderr: 2023-09-29 22:09:58,973 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
  2023-09-29 22:09:58 [VERBOSE]: stderr: 2023-09-29 22:09:58,973 | Sending Telemetry: {'metrics': [{'commandRun': {'requestId': 'c676ac68-9e8e-4bdb-b7a3-69488d5b31c3', 'installationId': 'fc43f435-e8ba-4f9c-a905-a5f64bbcb6a2', 'sessionId': 'ca07ba9c-c856-4a17-bad1-05d06ec0447a', 'executionEnvironment': 'AWSCodeBuild', 'ci': True, 'pyversion': '3.10.12', 'samcliVersion': '1.94.0', 'awsProfileProvided': False, 'debugFlagProvided': True, 'region': '', 'commandName': 'sam build', 'metricSpecificAttributes': {'projectType': 'CFN', 'gitOrigin': None, 'projectName': 'b404669a6e496d0c9cc117229b17f443fb375168d14b426dc94025e393f40113', 'initialCommit': None}, 'duration': 26295, 'exitReason': 'success', 'exitCode': 0}}]}
  2023-09-29 22:09:58,973 | Unable to find Click Context for getting session_id.
[...]

PR #3170, opened 2023-02-15
telemetry: add UA environment variable to SAM CLI invocations

JadenSimon:
## Problem
SAM CLI doesn't know which products are executing it on the user's behalf

## Solution
Add an environment variable containing additional context

## License

### DIFF
--- a/src/shared/sam/cli/samCliInvokerUtils.ts
+++ b/src/shared/sam/cli/samCliInvokerUtils.ts
@@ -105,8 +105,8 @@ export async function addTelemetryEnvVar(options: SpawnOptions | undefined): Pro
     return {
         ...options,
         env: {
-            ...options?.env,
             SAM_CLI_TELEMETRY_FROM_IDE: await getUserAgent({ includeClientId: false }),
+            ...options?.env,
         },
     }
 }
--- a/src/shared/sam/cli/samCliInvokerUtils.ts
+++ b/src/shared/sam/cli/samCliInvokerUtils.ts
@@ -5,6 +5,7 @@
 
 import { SpawnOptions } from 'child_process'
 import { getLogger } from '../../logger'
+import { getUserAgent } from '../../telemetry/util'
 import { ChildProcessResult, ChildProcessOptions } from '../../utilities/childProcess'
 
 export interface SamCliProcessInvokeOptions {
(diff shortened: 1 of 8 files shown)


src/shared/sam/cli/samCliInvokerUtils.ts
```diff
@@ -99,3 +100,13 @@ function startsWithEscapeSequence(text: string, sequences = acceptedSequences):
 function startsWithError(text: string): boolean {
     return text.startsWith('Error:')
 }
+
+export async function addTelemetryEnvVar(options: SpawnOptions | undefined): Promise<SpawnOptions> {
+    return {
+        ...options,
+        env: {
+            SAM_CLI_TELEMETRY_FROM_IDE: await getUserAgent({ includeClientId: false }),
+            ...options?.env,
```
justinmk3, line 109: related: https://github.com/aws/aws-toolkit-vscode/pull/2094

src/shared/sam/cli/samCliInvokerUtils.ts
```diff
@@ -99,3 +100,13 @@ function startsWithEscapeSequence(text: string, sequences = acceptedSequences):
 function startsWithError(text: string): boolean {
     return text.startsWith('Error:')
 }
+
+export async function addTelemetryEnvVar(options: SpawnOptions | undefined): Promise<SpawnOptions> {
```
justinmk3, line 104: should this live in chidProcess.ts ?

justinmk3 (approved):

src/shared/sam/cli/samCliInvokerUtils.ts
JadenSimon: It could be generalized to `addEnvVar` where the key and value are passed as args. But IMO that'd only be more ergonomic if added as a method that mutates the options prior to spawning. Which could be nice?

src/shared/sam/cli/samCliInvokerUtils.ts
JadenSimon: So I guess now we probably want to conditionally disable SAM CLI telemetry based off the global Toolkit setting as SAM CLI will expect some telemetry data from the Toolkits.

PR #3175, opened 2023-02-16
CWL: Code lenses on 'Search Log Stream' document

nkomonen-amazon:
## Problem

Previously, we used an inefficient implementation of Go To Definition (Ctrl + Click) to go to
open a specific Log Stream document from a document that had many Log Streams.

It was inefficient since we could not lazily load the Go To Definition objects and
it caused performance issues due to Go To Definition pre-loading pages
as the user hovered over the lines while holding Ctrl.

Additionally it was not obvious that we had this functionality available.

## Solution

- Drop the Go To Definition feature
- Replace it with Code Lenses that will appear at the beginning of
  each Log Stream. When clicked they will load the specific events
  of that log stream and open a new document

## Steps to Reproduce

1. Open the AWS Explorer
2. Drop down `CloudWatch Logs`
3. Right click one of them and select `Search Log Group`
4. Once fully loaded AND if you have multiple Log Streams, you will be able to see multiple Code Lenses that say `Load following Log Stream...`

Screen Shot 2023-02-06 at 11 19 39 AM

## 160,000 Lines Gif
The following is a 160,000 line log file with ~10,000 code lenses. The gif is 10 fps so it wont look as good.
Also with pagination, we can only load 10,000 events max per page. So it will take a good amount of loads to start impacting performance.

Kapture 2023-02-16 at 23 53 47

From some quick testing once I get around 400k lines, scrolling isn't as smooth as if there were no code lenses but it is still completely usable.

## Future performance optimizations

The vscode API allows us to see where the user currently is in the document. This can allow us to load code lenses in the surrounding of where the user currently is in the document. This should give a big performance increase if we find users need it.

## License

### DIFF
--- a/src/cloudWatchLogs/activation.ts
+++ b/src/cloudWatchLogs/activation.ts
@@ -19,6 +19,7 @@ import { Commands } from '../shared/vscode/commands2'
 import { searchLogGroup } from './commands/searchLogGroup'
 import { changeLogSearchParams } from './changeLogSearch'
 import { CloudWatchLogsNode } from './explorer/cloudWatchLogsNode'
+import { loadAndOpenInitialLogStreamFile, LogStreamCodeLensProvider } from './document/logStreamsCodeLensProvider'
 
 export async function activate(context: vscode.ExtensionContext, configuration: Settings): Promise<void> {
     const settings = new CloudWatchLogsSettings(configuration)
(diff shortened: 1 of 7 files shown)


justinmk3:
Seems reasonable. Main reason we avoided codelenses initially, was because there might be hundreds of them if search results come from different logstreams.

What's the performance like in that case?

nkomonen-amazon:
@justinmk3 I've updated the PR description with more info regarding large files

justinmk3 (approved):

justinmk3:
Let's try it!

PR #3181, opened 2023-02-20
fix(tests): coverage information found on head

andrewyuq:
## License

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -42,6 +42,16 @@ jobs:
                   verbose: true
                   file: ./coverage/coverage-final.json
                   flags: macos-unittests
+            - name: Code coverage (CodeWhisperer)
+              env:
+                  # Unset NODE_OPTIONS because of https://github.com/codecov/uploader/issues/475
+                  NODE_OPTIONS: ''
+              if: ${{ github.repository == 'aws/aws-toolkit-vscode' && ( github.ref == 'master' || github.event_name == 'pull_request' ) }}
+              uses: codecov/codecov-action@v2
+              with:
+                  verbose: true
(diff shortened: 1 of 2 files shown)


.github/workflows/node.js.yml
```diff
@@ -42,6 +42,16 @@ jobs:
                   verbose: true
                   file: ./coverage/coverage-final.json
                   flags: macos-unittests
+            - name: Code coverage (CodeWhisperer)
```
justinmk3, line 45: what is the purpose of this? why isn't the existing codecov upload good enough? It is not scalable to add different codecov accounts for every toolkit feature.

andrewyuq: It's not adding a different codecov account but just doing a different upload for a new [flag](https://docs.codecov.com/docs/flags). See the following notes in the doc:
> It is important to note that for flagging to work, separate coverage reports must be uploaded and flagged per project in the mono repo
We are still running the build/test once. This action just does another upload for the flag tracker. There's only one codecov account involved.

andrewyuq: And now it shows another line of github check for `codecov/project/codewhisperer` showing code coverage

justinmk3: > And now it shows another line of github check for `codecov/project/codewhisperer` showing code coverage

But we don't want CI to be constantly red. We intentionally removed this CI check in the past because it was noisy.

justinmk3 (approved):

.github/workflows/node.js.yml
andrewyuq: We are planning to reach 75% in the next 1-2 weeks @crossrac is working on it.

zixlin7 (approved):

PR #3182, opened 2023-02-20
tests: remove `Window.vscode()` and related logic

JadenSimon:
## Problem
Writing wrappers for the sole purpose of making an implementation easier to test is not a sustainable approach.

## Solution
* Remove `Window.vscode()` and `FakeWindow`
* Extend `TestWindow` to cover the same scenarios that `FakeWindow` was used for

### Important Notes
This PR uses `Proxy` heavily in order to make "testable" versions of native `vscode` objects. Some things like `showInformationMessage` and `withProgress` do _not_ have an API that can be manipulated programmatically. For these methods, we have to create in-mem representations that can be tested. These currently live in `src/test/shared/vscode/message.ts`.

## License


src/extension.ts
```diff
@@ -92,6 +92,10 @@ export async function activate(context: vscode.ExtensionContext) {
         handleError(error, info.id, defaultMessage)
     })
 
+    if (isCloud9()) {
+        vscode.window.withProgress = wrapWithProgressForCloud9(globals.outputChannel)
+    }
+
```
JadenSimon, line 98: This is the only non-test related change. Previously the output channel was wrapped for VSC and C9. Now it's just C9.

src/test/globalSetup.test.ts
```diff
@@ -22,15 +22,15 @@ import { activateExtension } from '../shared/utilities/vsCodeUtils'
 import { FakeExtensionContext, FakeMemento } from './fakeExtensionContext'
 import { TestLogger } from './testLogger'
 import * as testUtil from './testUtil'
-import { createTestWindow } from './shared/vscode/window'
+import { getTestWindow, resetTestWindow } from './shared/vscode/window'
 
 const testReportDir = join(__dirname, '../../../.test-reports')
 const testLogOutput = join(testReportDir, 'testLog.log')
 const globalSandbox = sinon.createSandbox()
+const maxTestDuration = 30_000
```
JadenSimon, line 30: Arbitrary max duration for unit tests. 30 seconds is already very generous but we have some tests that are taking ~7-8 seconds on my machine. Those should probably be moved or optimized.

src/extension.ts
```diff
@@ -317,6 +321,35 @@ function recordToolkitInitialization(activationStartedOn: number, logger?: Logge
     }
 }
 
+/**
+ * Wraps the `vscode.window.withProgress` functionality with functionality that also writes to the output channel.
+ *
+ * Cloud9 does not show a progress notification.
+ */
+function wrapWithProgressForCloud9(channel: vscode.OutputChannel): typeof vscode.window['withProgress'] {
```
justinmk3, line 329: should this live in shared/vscode/ or somewhere like that?

justinmk3 (approved):

src/extension.ts
JadenSimon: Depends on how visible we want this behavior/edge-case. Previously it was kind of tucked away so putting it in the extension entry point makes it more visible, maybe too much?

PR #3183, opened 2023-02-21
S3: add most recently used folder to upload files wizard

gsweetwood:
## Problem
The Upload Files wizard allows the user to choose a bucket, but
not folders.
## Solution
Add a 'last-expanded' folder and a 'last-uploaded-to' folder
to the wizard and upload to those folders.

##Screenshots
folders option

## License


src/test/s3/commands/uploadFile.test.ts
```diff
@@ -175,6 +189,28 @@ describe('uploadFileCommand', function () {
         ])
     })
 
+    it('successfully upload file to selected folder', async function () {
```
justinmk3: This 20 lines of code is almost identical to the test above it. I pushed a commit showing how this redundant code can be avoided:

- because the tests are very similar, do both the "bucket" and "folder" cases in the same test. It doesn't change the meaning of the test.
- the assert also shows how the output channel accumlates messages, which is a nice bonus.

src/test/s3/commands/uploadFile.test.ts
```diff
@@ -27,10 +28,23 @@ describe('uploadFileCommand', function () {
     const sizeBytes = 16
     const fileLocation = vscode.Uri.file('/file.jpg')
     const statFile: FileSizeBytes = _file => sizeBytes
+    const bucketResponse = { label: 'label', bucket: { Name: bucketName } }
+    const folderResponse = {
+        label: 'label',
+        bucket: { Name: bucketName },
+        folder: { name: 'folderA', path: 'folderA/', arn: 'arn' },
+    }
+    const getBucketFolderResponse: (
```
justinmk3: The existing sibling function on line 47 is called `getBucket`. Similarly, this could be called `getFolder`.

The name "getBucketFolderResponse" is explaining too much. The extra details in the name imply that something important or new is happening, which can be confusing.

src/test/s3/commands/uploadFile.test.ts
```diff
@@ -11,6 +11,7 @@ import {
     getFilesToUpload,
```
justinmk3, line 11: Would also be worth having test coverage for the use of globalState, similar to https://github.com/aws/aws-toolkit-vscode/blob/e9558c8776f41c152fd77f2f3de9839a1b19887c/src/test/s3/commands/downloadFileAs.test.ts#L60

src/test/s3/commands/uploadFile.test.ts
justinmk3: I pushed a commit with this change

justinmk3 (approved):

src/s3/commands/uploadFile.ts
```diff
@@ -418,15 +440,77 @@ export async function promptUserForBucket(
         }
     })
 
+    const lastTouchedFolder = globals.context.globalState.get<SavedFolder | undefined>('aws.lastTouchedS3Folder')
+    let lastFolderItem: BucketQuickPickItem | undefined = undefined
+    if (lastTouchedFolder) {
+        lastFolderItem = {
+            label: lastTouchedFolder.folder.name,
```
justinmk3, line 447: Should the "Folder: " prefix also be here, like with `lastUploadedToFolder` ?

src/s3/commands/uploadFile.ts
```diff
@@ -418,15 +440,77 @@ export async function promptUserForBucket(
...
+            bucket: { Name: lastTouchedFolder.bucket.name },
+            folder: lastTouchedFolder.folder,
+        }
+    }
+
+    const lastUploadedToFolder = globals.context.globalState.get<SavedFolder | undefined>('aws.lastUploadedToS3Folder')
+    let lastUploadedFolderItem: BucketQuickPickItem | undefined = undefined
+    if (lastUploadedToFolder) {
+        lastUploadedFolderItem = {
+            label: `Folder: ${lastUploadedToFolder.folder.name}`,
```
justinmk3: I removed this "Folder:" prefix because 

1. inconsistent with `lastFolderItem` 
2. the `description` gives similar information

PR #3185, opened 2023-02-21
refactor(package.json): viewItem format

justinmk3:
Problem:
Not scalable to match enumerated viewItem names for standard features like "Copy URL", "Copy Name", etc.

Solution:
Define a naming convention which switches on standard features ("Copy URL", "Copy Name", …) as follows:

awsFoo-ANUED

    /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable|awsCloudFormationFunctionNode|awsStateMachineNode|awsCloudFormationNode|awsS3BucketNode|awsS3FolderNode|awsS3FileNode|
    awsApiGatewayNode|awsIotThingNode)$|^(awsAppRunnerServiceNode|awsIotCertificateNode|awsIotPolicyNode|awsIotPolicyVersionNode)/

    awsCloudWatchLogNode|
    awsEcrRepositoryNode|
    awsEcsServiceNode|

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -1280,42 +1280,42 @@
             "view/item/context": [
                 {
                     "command": "aws.apig.invokeRemoteRestApi",
-                    "when": "view == aws.explorer && viewItem =~ /^(awsApiGatewayNode)$/",
+                    "when": "view == aws.explorer && viewItem =~ /^(awsApiGatewayNode)[-ANU]*$/",
                     "group": "0@1"
                 },
                 {
                     "command": "aws.codecatalyst.removeConnection",
-                    "when": "viewItem == awsCodeCatalystNodeSaved",
+                    "when": "viewItem =~ /^awsCodeCatalystNodeSaved[-ANU]*$/",
                     "group": "0@1"
                 },
                 {
                     "command": "aws.codecatalyst.removeConnection",
-                    "when": "viewItem == awsCodeCatalystNodeSaved",
+                    "when": "viewItem =~ /^awsCodeCatalystNodeSaved[-ANU]*$/",
                     "group": "inline@1"
                 },
                 {
                     "command": "aws.ecr.createRepository",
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -1525,62 +1525,62 @@
...
                 {
                     "command": "aws.iot.copyEndpoint",
-                    "when": "view == aws.explorer && viewItem == awsIotNode",
+                    "when": "view == aws.explorer && viewItem =~ /^awsIotNode[-ANU]*$/",
                     "group": "2@1"
                 },
                 {
                     "command": "aws.copyName",
-                    "when": "view == aws.explorer && viewItem =~ /^(awsRegionFunctionNode|awsRegionFunctionNodeDownloadable|awsCloudFormationFunctionNode|awsStateMachineNode|awsCloudFormationNode|awsS3BucketNode|awsS3FolderNode|awsS3FileNode|awsApiGatewayNode|awsIotThingNode)$|^(awsAppRunnerServiceNode|awsIotCertificateNode|awsIotPolicyNode|awsIotPolicyVersionNode)/",
+                    "when": "view == aws.explorer && viewItem =~ /^aws[^-]*-([AU]*N)/",
```
justinmk3, line 1578: copyName is enabled by the viewItem name instead of enumerating every viewItem here.

Could also do this for "DISABLED"/"ENABLED".

JadenSimon: Another idea: use the same pattern as CSS classes and append dot-separated (or whatever other delimiter) keys to the end of a node's context value. Then you can pattern match for `/\.${contextKey}/`, e.g. `awsRegionFunctionNode.copyName.copyArn`

The resulting strings could get pretty big though.

justinmk3: context value would be the natural mechanism, but AFAIR it doesn't work in cloud9.

PR #3187, opened 2023-02-22
file explorer: add upoloadToAws command

gsweetwood:
## Problem
The toolkit can upload files and folders to Lambda and S3, but the commands are only available from the AWS Explorer or the Command Palette.
## Solution
Add a context menu item to all files and folders that prompts for the resource to upload to. Initiate the appropriate wizard passing the file/folder. Current options are Lambda and S3.

## Screenshot
upload to aws

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -1100,6 +1100,10 @@
                     "when": "false"
                 },
                 {
+                    "command": "aws.uploadToAws",
+                    "when": "false"
+                },
+                {
                     "command": "aws.dev.openMenu",
                     "when": "aws.isDevMode || isCloud9"
                 }
@@ -1272,8 +1276,8 @@
                     "group": "z_aws@2"
                 },
                 {
-                    "command": "aws.uploadLambda",
-                    "when": "explorerResourceIsFolder || isFileSystemResource && resourceFilename =~ /^template\\.(json|yml|yaml)$/",
+                    "command": "aws.uploadToAws",
+                    "when": "explorerResourceIsFolder || isFileSystemResource",
                     "group": "z_aws@3"
                 }
             ],
@@ -2397,6 +2401,16 @@
             {
                 "command": "aws.uploadLambda",
                 "title": "%AWS.command.uploadLambda%",
+                "category": "%AWS.title%",
+                "cloud9": {
+                    "cn": {
+                        "category": "%AWS.title.cn%"
+                    }
+                }
+            },
+            {
+                "command": "aws.uploadToAws",
+                "title": "%AWS.command.uploadToAws%",
                 "category": "%AWS.title%",
                 "cloud9": {
                     "cn": {
--- a/package.nls.json
+++ b/package.nls.json
@@ -101,6 +101,7 @@
     "AWS.command.aboutToolkit": "About Toolkit",
     "AWS.command.downloadLambda": "Download...",
     "AWS.command.uploadLambda": "Upload Lambda...",
+    "AWS.command.uploadToAws": "Upload to AWS...",
     "AWS.command.invokeLambda": "Invoke on AWS",
     "AWS.command.invokeLambda.cn": "Invoke on Amazon",
     "AWS.command.refreshAwsExplorer": "Refresh Explorer",
--- a/src/lambda/activation.ts
+++ b/src/lambda/activation.ts
@@ -15,6 +15,7 @@ import { registerSamInvokeVueCommand } from './vue/configEditor/samInvokeBackend
 import { Commands } from '../shared/vscode/commands2'
 import { DefaultLambdaClient } from '../shared/clients/lambdaClient'
 import { copyLambdaUrl } from './commands/copyLambdaUrl'
+import { uploadToAwsCommand } from './commands/uploadToAwsCommand'
 
 /**
  * Activates Lambda components.
@@ -57,6 +58,7 @@ export async function activate(context: ExtContext): Promise<void> {
             'aws.copyLambdaUrl',
             async (node: LambdaFunctionNode) => await copyLambdaUrl(node, new DefaultLambdaClient(node.regionCode))
         ),
+        Commands.register('aws.uploadToAws', async (arg: vscode.Uri) => await uploadToAwsCommand(arg)),
         registerSamInvokeVueCommand(context)
     )
 }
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -1272,8 +1276,8 @@
                     "group": "z_aws@2"
                 },
                 {
-                    "command": "aws.uploadLambda",
-                    "when": "explorerResourceIsFolder || isFileSystemResource && resourceFilename =~ /^template\\.(json|yml|yaml)$/",
+                    "command": "aws.uploadToAws",
```
justinmk3, line 1279: For now, let's leave `aws.uploadLambda` as it was, and add a new item for `aws.uploadToAws` that only works on folders.

package.nls.json
```diff
@@ -101,6 +101,7 @@
     "AWS.command.aboutToolkit": "About Toolkit",
     "AWS.command.downloadLambda": "Download...",
     "AWS.command.uploadLambda": "Upload Lambda...",
+    "AWS.command.uploadToAws": "Upload to AWS...",
```
justinmk3, line 104: Needs a `.cn` variant. Example: https://github.com/aws/aws-toolkit-vscode/blob/b20e150914a817600d2ca416b6623b1bb5440443/package.nls.json#L9-L10

package.json
```diff
@@ -2404,6 +2408,16 @@
                     }
                 }
             },
+            {
+                "command": "aws.uploadToAws",
+                "title": "%AWS.command.uploadToAws%",
+                "category": "%AWS.title%",
+                "cloud9": {
+                    "cn": {
+                        "category": "%AWS.title.cn%"
```
justinmk3, line 2417: also:

    "title": "%AWS.command.uploadToAws.cn%"

src/lambda/commands/uploadToAwsCommand.ts
```diff
@@ -0,0 +1,59 @@
+/*!
```
justinmk3, line 1: Should live in `src/shared/` not `.../lambda/`. This file is not lambda-specific.

src/lambda/commands/uploadToAwsCommand.ts
```diff
@@ -0,0 +1,59 @@
...
+    })
+}
+
+export class UploadToAwsWizard extends Wizard<UploadToAwsWizardState> {
+    constructor() {
+        super()
+        this.form.resource.bindPrompter(() => createChooseResourcePrompter())
+        // region is needed to create an S3 client for the upload command
+        this.form.region.bindPrompter(() => createRegionPrompter().transform(region => region.id), {
+            showWhen: ({ resource }) => resource === 's3',
```
justinmk3, line 37: Despite the comment, I don't understand why we are creating a region prompter here. 

I would expect the "upload to s3" logic here: https://github.com/aws/aws-toolkit-vscode/blob/b20e150914a817600d2ca416b6623b1bb5440443/src/s3/activation.ts#L59-L72

to be extracted to a function, and that function would be called if "Upload to S3" is chosen from this wizard.

The idea is that "Upload to AWS" is a single menu that dispatches to various wizards. We don't want to duplicate parts of the various wizards here.

src/lambda/commands/uploadToAwsCommand.ts
```diff
@@ -0,0 +1,59 @@
...
+    const response = await new UploadToAwsWizard().run()
+
+    if (!response) {
+        throw new CancellationError('user')
+    }
+
+    if (response.resource === 's3') {
+        if (response.region) {
+            const s3 = new DefaultS3Client(response.region)
+            await uploadFileCommand(s3, fileOrFolder)
```
justinmk3, line 52: This is missing some key parts of https://github.com/aws/aws-toolkit-vscode/blob/b20e150914a817600d2ca416b6623b1bb5440443/src/s3/activation.ts#L5 . That's one reason why we don't want to duplicate that logic here.

justinmk3 (changes requested):
comments

src/lambda/commands/uploadToAwsCommand.ts
gsweetwood: This logic forces them to use a default region, which is what I was originally avoiding, but if no one is complaining about the current logic then I'll go ahead and refactor this a little.

package.nls.json
```diff
@@ -101,6 +101,8 @@
...
+    "AWS.command.uploadToAws.cn": "Upload to AWS...",
```
justinmk3, line 105: Like the `.cn` items, "Amazon" must be used instead of "AWS".

```suggestion
    "AWS.command.uploadToAws.cn": "Upload to Amazon...",
```

src/lambda/activation.ts
```diff
@@ -57,6 +58,7 @@ export async function activate(context: ExtContext): Promise<void> {
             'aws.copyLambdaUrl',
             async (node: LambdaFunctionNode) => await copyLambdaUrl(node, new DefaultLambdaClient(node.regionCode))
         ),
+        Commands.register('aws.uploadToAws', async (arg: vscode.Uri) => await uploadToAwsCommand(arg)),
```
justinmk3, line 61: this is not a lambda-specific command, is there somewhere more relevant it could live?

src/test/shared/uploadToAwsCommand.test.ts
```diff
@@ -0,0 +1,24 @@
...
+import { UploadToAwsWizard, UploadToAwsWizardState } from '../../shared/uploadToAwsCommand'
+
+describe('UploadToAwsWizard', function () {
+    let tester: WizardTester<UploadToAwsWizardState>
+
+    beforeEach(function () {
+        tester = createWizardTester(new UploadToAwsWizard())
+    })
+
+    it('shows region prompt when S3 is selected', function () {
```
justinmk3, line 15: the region menu is not really the main point of this wizard, so these test names are confusing

src/lambda/commands/uploadToAwsCommand.ts
justinmk3: > This logic forces them to use a default region

Could add a "switch region" item to the menu similar to: https://github.com/aws/aws-toolkit-vscode/blob/2cfed26f0701afc0230ce39246772aeeeb6c6334/src/shared/ui/common/regionSubmenu.ts#L43

The default region is usually what is wanted. And the "Switch region" item gives an escape hatch for other cases. And avoids an extra menu.

PR #3191, opened 2023-02-24
prompters: add a common S3 bucket prompter that recalls recent buckets

gsweetwood:
## Problem
Prompting for an S3 bucket is common in our command wizards, but each time they were created from scratch. They also do not have logic to recall the last bucket that was used (except sam sync).
## Solution
Create a function that returns a bucket prompter and recalls the 'recently used' bucket.

## License


justinmk3 (changes requested):
> src/lambda/wizards/samDeployWizard.ts:S3_BUCKET has a bucket prompt that allows (1) creating a new bucket, (2) entering bucket by name (no list), (3) selecting from list (if user has list permissions), (4) select recently-used buckets.
>
> ... We should lift that logic out of src/lambda/wizards/samDeployWizard.ts so that we can use it whenever a bucket name is needed.

Doesn't look like this PR implements 1-3?

PR #3195, opened 2023-02-27
tests(codewhisperer): add more unit tests

crossrac:
PR adding unit additional CodeWhisperer unit tests in order to increase test coverage percentage.

## License


src/test/codewhisperer/util/authUtil.test.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    afterEach(async function () {
+        sinon.restore()
+    })
+
+    //TODO
+    //factory function to create connectionlists dynamically
+    //stub secondaryAuth useNewConnection? vs auth instance use connection
+    //remove auth and profile from beforeEach and see if it breaks
+    //todo: figure out what sinon.replace in beforeEach does
```
JadenSimon, line 123: Oh it just removes noisy error messages from the test output when calling `Auth.logout`

src/test/codewhisperer/util/authUtil.test.ts
```diff
@@ -0,0 +1,180 @@
...
+        sinon.stub(Auth.instance, 'listConnections').resolves(mockConnListNoBuilder)
+        sinon.stub(Auth.instance, 'createConnection').resolves(builderIdConn)
+        const authSpy = sinon.stub(Auth.instance, 'useConnection')
+
+        getTestWindow().onDidShowQuickPick(async picker => {
+            await picker.untilReady()
+            picker.acceptItem(picker.items[1])
+        })
+
+        await AuthUtil.instance.connectToAwsBuilderId()
```
JadenSimon, line 138: You shouldn't need to stub anything
```suggestion
        const authUtil = new AuthUtil(auth)
        await authUtil.connectToAwsBuilderId()
```

src/test/codewhisperer/util/authUtil.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    
+        assert.ok(authSpy.called)
+        assert.strictEqual(authSpy.args[0][0], builderIdConn)
```
JadenSimon, line 141: Use `auth.activeConnection` instead of a spy

src/test/codewhisperer/commands/startSecurityScan.test.ts
```diff
@@ -163,7 +164,8 @@ describe('startSecurityScan', function () {
         )
         assert.ok(commandSpy.calledWith('workbench.action.problems.focus'))
         assert.ok(securityScanRenderSpy.calledOnce)
-        const warnings = getTestWindow().shownMessages.filter(m => m.severity === SeverityLevel.Warning)
-        assert.strictEqual(warnings.length, 0)
+        //TODO: below assertion fails when running full test suite succeeds when running test on individual file
```
JadenSimon, line 167: Usually that means some tests aren't have their state cleaned-up correctly.

JadenSimon (commented):
A few tips/comments

src/test/codewhisperer/commands/startSecurityScan.test.ts
crossrac: That's helpful, thank you! I'll review the tests I've added to see where I can fix the state cleanup.

src/test/codewhisperer/util/authUtil.test.ts
crossrac: Hi @JadenSimon , thanks for your comment!

I updated the tests to use `new Auth(auth)` per your note above and in doing so was able to remove the majority of the Auth stubs. 

However, I have still not found a way to implement the tests without the `useConnection` stub, due to the issue described below. Please let me know if the updated implementation works. 



**Detail:**
- When I use `AuthUtil(auth)` it creates the builderID connection, but generates the following error when the [secondaryAuth calls useConnection at the end of the function](https://github.com/aws/aws-toolkit-vscode/blob/master/src/codewhisperer/util/authUtil.ts#L95):

     - `Error: Connection does not exist: https://view.awsapps.com/start?scopes=codecatalyst:read_write,codewhisperer:analysis,codewhisperer:completions`

- It appears the `AuthUtil` is using the auth instance created in/for the test, but the `secondaryAuth` is not, and so it can’t find the profile associated with the new connection.

With that in mind:
- In the latest update, I used `new AuthUtil(auth)` as you mentioned and so was able to remove the stubs for `listConnection` and `createConnection`.  
- I still stub `useConnection`, in order to not get the above error and but still verify that the `AuthUtil` is calling it with the correct connection.

Does this work in your view? Alternatively, is there a good way to get the `secondaryAuth` to use the correct `Auth` obj (and by extension ProfileStore)?


Side note: 
- I am using ID as the connection property to check in the new assert, because when I try to use `type` or `label` I get the following error:
`Property 'type' does not exist on type 'Pick<IamConnection, "id”>’.` 
- I am not sure why `useConnection` is using `Pick<IamConnection, 'id'>` instead of `Pick<SsoConnection, 'id'>`

JadenSimon: > It appears the AuthUtil is using the auth instance created in/for the test, but the secondaryAuth is not, and so it can’t find the profile associated with the new connection.

Yup you're right. I've fixed this in https://github.com/aws/aws-toolkit-vscode/pull/3220/files#diff-b04b37a57656fe70020045378831b4b8df1ef84d02b18417f7b1599b2862cf1cR43-R48

> Does this work in your view?

So the thing with stubbing is that it _does_ work. There's nothing inherently wrong with using stubs and mocks. But each stubbed function or mocked object is a divergence from the original implementation. I think for our purposes stubbing `useConnection` is probably ok if my fix doesn't work.

> I am using ID as the connection property to check in the new assert
Are you looking at the captured param? The type signature is written intentionally loose but you can always cast it with `as Connection`. Also `Pick<IamConnection, 'id'>` is equivalent to `Pick<SsoConnection, 'id'>` in this context. I wouldn't worry too much about the exact names of things in TypeScript as it's a structurally-typed language. The only thing that matters are the shapes, which in this case is `{ id: string }` for both.

crossrac:
> /runintegrationtests

@justinmk3 thanks for your comment! I don't follow - is this meant to note that integration tests should be run prior to opening PR for review as well?

JadenSimon:
>  I don't follow - is this meant to note that integration tests should be run prior to opening PR for review as well?

It's just a command to run integration tests against this PR. I wouldn't worry about it.

PR #3207, opened 2023-03-02
test: Add code catalyst integ test

nkomonen-amazon:
## Problem

We want to ensure that Code Catalyst apis work as expected. If only we had
integration tests that could do this.

## Solution

This commit creates integ tests that test the CC APIs
this extension uses:

- creating a cc project
- creating a cc dev env
- connecting to the dev env through ssh

One thing to note is that the builderid/sso
configuration is required to be setup prior
to this test running (this test will fail until that
component is implemented in our CI, but is
being worked on separately).

## License

### DIFF
--- a/src/integrationTest/codecatalyst/client.test.ts
+++ b/src/integrationTest/codecatalyst/client.test.ts
@@ -1,0 +1,234 @@
+/*!
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as assert from 'assert'
+import {
+    CodeCatalystClient,
+    CodeCatalystOrg,
+    CodeCatalystProject,
+    createClient as createCodeCatalystClient,
+    DevEnvironment,
+} from '../../shared/clients/codecatalystClient'
+import { DeleteDevEnvironmentResponse } from 'aws-sdk/clients/codecatalyst'
+import { prepareDevEnvConnection } from '../../codecatalyst/model'
+import { Auth, SsoConnection } from '../../credentials/auth'
+import { CodeCatalystAuthenticationProvider, isValidCodeCatalystConnection } from '../../codecatalyst/auth'
+import { CodeCatalystCommands, DevEnvironmentSettings } from '../../codecatalyst/commands'
+import globals from '../../shared/extensionGlobals'
+
+let spaceName: CodeCatalystOrg['name']
+let projectName: CodeCatalystProject['name']
+let auth: Auth
+
+/**
+ * Key Information:
+ *
+ * This test can be run locally under the following conditions:
+ *
+ *   - You have previously configured BuilderId using the vscode extension.
+ *     This is because this test uses BuilderId/SSO information
+ *     that the extension would have naturally created in that process.
+ *
+ *   - A Space already exists in your Code Catalyst account.
+ *     We cannot currently automate this due to space creation
+ *     not being available in the Code Catalyst API.
+ *
+ *     TODO: Create a new space instead of using an existing one,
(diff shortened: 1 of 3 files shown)


src/integrationTest/codecatalyst/client.test.ts
```diff
@@ -0,0 +1,169 @@
...
+ * This provides a best effort to test the api functionality.
+ *
+ *   - We cannot currently perform integ tests with UI components,
+ *     so the underlying non-UI components that would be used in
+ *     those processes are being tested.
+ */
+describe('Test how this codebase uses the Code Catalyst API', () => {
+    let client: CodeCatalystClient
+
+    before(async () => {
```
JadenSimon, line 60: Need to skip these tests during CI:
```suggestion
    before(async function () {
        if (process.env['AWS_TOOLKIT_AUTOMATION'] !== 'local') {
            this.skip()
        }
```
If this doesn't work it might need to be in a `beforeEach` block

src/integrationTest/codecatalyst/client.test.ts
```diff
@@ -0,0 +1,169 @@
...
+ *     deleted manually.
+ *
+ *     TODO: Delete the project when the test is done,
+ *     if the api eventually allows us to do so.
+ *
+ * This provides a best effort to test the api functionality.
+ *
+ *   - We cannot currently perform integ tests with UI components,
+ *     so the underlying non-UI components that would be used in
+ *     those processes are being tested.
```
JadenSimon, line 55: `getTestWindow` _could_ be used in the integration test suite, it just requires duplicating some of the logic in `src/test/globalSetup.test.ts` to `src/integrationTest/globalSetup.test.ts`.

src/integrationTest/codecatalyst/client.test.ts
```diff
@@ -0,0 +1,169 @@
...
+import { DeleteDevEnvironmentResponse } from 'aws-sdk/clients/codecatalyst'
+import { prepareDevEnvConnection } from '../../codecatalyst/model'
+import { Auth, createBuilderIdConnection } from '../../credentials/auth'
+import { isValidCodeCatalystConnection } from '../../codecatalyst/auth'
+
+let spaceName: CodeCatalystOrg['name']
+let projectName: CodeCatalystProject['name']
+
+/**
+ * Key Information:
```
JadenSimon, line 23: Great notes!

src/integrationTest/codecatalyst/client.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * Creates a code catalyst client.
+     *
+     * This relies on BuilderId/SSO information already being configured.
+     * See the above "Key Information" comment for more info.
+     */
+    async function createTestCodeCatalystClient(): Promise<CodeCatalystClient> {
+        const conn = await createBuilderIdConnection(Auth.instance)
+        assert(isValidCodeCatalystConnection(conn))
```
JadenSimon, line 75: `createBuilderIdConnection` may cause the test to hang until the user logs into an account which we can't do in CI:
```suggestion
        const conn = (await Auth.instance.listConnections()).find(isValidCodeCatalystConnection)
        assert.ok(conn, 'No valid CodeCatalyst connection found.')
```

src/integrationTest/codecatalyst/client.test.ts
nkomonen-amazon: Thanks for the info, I've updated this comment to better reflect the reality.

src/integrationTest/codecatalyst/client.test.ts
nkomonen-amazon: Thanks, good so see they got some use :)

PR #2996, opened 2022-11-19
sam: throw error when no templates are found

gsweetwood:
## Problem
When running "Deploy SAM Application" command, if there are no SAM applications found, the menu is empty without any further feedback. This is confusing for new users or users just trying things out.
## Solution
Throw an error and provide the user with options to learn more via a link to the SAM docs and an option to create a new SAM application.

## Screenshot
image

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3211, opened 2023-03-06
feat(sam): "Sync SAM" shows more info if no template.yaml found

justinmk3:
## Problem:
When trying the "Sync SAM" command, the "No items found" message is confusing.

## Solution:
- Show a more informative message.
- Visit the help page if user hits Enter.

image

image

image

image

related: #2996

## License

### DIFF
--- a/src/shared/awsConsole.ts
+++ b/src/shared/awsConsole.ts
@@ -1,0 +1,17 @@
+/*!
+ * Copyright 2018-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+
+export function getAwsConsoleUrl(service: 'ecr' | 'cloudformation', region: string): vscode.Uri {
+    switch (service) {
+        case 'ecr':
+            return vscode.Uri.parse(`https://${region}.console.aws.amazon.com/ecr/repositories?region=${region}`)
+        case 'cloudformation':
+            return vscode.Uri.parse(`https://${region}.console.aws.amazon.com/cloudformation/home?region=${region}`)
+        default:
+            throw Error()
+    }
+}
--- a/src/shared/logger/winstonToolkitLogger.ts
+++ b/src/shared/logger/winstonToolkitLogger.ts
@@ -119,11 +119,18 @@ export class WinstonToolkitLogger implements Logger, vscode.Disposable {
     }
 
     private mapError(level: LogLevel, err: Error): Error | string {
+        // Use ToolkitError.trace even if we have source mapping (see below), because:
+        // 1. it is what users will see, we want visibility into that when debugging
+        // 2. it is often more useful than the stacktrace anyway
+        if (err instanceof ToolkitError) {
+            return err.trace
+        }
+
         if (isSourceMappingAvailable() && level === 'error') {
             return err
         }
 
-        return err instanceof ToolkitError ? err.trace : formatError(UnknownError.cast(err))
+        return formatError(UnknownError.cast(err))
     }
 
     private writeToLogs(level: LogLevel, message: string | Error, ...meta: any[]): number {
--- a/src/shared/sam/sync.ts
+++ b/src/shared/sam/sync.ts
@@ -7,6 +7,7 @@ import globals from '../extensionGlobals'
 
 import * as vscode from 'vscode'
 import * as path from 'path'
+import * as nls from 'vscode-nls'
 import * as localizedText from '../localizedText'
 import { DefaultS3Client } from '../clients/s3Client'
 import { Wizard } from '../wizards/wizard'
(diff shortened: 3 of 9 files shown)


JadenSimon (approved):

src/shared/logger/winstonToolkitLogger.ts
```diff
@@ -119,11 +119,18 @@ export class WinstonToolkitLogger implements Logger, vscode.Disposable {
...
 
     private mapError(level: LogLevel, err: Error): Error | string {
+        // Use ToolkitError.trace even if we have source mapping (see below), because:
+        // 1. it is what users will see, we want visibility into that when debugging
+        // 2. it is often more useful than the stacktrace anyway
+        if (err instanceof ToolkitError) {
+            return err.trace
+        }
+
         if (isSourceMappingAvailable() && level === 'error') {
```
justinmk3, line 129: @JadenSimon I changed the logic to prefer ToolkitError.trace because (1) it's closer to what the user will see (which is useful for deciding if a failure provides enough info) and (2) the stacktrace isn't that useful.

src/shared/sam/sync.ts
```diff
@@ -96,17 +101,27 @@ function createStackPrompter(client: DefaultCloudFormationClient) {
...
         title: 'Select a CloudFormation Stack',
-        placeholder: 'Filter or enter a new stack name',
-        filterBoxInputSettings: {
-            label: 'Create a New Stack',
-            transform: v => v,
+        placeholder: 'Select a stack',
+        // filterBoxInputSettings: {
+        //     label: 'Create a New Stack',
+        //     transform: v => v,
+        // },
```
justinmk3, line 108: @JadenSimon  what is this doing? Is it ok to comment-out these pass-through `filterBoxInputSettings` fields for now? Seems like `noItemsFoundItem` does not work if they are present.

src/shared/sam/sync.ts
```diff
@@ -96,17 +101,27 @@ function createStackPrompter(client: DefaultCloudFormationClient) {
...
+        buttons: createCommonButtons(samSyncUrl, consoleUrl),
+        noItemsFoundItem: {
+            label: localize(
+                'aws.cfn.noStacks',
+                'No stacks in region "{0}". Click to visit CloudFormation console.',
+                client.regionCode
+            ),
+            data: undefined,
+            onClick: () => vscode.env.openExternal(consoleUrl),
```
justinmk3, line 117: Will do a followup PR to add a openExternal() util that adds telemetry.

src/shared/awsConsole.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+import * as vscode from 'vscode'
+
+export function getAwsConsoleUrl(service: 'ecr' | 'cloudformation', region: string): vscode.Uri {
+    switch (service) {
+        case 'ecr':
+            return vscode.Uri.parse(`https://${region}.console.aws.amazon.com/ecr/repositories?region=${region}`)
+        case 'cloudformation':
+            return vscode.Uri.parse(`https://${region}.console.aws.amazon.com/cloudformation/home?region=${region}`)
```
justinmk3, line 13: Not an attempt to "deep link". Just an attempt to provide something useful if there are no CFN stacks or ECR repos.

src/shared/logger/winstonToolkitLogger.ts
JadenSimon: Yeah the native stack trace also doesn't show all exceptions which makes it even less useful. Hopefully that won't be as much of an issue when 'cause' is added to native ES exceptions.

src/shared/sam/sync.ts
JadenSimon: I could never come up with a good name for `filterBoxInputSettings` but it's intent is to create an item from the user's input. So in this case the user can choose to create a new stack from their input if one isn't already available in the picker.

jparkhurst3:
Should the string in the first GIF ("No SAM template.yaml file(s) found. Click for help") mirror the "Click to visit [site]" strings in the other two changes?

Optional suggestion: We could use "Select" instead of "Click" to cover click and enter as the interaction mechanism.

justinmk3:
see updated screenshots

src/shared/ui/pickerPrompter.ts
```diff
@@ -137,7 +137,7 @@ export function createQuickPick<T>(
 
     const prompter =
         mergedOptions.filterBoxInputSettings !== undefined
-            ? new FilterBoxQuickPickPrompter<T>(picker, mergedOptions.filterBoxInputSettings)
+            ? new FilterBoxQuickPickPrompter<T>(picker, mergedOptions)
             : new QuickPickPrompter<T>(picker, mergedOptions)
```
justinmk3, line 141: @JadenSimon was this intentional? In order to show `noItemsFoundItem` simultaneously with `filterBoxInputSettings` I changed `FilterBoxQuickPickPrompter` to pass the full `options` to its baseclass.

PR #3220, opened 2023-03-08
feat(auth): UX improvements for expired connections

JadenSimon:
## Problem
* It's not always clear how to re-authenticate expired connections, especially when using multiple at the same time
* The "Switch Connections" menu doesn't display much info about the status of the connection

## Solution
* Update the "Switch Connections" picker to:
  * Add a description to expired/invalid connections
  * Show which tools (if any) a connection is attached to
  * Re-authenticate expired connections instead of switching to them
* Revamp the status bar to account for multiple connection scenarios

## License


src/shared/telemetry/spans.ts
```diff
@@ -162,6 +163,7 @@ export class TelemetrySpan {
 interface TelemetryContext {
     readonly spans: TelemetrySpan[]
     readonly activeSpan?: TelemetrySpan
+    readonly attributes?: Partial<MetricShapes[MetricName]>
```
JadenSimon, line 166: Something I've been meaning to add. I came across a use-case here so I just added it now.

src/credentials/wizards/connections.ts
```diff
@@ -0,0 +1,106 @@
...
+import { telemetry } from '../../shared/telemetry/telemetry'
+import { createRefreshButton, createExitButton } from '../../shared/ui/buttons'
+import { createQuickPick, DataQuickPickItem } from '../../shared/ui/pickerPrompter'
+import { Auth, Connection, isIamConnection, reauthCommand } from '../auth'
+import { getDependentAuths } from '../secondaryAuth'
+
+export const getConnectionIcon = (conn: Connection) =>
+    conn.type === 'sso' ? getIcon('vscode-account') : getIcon('vscode-key')
+
+export function createConnectionPrompter(auth: Auth, type?: 'iam' | 'sso') {
```
JadenSimon, line 19: Had to split this out due to potential circular dependency issues. `secondaryAuth.ts` could possibly be consolidated into `auth.ts` to simplify things.

src/credentials/wizards/connections.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    const prompter = createQuickPick(loadItems(), {
+        placeholder,
+        title: localize('aws.auth.promptConnection.title', 'Switch Connection'),
+        buttons: [refreshButton, createExitButton()],
+    })
+
+    return prompter
+
+    async function loadItems(): Promise<DataQuickPickItem<Connection | 'addNewConnection' | 'editCredentials'>[]> {
```
JadenSimon, line 36: Placing function declarations _after_ the return statement of the outer scope is common in some codebases. Not sure if I like it or not though.

src/test/credentials/auth.test.ts
```diff
@@ -14,7 +22,7 @@ import { ToolkitError } from '../../shared/errors'
 import { FakeMemento } from '../fakeExtensionContext'
 import { assertTreeItem } from '../shared/treeview/testUtil'
 import { getTestWindow } from '../shared/vscode/window'
-import { captureEvent } from '../testUtil'
+import { assertTelemetry, captureEvent, captureEventOnce } from '../testUtil'
 import { stub } from '../utilities/stubber'
 
 function createSsoProfile(props?: Partial<Omit<SsoProfile, 'type'>>): SsoProfile {
```
JadenSimon, line 28: TODO: extract some of these functions out so they can be used elsewhere.

src/credentials/statusBarItem.ts
```diff
@@ -49,26 +52,40 @@ function handleDevSettings(devSettings: DevSettings, statusBarItem: vscode.Statu
...
 }
 
-// Resolves when the status bar reaches its final state
-export async function updateCredentialsStatusBarItem(
-    statusBarItem: vscode.StatusBarItem,
-    credentialsId?: string,
-    color?: vscode.ThemeColor
-): Promise<void> {
-    const connectedMsg = localize(
+function updateItem(statusBarItem: vscode.StatusBarItem): void {
```
JadenSimon, line 55: I deleted the old test file as it really didn't do much. Need to extract some `Auth` test util functions before it's worth writing tests for this.

src/credentials/wizards/connections.ts
JadenSimon: Nevermind. Splitting up the file isn't a huge amount of work but it does require moving around a lot of stuff which I'd rather do in a separate PR.

src/webviews/components/settingsPanel.vue
```diff
@@ -141,6 +141,7 @@ input[type='checkbox'] ~ label .collapse-button {
     color: var(--vscode-foreground);
     transition: transform 0.5s;
     transform: rotate(180deg);
+    text-align: right;
```
JadenSimon, line 144: Little bug I noticed. Same with the icons/css changes.

justinmk3 (approved):

justinmk3:
Is the attributes idea discussed more somewhere else?

JadenSimon:
> Is the attributes idea discussed more somewhere else?

No, I'll have to write something up. The goal is to generalize what we already do for "common metadata" i.e. attaching extra data to all future actions in the current context.

PR #3227, opened 2023-03-13
fix(codewhisperer): update the total recommendation count in visual cue

zixlin7:
## Problem
the total number of recommendations on the visual cue is not correct due to pagination
## Solution
mitigation to allow the total index to be correct while the long-term fix rely on inline api's support of pagination

https://user-images.githubusercontent.com/60411978/224853441-ecf21063-0f84-4752-934c-03a2593b0e0c.mov

## License

### DIFF
--- a/src/codewhisperer/service/inlineCompletionService.ts
+++ b/src/codewhisperer/service/inlineCompletionService.ts
@@ -334,6 +334,7 @@ export class InlineCompletionService {
             return
         }
         if (this.isSuggestionVisible()) {
+            // to force refresh the visual cue so that the total recommendation count can be updated
             const index = this.inlineCompletionProvider?.getActiveItemIndex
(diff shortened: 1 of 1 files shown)


src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -326,7 +330,12 @@ export class InlineCompletionService {
 
     async tryShowRecommendation() {
         const editor = vscode.window.activeTextEditor
-        if (this.isSuggestionVisible() || editor === undefined) {
+        if (editor === undefined) {
```
leigaol, line 333: Please add some doc string to explain what this change does in this function.

justinmk3 (approved):

Will-ShaoHua (approved):

leigaol (approved):

PR #3232, opened 2023-03-14
auth: prompt existing builder id signout on new builder id

nkomonen-amazon:
# Problem:

When a user is already signed in to their Builder ID, but they want to use a different one, they are not able to use a different ID.

This is confusing since nothing happens when they try to add another one.

# Solution:

When attempting to add a new Builder ID but one already exists, this will prompt the user to choose if they wish to sign out of the existing one, and will then complete the builder id verification process.

# Demo
Kapture 2023-03-14 at 15 21 24

## License

### DIFF
--- a/src/credentials/auth.ts
+++ b/src/credentials/auth.ts
@@ -770,9 +770,7 @@ const switchConnections = Commands.register('aws.auth.switchConnections', (auth:
     }
 })
 
-async function signout(auth: Auth) {
-    const conn = auth.activeConnection
-
+async function signout(auth: Auth, conn: Connection | undefined = auth.activeConnection) {
     if (conn?.type === 'sso') {
         // TODO: does deleting the connection make sense UX-wise?
         // this makes it disappear from the list of available connections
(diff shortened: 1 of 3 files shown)


justinmk3:
I would prefix this as `feat` not `chore` :) (In fact I'd say there's never a reason to label "chore": it's usually a `fix`, a `feat`, a `build`, or `docs`).

I'd also say this change is worth calling out in the changelog, so:

    npm run newChange

src/credentials/auth.ts
```diff
@@ -902,15 +902,64 @@ export async function createStartUrlPrompter(title: string, ignoreScopes = true)
...
+    const builderIdConn = (await auth.listConnections()).find(isBuilderIdConnection)
+
+    if (!builderIdConn) {
+        return undefined
+    }
+
+    const items: DataQuickPickItem<string>[] = [
+        {
+            data: 'signout',
+            label: `You already signed in with an ${getIdeProperties().company} Builder ID. Sign out to add another?`,
```
justinmk3, line 933: Unfortunately the builder-id doesn't have a user-facing "name" , else it would be good to show here too.

```suggestion
            label: `Currently signed in with ${getIdeProperties().company} Builder ID. Sign out to add another?`,
```

justinmk3 (approved):

src/credentials/auth.ts
```diff
@@ -809,8 +809,8 @@ const switchConnections = Commands.register('aws.auth.switchConnections', (auth:
     }
 })
 
-async function signout(auth: Auth) {
-    const conn = auth.activeConnection
+async function signout(auth: Auth, conn?: Connection) {
+    conn = conn ?? auth.activeConnection
```
JadenSimon, line 813: nit:
```suggestion
async function signout(auth: Auth, conn = auth.activeConnection) {
```

src/credentials/auth.ts
```diff
@@ -902,15 +902,64 @@ export async function createStartUrlPrompter(title: string, ignoreScopes = true)
...
+ *          in the first place.
+ */
+async function promptLogoutExistingBuilderIdConnection(auth: Auth): Promise<SsoConnection | undefined> {
+    const builderIdConn = (await auth.listConnections()).find(isBuilderIdConnection)
+
+    if (!builderIdConn) {
+        return undefined
+    }
+
+    const items: DataQuickPickItem<string>[] = [
```
JadenSimon, line 930: nit:
```suggestion
    const items: DataQuickPickItem<'signout' | 'cancel'>[] = [
```

TS might yell at you with this change. If so, do this to relevant fields:
```ts
data: 'signout' as const,
```

src/credentials/auth.ts
```diff
@@ -902,15 +902,64 @@ export async function createStartUrlPrompter(title: string, ignoreScopes = true)
...
+
+    switch (resp) {
+        case 'signout': {
+            await signout(auth, builderIdConn)
+
+            // Sanity check signout worked
+            const sanityCheckConn = (await auth.listConnections()).find(isBuilderIdConnection)
+            if (sanityCheckConn !== undefined) {
+                throw new Error('Failed to sign out of Builder ID.')
+            }
```
JadenSimon, line 953: This is nice for development/debugging but it makes the assumption that only 1 builder ID connection can exist. I would either remove this _or_ use a more precise search such as checking the `id`.

src/credentials/auth.ts
```diff
@@ -902,15 +902,64 @@ export async function createStartUrlPrompter(title: string, ignoreScopes = true)
...
+ *
+ *          Otherwise, undefined, due to user selecting logout or the connection never existing
+ *          in the first place.
+ */
+async function promptLogoutExistingBuilderIdConnection(auth: Auth): Promise<SsoConnection | undefined> {
+    const builderIdConn = (await auth.listConnections()).find(isBuilderIdConnection)
+
+    if (!builderIdConn) {
+        return undefined
+    }
```
JadenSimon, line 928: This would flow better if we checked prior to calling `promptLogoutExistingBuilderIdConnection`. Example:
```ts
    const existingConn = (await auth.listConnections()).find(isBuilderIdConnection)
    if (existingConn) {
        await promptLogoutExistingBuilderIdConnection(auth, existingConn)
    }
```

Then we can delete these three lines as the connection is passed in instead.

src/credentials/auth.ts
```diff
@@ -902,15 +902,64 @@ export async function createStartUrlPrompter(title: string, ignoreScopes = true)
...
+            const sanityCheckConn = (await auth.listConnections()).find(isBuilderIdConnection)
+            if (sanityCheckConn !== undefined) {
+                throw new Error('Failed to sign out of Builder ID.')
+            }
+
+            return undefined
+        }
+        default: {
+            return builderIdConn
+        }
```
JadenSimon, line 959: Throwing an error is more explicit. Returning the connection causes an error to get thrown later, might as well fail early with more appropriate semantics:
```suggestion
        default: {
            throw new CancellationError('user')
        }
```

JadenSimon (approved):

src/credentials/auth.ts
nkomonen-amazon: Maybe I'm misunderstanding but only the function `createBuilderIdConnection()` calls this function. Returning the `builderIdConn` is intentional since `createBuilderIdConnection()` sometimes expects a connection from the call.

nkomonen-amazon:
Don't merge yet, I came across a bug with the recent change

PR #3239, opened 2023-03-16
fix(codewhisperer): bug with right context handling

zixlin7:
## Problem
right context handling currently don't take the typeahead into account, which result in some cases are not handled:
Screenshot 2023-03-16 at 9 01 10 AM
## Solution
update the start position for right context to be the cursor position and fall back to invocation position
Screenshot 2023-03-16 at 9 02 42 AM

## License

### DIFF
--- a/src/codewhisperer/service/inlineCompletionService.ts
+++ b/src/codewhisperer/service/inlineCompletionService.ts
@@ -89,7 +89,7 @@ class CWInlineCompletionItemProvider implements vscode.InlineCompletionItemProvi
 
     truncateOverlapWithRightContext(document: vscode.TextDocument, suggestion: string): string {
         let rightContextRange: vscode.Range | undefined = undefined
-        const pos = RecommendationHandler.instance.startPos
+        const pos = vscode.window.activeTextEditor?.selection.active || RecommendationHandler.instance.startPos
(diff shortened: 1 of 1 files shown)


leigaol (approved):

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -89,7 +89,7 @@ class CWInlineCompletionItemProvider implements vscode.InlineCompletionItemProvi
 
     truncateOverlapWithRightContext(document: vscode.TextDocument, suggestion: string): string {
         let rightContextRange: vscode.Range | undefined = undefined
-        const pos = RecommendationHandler.instance.startPos
+        const pos = vscode.window.activeTextEditor?.selection.active || RecommendationHandler.instance.startPos
```
leigaol, line 92: Does the `suggestion: string` contain typeahead?

zixlin7: yes in this case, it's usually when user type a html tag like `<button>` or quotes and the IDE filled in the closing tag or quotes

justinmk3 (approved):

justinmk3:
Is this something worth mentioning in the changelog?

PR #3241, opened 2023-03-17
docs: marketplace page

justinmk3:
## Todo

- it looks like gitlens uses external URLs in https://raw.githubusercontent.com/gitkraken/vscode-gitlens/main/README.md
    - Example: ``
    - We could do this to (greatly) reduce the size of the VSIX.
    - Edit: confirmed that this works.
- Looks like `.webm` does _not_ work :( , so we must use `.gif`
    - https://github.com/microsoft/vscode/issues/32540#issuecomment-381847598
      > we strip out ffmpeg and other media support from our release builds. This is by-design and will not change ... many types of media playback will never be supported in webviews inside vscode.

## License


justinmk3:
ECS run command / terminal :

[ecs-terminal.webm](https://user-images.githubusercontent.com/55561878/225943531-ed087ef9-21e7-4b2f-a26e-ce8f2ec7f82a.webm)

JadenSimon (approved):

JadenSimon:
> Actually, I just learned that we strip out ffmpeg and other media support from our release builds.

I wonder why, maybe to keep things lean?

PR #3243, opened 2023-03-17
feat(IDE-9298): Dev Env Project Quick Pick selector

nkomonen-amazon:
## Problem:

When using the Create Dev Env webview we used
a drop down menu to provide all possible Projects.

The problem is that querying all Projects takes time and nothing would show on the screen for a while.

Additionally, if the user had many projects from
multiple spaces it was not easy to find

## Solution:

Instead of a dropdown menu use an async quick pick which will continuously load new Projects as they
are returned in the the paginated query for all Projects.

- Project loading is started immediately upon the webview loading. This will provide Projects quickly when the quick pick menu is opened
- Along with using a quick pick, the user can easily filter the Projects using the quick picks built in filtering mechanism.
- If the user specifically edits a Project, only Projects from the current space will appear in the quick pick

Kapture 2023-03-17 at 16 06 14

## License

### DIFF
--- a/src/codecatalyst/vue/create/backend.ts
+++ b/src/codecatalyst/vue/create/backend.ts
@@ -32,6 +32,8 @@ import { isCloud9 } from '../../../shared/extensionUtilities'
 import { telemetry } from '../../../shared/telemetry/telemetry'
 import { isNonNullable } from '../../../shared/utilities/tsUtils'
 import { recordSource } from '../../utils'
+import { QuickPickPrompter } from '../../../shared/ui/pickerPrompter'
+import { createProjectPrompter } from '../../wizards/selectResource'
 
 interface LinkedResponse {
     readonly type: 'linked'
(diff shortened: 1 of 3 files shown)


justinmk3:
Can we have a single flat list instead of having to select org + project?

nkomonen-amazon:
@justinmk3 

This was the design that was decided in a quick sync between myself and Jo.

Though I added some additional functionality now when the user edits an existing Project, it will only show Projects from the specific space. See the example in the GIF.

src/codecatalyst/vue/create/backend.ts
```diff
@@ -57,15 +62,74 @@ export class CodeCatalystCreateWebview extends VueWebview {
...
+            return projects.map(project => ({ label: `${project.org.name}/${project.name}`, data: project }))
+        })
+
+        // Note that the fetching of Projects is **started**
+        // immediately upon **creation** of the prompter.
+        const projectPrompter = createQuickPick<CodeCatalystProject>(projectQuickPickItems, {
+            placeholder: 'Select Space/Project',
+            buttons: [createExitButton()],
+        })
+        projectPrompter.quickPick.title = 'Select Space/Project'
```
JadenSimon, line 131: Can we re-use `createProjectPrompter` here?

nkomonen-amazon: Good call, was able to get rid of this entire function with this.

justinmk3:
Nice performance improvement! I think I noticed a bug: the Space selector shows all projects, not just spaces.

PR #3246, opened 2023-03-17
fix(codecatalyst): fix CC refresh button on some pickers

JadenSimon:
## Problem
https://github.com/aws/aws-toolkit-vscode/pull/3234 tried to make things more ergonomic by converting async iterables into async collections implicitly. This does work but unfortunately it consumes the iterable and thus cannot be re-used for things like refreshing

## Solution
* Remove the implicit conversion
* Add regression test

## License


justinmk3 (approved):

JadenSimon:
Oops I accidentally deleted the upstream branch instead of my local one.

PR #3247, opened 2023-03-17
test(integ): assert telemetry list

justinmk3:
Problem:
Checking a _sequence_ of metrics is awkward, leading to problems like: 9be5bbfcc11e.

Solution:
Enhance assertTelemetry to support checking a sequence.

## License

### DIFF
--- a/src/integrationTest/schema/schema.test.ts
+++ b/src/integrationTest/schema/schema.test.ts
@@ -3,7 +3,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import globals from '../../shared/extensionGlobals'
 import { getDefaultSchemas, samAndCfnSchemaUrl } from '../../shared/schemas'
 import {
     getCITestSchemas,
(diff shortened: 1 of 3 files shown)


src/test/testUtil.ts
```diff
@@ -146,46 +146,69 @@ export function getMetrics<K extends MetricName>(
...
+                return true
+            })
+            assert.ok(
+                found >= 0,
+                `telemetry item ${i + 1} (of ${
+                    expectedList.length
+                }) not found (in the expected order) for metric name: "${name}" `
+            )
+            // Remove all items up to the found item, to ensure we are checking the expected _order_.
+            metadata = metadata.splice(found, metadata.length - 1)
```
justinmk3, line 204: Checking "non-continguous order" is something generalizable (maybe already is). This is a hacky way to do it for now, but at least the interface is here for use in tests.

src/test/testUtil.ts
```diff
@@ -146,46 +146,69 @@ export function getMetrics<K extends MetricName>(
...
 export function assertTelemetry<K extends MetricName>(
     name: K,
-    expected: MetricShapes[MetricName],
-    eventNum?: number
+    expected: MetricShapes[MetricName] | MetricShapes[MetricName][]
 ): void | never
 export function assertTelemetry<K extends MetricName>(
     name: K,
-    expected: MetricShapes[K],
-    eventNum: number = 0
```
justinmk3, line 168: Eliminate `eventNum`. Now the tests can pass a list to assert a particular (non-contiguous) metrics order.

src/integrationTest/schema/schema.test.ts
```diff
@@ -65,47 +64,20 @@ describe('Sam Schema Regression', function () {
...
-        )
-        // Schema is retrieved from cache on subsequent retrieval
-        assertTelemetry(
-            'toolkit_getExternalResource',
-            { url: samAndCfnSchemaUrl, passive: true, result: 'Cancelled', reason: 'Cache hit' },
-            1
-        )
-    })
-
-    it('uses cache for all requests on second function call', async () => {
```
justinmk3, line 90: Don't need this 2nd test because we can test metrics order in one test.

However, a future TODO is to delete the cached item in globalStorage so that the test works better for local development (not only CI).

src/test/testUtil.ts
justinmk3: `metadata` is copied though, so no harm in mutating it? Will revisit later, want to unblock this PR for now.

PR #3252, opened 2023-03-20
fix(CoCa): do not show 3p repos

nkomonen-amazon:
## Problem:

When a user is selecting a repository to clone we
do not want to give users the option to select
a third party repo (github) to clone

## Solution:

Do not show them third party repos when selecting.

Fixes IDE-10273

---

In one of my projects I have a Github repo + Internal repo. Once I select 'Clone Repository` only the internal repo will appear. Before both the Github + internal repo would appear.

Kapture 2023-03-20 at 16 03 38

## License


src/shared/clients/codecatalystClient.ts
```diff
@@ -411,12 +413,45 @@ class CodeCatalystClientInternal {
 
     /**
      * Gets a flat list of all repos for the given CodeCatalyst user.
+     * @param internalOnly True if you want only internal repositories to be returned.
      */
     public listSourceRepositories(
-        request: CodeCatalyst.ListSourceRepositoriesRequest
+        request: CodeCatalyst.ListSourceRepositoriesRequest,
+        internalOnly: boolean = false
```
nkomonen-amazon, line 420: This is the core of the change. When listing repositories the user has the option to indicate they only want internal repos.

src/shared/clients/codecatalystClient.ts
```diff
@@ -411,12 +413,45 @@ class CodeCatalystClientInternal {
 
     /**
      * Gets a flat list of all repos for the given CodeCatalyst user.
+     * @param internalOnly True if you want only internal repositories to be returned.
```
justinmk3, line 416:
```suggestion
     * @param internalOnly Get only internal (non-3P) repositories.
```

src/shared/clients/codecatalystClient.ts
```diff
@@ -429,6 +464,18 @@ class CodeCatalystClientInternal {
         )
     }
 
+    /**
+     * Determines if a repo is internal versus non-internal.
```
justinmk3, line 468:
```suggestion
     * Determines if a repo is internal (3P) versus non-internal.
```

src/testE2E/codecatalyst/client.test.ts
```diff
@@ -279,6 +279,25 @@ describe('Test how this codebase uses the Code Catalyst API', function () {
         }
     })
 
+    describe('Repositories', function () {
+        it('can filter out non-internal repositories (eg: github)', async function () {
```
justinmk3, line 283: The filter logic is the non-trivial part, right? So this is an example of something that doesn't need end-to-end coverage. We just need to extract the filter logic out of `listSourceRepositories` and test the logic. Testing the call to `listSourceRepositories` is less interesting.

src/shared/clients/codecatalystClient.ts
```diff
@@ -445,12 +492,17 @@ class CodeCatalystClientInternal {
 
     /**
      * Lists ALL of the given resource in the current account
+     *
+     * @param internalOnly If true, only internal repos will be returned.
      */
     public listResources(resourceType: 'org'): AsyncCollection<CodeCatalystOrg[]>
     public listResources(resourceType: 'project'): AsyncCollection<CodeCatalystProject[]>
-    public listResources(resourceType: 'repo'): AsyncCollection<CodeCatalystRepo[]>
+    public listResources(resourceType: 'repo', internalOnly?: boolean): AsyncCollection<CodeCatalystRepo[]>
```
justinmk3, line 500: It's always `true`, do we need this param?

justinmk3: When do we show all repos?

src/shared/clients/codecatalystClient.ts
justinmk3: Resolved but the change wasn't made :P

nkomonen-amazon: I renamed from 'internal' to 'firstParty' since that seemed to be the right convention to use. But I updated the change to reflect your comment.

src/shared/clients/codecatalystClient.ts
nkomonen-amazon: Yeah, there are still places that want all repos. Plus its probably expected that you'd get all repos by default.

src/shared/clients/codecatalystClient.ts
```diff
@@ -445,12 +467,17 @@ class CodeCatalystClientInternal {
 
     /**
      * Lists ALL of the given resource in the current account
+     *
+     * @param firstPartOnly If true, only first party repos (code catalyst owned) will be returned.
```
justinmk3, line 471: CodeCatalyst is one word

```suggestion
     * @param firstPartOnly If true, only first party repos (CodeCatalyst owned) will be returned.
```

src/shared/clients/codecatalystClient.ts
```diff
@@ -676,3 +704,48 @@ class CodeCatalystClientInternal {
...
+                    : undefined
+            })
+        )
+    ).filter(item => item !== undefined) as CodeCatalyst.ListSourceRepositoriesItem[]
+}
+
+/**
+ * Determines if a repo is first party (1P) vs third party (3P).
+ *
+ * 1P is Code Catalyst, 3P is something like Github.
```
justinmk3, line 741:
```suggestion
 * 1P is CodeCatalyst, 3P is something like Github.
```

src/test/shared/clients/codeCatalystClient.test.ts
```diff
@@ -0,0 +1,53 @@
+/*!
```
justinmk3, line 1: Tests for the CC client already live in `./src/test/caws/codecatalystClient.test.ts`

src/shared/clients/codecatalystClient.ts
```diff
@@ -411,12 +417,28 @@ class CodeCatalystClientInternal {
 
     /**
      * Gets a flat list of all repos for the given CodeCatalyst user.
+     * @param firstPartyOnly Get only 1P repositories. 1P being repos that are
+     *                       created in CodeCatalyst.
      */
     public listSourceRepositories(
-        request: CodeCatalyst.ListSourceRepositoriesRequest
+        request: CodeCatalyst.ListSourceRepositoriesRequest,
+        firstPartyOnly: boolean = false
```
justinmk3, line 425: nit: is "third party" a more common term and more discoverable term than "first party" ?

```suggestion
        thirdparty: boolean = true
```

src/shared/clients/codecatalystClient.ts
```diff
@@ -676,3 +704,48 @@ class CodeCatalystClientInternal {
         return devenv
     }
 }
+
+/**
+ * Returns only the first-party repos from the given
+ * list of repository items.
+ */
+export async function getFirstPartyRepos(
```
justinmk3, line 712:
```suggestion
export async function excludeThirdpartyRepos(
```

src/shared/clients/codecatalystClient.ts
```diff
@@ -487,7 +514,8 @@ class CodeCatalystClientInternal {
         const r = await this.call(this.sdkClient.getSourceRepositoryCloneUrls(args), false)
```
justinmk3, line 514: Please update the docstring to:

    Gets the git source host URL for the given CodeCatalyst or third-party repo.

src/shared/clients/codecatalystClient.ts
```diff
@@ -676,3 +704,48 @@ class CodeCatalystClientInternal {
...
+/**
+ * Determines if a repo is first party (1P) vs third party (3P).
+ *
+ * 1P is Code Catalyst, 3P is something like Github.
+ */
+async function isfirstPartyRepo(
+    client: CodeCatalystClient,
+    codeCatalystRepo: GetSourceRepositoryCloneUrlsRequest
+): Promise<boolean> {
+    const url = await client.getRepoCloneUrl(codeCatalystRepo)
```
justinmk3, line 747: 👀  This is a O(N) performance cost for all customers, even if they have _zero_ third-party repos, right?

src/shared/clients/codecatalystClient.ts
nkomonen-amazon: Makes sense to me, refactored

justinmk3 (approved):

PR #3277, opened 2023-03-27
fix(s3): file system provider is not case-sensitive

### LINKED ISSUE #3276: S3 "Edit" feature treats paths as case-insensitive
## Problem
I have 2 folders in my S3 bucket (Test and test). Both these folders have a file main.json each. By default the folders are sorted such that "Test" bucket is above "test" bucket. When I click the edit icon for test/main.json it opens Test/main.json which is a different file.

My folder and file structure inside s3 (notice difference in case for first letter in folder names) -

**bucket1/Test/main.json
bucket1/test/main.json**

1. Create a bucket in S3.
2. Create 2 folders with same name with different case for first letter. (Test and test)
3. Add 2 files main.json in both the folder with different content inside both these files.
4. Try to edit **bucket1/test/main.json** but it opens **bucket1/Test/main.json** instead.
 
## Expected behavior

If I try to open or edit **bucket1/test/main.json** then it should open that file only and not the file from **bucket1/Test/main.json**

## System details (run the `AWS: About Toolkit` command)

OS: Windows_NT x64 10.0.22621
Visual Studio Code extension host:  1.76.2
AWS Toolkit:  1.66.0
node: 16.14.2
electron: 19.1.11

nkomonen-amazon:
## Problem:

When the s3 file system provider was registered it was not set to be case sensitive. So paths like
`/aws/Test/temp.json` and `/aws/test/temp.json` were
seen as the same thing.

## Solution:

Set the case sensitive flag when registering the
file system provider so that it will uniquely
identify files with the same text but different
case sensitivity.

Fixes #3276 

### Additional Info

Before setting case sensitivity, this line would return an existing text document instead of a new one (that had a different case). It was the hint that VSCode was handling it different from our expectations.
https://github.com/aws/aws-toolkit-vscode/blob/3a22329dbc6d1eb819a65ea6ef84599f24d2cb48/src/s3/fileViewerManager.ts#L171

## License

### DIFF
--- a/src/s3/activation.ts
+++ b/src/s3/activation.ts
@@ -39,8 +39,8 @@ export async function activate(ctx: ExtContext): Promise<void> {
 
     ctx.extensionContext.subscriptions.push(manager)
     ctx.extensionContext.subscriptions.push(
-        vscode.workspace.registerFileSystemProvider(s3EditScheme, fs),
-        vscode.workspace.registerFileSystemProvider(s3ReadScheme, fs, { isReadonly: true }),
+        vscode.workspace.registerFileSystemProvider(s3EditScheme, fs, { isCaseSensitive: true }),
+        vscode.workspace.registerFileSystemProvider(s3ReadScheme, fs, { isReadonly: true, isCaseSensitive: true }),
         Commands.register('aws.s3.copyPath', async (node: S3FolderNode | S3FileNode) => {
             await copyPathCommand(node)
         }),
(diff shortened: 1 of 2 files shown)


nkomonen-amazon:
> A basic regression test for this would be great if possible

@JadenSimon did you have something in mind? An integration test?

justinmk3:
> An integration test?

No, tests for the s3 fs provider are here: https://github.com/aws/aws-toolkit-vscode/blob/3a22329dbc6d1eb819a65ea6ef84599f24d2cb48/src/test/s3/util/fileViewerManager.test.ts#L90

PR #3285, opened 2023-03-28
feat(cwl): Remember previous selection for search/filter

nkomonen-amazon:
## Problem:

To search a CWL log group you must enter a date
range, and an optional string to filter the
event text.

Once openend there are buttons in the editor/title that are the calendar and magnifying glass.

If you click on one of these you can re-adjust their value through a quickpick/input prompter.

The problem is that users cannot see what they previously selected, it is not that good for UX

## Solution:

When the user attempts to re-adjust their previous selection, autofill the previous values in the prompter so they can see what they previously chose.

### Before 
BeforeCwl

### After
Notice how `Last 12 Hours` and `Hello` are autofilled once I go to re-adjust the values.

AfterCwl

Fixes IDE-8248

## License

### DIFF
--- a/src/cloudWatchLogs/timeFilterSubmenu.ts
+++ b/src/cloudWatchLogs/timeFilterSubmenu.ts
@@ -5,11 +5,12 @@
 
 import { Prompter, PromptResult } from '../shared/ui/prompter'
 import { InputBoxPrompter } from '../shared/ui/inputPrompter'
-import { ItemLoadTypes, QuickPickPrompter, DataQuickPickItem, createQuickPick } from '../shared/ui/pickerPrompter'
+import { QuickPickPrompter, DataQuickPickItem, createQuickPick } from '../shared/ui/pickerPrompter'
 import { createInputBox } from '../shared/ui/inputPrompter'
 import { isValidResponse, StepEstimator } from '../shared/wizards/wizard'
 import { createCommonButtons } from '../shared/ui/buttons'
 import * as nls from 'vscode-nls'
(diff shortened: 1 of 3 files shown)


src/cloudWatchLogs/timeFilterSubmenu.ts
```diff
@@ -107,7 +132,7 @@ export class TimeFilterSubmenu extends Prompter<TimeFilterResponse> {
 
                         return { start: startTime.valueOf(), end: endTime.valueOf() }
                     } else {
-                        return resp
+                        return undefined
```
nkomonen-amazon, line 135: This scenario triggers if the user cancels/exits. So we do not want to return anything, even the previous value since it triggers a reload.

src/cloudWatchLogs/timeFilterSubmenu.ts
```diff
@@ -163,6 +188,27 @@ export class TimeFilterSubmenu extends Prompter<TimeFilterResponse> {
         return [new Date(parts[0]), new Date(parts[1])]
     }
 
+    /**
+     * Formats a given start and end time in milliseconds to
+     * a string in the format 'YYYY/MM/DD-YYYY/MM/DD'
```
justinmk3, line 193: This "custom" case is the only datetime value that should be pre-filled. I don't think we should pre-fill the "relative" times because they are just a button-click.

nkomonen-amazon: Sounds good, updated so it only does stuff with the ranges

justinmk3 (approved):

PR #3303, opened 2023-04-02
fix(codecatalyst): disallow non-vscode Dev Environment

justinmk3:
Problem:
If customer changed the "IDE runtime" (in the CodeCatalyst console), AWS Toolkit will attempt to reconnect to the Dev Environment, which is inconsistent with other IDEs (jetbrains, cloud9).

Solution:
Check the `ides` field on connect. Disconnect if the ide runtime is not "vscode".

## License

### DIFF
--- a/src/shared/clients/devenvClient.ts
+++ b/src/shared/clients/devenvClient.ts
@@ -38,11 +38,11 @@ export class DevEnvClient implements vscode.Disposable {
                 if (this.lastStatus !== r.status) {
                     const newStatus = r.status ?? 'NULL'
                     getLogger().info(
-                        'codecatalyst: DevEnvClient: status change (old=%s new=%s) action=%s: %s',
+                        'codecatalyst: DevEnvClient: status change (old=%s new=%s)%s%s',
                         this.lastStatus,
                         newStatus,
-                        r.actionId ?? '?',
-                        r.message ?? '(no message)'
+                        r.actionId ? ` action=${r.actionId}` : '',
+                        r.message ? `: "${r.message}"` : ''
                     )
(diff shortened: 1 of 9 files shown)


src/shared/clients/devenvClient.ts
```diff
@@ -3,14 +3,65 @@
...
+            getLogger().debug('codecatalyst: DevEnvClient skipped (local)')
+            this.timer = undefined
+        } else {
+            getLogger().info('codecatalyst: DevEnvClient started')
+            this.timer = globals.clock.setInterval(async () => {
+                const r = await this.getStatus()
+                if (this.lastStatus !== r.status) {
+                    const newStatus = r.status ?? 'NULL'
+                    getLogger().info(
+                        'codecatalyst: DevEnvClient: status change (old=%s new=%s) action=%s: %s',
```
justinmk3, line 41: Currently this just logs status changes. Later could be used to handle `IMAGES-UPDATE-AVAILABLE`.

src/shared/clients/devenvClient.ts
```diff
@@ -3,14 +3,65 @@
...
+    private lastStatus = ''
+    private onStatusChangeFn: undefined | ((oldStatus: string, newStatus: string) => void)
+
+    /** Singleton instance (to avoid multiple polling workers). */
+    public static get instance() {
+        return (this.#instance ??= new this())
+    }
+
+    /** @internal */
+    public constructor(private readonly endpoint: string = environmentEndpoint) {
```
justinmk3, line 30: should be private but currently used in tests.

JadenSimon (approved):

PR #3306, opened 2023-04-03
fix(codecatalyst): No solution for 'connection is invalid or expired'

nkomonen-amazon:
## Problem:

When using an MDE + CodeCatalyst, if the BuilderId connection became expired, on attempts to use the CodeCatalyst client users would get an error 'Connection is invalid...' with no obvious way to resolve it.

## Solution:

Provide the user the ability to re-authenticate their BuilderId.

This commit verifies the BuilderId connection is valid before setting up the CodeCatalyst client. If it is not valid, users will be prompted to login again. They can click the login button on a pop up window, and be redirected through the BuilderId authentication process again.

Screen Shot 2023-04-03 at 6 12 05 PM

## License

### DIFF
--- a/src/codecatalyst/commands.ts
+++ b/src/codecatalyst/commands.ts
@@ -18,11 +18,12 @@ import { DevEnvironmentId, getConnectedDevEnv, openDevEnv } from './model'
 import { showConfigureDevEnv } from './vue/configure/backend'
 import { showCreateDevEnv } from './vue/create/backend'
 import { CancellationError } from '../shared/utilities/timeoutUtils'
-import { ToolkitError } from '../shared/errors'
+import { ToolkitError, errorCode } from '../shared/errors'
 import { telemetry } from '../shared/telemetry/telemetry'
 import { showConfirmationMessage } from '../shared/utilities/messages'
 import { AccountStatus } from '../shared/telemetry/telemetryClient'
 import { CreateDevEnvironmentRequest, UpdateDevEnvironmentRequest } from 'aws-sdk/clients/codecatalyst'
+import { Auth, SsoConnection } from '../credentials/auth'
 
 /** "List CodeCatalyst Commands" command. */
 export async function listCommands(): Promise<void> {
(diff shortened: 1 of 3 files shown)


JadenSimon (approved):

src/codecatalyst/commands.ts
```diff
@@ -135,13 +136,43 @@ function createClientInjector(authProvider: CodeCatalystAuthenticationProvider):
...
+
+    if (result !== 'Login') {
+        throw new ToolkitError('User cancelled login.', { cancelled: true, code: invalidConnection })
+    }
+
+    conn = await auth.reauthenticate(conn)
+
+    // Log in attempt failed
+    if (auth.getConnectionState(conn) !== 'valid') {
+        throw new ToolkitError('Connection could not be authenticated.', { code: invalidConnection })
```
justinmk3, line 170: this is user-facing, let's use more intuitive wording. "Failed to connect", or "Connection failed", or...

```suggestion
        throw new ToolkitError('Could not connect.', { code: invalidConnection })
```

src/credentials/auth.ts
```diff
@@ -301,6 +301,8 @@ interface ConnectionStateChangeEvent {
     readonly state: ProfileMetadata['connectionState']
 }
 
+export const invalidConnection = 'InvalidConnection'
```
justinmk3, line 304: should this live in errors.ts ?

src/codecatalyst/commands.ts
```diff
@@ -135,13 +136,43 @@ function createClientInjector(authProvider: CodeCatalystAuthenticationProvider):
...
     }
 }
 
+/**
+ * Returns a connection that is ensured to be authenticated.
+ *
+ * Provides the user the ability to re-authenticate if needed,
+ * otherwise throwing an error.
+ */
+async function authenticateConnection(conn: SsoConnection, auth: Auth): Promise<SsoConnection> {
```
justinmk3, line 153: elsewhere we have called this "validate connection", is it intentional that we are using different jargon here? i.e. are the semantics different?
```suggestion
async function validateConnection(conn: SsoConnection, auth: Auth): Promise<SsoConnection> {
```

src/codecatalyst/commands.ts
```diff
@@ -135,13 +136,43 @@ function createClientInjector(authProvider: CodeCatalystAuthenticationProvider):
 
         await authProvider.restore()
         const conn = authProvider.activeConnection ?? (await authProvider.promptNotConnected())
-        const client = await createClient(conn)
+        const authenticatedConn = await authenticateConnection(conn, authProvider.auth)
+        const client = await createClient(authenticatedConn)
```
justinmk3, line 140: This implies that `authenticateConnection` either throws or always succeeds. Should we do this in `createClient` itself then? Maybe this is out of scope of for this PR. It seems like `createClientInjector` might be applicable beyond just "commands.ts".

src/codecatalyst/commands.ts
nkomonen-amazon: From my understanding we have a 'valid' connection since we know it has valid properties (eg: startUrl), but due to the connection expiring we need to authenticate the existing connection with the server.

Successfully authenticating ourselves implies our connection properties were valid??

src/credentials/auth.ts
nkomonen-amazon: Should I make an Enum in `errors.ts`?

```
enum ErrorCodes {
    InvalidConnection = 'InvalidConnection'
}
```

src/codecatalyst/commands.ts
JadenSimon: I've been using "validation" when checking local state/configuration whereas "authentication" involves API calls. Right now there's no "who am I" API call for bearer tokens so the word "authenticate" is used only for the browser login flow. There's obviously a lot of overlap though `validateConnection` is more aligned with the impl. in `auth.ts`

src/codecatalyst/commands.ts
JadenSimon: > IMO any client shouldn't need to worry if the connection is valid.

There are many, many edge-cases involved with auth which is why a lot of decisions are delegated. For example, we likely wouldn't want to show a notification for a background task, at least not the same notification. CodeWhisperer is another good example because they might make API requests every time the user types. 

Logic [already exists](https://github.com/aws/aws-toolkit-vscode/blob/965cffb7f317af28369e3a5f7a4f187b91ad9f7b/src/credentials/auth.ts#L684-L694) that checks the local state of _all_ connections before they're used but we are conservative about when notifications are shown to avoid spamming the user.

Another really important thing to keep in mind is that a connection can only really be deemed "valid" _after_ making the desired request. I did not implement any logic to back-propagate exceptions from client requests to `Auth.ts` though this would help a lot with keeping states in-sync.

src/credentials/auth.ts
justinmk3: sure, though enums are never needed (essentially deprecated) in typescript, see https://www.executeprogram.com/blog/typescript-features-to-avoid

justinmk3 (approved):

PR #3307, opened 2023-04-03
tests: hook up E2E tests to CI

JadenSimon:
## Problem
E2E tests must be ran manually

## Solution
Run them automatically through CodeBuild

## License

### DIFF
--- a/buildspec/linuxE2ETests.yml
+++ b/buildspec/linuxE2ETests.yml
@@ -1,0 +1,39 @@
+version: 0.2
+
+env:
+    variables:
+        AWS_TOOLKIT_TEST_NO_COLOR: '1'
+        NO_COVERAGE: 'true'
+        # Suppress noisy apt-get/dpkg warnings like "debconf: unable to initialize frontend: Dialog").
+        DEBIAN_FRONTEND: 'noninteractive'
+        LOGIN_SECRET_ARN: 'codecatalyst-test-account'
+
+phases:
+    install:
+        commands:
+            - '>/dev/null add-apt-repository universe'
+            - '>/dev/null apt-get -qq install -y apt-transport-https'
+            - '>/dev/null apt-get -qq update'
+            - '>/dev/null apt-get -qq install -y ca-certificates'
+            - 'apt-get install --reinstall ca-certificates'
+            # Dependencies for running vscode.
+            - '>/dev/null apt-get -yqq install libatk1.0-0 libgtk-3-dev libxss1 xvfb libnss3-dev libasound2 libasound2-plugins libsecret-1-0'
+
+    build:
+        commands:
+            - npm ci --unsafe-perm
+            # We cannot run `code` as root during tests
+            # From: https://github.com/aws/aws-codebuild-docker-images/blob/2f796bb9c81fcfbc8585832b99a5f780ae2b2f52/ubuntu/standard/6.0/Dockerfile#L56
+            - mkdir -p /home/codebuild-user
+            - chown -R codebuild-user:codebuild-user /tmp /home/codebuild-user .
+            - su codebuild-user -c "xvfb-run npm run testE2E"
+            - VCS_COMMIT_ID="${CODEBUILD_RESOLVED_SOURCE_VERSION}"
+            - CI_BUILD_URL=$(echo $CODEBUILD_BUILD_URL | sed 's/#/%23/g')
+            - CI_BUILD_ID="${CODEBUILD_BUILD_ID}"
+        finally:
+            - rm -rf ~/.aws/sso/cache || true
+reports:
+    integ-test:
+        files:
+            - '*'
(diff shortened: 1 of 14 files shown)


buildspec/linuxE2ETests.yml
```diff
@@ -0,0 +1,39 @@
...
+phases:
+    install:
+        commands:
+            - '>/dev/null add-apt-repository universe'
+            - '>/dev/null apt-get -qq install -y apt-transport-https'
+            - '>/dev/null apt-get -qq update'
+            - '>/dev/null apt-get -qq install -y ca-certificates'
+            - 'apt-get install --reinstall ca-certificates'
+            # Dependencies for running vscode.
+            - '>/dev/null apt-get -yqq install libatk1.0-0 libgtk-3-dev libxss1 xvfb libnss3-dev libasound2 libasound2-plugins libsecret-1-0'
```
justinmk3, line 20: nit: might be time to "include" common buildspec parts to avoid duplicate code in the buildspecs

package.json
```diff
@@ -3422,6 +3422,7 @@
         "postinstall": "npm run generateTelemetry && npm run generateConfigurationAttributes",
         "testCompile": "npm run buildScripts && tsc -p ./ && npm run instrument",
         "test": "npm run testCompile && ts-node ./scripts/test/test.ts && npm run report",
+        "testE2E": "npm run testCompile && ts-node ./scripts/test/testE2E.ts && npm run report",
         "integrationTest": "npm run testCompile && ts-node ./scripts/test/integrationTest.ts && npm run report",
```
justinmk3, line 3426: future: we should rename this to `testinteg` as well as `src/integrationTest/` => `src/testinteg`

src/integrationTest/globalSetup.test.ts
```diff
@@ -6,48 +6,30 @@
...
-}
-
-// Before all tests begin, once only:
-before(async function () {
+export async function mochaGlobalSetup(this: Mocha.Runner) {
     console.log('globalSetup: before()')
+
+    // Prevent CI from hanging by forcing a timeout on both hooks and tests
+    this.on('hook', hook => setRunnableTimeout(hook, maxTestDuration))
+    this.on('test', test => setRunnableTimeout(test, maxTestDuration))
```
justinmk3, line 28: very nice

src/test/setupUtil.ts
```diff
@@ -0,0 +1,150 @@
...
+        throw error
+    }
+
+    return respPayload
+}
+
+function maskArns(text: string) {
+    return text.replace(/arn:aws:(?:.*?):(.*?):(.*?):./g, (match, region, account) => {
+        if (region) {
+            match = match.replace(region, '[omitted]')
```
justinmk3, line 142: very nice

src/testE2E/codecatalyst/client.test.ts
```diff
@@ -147,6 +160,37 @@ describe('Test how this codebase uses the CodeCatalyst API', function () {
...
+            await assert.rejects(
+                prepareDevEnvConnection(client, { ...defaultDevEnv }),
+                /Remote SSH extension not installed/
+            )
+            const notification = await getTestWindow().waitForMessage(
+                /Connecting to Dev Environment requires the Remote SSH extension/
+            )
+
+            // Normally selecting `Install...` would open up the extensions tab
+            // But for this test we actually do want to install the extension programtically
```
justinmk3, line 177: is this because the tests need the extension? or why don't we just test whatever the actual behavior is?

nkomonen-amazon (approved):

src/test/setupUtil.ts
```diff
@@ -0,0 +1,150 @@
...
+        error.message = maskArns(respPayload.errorMessage || error.message)
+
+        throw error
+    }
+
+    return respPayload
+}
+
+function maskArns(text: string) {
+    return text.replace(/arn:aws:(?:.*?):(.*?):(.*?):./g, (match, region, account) => {
```
rli, line 140: `aws` can also be `aws-cn` or `aws-us-gov` depending on the partition

buildspec/linuxE2ETests.yml
JadenSimon: Doesn't seem like buildspecs support 'extending' from a base file. There are other solutions of course, backlogging.

src/testE2E/codecatalyst/client.test.ts
JadenSimon: > is this because the tests need the extension? or why don't we just test whatever the actual behavior is?

Yes, the subsequent tests fail without the extension. We can install the extension prior to running the tests but that means we can't easily test the behavior without the extension. Which I think is import when considering the new user experience. 

We are testing (most) of the actual behavior but because we can't manipulate the extensions tab we instead simulate what a user might do e.g. install the extension.

PR #3310, opened 2023-04-05
fix(codecatalyst): Notify ssh extension min version required

nkomonen-amazon:
## Problem:

When a user attempted to start a dev env, it failed since the ssh extension version was too old.

## Solution:

Enforce a minumum version of the ssh extension.
If the version is unsupported we will notify the user through an error message.

This solution can be used by any future extension
as well.

Screen Shot 2023-04-05 at 12 20 27 PM

## License

### DIFF
--- a/src/codecatalyst/tools.ts
+++ b/src/codecatalyst/tools.ts
@@ -24,7 +24,7 @@ import { getLogger } from '../shared/logger'
 import { getIdeProperties } from '../shared/extensionUtilities'
 import { showConfirmationMessage } from '../shared/utilities/messages'
 import { getSshConfigPath } from '../shared/extensions/ssh'
-import { VSCODE_EXTENSION_ID } from '../shared/extensions'
+import { VSCODE_EXTENSION_ID, vscodeExtensionMinVersion } from '../shared/extensions'
 
 interface DependencyPaths {
     readonly vsc: string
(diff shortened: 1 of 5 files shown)


JadenSimon (approved):

src/codecatalyst/tools.ts
```diff
@@ -38,10 +38,16 @@ interface MissingTool {
 }
 
 export const hostNamePrefix = 'aws-devenv-'
+export const sshExtMinVer = '0.98.0'
```
justinmk3, line 41: this could live  wherever `VSCODE_EXTENSION_ID` lives

src/shared/utilities/vsCodeUtils.ts
```diff
@@ -53,8 +55,37 @@ export function isExtensionActive(extId: string): boolean {
     return !!extension && extension.isActive
 }
 
-export function isExtensionInstalled(extId: string): boolean {
-    return !!vscode.extensions.getExtension(extId)
+/**
+ * Checks if an extension is installed and meets the version requirement
+ * @param minVersion The minimum semver required for the extension
+ */
+export function isValidExtensionInstalled(extId: string, minVersion?: string): boolean {
```
justinmk3, line 62: "valid" is redundant in the function name. the paramters and docstring are sufficient hints about more detailed behavior.

```suggestion
export function isExtensionInstalled(extId: string, minVersion?: string): boolean {
```

src/shared/utilities/vsCodeUtils.ts
```diff
@@ -53,8 +55,37 @@ export function isExtensionActive(extId: string): boolean {
...
+    const ext = vscode.extensions.getExtension(extId)
+    if (ext === undefined) {
+        return false
+    }
+
+    return isExtensionValid(ext, minVersion)
+}
+
+export function isExtensionValid(ext: vscode.Extension<any>, minVersion?: string): boolean {
```
justinmk3, line 71: i see that this is used in tests, else it would be unnecessary indirection. Though, we're not really testing much, except that semver does it it says it does :) So, a single test to "exercise" `isExtensionInstalled` would be enough imo, rather than deeper coverage. That could eliminate the need for this indirection.

src/shared/utilities/vsCodeUtils.ts
```diff
@@ -90,6 +117,24 @@ export function showInstallExtensionMsg(
     return false
 }
 
+export function buildMissingExtensionMessage(
+    extId: string,
+    extName: string,
+    minVersion?: string,
+    feat = `${getIdeProperties().company} Toolkit`
+): string {
+    const expectedVersionMsg = minVersion ? ` of version >=${getSemver(minVersion)}` : ''
```
justinmk3, line 126: `getSemver` can throw here, which is not really helpful since we are already in the middle of handling an error.

It also isn't useful for `getSemver` to throw from `isExtensionInstalled`, instead just handle the nullish result in `isExtensionInstalled` and treat it as an invalid  version.

IOW: `getSemver` isn't needed, it can be elimiated. Just use `semver.coerce` directly.

PR #3313, opened 2023-04-05
fix(codecatalyst): codecatalyst login prompt message not clear

nkomonen-amazon:
This makes the prompt message more clear to users based off of IDE-10429
### Now
Screen Shot 2023-04-05 at 1 56 09 PM

## License


JadenSimon (approved):

src/shared/localizedText.ts
```diff
@@ -13,6 +13,7 @@ export const ok = localize('AWS.generic.response.ok', 'OK')
 export const retry = localize('AWS.generic.response.retry', 'Retry')
 export const skip = localize('AWS.generic.response.skip', 'Skip')
 export const confirm = localize('AWS.generic.confirm', 'Confirm')
+export const proceed = localize('AWS.generic.proceed', 'Proceed')
```
justinmk3, line 16: "Continue" is a more common verb, no?

src/codecatalyst/auth.ts
```diff
@@ -106,18 +106,18 @@ export class CodeCatalystAuthenticationProvider {
...
             // TODO: change to `satisfies` on TS 4.9
             telemetry.record({ codecatalyst_connectionFlow: isNewUser ? 'Create' : 'Upgrade' } as ConnectionFlowEvent)
 
-            const message = isNewUser
-                ? 'CodeCatalyst requires an AWS Builder ID connection. Creating a connection opens your browser to login.\n\n Create one now?'
-                : 'Your AWS Builder ID connection does not have access to CodeCatalyst. Upgrading the connection requires another login.\n\n Upgrade now?'
-            const resp = await vscode.window.showInformationMessage(message, { modal: true }, okItem, cancelItem)
-            if (resp !== okItem) {
+            const message =
+                'The AWS Toolkit for VSCode requires a connection for CodeCatalyst to begin.\n\n Proceed to the browser to allow access?'
```
justinmk3, line 117: looks for usages of `getIdeProperties().company` to see how we normally refer to the Toolkit in messages. Needs to be "Amazon", not "AWS", in some regions, thus `getIdeProperties().company` instead of "AWS".

If there's a chance this may show in non-vscode contexts such as Cloud9, I'd drop the mention of that also. If for some reason it's necessary to mention "for VSCode" here, could use `getIdeProperties().shortName`

src/shared/localizedText.ts
nkomonen-amazon: "Proceed" was the specific word Ankit used in the ticket. I'm assuming it was intentional

nkomonen-amazon: Now updated to "Continue"

src/shared/localizedText.ts
```diff
@@ -13,6 +13,7 @@ export const ok = localize('AWS.generic.response.ok', 'OK')
 export const retry = localize('AWS.generic.response.retry', 'Retry')
 export const skip = localize('AWS.generic.response.skip', 'Skip')
 export const confirm = localize('AWS.generic.confirm', 'Confirm')
+export const continueText = localize('AWS.generic.continue', 'Continue')
```
nkomonen-amazon, line 16: TS does not allow the name "continue", which is why "continueText"

PR #3316, opened 2023-04-06
feat(cwl): jump to search log group from log stream quick pick

nkomonen-amazon:
## Problem:

By default when a user clicks on a Log Group in the explorer
they will be shown a quick pick of only Log Streams.

If they want to search the whole Log Group, they have to
exit out of it, right click, then select search log group.

## Solution:

Add a quickpick item as the first item of the Log Stream
selection. This item if chosen will redirect them to the
Search Log Stream quick pick flow.

- When a Log Group is left clicked the user is shown the following:

Screen Shot 2023-04-17 at 11 33 54 AM

- If they select the first item they are brought to the following:

Screen Shot 2023-04-06 at 5 38 01 PM

## License

### DIFF
--- a/src/cloudWatchLogs/activation.ts
+++ b/src/cloudWatchLogs/activation.ts
@@ -83,11 +83,15 @@ export async function activate(context: vscode.ExtensionContext, configuration: 
         // Here instead of in ../awsexplorer/activation due to dependence on the registry.
         Commands.register('aws.cwl.viewLogStream', async (node: LogGroupNode) => await viewLogStream(node, registry)),
 
-        Commands.register(
-            'aws.cwl.searchLogGroup',
-            async (node: LogGroupNode | CloudWatchLogsNode) =>
-                await searchLogGroup(node instanceof LogGroupNode ? node : undefined, registry)
-        ),
+        Commands.register('aws.cwl.searchLogGroup', async (node: LogGroupNode | CloudWatchLogsNode) => {
+            let logGroupInfo = undefined
+
+            if (node instanceof LogGroupNode) {
+                logGroupInfo = { regionName: node.regionCode, groupName: node.logGroup.logGroupName! }
+            }
+
+            await searchLogGroup(registry, logGroupInfo)
+        }),
 
         Commands.register('aws.cwl.changeFilterPattern', async () => changeLogSearchParams(registry, 'filterPattern')),
 
--- a/src/cloudWatchLogs/commands/viewLogStream.ts
+++ b/src/cloudWatchLogs/commands/viewLogStream.ts
@@ -24,25 +24,22 @@ import {
     filterLogEventsFromUri,
 } from '../registry/logDataRegistry'
 import { createURIFromArgs } from '../cloudWatchLogsUtils'
-import { prepareDocument } from './searchLogGroup'
+import { prepareDocument, searchLogGroup } from './searchLogGroup'
 import { telemetry, Result } from '../../shared/telemetry/telemetry'
-
-export interface SelectLogStreamResponse {
-    region: string
-    logGroupName: string
-    logStreamName: string
-}
 
 export async function viewLogStream(node: LogGroupNode, registry: LogDataRegistry): Promise<void> {
     let result: Result = 'Succeeded'
     const logStreamResponse = await new SelectLogStreamWizard(node).run()
-    if (!logStreamResponse) {
-        telemetry.cloudwatchlogs_open.emit({
-            result: 'Cancelled',
-            cloudWatchResourceType: 'logStream',
-            source: 'Explorer',
-        })
+    if (
+        logStreamResponse === undefined ||
+        logStreamResponse.kind === 'cancelled' ||
+        logStreamResponse.kind === 'failed'
+    ) {
         return
+    }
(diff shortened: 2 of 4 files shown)


src/cloudWatchLogs/activation.ts
```diff
@@ -83,11 +83,15 @@ export async function activate(context: vscode.ExtensionContext, configuration:
...
-                await searchLogGroup(node instanceof LogGroupNode ? node : undefined, registry)
-        ),
+        Commands.register('aws.cwl.searchLogGroup', async (node: LogGroupNode | CloudWatchLogsNode) => {
+            let logGroupInfo = undefined
+
+            if (node instanceof LogGroupNode) {
+                logGroupInfo = { regionName: node.regionCode, groupName: node.logGroup.logGroupName! }
+            }
+
+            await searchLogGroup(registry, logGroupInfo)
```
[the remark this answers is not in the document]
justinmk3: would it make sense to define this as a `Partial<LogGroupNode ...>` ? or vice-versa? Then we could pass `LogGroupNode`

src/cloudWatchLogs/commands/viewLogStream.ts
```diff
@@ -87,6 +94,17 @@ export class DefaultSelectLogStreamWizardContext implements SelectLogStreamWizar
                 totalSteps: this.totalSteps,
             },
         })
+
+        // Add item to top of quick pick which indicates user instead wants to do Search Log Group quickpick flow
+        const separator = picker.quickPickSeparator()
+        const searchLogGroupItems: vscode.QuickPickItem[] = separator
+            ? [
+                  { label: 'Search Log Group', detail: 'Search within all Log Streams of the current Log Group' },
+                  { label: 'Log Streams', ...separator },
```
justinmk3, line 103: similar to how "Log Streams" shows in the right column, I think we should use "Actions" as the common title for action items.

src/test/cloudWatchLogs/commands/viewLogStream.test.ts
```diff
@@ -53,14 +54,20 @@ describe('viewLogStreamWizard', async function () {
...
         const result = await wizard.run()
 
         assert.ok(result)
-        assert.strictEqual(result?.logGroupName, groupName)
-        assert.strictEqual(result?.logStreamName, streamName)
-        assert.strictEqual(result?.region, region)
+        // HACK: using '===' since following this assertion the exact
+        // type of 'result' is narrowed down to 'selectedLogStream'.
+        // Otherwise we get TS2339 error.
+        assert.ok(result.kind === 'selectedLogStream')
```
JadenSimon, line 67: This is currently the only way to narrow the type without writing an overloaded type for `strictEqual`. We can add one though.

src/cloudWatchLogs/commands/viewLogStream.ts
```diff
@@ -87,6 +94,18 @@ export class DefaultSelectLogStreamWizardContext implements SelectLogStreamWizar
                 totalSteps: this.totalSteps,
             },
         })
+
+        // Add item to top of quick pick which indicates user instead wants to do Search Log Group quickpick flow
+        const separator = picker.quickPickSeparator()
+        const searchLogGroupItems: vscode.QuickPickItem[] = separator
+            ? [
+                  { label: 'Actions', ...separator },
+                  { label: 'Search Log Group', detail: 'Search within all Log Streams of the current Log Group' },
```
justinmk3, line 103:
```suggestion
                  { label: 'Search Log Group', detail: 'Search all Log Streams in this Log Group' },
```

src/cloudWatchLogs/commands/viewLogStream.ts
```diff
@@ -110,21 +129,55 @@ export class DefaultSelectLogStreamWizardContext implements SelectLogStreamWizar
...
+        // If the user wants to search the log group instead we return a different result
+        // compared to if they actually selected a log stream
+        return this.doSearchLogGroupProcess(result) ?? this.resolveSelectedLogStream(result)
+    }
+
+    /**
+     * Returns an object indicating the user wants to search the log groups,
+     * otherwise undefined
+     */
+    doSearchLogGroupProcess(quickPickResult?: string): LogSearchChoice | undefined {
```
justinmk3, line 143: If we add indirection like this, the function name should be intuitive, else what's the purpose? This function name is inscrutable, so maybe the indirection should be dropped.

src/test/techdebt.test.ts
```diff
@@ -34,6 +35,11 @@ describe('tech debt', function () {
             semver.lt(minVscode, '1.56.0'),
             'remove AsyncLocalStorage polyfill used in `spans.ts` if Cloud9 is on node 14+'
         )
+
+        assert.ok(
+            semver.lt(minVscode, '1.64.0'),
+            `remove ${quickPickSeparator.name}() since the vscode API now supports it`
```
justinmk3, line 41: there is alrady a test on line 32. it needs to be updated

justinmk3: Since https://github.com/aws/aws-toolkit-vscode/pull/3328 was merged we don't need this test now.

PR #3317, opened 2023-04-07
feat(auth): add support for `sso-session`

JadenSimon:
## Problem
Users are unable to use SSO configurations created by newer versions of the AWS CLI. 

For example, `aws configure sso` might produce something like this:
```ini
[profile default]
sso_session = default
sso_account_id = 012345678910
sso_role_name = MyRole

[sso-session default]
sso_region = us-east-1
sso_start_url = https://d-xxxxxxxxx.awsapps.com/start
sso_registration_scopes = sso:account:access
```

Users should be able to select the `default` profile in the Toolkit.

## Solution
Support `sso-session` sections and the corresponding `sso_session` property.

This PR does _not_ surface SSO sections in the UI. Only profiles are valid options.

## License


src/credentials/providers/sharedCredentialsProvider.ts
```diff
@@ -139,6 +141,36 @@ export class SharedCredentialsProvider implements CredentialsProvider {
         return true
     }
 
+    private getProfile(name: string) {
+        return getSectionDataOrThrow(this.sections, name, 'profile')
+    }
+
+    private getSsoProfileFromProfile(profile = this.profile): SsoProfile {
```
nkomonen-amazon: Is the `profile` parameter redundant here since it doesn't seem there is a need for it?

src/test/credentials/provider/sharedCredentialsProvider.test.ts
```diff
@@ -56,190 +68,330 @@ describe('SharedCredentialsProvider', async function () {
     })
 
     it('gets profile properties', async function () {
-        const sut = new SharedCredentialsProvider(
-            'default',
-            new Map<string, Profile>([
-                ['default', { aws_access_key_id: 'x', aws_secret_access_key: 'y', region: 'foo' }],
-            ])
-        )
+        const sections = await createTestSections(`
```
nkomonen-amazon, line 71: Nice addition with `createTestSections`. Makes it much easier to read the test

nkomonen-amazon (approved):

src/credentials/providers/sharedCredentialsProvider.ts
JadenSimon: Yup.

PR #3323, opened 2023-04-11
fix(auth): address several edge-cases

JadenSimon:
## Problem
* Builder ID profiles are showing in the connection picker
* Users can rarely be left in a bad auth state

## Solution
* Hide builder ID profiles 
* Clean profiles left in a bad state on reload
* Free invalid credentials lock after reauth

## License

### DIFF
--- a/src/credentials/auth.ts
+++ b/src/credentials/auth.ts
@@ -16,7 +16,7 @@ import { codicon, getIcon } from '../shared/icons'
 import { Commands } from '../shared/vscode/commands2'
 import { createQuickPick, DataQuickPickItem, showQuickPick } from '../shared/ui/pickerPrompter'
 import { isValidResponse } from '../shared/wizards/wizard'
-import { CancellationError } from '../shared/utilities/timeoutUtils'
+import { CancellationError, Timeout } from '../shared/utilities/timeoutUtils'
 import { errorCode, formatError, ToolkitError, UnknownError } from '../shared/errors'
 import { getCache } from './sso/cache'
 import { createFactoryFunction, Mutable } from '../shared/utilities/tsUtils'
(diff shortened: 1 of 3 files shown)


src/credentials/auth.ts
```diff
@@ -680,10 +682,16 @@ export class Auth implements AuthService, ConnectionManager {
...
             })
         }
-        // TODO: cancellable notification?
         if (previousState === 'valid') {
+            const timeout = new Timeout(60000)
+            this.#invalidCredentialsTimeouts.set(id, timeout)
+
             const message = localize('aws.auth.invalidConnection', 'Connection is invalid or expired, login again?')
-            const resp = await vscode.window.showInformationMessage(message, localizedText.yes, localizedText.no)
+            const resp = await Promise.race([
```
justinmk3, line 690: should we prefer `Promise.any` ?

src/credentials/auth.ts
```diff
@@ -701,6 +709,16 @@ export class Auth implements AuthService, ConnectionManager {
             return
         }
 
+        // Clear anything stuck in an 'authenticating...' state
+        // This can rarely happen when closing VS Code during authentication
+        await Promise.all(
```
justinmk3, line 714: [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) may be useful here if we want to handle errors.

justinmk3 (approved):

src/credentials/auth.ts
JadenSimon: Yeah it'll throw when updating a profile that no longer exists. Which could only happen in some sort of race condition. At least here only `vscode.Memento.update` could throw as all `getProfileOrThrow` calls happen within the same event loop.

A safer and more robust solution might be to store all state in-mem and add an 'uninitialized' state so it can be lazily eval'd as-needed. I'll try making that change when we address "linked accounts"

src/credentials/auth.ts
JadenSimon: We're trying to bypass the prompt here so `Promise.race` is preferred. If we could programmatically close the notification this would be so much cleaner :/

PR #3326, opened 2023-04-12
dev: introduce `Tasks` module

JadenSimon:
## Problems
* Lack of cohesion for cancellations
* No easy way to instrument logic without using telemetry
* It can be hard to understand the relationship between functions, even with logs

## Solution
Add a new module that allows callers to create arbitrary "tasks". Tasks do three important things:
* Handle cancellations as gracefully as possible
* Allows logic within a task to deal with cancellations isolated to that particular task
* Tracks when tasks are started within an existing task, creating a hierarchy of tasks

For functionality beyond this, callers can use lifecycle hooks to do whatever they need to do. 

### Other Changes
Several existing implementations have been adapted to use tasks. More testing is needed to ensure correctness.

## License

### DIFF
--- a/src/dev/activation.ts
+++ b/src/dev/activation.ts
@@ -16,8 +16,12 @@ import { Wizard } from '../shared/wizards/wizard'
 import { deleteDevEnvCommand, installVsixCommand, openTerminalCommand } from './codecatalyst'
 import { watchBetaVSIX } from './beta'
 import { isCloud9 } from '../shared/extensionUtilities'
-import { entries } from '../shared/utilities/tsUtils'
+import { entries, isNonNullable } from '../shared/utilities/tsUtils'
 import { isReleaseVersion } from '../shared/vscode/env'
+import { ResourceTreeNode } from '../shared/treeview/resource'
+import { isCancellable, isCompleted, Task, Tasks } from '../shared/tasks'
+import { addColor, getIcon } from '../shared/icons'
+import { RootNode } from '../awsexplorer/localExplorer'
 
 interface MenuOption {
     readonly label: string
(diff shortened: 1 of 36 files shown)


JadenSimon:
`esbuild` bug:
```ts
new Map<Task['id'], Task & { info: TaskInfo }>
```
results in:
> ERROR: Expected ";" but found ","

src/shared/tasks.ts
```diff
@@ -0,0 +1,365 @@
...
+                setPromise(promise)
+                return task
+            },
+        } as unknown as Task<T, U>
+
+        this.addTask(task)
+        return task as StoppedTask<T, U>
+    }
+
+    public getChildren(task: Pick<Task, 'id'>) {
```
justinmk3, line 235: (maybe future) do we need a traverse() which visits all descendants?

Is there a `.then()` which "chains" a task to the completion of its parent? So a task can have many children, and then we can `.waitAll()` or `.waitAny()` on the children (or maybe descendants).

justinmk3 (approved):

JadenSimon:
Need VSC 1.66+ for Node 16 support. Might be able to update the `AsyncLocalStorage` shim to handle context binding to support node 14.

src/dev/activation.ts
```diff
@@ -276,3 +280,320 @@ async function showGlobalState() {
...
+
+        return item
+    }
+
+    public toTreeNode(): ResourceTreeNode<this> {
+        return new ResourceTreeNode(this)
+    }
+}
+
+const clearMemory = Commands.register({ id: '_aws.dev.tasks.clearMemory', runAsTask: false }, (memory: TaskMemory) =>
```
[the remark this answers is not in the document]
JadenSimon: Yeah it's used within their own codebase to mark "private" commands. I believe `vscode.commands` actually filters out commands prefixed with `_` by default.

src/shared/tasks.ts
JadenSimon: > Is there a .then() which "chains" a task to the completion of its parent?

Right now child/parent relationship is strictly based off the asynchronous context so no. But I think it could make sense to add something like that because right now you have to nest things. Tasks can complete independently from one another which might be confusing at first but this is an accurate view of how the event loop works.

> So a task can have many children, and then we can .waitAll() or .waitAny() on the children (or maybe descendants).

I considered added more logic for these sorts of use-cases but decided to leave it as-is for now.

PR #3327, opened 2023-04-13
auth: Setup initial Auth Add Connections webview

nkomonen-amazon:
## Problem
We want to create a webview for Auth setup

## Solution
This PR does the following:

- Creates a vscode command which can be executed in the command palette to open the webview
- Sets up the initial code for a webview to be shown

The actual content will be added in future commits

Kapture 2023-04-13 at 10 18 52

### End goal for context
image

Fixes IDE-10461

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -1984,6 +1984,11 @@
                 "category": "%AWS.title%"
             },
             {
+                "command": "aws.auth.showConnectionsPage",
+                "title": "%AWS.command.auth.showConnectionsPage%",
+                "category": "%AWS.title%"
+            },
+            {
                 "command": "aws.auth.switchConnections",
                 "title": "%AWS.command.auth.switchConnections%",
                 "category": "%AWS.title%"
--- a/package.json
+++ b/package.json
@@ -1100,6 +1100,10 @@
                     "when": "false"
                 },
                 {
+                    "command": "aws.auth.showConnectionsPage",
+                    "when": "aws.isDevMode"
+                },
+                {
                     "command": "aws.dev.openMenu",
                     "when": "aws.isDevMode || isCloud9"
                 }
@@ -1986,7 +1990,8 @@
             {
                 "command": "aws.auth.showConnectionsPage",
                 "title": "%AWS.command.auth.showConnectionsPage%",
-                "category": "%AWS.title%"
+                "category": "%AWS.title%",
+                "enablement": "aws.isDevMode"
             },
             {
                 "command": "aws.auth.switchConnections",
--- a/src/shared/vscode/commands2.ts
+++ b/src/shared/vscode/commands2.ts
@@ -70,25 +70,24 @@ interface DeclaredCommand<T extends Callback = Callback, U extends any[] = any[]
  *
  * @param T The class that declares the commands
  */
-interface CommandDeclarations<T> {
-    readonly backend: T
-
+export interface CommandDeclarations<T> {
     readonly declared: { [K in FunctionKeys<T>]: DeclaredCommand<Functions<T>[K], [T]> }
 }
 
-export abstract class BaseCommandDeclarations<T> implements CommandDeclarations<T> {
-    public readonly declared: { [K in FunctionKeys<T>]: DeclaredCommand<Functions<T>[K], [T]> } = {} as any
-
-    public constructor(readonly extContext: vscode.ExtensionContext, readonly backend: T) {}
-
-    /**
-     * Registers all declared commands with VS Code commands api.
-     */
-    public registerCommandsWithVSCode(): void {
-        this.extContext.subscriptions.push(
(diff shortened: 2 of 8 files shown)


src/shared/vscode/commands2.ts
```diff
@@ -43,14 +43,54 @@ export interface Command<T extends Callback = Callback> {
...
+    public constructor(readonly extContext: vscode.ExtensionContext, readonly backend: T) {}
+
+    /**
+     * Registers all declared commands with VS Code commands api.
+     */
+    public registerCommandsWithVSCode(): void {
+        this.extContext.subscriptions.push(
+            ...Object.values<DeclaredCommand>(this.declared).map(c => c.register(this.backend))
+        )
+    }
```
JadenSimon, line 91: This implementation requires an instance of the class on construction, but declarations are meant to support late binding _and_ initialization.

Try this instead:
```suggestion
    public constructor(readonly extContext: vscode.ExtensionContext) {}

    /**
     * Registers all declared commands with VS Code commands api.
     */
    public registerCommandsWithVSCode(backend: T): void {
        this.extContext.subscriptions.push(
            ...Object.values<DeclaredCommand>(this.declared).map(c => c.register(backend))
        )
    }
```

src/shared/vscode/commands2.ts
```diff
@@ -43,14 +43,54 @@ export interface Command<T extends Callback = Callback> {
...
+ * Classes that have "declared" commands which wrap the underlying
+ * logic provided by 'T'.
+ *
+ * These declared commands can then be used to do things like register
+ * it as a VS Code command, or to build a tree node that uses the command.
+ *
+ * @param T The class that declares the commands
+ */
+interface CommandDeclarations<T> {
+    readonly backend: T
```
JadenSimon, line 74: I think we want to decouple here. Commands are top-level interfaces.
```suggestion
```

src/shared/vscode/commands2.ts
```diff
@@ -43,14 +43,54 @@ export interface Command<T extends Callback = Callback> {
...
+
+    readonly declared: { [K in FunctionKeys<T>]: DeclaredCommand<Functions<T>[K], [T]> }
+}
+
+export abstract class BaseCommandDeclarations<T> implements CommandDeclarations<T> {
+    public readonly declared: { [K in FunctionKeys<T>]: DeclaredCommand<Functions<T>[K], [T]> } = {} as any
```
JadenSimon, line 80:
```suggestion
    public abstract readonly declared: { [K in FunctionKeys<T>]: DeclaredCommand<Functions<T>[K], [T]> }
```

src/shared/vscode/commands2.ts
```diff
@@ -43,14 +43,54 @@ export interface Command<T extends Callback = Callback> {
...
+
+    public constructor(readonly extContext: vscode.ExtensionContext, readonly backend: T) {}
+
+    /**
+     * Registers all declared commands with VS Code commands api.
+     */
+    public registerCommandsWithVSCode(): void {
```
JadenSimon, line 87: If you apply my recommendations you might notice that a base class isn't really necessary because this function reduces nicely into:
```ts
function registerCommandsWithVSCode<T>(declarations: CommandDeclaraions<T>, backend: T): void
```

Although if you're planning on adding more logic to this class then I'd leave it as a class.

package.json
```diff
@@ -1974,6 +1974,11 @@
                 "title": "%AWS.command.auth.addConnection%",
                 "category": "%AWS.title%"
             },
+            {
+                "command": "aws.auth.showConnectionsPage",
+                "title": "%AWS.command.auth.showConnectionsPage%",
+                "category": "%AWS.title%"
+            },
```
JadenSimon, line 1981: You shouldn't need to add anything to `package.json` until we start changing the UI. Add a command to `src/dev/activation.ts` or create an experimental flag (see the `Experiments` class).

We don't want to surface anything unfinished to users. If this makes things hard to work with then we can use a feature branch.

JadenSimon:
Cloud9 won't respect the enablement clause. Need this under the `commandPalette` menu:
```json
{
    "command": "aws.auth.showConnectionsPage",
    "when": "aws.isDevMode"
}
```

JadenSimon (approved):

src/shared/vscode/commands2.ts
```diff
@@ -43,14 +43,53 @@ export interface Command<T extends Callback = Callback> {
...
+    readonly declared: { [K in FunctionKeys<T>]: DeclaredCommand<Functions<T>[K], [T]> }
+}
+
+/**
+ * Using the given inputs will register the given commands with VS Code.
+ *
+ * @param declarations Has the mapping of command names to the backend logic
+ * @param backend The backend logic of the commands
+ */
+export function registerCommandsWithVSCode<T>(
```
justinmk3, line 83: seems like unnecessary indirection. Also "with vscode" is a bit redundant.

PR #3332, opened 2023-04-14
fix(icon): change custom unicode range

nkomonen-amazon:
## Problem:

With the current implementation we write to a certain
unicode range which intentionally offsets the alignment
of the unicode character.

Screen Shot 2023-04-14 at 3 58 37 PM

## Solution:

Use the safe range starting a 0xE000-0xF8FF.

- VSCode unicode characters will be in the 0xF000 range
- Other unicode characters will be in the 0xE000 range

Due to this we shouldn't have a collision

## License

### DIFF
--- a/.changes/next-release/Bug Fix-a401606f-1768-452b-967b-ba24fc834b27.json
+++ b/.changes/next-release/Bug Fix-a401606f-1768-452b-967b-ba24fc834b27.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "offset icons in webviews"
+	"description": "Incorrect offset icons in webviews"
 }
--- a/.changes/next-release/Bug Fix-a401606f-1768-452b-967b-ba24fc834b27.json
+++ b/.changes/next-release/Bug Fix-a401606f-1768-452b-967b-ba24fc834b27.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "offset icons in webviews"
+}
--- a/scripts/build/generateIcons.ts
+++ b/scripts/build/generateIcons.ts
@@ -102,7 +102,7 @@ async function generate(mappings: Record<string, number | undefined> = {}) {
         files: iconSources,
         fontName: fontId,
         formats: ['woff'],
-        startUnicode: 0xe000,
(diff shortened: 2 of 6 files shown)


scripts/build/generateIcons.ts
```diff
@@ -201,9 +201,11 @@ async function loadCodiconMappings(): Promise<Record<string, number | undefined>
     const codicons = path.join(projectDir, 'node_modules', '@vscode', 'codicons', 'src')
     const data = JSON.parse(await fs.readFile(path.join(codicons, 'template', 'mapping.json'), 'utf-8'))
     const mappings: Record<string, number | undefined> = {}
+    let counter = 0xf000
```
JadenSimon, line 204: I think we should change line 105 to `0xf000` instead. That keeps the codicon mappings 1:1. It'll regenerate `package.json` but that's OK.

.changes/next-release/Bug Fix-a401606f-1768-452b-967b-ba24fc834b27.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "offset icons in webviews"
```
JadenSimon, line 3:
```suggestion
	"description": "Incorrect icon placement in webviews"
```

scripts/build/generateIcons.ts
```diff
@@ -203,7 +203,8 @@ async function loadCodiconMappings(): Promise<Record<string, number | undefined>
     const mappings: Record<string, number | undefined> = {}
     for (const [k, v] of Object.entries(data)) {
         if (typeof k === 'string' && typeof v === 'number') {
-            mappings[`vscode-${k}`] = v - 0xd000
+            // Assumption in the 0XE000 range, verify in mapping.json if needed
```
JadenSimon, line 206: Since this is a build step we should assert our assumptions whenever possible:
```suggestion
            if (v < 0xe000 || v >= 0xf000) {
                throw new Error(`Codicon "${k}" has unexpected codepoint: ${v}`)
            }
```
And then even better would be to dedupe `0xf000` to a new const `startUnicode`. This establishes the connection between the two things.

JadenSimon (approved):

PR #3333, opened 2023-04-14
fix(codecatalyst): reduce noisy `getUserDetails` API calls

JadenSimon:
## Problem
We try to call `getDevEnvironment` (and by proxy `getUserDetails`) even though we can check if it will fail ahead of time. This happens when rendering the developer tools view.

## Solution
Check our environment variables before making the call. Also move error handling up so the meaning of `undefined` is not conflated with errors. 

## License

### DIFF
--- a/src/codecatalyst/explorer.ts
+++ b/src/codecatalyst/explorer.ts
@@ -15,6 +15,7 @@ import { CodeCatalystAuthenticationProvider } from './auth'
 import { CodeCatalystCommands } from './commands'
 import { ConnectedDevEnv, getDevfileLocation, getThisDevEnv } from './model'
 import * as codecatalyst from './model'
+import { getLogger } from '../shared/logger'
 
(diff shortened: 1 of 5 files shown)


src/codecatalyst/explorer.ts
```diff
@@ -125,7 +126,10 @@ export class CodeCatalystRootNode implements RootNode {
     }
 
     public async getTreeItem() {
-        this.devenv = await getThisDevEnv(this.authProvider)
+        this.devenv = (await getThisDevEnv(this.authProvider))?.unwrapOrElse(err => {
```
justinmk3, line 129: off topic: `unwrapOrCatch` might be a more intuitive name

justinmk3:
probably worth a changelog item, like "CodeCatalyst: improved efficiency by reducing unnecessary service calls"

PR #3340, opened 2023-04-17
feat(codecatalyst): Show README.md on first dev env load

nkomonen-amazon:
Shows the README.md in its rendered form the first time a dev env is loaded.

### VSCode 
Kapture 2023-04-17 at 16 20 38

### Dev Env w/ C9
Screen Shot 2023-04-17 at 6 35 49 PM

## License

### DIFF
--- a/src/codecatalyst/activation.ts
+++ b/src/codecatalyst/activation.ts
@@ -100,6 +100,10 @@ export async function activate(ctx: ExtContext): Promise<void> {
 }
 
 async function showReadmeFileOnFirstLoad(workspaceState: vscode.ExtensionContext['workspaceState']): Promise<void> {
+    if (isCloud9()) {
+        return
+    }
+
     getLogger().info('codecatalyst: showReadmeFileOnFirstLoad()')
     // Check dev env state to see if this is the first time the user has connected to a dev env
     const isFirstLoad = workspaceState.get('aws.codecatalyst.devEnv.isFirstLoad', true)


src/codecatalyst/activation.ts
```diff
@@ -96,3 +98,29 @@ export async function activate(ctx: ExtContext): Promise<void> {
...
+    }
+
+    const readmeUri = vscode.workspace.findFiles('README.md').then(files => {
+        if (files.length === 0) {
+            return undefined
+        }
+        return files[0]
+    })
+
+    if (readmeUri === undefined) {
```
justinmk3, line 117: does `findFiles..then` need to be await'd ?

nkomonen-amazon: My bad, made a change right before pushing. I'm setting up to test in C9 right now, so will catch if anything else breaks.

Fixing this

src/codecatalyst/activation.ts
```diff
@@ -96,3 +98,29 @@ export async function activate(ctx: ExtContext): Promise<void> {
         authProvider.removeSavedConnection()
     })
 }
+
+async function showReadmeFileOnFirstLoad(workspaceState: vscode.ExtensionContext['workspaceState']): Promise<void> {
+    // Check dev env state to see if this is the first time the user has connected to a dev env
+    const isFirstLoad = workspaceState.get('aws.codecatalyst.devEnv.isFirstLoad', true)
```
justinmk3, line 104: Not a blocker: thinking about edge cases: does workspaceState store in the project-local .vscode folder, rather than the user-global (but remote) state? Ideally it would be stored in the user-global (but remote) globalstate, since this is an environment-global state. And it's possible that the workspace state or "main project" changes, e.g. on branch-switch or other manual actions.

Maybe a future idea, could we avoid this state if we checked that the env was created in the last ~1 hour? I guess that's too flaky though.

src/codecatalyst/activation.ts
```diff
@@ -96,3 +98,29 @@ export async function activate(ctx: ExtContext): Promise<void> {
...
+        return files[0]
+    })
+
+    if (readmeUri === undefined) {
+        getLogger().info('codecatalyst: README.md not found in project root')
+        return
+    }
+
+    // Show rendered readme file to user
+    await vscode.commands.executeCommand('markdown.showPreview', readmeUri)
```
justinmk3, line 123: Hopefully this works in Cloud9. Else, `markdown.showPreviewToSide` is known to work: https://github.com/aws/aws-toolkit-vscode/blob/61bbc248221ac21e30fa74aaec20a12cae59c10a/src/lambda/commands/createNewSamApp.ts#L94

nkomonen-amazon:
@justinmk3 Looks like the parent dir of the repo (`spa-app`) is opened in C9, so we can't assume README.md is in the root. I will have to look in to this, but any ideas on your end?

C9:
Screen Shot 2023-04-17 at 5 22 38 PM

VSCode:
Screen Shot 2023-04-17 at 5 27 03 PM

src/codecatalyst/activation.ts
nkomonen-amazon: I was able to trigger `markdown.showPreview` from the UI so it should work.

src/codecatalyst/activation.ts
```diff
@@ -96,3 +98,39 @@ export async function activate(ctx: ExtContext): Promise<void> {
...
+    const isFirstLoad = workspaceState.get('aws.codecatalyst.devEnv.isFirstLoad', true)
+
+    if (!isFirstLoad) {
+        return
+    }
+
+    // Determine expected readme file location
+    let readmePath = `README.md`
+    if (isCloud9()) {
+        readmePath = projectName + '/' + readmePath
```
justinmk3, line 116: I suppose this should be reliable. Else we could tell findFiles() to recursive into child folders.

justinmk3:
Needs a changelog. Might as well rebase too then :)

src/codecatalyst/activation.ts
justinmk3: should we use path.join() here?

justinmk3:
Downside of doing this in Cloud9-MDE is that it shows this alert (known issue): 
image

.changes/next-release/Feature-865fc431-55f8-4156-b3f2-dfdcb898241e.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "show readme on first load of a dev env"
```
justinmk3, line 3:
```suggestion
	"description": "CodeCatalyst: show readme on first load of a Dev Environment"
```

src/codecatalyst/activation.ts
JadenSimon: `workspaceState` _might_ live in the home directory of the remote. If so, it shouldn't be used in a dev environment. One way to check is to nuke `$HOME` in the remote and then restart the environment.

src/codecatalyst/activation.ts
```diff
@@ -96,3 +98,35 @@ export async function activate(ctx: ExtContext): Promise<void> {
         authProvider.removeSavedConnection()
     })
 }
+
+async function showReadmeFileOnFirstLoad(workspaceState: vscode.ExtensionContext['workspaceState']): Promise<void> {
+    getLogger().info('codecatalyst: showReadmeFileOnFirstLoad()')
+    // Check dev env state to see if this is the first time the user has connected to a dev env
+    const isFirstLoad = workspaceState.get('aws.codecatalyst.devEnv.isFirstLoad', true)
+
+    if (!isFirstLoad) {
```
justinmk3, line 107: skip if `isCloud9` ?

nkomonen-amazon: Before this function is called we exit early if the IDE is not VSCode, but I'll add this check for safety

PR #3342, opened 2023-04-18
fix(codecatalyst): do not force-close workspace in Cloud9 devenv

justinmk3:
Problem:
Since eb380556b6b4, Toolkit actively exits if the CC devenv is not a "vscode" devenv. But this is incorrect when running in the context of a Cloud9 devenv.

NOTE: `executeCommand('workbench.action.remote.close')` has no effect in Cloud9, but this the code should be corrected anyway.

Solution:
Skip the condition in Cloud9.

## License

### DIFF
--- a/src/codecatalyst/activation.ts
+++ b/src/codecatalyst/activation.ts
@@ -66,7 +66,9 @@ export async function activate(ctx: ExtContext): Promise<void> {
         return undefined
     })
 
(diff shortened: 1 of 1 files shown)


justinmk3:
Tested with vscode, vscode-mde, cloud9-classic, cloud9-mde.

PR #3346, opened 2023-04-19
Make CW generates code based on the language of each notebook cell

andrewyuq:
## License

### DIFF
--- a/src/codewhisperer/util/editorContext.ts
+++ b/src/codewhisperer/util/editorContext.ts
@@ -63,7 +63,7 @@ export function getFileNameForRequest(editor: vscode.TextEditor): string {
             return filenameWithNewExtension.substring(0, CodeWhispererConstants.filenameCharsLimit)
         }
     }
-    return fileName
+    return fileName.substring(0, CodeWhispererConstants.filenameCharsLimit)
 }
 
 export function buildListRecommendationRequest(
--- a/src/codewhisperer/util/editorContext.ts
+++ b/src/codewhisperer/util/editorContext.ts
@@ -35,7 +35,7 @@ export function extractContextForCodeWhisperer(editor: vscode.TextEditor): codew
     )
 
     return {
-        filename: getFileName(editor),
+        filename: getFileNameForRequest(editor),
         programmingLanguage: {
             languageName:
                 runtimeLanguageContext.mapVscLanguageToCodeWhispererLanguage(editor.document.languageId) ??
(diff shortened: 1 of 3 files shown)


src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -39,6 +46,25 @@ export class RuntimeLanguageContext {
             shellscript: 'shell',
             sql: 'sql',
         })
+        this.supportedLanguageExtensionMap = createConstantMap<CodeWhispererConstants.SupportedLanguage, string>({
```
justinmk3, line 49: some languages like yaml have extension aliases (.yaml, .yml) so this might need to be `string[]`

justinmk3 (approved):

JadenSimon (approved):

zixlin7 (approved):

src/codewhisperer/util/runtimeLanguageContext.ts
andrewyuq: This is actually fine, what we are doing here to try to map the language type to any appropriate extension to be used by the service to generate completion for that language, so in yaml case, we can choose either .yaml or .yml the service would know to generate code for yaml(yaml here just an example since we don't support yaml yet)

justinmk3:
does this have a user-facing behavior change that should be mentioned in a changelog item?

Else, ready to merge?

andrewyuq:
> does this have a user-facing behavior change that should be mentioned in a changelog item?
> 
> Else, ready to merge?

This is not user-facing in a way that it needs a message, yes this is good to merge now.

src/codewhisperer/util/editorContext.ts
```diff
@@ -51,6 +51,21 @@ export function getFileName(editor: vscode.TextEditor): string {
     return fileName.substring(0, CodeWhispererConstants.filenameCharsLimit)
 }
 
+export function getFileNameForRequest(editor: vscode.TextEditor): string {
+    const fileName = path.basename(editor.document.fileName)
+
+    // For notebook files, we want to use the programming language for each cell for the code suggestions, so change
+    // the filename sent in the request to reflect that language
+    if (fileName.endsWith('.ipynb')) {
```
leigaol, line 59: What is the languageId in a ipynb file?

andrewyuq: It's related to the kernel, if I specify the rust kernel it's rust, if it's typescript it will correctly give me typescript.

PR #3347, opened 2023-04-19
fix(dev): `husky` does not install correctly

JadenSimon:
## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -3456,7 +3456,8 @@
         "createRelease": "ts-node ./scripts/build/createRelease.ts",
         "serve": "webpack serve --config-name vue-hmr --mode development",
         "instrument": "nyc instrument --in-place ./dist/src",
-        "report": "nyc report --reporter=html --reporter=json"
+        "report": "nyc report --reporter=html --reporter=json",
+        "prepare": "husky install"
     },
     "devDependencies": {
         "@aws-toolkits/telemetry": "^1.0.120",


package.json
```diff
@@ -3456,7 +3456,8 @@
         "createRelease": "ts-node ./scripts/build/createRelease.ts",
         "serve": "webpack serve --config-name vue-hmr --mode development",
         "instrument": "nyc instrument --in-place ./dist/src",
-        "report": "nyc report --reporter=html --reporter=json"
+        "report": "nyc report --reporter=html --reporter=json",
+        "prepare": "husky install"
```
justinmk3, line 3460: I removed this in https://github.com/aws/aws-toolkit-vscode/commit/c7637c261b90e175e9aac8dc66508dc02f4fa563 to fix CI (specifically `PushToGithub` in the release pipeline).

We don't really want husky in CI, need a way to only do this for developers?

justinmk3: @JadenSimon I'll do this. It's a followup to my previous change anyway.

justinmk3: > Yeah it's just for developers. husky is only used for git-secrets though so if we had an alternative for that we could get rid of husky

hmm, what about `prettier`? https://github.com/aws/aws-toolkit-vscode/blob/d2d8f14975c81f60decf481f6775261589703b8a/.husky/pre-commit#L6

PR #3353, opened 2023-04-19
build(deps): bump `typescript` to `5.0.4`

JadenSimon:
## License


src/apprunner/wizards/apprunnerCreateServiceWizard.ts
```diff
@@ -39,7 +39,7 @@ const validateName = (name: string) => {
         )
     }
 
-    let matches = name.match(badNameRegExp)
+    let matches: string[] | undefined = name.match(badNameRegExp) ?? undefined
```
justinmk3, line 42: seems like a type interface regression?

JadenSimon: Previously Typescript was allowing `string[]` to be assigned in-place of `RegExpMatchArray` which is incorrect because `RegExpMatchArray` is typed as `[string, string?, string?, ...]`. The first element is always present.

PR #3328, opened 2023-04-13
package: increase minimum vscode to 1.65

justinmk3:
## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3315, opened 2023-04-06
perf: startup profiling

justinmk3:
## Problem

Slow startup. #1918 #3467

## Solution

Instrument startup routines.

## Future

- Change `WatchedFiles` (or introduce an alternative) to use `vscode.workspace.onDidOpenTextDocument` instead of fs file watchers.
- Move `profiling.ts` logic into new function `TelemetryTracer.runTimed()`
    - Introduce a metric similar to `vscode_executeCommand`, e.g. `executeFunction` ?
- Fail CI if any startup routine exceeds X ms.

## Example

```
2023-11-10 06:59:47 [VERBOSE]: Toolkit startup performance
<toolkit-startup>
    initializeComputeRegion                 : 0.2
    manifestpaths                           : 0.0
    logger                                  : 109.6
    credentials-manager                     : 0.1
    statusbar                               : 1.0
    credentials-init                        : 0.5
    telemetry                               : 55.4
    schemaService.start()                   : 556.3
    aws-filetypes                           : 0.0
    devtools                                : 0.6
    codecatalyst                            : 1.7
    cloudformation-templateregistry         : 0.2
        AsyncCloudFormationTemplateRegistry : 0.0
    codewhisperer                           : 2724.1
        CodeWhispererSettings.instance      : 0.1
        AuthUtil.instance                   : 0.1
        new SecurityPanelViewProvider()     : 0.0
        activateSecurityScan                : 0.1
        new codewhispererClient.DefaultCodeWhispererClient(): 0.0
        auth.restore()                      : 2719.6
        setSubscriptionsforInlineCompletion : 0.1
    awsexplorer                             : 1.4
    apprunner                               : 0.3
    apigateway                              : 0.1
    lambda                                  : 0.2
    ssmdocuments                            : 120.2
    sam                                     : 1.2
        createYamlExtensionPrompt           : 0.2
        SamCliSettings.instance             : 0.1
        registerCommands                    : 0.4
        registerSync                        : 0.0
</toolkit-startup>
```

PR #3357, opened 2023-04-20
fix(codecatalyst): e2e test updateDevEnv failing

nkomonen-amazon:
For some reason the cc backend did not receive at
dev env id from our e2e test updateDevEnv api call.

This commit verifies we are giving a dev env id
before sending the api request.

Additional:

We used the variable 'globals' before it could
be properly initialized and due to the order of
execution this variable was accessed before it
could be properly initialized.

To fix this it was moved inside a 'before' block
which will ensure it gets executed AFTER all the
pre-test hooks are run, which includes initializing the global variable

## License

### DIFF
--- a/src/testE2E/codecatalyst/client.test.ts
+++ b/src/testE2E/codecatalyst/client.test.ts
@@ -132,7 +132,11 @@ describe('Test how this codebase uses the CodeCatalyst API', function () {
         })
 
         describe('getThisDevEnv', function () {
-            const ccAuth = CodeCatalystAuthenticationProvider.fromContext(globals.context)
+            let ccAuth: CodeCatalystAuthenticationProvider
+
+            before(function () {
+                ccAuth = CodeCatalystAuthenticationProvider.fromContext(globals.context)
+            })
 
             it('returns `undefined` if not in a dev environment', async function () {
                 const result = await getThisDevEnv(ccAuth)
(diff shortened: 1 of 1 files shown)


src/testE2E/codecatalyst/client.test.ts
```diff
@@ -132,7 +132,11 @@ describe('Test how this codebase uses the CodeCatalyst API', function () {
         })
 
         describe('getThisDevEnv', function () {
-            const ccAuth = CodeCatalystAuthenticationProvider.fromContext(globals.context)
```
justinmk3, line 135: Yeah we should never do non-constant setup in the describe() scope because its scope is not deterministically evaluated. Always must be done in before().

justinmk3 (approved):

JadenSimon:
> We used the variable 'globals' before it could
be properly initialized and due to the order of
execution this variable was accessed before it
could be properly initialized.

If this was happening the suite would fail with a type error, something like ` TypeError: Cannot read properties of undefined`. Did you see this error at some point? If so that means the test runner doesn't wait for `mochaGlobalSetup` to complete prior to loading each test file.

JadenSimon (approved):

JadenSimon:
Try adding `inactivityTimeoutMinutes: 15` to `buildDevEnvSettings` to fix the tests.

justinmk3:
> If so that means the test runner doesn't wait for `mochaGlobalSetup` to complete prior to loading each test file.

wonder if that is the cause of https://github.com/aws/aws-toolkit-vscode/pull/3315#issuecomment-1506896655

justinmk3 (approved):

justinmk3:
`E2E Tests - minimum — Build Status - FAILED` logs:

```
     Error: Timed out waiting to install remote SSH extension
      at /codebuild/output/src879936470/src/github.com/aws/aws-toolkit-vscode/src/testE2E/codecatalyst/client.test.ts:213:27
```

Also saw this on https://github.com/aws/aws-toolkit-vscode/pull/3328

PR #3359, opened 2023-04-20
fix(codewhisperer): auto-connect on vscode startup

andrewyuq:
This PR also fixes some other issues:
1. Clear the inlineCompletionState on focus change so that they should still be able to invoke CW when they invoked CW and then switched to another app and switched back.
2. put activateCodeWhipserer() to be before activateAwsExplorer() since the refreshStatusBar() and showReferenceLog() commands are used within activateAwsExplorer() when the connection is changed. These
 commands are not registered until activateCodeWhisperer() has been called.
3. Remove unnecessary call to refreshCodeWhisperer(), this only needs to be called when there's a connection change.

Before:

https://user-images.githubusercontent.com/89420755/233482893-7c97b057-043d-4d1c-a761-9112c09eaaa0.mov

After:

https://user-images.githubusercontent.com/89420755/233482649-62aac6cc-2ca5-4016-ad89-54a6f7197fd4.mov

## License

### DIFF
--- a/src/codewhisperer/activation.ts
+++ b/src/codewhisperer/activation.ts
@@ -202,10 +202,7 @@ export async function activate(context: ExtContext): Promise<void> {
         )
     )
 
-    if (!isCloud9() && !auth.isConnectionValid()) {
-        // this is to proactively check and reflect the state if user's connection is expired
-        auth.refreshCodeWhisperer()
-    }
+    await auth.restore()
 
     function activateSecurityScan() {
         context.extensionContext.subscriptions.push(
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -195,6 +195,8 @@ export async function activate(context: vscode.ExtensionContext) {
 
         await activateCloudFormationTemplateRegistry(context)
 
+        await activateCodeWhisperer(extContext)
+
         await activateAwsExplorer({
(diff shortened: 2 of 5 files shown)


JadenSimon (approved):

src/codewhisperer/activation.ts
```diff
@@ -202,10 +202,7 @@ export async function activate(context: ExtContext): Promise<void> {
         )
     )
 
-    if (!isCloud9() && !auth.isConnectionValid()) {
-        // this is to proactively check and reflect the state if user's connection is expired
-        auth.refreshCodeWhisperer()
-    }
+    await auth.restore()
```
zixlin7, line 205: do we need a Cloud9 check here?

leigaol: I believe this has to be tested in C9 to ensure it is not breaking.

src/extension.ts
```diff
@@ -195,6 +195,8 @@ export async function activate(context: vscode.ExtensionContext) {
 
         await activateCloudFormationTemplateRegistry(context)
 
+        await activateCodeWhisperer(extContext)
```
leigaol, line 198: Could you explain why you are making this change? Thanks!

PR #3362, opened 2023-04-21
fix(auth): invalid sso tokens being loaded

nkomonen-amazon:
## Problem:

At a certain point in time sso tokens became invalid.
We stored those tokens in cache and eventually they will
throw errors to the user when the are attempted to be
refreshed.

## Solution:

When the cache is being setup delete any invalid files.
Invalid files are files that were created before
a certain date.

## License

### DIFF
--- a/src/credentials/sso/cache.ts
+++ b/src/credentials/sso/cache.ts
@@ -12,6 +12,7 @@ import { hasProps, selectFrom } from '../../shared/utilities/tsUtils'
 import { SsoToken, ClientRegistration } from './model'
 import { SystemUtilities } from '../../shared/systemUtilities'
 import { DevSettings } from '../../shared/settings'
+import { statSync, Stats, readdirSync, unlinkSync } from 'fs'
 
 interface RegistrationKey {
     readonly region: string
@@ -33,11 +34,44 @@ export interface SsoCache {
 const defaultCacheDir = path.join(SystemUtilities.getHomeDirectory(), '.aws', 'sso', 'cache')
 export const getCacheDir = () => DevSettings.instance.get('ssoCacheDirectory', defaultCacheDir)
 
-export function getCache(directory = getCacheDir()): SsoCache {
+export function getCache(directory = getCacheDir(), statFunc = getFileStats): SsoCache {
+    try {
+        deleteOldFiles(directory, statFunc)
+    } catch (e) {
+        getLogger().warn('auth: error deleting old files in sso cache: %s', e)
+    }
+
     return {
         token: getTokenCache(directory),
         registration: getRegistrationCache(directory),
     }
+}
+
+function deleteOldFiles(directory: string, statFunc: typeof getFileStats) {
+    if (!isDirSafeToDeleteFrom(directory)) {
+        getLogger().warn(`Skipped deleting files in directory: ${path.resolve(directory)}`)
+        return
+    }
+
+    const fileNames = readdirSync(directory)
+    fileNames.forEach(fileName => {
+        const filePath = path.join(directory, fileName)
+        if (path.extname(filePath) === '.json' && isOldFile(filePath, statFunc)) {
+            unlinkSync(filePath)
+            getLogger().warn(`auth: removed old cache file: ${filePath}`)
+        }
+    })
+}
+
+export function isDirSafeToDeleteFrom(dirPath: string): boolean {
+    const resolvedPath = path.resolve(dirPath)
+    const isRoot = resolvedPath === path.resolve('/')
+    const isCwd = resolvedPath === path.resolve('.')
+    const isAbsolute = path.isAbsolute(dirPath)
+    const pathDepth = resolvedPath.split(path.sep).length
+
+    const isSafe = !isRoot && !isCwd && isAbsolute && pathDepth >= 5
+    return isSafe
 }
 
 export function getRegistrationCache(directory = getCacheDir()): KeyedCache<ClientRegistration, RegistrationKey> {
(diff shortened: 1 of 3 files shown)


src/credentials/sso/cache.ts
```diff
@@ -31,32 +32,40 @@ export interface SsoCache {
 
 const cacheDir = join(homedir(), '.aws', 'sso', 'cache')
 
-export function getCache(directory = cacheDir): SsoCache {
+export function getCache(directory = cacheDir, statFunc = getFileStats): SsoCache {
+    // TODO: Delete this in a few months since we will most likely not have many invalid files left
+    deleteInvalidFiles(directory, statFunc)
```
justinmk3, line 37: maybe we should add a test to `techdebt.test.ts` that checks a certain date

src/credentials/sso/cache.ts
```diff
@@ -31,32 +32,40 @@ export interface SsoCache {
...
-        scopes.forEach(s => shasum.update(s))
-        return shasum.digest('hex')
-    }
-
-    const getTarget = (key: RegistrationKey) => {
-        const suffix = `${key.region}${key.scopes && key.scopes.length > 0 ? `-${hashScopes(key.scopes)}` : ''}`
-        return join(directory, `aws-toolkit-vscode-client-id-${suffix}.json`)
-    }
+// Iterate over each file in the directory variable, run isFileValid in each file and if it is invalid delete the file
+function deleteInvalidFiles(directory: string, statFunc: typeof getFileStats) {
```
JadenSimon, line 48:
```suggestion
function deleteOldFiles(directory: string, statFunc: typeof getFileStats) {
```

src/credentials/sso/cache.ts
```diff
@@ -110,24 +119,59 @@ export function getTokenCache(directory = cacheDir): KeyedCache<SsoAccess> {
...
+    return statSync(file)
+}
+
+export const firstValidDate = new Date(2023, 0, 1) // April 1, 2023
+
+/**
+ * @param file path to file
+ * @returns false if we can verify the file was created before the first valid date, true otherwise
+ */
+function isFileValid(file: string, statFunc: typeof getFileStats): boolean {
```
JadenSimon, line 138:
```suggestion
function isOldFile(file: string, statFunc: typeof getFileStats): boolean {
```

src/credentials/sso/cache.ts
```diff
@@ -31,32 +32,40 @@ export interface SsoCache {
...
-
-    const getTarget = (key: RegistrationKey) => {
-        const suffix = `${key.region}${key.scopes && key.scopes.length > 0 ? `-${hashScopes(key.scopes)}` : ''}`
-        return join(directory, `aws-toolkit-vscode-client-id-${suffix}.json`)
-    }
+function deleteOldFiles(directory: string, statFunc: typeof getFileStats) {
+    const fileNames = readdirSync(directory)
+    fileNames.forEach(fileName => {
+        const filePath = join(directory, fileName)
+        if (!isOldFile(filePath, statFunc)) {
```
justinmk3, line 52: When deleting files in a loop it's worth being paranoid. If somehow `/` or `.` is passed as `directory`, or any other weird case, it would be catastrophic. 

Some guards that can prevent catastrophe:

- check that `directory` is absolute
- check that the trimmed result of `directory` is non-empty and has length greater than, say, 5

src/credentials/sso/cache.ts
```diff
@@ -31,32 +32,40 @@ export interface SsoCache {
 
 const cacheDir = join(homedir(), '.aws', 'sso', 'cache')
 
-export function getCache(directory = cacheDir): SsoCache {
+export function getCache(directory = cacheDir, statFunc = getFileStats): SsoCache {
+    try {
+        deleteOldFiles(directory, statFunc)
```
justinmk3, line 37: how often is `getCache` called? will iterating all the files cause a lot of unnecessary filesystem activity?

src/credentials/sso/cache.ts
```diff
@@ -110,24 +119,59 @@ export function getTokenCache(directory = cacheDir): KeyedCache<SsoAccess> {
...
+function isOldFile(file: string, statFunc: typeof getFileStats): boolean {
+    try {
+        const statResult = statFunc(file)
+        // Depending on the Windows filesystem, birthtime may be 0, so we fall back to ctime (last time metadata was changed)
+        // https://nodejs.org/api/fs.html#stat-time-values
+        return statResult.birthtimeMs !== 0
+            ? statResult.birthtimeMs >= firstValidDate.getTime()
+            : statResult.ctime >= firstValidDate
+    } catch (err) {
+        getLogger().debug(`SSO cache file ${file} could not be validated: ${err}`)
```
justinmk3, line 147:
```suggestion
        getLogger().debug(`SSO cache file could not be validated: ${file} : ${err}`)
```

src/credentials/sso/cache.ts
JadenSimon: `deleteOldFiles` will have some impact on performance though it shouldn't be massive. Could add a module-scoped flag to skip if set. And then only set it when not running as a test.

nkomonen-amazon: Also we will be deleting this functionality in the future

justinmk3: Yeah, wonder if `getCache` itself is hot enough to warrant an optimization.  Out of scope for this PR, though.

src/credentials/sso/cache.ts
```diff
@@ -33,32 +34,56 @@ export interface SsoCache {
...
 
-    const getTarget = (key: RegistrationKey) => {
-        const suffix = `${key.region}${key.scopes && key.scopes.length > 0 ? `-${hashScopes(key.scopes)}` : ''}`
-        return path.join(directory, `aws-toolkit-vscode-client-id-${suffix}.json`)
-    }
+    const fileNames = readdirSync(directory)
+    fileNames.forEach(fileName => {
+        const filePath = path.join(directory, fileName)
+        if (!isOldFile(filePath, statFunc)) {
+            unlinkSync(filePath)
```
JadenSimon, line 60:
```suggestion
            unlinkSync(filePath)
            getLogger().warn(`auth: removed old cache file: ${filePath}`)
```

src/credentials/sso/cache.ts
```diff
@@ -33,32 +34,56 @@ export interface SsoCache {
...
+}
+
+export function isSafeFilePath(filePath: string): boolean {
+    const resolvedPath = path.resolve(filePath)
+    const isRoot = resolvedPath === path.resolve('/')
+    const isCwd = resolvedPath === path.resolve('.')
+    const isAbsolute = path.isAbsolute(filePath)
+    const pathDepth = resolvedPath.split(path.sep).length
+
+    const isSafe = !isRoot && !isCwd && isAbsolute && pathDepth >= 5
```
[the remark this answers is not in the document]
nkomonen-amazon: I renamed this func, it is checking that the directory is 'safe' not the file. But I will add this ext check to the correct part of the code

src/credentials/sso/cache.ts
```diff
@@ -112,24 +137,59 @@ export function getTokenCache(directory = getCacheDir()): KeyedCache<SsoAccess>
...
+    return statSync(file)
+}
+
+export const firstValidDate = new Date(2023, 0, 1) // April 1, 2023
+
+/**
+ * @param file path to file
+ * @returns false if we can verify the file was created before the first valid date, true otherwise
+ */
+function isOldFile(file: string, statFunc: typeof getFileStats): boolean {
```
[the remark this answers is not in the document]
nkomonen-amazon: Fixed the conditional so this makes sense

PR #3367, opened 2023-04-24
fix(codewhisperer): credentials refresh does not happen on C9

JadenSimon:
## Problem
We can't tell if the default connection is expired or not on C9 until after making an API call. So the `reauthenticate` logic in `authUtil.ts` does not work on C9

## Solution
Remove the check to always force a refresh

## License

### DIFF
--- a/scripts/newChange.ts
+++ b/scripts/newChange.ts
@@ -7,6 +7,7 @@ import * as child_process from 'child_process'
 import * as fs from 'fs-extra'
 import { join } from 'path'
 import * as readlineSync from 'readline-sync'
+import * as crypto from 'crypto'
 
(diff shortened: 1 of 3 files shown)


scripts/newChange.ts
```diff
@@ -7,6 +7,7 @@ import * as child_process from 'child_process'
 import * as fs from 'fs-extra'
 import { join } from 'path'
 import * as readlineSync from 'readline-sync'
+import * as crypto from 'crypto'
```
justinmk3, line 10: I definitely tested this. Wonder why it worked for me without this.

justinmk3 (approved):

PR #3375, opened 2023-04-25
auth refactoring

nkomonen-amazon:
Doing some refactoring of the auth related code.

Includes naming changes, extracting logic in to its own function, centralizing code.

See commit messages for specific info on each change.

## License


src/credentials/wizards/templates.ts
```diff
@@ -11,30 +11,20 @@ import { getIdeProperties } from '../../shared/extensionUtilities'
...
-    Process = 'credential_process',
-}
+import { BaseCredentialsProfileData } from '../sharedCredentials'
+import { sharedCredentialProperties } from '../providers/sharedCredentialsProvider'
 
 function getTitle(profileName: string): string {
     return localize('AWS.title.createCredentialProfile', 'Creating new profile "{0}"', profileName)
 }
 
-interface StaticProfile {
```
nkomonen-amazon, line 26: This is replaced by `BaseCredentialsProfileData` in this commit, we will be centralizing these types since they will have use outside of this file.

src/credentials/sharedCredentials.ts
```diff
@@ -234,6 +234,13 @@ async function loadCredentialsFile(credentialsUri?: vscode.Uri): Promise<ReturnT
     return parseIni(await SystemUtilities.readFile(credentialsUri), credentialsUri)
 }
 
+/**
+ * The name of a section in a credentials/config file
+ *
+ * The is the value of `{A}` in `[ {A} ]` or `[ {B} {A} ]`.
+ */
+export type ProfileName = string
```
JadenSimon, line 242: `SectionName` might be less confusing. `profile` sections need to be handled differently from `sso-session` sections. But both are named.

src/credentials/types.ts
```diff
@@ -0,0 +1,35 @@
+/*!
```
JadenSimon, line 1: Calling this file "types" is somewhat misleading but I can't really think of anything else.

JadenSimon (approved):

src/credentials/types.ts
nkomonen-amazon: One thing I was thinking is that this would instead be named 'sharedCredentials.ts`, and the existing `sharedCredentials.ts` which holds the parsing related code of credentials could go to a new file like `sharedCredentialsParser.ts`

JadenSimon: I've changed my mind. `types.ts` is appropriate because `SharedCredentialsKeys` is being used as a enum. The typescript compiler follows a similar pattern.

PR #3384, opened 2023-04-26
fix(codewhisperer): Fixes some tests that fail when run locally

nkomonen-amazon:
When unit tests are run locally + CW is setup with Builder ID the `credentialStartUrl` value will be defined, this causes unit tests to fail.

To fix it, this change adds to the assertion whatever the current `startUrl` is.

- Additionally, some tests assume that CW connection is not configured automatically. This ensures the connection is not set up and prevents any failures due to the test env.

For more info see sim IDE-10578

## License

### DIFF
--- a/src/test/codewhisperer/explorer/codewhispererNode.test.ts
+++ b/src/test/codewhisperer/explorer/codewhispererNode.test.ts
@@ -10,6 +10,17 @@ import { codewhispererNode } from '../../../codewhisperer/explorer/codewhisperer
 import { AuthUtil } from '../../../codewhisperer/util/authUtil'
 
 describe('codewhispererNode', function () {
+    let isConnectionValid: sinon.SinonStub
+    let isConnected: sinon.SinonStub
+
+    beforeEach(function () {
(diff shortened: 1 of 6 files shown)


justinmk3 (approved):
waiting for CW approval

src/test/codewhisperer/explorer/codewhispererNode.test.ts
```diff
@@ -10,6 +10,17 @@ import { codewhispererNode } from '../../../codewhisperer/explorer/codewhisperer
 import { AuthUtil } from '../../../codewhisperer/util/authUtil'
 
 describe('codewhispererNode', function () {
+    let isConnectionValid: sinon.SinonStub
+    let isConnected: sinon.SinonStub
+
+    beforeEach(function () {
+        isConnectionValid = sinon.stub(AuthUtil.instance, 'isConnectionValid')
```
nkomonen-amazon, line 17: This block of code sets lots of auth stuff to not being connected initially, any following test that needs it to be connected (returns true) will set it on its own.

leigaol (approved):

PR #3387, opened 2023-04-26
test: run git-secrets during CI

nkomonen-amazon:
This PR

- Creates a new test directory called `testLint/` which will hold our code linting related tests
  - Creates a git-secrets test in `testLint/` which runs `git-secrets` on our codebase
  - Creates a eslint test in `testLint/` which runs `eslint` on our codebase
- Adds a new CI job that runs the `testLint/` tests as well as the existing code coverage tests which were running in the regular tests before
- Can run the lint tests with the new `launch.json` command or `npm run lint` from cli

## License

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -83,3 +83,22 @@ jobs:
                   verbose: true
                   file: ./coverage/coverage-final.json
                   flags: windows-unittests
+
+    lint:
+        name: Ubuntu nodejs Lint
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                node-version: [16.x]
+                vscode-version: [stable]
+        env:
+            NODE_OPTIONS: '--max-old-space-size=8192'
+        steps:
+            - uses: actions/checkout@v3
+            - name: Use Node.js ${{ matrix.node-version }}
+              uses: actions/setup-node@v3
+              with:
+                  node-version: ${{ matrix.node-version }}
+            - run: npm ci
+            - run: npm run testCompile
+            - run: npm run lint
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -137,6 +137,14 @@
             },
             "outFiles": ["${workspaceFolder}/dist/**/*.js"],
             "preLaunchTask": "${defaultBuildTask}"
+        },
+        {
+            "name": "Test Lint",
+            "type": "node",
+            "request": "launch",
+            "program": "${workspaceFolder}/scripts/lint/testLint.ts",
+            "outFiles": ["${workspaceFolder}/dist/**/*.js"],
+            "preLaunchTask": "${defaultBuildTask}"
         }
     ],
     "compounds": [
--- a/package.json
+++ b/package.json
@@ -3431,7 +3431,7 @@
         }
     },
     "scripts": {
-        "vscode:prepublish": "npm run clean && npm run buildScripts && npm run lint && webpack --mode production && npm run copyFiles -- --webpacked",
+        "vscode:prepublish": "npm run clean && npm run buildScripts && webpack --mode production && npm run copyFiles -- --webpacked",
         "clean": "ts-node ./scripts/clean.ts dist",
         "reset": "npm run clean -- node_modules && npm install",
         "copyFiles": "ts-node ./scripts/build/copyFiles.ts",
@@ -3443,7 +3443,7 @@
         "test": "npm run testCompile && ts-node ./scripts/test/test.ts && npm run report",
         "testE2E": "npm run testCompile && ts-node ./scripts/test/testE2E.ts && npm run report",
         "integrationTest": "npm run testCompile && ts-node ./scripts/test/integrationTest.ts && npm run report",
-        "lint": "eslint -c .eslintrc.js --ext .ts .",
+        "lint": "ts-node ./scripts/lint/testLint.ts",
         "lintfix": "eslint -c .eslintrc.js --fix --ext .ts .",
         "package": "ts-node ./scripts/build/package.ts",
         "install-plugin": "vsce package -o aws-toolkit-vscode-test.vsix && code --install-extension aws-toolkit-vscode-test.vsix",
--- a/scripts/lint/testLint.ts
+++ b/scripts/lint/testLint.ts
@@ -1,0 +1,29 @@
+/*!
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { glob } from 'glob'
+import * as Mocha from 'mocha'
+;(async () => {
+    try {
+        console.log('Running linting tests...')
+
+        const mocha = new Mocha()
+
+        const testFiles = glob.sync('dist/src/testLint/**/*.test.js')
+        testFiles.forEach(file => {
+            mocha.addFile(file)
+        })
+
+        mocha.run(failures => {
+            const exitCode = failures ? 1 : 0
+            console.log(`Finished running Main test suite with result code: ${exitCode}`)
+            process.exit(exitCode)
+        })
+    } catch (err) {
+        console.error(err)
+        console.error('Failed to run tests')
+        process.exit(1)
+    }
+})()
--- a/src/test/gitSecrets.test.ts
+++ b/src/test/gitSecrets.test.ts
@@ -1,0 +1,146 @@
+/*!
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe } from 'mocha'
+
+import { SpawnSyncOptions, spawnSync } from 'child_process'
+import * as assert from 'assert'
+import { getProjectDir } from './testUtil'
+import * as path from 'path'
+import { platform } from 'os'
+import { existsSync, mkdirSync, unlinkSync, writeFileSync } from 'fs'
+
+/**
+ * NOTES:
+ * - git-secrets patterns are set in the project's `.git/config` file
+ */
+describe('git-secrets', function () {
+    let accessKeyFilePath: string
+    let toolkitProjectDir: string
+    let testFixturesPath: string
+    let gitSecrets: string // path to the git-secrets executable
+
+    /** git-secrets patterns that will not cause a failure during the scan */
+    function setAllowListPatterns(gitSecrets: string) {
+        const allowListPatterns: string[] = ['"accountId": "123456789012"']
+
+        allowListPatterns.forEach(pattern => {
+            // Returns non-zero exit code if pattern already exists
+            runCmd([gitSecrets, '--add', '--allowed', pattern], { cwd: toolkitProjectDir, throws: false })
+        })
+    }
+
+    function setDenyListPatterns(gitSecrets: string) {
+        const denyListPatterns: string[] = []
+
+        denyListPatterns.forEach(pattern => {
+            // Returns non-zero exit code if pattern already exists
+            runCmd([gitSecrets, '--add', pattern], { cwd: toolkitProjectDir, throws: false })
+        })
+    }
+
+    function setupTestFixturesDir(toolkitProjectDir: string) {
+        const testFixturesPath = path.join(toolkitProjectDir, 'src', 'testFixtures', 'bin')
+        mkdirSync(testFixturesPath, { recursive: true })
+        return testFixturesPath
+    }
+
+    function setupAccessKeyFile(testFixturesPath: string) {
+        const accessKeyFilePath = path.join(testFixturesPath, 'fileWithAccessKey.ts')
+        deleteFileIfExists(accessKeyFilePath)
+        return accessKeyFilePath
+    }
+
+    function setupGitSecretsExecutable(testFixturesPath: string) {
+        const gitSecretsExecutablePath = path.join(testFixturesPath, 'git-secrets')
+
+        if (existsSync(gitSecretsExecutablePath)) {
+            console.log('INFO: git-secrets already installed')
+        } else {
+            console.log('INFO: Installing git-secrets...')
+            runCmd(['mkdir', '-p', path.parse(gitSecretsExecutablePath).dir])
+            runCmd([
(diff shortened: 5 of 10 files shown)


.github/workflows/node.js.yml
```diff
@@ -21,6 +21,11 @@ jobs:
             VSCODE_TEST_VERSION: ${{ matrix.vscode-version }}
             NODE_OPTIONS: '--max-old-space-size=8192'
         steps:
+            - name: Install git-secrets
+              run: |
+                  brew update
+                  brew install git-secrets
+                  brew upgrade
```
justinmk3, line 28: brew update/upgrade can be really slow, so this is a high time-cost in our CI just for git secrets :) 

isn't it just a script? can we `curl` the raw script perhaps?

.github/workflows/node.js.yml
```diff
@@ -21,6 +21,13 @@ jobs:
             VSCODE_TEST_VERSION: ${{ matrix.vscode-version }}
             NODE_OPTIONS: '--max-old-space-size=8192'
         steps:
+            # Setup for git-secrets
+            - name: Setup git-secrets executable
+              run: |
+                  mkdir $HOME/bin
+                  curl -o $HOME/bin/git-secrets https://raw.githubusercontent.com/awslabs/git-secrets/99d01d58ebcc06e237c0e3f3ff5ae628aeef6aa6/git-secrets
+                  chmod +x $HOME/bin/git-secrets
```
justinmk3, line 29: not in $PATH.

But anyway, since we have a test file dedicate for this, let's move the `curl` logic into the test file. That avoids the ambiguity related to $PATH (which could result in different git-secrets versions if run locally, for example)

src/test/gitSecrets.test.ts
```diff
@@ -0,0 +1,63 @@
...
+import { spawnSync } from 'child_process'
+import * as assert from 'assert'
+import { getProjectDir } from './testUtil'
+import * as path from 'path'
+import { platform } from 'os'
+
+/**
+ * secrets patterns are set in the project root, .git/config
+ */
+describe('Test git-secrets scan', function () {
```
justinmk3, line 17: "test" is redundant in a test description
```suggestion
describe('git-secrets', function () {
```

src/test/gitSecrets.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        projectDir = path.join(getProjectDir(), '..', '..')
+        const result = spawnSync('cat', ['.git/config'], { cwd: projectDir })
+        console.log(`config ls: ${result.stdout.toString()}`)
+        spawnSync('git-secrets', ['--register-aws'], { cwd: projectDir })
+        setAllowListPatterns()
+        setDenyListPatterns()
+    })
+
+    it('ensures no git secrets are found', function () {
+        let result = spawnSync('which', ['git-secrets'], { cwd: projectDir })
```
justinmk3, line 49: by doing the `curl` in this test, we can avoid hacks like `which`. We can download the script to an explicit location and call that location.

src/test/gitSecrets.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        assert.strictEqual(result.status, 0, 'Could not find `git-secrets` in PATH')
```
justinmk3, line 50: by doing the `curl` in this test, we can avoid hacks like `which`. We can download the script to an explicit location and call that location.

src/test/gitSecrets.test.ts
```diff
@@ -0,0 +1,108 @@
...
+        const result = spawnSync(gitSecrets, ['--scan', '--no-index'], { cwd: toolkitProjectDir })
+        assert.strictEqual(result.status, 0, `Failure output: ${result.stderr.toString()}`)
+    })
+
+    it('sanity check it finds secrets', function () {
+        // Create file in project that has secret key value.
+        // Need to build access key string incrementally to not trigger git-secrets.
+        const keyValue = 'yAki21XLhAIBiKvyaxr4p/ltr8OxkZTHISISFAKE'
+        const mySecretAccessKey = `const x = { "aws_secret_access_key": "${keyValue}" }`
+        const fileContent = dedent(
```
justinmk3, line 93: This is nice to have, but is it worth an extra dependency? looks like it has no transitive dependencies, but still not sure.

src/test/gitSecrets.test.ts
```diff
@@ -0,0 +1,108 @@
...
+            unlinkSync(accessKeyFilePath)
+        }
+    })
+
+    it('ensures no git secrets are found', function () {
+        const result = spawnSync(gitSecrets, ['--scan', '--no-index'], { cwd: toolkitProjectDir })
+        assert.strictEqual(result.status, 0, `Failure output: ${result.stderr.toString()}`)
+    })
+
+    it('sanity check it finds secrets', function () {
```
justinmk3, line 88: good idea

src/test/gitSecrets.test.ts
```diff
@@ -0,0 +1,108 @@
...
+import * as path from 'path'
+import { platform } from 'os'
+import { existsSync, mkdirSync, unlinkSync, writeFileSync } from 'fs'
+import dedent from 'ts-dedent'
+
+/**
+ * NOTES:
+ * - git-secrets patterns are set in the project's `.git/config` file
+ */
+describe('Ensure no secrets detected in project', function () {
```
justinmk3, line 20: this gets prepended to the `it('ensures no git secrets are found'` below. so it's usually fine to keep the describe() titles very terse.

```suggestion
describe('git-secrets', function () {
```

src/test/gitSecrets.test.ts
nkomonen-amazon: Not a necessity to me, will remove

src/test/gitSecrets.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    before(function () {
+        if (platform() === 'win32') {
+            this.skip()
+        }
+
+        toolkitProjectDir = path.join(getProjectDir(), '..', '..')
+        testFixturesPath = path.join(toolkitProjectDir, 'src', 'testFixtures', 'resources')
```
justinmk3, line 72:
```suggestion
        testFixturesPath = path.join(toolkitProjectDir, 'src', 'testFixtures', 'bin')
```

src/test/gitSecrets.test.ts
```diff
@@ -0,0 +1,133 @@
...
+`.trim()
+        writeFileSync(accessKeyFilePath, fileContent)
+
+        const result = runCmd([gitSecrets, '--scan', '--no-index'], { cwd: toolkitProjectDir, throws: false })
+
+        assert.strictEqual(result.status, 1)
+    })
+})
+
+function runCmd(args: string[], options?: SpawnSyncOptions & { throws?: boolean }) {
```
justinmk3, line 117: does using our existing `src/shared/utilities/childProcess.ts` avoid the need for this?

nkomonen-amazon:
@justinmk3 See the PR description with new updated information

scripts/test/testQA.ts
```diff
@@ -0,0 +1,38 @@
...
+        console.log('Running Main test suite...')
+
+        const vsCodeExecutablePath = await setupVSCodeTestInstance()
+        const rootDir = process.cwd()
+        const testEntrypoint = resolve(rootDir, 'dist/src/testQA/index.js')
+        const testWorkspace = resolve(rootDir, 'src/testFixtures/workspaceFolder')
+        console.log(`Starting tests: ${testEntrypoint}`)
+
+        const result = await runTests({
+            vscodeExecutablePath: vsCodeExecutablePath,
```
justinmk3, line 20: Not sure we want the lint job to be driven by the mocha test runner. When we later add linting, that will be driven by eslint. Invoking that from mocha seems like too many layers. It also requires starting vscode, which is slow.

Instead this "test" could live in `scripts/lint/`. Later we may add more lint rules there. That makes linting faster than invoking via mocha+vscode. 

Are there advantages to using the test runner? 

- nice reporting
- common structure via it(), describe()
- ...?

.github/workflows/node.js.yml
```diff
@@ -51,15 +62,15 @@ jobs:
...
                   file: ./coverage/coverage-final.json
-                  flags: codewhisperer
+                  flags: windows-unittests
 
-    windows:
-        name: Windows nodejs
-        runs-on: windows-2019
+    qa:
+        name: macOS nodejs QA
+        runs-on: macos-latest
```
justinmk3, line 69: linux will have a faster filesystem and faster package installs. any reason to macos?

scripts/test/testQA.ts
nkomonen-amazon: I updated this so that we don't used the vscode test runner, instead running directly through the Mocha runner which is instantaneous.

The advantages I see are what you listed. It matches the same format as all our other tests, so easy to understand.
- Any necessary setup can be easily done and is isolated to that test (eg: setting up fake files to test against)

.vscode/launch.json
```diff
@@ -137,6 +137,24 @@
             },
             "outFiles": ["${workspaceFolder}/dist/**/*.js"],
             "preLaunchTask": "${defaultBuildTask}"
+        },
+        {
+            "name": "Test QA (current file)",
```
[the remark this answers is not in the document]
nkomonen-amazon: I've updated this to use the 'lint' terminology. I find this useful so we can debug the tests if needed.

.github/workflows/node.js.yml
```diff
@@ -32,26 +32,6 @@ jobs:
               uses: GabrielBB/xvfb-action@v1
               with:
                   run: npm test
-            - name: Code coverage
```
nkomonen-amazon, line 35: These are all moved to the new 'lint' job

scripts/lint/testLint.ts
```diff
@@ -0,0 +1,30 @@
+/*!
```
nkomonen-amazon, line 1: Runs all our lint tests without the need to spin up vscode. Much faster.

src/testLint/testUtils.ts
```diff
@@ -0,0 +1,24 @@
+/*!
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { SpawnSyncOptions, spawnSync } from 'child_process'
+
+export function runCmd(args: string[], options?: SpawnSyncOptions & { throws?: boolean }) {
```
nkomonen-amazon, line 8: Cannot reuse existing `ChildProcess` class since that relies on vscode api

package.json
```diff
@@ -3431,7 +3431,7 @@
         }
     },
     "scripts": {
-        "vscode:prepublish": "npm run clean && npm run buildScripts && npm run lint && webpack --mode production && npm run copyFiles -- --webpacked",
```
nkomonen-amazon, line 3434: Removing lint from this command since linting errors will be caught in the new linting CI tests

.github/workflows/node.js.yml
justinmk3: codecoverage must stay here because it depends on `npm test`

.github/workflows/node.js.yml
```diff
@@ -83,3 +63,47 @@ jobs:
...
+            NODE_OPTIONS: '--max-old-space-size=8192'
+        steps:
+            - uses: actions/checkout@v3
+            - name: Use Node.js ${{ matrix.node-version }}
+              uses: actions/setup-node@v3
+              with:
+                  node-version: ${{ matrix.node-version }}
+            - run: npm ci
+
+            - name: Test Lint
```
justinmk3, line 84:
```suggestion
            - name: Lint
```

src/testLint/README.md
```diff
@@ -0,0 +1,3 @@
+# Lint Tests
+
+These are linting related tests that must pass in CI
```
justinmk3, line 3:
```suggestion
These are linting related tests that must pass in CI. Lint code is nodejs-only, must not use vscode API.
```

.github/workflows/node.js.yml
```diff
@@ -83,3 +83,25 @@ jobs:
...
+        steps:
+            - uses: actions/checkout@v3
+            - name: Use Node.js ${{ matrix.node-version }}
+              uses: actions/setup-node@v3
+              with:
+                  node-version: ${{ matrix.node-version }}
+            - run: npm ci
+
+            - name: Lint
+              uses: GabrielBB/xvfb-action@v1
```
justinmk3, line 105: xvfb is for graphics. Shouldn't be needed since we don't run vscode in this job.

package.json
```diff
@@ -3443,7 +3443,7 @@
         "test": "npm run testCompile && ts-node ./scripts/test/test.ts && npm run report",
         "testE2E": "npm run testCompile && ts-node ./scripts/test/testE2E.ts && npm run report",
         "integrationTest": "npm run testCompile && ts-node ./scripts/test/integrationTest.ts && npm run report",
-        "lint": "eslint -c .eslintrc.js --ext .ts .",
+        "lint": "npm run testCompile && ts-node ./scripts/lint/testLint.ts",
```
justinmk3, line 3446: Is it possible to avoid `testCompile` here? That should normally be the responsibility of the caller. So `npm run lint` will be fast if I know that I already have an up-to-date `dist/`. It might be worth adding a sanity check in `testLint.ts` which checks that `dist/` exists and is non-empty.

nkomonen-amazon: Updated to not run compile first, will look in to ensuring `dist` exists in a separate pr

src/testLint/gitSecrets.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        setDenyListPatterns(gitSecrets)
+        setAllowListPatterns(gitSecrets)
+    })
+
+    afterEach(function () {
+        deleteFileIfExists(accessKeyFilePath)
+    })
+
+    it('ensures no git secrets are found', function () {
+        const result = runCmd([gitSecrets, '--scan'], { cwd: toolkitProjectDir })
```
justinmk3, line 118: ✔ ensures no git secrets are found (68ms)

68ms to scan the whole codebase is fast, so I guess we don't need to pass the "diff".

justinmk3 (approved):

PR #3388, opened 2023-04-26
refactor: add basic module system

JadenSimon:
## Problem
No isolation between components of the extension. 

## Solution
Add `modules.ts` to handle registration and activation of modules. A build script discovers all modules (with some restrictions) and renders them into a single file. Each module is expected to have an `activate` function just like a normal VS Code extension. Modules can declare additional parameters beyond `vscode.ExtensionContext` but they must be typed as return values of other modules. 

Modules are registered with dynamic imports so they are always loaded lazily (even with `webpack`!). That is, the file itself is not loaded into memory until the `activate` is called. Of course we activate every single thing on startup right now so it doesn't do much yet.

I've refactored many of our components but not all of them. I primarily stuck to the simpler ones to avoid making massive changes.

This is what's currently generated:
```ts
/*!
 * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 */

import * as vscode from 'vscode';
import { register, activateAll, ModuleApi } from './shared/modules';

export const extcontextModule = register('extcontext', () => import('./extcontext'));
export const apigatewayModule = register('apigateway', () => import('./apigateway'), extcontextModule);
export const apprunnerModule = register('apprunner', () => import('./apprunner'));
export const awsexplorerModule = register('awsexplorer', () => import('./awsexplorer'), extcontextModule);
export const cdkModule = register('cdk', () => import('./cdk'), awsexplorerModule);
export const cloudWatchLogsModule = register('cloudWatchLogs', () => import('./cloudWatchLogs'));
export const codecatalystModule = register('codecatalyst', () => import('./codecatalyst'), extcontextModule, awsexplorerModule);
export const codewhispererModule = register('codewhisperer', () => import('./codewhisperer'), awsexplorerModule);
export const devModule = register('dev', () => import('./dev'));
export const dynamicResourcesModule = register('dynamicResources', () => import('./dynamicResources'));
export const ecrModule = register('ecr', () => import('./ecr'));
export const ecsModule = register('ecs', () => import('./ecs'));
export const eventSchemasModule = register('eventSchemas', () => import('./eventSchemas'), extcontextModule);
export const iotModule = register('iot', () => import('./iot'), extcontextModule);
export const lambdaModule = register('lambda', () => import('./lambda'), extcontextModule);
export const s3Module = register('s3', () => import('./s3'), extcontextModule);
export const ssmDocumentModule = register('ssmDocument', () => import('./ssmDocument'), extcontextModule);
export const stepFunctionsModule = register('stepFunctions', () => import('./stepFunctions'), extcontextModule);

export type extcontext = ModuleApi<typeof extcontextModule>;
export type apigateway = ModuleApi<typeof apigatewayModule>;
export type apprunner = ModuleApi<typeof apprunnerModule>;
export type awsexplorer = ModuleApi<typeof awsexplorerModule>;
export type cdk = ModuleApi<typeof cdkModule>;
export type cloudWatchLogs = ModuleApi<typeof cloudWatchLogsModule>;
export type codecatalyst = ModuleApi<typeof codecatalystModule>;
export type codewhisperer = ModuleApi<typeof codewhispererModule>;
export type dev = ModuleApi<typeof devModule>;
export type dynamicResources = ModuleApi<typeof dynamicResourcesModule>;
export type ecr = ModuleApi<typeof ecrModule>;
export type ecs = ModuleApi<typeof ecsModule>;
export type eventSchemas = ModuleApi<typeof eventSchemasModule>;
export type iot = ModuleApi<typeof iotModule>;
export type lambda = ModuleApi<typeof lambdaModule>;
export type s3 = ModuleApi<typeof s3Module>;
export type ssmDocument = ModuleApi<typeof ssmDocumentModule>;
export type stepFunctions = ModuleApi<typeof stepFunctionsModule>;

export function activateModules(ctx: vscode.ExtensionContext) {
    return activateAll(ctx);
}
```

## TODO
Need architecture documentation

## License


JadenSimon:
VSIX size went up by ~1.2MB because of webpack code-splitting. Which is more than what I would expect. Looking at the emits, I think webpack is still placing most of the source code in a single blob. Some changes to the config are needed.

JadenSimon:
With code splitting the VSIX is ~300KB bigger compared to mainline. No change in the time it takes to package things.

The activation time appears to have been cut in half. An artifact from mainline usually takes over 800ms whereas I consistently see <400ms with this branch.

PR #3391, opened 2023-04-27
tests(codecatalyst): improve reporting when tests fail

JadenSimon:
## License


justinmk3 (approved):

justinmk3:
> Error: Remote SSH extension version is too low [ExtensionVersionTooLow] (expected: 0.98.0)

Oh, do we actually need 0.98 or could we lower the min version?

nkomonen-amazon:
@justinmk3 from IDE-10402 it was mentioned the user had an error with 0.68 and that when they upgraded to 0.98 it was fixed. So no explicit need for 0.98.

If we have a version we want to check we can install that version and reload vscode to see if it breaks

JadenSimon:
Remote SSH `0.74.0` is compatible with minver. Installing this on stable still allows me to connect.

PR #3407, opened 2023-05-02
fix(auth): static credentials without an expiration are not coherent with the source 

JadenSimon:
## Problem
It's possible to load temporary credentials without an explicit expiration time. Cloud9 writes credentials in this way. Without an expiration time, we assume they are good forever even when the underlying source changes. 

## Solution
Fetch a new provider on every call. `CredentialsProviderManager` calls `fs.stat` every time which might have performance implications. This really only affects CW on C9, need to test to see if it's noticeable.

## License


src/test/credentials/auth.test.ts
```diff
@@ -360,6 +365,50 @@ describe('Auth', function () {
         })
     })
 
+    describe('Shared ini files', function () {
+        let tmpDir: string
+
+        beforeEach(async function () {
+            tmpDir = await makeTemporaryToolkitFolder()
+            sinon.stub(SystemUtilities, 'getHomeDirectory').returns(tmpDir)
```
justinmk3, line 373: would it be more robust to temporarily set $HOME?

src/credentials/auth.ts
```diff
@@ -760,16 +762,17 @@ export class Auth implements AuthService, ConnectionManager {
         )
     }
 
-    private getIamConnection(id: Connection['id'], provider: CredentialsProvider): IamConnection & StatefulConnection {
-        const profile = this.store.getProfileOrThrow(id)
-
+    private getIamConnection(
```
justinmk3, line 765: is it worth adding a comment here to call out the perf implications?

```suggestion
    /** Reads the filesystem on each call. */
    private getIamConnection(
```

justinmk3 (approved):

src/test/credentials/auth.test.ts
JadenSimon: Home directory behavior is dependent on the operating system so we'd need to change `USERPROFILE` too. I think a stub is OK here as we're not really stubbing out much logic anyway

src/credentials/auth.ts
JadenSimon: So this specific method doesn't make any `fs` calls, it's only when you do `getCredentials` on shared ini credentials. I think callers should be mindful of _any_ `async` operation as potentially impacting perf.

In general, VS Code API designs avoid making `get` operations asynchronous. Which is something we should strive for as well. Designing these types of systems can be a lot of effort though because we have to be aware of all operations, including those not within our program, that affect the state of the system.

PR #3410, opened 2023-05-02
cwl: record "Search Log Group" telemetry as a flow/span

justinmk3:
## License

### DIFF
--- a/src/cloudWatchLogs/changeLogSearch.ts
+++ b/src/cloudWatchLogs/changeLogSearch.ts
@@ -2,6 +2,7 @@
  * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
+import { CancellationError } from '../shared/utilities/timeoutUtils'
 import { telemetry, CloudWatchResourceType } from '../shared/telemetry/telemetry'
 import { showInputBox } from '../shared/ui/inputPrompter'
 import { createURIFromArgs, isLogStreamUri, recordTelemetryFilter } from './cloudWatchLogsUtils'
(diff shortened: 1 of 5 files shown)


src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -99,25 +99,19 @@ export async function searchLogGroup(
...
+        if (!response) {
+            throw new CancellationError('user')
+        }
 
-    result = await prepareDocument(uri, userResponse, registry)
-    telemetry.cloudwatchlogs_open.emit({ result: result, cloudWatchResourceType: 'logGroup', source: source })
+        const userResponse = handleWizardResponse(response, registry)
+        const uri = createURIFromArgs(userResponse.logGroupInfo, userResponse.parameters)
+        const result = await prepareDocument(uri, userResponse, registry)
+        telemetry.record({ result: result })
```
JadenSimon, line 113: Oh this will lead to unexpected behavior, I didn't implement any logic in `spans.ts` for "overriding" `result`.

The assumption is that, when using `run`, the instrumented code should not (and actually cannot) accurately record the outcome of its own execution.

justinmk3: > The assumption is that, when using run, the instrumented code should not (and actually cannot) accurately record the outcome of its own execution.

In this case I assume this means just that `prepareDocument` should either throw or not throw rather than return a `Result`.

JadenSimon: > In this case I assume this means just that prepareDocument should either throw or not throw rather than return a Result.

Yes sorry this is what I meant, my comment was a little cryptic. If a function cannot proceed it should raise an exception. If callers frequently need to handle those exceptions then using a monad like `Result` can make things more explicit. But either way exceptions are a divergence from the expected behavior of a function.

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -74,42 +75,38 @@ export async function changeLogSearchParams(
...
-    }
-
-    const oldData = registry.getRegisteredLog(oldUri)
-    const newData = await getNewData(param, oldData)
+    telemetry.cloudwatchlogs_open.run(async span => {
+        span.record({ source: 'Editor' })
+        const oldUri = getActiveDocumentUri(registry)
+        if (!registry.isRegistered(oldUri)) {
+            telemetry.cloudwatchlogs_filter.record({
+                source: 'Editor',
```
justinmk3, line 83: Should `result` be included in this `telemetry.cloudwatchlogs_filter.record()` call?

src/cloudWatchLogs/commands/searchLogGroup.ts
```diff
@@ -66,31 +67,26 @@ export async function prepareDocument(
     uri: vscode.Uri,
     logData: CloudWatchLogsData,
     registry: LogDataRegistry
-): Promise<Result> {
+) {
```
justinmk3, line 70: throw instead of return

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -74,42 +75,40 @@ export async function changeLogSearchParams(
...
-            cloudWatchResourceType: isLogStreamUri(oldUri) ? 'logStream' : 'logGroup',
-            hasTimeFilter: param === 'timeFilter',
-            hasTextFilter: param === 'filterPattern',
-        })
-        throw new Error(`cwl: Unable to find data for active URI ${oldUri}`)
-    }
-
-    const oldData = registry.getRegisteredLog(oldUri)
-    const newData = await getNewData(param, oldData)
+    telemetry.cloudwatchlogs_open.run(async span => {
```
JadenSimon, line 78:
```suggestion
    await telemetry.cloudwatchlogs_open.run(async span => {
```

src/cloudWatchLogs/changeLogSearch.ts
```diff
@@ -74,42 +75,40 @@ export async function changeLogSearchParams(
...
+        span.record({ source: 'Editor' })
+        const oldUri = getActiveDocumentUri(registry)
+        if (!registry.isRegistered(oldUri)) {
+            telemetry.cloudwatchlogs_filter.emit({
```
JadenSimon, line 82: Seems like `cloudwatchlogs_filter` is redundant now? Because we can record `hasTimeFilter` and `hasTextFilter` as soon as the user completes the prompt?

JadenSimon (approved):

src/cloudWatchLogs/changeLogSearch.ts
justinmk3: Good idea

JadenSimon (approved):

PR #3432, opened 2023-05-08
feat(telemetry): add `attributes` and `increment`

JadenSimon:
## Problem
* `spans.ts` can be simplified
* Some telemetry/logging is too noisy
* `vscode_executeCommand` is not a span

## Solution
Change the behavior to make it less confusing. Spans are now more independent. It's also now possible to record information on all future spans bound to a context.

### Other Changes
* Silence 4 CW commands that are triggered whenever auth changes
* Remove throttle on `aws.auth.addConnection`
    * I'd like to remove the throttle on everything eventually

## License


src/codewhisperer/commands/basicCommands.ts
```diff
@@ -132,9 +132,12 @@ export const showFreeTierLimit = Commands.declare('aws.codeWhisperer.freeTierLim
...
 })
 
-export const updateReferenceLog = Commands.declare('aws.codeWhisperer.updateReferenceLog', () => () => {
-    ReferenceLogViewProvider.instance.update()
-})
+export const updateReferenceLog = Commands.declare(
+    { id: 'aws.codeWhisperer.updateReferenceLog', logging: false },
+    () => () => {
+        ReferenceLogViewProvider.instance.update()
+    }
```
justinmk3, line 139: One thing the logs made obvious, is that this is doing stuff before the extension has had a chance to initialize. This declare() should probably be done in activate() somewhere rather than module-load time?

justinmk3: Probably out of scope for this PR though. But wanted to note it since we won't notice it in the logs now.

src/shared/telemetry/spans.ts
```diff
@@ -159,16 +167,25 @@ export class TelemetrySpan {
...
+    readonly attributes: Attributes
 }
 
 // This class is called 'Telemetry' but really it can be used for any kind of tracing
 // You would need to make `Span` a template type and reduce the interface to just create/start/stop
 export class TelemetryTracer extends TelemetryBase {
     readonly #context = new AsyncLocalStorage<TelemetryContext>()
 
+    /**
+     * `record` may be called prior to entering any span. This field simulates the
```
justinmk3, line 183: nit: avoid confusion with `Record` type
```suggestion
     * `record()` may be called prior to entering any span. This field simulates the
```

src/shared/telemetry/spans.ts
```diff
@@ -159,16 +167,25 @@ export class TelemetrySpan {
...
+     * effect of having an ephemeral "root" span that only extends until we enter
```
justinmk3, line 184: So this is like a "staging" area and can be used even before a span is entered?

src/shared/telemetry/spans.ts
```diff
@@ -183,33 +200,39 @@ export class TelemetryTracer extends TelemetryBase {
...
+    /**
+     * State that is applied to all new spans within the current or subsequent executions.
+     */
+    public get attributes(): Readonly<Attributes> {
+        return this.#context.getStore()?.attributes ?? this.#syncAttributes
     }
 
     /**
-     * Records information on all _current_ spans in the current execution context.
+     * Records information on all current and future spans in the execution context.
```
justinmk3, line 214: Does it apply to all future spans or only the _next_ span/context (which "pops it off the stack")?

src/shared/telemetry/spans.ts
```diff
@@ -146,11 +147,18 @@ export class TelemetrySpan {
...
         this.#startTime = undefined
     }
 
+    public increment(data: { [P in NumericKeys<T>]+?: number }): void {
+        for (const [k, v] of entries(data)) {
+            ;(this.state as Record<typeof k, number>)[k] = ((this.state[k] as number) ?? 0) + v!
+        }
+    }
+
+    // TODO: implement copy-on-write abstraction if this method causes perf issues
```
justinmk3, line 161: At that point we should look at a "persistent datatype" library rather than implementing this ourselves.

src/shared/utilities/tsUtils.ts
```diff
@@ -145,3 +145,5 @@ export type Mutable<T> = { -readonly [P in keyof T]: T[P] }
 export type FactoryFunction<T extends abstract new (...args: any[]) => any> = (
     ...args: ConstructorParameters<T>
 ) => InstanceType<T>
+
+export type NumericKeys<T> = { [P in keyof T]-?: T[P] extends number | undefined ? P : never }[keyof T]
```
justinmk3, line 149: is there a brief docstring that can give an intuitive description of the practical purpose of this

justinmk3 (approved):

src/codewhisperer/commands/basicCommands.ts
JadenSimon: > These declare() calls should probably be done in activate() somewhere, not at module-load?

`declare` doesn't register the command. Registration still happens in `activate`, `declare` simply allows you to reference commands in a type-safe way.

JadenSimon: Ah ok so those errors we were seeing was because some CW commands are only registered in `activate` but they may be invoked _before_ activation. Module-level registration would stop those errors from happening because it guarantees registration on initialization.

I don't see many downsides to registering commands on module-load because:
* Registering commands is cheap
* There is little need to register different functions dynamically on start-up
* Command registration doesn't trigger side-effects besides binding a function to an id

Commands are an extremely important part of the VS Code API, and to me it feels odd to hide them away in a single function (per-module). Also, if we _really_ need to, we can always defer actual command registration until extension initialization. The `Commands` class can simply wait to call `vscode.registerCommand` until told to do so.

src/shared/telemetry/spans.ts
JadenSimon: Yup. The ideal scenario is that all entry-points are covered by a span, but this isn't always practical.

src/shared/telemetry/spans.ts
JadenSimon: All future spans within the current and any nested contexts. 

The only exception is when you're not in a context. In this case it only affects the next span/context. But I think I'll remove this behavior and require that the caller be within a context.

src/shared/telemetry/spans.ts
JadenSimon: Yeah this comment isn't relevant anymore either. In one commit I had `clone()` being called a lot more.

src/shared/telemetry/spans.ts
```diff
@@ -232,6 +260,19 @@ export class TelemetryTracer extends TelemetryBase {
         }
     }
 
+    /**
+     * Executes the given function within an anonymous 'root' span which does not emit
+     * any telemetry on its own.
+     *
+     * This can be used as a 'staging area' for adding attributes prior to creating spans.
```
justinmk3, line 267: If runRoot() is called in multiple places , do all of their staged items get consumed by the first "real" span?

justinmk3 (approved):

src/shared/telemetry/spans.ts
JadenSimon: No, the staging area only exists within the callback function, and so all spans _within the context_ will inherit the staged items. The purpose of `runRoot()` is to guarantee that we are in a context. This keeps the behavior of everything consistent: `record()` should behave the same regardless of your "parent" span.

Now I just need to change the behavior of `record()` to not back-propagate attributes.

JadenSimon: > When would a caller not call runRoot()? Why do the callers need to decide when to call run() vs runRoot()?

I'm basically only using `runRoot()` for the purpose of establishing `source`. It doesn't need to exist if we add definitions for these entry points and attach the parent span name to children spans. The majority of callers _shouldn't_ use `runRoot()`, it's just a means to defer adding more definitions.

> Also wondering if we can rename emit to runFoo (or vice-versa: run => emitFoo, or let emit take a function parameter which causes it to behave like run()) so that developers understand the basic equivalence of all these things.

Yeah `emit` was only added to avoid refactoring everything to use `run`. `emit` can be removed once things are converted.

justinmk3: > Yeah `emit` was only added to avoid refactoring everything to use `run`. `emit` can be removed once things are converted.

Will help to mention practical stuff like that in the docstring on emit()

src/test/testUtil.ts
```diff
@@ -136,21 +137,30 @@ export function installFakeClock(): FakeTimers.InstalledClock {
...
+}
+
 /**
  * Finds the emitted telemetry metrics with the given `name`, then checks if the metadata fields
- * match the expected values, in the (non-continguous) order specified by `expected`.
+ * match the expected values, in the order specified by `expected`. Comparisons are done using
+ * {@link partialDeepCompare}. **Only fields present in {@link expected} will be checked.**
  *
  * @param name Metric name
- * @param expected Metric(s) shape(s) which are compared _in order_ (non-continguous) to metrics matching `name`.
```
JadenSimon, line 153: @justinmk3 The non-contiguous comparison wasn't working correctly for me. Was it originally added to deal with flakiness?

src/webviews/server.ts
```diff
@@ -29,48 +30,54 @@ export function registerWebviewServer(webview: vscode.Webview, commands: Protoco
...
 
-    const messageListener = webview.onDidReceiveMessage(async (event: Message) => {
-        const { id, command, data } = event
-        const metadata: Omit<Message, 'id' | 'command' | 'data'> = {}
+    const messageListener = webview.onDidReceiveMessage(
+        // XXX: In earlier versions of Node the first parameter was used as
+        // the `thisArg` for calling the bound function.
+        //
+        // Fixed in https://github.com/nodejs/node/commit/324a6c235a5bfcbcd7cc7491d55461915c10af34
+        AsyncResource.bind(async function (this: any, event: Message) {
```
JadenSimon, line 38: I almost didn't notice this bug. I'm surprised that Node didn't have any tests that execute the bound function with parameters.

src/test/testUtil.ts
justinmk3: it was added so that the caller didn't need to care about the exact sequence of metrics emitted

src/test/testUtil.ts
```diff
@@ -136,21 +136,30 @@ export function installFakeClock(): FakeTimers.InstalledClock {
...
 }
 
+// Note that Typescript is unable to describe the set of all supertypes of a type.
+// This should be typed something like `asserts actual is * extends T`
+export function partialDeepCompare<T>(actual: unknown, expected: T, message?: string): asserts actual is T {
+    if (typeof actual !== 'object' || !actual || typeof expected !== 'object' || !expected) {
+        assert.deepStrictEqual(actual, expected, message)
+    }
+
+    const partial = selectFrom(actual, ...keys(expected as object))
```
justinmk3, line 152: does this preserve order, so the assertion still asserts the (potentially non-contiguous) order?

src/webviews/server.ts
justinmk3: Nice find! I wonder if this explains https://github.com/aws/aws-toolkit-vscode/pull/3315#issuecomment-1506896655

src/test/testUtil.ts
JadenSimon: Ah ok, maybe some other change was causing issues. I'll try to add it back because it would be useful for testing sequences of events with different names.

src/test/testUtil.ts
JadenSimon: Yes it's order-preserving.

PR #3437, opened 2023-05-09
authWebview: Credentials Form semi-working

nkomonen-amazon:
# IGNORE THIS PR
See: https://github.com/aws/aws-toolkit-vscode/pull/3448

# Info

This PR sets up the form for the user to enter their credentials data, click submit, have their credentials saved to the `~/.aws/credentials` file + have the extension connect using them.

This is the first iteration of the form and can be improved/cleaned up. I will look to improve them in the next PR.

## How to flow works
- User can enter credentials details
- Shows errors if there is an error
  - If the profile name already exists
  - If the other data is improperly formatted
- Can only click submit button once all data is
  validated
- When submit is clicked it will add the credentials
  data to the credentials file, then use those credentials
  as the new connection
- If a connection exists and is already connected the form
  will show 'connected' in the title

### What is not addressed in this PR

- When successfully submitted it will propogate the
  result so that we can adjust other parts of the webview
  to show it is connected
- html/css fixes for layout
- Additional content in the container adjacent to the form

### The flow of entering credentials info
Kapture 2023-05-09 at 14 14 12

### Example error messages
Screen Shot 2023-05-09 at 2 29 45 PM

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.~~


src/credentials/vue/authForms/CredentialsForm.vue
```diff
@@ -0,0 +1,206 @@
+<template>
+    <div class="auth-form container-background border-common" id="credentials-form">
+        <FormTitle :isConnected="isConnected">IAM Credentials</FormTitle>
+
+        <div class="form-section">
+            <label class="small-description">Credentials will be added to the appropriate `.aws/` files.</label>
```
justinmk3, line 6: Looks like a relative path. "~/" is understood to mean home even on Windows. Could also consider showing the resolved path here.

```suggestion
            <label class="small-description">Credentials will be added to the appropriate `~/.aws/` files.</label>
```

src/credentials/vue/authForms/CredentialsForm.vue
```diff
@@ -0,0 +1,206 @@
+<template>
```
justinmk3, line 1: Existing webviews/forms are not named "form", the `.vue` extension makes it clear enough. Using `feedback/vue/submitFeedback.vue` as a precedent, this would be named `editCredentials.vue` or something like that.

nkomonen-amazon: I've re-opened a new PR since this was a draft.

What is the rationale for naming components verbs instead of nouns? I see a component as a piece of the page like a `footer` or `nav`.

> Existing webviews/forms are not named "form", the .vue extension makes it clear enough.
For example I have a component that is dedicated to being the title of a form since it is dynamic (can change to show an auth connected logo). Or a component that is a button. So the distinction seems necessary to me

justinmk3: The stakes are admittedly low in either case :) but in general, YAGNI (i.e. "don't worry").

> For example I have a component that is dedicated to being the title of a form since it is dynamic (can change to show an auth connected logo). Or a component that is a button. So the distinction seems necessary to me

`title.vue` or `fooTitle.vue`, right? Doesn't necessarily need to be a verb and the lack of "form" in the name doesn't risk confusion (though it's still technically a form).

Moreover, this particular file already lives in `authForms/` (how many auth forms are we going to have? probably not many).

PR #3438, opened 2023-05-09
codecatalyst: logging

justinmk3:
## Problem

- old outdated fields listed in codecatalyst API logs
- no timing/profiling info for codecatalyst API calls
- very long fields like `nextToken` are noisy in the logs

## Solution

- add timing/profiling info to logs
- truncate some fields like `nextToken`
    - this is experimental, maybe the field should just be omitted if the value is not useful. OTOH, truncating it is somewhat equivalent to writing `[omitted]` but gives a hint about the "shape" of the value.

## License

### DIFF
--- a/src/shared/logger/logger.ts
+++ b/src/shared/logger/logger.ts
@@ -4,6 +4,7 @@
  */
 
 import { Uri } from 'vscode'
+import globals from '../extensionGlobals'
 
 const toolkitLoggers: {
     main: Logger | undefined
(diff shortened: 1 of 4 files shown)


src/shared/logger/logger.ts
```diff
@@ -144,22 +145,24 @@ export function setLogger(logger: Logger | undefined, type?: 'channel' | 'debugC
 }
 
 export class PerfLog {
```
justinmk3, line 147: idk where this class is going, it's a minor experiment

Issue #3216, opened 2023-03-07
performance: avoid file watchers


justinmk3:
## Problem

Toolkit has various file watchers for SAM template.yaml files, CDK, etc. These can negatively affect performance.

## Solution

1. ~~Introduce settings to limit the cost of Toolkit file watchers:~~
1. ✅  exclude files/directories specified by vscode `search.exclude`, `files.exclude` settings. https://github.com/aws/aws-toolkit-vscode/pull/3770
1. ✅ Change `template.yaml` detect to be _on-demand_ instead of eager (omnipresent, filewatcher-driven)
    - When a `template.yaml` is needed but not found (such as "SAM run/debug", and "Sync SAM" command), mention this in the info message. (Related: [#3211](https://github.com/aws/aws-toolkit-vscode/pull/3211))
    - Implemented in #3931 + #3962

justinmk3:
The "schema service" in particular can be a complete show-stopper if your workspace has many big projects (which... may contain... yaml files):

```
2023-03-27 19:46:16 [DEBUG]: schema service: handle yaml mapping: sam -> file:///projects/sam-installation/aws-sam-cli-src/tests/smoke/templates/sar/s3-uncompressor-template.yaml
2023-03-27 19:46:16 [DEBUG]: schema service: handle yaml mapping: sam -> file:///projects/sam-installation/aws-sam-cli-src/tests/smoke/templates/sar/alexa-skills-kit-color-expert-template.yaml
2023-03-27 19:46:16 [DEBUG]: schema service: handle yaml mapping: sam -> file:///projects/sam-installation/aws-sam-cli-src/tests/smoke/templates/sar/SecretsManagerRedshiftRotationSingleUser-template.yaml
2023-03-27 19:46:17 [DEBUG]: schema service: handle yaml mapping: sam -> file:///projects/sam-installation/aws-sam-cli-src/tests/smoke/templates/sar/magic-8-ball-template.yaml
2023-03-27 19:46:17 [DEBUG]: schema service: handle yaml mapping: sam -> file:///projects/sam-installation/aws-sam-cli-src/tests/smoke/templates/sar/SecretsManagerMongoDBRotationSingleUser-template.yaml
2023-03-27 19:46:17 [DEBUG]: schema service: handle yaml mapping: sam -> file:///projects/sam-installation/aws-sam-cli-src/tests/smoke/templates/sar/AmazonConnectSalesforceLambda-template.yaml
<100s more...>
```

[...]

PR #3440, opened 2023-05-10
perf: do not watch files in node_modules

justinmk3:
Problem:
When setting up file watchers, Toolkit includes unwanted directories such as "node_modules". The default vscode setting `search.exclude` excludes "node_modules", but `vscode.workspace.findFiles()` uses the `files.exclude` as its default exclude pattern. #3216

Solution:
Specify an exclude pattern that matches the defaults from both `search.exclude` and `files.exclude`, plus some other things like ".aws-sam/". Do this for all `WatchedFiles` subclasses because there is no case where we want to watch files in any of these folders.

BEFORE:

    2023-05-10 04:24:34 [VERBOSE]: cfn: template registry setup took 257ms

AFTER:

    2023-05-10 04:23:22 [VERBOSE]: cfn: template registry setup took 53ms

## License


src/shared/fs/templateRegistry.ts
```diff
@@ -120,11 +121,11 @@ export class AsyncCloudFormationTemplateRegistry {
                         // Allows for new message to be created if templateRegistry variable attempted to be used again
                         this.setupProgressMessage = undefined
                     })
```
justinmk3, line 123: @nkomonen-amazon Possible bug: seems like Cancel does not actually cancel the work being done by https://github.com/aws/aws-toolkit-vscode/blob/a0a9280060e17ac8983cc4a6be83c28656011955/src/shared/cloudformation/activation.ts#L68-L69

PR #3446, opened 2023-05-10
fix(codecatalyst): skip DevEnv connect if quota exceeded

justinmk3:
Problem:
When customer's Space has reached quota, trying to connect to a DevEnv waits a long time instead of failing immediately.

Solution:
Cancel the wait loop if `ServiceQuotaExceededException` is detected.

## License


JadenSimon (approved):

src/shared/clients/codecatalystClient.ts
```diff
@@ -700,9 +700,16 @@ class CodeCatalystClientInternal {
                         await this.startDevEnvironment(args)
                     } catch (e) {
                         const err = e as AWS.AWSError
+                        // - ServiceQuotaExceededException: account billing limit reached
                         // - ValidationException: "… creation has failed, cannot start"
                         // - ConflictException: "Cannot start … because update process is still going on"
                         //   (can happen after "Update Dev Environment")
+                        if (err.code === 'ServiceQuotaExceededException') {
+                            throw new ToolkitError('Dev Environment failed: quota exceeded', {
+                                code: 'ServiceQuotaExceeded',
```
JadenSimon, line 717: You can omit `code`, it'll report the code of `cause` instead.

justinmk3: wondering if we should be explicit here, so that the metrics have the same code, even cross-toolkit? otoh, if other toolkits don't do this then we could end up with _less_ comparable/consistent metrics.

src/shared/clients/codecatalystClient.ts
```diff
@@ -680,17 +681,25 @@ class CodeCatalystClientInternal {
...
 
                 if (
-                    lastStatus &&
+                    startAttempts > 2 &&
+                    elapsed > 10000 &&
                     ['STOPPED', 'FAILED'].includes(lastStatus.status) &&
-                    ['STOPPED', 'FAILED'].includes(resp.status) &&
-                    elapsed > 60000 &&
-                    startAttempts > 2
+                    ['STOPPED', 'FAILED'].includes(resp.status)
```
justinmk3, line 691: Only wait 10 seconds instead of 1 minute before deciding that a stopped/failed devenv is not going to recover.

src/shared/clients/codecatalystClient.ts
```diff
@@ -729,23 +745,17 @@ class CodeCatalystClientInternal {
...
+            if (isUserCancelledError(e)) {
                 doLog('info', 'devenv failed to start (user cancelled)')
-                err.message = failedStartMsg()
-                throw err
+                e.message = failedStartMsg()
+                throw e
+            } else if (e instanceof ToolkitError) {
+                doLog('error', 'devenv failed to start')
+                throw e
             }
```
justinmk3, line 755: pass through all `ToolkitError`s instead of making decisions in this `waitTimeout()` catch.

PR #3448, opened 2023-05-11
authWebview: Can connect with credentials

nkomonen-amazon:
Ability to enter credentials from auth webview
- Validation of inputs
  - No duplicate profile name
  - Checks credentials can be authenticated before submission
- Connects to credentials upon submission

### Updates
- The secret key input shows `******` instead of the text

### Happy Path
Kapture 2023-05-11 at 10 57 54

### Validation Examples
Kapture 2023-05-11 at 10 53 36

## License


src/credentials/auth.ts
```diff
@@ -586,6 +587,16 @@ export class Auth implements AuthService, ConnectionManager {
...
 
+    public addTempCredentials(data: StaticProfile): CredentialsId {
+        const tempProvider = new TempCredentialsProvider(data)
+        this.iamProfileProvider.addProvider(tempProvider)
+        return tempProvider.getCredentialsId()
+    }
+
+    public removeTempCredentials(id: CredentialsId) {
+        this.iamProfileProvider.removeProvider(id)
+    }
```
JadenSimon, line 598: Exposing methods to temporarily change state in order to do other operations is a sign that these methods shouldn't be public. It would be better to consolidate them into a single method. Callers shouldn't need to worry about "temporary" state.

src/credentials/vue/root.vue
```diff
@@ -2,6 +2,34 @@
...
...,1,.67-.16H26.4a1.1,1.1,0,0,1,.67.16,1.17,1.17,0,0,1,.32.65L30,12.39,32.88,1.25A1.39,1.39,0,0,1,33.22.6a1,1,0,0,1,.65-.16h1.54a.36.36,0,0,1,.41.41,1.36,1.36,0,0,1,0,.26,3.64,3.64,0,0,1-.12.41l-4,12.86a1.3,1.3,0,0,1-.35.65,1,1,0,0,1-.64.16H29.25a1,1,0,0,1-.67-.17,1.26,1.26,0,0,1-.32-.67L25.67,3.64,23.11,14.34a1.26,1.26,0,0,1-.32.67,1,1,0,0,1-.67.17Zm21.36.44a11.28,11.28,0,0,1-2.56-.29,7.44,7.44,0,0,1-1.92-.67,1,1,0,0,1-.61-.93v-.84q0-.52.38-.52a.9.9,0,0,1,.31.06l.42.17a8.77,8.77,0,0,0,1.83.58,9.78,9.78,0,0,0,2,.2,4.48,4.48,0,0,0,2.43-.55,1.76,1.76,0,0,0,.86-1.57,1.61,1.61,0,0,0-.45-1.16A4.29,4.29,0,0,0,43,9.22l-2.41-.76A5.15,5.15,0,0,1,38,6.78a3.94,3.94,0,0,1-.83-2.41,3.7,3.7,0,0,1,.45-1.85,4.47,4.47,0,0,1,1.19-1.37A5.27,5.27,0,0,1,40.51.29,7.4,7.4,0,0,1,42.6,0a8.87,8.87,0,0,1,1.12.07q.57.07,1.08.19t.95.26a4.27,4.27,0,0,1,.7.29,1.59,1.59,0,0,1,.49.41.94.94,0,0,1,.15.55v.79q0,.52-.38.52a1.76,1.76,0,0,1-.64-.2,7.74,7.74,0,0,0-3.2-.64,4.37,4.37,0,0,0-2.21.47,1.6,1.6,0,0,0-.79,1.48,1.58,1.58,0,0,0,.49,1.18,4.94,4.94,0,0,0,1.83.92L44.55,7a5.08,5.08,0,0,1,2.57,1.6A3.76,3.76,0,0,1,47.9,11a4.21,4.21,0,0,1-.44,1.93,4.4,4.4,0,0,1-1.21,1.47,5.43,5.43,0,0,1-1.85.93A8.25,8.25,0,0,1,42.05,15.62Z"
```
nkomonen-amazon, line 16: This is the placeholder AWS logo, I will look at alternatives later

justinmk3: can this live in a file somewhere so we don't have to paste it into the views ?

src/credentials/auth.ts
nkomonen-amazon: Good point, I updated this commit and made a single public method in the Auth class that uses the above methods which are now private.

src/credentials/vue/root.vue
nkomonen-amazon: Yup this will be the most likely case. I just haven't looked in to how we add the file in to a webview.

src/credentials/auth.ts
```diff
@@ -586,6 +587,48 @@ export class Auth implements AuthService, ConnectionManager {
         return this.#validationErrors.get(connection.id)
     }
 
+    /**
+     * Returns true if credential data can be authenticated successfully
+     */
+    public async canAuthenticateRawData(data: StaticProfile): Promise<boolean> {
```
justinmk3, line 593: "raw data" as opposed to what?

src/credentials/auth.ts
```diff
@@ -586,6 +587,48 @@ export class Auth implements AuthService, ConnectionManager {
...
+        const tempId = await this.addTempCredential(data)
+        const tempIdString = asString(tempId)
+        try {
+            await this.reauthenticate({ id: tempIdString })
+        } catch (e) {
+            if (e instanceof Error && e.name === 'InvalidClientTokenId') {
```
justinmk3, line 599: is this checking a ToolkitError or is the `name` coming from the service? is there a `instanceof` comparison we can use instead?

src/credentials/providers/tempCredentialsProvider.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * A credentials provider for a temporary use case.
+ *
+ * This provides the bare minimum way to use credentials
+ * in the rest of the system.
+ *
+ * @see CredentialsProviderType
+ */
+export class TempCredentialProvider implements CredentialsProvider {
```
justinmk3, line 20: This is a lot of code for what looks like a hack(?). If this is not intended for use anywhere except the `addTempCredential` hack, the docstring here should make that very clear. Perhaps better would be to define this subclass in the same module as `addTempCredential` so that it's not exported.

src/credentials/providers/credentials.ts
```diff
@@ -67,11 +67,12 @@ export function isEqual(idA: CredentialsId, idB: CredentialsId): boolean {
...
  *   valid SSO profiles may not necessarily be considered valid among all tools.
- *
+ * - "temp" refers to credentials that are used temporarily within the code.
+ *   This can be for something like testing raw credentials data
  * Compare the similar concept `telemetry.CredentialSourceId`.
  */
-export type CredentialsProviderType = typeof credentialsProviderType[number]
-export const credentialsProviderType = ['profile', 'ec2', 'ecs', 'env', 'sso'] as const
+export type CredentialsProviderType = (typeof credentialsProviderType)[number]
+export const credentialsProviderType = ['profile', 'ec2', 'ecs', 'env', 'sso', 'temp'] as const
```
justinmk3, line 75: Does this `temp` value reach telemetry? I would guess it shouldn't, and in that case the providertype shouldn't matter.

src/credentials/vue/authForms/CredentialsForm.vue
```diff
@@ -0,0 +1,280 @@
+<template>
```
justinmk3, line 1: A name like `manageCredentials.vue` (or whatever is more appropriate) would express the purpose of this form more clearly than `CredentialsForm.vue`

justinmk3 (changes requested):
All of the filenames with `TitleCase` should be renamed to `camelCase` to match the convention of almost all(?) files in the codebase.

src/credentials/vue/root.vue
JadenSimon: I think if you use relative file paths they'll get bundled into the final build.

src/credentials/auth.ts
nkomonen-amazon: With the other available providers, the 'raw data' is encapsulated in something like env vars or the credentials file and the specific provider understands how to retrieve the data from them.

Let me know if you have a better name in mind

src/credentials/auth.ts
nkomonen-amazon: This is an error from the sdk, but I cannot find an error object for `InvalidClientTokenId` in the sdk types.

I've updated to use `isAwsError()` instead and then check the `code` instead of `name`

src/credentials/providers/credentials.ts
nkomonen-amazon: By the looks of it, it gets reported as 'other' in telemetry in `loginManager.ts#103`

nkomonen-amazon:
@justinmk3 I've updated the .vue file names to be camelCase and modified the form names

justinmk3 (approved):

Issue #3381, opened 2023-04-26
SAM: failed to run: "python3.10 runtime is not supported"


Rybalochka-MS:
## Problem
Hi there, I trying to locally run a Lambda function for Python 3.10. Previously I update all requirements: AWS SAM CLI, AWS CLI, CloudFormation template. But I got a `Error: 'python3.10' runtime is not supported` error.

Full output log:

> 2023-04-26 10:38:53 [ERROR]: log level: info
> 2023-04-26 10:38:53 [INFO]: endpoints: retrieving AWS endpoints data
> 2023-04-26 10:38:53 [INFO]: OS: Darwin arm64 22.2.0
> 2023-04-26 10:38:53 [INFO]: Visual Studio Code extension host:  1.77.3
> 2023-04-26 10:38:53 [INFO]: AWS Toolkit:  1.70.0
> 2023-04-26 10:38:53 [INFO]: node: 16.14.2
> 2023-04-26 10:38:53 [INFO]: electron: 19.1.11
> 2023-04-26 10:40:38 [INFO]: SAM CLI location: [object Object]
> 2023-04-26 10:40:38 [INFO]: SAM CLI location: [object Object]
> 2023-04-26 10:40:39 [INFO]: SAM CLI location: [object Object]
> 2023-04-26 10:40:39 [INFO]: Preparing to debug locally: Lambda "app.lambda_handler"
> 2023-04-26 10:40:39 [INFO]: Building SAM application...
> 2023-04-26 10:40:39 [INFO]: SAM CLI location: [object Object]
> 2023-04-26 10:40:39 [INFO]: Command: (not started) [/usr/local/bin/sam build --build-dir /tmp/aws-toolkit-vscode/vsctkrNAWTQ/output --template /tmp/aws-toolkit-vscode/vsctkrNAWTQ/app___vsctk___template.yaml --base-dir /root/functions/check_bpy_file --manifest /tmp/aws-toolkit-vscode/vsctkrNAWTQ/debug-requirements.txt]
> 2023-04-26 10:40:39 [INFO]: SAM CLI not configured, using SAM found at: '/usr/local/bin/sam'
> 2023-04-26 10:40:41 [ERROR]: Unexpected exitcode (1), expecting (0)
> 2023-04-26 10:40:41 [ERROR]: Error: undefined
[...]

PR #3452, opened 2023-05-12
fix(sam): "Detect SAM CLI" uses cached sam location

justinmk3:
Problem:
- Cached sam location is not evicted by the "AWS: Detect SAM CLI" command.
- Toolkit logs the sam cli location but not its version, which makes troubleshooting less obvious. #3381

Solution:
- Specify `forceSearch` when "AWS: Detect SAM CLI" calls `getOrDetectSamCli()`.
- Log the sam version.

## License

### DIFF
--- a/src/shared/sam/cli/samCliInfo.ts
+++ b/src/shared/sam/cli/samCliInfo.ts
@@ -5,11 +5,7 @@
 
 import { getLogger, Logger } from '../../logger'
 import { ChildProcess } from '../../utilities/childProcess'
-import { logAndThrowIfUnexpectedExitCode } from './samCliInvokerUtils'
 
-/**
- * Maps out the response text from the sam cli command `sam --info`
- */
 export interface SamCliInfoResponse {
     version: string
 }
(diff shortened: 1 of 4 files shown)


src/shared/sam/cli/samCliInfo.ts
```diff
@@ -18,13 +14,16 @@ export class SamCliInfoInvocation {
...
 
     public async execute(): Promise<SamCliInfoResponse> {
-        const childProcessResult = await new ChildProcess(this.samPath, ['--info'], { logging: 'no' }).run()
+        const r = await new ChildProcess(this.samPath, ['--info'], { logging: 'no' }).run()
 
-        logAndThrowIfUnexpectedExitCode(childProcessResult, 0)
-        const response = this.convertOutput(childProcessResult.stdout)
+        if (r.exitCode !== 0) {
+            // getVersionValidatorResult() will return `SamCliVersionValidation.VersionNotParseable`.
+            return { version: '' }
```
justinmk3, line 21: No need to throw here.

src/shared/sam/cli/samCliLocator.ts
```diff
@@ -29,9 +29,9 @@ export class DefaultSamCliLocationProvider implements SamCliLocationProvider {
...
      * `sam` wasn't previously found and cached.
      */
-    public async getLocation() {
+    public async getLocation(forceSearch?: boolean) {
         const perflog = new PerfLog('samCliLocator: getLocation')
-        const cachedLoc = DefaultSamCliLocationProvider.cachedSamLocation
+        const cachedLoc = forceSearch ? undefined : DefaultSamCliLocationProvider.cachedSamLocation
 
         // Avoid searching the system for `sam` (especially slow on Windows).
         if (cachedLoc && (await DefaultSamCliLocationProvider.isValidSamLocation(cachedLoc.path))) {
```
justinmk3, line 37: Probably will remove `isValidSamLocation` in a future PR. In the worst case (somehow the cached `sam` breaks), the user must either restart vscode or run "Detect SAM CLI".

JadenSimon (approved):

src/shared/sam/cli/samCliInfo.ts
JadenSimon: Isn't it better to fail early though? If we aren't going to throw then I think it makes more sense to return `undefined` instead of an empty value because it forces callers to deal with the difference.

justinmk3: yes, in this case the caller is `getVersionValidatorResult()` and it signals "undefined" as `VersionNotParseable`

PR #3453, opened 2023-05-12
codecatalyst: "Open Dev Environment": filter by space, project

justinmk3:
## Problem:
"Open Dev Environment" menu:
- shows env id prominently which is not human-friendly
- user cannot filter by space or project name
- "Last used" string takes unnecessary space
- places "RUNNING" environments at the bottom of the list.

## Solution:
- Set `matchOnDetail=true` so the list can be filtered by the detail description, not just the label.
- Drop "Last used" label; the timestamp is clear enough on its own.
- When sorting the list, only compare "last used".

### Before

listdevenvs-before

### After

Screenshot 2023-05-12 at 06 44 07

## License


nkomonen-amazon (approved):
Nice change, much more intuitive

JadenSimon (approved):

PR #3460, opened 2023-05-15
fix(ci): exclude testE2E/ from coverage analysis

justinmk3:
- fix(ci): exclude `testE2E/**` and `testLint/**` from coverage analysis
- rename `integrationTest/` => `testInteg/`

## License


.nycrc.json
```diff
@@ -1,5 +1,5 @@
 {
     "all": true,
     "include": ["src/**", "dist/src/**"],
-    "exclude": ["test/**", "integrationTest/**", "testFixtures/**"]
+    "exclude": ["test/**", "testE2E/**", "testLint/**", "testInteg/**", "testFixtures/**"]
```
justinmk3, line 4: should these be prefixed with `src/` ?
```suggestion
    "exclude": ["src/test/**", "src/testE2E/**", "src/testLint/**", "src/testInteg/**", "src/testFixtures/**"]
```

nkomonen-amazon (approved):

.nycrc.json
JadenSimon: I think the patterns are applied relative to `include` but not 100% sure

justinmk3: that seems to be what the examples in https://github.com/istanbuljs/nyc#selecting-files-for-coverage imply, but it doesn't clearly say.

PR #3471, opened 2023-05-16
fix(codewhisperer): Re-enable the suggestion navigation after 1.78 VS Code inline completion API breaking changes

leigaol:
## Problem

1. Suggestion navigation is no longer working when using VS Code 1.78+
2. Users need to hit Esc twice to clear/reject a code suggestion.

Root cause:

The dispose method of a inlineCompletionProvider used to clear the inline Completion widget, which allows us to update the widget whenever we want to by doing a dispose and doing another register with a new inlineCompletionProvider. However, after VS Code 1.78+, disposing the provider no longer clears the active inline completion widget, which is a regression because disposing a object should clear the states it created.

## Solution

Manually refresh the suggestion ghost text after each dispose of `inlineCompletionProvider` only for VS Code 1.78+.

## License


src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -18,13 +18,15 @@ import { getLogger } from '../../shared/logger/logger'
...
-import { getPrefixSuffixOverlap } from '../util/commonUtil'
+import { getPrefixSuffixOverlap, isVscHavingRegressionInlineCompletionApi } from '../util/commonUtil'
 import globals from '../../shared/extensionGlobals'
 import { AuthUtil } from '../util/authUtil'
 import { shared } from '../../shared/utilities/functionUtils'
 import { ImportAdderProvider } from './importAdderProvider'
+import { Mutex } from 'async-mutex'
 
 const performance = globalThis.performance ?? require('perf_hooks').performance
+const mutex = new Mutex()
```
JadenSimon, line 29: Use `async-lock` instead. Adding a new dependency is unnecessary.

leigaol: ACK. I will update it in the next commit.

leigaol: I have switched to async-lock instead.

JadenSimon (approved):
Needs changelog entry right?

leigaol:
> Needs changelog entry right?

Yes, let me add one.

zixlin7 (approved):

JadenSimon (approved):

.changes/next-release/Bug Fix-65050370-fd7c-49e6-86ba-00766572400c.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "CodeWhisperer: Fix a bug when user cannot navigate to next or previous code suggestion in VS Code 1.78+"
```
justinmk3, line 3:
```suggestion
	"description": "CodeWhisperer: cannot navigate to next or previous code suggestion in VS Code 1.78+"
```

justinmk3 (approved):

PR #3482, opened 2023-05-18
fix(codewhisperer): Remove deprecated code path

leigaol:
## Problem

Customers's plugin occasionally runs into the deprecated code path in CodeWhisperer, causing bugs and unexpected errors. 

## Solution

This PR is to remove the deprecated code path: 
1. The custom made `editor.edit` API based inline completion.
2. The old, deprecated cognital identity authentication.   

## License


zixlin7 (approved):

justinmk3:
Should this have a changelog item mentioning the fix for `requested operation is not recognized by the service`

leigaol:
> Should this have a changelog item mentioning the fix for `requested operation is not recognized by the service`

I added a change log in the latest commit

PR #3485, opened 2023-05-19
fix(stepfunctions): offline sfn visualizer

### LINKED ISSUE #2922: Despite instructions to run toolkit in c9 without internet access, visualize step function graph requires internet connectivity that is easily patched.

yaythomas:
Step Functions Render Graph visualizer will now work offline, as long as there is a cached local copy of the generating js and css files.

Closes #2922.

There is now a  graceful fallback when internet connectivity isn't available to a c9 or vscode host but where the required files are available in the local filesystem cache (either by previous automatic caching, or by manual user intervention).

Implementation avoids confusing the return type of 'updateCache' or 'updateCachedFile' and instead only checks for local files after a updateCache call fails.

This commit merges 3 previous work in progress commits made for PR #3098. The current change submitted in this PR incorporates all the review comments from the previous PR.

Once this PR merges to main you can close #3098.

With great thanks for the original work by @simis2626!

## License

### DIFF
--- a/.changes/next-release/Bug Fix-5dd44ab8-9083-4dd8-ac28-585d99895aca.json
+++ b/.changes/next-release/Bug Fix-5dd44ab8-9083-4dd8-ac28-585d99895aca.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Add fallback to local cached files for Step Function Visualizations where host doesn't have internet."
+}
--- a/src/stepFunctions/commands/visualizeStateMachine/abstractAslVisualizationManager.ts
+++ b/src/stepFunctions/commands/visualizeStateMachine/abstractAslVisualizationManager.ts
@@ -56,4 +56,14 @@ export abstract class AbstractAslVisualizationManager<T extends AslVisualization
     protected getExistingVisualization(key: string): T | undefined {
         return this.managedVisualizations.get(key)
     }
+
+    protected async updateCache(globalStorage: vscode.Memento, logger: Logger): Promise<void> {
+        try {
+            await this.cache.updateCache(globalStorage)
+        } catch (err) {
+            // So we can't update the cache, but can we use an existing on disk version.
+            logger.warn('Updating State Machine Graph Visualisation assets failed, checking for fallback local cache.')
+            await this.cache.confirmCacheExists()
+        }
+    }
 }
--- a/src/stepFunctions/commands/visualizeStateMachine/aslVisualizationManager.ts
+++ b/src/stepFunctions/commands/visualizeStateMachine/aslVisualizationManager.ts
@@ -33,8 +33,7 @@ export class AslVisualizationManager extends AbstractAslVisualizationManager {
 
         // Existing visualization does not exist, construct new visualization
         try {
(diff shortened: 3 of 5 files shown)


JadenSimon (approved):
Looks good! I don't see any major issues.

PR #3486, opened 2023-05-19
fix(codewhisperer): Remove extra space at the beginning of ghost text

leigaol:
## Problem

When user types with SPACE at certain speed, both the service and user will add extra SPACE after the CodeWhisperer invocation position, resulting in multiple extra space at the start of the suggestion ghost text, which breaks the indentation of the current file.  

https://github.com/aws/aws-toolkit-vscode/assets/97199248/29467bc7-8192-44f6-ad64-40abc8c6142b

The root cause is a dated mechanism that moves the invocation start position if the user typeahead is while-space characters.

## Solution

Remove the dated start position moving mechanism. Make sure the space is included in the typeahead computation.

## License


.changes/next-release/Bug Fix-de534112-c087-411d-93e0-e38a7d76e10b.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "CodeWhisperer: Fix a bug when CodeWhisperer sometimes add extra space at the beginning of suggestion"
```
justinmk3, line 3: "bug fix" is already the item prefix so it is never necessary to repeat that in the description
```suggestion
	"description": "CodeWhisperer sometimes add extra space at the beginning of suggestion"
```

justinmk3 (approved):

zixlin7 (approved):

Issue #3490, opened 2023-05-22
CodeWhisperer errors when Not Enabled


mousedownmike:
## Steps to reproduce the issue

1. Install AWS Toolkit alongside GitHub Copilot
2. Leave Code Whisperer disabled (Should show "Start" button in the extension configuration).
3. Write some code and observe suggestions appear (assumed to be from CoPilot)
4. Hit Tab to accept the suggestions.
5. Observe: 
  * Suggested code is not inserted.
  * Code Whisperer error message appears: `command 'aws.codeWhisperer.rejectCodeSuggestion' not found`

## Expected behavior

Code Whisperer should NOT be running.
CoPilot suggestion should be inserted.

## System details (run the `AWS: About Toolkit` command)

OS: Linux x64 5.19.0-41-generic
Visual Studio Code extension host:  1.78.2
AWS Toolkit:  1.74.0
node: 16.17.1
electron: 22.5.2

peterabbott:
+1 seeing this too, had to disable the entire plugin. Was getting in the way.

bombillazo:
Please fix this! This is tampering with our productivity as the cursor gets stuck in place whenever this error prompts up! We use the AWS toolkit to connect to our cloud account, we don't have anything regarding Code Whisperer set up. 😫

laygir:
This is very annoying, keeps breaking my productivity for the whole last week. Suddenly getting stuck in hitting esc key and unable to escape from this error 🤯

rosmcmahon:
yes, very annoying, disabling the entire plugin seems to be the only course of action.

hauboldj:
I was able to fix this by removing the keybinding for `aws.codeWhisperer.rejectCodeSuggestion` command. I don't have a screenshot unfortunately. After removing this keybinding I don't see the command anymore. I guess that the AWS Toolkit has removed this command, but didn't remove the keybinding and VS Code evaluates if the command exists before evaluating the When clause for keybinding.

bombillazo:
@hauboldj  thanks! Here are all the bindings, I guess Ill be removing them all:
![image](https://github.com/aws/aws-toolkit-vscode/assets/44002321/90dc421d-20f3-404b-8164-9199cd151ca7)

samuelbutler:
I was able to just remove the `aws.codeWhisperer.rejectCodeSuggestion` keybinding, and after doing that the escape key works properly and actually rejects the code suggestion. 

[...]

Issue #3509, opened 2023-05-30
"command aws.amazonq.rejectCodeSuggestion not found" if connection expired


DustyLamp:
## Problem
`command 'aws.codeWhisperer.rejectCodeSuggestion' not found` keeps popping up when I press escape when a suggestion is made and I want to dismiss it.

Note: I'm using Copilot and Code Whisperer at the same time.

## System details (run the `AWS: About Toolkit` command)
I'm using copilot and aws code whisperer at the same time.

-   OS: mac
-   Visual Studio Code version: 1.78.2
-   AWS Toolkit version: 1.74.0

disbelief:
This happens for me even when I've explicitly disabled code whisperer in the AWS extension config

dreamorosi:
Hi @justinmk3 I'm using AWS Toolkit 1.76 like you suggested in the previous comment but I'm still seeing the error, even though I have disabled code whisperer in the workspace:

<img width="463" alt="image" src="https://github.com/aws/aws-toolkit-vscode/assets/7353869/a3e8020c-15b8-45e9-9954-b8a754c85015">


<img width="506" alt="Screenshot 2023-06-10 at 12 17 12" src="https://github.com/aws/aws-toolkit-vscode/assets/7353869/a8e6a34a-588c-4e92-92be-4d5e5da215fe">

Only thing left to do is disable the extension entirely.

Will-ShaoHua:
hi @dreamorosi, by "I have disable codewhisperer in the workspace", do you mean you have signed out from either Aws-builderId or Aws-Identity-center and also curious at that point of time, what's the UI you can see on CodeWhisperer node in the toolkit panel? Thanks!

Will-ShaoHua:
hi @disbelief, could you elaborate more on the step you mentioned ` disabled code whisperer in the AWS extension config`, what config did you change then? Thanks!

disbelief:
@Will-ShaoHua I actually ended up just disabling the entire AWS extension for my workspace. I thought it was possible to disable code whisperer in the extension settings, but you can only turn off some of its features:

<img width="622" alt="Screenshot 2023-06-11 at 09 03 56" src="https://github.com/aws/aws-toolkit-vscode/assets/295574/ba4b1eed-d753-4c51-b081-f492ef375066">

However after updating to the latest version of the AWS extension (and restarting VS Code) I'm not having this problem anymore.

Will-ShaoHua:
[...]

PR #3518, opened 2023-06-02
fix codewhisperer command not found error when codewhisperer is disabled

Will-ShaoHua:
## Problem
fix https://github.com/aws/aws-toolkit-vscode/issues/3490 
## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -3157,7 +3157,7 @@
                 "command": "aws.codeWhisperer",
                 "key": "alt+c",
                 "mac": "alt+c",
-                "when": "editorTextFocus"
+                "when": "editorTextFocus && CODEWHISPERER_ENABLED"
             },
             {
                 "command": "aws.codeWhisperer.rejectCodeSuggestion",
--- a/.changes/next-release/Bug Fix-185bc9a6-c99a-4c9a-b230-2dd7c90e6162.json
+++ b/.changes/next-release/Bug Fix-185bc9a6-c99a-4c9a-b230-2dd7c90e6162.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Fix CodeWhisperer not found error when it is disabled by users issue #3490"
+}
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -3157,7 +3157,7 @@
                 "command": "aws.codeWhisperer",
                 "key": "alt+c",
                 "mac": "alt+c",
-                "when": "editorTextFocus"
+                "when": "editorTextFocus && CODEWHISPERER_ENABLED"
```
Will-ShaoHua, line 3160: Unrelevant to the bug. But it will fix: when codewhisperer in disabled, users are able to press option + c and invoke CW service getting recommendation, the recommendations will not be rendered tho.

zixlin7 (approved):

justinmk3:
Needs a changelog item :)

package.json
andrewyuq: when codewhisperer is disabled, it shouldn't even get suggestions correct?

Will-ShaoHua: yes correct

Will-ShaoHua:
> Needs a changelog item :)

ack

Will-ShaoHua:
btw @justinmk3 have a q: what's the best practice using Toolkit's `Commands` APIs (for example `Cmmands.declare(...)`) for service like CodeWhisperer (users have to enable it and sometimes it's disable), for the time being, we kinda declare all the commands regardless if users enable or disable our service.

justinmk3:
> for the time being, we kinda declare all the commands regardless if users enable or disable our service.

`Commands.declare` is safe to use always. In this case, `alt+c` triggered the command declared in package.json, not related to `Commands.declare`.

nkomonen-amazon (approved):

dreamorosi:
I still see the warning on v 1.76 of the extension, see [my comment](https://github.com/aws/aws-toolkit-vscode/issues/3509#issuecomment-1585597374) in this issue for screenshots.

PR #3525, opened 2023-06-05
refactor: auth code

nkomonen-amazon:
## Problem
Currently, our `auth.ts` module is 1500+ lines.
Most of the code in this module can be separated in to other modules that have narrower concerns.

Additionally, with separate PRs I was running in to circular dependencies with the Auth client due to the way this module is currently structured.

## How to review PR
I've created many commits, each with a specific refactor. See each commit message for more info on each.

### CodeWhisperer Team
The only thing that changed related to your code was the imports. Verifying the changes in `src/codewhisperer` will be all you need to do.

## Solution

Refactor the code to organized modules.

### Overview
- The overall folder named is now `src/auth/` instead of `src/credentials/`
- The `auth.ts` module now focuses only on the `Auth` client and does not have other related code in the same module
- When dealing with a `Connection` object like an Sso, BuilderId, or Credentials/IAM connection, related code was moved from `auth.ts` to `connection.ts`. This helps to avoid circular dependencies when working with connections + auth.
- The functional code previously in `auth.ts` was moved to `src/auth/utils.ts` to create separation between the functions and the Auth client.
- Existing credentials related files were moved in to `src/auth/credentials/` to consolidate them
- A new folder was made in `src/auth/ui/` that will hold all auth related UI code, including the status bar item and add connection webview.
  
#### Old
Screen Shot 2023-06-05 at 2 39 48 PM

#### New
Screen Shot 2023-06-07 at 12 49 34 PM

## License


src/auth/auth.ts
```diff
@@ -0,0 +1,758 @@
...
+    StatefulConnection,
+    StoredProfile,
+    codecatalystScopes,
+    createBuilderIdProfile,
+    hasScopes,
+    isBuilderIdConnection,
+    loadIamProfilesIntoStore,
+    loadLinkedProfilesIntoStore,
+    ssoAccountAccessScopes,
+} from './connection'
```
JadenSimon, line 55: nit: could make sense to use a namespace import e.g. `import * as connection from './connection`. Not sure if the Typescript language server supports this as a refactoring option.

JadenSimon (approved):

zixlin7 (approved):

src/auth/auth.ts
nkomonen-amazon: Makes sense, I'll keep this in mind since I'll probably be doing some other auth refactors in the future.

src/auth/commands.ts
```diff
@@ -1,11 +1,10 @@
 /*!
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
justinmk3, line 2: Dates are not required. If you find the need to update these, just remove the date.

PR #3534, opened 2023-06-07
ec2: add prompt for instance id to command palette.

Hweinstock:
## Problem
Initial piece of the project: expose command on command palette to allow user to select an EC2 instance. 

## Solution
Notes/Details:
- we log the instanceId selected to the console. 
- the region is automatically chosen based on `regionProvider.guessDefaultRegion`. Escape hatch to come in future PR. 
- tests are provided for basic cases, more to come for instance prompter when region prompter is integrated. 
- reliant on the refactor of test utils (https://github.com/aws/aws-toolkit-vscode/pull/3535)

image

image

## License


package.json
```diff
@@ -2020,6 +2020,16 @@
                     }
                 }
             },
+            {
```
nkomonen-amazon, line 2023: I had a previous conversation with Justin and it might be a good idea to contribute directly to `master` to avoid the `feature/*` branch merging process completely since this feature does not need to be hidden.

For the command palette command I used `"when": "aws.isDevMode"` for my command `aws.auth.showConnectionsPage` in package.json

src/ec2/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+import { AsyncCollection } from "../shared/utilities/asyncCollection"
+
+export function extractInstanceIdsFromReservations(reservations: AsyncCollection<EC2.ReservationList | undefined>): AsyncCollection<string> {
+    return reservations
+        .flatten()
+        .map(instanceList => instanceList?.Instances)
+        .flatten()
+        .map(instance => instance?.InstanceId)
+        .filter(instanceId => instanceId !== undefined)
+}
```
justinmk3, line 16: Make sure `files.insertFinalNewline` is enabled in your vscode settings. If so, maybe `git` is removing the final newline?

package.json
Hweinstock: The "when" tells the toolkit what conditions must be met to show the command right? But what does the `"aws.isDevMode"` condition mean?

src/test/utilities/collectionUtils.ts
```diff
@@ -10,6 +10,12 @@ export async function* asyncGenerator<T>(items: T[]): AsyncIterableIterator<T> {
     yield* items
 }
 
+export function intoCollection<T>(arr: T[]): AsyncCollection<T> {
+    return toCollection(async function* () {
+        yield* arr
+    })
+}
+
 export function createCollectionFromPages<T>(...pages: T[]): AsyncCollection<T> {
```
justinmk3, line 19: what is test-specific about these utils? why don't they live in `src/shared/utilities/collectionUtils.ts` ?

src/ec2/utils.ts
Hweinstock: Just enabled it, guess its not default?

src/test/utilities/collectionUtils.ts
Hweinstock: I think all the methods in here are only currently used in tests. My understanding of these methods is that they allow us to create a `AsyncCollection` from non-async objects.  Would there ever be a use case for this outside of testing?

justinmk3: Not sure, but if there is a use-case outside of testing they would be less likely to be discovered. When choosing to create "test libraries" it should be obvious that the library stuff is only relevant to tests. If it's not obvious, why not put it in the common shared library?

package.json
JadenSimon: Add `"aws.dev.forceDevMode": true` to your VS Code settings.

src/test/utilities/collectionUtils.ts
Hweinstock: I setup a separate PR for the refactor which could be merged into here first, before merging it all together. Let me know if I understood what your saying correctly. https://github.com/aws/aws-toolkit-vscode/pull/3535

src/ec2/commands.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+import { EC2 } from 'aws-sdk'
+import globals from '../shared/extensionGlobals'
+import { pageableToCollection } from '../shared/utilities/collectionUtils'
+import { CancellationError } from '../shared/utilities/timeoutUtils'
+import { extractInstanceIdsFromReservations } from "./utils"
+import { selectInstance } from './prompter'
+
+export async function attemptConnection(defaultRegion: string): Promise<void> {
```
justinmk3, line 13:
```suggestion
export async function tryConnection(defaultRegion: string): Promise<void> {
```

src/ec2/utils.ts
```diff
@@ -0,0 +1,16 @@
+/*!
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
justinmk3, line 2:
```suggestion
 * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

justinmk3 (approved):

justinmk3 (approved):

justinmk3:
CI failing

PR #3535, opened 2023-06-07
Refactor: test CollectionUtils -> general collectionUtils. 

Hweinstock:
## Problem
The difference between the collection utils and the general collection utils is blurry and not well defined. While some utility functions are only used in testing, this is not obvious. Follow up from https://github.com/aws/aws-toolkit-vscode/pull/3534#discussion_r1222241097. 
## Solution
Move all of the utility functions regarding collections into one place, so that they don't get lost. 

## License


src/shared/utilities/collectionUtils.ts
```diff
@@ -488,3 +488,25 @@ export async function* joinAll<T>(iterable: AsyncIterable<AsyncIterable<T>>): As
         }
     } while (!iterables.completed)
 }
+
+// TODO: what is the point of this? should it live in src/shared/utilities/collectionUtils.ts ?
```
JadenSimon, line 492:
```suggestion

```

JadenSimon (approved):

JadenSimon:
Target `master`?

Hweinstock:
This branch also includes the changes for the EC2 prompter. I'll remove the changes from here regarding the EC2 prompter, then can target master with this PR.

Hweinstock:
Think I removed any remnants from ec2. However, first time using `rebase` so lmk if I messed something up.

src/shared/utilities/collectionUtils.ts
```diff
@@ -488,3 +488,24 @@ export async function* joinAll<T>(iterable: AsyncIterable<AsyncIterable<T>>): As
         }
     } while (!iterables.completed)
 }
+
+export async function* asyncGenerator<T>(items: T[]): AsyncIterableIterator<T> {
```
justinmk3, line 492: looks like the `test/utilities/collectionUtils.ts` file still exists?

PR #3538, opened 2023-06-08
authWebview: Various small changes

nkomonen-amazon:
## Problem
Various fixes related to the Auth webview

## Solution
Each commit makes a minor specific change

## License


src/auth/ui/vue/backend/types.ts
```diff
@@ -0,0 +1,6 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export type ServiceItemId = 'DOCUMENT_TYPE_SUPPORT' | 'RESOURCE_EXPLORER' | 'CODE_WHISPERER' | 'CODE_CATALYST'
```
JadenSimon, line 6: You should be able to combine this w/ the frontend types. It's ok to use the same file as long as the code is runtime agnostic (node vs. browser). Types are always safe to share.

JadenSimon (approved):

src/auth/ui/vue/backend/types.ts
nkomonen-amazon: Thanks, moved out of the "backend" folder in to the root

PR #3547, opened 2023-06-08
ec2: add escape hatch option on prompt

Hweinstock:
## Problem
The current EC2 prompt does not provide an ability to customize region selection. 
## Solution
Leverage the `regionSubmenu`  to accomplish a defaulted region prompt with an escape hatch to change if needed. 

image

image

## License


justinmk3:
> * necessary files for `regionSubmenu` were copy-pasted from the `feature/cwl` branch. 

can revert since `feature/cwl` was merged

> The current EC2 prompt does not provide an ability to customize region selection.

Can we somehow make this "switch region" quickpick menu re-usable? Maybe a base class like `ServiceResourcePrompter` which would be the common prompter used for the first step of most service actions.

src/ec2/commands.ts
```diff
@@ -3,23 +3,15 @@
...
 
-            const selection = await selectInstance(collection)
-            if(!selection){
-                throw new CancellationError('user')
-            } else {
-                console.log(selection)
-            }
+    if(isValidResponse(response)){
+        const selection = handleEc2ConnectPrompterResponse(response)
+        console.log(selection)
```
JadenSimon, line 15: I've seen people logging the selection pretty often. Would it be helpful if responses (quickpick items only, not text input) were logged automatically at a debug level?

src/ec2/prompter.ts
```diff
@@ -3,28 +3,36 @@
...
+import { getInstanceIdsFromRegion } from "./utils";
+import { DataQuickPickItem } from "../shared/ui/pickerPrompter";
 
-function asQuickpickItem(stringItem: string): DataQuickPickItem<string>[] 
-{
-    return [{
-        label: stringItem, 
-        data: stringItem
-    }]
+type EC2InstanceId = string
```
JadenSimon, line 10: I would avoid aliasing primitives because it gives a false sense of robustness. Let identifiers (properties, parameters, etc.) describe the expected shape instead.

src/ec2/prompter.ts
```diff
@@ -3,28 +3,36 @@
...
+
+type EC2Selection = {
```
JadenSimon, line 12: nit: use `interface` for object-like shapes

It's _mostly_ a style thing.

JadenSimon (commented):
You might need to install `husky` to get auto-formatting to work. You can use `npx pretty-quick` to do the formatting too.

src/ec2/commands.ts
Hweinstock: I would think so. The main reason I am logging here is since the next-step isn't implemented. I imagine this is a decently common use case. Is there any downside other than slight clutter increase to logging it at the debug level?

src/ec2/prompter.ts
Hweinstock: So should `type` be reserved for enum-like shapes. Ex. `type currency = 'USD' | 'EUR'`. Are there other cases where `type` is preferred?

JadenSimon: > So should type be reserved for enum-like shapes. 

I pretty much only use `type` when I can't use `interface`. Which is actually pretty frequent with "advanced" types.

src/ec2/commands.ts
JadenSimon: Noise is the primary reason for not doing it. But it shouldn't be too noisy, at least compared to something like SDK API calls.

src/ec2/utils.ts
```diff
@@ -4,13 +4,27 @@
  */
 
 import { EC2 } from 'aws-sdk'
-import { AsyncCollection } from "../shared/utilities/asyncCollection"
+import { AsyncCollection } from '../shared/utilities/asyncCollection'
+import globals from '../shared/extensionGlobals'
+import { pageableToCollection } from '../shared/utilities/collectionUtils'
 
-export function extractInstanceIdsFromReservations(reservations: AsyncCollection<EC2.ReservationList | undefined>): AsyncCollection<string> {
+export function extractInstanceIdsFromReservations(
```
justinmk3, line 11: this could live in the "client". Do we have that for ec2 yet? That's the usual pattern we have done thus far. (We have some improvements planned, but until then it probably makes sense to continue the pattern. For ec2 that would mean creating a client in `src/shared/clients/ec2Client.ts`

src/ec2/utils.ts
```diff
@@ -4,13 +4,27 @@
...
     return reservations
         .flatten()
         .map(instanceList => instanceList?.Instances)
         .flatten()
         .map(instance => instance?.InstanceId)
         .filter(instanceId => instanceId !== undefined)
-} 
+}
+
+export async function getInstanceIdsFromRegion(regionCode: string): Promise<AsyncCollection<string>> {
```
justinmk3, line 22: could live in the client, `src/shared/clients/ec2Client.ts`

justinmk3 (approved):

src/shared/clients/ec2Client.ts
```diff
@@ -0,0 +1,37 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { EC2 } from 'aws-sdk'
```
justinmk3, line 6: Should probably use v3 sdk instead.

Hweinstock: Will make this its own PR to make sure I don't break anything before merge.

PR #3558, opened 2023-06-09
fix(cwl): "Copy Log Stream Name" fails on search results editor

Hweinstock:
## Problem
The option to copy the log stream name appears on both log group results and single log stream results. When used with a log group open, the command returns an error.  
## Solution
Change the "copy log stream" option to "Copy Log Stream or Group" so that it fits to both cases. 

image

Also, update the tests to isolate each desired outcome/behavior. 

## License


justinmk3:
I think when viewing a log stream we still want the "Copy Log Stream" feature. 

If it's a lot of code to add "Copy Log Group", can we just _remove_ "Copy Log Stream" for the search results case?

Hweinstock:
Yeah, I can look into it. So the desired behavior is:
- when viewing a log stream we get copy log stream. 
- when viewing a log group we get copy log group. 

I think this can be done by creating a dynamic `copyLogResource` command that checks the open document's Uri to determine if its logStream or logGroup.

justinmk3:
Could name it `Copy Log Stream or Group`. Some vscode commands have that precedent, e.g. `Run and Debug`

package.json
```diff
@@ -844,7 +844,7 @@
                     "when": "false"
                 },
                 {
-                    "command": "aws.copyLogStreamName",
+                    "command": "aws.copyLogResource",
```
justinmk3, line 847: nit: let's keep "name" in the command name since that is clearer and matches many existing similar commands
```suggestion
                    "command": "aws.copyLogName",
```

src/cloudWatchLogs/commands/copyLogResource.ts
```diff
@@ -7,31 +7,37 @@ import * as nls from 'vscode-nls'
...
-        }
+        } 
         const parsedUri = parseCloudWatchLogsUri(uri)
-        const streamName = parsedUri.logGroupInfo.streamName
+        let resourceName
+        if(isLogStreamUri(uri)){
+            resourceName = parsedUri.logGroupInfo.streamName
+        } else {
+            resourceName = parsedUri.logGroupInfo.groupName
+        }
```
justinmk3, line 30: ternary is clearer, more concise, and allows use of `const`

```suggestion
        const resourceName = isLogStreamUri(uri) ? parsedUri.logGroupInfo.streamName : parsedUri.logGroupInfo.groupName
```

src/cloudWatchLogs/commands/copyLogResource.ts
```diff
@@ -7,31 +7,37 @@ import * as nls from 'vscode-nls'
...
 
-        if (!streamName) {
-            throw new Error(`Unable to copy stream name for Uri that doesn\'t have stream. Attempted copy on ${uri}`)
+        if (!resourceName) {
+            throw new Error(`Unable to copy log resource name for Uri that doesn\'t have resource. Attempted copy on ${uri}`)
```
justinmk3, line 33: while we're here let's make this message less verbose, it's almost hard to understand

```suggestion
            throw new Error(`Failed to copy log resource name for URI (missing stream or group?): ${uri}`)
```

src/cloudWatchLogs/commands/copyLogResource.ts
```diff
@@ -7,31 +7,37 @@ import * as nls from 'vscode-nls'
...
         }
-        await copyToClipboard(streamName)
+        await copyToClipboard(resourceName)
     } catch (e) {
         vscode.window.showErrorMessage(
             localize(
                 'AWS.cwl.invalidEditor',
-                'Not a Cloudwatch Log stream: {0}',
+                'Not a Cloudwatch Log group: {0}',
```
justinmk3, line 40:
```suggestion
                'Invalid Cloudwatch Log stream or group: {0}',
```

src/cloudWatchLogs/commands/copyLogResource.ts
```diff
@@ -7,31 +7,37 @@ import * as nls from 'vscode-nls'
...
+export async function copyLogResource(uri?: vscode.Uri): Promise<void> {
     try {
+
         if (!uri) {
             // No URI = used command palette as entrypoint, attempt to get URI from active editor
             // should work correctly under any normal circumstances since the action only appears in command palette when the editor is a CloudWatch Logs editor
             uri = vscode.window.activeTextEditor?.document.uri
             if (!uri) {
-                throw new Error("Attempt to copy Uri that doesn't exist.")
+                throw new Error("Attempt to copy resource name for Uri that doesn't exist.")
```
justinmk3, line 21:
```suggestion
                throw new Error("no active text editor, or undefined URI")
```

justinmk3 (approved):

justinmk3 (approved):

PR #3574, opened 2023-06-13
feat: Update the 'forceDevMode' setting

nkomonen-amazon:
### Problem:

There was not a way to explicitly enable and disable dev mode with a single setting

### Solution:

Update the logic so that aws.dev.forceDevMode takes precedence over all other settings and will honor
whatever that value is in relation to if the user
is in dev mode.

Now, a user can have all their other dev settings
in place, but when they set 'aws.dev.forceDevMode: false' they will not be in dev mode anymore.

Fix: IDE-10948

## License

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -238,7 +238,9 @@ generating SDKs, etc.
 The [DevSettngs](https://github.com/aws/aws-toolkit-vscode/blob/479b9d45b5f5ad30fc10567e649b59801053aeba/src/shared/settings.ts#L553) class defines various developer-only settings that change the behavior of the
 Toolkit for testing and development purposes. To use a setting just add it to
 your `settings.json`. At runtime, if the Toolkit reads any of these settings,
-the "AWS" statusbar item will [change its color](https://github.com/aws/aws-toolkit-vscode/blob/479b9d45b5f5ad30fc10567e649b59801053aeba/src/credentials/awsCredentialsStatusBarItem.ts#L45). Use the setting `aws.dev.forceDevMode` to trigger this effect on start-up.
+the "AWS" statusbar item will [change its color](https://github.com/aws/aws-toolkit-vscode/blob/479b9d45b5f5ad30fc10567e649b59801053aeba/src/credentials/awsCredentialsStatusBarItem.ts#L45).
+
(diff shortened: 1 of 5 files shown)


src/shared/settings.ts
```diff
@@ -632,6 +632,20 @@ export class DevSettings extends Settings.define('aws.dev', devSettings) {
         return this.trappedSettings
     }
 
+    public isDevMode(): boolean {
+        // This setting takes precedence over everything.
+        // It must be removed completely from the settings to not be considered.
+        const forceDevMode: boolean | undefined = !this.isSet('forceDevMode')
+            ? undefined
+            : this.get('forceDevMode', false)
```
JadenSimon, line 640: Minor style nit: prefer placing `undefined` in the "whenFalse" expression:
```suggestion
        const forceDevMode: boolean | undefined = this.isSet('forceDevMode')
            ? this.get('forceDevMode', false)
            : undefined
```

JadenSimon (approved):

CONTRIBUTING.md
```diff
@@ -238,7 +238,9 @@ generating SDKs, etc.
 The [DevSettngs](https://github.com/aws/aws-toolkit-vscode/blob/479b9d45b5f5ad30fc10567e649b59801053aeba/src/shared/settings.ts#L553) class defines various developer-only settings that change the behavior of the
 Toolkit for testing and development purposes. To use a setting just add it to
 your `settings.json`. At runtime, if the Toolkit reads any of these settings,
-the "AWS" statusbar item will [change its color](https://github.com/aws/aws-toolkit-vscode/blob/479b9d45b5f5ad30fc10567e649b59801053aeba/src/credentials/awsCredentialsStatusBarItem.ts#L45). Use the setting `aws.dev.forceDevMode` to trigger this effect on start-up.
+the "AWS" statusbar item will [change its color](https://github.com/aws/aws-toolkit-vscode/blob/479b9d45b5f5ad30fc10567e649b59801053aeba/src/credentials/awsCredentialsStatusBarItem.ts#L45).
+
+The setting `aws.dev.forceDevMode` will take precedence over all dev settings and enable dev mode on `"aws.dev.forceDevMode": true` or disable on `"aws.dev.forceDevMode": false`.
```
justinmk3, line 243: Should probably rename it to `devMode`, that will be less confusing.

nkomonen-amazon: Makes sense. I will make another PR for this

PR #3575, opened 2023-06-13
ec2: connect to terminal w/o error handling. 

Hweinstock:
## Problem
The selection menu currently logs the response, but doesn't open a terminal or use the selection information. 
## Solution
Leverage the SSM SDK to establish a connection to the desired instanceId. 

Notes:
- SSM SDK error response appears to be the same in many situations that lead to failed connections. For example, when instance is currently booting up/down, instance isn't running, or IAM role not properly configured. Throws following general error:
image

- This means handling the error will involve leveraging other pieces of the SDK to pinpoint source of the problem. 
- For this PR, the toolkit throws a general error that is copy-paste of console error when taking same steps. 
image

## License


src/ec2/model.ts
```diff
@@ -0,0 +1,69 @@
...
+import * as vscode from 'vscode'
+import { Session } from 'aws-sdk/clients/ssm'
+import { Ec2Selection } from './utils'
+import { getOrInstallCli } from '../shared/utilities/cliUtils'
+import { isCloud9 } from '../shared/extensionUtilities'
+import { withoutShellIntegration } from '../ecs/commands'
+import { ToolkitError } from '../shared/errors'
+import { DefaultSsmClient } from '../shared/clients/ssmClient'
+
+export class Ec2ConnectClient {
```
justinmk3, line 14: will this be moved to the `.../clients/` folder? If it's meant to be a client, consider naming it `Ec2Client` since it's likely that we'll put all ec2-related client-layer stuff there.

If this is meant to hold higher-layer business logic, could name it simply `ec2` or `ec2Model`.

src/ec2/model.ts
```diff
@@ -0,0 +1,69 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import * as vscode from 'vscode'
+import { Session } from 'aws-sdk/clients/ssm'
+import { Ec2Selection } from './utils'
+import { getOrInstallCli } from '../shared/utilities/cliUtils'
+import { isCloud9 } from '../shared/extensionUtilities'
+import { withoutShellIntegration } from '../ecs/commands'
```
justinmk3, line 10: Followup: does it make sense to extract most of `ecs/util.ts` and `ecs/commands.ts` into something like `shared/remoteSession.ts` ... most of it looks re-usable. Just need to parameterize `client` and telemetry in a few places.

src/ec2/model.ts
Hweinstock: Its meant to hold the logic specific for connecting to remote instance which combines working with the SSM client and EC2 client. I wasn't sure if this qualified as business logic or if it should itself be a higher-level client built on others.

src/ec2/model.ts
Hweinstock: That's a good point. I was using that code as a reference. The main difference is how we start and work with a session. However, splitting this up into smaller units should yield some shared components.

src/ec2/model.ts
justinmk3: That is business logic for EC2. The common parts of "remote connect" could be extracted though (see other comments)

src/shared/clients/ssmClient.ts
```diff
@@ -0,0 +1,39 @@
...
+            .promise()
+            .catch(err => {
+                getLogger().warn(`ssm: failed to terminate session "${sessionId}": %s`, err)
+            })
+        return termination
+    }
+
+    public async startSession(
+        target: string,
+        callback?: ((err: AWSError, data: SSM.StartSessionResponse) => void) | undefined
```
JadenSimon, line 34: Prefer `async/await` over callbacks.

src/shared/remoteSession.ts
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+import { Settings } from '../shared/settings'
+
+export async function openRemoteTerminal(
+    options: vscode.TerminalOptions,
+    onClose: () => void,
+    onError: (err: unknown) => void
```
JadenSimon, line 12: Let the error bubble up and use `.catch`?

src/shared/clients/ssmClient.ts
Hweinstock: I was having trouble detecting if it was an error or not, but just changed to `async/await` and checking `response.$response.error` which seems to work. Is that a reliable way to check if the response is an AWSError or is there easier/better way?

src/shared/remoteSession.ts
Hweinstock: Is the main advantage of using `.catch` here to reduce the # of parameters, and therefore making the function easier to understand and simpler?

JadenSimon: Both. It's simpler and now your error handler can be async if needed.

src/shared/clients/ssmClient.ts
JadenSimon: Use `isAwsError` from `src/shared/errors.ts` in a catch block. If there's an error it should throw, are you using `.promise()`? Sorry forgot to mention that.

Hweinstock: Yeah, that's exactly what I ended up doing. I saw a few different examples of handling SDK errors in the code base so wasn't sure which to follow, but this seemed to be most prominent.

src/shared/clients/ssmClient.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { AWSError, SSM } from 'aws-sdk'
+import globals from '../extensionGlobals'
+import { PromiseResult } from 'aws-sdk/lib/request'
+import { getLogger } from '../logger/logger'
+
+export class DefaultSsmClient {
```
justinmk3, line 11: the DefaultXX naming pattern is left over from old code where interfaces were heavily used. If there is no `SsmClient` interface then we don't need "Default"
```suggestion
export class SsmClient {
```

justinmk3 (approved):
`Ec2ConnectClient` is still in model.ts , is that intentional?

Hweinstock:
Yea, thats the piece that leverages the SSM and EC2 clients to make the connection. It's also where we will eventually handle the errors. I suspect the name client is confusing. I could drop that piece and go with `Ec2SsmConnecter`? Any other ideas you think more clearly illustrate what its doing?

PR #3579, opened 2023-06-14
ui: regionSubmenu tests don't need to be cwl specific. 

Hweinstock:
## Problem
The tests currently use language specific to cwl, implying the component is only used for that purpose. However, nothing in the design of the UI component is specific to cwl and can be used outside that context.
## Solution
Change the language to be more general, and avoid cwl terms like `group` or `logData`. 

## License


JadenSimon (approved):

justinmk3 (approved):
Usually useful to put the problem/solution in the commit message too

PR #3581, opened 2023-06-14
fix: vscode command gets wrong arg

nkomonen-amazon:
## Problem

When a vscode command is called it is possible for vscode to pass in a completely separate type of object as the argument, instead of the expected auth object. This would cause it to fail.

How to recreate:

- select a node in the explorer
- then indirectly call this command using `Connect to AWS` from the explorer kebab context menu

it will then pass in that node as context instead of the expected auth.

## Solution

In this case we do not need auth as an argument and can instead define it inside the function.

## License

### DIFF
--- a/src/auth/utils.ts
+++ b/src/auth/utils.ts
@@ -468,7 +468,8 @@ export const useIamCredentials = Commands.register('_aws.auth.useIamCredentials'
 })
 
 // Legacy commands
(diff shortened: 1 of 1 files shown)


justinmk3 (approved):
the pr title / commit subject are very generic, but it seems like this is specific to the "aws.login" command

PR #3583, opened 2023-06-15
(feat codewhisperer) refactor BM25 algorithm impl, prioritize opening files when generating supplemental context

Will-ShaoHua:
* fix cross file not prioritize opened files in editors
* trimEnd() on every code chunk of cross file
* fix BM25 algorithm counting logic throwing NaN

## License


src/codewhisperer/util/supplementalContext/rankBm25.ts
```diff
@@ -5,46 +5,126 @@
...
+        const sorted = notSorted.sort((a, b) => b.score - a.score)
+        return sorted.slice(0, Math.min(n, sorted.length))
+    }
 }
 
-const BM25Configs = {
-    epsilon: 0.25,
-    k1: 1.5,
-    b: 0.75,
+export class BM25Okapi extends BM25 {
```
JadenSimon, line 82: Why the abstract base? Will there be many concrete implementations? There's _a lot_ of protected fields too, seems like the class only exists for initialization.

Will-ShaoHua: yea, it's aligning with JB implementation and make it abstract because there are different variants of bm25 exist, I want it extendable so that if we decide to use different variants later it would be easier to just adding another subclass with its own calIdf.

Will-ShaoHua: like there variants https://github.com/dorianbrown/rank_bm25/blob/master/rank_bm25.py

Will-ShaoHua: is "many protected fields" an anti-pattern Jaden? I thought portected is fine since only subclass has access to it.

src/codewhisperer/util/supplementalContext/rankBm25.ts
```diff
@@ -5,46 +5,126 @@
...
+    ) {
+        this.corpus = corpus
+        this.corpusSize = corpus.length
+        this.docLen = []
+        this.idf = new Map()
+        this.docFreqs = []
+        this.nd = new Map()
+        this.k1 = k1
+        this.b = b
+        this.epsilon = epsilon
```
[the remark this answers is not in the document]
Will-ShaoHua: gotcha, will update

src/codewhisperer/util/supplementalContext/rankBm25.ts
Will-ShaoHua: basically what I did on JB and VSC change now is to do an exact same implementation as the original python library.

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -158,21 +166,38 @@ async function getRelevantCrossFiles(editor: vscode.TextEditor, dependencyGraph:
...
+        } else {
+            body.push(file)
+        }
+    })
 
-    return mergedCrossFileList
+    return [...head, ...body]
+}
+
+function openFilesInWindow(): string[] {
```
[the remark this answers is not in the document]
Will-ShaoHua: agree, have't checked the behavior on C9, will double check this. Is it make sense to do a `try` and doesn't do this prioritization if it goes in `catch` block?

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -158,21 +166,38 @@ async function getRelevantCrossFiles(editor: vscode.TextEditor, dependencyGraph:
...
-        .filter(file => srcDependencies.includes(file.fsPath))
-        .map(file => file.fsPath)
+    return mergedCrossFileList
+}
 
-    const mergedCrossFileList = [
-        ...new Set([...filteredRelevantOpenFiles, ...srcDependencies, ...samePackageRelevantFiles]),
-    ]
+// Util to move selected files to the front of the input array if it exists
+function moveToFront(files: string[], picked: Set<string>) {
```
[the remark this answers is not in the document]
Will-ShaoHua: ack

src/codewhisperer/util/editorContext.ts
```diff
@@ -109,12 +109,21 @@ export async function buildListRecommendationRequest(
...
 
+    const supplementalContext: codewhispererClient.SupplementalContext[] = supplementalContexts
+        ? supplementalContexts.contents.map(v => {
+              return {
+                  filePath: v.filePath,
+                  content: v.content,
+              }
+          })
+        : []
+
```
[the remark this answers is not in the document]
Will-ShaoHua: oh because here I need to return a different pojo which is `codewhispererClient.SupplementalContext`
```
{
  content: string,
  filePath: string
} 

```
where as my plugin Pojo has 3 fields (1 extra)
```
{
   content: string,
   filePath: string,
  score: number
}
```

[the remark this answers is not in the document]
Will-ShaoHua: is there an easier cleaner syntax to achieve this purpose?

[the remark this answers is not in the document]
JadenSimon: They're just called "objects" :)

You can use `selectFrom` in `src/shared/utilities/tsUtils.ts`:
```ts
const supplementalContext = supplementalContexts?.map(x => selectFrom(x, 'content', 'filePath') ?? []
```

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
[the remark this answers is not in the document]
JadenSimon: Yeah `try/catch` would work.

src/codewhisperer/util/supplementalContext/rankBm25.ts
JadenSimon: > I thought portected is fine since only subclass has access to it.

I view having lots of `protected` state an anti-pattern because it becomes harder to reason about the entire state of an object when multiple things can change it. Protected immutable state doesn't have this issue.

src/codewhisperer/util/editorContext.ts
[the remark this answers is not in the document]
Will-ShaoHua: gotchaa

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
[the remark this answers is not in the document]
Will-ShaoHua: ack

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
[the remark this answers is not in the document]
Will-ShaoHua: @JadenSimon I just realized the reason I didn't use this approach is that I want to keep the natural order if both a, b are in the picked, they will not be in the same order as before sorting.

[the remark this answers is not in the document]
Will-ShaoHua: for example 
```
arr = [1, 9, 4, 5, 8, 2]
picked = [9, 5, 4]

actual = [5, 4, 9, 1, 8, 2]

expected = [9, 4, 5, 1, 8, 2]
```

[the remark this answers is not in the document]
JadenSimon: Hmm maybe this then:
```ts
function moveToFront<T>(arr: T[], picked: Set<T>) {
    return [...arr].sort((a, b) => (picked.has(b) ? 1 : 0) - (picked.has(a) ? 1 : 0))
}
```

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -158,21 +166,41 @@ async function getRelevantCrossFiles(editor: vscode.TextEditor, dependencyGraph:
...
+    const filesOpenedInEditor: string[] = []
+
+    try {
+        const tabArrays = vscode.window.tabGroups.all
+        tabArrays.forEach(tabArray => {
+            tabArray.tabs.forEach(tab => {
+                filesOpenedInEditor.push((tab.input as any).uri.path)
+            })
+        })
+    } catch (e) {}
```
JadenSimon, line 201:
```suggestion
    } catch (e) {
        // Older versions of VSC do not have the tab API
    }
```

JadenSimon (approved):

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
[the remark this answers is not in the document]
Will-ShaoHua: gotcha let me try this, I have experimented several combination and didn't work out lol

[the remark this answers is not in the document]
Will-ShaoHua: yea, should be this one

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -38,59 +36,66 @@ export async function fetchSupplementalContextForSrc(
...
-function findBestKChunkMatches(chunkInput: Chunk, chunkReferences: Chunk[], k: number) {
+function findBestKChunkMatches(chunkInput: Chunk, chunkReferences: Chunk[], k: number): Chunk[] {
     const chunkContentList = chunkReferences.map(chunk => chunk.content)
     //performBM25Scoring returns the output in a sorted order (descending of scores)
-    const output: BMDocument[] = performBM25Scoring(chunkContentList, chunkInput.content) as BMDocument[]
-    const bestChunks: Chunk[] = []
-    //pick Top 3
-    for (let i = 0; i < Math.min(k, output.length); i++) {
-        const chunkIndex = output[i].index
+    // const output: BMDocument[] = performBM25Scoring(chunkContentList, chunkInput.content) as BMDocument[]
```
leigaol, line 77: You can remove this unused code

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -4,30 +4,28 @@
...
     fileName: string
     content: string
     nextContent: string
+    score?: number
 }
-const chunkSize = 10
-const chunkCount = 60
-const topK = 3
 
 export async function fetchSupplementalContextForSrc(
```
leigaol, line 24: Do you have unit tests for `fetchSupplementalContextForSrc`?

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -4,30 +4,28 @@
...
     editor: vscode.TextEditor,
     dependencyGraph: DependencyGraph,
     cancellationToken: vscode.CancellationToken
-) {
+): Promise<CodeWhispererSupplementalContextItem[] | undefined> {
```
leigaol, line 28: You are declaring this function as an async function, but inside this function there is no await. Is this intentional?

src/shared/vscode/tabGroup.ts
```diff
@@ -0,0 +1,305 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
```
leigaol, line 3: Can you leave a comment of the source?

src/codewhisperer/util/supplementalContext/utgUtils.ts
```diff
@@ -73,20 +73,19 @@ export async function fetchSupplementalContextForTest(
 function generateSupplementalContextFromFocalFile(
     filePath: string,
     cancellationToken: vscode.CancellationToken
-): codewhispererClient.SupplementalContext[] {
-    const supplementalContexts: codewhispererClient.SupplementalContext[] = []
+): CodeWhispererSupplementalContextItem[] {
     const fileContent = fs.readFileSync(vscode.Uri.file(filePath!).fsPath, 'utf-8')
     const segmentSize = 10200
```
leigaol, line 78: make it a constant

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
Will-ShaoHua: updated

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
Will-ShaoHua: not for now

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
Will-ShaoHua: Nope, there is an await at L34  https://github.com/aws/aws-toolkit-jetbrains/blob/main/jetbrains-core/tst/software/aws/toolkits/jetbrains/services/codewhisperer/CodeWhispererFileContextProviderTest.kt

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
Will-ShaoHua: In fact, I think it's even better to refactor the overall structure of the cross file context / utg feature in vscode, I would prefer more abstraction instead of exporting functions everywhere.

Will-ShaoHua: I will take it as a follow up once I have bandwidth to do it.

src/codewhisperer/util/supplementalContext/utgUtils.ts
Will-ShaoHua: ack

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -158,21 +166,31 @@ async function getRelevantCrossFiles(editor: vscode.TextEditor, dependencyGraph:
...
-    const mergedCrossFileList = [
-        ...new Set([...filteredRelevantOpenFiles, ...srcDependencies, ...samePackageRelevantFiles]),
-    ]
+// Util to move selected files to the front of the input array if it exists
+function moveToFront<T>(arr: T[], picked: Set<T>) {
+    return [...arr].sort((a, b) => (picked.has(b) ? 1 : 0) - (picked.has(a) ? 1 : 0))
+}
 
-    return mergedCrossFileList
+function openFilesInWindow(): string[] {
```
leigaol, line 179: Could you rename this to `getCurrentOpenFilesInWindow` or similar and write a doc string to explain what the return object is?

`openFilesInWindow` seems like a name that opens a file as a new tab. But what you are doing is to get the file path of all opened tabs.

Will-ShaoHua: sg

Will-ShaoHua:
Will not merged it right now, need to another round of sanity check before merging.

leigaol (approved):

Will-ShaoHua:
@JadenSimon @justinmk3 this one is safe to merge, I have done the check in the follow up PR https://github.com/aws/aws-toolkit-vscode/pull/3615

PR #3592, opened 2023-06-19
ui: add optional parameter for separator label in RegionSubmenu

Hweinstock:
## Problem
Regardless of the use case of the `RegionSubmenu` component, the separator label is fixed as `"Selections"`. Since this component is now used in EC2 and CWL, this is vague and not a very helpful label. 

## Solution
Default the label to `"Selections"`, but allow for a constructor parameter to add case-specific text. 
First example: CWL can use `"Log Groups"` instead of `"Selections"`:
image
Second Example: EC2 can use `"Instances"` instead of `"Selections"`:
image

## License

### DIFF
--- a/src/shared/ui/common/regionSubmenu.ts
+++ b/src/shared/ui/common/regionSubmenu.ts
@@ -27,6 +27,7 @@ export class RegionSubmenu<T> extends Prompter<RegionSubmenuResponse<T>> {
         private readonly itemsProvider: (region: string) => ItemLoadTypes<T>,
         private readonly dataOptions?: ExtendedQuickPickOptions<T>,
         private readonly regionOptions?: ExtendedQuickPickOptions<T>,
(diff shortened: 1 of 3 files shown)


src/shared/ui/common/regionSubmenu.ts
```diff
@@ -27,6 +27,7 @@ export class RegionSubmenu<T> extends Prompter<RegionSubmenuResponse<T>> {
         private readonly itemsProvider: (region: string) => ItemLoadTypes<T>,
         private readonly dataOptions?: ExtendedQuickPickOptions<T>,
         private readonly regionOptions?: ExtendedQuickPickOptions<T>,
+        private readonly separatorLabel: string = "Selections",
```
justinmk3, line 30: Maybe "Items" is more common?

```suggestion
        private readonly separatorLabel: string = "Items",
```

justinmk3 (approved):

PR #3596, opened 2023-06-20
ec2: improve QuickPick UI

Hweinstock:
## Problem
The UI for selecting an instance lists the instance ids, which customers may not know. 
## Solution
List the instanceIds as `detail` and set the primary label to the name of the instance if one exists. Otherwise, use instanceId. 

Old Version:
image

New Version: 
image

Technical Notes:
- In order to do this we parse the `Tags` attribute returned with the instance where the `Name` key appears to match the name assigned in console. 
- If an instance does not have a name (which appears possible after experimenting), the instance id is used in both places. 

## License


justinmk3:
> If an instance does not have a name (which appears possible after experimenting), the instance id is used in both places.

let's try reporting the no-name cases as `(no name)` (or similar), which has precedent in e.g. our CodeCatalyst menu.

src/shared/clients/ec2Client.ts
```diff
@@ -11,35 +11,48 @@ import {
     Reservation,
     DescribeInstanceStatusRequest,
     InstanceStateName,
+    Tag,
 } from '@aws-sdk/client-ec2'
 import { AsyncCollection } from '../utilities/asyncCollection'
 import { pageableToCollection } from '../utilities/collectionUtils'
 import { IamInstanceProfile } from 'aws-sdk/clients/ec2'
 
+export interface Ec2Instance {
```
justinmk3, line 20: would it make senes to `extends` the sdk types, to avoid having to re-map properties. Example from the codecatalyst client: https://github.com/aws/aws-toolkit-vscode/blob/38e13ea204d375ea9b1a4ee1e62075257760c432/src/shared/clients/codecatalystClient.ts#L83

src/shared/clients/ec2Client.ts
```diff
@@ -11,35 +11,48 @@ import {
...
+        const instances = this.extractInstancesFromReservations(collection)
+        return instances
+    }
+
+    private lookupTagKey(tags: Tag[], targetKey: string): string | undefined {
+        return tags.filter(tag => tag.Key == targetKey)[0].Value
     }
 
-    public extractInstanceIdsFromReservations(
+    public extractInstancesFromReservations(
```
justinmk3, line 43:
```suggestion
    public getInstancesFromReservations(
```

justinmk3 (approved):

src/shared/clients/ec2Client.ts
Hweinstock: Yeah, that makes sense. I changed it to extend the sdk type, but add the name field. The name is traditionally buried in the tags so this makes it easier to work with.

PR #3599, opened 2023-06-20
authWebview: minor changes

nkomonen-amazon:
This includes:
- Add/update url links
- During SSO (BuilderID and Identity Center) when something fails in the process we show an error message under the submit button indicating what happened.
- In the scenario where an Identity Center connection already exists and the user wants to use the same IC start url to connect to CodeWhisperer we previously wouldn't allow them since the start url already existed. Now, we will instead allow them to submit using that start url and it will take them through a flow which adds the new scopes to that sso profile.

## License

### DIFF
--- a/src/auth/ui/vue/show.ts
+++ b/src/auth/ui/vue/show.ts
@@ -32,7 +32,7 @@ import {
     isIamConnection,
     isSsoConnection,
 } from '../../connection'
-import { tryAddCredentials, signout, showRegionPrompter, addConnection, promptForConnection } from '../../utils'
+import { tryAddCredentials, signout, showRegionPrompter, addConnection, promptAndUseConnection } from '../../utils'
 import { Region } from '../../../shared/regions/endpoints'
 import { CancellationError } from '../../../shared/utilities/timeoutUtils'
 import { validateSsoUrl, validateSsoUrlFormat } from '../../sso/validation'
@@ -317,7 +317,7 @@ export class AuthWebview extends VueWebview {
     }
 
     showConnectionQuickPick() {
-        return promptForConnection(Auth.instance)
+        return promptAndUseConnection(Auth.instance)
     }
 }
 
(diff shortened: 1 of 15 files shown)


src/codewhisperer/util/authUtil.ts
```diff
@@ -127,7 +127,8 @@ export class AuthUtil {
         } else if (isValidCodeWhispererConnection(existingConn)) {
             return this.secondaryAuth.useNewConnection(existingConn)
         } else if (isSsoConnection(existingConn)) {
```
nkomonen-amazon, line 129: We added the scope to the connection but didn't actually start using it.

JadenSimon (approved):

src/auth/ui/vue/show.ts
```diff
@@ -188,16 +184,19 @@ export class AuthWebview extends VueWebview {
         return this.ssoSetup(setupFunc)
     }
 
-    private async ssoSetup(setupFunc: () => Promise<any>) {
+    private async ssoSetup(setupFunc: () => Promise<any>): Promise<string> {
         try {
             await setupFunc()
+            return ''
```
JadenSimon, line 190: Can't remember now if this was a problem with the webview impl. but was something weird happening when returning `void`/`undefined`?

nkomonen-amazon: In this case with it being a string this isn't an issue. It was in the scenario where an arg would be like:

```
{ a: 'A', b: undefined }
```

and we would only get

```
{ a: 'A' }
```

JadenSimon: Oh that's because everything is serialized between the extension and the webview. `undefined` has no JSON representation and so is equivalent to the absence of a field.

justinmk3 (approved):

andrewyuq (approved):

PR #3604, opened 2023-06-22
ec2: add progress bar on ssm startSession. 

Hweinstock:
## Problem 
Sometimes the `startSession` command can take ~5 seconds. This has only been observed when working with windows instances. The leads to poor UX. 
## Solution 
Provide a progress bar w/ cancel when loading the new terminal. 
image

Technical Notes: 
- This PR also adds a status bar for ECS `open terminal` functionality as they now use the same code. 

## License


src/shared/remoteSession.ts
```diff
@@ -18,6 +23,8 @@ export async function openRemoteTerminal(options: vscode.TerminalOptions, onClos
 
         terminal.show()
     })
+
+    timeout.cancel()
```
JadenSimon, line 27: It's good practice to dispose of things in a `finally` block/callback:
```suggestion
    }).finally(() => timeout.cancel())
```

JadenSimon (approved):

JadenSimon:
Do you plan on squashing this PR? There's a lot of redundant commits.

src/shared/remoteSession.ts
Hweinstock: Is the advantage of placing this into a finally block that it will run regardless of whether or not the main body throws an exception or not?

src/shared/remoteSession.ts
```diff
@@ -5,9 +5,14 @@
...
 import * as vscode from 'vscode'
 import { Settings } from '../shared/settings'
+import { showMessageWithCancel } from './utilities/messages'
+import { Timeout } from './utilities/timeoutUtils'
 
 export async function openRemoteTerminal(options: vscode.TerminalOptions, onClose: () => void) {
-    await withoutShellIntegration(() => {
+    const timeout = new Timeout(60000)
+
+    await showMessageWithCancel('AWS: Starting session...', timeout)
```
justinmk3, line 14: Could use the `showAfterMs` param to skip this message if the connection is fast enough.
```suggestion
    await showMessageWithCancel('AWS: Starting session...', timeout, 1000)
```

justinmk3 (approved):

src/shared/remoteSession.ts
JadenSimon: Yes. Anything that you want to happen regardless of the outcome should go in `finally`.

PR #3608, opened 2023-06-23
refactor: Remove additional prompt for CC Builder ID

nkomonen-amazon:
## Problem

When connection to CodeCatalyst Builder ID there is a prompt that asks if you would like to go to the browser but then immediately follows up with a prompt to copy a code then open the browser.

The initial prompt was created before the second one existed.

## Solution

Remove the initial prompt since it is not needed.

#### Before
BuilderIdBefore

#### After
BuilderIdCWAfter

## License

### DIFF
--- a/src/codecatalyst/auth.ts
+++ b/src/codecatalyst/auth.ts
@@ -5,7 +5,6 @@
 
 import * as vscode from 'vscode'
 import { CodeCatalystClient, createClient } from '../shared/clients/codecatalystClient'
-import { getIdeProperties } from '../shared/extensionUtilities'
 import { Auth } from '../auth/auth'
 import { getSecondaryAuth } from '../auth/secondaryAuth'
 import { getLogger } from '../shared/logger'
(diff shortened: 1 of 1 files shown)


src/codecatalyst/auth.ts
```diff
@@ -138,15 +137,6 @@ export class CodeCatalystAuthenticationProvider {
...
-            const message = `The ${
-                getIdeProperties().company
-            } Toolkit extension requires a connection for CodeCatalyst to begin.\n\n Proceed to the browser to allow access?`
-
-            const resp = await vscode.window.showInformationMessage(message, { modal: true }, continueItem, cancelItem)
-            if (resp !== continueItem) {
-                throw new ToolkitError('Not connected to CodeCatalyst', { code: 'NoConnection', cancelled: true })
-            }
-
             const newConn = await createBuilderIdConnection(this.auth, defaultScopes)
```
nkomonen-amazon, line 140: This call contains the 2nd prompt and throws if cancelled.

JadenSimon (approved):

src/codecatalyst/auth.ts
JadenSimon: Is `promptNotConnected` called anywhere where the first prompt adds context? If so, you could change those callsites to take users to the auth page.

justinmk3 (approved):

PR #3611, opened 2023-06-23
ec2: add telemetry to connect flow. 

Hweinstock:
## Problem
The current flow of connecting to an EC2 instance does not emit any telemetry. 
## Solution
- Add `ssm` to `EC2ConnectionType`.
- Emit `Ec2ConnectToInstance` on `ssm` connection.
- Emit `Ec2ConnectToInstance` on failed connections with relevant reasons.
- Emit `Ec2ConnectToInstance` on cancel with `result: 'cancelled'`. 

## License


src/ec2/model.ts
```diff
@@ -63,6 +64,7 @@ export class Ec2ConnectionManager {
         const hasProperPolicies = await this.hasProperPolicies(selection.instanceId)
 
         if (!isInstanceRunning) {
+            telemetry.record({ result: 'Failed', reason: 'EC2SSMStatus' })
```
JadenSimon, line 67: You shouldn't need to do this if the error is bubbled up.

justinmk3: same for the other `record({ result: 'Failed' })` cases here, right?

Hweinstock: Removed them, and retested. Found a missing `await` while I was at it.

JadenSimon (approved):

PR #3613, opened 2023-06-24
fix(ec2): check if instance connectable before connecting. 

Hweinstock:
## Problem
If the user changes the IAM role associated with the instance, the connect command fails whether it be adding or removing the necessary policies. 

The root of this problem is ultimately a bug in the usage of the `startSession` SDK. That is, there are ways this command can fail such that it doesn't throw an error, but the terminal will crash on entry. 
## Solution
We check for ways the `startSession` command can fail before running it. If any are detected, we throw a relevant error. This way, we only call this command when we are confident it will succeed. 

### Notes
- In the process of discovering this bug, debugging effort involved refactoring some of the `ec2Client` code. The changes here aim to make it easier to add more SDK related functionality in this class later on. 

## License


src/ec2/model.ts
```diff
@@ -107,13 +115,13 @@ export class Ec2ConnectionManager {
     public async attemptEc2Connection(selection: Ec2Selection): Promise<void> {
         try {
             const response = await this.ssmClient.startSession(selection.instanceId)
+            const isConnectable = await this.isInstanceConnectable(selection.instanceId)
+            if (!isConnectable) {
+                throw new Error('Instance is not connectable.')
+            }
```
JadenSimon, line 121: Would this be user-facing? Should we use `ToolkitError` to show the user a more helpful message?

src/ec2/model.ts
```diff
@@ -41,25 +41,33 @@ export class Ec2ConnectionManager {
...
     }
 
-    public async handleStartSessionError(err: AWSError, selection: Ec2Selection): Promise<Error> {
-        const isInstanceRunning = (await this.ec2Client.getInstanceStatus(selection.instanceId)) == 'running'
+    private async isInstanceConnectable(instanceId: string): Promise<boolean> {
+        const isInstanceRunning = (await this.ec2Client.getInstanceStatus(instanceId)) == 'running'
+        const hasProperPolicies = await this.hasProperPolicies(instanceId)
+
+        return isInstanceRunning && hasProperPolicies
+    }
```
JadenSimon, line 65: It would be helpful to return (or throw) _why_ the instance cannot be connected to.

src/shared/clients/ec2Client.ts
```diff
@@ -19,28 +19,28 @@ export class Ec2Client {
...
-    public async getInstances(): Promise<AsyncCollection<Ec2Instance>> {
+
+    public async getInstances(filters?: EC2.Filter[]): Promise<AsyncCollection<EC2.Instance>> {
         const client = await this.createSdkClient()
+
         const requester = async (request: EC2.DescribeInstancesRequest) => client.describeInstances(request).promise()
-        const collection = pageableToCollection(requester, {}, 'NextToken', 'Reservations')
+        const collection = filters
+            ? pageableToCollection(requester, { Filters: filters }, 'NextToken', 'Reservations')
+            : pageableToCollection(requester, {}, 'NextToken', 'Reservations')
```
justinmk3, line 29: syntax nit: can this be less verbose, something like (untested):

```suggestion
        const collection =pageableToCollection(requester, filters ? { Filters: filters } : {}, 'NextToken', 'Reservations')
```

src/shared/clients/ec2Client.ts
```diff
@@ -62,19 +62,48 @@ export class Ec2Client {
...
+     * @param instanceId target EC2 instance ID
+     * @returns Date object for launch time associated with instance, or undefined if instance doesn't exists or doesn't have one.
+     */
+    public async getInstanceLaunchTime(instanceId: string): Promise<Date | undefined> {
+        const singleInstanceFilter = this.getInstancesFilter([instanceId])
+        try {
+            const instance = (await (await this.getInstances(singleInstanceFilter)).promise())[0]
+            return instance.LaunchTime!
+        } catch (err: unknown) {
+            return undefined
```
justinmk3, line 85: silencing errors doesn't seem like the right pattern here. 

If an error occurs does getInstances() log it?

src/ec2/model.ts
Hweinstock: This isn't user facing, this error is immediately caught and then passed into the `handleStartSessionError`. I imagine there this is a better way to handle this. Will try to refactor to make this flow clearer.

src/ec2/model.ts
Hweinstock: Right now I have these as two separate methods, but it sounds like it makes more sense to combine them into one.

src/shared/clients/ec2Client.ts
Hweinstock: It doesn't. This and the `IAMAttachDate` functions are both a little half-baked. I used them to debug the source of the error, but didn't up using them in the actual solution. I think its better if I just delete them.

Hweinstock:
I just refactored how we check for errors in the `model.ts` file. The new version has a single function to pre-check the start session errors, rather than checking both before and after the call.

src/ec2/model.ts
```diff
@@ -105,15 +107,17 @@ export class Ec2ConnectionManager {
...
-                await this.handleStartSessionError(err, selection)
-            } else {
-                throw err
-            }
+            // Default error if pre-check fails.
+            this.throwConnectionError(
+                'Check that the SSM Agent is running on target instance',
+                'EC2SSMConnect',
+                selection
+            )
```
JadenSimon, line 120: We should add `err` as a `cause` for this new error (it can be added to `ToolkitError`). That way we don't swallow the original error which might contain helpful debugging info.

src/ec2/model.ts
```diff
@@ -41,53 +41,55 @@ export class Ec2ConnectionManager {
...
+        documentationUri?: vscode.Uri
+    ) {
         const generalErrorMessage = `Unable to connect to target instance ${selection.instanceId} on region ${selection.region}. `
+        throw new ToolkitError(
+            generalErrorMessage + message,
+            documentationUri ? { code: code, documentationUri: documentationUri } : { code: code }
+        )
+    }
+
+    public async checkForStartSessionError(selection: Ec2Selection): Promise<void> {
```
JadenSimon, line 77: Pre-checking works although a reactive approach may be more robust. Let's leave this as-is and come back to it later if we have time.

Hweinstock: I originally had this as a reactive approach, but it had some edge cases. There were ways we could get the `ssm.startSession` to not throw any errors, but the terminal immediately crashes. Thats why I moved it to checking the errors before. Should I be looking for errors both before and after if this is the case?

justinmk3 (approved):

PR #3617, opened 2023-06-26
ec2: explicitly check if SSM agent is running

Hweinstock:
## Problem
If the error is not related to permissions or instance status, we throw a general error about the SSM agent not running. However, this may not always be the case and we shouldn't assume other errors couldn't occur.

## Solution
We can leverage `DescribeInstanceInformation` from the SSM SDK to check the ping status of the SSM agent of target EC2 instance. If this call returns an error, or the ping status is outdated, we can conclude the SSM agent isn't running.
https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-ssm/classes/describeinstanceinformationcommand.html

Now, there is a specific error message for when an instance is not able to connect. Therefore, the default error message has been changed so that it does not mention the SSM agent (since we would have already checked that). 

## License


src/ec2/model.ts
```diff
@@ -90,11 +92,17 @@ export class Ec2ConnectionManager {
             )
         }
 
-        throw new ToolkitError('Is SSM running on the target instance?', {
+        if (!isSsmAgentRunning) {
+            throw new ToolkitError('Is SSM running on the target instance?', {
```
justinmk3, line 96:
```suggestion
            throw new ToolkitError('Is SSM Agent running on the target instance?', {
```

src/ec2/model.ts
```diff
@@ -90,11 +92,17 @@ export class Ec2ConnectionManager {
...
+                code: 'EC2SSMAgentStatus',
+                documentationUri: vscode.Uri.parse(
+                    'https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started.html'
```
justinmk3, line 99: since this message is about the agent would this be a more relevant url:

https://docs.aws.amazon.com/systems-manager/latest/userguide/ssm-agent.html

src/shared/clients/ssmClient.ts
```diff
@@ -28,4 +36,29 @@ export class SsmClient {
         const response = await client.startSession({ Target: target })
         return response
     }
+
+    public async describeInstanceInformation(target: string): Promise<InstanceInformation> {
```
justinmk3, line 40: it's good to avoid needless differences, but we can truncate verbose things when writing wrappers (which isn't quite the same as "changing for the sake of change"). "Information" adds nothing in most cases when it is on a function name.

```suggestion
    public async describeInstance(target: string): Promise<InstanceInformation> {
```

justinmk3 (approved):

justinmk3 (approved):

PR #3620, opened 2023-06-27
authWebview: Bug Bash fixes

nkomonen-amazon:
## Problem
Various fixes from the bug bash.

## Solution

See each commit for info on the specific fix

### Additional

- Changed the mechanism for detecting if user is a first time user or not. See commit `refactor: is first extension use logic`

## License

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -68,8 +68,8 @@ import { Logging } from './shared/logger/commands'
 import { UriHandler } from './shared/vscode/uriHandler'
 import { telemetry } from './shared/telemetry/telemetry'
 import { Auth } from './auth/auth'
-import { openUrl } from './shared/utilities/vsCodeUtils'
 import { showMessageWithUrl } from './shared/utilities/messages'
+import { openUrl } from './shared/utilities/vsCodeUtils'
 
 let localize: nls.LocalizeFunc
 
--- a/src/shared/utilities/vsCodeUtils.ts
+++ b/src/shared/utilities/vsCodeUtils.ts
@@ -12,6 +12,7 @@ import { CancellationError, Timeout, waitTimeout, waitUntil } from './timeoutUti
 import { telemetry } from '../telemetry/telemetry'
 import * as semver from 'semver'
 import { isNonNullable } from './tsUtils'
+import globals from '../extensionGlobals'
 
 // TODO: Consider NLS initialization/configuration here & have packages to import localize from here
 export const localize = nls.loadMessageBundle()
@@ -44,6 +45,36 @@ export async function closeAllEditors() {
                 vscode.window.activeTextEditor!.document.fileName
             }" was still open after executing "closeAllEditors"`
         )
+    }
+}
+
+/**
+ * Class to get info about the extension being first used
+ */
+export class ExtensionUse {
+    private readonly isExtensionFirstUseKey = 'isExtensionFirstUse'
+
+    // The result of if is first use for the remainder of the extension session.
+    // This will reset on next startup.
+    private isFirstUseCurrentSession: boolean | undefined
+
+    isFirstUse(state: vscode.Memento = globals.context.globalState): boolean {
+        if (this.isFirstUseCurrentSession !== undefined) {
+            return this.isFirstUseCurrentSession
+        }
+
+        this.isFirstUseCurrentSession = state.get(this.isExtensionFirstUseKey, true)
+
+        // Update state, so next time it is not first use
+        state.update(this.isExtensionFirstUseKey, false)
+
+        return this.isFirstUseCurrentSession
+    }
+
+    static #instance: ExtensionUse
+
+    static get instance(): ExtensionUse {
+        return (this.#instance ??= new ExtensionUse())
     }
 }
 
--- a/src/test/shared/utilities/vscodeUtils.test.ts
+++ b/src/test/shared/utilities/vscodeUtils.test.ts
@@ -7,6 +7,8 @@ import * as assert from 'assert'
 import { VSCODE_EXTENSION_ID } from '../../../shared/extensions'
 import * as vscodeUtil from '../../../shared/utilities/vsCodeUtils'
 import * as vscode from 'vscode'
+import { FakeExtensionContext } from '../../fakeExtensionContext'
+import { ExtensionUse } from '../../../shared/utilities/vsCodeUtils'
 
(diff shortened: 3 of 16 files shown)


src/test/shared/utilities/vscodeUtils.test.ts
```diff
@@ -7,6 +7,9 @@ import * as assert from 'assert'
 import { VSCODE_EXTENSION_ID } from '../../../shared/extensions'
 import * as vscodeUtil from '../../../shared/utilities/vsCodeUtils'
 import * as vscode from 'vscode'
+import { FakeExtensionContext } from '../../fakeExtensionContext'
+import { beforeEach } from 'mocha'
```
JadenSimon, line 11:
```suggestion
```

`mocha` should be global

src/shared/utilities/vsCodeUtils.ts
```diff
@@ -47,6 +48,49 @@ export async function closeAllEditors() {
...
+     */
+    updateIsFirstUse(ctx: vscode.ExtensionContext): void {
+        const isFirst = ctx.globalState.get(ExtensionUse.isExtensionFirstUseKey, true)
+
+        if (isFirst) {
+            ctx.globalState.update(ExtensionUse.isExtensionFirstUseKey, false)
+        }
+
+        this.isFirstUseState = isFirst
+    }
```
JadenSimon, line 73: Should we use `globals` to do this implicitly?

src/shared/utilities/vsCodeUtils.ts
```diff
@@ -47,6 +48,49 @@ export async function closeAllEditors() {
...
+     * IMPORTANT: This must be run first.
+     *
+     * Updates the {@link isFirstUse} variable to indicate
+     * if this is the first use of the extension.
+     */
+    updateIsFirstUse(ctx: vscode.ExtensionContext): void {
+        const isFirst = ctx.globalState.get(ExtensionUse.isExtensionFirstUseKey, true)
+
+        if (isFirst) {
+            ctx.globalState.update(ExtensionUse.isExtensionFirstUseKey, false)
```
justinmk3, line 69: (off topic) this is another example hinting that we need a globalState shared module, similar to `shared/settings`. Doesn't look like there is much need for a dedicated `ExtensionUse` class besides the lack of that

src/extension.ts
```diff
@@ -89,6 +89,8 @@ export async function activate(context: vscode.ExtensionContext) {
     )
     globals.outputChannel = toolkitOutputChannel
 
+    ExtensionUse.instance.updateIsFirstUse(context)
```
justinmk3, line 92: this is pretty early, would it make sense to be conservative and only set this after the New Connections page successfully loaded?

src/test/shared/utilities/vscodeUtils.test.ts
```diff
@@ -99,3 +102,60 @@ describe('buildMissingExtensionMessage()', function () {
         )
     })
 })
+
+describe('isExtensionFirstUse', function () {
```
justinmk3, line 106: Could probably avoid this test code if we had a globalState module

src/extension.ts
nkomonen-amazon: I figured it didn't hurt to do this early since this could be used in other places eventually and the earlier we initialized it, the less likely to run in to a race condition.

src/shared/utilities/vsCodeUtils.ts
nkomonen-amazon: Makes things much simpler using this. Updating to not need to use a class at all.

src/test/shared/utilities/vscodeUtils.test.ts
nkomonen-amazon: This test was simplified, so not as bad as before. There was a small intricacy managing some state for the current extension session (until extension is stopped) which required this more complex solution.

src/test/shared/utilities/vscodeUtils.test.ts
```diff
@@ -99,3 +101,32 @@ describe('buildMissingExtensionMessage()', function () {
...
+    it('indicates the extension is loaded for the first time', function () {
+        assert.strictEqual(instance.isFirstUse(fakeState), true)
+    })
+
+    it('indicates the extension is loaded for the first time on subsequent calls', function () {
+        assert.strictEqual(instance.isFirstUse(fakeState), true)
+        assert.strictEqual(instance.isFirstUse(fakeState), true)
+    })
+
+    it('indicates the extension is NOT loaded for the first time on a new startup', function () {
```
justinmk3, line 123:
```suggestion
    it('is true on first startup', function () {
```

src/test/shared/utilities/vscodeUtils.test.ts
```diff
@@ -99,3 +101,32 @@ describe('buildMissingExtensionMessage()', function () {
...
+    let fakeState: vscode.Memento
+    let instance: ExtensionUse
+
+    beforeEach(async function () {
+        fakeState = (await FakeExtensionContext.create()).globalState
+        instance = new ExtensionUse()
+    })
+
+    it('indicates the extension is loaded for the first time', function () {
+        assert.strictEqual(instance.isFirstUse(fakeState), true)
```
justinmk3, line 115: looks redundant with the test(s) below

src/test/shared/utilities/vscodeUtils.test.ts
```diff
@@ -99,3 +101,32 @@ describe('buildMissingExtensionMessage()', function () {
...
+    })
+
+    it('indicates the extension is loaded for the first time on subsequent calls', function () {
```
justinmk3, line 118:
```suggestion
    it('is false after first startup`', function () {
```

src/test/shared/utilities/vscodeUtils.test.ts
```diff
@@ -99,3 +101,32 @@ describe('buildMissingExtensionMessage()', function () {
...
+        assert.strictEqual(instance.isFirstUse(fakeState), true)
+        assert.strictEqual(instance.isFirstUse(fakeState), true)
```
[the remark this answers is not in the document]
nkomonen-amazon: Makes sense, I've shrunk to a single test which covers all cases

nkomonen-amazon:
@justinmk3 I fixed the tests, should be good to go now!

PR #3622, opened 2023-06-27
ec2: add explorer node w/ dropdown for instances. 

Hweinstock:
## Problem
Feature currently only accessible from the command palette, and this therefore not discoverable. 
## Solution
As a first step, add a node to the explorer that lists the ec2 instances associated with the account. On hover, we show the instance Id, but show the name in the explorer. 

image

## License


src/ec2/explorer/ec2InstanceNode.ts
```diff
@@ -0,0 +1,44 @@
...
+    public get name(): string {
+        return getNameOfInstance(this.instance) ?? 'Unnamed instance'
+    }
+
+    public get InstanceId(): string {
+        return this.instance.InstanceId!
+    }
+
+    public get arn(): string {
+        return `arn:aws:ec2:${this.regionCode}:${globals.awsContext.getCredentialAccountId()}:instance/${
```
justinmk3, line 40: does `Ec2Instance` or some other SDK function not have a arn() method?

Hweinstock: I don't believe so, or at least I couldn't find it. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html.

justinmk3: ok, as noted in that post we should avoid hardcoding "aws". 

- example: https://github.com/aws/aws-toolkit-vscode/blob/d66f115ee4e1fed8d87fa6a59bd83a1c5925ac05/src/apigateway/explorer/apiNodes.ts#L45
- the partition is passed from the explorer tree root: https://github.com/aws/aws-toolkit-vscode/blob/d66f115ee4e1fed8d87fa6a59bd83a1c5925ac05/src/awsexplorer/regionNode.ts#L33
- it might be more robust if we use the sdk's "builder" https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/functions/_aws_sdk_util_arn_parser.build.html

src/test/ec2/explorer/ec2ParentNode.test.ts
```diff
@@ -18,6 +18,7 @@ describe('ec2ParentNode', function () {
     let testNode: Ec2ParentNode
     let instanceNames: string[]
     const testRegion = 'testRegion'
+    const testPartition = ''
```
justinmk3, line 21: why empty string? usually a good idea to have a realistic test value.

src/awsexplorer/regionNode.ts
```diff
@@ -107,6 +115,9 @@ export class RegionNode extends AWSTreeNodeBase {
         const partitionId = this.regionProvider.getPartitionId(this.regionCode) ?? defaultPartition
         const childNodes: AWSTreeNodeBase[] = []
         for (const { serviceId, createFn } of serviceCandidates) {
```
justinmk3, line 117: if this for-loop was less clever we could have a more ergonomic way to define "conditional" items.
```suggestion
        for (const service of serviceCandidates) {
            if (service.when === undefined || !service.when()) {
                continue
            }
```

Then in `serviceCandidates` we can define the item like this:

```
    {
        serviceId: 'ec2',
        when: DevSettings.instance.isDevMode(),
        createFn: (regionCode: string, partitionId: string) =>
            new Ec2ParentNode(regionCode, partitionId, new Ec2Client(regionCode)),
    },
```

src/awsexplorer/regionNode.ts
```diff
@@ -26,54 +26,75 @@ import { DefaultS3Client } from '../shared/clients/s3Client'
...
+            new Ec2ParentNode(regionCode, partitionId, new Ec2Client(regionCode)),
+    },
     {
         serviceId: 'ecr',
+        when: () => true,
         createFn: (regionCode: string) => new EcrNode(new DefaultEcrClient(regionCode)),
     },
     {
         serviceId: 'ecs',
+        when: () => true,
```
justinmk3, line 67: I was hoping this wouldn't be needed, that's what the `service.when === undefined` check is for.

justinmk3 (approved):

src/awsexplorer/regionNode.ts
justinmk3: Oh that was just my logic error :) `when` should not be required

src/awsexplorer/regionNode.ts
```diff
@@ -26,6 +26,9 @@ import { DefaultS3Client } from '../shared/clients/s3Client'
 import { DefaultSchemaClient } from '../shared/clients/schemaClient'
 import { getEcsRootNode } from '../ecs/model'
 import { compareTreeItems, TreeShim } from '../shared/treeview/utils'
+import { Ec2ParentNode } from '../ec2/explorer/ec2ParentNode'
+import { DevSettings } from '../shared/settings'
+import { Ec2Client } from '../shared/clients/ec2Client'
 
 const serviceCandidates = [
```
justinmk3, line 33: Let's leave a comment here giving a hint about the `when` feature:

```
// You can mark a feature as "dev-mode only" by adding a `when` field:
//    {
//        serviceId: 'foo',
//        when: () => DevSettings.instance.isDevMode(),
//        createFn: ...
//    },
```

PR #3630, opened 2023-06-29
authWebview: Webview final release

nkomonen-amazon:
- Includes Telemetry
- Removes our old walkthrough that gets shown to new users
- upgrades to use a newer telemetry package
- adds a function to check if we are in a dev env
- Disables the Dev mode, so now users will see it

## License


package.json
```diff
@@ -3288,48 +3287,7 @@
                 }
             }
         ],
-        "walkthroughs": [
```
nkomonen-amazon, line 3291: Removes walkthrough since we will be showing the Add Connection page on first startup

src/auth/ui/vue/show.ts
```diff
@@ -79,8 +78,25 @@ export class AuthWebview extends VueWebview {
...
+                return false
+            }
+            telemetry.auth_addConnection.emit({
+                source: this.getSource() || '',
+                credentialSourceId: 'sharedCredentials',
+                result: 'Failed',
+                reason: e.message,
+            })
+            return false
+        }
```
JadenSimon, line 99: Use `run` when instrumenting workflows:
```suggestion
        try {
            await telemetry.auth_addConnection.run(span => {
                span.record({
                    source: this.getSource() || '',
                    credentialSourceId: 'sharedCredentials' 
                })
            
                return tryAddCredentials(profileName, data, true)    
            })
        
            return true
        } catch {
            return false
        }
```

src/auth/ui/vue/show.ts
```diff
@@ -332,18 +358,232 @@ export class AuthWebview extends VueWebview {
...
+     */
+    async emitOpened() {
+        telemetry.auth_addConnection.emit({
+            source: this.getSource() ?? '',
+            reason: 'opened',
+            authConnectionsCount: await this.getConnectionCount(),
+            authEnabledAreas: builderCommaDelimitedString(this.getUnlockedFeatures()),
+            result: 'Cancelled',
+        })
+    }
```
JadenSimon, line 475: Not a fan of using `Cancelled` when nothing was cancelled. It would make more sense to attach this data to one of the entrypoint events.

src/auth/ui/vue/show.ts
```diff
@@ -332,18 +358,232 @@ export class AuthWebview extends VueWebview {
...
+    /**
+     * The metric emitted when the webview is closed by the user.
+     */
+    async emitClosed() {
+        const previousAuthType = this.previousAuthType
+        const previousFeatureType = this.previousFeatureType
+        if (this.previousFeatureType && this.previousAuthType) {
+            // We are closing the webview, emit a final cancellation if they were
+            // interacting with a form but failed to complete it.
+            this.endExistingAuthFormInteraction(this.previousFeatureType, this.previousAuthType)
```
JadenSimon, line 486: Why can't we just attach `this.invalidFields` to the event later in this method?

zixlin7 (approved):

src/auth/ui/vue/show.ts
```diff
@@ -332,18 +358,232 @@ export class AuthWebview extends VueWebview {
...
+            source: this.getSource() ?? '',
+            authUiElement: previousFeatureType,
+            credentialSourceId: previousAuthType,
+            reason: 'closed',
+            authConnectionsCount: await this.getConnectionCount(),
+            authEnabledAreas: builderCommaDelimitedString(this.getUnlockedFeatures()),
+            result: 'Cancelled',
+        })
+
+        this.setSource(undefined)
```
JadenSimon, line 499: What's the purpose of clearing this when it's only set on init?

src/auth/activation.ts
```diff
@@ -48,8 +51,74 @@ export async function initialize(
...
+
+    // Metrics that are emitted if existing connections are detected
+    const allConnections = await Auth.instance.listConnections() // implicitly loads user credentials
+    const reason = 'alreadyHadAuth'
+
+    const credentialsConnections = allConnections.filter(isIamConnection)
+    if (credentialsConnections.length > 0) {
+        telemetry.auth_addConnection.emit({
+            source: 'firstStartup' as AuthSource,
+            reason,
```
JadenSimon, line 118: The user might not have added the connection through the Toolkit though? Emitting this event is misleading.

src/auth/ui/vue/authForms/manageIdentityCenter.vue
```diff
@@ -155,25 +181,50 @@ export default defineComponent({
             this.emitAuthConnectionUpdated({ id: this.state.id, isConnected: actualIsConnected, cause })
         },
         async getRegion() {
+            client.startAuthFormInteraction(this.state.featureType, 'iamIdentityCenter')
             const region = await this.state.getRegion()
+            if (!region) {
+                return
+            }
+            this.errors.region = ''
```
JadenSimon, line 189: In what situation would there ever be a region error? Shouldn't it default to something?

src/auth/activation.ts
```diff
@@ -48,8 +51,74 @@ export async function initialize(
...
+     * on their local machine, it looks at the current values in its internal store.
+     * So if this returns false it simply means the extension is not aware of it yet,
+     * but credentials could exist in something like a `credentials` file.
+     *
+     * If the user has existing credentials on their system, but this returns false,
+     * we can assume this extension has not run before since it would
+     * have discovered them and added to the extensions internal store, resulting in
+     * this returning true.
+     */
+    return Auth.instance.hasConnections
```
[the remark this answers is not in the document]
nkomonen-amazon: Yeah this is old, removing

src/auth/ui/vue/telemetry.vue
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
+import { defineComponent } from 'vue'
+import { WebviewClientFactory } from '../../../webviews/client'
+import { AuthUiClick, AuthWebview } from './show'
+
+const client = WebviewClientFactory.create<AuthWebview>()
+
+export default defineComponent({
```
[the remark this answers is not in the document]
nkomonen-amazon: Yeah, I was using it so that other Components could extend it, allowing us to reuse the `emitUiClick()` method it defines within the HTML. Otherwise we need to define the same method for each component.

src/auth/ui/vue/show.ts
nkomonen-amazon: Right now the `result` field is required, but I don't think we need it to be. `Cancelled` is a placeholder for now, since if we see `reason: 'opened'` we know that `result` is irrelevant.

src/auth/ui/vue/show.ts
nkomonen-amazon: This is just some clean up I had, but isn't necessary since the value is overridden on next startup.

src/auth/ui/vue/authForms/manageIdentityCenter.vue
nkomonen-amazon: This is for errors like "please select a region"

src/auth/ui/vue/show.ts
nkomonen-amazon: The event this method emits is about a specific auth form's failed interaction, so the invalid fields are tied directly to this specific auth form. The event that comes later is a general 'closed' event.

src/auth/activation.ts
```diff
@@ -48,8 +51,74 @@ export async function initialize(
...
+        interval: 500,
+        timeout: 30000,
+    })
+
+    // Metric that is emitted for ALL new users
+    telemetry.auth_addConnection.emit({
+        source: 'firstStartup' as AuthSource,
+        reason: 'firstStartup',
+        result: 'Cancelled',
+    })
```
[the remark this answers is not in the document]
nkomonen-amazon: This is just a default metric that gets emitted for all new users, the subsequent one is for specifics about the existing credentials count.

src/auth/activation.ts
```diff
@@ -48,8 +51,74 @@ export async function initialize(
  */
 async function showManageConnectionsOnStartup() {
     if (!ExtensionUse.instance.isFirstUse()) {
+        getLogger().debug(
+            'firstStartup: This is not the users first use of the extension, skipping showing Add Connections page.'
```
justinmk3, line 55: Avoid verbose messages in logs and UI. 
```suggestion
            'firstStartup: not first use, skipping showing Add Connections page.'
```

src/auth/commands.ts
```diff
@@ -14,13 +14,17 @@ import { showConnectionsPageCommand } from './utils'
 export class AuthCommandBackend {
     constructor(private readonly extContext: vscode.ExtensionContext) {}
 
-    public showConnectionsPage(serviceToShow?: ServiceItemId) {
+    public showConnectionsPage(source: AuthSource, serviceToShow?: ServiceItemId) {
         // Edge case where called by vscode UI and non ServiceItemId object
         // is passed in.
+        if (typeof source !== 'string') {
+            source = 'unknown'
```
justinmk3, line 21: 👀

src/auth/ui/vue/show.ts
```diff
@@ -332,18 +358,232 @@ export class AuthWebview extends VueWebview {
...
+    }
+}
+
+export type AuthUiClick =
+    | 'auth_signUpForFree'
+    | 'auth_infoIAMIdentityCenter'
+    | 'auth_learnMoreAWSResources'
+    | 'auth_learnMoreCodeWhisperer'
+    | 'auth_learnMoreCodeCatalyst'
+    | 'auth_learnMoreBuilderId'
```
justinmk3, line 549: This looks like a maintenance risk. Also makes me wonder how much of the code added in this PR is view-side duplication for telemetry?

nkomonen-amazon: The UI will be different between IDEs so we will need our own custom names for things since one implementation may have things the other does not.

src/auth/activation.ts
```diff
@@ -48,8 +51,74 @@ export async function initialize(
...
+     * we can assume this extension has not run before since it would
+     * have discovered them and added to the extensions internal store, resulting in
+     * this returning true.
+     */
+    return Auth.instance.hasConnections
+}
+
+async function emitFirstStartupMetrics() {
+    // HACK: Telemetry client may not be initialized yet, wait until it exists
+    await waitUntil(async () => telemetry, {
```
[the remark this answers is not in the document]
nkomonen-amazon: New pr changes do use telemetry here anymore, so not an issue atm.

src/auth/activation.ts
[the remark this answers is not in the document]
nkomonen-amazon: Removed these metrics in the new PR

src/auth/ui/vue/show.ts
nkomonen-amazon: Removed this emit, not needed anymore

src/auth/activation.ts
nkomonen-amazon: Removed this, not needed anymore

src/auth/ui/vue/show.ts
```diff
@@ -332,18 +351,382 @@ export class AuthWebview extends VueWebview {
...
+    }) {
+        telemetry.auth_addConnection.emit({
+            source: this.#authSource ?? '',
+            credentialSourceId: args.authType,
+            featureId: args.featureType,
+            result: args.result,
+            reason: args.reason,
+            invalidInputFields: args.invalidFields ? builderCommaDelimitedString(args.invalidFields) : undefined,
+            attempts: args.attempts,
+            isAggregated: true,
```
JadenSimon, line 575: I'm a bit confused about this field. If the event is emitted on every attempt, how could it be aggregated? I thought `auth_addedConnections` was the aggregate event.

src/auth/ui/vue/authForms/manageBuilderId.vue
```diff
@@ -77,10 +79,23 @@ export default defineComponent({
...
-            this.error = await this.state.startBuilderIdSetup()
-            if (this.error) {
+            client.startAuthFormInteraction(this.state.featureType, 'awsId')
+            const authError = await this.state.startBuilderIdSetup()
+
+            if (authError) {
+                this.error = authError.text
                 this.stage = await this.state.stage()
+
+                client.failedAuthAttempt({
```
JadenSimon, line 89: For future work I would suggest avoiding this "round-tripping" pattern by moving more logic/state out of the webview. Shorter data flow paths are easier to understand.

src/auth/ui/vue/root.vue
```diff
@@ -221,16 +229,10 @@ export default defineComponent({
...
-            if (this.successfulAuthConnection === undefined) {
-                return ''
-            }
-            return AuthFormDisplayName[this.successfulAuthConnection]
-        },
     },
     methods: {
         isLandscape() {
-            return this.currWindowWidth > 1200
+            return this.currWindowWidth > 1120
```
JadenSimon, line 235: I think usually you would check this using viewport ratios and/or orientation. But I'm sure this works fine.

src/auth/ui/vue/show.ts
```diff
@@ -332,18 +351,382 @@ export class AuthWebview extends VueWebview {
...
+
+        if (numNewAuthConnections < 0) {
+            // Edge Case:
+            // numAuthConnections gets the latest number of connections, and it is
+            // possible that the user signed out or removed connections they initially had.
+            // If this is the case, we will set the total connections to what it initially was
+            // since we don't consider removing connections as not having existed.
+            numAuthConnections = this.initialNumConnections!
+            numNewAuthConnections = newConnectedAuths.size // best effort guess but can be wrong
+        }
```
JadenSimon, line 616: I don't think it would be unreasonable to leave it as a negative number. They removed a connection, so the # of new connections is negative.

JadenSimon (approved):

src/auth/ui/vue/show.ts
nkomonen-amazon: `auth_addConnection` will have 2 different types essentially. If `isAggregated` is `true` it has information regarding the # of attempts prior to it and is the final result.

I did it this way to we can both emit a metric for every single interaction (isAggregated === false). Additionally, a final one (isAggregated === true) which is for cases like the user failed multiple times, but their final one was a success.

Once we want to know their final result of all attempts, we only need to look at the ones with `isAggregated === true`

src/auth/ui/vue/authForms/manageBuilderId.vue
nkomonen-amazon: I agree, if we want to invest more in to webviews it'll be useful to add functionality to be able to use multiple WebviewClients so we don't need to cram everything in to one `VueWebview` class

PR #3634, opened 2023-06-30
fix(messages): "first use" missing fallback metric

nkomonen-amazon:
If the user has not had this run on their system yet, they will
not have the state variable existing yet. It will assume they
are a first time user.

### Solution:

- If they have no state, then fallback to checking if we are
aware of existing connections.

- Had to move the ExtensionUse class since we now have a circular
  dependency due to using the Auth class

## License

### DIFF
--- a/src/auth/utils.ts
+++ b/src/auth/utils.ts
@@ -43,6 +43,7 @@ import { Auth } from './auth'
 import { validateIsNewSsoUrl, validateSsoUrlFormat } from './sso/validation'
 import { openUrl } from '../shared/utilities/vsCodeUtils'
 import { AuthSource } from './ui/vue/show'
+import { getLogger } from '../shared/logger'
 
 // TODO: Look to do some refactoring to handle circular dependency later and move this to ./commands.ts
 export const showConnectionsPageCommand = 'aws.auth.manageConnections'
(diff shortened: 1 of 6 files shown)


src/test/credentials/utils.test.ts
```diff
@@ -0,0 +1,58 @@
...
+        assert.strictEqual(instance.isFirstUse(fakeState), true, 'Failed on second call.')
+
+        // Mimic new extension startup, since a new instance
+        // is created on each load of the extension.
+        const secondInstance = new ExtensionUse()
+
+        assert.strictEqual(secondInstance.isFirstUse(fakeState), false, 'Failed on new startup.')
+    })
+
+    it('indicates IS first time user when: (state value not exists + NOT has existing connections)', async function () {
```
justinmk3, line 31:
```suggestion
    it('true when: (state value not exists + NOT has existing connections)', async function () {
```

src/test/credentials/utils.test.ts
```diff
@@ -0,0 +1,58 @@
...
+        await makeStateValueNotExist()
+        const notHasExistingConnections = () => false
+        assert.strictEqual(
+            instance.isFirstUse(fakeState, notHasExistingConnections),
+            true,
+            'No existing connections, should be first use'
+        )
+    })
+
+    it('indicates IS NOT first time user when: (state value not exists + has existing connections)', async function () {
```
justinmk3, line 41:
```suggestion
    it('false when: (state value not exists + has existing connections)', async function () {
```

src/auth/utils.ts
```diff
@@ -556,3 +557,55 @@ export class AuthNode implements TreeNode<Auth> {
...
+        this.isFirstUseCurrentSession = state.get(this.isExtensionFirstUseKey)
+
+        if (this.isFirstUseCurrentSession === undefined) {
+            // The variable in the store is not defined yet, fallback to checking if they have existing connections.
+            this.isFirstUseCurrentSession = !hasExistingConnections()
+
+            getLogger().info(
+                `isFirstUse: State not found, marking them as '${
+                    this.isFirstUseCurrentSession ? '' : 'NOT '
+                }first use' since they 'did ${this.isFirstUseCurrentSession ? 'NOT ' : ''}have existing connections'.`
```
justinmk3, line 596: For NOT first-time use this should be logged at verbose or debug log-level. Else it will always appear in the logs for most users.

nkomonen-amazon: This scenario only triggers when the global state variable is `undefined`, when the function has never run. After this function runs it will set the global state variable to `false`.

So this only runs the first ever time this function is run for a user, then never again since the value will either be `true` or `false. Did you still not want to log it with `info` for all first time users?

justinmk3: ok, info is fine then.

PR #3635, opened 2023-07-01
codewhisperer: update completion type Line/Block aggregation

Will-ShaoHua:
## Problem

As per our sci team's request, update session level `CompletionType`.

### Before 
- `Block` if there is a `Block` completion present in the session, otherwise `Line`

### After
- Only check the 'first' completion, if it's `Block`, then the session is `Block` otherwise `Line`

## License

### DIFF
--- a/src/codewhisperer/service/recommendationHandler.ts
+++ b/src/codewhisperer/service/recommendationHandler.ts
@@ -202,11 +202,9 @@ export class RecommendationHandler {
                 TelemetryHelper.instance.triggerType = triggerType
                 TelemetryHelper.instance.CodeWhispererAutomatedtriggerType =
                     autoTriggerType === undefined ? 'KeyStrokeCount' : autoTriggerType
-                if (
-                    recommendation.length > 0 &&
-                    recommendation[0].content.search(CodeWhispererConstants.lineBreak) !== -1
-                ) {
-                    completionType = 'Block'
+                if (page === 0 && recommendation.length > 0) {
+                    completionType =
+                        recommendation[0].content.search(CodeWhispererConstants.lineBreak) !== -1 ? 'Block' : 'Line'
                 }
                 TelemetryHelper.instance.completionType = completionType
                 requestId = resp?.$response && resp?.$response?.requestId


leigaol (approved):

Will-ShaoHua:
I am wondering since kotlin generateTelemetry will always generate a `Unknown` when generating an enum field whereas typescript doesn't. The question here is that jetbrains does use the `Unknown` type, therefore if I want to 100% align both IDEs, I may need to update the schema for typescript to have `Unknown`.

For example in typescript, 
```
type CodeWhispererCompletionType = 'Line' || 'Block'
```

in Kotlin
```
enum class CodeWhispererCompletionType {
    Line, 
    Block,
    Unknown
}
```

justinmk3:
> The question here is that jetbrains does use the `Unknown` type, therefore if I want to 100% align both IDEs, I may need to update the schema for typescript to have `Unknown`.

I think they both treated as empty when data is sent to ES. Are you seeing something different in ES?

Will-ShaoHua:
> 

yea @justinmk3 , we can see `unknown` type on ES https://telemetry-externalprod.ide-toolkits.dev-tools.aws.dev/_plugin/kibana/goto/840d71e81c7f303ec5582dc8e6837b8c?security_tenant=global.

I think the correct way is not to directly use this on JB, it is just a fallback value
```
public enum class CodewhispererAutomatedTriggerType(
    private val `value`: String,
) {
    KeyStrokeCount("KeyStrokeCount"),
    SpecialCharacters("SpecialCharacters"),
    Enter("Enter"),
    IntelliSenseAcceptance("IntelliSenseAcceptance"),
    IdleTime("IdleTime"),
    Classifier("Classifier"),
    Unknown("unknown"),
    ;

    public override fun toString(): String = value

    public companion object {
        public fun from(type: String): CodewhispererAutomatedTriggerType =
                values().firstOrNull { it.value == type } ?: Unknown
    }
}
```

Will-ShaoHua:
however we do need a 3rd type of completion type for an empty recommendation, that's where we send unknown from JB, therefore, I may need to update telemetry schema and add 3rd type e.g. `empty` or `unknown`. (personally prefer empty as unknown is a fallback for JB)

justinmk3:
For reference here is an existing precedent: https://github.com/aws/aws-toolkit-vscode/blob/d66f115ee4e1fed8d87fa6a59bd83a1c5925ac05/src/shared/telemetry/telemetryClient.ts#L26C1-L28

JadenSimon (approved):

Will-ShaoHua:
i will take the discussion as followup @JadenSimon @justinmk3 thanks!

PR #3640, opened 2023-07-03
fix(ec2): explorer assumes names are unique

Hweinstock:
## Problem
In the explorer, assuming the names are unique has the following consequences:
- it is used as a key in the map connecting parent -> child nodes.
- it is displayed by name, meaning each instance with same name appears identical.
Ex. image

## Solution
- Remove this assumption, and use the instanceIds to key the child nodes. 
- add instance Ids following the names in explorer. 
image

## License


justinmk3:
> With duplicate named EC2 instances, there is no way to distinguish other than hovering or right clicking to get the instance Id.

Good point, so let's always mention the instance id in parentheses.

justinmk3 (approved):

src/ec2/explorer/ec2InstanceNode.ts
```diff
@@ -21,8 +21,8 @@ export class Ec2InstanceNode extends AWSTreeNodeBase implements AWSResourceNode
 
     public update(newInstance: Ec2Instance) {
         this.setInstance(newInstance)
-        this.label = this.name
-        this.tooltip = this.InstanceId
+        this.label = this.name + ` (${this.InstanceId})`
+        this.tooltip = this.name + '\n' + this.InstanceId + '\n' + this.arn
```
justinmk3, line 25: nit
```suggestion
        this.tooltip = `${this.name}\n${this.InstanceId}\n${this.arn}`
```

justinmk3 (approved):

PR #3642, opened 2023-07-05
fix(codewhisperer): filter out blank code chunk

Will-ShaoHua:
## License

### DIFF
--- a/.changes/next-release/Bug Fix-16b028ca-252d-4bbe-9702-a3566377aa9b.json
+++ b/.changes/next-release/Bug Fix-16b028ca-252d-4bbe-9702-a3566377aa9b.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Fix inproper request format when sending empty supplemental context"
+	"description": "codewhisperer: improper request format when sending empty supplemental context"
 }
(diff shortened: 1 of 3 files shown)


src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -68,7 +68,10 @@ export async function fetchSupplementalContextForSrc(
         })
     }
 
-    return supplementalContexts
+    // DO NOT send code chunk with empty content
```
leigaol, line 71: You can check this BEFORE you push anything into this supplementalContexts list. Doing it in the end will run the for loop twice.

Will-ShaoHua: yea make sense, but I prefer keeping it for now so that we have a similar place/structure filter out blank strings between 2 implementation. Plus, the size of the code chunk is constant (1~3), the runtime shouldn't be an issue.

Will-ShaoHua:
need a changelog

leigaol (approved):

.changes/next-release/Bug Fix-16b028ca-252d-4bbe-9702-a3566377aa9b.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Fix inproper request format when sending empty supplemental context"
```
JadenSimon, line 3:
```suggestion
	"description": "codewhisperer: improper request format when sending empty supplemental context"
```

JadenSimon (approved):

Will-ShaoHua:
@JadenSimon let me sync up with Mahit and merge it once we are sure if we can merge the jb corresponding one or not thanks!

Will-ShaoHua:
@JadenSimon @justinmk3 we are good to merge this one thanks!

PR #3644, opened 2023-07-05
feat(ec2): add utility functions to explorer

Hweinstock:
Currently a draft as it builds off work done in https://github.com/aws/aws-toolkit-vscode/pull/3640

## Problem
User's expect to be able to have this information readily available (as in line with other services). 
## Solution
commands added on right click of explorer node:
- copy `ARN`
- copy `Name`
- copy `instanceId` 

image

## License

### DIFF
--- a/src/ec2/commands.ts
+++ b/src/ec2/commands.ts
@@ -34,13 +34,12 @@ export async function tryConnect(selection?: Ec2Selection): Promise<void> {
 }
 
 export async function copyInstanceId(instanceId: string): Promise<void> {
-    console.log('copied it!')
     try {
         if (!instanceId) {
             throw new ToolkitError(`Attempting to copy undefined instanceId.`)
         }
         await copyToClipboard(instanceId)
     } catch (e) {
-        throw new ToolkitError('Failed to copy instanceId', { cause: e as Error })
+        throw new ToolkitError('Failed to copy instanceId', { code: 'InvalidResource', cause: e as Error })
     }
 }
--- a/src/ec2/commands.ts
+++ b/src/ec2/commands.ts
@@ -10,6 +10,8 @@ import { Ec2Selection } from './utils'
 import { RegionSubmenuResponse } from '../shared/ui/common/regionSubmenu'
 import { PromptResult } from '../shared/ui/prompter'
 import { CancellationError } from '../shared/utilities/timeoutUtils'
+import { ToolkitError } from '../shared/errors'
+import { copyToClipboard } from '../shared/utilities/messages'
 
 function getSelectionFromResponse(response: PromptResult<RegionSubmenuResponse<string>>): Ec2Selection {
     if (isValidResponse(response)) {
@@ -29,4 +31,16 @@ export async function tryConnect(selection?: Ec2Selection): Promise<void> {
 
     const ec2Client = new Ec2ConnectionManager(selection.region)
     await ec2Client.attemptEc2Connection(selection)
+}
+
+export async function copyInstanceId(instanceId: string): Promise<void> {
+    console.log('copied it!')
+    try {
+        if (!instanceId) {
+            throw new ToolkitError(`Attempting to copy undefined instanceId.`)
+        }
+        await copyToClipboard(instanceId)
+    } catch (e) {
+        throw new ToolkitError('Failed to copy instanceId', { cause: e as Error })
+    }
 }
--- a/package.json
+++ b/package.json
@@ -1404,6 +1404,11 @@
                     "group": "1@2"
                 },
                 {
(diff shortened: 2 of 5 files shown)


src/ec2/commands.ts
```diff
@@ -30,3 +32,14 @@ export async function tryConnect(selection?: Ec2Selection): Promise<void> {
     const ec2Client = new Ec2ConnectionManager(selection.region)
     await ec2Client.attemptEc2Connection(selection)
 }
+
+export async function copyInstanceId(instanceId: string): Promise<void> {
+    try {
+        if (!instanceId) {
+            throw new ToolkitError(`Attempting to copy undefined instanceId.`)
+        }
+        await copyToClipboard(instanceId)
```
justinmk3, line 41:
```suggestion
        await copyToClipboard(instanceId, 'Id')
```

src/ec2/commands.ts
```diff
@@ -30,3 +32,14 @@ export async function tryConnect(selection?: Ec2Selection): Promise<void> {
...
+    } catch (e) {
+        throw new ToolkitError('Failed to copy instanceId', { code: 'InvalidResource', cause: e as Error })
```
justinmk3, line 43: This shouldn't really be possible for copy-to-clipboard, but if it happens, the global exception handler is good enough. 

The try/catch in copyArn.ts is likely old/legacy and could be cleaned up: https://github.com/aws/aws-toolkit-vscode/blob/926fca0d358bd36c473febf182e9c30e4a2af198/src/awsexplorer/commands/copyArn.ts#L28-L30

src/ec2/commands.ts
```diff
@@ -30,3 +32,14 @@ export async function tryConnect(selection?: Ec2Selection): Promise<void> {
     const ec2Client = new Ec2ConnectionManager(selection.region)
     await ec2Client.attemptEc2Connection(selection)
 }
+
+export async function copyInstanceId(instanceId: string): Promise<void> {
+    try {
+        if (!instanceId) {
+            throw new ToolkitError(`Attempting to copy undefined instanceId.`)
```
justinmk3, line 39: param is defined as string so shouldn't be undefined. The [global handler](https://github.com/aws/aws-toolkit-vscode/blob/926fca0d358bd36c473febf182e9c30e4a2af198/src/extension.ts#L279) is good enough here though, since this is unexpected, a special error is not worth the extra code.

src/ec2/commands.ts
```diff
@@ -30,3 +32,14 @@ export async function tryConnect(selection?: Ec2Selection): Promise<void> {
     const ec2Client = new Ec2ConnectionManager(selection.region)
     await ec2Client.attemptEc2Connection(selection)
 }
+
+export async function copyInstanceId(instanceId: string): Promise<void> {
```
justinmk3, line 36: This is doing the same thing as [copyName](https://github.com/aws/aws-toolkit-vscode/blob/926fca0d358bd36c473febf182e9c30e4a2af198/src/awsexplorer/commands/copyName.ts#L14) and [copyArn](https://github.com/aws/aws-toolkit-vscode/blob/926fca0d358bd36c473febf182e9c30e4a2af198/src/awsexplorer/commands/copyArn.ts#L21).

This might be the right time to generalize them to a `src/awsexplorer/commands/copyText.ts` command. Then we only need test coverage for the generalized file. And we can delete `src/awsexplorer/commands/copyArn.ts` (all of its call sites would change to `await copyTextCommand(node, env, 'ARN')`).

Hweinstock: I gave the refactor a shot, but was unsure whether to include the instanceId as part of the same functionality. I figured it only makes sense if multiple services carry a general Id tag that could be helpful to copy.

src/shared/extensionUtilities.ts
justinmk3, line 1: did prettier reformat this file? if there are no code changes, better to avoid formatting it in this PR

src/shared/treeview/nodes/awsResourceNode.ts
```diff
@@ -13,4 +13,6 @@ export interface AWSResourceNode {
      * Returns the name of the AWS resource.
      */
     readonly name: string
+
+    readonly id?: string
```
justinmk3, line 17:
```suggestion
    /** Service-defined idenifier, e.g. for EC2 this is the "Instance Id". */
    readonly id?: string
```

justinmk3 (approved):
still some redundancy in the tests, but can improve later. 

LGTM

src/shared/extensionUtilities.ts
Hweinstock: Yea, believe so. I noticed a few others in the PR it did the same with so went ahead and undid formatting for those files.

PR #3645, opened 2023-07-05
feat: User Activity event

nkomonen-amazon:
This adds a new class which provides an event that triggers
upon user activity.

Contributes to: IDE-8671

## License


src/shared/extensionUtilities.ts
```diff
@@ -364,3 +364,53 @@ export function getComputeRegion(): string | undefined {
...
+    static instance(delay: number = 500, activityEvents = ExtensionUserActivity.activityEvents): ExtensionUserActivity {
+        if (!ExtensionUserActivity.instances[delay]) {
+            const newInstance = new ExtensionUserActivity(delay, activityEvents)
+            ExtensionUserActivity.instances[delay] = newInstance
+        }
+        return ExtensionUserActivity.instances[delay]
+    }
+
+    private constructor(delay: number, private readonly activityEvents: vscode.Event<any>[]) {
+        const throttledEvent = _.throttle(() => this.activityEvent.fire(), delay)
```
justinmk3, line 394: Does `_.throttle` "collapse" redundant events? Don't want to merely spread out the triggers, but collapse them.

justinmk3 (approved):

src/shared/extensionUtilities.ts
nkomonen-amazon: Yup it collapses them in to 1 event firing. If multiple vscode events trigger at a similar time within the `delay`, all successive calls will do nothing. So only 1 event firing per `delay`

PR #3648, opened 2023-07-07
codecatalyst: Show warning messages when CC dev env is close to inactivity shutdown

nkomonen-amazon:
## Problem

If a user is using a dev env with something like vscode, and the user stops interacting with the ide
we want to notify the dev env about this inactivity so it can automatically shutdown after a certain
amount of user inactivity.

## Solution

- A cancellable message will be shown a few minutes before shutdown, and will update
   the remaining time each minute until the final minute before shutdown.
- On the final minute a different message will be show that indicates the dev env
   will shut down in 1 minute.
- We will show messages 5 minutes before shutdown.
----
- For each message that pops up if they interact with the IDE or click one of the buttons
in the message it will refresh the shutdown timer to whatever was initially configured.
- General interactions with the IDE will cause the ide to send a heartbeat to the dev env (see `startTrackingUserActivity()`)

#### First message
Screen Shot 2023-07-10 at 3 22 24 PM

#### Updated message after a minute
Screen Shot 2023-07-10 at 3 23 21 PM

#### Final Message a minute before shutdown
Screen Shot 2023-07-10 at 3 24 22 PM

## License

### DIFF
--- a/src/codecatalyst/activation.ts
+++ b/src/codecatalyst/activation.ts
@@ -12,16 +12,17 @@ import { CodeCatalystCommands } from './commands'
 import { GitExtension } from '../shared/extensions/git'
 import { CodeCatalystAuthenticationProvider } from './auth'
 import { registerDevfileWatcher } from './devfile'
-import { DevEnvClient } from '../shared/clients/devenvClient'
+import { DevEnvClient, DevEnvActivity } from '../shared/clients/devenvClient'
 import { watchRestartingDevEnvs } from './reconnect'
 import { PromptSettings } from '../shared/settings'
 import { dontShow } from '../shared/localizedText'
 import { getIdeProperties, isCloud9 } from '../shared/extensionUtilities'
 import { Commands } from '../shared/vscode/commands2'
 import { getCodeCatalystConfig } from '../shared/clients/codecatalystClient'
-import { getThisDevEnv } from './model'
 import { isDevenvVscode } from './utils'
+import { getThisDevEnv } from './model'
 import { getLogger } from '../shared/logger/logger'
+import { InactivityMessage, shouldTrackUserActivity } from './devEnv'
 
 const localize = nls.loadMessageBundle()
 
@@ -98,6 +99,16 @@ export async function activate(ctx: ExtContext): Promise<void> {
                     CodeCatalystCommands.declared.openDevEnvSettings.execute()
                 }
             })
+        }
+
+        const maxInactivityMinutes = thisDevenv.summary.inactivityTimeoutMinutes
+        const devEnvClient = thisDevenv.devenvClient
+        const devEnvActivity = await DevEnvActivity.instanceIfActivityTrackingEnabled(devEnvClient)
+        if (shouldTrackUserActivity(maxInactivityMinutes) && devEnvActivity) {
+            const inactivityMessage = new InactivityMessage()
+            inactivityMessage.setupMessage(maxInactivityMinutes, devEnvActivity)
+
+            ctx.extensionContext.subscriptions.push(inactivityMessage, devEnvActivity)
         }
     }
 }
--- a/src/codecatalyst/activation.ts
+++ b/src/codecatalyst/activation.ts
@@ -38,6 +38,10 @@ export async function activate(ctx: ExtContext): Promise<void> {
         ...Object.values(CodeCatalystCommands.declared).map(c => c.register(commands))
     )
 
+    Commands.register('aws.codecatalyst.removeConnection', () => {
+        authProvider.removeSavedConnection()
+    })
+
     if (!isCloud9()) {
         GitExtension.instance.registerRemoteSourceProvider(remoteSourceProvider).then(disposable => {
             ctx.extensionContext.subscriptions.push(disposable)
@@ -56,11 +60,6 @@ export async function activate(ctx: ExtContext): Promise<void> {
         watchRestartingDevEnvs(ctx, authProvider)
     }
 
-    ctx.extensionContext.subscriptions.push(DevEnvClient.instance)
-    if (DevEnvClient.instance.id) {
-        ctx.extensionContext.subscriptions.push(registerDevfileWatcher(DevEnvClient.instance))
-    }
-
     const thisDevenv = (await getThisDevEnv(authProvider))?.unwrapOrElse(err => {
         getLogger().warn('codecatalyst: failed to get current Dev Enviroment: %s', err)
         return undefined
@@ -69,6 +68,11 @@ export async function activate(ctx: ExtContext): Promise<void> {
     if (!thisDevenv) {
         getLogger().verbose('codecatalyst: not a devenv, getThisDevEnv() returned empty')
     } else {
+        ctx.extensionContext.subscriptions.push(DevEnvClient.instance)
+        if (DevEnvClient.instance.id) {
+            ctx.extensionContext.subscriptions.push(registerDevfileWatcher(DevEnvClient.instance))
+        }
+
         getLogger().info('codecatalyst: Dev Environment ides=%O', thisDevenv?.summary.ides)
         if (!isCloud9() && thisDevenv && !isDevenvVscode(thisDevenv.summary.ides)) {
             // Prevent Toolkit from reconnecting to a "non-vscode" devenv by actively closing it.
@@ -96,10 +100,6 @@ export async function activate(ctx: ExtContext): Promise<void> {
             })
         }
     }
-
-    Commands.register('aws.codecatalyst.removeConnection', () => {
-        authProvider.removeSavedConnection()
-    })
 }
 
 async function showReadmeFileOnFirstLoad(workspaceState: vscode.ExtensionContext['workspaceState']): Promise<void> {
--- a/src/codecatalyst/devEnv.ts
+++ b/src/codecatalyst/devEnv.ts
@@ -1,0 +1,230 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { DevEnvironment } from '../shared/clients/codecatalystClient'
+import { DevEnvActivity } from '../shared/clients/devenvClient'
+import globals from '../shared/extensionGlobals'
+import * as vscode from 'vscode'
+import { Timeout } from '../shared/utilities/timeoutUtils'
+import { showMessageWithCancel } from '../shared/utilities/messages'
+import { isCloud9 } from '../shared/extensionUtilities'
+
+/** If we should be sending the dev env activity timestamps to track user activity */
+export function shouldTrackUserActivity(maxInactivityMinutes: DevEnvironment['inactivityTimeoutMinutes']): boolean {
+    // This value is a static value initialized when the dev env is first created.
+    //
+    // If it is updated, the dev env is restarted and this extension will restarted and grab the latest value.
+    // Due to this, we do not need to keep track of this value since the
+    //
+    // For more info see: https://docs.aws.amazon.com/codecatalyst/latest/APIReference/API_UpdateDevEnvironment.html#codecatalyst-UpdateDevEnvironment-request-inactivityTimeoutMinutes
+    return maxInactivityMinutes > 0
+}
+
+export class InactivityMessage implements vscode.Disposable {
+    #message: Message | undefined
+    #nextWholeMinute: NodeJS.Timeout | undefined
+    #beforeMessageShown: NodeJS.Timeout | undefined
+
+    /** Indicates to show first message 5 minutes before shutdown */
+    static readonly firstMessageBeforeShutdown = 5
+
+    /**
+     * Sets up all messages that are displayed to the user when the dev env is inactive and starts to get close to shutdown.
+     *
+     * @param maxInactivityMinutes The # of minutes a CodeCatalyst Dev Env can be inactive till it shuts down.
+     * @param devEnvActivity The DevEnvActivity client
+     * @param relativeMinuteMillis How many milliseconds we want a "minute" to be. This is
+     *                             useful for testing purposes.
+     */
+    async setupMessage(
+        maxInactivityMinutes: DevEnvironment['inactivityTimeoutMinutes'],
+        devEnvActivity: DevEnvActivity,
+        relativeMinuteMillis: number = 60_000
+    ) {
+        devEnvActivity.onActivityUpdate(async latestActivityTimestamp => {
+            this._setupMessage(maxInactivityMinutes, latestActivityTimestamp, devEnvActivity, relativeMinuteMillis)
+        })
+
+        // Send an initial update to the dev env on startup
+        devEnvActivity.sendActivityUpdate()
+    }
+
+    private _setupMessage(
+        maxInactivityMinutes: number,
+        latestTimestamp: number,
+        devEnvActivity: DevEnvActivity,
+        relativeMinuteMillis: number
+    ) {
+        this.clearOldMessage()
+
+        const { millisToWait, minutesSinceTimestamp } = this.millisUntilNextWholeMinute(
+            latestTimestamp,
+            relativeMinuteMillis
+        )
+        const minutesUntilShutdown = maxInactivityMinutes - minutesSinceTimestamp
+        const minutesUntilFirstMessage = minutesUntilShutdown - InactivityMessage.firstMessageBeforeShutdown
+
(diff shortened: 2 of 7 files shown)


src/codecatalyst/devEnv.ts
```diff
@@ -0,0 +1,206 @@
...
+        shutdownMessage: InactivityShutdownMessage
+    ) {
+        const maxInactivityMinutes = inactivityMinutes
+
+        devEnv.onUserActivity(() => {
+            const userIsActive = () => devEnv.updateUserActivity()
+            warningMessage.setupMessage(maxInactivityMinutes, userIsActive)
+            shutdownMessage.setupMessage(maxInactivityMinutes, userIsActive, () =>
+                warningMessage.clearExistingMessage()
+            )
```
JadenSimon, line 47: Would the warning message always be shown before the shutdown? Having these messages as two separate entities feels unnatural when it's actually a sequence of events.

src/codecatalyst/activation.ts
```diff
@@ -95,11 +105,20 @@ export async function activate(ctx: ExtContext): Promise<void> {
...
-    })
+        const maxInactivityMinutes = thisDevenv.summary.inactivityTimeoutMinutes
+        const devEnvClient = thisDevenv.devenvClient
+        if (shouldTrackUserActivity(maxInactivityMinutes)) {
+            InactivityMessages.instance.setupMessages(
+                maxInactivityMinutes,
+                devEnvClient,
+                InactivityWarningMessage.instance,
+                InactivityShutdownMessage.instance
+            )
```
JadenSimon, line 117: Feels like too many singletons. Why can't `InactivityMessages` create the messages?

src/codecatalyst/devEnv.ts
```diff
@@ -0,0 +1,206 @@
...
+     * @param devEnv The DevEnv client
+     * @returns
+     */
+    async setupMessages(
+        inactivityMinutes: DevEnvironment['inactivityTimeoutMinutes'],
+        devEnv: DevEnvClient,
+        warningMessage: InactivityWarningMessage,
+        shutdownMessage: InactivityShutdownMessage
+    ) {
+        const maxInactivityMinutes = inactivityMinutes
```
JadenSimon, line 40: Perhaps rename `inactivityMinutes` instead

src/codecatalyst/devEnv.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    async setupMessage(maxInactivityMinutes: number, userIsActive: () => Promise<any> | any): Promise<void> {
+        this.clearExistingMessage()
+
+        const showMessageAfterMinutes = this.getMinutesUntilShowWarningMessage(maxInactivityMinutes)
+
+        this.beforeWarningMessageShownTimeout = globals.clock.setTimeout(() => {
+            const inactiveMinutes = showMessageAfterMinutes
+            const remainingMinutes = maxInactivityMinutes - inactiveMinutes
+            this.showInactivityWarningMessage(inactiveMinutes, remainingMinutes, userIsActive)
```
JadenSimon, line 79:
```suggestion
            const remainingMinutes = maxInactivityMinutes - showMessageAfterMinutes
            this.showInactivityWarningMessage(showMessageAfterMinutes, remainingMinutes, userIsActive)
```

Or rename `showMessageAfterMinutes`

src/codecatalyst/devEnv.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    private getMinutesUntilShowWarningMessage(maxInactivityMinutes: number): number {
+        if (maxInactivityMinutes >= 10) {
+            // Show warning messages 5 minutes before shutdown
+            return maxInactivityMinutes - 5
+        } else {
+            // Show warning message 3 minutes before shutdown
+            return maxInactivityMinutes - 3
+        }
+    }
```
JadenSimon, line 103: This is a hint that the class could be modeled as a state machine

src/codecatalyst/devEnv.ts
```diff
@@ -0,0 +1,206 @@
...
+        inactiveMinutes: number,
+        remainingMinutes: number,
+        userIsActive: () => Promise<any> | any,
+        interval: Timeout = this.warningMessageInterval()
+    ) {
+        if (remainingMinutes <= 1) {
+            // Recursive base case, with only 1 minute left we do not want to show warning messages anymore.
+            interval.dispose()
+            return
+        }
```
JadenSimon, line 120: And that's because we show the final shutdown message right? That's a sign that the two classes shouldn't be separate.

src/shared/extensionUtilities.ts
```diff
@@ -379,32 +383,60 @@ export class ExtensionUserActivity {
...
-     * @param activityEvents The events that trigger the user activity.
+     * @param activityEvents For testing purposes. The events that trigger the user activity.
      * @returns
      */
-    static instance(delay: number = 500, activityEvents = ExtensionUserActivity.activityEvents): ExtensionUserActivity {
+    static instance(delay: number = 10_000, activityEvents?: ActivityEvent[]): ExtensionUserActivity {
+        if (activityEvents) {
+            // On custom events we don't cache the instance
+            return new ExtensionUserActivity(delay, activityEvents)
+        }
```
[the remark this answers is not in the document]
nkomonen-amazon: Makes sense, this was a premature optimization so we can come back to something like this if needed. Removing this part.

src/codecatalyst/devEnv.ts
nkomonen-amazon: Good call, refactoring this stuff

src/codecatalyst/activation.ts
nkomonen-amazon: Removed all singletons

src/codecatalyst/devEnv.ts
nkomonen-amazon: Renamed the param to `maxInactivityMinutes` instead

src/codecatalyst/devEnv.ts
nkomonen-amazon: Good call, refactored so the message is in this call

src/shared/clients/devenvClient.ts
```diff
@@ -103,6 +108,72 @@ export class DevEnvClient implements vscode.Disposable {
...
+        this.#lastKnownActivityTimestamp = timestamp
+        this.#updatedActivityEventEmitter.fire(timestamp)
+        return timestamp
+    }
+
+    async getUserActivity(): Promise<number | undefined> {
+        const response = await this.got<GetActivityResponse>('activity')
+        const latestTimestamp = response.body.timestamp ? parseInt(response.body.timestamp) : undefined
+
+        // There is a newer activity timestamp that this client did not know of
```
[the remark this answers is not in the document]
nkomonen-amazon: A Dev Env can have multiple clients connected to it, so if one is active it will update the timestamp, and later if another one calls this get method, it will be different from what it currently has.

I'll update the comment to explain 'why'.

src/codecatalyst/devEnv.ts
```diff
@@ -0,0 +1,250 @@
...
+            this.#warningMessage = undefined
+        }
+    }
+
+    /**
+     * The latest activity timestamp may not always be the current time, it may be from the past.
+     * So the amount of time that has passed since that timestamp may not be a whole minute eg: 1 minute vs 1 minutes and 1 second.
+     *
+     * Since we want to show messages on a minute by minute basis, we can wait until the next whole minute so that going forward
+     * the warning message code does not need to care about seconds.
```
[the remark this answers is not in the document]
nkomonen-amazon: We need to handle where we are given a timestamp that might not be the current time, so we will have to account for time already passed. The api can give any possible time. So this is why the code has the added complexity

src/codecatalyst/devEnv.ts
```diff
@@ -0,0 +1,250 @@
...
+     * It is updated each {@link interval} with the latest remaining minutes.
+     *
+     * @param checkNewerTimestamp sanity checks with the dev env api that the latest activity timestamp
+     *                          is the same as what this client has locally.
+     */
+    private async showInactivityWarningMessage(
+        inactiveMinutes: number,
+        remainingMinutes: number,
+        userIsActive: () => Promise<any> | any,
+        checkNewerTimestamp: () => Promise<boolean>,
```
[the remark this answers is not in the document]
nkomonen-amazon: I renamed this to indicate that it will trigger a refresh, `willRefreshOnStaleTimestamp()`

src/codecatalyst/devEnv.ts
```diff
@@ -0,0 +1,231 @@
...
+
+        if (minutesUntilShutdown <= 1) {
+            // Recursive base case, with only 1 minute left we do not want to show warning messages anymore,
+            // since we will show a shutdown message instead.
+            this.clearExistingMessage()
+
+            const imHere = `I'm here!`
+            vscode.window
+                .showWarningMessage(
+                    `Your CodeCatalyst Dev Environment has been inactive for ${minutesUserWasInactive} minutes, and will be shutdown soon.`,
```
justinmk3, line 172:
```suggestion
                    `Your CodeCatalyst Dev Environment has been inactive for ${minutesUserWasInactive} minutes, and will stop soon.`,
```

src/shared/clients/devenvClient.ts
```diff
@@ -103,6 +108,67 @@ export class DevEnvClient implements vscode.Disposable {
...
+    }
+
+    async getUserActivity(): Promise<number | undefined> {
+        const response = await this.got<GetActivityResponse>('activity')
+        const latestTimestamp = response.body.timestamp ? parseInt(response.body.timestamp) : undefined
+
+        // A single Dev Env can have multiple clients connected to it.
+        // So if one client updates the timestamp, it will be different from what the
+        // other clients assumed the last activity was.
+        if (latestTimestamp && latestTimestamp !== this.#lastKnownActivityTimestamp) {
```
justinmk3, line 128:
```suggestion
        if (latestTimestamp && latestTimestamp > this.#lastKnownActivityTimestamp) {
```

src/shared/clients/devenvClient.ts
```diff
@@ -103,6 +108,67 @@ export class DevEnvClient implements vscode.Disposable {
         // `Authorization` _should_ have two parameters (RFC 7235), MDE should probably fix that
         headers: { Authorization: this.authToken },
     })
+
+    #lastKnownActivityTimestamp: number | undefined
```
justinmk3, line 112: Can we avoid dealing with undefined for this variable:

```suggestion
    #lastKnownActivityTimestamp: number | 0
```

src/shared/extensionUtilities.ts
```diff
@@ -374,37 +375,60 @@ export class ExtensionUserActivity {
...
+                    if (!uriOfDocumentUserIsEditing) {
+                        // User was not editing any file themselves
+                        return
+                    }
+                    if (uriOfDocumentUserIsEditing !== (obj as vscode.TextDocumentChangeEvent).document.uri) {
+                        // File changed was not the one the user was actively working on
+                        return
+                    }
+                }
+                this.activityEvent.fire()
```
justinmk3, line 403: Do we have a log message for this? We need a log message like:

    ExtensionUserActivity: event "${eventName}" 

so that we can troubleshoot in case we have "spurious activity", other than the special-cases already checked above.

src/testInteg/codecatalyst/devEnv.test.ts
```diff
@@ -0,0 +1,236 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as assert from 'assert'
+import { InactivityMessage, shouldTrackUserActivity } from '../../codecatalyst/devEnv'
```
justinmk3, line 7: Looks like the devfileLocation logic should also live in this new devEnv.ts file. And its tests should be moved here. 

https://github.com/aws/aws-toolkit-vscode/blob/bd58f9e5e48bc109827d019442c2cdf97bc0afb6/src/test/codecatalyst/devfileLocation.test.ts#L13

justinmk3 (approved):

src/shared/clients/devenvClient.ts
nkomonen-amazon: It is possible an older timestamp is set, so the `latestTimestamp` could be older than the one we previously knew about.

src/shared/clients/devenvClient.ts
nkomonen-amazon: 0 implies the last user activity was at 0, and undefined implies no user activity was set yet. This distinction is important.

src/shared/clients/devenvClient.ts
```diff
@@ -103,6 +108,69 @@ export class DevEnvClient implements vscode.Disposable {
...
+    }
+
+    async getUserActivity(): Promise<number | undefined> {
+        const response = await this.got<GetActivityResponse>('activity')
+        const latestTimestamp = response.body.timestamp ? parseInt(response.body.timestamp) : undefined
+
+        // A single Dev Env can have multiple clients connected to it.
+        // So if one client updates the timestamp, it will be different from what the
+        // other clients assumed the last activity was.
+        if (latestTimestamp && latestTimestamp !== this.#lastKnownActivityTimestamp) {
```
justinmk3, line 130: Choosing different variable names would help me understand this, does this make sense:

```suggestion
        if (lastServiceActivity && lastServiceActivity !== this.#lastLocalActivity) {
```

I'm wondering: why would we enter this block if `lastLocalActivity` is _older than_ `lastServiceActivity` ? 

Are there other places where it we check the "last activity reported by MDE"?

src/testInteg/codecatalyst/devEnv.test.ts
nkomonen-amazon: I see we also have a `devfile.ts` file as well. Will look later to see how to refactor these properly.

src/shared/extensionUtilities.ts
```diff
@@ -369,42 +369,108 @@ export function getComputeRegion(): string | undefined {
...
     ]
+
+    private register(disposable: vscode.Disposable) {
+        this.disposables.push(disposable)
+    }
+
+    /**
+     * Registers events that have conditions if they should even fire
+     */
+    private registerEventsWithEdgeCases(throttledEmit: (e: vscode.Event<any>) => any) {
```
justinmk3, line 434: Could be just `registerAll`, so all cases live in one place. Separating them adds indirection without much benefit.

src/shared/clients/devenvClient.ts
```diff
@@ -103,6 +104,123 @@ export class DevEnvClient implements vscode.Disposable {
...
+    /** Send heartbeat timestamp to the Dev Env */
+    async sendHeartbeat(timestamp: number = Date.now()): Promise<number> {
+        await this.client.updateActivity()
+        this.lastLocalHeartbeat = timestamp
+        this.heartbeatEventEmitter.fire(timestamp)
+        return timestamp
+    }
+
+    /** Get the latest heartbeat timestamp from the Dev Env */
+    async getLatestHeartbeat(): Promise<number | undefined> {
```
justinmk3, line 169: have the "activity" concept, so introducing another name for it adds indirection but doesn't gain much. Can we avoid adding "heartbeat" as a quasi-synonymous concept?

nkomonen-amazon: I like using the "heartbeat" keyword, but the api is called "activity". Should I switch everything to just match the API (use "activity")?

justinmk3: If we can avoid extra jargon we should. If heartbeat is a necessary _new_ concept in addition to "activity" then it may be justified.

src/test/shared/clients/devenvClient.test.ts
```diff
@@ -0,0 +1,99 @@
...
+            }
+            return {
+                dispose: sandbox.stub()
+            }
+        })
+        
+        devEnvClientStub = createStubInstance(DevEnvClient)
+
+        devEnvActivity = (await DevEnvActivity.instanceIfActivityTrackingEnabled(devEnvClientStub as unknown as DevEnvClient, new ExtensionUserActivity(0, [userActivityEvent])))!
+        sendActivityUpdateSpy = sandbox.spy(devEnvActivity, 'sendActivityUpdate')
```
justinmk3, line 49: We fully control this. Is a spy really needed? We are using mocks and spies a lot, this should not be the default way of testing things.

justinmk3 (approved):

src/test/shared/clients/devenvClient.test.ts
nkomonen-amazon: Good find. This wasn't needed anymore, I've removed the test that uses it.

PR #3649, opened 2023-07-07
ec2: add new command for remote connect

Hweinstock:
## Problem 
We need an entry point for the remote connect functionality and a clear distinction between `openTerminal` and `openRemoteConnection`.

## Solution
- change `aws.ec2.connectToInstance` to `aws.ec2.openTerminal`. (inline with ecs naming.)
- add new command `aws.ec2.openRemoteConnecton`. (inline with CodeCatalyst naming)
- make command accessible from the command palette. 
- allow `aws.ec2.openRemoteConnection` to be empty. 
- refactor code to make clear which parts are specific to terminal and which are for general connection. 

## License


src/ec2/prompter.ts
```diff
@@ -16,6 +19,21 @@ function asQuickpickItem(instance: Ec2Instance): DataQuickPickItem<string> {
...
+    } else {
+        throw new CancellationError('user')
+    }
+}
+
+export async function promptUserForEc2Selection(): Promise<Ec2Selection> {
+    const prompter = createEc2ConnectPrompter()
+    const response = await prompter.prompt()
+
+    return getSelectionFromResponse(response)
```
JadenSimon, line 34: nit: `getSelectionFromResponse` isn't generic and seems unlikely to be re-used. Could be inlined.

package.nls.json
```diff
@@ -108,7 +108,8 @@
     "AWS.command.refreshAwsExplorer": "Refresh Explorer",
     "AWS.command.refreshCdkExplorer": "Refresh CDK Explorer",
     "AWS.command.cdk.help": "View CDK Documentation",
-    "AWS.command.ec2.connectToInstance": "Connect to EC2 Instance...",
+    "AWS.command.ec2.openTerminal": "Open terminal in EC2 instance...",
```
JadenSimon, line 111: Wording feels clunky. Maybe this is better?
```suggestion
    "AWS.command.ec2.openTerminal": "Open terminal to EC2 instance...",
```
Could also use "for" instead of "to"

package.nls.json
```diff
@@ -108,7 +108,8 @@
...
+    "AWS.command.ec2.openRemoteConnection": "Open EC2 instance in VSCode remote...",
```
JadenSimon, line 112: The Remote SSH extension uses the phrase "Connect in New Window...", we should try to use similar terminology.

JadenSimon (approved):

package.nls.json
```diff
@@ -108,7 +108,8 @@
     "AWS.command.refreshAwsExplorer": "Refresh Explorer",
     "AWS.command.refreshCdkExplorer": "Refresh CDK Explorer",
     "AWS.command.cdk.help": "View CDK Documentation",
-    "AWS.command.ec2.connectToInstance": "Connect to EC2 Instance...",
+    "AWS.command.ec2.openTerminal": "Open terminal to EC2 instance...",
+    "AWS.command.ec2.openRemoteConnection": "Connect to EC2 instance in New Window...",
```
justinmk3, line 112: This is for connecting vscode-remote, right? Then we don't need to mention "new window" because we don't have control over that. But we should mention "IDE" perhaps.

    Connect IDE to EC2 instance...

Future note: We eventually will want to think about being consistent across all of our "remote" features. For CodeCatalyst this feature is called `Open Dev Environment`, but maybe it should be called `Connect IDE to Dev Environment`.

PR #3654, opened 2023-07-11
IGNORE: feat(ec2): automatically configure /ssh_config

Hweinstock:
Currently a draft as it relies on work done in https://github.com/aws/aws-toolkit-vscode/pull/3652. 

## Problem
Building off of https://github.com/aws/aws-toolkit-vscode/pull/3652, this step is assumed to have already been done. However, it should be automated as part of the flow.

## Solution
- Before establishing Connection, check if /ssh_config already configured.
- If it not configured, automatically configure it. 
- Follows exact same steps, and code, that CodeCatalyst does to accomplish the same task. 

Notes:
- This PR also refactors the CodeCatalyst work that does this same task. To avoid duplicate code, the general functionality is placed into the new `VscodeRemoteSshConfig` object in `src/shared/extensions/ssh.ts`, with each of CodeCatalyst and EC2 extending this shared functionality for their needs. 
- This PR also tests this object through mocking it, focusing on the logic and flow of the underlying flows, rather than the technical details, such as finding ssh path. 

## License


src/ec2/tools.ts
```diff
@@ -0,0 +1,36 @@
...
+            return verifyHost
+        }
+
+        return Result.ok()
+    }
+
+    protected override createSSHConfigSection(proxyCommand: string): string {
+        return `
+# Created by AWS Toolkit for VSCode. https://github.com/aws/aws-toolkit-vscode
+Host i-* mi-*
```
justinmk3, line 31: - let's use `aws-ec2-` as the prefix, for now.
- only need one value for the hostname. the only purpose of the hostname is to "route" to our ProxyCommand.

```suggestion
Host aws-ec2-*
```

src/shared/extensions/ssh.ts
```diff
@@ -143,3 +149,132 @@ export async function startVscodeRemote(
 
     await new ProcessClass(vscPath, ['--folder-uri', workspaceUri]).run()
 }
+
+export abstract class VscodeRemoteSshConfig {
```
justinmk3, line 153: This code is for "vanilla ssh" so we should probably move it somewhere like `shared/ssh.ts`

```suggestion
export abstract class SshConfig {
```

PR #3658, opened 2023-07-14
feat(ec2): add commands to start, stop, and reboot instances. 

Hweinstock:
## Problem
Direct Feature Request. 
## Solution
- Add three new commands: `aws.ec2.startInstance`, `aws.ec2.stopInstance`, and `aws.ec2.rebootInstance`. 
- QuickPick only lists valid commands for action. Ex. doesn't show stopped containers for `aws.ec2.stopinstance`. 
- Commands will throw error if status already achieved. (Ex: stopping an already stopped instance). Can only be achieved on right-click of node. 
- Accessible from both command palette and the explorer.  
- loading bar w/ cancel for all operations. 
image
image

Notes:
- This PR also involves a refactoring of the `prompter` code into its own object. This is done to promote flexibility and testability (the prompter now has tests as well). 
- All the operations are done in new `InstanceStatusManager` class to minimize duplicate implementations. 
- Currently no way to check the status of an instance before starting/stopping. Plan to implement this in future PR. 
- Does not provide feedback on successful start/stop/reboot only throws error if something went wrong. 
- We don't throw error if user runs reboot on stopped instance, we just start it. 

## License


package.json
```diff
@@ -1120,6 +1120,18 @@
                     "command": "aws.ec2.openTerminal",
                     "when": "aws.isDevMode"
                 },
+                {
+                    "command": "aws.ec2.startInstance",
+                    "whem": "aws.isDevMode"
```
justinmk3, line 1125: typo (also other cases below). Not sure this is necessary though, unless these commands are "global"

```suggestion
                    "when": "aws.isDevMode"
```

package.nls.json
```diff
@@ -110,6 +110,9 @@
     "AWS.command.cdk.help": "View CDK Documentation",
     "AWS.command.ec2.openTerminal": "Open terminal to EC2 instance...",
     "AWS.command.ec2.openRemoteConnection": "Connect to EC2 instance in New Window...",
+    "AWS.command.ec2.startInstance": "Start EC2 Instance...",
+    "AWS.command.ec2.stopInstance": "Stop EC2 Instance...",
+    "AWS.command.ec2.rebootInstance": "Reboot EC2 Instance...",
```
justinmk3, line 115: "..." indicates that more steps are involved, is that the case for all of these? (A confirm dialog wouldn't count as "more steps".)

Hweinstock: They don't have more steps, so I went ahead and removed them.

package.json
justinmk3: still present in other cases

src/ec2/instanceStateManager.ts
```diff
@@ -0,0 +1,89 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Ec2Client } from '../shared/clients/ec2Client'
+import { ToolkitError, isAwsError } from '../shared/errors'
+import { showMessageWithCancel } from '../shared/utilities/messages'
+import { Timeout } from '../shared/utilities/timeoutUtils'
+import { Ec2Selection } from './utils'
+
+export class InstanceStateManager {
```
justinmk3, line 12: ins "Manager" communicating anything useful in these class names?

```suggestion
export class InstanceState {
```

Hweinstock: Instance state made me think of the physical state itself, whereas manager makes me think of "managing" that state. Do you think its clearer to leave out the Manager piece? Is another suffix that you see as more appropriate?

src/ec2/instanceStateManager.ts
```diff
@@ -0,0 +1,89 @@
...
+    private readonly client: Ec2Client
+
+    public constructor(protected readonly instanceId: string, protected readonly regionCode: string) {
+        this.client = this.getEc2Client()
+    }
```
justinmk3, line 17: This class doesn't really hold any non-trivial state. Can its functions live on `src/shared/clients/ec2Client.ts` ?

The EC2 service is starting to accumulate a lot of classes. As a developer if I want to use the Toolkit's EC2 layer, it's nice to be able to reach for `ec2Client` and just look for functions in a flat list. In contrast, having to hunt around for this or that class is an extra step (or multiple steps if there are many classes).

New classes are useful for non-trivial topics. But when things are trivial, that's a gift, let's keep them as trivial as possible.

src/ec2/prompter.ts
```diff
@@ -4,43 +4,62 @@
...
+import { AsyncCollection } from '../shared/utilities/asyncCollection'
 
-function asQuickpickItem(instance: Ec2Instance): DataQuickPickItem<string> {
-    return {
-        label: '$(terminal) \t' + (instance.name ?? '(no name)'),
-        detail: instance.InstanceId,
-        data: instance.InstanceId,
+export type instanceFilter = (instance: Ec2Instance) => boolean
+
+export class Ec2Prompter {
```
justinmk3, line 16: should this `extends Prompter` like most other FooPrompter classes?

    git grep -P 'class.*Prompter'

src/ec2/utils.ts
```diff
@@ -3,15 +3,14 @@
...
     instanceId: string
     region: string
 }
 
-export async function getInstancesFromRegion(regionCode: string): Promise<AsyncCollection<Ec2Instance>> {
-    const client = new Ec2Client(regionCode)
-    return await client.getInstances()
+export async function isEc2SelectionRunning(selection: Ec2Selection): Promise<boolean> {
+    const client = new Ec2Client(selection.region)
+    return await client.isInstanceRunning(selection.instanceId)
```
justinmk3, line 15: why is `isEc2SelectionRunning` needed (or util.ts)?

src/shared/clients/ec2Client.ts
```diff
@@ -30,8 +32,27 @@ export class Ec2Client {
...
-        const instances = this.getInstancesFromReservations(collection)
-        return instances
+        const extractedInstances = this.getInstancesFromReservations(collection)
+        const instancesWithStatuses = await this.addStatusesToInstances(extractedInstances)
+        const instancesWithNames = await this.addNamesToInstances(instancesWithStatuses)
+
+        return instancesWithNames
+    }
+
+    protected async addStatusesToInstances(
```
justinmk3, line 42: -  `addStatusesToInstances` and `addNamesToInstances` are separate functions, but could they be combined? is there a case where we would want to "update statues" but not "update names"? Combining them helps discoverability and readability.
- "add" as the verb confused me because I expected to see something being appended/inserted/concat'd, but actually we're updating-in-place.

```suggestion
    protected async updateStatuses(
```

src/shared/clients/ec2Client.ts
```diff
@@ -30,8 +32,27 @@ export class Ec2Client {
...
+        instances: AsyncCollection<EC2.Instance>
+    ): Promise<AsyncCollection<EC2.Instance>> {
+        return instances.map(async instance => {
+            return { ...instance, status: await this.getInstanceStatus(instance.InstanceId!) }
+        })
+    }
+
+    protected async addNamesToInstances(
```
justinmk3, line 50:
```suggestion
    /** Updates names in-place for displaying to humans. */
    protected async updateNames(
```

src/shared/extensionUtilities.ts
```diff
@@ -410,48 +411,58 @@ export class ExtensionUserActivity implements vscode.Disposable {
...
+        )
 
-        this.register(vscode.window.onDidChangeTextEditorVisibleRanges(e => {
-            if(e.textEditor.document.uri.scheme === 'output') {
-                // Ignore `output` scheme as text editors from output panel autoscroll
-                return
-            }
-            throttledEmit(vscode.window.onDidChangeTextEditorVisibleRanges)
-        }))
+        this.register(
```
justinmk3, line 447: https://github.com/aws/aws-toolkit-vscode/pull/3700 will fix the formatting issues

src/ec2/instanceStateManager.ts
Hweinstock: I was imagining the EC2 client as purely a wrapper of the SDK and thought I should separate the business logic from the SDK logic. I was trying to avoid overloading the EC2 client in general as well. However, I definitely see what you mean about making this harder to find. Going to work on integrating this logic into the EC2Client itself.

src/test/ec2/prompter.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        public instances: Ec2Instance[] = []
+
+        public testAsQuickPickItem(testInstance: Ec2Instance) {
+            return Ec2Prompter.asQuickPickItem(testInstance)
+        }
+
+        public testGetSelectionFromResponse(response: RegionSubmenuResponse<string>): Ec2Selection {
+            return Ec2Prompter.getSelectionFromResponse(response)
+        }
+        public async testGetInstancesAsQuickPickItems(region: string): Promise<DataQuickPickItem<string>[]> {
```
justinmk3, line 25: Can get rich testing of wizard behavior using `wizardTestUtils`. 

See `src/test/cloudWatchLogs/commands/searchLogGroup.test.ts`, here's a commit that migrated the tests to use `wizardTestUtils`:

https://github.com/aws/aws-toolkit-vscode/commit/8eed0da9990b0b699faa50f9125a71d8854ddca8

src/test/shared/clients/defaultEc2Client.test.ts
justinmk3, line 1: rename file to `src/test/shared/clients/ec2Client.test.ts`. The "defaultFooClient" stuff is old

src/ec2/instanceStateManager.ts
justinmk3: > I was imagining the EC2 client as purely a wrapper of the SDK

If it's pure then we don't need it :) We need a Toolkit-layer wrapper around the SDK that does useful things, not a passthrough interface.

Hweinstock: Yeah that makes sense. This seems to be an example of trying to over-modularize, to the point where its difficult to find what you want to use. I just committed the refactor of moving to EC2Client.

src/ec2/prompter.ts
Hweinstock: This case might be unique. Since we're leveraging the `regionSubmenu`, we have already extended the prompter class there. This class serves as a way to organize the methods relating to the EC2 implementation of `regionSubmenu`.

src/ec2/utils.ts
Hweinstock: good catch. I think this is outdated. The `util.ts` was originally for avoiding some circular imports with the `Ec2Selection`, but I moved it to `prompter.ts` which should also solve that problem.

src/test/ec2/prompter.test.ts
Hweinstock: The relevant wizard testing lives in `src/test/shared/ui/prompters/regionSubmenu.test.ts`. This test file is aimed at testing the utility functions surrounding the prompter, since the prompter is just implementing the already tested component of `regionSubmenu`.

src/shared/clients/ec2Client.ts
Hweinstock: No use case that I can think of at least right now of wanting them to be separated. I definitely see how the `add` language is confusing here. I combined them into `updateInstancesDetail`, which does both the adding of status and name.

src/shared/clients/ec2Client.ts
```diff
@@ -30,8 +35,25 @@ export class Ec2Client {
...
+    ): Promise<AsyncCollection<EC2.Instance>> {
         return instances
+            .map(async instance => {
+                return { ...instance, status: await this.getInstanceStatus(instance.InstanceId!) }
+            })
+            .map(instance => {
+                return instanceHasName(instance!)
+                    ? { ...instance, name: lookupTagKey(instance!.Tags!, 'Name') }
+                    : instance!
+            })
```
justinmk3, line 56: Each map() call re-scans the entire collection. Can both fields be updated in the same map() callback?

justinmk3:
> - Does not show feedback on successful start/stop/reboot, only on failure. 
> - Does not throw error if user runs reboot on stopped instance, just start it. 

Both of those seem like a good choice. But for "positive confirmation" in both cases, we should:

- Log something to the AWS Toolkit "Output channel"
- Log to the log file
- Future: "passive" messages/status in Toolkit-owned statusbar item.

PR #3661, opened 2023-07-17
feat(ec2): status icons in QuickPick, Explorer

Hweinstock:
## Problem
While we have added functionality to start/stop/restart instances from explorer, we don't have anyway of knowing the existing state. 
## Solution
First, I want to point out that the icons used here are temporary and I don't believe they are the best fit. They are put there currently to demonstrate the functionality. 
image
The three different icons to the left of the label indicate the status of the instance with the 'X' meaning stopped, the check mark meaning 'running', and the wheel indicating that its 'pending' (starting up, booting down). 

image
We also added icons to run/stop/restart the instances, such that the icons are only visible for valid actions (i.e. we cannot stop an already stopped instance). Additionally, we only display the `openTerminal` command next to `running` instances.

### Notes on Explorer Updates:
- Currently the command is not accessible from the command palette, because we have no way to link the selection back to the node to update. One approach could be to refresh the entire explorer or find a way to expose the `ec2ParentNode` such that we can refresh just that node. 
- When executing a command, the node will visually update, but only once. Thus, starting a node will leave it stuck in pending status until another refresh from elsewhere allows it to update. This issue will be addressed in a subsequent PR. 
- Explorer updates are limited to the EC2ParentNode, so any actions here will not impact updates for other nodes. 

## License


.changes/1.82.0.json
```diff
@@ -0,0 +1,22 @@
+{
+	"date": "2023-07-26",
```
justinmk3: This file (and others) is already on master. Not sure what's going on here, maybe a github bug

Hweinstock: looks like it went away? Not sure if you did something or if git fixed it.

src/ec2/utils.ts
```diff
@@ -0,0 +1,22 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Ec2Instance } from '../shared/clients/ec2Client'
+
+export function getIconForInstanceStatus(instance: Ec2Instance) {
+    return `$(${getIconCodeForInstanceStatus(instance)})`
```
justinmk3, line 9: unnecessary indirection

src/ec2/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export function getIconCodeForInstanceStatus(instance: Ec2Instance) {
```
justinmk3, line 12: Premature namespacing. Unlikely to be other icons for EC2 nodes, so we don't need to over-explain in the function name.

```suggestion
export function getIconCode(instance: Ec2Instance) {
```

justinmk3 (approved):

justinmk3 (approved):

PR #3668, opened 2023-07-20
feat(codewhisperer): add crossfile js, ts jsx, tsx support

Will-ShaoHua:
* restructure crossfile/utg code path
* add ts js tsx jsx crossfile support

## License


src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -158,29 +181,19 @@ function splitFileToChunks(filePath: string, chunkSize: number): Chunk[] {
...
-                filesOpenedInEditor.push((tab.input as any).uri.path)
-            })
-        })
-    } catch (e) {
-        // Older versions of VSC do not have the tab API
-    }
-
-    return filesOpenedInEditor
-}
-
```
Will-ShaoHua, line 183: Moved to supplementalContextUtil.ts as it will be used by both crossfile and utg features 

https://github.com/aws/aws-toolkit-vscode/pull/3668/files#diff-752c5d36a246334f6da7cd1a91d445bd971aa71573a337a51c293e5cc1ac53a4R84

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -110,6 +112,27 @@ function getInputChunk(editor: vscode.TextEditor, chunkSize: number) {
...
+
+    if (languageId === 'java') {
+        return true
+    } else if (crossFileLanguageConfigs.includes(languageId) && userGroup === UserGroup.CrossFile) {
+        return true
+    } else {
+        return false
+    }
+}
+
```
Will-ShaoHua, line 135: support 
1. 100% java users
2. 50% ts, tsx, js, jsx users

src/codewhisperer/util/supplementalContext/supplementalContextUtil.ts
```diff
@@ -35,19 +35,18 @@ export async function fetchSupplementalContext(
     const timesBeforeFetching = performance.now()
     const dependencyGraph = DependencyGraphFactory.getDependencyGraph(editor.document.languageId)
 
-    if (dependencyGraph === undefined) {
-        // This is a general check for language support of CW.
-        // We perform feature level language filtering later.
-        return undefined
-    }
```
Will-ShaoHua, line 42: makes dependency graph as optional since we are not actually utilizing all the APIs there

src/codewhisperer/util/supplementalContext/utgUtils.ts
```diff
@@ -112,34 +110,44 @@ async function findSourceFileByContent(
...
+}
+
 async function findSourceFileByName(
     editor: vscode.TextEditor,
-    languageConfig: utgLanguageConfig,
-    dependencyGraph: DependencyGraph
+    languageConfig: utgLanguageConfig
 ): Promise<string | undefined> {
-    const uri = dependencyGraph.getRootFile(editor)
-    const projectPath = dependencyGraph.getProjectPath(uri)
```
Will-ShaoHua, line 142: here is the only dependencyGraph APIs we will be using. Thus simply extract them.

src/test/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -0,0 +1,168 @@
...
+
+        it('should be empty if userGroup is control', async function () {
+            if (!shouldRunTheTest) {
+                this.skip()
+            }
+
+            sinon.stub(userGroupSettings, 'userGroup').returns(UserGroup.Control)
+
+            // TODO: flaky, is there other approach to achive this? need this probably because it's an async test
+            while (userGroupSettings.userGroup !== UserGroup.Control) {}
```
Will-ShaoHua, line 82: @justinmk3 @JadenSimon, naive q, I need this waiting probably because it's an async test, is there better approach to deal with it? thanks!

src/test/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -0,0 +1,168 @@
...
+            if (!shouldRunTheTest) {
+                this.skip()
+            }
+
+            sinon.stub(userGroupSettings, 'userGroup').returns(UserGroup.CrossFile)
+
+            while (userGroupSettings.userGroup !== UserGroup.CrossFile) {}
+
+            // TODO: flaky, is there other approach to achive this? need this probably because it's an async test
+            console.log(`in crossfile group test, ${userGroupSettings.userGroup}`)
```
Will-ShaoHua, line 122: @justinmk3 @JadenSimon, naive q, I need this waiting probably because it's an async test, is there better approach to deal with it? thanks!

src/test/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -0,0 +1,168 @@
...
+        })
+
+        it('should be non empty if userGroup is crossfile', async function () {
+            if (!shouldRunTheTest) {
+                this.skip()
+            }
+
+            sinon.stub(userGroupSettings, 'userGroup').returns(UserGroup.CrossFile)
+
+            while (userGroupSettings.userGroup !== UserGroup.CrossFile) {}
```
Will-ShaoHua, line 119: @justinmk3 @JadenSimon, here as well

src/test/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+            const editor = await openATextEditorWithText('content-1', 'file-1.js', tempFolder, { preview: false })
+            await openATextEditorWithText('content-2', 'file-2.js', tempFolder, { preview: false })
+            await openATextEditorWithText('content-3', 'file-3.js', tempFolder, { preview: false })
+            await openATextEditorWithText('content-4', 'file-4.js', tempFolder, { preview: false })
+
+            await assertTabCount(4)
+
+            const actual = await crossFile.fetchSupplementalContextForSrc(editor, fakeCancellationToken)
+            console.log(`in crossfile test, ${CodeWhispererUserGroupSettings.instance.userGroup}`)
```
Will-ShaoHua, line 118: @justinmk3 @JadenSimon, do you know how to avoid the race condition here, I checked the value here is still `Control` which is set up at L68 (same file) https://github.com/aws/aws-toolkit-vscode/pull/3668/files#diff-04181568581b3ad5742017095727f66f2749885f6dab8c555d8d2cb0b565f0ceR68.

Will-ShaoHua: and also it's affecting other test code path

Will-ShaoHua: what's your recommendation for this kinda of code `codewhispererUserGroup` which is used by lots of other code path and I need to test all different values of it https://github.com/aws/aws-toolkit-vscode/blob/master/src/codewhisperer/util/userGroupUtil.ts#L14

Will-ShaoHua: like here I want to test the behavior of `fetchSupplementalContextForSrc` under `Control` and `CrossFile` group

src/codewhisperer/util/supplementalContext/codeParsingUtil.ts
```diff
@@ -68,29 +67,21 @@ export function countSubstringMatches(arr1: string[], arr2: string[]): number {
...
-        // as we depend on Regex for this validation.
-        return false
-    }
+export async function isTestFile(
+    filePath: string,
+    languageConfig: { languageId: string; dependencyGraph?: DependencyGraph; fileContent?: string }
+): Promise<boolean> {
+    const byPath = filePath.includes(`tests/`) || filePath.includes('test/') || filePath.includes('tst/')
+    const byName = isTestFileByName(filePath, languageConfig.languageId)
+    const byDependencyGraph =
```
leigaol, line 76: If `await languageConfig.dependencyGraph.isTestFile(languageConfig.fileContent)` is slow, you can consider 

```
if ( byPath || by Name) return true.
return byDependencyGraph
```

Consider rename byPath --> filePathContainsTest, byName --> fileNameFitsTestPattern, 
byDependencyGraph --> 

To make this code more readable

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -5,16 +5,27 @@
...
-import { CodeWhispererSupplementalContextItem } from './supplementalContextUtil'
+import { CodeWhispererSupplementalContextItem, getOpenFilesInWindow } from './supplementalContextUtil'
 import { CodeWhispererUserGroupSettings } from '../userGroupUtil'
+import { isTestFile } from './codeParsingUtil'
+import * as CodeWhispererConstants from '../../models/constants'
+
+// TODO: ugly, can we make it prettier?
+// TODO: Move to another config file or constants file
+// Supported language to its dialects
+const supportedLanguageToDialects: Record<string, Set<string>> = {
```
leigaol, line 21: const supportedLanguageToDialects = {
         CodeWhispererConstants.java: new Set<string>(CodeWhispererConstants.java)

}

You don't have to use Record

leigaol: Better implement a singleton class for cross file context.

class CrossFileLanguageUtil {
       const supportedLanguageToDialects: Record<string, Set<string>> = {}

      isSupportedLanguage(): boolean {
      }
      
      
}

Use this class to control 
1. which language is supported for all
2. which dialect is supported
3. which language  is supported only for user group

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -24,20 +35,21 @@ interface Chunk {
 
 export async function fetchSupplementalContextForSrc(
     editor: vscode.TextEditor,
-    dependencyGraph: DependencyGraph,
```
Will-ShaoHua, line 27: `fetchSupplementalContextForSrc` is not using dependencyGraph, thus removed

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
leigaol: In general, aim for a high cohesion code

src/test/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+        it('should be empty if userGroup is control', async function () {
+            if (!shouldRunTheTest()) {
+                this.skip()
+            }
+
+            await globals.context.globalState.update(userGroupKey, {
+                group: UserGroup.Control,
+                version: extensionVersion,
+            })
```
Will-ShaoHua, line 71: @zixlin7

src/test/codewhisperer/util/crossFileContextUtil.test.ts
Will-ShaoHua: @zixlin7

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
Will-ShaoHua: yea agree, I prefer restructuring the whole into more OOP style, but it might take time

Will-ShaoHua: will leave it as is now @leigaol

src/codewhisperer/util/supplementalContext/codeParsingUtil.ts
Will-ShaoHua: updated

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
Will-ShaoHua: I aim to abstract the whole crossfile & utg into a single `supplementalContextProvider` or do it seperately `crossfileContextProvider` `utgContextProvider`

src/codewhisperer/models/constants.ts
```diff
@@ -76,6 +76,10 @@ export const javascript = 'javascript'
 
 export const typescript = 'typescript'
 
+export const jsx = 'javascriptreact'
+
+export const tsx = 'typescriptreact'
+
```
Will-ShaoHua, line 82: not being used, can remove tho

src/codewhisperer/util/supplementalContext/codeParsingUtil.ts
```diff
@@ -68,29 +67,26 @@ export function countSubstringMatches(arr1: string[], arr2: string[]): number {
...
-    }
-
-    return false
+    // This run slowly thus lazily execute
+    const fileHasTestDependency =
+        languageConfig.dependencyGraph && languageConfig.fileContent
+            ? await languageConfig.dependencyGraph.isTestFile(languageConfig.fileContent)
+            : false
+    return fileHasTestDependency
 }
```
Will-ShaoHua, line 87: make dependencyGraph an optional graph since 
1. it's only used for 1 case which is unit test generator but not crossfile.
2. only a few simple APIs of it are used, and these APIs can be replaced with simple VSCode APIs.

Will-ShaoHua: instead of making fake and hollow Typescript dependencyGraph object, making this parameter an optional is easier to maintain.

src/codewhisperer/util/supplementalContext/codeParsingUtil.ts
```diff
@@ -68,29 +67,26 @@ export function countSubstringMatches(arr1: string[], arr2: string[]): number {
...
-export async function isTestFile(editor: vscode.TextEditor, dependencyGraph: DependencyGraph): Promise<boolean> {
-    const languageConfig = utgLanguageConfigs[editor.document.languageId]
-    if (!languageConfig) {
-        // We have enabled the support only for python and Java for this check
-        // as we depend on Regex for this validation.
-        return false
-    }
+export async function isTestFile(
+    filePath: string,
+    languageConfig: { languageId: string; dependencyGraph?: DependencyGraph; fileContent?: string }
```
nkomonen-amazon, line 72: NIT: This makes it easier to know where the value should come from

```suggestion
    languageConfig: { languageId: vscode.TextDocument['languageId']; dependencyGraph?: DependencyGraph; fileContent?: string }
```

src/codewhisperer/util/supplementalContext/codeParsingUtil.ts
```diff
@@ -68,29 +67,26 @@ export function countSubstringMatches(arr1: string[], arr2: string[]): number {
...
+    // This run slowly thus lazily execute
+    const fileHasTestDependency =
+        languageConfig.dependencyGraph && languageConfig.fileContent
+            ? await languageConfig.dependencyGraph.isTestFile(languageConfig.fileContent)
+            : false
+    return fileHasTestDependency
 }
 
-export function isTestFileByName(filePath: string, language: string): boolean {
+function isTestFileByName(filePath: string, language: string): boolean {
```
nkomonen-amazon, line 89: NIT:
```suggestion
function isTestFileByName(filePath: string, language: vscode.TextEditor['languageId']): boolean {
```

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -5,16 +5,40 @@
...
+    | 'python'
+    | 'javascript'
+    | 'typescript'
+    | 'javascriptreact'
+    | 'typescriptreact'
+
+// TODO: ugly, can we make it prettier? like we have to manually type 'java', 'javascriptreact' which is error prone
+// TODO: Move to another config file or constants file
+// Supported language to its corresponding file ext
+const supportedLanguageToDialects: Record<CrossFileSupportedLanguage, Set<string>> = {
```
nkomonen-amazon, line 29:
```suggestion
const supportedLanguageToDialects: Readonly<Record<CrossFileSupportedLanguage, Set<string>>> = {
```

src/test/shared/utilities/editorUtilities.test.ts
```diff
@@ -0,0 +1,74 @@
...
+        before(async function () {
+            this.timeout(60000)
+        })
+
+        beforeEach(async function () {
+            tempFolder = (await createTestWorkspaceFolder()).uri.fsPath
+        })
+
+        afterEach(async function () {
+            await closeAllEditors()
```
nkomonen-amazon, line 31: I'd move this to the `beforeEach` to ensure there are no open editors before each test runs, a completely separate test may have opened an editor and the first test would get the wrong result.

src/shared/utilities/editorUtilities.ts
```diff
@@ -15,3 +16,35 @@ export function getTabSizeSetting(): number {
...
+export async function getOpenFilesInWindow(
+    filterPredicate?: (filePath: string) => Promise<boolean>
+): Promise<string[]> {
+    const filesOpenedInEditor: string[] = []
+
+    try {
+        const tabArrays = vscode.window.tabGroups.all
+        tabArrays.forEach(tabArray => {
+            tabArray.tabs.forEach(tab => {
+                filesOpenedInEditor.push((tab.input as any).uri.fsPath)
```
justinmk3, line 29: Is there a reason not to use `vscode.window.visibleTextEditors` ?
https://github.com/aws/aws-toolkit-vscode/blob/bfd220fb09749ef345fb94c58fabb668ae9fa5b3/src/shared/utilities/vsCodeUtils.ts#L28

src/test/codewhisperer/util/codeParsingUtil.test.ts
```diff
@@ -63,47 +61,51 @@ describe('RegexValidationForJava', () => {
     })
 })
 
-describe('isTestFileByNameValidation', () => {
-    it('should return true if the file name matches the test filename pattern - Java', () => {
+describe('isTestFileValidation', () => {
```
justinmk3, line 64: there is no symbol named `isTestFileValidation`. should this be 
```suggestion
describe('isTestFileByName', () => {
```

src/test/shared/utilities/editorUtilities.test.ts
```diff
@@ -0,0 +1,74 @@
...
+import * as vscode from 'vscode'
+import * as assert from 'assert'
+import * as path from 'path'
+import * as semver from 'semver'
+import { closeAllEditors, assertTabCount, createTestWorkspaceFolder, openATextEditorWithText } from '../../testUtil'
+import { getOpenFilesInWindow } from '../../../shared/utilities/editorUtilities'
+
+// VSCode tab APIs are available since 1.68.0
+function shouldRunTheTest(): boolean {
+    return (semver.valid(vscode.version) && semver.gte(vscode.version, '1.68.0')) as boolean
```
justinmk3, line 15: before this line, we can check `getMinVscodeVersion()` and throw if it is >= 1.68. That will let us know we can remove `shouldRunTheTest()`.

https://github.com/aws/aws-toolkit-vscode/blob/bfd220fb09749ef345fb94c58fabb668ae9fa5b3/src/shared/vscode/env.ts#L95

src/test/testUtil.ts
```diff
@@ -240,6 +240,47 @@ export async function assertTextEditorContains(contents: string): Promise<void |
     }
 }
 
+/**
+ * Create and open an editor with provided fileText, fileName and options. If folder is not provided,
+ * will create a temp worksapce folder which will be automatically deleted in testing environment
+ */
+export async function openATextEditorWithText(
```
justinmk3, line 247: Now that this is lifted, we can remove the old one: https://github.com/aws/aws-toolkit-vscode/blob/bfd220fb09749ef345fb94c58fabb668ae9fa5b3/src/testInteg/stepFunctions/visualizeStateMachine.test.ts#L74

src/shared/utilities/editorUtilities.ts
Will-ShaoHua: yea simply `vscode.window.visibleTextEditors` won't work. It's not returning all opened files.

leigaol: // This will provide all open editors (like split view with mulitple windows)
    // with one active file from each editor.
    
    I saw this comment in the existing code.  tabs not on the top are not visible, not in the returned files.

justinmk3: Does the docstring for this function mention that? That kind of context is very helpful.

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
leigaol: ACK. By when do you need this merged? I can approve once blocking comments are resolved.

src/test/shared/utilities/editorUtilities.test.ts
Will-ShaoHua: @nkomonen-amazon @justinmk3 should we keep this cleanup `closeAllEditors` both on `beforeEach` and `afterEach`?

Will-ShaoHua: asking because I saw it's used in `after` in `visualizeStateMachine.test.ts`
https://github.com/aws/aws-toolkit-vscode/blob/bfd220fb09749ef345fb94c58fabb668ae9fa5b3/src/testInteg/stepFunctions/visualizeStateMachine.test.ts#L116

Will-ShaoHua: not sure if it will cause unclean states if it's removed from `afterEach`

src/test/shared/utilities/editorUtilities.test.ts
Will-ShaoHua: @justinmk3 sorry didn't follow this one, did you mean remove this func and repalce it with 
```
semver.gte(getMinVsCodeVersion(), '1.68.0')
```
?

src/codewhisperer/util/supplementalContext/codeParsingUtil.ts
Will-ShaoHua: yea sg, will update

src/test/codewhisperer/util/codeParsingUtil.test.ts
Will-ShaoHua: yea renamed it to `isTestFile` sorry didn't cache that

src/shared/utilities/editorUtilities.ts
Will-ShaoHua: @justinmk3 not sure if it helps https://github.com/microsoft/vscode/issues/8886#issuecomment-259158438 (not a docstring in the official API)

Will-ShaoHua: I can provide a screenshot as well if you want

Will-ShaoHua: basically `vscode.window.visibleTextEditors` will only provide the tab which **_"users are able to see in the IDE"_**

Will-ShaoHua: so if the user is not on a split view, there will be only 1 visibleTextEditor. If they are on split view, the amount of visibleTextEditor will be no more than # of split users have

Will-ShaoHua: that's why we need tabGroup API here., which allow us to access all "opened" files

src/test/shared/utilities/editorUtilities.test.ts
justinmk3: > should we keep this cleanup `closeAllEditors` both on `beforeEach` and `afterEach`?

sure

nkomonen-amazon: I would say in general we should clean the state in the `before` to make tests less fragile from external changes. And doing it in the `after` is a courtesy for other tests. But each test in general shouldn't rely on a clean state when first starting.

I'd put `closeAllEditors()` in the `beforeEach` for your tests, and `after` as the courtesy for subsequent tests. If tests are failing after this change then we'll need to fix those tests.

src/shared/utilities/editorUtilities.ts
```diff
@@ -15,3 +16,35 @@ export function getTabSizeSetting(): number {
 export function getInlineSuggestEnabled(): boolean {
     return Settings.instance.getSection('editor').get('inlineSuggest.enabled', true)
 }
+
+export async function getOpenFilesInWindow(
```
justinmk3, line 20:
```suggestion
// Unlike vscode.window.visibleTextEditors, this function is needed because ...
// 
// See also: https://github.com/microsoft/vscode/issues/8886#issuecomment-259158438
export async function getOpenFilesInWindow(
```

src/test/shared/utilities/editorUtilities.test.ts
justinmk3: No. I mean just add a check like:

```
function shouldRunTheTest(): boolean {
  if (semver.gte(getMinVsCodeVersion(), '1.68.0')) {
    throw Error('this function is no longer needed, remove it')
  }
  return (semver.valid(vscode.version) && semver.gte(vscode.version, '1.68.0'))
```

src/test/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -0,0 +1,143 @@
...
+import { CodeWhispererUserGroupSettings } from '../../../codewhisperer/util/userGroupUtil'
+import { UserGroup } from '../../../codewhisperer/models/constants'
+import { assertTabCount, closeAllEditors, createTestWorkspaceFolder, openATextEditorWithText } from '../../testUtil'
+
+const userGroupSettings = CodeWhispererUserGroupSettings.instance
+let tempFolder: string
+
+// VSCode tab APIs are available since 1.68.0
+function shouldRunTheTest(): boolean {
+    return (semver.valid(vscode.version) && semver.gte(vscode.version, '1.68.0')) as boolean
```
justinmk3, line 21: `as boolean` is suspicious. It should never been needed, and it's useless. If you need to force a boolean, then actually force the value to a boolean using `!!`.

```suggestion
    return !!(semver.valid(vscode.version) && semver.gte(vscode.version, '1.68.0'))
```

Will-ShaoHua: ack, updated

src/test/shared/utilities/editorUtilities.test.ts
Will-ShaoHua: updated

src/test/shared/utilities/editorUtilities.test.ts
Will-ShaoHua: gotcha, yea added on both sites

src/shared/utilities/editorUtilities.ts
```diff
@@ -15,3 +16,41 @@ export function getTabSizeSetting(): number {
 export function getInlineSuggestEnabled(): boolean {
     return Settings.instance.getSection('editor').get('inlineSuggest.enabled', true)
 }
+
+/**
+ * Unlike vscode.window.visibleTextEditors, which will only return the "visible" files in the IDE, this function will return all files opened in the IDE.
```
justinmk3, line 21: Why doesn't `vscode.workspace.textDocuments` work for this use-case?

justinmk3 (approved):

src/shared/utilities/editorUtilities.ts
Will-ShaoHua: let me verify it, I remember it doesn't work but kinda forget hot it failed

Will-ShaoHua: @justinmk3 , `vscode.workspace.textDocuments` kinda return similar result to `vscode.workspace.visibleTextEditors`
image

Will-ShaoHua: in summary, either `vscode.workspace.textDocuments` or `vscode.worksapce.visibleTextEditors` will not return all opened files like we have with tabGroups API

leigaol (approved):

justinmk3:
Note for future reference: ` closeAllEditors()` is not needed in beforeEach, it is already done in the global beforeEach:

https://github.com/aws/aws-toolkit-vscode/blob/fa75a9df8b18d75e362f997ead2faa5596f01857/src/test/globalSetup.test.ts#L88

.changes/next-release/Feature-5baff9c0-54b5-4ed3-8a3e-a74800cc5af5.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "CodeWhisperer: Improve Java suggestion quality with enhanced file context fetching"
```
justinmk3, line 3: The PR title mentions js, ts jsx, tsx. Do we need a changelog for those too?

zixlin7 (approved):

.changes/next-release/Feature-5baff9c0-54b5-4ed3-8a3e-a74800cc5af5.json
Will-ShaoHua: yea that make sense... let me discuss this with the team Justin

PR #3669, opened 2023-07-20
feat(codewhisperer): factor in filedistance when fetching crossfile context

Will-ShaoHua:
### Before
- Crossfile file candidates are fetched by opened tabs without any sorting, e.g. files will be returned based on theirs orders in the IDE tabs (from the left to the right)

### After
- Crossfile file candidates fetched will be sorted by their file distance ascendingly. Files with smaller file distance will be prioritized than others.

## License

### DIFF
--- a/src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
+++ b/src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
@@ -13,6 +13,7 @@ import { CancellationError } from '../../../shared/utilities/timeoutUtils'
 import { CodeWhispererSupplementalContextItem } from './supplementalContextUtil'
 import { CodeWhispererUserGroupSettings } from '../userGroupUtil'
 import { isTestFile } from './codeParsingUtil'
+import { getFileDistance } from '../../../shared/filesystemUtilities'
 import { getOpenFilesInWindow } from '../../../shared/utilities/editorUtilities'
 
 type CrossFileSupportedLanguage =
@@ -201,10 +202,10 @@ function splitFileToChunks(filePath: string, chunkSize: number): Chunk[] {
 }
 
 /**
- * This function will return relevant cross files for the given editor file
+ * This function will return relevant cross files sorted by file distance for the given editor file
  * by referencing open files, imported files and same package files.
  */
-async function getCrossFileCandidates(editor: vscode.TextEditor): Promise<string[]> {
+export async function getCrossFileCandidates(editor: vscode.TextEditor): Promise<string[]> {
     const targetFile = editor.document.uri.fsPath
     const language = editor.document.languageId as CrossFileSupportedLanguage
     const dialects = supportedLanguageToDialects[language]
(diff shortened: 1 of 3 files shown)


src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -215,14 +216,28 @@ async function getCrossFileCandidates(editor: vscode.TextEditor): Promise<string
...
+                file: candidate,
+                fileDistance: getFileDistance(targetFile, candidate),
+            }
+        })
+        .sort((file1, file2) => {
+            return file1.fileDistance - file2.fileDistance
+        })
+        .map(fileToDistance => {
+            return fileToDistance.file
+        })
```
zixlin7, line 240: nit: the mappings seem to be unnecessary? we can do `sort((file1, file2) => { return  getFileDistance(targetFile, file1) - getFileDistance(targetFile, file2) }`?

src/test/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -49,6 +57,71 @@ describe('crossFileContextUtil', function () {
...
+    describe('getCrossFileCandidate', function () {
+        before(async function () {
+            this.timeout(60000)
+        })
+
+        beforeEach(async function () {
+            tempFolder = (await createTestWorkspaceFolder()).uri.fsPath
+        })
+
+        afterEach(async function () {
```
zixlin7, line 69: Justin mentioned this is not needed

justinmk3 (approved):

src/test/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -49,6 +57,71 @@ describe('crossFileContextUtil', function () {
         })
     })
 
+    describe('getCrossFileCandidate', function () {
+        before(async function () {
+            this.timeout(60000)
```
justinmk3, line 62: Why does this test need so much time? We should avoid assigning giant timeouts for tests unless they really are doing a lot of work.

src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
justinmk3: Would calling `getFileDistance` in sort() would incur redundant calculations?

zixlin7: I see, map makes sense if `getFileDistance` does heavy calculations

src/test/codewhisperer/util/crossFileContextUtil.test.ts
Will-ShaoHua: gotcha, let me remove it, i referred to `visualizeStateMachine.test.ts` and thought this timeout is needed for tests manipulating editos or something lol

src/test/codewhisperer/util/crossFileContextUtil.test.ts
Will-ShaoHua: wait, I thought this is needed, and I should add it in `beforeEach()` too as per the discussion here with Nick https://github.com/aws/aws-toolkit-vscode/pull/3668#discussion_r1270805968 @zixlin7 @justinmk3

zixlin7: Oh beforeEach is the one that has this done globally, it doesn't hurt to keep it in afterEach then

zixlin7 (approved):

PR #3671, opened 2023-07-21
feat(ec2): remoteConnect functionality with automatic key setup, ssh config, and connection. 

Hweinstock:
## Problem
New feature as in title. 
## Solution
There are three key pieces of this PR:
- setting up ssh keys. 
- configuring ssh config file. 
- establishing connection. 

### Setting up SSH Keys 
After many attempts, this step is deemed as essential with no current workarounds known. We do the following: 
- generate ssh key pair in global storage (don't overwrite if already exists). 
- send the public key to the target instance before attempting connection. 
- eventually, we want to also delete the key after the connection has been made, however this step is not part of this PR. 

### Configuring SSH Config file 
Similar to CodeCatalyst, we leverage a proxy command configured in this script to establish an SSH connection. The currently generated config section looks like the following: 
```
# Created by AWS Toolkit for VSCode. https://github.com/aws/aws-toolkit-vscode
Host ec2-*
    ForwardAgent yes
    AddKeysToAgent yes
    StrictHostKeyChecking accept-new
    ProxyCommand '/Users/hkobew/Library/Application Support/Code/User/globalStorage/amazonwebservices.aws-toolkit-vscode/ec2_connect' '%h'
    IdentityFile '/Users/hkobew/Library/Application Support/Code/User/globalStorage/amazonwebservices.aws-toolkit-vscode/aws-ec2-key'
    User '%r'
```
All fields are the same as CodeCatalyst with two exceptions:
- `IdentityFile` points toward the key we generated in global storage. 
- `User` is the remote username, which is OS dependent because this field tells the remote machine where to look for the public key. Currently Amazon Linux and Ubuntu instances are supported. 

### Establishing Connection 
We use the same process and code from CodeCatalyst. Much of this code has been moved to the `src/shared/sshConfig.ts` file. Similar to CodeCatalyst, we leverage a connect script to give the added benefit of logging. 

### Additional Notes
Sending the keys to the remote instance requires determining the OS of the remote instance. Each OS requires a different command to send the keys. With this PR, both amazon linux and ubuntu have been implemented and tested. 

Additionally, the current functionality is only tested on MacOS client, but tests appear to pass in Windows CI. 

## License


src/ec2/sshKeyPair.ts
```diff
@@ -0,0 +1,39 @@
...
+            await SshKeyPair.generateSshKeyPair(keyPath)
+        }
+        return new SshKeyPair(keyPath)
+    }
+
+    public static async generateSshKeyPair(keyPath: string): Promise<void> {
+        const process = new ChildProcess(`ssh-keygen`, ['-t', 'rsa', '-N', '', '-q', '-f', keyPath])
+        const result = await process.run()
+        if (result.exitCode !== 0) {
+            throw new ToolkitError('ec2: Failed to generate ssh key')
```
JadenSimon, line 27: Would it make sense to attach stdout to the error, assuming the error message is in stdout? Like this:
```suggestion
            throw new ToolkitError('ec2: Failed to generate ssh key', { details: { stdout: result.stdout } })
```

src/ec2/sshKeyPair.ts
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    public getPublicKeyPath(): string {
+        return this.publicKeyPath
+    }
+
+    public async getPublicKey(): Promise<string> {
+        const contents = await fs.readFile(this.publicKeyPath, 'utf-8')
+        return contents
```
JadenSimon, line 37: Are callers expected to handle any exceptions if the public key is deleted while the extension is running? Also, use `SystemUtilities` for fs calls, it'll automatically display permissions issues.

src/shared/clients/ssmClient.ts
```diff
@@ -51,10 +59,44 @@ export class SsmClient {
             .flatten()
             .flatten()
             .promise()
-
         return response[0]!
     }
 
+    public async getTargetPlatformName(target: string): Promise<string> {
+        const instanceInformation = await this.describeInstance(target)
+        return instanceInformation.PlatformName!
```
JadenSimon, line 67: style nit (does `prettier` have a setting for this??): 
```suggestion
        const instanceInformation = await this.describeInstance(target)

        return instanceInformation.PlatformName!
```

IMO it's nice to have return statements separated from other statements.

src/shared/extensions/ssh.ts
```diff
@@ -123,9 +123,10 @@ export async function startVscodeRemote(
     ProcessClass: typeof ChildProcess,
     hostname: string,
     targetDirectory: string,
-    vscPath: string
+    vscPath: string,
+    user?: string
 ): Promise<void> {
-    const workspaceUri = `vscode-remote://ssh-remote+${hostname}${targetDirectory}`
+    const workspaceUri = `vscode-remote://ssh-remote+${`${user}@` ?? ''}${hostname}${targetDirectory}`
```
JadenSimon, line 129: Does this cause any issues with CodeCatalyst? Our e2e tests wouldn't cover this.

If you haven't tested this yet I can give it a try.

src/ec2/sshKeyPair.ts
Hweinstock: What exactly does attaching stdout do? If it is attached, will it not log the error to console?

src/shared/extensions/ssh.ts
Hweinstock: I tested it, it doesn't seem to cause any issues with CodeCatalyst. I was still able to open dev environments in the sam way as before. Is there any other key functionality I should be testing here?

src/ec2/sshKeyPair.ts
JadenSimon: It should still get logged. Attaching it establishes a clear relationship between the message and the output whereas right now the two things aren't explicitly connected.

src/ec2/sshKeyPair.ts
JadenSimon: The caller doesn't necessarily need to handle the exception. We just want to make sure that if this ever does happen, it should be obvious what went wrong. That might be as simple as re-throwing the error with extra information.

src/shared/extensions/ssh.ts
JadenSimon: > Is there any other key functionality I should be testing here?

No, as long as you can connect in a new window then it should be good.

src/ec2/sshKeyPair.ts
Hweinstock: Should this be something we do on each error we throw? Or what is it about this error that encourages us to explicitly link them?

resources/ec2_connect
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
```
justinmk3, line 1: Not a blocker for this PR, but will also need a powershell script for Windows clients.

src/codecatalyst/model.ts
```diff
@@ -30,8 +29,12 @@ import { fileExists } from '../shared/filesystemUtilities'
 import { CodeCatalystAuthenticationProvider } from './auth'
 import { ToolkitError } from '../shared/errors'
 import { Result } from '../shared/utilities/result'
+import { VscodeRemoteConnection, ensureDependencies } from '../shared/remoteSession'
+import { VscodeRemoteSshConfig, sshLogFileLocation } from '../shared/vscodeRemoteSshConfig'
 
 export type DevEnvironmentId = Pick<DevEnvironment, 'id' | 'org' | 'project'>
+export const hostNamePrefix = 'aws-devenv-'
+export const connectScriptPrefix = 'codecatalyst_connect'
```
justinmk3, line 37: No need to define this module-level. Should be a literal arg to `sshLogFileLocation`

src/ec2/model.ts
```diff
@@ -145,11 +167,117 @@ export class Ec2ConnectionManager {
...
+
+        await this.ssmClient.sendCommandAndWait(selection.instanceId, documentName, {
+            commands: [command],
+        })
+    }
+
+    protected async getRemoteUser(instanceId: string) {
+        const osName = await this.ssmClient.getTargetPlatformName(instanceId)
+        if (osName === 'Amazon Linux') {
+            return 'ec2-user'
```
justinmk3, line 260: Does EC2/SSM have an API that could give this info?

src/ec2/model.ts
```diff
@@ -145,11 +167,117 @@ export class Ec2ConnectionManager {
...
+    }
+
+    public async sendSshKeyToInstance(
+        selection: Ec2Selection,
+        sshKeyPair: SshKeyPair,
+        remoteUser: string
+    ): Promise<void> {
+        const sshKey = await sshKeyPair.getPublicKey()
+
+        const remoteAuthorizedKeysPaths = `/home/${remoteUser}/.ssh/authorized_keys`
```
justinmk3, line 248: Should probably first do a `sendCommandAndWait` call that checks that this file exists and the user can write to it. So we can show a useful message if that fails.

src/ec2/model.ts
```diff
@@ -145,11 +167,117 @@ export class Ec2ConnectionManager {
...
+        await this.sendSshKeyToInstance(selection, keyPair, remoteUser)
+        return keyPath
+    }
+
+    public async sendSshKeyToInstance(
+        selection: Ec2Selection,
+        sshKeyPair: SshKeyPair,
+        remoteUser: string
+    ): Promise<void> {
+        const sshKey = await sshKeyPair.getPublicKey()
```
justinmk3, line 246:
```suggestion
        const sshPubKey = await sshKeyPair.getPublicKey()
```

src/ec2/model.ts
```diff
@@ -145,11 +167,117 @@ export class Ec2ConnectionManager {
...
+
+        const remoteAuthorizedKeysPaths = `/home/${remoteUser}/.ssh/authorized_keys`
+        const command = `echo "${sshKey}" > ${remoteAuthorizedKeysPaths}`
```
justinmk3, line 249: more robust script: https://github.com/elpy1/ssh-over-ssm/blob/e439729a80dd65990bd3bf7b96f4a180dc93c7e3/ssh-ssm.sh#L29-L37

src/shared/extensions/ssh.ts
justinmk3, line 129: The `??` in `${user}@ ?? ''` never happens because of the "@". The string is getting complicated so better treat the user can separately

```suggestion
    const userAt = user ? `${user}@` : ''
    const workspaceUri = `vscode-remote://ssh-remote+${userAt}${hostname}${targetDirectory}`
```

src/shared/vscodeRemoteSshConfig.ts
```diff
@@ -0,0 +1,229 @@
...
+import { getIdeProperties } from './extensionUtilities'
+import { showConfirmationMessage } from './utilities/messages'
+import { CancellationError } from './utilities/timeoutUtils'
+import { getSshConfigPath } from './extensions/ssh'
+import globals from './extensionGlobals'
+import { fileExists, readFileAsString } from './filesystemUtilities'
+
+const localize = nls.loadMessageBundle()
+
+export class VscodeRemoteSshConfig {
```
justinmk3, line 23: Not really vscode-specific, just normal ssh stuff

```suggestion
export class SshConfig {
```

src/test/shared/vscodeRemoteSshConfig.test.ts
justinmk3, line 1: rename to `src/test/shared/sshConfig.test.ts`

src/codecatalyst/model.ts
Hweinstock: I copied this format from the existing CodeCatalyst code, but just refactored it to minimize the number of module-level constants. All module level constants were removed with the exception of `codecatalyst_connect` script since it is imported in the test files.

src/shared/vscodeRemoteSshConfig.ts
justinmk3, line 1: rename to `sshConfig.ts`

src/ec2/model.ts
justinmk3: This can wait until future. Created a backlog issue.

src/ec2/model.ts
Hweinstock: I don't believe so. There are a few ways of getting this information, but this method from SSM is what I found to be the most reliable. The defaults are listed in a table here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/TroubleshootingInstancesConnecting.html

justinmk3:
Found no regressions in CC functionality, test scenarios:

- missing ssh
- empty sshconfig
- missing session-manager-plugin
- connect to existing CC dev-env
- create new CC dev-env

PR #3677, opened 2023-07-24
fix(ec2): openTerminal throws error on Cloud9 Instances

Hweinstock:
## Problem
Steps to reproduce:
- create a cloud9 ec2 instance
- try to use "open terminal" feature on the instance
- command failed error :(
### Important Questions
- Why did the toolkit try to use the `AWSCloud9SSMInstanceProfile` role if it's not valid?
    - In order to check if an EC2 instance has the proper permissions, we use the EC2 SDK to get the role name associated with the instance. Then, we use the IAM SDK to list the policies associated with that role. Therefore, the toolkit got the `AWSCloud9SSMInstanceProfile` role from the EC2 SDK, but then the IAM threw an error saying the role does not exist. Currently unsure why the role is invalid. 
- or is it valid, but we need to pass the full ARN to whatever API is throwing the error?
    - The docs: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/IAM.html#listAttachedRolePolicies-property, only accept the `friendly name` which is the role name and not the ARN. Once we know why the role doesn't exist (despite being attached to an EC2 instance), that will likely lead us to how we can request more info on it via SDK. 
- or do we need to find a different role and ignore this one?
    - Only one IAM role can be attached to each EC2 instance. Since this is the role we get from the EC2 SDK, there is no other role we can use. 
- or can we tell the user something more meaningful about what the problem actually is, instead of just "NoSuchEntity ... the role cannot be found"?
We can throw a specific error to the user about what happened. Rather than throwing the general lacking permission error, we can mention that there was a failure in retrieving the policies attached to a given instance. Additionally, we can log the ARN of the role we receive from the EC2 SDK. This is the solution implemented in this PR. 

Note that this does have an important consequence. Because we check for policies before executing `startSession`, it is possible that the proper policies do exist on the instance, but we fail the pre-check. Thus, we don't connect despite having the proper polices. One way around this could be to attempt connection despite a missing role, however this presents another challenge of what error message to throw if that connection fails. That is, if we don't find a valid role on the instance, and the attempted connection failed, can we assume a permission error or should a more general error be thrown? Additionally, this approach would require checking for the errors twice, which seems unnecessary. A better solution can likely be determined once we know why the role is invalid. 

New Error:
image

## License

### DIFF
--- a/src/ec2/model.ts
+++ b/src/ec2/model.ts
@@ -8,7 +8,7 @@ import { IAM } from 'aws-sdk'
 import { Ec2Selection } from './utils'
 import { getOrInstallCli } from '../shared/utilities/cliUtils'
 import { isCloud9 } from '../shared/extensionUtilities'
-import { ToolkitError } from '../shared/errors'
+import { ToolkitError, isAwsError } from '../shared/errors'
 import { SsmClient } from '../shared/clients/ssmClient'
 import { Ec2Client } from '../shared/clients/ec2Client'
 
@@ -51,9 +51,13 @@ export class Ec2ConnectionManager {
             const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole.Arn)
             return attachedPolicies
         } catch (e) {
-            const errorMessage = `No policies attached to role: ${IamRole.Arn}.`
-            getLogger().error(`ec2: ${errorMessage}`)
-            throw ToolkitError.chain(e, errorMessage, { code: 'NoSuchEntity' })
+            if (isAwsError(e)) {
+                console.log(e)
+                const errorMessage = `No policies attached to role: ${IamRole.Arn}.`
+                getLogger().error(`ec2: ${errorMessage}`)
+                throw ToolkitError.chain(e, errorMessage, { code: e.code })
+            }
+            throw e
         }
     }
 
@@ -70,9 +74,8 @@ export class Ec2ConnectionManager {
                 )
                 return false
             }
-            throw new ToolkitError(`An unknown error occurred when checking the policies for ${instanceId}`, {
-                cause: e as Error,
-                code: 'PolicyCheckError',
+            throw ToolkitError.chain(e as Error, `Failed to check policies for EC2 instance: ${instanceId}`, {
+                code: 'PolicyCheck',
             })
         }
     }
--- a/src/ec2/model.ts
+++ b/src/ec2/model.ts
@@ -41,14 +41,17 @@ export class Ec2ConnectionManager {
         return new DefaultIamClient(this.regionCode)
     }
 
-    protected async getAttachedPolicies(instanceId: string): Promise<IAM.attachedPoliciesListType> {
+    protected async getAttachedPolicies(instanceId: string): Promise<IAM.AttachedPolicy[]> {
         const IamRole = await this.ec2Client.getAttachedIamRole(instanceId)
         if (!IamRole) {
             return []
         }
-        const iamResponse = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
-
-        return []
+        try {
+            const attachedRoles = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
+            return attachedRoles
+        } catch (e) {
+            return []
+        }
     }
 
     public async hasProperPolicies(instanceId: string): Promise<boolean> {
--- a/src/ec2/model.ts
+++ b/src/ec2/model.ts
@@ -51,16 +51,24 @@ export class Ec2ConnectionManager {
             const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
             return attachedPolicies
         } catch (e) {
-            getLogger().error(`ec2: failed to find policies attached to role ${IamRole.Arn}.`)
-            return []
+            const errorMessage = `Failed to find policies attached to role with ARN ${IamRole.Arn}.`
+            getLogger().error(`ec2: ${errorMessage}`)
+            throw new ToolkitError(errorMessage)
         }
     }
 
     public async hasProperPolicies(instanceId: string): Promise<boolean> {
-        const attachedPolicies = (await this.getAttachedPolicies(instanceId)).map(policy => policy.PolicyName!)
-        const requiredPolicies = ['AmazonSSMManagedInstanceCore', 'AmazonSSMManagedEC2InstanceDefaultPolicy']
+        try {
+            const attachedPolicies = (await this.getAttachedPolicies(instanceId)).map(policy => policy.PolicyName!)
+            const requiredPolicies = ['AmazonSSMManagedInstanceCore', 'AmazonSSMManagedEC2InstanceDefaultPolicy']
 
-        return requiredPolicies.length !== 0 && requiredPolicies.every(policy => attachedPolicies.includes(policy))
+            return requiredPolicies.length !== 0 && requiredPolicies.every(policy => attachedPolicies.includes(policy))
+        } catch (e) {
+            getLogger().warn(
+                `ec2: due to error in checking policies attached to instance, assuming necessary policies do not exist for instance ${instanceId}.`
+            )
+            return false
+        }
     }
 
     public async isInstanceRunning(instanceId: string): Promise<boolean> {
--- a/src/ec2/model.ts
+++ b/src/ec2/model.ts
@@ -52,7 +52,7 @@ export class Ec2ConnectionManager {
             return attachedPolicies
         } catch (e) {
             if (isAwsError(e) && e.code == 'NoSuchEntity') {
-                const errorMessage = `No policies attached to role: ${IamRole.Arn}.`
+                const errorMessage = `Attached role does not exist in IAM: ${IamRole.Arn}.`
                 getLogger().error(`ec2: ${errorMessage}`)
                 throw ToolkitError.chain(e, errorMessage, { code: e.code })
             }
--- a/src/ec2/model.ts
+++ b/src/ec2/model.ts
@@ -17,6 +17,7 @@ export type Ec2ConnectErrorCode = 'EC2SSMStatus' | 'EC2SSMPermission' | 'EC2SSMC
 import { openRemoteTerminal } from '../shared/remoteSession'
 import { DefaultIamClient } from '../shared/clients/iamClient'
 import { ErrorInformation } from '../shared/errors'
+import { getLogger } from '../shared/logger'
 
 export class Ec2ConnectionManager {
     private ssmClient: SsmClient
@@ -47,9 +48,10 @@ export class Ec2ConnectionManager {
             return []
         }
         try {
-            const attachedRoles = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
-            return attachedRoles
+            const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
+            return attachedPolicies
         } catch (e) {
+            getLogger().error(`ec2: failed to find policies attached to role ${IamRole.Arn}.`)
             return []
         }
     }
--- a/src/ec2/model.ts
+++ b/src/ec2/model.ts
@@ -48,12 +48,12 @@ export class Ec2ConnectionManager {
             return []
         }
         try {
-            const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
+            const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole.Arn)
             return attachedPolicies
         } catch (e) {
-            const errorMessage = `Failed to find policies attached to role with ARN ${IamRole.Arn}.`
+            const errorMessage = `No policies attached to role: ${IamRole.Arn}.`
             getLogger().error(`ec2: ${errorMessage}`)
-            throw new ToolkitError(errorMessage, { code: 'NoSuchEntity' })
+            throw ToolkitError.chain(e, errorMessage, { code: 'NoSuchEntity' })
         }
     }
 
--- a/src/ec2/model.ts
+++ b/src/ec2/model.ts
@@ -51,7 +51,7 @@ export class Ec2ConnectionManager {
             const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole.Arn)
             return attachedPolicies
         } catch (e) {
-            if (isAwsError(e)) {
+            if (isAwsError(e) && e.code == 'NoSuchEntity') {
                 const errorMessage = `No policies attached to role: ${IamRole.Arn}.`
                 getLogger().error(`ec2: ${errorMessage}`)
                 throw ToolkitError.chain(e, errorMessage, { code: e.code })
--- a/src/ec2/model.ts
+++ b/src/ec2/model.ts
@@ -52,7 +52,6 @@ export class Ec2ConnectionManager {
             return attachedPolicies
         } catch (e) {
             if (isAwsError(e)) {
-                console.log(e)
                 const errorMessage = `No policies attached to role: ${IamRole.Arn}.`
                 getLogger().error(`ec2: ${errorMessage}`)
                 throw ToolkitError.chain(e, errorMessage, { code: e.code })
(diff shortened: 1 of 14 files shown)


src/ec2/model.ts
```diff
@@ -41,14 +41,17 @@ export class Ec2ConnectionManager {
         return new DefaultIamClient(this.regionCode)
     }
 
-    protected async getAttachedPolicies(instanceId: string): Promise<IAM.attachedPoliciesListType> {
+    protected async getAttachedPolicies(instanceId: string): Promise<IAM.AttachedPolicy[]> {
         const IamRole = await this.ec2Client.getAttachedIamRole(instanceId)
         if (!IamRole) {
```
justinmk3, line 46:
```suggestion
        if (!IamRole?.Arn) {
```

this would avoid the need for `IamRole!.Arn!` below

src/ec2/model.ts
```diff
@@ -41,14 +41,17 @@ export class Ec2ConnectionManager {
...
             return []
         }
-        const iamResponse = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
-
-        return iamResponse.AttachedPolicies ?? []
+        try {
+            const attachedRoles = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
+            return attachedRoles
+        } catch (e) {
+            return []
```
justinmk3, line 53: does iamClient already log the failure? if not, we should log something here

src/shared/clients/iamClient.ts
```diff
@@ -77,11 +77,20 @@ export class DefaultIamClient {
...
-        const response = await client.listAttachedRolePolicies({ RoleName: roleName }).promise()
 
-        return response
+        const requester = async (request: IAM.ListAttachedRolePoliciesRequest) =>
+            client.listAttachedRolePolicies(request).promise()
+
+        const collection = pageableToCollection(requester, { RoleName: roleName }, 'Marker', 'AttachedPolicies')
+            .flatten()
+            .filter(p => p !== undefined)
+            .map(p => p!)
```
justinmk3, line 90: Tradeoff is that this defeats any callers that would want to do paging themselves. But if we don't expect `listAttachedRolePolicies` to ever have many pages, that may be acceptable in this case.

src/test/ec2/model.test.ts
```diff
@@ -189,4 +195,28 @@ describe('Ec2ConnectClient', function () {
...
     })
+
+    describe('getAttachedPolicies', async function () {
+        let client: MockEc2ConnectClient
+
+        before(async function () {
+            client = new MockEc2ConnectClient()
+        })
+
+        it('returns empty list when IamRole does not exist', async function () {
```
justinmk3, line 206:
```suggestion
        it('returns empty when IamRole not found', async function () {
```

justinmk3:
> Currently if no role is associated with the EC2 instances, we treat the instance as having no attached policies. I believe the same treatment for roles that can not be found is fitting. 

"roles that can not be found" means "_attached role_ that is not found by IAM"? 

We can at least say in the user-facing message: "role X was attached to instance but not found in IAM".

src/ec2/model.ts
```diff
@@ -41,14 +41,17 @@ export class Ec2ConnectionManager {
...
+    protected async getAttachedPolicies(instanceId: string): Promise<IAM.AttachedPolicy[]> {
         const IamRole = await this.ec2Client.getAttachedIamRole(instanceId)
         if (!IamRole) {
             return []
         }
-        const iamResponse = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
-
-        return iamResponse.AttachedPolicies ?? []
+        try {
+            const attachedRoles = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
```
justinmk3, line 50: In the docstring for `getAttachedPolicies` we could document that "error Foo is raised if IAM could not find the role". Then the caller can handle that error. This would allow us to show a better error message to the user.

Alternatively we could throw a ToolkitError. That is probably more idiomatic and also will be automatically handled even if the caller doesn't handle it. It also serves as a kind of documentation.

src/ec2/model.ts
Hweinstock: Makes sense, I changed it to log the ARN here.

src/shared/clients/iamClient.ts
Hweinstock: Yea, I would think the policies attached to an IAM role wouldn't require many pages since each policy can list as many permissions as it wants.

src/test/ec2/model.test.ts
```diff
@@ -189,4 +195,28 @@ describe('Ec2ConnectClient', function () {
...
+            sinon.stub(Ec2Client.prototype, 'getAttachedIamRole').resolves(undefined)
+            const response = await client.testGetAttachedPolicies('test-instance')
+            assert.deepStrictEqual(response, [])
+
+            sinon.restore()
+        })
+
+        it('returns empty list when IamRole returned does not exist', async function () {
```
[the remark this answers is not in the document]
Hweinstock: Yea, that describes what I was going for more. The case I am focused on here is specifically the cloud9 one. That is, we are given a role from the EC2 SDK that doesn't register as valid in the IAM SDK.

src/ec2/model.ts
Hweinstock: changed it to throw a ToolkitError so that it bubbles the error up a level. Now, it logs what happened, but still assumes necessary policies don't exist.

src/ec2/model.ts
```diff
@@ -51,16 +51,24 @@ export class Ec2ConnectionManager {
             const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
             return attachedPolicies
         } catch (e) {
-            getLogger().error(`ec2: failed to find policies attached to role ${IamRole.Arn}.`)
-            return []
+            const errorMessage = `Failed to find policies attached to role with ARN ${IamRole.Arn}.`
```
justinmk3, line 54:
```suggestion
            const errorMessage = `No policies attached to role with ARN: ${IamRole.Arn}.`
```

src/ec2/model.ts
```diff
@@ -41,21 +42,33 @@ export class Ec2ConnectionManager {
...
         const IamRole = await this.ec2Client.getAttachedIamRole(instanceId)
-        if (!IamRole) {
+        if (!IamRole?.Arn) {
             return []
         }
-        const iamResponse = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
-
-        return iamResponse.AttachedPolicies ?? []
+        try {
+            const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
```
justinmk3, line 51: does the if-check above avoid this?

```suggestion
            const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole.Arn)
```

src/ec2/model.ts
```diff
@@ -41,21 +42,33 @@ export class Ec2ConnectionManager {
...
+            return attachedPolicies
+        } catch (e) {
+            const errorMessage = `Failed to find policies attached to role with ARN ${IamRole.Arn}.`
+            getLogger().error(`ec2: ${errorMessage}`)
+            throw new ToolkitError(errorMessage)
```
justinmk3, line 56:
```suggestion
            throw new ToolkitError.chain(e, errorMessage)
```

src/ec2/model.ts
```diff
@@ -41,21 +42,33 @@ export class Ec2ConnectionManager {
...
+        if (!IamRole?.Arn) {
             return []
         }
-        const iamResponse = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
-
-        return iamResponse.AttachedPolicies ?? []
+        try {
+            const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
+            return attachedPolicies
+        } catch (e) {
```
justinmk3, line 53: does the relevant error set a `code` we can check?

```
if (e.code = 'NoSuchEntity') {

}
```

src/ec2/model.ts
```diff
@@ -41,21 +42,33 @@ export class Ec2ConnectionManager {
...
-        const attachedPolicies = (await this.getAttachedPolicies(instanceId)).map(policy => policy.PolicyName!)
-        const requiredPolicies = ['AmazonSSMManagedInstanceCore', 'AmazonSSMManagedEC2InstanceDefaultPolicy']
+        try {
+            const attachedPolicies = (await this.getAttachedPolicies(instanceId)).map(policy => policy.PolicyName!)
+            const requiredPolicies = ['AmazonSSMManagedInstanceCore', 'AmazonSSMManagedEC2InstanceDefaultPolicy']
 
-        return requiredPolicies.length !== 0 && requiredPolicies.every(policy => attachedPolicies.includes(policy))
+            return requiredPolicies.length !== 0 && requiredPolicies.every(policy => attachedPolicies.includes(policy))
+        } catch (e) {
+            getLogger().warn(
```
justinmk3, line 67:
```suggestion
            if (e instance of ToolkitError && e.code == 'NoSuchEntity') {
                getLogger().warn(
```

src/ec2/model.ts
Hweinstock: changed it so that it does.

src/ec2/model.ts
```diff
@@ -41,21 +42,39 @@ export class Ec2ConnectionManager {
...
-        const iamResponse = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
-
-        return iamResponse.AttachedPolicies ?? []
+        try {
+            const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole.Arn)
+            return attachedPolicies
+        } catch (e) {
+            const errorMessage = `No policies attached to role: ${IamRole.Arn}.`
+            getLogger().error(`ec2: ${errorMessage}`)
+            throw ToolkitError.chain(e, errorMessage, { code: 'NoSuchEntity' })
```
justinmk3, line 56: But we should also check the code of the `e` in this scope. We can't assume that is the only error returned by `listAttachedRolePolicies`. What is `e.code` when `listAttachedRolePolicies` fails for the case concerning this PR?

src/ec2/model.ts
```diff
@@ -41,21 +42,39 @@ export class Ec2ConnectionManager {
...
+            return requiredPolicies.length !== 0 && requiredPolicies.every(policy => attachedPolicies.includes(policy))
+        } catch (e) {
+            if (e instanceof ToolkitError && e.code == 'NoSuchEntity') {
+                getLogger().warn(
+                    `ec2: due to error in checking policies attached to instance, assuming necessary policies do not exist for instance ${instanceId}.`
+                )
+                return false
+            }
+            throw new ToolkitError(`An unknown error occurred when checking the policies for ${instanceId}`, {
+                cause: e as Error,
```
justinmk3, line 74: - I think `ToolkitError.chain()` sets `cause`.
- the `code` never needs to have "Error" in its name

    throw new ToolkitError.chain(e, `Failed to check policies for EC2 instance: ${instanceId}`, { code: 'PolicyCheck' })

src/ec2/model.ts
```diff
@@ -41,21 +42,39 @@ export class Ec2ConnectionManager {
...
+            const attachedPolicies = (await this.getAttachedPolicies(instanceId)).map(policy => policy.PolicyName!)
+            const requiredPolicies = ['AmazonSSMManagedInstanceCore', 'AmazonSSMManagedEC2InstanceDefaultPolicy']
+
+            return requiredPolicies.length !== 0 && requiredPolicies.every(policy => attachedPolicies.includes(policy))
+        } catch (e) {
+            if (e instanceof ToolkitError && e.code == 'NoSuchEntity') {
+                getLogger().warn(
+                    `ec2: due to error in checking policies attached to instance, assuming necessary policies do not exist for instance ${instanceId}.`
+                )
+                return false
```
justinmk3, line 71: It looks like we are still silencing the error. We _want_ to show the user an error like:

    Role "foo" attached to EC2 instance has no policies attached.

Are we doing that?

src/ec2/model.ts
Hweinstock: that code is `NoSuchEntity`. I changed it so that the ToolkitError inherits the error code from the AWSError.

src/ec2/model.ts
Hweinstock: Oh, yes I was silencing it to show the same general error of lacking permissions. But just changed it to throw the more specific error instead.

src/ec2/model.ts
```diff
@@ -41,21 +42,35 @@ export class Ec2ConnectionManager {
...
         }
-        const iamResponse = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
-
-        return iamResponse.AttachedPolicies ?? []
+        try {
+            const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole.Arn)
+            return attachedPolicies
+        } catch (e) {
+            if (isAwsError(e)) {
+                const errorMessage = `No policies attached to role: ${IamRole.Arn}.`
```
justinmk3, line 55: Can't assume this is the reason for the error unless we check the code.
```suggestion
                if (e.code == ...) {
                  const errorMessage = `No policies attached to role: ${IamRole.Arn}.`
```

justinmk3:
Currently for the "invalid" case I see this:

image

Ideally for the "invalid" case, it should show `No policies attached to role: ...AWSCloud9SSMInstanceProfile.` Because that's the more specific, useful, actionable message.

The logs show the full chain:

```
2023-07-25 14:32:48 [ERROR]: aws.ec2.openTerminal: Error: Failed to check policies for EC2 instance: i-07861ebf13e1e3f88 [PolicyCheck]
	 -> Error: No policies attached to role: arn:aws:iam::541201481031:instance-profile/cloud9/AWSCloud9SSMInstanceProfile. [NoSuchEntity]
	 -> NoSuchEntity: The role with name AWSCloud9SSMInstanceProfile cannot be found. (statusCode: 404; requestId: 3aed1d5e-fa7c-4915-89e3-62e4f9d74f15)
```

Wherever the `PolicyCheck` error is wrapping the `NoSuchEntity`, we can skip wrapping if `code == NoSuchEntity`.

justinmk3:
About this message:
image
It should mention the policy name. Every little thing we can do to save time helps, especially if we haven't yet implemented auto-configuration.

Hweinstock:
Current errors being thrown:
- `Invalid` case (ex. c9):
image

- If it fails to find proper policies, but does find a role: 
image

- If it doesn't even find a role: 
image

Is this the distinction you had in mind? Also, would it be better to only to display the "friendlyName" of the ARNs, or entire ARN? The "friendlyName" is the name at the end.

justinmk3:
Looks good! Ideally in the "doesn't even find a role" case, the message shouldn't mention `Ensure the IAM role attached...` 

> Also, would it be better to only to display the "friendlyName" of the ARNs

Friendly name is probably better but we can revisit that later.

src/ec2/model.ts
```diff
@@ -51,38 +51,54 @@ export class Ec2ConnectionManager {
...
-    private throwConnectionError(message: string, selection: Ec2Selection, errorInfo: ErrorInformation) {
+    protected throwConnectionError(message: string, selection: Ec2Selection, errorInfo: ErrorInformation) {
         const generalErrorMessage = `Unable to connect to target instance ${selection.instanceId} on region ${selection.region}. `
         throw new ToolkitError(generalErrorMessage + message, errorInfo)
     }
 
+    protected async throwPolicyError(selection: Ec2Selection) {
+        const role = await this.ec2Client.getAttachedIamRole(selection.instanceId)
+
+        const baseMessage = 'Ensure the IAM role attached to the instance has the required policies.'
```
justinmk3, line 85:
```suggestion
        const baseMessage = 'Ensure an IAM role is attached to the instance and has the required policies.'
```

justinmk3 (approved):

src/ec2/model.ts
```diff
@@ -41,14 +42,24 @@ export class Ec2ConnectionManager {
...
         }
-        const iamResponse = await this.iamClient.listAttachedRolePolicies(IamRole!.Arn!)
-
-        return iamResponse.AttachedPolicies ?? []
+        try {
+            const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole.Arn)
+            return attachedPolicies
+        } catch (e) {
+            if (isAwsError(e) && e.code == 'NoSuchEntity') {
+                const errorMessage = `No policies attached to role: ${IamRole.Arn}.`
```
justinmk3, line 55: Based on the full message, `NoSuchEntity` does not mean the role has no policies:

    2023-07-18 08:59:15 [ERROR]: aws.ec2.openTerminal: [NoSuchEntity: The role with name AWSCloud9SSMInstanceProfile cannot be found.

```suggestion
                const errorMessage = `Attache role does not exist in IAM: ${IamRole.Arn}.`
```

src/ec2/model.ts
```diff
@@ -41,14 +42,24 @@ export class Ec2ConnectionManager {
...
+                getLogger().error(`ec2: ${errorMessage}`)
+                throw ToolkitError.chain(e, errorMessage, { code: e.code })
```
justinmk3, line 57: This also needs the `documentationUri`, so the View Documentation button is shown

justinmk3 (approved):

PR #3680, opened 2023-07-25
feat: support only .mjs extension for node18+ lambdas

nkomonen-amazon:
By default when a nodejs lambda was created,
the default file would be .js, but now with
node 18 the default file is .mjs.

The extension could not find node18 users files since the extension was
in .mjs but we were expecting it to be .js

### Solution:

Only support .mjs files for node18+ lambdas

## License

### DIFF
--- a/src/lambda/models/samLambdaRuntime.ts
+++ b/src/lambda/models/samLambdaRuntime.ts
@@ -34,6 +34,20 @@ export const nodeJsRuntimes: ImmutableSet<Runtime> = ImmutableSet<Runtime>([
     'nodejs14.x',
     'nodejs12.x',
 ])
+export function getNodeMajorVersion(version?: string): number | undefined {
+    if (!version) {
+        return undefined
+    }
+
+    const match = version.match(/^nodejs(\d+)\./)
+
+    if (match) {
+        return Number(match[1])
+    } else {
+        return undefined
+    }
+}
+
 export const pythonRuntimes: ImmutableSet<Runtime> = ImmutableSet<Runtime>([
     'python3.10',
     'python3.9',
--- a/src/test/lambda/models/samLambdaRuntime.test.ts
+++ b/src/test/lambda/models/samLambdaRuntime.test.ts
@@ -13,6 +13,8 @@ import {
     RuntimeFamily,
     samImageLambdaRuntimes,
     samLambdaCreatableRuntimes,
+    getNodeMajorVersion,
+    nodeJsRuntimes,
 } from '../../../lambda/models/samLambdaRuntime'
 
 describe('compareSamLambdaRuntime', async function () {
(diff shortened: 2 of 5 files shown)


src/test/lambda/models/samLambdaRuntime.test.ts
```diff
@@ -123,3 +124,20 @@ describe('runtimes', function () {
...
+        const version = getNodeMajorVersion('nodejs12.x')
+        assert.strictEqual(version, 12)
+    })
+
+    it('should extract major version from nodejs18.x', () => {
+        const version = getNodeMajorVersion('nodejs18.x')
+        assert.strictEqual(version, 18)
+    })
+
+    it('should return undefined on invalid input', () => {
```
justinmk3, line 139: "should" is always redundant in test descriptions. 
```suggestion
    it('returns undefined on invalid input', () => {
```

src/test/lambda/models/samLambdaRuntime.test.ts
```diff
@@ -123,3 +124,18 @@ describe('runtimes', function () {
         ])
     })
 })
+
+describe('getNodeMajorVersion()', () => {
+    it(`extracts major version from 'nodejs{VERSION}.x'`, () => {
+        for (let i = 1; i < 1000; i++) {
+            const versionString = `nodejs${i}.x`
```
justinmk3, line 131: Sorry for the confusion. I was suggesting to use the `nodeJsRuntimes` list from `src/lambda/models/samLambdaRuntime.ts`. That way, as we add support for new runtimes, we have coverage for all new runtime names.

src/test/lambda/models/samLambdaRuntime.test.ts
```diff
@@ -123,3 +124,18 @@ describe('runtimes', function () {
...
+            const version = getNodeMajorVersion(versionString)
```
justinmk3, line 132: My suggestion about a "range" test was to check the `version` result.

    assert(version > 0 && version < 999)

src/test/lambda/models/samLambdaRuntime.test.ts
nkomonen-amazon: Ah I see now, updated

src/test/lambda/models/samLambdaRuntime.test.ts
nkomonen-amazon: Done!

src/lambda/models/samLambdaRuntime.ts
```diff
@@ -34,6 +34,20 @@ export const nodeJsRuntimes: ImmutableSet<Runtime> = ImmutableSet<Runtime>([
     'nodejs14.x',
     'nodejs12.x',
 ])
+export function getNodeMajorVersion(version?: string): number | undefined {
+    if (!version) {
+        return undefined
+    }
+
+    const match = version.match(/^nodejs(\d+)\./)
```
justinmk3, line 42: oh, we probably want this to be case insensitive. OTOH, the test will catch it, so I guess doesn't matter.
```suggestion
    const match = version.match(/^nodejs(\d+)\./i)
```

justinmk3 (approved):

PR #3681, opened 2023-07-25
ci: don't fail fast on macos tests

nkomonen-amazon:
Usually if one of the macos tests fails, we
will abort all other ones.

But, recently vs code insiders is failing exclusively and it aborts the other tests even though they should be fine. Due to this we cannot easily see that everything else is passing.

Solution:

Don't fail fast on mac os tests, let all of them
run and maybe we can tell if it is just a vscode
insiders problem more easily.

## License


justinmk3:
Good idea

PR #3684, opened 2023-07-25
fix(codewhisperer): use stored session id in aggregated user decision events

leigaol:
## Problem
The session id can be empty in the aggregated user decision events.

To reproduce it, you can invoke, and keep pressing Esc. 

When you press Esc frequently, multiple clear state function can be run in parallel, causing race conditions that set the session id to empty string.

## Solution

Use the pre-existing, stored, non-empty session id in the aggregated user decision events. 

This solution only works for aggregated user decision events, it does not solve the global state problem in this code base. 

## License

### DIFF
--- a/src/codewhisperer/util/telemetryHelper.ts
+++ b/src/codewhisperer/util/telemetryHelper.ts
@@ -219,7 +219,7 @@ export class TelemetryHelper {
             (this.lastRequestId && this.lastRequestId === requestId) ||
             (this.sessionDecisions.length && this.sessionDecisions.length === this.numberOfRequests)
         ) {
-            this.sendUserTriggerDecisionTelemetry(sessionId, supplementalContextMetadata)
+            this.sendUserTriggerDecisionTelemetry(supplementalContextMetadata)
         }
     }
 
@@ -260,7 +260,6 @@ export class TelemetryHelper {
     }
 
     private sendUserTriggerDecisionTelemetry(
-        sessionId: string,
         supplementalContextMetadata?: Omit<CodeWhispererSupplementalContext, 'supplementalContextItems'> | undefined
     ) {
         // the user trigger decision will aggregate information from request level user decisions within one session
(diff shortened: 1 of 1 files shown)


justinmk3 (approved):

zixlin7 (approved):

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -273,7 +272,7 @@ export class TelemetryHelper {
         const autoTriggerType = this.sessionDecisions[0].codewhispererAutomatedTriggerType
         const language = this.sessionDecisions[0].codewhispererLanguage
         const aggregated: CodewhispererUserTriggerDecision = {
-            codewhispererSessionId: sessionId,
+            codewhispererSessionId: this.sessionDecisions[0].codewhispererSessionId,
```
andrewyuq, line 275: can I know when exactly is this `sessionDecisions` being reset?

leigaol: It is reset in function `resetUserTriggerDecisionTelemetry`, which is the last line of function `sendUserTriggerDecisionTelemetry`

andrewyuq: ok, but if the passed `sessionId` is empty, this means the caller `recordUserDecisionTelemetry` will also have this empty sessionId and send it with the `userDecision` event/

leigaol: From my previous investigation, the pre-existing UserDecision telemetry has less than 1% empty `sessionId`. 

This change is not perfect, however, while being a small change, it can effectively reduce the percentage of empty session id in Aggregated User Decision events. According to my manual testing, without this change, if you press Esc at high frequency after invoking, there is a great chance of seeing one with empty session id. With this change, I have yet found a case when session id is not empty. 

Removing those global states will take much more work.

PR #3686, opened 2023-07-26
fix(ec2): use IAM Role instead of instance profile. 

Hweinstock:
## Problem
Currently, we retrieve the IAM Instance Profile associated with a target EC2 instance. This is not the same as the IAM Role attached to the EC2 instance. This issue is well hidden because the console (by-default) creates an instance profile for us when we attach a role to an EC2 instance. Specifically, 

> If you use the AWS Management Console to create a role for Amazon EC2, the console automatically creates an instance profile and gives it the same name as the role. When you then use the Amazon EC2 console to launch an instance with an IAM role, you can select a role to associate with the instance. In the console, the list that's displayed is actually a list of instance profile names. The console does not create an instance profile for a role that is not associated with Amazon EC2.
> 
More info: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html

Because the IAM instance profile and the IAM Role have the same name when the console creates it for us and the IAM SDK works with the "friendlyName" and not the entire ARN, their distinction is completely hidden for most cases. 

However, the distinction is revealed in the C9 case, where I suspect the instance profile and IAM role were manually created separately with distinct names. 

## Solution
We make an explicit distinction in the code between IAM Instance Profile and IAM Roles. Specifically, the EC2 SDK gives us the IAM Instance Profile, which we then must use to retrieve the IAM role via the IAM SDK. This extra step is needed to accommodate the case where the name of the two differ (as in C9). 

The effects of this are that the error message for c9 instances no longer gives an instance profile name instead of a role name: 
image

Additionally, we no longer need to worry about the case of explicitly handling non-existent roles from the EC2 SDK, since the underlying cause now appears to be trying to use the name of the instance profile in place of the name of the IAM role. 

## License

### DIFF
--- a/src/ec2/model.ts
+++ b/src/ec2/model.ts
@@ -17,7 +17,6 @@ export type Ec2ConnectErrorCode = 'EC2SSMStatus' | 'EC2SSMPermission' | 'EC2SSMC
 import { openRemoteTerminal } from '../shared/remoteSession'
 import { DefaultIamClient } from '../shared/clients/iamClient'
 import { ErrorInformation } from '../shared/errors'
-import { getLogger } from '../shared/logger'
 
 export class Ec2ConnectionManager {
     private ssmClient: SsmClient
@@ -50,35 +49,18 @@ export class Ec2ConnectionManager {
         return new DefaultIamClient(this.regionCode)
     }
 
-    public async getAttachedIamRole(instanceId: string): Promise<IAM.Role> {
+    public async getAttachedIamRole(instanceId: string): Promise<IAM.Role | undefined> {
         const IamInstanceProfile = await this.ec2Client.getAttachedIamInstanceProfile(instanceId)
         if (IamInstanceProfile && IamInstanceProfile.Arn) {
             const IamRole = await this.iamClient.getIAMRoleFromInstanceProfile(IamInstanceProfile.Arn)
             return IamRole
         }
-        throw new ToolkitError(`No IAM instance profile attached to instance ${instanceId}`, {
-            code: 'NoIamInstanceProfile',
-        })
     }
 
-    protected async getAttachedPolicies(instanceId: string): Promise<IAM.AttachedPolicy[]> {
-        try {
-            const IamRole = await this.getAttachedIamRole(instanceId)
-            const attachedPolicies = await this.iamClient.listAttachedRolePolicies(IamRole.Arn)
-            return attachedPolicies
-        } catch (e) {
-            if (e instanceof ToolkitError && e.code == 'NoIamInstanceProfile') {
-                getLogger().warn(
-                    `ec2: failed to find IAM Instance Profile associated with instance. Returning no policies attached for instance: ${instanceId}`
-                )
-                return []
-            }
-            throw e
-        }
-    }
-
-    public async hasProperPolicies(instanceId: string): Promise<boolean> {
-        const attachedPolicies = (await this.getAttachedPolicies(instanceId)).map(policy => policy.PolicyName!)
+    public async hasProperPolicies(IamRoleArn: string): Promise<boolean> {
+        const attachedPolicies = (await this.iamClient.listAttachedRolePolicies(IamRoleArn)).map(
+            policy => policy.PolicyName!
+        )
         const requiredPolicies = ['AmazonSSMManagedInstanceCore', 'AmazonSSMManagedEC2InstanceDefaultPolicy']
 
         return requiredPolicies.length !== 0 && requiredPolicies.every(policy => attachedPolicies.includes(policy))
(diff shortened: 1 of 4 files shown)


src/shared/clients/ec2Client.ts
```diff
@@ -100,11 +100,11 @@ export class Ec2Client {
     }
 
     /**
-     * Retrieve IAM role attached to given EC2 instance.
+     * Retrieve IAM Instance Profile attached to given EC2 instance.
```
justinmk3, line 103:
```suggestion
     * Gets the IAM Instance Profile (not role) attached to given EC2 instance.
```

src/shared/clients/iamClient.ts
```diff
@@ -93,4 +93,11 @@ export class DefaultIamClient {
 
         return policies
     }
+
+    public async getIAMRoleFromInstanceProfile(instanceProfileArn: string): Promise<IAM.Role> {
+        const client = await this.createSdkClient()
+        const instanceProfileName = this.getFriendlyName(instanceProfileArn)
+        const response = await client.getInstanceProfile({ InstanceProfileName: instanceProfileName }).promise()
+        return response.InstanceProfile.Roles[0]
```
justinmk3, line 101: Could this return undefined? If so, the return type seems wrong. Or maybe we should throw if `Roles.length == 0`

src/ec2/model.ts
```diff
@@ -86,16 +94,19 @@ export class Ec2ConnectionManager {
         throw new ToolkitError(generalErrorMessage + message, errorInfo)
     }
 
-    protected async throwPolicyError(selection: Ec2Selection) {
-        const role = await this.ec2Client.getAttachedIamRole(selection.instanceId)
+    private async getPolicyErrorText(selection: Ec2Selection): Promise<string> {
```
justinmk3, line 97: this method is only used in one place and doesn't seem worth the indirection

src/ec2/model.ts
```diff
@@ -86,16 +94,19 @@ export class Ec2ConnectionManager {
...
+        try {
+            const role = await this.getAttachedIamRole(selection.instanceId)
```
justinmk3, line 99: Making another (redundant) service call in logic that is trying to _build_ an error message, should be avoided. It's another place that can fail, while already handling a failure. It's also inefficient.

Can't `throwPolicyError` get enough info to make this decision without another service call?

src/shared/clients/iamClient.ts
Hweinstock: I was under the understanding that an instance profile is created to link a role to an instance, so it seems strange it would have no roles. However, to be safe I changed it to throw here.

src/ec2/model.ts
```diff
@@ -46,26 +50,30 @@ export class Ec2ConnectionManager {
...
-        if (!IamRole?.Arn) {
-            return []
+    public async getAttachedIamRole(instanceId: string): Promise<IAM.Role> {
+        const IamInstanceProfile = await this.ec2Client.getAttachedIamInstanceProfile(instanceId)
+        if (IamInstanceProfile && IamInstanceProfile.Arn) {
+            const IamRole = await this.iamClient.getIAMRoleFromInstanceProfile(IamInstanceProfile.Arn)
+            return IamRole
         }
+        throw new ToolkitError(`No IAM instance profile attached to instance ${instanceId}`, {
+            code: 'NoIamInstanceProfile',
```
justinmk3, line 60: If `getAttachedIamRole` returns `undefined`, the caller knows unambiguously that nothing was found. Don't need to throw here. 

A caller may choose to throw, however.

src/ec2/model.ts
Hweinstock: Yea, that makes a lot of sense. I think this is a symptom of the structure of the error handling. I restructured it so that not having a role is its own error. I think the new structure is a little cleaner. I had to refactor the tests to make it work with new structure.

src/ec2/model.ts
Hweinstock: Yea, the refactored flow makes use of this exactly.

src/ec2/model.ts
```diff
@@ -86,46 +76,55 @@ export class Ec2ConnectionManager {
...
-    }
-
-    public async checkForStartSessionError(selection: Ec2Selection): Promise<void> {
+    private async checkForInstanceStatusError(selection: Ec2Selection): Promise<void> {
         const isInstanceRunning = await this.isInstanceRunning(selection.instanceId)
-        const hasProperPolicies = await this.hasProperPolicies(selection.instanceId)
-        const isSsmAgentRunning = (await this.ssmClient.getInstanceAgentPingStatus(selection.instanceId)) == 'Online'
 
         if (!isInstanceRunning) {
             const message = 'Ensure the target instance is running and not currently starting, stopping, or stopped.'
```
justinmk3, line 83:
```suggestion
            const message = 'Ensure the target instance is running.'
```

justinmk3 (approved):

justinmk3:
The failing "stable" CI jobs are related to https://github.com/aws/aws-toolkit-vscode/pull/3690

Issue #2205, opened 2021-10-20
support vscode web


justinmk3:
Toolkit currently doesn't support "web" version of vscode https://vscode.dev/

![image](https://user-images.githubusercontent.com/55561878/138169490-5f6bf663-edde-452e-a0d1-1c2eb17c99c4.png)

PR #3688, opened 2023-07-27
browser: Initial structure setup for toolkit in browser

nkomonen-amazon:
- In the VS Code Launch panel we can run the extension
  in a browser context using `Run Web Extension`
- It will run the code, with `src/extensionWeb.ts` being the
  entrypoint
- Webpack is configured to bundle the browser code
- Our extension will be seen as supporting the browser since we have a `browser` entry in `package.json`

ref #2205

## License


justinmk3 (approved):
🚀

PR #3694, opened 2023-07-27
fix(ec2): check for actions instead of policies.

Hweinstock:
## Problem 
We currently check if certain policies are attached to the IAM role in order to determine if the necessary permissions exist on the instance to open an SSM session. However, this can be done more fine-grained. 

## Solution 
https://github.com/aws/aws-toolkit-vscode/blob/a9d64f3a448480f8fed9ba96b3da33631c298265/src/ecs/util.ts#L33-L52 
Leverage the approach done by ECS to check for specific actions rather than policies. 

- Restructure testing framework to look at actions instead of policies. 
- Refactor ECS code to pull out shared functionality to `src/shared/remoteSession.ts`.
- Refactor existing code in the `src/ec2/model.ts` file to utilize the approach demonstrated above. 
- Add existing testing coverage if necessary. 

## License

### DIFF
--- a/src/ec2/model.ts
+++ b/src/ec2/model.ts
@@ -77,19 +77,10 @@ export class Ec2ConnectionManager {
         }
     }
 
-    // public async hasProperPolicies(IamRoleArn: string): Promise<boolean> {
-    //     const attachedPolicies = (await this.iamClient.listAttachedRolePolicies(IamRoleArn)).map(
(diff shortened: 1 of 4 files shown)


JadenSimon (approved):

justinmk3 (approved):
Great choices!

PR #3697, opened 2023-07-28
feat(ec2): provide link to launch instance. 

Hweinstock:
## Problem
User Request
## Solution
Add option to launch instance from toolkit. The option links to console page for launching EC2 instances. Currently, the functionality is available as an icon-button and on right-click. Additionally, we have it in the command-palette. In the command palette, the region is inferred using `guessDefaultRegion`. I see three potential alternatives:
- (1)provide a region prompter when used from the command palette that defaults to the region given by `guessDefaultRegion`. 
- (2)Don't prompt for region, and assume the region based on `guessDefaultRegion`. 
- (3)Don't allow the command to be used from command palette and only see it on the explorer. 
Let me know if you think either of the alternatives presented above would be preferred. 

The current implementation (1) because it seems like the optimal middle ground. We allow the customer to select their region, while also providing an intuitive and helpful default. 

Below are some screenshots of the feature:
image
image
image

## License


justinmk3:
> We allow the customer to select their region, while also providing an intuitive and helpful default.

If the region prompt has the "default" (guessed) region at the top, then that's perfect.

packages/toolkit/.changes/next-release/Feature-ddc43dca-284c-4ab6-8857-e207bcdd9858.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "provide link to launch ec2 instances from explorer"
```
justinmk3, line 3: The changelog is a "mini blog" and is the single most high-traffic way that customers get news about the product. So when writing a changelog, I imagine I'm writing a tweet or a blog post :)

```suggestion
	"description": "EC2: Launch ec2 instances from AWS Explorer"
```

justinmk3 (approved):

PR #3699, opened 2023-07-31
browser: start webpack setup to get logging working

nkomonen-amazon:
## Problem
We have the browser entry point setup already, and now we want to start with getting something small from our
existing code running.

## Solution
Get the logging working. This PR doesn't get it working :(

This commit:
- Creates a new cli command to build the webpack bundle for the browser `npm run buildBrowser`
- polyfills node specific modules with browser compatible ones. This was automatically done in webpack 4 (we are on 5), but now it needs to be done explicitly

#### Remaining issues
- Certain code needs to be modified to handle being run in the browser
  - Winston
  - Telemetry
- `child_process` and `async_hooks` do not have a good browser supported alternative

## License


package.json
```diff
@@ -3567,22 +3568,33 @@
         "@iarna/toml": "^2.2.5",
         "adm-zip": "^0.5.10",
         "amazon-states-language-service": "^1.10.0",
+        "assert": "^2.0.0",
         "async-lock": "^1.4.0",
         "aws-sdk": "^2.1384.0",
         "aws-ssm-document-language-service": "^1.0.0",
+        "browserify-fs": "^1.0.0",
+        "browserify-zlib": "^0.2.0",
```
justinmk3, line 3576: It will be hard to reason about which of these interleaved dependencies is needed for which "mode". Can we create a sub-project in a new folder like `srcWeb/` which has its own package.json? 

Example of a project "inheriting" a parent project: https://github.com/aws/aws-toolkit-common/blob/147df8f44f08e081675e01cebf9f957ca9658add/lsp/core/aws-lsp-yaml-common/package.json#L10

justinmk3 (changes requested):
before merging this I think we need to creating a `srcWeb/package.json` which inherits the "base" package.json.

webpack.config.js
```diff
@@ -102,11 +102,43 @@ const webConfig = {
...
+            http: require.resolve('stream-http'),
+            https: require.resolve('https-browserify'),
+            os: require.resolve('os-browserify/browser'),
+            path: require.resolve('path-browserify'),
+            stream: require.resolve('stream-browserify'),
+            zlib: require.resolve('browserify-zlib'),
+            process: require.resolve('process/browser'),
+            // TODO: these do not have a straight-forward replacement and cause browser mode not to work
+            child_process: false, // Reason for error: 'TypeError: The "original" argument must be of type Function'
+            async_hooks: false,
```
JadenSimon, line 136: It's only used for telemetry, but yeah a replacement is very difficult to do without access to the engine's APIs. There's a partial solution in `src/shared/telemetry/spans.ts` that could be used. The behavior only diverges when two separate async tasks are trying to record telemetry to the context in parallel. CodeWhisperer doesn't do this though.

package.json
nkomonen-amazon: I've created a new `package.json` with the browser dependencies in `src/browser/`. This should be good enough for now to separate things, but when it comes to naming we can revisit later and come to a final agreement.

src/browser/README.md
```diff
@@ -0,0 +1,25 @@
...
+Functionality will not be immediately available in the browser since the underlying code will need
+to be modified to get things working. Functionality will be provided incrementally.
+
+## `package.json`
+
+We separate browser specific dependencies from the main `package.json`
+in the root to reduce complexity.
+
+These browser specific dependencies are consumed in the
+root `package.json` by using `"aws-toolkit-browser": "file:src/browser"`
```
justinmk3, line 16: Should it be the other way around, i.e. browser/package.json consumes a base package.json? 

Is this only a temporary step towards that?

nkomonen-amazon: This is the correct direction. The "base" package.json consumes the "browser" package.json since our base `package.json` is the only one that gets used to build the final VSIX.

justinmk3: That's the inverse of this approach, right? https://github.com/aws/aws-toolkit-common/blob/147df8f44f08e081675e01cebf9f957ca9658add/lsp/core/aws-lsp-yaml-common/package.json#L10

package.json
```diff
@@ -3541,6 +3543,7 @@
         "mocha-junit-reporter": "^2.2.0",
         "mocha-multi-reporters": "^1.5.1",
         "nyc": "^15.1.0",
+        "path-browserify": "^1.0.1",
```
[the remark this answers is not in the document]
nkomonen-amazon: Done

src/browser/README.md
nkomonen-amazon: Yup, those have a "core" package which is consumed by other modules. But in our case everything goes in to the root `package.json` which has a dependency on the nested `src/browser/package.json`

justinmk3: Ok, can you add a note about this specifically in this readme? And link to the above URL example

nkomonen-amazon: Updated

nkomonen-amazon: @justinmk3 updated to fix the package.json import issue

justinmk3 (approved):

justinmk3:
the failing `VSCode Package Test` CI looks relevant

PR #3705, opened 2023-08-04
feat(ec2): prompt to add inline policies. 

Hweinstock:
## Problem
By default, the ec2 instances do not have the proper permissions to allow for connect via SSM. Therefore, users will have to complete this step manually. 
## Current Solution
Check if IAM role exists on instance, if it does, prompt to add necessary actions as inline policy. 
### Details and Problems with Current Solution
- actions in prompt are poorly formatted. 
- attaching the inline policy, then immediately attempting connection does not appear to work reliably. The cause of this is currently unknown, but one potential explanation is the delay of propagating permissions across all regions. When immediately attempting connection after adding the inline policy, the terminal will open, but will immediately crash. Note that no error is thrown by the SDK here, meaning we have no way of knowing if the terminal will crash or not before opening the terminal. Additionally, while we do get a terminal close event on the crash, we have no way of distinguishing this from a user close event. The current solution to this problem is to add a 5 second delay between adding the policies and attempting connection. During this process we provide a progress bar with cancel. This solution is far from ideal, but does appear to work. A polling approach was also attempted, but seemed to fail as the SDK calls returned that the necessary policies were attached before a connection can be successfully made. 
- after attaching the inline policy, there is often a delay in the SSM agent starting on the instance. This results in receiving the `is SSM agent running error?` if connection is attempted immediately following adding the inline-policy (only in remote window, this error doesn't happen when trying to establish a terminal). This can take up to a few minutes to go away, or can be immediately resolved by fully restarting (not rebooting) the instance. Giving the customer an option to restart is not a feasible solution as it comes with many side effects (new billing cycle, new ip address, etc.). 
- Actions added include 
```
    'ssmmessages:CreateControlChannel',
    'ssmmessages:CreateDataChannel',
    'ssmmessages:OpenControlChannel',
    'ssmmessages:OpenDataChannel',
    'ssm:DescribeAssociation',
    'ssm:ListAssociations',
    'ssm:UpdateInstanceInformation'
```
- The bottom three, while not necessary for the connection, appear to help in starting the SSM agent on initial connection. With these policies attached, and restarting the instance, the connection works every time. 
- Actions added are expanded such that SSM agent starts on instance. 
- If no IAM role attached, we link to documentation. 

### Visuals

progress bar with cancel:
image

prompt with actions: 
image

## License


src/ec2/model.ts
```diff
@@ -113,19 +115,29 @@ export class Ec2ConnectionManager {
...
+            if (!policiesAdded) {
+                const message = `Did not add permissions. Ensure an IAM role with the proper permissions is attached to the instance. Found attached role: ${
+                    IamRole!.Arn
+                }`
+                this.throwConnectionError(message, selection, {
+                    code: 'EC2SSMPermission',
+                    documentationUri: this.policyDocumentationUri,
+                })
+            } else {
+                const timeout = new Timeout(policyAttachDelay)
```
nkomonen-amazon, line 128: Is it cheap to poll an api until we can confirm the policy is attached?

Hweinstock: I tried a polling approach on if the policy is attached, but it appears the SDK call will tell me its attached, before it actually works. When I stepped through with debugger, the polling would always succeed on the first attempt, then open terminal and it still crashes.

Hweinstock: It's possible I had polling implemented incorrectly, so I committed the function I used in the recent commit.

src/ec2/model.ts
```diff
@@ -93,6 +95,19 @@ export class Ec2ConnectionManager {
...
+    private async pollForPermissions(roleArn: string): Promise<void> {
+        const timeout = new Timeout(-1)
+        await showMessageWithCancel(`Adding inline policy to IAM role ${roleArn}`, timeout)
+        while (true) {
+            const permissionsAdded = await this.hasProperPermissions(roleArn)
+            if (permissionsAdded) {
+                break
+            }
+            await sleep(1000)
+        }
```
justinmk3, line 107: what if `hasProperPermissions` never succeeds? the while-loop needs to break if the `timeout` was canceled.


    while(!timeout.completed)

src/ec2/model.ts
```diff
@@ -113,19 +128,29 @@ export class Ec2ConnectionManager {
...
+                    code: 'EC2SSMPermission',
+                    documentationUri: this.policyDocumentationUri,
+                })
+            } else {
+                const timeout = new Timeout(policyAttachDelay)
+
+                function delay(ms: number) {
+                    return new Promise(resolve => setTimeout(resolve, ms))
+                }
+                await showMessageWithCancel(`Adding Inline Policy to ${IamRole!.Arn}`, timeout)
```
justinmk3, line 146: `showMessageWithCancel` has a `showAfterMs` parameter. 
```suggestion
                await showMessageWithCancel(`Adding Inline Policy to ${IamRole!.Arn}`, timeout, policyAttachDelay)
```

src/ec2/model.ts
```diff
@@ -113,19 +128,29 @@ export class Ec2ConnectionManager {
...
+            const policiesAdded = await promptToAddPolicies(this.iamClient, IamRole!.Arn!)
+            if (!policiesAdded) {
+                const message = `Did not add permissions. Ensure an IAM role with the proper permissions is attached to the instance. Found attached role: ${
+                    IamRole!.Arn
+                }`
+                this.throwConnectionError(message, selection, {
+                    code: 'EC2SSMPermission',
+                    documentationUri: this.policyDocumentationUri,
+                })
+            } else {
```
justinmk3, line 140: if we are throwing above, don't need `else` here.
```suggestion
            }
```

src/ec2/model.ts
justinmk3, line 146: this is a strange place for this logic. Where is the code that actually adds the policy?

src/shared/remoteSession.ts
```diff
@@ -173,13 +180,47 @@ export async function handleMissingTool(tools: Err<MissingTool[]>) {
...
-            'ssmmessages:OpenDataChannel',
-        ],
+        ActionNames: minimumSsmActions,
     })
 
     return deniedActions
 }
+
+export async function promptToAddPolicies(client: IamClient, roleArn: string): Promise<boolean> {
+    const promptText = `In order to connect, toolkit will add actions ${getFormattedSsmActions()} to role ${roleArn}`
```
justinmk3, line 190: The most verbose variable should usually go at the end of a message, else it crowds out parts that a human actually needs to read.

```suggestion
    const promptText = `${getIdeProperties().company} Toolkit will add required actions to role ${roleArn}:  ${getFormattedSsmActions()}`
```

src/shared/remoteSession.ts
```diff
@@ -173,13 +180,47 @@ export async function handleMissingTool(tools: Err<MissingTool[]>) {
...
+
+function getFormattedSsmActions() {
+    const formattedActions = minimumSsmActions.map(action => `${action},\n`).reduce((l, r) => l + r)
+
+    return formattedActions.slice(0, formattedActions.length - 2)
+}
+
+function getSsmPolicyDocument() {
+    return `{
+            \"Version\": \"2012-10-17\",
```
justinmk3, line 208: is this escaping necessary?
```suggestion
            "Version": "2012-10-17",
```

src/ec2/model.ts
Hweinstock: Currently not using this method since it appeared to give the same problems. That is, the API would say the permissions are attached, but terminal would crash. I still updated it in case there is a way to modify it to avoid this.

src/ec2/model.ts
Hweinstock: The logic for adding a policy is on line 131 just above with a call to `promptToAddPolicies` declared in `src/shared/remoteSession.ts`.  We prompt the user to see if they want to add the policy, then handle the response here. I moved the `promptToAddPolicies` to `src/shared/remoteSession.ts` since it seemed like a general utility function for establishing SSM connection, and not specific to EC2.

src/shared/remoteSession.ts
Hweinstock: Oh, good catch. The example in the documentation made me think it was needed, but appears to work without. https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/IAM.html#putRolePolicy-property

src/ec2/model.ts
justinmk3: should the "Adding policy..." progress message start before the call to `addSsmActionsToInlinePolicy` ?

src/shared/remoteSession.ts
```diff
@@ -173,13 +180,47 @@ export async function handleMissingTool(tools: Err<MissingTool[]>) {
...
 
     return deniedActions
 }
+
+export async function promptToAddPolicies(client: IamClient, roleArn: string): Promise<boolean> {
+    const promptText = `In order to connect, toolkit will add actions ${getFormattedSsmActions()} to role ${roleArn}`
+    const confirmation = await showConfirmationMessage({ prompt: promptText, confirm: 'Approve' })
+
+    if (confirmation) {
+        await addSsmActionsToInlinePolicy(client, roleArn)
```
justinmk3, line 197: If you _don't_ `await` here, could immediately return and the `Adding Inline Policy...` progress message could display sooner. 

With the current approach, there's an invisible delay and then the progress message shows later?

Hweinstock: I like that idea. I just tried it out and it is noticeably much smoother.

src/ec2/model.ts
Hweinstock: I refactored the delay logic to be in the `remoteSession.ts` file to make this change.

Hweinstock:
I noticed that the cancel button only cancels adding the policies, but not the entire process of connecting. I'm confused as to why this is happening and what I might need to add to avoid this behavior.

justinmk3:
> I noticed that the cancel button only cancels adding the policies, but not the entire process of connecting. I'm confused as to why this is happening

Is the `timeout` hooked to anything that controls the connection process? That would be one way to communicate the cancellation. Another way, less elegant (but common) is to throw a CancellationError: https://github.com/aws/aws-toolkit-vscode/blob/8fdd8d3197d26c4959db327e4c4e62242b5e0fd5/src/auth/secondaryAuth.ts#L78

packages/core/src/shared/remoteSession.ts
```diff
@@ -175,6 +181,61 @@ export async function handleMissingTool(tools: Err<MissingTool[]>) {
     )
 }
 
+function getFormattedSsmActions() {
+    const formattedActions = minimumSsmActions.map((action) => `"${action}",\n`).reduce((l, r) => l + r)
+
+    return formattedActions.slice(0, formattedActions.length - 2)
+}
+
+async function addInlinePolicyWithDelay(client: IamClient, roleArn: string) {
```
justinmk3, line 190: a short docstring would be helpful, like:

/** Shows a message for X seconds then does Y if user cancels, else does Z. */

justinmk3 (approved):

packages/core/src/shared/remoteSession.ts
```diff
@@ -175,6 +181,61 @@ export async function handleMissingTool(tools: Err<MissingTool[]>) {
...
+    const timeout = new Timeout(policyAttachDelay)
+    const message = `Adding Inline Policy to ${roleArn}`
+    await showMessageWithCancel(message, timeout)
```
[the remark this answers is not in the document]
Hweinstock: I experimented with this, but the behavior is slightly different. `showAfterMs` will delay the progress bar from popping up, whereas `delay` will extend the progress bar being up for some delay. The advantage of the second is that the moment the customer presses the button they receive feedback in a progress bar, rather than waiting for `policyAttachDelay` seconds to see any indication something is happening.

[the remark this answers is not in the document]
justinmk3: Maybe that's worth adding to `showMessageWithCancel` as an optional feature?

PR #3710, opened 2023-08-08
codewhisperer: dial up java utg support

Will-ShaoHua:
## Before
java 50% 50%
python 50% 50%

## After
java 100%
python 50% 50%

## License

### DIFF
--- a/src/codewhisperer/util/supplementalContext/utgUtils.ts
+++ b/src/codewhisperer/util/supplementalContext/utgUtils.ts
@@ -24,6 +24,27 @@ import { CodeWhispererUserGroupSettings } from '../userGroupUtil'
 import { UserGroup } from '../../models/constants'
 import { getOpenFilesInWindow } from '../../../shared/utilities/editorUtilities'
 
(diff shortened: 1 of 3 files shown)


src/codewhisperer/util/supplementalContext/utgUtils.ts
```diff
@@ -24,6 +24,27 @@ import { CodeWhispererUserGroupSettings } from '../userGroupUtil'
...
+    if (!isUtgSupportedLanguage(languageId)) {
+        return undefined
+    }
+
+    if (languageId === 'java') {
+        return true
+    } else {
+        return userGroup === UserGroup.CrossFile
+    }
+}
```
Will-ShaoHua, line 46: should return `undefined` if the language is not supported.

Will-ShaoHua: should return false if the language is supported but userGroup is `Control`

justinmk3 (approved):

andrewyuq (approved):

PR #3714, opened 2023-08-09
auth: remote workspaces use the global state, showing incorrect profile info. 

Hweinstock:
## Problem
In Title. 
## Solution
We check if we are on remote by using `vscode.env.remoteName`. Then we use workspace state instead of global state when on a non-CodeCatalyst remote instance. 

## License


JadenSimon (approved):
Works correctly on CodeCatalyst.

justinmk3:
Which commits are relevant? Github is showing change that are already on master, for some reason. Try squash-merge and rebase.

Hweinstock:
Believe I fixed it by merging in master. Will make sure to squash and merge when doing final merge.

src/auth/auth.ts
```diff
@@ -743,11 +743,21 @@ export class Auth implements AuthService, ConnectionManager {
 
     static #instance: Auth | undefined
     public static get instance() {
-        const devEnvId = getCodeCatalystDevEnvId()
-        const memento =
-            devEnvId !== undefined ? partition(globals.context.globalState, devEnvId) : globals.context.globalState
+        return (this.#instance ??= new Auth(new ProfileStore(getMemento())))
 
-        return (this.#instance ??= new Auth(new ProfileStore(memento)))
+        function getMemento() {
```
justinmk3, line 748: looks strange to define a function-scoped function after `return`

justinmk3:
Is this a correct Problem/Solution framing:

> Problem:
> When connected to remote EC2 instance, auth info is wrong.
>
> Solution:
> In a non-CodeCatalyst remote context, get auth info from "workspace state" instead of "global state".

PR #3715, opened 2023-08-09
UX: small issues with the regionSubmenu fix. 

Hweinstock:
## Problems
This PR fixes the following two issues:
- "Switch region" --> "Switch Regions"
- Duplicate "recently used" tag is replaced by "default region" and "recently used" tags. 
## Solution
Old Version:
image
New Version:
image

## License

### DIFF
--- a/src/shared/ui/common/region.ts
+++ b/src/shared/ui/common/region.ts
@@ -45,8 +45,8 @@ export function createRegionPrompter(
 
     const defaultRegionItem = items.find(item => item.detail === defaultRegion)
 
-    if (defaultRegionItem !== undefined) {
-        defaultRegionItem.description = localize('AWS.generic.recentlyUsed', '(recently used)')
+    if (defaultRegionItem !== undefined && !defaultRegionItem.recentlyUsed) {
+        defaultRegionItem.description = localize('AWS.generic.defaultRegion', '(default region)')
     }
 
     const prompter = createQuickPick(items, {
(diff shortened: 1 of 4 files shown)


src/shared/ui/common/regionSubmenu.ts
```diff
@@ -46,7 +46,7 @@ export class RegionSubmenu<T> extends Prompter<RegionSubmenuResponse<T>> {
                 data: undefined,
             },
             {
-                label: 'Switch region',
+                label: 'Switch Regions',
```
justinmk3, line 49: kind of strange to make it plural?

src/shared/ui/common/region.ts
```diff
@@ -45,8 +45,8 @@ export function createRegionPrompter(
 
     const defaultRegionItem = items.find(item => item.detail === defaultRegion)
 
-    if (defaultRegionItem !== undefined) {
-        defaultRegionItem.description = localize('AWS.generic.recentlyUsed', '(recently used)')
+    if (defaultRegionItem !== undefined && !defaultRegionItem.recentlyUsed) {
+        defaultRegionItem.description = localize('AWS.generic.defaultRegion', '(default region)')
```
justinmk3, line 49: great to have this fixed! would you mind doing `npm run newChange` and mark it as a "fix", something like:

    Regions quickpick menu shows duplicate "recently used" labels

justinmk3:
test failure:

```
  1) regionSubmenu
       allow users to swap regions via escape hatch:
     AssertionError [ERR_ASSERTION]: Unable to find item: Switch region
```

src/shared/ui/common/regionSubmenu.ts
Hweinstock: originally suggested by Jo, but double checking about the plural piece.

justinmk3 (approved):

PR #3717, opened 2023-08-10
CodeWhisperer: Fix inline suggestions not showing in the editor

andrewyuq:
There's a weird case that when an empty suggestion is provided to the inline suggestion provider, the inline suggestion UI won't show afterward even executing commands to hide/show it again.

Given that empty suggestions don't need to be shown to the users, don't call onDidReceiveRecommendations until there's at least one non-empty suggestion for the inline provider(so inline UI will be able to pick this and show up correctly)

Before: 
1. Trigger to get two suggestions, the first one is an empty string
2. The suggestions won't show in the editor, I have to do the `tab` and `backspace` trick to see the suggestions again

https://github.com/aws/aws-toolkit-vscode/assets/89420755/ae0d6315-d5e6-48cb-a21a-4bc15f8f86c3

After:
1. Trigger at the same point to get two suggestions, frist one being an empty string
2. The suggestion shows in the editor, only the second one shows (since first one is empty)

https://github.com/aws/aws-toolkit-vscode/assets/89420755/b4470cdf-5b5c-45b1-970d-b4b64c8dff6b

Fix to another case:

Before:
1. trigger by hitting "Enter", and then hit "Enter" once again
2. Some states got stuck and manual trigger at this position don't do anything

https://github.com/aws/aws-toolkit-vscode/assets/89420755/c7eb6ada-ccfd-4447-abcd-17aa535874e4

After:
1. trigger by hitting "Enter", and then hit "Enter" once again
2. The suggestion that should show got successfully shown in the editor.

https://github.com/aws/aws-toolkit-vscode/assets/89420755/696e7c17-4dea-4306-a71a-e2c867be42fc

## License

### DIFF
--- a/.changes/next-release/Bug Fix-1bdf6b9c-d49a-451d-acdc-1d4fb233e02b.json
+++ b/.changes/next-release/Bug Fix-1bdf6b9c-d49a-451d-acdc-1d4fb233e02b.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "CodeWhisperer: Fix in some cases the inline suggestions are not showing in the editor"
+}
--- a/src/codewhisperer/service/inlineCompletionService.ts
+++ b/src/codewhisperer/service/inlineCompletionService.ts
@@ -51,31 +51,6 @@ export class CWInlineCompletionItemProvider implements vscode.InlineCompletionIt
         this.activeItemIndex = undefined
     }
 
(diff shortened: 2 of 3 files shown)


justinmk3:
changelog item?

Will-ShaoHua (approved):

justinmk3 (approved):

nkomonen-amazon (approved):

PR #3723, opened 2023-08-10
telemetry(ec2): add more detailed metric on remote VSCode connection. 

Hweinstock:
## Solution
- Add a connection type for `remoteWorkspace` for opening VSCode window within instance. 
- Commons PR: https://github.com/aws/aws-toolkit-common/pull/885
- Add tests for telemetry emitted by the core ec2 functions. 

## License


justinmk3:
> The disadvantage is that its not exactly a `remoteDesktop` connection, but rather an `ssm` connection.

"ssm" is available on master now: https://github.com/aws/aws-toolkit-vscode/commit/d66f115ee4e1fed8d87fa6a59bd83a1c5925ac05#diff-9d82ddb9c2d4467395278e91f226d45aabe8e478e2a094567ef4183f7be32776R6

Hweinstock:
Went ahead and added an `ec2ConnectionInterface` type that is either `remote-dev` or `terminal`. Seemed like the most reasonable name, but open to other ideas.

justinmk3:
Oh I get it now :) I think we should add new `ec2ConnectionType` rather than a new field. Maybe `remoteWorkspace` or just `remote`

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1,5 +1,11 @@
 {
     "types": [
+        {
+            "name": "ec2ConnectionType",
+            "type": "string",
+            "allowedValues": ["remoteDesktop", "ssh", "scp", "ssm", "remoteWorkspace"],
+            "description": "Ways to connect to an EC2 Instance"
```
justinmk3, line 7: Add to https://github.com/aws/aws-toolkit-common/blob/baeeebcb52f45b2bf93ad9f3752efcbeeb6e83eb/telemetry/definitions/commonDefinitions.json#L1118-L1125 ?

P.S. seems like this could be a general `remoteConnectType` metric, but too late :/

Hweinstock: Oh, like a single metric for all connections with like some metadata for the type (i.e. ec2, ecs, codecatalyst, etc.)?

Also based on https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md#development, should I wait for EC2 release to make commons change or no reason to wait?

justinmk3: > Also based on https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md#development, should I wait for EC2 release to make commons change or no reason to wait?

Mostly a question of whether the metrics are still in development. I think these metrics are fairly settled by now.

justinmk3: can remove this now

justinmk3:
https://github.com/aws/aws-toolkit-common/commit/533415f0c79359438593793cf7ec5a7866672e96 was released in https://github.com/aws/aws-toolkit-common/commit/49de773b65efdeaadea5c30859cf286ccd74ec5b (1.0.273).

So updating the dependency here to 1.0.273 or later should work.

justinmk3:
linux ci found unhandled promise rejections

```
rejected promise not handled within 1 second: Error: Node with id testId from polling set not found
rejected promise not handled within 1 second: Error: Node with id testId from polling set not found

ERROR: Test output matched this pattern 3 times:
```

Hweinstock:
Ahhhh, forgot to disable the timer in the new tests. I feel like how I implemented the PollingSet as part of the class is becoming a code smell, with how it makes tests flaky if you aren't careful. Do you see a way to simplify it, and avoid the need the disable the polling whenever testing?

packages/core/src/test/awsService/ec2/activation.test.ts
```diff
@@ -0,0 +1,73 @@
...
+import * as vscode from 'vscode'
+import * as sinon from 'sinon'
+import { assertTelemetry } from '../../testUtil'
+import { Ec2InstanceNode } from '../../../awsService/ec2/explorer/ec2InstanceNode'
+import { Ec2ParentNode } from '../../../awsService/ec2/explorer/ec2ParentNode'
+import { Ec2Client } from '../../../shared/clients/ec2Client'
+import { Ec2ConnectionManager } from '../../../awsService/ec2/model'
+import { PollingSet } from '../../../shared/utilities/pollingSet'
+
+describe('ec2 telemetry', function () {
```
justinmk3, line 14: Telemetry is generally just logging that happens when code runs. Asserting telemetry is thus asserting that code was executed. So in most cases, we rarely need explicit "telemetry tests" (except for the telemetry module itself). 

Maybe there's not much to test right now, but these tests can be positioned as general tests for ec2 or whatever, not "telemetry tests". That makes it clear for future readers that this is where new tests for this module should go.

```suggestion
describe('ec2 activation', function () {
```

packages/core/src/test/awsService/ec2/activation.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        testNode = new Ec2InstanceNode(parentNode, testClient, testRegion, testPartition, {
+            InstanceId: 'testId',
+            LastSeenStatus: 'status',
+        })
+    })
+
+    after(function () {
+        sinon.restore()
+    })
+    it('emits correct telemetry on terminal open', async function () {
```
justinmk3, line 34:
```suggestion
    it('terminal open', async function () {
```

packages/core/src/test/awsService/ec2/activation.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        const terminalStub = sinon.stub(Ec2ConnectionManager.prototype, 'attemptToOpenEc2Terminal')
+        await vscode.commands.executeCommand('aws.ec2.openTerminal', testNode)
+
+        assertTelemetry('ec2_connectToInstance', { ec2ConnectionType: 'ssm' })
+        terminalStub.restore()
+    })
+
+    it('emits correct telemetry on remote window open', async function () {
```
justinmk3, line 42: same here, and below

packages/core/src/test/awsService/ec2/activation.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        terminalStub.restore()
+    })
+
+    it('remote window open', async function () {
+        const remoteWindowStub = sinon.stub(Ec2Connecter.prototype, 'tryOpenRemoteConnection')
+        await vscode.commands.executeCommand('aws.ec2.openRemoteConnection', testNode)
+
+        assertTelemetry('ec2_connectToInstance', { ec2ConnectionType: 'remoteWorkspace' })
+        remoteWindowStub.restore()
+    })
```
justinmk3, line 48: if these tests have stubbed the command so that it does nothing useful, then they could all be one `it('telemetry')` test.

Alternatively, if there are existing tests for these commands, can't those existing tests call `assertTelemetry` ?

Hweinstock: The high-level commands are simply wrappers around the logic that is tested in https://github.com/aws/aws-toolkit-vscode/blob/142761e268dd48687d0f696f6ded22819c74ffc7/packages/core/src/test/awsService/ec2/model.test.ts. The only difference at this level is the prompting logic (tested in https://github.com/aws/aws-toolkit-vscode/blob/142761e268dd48687d0f696f6ded22819c74ffc7/packages/core/src/test/awsService/ec2/prompter.test.ts) and telemetry (tested here). It doesn't seem to me to make sense to test any E2E aspects of the code beyond telemetry here.

justinmk3 (approved):

PR #3725, opened 2023-08-11
CodeWhisperer: Correct completion type for every completion in user decision events

andrewyuq:
1. Previously we just use the completion type of the first suggestion, as the completion type for all suggestions. Now since this is no longer true, plugin side also needs to make the change to compute the completion type for every suggestion. Since this is per-suggestion level this will be recorded in the userDecision event.

2. Note that we still have 'completionType' field in other events( serviceInvocation, clientComponentLatency, etc) these values no longer make sense for analysis.

3. For actually computing the completionType for each suggestion, the rule is:
If a suggestion has multi-lines and they are non-blank lines, it's a block suggestion, otherwise it's a line suggestion.

## License


src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -83,6 +85,15 @@ export class RecommendationHandler {
         )
     }
 
+    setCompletionType(index: number, recommendation: Recommendation) {
+        const nonBlankLines = recommendation.content.split('\n').filter(line => line.trim() !== '').length
+        this.completionTypes.set(index, nonBlankLines > 1 ? 'Block' : 'Line')
+    }
+
+    getCompletionType(index: number): CodewhispererCompletionType {
+        return this.completionTypes.get(index) || 'Line'
```
Will-ShaoHua, line 94: why would we do 
```
return this.completionTypes.get(index) || undefined
```

I think theoretically we never should reach `||` which is the fallback value right?
If we set it to `Line`, we are unlikely to know if there is a bug/issue with this?

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -417,6 +412,10 @@ export class TelemetryHelper {
         return i === acceptIndex ? 'Accept' : 'Ignore'
     }
 
+    public getCompletionType(i: number, completionTypes: Map<number, CodewhispererCompletionType>) {
```
Will-ShaoHua, line 415: don't think we need another method for its purpose?

src/test/codewhisperer/service/recommendationHandler.test.ts
```diff
@@ -141,11 +142,10 @@ describe('recommendationHandler', function () {
             assertTelemetry({
                 codewhispererRequestId: 'test_request',
                 codewhispererSessionId: 'test_request',
-                codewhispererLastSuggestionIndex: -1,
+                codewhispererLastSuggestionIndex: 1,
```
Will-ShaoHua, line 145: i might missed the context, Andrew could you point me out where is the change that requires this change

Will-ShaoHua:
seems like we will have 1 PR for JB right?

src/codewhisperer/service/recommendationHandler.ts
andrewyuq: If we set to `undefined` -- some of the events are marking this value as a necessary value, so it difficult to get away with it. This is an issue about using a separate map structure to store this value which shouldn't be the case, there should be a context object that stores every information about a suggestion, like (`DetailContext` in JB), not several different maps. Agreeing this can be a problem for now but I want to address this in the refactor work.

Will-ShaoHua (approved):

src/test/codewhisperer/service/recommendationHandler.test.ts
andrewyuq: It's the movement of this part of code `recordServiceInvocationTelemetry`. It's moved after `this.recommendations` are updated. If we don't do that, a serviceInvocation will use an empty `this.recommendations` value and always send a `-1` in the first serviceInvocation event for the session, which is incorrect. This value in the first event should record the # of suggestions received so far minus one. (including the first response).

src/codewhisperer/service/recommendationHandler.ts
Will-ShaoHua: sgtm, yea we need a refactor work to get rid of these states

src/codewhisperer/util/telemetryHelper.ts
andrewyuq: It's just a wrapper to hide the `||` logic so caller don't have to worry about it.

src/test/codewhisperer/service/recommendationHandler.test.ts
Will-ShaoHua: sg

justinmk3 (approved):

PR #3730, opened 2023-08-14
test: Skip CodeCatalyst E2E tests

nkomonen-amazon:
These tests always fail due to a throttling issue with CodeCatalyst, see the SIM IDE-10496 for more info.

We will eventually re-enable these tests once the CC side fixes the issues causing this.

## License

### DIFF
--- a/src/testE2E/codecatalyst/client.test.ts
+++ b/src/testE2E/codecatalyst/client.test.ts
@@ -81,7 +81,7 @@ let projectName: CodeCatalystProject['name']
  *     integ tests, but using the ssh hostname that we get from
  *     {@link prepareDevEnvConnection}.
  */
-describe('Test how this codebase uses the CodeCatalyst API', function () {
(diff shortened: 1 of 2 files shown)


src/testE2E/codecatalyst/client.test.ts
```diff
@@ -81,7 +81,7 @@ let projectName: CodeCatalystProject['name']
  *     integ tests, but using the ssh hostname that we get from
  *     {@link prepareDevEnvConnection}.
  */
-describe('Test how this codebase uses the CodeCatalyst API', function () {
+describe.skip('Test how this codebase uses the CodeCatalyst API', function () {
```
justinmk3, line 84: Perhaps we could catch exception and skip only if the error matches `ThrottlingException`.

A data-based check is also ok, but 1 month is probably too short.

justinmk3 (approved):

PR #3740, opened 2023-08-15
fileSystem: Class that works with browser or desktop file system

nkomonen-amazon:
Create Browser + Desktop compatible file system.

Future commits will make use of this, and existing uses of a different filesystem implementation (fs, fs-extra) should be
swapped to use these over time if possible.

## License

### DIFF
--- a/src/srcShared/fs.ts
+++ b/src/srcShared/fs.ts
@@ -1,0 +1,134 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import * as vscode from 'vscode'
+
+const fs = vscode.workspace.fs
+type Uri = vscode.Uri
+
+/**
+ * @warning Do not import this class specifically, instead import the instance {@link fsCommon}.
+ *
+ * This class contains file system methods that are "common", meaning
+ * it can be used in the browser or desktop.
+ *
+ * Technical Details:
+ * TODO: Verify the point below once I get this hooked up to the browser code.
+ * - vscode.workspace.fs dynamically resolves the correct file system provider
+ *   to use. This means that in the browser it will attempt to use the File System
+ *   Access API.
+ *
+ *
+ * MODIFYING THIS CLASS:
+ * - All methods must work for both browser and desktop
+ * - Do not use 'fs' or 'fs-extra' since they are not browser compatible.
+ *   If they have functionality that cannot be achieved with a
+ *   browser+desktop implementation, then create it in the module {@link TODO}
+ */
+export class FileSystemCommon {
+    private constructor() {}
+    static #instance: FileSystemCommon
+    static get instance(): FileSystemCommon {
+        return (this.#instance ??= new FileSystemCommon())
+    }
+
+    async mkdir(path: Uri | string): Promise<void> {
+        path = FileSystemCommon.getUri(path)
+        return fs.createDirectory(path)
+    }
+
+    async readFile(path: Uri | string): Promise<Uint8Array> {
+        path = FileSystemCommon.getUri(path)
+        return fs.readFile(path)
+    }
+
+    async readFileAsString(path: Uri | string): Promise<string> {
+        path = FileSystemCommon.getUri(path)
+        return FileSystemCommon.arrayToString(await this.readFile(path))
+    }
+
+    async appendFile(path: Uri | string, content: Uint8Array | string): Promise<void> {
+        path = FileSystemCommon.getUri(path)
+
+        const currentContent: Uint8Array = (await this.fileExists(path)) ? await this.readFile(path) : new Uint8Array(0)
+        const currentLength = currentContent.length
+
+        const newContent = FileSystemCommon.asArray(content)
+        const newLength = newContent.length
+
+        const finalContent = new Uint8Array(currentLength + newLength)
+        finalContent.set(currentContent)
+        finalContent.set(newContent, currentLength)
+
+        return this.writeFile(path, finalContent)
+    }
+
+    async fileExists(path: Uri | string): Promise<boolean> {
+        path = FileSystemCommon.getUri(path)
+        const stat = await this.stat(path)
+        return stat === undefined ? false : stat.type === vscode.FileType.File
+    }
+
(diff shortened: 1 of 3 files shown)


src/shared/fs/fileSystem.ts
justinmk3, line 1: Can be named just `shared/fs.ts`. Unlikely that we'll have many other things in the fs/ dir

src/shared/fs/fileSystem.ts
```diff
@@ -0,0 +1,126 @@
...
+ * - vscode.workspace.fs dynamically resolves the correct file system provider
+ *   to use. This means that in the browser it will attempt to use the File System
+ *   Access API.
+ *
+ *
+ * MODIFYING THIS CLASS:
+ * - All methods must work for both browser and desktop
+ * - Do not use 'fs' or 'fs-extra' since they are not browser compatible.
+ *   If they have functionality that cannot be achieved with a
+ *   browser+desktop implementation, then create it in the module {@link TODO}
```
justinmk3, line 25: This class should live in a new folder named something like `srcShared/`. 

We should pretty soon add a linter that checks imports. We should do this now, instead of putting such modules in the existing `src/` directory.

nkomonen-amazon: Moved to `src/srcShared` for now

src/srcShared/fs.ts
```diff
@@ -0,0 +1,133 @@
...
+    private static getUri(path: string | vscode.Uri): vscode.Uri {
+        if (path instanceof vscode.Uri) {
+            return path
+        }
+        return vscode.Uri.file(path)
+    }
+}
+
+const fsCommon = new FileSystemCommon()
+export default fsCommon
```
justinmk3, line 133: We generally don't use `export default`. Is there a reason to introduce it? Can add friction to refactoring.

src/srcShared/fs.ts
```diff
@@ -0,0 +1,133 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import * as vscode from 'vscode'
+
+const fs = vscode.workspace.fs
+type Uri = vscode.Uri
+
+/**
+ * @warning Do not import this class specifically, instead import the instance {@link fsCommon}.
```
justinmk3, line 11: could use a private constructore, and this instance pattern that is used in a few places already: https://github.com/aws/aws-toolkit-vscode/blob/4dcee33931693380739eaa5d44e92fa4545a9666/src/shared/settings.ts#L167-L174

src/srcShared/fs.ts
```diff
@@ -0,0 +1,133 @@
...
+ * - All methods must work for both browser and desktop
+ * - Do not use 'fs' or 'fs-extra' since they are not browser compatible.
+ *   If they have functionality that cannot be achieved with a
+ *   browser+desktop implementation, then create it in the module {@link TODO}
+ */
+export class FileSystemCommon {
+    constructor() {}
+
+    async mkdir(path: Uri | string): Promise<void> {
+        path = FileSystemCommon.getUri(path)
```
justinmk3, line 33: In general we probably want to force callers to provide a Uri. However this fs module is probably used in many legacy places so we probably need to support string names as you have done.

src/srcShared/fs.ts
```diff
@@ -0,0 +1,133 @@
...
+        return stat === undefined ? false : stat.type === vscode.FileType.File
+    }
+
+    async writeFile(path: Uri | string, data: string | Uint8Array): Promise<void> {
+        path = FileSystemCommon.getUri(path)
+        return fs.writeFile(path, FileSystemCommon.asArray(data))
+    }
+
+    /**
+     * @returns the stat of the file or undefined if it does not exist
```
justinmk3, line 75: this docstring is not telling us anything the signature didn't already tell us :)

src/test/srcShared/fs.test.ts
```diff
@@ -0,0 +1,230 @@
...
+            const path = await makeFile('test.txt', 'hello world')
+            const content = await fsCommon.readFileAsString(path)
+            assert.strictEqual(content, 'hello world')
+        })
+
+        it('reads a file from a URI', async function () {
+            const path = await makeFile('test.txt', 'hello world')
+            const uri = vscode.Uri.file(path)
+            const content = await fsCommon.readFileAsString(uri)
+            assert.strictEqual(content, 'hello world')
```
justinmk3, line 45: Could cover the string/URI variants in one test. That saves a lot of code (given all the string/URI variants in fs.ts), which allows the reader to focus on the more interesting and relevant things to be tested.

The variants are probably not even needed, since passing a string implicitly tests both (and the typescript type system tells us the possible inputs).
```suggestion
            const content = await fsCommon.readFileAsString(uri)
            assert.strictEqual(content, 'hello world')
            assert.strictEqual((await fsCommon.readFileAsString(path)), 'hello world')
```

src/srcShared/fs.ts
nkomonen-amazon: I did the export of the instance so that for each use we don't always need to call `FileSystem.instance.{action}()` and could instead do `fsCommon.{action}()`, similar to `fs`. But makes sense to follow what we've already been doing, will update.

src/srcShared/fs.ts
nkomonen-amazon: Related to the instance getter, this code has been removed

src/srcShared/fs.ts
nkomonen-amazon: Yeah, this is to support legacy for now. It will be one of those things where in PR review we'll want to ensure a `vscode.Uri` object instead of a string.

src/srcShared/fs.ts
nkomonen-amazon: I figured this was different since it may not be clear `undefined` is only for the case where it does not exist. If there was an error, eg permissions, we would throw

src/test/srcShared/fs.test.ts
nkomonen-amazon: Makes sense. I'll just keep this single test that takes the URI input as a sanity check, but will merge the existing test in to one.

justinmk3 (approved):

PR #3744, opened 2023-08-16
remove log message

nkomonen-amazon:
Fixes IDE-11421

## License

### DIFF
--- a/src/auth/ui/vue/show.ts
+++ b/src/auth/ui/vue/show.ts
@@ -69,7 +69,6 @@ export class AuthWebview extends VueWebview {
     }
 
     getCredentialFormatError(key: CredentialsKey, value: string | undefined): string | undefined {
-        getLogger().warn('getCredentialFormatError(): %s %s', key, value)
         return getCredentialFormatError(key, value)
     }
 


justinmk3 (approved):
We have logic to omit secrets, but that didn't case this case.

https://github.com/aws/aws-toolkit-vscode/blob/3c4658c7319330c1c4afcd44f3005deeb318acef/src/shared/utilities/typeConstructors.ts#L64

https://github.com/aws/aws-toolkit-vscode/blob/3c4658c7319330c1c4afcd44f3005deeb318acef/src/auth/sso/clients.ts#L202-L206

src/auth/ui/vue/show.ts
```diff
@@ -69,7 +69,6 @@ export class AuthWebview extends VueWebview {
     }
 
     getCredentialFormatError(key: CredentialsKey, value: string | undefined): string | undefined {
-        getLogger().warn('getCredentialFormatError(): %s %s', key, value)
         return getCredentialFormatError(key, value)
```
justinmk3, line 72: After removing the line, is the error logged in some other fashion (without secret values)?

nkomonen-amazon: We don't log the error elsewhere in this flow. The actual output of `getCredentialFormatError()` just has an error message but doesn't include the actual secret values.

The object this function returns is displayed in the webview UI to the user.

justinmk3: If we aren't logging this failure case at all, we should be... Logs are often the best hope of troubleshooting problems, not screenshots or UI messages

PR #3745, opened 2023-08-16
doc: create faq page

nkomonen-amazon:
This will be our new FAQ page for common issues that we can easily update.

## License


FAQ.md
justinmk3, line 1: Let's put it in `docs/faq-credentials.md` in case we need more pages for non-credentials topics.

justinmk3 (approved):

PR #3747, opened 2023-08-16
auth: Helper methods to check if connection exists

nkomonen-amazon:
This creates simple helper methods to be able to check if a connection of a certain type exists.

The methods answer questions similar to: `Is the toolkit extension aware of a Builder ID connection for CodeWhisperer, regardless of if the connection is active?`

- CodeCatalyst and CodeWhisperer have functions in their auth classes to check for connections specific to them

### Why?
These will be used in future commits by the Add Connections page for telemetry purposes.

## License

### DIFF
--- a/src/auth/connection.ts
+++ b/src/auth/connection.ts
@@ -23,11 +23,34 @@ export const ssoAccountAccessScopes = ['sso:account:access']
 export const codewhispererScopes = ['codewhisperer:completions', 'codewhisperer:analysis']
 export const defaultSsoRegion = 'us-east-1'
 
+type SsoType =
+    | 'any' // any type of sso
+    | 'base' // AWS Identity Center
+    | 'builderId'
+
 export const isIamConnection = (conn?: Connection): conn is IamConnection => conn?.type === 'iam'
-export const isSsoConnection = (conn?: Connection): conn is SsoConnection => conn?.type === 'sso'
-export const isBuilderIdConnection = (conn?: Connection): conn is SsoConnection =>
-    isSsoConnection(conn) && conn.startUrl === builderIdStartUrl
-export const isIdcConnection = (conn?: Connection) => isSsoConnection(conn) && !isBuilderIdConnection(conn)
+export const isSsoConnection = (conn?: Connection, type: SsoType = 'any'): conn is SsoConnection => {
+    if (conn?.type !== 'sso') {
+        return false
+    }
+    // At this point the conn is an SSO conn, but now we must determine the specific type
+    switch (type) {
+        case 'base':
+            // Any SSO type is considered an SSO connection, but in this case
+            // we only want an Identity Center connection. So we return true
+            // as long as the connection is none of the other connection types.
+
+            // IMPORTANT: This condition should grow as more SsoType's get added.
+            return !isBuilderIdConnection(conn)
+        case 'builderId':
+            return conn.startUrl === builderIdStartUrl
+        case 'any':
+            return true
+    }
+}
+export const isAnySsoConnection = (conn?: Connection): conn is SsoConnection => isSsoConnection(conn, 'any')
+export const isBaseSsoConnection = (conn?: Connection): conn is SsoConnection => isSsoConnection(conn, 'base')
+export const isBuilderIdConnection = (conn?: Connection): conn is SsoConnection => isSsoConnection(conn, 'builderId')
 
 export function hasScopes(target: SsoConnection | SsoProfile, scopes: string[]): boolean {
     return scopes?.every(s => target.scopes?.includes(s))
--- a/src/auth/connection.ts
+++ b/src/auth/connection.ts
@@ -52,6 +52,9 @@ export const isAnySsoConnection = (conn?: Connection): conn is SsoConnection => 
 export const isBaseSsoConnection = (conn?: Connection): conn is SsoConnection => isSsoConnection(conn, 'base')
 export const isBuilderIdConnection = (conn?: Connection): conn is SsoConnection => isSsoConnection(conn, 'builderId')
 
+export const isValidCodeCatalystConnection = (conn: Connection): conn is SsoConnection =>
+    isBuilderIdConnection(conn) && hasScopes(conn, codecatalystScopes)
+
 export function hasScopes(target: SsoConnection | SsoProfile, scopes: string[]): boolean {
     return scopes?.every(s => target.scopes?.includes(s))
 }
--- a/src/auth/connection.ts
+++ b/src/auth/connection.ts
@@ -27,6 +27,7 @@ export const isIamConnection = (conn?: Connection): conn is IamConnection => con
 export const isSsoConnection = (conn?: Connection): conn is SsoConnection => conn?.type === 'sso'
 export const isBuilderIdConnection = (conn?: Connection): conn is SsoConnection =>
     isSsoConnection(conn) && conn.startUrl === builderIdStartUrl
+export const isIdcConnection = (conn?: Connection) => isSsoConnection(conn) && !isBuilderIdConnection(conn)
 
 export function hasScopes(target: SsoConnection | SsoProfile, scopes: string[]): boolean {
     return scopes?.every(s => target.scopes?.includes(s))
--- a/src/auth/connection.ts
+++ b/src/auth/connection.ts
@@ -25,7 +25,7 @@ export const defaultSsoRegion = 'us-east-1'
 
 type SsoType =
     | 'any' // any type of sso
-    | 'base' // AWS Identity Center
+    | 'idc' // AWS Identity Center
     | 'builderId'
 
 export const isIamConnection = (conn?: Connection): conn is IamConnection => conn?.type === 'iam'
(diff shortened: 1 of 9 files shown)


src/codecatalyst/auth.ts
```diff
@@ -69,6 +69,14 @@ export class CodeCatalystAuthenticationProvider {
         return this.secondaryAuth.isUsingSavedConnection
     }
 
+    /** True if a CC Builder ID connection is known by the Toolkit*/
+    public async builderIdExists(allConnections = () => this.auth.listConnections()): Promise<boolean> {
+        const predicate = ((conn?: Connection) => {
+            return isBuilderIdConnection(conn) && isValidCodeCatalystConnection(conn)
+        })
+        return (await allConnections()).some(predicate)
```
justinmk3, line 77: Instead of continuing to duplicate stuff in a spread-out fashion, it would help discovery if `auth/utilts.ts` had a single function that took a `kind='plain'|'codecatalyst'|'codewhisperer'` parameter.

Or something like that. If that introduces a circular dependency, create a new file `src/auth/utils2.ts`.

src/auth/connection.ts
```diff
@@ -28,6 +28,7 @@ export const isIamConnection = (conn?: Connection): conn is IamConnection => con
 export const isSsoConnection = (conn?: Connection): conn is SsoConnection => conn?.type === 'sso'
 export const isBuilderIdConnection = (conn?: Connection): conn is SsoConnection =>
     isSsoConnection(conn) && conn.startUrl === builderIdStartUrl
+export const isEnterpriseSsoConnection = (conn?: Connection) => isSsoConnection(conn) && !isBuilderIdConnection(conn)
```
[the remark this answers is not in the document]
nkomonen-amazon: If we just did the inverse of `isBuilderIdConnection` it could imply enterprise sso or IAM credentials.
This is why we need the granularity of this function for non-builder id sso connections.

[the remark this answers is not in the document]
nkomonen-amazon: I saw "enterprise" being used in the CodeWhisperer code, so I assumed that this is the new standard.

Instead of "enterprise" I could also see using "identityCenter".
So we would have:
-  "sso" that could be Builder ID or Identity Center
- "builderId" which is only builder id sso's
- "identityCenter" which is only identityCenter sso's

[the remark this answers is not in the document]
justinmk3: `identityCenter` or `Idc` is better. "enterprise" is specific to CW's particular currently supported concepts. Their functions should also be renamed, but that should be a separate PR.

justinmk3:
Is there existing code that can use these functions and perhaps delete some older code?

src/codecatalyst/auth.ts
nkomonen-amazon: Do we want to move context specific code (eg: CodeWhisperer Identity Center, CodeCatalyst Builder ID) outside of their dedicated auth classes? I'd assume this would make discovery harder since you would have to know to look in `src/auth/utils.ts` for a specific CodeWhisperer functionality, instead of in the existing `src/codewhisperer/util/authUtil.ts` module.

nkomonen-amazon:
@justinmk3 There is nothing that currently uses it since we never really cared if a connection existed but wasn't in use.

But with the Add Connection telemetry we now care about what connections existed before the Add Connection webview was interacted with since we can get an overall snapshot of all the existing connections in the toolkit.

src/codecatalyst/auth.ts
justinmk3: > outside of their dedicated auth classes? ...   would make discovery harder since you would have to know to look in src/auth/utils.ts for a specific CodeWhisperer functionality, instead of src/codewhisperer/util/authUtil.ts module.

That is always the case for any generic functionality: generic functionality is found in generic modules, I don't see a discovery problem there. 

The functionality in this PR is generalizable. The special-case auth modules are an accident, not something we want to grow.

nkomonen-amazon: Also regarding the duplicate test code, these tests are more-so testing the `predicate` of each method instead of the `some` since each implementation has a slightly different predicate.

nkomonen-amazon: > The special-case auth modules are an accident, not something we want to grow.

I see, will update.

src/auth/connection.ts
```diff
@@ -24,10 +24,33 @@ export const ssoAccountAccessScopes = ['sso:account:access']
...
+    }
+    // At this point the conn is an SSO conn, but now we must determine the specific type
+    switch (type) {
+        case 'base':
+            // Any SSO type is considered an SSO connection, but in this case
+            // we only want an Identity Center connection. So we return true
+            // as long as the connection is none of the other connection types.
+
+            // IMPORTANT: This condition should grow as more SsoType's get added.
+            return !isSsoConnection(conn, 'builderId')
```
justinmk3, line 42: nit: does recursion reduce the clarity

nkomonen-amazon: The alternative to this is to call `!isBuilderIdConnection()` which is a function a few lines below this. But internally it is calling this exact line: `isSsoConnection(conn, 'builderId')`.

Would using `isBuilderIdConnection()` instead be more clear in your opinion?

src/auth/connection.ts
```diff
@@ -23,10 +23,33 @@ export const ssoAccountAccessScopes = ['sso:account:access']
...
+        case 'builderId':
+            return conn.startUrl === builderIdStartUrl
+        case 'any':
+            return true
+    }
+}
+export const isBaseSsoConnection = (conn?: Connection): conn is SsoConnection => isSsoConnection(conn, 'base')
+export const isBuilderIdConnection = (conn?: Connection): conn is SsoConnection => isSsoConnection(conn, 'builderId')
+
+export const isValidCodeCatalystConnection = (conn: Connection): conn is SsoConnection =>
```
justinmk3, line 51: Not a blocker for this PR, but `isValidCodeWhispererConnection` could also live here.

Also `getBearerToken` and `getCredentials` have no business being in the CW special-case auth module: https://github.com/aws/aws-toolkit-vscode/blob/dcf55e4d147638de82b026e31a9bef6dee70bdfa/src/codewhisperer/util/authUtil.ts#L154-L181

Probably almost nothing in the CW special-case needs to be special. Eliminating that kind of thing will reduce weird quirky behavior.

nkomonen-amazon: Yeah I agree. We'll be able to simplify a lot moving stuff out of the CW specific file

src/test/codewhisperer/util/authUtil.test.ts
```diff
@@ -128,11 +128,11 @@ describe('AuthUtil', async function () {
         await authUtil.connectToAwsBuilderId()
         assert.ok(authUtil.isConnected())
         assert.ok(authUtil.isConnectionValid())
-        assert.ok(isSsoConnection(authUtil.conn))
+        assert.ok(isBuilderIdConnection(authUtil.conn))
```
justinmk3, line 131: why wasn't this changed to `isAnySsoConnection` like the other cases above?

src/auth/connection.ts
```diff
@@ -23,10 +23,37 @@ export const ssoAccountAccessScopes = ['sso:account:access']
...
+
+            // IMPORTANT: This condition should grow as more SsoType's get added.
+            return !isBuilderIdConnection(conn)
+        case 'builderId':
+            return conn.startUrl === builderIdStartUrl
+        case 'any':
+            return true
+    }
+}
+export const isAnySsoConnection = (conn?: Connection): conn is SsoConnection => isSsoConnection(conn, 'any')
```
justinmk3, line 51: `type` defaults to 'any' so is this function needed?

src/auth/connection.ts
```diff
@@ -23,10 +23,37 @@ export const ssoAccountAccessScopes = ['sso:account:access']
...
+export const isBaseSsoConnection = (conn?: Connection): conn is SsoConnection => isSsoConnection(conn, 'base')
```
justinmk3, line 52: - Ideally we would avoid "enterprise sso" or "idc" or other marketing names, but using yet another term "base" probably is a higher cost than just using "idc".
- If wonder if we'll later need IdC sub-types like `idc.federated`.


```suggestion
export const isIdcSsoConnection = (conn?: Connection): conn is SsoConnection => isSsoConnection(conn, 'idc')
```

src/auth/utils.ts
```diff
@@ -558,6 +563,81 @@ export class AuthNode implements TreeNode<Auth> {
...
+
+export type BuilderIdKind = 'any' | 'codewhisperer' | 'codecatalyst'
+
+/**
+ * Returns true if a Builder ID connection exists.
+ *
+ * @param kind A Builder ID connection that has the scopes of this kind.
+ * @param allConnections func to get all connections that exist
+ */
+export async function builderIdExists(
```
justinmk3, line 612: would `has` be a more discoverable prefix for these functions:

```
hasBuilderId
hasSso
hasIamCredentials
```

That also parallels nicely with the `isIamConnection`/`isSsoConnection`/etc functions.

src/test/codewhisperer/util/authUtil.test.ts
nkomonen-amazon: This test uses a builder id in this case

src/auth/connection.ts
nkomonen-amazon: This is used in a few places with `filter()` and lets us do `filter(isAnySsoConnection)`, instead of `filter((e) => isSsoConnection(e))`

justinmk3: can't `filter(isSsoConnection)` work if `type` defaults to 'any' ?

src/auth/utils.ts
nkomonen-amazon: Updated

src/auth/connection.ts
nkomonen-amazon: Updated

src/auth/connection.ts
nkomonen-amazon: I wish that was the case :(
The argument for `filter()` is `predicate: (value: never, index: number, array: never[])` so the `index` argument will incorrectly be passed to `type`

justinmk3 (approved):

PR #3748, opened 2023-08-17
fix: misleading "Permission set" warning for OIDC SSO/IdC connection

justinmk3:
Problem:
Some IdC/SSO connections have OIDC "scopes" and don't have IAM "roles" (returning IAM credentials / "Permission sets"). Such connections are useful for scope-based applications, even though they don't have IAM credentials.

Solution:
Don't warn about missing roles/permission-sets for connections that have OIDC scopes.

## License

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -223,12 +223,16 @@ export class Auth implements AuthService, ConnectionManager {
 
             const stream = toStream(this.store.listProfiles().map(entry => this.getConnectionFromStoreEntry(entry)))
 
+            /** Decides if SSO service should be queried for "linked" IAM roles/credentials for the given SSO connection. */
             const isLinkable = (
                 entry: [string, StoredProfile<Profile>]
-            ): entry is [string, StoredProfile<SsoProfile>] =>
-                entry[1].type === 'sso' &&
-                hasScopes(entry[1], ssoAccountAccessScopes) &&
-                entry[1].metadata.connectionState === 'valid'
+            ): entry is [string, StoredProfile<SsoProfile>] => {
+                const r =
+                    entry[1].type === 'sso' &&
+                    hasScopes(entry[1], ssoAccountAccessScopes) &&
+                    entry[1].metadata.connectionState === 'valid'
+                return r
+            }
 
             const linked = this.store
                 .listProfiles()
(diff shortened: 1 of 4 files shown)


src/auth/connection.ts
```diff
@@ -289,17 +289,17 @@ export async function* loadLinkedProfilesIntoStore(
...
             ssoAccountId: info.accountId,
         })
 
         yield [id, profile] as const
     }
 
-    const isBuilderId = startUrl === builderIdStartUrl // Special case.
-    if (!isBuilderId && (accounts.size === 0 || found.size === 0)) {
-        const name = truncateStartUrl(startUrl)
+    const hasScopes = !!ssoProfile.scopes?.some(s => !ssoAccountAccessScopes.includes(s))
```
nkomonen-amazon, line 300: Should the `!` not be in `!ssoAccountScopes...` since this implies the ssoProfile has no `ssoAccountAccessScopes` and `hasScopes` means the ssoProfile has `ssoAccountAccessScopes`?

src/auth/auth.ts
```diff
@@ -225,10 +225,13 @@ export class Auth implements AuthService, ConnectionManager {
 
             const isLinkable = (
                 entry: [string, StoredProfile<Profile>]
-            ): entry is [string, StoredProfile<SsoProfile>] =>
-                entry[1].type === 'sso' &&
-                hasScopes(entry[1], ssoAccountAccessScopes) &&
-                entry[1].metadata.connectionState === 'valid'
+            ): entry is [string, StoredProfile<SsoProfile>] => {
+                const r =
```
nkomonen-amazon, line 229: Can we just return instead of assigning to `const r`?

nkomonen-amazon (approved):

src/auth/connection.ts
justinmk3: > since this implies the ssoProfile has no `ssoAccountAccessScopes`

The query is asking "are there any scopes which are _not_ `ssoAccountAccessScopes`"? But it's hard to read, will try to improve it.

src/auth/auth.ts
justinmk3: This is intentional for debugging. Hard to debug dense one-liners.

src/auth/connection.ts
justinmk3: Added a comment, but I can't find a better way to express the query. Also considered renaming `hasScopes` but nothing else seems better.

PR #3760, opened 2023-08-22
refactor: enable esModuleInterop + fix imports

nkomonen-amazon:
Enabling the [esModuleInterop](https://www.typescriptlang.org/tsconfig#esModuleInterop) field in tsconfig.json enables module import compatibility with CommonJS
modules.

Now we do not need to use `import * as x from 'x'` and can just do `import x from x` on modules that have a default export.

Resolves IDE-6382

### Why?
The latest version of the `got` module requires this flag enabled and we need to test if the latest version of `got` works in C9.

Also this was a separate ticket that needed to be done regardless.

## License


justinmk3:
Nice! Related: https://github.com/aws/aws-toolkit-vscode/pull/2591#issuecomment-1108842773

justinmk3 (approved):

justinmk3:
Should we also set the `module` tsconfig setting? https://www.typescriptlang.org/tsconfig#esModuleInterop mentions:

> Recommended Default:
> true if [module](https://www.typescriptlang.org/tsconfig#module) is node16 or nodenext; 
> false otherwise.

andrewyuq (approved):

PR #3762, opened 2023-08-22
codewhisperer: add  logs to supplemental context feature

Will-ShaoHua:
## License

### DIFF
--- a/src/codewhisperer/util/editorContext.ts
+++ b/src/codewhisperer/util/editorContext.ts
@@ -226,13 +226,14 @@ function logSupplementalContext(supplementalContext: CodeWhispererSupplementalCo
             latency: ${supplementalContext.latency},
         `)
 
+    let logString = ''
     supplementalContext.supplementalContextItems.forEach((context, index) => {
-        getLogger().debug(`
-                -----------------------------------------------
-                Path: ${context.filePath}
-                Score: ${context.score}
-                Chunk: ${index}:${context.content}
-                -----------------------------------------------
-            `)
+        logString += `Chunk ${index}:
+        Path: ${context.filePath}
+        Content: ${index}:${context.content}
+        Score: ${context.score}
+        -----------------------------------------------`
     })
+
+    getLogger().debug(logString)
 }
--- a/src/codewhisperer/util/editorContext.ts
+++ b/src/codewhisperer/util/editorContext.ts
@@ -219,7 +219,7 @@ function logSupplementalContext(supplementalContext: CodeWhispererSupplementalCo
         return
     }
 
-    getLogger().verbose(`
+    getLogger().debug(`CodeWhispererSupplementalContext:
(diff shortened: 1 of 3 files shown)


andrewyuq (approved):

src/codewhisperer/util/editorContext.ts
```diff
@@ -219,15 +219,15 @@ function logSupplementalContext(supplementalContext: CodeWhispererSupplementalCo
...
+    getLogger().debug(`CodeWhispererSupplementalContext:
             isUtg: ${supplementalContext.isUtg},
             isProcessTimeout: ${supplementalContext.isProcessTimeout},
             contentsLength: ${supplementalContext.contentsLength},
             latency: ${supplementalContext.latency},
         `)
 
     supplementalContext.supplementalContextItems.forEach((context, index) => {
-        getLogger().verbose(`
+        getLogger().debug(`
```
justinmk3, line 230: Logging related messages in separate logger calls should not be done in a loop. Build _one_ string, then log that single string.

Will-ShaoHua: understood

src/codewhisperer/util/editorContext.ts
```diff
@@ -219,20 +219,21 @@ function logSupplementalContext(supplementalContext: CodeWhispererSupplementalCo
...
-                -----------------------------------------------
-            `)
+        logString += `Chunk ${index}:
+        Path: ${context.filePath}
+        Content: ${index}:${context.content}
+        Score: ${context.score}
+        -----------------------------------------------`
     })
+
+    getLogger().debug(logString)
```
justinmk3, line 238: This is an improvement, thanks. But there is still a disconnected `getLogger().debug()` call on line 222. Only one one message is needed, not multiple disconnected ones. That also has the advantage of putting a single timestamp on the message.

Will-ShaoHua: yea make sense, let me update it

Will-ShaoHua: knowledge learned thanks Justin

src/codewhisperer/util/editorContext.ts
```diff
@@ -219,20 +219,21 @@ function logSupplementalContext(supplementalContext: CodeWhispererSupplementalCo
...
-    getLogger().verbose(`
-            isUtg: ${supplementalContext.isUtg},
-            isProcessTimeout: ${supplementalContext.isProcessTimeout},
-            contentsLength: ${supplementalContext.contentsLength},
-            latency: ${supplementalContext.latency},
-        `)
-
+    let logString = ''
+
+    logString += `CodeWhispererSupplementalContext:
```
justinmk3, line 224:
```suggestion
    let logString = `CodeWhispererSupplementalContext:
```

justinmk3 (approved):

PR #3763, opened 2023-08-22
codewhisperer: patch java test file regex pattern MyClassTests.java

Will-ShaoHua:
## Before
- TestMyClass.java
- MyClassTest.java

## After
- TestMyClass.java
- MyClassTest.java
- MyClassTests.java

## License

### DIFF
--- a/.changes/next-release/Bug Fix-d47847cc-bbb7-4c26-a102-82943152654a.json
+++ b/.changes/next-release/Bug Fix-d47847cc-bbb7-4c26-a102-82943152654a.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "CodeWhisperer: patch java test file name MyClassTests.java will not be captured if file is not living in a test folder"
+	"description": "CodeWhisperer: fix java test file name MyClassTests.java (plural) will not be captured if file is not living in a test folder"
 }
--- a/.changes/next-release/Bug Fix-d47847cc-bbb7-4c26-a102-82943152654a.json
+++ b/.changes/next-release/Bug Fix-d47847cc-bbb7-4c26-a102-82943152654a.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "CodeWhisperer: patch java test file name MyClassTests.java will not be captured if file is not living in a test folder"
+}
--- a/src/codewhisperer/util/supplementalContext/codeParsingUtil.ts
+++ b/src/codewhisperer/util/supplementalContext/codeParsingUtil.ts
@@ -19,7 +19,7 @@ export const utgLanguageConfigs: Record<string, utgLanguageConfig> = {
     // Java regexes are not working efficiently for class or function extraction
     java: {
         extension: '.java',
-        testFilenamePattern: /(?:Test([^/\\]+)\.java|([^/\\]+)Test\.java)$/,
+        testFilenamePattern: /(?:Test([^/\\]+)\.java|([^/\\]+)Test\.java|([^/\\]+)Tests\.java)$/,
         functionExtractionPattern:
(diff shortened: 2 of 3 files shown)


src/codewhisperer/util/supplementalContext/codeParsingUtil.ts
```diff
@@ -19,7 +19,7 @@ export const utgLanguageConfigs: Record<string, utgLanguageConfig> = {
     // Java regexes are not working efficiently for class or function extraction
     java: {
         extension: '.java',
-        testFilenamePattern: /(?:Test([^/\\]+)\.java|([^/\\]+)Test\.java)$/,
+        testFilenamePattern: /(?:Test([^/\\]+)\.java|([^/\\]+)Test\.java|([^/\\]+)Tests\.java)$/,
```
Will-ShaoHua, line 22: I still prefer a list of smaller regex like https://github.com/aws/aws-toolkit-jetbrains/blob/main/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codewhisperer/util/JavaCodeWhispererFileCrawler.kt#L25, which is easier to work with. But I will take this as follow up and don't want to bring too much changes at one shot. Any suggestion on which one is better? @justinmk3 @andrewyuq

justinmk3: For lists of "or" (`|`) cases, a list of regexes is usually a good idea. Especially if the list will continue to grow.

justinmk3 (approved):
Does this have a customer-facing impact worth mentioning in the changelog?

src/codewhisperer/util/supplementalContext/codeParsingUtil.ts
Will-ShaoHua: yea agree, it's quite error prone when working with this giant regex. I will take refactoring as a follow up.

andrewyuq: TODO comment?
and it looks like these two can be combined?
`([^/\\]+)Test\.java|([^/\\]+)Tests\.java`

Do we want to match aggressively or preservably? If it's the former do want to include any filename that contains the case-insensitive string `test`?

.changes/next-release/Bug Fix-d47847cc-bbb7-4c26-a102-82943152654a.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "CodeWhisperer: patch java test file name MyClassTests.java will not be captured if file is not living in a test folder"
```
justinmk3, line 3: Do customers know what "patch java test file" means?

```suggestion
	"description": "CodeWhisperer: java test file name MyClassTests.java will not be captured if file is not living in a test folder"
```

src/codewhisperer/util/supplementalContext/codeParsingUtil.ts
Will-ShaoHua: I prefer not combining these 2 regex, I think it's much error prone. The final format should be more similar to JB's implementation which use a list of small regex, it's also for readability.

Will-ShaoHua: I think case sensitive is fine, let's do it case sensitive for now. We can update it if needed.

.changes/next-release/Bug Fix-d47847cc-bbb7-4c26-a102-82943152654a.json
Will-ShaoHua: sgg

justinmk3 (approved):

andrewyuq (approved):

Issue #3510, opened 2023-05-30
"Scanning CloudFormation templates..." notification never goes away/finishes


ConnorChristie:
## Steps to reproduce the issue

1. Install AWS Toolkit
2. Open a large workspace with lots of files and folders.
3. Observe that on any file open, you see this "Scanning CloudFormation templates..." notification.
4. Close the notification.
5. Open a new file.
6. And the notification re-appears with the same loading bar and never finishes.

## Expected behavior

My assumption is that the file scan is just never completing because there are so many files and if so, perhaps you could utilize the user's search exclude list to ignore irrelevant files.

Or give me an option to fully disable this feature and prevent CFN template discovery / watching.

## System details (run the `AWS: About Toolkit` command)

-   OS: Linux x64 5.4.242-162.348.amzn2int.x86_64
-   Visual Studio Code version: 1.78.2
-   AWS Toolkit version: 1.74.0

justinmk3:
> Or give me an option to fully disable this feature and prevent CFN template discovery / watching.

Disabling the `aws.samcli.enableCodeLenses` should [help with that](https://github.com/aws/aws-toolkit-vscode/pull/3445) , but it doesn't fully disable this feature.

We are working on more total performance improvements: https://github.com/aws/aws-toolkit-vscode/issues/3216

ConnorChristie:
> > Or give me an option to fully disable this feature and prevent CFN template discovery / watching.
> 
> Disabling the `aws.samcli.enableCodeLenses` should [help with that](https://github.com/aws/aws-toolkit-vscode/pull/3445) , but it doesn't fully disable this feature.
> 
> We are working on more total performance improvements: #3216

Thank you but it seems that option was already disabled for me. Tried toggling it and that did not make a difference.

GPhilo:
Adding to this, it seems that whenever the scan is in progress, all other extensions stop working. This is making it impossible for me to continue using the extension, as even with `aws.samcli.enableCodeLenses` set to false, the scan keeps getting triggered and blocks my workspace for several minutes every time I open it.

justinmk3:
> whenever the scan is in progress, all other extensions stop working. ... even with `aws.samcli.enableCodeLenses` set to false

@GPhilo can you try vscode's `extensions.experimental.affinity` setting? https://github.com/aws/aws-toolkit-vscode/issues/2703#issuecomment-1202215085 :

1. Run the vscode command `Preferences: Open User Settings (JSON)`
[...]

PR #3770, opened 2023-08-24
perf(sam): do not watch files excluded in user settings

justinmk3:
## Problem

- "Scanning CloudFormation templates..." message takes a long time on big workspaces, and no way to workaround this.
    - #3510

## Solution

- When scanning for CFN/SAM templates, exclude all directories/files specified in _all_ of the vscode settings `files.exclude`, `search.exclude`, or `files.watcherExclude`.

## License

### DIFF
--- a/src/shared/fs/watchedFiles.ts
+++ b/src/shared/fs/watchedFiles.ts
@@ -7,14 +7,27 @@ import * as vscode from 'vscode'
 import { getLogger } from '../logger/logger'
 import * as pathutils from '../utilities/pathUtils'
 import * as path from 'path'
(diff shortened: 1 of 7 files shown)


src/shared/fs/watchedFiles.ts
```diff
@@ -7,14 +7,27 @@ import * as vscode from 'vscode'
...
+    '.aws-sam': true,
+    '.git': true,
+    '.svn': true,
+    '.hg': true,
+    '.rvm': true,
+    '.gem': true,
+    '.project': true,
+    node_modules: true,
+    venv: true,
+    bower_components: true,
```
justinmk3, line 29: prettier being clever :/

PR #3774, opened 2023-08-24
fix(codewhisperer): incomplete log message

justinmk3:
# Problem:
Log messages are missing useful info about the exception:

    2023-08-24 17:32:05 [ERROR]: CodeWhisperer Invocation Exception :
    2023-08-24 17:32:07 [ERROR]: CodeWhisperer Invocation Exception :
    2023-08-24 17:32:09 [ERROR]: CodeWhisperer Invocation Exception :
    2023-08-24 17:32:09 [ERROR]: CodeWhisperer Invocation Exception :

# Solution:
Fix the logger call.

**Note:** We are looking at ways to improve the logger to avoid this kind of mistake.

## License

### DIFF
--- a/src/codewhisperer/service/recommendationHandler.ts
+++ b/src/codewhisperer/service/recommendationHandler.ts
@@ -240,8 +240,7 @@ export class RecommendationHandler {
             if (latency === 0) {
                 latency = startTime !== 0 ? performance.now() - startTime : 0
             }
-            getLogger().error('CodeWhisperer Invocation Exception : ', error)
-            getLogger().verbose(`CodeWhisperer Invocation Exception : ${error}`)
+            getLogger().error('CodeWhisperer Invocation Exception : %s', (error as Error).message)
             if (isAwsError(error)) {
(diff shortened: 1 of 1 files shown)


Will-ShaoHua (approved):

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -240,8 +240,7 @@ export class RecommendationHandler {
             if (latency === 0) {
                 latency = startTime !== 0 ? performance.now() - startTime : 0
             }
-            getLogger().error('CodeWhisperer Invocation Exception : ', error)
-            getLogger().verbose(`CodeWhisperer Invocation Exception : ${error}`)
+            getLogger().error('CodeWhisperer Invocation Exception : %s', (error as Error).message)
```
justinmk3, line 243: stringifying the entire error usually isn't useful, the stacktrace is mangled in production. Therefore it's common to only mention the  `message` field .

PR #3777, opened 2023-08-25
fix: package.json invalid after 'npm run generatePackage'

nkomonen-amazon:
This is related to us enabling esModuleInterop in the tsconfig. When 'npm run generatePackage' was run it would add extra values to the package.json.

As a solution it was narrowed down that we can not do namespace imports (import * as assert from 'assert') and must instead do default imports (import assert from 'assert') for the package.json file.

## License


justinmk3:
The problem/solution format for this PR description would be:

```
Problem:
After enabling esModuleInterop in the tsconfig, 'npm run generatePackage' adds extra values to the package.json.

Solution:
Avoid namespace imports (import * as assert from 'assert').
Instead do default imports (import assert from 'assert') for the package.json file.
```

That is much clearer.

PR #3782, opened 2023-08-29
browser: logging works in the browser

nkomonen-amazon:
## Problem

In the browser logging does not work with our existing code.

## Solution

- Change how we handle logging to a File since the built in "winston file transport" for logging to a file is not browser compatible
- Update the winston transport to allow us to log to the browser console

## License

### DIFF
--- a/src/extensionWeb.ts
+++ b/src/extensionWeb.ts
@@ -7,7 +7,7 @@ import * as vscode from 'vscode'
 import { setInBrowser } from './common/browserUtils'
 
 export async function activate(context: vscode.ExtensionContext) {
-    setInBrowser(true)    
+    setInBrowser(true)
 
     vscode.window.showInformationMessage(
         'AWS Toolkit: Browser Mode Under Development. No features are currently provided',
--- a/src/extensionWeb.ts
+++ b/src/extensionWeb.ts
@@ -5,14 +5,30 @@
 
 import * as vscode from 'vscode'
 import { setInBrowser } from './common/browserUtils'
+import { activate as activateLogger } from './shared/logger/activation'
+import { initializeComputeRegion } from './shared/extensionUtilities'
+import globals from './shared/extensionGlobals'
+import { TelemetryService } from './shared/telemetry/telemetryService'
+import { TelemetryLogger } from './shared/telemetry/telemetryLogger'
 
 export async function activate(context: vscode.ExtensionContext) {
+    // This is temporary and required for the logger to run.
+    // It assumes the following exists and uses it during execution.
+    globals.telemetry = {
+        record: (event: any, awsContext?: any) => {},
+    } as TelemetryService & { logger: TelemetryLogger }
+
     setInBrowser(true)
+    await initializeComputeRegion()
 
     vscode.window.showInformationMessage(
         'AWS Toolkit: Browser Mode Under Development. No features are currently provided',
         { modal: false }
     )
+
+    // Disabling for now since this has webpack issues that will be fixed in future commits:
+    const toolkitOutputChannel = vscode.window.createOutputChannel('AWS Toolkit')
+    await activateLogger(context, toolkitOutputChannel)
 }
 
 export async function deactivate() {}
--- a/src/extensionWeb.ts
+++ b/src/extensionWeb.ts
@@ -26,7 +26,6 @@ export async function activate(context: vscode.ExtensionContext) {
         { modal: false }
     )
 
-    // Disabling for now since this has webpack issues that will be fixed in future commits:
     const toolkitOutputChannel = vscode.window.createOutputChannel('AWS Toolkit')
     await activateLogger(context, toolkitOutputChannel)
 }
--- a/src/shared/logger/activation.ts
+++ b/src/shared/logger/activation.ts
@@ -127,7 +127,7 @@ export function makeLogger(
     if (opts.useDebugConsole) {
         logger.logToDebugConsole()
     }
-    if (opts.useConsoleLog && isInBrowser()) {
+    if (opts.useConsoleLog) {
         logger.logToConsole()
     }
 
--- a/src/shared/logger/activation.ts
+++ b/src/shared/logger/activation.ts
@@ -43,6 +43,7 @@ export async function activate(
         {
             logPaths: [logUri.fsPath],
             outputChannels: [chan],
+            useConsoleLog: true,
         },
         extensionContext.subscriptions
     )
@@ -108,6 +109,7 @@ export function makeLogger(
         logPaths?: string[]
         outputChannels?: vscode.OutputChannel[]
         useDebugConsole?: boolean
+        useConsoleLog?: boolean
     },
     disposables?: vscode.Disposable[]
 ): Logger {
@@ -124,6 +126,9 @@ export function makeLogger(
     }
     if (opts.useDebugConsole) {
         logger.logToDebugConsole()
+    }
+    if (opts.useConsoleLog && isInBrowser()) {
+        logger.logToConsole()
     }
 
     if (!opts.staticLogLevel) {
--- a/src/shared/logger/sharedFileTransport.ts
+++ b/src/shared/logger/sharedFileTransport.ts
@@ -1,0 +1,110 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import TransportStream from 'winston-transport'
+import { FileSystemCommon } from '../../srcShared/fs'
+import * as vscode from 'vscode'
+import globals from '../extensionGlobals'
+import { MESSAGE } from './debugConsoleTransport'
+import { WinstonToolkitLogger } from './winstonToolkitLogger'
+
+interface LogEntry {
+    level: string
+    message: string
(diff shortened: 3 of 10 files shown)


src/browser/package.json
```diff
@@ -3,7 +3,7 @@
     "description": "Browser specific modules to separate concerns from main package.json",
     "license": "Apache-2.0",
     "dependencies": {
-        "filer": "^1.4.1",
+        "memfs": "^4.2.1",
```
nkomonen-amazon, line 6: This is temporary, we will delete this dependency eventually

src/shared/logger/activation.ts
```diff
@@ -116,8 +117,10 @@ export function makeLogger(
     // if we're outputting to an output channel, any other output doesn't need ANSI color codes since we have a better display from them in the IDE
     // don't alter logfile output for now since that should be more diagnostic. On the fence about this...
     const stripAnsi = opts.useDebugConsole || false
-    for (const logPath of opts.logPaths ?? []) {
-        logger.logToFile(logPath)
+    if (!isInBrowser()) {
```
nkomonen-amazon: A future commit in this PR removes this logic and will log to the file (but does it differently if we are in a browser)

src/shared/logger/universalFileTransport.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import winston from 'winston'
+import TransportStream from 'winston-transport'
+import { FileSystemCommon } from '../../srcShared/fs'
+import * as vscode from 'vscode'
+
+/**
+ * Inspired from: https://github.com/MarcoMedrano/winston-transport-browserconsole/blob/master/src/lib/BrowserConsole.ts
+ */
+export class UniversalFileTransport extends TransportStream {
```
justinmk3, line 14: Let's stick to "shared" where we mean "universal", to avoid multiple synonyms. But also generally, anything in /shared/ is intended to be universal so we don't need to mention it explicitly in the class name.
```suggestion
export class FileTransport extends TransportStream {
```

package.json
```diff
@@ -3607,6 +3607,7 @@
         "@types/uuid": "^9.0.1",
         "@types/vscode": "^1.65.0",
         "@types/vscode-webview": "^1.57.1",
+        "@types/webpack-env": "^1.18.1",
```
justinmk3, line 3610: is this for `webpack.EnvironmentPlugin` or something else?

src/shared/logger/sharedFileTransport.ts
```diff
@@ -0,0 +1,110 @@
...
+ *   it would not have any of the previous logs in it.
+ * - The `next()` method is intended to signal that the current log is done being
+ *   written to the file but that is not working by the looks of it.
+ * Temporary Solution:
+ * - I created a buffer that buffers incoming logs and batch writes them after
+ *   a certain timeout. This seems to improve things
+ * - Also the current append implementation is inefficient due to the underlying api
+ *   (reads whole file, concatenates, then writes) so buffering is beneficial here.
+ */
+export class SharedFileTransport extends TransportStream {
```
justinmk3, line 38: Why not just `FileTransport`? I don't think we want to prefix every class with "shared"/browser/node?

nkomonen-amazon: There is a File transport built in to winston already, so I wanted to put "shared" in this case to differentiate it

src/extensionWeb.ts
```diff
@@ -5,14 +5,30 @@
...
 export async function activate(context: vscode.ExtensionContext) {
-    setInBrowser(true)    
+    // This is temporary and required for the logger to run.
+    // It assumes the following exists and uses it during execution.
+    globals.telemetry = {
+        record: (event: any, awsContext?: any) => {},
+    } as TelemetryService & { logger: TelemetryLogger }
+
+    setInBrowser(true)
+    await initializeComputeRegion()
```
justinmk3, line 22: 🎉

src/extensionWeb.ts
```diff
@@ -5,14 +5,30 @@
...
 
     vscode.window.showInformationMessage(
         'AWS Toolkit: Browser Mode Under Development. No features are currently provided',
         { modal: false }
     )
+
+    // Disabling for now since this has webpack issues that will be fixed in future commits:
+    const toolkitOutputChannel = vscode.window.createOutputChannel('AWS Toolkit')
+    await activateLogger(context, toolkitOutputChannel)
```
justinmk3, line 31: what is the comment referring to? what's being "disabled"? outputchannel works in browser

src/shared/logger/activation.ts
```diff
@@ -125,6 +127,9 @@ export function makeLogger(
     if (opts.useDebugConsole) {
         logger.logToDebugConsole()
     }
+    if (opts.useConsoleLog && isInBrowser()) {
```
justinmk3, line 130: is the isInBrowser check still needed? if something sets `useConsoleLog` why ignore it?

```suggestion
    if (opts.useConsoleLog) {
```

src/test/testUtil.ts
```diff
@@ -68,6 +68,13 @@ export async function createTestWorkspaceFolder(name?: string): Promise<vscode.W
     }
 }
 
+export async function createTestFile(fileName: string): Promise<vscode.Uri> {
+    const tempFolder = await makeTemporaryToolkitFolder()
```
justinmk3, line 72: this will let the test teardown delete the temp dir:

    testTempDirs.push(tempFolder)

src/test/shared/logger/sharedFileTransport.test.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import vscode from 'vscode'
+import assert from 'assert'
+import { SharedFileTransport, flushIntervalMillis } from '../../../shared/logger/sharedFileTransport'
+import { FileSystemCommon } from '../../../srcShared/fs'
+import { stub, SinonStub } from 'sinon'
+import { MESSAGE } from '../../../shared/logger/debugConsoleTransport'
```
justinmk3, line 11: off-topic: I wonder if there's a way we can hide this `MESSAGE` symbol from callers, it's a strange interface.

src/test/shared/logger/consoleLogLogger.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        )
+        await untilLogged
+        assert.strictEqual(fakeConsole.log.callCount, 1)
+        assert.strictEqual(fakeConsole.log.getCall(0).args[0], 'myMessageFormatted')
+        assert.strictEqual(next.callCount, 1)
+    })
+
+    function createBrowserFakeConsole() {
+        return {
+            info: stub(),
```
justinmk3, line 46: off-topic: if our logger were queryable [like our telemetry](https://github.com/aws/aws-toolkit-vscode/blob/a5031452fb1333429a6dcf26da19492e432c2612/src/test/testUtil.ts#L198) then we wouldn't need to mock or fake it, instead we could just query it in tests.

I think winston supports queries but not sure if we are leveraging it.

src/test/shared/logger/sharedFileTransport.test.ts
nkomonen-amazon: The `MESSAGE` symbol really confused me at first.

This doc says that 2 separate symbols, even if created with the same string cannot be equal: [see the 2nd code snippet](https://www.typescriptlang.org/docs/handbook/symbols.html). But then it works when the winston backend injects their own property with their [own `MESSAGE` symbol](https://github.com/winstonjs/triple-beam/blob/master/index.js#L23).

I'm not sure if we can change, seems like a built in thing: 
"[Symbol.for('message'): complete string message set by "finalizing formats"](https://github.com/winstonjs/winston#streams-objectmode-and-info-objects)"

src/shared/logger/activation.ts
nkomonen-amazon: We only used the console log for testing previously, but yeah this makes sense we should just enable it permanently for now.

justinmk3 (approved):

src/shared/logger/activation.ts
```diff
@@ -43,6 +43,7 @@ export async function activate(
         {
             logPaths: [logUri.fsPath],
             outputChannels: [chan],
+            useConsoleLog: true,
```
justinmk3, line 46: @nkomonen-amazon why does this default to true? does that change the nodejs behavior compared to before?

nkomonen-amazon: Yeah this will have nodejs log to the vscode Developer Tools console as well

justinmk3: some reasons to default to `useConsoleLog=false`

- it matches the previous behavior (for nodejs), less risk. (set to true when in browser-mode.)
- in cloud9 this will cause duplicate logs because cloud9 already forwards our logs to the browser console. 
- in the integ tests CI logs, it looks like this is causing duplicate messages:
  ```
  2023-09-07 13:42:23 [VERBOSE]: stderr: 
  Fetching public.ecr.aws/sam/build-java8.al2:latest-x86_64 Docker container image....
  2023-09-07 13:42:23 [VERBOSE]: stderr: 
  Fetching public.ecr.aws/sam/build-java8.al2:latest-x86_64 Docker container image....
  ```

PR #3785, opened 2023-08-29
refactor: Various browser changes

nkomonen-amazon:
See each commit for description on each specific change

## License


src/shared/logger/activation.ts
```diff
@@ -168,23 +168,36 @@ async function createLogWatcher(logFile: vscode.Uri): Promise<vscode.Disposable>
...
-            )
-            watcher.close()
-        }
-        checking = false
-    })
+    const disposables: vscode.Disposable[] = []
+
+    // Notify if log file is renamed
+    disposables.push(
+        vscode.workspace.onDidRenameFiles((e: vscode.FileRenameEvent) => {
```
justinmk3, line 175: Nice change. Not a blocker for this PR, but we should consider dropping the log file watcher entirely. Instead just handle the failure if writing to the logfile fails. No need to create a watcher for this edge case...

src/browser/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export function setInBrowser(value: boolean) {
+    inBrowser = value
+    return isInBrowser()
+}
+
+/**
+ * Return true if we are running in the browser, false otherwise.
+ */
+export function isInBrowser() {
```
justinmk3, line 19: seems like this shoudl live in /shared/, not /browser/. Because non-browser code needs to call this.

justinmk3 (approved):
LGTM. In favor of merging this and continuing in followup PRs or on `feature/web`

src/extensionWeb.ts
```diff
@@ -4,17 +4,17 @@
...
-        'AWS Toolkit: Browser Mode Under Development. No features are currently provided',
-        { modal: true }
-    )
+    setInBrowser(true)
+    
+    // This causes the Extension Browser tests to fail.
+    // See the comment in src/testBrowser/testRunner.ts#stubVscodeWindow() for more info.
+    // vscode.window.showInformationMessage(
+    //     'AWS Toolkit: Browser Mode Under Development. No features are currently provided',
+    //     { modal: true }
```
justinmk3, line 16: do the tests work if it's not modal?
```suggestion
    //     { modal: false }
```

src/browser/utils.ts
nkomonen-amazon: I was going through `vscode` source code and they structure their project like:

Example: https://github.com/microsoft/vscode/tree/main/src/vs/platform/request)
```
  {WHATEVER TOPIC}/
  - browser/
  - node/
  - common/
```
and this structure appears throughout the project in each topic folder.

---

So related to this comment, what do you think about creating a `src/common/` folder and then add in `browserUtils.ts`. So now we start using the `common` keyword for browser+node compatible code. And `shared` would instead be `utils` since "shared" may be confused with browser+node compatibility.

We'd document this new structure of course

nkomonen-amazon: Sounds good to me!

src/shared/logger/activation.ts
nkomonen-amazon: Sounds good, created IDE-11538

src/extensionWeb.ts
nkomonen-amazon: Oh nice, didn't realize this was the core of the issue. Fixed it

src/common/browserUtils.ts
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+let inBrowser = false
+/** Set the value of if we are in the browser. Impacts {@link isInBrowser}. */
+export function setInBrowser(value: boolean) {
+    inBrowser = value
+}
+/** Return true if we are running in the browser, false otherwise. */
+export function isInBrowser() {
```
justinmk3, line 12: These "in a browser?" functions could live in `src/shared/extensionUtilities.ts`.  Compare `getIdeProperties()` which lives in `extensionUtil.ts`.

justinmk3: Also, [we don't need /common/](https://github.com/aws/aws-toolkit-vscode/pull/3785#discussion_r1309314009) if we have /shared/, right? Instead, we can incrementally _move_ browser-only or node-only things _out of_ /shared/, so eventually /shared/ will have only "platform-agnostic" code in it.

nkomonen-amazon: The issue I found was that with any import of a /shared/ module one of the transitive dependencies caused issues and it was a hassle to chase it down. 

- it will become a whack-a-mole if we import something from shared but some separate code on the module causes issues and we need to find all node/browser specific code and move it out
- On the other hand, if we use some code from `src/shared/` and realize it is platform agnostic, we can move it to `src/common`

justinmk3 (approved):
🚀

PR #3792, opened 2023-08-31
Enable SendTelemetryEvent for all SSO users

andrewyuq:
This includes calling the API with three different types of events: codeScan, codePercentage and userTriggerDecision.

Have tested to make sure the requests successfully reached to the service.
## License


src/codewhisperer/util/codewhispererStates.ts
```diff
@@ -0,0 +1,29 @@
...
+    sdkApiCallStartTime = 0
+    invokeSuggestionStartTime = 0
+
+    public static get instance() {
+        return (this.#instance ??= new CodeWhispererStates())
+    }
+
+    setFetchCredentialStart() {
+        if (this.fetchCredentialStartTime === 0 && this.invokeSuggestionStartTime !== 0) {
+            this.fetchCredentialStartTime = performance.now()
```
justinmk3, line 20: `performance` is not available in node 14 (test in cloud9-codecatalyst)

https://github.com/aws/aws-toolkit-vscode/blob/a5031452fb1333429a6dcf26da19492e432c2612/src/codewhisperer/commands/startSecurityScan.ts#L36

src/codewhisperer/util/codewhispererStates.ts
```diff
@@ -0,0 +1,29 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export class CodeWhispererStates {
```
justinmk3, line 6: should `vsCodeState` be merged into this class? https://github.com/aws/aws-toolkit-vscode/blob/a5031452fb1333429a6dcf26da19492e432c2612/src/codewhisperer/models/model.ts#L33

there is also `globalStateUtil.ts`

andrewyuq: This `CodeWhispererStates` is aiming for a CW session(per trigger) scope and this is only part of the progress to group some of the variables that have the same scope together. Later on we will call `CodeWhispererStates.reset()` for every CW session in which case I think it's more appropriate to separate the these states vs the `vsCodeState` because they are global scopes. I will change it class name to be `CodeWhispererSessionStates` to make it clear.

src/codewhisperer/util/codewhispererStates.ts
andrewyuq: Fixing

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -486,18 +551,81 @@ export class TelemetryHelper {
...
         })
-        this.resetClientComponentLatencyTime()
+    }
+
+    public sendUserModificationEvent(suggestion: AcceptedSuggestionEntry, percentage: number) {
+        client
+            .sendTelemetryEvent({
+                telemetryEvent: {
+                    userModificationEvent: {
+                        sessionId: suggestion.sessionId ? suggestion.sessionId : 'undefined',
```
leigaol, line 580: Is it necessary to send it when session id is undefined?

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -421,10 +498,10 @@ export class TelemetryHelper {
     }
 
     public resetClientComponentLatencyTime() {
-        this.invokeSuggestionStartTime = 0
-        this.sdkApiCallStartTime = 0
+        CodeWhispererSessionStates.instance.invokeSuggestionStartTime = 0
+        CodeWhispererSessionStates.instance.sdkApiCallStartTime = 0
         this.sdkApiCallEndTime = 0
-        this.fetchCredentialStartTime = 0
```
leigaol, line 427: Thanks for simplifying the states. Any chance to move `sdkApiCallEndTime = 0` inside `CodeWhispererSessionStates.instance` ?

src/codewhisperer/util/telemetryHelper.ts
andrewyuq: no and in fact we shouldn't send if sessionId is undefined.

src/codewhisperer/util/telemetryHelper.ts
andrewyuq: Yea definitely, simple not a required change for this PR.

src/codewhisperer/util/telemetryHelper.ts
andrewyuq: ```suggestion.sessionId ? suggestion.sessionId : 'undefined'``
I don't think this is even needed. sessionId will in the worst be an emptpy string but not undefined.

andrewyuq: Another thing is sessionId and requestId don't make sense for the userModificationEvent, and this function is also not get called currently, I will simply remove this method.

src/codewhisperer/util/codeWhispererSessionStates.ts
```diff
@@ -3,16 +3,18 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-export class CodeWhispererStates {
-    static #instance: CodeWhispererStates
+const performance = globalThis.performance ?? require('perf_hooks').performance
+
+export class CodeWhispererSessionStates {
```
justinmk3, line 8: Is "states" actually needed in the name?
```suggestion
export class CodeWhispererSession {
```

leigaol (approved):

src/codewhisperer/util/codeWhispererSession.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    setFetchCredentialStart() {
+        if (this.fetchCredentialStartTime === 0 && this.invokeSuggestionStartTime !== 0) {
+            this.fetchCredentialStartTime = performance.now()
+        }
+    }
+
+    setSdkApiCallStart() {
+        if (this.sdkApiCallStartTime === 0 && this.fetchCredentialStartTime !== 0) {
```
leigaol, line 27: Does this mean `setSdkApiCallStart` can only be invoked once in this singleton? The second time when I invoke `setSdkApiCallStart()`, `this.sdkApiCallStartTime === 0` is no longer true, but I do need a sdk api call start time. 

Rest of the changes looks good.

andrewyuq: It should be invoked only once in the beginning of a trigger.
The current logic is somehow complicated for me to follow, making me worrying that we may reset the this state when we still need it, espeically when we delay sending some of the metrics at the start of next trigger.

justinmk3 (approved):

src/codewhisperer/client/codewhisperer.ts
```diff
@@ -189,6 +191,16 @@ export class DefaultCodeWhispererClient {
             .listCodeScanFindings(request as CodeWhispererClient.ListCodeScanFindingsRequest)
             .promise()
     }
+
+    public async sendTelemetryEvent(request: SendTelemetryEventRequest) {
+        if (!AuthUtil.instance.isValidEnterpriseSsoInUse()) {
+            return
+        }
+        const response = await (await this.createUserSdkClient()).sendTelemetryEvent(request).promise()
+        getLogger().debug(`send telemetry event requestID: ${response.$response.requestId}`)
```
justinmk3, line 200:
```suggestion
        getLogger().debug(`codewhisperer: sendTelemetryEvent requestID: ${response.$response.requestId}`)
```

andrewyuq:
There's another bug that we found it would send negative numbers for the latency field in the `userTriggerDecision` payload due to an incorrect global state manipulation. I am working on it and will post a fix before I merge this one.

src/codewhisperer/util/codeWhispererSession.ts
leigaol: >The current logic is somehow complicated for me to follow
Understood. This is the tech debt we had a long time ago.

src/codewhisperer/util/codeWhispererSession.ts
```diff
@@ -0,0 +1,68 @@
...
+        const nonBlankLines = recommendation.content.split('\n').filter(line => line.trim() !== '').length
+        this.completionTypes.set(index, nonBlankLines > 1 ? 'Block' : 'Line')
+    }
+
+    getCompletionType(index: number): CodewhispererCompletionType {
+        return this.completionTypes.get(index) || 'Line'
+    }
+}
+
+export const session = CodeWhispererSession.instance
```
justinmk3, line 68: Does this call `instance()` at moduel-load time? That can lead to strange order-of-operations issues, because the module load order will affect when `instance()` is called. That can be non-obvious to troubleshoot.

Not a blocker for this PR, but as a followup, this could be fixed by converting `session` to a function.

src/codewhisperer/util/authUtil.ts
```diff
@@ -91,6 +85,12 @@ export class AuthUtil {
         return this.secondaryAuth.activeConnection
     }
 
+    public get startUrl(): string | undefined {
+        // Reformat the url to remove any trailing '/' and `#`
+        // e.g. https://view.awsapps.com/start/# will become https://view.awsapps.com/start
+        return isSsoConnection(this.conn) ? this.reformatStartUrl(this.conn?.startUrl) : undefined
```
justinmk3, line 91: Future followup: We need to stop adding special things to `codewhisperer/authUtil.ts`. Most auth-related code should live in the `shared/.../auth.ts`.

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -202,19 +200,30 @@ export class TelemetryHelper {
...
         const aggregatedEvent = this.aggregateUserDecisionByRequest(events, requestId, sessionId)
         if (aggregatedEvent) {
             this.sessionDecisions.push(aggregatedEvent)
         }
 
+        let acceptedRecommendationContent
+        if (acceptIndex != -1 && recommendations[acceptIndex] != undefined) {
+            acceptedRecommendationContent = recommendations[acceptIndex].content
+        } else {
+            acceptedRecommendationContent = ''
```
justinmk3, line 217: using a ternary would allow this to be `const` and is generally more readable and less verbose

justinmk3 (approved):

PR #3793, opened 2023-09-01
refactor(cloudwatch): simplify saveCurrentLogDataContent()

justinmk3:
Problem:
saveCurrentLogDataContent() re-renders the cloudwatch logs document from the "registry", even though this command only makes sense on an existing, open editor document. Also even in the case where a document isn't open, it should just be opened instead of using the low-level registry mechanics to get the file contents.

Solution:
- Use `vscode.window.activeTextEditor?.document.getText()` to get the file contents.
- Make the test more meanginful by opening an editor document instead of passing a URI.

## License

### DIFF
--- a/src/cloudWatchLogs/commands/saveCurrentLogDataContent.ts
+++ b/src/cloudWatchLogs/commands/saveCurrentLogDataContent.ts
@@ -12,7 +12,7 @@ import { isLogStreamUri, parseCloudWatchLogsUri } from '../cloudWatchLogsUtils'
 import { telemetry, CloudWatchResourceType, Result } from '../../shared/telemetry/telemetry'
 import { FileSystemCommon } from '../../srcShared/fs'
 
-/** Saves the current "aws-cwl:" document. */
+/** Prompts the user to select a file location to save the currently visible "aws-cwl:" document to. */
 export async function saveCurrentLogDataContent(): Promise<void> {
     let result: Result = 'Succeeded'
     let resourceType: CloudWatchResourceType = 'logStream' // Default to stream if it fails to find URI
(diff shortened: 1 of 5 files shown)


src/cloudWatchLogs/commands/saveCurrentLogDataContent.ts
```diff
@@ -49,8 +44,8 @@ export async function saveCurrentLogDataContent(uri: vscode.Uri | undefined, reg
 
         if (selectedUri) {
             try {
-                await fs.writeFile(selectedUri.fsPath, content)
-                // TODO: Open file and close virtual doc? Is this possible?
+                const bytes = new TextEncoder().encode(content)
+                await vscode.workspace.fs.writeFile(selectedUri, bytes)
```
nkomonen-amazon, line 48: Use the new fs class which does this under the hood

```suggestion
                await FileSystemCommon.instance.writeFile(selectedUri, content)
```
```

File: src/cloudWatchLogs/commands/saveCurrentLogDataContent.ts [hunk 2]
```diff
@@ -7,29 +7,25 @@ import * as vscode from 'vscode'
...
-import * as fs from 'fs-extra'
 import { SystemUtilities } from '../../shared/systemUtilities'
 import { isLogStreamUri, parseCloudWatchLogsUri } from '../cloudWatchLogsUtils'
-import { LogDataRegistry } from '../registry/logDataRegistry'
 import { telemetry, CloudWatchResourceType, Result } from '../../shared/telemetry/telemetry'
-import { generateTextFromLogEvents } from '../document/textContent'
+import { FileSystemCommon } from '../../srcShared/fs'
 
-export async function saveCurrentLogDataContent(uri: vscode.Uri | undefined, registry: LogDataRegistry): Promise<void> {
+/** Saves the current "aws-cwl:" document. */
```
nkomonen-amazon, line 15: Could we give more detail with something like:

```suggestion
/** Prompts the user to select a file location to save the currently visible "aws-cwl:" document to. */
```

nkomonen-amazon (approved):

PR #3799, opened 2023-09-05
codewhisperer: add supplemental context strategy telemetry field

Will-ShaoHua:
## Problem
We are already able to analyze acceptance rate on either CrossFile or UTG or an aggregated number. Now we need more granular metrics for UTG `resolve source file by file name` vs. `resolve source file by file content` to help us understanding deeper the effectiveness of UTG, thus adding a new field `codewhispererSupplementalContextStrategyId` field representing the strategies being used for Crossfile and UTG.

* Crossfile
    * OpenTabs + BM25

* UTG
   * Resolve by file name (regex)
   * Resolve by file content  

there are 4 values for now
`Empty`, `ByName`, `ByContent`, `OpenTabs_BM25`

## License

### DIFF
--- a/src/codewhisperer/util/telemetryHelper.ts
+++ b/src/codewhisperer/util/telemetryHelper.ts
@@ -299,6 +299,7 @@ export class TelemetryHelper {
             codewhispererSupplementalContextTimeout: supplementalContextMetadata?.isProcessTimeout,
             codewhispererSupplementalContextIsUtg: supplementalContextMetadata?.isUtg,
             codewhispererSupplementalContextLength: supplementalContextMetadata?.contentsLength,
+            // eslint-disable-next-line @typescript-eslint/naming-convention
             codewhispererSupplementalContextStrategyId: supplementalContextMetadata?.strategy,
         }
         telemetry.codewhisperer_userTriggerDecision.emit(aggregated)
--- a/src/codewhisperer/util/telemetryHelper.ts
+++ b/src/codewhisperer/util/telemetryHelper.ts
@@ -299,6 +299,7 @@ export class TelemetryHelper {
             codewhispererSupplementalContextTimeout: supplementalContextMetadata?.isProcessTimeout,
             codewhispererSupplementalContextIsUtg: supplementalContextMetadata?.isUtg,
             codewhispererSupplementalContextLength: supplementalContextMetadata?.contentsLength,
+            codewhispererSupplementalContextStrategyId: supplementalContextMetadata?.strategy,
         }
         telemetry.codewhisperer_userTriggerDecision.emit(aggregated)
         this.prevTriggerDecision = this.getAggregatedUserDecision(this.sessionDecisions)
--- a/src/codewhisperer/util/telemetryHelper.ts
+++ b/src/codewhisperer/util/telemetryHelper.ts
@@ -299,7 +299,7 @@ export class TelemetryHelper {
             codewhispererSupplementalContextTimeout: supplementalContextMetadata?.isProcessTimeout,
             codewhispererSupplementalContextIsUtg: supplementalContextMetadata?.isUtg,
             codewhispererSupplementalContextLength: supplementalContextMetadata?.contentsLength,
-            // eslint-disable-next-line @typescript-eslint/naming-convention
+            // eslint-disable-next-line id-length
             codewhispererSupplementalContextStrategyId: supplementalContextMetadata?.strategy,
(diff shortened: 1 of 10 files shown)


src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -40,6 +45,125 @@
         }
     ],
     "metrics": [
+        {
```
Will-ShaoHua, line 48: will revert vscodeTelemetry.json once common definition is updated

andrewyuq: Do we want to do the common repo update first and then update the lib dep version?

andrewyuq: >once common definition is updated
If this is ongoing then it's good.

Will-ShaoHua: yep will update common repo first then bump the telemetry version here instead of overwriting vscode.telemtry

src/test/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -169,7 +169,10 @@ describe('crossFileContextUtil', function () {
 
                 const actual = await crossFile.fetchSupplementalContextForSrc(editor, fakeCancellationToken)
 
-                assert.ok(actual?.length !== undefined && actual.length === 0)
+                assert.ok(
+                    actual?.supplementalContextItems.length !== undefined &&
+                        actual.supplementalContextItems.length === 0
+                )
```
justinmk3, line 175: nit, I think this works and is equivalent?

```suggestion
                assert.ok(actual?.supplementalContextItems.length === 0)
```

or if that doesn't work,

```suggestion
                assert.ok(actual !== undefined && actual.supplementalContextItems.length === 0)
```

andrewyuq (approved):

src/test/codewhisperer/util/crossFileContextUtil.test.ts
Will-ShaoHua: yea make sense

justinmk3:
Lint is failing. Why are these names so long?

      302:13  error  Identifier name 'codewhispererSupplementalContextStrategyId' is too long (> 40)  id-length

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -299,6 +299,7 @@ export class TelemetryHelper {
             codewhispererSupplementalContextTimeout: supplementalContextMetadata?.isProcessTimeout,
             codewhispererSupplementalContextIsUtg: supplementalContextMetadata?.isUtg,
             codewhispererSupplementalContextLength: supplementalContextMetadata?.contentsLength,
+            codewhispererSupplementalContextStrategyId: supplementalContextMetadata?.strategy,
```
justinmk3, line 302: Since this name is from the telemetry definitions we will have to ignore the lint rule in this case.
```suggestion
            // eslint-disable-next-line @typescript-eslint/naming-convention
            codewhispererSupplementalContextStrategyId: supplementalContextMetadata?.strategy,
```

In the future, such long names should not be used. There really is no reason that names need to be that long.

Will-ShaoHua: yea https://github.com/aws/aws-toolkit-common/commit/a826622c0d6544cb2da0abe87f52e7105cb2f971#diff-f9ad596fbedec6dc08aa26f65b8a0f9358ae2be13476b00a82557f7d7125c568R609

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -299,6 +299,7 @@ export class TelemetryHelper {
             codewhispererSupplementalContextTimeout: supplementalContextMetadata?.isProcessTimeout,
             codewhispererSupplementalContextIsUtg: supplementalContextMetadata?.isUtg,
             codewhispererSupplementalContextLength: supplementalContextMetadata?.contentsLength,
+            // eslint-disable-next-line @typescript-eslint/naming-convention
```
justinmk3, line 302:
```suggestion
            // eslint-disable-next-line id-length
```

src/codewhisperer/util/telemetryHelper.ts
justinmk3: oops should be `eslint-disable-next-line id-length`. i pushed a fix

PR #3805, opened 2023-09-06
Getting Started for CodeWhisperer

LaxmanReddyAileni:
## Problem
Customers' are unfamiliar with best Practices of CodeWhisperer to get suggestions.
Currently, CodeWhisperer offers a text-heavy instructional markdown file in the IDEs after user signing in for the first time. 

## Solution

This project is to provide user a quick and interactive way to learn about the CodeWhisperer.

In this pull request, I implemented the following features

1. "Learn" Child Node under CodeWhisperer Node.
image
2. Getting Started Page will open for first time Sign-in users with banner section and can also access from "Learn" node without banner section.
image

3. Getting Started Page UI
    - Header section.
    image
    - Generate suggestions with examples for popular languages like TypeScript, JavaScript, Python, Java and C#. According to the language selected, example files are created with default code and automatically saved every time.
image
image

    - Code Scan section.
image
    - Shortcuts section with Keyboard shortcuts editor
image
    - Resources section with Workshop, Documentation and Feedback to CodeWhisperer.
image
   
4. Feedback Page for CodeWhisperer.
image

## License


src/testFixtures/workspaceFolder/.vscode/settings.json
justinmk3, line 1: was this file intentionally added?

src/codewhisperer/vue/feedback/submitFeedbackCW.vue
justinmk3, line 1: Please don't copy-paste large amounts of code. Can you instead parameterize the existing `src/feedback/vue/submitFeedback.vue` and related files?

justinmk3 (changes requested):
redundant code

src/testFixtures/workspaceFolder/.vscode/settings.json
nkomonen-amazon: This file is added when running `npm run test` on the CLI

LaxmanReddyAileni: Yes, 
It was added by running test command(npm run test) in VSC Terminal and I removed that file in latest commit.
TY.

package.json
```diff
@@ -3452,103 +3452,110 @@
...
-            "aws-lambda-function": {
+            "aws-iot-thing": {
                 "description": "AWS Contributed Icon",
                 "default": {
                     "fontPath": "./resources/fonts/aws-toolkit-icons.woff",
                     "fontCharacter": "\\f1b7"
                 }
             },
-            "aws-s3-bucket": {
+            "aws-lambda-function": {
```
leigaol, line 3518: Is it possible to avoid making this change?

LaxmanReddyAileni: I added a **svg** icon for the"Learn" child node UX thats causing change in package.json.

src/codewhisperer/models/constants.ts
```diff
@@ -123,7 +123,9 @@ export const welcomeCodeWhispererReadmeFileSource = 'resources/markdown/WelcomeT
 
 export const welcomeCodeWhispererCloud9Readme = 'resources/markdown/WelcomeToCodeWhispererCloud9.md'
 
-export const welcomeMessageKey = 'CODEWHISPERER_WELCOME_MESSAGE'
+export const welcomeMessageKeyPrevious = 'CODEWHISPERER_WELCOME_MESSAGE'
```
leigaol, line 126: Can you add a comment, explain that `welcomeMessageKeyPrevious` is a key for the welcome page of previous v1.91.0, `welComeMessageKeyShown` is a new key for the new getting started experience, it is used to track if user has seen the getting started page or not.

src/codewhisperer/activation.ts
```diff
@@ -64,6 +65,28 @@ export async function activate(context: ExtContext): Promise<void> {
         await enableDefaultConfigCloud9()
     }
 
+    //This logic is to show the Getting Started Page only once for the first time SignIn Users
+    let start = ''
+
+    const hasShownWelcomeMsgBefore = get(
```
leigaol, line 71: Can you move line 68 to line 88 to a function named `tryShowGettingStartedPage`?

Then in this function, clearly call out that `hasShownWelcomeMsgBefore` is for users who have signed in before v1.91.0. Make it clear that `hasShownWelcomeMsgBeforeToExistingUsers` is for users who has seen or not see the new getting started experience since v1.92.0

src/codewhisperer/vue/backend.ts
```diff
@@ -0,0 +1,155 @@
...
+        vscode.workspace.openTextDocument(localFilePath).then(doc => {
+            vscode.window.showTextDocument(doc, vscode.ViewColumn.Active).then(editor => {
+                const endOfDocument = new vscode.Position(doc.lineCount - 1, doc.lineAt(doc.lineCount - 1).text.length)
+                editor.selection = new vscode.Selection(endOfDocument, endOfDocument)
+            })
+        })
+    }
+
+    //This function returns the OS type of the machine used in Shortcuts and Generate Suggestion Sections
+    public getOSType() {
```
leigaol, line 70: Linux is one of the supported platforms

src/codewhisperer/models/constants.ts
LaxmanReddyAileni: Updated accordingly.

src/codewhisperer/vue/backend.ts
LaxmanReddyAileni: As per implementation apart from **Mac** everything is shown as same Key

For Mac: Option+C
Rest : Alt+C

```
 <div class="generateSuggestionTabMachine">
                                    <div class="col2" v-if="osState === 'Mac'">Option</div>
                                    <div class="col2" v-else>Alt</div>
                                    <div class="generateSuggestionTabMachineText">+</div>
                                    <div class="col2">C</div>
                                </div>
```

src/codewhisperer/activation.ts
LaxmanReddyAileni: Updated accordingly.

src/codewhisperer/vue/backend.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+    //This function is called when the extension is activated to check whether is it the first time the user is using the extension or not
+    public async showAtStartUp(): Promise<string | void> {
+        return this.start
+    }
+
+    private isFileSaved: boolean = false
+    private getLocalFilePath(fileName: string): string {
+        //This will store the files in the global storage path of VSCode
+        return path.join(globals.context.globalStoragePath, fileName)
```
nkomonen-amazon, line 31: `globalStoragePath` is deprecated, use `globalStorageUri` instead. Also since the `globalStorageUri` folder is globally used you should created a subfolder `return path.join(globals.context.globalStorageUri, 'codeWhispererStartup', fileName)` so all of this code is contained to a folder.

src/codewhisperer/vue/backend.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    // This function opens Python/JavaScript/C# file in the editor.
+    async openFile(name: string[]): Promise<void> {
```
nkomonen-amazon, line 35: Try this for better types:
```suggestion
    async openFile(name: [fileName: string, fileContent: string]): Promise<void> {
```

package.json
leigaol: will this affect toolkit-team's code path other than CodeWhisperer? If you are adding resources, it should only show as addition in the PR diff.

src/codewhisperer/vue/backend.ts
```diff
@@ -0,0 +1,153 @@
...
+        const fileName = name[0]
+        const fileContent = name[1]
+
+        const localFilePath = this.getLocalFilePath(fileName)
+
+        if (fs.existsSync(localFilePath) && this.isFileSaved) {
```
nkomonen-amazon, line 41: For `fs` related code try and use functions like `FileSystemCommon.instance.fileExists()` where possible. FileSystemCommon enables better support for the extension running in the Browser or with Node.js on the desktop. But not all functionality is supported by `FileSystemCommon` yet, so use it when you can.

src/codewhisperer/vue/backend.ts
```diff
@@ -0,0 +1,153 @@
...
+        vscode.workspace.openTextDocument(localFilePath).then(doc => {
+            vscode.window.showTextDocument(doc, vscode.ViewColumn.Active).then(editor => {
+                const endOfDocument = new vscode.Position(doc.lineCount - 1, doc.lineAt(doc.lineCount - 1).text.length)
+                editor.selection = new vscode.Selection(endOfDocument, endOfDocument)
+            })
+        })
+    }
+
+    //This function returns the OS type of the machine used in Shortcuts and Generate Suggestion Sections
+    public getOSType() {
```
nkomonen-amazon, line 70: Try and create an explicit `type` for what this returns and it will make things easier in the future.

`export type OSType = 'Mac' | 'RestOfOS'` and then you can import this type in to your `vue.js` code if it makes sense.

  ```suggestion
    public getOSType(): OSType {
```

File: src/codewhisperer/activation.ts [hunk 9]
```diff
@@ -202,6 +206,39 @@ export async function activate(context: ExtContext): Promise<void> {
...
+        const hasShownWelcomeMsgBeforeToExistingUsers = get(
+            /**
+             *  welcomeMessageKeyShown is a new key for the new getting started experience, it is used to track if first time signIn new user/ existing user has seen the getting started page or not.
+             */
+            CodeWhispererConstants.welcomeMessageKeyShown,
+            context.extensionContext.globalState
+        )
+
+        if (!hasShownWelcomeMsgBeforeToExistingUsers) {
+            start = hasShownWelcomeMsgBefore ? 'ExistingUser' : 'startUpOnly'
```
andrewyuq, line 232: s/startUpOnly/NewUser ?

src/codewhisperer/activation.ts
```diff
@@ -202,6 +206,39 @@ export async function activate(context: ExtContext): Promise<void> {
...
+         * hasShownWelcomeMsgBefore defines whether user is existing user or new user.
+         */
+        const hasShownWelcomeMsgBefore = get(
+            /** welcomeMessageKeyPrevious is a key for the previous welcome page of version v1.91.0,*/
+            CodeWhispererConstants.welcomeMessageKeyPrevious,
+            context.extensionContext.globalState
+        )
+
+        //hasShownWelcomeMsgBeforeToExistingUsers defines whether getting started page has been shown to existing users or new users since v1.92.0.
+        const hasShownWelcomeMsgBeforeToExistingUsers = get(
```
andrewyuq, line 223: s/hasShownWelcomeMsgBeforeToExistingUsers/hasShownNewWelcomeMsg

src/codewhisperer/models/constants.ts
andrewyuq: `oldWelcomeMessageKey`

src/codewhisperer/models/constants.ts
```diff
@@ -122,8 +122,15 @@ export const licenseFilter = 'CodeWhisperer suggestions were filtered due to ref
...
+/**
+ * welcomeMessageKeyPrevious is a key for the previous welcome page of version v1.91.0,
+ */
+export const welcomeMessageKeyPrevious = 'CODEWHISPERER_WELCOME_MESSAGE'
 
-export const welcomeMessageKey = 'CODEWHISPERER_WELCOME_MESSAGE'
+/**
+ *  welcomeMessageKeyShown is a new key for the new getting started experience, it is used to track if first time signIn new user/ existing user has seen the getting started page or not.
+ */
+export const welcomeMessageKeyShown = 'CODEWHISPERER_WELCOME_MESSAGE_SHOWN'
```
andrewyuq, line 133: `newWelcomeMessageKey`

src/testFixtures/workspaceFolder/.vscode/settings.json
andrewyuq, line 1: Can we remove this file? Doesn't look like it should be here.

src/codewhisperer/vue/backend.ts
```diff
@@ -0,0 +1,153 @@
...
+// export type codewhispererLanguage = 'python' | 'javascript' | 'csharp'
+//List of all events that are emitted from the webview of CodeWhisperer
+export type CodeWhispererUiClick =
+    | 'codeWhisperer_Resources_Documentation'
+    | 'codeWhisperer_Resources_Feedback'
+    | 'codeWhisperer_Workshop_LearnMore'
+    | 'codeWhisperer_Commands_KeyboardShortcutsEditor'
+    | 'codeWhisperer_ScanCode_LearnMore'
+    | 'codeWhisperer_GenerateSuggestions_LearnMore'
+    | 'codeWhisperer_GenerateSuggestions_TryExample'
```
andrewyuq, line 122: Where do we capture the page open click event? I don't see that
`codeWhisperer_GettingStarted_PageOpen`

andrewyuq: do you also want to put `codewhisperer_learn_buttonclick` here?

src/codewhisperer/vue/backend.ts
LaxmanReddyAileni: Tried using `globalStorageUri` 
`  return path.join(globals.context.globalStorageUri.toString(), fileName)`
 but I got this error which shows not able to open 
image

So, Opted to go with globalStorage.

src/codewhisperer/vue/backend.ts
LaxmanReddyAileni: Yeah, Updated Accordingly.

src/testFixtures/workspaceFolder/.vscode/settings.json
LaxmanReddyAileni: Deleted File

src/codewhisperer/vue/backend.ts
LaxmanReddyAileni: Updated Accordingly.

src/codewhisperer/vue/backend.ts
LaxmanReddyAileni: Its just comparing the ostype and displaying Option or Alt. Even 
If we try to write logic inside Vue JS file which is not supporting to
 **import OS library**
 So to keep it simple In implemented in this way.
 
 But for the room for improvement I updated this code with 
 
```
  public getOSType():OSType {
        return os.platform() === 'darwin' ? 'Mac' : 'RestOfOS';
    }
```

src/codewhisperer/activation.ts
LaxmanReddyAileni: I will rename it to NewUser

src/codewhisperer/models/constants.ts
LaxmanReddyAileni: Updated

src/codewhisperer/vue/backend.ts
LaxmanReddyAileni: `codeWhisperer_GettingStarted_PageOpen`

That was in other branch as that requires additional files from **aws:Master**
Which was not part of my PR.
I will raise a separate PR for the Telemetry of `codeWhisperer_GettingStarted_PageOpen` after this PR was resolved

LaxmanReddyAileni: > do you also want to put `codewhisperer_learn_buttonclick` here? And also to make it consistent `codeWhisperer_Learn_ButtonClick`

Updated accordingly.

src/codewhisperer/activation.ts
LaxmanReddyAileni: Updated with `hasShownNewWelcomeMsg`

src/codewhisperer/activation.ts
```diff
@@ -41,15 +41,16 @@ import { ReferenceInlineProvider } from './service/referenceInlineProvider'
...
 import { InlineCompletionService } from './service/inlineCompletionService'
 import { isInlineCompletionEnabled } from './util/commonUtil'
 import { CodeWhispererCodeCoverageTracker } from './tracker/codewhispererCodeCoverageTracker'
 import { AuthUtil } from './util/authUtil'
 import { ImportAdderProvider } from './service/importAdderProvider'
 import { TelemetryHelper } from './util/telemetryHelper'
 import { openUrl } from '../shared/utilities/vsCodeUtils'
-
+import { CodeWhispererCommandBackend, CodeWhispererCommandDeclarations } from './commands/gettingStartedPageCommands'
+import { get } from './commands/basicCommands'
```
justinmk3, line 53: `get` is a strange thing to export from a module. It could mean anything.

src/codewhisperer/activation.ts
```diff
@@ -202,6 +206,39 @@ export async function activate(context: ExtContext): Promise<void> {
...
+        //This logic is to show the Getting Started Page only once for the first time SignIn Users
+        let start = ''
+
+        /**
+         * hasShownWelcomeMsgBefore defines whether user is existing user or new user.
+         */
+        const hasShownWelcomeMsgBefore = get(
+            /** welcomeMessageKeyPrevious is a key for the previous welcome page of version v1.91.0,*/
+            CodeWhispererConstants.welcomeMessageKeyPrevious,
+            context.extensionContext.globalState
```
justinmk3, line 219: `get` is a very common name so I would namespace it by importing the module name instead of the function itself.

Different question: instead of storing this in globalState, did CW team consider using the [PromptSettings](https://github.com/aws/aws-toolkit-vscode/blob/de8244c7172da804fdfc796914a7372b0c7c1d7a/src/shared/settings.ts#L530) feature? 

1. just add a new prompt kind [here](https://github.com/aws/aws-toolkit-vscode/blob/de8244c7172da804fdfc796914a7372b0c7c1d7a/package.json#L172)
2. then check if the prompt is enabled [like this](https://github.com/aws/aws-toolkit-vscode/blob/de8244c7172da804fdfc796914a7372b0c7c1d7a/src/codecatalyst/activation.ts#L88)

src/codewhisperer/vue/backend.ts
LaxmanReddyAileni: > do you also want to put `codewhisperer_learn_buttonclick` here? And also to make it consistent `codeWhisperer_Learn_ButtonClick`

We can not use here as `Learn` button is not a part of CodeWhispererWebView.

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -43,14 +57,17 @@ export const enableCodeSuggestions = Commands.declare(
         await vscode.commands.executeCommand('setContext', 'CODEWHISPERER_ENABLED', true)
         await vscode.commands.executeCommand('aws.codeWhisperer.refresh')
 
-        const hasShownWelcomeMsgBefore = get(
-            CodeWhispererConstants.welcomeMessageKey,
+        const hasShownWelcomeMsgBeforeToExistingUsers = get(
+            CodeWhispererConstants.welcomeMessageKeyShown,
             context.extensionContext.globalState
```
justinmk3, line 62: if [PromptSettings](https://github.com/aws/aws-toolkit-vscode/pull/3805/files#r1336497017) is used then this would become `settings.isPromptEnabled('...')`

src/feedback/vue/submitFeedback.vue
```diff
@@ -3,7 +3,7 @@
 <template>
     <div>
         <div class="container button-container" style="justify-content: space-between">
-            <h1>Feedback for AWS Toolkit</h1>
+            <h1>Feedback for {{ feedbackName }}</h1>
```
justinmk3, line 6: Glad this is being re-used, thanks

src/codewhisperer/activation.ts
LaxmanReddyAileni: Yes, `get` was an **existing function** in the code so I reused it.
I will change the name from "get" to "getState"

src/feedback/vue/submitFeedback.vue
```diff
@@ -70,22 +70,33 @@ export default defineComponent({
...
         async submitFeedback() {
             this.error = ''
             this.isSubmitting = true
             console.log('Submitting feedback...')
-
+            // identifier to help us (internally) know that feedback came from either CodeWhisperer or AWS Toolkit
             const resp = await client.submit({
-                comment: this.comment,
+                comment:
+                    this.feedbackName === 'CodeWhisperer' ? 'CodeWhisperer onboarding: ' + this.comment : this.comment,
```
justinmk3, line 87: is it really only for onboarding? would it make sense to add a "Send Feedback" item in the right-click menu of the CodeWhisperer node (in "developer tools" panel)?

LaxmanReddyAileni: This additional 'CodeWhisperer onboarding: ' was added to separate Feedback of CodeWhisperer from feedback of AWS Toolkit.

src/test/codewhisperer/explorer/codewhispererChildrenNodes.test.ts
```diff
@@ -50,6 +51,12 @@ describe('codewhisperer children nodes', function () {
         assert.strictEqual(node.resource.id, 'aws.auth.manageConnections')
     })
 
+    it('should build showGettingStartedPage command node', function () {
+        const node = createGettingStartedNode()
+
+        assert.strictEqual(node.resource.id, 'aws.codeWhisperer.gettingStarted')
```
justinmk3, line 57: are these tests doing anything meaningful? would it be less verbose and more meaningful to check the whole treeview somehow, instead of directly calling the nodes

src/test/codewhisperer/explorer/codewhispererNode.test.ts
```diff
@@ -78,5 +78,27 @@ describe('codewhispererNode', function () {
...
+            const getingStartedNode = children.find(c => c.resource.id == 'aws.codeWhisperer.gettingStarted')
+            const openCodeReferenceNode = children.find(c => c.resource.id == 'aws.codeWhisperer.openReferencePanel')
+            const runSecurityScanNode = children.find(c => c.resource.id == 'aws.codeWhisperer.security.scan')
+
+            assert.strictEqual(children.length, 4)
+
+            assert.ok(pauseAutoSuggestionsNode)
+            assert.ok(runSecurityScanNode)
+            assert.ok(openCodeReferenceNode)
+            assert.ok(getingStartedNode)
```
justinmk3, line 101: would it be less verbose and more readable to do something like:

```
assert.deepStrictEqual([...], children.map(o => o.resource.id)
```

justinmk3, line 101:
```suggestion
            const ids = node.getChildren().map(o => o.resource.id)
            assert.deepStrictEqual(ids, [
                 'aws.codeWhisperer.toggleCodeSuggestion',
                'aws.codeWhisperer.gettingStarted',
                'aws.codeWhisperer.openReferencePanel',
                'aws.codeWhisperer.security.scan',
            ])
```

src/test/feedback/commands/submitFeedbackListener.test.ts
```diff
@@ -32,7 +60,7 @@ describe('submitFeedbackListener', function () {
 
         when(mockTelemetry.postFeedback(anything())).thenThrow(new Error(error))
 
-        const webview = new FeedbackWebview(instance(mockTelemetry))
+        const webview = new FeedbackWebview(instance(mockTelemetry), 'CodeWhisperer')
```
justinmk3, line 63: The entire test code here is duplicated except for a single value. If this really needs to be tested, the test should be parameterized or deduplicated somehow.

src/test/codewhisperer/explorer/codewhispererChildrenNodes.test.ts
LaxmanReddyAileni: This test is written according to previous test case and this test case checks the 
**"createGettingStartedNode() function correctly creates a command node with the expected resource.id property"**

src/codewhisperer/activation.ts
justinmk3: Its name doesn't necessarily need to be changed. Just the import. Like this:

    import * as foo from '...'

then it can be used like this:

    foo.get()

that is easier to follow than un-namespaced. `get`. 

Renaming it to `getFoo()` is a different way of "namespacing".

justinmk3: however, the main suggestion above is regarding "PromptSettings", not get().

src/feedback/vue/submitFeedback.vue
justinmk3: Yes. So would it potentially be used for other kinds of codewhisperer feeding besides "onboarding"? If so, the `comment` could be set to "CodeWhisperer". No need to mention "onboarding" in the `comment`.

src/test/codewhisperer/explorer/codewhispererChildrenNodes.test.ts
justinmk3: I can see that, but in order to improve code we sometimes need to question existing code instead of just copying it.

package.json
LaxmanReddyAileni: Only this icon code is additional and other changes is due to it.
```
 "aws-codewhisperer-learn": {
                "description": "AWS Contributed Icon",
                "default": {
                    "fontPath": "./resources/fonts/aws-toolkit-icons.woff",
                    "fontCharacter": "\\f1af"
                }
            },
```

src/test/codewhisperer/explorer/codewhispererNode.test.ts
LaxmanReddyAileni: Updated !

LaxmanReddyAileni: Updated!

src/codewhisperer/vue/backend.ts
justinmk3: path.join() won't work on a URI. you would need to do something like

    path.join(globals.context.globalStorageUri.fsPath, fileName)

src/test/feedback/commands/submitFeedbackListener.test.ts
LaxmanReddyAileni: Reused the code into one test case.

src/feedback/vue/submitFeedback.vue
andrewyuq: It potentially can have other source from CodeWhisperer e.g. "CodeWhisperer suggestions" in the future, but not as of now.

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -114,7 +114,7 @@ export const reconnect = Commands.declare('aws.codeWhisperer.reconnect', () => a
     await AuthUtil.instance.reauthenticate()
 })
 
-export function get(key: string, context: vscode.Memento): any {
+export function getState(key: string, context: vscode.Memento): any {
     return context.get(key)
 }
```
justinmk3, line 119: What is the point of this function?

src/codewhisperer/vue/backend.ts
LaxmanReddyAileni: Updated !

src/test/codewhisperer/explorer/codewhispererChildrenNodes.test.ts
LaxmanReddyAileni: Yes, they are both test the same functionality.

src/codewhisperer/activation.ts
LaxmanReddyAileni: I need to look about "PromptSettings" neither I used nor explored on that.

src/codewhisperer/commands/basicCommands.ts
LaxmanReddyAileni: I thought it was about naming convention but I came to know about implementation of global state. 
Instead of storing global state we need to this prompt settings. 
I will check about this **Prompt Settings.**

src/test/codewhisperer/explorer/codewhispererChildrenNodes.test.ts
justinmk3: If `codewhispererNode.test.ts` is testing the same thing, then these redundant tests in `codewhispererNodeChildrenNodes.test.ts` should be deleted.

LaxmanReddyAileni: Yes, I deleted the test case which is duplicated and updated it.

src/codewhisperer/activation.ts
justinmk3: `oldWelcomeMessageKey` is a legacy thing. My comment is a general comment about the functionality, not about the "old message" specifically. It's true that PromptSettings can't be used for the "old" message. But it can be used for the "new message" functionality.

LaxmanReddyAileni: From what I understood, PromptSettings currently are used per user signIn session. If we sign out and sign-In, this will restore the values which is not expected in this functionality.
**Expected:-**
If a user once signIn, Welcome Page is shown and thats it even if the user sign out and sign-in this should not be shown again.
**By using PromptSettings:-**
If a user sign-in, Welcome Page is displayed and if user sign out and sign-in, again the Welcome Page is shown because for each sign out and sign in the values reset to default values.

justinmk3: > From what I understood, PromptSettings currently are used per user signIn session. If we sign out and sign-In, this will restore the values which is not expected in this functionality.

PromptSettings are stored in the user's vscode settings (settings.json). Not affected by sign out.

andrewyuq: To summarize:
Use the global state for the old(existing) state, since it's the place where the state is already stored by existing toolkit versions.
Use the PromptSettings for the new state

src/codewhisperer/models/constants.ts
```diff
@@ -122,8 +122,15 @@ export const licenseFilter = 'CodeWhisperer suggestions were filtered due to ref
...
+/**
+ * oldWelcomeMessageKey is a key for the previous welcome page of version v1.91.0,
+ */
+export const oldWelcomeMessageKey = 'CODEWHISPERER_WELCOME_MESSAGE'
 
-export const welcomeMessageKey = 'CODEWHISPERER_WELCOME_MESSAGE'
+/**
+ *  newWelcomeMessageKey is a new key for the new getting started experience, it is used to track if first time signIn new user/ existing user has seen the getting started page or not.
+ */
+export const newWelcomeMessageKey = 'CODEWHISPERER_WELCOME_MESSAGE_SHOWN'
```
andrewyuq, line 133: not a blocker but the actual string can be better to indicate it's key for the new UX.

src/codewhisperer/vue/backend.ts
```diff
@@ -0,0 +1,147 @@
...
+        return this.start
+    }
+
+    private isFileSaved: boolean = false
+    private getLocalFilePath(fileName: string): string {
+        //This will store the files in the global storage path of VSCode
+        return path.join(globals.context.globalStorageUri.fsPath, fileName)
+    }
+
+    // This function opens Python/JavaScript/C# file in the editor.
```
andrewyuq, line 34: comment is incorrect to not reflecting all 5 languages.

src/codewhisperer/vue/backend.ts
```diff
@@ -0,0 +1,147 @@
...
+    private async saveFileLocally(localFilePath: string, fileContent: string): Promise<void> {
+        await FileSystemCommon.instance.writeFile(localFilePath, fileContent)
+        this.isFileSaved = true
+        vscode.workspace.openTextDocument(localFilePath).then(doc => {
+            vscode.window.showTextDocument(doc, vscode.ViewColumn.Active).then(editor => {
+                const endOfDocument = new vscode.Position(doc.lineCount - 1, doc.lineAt(doc.lineCount - 1).text.length)
+                editor.selection = new vscode.Selection(endOfDocument, endOfDocument)
+            })
+        })
+    }
```
andrewyuq, line 66: I would suggest to have an error handling here in-case the saveFile fails (e.g. due to insufficient permission). And notify users through a notification? @justinmk3

src/codewhisperer/vue/backend.ts
```diff
@@ -0,0 +1,147 @@
...
+// This function is called when the extension is activated : Webview of CodeWhisperer
+export async function showCodeWhispererWebview(
+    ctx: vscode.ExtensionContext,
+    source: CodeWhispererSource,
+    start: string
+): Promise<void> {
+    activePanel ??= new Panel(ctx, start) // "start" Parameter is passed to the constructor of CodeWhispererWebview to seperate the user experience from first time user to regualr signIn user
+    activePanel.server.setSource(source)
+
+    const webview = await activePanel!.show({
```
andrewyuq, line 133: Does `activePanel!` assume/assert it needs to be non-null/not undefined? I would do an early return in case it's undefined instead of doing `activePanel!` here.

src/codewhisperer/vue/genSuggestionTab.vue
```diff
@@ -0,0 +1,498 @@
+<!-- This Vue File provides Sandbox files for experimenting CodeWhisperer and generating code suggestions -->
+<template>
+    <div class="generateSuggestionDiv">
+        <div class="generateSuggestionHeaderDiv">
+            <div class="generateSuggestionHeader">Generate code suggestions with examples</div>
+            <div class="generateSuggestionDescription">
```
andrewyuq, line 6: not a blocker. But usually style for className should be like `generate-suggestion-div`? Better check the existing code style for this.

src/codewhisperer/vue/root.vue
```diff
@@ -0,0 +1,223 @@
...
+                            id="paint0_linear_227_4962"
+                            x1="53.5957"
+                            y1="-883.34"
+                            x2="19.5336"
+                            y2="-883.327"
+                            gradientUnits="userSpaceOnUse"
+                        >
+                            <stop stop-color="#6EB1FF" />
+                            <stop offset="1" stop-color="#DA8AF7" />
+                        </linearGradient>
```
andrewyuq, line 49: I am curious how do we come up with all these magic numbers?

andrewyuq:
Please fix the conflicts and failing test if any, I will issue an approval after it gets resolved. This change looks good overall.

src/codewhisperer/models/constants.ts
LaxmanReddyAileni: We are not using this Keys anymore as we are showing the Banner for both new and old users all the time.

src/codewhisperer/vue/backend.ts
LaxmanReddyAileni: Updated to all the 5 languages  [ TypeScript/JavaScript/Python/Java/C# ]

src/codewhisperer/vue/root.vue
LaxmanReddyAileni: Its a SVG file provided by Ray. Just used it here.

src/codewhisperer/models/constants.ts
andrewyuq: just to confirm, we no longer need the `oldWelcomeMessageKey`, the new one is still needed.

src/codewhisperer/activation.ts
LaxmanReddyAileni: updated accordingly.

package.json
LaxmanReddyAileni: Updated !

src/codewhisperer/models/constants.ts
LaxmanReddyAileni: As we are using PromptSettings we are not using these keys.

src/codewhisperer/activation.ts
```diff
@@ -64,6 +64,13 @@ export async function activate(context: ExtContext): Promise<void> {
         await enableDefaultConfigCloud9()
     }
 
+    //Opens the getting started page if it is not already open.
```
nkomonen-amazon, line 67: Is this comment accurate? `registerCommandsWithVSCode` only registers the command but doesn't call it, the actual command would be done elsewhere.

LaxmanReddyAileni: Yeah That part of code was deleted and I forgot to remove this comment

andrewyuq (approved):

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -35,21 +39,31 @@ export const toggleCodeSuggestions = Commands.declare(
...
+        const prompts = PromptSettings.instance
 
-        const hasShownWelcomeMsgBefore = get(
-            CodeWhispererConstants.welcomeMessageKey,
-            context.extensionContext.globalState
-        )
-        if (!hasShownWelcomeMsgBefore) {
-            showCodeWhispererWelcomeMessage(context)
-            await set(CodeWhispererConstants.welcomeMessageKey, true, context.extensionContext.globalState)
+        const shouldShow = await prompts.isPromptEnabled('codeWhispererNewWelcomeMessageKey')
```
justinmk3, line 62: "key" is redundant in the name.
```suggestion
        const shouldShow = await prompts.isPromptEnabled('codeWhispererNewWelcomeMessage')
```

LaxmanReddyAileni: Updated Accordingly

justinmk3 (approved):

PR #3808, opened 2023-09-07
logging: only log to console in browser

nkomonen-amazon:
Problem:

From a previous change we are now logging to the Developer Tools console of a browser whether in Node.js or the Browser.

Solution:

Only log to the Developer Tools console when in the Browser and skip if we are in Node.js

## License

### DIFF
--- a/src/shared/logger/activation.ts
+++ b/src/shared/logger/activation.ts
@@ -43,7 +43,7 @@ export async function activate(
         {
             logPaths: [logUri.fsPath],
             outputChannels: [chan],
-            useConsoleLog: true,
+            useConsoleLog: isInBrowser(),
         },
         extensionContext.subscriptions
     )
@@ -100,7 +100,8 @@ export async function activate(
  * @param opts.staticLogLevel Static log level, overriding config value. Will persist overridden config value even if the config value changes.
  * @param opts.logPaths Array of paths to output log entries to
  * @param opts.outputChannels Array of output channels to log entries to
- * @param opts.useDebugConsole If true, outputs log entries to currently-active debug console. As per VS Code API, cannot specify a debug console in particular.
+ * @param opts.useDebugConsole If true, outputs log entries to `vscode.debug.activeDebugConsole`
+ * @param opts.useConsoleLog If true, outputs log entries to the nodejs or browser devtools console.
  * @param disposables Array of disposables to add a subscription to
  */
 export function makeLogger(


src/shared/logger/activation.ts
```diff
@@ -127,7 +128,7 @@ export function makeLogger(
     if (opts.useDebugConsole) {
         logger.logToDebugConsole()
     }
-    if (opts.useConsoleLog) {
+    if (opts.useBrowserConsole && isInBrowser()) {
```
justinmk3, line 131: I don't understand why isInBrowser() is needed here. We have extensionWeb.ts which can control things, instead of adding special-cases to inner components like this.

src/shared/logger/activation.ts
```diff
@@ -101,6 +101,7 @@ export async function activate(
  * @param opts.logPaths Array of paths to output log entries to
  * @param opts.outputChannels Array of output channels to log entries to
  * @param opts.useDebugConsole If true, outputs log entries to currently-active debug console. As per VS Code API, cannot specify a debug console in particular.
+ * @param opts.useBrowserConsole If true, outputs log entries to the Developer Tools console when in the browser.
```
justinmk3, line 104: This docstring is not really true because of the isInBrowser() check :) That's an example of why such special cases should be avoided: it makes explaining and reasoning about things harder.

src/shared/logger/activation.ts
nkomonen-amazon: Updated the code, let me know if that is better.

src/shared/logger/activation.ts
nkomonen-amazon: Updated the code and this string, should be more clear now

src/shared/logger/activation.ts
```diff
@@ -109,7 +110,7 @@ export function makeLogger(
         logPaths?: string[]
         outputChannels?: vscode.OutputChannel[]
         useDebugConsole?: boolean
-        useConsoleLog?: boolean
+        useDeveloperToolsConsole?: boolean
```
justinmk3, line 113: The old name was probably better. "console" is a generally known concept in the JS ecosystem and in nodejs "developer tools" is not involved (not mentioned on https://nodejs.org/api/console.html for example).

src/shared/logger/activation.ts
```diff
@@ -101,6 +101,7 @@ export async function activate(
  * @param opts.logPaths Array of paths to output log entries to
  * @param opts.outputChannels Array of output channels to log entries to
  * @param opts.useDebugConsole If true, outputs log entries to currently-active debug console. As per VS Code API, cannot specify a debug console in particular.
+ * @param opts.useDeveloperToolsConsole If true, outputs log entries to the Developer Tools console.
```
justinmk3, line 104:
```suggestion
 * @param opts.useDeveloperToolsConsole If true, outputs log entries to the nodejs or browser devtools console.
```

src/shared/logger/activation.ts
```diff
@@ -101,6 +101,7 @@ export async function activate(
  * @param opts.logPaths Array of paths to output log entries to
  * @param opts.outputChannels Array of output channels to log entries to
  * @param opts.useDebugConsole If true, outputs log entries to currently-active debug console. As per VS Code API, cannot specify a debug console in particular.
```
justinmk3, line 103: Some things are over-explained and that makes them more confusing. Yet other things have no comments at all...

```suggestion
 * @param opts.useDebugConsole If true, outputs log entries to `vscode.debug.activeDebugConsole`.
```

src/shared/logger/activation.ts
```diff
@@ -43,7 +43,7 @@ export async function activate(
         {
             logPaths: [logUri.fsPath],
             outputChannels: [chan],
-            useConsoleLog: true,
+            useDeveloperToolsConsole: isInBrowser(),
```
justinmk3, line 46: This is an OK compromise for now, but ideally this would be decided by `src/extensionWeb.ts` , right?

nkomonen-amazon: Yeah this is a compromise for now. We'd need to add a new argument to [`activate()`](https://github.com/aws/aws-toolkit-vscode/blob/master/src/shared/logger/activation.ts#L29) but there are different ways to do it which I didn't want to figure out in this PR

PR #3823, opened 2023-09-12
Various improvements to CodeWhisperer codebase

andrewyuq:
1. Refactor some duplicated code
2. Refactor some condition checking logic to improve readability
3. Moved some of the states from `RecommendationHandler` to `CodeWhispererSession` since they are session scope(continued work), as well as removed some others that are not needed.
4. Fixed an issue where CodeWhisperer would try to get editor context (left/right context) for every request within a trigger session. These info only need to be retrieved once in the beginning of the session. Otherwise the latest leftContext will contain latest users edits which is not intended.
5. Fixed an issue where CodeWhisperer would try to use the language type of the current editor for sending telemetry; There could be cases when the current editor is no longer the original editor where invocation happens, leading to the telemetry to have the incorrect language type.
6. Fixed an issue where CodeWhisperer would keep stale suggestions as global states after the pagination has been requested to cancel. Add a check right after the service response and return immediately if the current pagination session is cancelled.
7. Right before showing the inline suggestions, do a check for the current window focus state. If it's not focused, mark all suggestions as discarded and exit this session.
8. Created a `CodeWhispererApplication` class to managed some CW specific application-wide listeners/variables.

## License

### DIFF
--- a/src/codewhisperer/activation.ts
+++ b/src/codewhisperer/activation.ts
@@ -176,7 +176,7 @@ export async function activate(context: ExtContext): Promise<void> {
             if (isInlineCompletionEnabled() && e.uri.fsPath !== InlineCompletionService.instance.filePath()) {
                 return
             }
-            RecommendationHandler.instance.reportUserDecisionOfRecommendation(vscode.window.activeTextEditor, -1)
+            RecommendationHandler.instance.reportUserDecisions(-1)
         }),
 
         vscode.languages.registerHoverProvider(
@@ -259,49 +259,49 @@ export async function activate(context: ExtContext): Promise<void> {
                 await RecommendationHandler.instance.onCursorChange(e)
             }),
             vscode.workspace.onDidChangeTextDocument(async e => {
+                const editor = vscode.window.activeTextEditor
+                if (!editor) {
+                    return
+                }
+                if (e.document !== editor.document) {
+                    return
+                }
+                if (!runtimeLanguageContext.isLanguageSupported(e.document.languageId)) {
+                    return
+                }
+
                 /**
                  * CodeWhisperer security panel dynamic handling
                  */
-                if (e.document === vscode.window.activeTextEditor?.document) {
-                    disposeSecurityDiagnostic(e)
-                }
+                disposeSecurityDiagnostic(e)
 
                 CodeWhispererCodeCoverageTracker.getTracker(e.document.languageId)?.countTotalTokens(e)
 
                 /**
                  * Handle this keystroke event only when
-                 * 1. It is in current non plaintext active editor
                  * 2. It is not a backspace
                  * 3. It is not caused by CodeWhisperer editing
                  * 4. It is not from undo/redo.
                  */
-                if (
-                    e.document === vscode.window.activeTextEditor?.document &&
-                    runtimeLanguageContext.isLanguageSupported(e.document.languageId) &&
-                    e.contentChanges.length != 0 &&
-                    !vsCodeState.isCodeWhispererEditing
-                ) {
-                    if (vsCodeState.lastUserModificationTime) {
-                        TelemetryHelper.instance.setTimeSinceLastModification(
-                            performance.now() - vsCodeState.lastUserModificationTime
-                        )
-                    }
-                    vsCodeState.lastUserModificationTime = performance.now()
-                    /**
-                     * Important:  Doing this sleep(10) is to make sure
-                     * 1. this event is processed by vs code first
-                     * 2. editor.selection.active has been successfully updated by VS Code
-                     * Then this event can be processed by our code.
-                     */
-                    await sleep(CodeWhispererConstants.vsCodeCursorUpdateDelay)
-                    if (!RecommendationHandler.instance.isSuggestionVisible()) {
-                        await KeyStrokeHandler.instance.processKeyStroke(
-                            e,
-                            vscode.window.activeTextEditor,
-                            client,
-                            await getConfigEntry()
-                        )
-                    }
+                if (e.contentChanges.length === 0 || vsCodeState.isCodeWhispererEditing) {
+                    return
+                }
+
+                if (vsCodeState.lastUserModificationTime) {
+                    TelemetryHelper.instance.setTimeSinceLastModification(
+                        performance.now() - vsCodeState.lastUserModificationTime
+                    )
+                }
+                vsCodeState.lastUserModificationTime = performance.now()
+                /**
+                 * Important:  Doing this sleep(10) is to make sure
+                 * 1. this event is processed by vs code first
+                 * 2. editor.selection.active has been successfully updated by VS Code
+                 * Then this event can be processed by our code.
+                 */
+                await sleep(CodeWhispererConstants.vsCodeCursorUpdateDelay)
+                if (!RecommendationHandler.instance.isSuggestionVisible()) {
+                    await KeyStrokeHandler.instance.processKeyStroke(e, editor, client, await getConfigEntry())
                 }
             })
         )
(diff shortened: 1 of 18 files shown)


src/codewhisperer/activation.ts
```diff
@@ -259,49 +259,49 @@ export async function activate(context: ExtContext): Promise<void> {
...
-                    disposeSecurityDiagnostic(e)
-                }
+                disposeSecurityDiagnostic(e)
 
                 CodeWhispererCodeCoverageTracker.getTracker(e.document.languageId)?.countTotalTokens(e)
 
                 /**
                  * Handle this keystroke event only when
-                 * 1. It is in current non plaintext active editor
                  * 2. It is not a backspace
```
leigaol, line 282: nit: If you remove 1. you should rename 2,3,4

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -109,8 +118,8 @@ export class InlineCompletionService {
                     page
                 )
                 if (RecommendationHandler.instance.checkAndResetCancellationTokens()) {
-                    // RecommendationHandler.instance.reportUserDecisionOfRecommendation(editor, -1)
-                    // vscode.commands.executeCommand('aws.codeWhisperer.refreshStatusBar')
```
leigaol, line 113: Remove these 2 lines?

leigaol:
There is one test that fails
`
  1) inlineCompletionService
       getPaginatedRecommendation
         should call checkAndResetCancellationTokens before showing inline and next token to be null:
     TypeError: Cannot read properties of undefined (reading 'errorMessage')
      at InlineCompletionService.<anonymous> (src/codewhisperer/service/inlineCompletionService.ts:137:39)
      at Generator.next (<anonymous>)
      at fulfilled (dist/src/codewhisperer/service/inlineCompletionService.js:2:3366)`

src/codewhisperer/service/inlineCompletionService.ts
andrewyuq: can you ellaborate? I have uncommented these two lines since there's a need to end the session(report userDecisions) whenever we get pagination cancelled.

src/codewhisperer/activation.ts
```diff
@@ -259,49 +263,49 @@ export async function activate(context: ExtContext): Promise<void> {
...
+                    return
+                }
+
                 /**
                  * CodeWhisperer security panel dynamic handling
                  */
-                if (e.document === vscode.window.activeTextEditor?.document) {
-                    disposeSecurityDiagnostic(e)
-                }
+                disposeSecurityDiagnostic(e)
```
Will-ShaoHua, line 280: is the change on purpose? (removing the `if` check), not sure what this line is doing, just want do double check this change is intentional

justinmk3:
needs a rebase

src/codewhisperer/activation.ts
```diff
@@ -259,49 +263,49 @@ export async function activate(context: ExtContext): Promise<void> {
...
-                 * 2. It is not a backspace
-                 * 3. It is not caused by CodeWhisperer editing
-                 * 4. It is not from undo/redo.
+                 * 1. It is not a backspace
+                 * 2. It is not caused by CodeWhisperer editing
+                 * 3. It is not from undo/redo.
                  */
-                if (
-                    e.document === vscode.window.activeTextEditor?.document &&
-                    runtimeLanguageContext.isLanguageSupported(e.document.languageId) &&
```
Will-ShaoHua, line 280: just a sanity check if it's safe to remove this?

src/codewhisperer/models/constants.ts
```diff
@@ -217,7 +217,7 @@ export const artifactTypeSource = 'SourceCode'
 export const codeScanFindingsSchema = 'codescan/findings/1.0'
 
 // wait time for editor to update editor.selection.active (in milliseconds)
-export const vsCodeCursorUpdateDelay = 3
+export const vsCodeCursorUpdateDelay = 10
```
Will-ShaoHua, line 220: don't have context for this constant change, but if it's intentional, can just resolve my comment

src/codewhisperer/activation.ts
andrewyuq: The check is moved to above lines

src/codewhisperer/activation.ts
andrewyuq: The check is moved to above lines

src/codewhisperer/util/codeWhispererSession.ts
Will-ShaoHua, line 1: a general question but not specific to the changes in this PR, why do we need a singleton instance of CodeWhispererSession? Couldn't we make it a regular object and init a new one everytime user has a new trigger

Will-ShaoHua: Or there is some difficulty passing the object through the vscode inline API

andrewyuq: This will be a better approach, as this is what JB does, but passing it around for every to get access is another refactor, I am planning to address it as part of the second round factor after I finished cleaning up all the existing ones.

Will-ShaoHua: ^^ not blocking comment

Will-ShaoHua: sgtm

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -148,27 +142,25 @@ export class RecommendationHandler {
...
-    ) {
+    ): Promise<GetRecommendationsResponse> {
+        let invocationResult: 'Succeeded' | 'Failed' = 'Failed'
+        let errorMessage: string | undefined = undefined
+
         if (!editor) {
-            return
+            return Promise.resolve<GetRecommendationsResponse>({
+                result: invocationResult,
+                errorMessage: errorMessage,
```
Will-ShaoHua, line 152: nit: errorMessage here is `undefined`, make it `no active editor found`?

Will-ShaoHua (approved):
lgtm, don't have any major comment but we have to functionally test it

justinmk3 (approved):

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -192,55 +184,64 @@ export class RecommendationHandler {
...
+            latency = startTime !== 0 ? performance.now() - startTime : 0
+            if ('recommendations' in resp) {
+                recommendations = (resp && resp.recommendations) || []
+            } else {
+                recommendations = (resp && resp.completions) || []
+            }
+            invocationResult = 'Succeeded'
+            TelemetryHelper.instance.triggerType = triggerType
+            TelemetryHelper.instance.CodeWhispererAutomatedtriggerType =
+                autoTriggerType === undefined ? 'KeyStrokeCount' : autoTriggerType
```
zixlin7, line 234: nit: we should probably update the default here since there is no more 'KeyStrokeCount'

zixlin7 (approved):

src/codewhisperer/service/recommendationHandler.ts
andrewyuq: ah yes, turns out this is another globale that we don't need, will remove.

src/codewhisperer/service/recommendationHandler.ts
andrewyuq: Making it a non undefined will make it show a notification for this error, which changes the current behavior. When the user closes the current editor, I think we can just sliently exit without popuping a message.

PR #3829, opened 2023-09-14
sso: use new 'confirm match' device code flow

nkomonen-amazon:
Problem:

Before this commit, the BuilderID/Identity Center sso flow
would require the user to copy a code and paste it in the
browser window to confirm their device.

Solution:

Now, with new changes from the identity team, if we pass
the device code in the URL a different flow will be shown
in the browser which asks them to instead confirm the
code instead of pasting it.

- This commit has users go through the new flow
- Copying code functionality is removed
- The information messages are updated to mention the requirement
  to confirm the code in the browser

#### This appears first

Screenshot 2023-09-15 at 11 09 35 AM

#### Then the user is brought to the web page
Screenshot 2023-09-15 at 11 09 48 AM

#### If the user forgets the code before clicking `Confirm and Continue` and they go back to the IDE, they will see this message 
Screenshot 2023-09-15 at 11 09 17 AM

## License

### DIFF
--- a/src/auth/sso/model.ts
+++ b/src/auth/sso/model.ts
@@ -86,43 +86,51 @@ export function truncateStartUrl(startUrl: string) {
     return startUrl.match(/https?:\/\/(.*)\.awsapps\.com\/start/)?.[1] ?? startUrl
 }
 
-export async function openSsoPortalLink(
-    startUrl: string,
-    authorization: { readonly verificationUri: string; readonly userCode: string }
-): Promise<boolean> {
-    async function copyCodeAndOpenLink() {
-        await vscode.env.clipboard.writeText(authorization.userCode).then(undefined, err => {
-            getLogger().warn(`auth: failed to copy user code "${authorization.userCode}" to clipboard: %s`, err)
-        })
+type Authorization = { readonly verificationUri: string; readonly userCode: string }
 
-        const didOpen = await vscode.env.openExternal(vscode.Uri.parse(authorization.verificationUri))
-        if (!didOpen) {
+export const proceedToBrowser = localize('AWS.auth.loginWithBrowser.proceedToBrowser', 'Proceed To Browser')
+
+export async function openSsoPortalLink(startUrl: string, authorization: Authorization): Promise<boolean> {
+    /**
+     * Depending on the verification URL + parameters used, the way the sso login flow works changes.
+     * Previously, users were asked to copy and paste a device code in to the browser page.
+     *
+     * Now, with the URL this function creates, the user will instead be asked to confirm the device code
+     * in the browser.
+     */
+    function makeConfirmCodeUrl(authorization: Authorization): vscode.Uri {
+        return vscode.Uri.parse(`${authorization.verificationUri}?user_code=${authorization.userCode}`)
+    }
+
+    async function openSsoUrl() {
+        const ssoLoginUrl = makeConfirmCodeUrl(authorization)
+        const didOpenUrl = await vscode.env.openExternal(ssoLoginUrl)
+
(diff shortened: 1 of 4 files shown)


src/auth/sso/model.ts
```diff
@@ -86,43 +86,51 @@ export function truncateStartUrl(startUrl: string) {
...
-        const didOpen = await vscode.env.openExternal(vscode.Uri.parse(authorization.verificationUri))
-        if (!didOpen) {
+type Authorization = { readonly verificationUri: string; readonly userCode: string }
+
+export const proceedToBrowser = localize('AWS.auth.loginWithBrowser.proceedToBrowser', 'Proceed To Browser')
+
+export async function openSsoPortalLink(startUrl: string, authorization: Authorization): Promise<boolean> {
+    /**
+     * Depending on the verification URL + parameters used, the way the sso login flow works changes.
+     * Currently, users will be asked to copy and paste a device code in to the browser page.
```
justinmk3, line 96:
```suggestion
     * Previously, users were asked to copy and paste a device code in to the browser page.
```

src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -234,7 +234,7 @@ async function pollForTokenWithProgress<T>(
         {
             title: localize(
                 'AWS.auth.loginWithBrowser.messageDetail',
-                'Login page opened in browser. When prompted, provide this code: {0}',
+                'Login page opened in browser. Confirm the code matches: {0}',
```
justinmk3, line 237: This is probably a case where the code should be as prominent as possible in the message. "Login page opened in browser" has more prominence than it should.

```suggestion
                'Confirm code "{0}" from the login page opened in your web browser.',
```

src/auth/sso/model.ts
```diff
@@ -86,43 +86,51 @@ export function truncateStartUrl(startUrl: string) {
...
     }
 
     async function showLoginNotification() {
         const name = startUrl === builderIdStartUrl ? localizedText.builderId() : localizedText.iamIdentityCenterFull()
-        const title = localize('AWS.auth.loginWithBrowser.messageTitle', 'Copy Code for {0}', name)
+        const title = localize('AWS.auth.loginWithBrowser.messageTitle', 'Confirm Code for {0}', name)
         const detail = localize(
             'AWS.auth.loginWithBrowser.messageDetail',
-            'To proceed, open the login page and provide this code to confirm the access request: {0}',
+            'In the browser, confirm this code: {0}',
```
justinmk3, line 123:
```suggestion
            'Confirm this code in the browser: {0}',
```

justinmk3 (approved):

src/auth/sso/model.ts
```diff
@@ -86,43 +86,51 @@ export function truncateStartUrl(startUrl: string) {
...
             })
         }
-        return didOpen
+        return didOpenUrl
     }
 
     async function showLoginNotification() {
         const name = startUrl === builderIdStartUrl ? localizedText.builderId() : localizedText.iamIdentityCenterFull()
-        const title = localize('AWS.auth.loginWithBrowser.messageTitle', 'Copy Code for {0}', name)
+        const title = localize('AWS.auth.loginWithBrowser.messageTitle', 'Confirm Code for {0}', name)
```
justinmk3, line 120: Given the way that vscode displays the modal dialog, we may want to also show the code in the title.

Can revisit later though.

nkomonen-amazon: Yeah, I'm assuming most people will miss the code in the modal the first time anyways and the information message will be where they actually read it when they switch back to the IDE

PR #3832, opened 2023-09-14
CodeWhisperer: Fix multiple issues around reporting CodePercentage

andrewyuq:
1. Add the logic to compute generatedLines count, this will enable reporting the correct count of lines from CodeWhisperer. Making it a local computation.
2. Remove the 20 chars check in countTotalTokens(). This will enable the tracker timer to start whenever there's a change.
3. Remove an additional addTotalToken() call when accepting CodeWhisperer suggestions, this avoids double counting the total tokens when accepting CW suggetions.
4. Fix an issue where the referenceCount for each trigger is not reset after assigned. Removed the global and make it a local computation.
5. Ignore deletion events when computing total tokens, since it's non trivial to track the deletion of CW suggestions. This implictly makes the metric to track the code percentage of CW suggestion among all the code user generates.

## License

### DIFF
--- a/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
+++ b/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
@@ -64,7 +64,6 @@ export class CodeWhispererCodeCoverageTracker {
         }
         // generate accepted recommendation token and stored in collection
         this.addAcceptedTokens(filename, { range: range, text: text, accepted: text.length })
-        this.addTotalTokens(filename, text.length)
     }
 
     public incrementServiceInvocationCount() {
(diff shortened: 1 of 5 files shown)


src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -224,26 +223,19 @@ export class CodeWhispererCodeCoverageTracker {
 
     public countTotalTokens(e: vscode.TextDocumentChangeEvent) {
         // ignore no contentChanges. ignore contentChanges from other plugins (formatters)
-        // only include contentChanges from user action
+        // only include contentChanges from user action.
+        // Also ignore deletion events due to a known issue of tracking deleted CodeWhiperer tokens.
```
Will-ShaoHua, line 227: curious what's the known issue here

Will-ShaoHua (approved):

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
Will-ShaoHua: seems we have some problem tracking deletion token so remove it for now?

andrewyuq: I have a CW suggestion accepted there -- upon deletion of some texts, I am able to track the # of characters got deleted, but not able to track the # of characters deleted dynamically. To actually track this needs another non-trivial work. Not part of this PR.

justinmk3 (approved):

panshao612 (approved):

PR #3837, opened 2023-09-15
Feature/sm ide

JedML:
## Problem
New SM VSCode IDE with IAM role, we want to use the IAM role for Code Whisperer auth. Currently Code Whisperer auth used Builder ID/SSO/Cloud9(IAM). Current IAM auth is for Cloud9 IDE and does not allow for paginated results.

## Solution
Add new IDE type for SM VSCode that can be used for this feature and future modifications. Use the new IDE type for conditional on connection type to use, defaulting to IAM. Use IAM for Code Whisperer and make sure we use the correct, `generateRecommendations`, endpoint to allow for auth with IAM and paginated results.

## License


.changes/next-release/Feature-015cabbc-aec3-4b75-92e9-10489e8ee7be.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Add new IDE type SM VSCode"
```
justinmk3, line 3: The changelog is for customers, it's our lowest-friction form of marketing. So changelog items should be meaningful to customers. Is this description going to mean anything to customers that read it?

JedML: Understood, sent you a slack, let me know how you like me to proceed.

src/codewhisperer/explorer/codewhispererNode.ts
```diff
@@ -74,13 +78,13 @@ export class CodeWhispererNode implements RootNode {
             return [createSsoSignIn(), createLearnMore()]
         }
         if (this._showFreeTierLimitReachedNode) {
-            if (isCloud9()) {
+            if (isCloud9() || isSageMaker()) {
```
justinmk3, line 81: if this "pair" is found in many places, is there a common property that can encapsulate what's being checked, e.g.:
```suggestion
            if (hasVendedIamCredentials()) {
```

andrewyuq: If we do that, I would at least make `isSageMaker()` a separate callable just so that in case we have any SageMaker specific logics that don't apply to Cloud9.

src/codewhisperer/explorer/codewhispererNode.ts
```diff
@@ -55,9 +55,13 @@ export class CodeWhispererNode implements RootNode {
...
-            return AuthUtil.instance.isEnterpriseSsoInUse()
-                ? 'IAM Identity Center Connected'
-                : 'AWS Builder ID Connected'
+            if (AuthUtil.instance.isEnterpriseSsoInUse()) {
+                return 'IAM Identity Center Connected'
+            } else if (AuthUtil.instance.isBuilderIdInUse()) {
+                return 'AWS Builder ID Connected'
+            } else {
+                return 'IAM Connected'
+            }
```
andrewyuq, line 64: what's the logic for `isConnectionValid` when it's for SageMaker IAM? I don't see it's being changed. Does it work out of the box?

src/codewhisperer/explorer/codewhispererNode.ts
JedML: sounds good, will do

src/codewhisperer/explorer/codewhispererNode.ts
JedML: `isConnectionValid` just checks for a live, non-expired, connection. So this could be Builder ID, SSO, or IAM. So we get it out of the box

andrewyuq (approved):

src/codewhisperer/util/authUtil.ts
```diff
@@ -33,6 +33,10 @@ export const isValidCodeWhispererConnection = (conn: Connection): conn is Connec
         return isIamConnection(conn)
     }
 
+    if (isSageMaker()) {
```
justinmk3, line 36: could put this in the condition on line 32, above

src/shared/extensionUtilities.ts
```diff
@@ -80,12 +86,29 @@ export function getIdeProperties(): IdeProperties {
...
+            return createSageMakerProperties(company)
         // default is IDE.vscode
         default:
             return vscodeVals
     }
 }
 
+function createSageMakerProperties(company: string): IdeProperties {
+    return {
+        shortName: localize('AWS.vscode.shortName', 'AWS Code'),
```
justinmk3, line 103:
```suggestion
        shortName: localize('AWS.vscode.shortName', '{0} Code', company),
```

src/shared/extensionUtilities.ts
```diff
@@ -109,6 +132,14 @@ export function isCloud9(flavor: 'classic' | 'codecatalyst' | 'any' = 'any'): bo
     return (flavor === 'classic' && !codecat) || (flavor === 'codecatalyst' && codecat)
 }
 
+export function isSageMaker(): boolean {
+    return vscode.env.appName === sageMakerAppname
+}
+
+export function hasVendedIamCredentials(isC9: boolean = isCloud9(), isSM: boolean = isSageMaker()) {
```
justinmk3, line 139: a docstring can provide some useful insight
```suggestion
/** Credentials are placed in ~/.aws/ by the environment. */
export function hasVendedIamCredentials(isC9: boolean = isCloud9(), isSM: boolean = isSageMaker()) {
```

justinmk3: Note that calling isCloud9() etc in the default parameter values caused that function to be executed at module-load time. That can be problematic, especially for this module which is a low-level module used to initialize early parts of the app. 

To be defensive, instead of assigning default parameter values, make them optional and then check for undefined in the function body.

```
isC9 ??= isCloud()
...
```

src/shared/extensionUtilities.ts
```diff
@@ -338,10 +369,15 @@ export function getToolkitEnvironmentDetails(): string {
...
- * Returns the Cloud9 compute region or 'unknown' if we can't pull a region, or `undefined` if this is not Cloud9.
+ * Returns the Cloud9/SageMaker compute region or 'unknown' if we can't pull a region, or `undefined` if this is not Cloud9 or SageMaker.
  */
-export async function initializeComputeRegion(metadata?: Ec2MetadataClient, isC9: boolean = isCloud9()): Promise<void> {
-    if (isC9) {
+
+export async function initializeComputeRegion(
+    metadata?: Ec2MetadataClient,
+    isC9: boolean = isCloud9(),
+    isSM: boolean = isSageMaker()
```
justinmk3, line 378: same comment about default parameters

src/shared/extensionUtilities.ts
JedML: `company` is "Amazon" here, so this would result in a different short name then desired

src/codewhisperer/util/authUtil.ts
JedML: it could, I was just trying to avoid coupling them if I could

justinmk3:
The failing Insiders CI jobs are unrelated to this PR and are likely an upstream issue in VSCode.

```
Cannot read properties of null (reading 'finishWrite'): TypeError: Cannot read properties of null (reading 'finishWrite')
    at JSStreamSocket.finishWrite (node:internal/js_stream_socket:211:12)
```

src/shared/extensionUtilities.ts
```diff
@@ -22,13 +22,15 @@ const localize = nls.loadMessageBundle()
 const vscodeAppname = 'Visual Studio Code'
 const cloud9Appname = 'AWS Cloud9'
 const cloud9CnAppname = 'Amazon Cloud9'
+const sageMakerAppname = 'Code - OSS'
```
justinmk3, line 25: can this be `AWS SageMaker Code Editor` or something with "AWS SageMaker" in the name? 

`vscode.env.appName` is what is reported in telemetry, so a generic name like `Code - OSS` will be very obscure.

alternatively, we could special-case this [here](https://github.com/aws/aws-toolkit-vscode/blob/44a3bde971507267fd5dfb0832106fa40c6dafe4/src/shared/telemetry/telemetryClient.ts#L94) (and [here](https://github.com/aws/aws-toolkit-vscode/blob/44a3bde971507267fd5dfb0832106fa40c6dafe4/src/shared/telemetry/telemetryClient.ts#L120)), but even so, depending on a appName as generic as "Code - OSS" could run into false-positives in the wild.

JedML: revised to `SageMaker Code Editor`

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -107,24 +107,34 @@ export class RecommendationHandler {
...
-                    return await asyncCallWithTimeout(
-                        promise,
-                        timeoutMessage,
-                        CodeWhispererConstants.promiseTimeoutLimit * 1000
-                    )
-                }
+        const timeoutMessage = hasVendedIamCredentials()
+            ? `Generate recommendation timeout.`
+            : `List recommendation timeout`
+            if (
```
justinmk3, line 113: indent looks wrong. also unnecessary template strings (backticks instead of single-quotes)

src/shared/extensionUtilities.ts
```diff
@@ -44,7 +46,11 @@ export function getIdeType(): IDE {
         return IDE.cloud9
     }
 
-    // Theia doesn't necessarily have all env propertie
+    if (vscode.env.appName === sageMakerAppname) {
+        return IDE.sagemaker
+    }
+
+    // Theis doesn't necessarily have all env properties
```
justinmk3, line 53:
```suggestion
    // Theia doesn't necessarily have all env properties
```

this is referring to https://theia-ide.org/

src/shared/extensionUtilities.ts
```diff
@@ -109,6 +132,24 @@ export function isCloud9(flavor: 'classic' | 'codecatalyst' | 'any' = 'any'): bo
     return (flavor === 'classic' && !codecat) || (flavor === 'codecatalyst' && codecat)
 }
 
+export function isSageMaker(): boolean {
+    return vscode.env.appName === sageMakerAppname
+}
+
+/**
+ * Helper function to determine if Cloud9 or Sagemaker are being used. Both use IAM roles
+ * on hosting instance for auth
```
justinmk3, line 141: - describe the semantics not the implementation.
- "helper" is never needed in a description or name. 

```suggestion
 * Returns true if credentials are provided by the environment (typically via ~/.aws/).
```

src/codewhisperer/service/recommendationHandler.ts
JedML: updated

src/shared/extensionUtilities.ts
JedML: updated

src/shared/extensionUtilities.ts
JedML: updated

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -222,9 +225,8 @@ export class RecommendationHandler {
             startTime = performance.now()
             this.lastInvocationTime = startTime
             const mappedReq = runtimeLanguageContext.mapToRuntimeLanguage(request)
-            const codewhispererPromise = pagination
-                ? client.listRecommendations(mappedReq)
-                : client.generateRecommendations(mappedReq)
+            const codewhispererPromise =
+                pagination && !isSM ? client.listRecommendations(mappedReq) : client.generateRecommendations(mappedReq)
```
justinmk3, line 229: Why isn't this special-case needed here for cloud9? There are already various callsites such as [this](https://github.com/aws/aws-toolkit-vscode/blob/756ee6913a88d2a8fa4c0cbc3f46d4dd12703873/src/codewhisperer/commands/invokeRecommendation.ts#L68-L76) which set `pagination` depending on `isCloud9`. So it seems suspicious that this special-case is being added at this low level.

JedML: Cloud9 is special in that it uses the `generateRecommendations` endpoint (for IAM auth) but can't handle the pagination, due to UI limitations. So we are creating a path that allows for pagination and also IAM auth. All other paginated code paths with use `listRecommendations` and auth tokens.

Note: on the linked code it is sending `false` for pagination, since it is Cloud9.

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -62,6 +62,19 @@ export const enableCodeSuggestions = Commands.declare(
     }
 )
 
+export const disableCodeSuggestions = Commands.declare(
+    'aws.codeWhisperer.disableCodeSuggestions',
+    (context: ExtContext) => async () => {
+        await set(CodeWhispererConstants.autoTriggerEnabledKey, false, context.extensionContext.globalState)
+        await vscode.commands.executeCommand('setContext', 'CODEWHISPERER_ENABLED', true)
```
andrewyuq, line 69: This command says "disable" but here it's set to true -- is this intended?

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -266,6 +267,18 @@ export class RecommendationHandler {
                 requestId = error.requestId || ''
                 reason = `CodeWhisperer Invocation Exception: ${error?.code ?? error?.name ?? 'unknown'}`
                 await this.onThrottlingException(error, triggerType)
+
+                if (error?.code === 'AccessDeniedException') {
+                    getLogger().error('CodeWhisperer AccessDeniedException : %s', (error as Error).message)
```
andrewyuq, line 273: Are there any cases when this AccessDenied is not coming from IAM? Should we also check the errorMessage here?

arkaprava08: @andrewyuq I would not agree on parsing errorMessages as they can change in service side breaking dependencies. Any other way to detect IAM error here?

src/codewhisperer/commands/basicCommands.ts
JedML: revised

src/codewhisperer/service/recommendationHandler.ts
andrewyuq: Unfortunately this is the currently the only way we check it's due to IAM. All plugin receives is an AccessDeniedException with the IAM message.

src/shared/extensionUtilities.ts
```diff
@@ -109,6 +132,23 @@ export function isCloud9(flavor: 'classic' | 'codecatalyst' | 'any' = 'any'): bo
...
+}
+
+/**
+ * Returns true if credentials are provided by the environment (ex. via ~/.aws/)
+ *
+ * @param isC9 boolean for if Cloud9 is host
+ * @param isSM boolean for if SageMaker is host
+ * @returns boolean for if C9 "OR" SM
+ */
+export function hasVendedIamCredentials(isC9?: boolean, isSM?: boolean) {
```
justinmk3, line 146: would you mind moving this to [auth.ts](https://github.com/aws/aws-toolkit-vscode/blob/13ebf3aee88b2357a9a59b27d0308417b3f105c0/src/auth/auth.ts)

src/test/codewhisperer/explorer/codewhispererNode.test.ts
```diff
@@ -50,6 +51,20 @@ describe('codewhispererNode', function () {
             assert.strictEqual(treeItem.collapsibleState, vscode.TreeItemCollapsibleState.Collapsed)
         })
 
+        it('should create a node showing IAM connection', function () {
```
justinmk3, line 54: this test and its siblings are almost identical, this would be the right time to de-duplicate them

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -69,6 +69,19 @@ export const enableCodeSuggestions = Commands.declare(
     }
 )
 
+export const disableAutoTriggerOfCodeSuggestions = Commands.declare(
+    'aws.codeWhisperer.disableAutoTriggerOfCodeSuggestions',
```
justinmk3, line 73: unnecessarily verbose. there is no other "auto trigger" that needs to be disambiguated.
```suggestion
    'aws.codeWhisperer.disableAutoTrigger',
```

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -270,6 +273,18 @@ export class RecommendationHandler {
...
+                    vscode.window
+                        .showErrorMessage(`CodeWhisperer: ${error?.message}`, CodeWhispererConstants.settingsLearnMore)
+                        .then(async resp => {
+                            if (resp === CodeWhispererConstants.settingsLearnMore) {
+                                openUrl(vscode.Uri.parse(CodeWhispererConstants.learnMoreUri))
+                            }
+                        })
+                    await vscode.commands.executeCommand('aws.codeWhisperer.disableAutoTriggerOfCodeSuggestions')
```
justinmk3, line 286: if this is only called programmatically why isn't it just a function call? why was it defined as a vscode command?

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -270,6 +273,18 @@ export class RecommendationHandler {
                 errorCode = error.code
                 reason = `CodeWhisperer Invocation Exception: ${error?.code ?? error?.name ?? 'unknown'}`
                 await this.onThrottlingException(error, triggerType)
+
+                if (error?.code === 'AccessDeniedException') {
+                    getLogger().error('CodeWhisperer AccessDeniedException : %s', (error as Error).message)
+                    vscode.window
+                        .showErrorMessage(`CodeWhisperer: ${error?.message}`, CodeWhispererConstants.settingsLearnMore)
+                        .then(async resp => {
+                            if (resp === CodeWhispererConstants.settingsLearnMore) {
```
justinmk3, line 282: this is the third place where this logic was duplicated. please extract all 3 cases to a shared function

src/codewhisperer/service/recommendationHandler.ts
justinmk3: yes, the other cases in the codebase that check AccessDeniedException check other parts of the message: 

- https://github.com/aws/aws-toolkit-vscode/blob/13ebf3aee88b2357a9a59b27d0308417b3f105c0/src/codecatalyst/auth.ts#L215
- https://github.com/aws/aws-toolkit-vscode/blob/13ebf3aee88b2357a9a59b27d0308417b3f105c0/src/codewhisperer/service/recommendationHandler.ts#L305
- https://github.com/aws/aws-toolkit-vscode/blob/13ebf3aee88b2357a9a59b27d0308417b3f105c0/src/shared/clients/codecatalystClient.ts#L235

...so checking `AccessDeniedException` without also checking for message contents, could be a potential bug.

it's also probably time to extract this common logic into a shared function in [errors.ts](https://github.com/aws/aws-toolkit-vscode/blob/13ebf3aee88b2357a9a59b27d0308417b3f105c0/src/shared/errors.ts) , something like `errors.ts:isAccessDeniedError()`. It could live next to [this](https://github.com/aws/aws-toolkit-vscode/blob/13ebf3aee88b2357a9a59b27d0308417b3f105c0/src/shared/errors.ts#L373). This function may not do much but it serves as a placeholder for all of the places that are using this bit of esoteric knowledge.

src/codewhisperer/service/recommendationHandler.ts
JedML: it is setting and calling other commands and state similar to the pattern of the `enableCodeSuggestions` command. We were just trying to keep similar process with similar.

justinmk3: In fact it's almost identical to `enableCodeSuggestions`. So a parameter can be added to `enableCodeSuggestions` instead of defining this new command.

PR #3840, opened 2023-09-19
build: set target to "es2021"

justinmk3:
Problem:
"es6" is very old and all of our targets support es2021. Targeting es6 disallows new javascript features such as Promise.allSettled().

Solution:
- increase target to "es2021".
- improve logging
- invalidate(): use Promise.allSettled() instead of all()

No significant change in (unminified) artifact size:

```
:!du -hs */dist/src/main.js  
 33M    es2021/dist/src/main.js                                                                                                                                                                                                   
 33M    es6/dist/src/main.js
```

Sample of webpack'd source changes:

image

## License

### DIFF
--- a/scripts/build/package.ts
+++ b/scripts/build/package.ts
@@ -24,7 +24,7 @@ import * as fs from 'fs-extra'
 import { betaUrl } from '../../src/dev/config'
 
 const packageJsonFile = './package.json'
-const webpackConfigJsFile = './webpack.config.js'
+const webpackConfigJsFile = './webpack.base.config.js'
 
 function parseArgs() {
     // Invoking this script with argument "foo":


webpack.base.config.js
```diff
@@ -69,7 +69,7 @@ const baseConfig = {
                         loader: 'esbuild-loader',
                         options: {
                             loader: 'ts',
-                            target: 'es2018',
+                            target: 'es2021',
```
justinmk3, line 72: Why isn't this driven by tsconfig...?

nkomonen-amazon: It was a design choice by the webpack team. There was an edge case where you could use the same webpack config for multiple projects and it wouldn't know which of the projects targets it should use. So they just made it explicitly required.

nkomonen-amazon (approved):

justinmk3:
The failing Insiders CI jobs are unrelated to this PR and are likely an upstream issue in VSCode. Same failures happen on [other PRs](https://github.com/aws/aws-toolkit-vscode-staging/pull/1005).

```
  1) "before each" hook: beforeEach for "returns a webview with tokens replaced":
     Uncaught TypeError: Cannot read properties of null (reading 'finishWrite')
```

- https://github.com/microsoft/vscode/issues/188676
- vscode says it's a node/electron issue... https://github.com/nodejs/node/issues/46094

PR #3845, opened 2023-09-21
codewhisperer: loading status bar stuck on expired credentials

nkomonen-amazon:
## Problem:

- The function `getPaginatedRecommendation()` sets the CW status bar to loading when it starts to run.
- But it will return early if `AuthUtil.instance.isConnectionExpired()` returns true.
- The issue is when we return early we do not revert the loading status and the status bar is stuck loading.

## Solution:
- Set the status bar to loading after `AuthUtil.instance.isConnectionExpired()` check

### Gif of status bar stuck loading
Kapture 2023-09-21 at 12 40 37

## License

### DIFF
--- a/src/codewhisperer/service/inlineCompletionService.ts
+++ b/src/codewhisperer/service/inlineCompletionService.ts
@@ -85,7 +85,6 @@ export class InlineCompletionService {
         // Call report user decisions once to report recommendations leftover from last invocation.
         RecommendationHandler.instance.reportUserDecisions(-1)
 
-        this.setCodeWhispererStatusBarLoading()
         if (ClassifierTrigger.instance.shouldInvokeClassifier(editor.document.languageId)) {
             ClassifierTrigger.instance.recordClassifierResultForAutoTrigger(editor, autoTriggerType, event)
         }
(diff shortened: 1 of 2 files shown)


src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -98,6 +97,9 @@ export class InlineCompletionService {
             await AuthUtil.instance.notifyReauthenticate(isAutoTrigger)
             return
```
nkomonen-amazon, line 98: Maybe also we do not want to even return here and just continue?

justinmk3:
Nice! Should be mentioned in a changelog, something like "fix CodeWhisperer: status item may get stuck"

src/codewhisperer/service/inlineCompletionService.ts
andrewyuq: We want to continue only when the reauthenticate is a success, meaning sessionToken is still valid to get a new accessToken and use that token to invoke CW service, but in case it fails it should return.

andrewyuq: We can move it here as long as there isn't a noticable time difference betwen users trigger(especially manual trigger) and the icon changes. Otherwise we can also set it to OK before return.

nkomonen-amazon: Ah I see how the flow is intended to be now. We can ignore this, thanks for the info.

justinmk3 (approved):
need approval from CW team

andrewyuq (approved):

PR #3861, opened 2023-09-29
Getting Started UX

LaxmanReddyAileni:
## Problem
User can see Getting Started UX Even without Login
## Solution

- Fixed the problem and now user can see the Getting Started UX only if user is logged in.
- Made Telemetry changes to capture user clicks in getting started UX.

## License


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "aws-toolkit-vscode",
     "displayName": "AWS Toolkit",
     "description": "Including CodeWhisperer, CodeCatalyst, and support for Lambda, S3, CloudWatch Logs, and many other services",
-    "version": "1.92.0-SNAPSHOT",
+    "version": "1.92.0-656101ffc424",
```
justinmk3, line 5: revert

src/codewhisperer/commands/gettingStartedPageCommands.ts
```diff
@@ -6,13 +6,20 @@ import * as vscode from 'vscode'
 import { CommandDeclarations, Commands } from '../../shared/vscode/commands2'
 import { showCodeWhispererWebview, CodeWhispererSource } from '../vue/backend'
 import { telemetry } from '../../shared/telemetry/telemetry'
+
+import { PromptSettings } from '../../shared/settings'
```
andrewyuq, line 10: unnecessary extra line

src/codewhisperer/commands/gettingStartedPageCommands.ts
```diff
@@ -6,13 +6,20 @@ import * as vscode from 'vscode'
...
  */
 export class CodeWhispererCommandBackend {
     constructor(private readonly extContext: vscode.ExtensionContext) {}
-    public showGettingStartedPage(source: CodeWhispererSource) {
-        telemetry.ui_click.emit({ elementId: 'codewhisperer_Learn_ButtonClick', passive: true })
+    public async showGettingStartedPage(source: CodeWhispererSource) {
+        const prompts = PromptSettings.instance
+        //To check the condition If the user has already seen the welcome message
+        if (!(await prompts.isPromptEnabled('codeWhispererNewWelcomeMessage'))) {
+            telemetry.ui_click.emit({ elementId: 'codewhisperer_Learn_ButtonClick', passive: true })
```
justinmk3, line 21: If it is a user action (a click) then it's not passive.
```suggestion
            telemetry.ui_click.emit({ elementId: 'codewhisperer_Learn_ButtonClick', passive: false })
```

src/codewhisperer/commands/gettingStartedPageCommands.ts
```diff
@@ -6,13 +6,20 @@ import * as vscode from 'vscode'
...
+        }
```
andrewyuq, line 22: I am not following, why do we need to check the condition here?

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -136,13 +136,15 @@ export class RecommendationHandler {
     }
 
     async getTaskTypeFromEditorFileName(fileName: string): Promise<CodewhispererGettingStartedTask> {
-        if (fileName.startsWith('CodeWhisperer_generate_suggestion')) {
+        const parts = fileName.split('/')
+        const name = parts[parts.length - 1]
```
justinmk3, line 140: could `name` be undefined? what guarantees that `fileName` has slashes?

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -136,13 +136,15 @@ export class RecommendationHandler {
     }
 
     async getTaskTypeFromEditorFileName(fileName: string): Promise<CodewhispererGettingStartedTask> {
-        if (fileName.startsWith('CodeWhisperer_generate_suggestion')) {
+        const parts = fileName.split('/')
```
andrewyuq, line 139: does this work on windows path?

src/codewhisperer/util/authUtil.ts
```diff
@@ -72,7 +72,10 @@ export class AuthUtil {
                 vscode.commands.executeCommand('aws.codeWhisperer.refreshStatusBar'),
                 vscode.commands.executeCommand('aws.codeWhisperer.updateReferenceLog'),
             ])
-            await vscode.commands.executeCommand('aws.codeWhisperer.enableCodeSuggestions')
+            // To check valid connection
+            if (this.isValidEnterpriseSsoInUse() || this.isBuilderIdInUse()) {
+                await vscode.commands.executeCommand('aws.codeWhisperer.enableCodeSuggestions')
```
justinmk3, line 77: This looks like an important change but how is it related to the "getting started ux"? Should this be in a separate PR, possibly with a changelog item (unless it's fixing code that wasn't released yet)?

src/codewhisperer/util/authUtil.ts
```diff
@@ -72,7 +72,10 @@ export class AuthUtil {
                 vscode.commands.executeCommand('aws.codeWhisperer.refreshStatusBar'),
                 vscode.commands.executeCommand('aws.codeWhisperer.updateReferenceLog'),
             ])
-            await vscode.commands.executeCommand('aws.codeWhisperer.enableCodeSuggestions')
+            // To check valid connection
+            if (this.isValidEnterpriseSsoInUse() || this.isBuilderIdInUse()) {
```
andrewyuq, line 76: isBuilderIdInUse doesn't check connection expired. Does this work for the case when user is using Builder ID and expired(should not show)?

src/codewhisperer/commands/gettingStartedPageCommands.ts
LaxmanReddyAileni: This is triggered automatically when user updates the plugin
So I made this Passive:true

If we keep Passive:false then 
When user updates the plugin this shows this kind of error
image

src/codewhisperer/vue/backend.ts
```diff
@@ -156,5 +156,6 @@ export async function showCodeWhispererWebview(
                 subscriptions = undefined
             }),
         ]
+        telemetry.ui_click.emit({ elementId: 'codewhisperer_Learn_PageOpen', passive: true })
```
justinmk3, line 159: The user clicked the "Learn" button, right? So why is this passive?

src/codewhisperer/util/authUtil.ts
LaxmanReddyAileni: This is fixing code that was not release yet.

User should show Learn UX only If user is logged in

src/codewhisperer/commands/gettingStartedPageCommands.ts
LaxmanReddyAileni: Updated

src/codewhisperer/commands/gettingStartedPageCommands.ts
LaxmanReddyAileni: We need to check the condition because this function is called in both cases 
When user trying to open the Learn UX using Button click and 
When user is  trying open automatically for the first time 

To differentiate 2 scenarios we need a check for this

src/codewhisperer/vue/backend.ts
LaxmanReddyAileni: This is for users for both the scenarios 
with button click and auto open so I will write a check here too

LaxmanReddyAileni: if ((await prompts.isPromptEnabled('codeWhispererNewWelcomeMessage'))) {
            telemetry.ui_click.emit({ elementId: 'codewhisperer_Learn_PageOpen', passive: true })
        } else {
            telemetry.ui_click.emit({ elementId: 'codewhisperer_Learn_PageOpen', passive: false })
        }

        Check for both cases

src/codewhisperer/util/authUtil.ts
LaxmanReddyAileni: I think logic covers all the basics

` if (this.isValidEnterpriseSsoInUse() || (this.isBuilderIdInUse()&& !this.isConnectionExpired())) `

andrewyuq: yea, I mean the current code only has `this.isValidEnterpriseSsoInUse() || this.isBuilderIdInUse()`

src/codewhisperer/commands/gettingStartedPageCommands.ts
andrewyuq: ah ok.

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -136,13 +136,15 @@ export class RecommendationHandler {
     }
 
     async getTaskTypeFromEditorFileName(fileName: string): Promise<CodewhispererGettingStartedTask> {
```
nkomonen-amazon, line 138: Is `fileName` a path to a file or just the name. I'm assuming it is a path based on how the following code uses this. If it is a path then it would be better named as `filePath`.

src/codewhisperer/service/recommendationHandler.ts
nkomonen-amazon: You'll want to use `path.basename()` instead

src/codewhisperer/service/recommendationHandler.ts
LaxmanReddyAileni: Path to the file and I am using logic of
fileName.includes(name)

Yes changed to filePath

src/codewhisperer/service/recommendationHandler.ts
LaxmanReddyAileni: I used 
` if (filePath.includes('CodeWhisperer_generate_suggestion'))`

I think this should work for both

src/codewhisperer/service/recommendationHandler.ts
LaxmanReddyAileni: Updated

src/codewhisperer/util/authUtil.ts
LaxmanReddyAileni: Yes Updating  accordingly

src/codewhisperer/util/authUtil.ts
LaxmanReddyAileni: This checks 
If users if logged In either by builderId or EnterpriseSso and not has expired connection then user is able to see the Learn UX

andrewyuq (approved):
lgtm for now.

src/codewhisperer/util/authUtil.ts
justinmk3: how is `enableCodeSuggestions` related to showing the Learn UX??

LaxmanReddyAileni: We are checking whether user is logged in in enable suggestions.

justinmk3 (approved):

Issue #2347, opened 2021-11-30
lint: custom lint rules


justinmk3:
implement these custom lint rules:

- disallow `.only()` in tests https://github.com/aws/aws-toolkit-vscode/pull/950#discussion_r387819343
- require `awsToolkit.clock.setInterval`/`setTimeout`/etc (instead of `setInterval`/`setTimeout`/etc), `awsToolkit.clock.Date` (instead of `new Date()` or `Date.foo()`) #2343

JadenSimon:
https://github.com/aws/aws-toolkit-vscode/pull/1859 has the skeleton for custom linting rules
Right now writing your own rules is kind of tough since you need to know a lot about the Typescript AST, though I'm sure we can write some easy abstractions for our use-cases.

PR #3868, opened 2023-10-02
Webview errors

nkomonen-amazon:
## Problem

Before this commit, when webview backend/server code threw
an error it would only be logged. So during interactions in
the UI if the user did something and the backend threw
it would not be obvious to them that something failed.

## Solution
Now, for any error that is thrown we will show an
error message to the user and link to where it failed in the
logs.

#### Additional:

In some cases, such as `show.ts#ssoSetup()` of this PR, we
make an effort to handle the exception gracefully, telling the user
in the webview UI. But if there is an unhandled error we would show an error message.

### Before
We get a vague error under the `Sign Up or Sign In` button
Screenshot 2023-10-02 at 12 24 02 PM

### After
We get the same error as before, PLUS an information message
Screenshot 2023-10-02 at 12 24 08 PM

If the `View Logs` button is clicked they will be brought to here with the specific error automatically highlighted. In this case it indicates there was a file permission error.
Screenshot 2023-10-02 at 12 24 16 PM

## License


src/auth/ui/vue/show.ts
```diff
@@ -236,7 +244,11 @@ export class AuthWebview extends VueWebview {
                 return { id: 'badStartUrl', text: `Connection failed. Please verify your start URL.` }
             }
 
-            getLogger().error('AuthWebview: Failed to setup: %s', (e as Error).message)
+            // If SSO setup fails we want to be able to show the user an error in the UI, due to this we cannot
+            // throw an error here. So instead this will additionally show an error message that provides more
+            // detailed information.
+            showLoggedErrorToUser(e, this.id, methodName)
```
justinmk3, line 250: "to user" is never needed since that is implied by "show"
```suggestion
            showLoggedError(e, this.id, methodName)
```

src/auth/ui/vue/show.ts
```diff
@@ -236,7 +244,11 @@ export class AuthWebview extends VueWebview {
                 return { id: 'badStartUrl', text: `Connection failed. Please verify your start URL.` }
             }
 
-            getLogger().error('AuthWebview: Failed to setup: %s', (e as Error).message)
+            // If SSO setup fails we want to be able to show the user an error in the UI, due to this we cannot
```
justinmk3, line 247: `handleError` logs the error, right? so that's why we can remove the getLogger call here

src/extension.ts
```diff
@@ -276,7 +276,7 @@ export async function activate(context: vscode.ExtensionContext) {
 // This is only being used for errors from commands although there's plenty of other places where it
 // could be used. It needs to be apart of some sort of `core` module that is guaranteed to initialize
 // prior to every other Toolkit component. Logging and telemetry would fit well within this core module.
-async function handleError(error: unknown, topic: string, defaultMessage: string) {
+export async function handleError(error: unknown, topic: string, defaultMessage: string) {
```
justinmk3, line 279: instead of exporting this should we keep the same pattern where extension.ts calls `registerErrorHandler` in the relevant module? that will keep all the related code in one place

src/shared/logger/commands.ts
```diff
@@ -43,6 +43,14 @@ export class Logging {
             return
         }
 
+        // HACK: The editor.document.getText() was returning a stale cached version of the logs,
+        // and due to this highlighting the new log would fail on subsequent calls of this function.T
+        // The following performs an action on the currently active file
+        // (which is why this command is right after the document is shown)
+        // and causes it to not use the cached getText() result.
+        // There is no guarantee for this behaviour in the future
```
justinmk3, line 51: Using the past tense here makes the comment hard to understand. We can express the same idea with fewer words and more precision.

```suggestion
        // HACK: editor.document.getText() may return "stale" content,
        // then subsequent calls to openLogId() get confused.
        // Invoke "revert" on the current file to force vscode to read from disk.
```

src/webviews/server.ts
```diff
@@ -81,3 +91,22 @@ export function registerWebviewServer(webview: vscode.Webview, commands: Protoco
...
+ * @param webviewId Arbitrary value that identifies which webview had the error
+ * @param command The high level command/function that was run which triggered the error
+ */
+export function showLoggedErrorToUser(err: unknown, webviewId: string, command: string) {
+    // HACK: The following implementation is a hack, influenced by the implementation of handleError().
+    // The userFacingError message will be seen in the UI, and the detailedError message will provide the
+    // detailed information in the logs.
+    const detailedError = ToolkitError.chain(
+        err,
+        `The backend command "${command}()" failed when executed by the webview.`
```
justinmk3, line 108: - Variable part should go at the end where possible.
- Can express the same idea with fewer words.

```suggestion
        `webview backend command failed: ${command}()`
```

src/webviews/server.ts
```diff
@@ -81,3 +91,22 @@ export function registerWebviewServer(webview: vscode.Webview, commands: Protoco
...
+    )
+    const userFacingError = ToolkitError.chain(detailedError, 'A webview had an error.')
+    handleError(userFacingError, `webviewId="${webviewId}"`, 'A webview had an error.')
```
justinmk3, line 111: Using passive voice makes it sound even more mysterious.

```suggestion
    handleError(userFacingError, `webviewId="${webviewId}"`, 'Webview error')
```

src/webviews/server.ts
```diff
@@ -81,3 +91,22 @@ export function registerWebviewServer(webview: vscode.Webview, commands: Protoco
...
+ * Show an error to the user + button that links to the logged error
+ *
+ * @param err The error that was thrown in the backend
+ * @param webviewId Arbitrary value that identifies which webview had the error
+ * @param command The high level command/function that was run which triggered the error
+ */
+export function showLoggedErrorToUser(err: unknown, webviewId: string, command: string) {
+    // HACK: The following implementation is a hack, influenced by the implementation of handleError().
+    // The userFacingError message will be seen in the UI, and the detailedError message will provide the
+    // detailed information in the logs.
```
justinmk3, line 105: it sounds like the comment is trying to explain that this is fragile. if so, to guard against future changes in `handleError`, maybe we need a test

src/shared/utilities/errorUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @param error The error itself
+ * @param topic The prefix of the error message
+ * @param defaultMessage The message to show if once cannot be resolved from the given error
+ * 
+ * SIDE NOTE:
+ * This is only being used for errors from commands although there's plenty of other places where it
+ * could be used. It needs to be apart of some sort of `core` module that is guaranteed to initialize
+ * prior to every other Toolkit component. Logging and telemetry would fit well within this core module.
+ */
+export async function showErrorToUser(error: unknown, topic: string, defaultMessage: string) {
```
justinmk3, line 27: this could live in `errors.ts`, is errorUtils.ts needed?

the old `handleError` name conveyed the fact that it does more than just show an error.

src/shared/utilities/errorUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * The high level function for showing the error to the user.
+ * 
+ * @param error The error itself
+ * @param topic The prefix of the error message
+ * @param defaultMessage The message to show if once cannot be resolved from the given error
+ * 
+ * SIDE NOTE:
+ * This is only being used for errors from commands although there's plenty of other places where it
```
justinmk3, line 23:
```suggestion
 * This is only being used for errors from commands and webviews, there's other places (explorer nodes, ...) where it
```

src/webviews/util.ts
```diff
@@ -3,6 +3,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+/**
+ * IMPORTANT:
+ *
+ * This class is used in the frontend vue.js webview code.
+ * Do not put backend code in this module.
+ */
```
justinmk3, line 11: added an item in https://github.com/aws/aws-toolkit-vscode/issues/2347 to track adding a lint rule for this.

src/auth/ui/vue/show.ts
nkomonen-amazon: Yeah, handleError logs the error + shows a message to the user. I've renamed the method to make it more clear.

src/shared/utilities/errorUtils.ts
nkomonen-amazon: I tried that initially, but got a circular dependency issue.

> the old `handleError` name ...
Agreed, I renamed it to `logAndShowError()` and also update the docstring with more nuanced information.

src/webviews/server.ts
nkomonen-amazon: Makes sense, created a test.

src/extension.ts
```diff
@@ -273,29 +271,6 @@ export async function activate(context: vscode.ExtensionContext) {
     }
 }
 
-// This is only being used for errors from commands although there's plenty of other places where it
-// could be used. It needs to be apart of some sort of `core` module that is guaranteed to initialize
-// prior to every other Toolkit component. Logging and telemetry would fit well within this core module.
-async function handleError(error: unknown, topic: string, defaultMessage: string) {
```
justinmk3, line 279: If we kept this here for now, would that avoid the circular dep?

And then we could use the same `registerErrorHandler` pattern for registering the webview handler. Not sure if that's what we want long-term, but in the meantime using the same pattern makes it easy to understand in the future if we want to refactor.

src/extension.ts
```diff
@@ -91,7 +89,7 @@ export async function activate(context: vscode.ExtensionContext) {
 
     registerErrorHandler((info, error) => {
         const defaultMessage = localize('AWS.generic.message.error', 'Failed to run command: {0}', info.id)
-        handleError(error, info.id, defaultMessage)
+        logAndShowError(error, info.id, defaultMessage)
     })
```
justinmk3, line 93: if we use the same `registerErrorHandler` pattern for webview handler then we would call `webviews.registerErrorHandler()` just below here. 

then we have a consistent pattern for both, and the setup is done at the same time (here) during startup.

src/auth/ui/vue/show.ts
```diff
@@ -129,11 +130,11 @@ export class AuthWebview extends VueWebview {
     }
 
     async startCodeWhispererBuilderIdSetup(): Promise<AuthError | undefined> {
-        return this.ssoSetup(() => awsIdSignIn())
+        return this.ssoSetup('startCodeWhispererBuilderIdSetup', () => awsIdSignIn())
```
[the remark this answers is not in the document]
nkomonen-amazon: We can use the `myFunc.name` property, but we'd have to refactor the functions to do so.

src/extension.ts
nkomonen-amazon: I've updated the code to match this pattern in a new commit.

justinmk3 (approved):

justinmk3:
Squashed the commits since there was a "fixup" commit.

PR #3870, opened 2023-10-02
codeWhisperer: fix for auto-enable suggestion on window reload and changes in Telemetry events

LaxmanReddyAileni:
## Problem

- Auto-Enable suggestion on window reload
- Missing "CodewhispererGettingStartedTask" attribute in both "serviceInvocation" and "userTriggerDecision" events.

## Solution

- Fixed auto enable suggestion on window reload by adding check for Getting started function instead of enable suggestions.
- Added "CodewhispererGettingStartedTask" attribute to "serviceInvocation" and "userTriggerDecision" events.

## License

### DIFF
--- a/src/codewhisperer/commands/basicCommands.ts
+++ b/src/codewhisperer/commands/basicCommands.ts
@@ -22,7 +22,6 @@ import { openUrl } from '../../shared/utilities/vsCodeUtils'
 import { CodeWhispererCommandDeclarations } from '../commands/gettingStartedPageCommands'
 import { getIcon } from '../../shared/icons'
 import { localize } from '../../shared/utilities/vsCodeUtils'
-import { PromptSettings } from '../../shared/settings'
 
 export const toggleCodeSuggestions = Commands.declare(
     'aws.codeWhisperer.toggleCodeSuggestion',
@@ -57,14 +56,6 @@ export const enableCodeSuggestions = Commands.declare(
         await set(CodeWhispererConstants.autoTriggerEnabledKey, true, context.extensionContext.globalState)
         await vscode.commands.executeCommand('setContext', 'CODEWHISPERER_ENABLED', true)
         await vscode.commands.executeCommand('aws.codeWhisperer.refresh')
-        const prompts = PromptSettings.instance
-
-        const shouldShow = await prompts.isPromptEnabled('codeWhispererNewWelcomeMessage')
-        //If user login old or new, If welcome message is not shown then open the Getting Started Page after this mark it as SHOWN.
-        if (shouldShow) {
-            vscode.commands.executeCommand('aws.codeWhisperer.gettingStarted', createGettingStartedNode())
-            prompts.update('codeWhispererNewWelcomeMessage', true)
-        }
         if (!isCloud9()) {
             await vscode.commands.executeCommand('aws.codeWhisperer.refreshStatusBar')
         }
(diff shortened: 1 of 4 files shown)


src/codewhisperer/util/authUtil.ts
```diff
@@ -72,7 +72,17 @@ export class AuthUtil {
                 vscode.commands.executeCommand('aws.codeWhisperer.refreshStatusBar'),
                 vscode.commands.executeCommand('aws.codeWhisperer.updateReferenceLog'),
             ])
-
+            const prompts = PromptSettings.instance
+
+            const shouldShow = await prompts.isPromptEnabled('codeWhispererNewWelcomeMessage')
+            // To check valid connection
+            if (this.isValidEnterpriseSsoInUse() || (this.isBuilderIdInUse() && !this.isConnectionExpired())) {
```
mr-lee, line 79: shouldn't you check for both if the connection is not expired?

mr-lee (commented):
you might want to consider squashing your commits as well

src/codewhisperer/util/authUtil.ts
LaxmanReddyAileni: isValidEnterpriseSsoInUse() - function already checks the condition. So, i did not write the check again.

```
   public isValidEnterpriseSsoInUse(): boolean {
        return this.isEnterpriseSsoInUse() && !this.isConnectionExpired()
    }
```

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -57,14 +56,6 @@ export const enableCodeSuggestions = Commands.declare(
         await set(CodeWhispererConstants.autoTriggerEnabledKey, true, context.extensionContext.globalState)
         await vscode.commands.executeCommand('setContext', 'CODEWHISPERER_ENABLED', true)
         await vscode.commands.executeCommand('aws.codeWhisperer.refresh')
-        const prompts = PromptSettings.instance
-
-        const shouldShow = await prompts.isPromptEnabled('codeWhispererNewWelcomeMessage')
-        //If user login old or new, If welcome message is not shown then open the Getting Started Page after this mark it as SHOWN.
-        if (shouldShow) {
-            vscode.commands.executeCommand('aws.codeWhisperer.gettingStarted', createGettingStartedNode())
```
zixlin7, line 65: why are we removing `createGettingStartedNode` ? is there other places that this node gets created?

LaxmanReddyAileni: I think we do not need to declare which callback function to execute,
```
export const createGettingStartedNode = () =>
    CodeWhispererCommandDeclarations.instance.declared.showGettingStartedPage
        .build('codewhispererDeveloperTools')
        .asTreeNode({
            label: localize('AWS.explorerNode.codewhispererGettingStartedNode.label', 'Learn'),
            iconPath: getIcon('aws-codewhisperer-learn'),
        })
```

instead using this : 
[code](https://github.com/LaxmanReddyAileni/aws-toolkit-vscode/blob/feature/gettingstarted/src/codewhisperer/commands/gettingStartedPageCommands.ts)

```
     showGettingStartedPage: Commands.from(CodeWhispererCommandBackend).declareShowGettingStartedPage(
            'aws.codeWhisperer.gettingStarted'
        ),
```

Correct me if I am wrong!
I tried to generate package and it works in all test cases.

andrewyuq: I guess what's confusing here to me is:
here we want to show the page, what does it have to do with `createGettingStartedNode()` at all if `createGettingStartedNode()` is meant to just render to `Learn` node in the devTools panel?

LaxmanReddyAileni: Yeah thats the reason I removed it from the existing code.

andrewyuq (approved):

PR #3876, opened 2023-10-05
codewhisperer: fix error type get swllowed

Will-ShaoHua:
## License

### DIFF
--- a/src/codewhisperer/service/recommendationHandler.ts
+++ b/src/codewhisperer/service/recommendationHandler.ts
@@ -131,7 +131,7 @@ export class RecommendationHandler {
                 CodeWhispererConstants.promiseTimeoutLimit * 1000
             )
         } catch (error) {
-            throw new Error(`${error instanceof Error ? error.message : error}`)
+            throw error
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -131,7 +131,7 @@ export class RecommendationHandler {
                 CodeWhispererConstants.promiseTimeoutLimit * 1000
             )
         } catch (error) {
-            throw new Error(`${error instanceof Error ? error.message : error}`)
+            throw error
```
Will-ShaoHua, line 134: since we don't have any specific handling in the catch block, should we simply remove the entire try..catch block here ? @andrewyuq @leigaol @zixlin7

Will-ShaoHua: the error is already handled at `getRecommendations()` https://github.com/aws/aws-toolkit-vscode/blob/master/src/codewhisperer/service/recommendationHandler.ts#L229

zixlin7: yes we can remove the try catch here

zixlin7 (approved):

andrewyuq (approved):

PR #3886, opened 2023-10-11
codewhisperer: fix for cloud9 path

zixlin7:
## Problem
fix no recommendation in C9 path after the first invocation
## License

### DIFF
--- a/src/codewhisperer/commands/invokeRecommendation.ts
+++ b/src/codewhisperer/commands/invokeRecommendation.ts
@@ -61,10 +61,14 @@ export async function invokeRecommendation(
         if (RecommendationHandler.instance.isGenerateRecommendationInProgress) {
             return
         }
+        RecommendationHandler.instance.checkAndResetCancellationTokens()
         vsCodeState.isIntelliSenseActive = false
         RecommendationHandler.instance.isGenerateRecommendationInProgress = true
         try {
-            let response: GetRecommendationsResponse
+            let response: GetRecommendationsResponse = {
+                result: 'Failed',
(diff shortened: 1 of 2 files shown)


andrewyuq (approved):

src/codewhisperer/commands/invokeRecommendation.ts
```diff
@@ -61,10 +61,14 @@ export async function invokeRecommendation(
         if (RecommendationHandler.instance.isGenerateRecommendationInProgress) {
             return
         }
+        RecommendationHandler.instance.checkAndResetCancellationTokens()
```
justinmk3, line 64: Can you explain why this fixed the issue and how you did the troubleshooting

justinmk3 (approved):

src/codewhisperer/commands/invokeRecommendation.ts
zixlin7: I synced with Andrew, we suspected that it returned early [here](https://github.com/aws/aws-toolkit-vscode/blob/master/src/codewhisperer/service/recommendationHandler.ts#L311) as we were able to see the logs before this, and verified the fix in C9

justinmk3: Thanks. This is useful info to include in the commit message (and PR description):

> we suspected that it returned early [here](https://github.com/aws/aws-toolkit-vscode/blob/master/src/codewhisperer/service/recommendationHandler.ts#L311)

This helps in the future when others need to debug these issues. I was pretty lost when I was debugging this :)

PR #3889, opened 2023-10-12
Enable SentTelemetryEvent for Builder ID users

andrewyuq:
1. and the api will only be called when the user has optin telemetry.

## License


zixlin7 (approved):

justinmk3 (approved):

Will-ShaoHua (approved):
@andrewyuq  sanity check, do we ensure this function call will always be made while there is a CWSPR connection?

PR #3892, opened 2023-10-12
codewhisperer: suppress token filling from auto-trigger

zixlin7:
## Problem
We don't want to trigger when there is immediate right context on the same line, with "}" as an exception
This will be an A/B test
## License

### DIFF
--- a/src/test/codewhisperer/service/keyStrokeHandler.test.ts
+++ b/src/test/codewhisperer/service/keyStrokeHandler.test.ts
@@ -20,6 +20,7 @@ import { RecommendationHandler } from '../../../codewhisperer/service/recommenda
 import { isInlineCompletionEnabled } from '../../../codewhisperer/util/commonUtil'
 import { ClassifierTrigger } from '../../../codewhisperer/service/classifierTrigger'
 import { CodeWhispererUserGroupSettings } from '../../../codewhisperer/util/userGroupUtil'
+import * as CodeWhispererConstants from '../../../codewhisperer/models/constants'
 
 const performance = globalThis.performance ?? require('perf_hooks').performance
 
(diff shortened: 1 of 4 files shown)


justinmk3 (changes requested):
The test cases are all nearly identical. Redundant code should be avoided, it is a maintenance burden. Extract the common code to a function. This is a general guideline and should be a general practice for all code, not only this PR.

image

zixlin7:
Please don't merge this PR yet

src/test/codewhisperer/service/keyStrokeHandler.test.ts
```diff
@@ -213,6 +213,62 @@ describe('keyStrokeHandler', function () {
...
         })
+
+        it('Should skip invoking if there is immediate right context on the same line and not a single }', async function () {
+            const casesForSuppressTokenFilling = [
+                {
+                    fileContent: 'function addTwo',
+                    line: 1,
+                    char: 9,
+                    shouldInvoke: false,
+                },
```
Will-ShaoHua, line 224: just for my clarification on the test cases, for this one, it's like `{function addTwo` and the cursor is located between char `n` and char space ` ` right?

Will-ShaoHua: and keyStrokeHandler shouldn't proceed the invocation since next token is space ` `

zixlin7: the selection is at line 1 char 9 so it would be `function {addTwo`, it should not trigger because the `addTwo` is immediate right context and same line

zixlin7:
need another update

src/test/codewhisperer/service/keyStrokeHandler.test.ts
zixlin7: I updated the tests to be easier to understand

src/test/codewhisperer/service/keyStrokeHandler.test.ts
```diff
@@ -213,6 +214,62 @@ describe('keyStrokeHandler', function () {
...
+                await testIfRightContextShouldInvoke(
+                    rightContext,
+                    shouldInvoke,
+                    CodeWhispererConstants.UserGroup.RightContext
+                )
+            })
+        })
+
+        it('Should not skip invoking based on right context for control group', async function () {
+            await testIfRightContextShouldInvoke('add', true, CodeWhispererConstants.UserGroup.Control)
```
justinmk3, line 255: Thanks for de-duping this. Are there other cases (in old existing code) that could use `testIfRightContextShouldInvoke` ?

justinmk3 (approved):

Will-ShaoHua (approved):

src/test/codewhisperer/service/keyStrokeHandler.test.ts
zixlin7: yes I'll follow up with another PR to rector the other tests

PR #3894, opened 2023-10-13
auth: show wizard in c9 instead of webview

nkomonen-amazon:
- Will now show the auth wizard instead of the auth webview to users when in c9 classic or codecatalyst
- There will be no automatic opening of any auth setup when the user is a first time user in c9
- Renamed method to `showManageConnections()` since there are different outcomes depending on the IDE now.

## License

### DIFF
--- a/src/auth/activation.ts
+++ b/src/auth/activation.ts
@@ -11,8 +11,9 @@ import { fromString } from './providers/credentials'
 import { registerCommandsWithVSCode } from '../shared/vscode/commands2'
 import { AuthCommandBackend, AuthCommandDeclarations } from './commands'
 import { getLogger } from '../shared/logger'
-import { isInDevEnv } from '../codecatalyst/utils'
 import { ExtensionUse } from './utils'
+import { isCloud9 } from '../shared/extensionUtilities'
+import { isInDevEnv } from '../codecatalyst/utils'
 
 export async function initialize(
     extensionContext: vscode.ExtensionContext,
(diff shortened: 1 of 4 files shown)


src/auth/activation.ts
```diff
@@ -48,10 +48,10 @@ async function showManageConnectionsOnStartup() {
         return
     }
 
-    if (isInDevEnv()) {
-        getLogger().debug('firstStartup: Detected we are in Dev Env, skipping showing Add Connections page.')
+    if (isCloud9('any')) {
+        getLogger().debug('firstStartup: Detected we are in Cloud9, skipped showing Add Connections page.')
```
justinmk3, line 52: but what about non-cloud9 devenvs?

src/auth/activation.ts
```diff
@@ -48,10 +48,10 @@ async function showManageConnectionsOnStartup() {
         return
     }
 
-    if (isInDevEnv()) {
-        getLogger().debug('firstStartup: Detected we are in Dev Env, skipping showing Add Connections page.')
+    if (isCloud9('any')) {
```
justinmk3, line 51: i think it was intended that any kind of devenv (not only cloud9 devenvs) should not show this page. 

```suggestion
    if (isInDevEnv() || isCloud9('any')) {
```

nkomonen-amazon: Thanks for catching, the vscode connected dev env slipped my mind

src/auth/activation.ts
```diff
@@ -41,17 +42,20 @@ export async function initialize(
...
+        reason = 'This is not the users first use of the extension'
+    } else if (isInDevEnv()) {
+        reason = 'The user is in a Dev Evironment'
+    } else if (isCloud9('any')) {
+        reason = 'The user is in Cloud9'
     }
-
-    if (isInDevEnv()) {
-        getLogger().debug('firstStartup: Detected we are in Dev Env, skipping showing Add Connections page.')
+    if (!reason) {
```
justinmk3, line 54:
```suggestion
    if (reason) {
```

nkomonen-amazon: 🤦

justinmk3 (approved):

PR #3895, opened 2023-10-13
codewhisperer: fix bash not supported in C9

zixlin7:
## Problem
cloud9 has different language id for bash/shellscript
## Solution
add that id to supported list

## License

### DIFF
--- a/src/codewhisperer/models/constants.ts
+++ b/src/codewhisperer/models/constants.ts
@@ -101,6 +101,7 @@ export const supportedLanguages = [
     'rust',
     'scala',
     'shellscript',
+    'sh', // Cloud9 reports bash files with this language-id
     'sql',
 ] as const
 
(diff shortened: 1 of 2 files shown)


src/codewhisperer/models/constants.ts
leigaol, line 1: What does telemetry report when it comes to sh? Is it merged with shell?

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -66,6 +67,7 @@ export class RuntimeLanguageContext {
             scala: 'scala',
             shellscript: 'sh',
             sql: 'sql',
+            sh: 'sh',
```
justinmk3, line 70: alphbetical order is helpful

src/codewhisperer/models/constants.ts
zixlin7: yes it's shell

src/codewhisperer/util/runtimeLanguageContext.ts
zixlin7: done

leigaol (approved):

PR #3908, opened 2023-10-17
codewhisperer: security hover update range and format

ctlai95:
## Problem

Hovers positions are not being updated if the document changes.

## Solution

Add an event handler which either removes the hover or offsets the line number depending on what was changed (similar to how [diagnostics are updated](https://github.com/aws/aws-toolkit-vscode/blob/0974becd8d26bc12a640b34bfbefc2e20fe8cb17/src/codewhisperer/service/diagnosticsProvider.ts#L61-L93
)) 
## License

### DIFF
--- a/src/codewhisperer/service/securityIssueHoverProvider.ts
+++ b/src/codewhisperer/service/securityIssueHoverProvider.ts
@@ -5,6 +5,7 @@
 import * as vscode from 'vscode'
 import { AggregatedCodeScanIssue, CodeScanIssue } from '../models/model'
 import globals from '../../shared/extensionGlobals'
+import { getLineOffset } from './diagnosticsProvider'
 
 export class SecurityIssueHoverProvider implements vscode.HoverProvider {
     static #instance: SecurityIssueHoverProvider
@@ -39,6 +40,32 @@ export class SecurityIssueHoverProvider implements vscode.HoverProvider {
         }
 
         return new vscode.Hover(contents)
+    }
+
+    public updateRanges(event: vscode.TextDocumentChangeEvent) {
+        const changedRange = event.contentChanges[0].range
+        const changedText = event.contentChanges[0].text
+        const lineOffset = getLineOffset(changedRange, changedText)
+
+        this._issues = this._issues.map(issues => this._applyRangeOffset(event.document.fileName, issues, lineOffset))
+    }
+
+    private _applyRangeOffset(
+        fileName: string,
+        aggregatedIssues: AggregatedCodeScanIssue,
+        lineOffset: number
+    ): AggregatedCodeScanIssue {
+        if (aggregatedIssues.filePath !== fileName) {
+            return aggregatedIssues
+        }
+        return {
+            ...aggregatedIssues,
+            issues: aggregatedIssues.issues.map(issue => ({
+                ...issue,
+                startLine: issue.startLine + lineOffset,
+                endLine: issue.endLine + lineOffset,
+            })),
+        }
     }
 
     private _getContent(issue: CodeScanIssue) {
@@ -94,10 +121,24 @@ export class SecurityIssueHoverProvider implements vscode.HoverProvider {
      * @returns The markdown string
      */
     private _makeCodeBlock(code: string, language?: string) {
-        // Add some padding so that each line has the same number of chars
         const lines = code.split('\n').slice(1) // Ignore the first line for diff header
+        // Get the length of the longest line to pad each line to be the same length
         const maxLineChars = lines.reduce((acc, curr) => Math.max(acc, curr.length), 0)
-        const paddedLines = lines.map(line => line.padEnd(maxLineChars + 2))
+        // Get the number of leading whitespaces that can be removed to hide unnecessary indentation
+        const minLeadingWhitespaces = lines.reduce((acc, curr) => {
+            const numWhitespaces = curr.slice(1).search(/\S/)
+            if (numWhitespaces < 0) {
+                return acc
+            }
+            return Math.min(acc, numWhitespaces)
+        }, maxLineChars)
+        const paddedLines = lines.map((line, i) => {
+            const paddedLine = line.padEnd(maxLineChars + 2)
+            if (minLeadingWhitespaces > 1) {
+                return paddedLine[0] + paddedLine.slice(minLeadingWhitespaces)
+            }
+            return paddedLine
+        })
 
         // Group the lines into sections so consecutive lines of the same type can be placed in
         // the same span below
@@ -128,8 +169,9 @@ export class SecurityIssueHoverProvider implements vscode.HoverProvider {
 ${section}
 \`\`\`
 
-</span>`
+</span>
+`
             )
-            .join('')
+            .join('<br />')
     }
 }
--- a/src/codewhisperer/service/securityIssueHoverProvider.ts
+++ b/src/codewhisperer/service/securityIssueHoverProvider.ts
@@ -5,7 +5,6 @@
 import * as vscode from 'vscode'
 import { AggregatedCodeScanIssue, CodeScanIssue } from '../models/model'
 import globals from '../../shared/extensionGlobals'
-import { getLineOffset } from './diagnosticsProvider'
 
 export class SecurityIssueHoverProvider implements vscode.HoverProvider {
     static #instance: SecurityIssueHoverProvider
@@ -45,9 +44,15 @@ export class SecurityIssueHoverProvider implements vscode.HoverProvider {
     public updateRanges(event: vscode.TextDocumentChangeEvent) {
         const changedRange = event.contentChanges[0].range
         const changedText = event.contentChanges[0].text
-        const lineOffset = getLineOffset(changedRange, changedText)
+        const lineOffset = this._getLineOffset(changedRange, changedText)
 
         this._issues = this._issues.map(issues => this._applyRangeOffset(event.document.fileName, issues, lineOffset))
+    }
+
+    private _getLineOffset(range: vscode.Range, text: string) {
+        const originLines = range.end.line - range.start.line + 1
+        const changedLines = text.split('\n').length
+        return changedLines - originLines
     }
 
     private _applyRangeOffset(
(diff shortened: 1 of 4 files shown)


src/codewhisperer/service/securityIssueHoverProvider.ts
```diff
@@ -94,10 +126,24 @@ export class SecurityIssueHoverProvider implements vscode.HoverProvider {
      * @returns The markdown string
      */
     private _makeCodeBlock(code: string, language?: string) {
-        // Add some padding so that each line has the same number of chars
         const lines = code.split('\n').slice(1) // Ignore the first line for diff header
+        // Get the length of the longest line to pad each line to be the same length
         const maxLineChars = lines.reduce((acc, curr) => Math.max(acc, curr.length), 0)
-        const paddedLines = lines.map(line => line.padEnd(maxLineChars + 2))
+        // Get the number of leading whitespaces that can be removed to hide unnecessary indentation
+        const minLeadingWhitespaces = lines.reduce((acc, curr) => {
```
justinmk3, line 133: would you mind generalizing this as a `dedent()` function which can live next to [indent()](https://github.com/aws/aws-toolkit-vscode/blob/4a9fea7bd2383a9f9bd85839d57acc98a98fdba1/src/shared/utilities/textUtilities.ts#L40)

src/codewhisperer/service/securityIssueHoverProvider.ts
```diff
@@ -41,6 +41,38 @@ export class SecurityIssueHoverProvider implements vscode.HoverProvider {
         return new vscode.Hover(contents)
     }
 
+    public updateRanges(event: vscode.TextDocumentChangeEvent) {
```
justinmk3, line 44: This is a case where a docstring is desperately needed. What is the purpose of this function?
```suggestion
    /** .... */
    public updateRanges(event: vscode.TextDocumentChangeEvent) {
```

ctlai95: Sure, will add a docstring here. The purpose of this function is to offset the hover positions when the document changes, basically the same thing for [diagnostics](https://github.com/aws/aws-toolkit-vscode/blob/d2f03d52a22109bf6b4c71d04581be171b59a5e3/src/codewhisperer/service/diagnosticsProvider.ts#L83-L88).

src/codewhisperer/service/securityIssueHoverProvider.ts
ctlai95: After more consideration I decided to remove this, as it might cause some confusion about indentation in the suggested fix (especially in python where indentation is important)

src/codewhisperer/service/securityIssueHoverProvider.ts
```diff
@@ -41,6 +41,44 @@ export class SecurityIssueHoverProvider implements vscode.HoverProvider {
...
+     */
+    public updateRanges(event: vscode.TextDocumentChangeEvent) {
+        const changedRange = event.contentChanges[0].range
+        const changedText = event.contentChanges[0].text
+        const lineOffset = this._getLineOffset(changedRange, changedText)
+
+        this._issues = this._issues.map(issues => this._applyRangeOffset(event.document.fileName, issues, lineOffset))
+    }
+
+    private _getLineOffset(range: vscode.Range, text: string) {
```
mj632, line 58: What if the user deleted lines of code where an issue was showing previously? The offset will push the hover content on wrong line numbers. We should remove that issue. 

possible scenarios - 
1. If user removes all the lines where a issue was highlighted.
2. If user partially deletes the lines where a issue is present
3. If user deletes lines such that two issues are getting merged to each other.

src/codewhisperer/service/securityIssueHoverProvider.ts
```diff
@@ -41,6 +41,44 @@ export class SecurityIssueHoverProvider implements vscode.HoverProvider {
...
+        const originLines = range.end.line - range.start.line + 1
+        const changedLines = text.split('\n').length
```
mj632, line 60: Please verify if this split is not extra splitting on any `\n` which user has written as part of any string variables.

src/codewhisperer/service/securityIssueHoverProvider.ts
```diff
@@ -41,6 +41,44 @@ export class SecurityIssueHoverProvider implements vscode.HoverProvider {
...
+        lineOffset: number
+    ): AggregatedCodeScanIssue {
+        if (aggregatedIssues.filePath !== fileName) {
+            return aggregatedIssues
+        }
+        return {
+            ...aggregatedIssues,
+            issues: aggregatedIssues.issues.map(issue => ({
+                ...issue,
+                startLine: issue.startLine + lineOffset,
```
mj632, line 76: the line numbers should only be updated if the text change in file happened before this startLine. Seems like this will update the start and end lines for some issues which are not impacted by the text change in the file.

e.g., if change of text in a file is between line numbers 25 and 30 then the issues should only be updated starting from line number 25.

src/test/codewhisperer/service/securityIssueHoverProvider.test.ts
```diff
@@ -99,5 +102,30 @@ describe('securityIssueHoverProvider', () => {
             const actual = securityIssueHoverProvider.provideHover(mockDocument, new vscode.Position(2, 0), token.token)
             assert.strictEqual(actual.contents.length, 0)
         })
+
+        it('should update ranges if text document changed', () => {
```
mj632, line 106: Please add more test to verify various scenarios of text changes in a file.

src/codewhisperer/service/securityIssueHoverProvider.ts
ctlai95: Good catch, updating it

src/codewhisperer/service/securityIssueHoverProvider.ts
ctlai95: Verified it is working as expected

src/codewhisperer/service/securityIssueHoverProvider.ts
ctlai95: Good catch, updating it

src/test/codewhisperer/service/securityIssueHoverProvider.test.ts
```diff
@@ -99,5 +102,81 @@ describe('securityIssueHoverProvider', () => {
...
+                    issues: [makeIssue({ startLine: 1, endLine: 2 })],
+                },
+            ]
+            const token = new vscode.CancellationTokenSource()
+            let actual = securityIssueHoverProvider.provideHover(mockDocument, new vscode.Position(1, 0), token.token)
+            assert.strictEqual(actual.contents.length, 1)
+
+            const changeEvent = createTextDocumentChangeEvent(
+                mockDocument,
+                new vscode.Range(new vscode.Position(0, 0), new vscode.Position(0, 1)),
```
nkomonen-amazon, line 145: Does this range mean that the `d` in `def two_sum...` is replaced with a `\n`. Should it just be inserted?
```suggestion
                new vscode.Range(new vscode.Position(0, 0), new vscode.Position(0, 0)),
```

src/test/codewhisperer/service/securityIssueHoverProvider.test.ts
```diff
@@ -99,5 +102,81 @@ describe('securityIssueHoverProvider', () => {
...
+                '\n'
+            )
+            securityIssueHoverProvider.handleDocumentChange(changeEvent)
+
+            actual = securityIssueHoverProvider.provideHover(mockDocument, new vscode.Position(0, 0), token.token)
+            assert.strictEqual(actual.contents.length, 0)
+        })
+
+        it('should offset hover if document changed on a line above it', () => {
```
nkomonen-amazon, line 130:
```suggestion
        it('offsets the existing issue down a line if a new line is inserted above', () => {
```

src/test/codewhisperer/service/securityIssueHoverProvider.test.ts
```diff
@@ -99,5 +102,81 @@ describe('securityIssueHoverProvider', () => {
...
+            securityIssueHoverProvider.handleDocumentChange(changeEvent)
+
+            actual = securityIssueHoverProvider.provideHover(mockDocument, new vscode.Position(1, 0), token.token)
+            assert.strictEqual(actual.contents.length, 0)
+
+            actual = securityIssueHoverProvider.provideHover(mockDocument, new vscode.Position(2, 0), token.token)
+            assert.strictEqual(actual.contents.length, 1)
+        })
+
+        it('should not offset hover if document changed on a line below it', () => {
```
nkomonen-amazon, line 157:
```suggestion
        it('does not move the issue if the document changed below the line', () => {
```

src/test/codewhisperer/service/securityIssueHoverProvider.test.ts
```diff
@@ -99,5 +102,81 @@ describe('securityIssueHoverProvider', () => {
             const actual = securityIssueHoverProvider.provideHover(mockDocument, new vscode.Position(2, 0), token.token)
             assert.strictEqual(actual.contents.length, 0)
         })
+
+        it('should remove hover if document changed on the same line', () => {
```
nkomonen-amazon, line 106: When naming these the language should be more direct

```suggestion
        it('removes the issue hover if the document change on the same line', () => {
```

nkomonen-amazon (approved):

mj632 (approved):

src/codewhisperer/service/securityIssueHoverProvider.ts
```diff
@@ -41,6 +41,51 @@ export class SecurityIssueHoverProvider implements vscode.HoverProvider {
...
+            return {
+                ...group,
+                issues: group.issues
+                    .filter(issue => {
+                        const range = new vscode.Range(issue.startLine, 0, issue.endLine, 0)
+                        const intersection = changedRange.intersection(range)
+                        return !(intersection && (/\S/.test(changedText) || changedText === ''))
+                    })
+                    .map(issue => {
+                        const range = new vscode.Range(issue.startLine, 0, issue.endLine, 0)
```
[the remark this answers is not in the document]
ctlai95: Will update in the next PR

PR #3913, opened 2023-10-18
web: remove various commands

bryceitoc9:
## Problem
The web extension is showing a lot of commands that we don't want to show.

## Solution
Add a lot of `!isWeb` to command, debugger, and view registrations. This will render those code paths inaccessible directly via command, but will not fix any underlying compatibility issues if the code paths are exercised by alternate means.
* Using `isWeb` will still show these commands if we're running through the `Extension (browser)` debug target. We can create our own context, but since that requires the context to be written, we will see the views and commands pop in and out once the context is written. We cannot overwrite the `isWeb` context, as well.
  * For now, use `npm run runInBrowser` (no debugger support though)
* Limiting on the command (instead of at the menu level) should still prevent these from showing up in menus (like the command palette)

Currently-exposed commands, and proof that the AWS explorer view is removed:
image

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -281,6 +281,7 @@
         "debuggers": [
             {
                 "type": "aws-sam",
+                "when": "!isWeb",
                 "label": "%AWS.configuration.description.awssam.debug.label%",
                 "configurationAttributes": {
                     "direct-invoke": {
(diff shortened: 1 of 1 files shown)


justinmk3 (approved):
Thank you for the clear explanation!

PR #3916, opened 2023-10-18
web: launch web browser with web extension, attach debugger

bryceitoc9:
## Problem
It's a bit clunky to get a browser up with the extension, and it doesn't connect to VS Code's debugger

## Solution
This solves 2/3 of the problem:
* launch config for the web extension in a web browser!
* attaches the VS Code debugger to the chromium instance
* HOWEVER! This does not attach the debugger to the running extension code: VS Code doesn't appear to pick it up. I'm pretty sure this is running in a web worker, and I'm not sure if VS Code supports debugging these (code is running in the `WebExtensionHost` as `extensionWeb.js`): 
image
  * To see what VS Code attaches to, hover over the breakpoint, click "troubleshoot your launch config", and click "What scripts and sourcemaps are loaded" in the webview popup
image
image

Feel free to merge if F5 launch to the web browser feels useful, otherwise we can put this on ice until we can actually hook up the debugger to the worker.

## License


justinmk3:
Nice, should we add an item here: https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#run

PR #3918, opened 2023-10-19
feat(codewhisperer): apply security fix command

mj632:
## Problem
Apply fix button click should update the code with appropriate patch suggested by Security scan.

## Solution
Added a new function `patchVulnerableCode` to patch the code diff suggested by Security scan when user clicks on the Apply fix link in hover. This will run a patch command on the file where the user has clicked on Apply fix in hover. 
As a result user will see a notification of the operation.
 

https://github.com/aws/aws-toolkit-vscode/assets/26518489/cf288700-b53e-419f-94c0-d9c9a59e5ec3

## License


src/codewhisperer/activation.ts
```diff
@@ -4,57 +4,58 @@
...
-import { resetIntelliSenseState } from './util/globalStateUtil'
-import { CodeWhispererSettings } from './util/codewhispererSettings'
-import { ExtContext } from '../shared/extensions'
 import { TextEditorSelectionChangeKind } from 'vscode'
-import { CodeWhispererTracker } from './tracker/codewhispererTracker'
-import * as codewhispererClient from './client/codewhisperer'
-import { runtimeLanguageContext } from './util/runtimeLanguageContext'
-import { getLogger } from '../shared/logger'
 import { isCloud9 } from '../shared/extensionUtilities'
+import { ExtContext } from '../shared/extensions'
```
ctlai95, line 9: Can we undo these formatting changes, its causing too much noise

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -189,3 +189,14 @@ export const notifyNewCustomizationsCmd = Commands.declare(
         notifyNewCustomizations().then()
     }
 )
+
+export const applySecurityFix = Commands.declare(
+    'aws.codewhisperer.applySecurityFix',
+    () => async (codeScanIssue: CodeScanIssue) => {
+        const activeEditor = vscode.window.activeTextEditor
+        if (activeEditor) {
+            const filePath = activeEditor.document.uri.fsPath
```
ctlai95, line 198: What's the difference between this and `activeEditor.document.fileName`?

src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -201,3 +207,46 @@ export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrl
...
+            }
+        })
+        fs.stat(`${filePath}.rej`, function (_err, stat) {
+            if (stat) {
+                fsPromises.rm(`${filePath}.rej`)
+            }
+        })
+        SecurityIssueHoverProvider.instance.removeFixedIssue(filePath, issue)
+    } else {
+        vscode.window.showErrorMessage(`Error applying suggested fix. Exit code: ${code}`)
```
ctlai95, line 250: In this case, isn't there still a chance of `*.orig` or `*.rej` files being created?

src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -201,3 +207,46 @@ export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrl
...
     getLogger().debug(`StatusCode: ${response.statusCode}`)
 }
+
+export async function patchVulnerableCode({ filePath, issues: [issue] }: AggregatedCodeScanIssue) {
+    const [suggestedFix] = issue.suggestedFixes ?? []
+    if (!suggestedFix) {
+        return
+    }
+    const patch = suggestedFix.code
+    const process = new ChildProcess('patch', [filePath])
```
ctlai95, line 217: Did we consider how this runs on windows? or if `patch` doesn't exist?

ctlai95, line 217: Will it require the file to be saved? If the file is dirty can the patch still be applied?

src/codewhisperer/activation.ts
mj632: sure. Updating it

src/codewhisperer/commands/basicCommands.ts
mj632: Both are same. Shorthand notation for TextDocument.uri.fsPath. Updating to fileName

src/codewhisperer/service/securityScanHandler.ts
mj632: moving it out side `if..else`

src/codewhisperer/service/securityScanHandler.ts
mj632: `patch` is available as a built-in command under Windows Subsystem for Linux or any third-party tools like Git Bash, Cygwin, or MinGW. The command should work out in majority of the cases. However, will verify in which scenario the patch fails for windows.

mj632: If we apply a patch on unsaved file then if the code where patch is going to be applied is matching with the diff then it will update the vulnerable code similar to what any git pull or apply stash command does. For unsaved content, it will remain unsaved and when user tries to save it, it will show an error notification so that user can either discard the unsaved changes or compare the changes between  saved file after diff applied and the open unsaved file. User can manually override the changes similar to apply git stash command.

src/codewhisperer/service/securityIssueHoverProvider.ts
```diff
@@ -41,6 +41,19 @@ export class SecurityIssueHoverProvider implements vscode.HoverProvider {
         return new vscode.Hover(contents)
     }
 
+    public removeFixedIssue(filePath: string, fixedIssue: CodeScanIssue) {
```
ctlai95, line 44: We can probably remove this once my changes are in, since applying the fix should trigger the document change event and clean up the issues.

src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -201,3 +207,45 @@ export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrl
     })
     getLogger().debug(`StatusCode: ${response.statusCode}`)
 }
+
+export async function patchVulnerableCode({ filePath, issues: [issue] }: AggregatedCodeScanIssue) {
```
ctlai95, line 211: Why is this the function signature? From the caller it looks like you can apply multiple patches but only the first one is being taken?

src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -201,3 +207,45 @@ export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrl
...
+    fs.stat(`${filePath}.orig`, function (_err, stat) {
+        if (stat) {
+            fsPromises.rm(`${filePath}.orig`)
+        }
+    })
+    fs.stat(`${filePath}.rej`, function (_err, stat) {
+        if (stat) {
+            fsPromises.rm(`${filePath}.rej`)
+        }
+    })
```
ctlai95, line 250: I saw that there is a `FileSystemCommon` class, can we use that?

src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -201,3 +207,45 @@ export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrl
...
+    const process = new ChildProcess('patch', [filePath])
+
+    const r = process.run()
+    await process.send(patch)
+    await process.send(eof)
+    const code = (await r).exitCode
+
+    if (code === 0) {
+        vscode.window
+            .showInformationMessage(
```
ctlai95, line 226: Let's add this message to `src/codewhisperer/models/constants.ts`

src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -201,3 +207,45 @@ export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrl
...
+                'Suggested Code fix was applied successfully. Run the security scan to validate the fix. To revert back to the original code use the undo command.',
```
ctlai95, line 227: The casing/grammar looks a bit off here

```suggestion
                'Suggested code fix was applied successfully. Run a security scan to validate the fix. To revert back to the original code, use the undo command.',
```

src/codewhisperer/service/securityIssueHoverProvider.ts
mj632: yes, we no longer need this method. removing it.

src/codewhisperer/service/securityScanHandler.ts
mj632: updating to use `FileSystemCommon` class.

src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -201,3 +205,38 @@ export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrl
...
+            .showInformationMessage(CodeWhispererConstants.codeFixAppliedSuccessMessage, {
+                title: CodeWhispererConstants.runSecurityScanButtonTitle,
+            })
+            .then(res => {
+                if (res?.title === CodeWhispererConstants.runSecurityScanButtonTitle) {
+                    vscode.commands.executeCommand('aws.codeWhisperer.security.scan')
+                }
+            })
+    } else {
+        vscode.window.showErrorMessage(`Error applying suggested fix. Exit code: ${code}`)
```
justinmk3, line 233: ChildProcess [enables logging](https://github.com/aws/aws-toolkit-vscode/blob/8398a52552baffb991f99b18b02e6ec9403dac86/src/shared/utilities/childProcess.ts#L24) by default, so it's fine that we don't log here.

But what is the user supposed to do with an exit code here? Consider using showViewLogsMessage: https://github.com/aws/aws-toolkit-vscode/blob/8398a52552baffb991f99b18b02e6ec9403dac86/src/shared/utilities/messages.ts#L92

src/codewhisperer/models/constants.ts
```diff
@@ -192,7 +192,7 @@ export const referenceLogText = (
 ) =>
     `with code ${code} provided with reference under ${license} from repository ${repository}. Added to ${filePath} ${lineInfo}.`
 
-export const referenceLogPromptText = `Don\'t want suggestions that include code with references? Uncheck this option in 
+export const referenceLogPromptText = `Don\'t want suggestions that include code with references? Uncheck this option in
```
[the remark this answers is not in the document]
justinmk3: Er... please disable what extension is doing that.

src/codewhisperer/service/securityScanHandler.ts
mj632: Yes, after verifying in Windows platform. It seems like it is failing in windows OS. I am updating it to use `diff` package to avoid platform dependency.

src/codewhisperer/service/securityScanHandler.ts
mj632: After moving to [diff](https://www.npmjs.com/package/diff), `applyPatch` method does not throw any error instead it returns `false` to mention that the patch did not applied. I have updated the message to tell user about manually applying the patch.

ctlai95 (approved):

ctlai95 (approved):

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -197,3 +198,42 @@ export const notifyNewCustomizationsCmd = Commands.declare(
...
+        if (updatedContent) {
+            activeEditor.edit(editBuilder => {
+                const range = new vscode.Range(0, 0, lineCount, 0)
+                editBuilder.replace(range, updatedContent)
+            })
+            activeEditor.document.save().then(isSaved => {
+                if (isSaved) {
+                    vscode.window
+                        .showInformationMessage(CodeWhispererConstants.codeFixAppliedSuccessMessage, {
+                            title: CodeWhispererConstants.runSecurityScanButtonTitle,
```
justinmk3, line 226: Is showing a message necessary or could it be a statusbar message? Imagine if the editor showed a popup message every time you used a "quick fix" or other action--that would be very noisy.

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -197,3 +198,42 @@ export const notifyNewCustomizationsCmd = Commands.declare(
...
+
+        const updatedContent = applyPatch(fileContent, patch)
+
+        if (updatedContent) {
+            activeEditor.edit(editBuilder => {
+                const range = new vscode.Range(0, 0, lineCount, 0)
+                editBuilder.replace(range, updatedContent)
+            })
+            activeEditor.document.save().then(isSaved => {
+                if (isSaved) {
```
justinmk3, line 223: when would this be false? should details be logged for troubleshooting?

src/codewhisperer/models/constants.ts
```diff
@@ -286,6 +286,14 @@ export const isClassifierEnabledKey = 'CODEWHISPERER_CLASSIFIER_TRIGGER_ENABLED'
 
 export const supplemetalContextFetchingTimeoutMsg = 'codewhisperer supplemental context fetching timeout'
 
+export const codeFixAppliedSuccessMessage =
+    'Suggested code fix was applied successfully. Run a security scan to validate the fix. To revert back to the original code, use the undo command.'
```
justinmk3, line 290: - Never need to mention "successfully", it's already implied.
- Is it worth the text to mention "undo"? That is a basic editor feature.
```suggestion
    'Code fix was applied. Run a security scan to validate the fix. To revert back to the original code, use the undo command.'
```

src/test/codewhisperer/commands/basicCommands.test.ts
```diff
@@ -131,5 +135,67 @@ describe('CodeWhisperer-basicCommands', function () {
...
+                selection: undefined,
+            }
+            const sandbox = sinon.createSandbox()
+            sandbox.stub(vscode.window, 'activeTextEditor').value(activeTextEditorMock)
+
+            targetCommand = testCommand(applySecurityFix)
+            await targetCommand.execute(createCodeScanIssue())
+
+            assert.ok(documentEditStub.calledOnce)
+            assert.ok(documentSaveStub.calledOnce)
```
justinmk3, line 161: why not check the editor state? checking internal details with mocks/spies is brittle. vscode editor has a "modified" flag that you can check.

src/test/codewhisperer/commands/basicCommands.test.ts
```diff
@@ -131,5 +135,67 @@ describe('CodeWhisperer-basicCommands', function () {
...
+            assert.strictEqual(
+                getTestWindow().shownMessages[0].message,
+                'Suggested code fix was applied successfully. Run a security scan to validate the fix. To revert back to the original code, use the undo command.'
+            )
+
+            sandbox.restore()
+        })
+        it('should fail to apply patch', async function () {
```
justinmk3, line 169: please make sure that test names are meaningful.
```suggestion
        it('handles patch failure', async function () {
```

src/codewhisperer/models/constants.ts
```diff
@@ -286,6 +286,14 @@ export const isClassifierEnabledKey = 'CODEWHISPERER_CLASSIFIER_TRIGGER_ENABLED'
...
+
+export const codeFixAppliedFailedMessage =
+    'Error while applying the suggested code fix. Please copy the suggested code and replace the vulnerable lines of code.'
```
justinmk3, line 293: - "Error" is misleading and implies there was an exception or other failure details that can be inspected. But instead, IIUC, the unified diff simply failed to apply to the code.
- "Please" is never useful in messages, and uses valuable space. "Copy and paste the code" also seems very obvious...

```suggestion
    'Failed to apply suggested code fix.'
```

src/codewhisperer/commands/basicCommands.ts
mj632: https://github.com/aws/aws-toolkit-vscode/assets/26518489/4e64ebc4-f432-4aa8-bf07-ebd2d2da5e5d

justinmk3: If it happens frequently then don't log. If it's rare, I'd still add a log

src/codewhisperer/commands/basicCommands.ts
mj632: We discussed about showing notification, the reason for showing this notification is that user can run a new scan quickly from the notification popup only. Also, since the popup gets dismissed after ~10 seconds or re-uses the same notification model when applying multiple fixes for different issues, it won't stack up the notification.

We are targeting to keep this notification until we have proactive scanning feature. With that we will remove this notification and only show notification in case of error. In addition, we are going to collect the user data from upcoming testing to see if they find it useful or not.

justinmk3 (approved):

Issue #3910, opened 2023-10-18
How to Disable the 'Learn CodeWhisperer' Page from Appearing on VSCode Startup


LHR-lhr:
## System details (run the `AWS: About Toolkit` command)

-   OS: Windows_NT x64 10.0.19045
-   Visual Studio Code version: 1.83.1
-   AWS Toolkit version: 1.95.0

## Question

Every time I launch Visual Studio Code (VSCode), a 'Learn CodeWhisperer' page automatically pops up. I am looking for a solution to disable this page from appearing on startup.

### Steps to Reproduce:

1、Launch Visual Studio Code.
2、Observe the automatic appearance of the 'Learn CodeWhisperer' page.

### Expected Behavior:
I would like to find a way to prevent the 'Learn CodeWhisperer' page from appearing every time I start VSCode.

![image](https://github.com/aws/aws-toolkit-vscode/assets/48155416/7624dbfb-2a3c-414c-b049-d9c6bacacd02)

justinmk3:
This page should not show after the first start. That decision is remembered in vscode setting `aws.suppressPrompts`. 

In vscode settings, search for `aws.suppressPrompts` , what does it show? Is `codeWhispererNewWelcomeMessage` checked? 

<img width="652" alt="image" src="https://github.com/aws/aws-toolkit-vscode/assets/55561878/0c854bac-79c4-493c-8196-3722699e63fe">

Is it possible your settings.json file is getting overwritten by some process?

LHR-lhr:
Thank you so much! This solved my issue! 

I couldn't find the `codeWhispererNewWelcomeMessage` setting when I searched for `aws.suppressPrompt`s in VSCode settings. 
![image](https://github.com/aws/aws-toolkit-vscode/assets/48155416/3546fc98-540c-4e53-a69e-46f22d07597b)

So, I added the following to my `settings.json`:
```
"aws.suppressPrompts": {
     "codeWhispererNewWelcomeMessage": true,
 }
```

justinmk3:
That suggests something is going wrong with the vscode settings system. 

Can you share your logs for `Extension Host`:
![image](https://github.com/aws/aws-toolkit-vscode/assets/55561878/0e4ca110-953b-4f48-9b55-c4d01e92dacc)

and also the AWS Toolkit logs (check that private info is removed)? 

1. In VSCode settings, ensure that "aws log level" is "debug".
2. Perform the steps to reproduce the issue.
[...]

PR #3924, opened 2023-10-20
feat(settings): check that user settings.json is valid

justinmk3:
Problem:
vscode notifies the user if `settings.json` is complete nonsense, but is silent if there are only "recoverable" JSON syntax errors. So _reading_ settings will appear to work, but _writing_ settings will throw an exception. #3910

Solution:
On startup, check that we can actually _write_ settings, and show an error message if not.

image

## License


nkomonen-amazon:
Nice find

PR #3925, opened 2023-10-20
auth: ellipsis + right click options for `Developer Tools`

nkomonen-amazon:
This commit allows user to right click to root CW or CC
Node or click the new ellipsis icon, and they will get
the options to Add Connection or Signout (if appropriate)

Screenshot 2023-10-20 at 10 50 42 AM

Kapture 2023-10-20 at 10 36 27

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -785,7 +785,7 @@
                     "when": "false"
                 },
                 {
-                    "command": "aws.codeWhisperer.removeConnection",
+                    "command": "aws.codeWhisperer.signout",
                     "when": "false"
                 },
                 {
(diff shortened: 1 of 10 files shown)


src/codecatalyst/auth.ts
```diff
@@ -45,6 +45,10 @@ export const isValidCodeCatalystConnection = (conn: Connection): conn is SsoConn
 export const isUpgradeableConnection = (conn: Connection): conn is SsoConnection =>
     isBuilderIdConnection(conn) && !isValidCodeCatalystConnection(conn)
 
+export function setCodeCatalystConnectedContext(isConnected: boolean) {
+    return vscode.commands.executeCommand('setContext', 'CODECATALYST_CONNECTED', isConnected)
```
justinmk3, line 49:
```suggestion
    return vscode.commands.executeCommand('setContext', 'aws.codecatalyst.connected', isConnected)
```

package.json
```diff
@@ -1981,6 +1998,32 @@
...
                 }
+            ],
+            "aws.codecatalyst.submenu": [
+                {
+                    "command": "aws.codecatalyst.manageConnections",
+                    "group": "0@1"
+                },
+                {
+                    "command": "aws.codecatalyst.signout",
+                    "when": "!isCloud9 && CODECATALYST_CONNECTED",
```
justinmk3, line 2009: setContext doesn't actually work in cloud9, but it looks like all of these cases are `!isCloud9` so it doesn't matter in this case.

src/codecatalyst/auth.ts
```diff
@@ -101,10 +112,6 @@ export class CodeCatalystAuthenticationProvider {
         }
     }
 
-    public async removeSavedConnection() {
-        await this.secondaryAuth.removeConnection()
```
justinmk3, line 105: thank you for lifting these special cases into a higher layer!

justinmk3 (approved):

andrewyuq (approved):

package.json
nkomonen-amazon: Better to have less uses of `isCloud9`, will remove. Good callout

PR #3928, opened 2023-10-20
fix(codewihsperer): mapping language code before calling sendTelemetry API

Will-ShaoHua:
## License

### DIFF
--- a/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
+++ b/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
@@ -132,13 +132,21 @@ export class CodeWhispererCodeCoverageTracker {
             codewhispererUserGroup: CodeWhispererUserGroupSettings.getUserGroup().toString(),
             codewhispererCustomizationArn: selectedCustomization.arn === '' ? undefined : selectedCustomization.arn,
         })
+
+        let codewhispererRuntimeLanguage: string = this._language
+        if (this._language === 'jsx') {
+            codewhispererRuntimeLanguage = 'javascript'
+        } else if (this._language === 'tsx') {
+            codewhispererRuntimeLanguage = 'typescript'
+        }
+
         client
             .sendTelemetryEvent({
                 telemetryEvent: {
                     codeCoverageEvent: {
                         customizationArn: selectedCustomization.arn === '' ? undefined : selectedCustomization.arn,
                         programmingLanguage: {
-                            languageName: this._language,
+                            languageName: codewhispererRuntimeLanguage,
                         },
                         acceptedCharacterCount: acceptedTokens,
                         totalCharacterCount: totalTokens,
--- a/src/codewhisperer/util/telemetryHelper.ts
+++ b/src/codewhisperer/util/telemetryHelper.ts
@@ -340,6 +340,13 @@ export class TelemetryHelper {
             e2eLatency = performance.now() - session.invokeSuggestionStartTime
         }
 
+        let codewhispererRuntimeLanguage: string = this.sessionDecisions[0].codewhispererLanguage
+        if (codewhispererRuntimeLanguage === 'jsx') {
+            codewhispererRuntimeLanguage = 'javascript'
+        } else if (codewhispererRuntimeLanguage === 'tsx') {
+            codewhispererRuntimeLanguage = 'typescript'
+        }
+
         client
             .sendTelemetryEvent({
                 telemetryEvent: {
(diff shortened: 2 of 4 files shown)


src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -340,6 +340,13 @@ export class TelemetryHelper {
             e2eLatency = performance.now() - session.invokeSuggestionStartTime
         }
 
+        let codewhispererRuntimeLanguage: string = this.sessionDecisions[0].codewhispererLanguage
```
zixlin7, line 343: I think we already have a method `mapToRuntimeLanguage` can directly use that?

andrewyuq: In telemetry we do want to report if it's tsx or jsx specific.

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -132,13 +132,21 @@ export class CodeWhispererCodeCoverageTracker {
...
             codewhispererCustomizationArn: selectedCustomization.arn === '' ? undefined : selectedCustomization.arn,
         })
+
+        let codewhispererRuntimeLanguage: string = this._language
+        if (this._language === 'jsx') {
+            codewhispererRuntimeLanguage = 'javascript'
+        } else if (this._language === 'tsx') {
+            codewhispererRuntimeLanguage = 'typescript'
+        }
+
```
andrewyuq, line 142: I think instead of changing here we may want to change the one that's something called "VSCToCodeWhispererRuntimeLanguage"

Will-ShaoHua: this one is for patch only. I am refactoring the whole language mapping stuff

src/codewhisperer/util/telemetryHelper.ts
zixlin7: I thought we are mapping it to typescript/javascript here?

Will-ShaoHua: for that function 
```
public mapToRuntimeLanguage<
        T extends codewhispererClient.ListRecommendationsRequest | codewhispererClient.GenerateRecommendationsRequest
    >(request: T): T
```


```
return {
            ...request,
            fileContext: {
                ...fileContext,
                programmingLanguage: parentLanguage,
            },
        }
```

andrewyuq: we only map for the request to the service but for telemetry we are stil reporting tsx if it's tsx.

andrewyuq: actually nvm, I thought we are changing for Kibana metrics but this is STE. So this is good then.

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
andrewyuq: Ok, when available please create such ticket and briefly describe what's going to be refactored.

andrewyuq (approved):

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
Will-ShaoHua: yea I will, I think I will follow similarly what we have in JB implementation

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -581,6 +588,14 @@ export class TelemetryHelper {
     }
     public sendCodeScanEvent(languageId: string, jobId: string) {
         getLogger().debug(`start sendCodeScanEvent: jobId: "${jobId}", languageId: "${languageId}"`)
+
+        let codewhispererRuntimeLanguage: string = languageId
+        if (codewhispererRuntimeLanguage === 'jsx') {
+            codewhispererRuntimeLanguage = 'javascript'
+        } else if (codewhispererRuntimeLanguage === 'tsx') {
+            codewhispererRuntimeLanguage = 'typescript'
+        }
```
justinmk3, line 597: Logic is repeated in 4 different places. Could be extracted to a function like toTelemetryLang

Will-ShaoHua: @justinmk3 yea, actually I am thinking to refactor the language mapping stuff with this PR https://github.com/aws/aws-toolkit-vscode/pull/3933/files

PR #3931, opened 2023-10-20
watchedfiles: add watch patterns once to avoid rebuilding n times

bryceitoc9:
## Problem
WatchedFiles takes a long time to process the watched files

## Solution
There's probably a lot of problems here, but this is definitely one of them: rebuilding after each additional watch pattern meant we were processing globs n times, with each run containing `{1, 2, 3, ... , n-1, n}` patterns. Swapping this to an array and enforcing adding the patterns once ensures we only process 1 time, for n globs on initial load. We aren't adding globs dynamically at any point; if we ever feel the need to, we can implement some fancy caching/expiry business.

FWIW it seems like getting the files in the glob is fairly fast, so if processing files is the bottleneck, this might be the cure? I'm sure we can improve more going forward regardless.

## License

### DIFF
--- a/.changes/next-release/Bug Fix-b3e30500-5dbe-4b08-ae36-aaba6b020c8f.json
+++ b/.changes/next-release/Bug Fix-b3e30500-5dbe-4b08-ae36-aaba6b020c8f.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Improved CloudFormation file watcher startup"
+}
--- a/.changes/next-release/Bug Fix-b3e30500-5dbe-4b08-ae36-aaba6b020c8f.json
+++ b/.changes/next-release/Bug Fix-b3e30500-5dbe-4b08-ae36-aaba6b020c8f.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Improved CloudFormation file watcher startup"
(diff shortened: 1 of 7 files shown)


justinmk3:
integ failure

```
  1) CloudFormation Template Registry
       fails if you set watch patterns multiple times:
     TypeError [ERR_AMBIGUOUS_ARGUMENT]: The "error/message" argument is ambiguous. The error message "CloudFormationTemplateRegistry: watch patterns have already been established" is identical to the message.
      at new NodeError (node:internal/errors:399:5)
      at expectsError (node:assert:817:15)
      at Function.rejects (node:assert:922:3)
      at async Context.<anonymous> (src/testInteg/cloudformation/templateRegistry.test.ts:104:9)
```

.changes/next-release/Bug Fix-b3e30500-5dbe-4b08-ae36-aaba6b020c8f.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Improved CloudFormation file watcher startup"
```
justinmk3, line 3:
```suggestion
	"description": "Improved performance of CloudFormation file watcher startup"
```

justinmk3 (approved):

justinmk3:
Followup: https://github.com/aws/aws-toolkit-vscode/pull/3962

PR #3932, opened 2023-10-21
codewhisperer: add quick fix code actions for issues

ctlai95:
## Problem

Currently the Apply Fix action is hidden behind the hover panel and requires user to mouse over the issue. It should be more accessible without having to leave the keyboard.

## Solution

- Added code action provider to provide a Quick Fix action for each issue (that has at least 1 security patch available).
- The menu can be accessed using the  ⌘ + . keyboard shortcut or by clicking the light bulb icon
- Added an abstract class `SecurityIssueProvider` which has text document change handler common for `SecurityIssueHoverProvider` and `SecurityIssueCodeActionProvider`

image

## License

### DIFF
--- a/src/codewhisperer/service/securityIssueCodeActionProvider.ts
+++ b/src/codewhisperer/service/securityIssueCodeActionProvider.ts
@@ -1,0 +1,44 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+import { SecurityIssueProvider } from './securityIssueProvider'
+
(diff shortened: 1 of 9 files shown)


src/codewhisperer/service/securityIssueCodeActionProvider.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    public provideCodeActions(
+        document: vscode.TextDocument,
+        range: vscode.Range | vscode.Selection,
+        _context: vscode.CodeActionContext,
+        _token: vscode.CancellationToken
+    ): vscode.CodeAction[] {
+        const codeActions: vscode.CodeAction[] = []
+
+        for (const group of this.issues) {
```
justinmk3, line 24: strange interface. why isn't it something like `this.groups` ?

justinmk3 (approved):

mj632 (approved):

PR #3938, opened 2023-10-23
filewatcher: template watcher lazy-loads if enableCodeLenses is off

bryceitoc9:
## Problem
The template file watcher always ran regardless of SAM hint setting

## Solution
The template file watcher now loads on demand. This appears to work even when toggling from off to on (codelenses appear on the template file)
* ~~This may hit the startup function multiple times? Will do some more testing~~ should be resolved with https://github.com/aws/aws-toolkit-vscode/pull/3938/commits/0b1df9d39df91638e749d4825735f548e45ece28
* Confirmed SAM debug works, when coming from startup with codelenses disabled
* Confirmed the registry builds when codelenses are enabled

## License

### DIFF
--- a/src/shared/cloudformation/activation.ts
+++ b/src/shared/cloudformation/activation.ts
@@ -5,7 +5,7 @@
 
 import * as vscode from 'vscode'
 import { getLogger } from '../logger'
-import { localize } from '../utilities/vsCodeUtils'
+import { isToolkitActive, localize } from '../utilities/vsCodeUtils'
 
 import { AsyncCloudFormationTemplateRegistry, CloudFormationTemplateRegistry } from '../fs/templateRegistry'
 import { getIdeProperties } from '../extensionUtilities'
@@ -13,6 +13,7 @@ import { NoopWatcher } from '../fs/watchedFiles'
 import { createStarterTemplateFile } from './cloudformation'
 import { Commands } from '../vscode/commands2'
 import globals from '../extensionGlobals'
+import { SamCliSettings } from '../sam/cli/samCliSettings'
 
 export const templateFileGlobPattern = '**/*.{yaml,yml}'
 
@@ -48,7 +49,7 @@ export async function activate(extensionContext: vscode.ExtensionContext): Promi
         getLogger().error('Failed to activate template registry: %s', e)
         // This prevents us from breaking for any reason later if it fails to load. Since
         // Noop watcher is always empty, we will get back empty arrays with no issues.
-        globals.templateRegistry = new NoopWatcher() as unknown as CloudFormationTemplateRegistry
+        globals.templateRegistry = (async () => new NoopWatcher() as unknown as CloudFormationTemplateRegistry)()
     }
     // If setting it up worked, add it to subscriptions so it is cleaned up at exit
     extensionContext.subscriptions.push(
@@ -71,6 +72,8 @@ function setTemplateRegistryInGlobals(registry: CloudFormationTemplateRegistry) 
         await registry.addExcludedPattern(templateFileExcludePattern)
         await registry.addWatchPatterns([templateFileGlobPattern])
         await registry.watchUntitledFiles()
+
+        return registry
     }
 
     const asyncRegistry = new AsyncCloudFormationTemplateRegistry(registry, registrySetupFunc)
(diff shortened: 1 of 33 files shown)


src/shared/fs/templateRegistry.ts
```diff
@@ -108,31 +102,37 @@ export class AsyncCloudFormationTemplateRegistry {
...
-                    })
-                    getLogger().debug('cfn: getInstance() requested, still initializing')
-                    while (!this.isSetup) {
-                        await sleep(2000)
+        const config = SamCliSettings.instance
+        if (config.get('enableCodeLenses', false)) {
+            this.asyncSetupFunc(this.instance).then(() => {
+                this.isSetup = true
+                return this.instance
+            })
```
justinmk3, line 110: Isn't the PerfLog still useful to have? It's not only for startup performance, it's for measuring any kind of performance.

src/shared/fs/templateRegistry.ts
```diff
@@ -108,31 +102,37 @@ export class AsyncCloudFormationTemplateRegistry {
...
-                async (progress, token) => {
-                    token.onCancellationRequested(() => {
-                        // Allows for new message to be created if templateRegistry variable attempted to be used again
-                        this.setupProgressMessage = undefined
-                    })
-                    getLogger().debug('cfn: getInstance() requested, still initializing')
-                    while (!this.isSetup) {
-                        await sleep(2000)
+        const config = SamCliSettings.instance
+        if (config.get('enableCodeLenses', false)) {
```
justinmk3, line 106: does sam debugging work with `enableCodeLenses=false` ?

sam debugging depends on the registry to resolve stuff: https://github.com/aws/aws-toolkit-vscode/blob/1db7975286b61ed462e782d7dfc2573a53c615df/src/shared/sam/debugger/awsSamDebugger.ts#L257-L259

bryceitoc9: Will take a look. If we're getting the instance, it should (it'll build the registry on demand).

src/shared/fs/templateRegistry.ts
bryceitoc9: I'll see if I can worm it in there. It's a little trickier now that this builds on demand and not on the single constructor call.

src/shared/fs/templateRegistry.ts
bryceitoc9: Looks like a no. Will take a look.

justinmk3: At worst, perhaps remove the `config.get('enableCodeLenses')` check for now. Then we need see what's calling `CloudFormationTemplateRegistry.instance` and try to minimize those calls.

bryceitoc9: This triggered a larger, but necessary change: essentially, `globals.templateRegistry` needed to be async to ensure the registry could build.

src/shared/cloudformation/activation.ts
```diff
@@ -79,15 +82,19 @@ function setTemplateRegistryInGlobals(registry: CloudFormationTemplateRegistry)
...
             // already set a mock object before activation.
             // Though in prod nothing should be calling this 'set' function.
             if (this.cfnInstance) {
                 return this.cfnInstance
             }
 
-            return asyncRegistry.getInstance()
+            // prevent eager load if codelenses are off
+            const config = SamCliSettings.instance
+            if (config.get('enableCodeLenses', false) || isToolkitActive()) {
```
justinmk3, line 95: Ideally we would fix the callers that are actually calling this during startup. But this is a good compromise for now.

justinmk3 (approved):
These changes will be helpful, since allowing the global to be async is probably useful. 

But,
- something is still triggering `getInstance()`, even with `aws.samcli.enableCodeLenses = false`. 
- I don't see the "Scanning..." message anymore, even though the logs show that it's scanning.

bryceitoc9:
Going to take some time to figure out why the tests are busted.

> something is still triggering getInstance(), even with aws.samcli.enableCodeLenses = false
On load? This might be the `awsFiletypes` activation. The whole key is that `getInstance` can get hit, but not do anything until it gets hit after activation is finished (so someone booting VS Code and mashing F5 before the activation can load might hit issues)

> I don't see the "Scanning..." message anymore, even though the logs show that it's scanning.
Optimistic me says that this is because it's going so fast that it's missing it (now that it's not necessarily running at the same time as the other filewatcher). I have seen it pop up on my side with the AWS SAM repo.

justinmk3 (approved):

Issue #3930, opened 2023-10-20
Extension always opens OUTPUT panel at VS Code start


rafrafek:
## Problem

OUTPUT panel always opens at VS Code launch

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/aws/aws-toolkit-vscode/assets/23004737/4dc20bff-a1bb-43f7-99f9-10e0b61015e9">
  <source media="(prefers-color-scheme: light)" srcset="https://github.com/aws/aws-toolkit-vscode/assets/23004737/43e4e9c3-97cd-4aee-a37c-3e4a3c78838d">
  <img width="1390" alt="image" src="https://github.com/aws/aws-toolkit-vscode/assets/23004737/4dc20bff-a1bb-43f7-99f9-10e0b61015e9">
</picture>

## Steps to reproduce the issue

1. Install the extension
2. Close all panels and close VS Code
3. Open VS Code and see the OUTPUT panel

Disabling this extension fixes the issue.

## Expected behavior

No OUTPUT panel at VS Code launch

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/aws/aws-toolkit-vscode/assets/23004737/f0424f39-2bdc-4716-b802-fd8072317dc8">
  <source media="(prefers-color-scheme: light)" srcset="https://github.com/aws/aws-toolkit-vscode/assets/23004737/54028813-6cc6-4c43-b8cb-b015e1dcf989">
  <img width="1346" alt="image" src="https://github.com/aws/aws-toolkit-vscode/assets/23004737/f0424f39-2bdc-4716-b802-fd8072317dc8">
</picture>

## System details (run the `AWS: About Toolkit` command)

-   OS: MacOS Sonoma 14.0
-   Visual Studio Code version: 1.83.1 Darwin arm64
-   AWS Toolkit version: v1.95.0

`AWS: About Toolkit` command output:
OS: Darwin arm64 23.0.0
Visual Studio Code extension host:  1.83.1
AWS Toolkit:  1.95.0
node: 18.15.0
electron: 25.8.4

rafrafek:
Fun fact: my screenshots are both light- and dark- theme compatible 😊

object-Object:
This is the code at fault (added in 029f1d6d8644ae4697e4522a21492f815c383159):

[...]

PR #3939, opened 2023-10-23
fix(redshift): VS Code always opens Output panel

yi-jin-0:
## Problem
Output panel always opens at VS Code launch. Issue reported at https://github.com/aws/aws-toolkit-vscode/issues/3930. 
Screenshot 2023-10-23 at 11 04 57 AM

## Solution
Instead of showing Output panel at Redshift activation, change to show Output panel only when a message is added.
Screenshot 2023-10-23 at 11 06 40 AM
Screenshot 2023-10-23 at 11 08 37 AM

## License

### DIFF
--- a/.changes/next-release/Bug Fix-fff4dbbd-2664-43c3-bf6f-98cf1440ace2.json
+++ b/.changes/next-release/Bug Fix-fff4dbbd-2664-43c3-bf6f-98cf1440ace2.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Redshift: Fix the issue that Output panel always opens at VS Code launch."
+	"description": "Output panel always opens at VS Code launch."
 }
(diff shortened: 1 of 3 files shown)


.changes/next-release/Bug Fix-fff4dbbd-2664-43c3-bf6f-98cf1440ace2.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Redshift: Fix the issue that Output panel always opens at VS Code launch."
```
justinmk3, line 3: - The item type is "Bug Fix" so the description just needs to describe what's being fixed.
- Prefixing Redshift changes with `Redshift:` is usually a good idea, but in this case it's a Toolkit-wide behavior that's being fixed, so we can leave out the prefix.


```suggestion
	"description": "Output panel always opens at VS Code launch."
```

yi-jin-0: Yes, I have made the suggested change.

justinmk3 (approved):

PR #3940, opened 2023-10-23
fix(codewhisperer): refactor runtimeLanguageContext.ts (languageMapping)

Will-ShaoHua:
## Problem
Too many different lands of "Language Identifier" (languageId), and it's quite confusing to handle mapping back and forward between different set of language id. For CodeWhisperer, we have the following 3 types of language identifier
1. CodewhispererLanguage (Telemetry / Plugin)
2. PlatformLanguage  which is defined by the client platform e.g. C9, VSCode
3. Codewhisperer runtime language (Service)
  * In most of the cases, CodewhispererLanguageId equals to Codewhisperer runtime language except the following
  * TSX which will be mapped to regular Typescrript at CodeWhisperer runtime
  * JSX which will be mapped to regular Javascript at CodeWhisperer runtime

## Solution
1. rename type `SupportedLanguage` to be `PlatformLanguageId` to avoid confusion
2. refactor runtimeLanguageContext.ts 
  *  Simplify interface `toTelemetryLanguage` to consume a string "Language ID" which could be either `CodeWhispererLanguageId` or `PlatforLanguageId`, the component abstracts the mapping logics and will *ALWAYS* return a CodeWhispererLanguage if any, otherwise undefined
  * Add `toRuntimeLanguage` API which is the *ONLY* one function which takes `CodewhispererLanguage` as the parameter given the fact that we will only invoke CW service after we handle WELL the mapping between `CodewhispererLanguage`and `PlatformLanguageId`

#### Java

java

#### Python
python

#### C
c

#### CPP
cpp

#### Go
go

#### Javascript
js

#### Jsx
jsx

#### K
kt
otlin

#### Plaintext
text

#### PHP
php

#### Ruby
ruby

#### Rust
rust

#### Scala
scala

#### Shellscript
shell

#### Tsx
tsx

#### Typescript
ts

#### Java

https://github.com/aws/aws-toolkit-vscode/assets/96078566/7f47fc67-3a68-4f09-bcf7-60a3a2821da9

#### JSX

https://github.com/aws/aws-toolkit-vscode/assets/96078566/eb8ea00b-bbb5-4a39-b51d-7d20f8d3cd89

#### Python

https://github.com/aws/aws-toolkit-vscode/assets/96078566/ad9381c5-6928-4deb-9bcd-44fd5a0c08ea

#### Ruby

https://github.com/aws/aws-toolkit-vscode/assets/96078566/506dd31c-829d-4066-ae8a-5322740fe052

#### Shell

https://github.com/aws/aws-toolkit-vscode/assets/96078566/6eba1376-75f2-47ae-8b45-d131dba0d38e

#### TS

https://github.com/aws/aws-toolkit-vscode/assets/96078566/fef15ba6-794b-4402-9fe8-93d3ee63c4ce

#### CPP

https://github.com/aws/aws-toolkit-vscode/assets/96078566/94308610-f00b-44e7-aa8e-708a5a282c68

#### C

https://github.com/aws/aws-toolkit-vscode/assets/96078566/6b1706cd-d0a9-4a1b-9d5d-156acd4b7381

#### CSharp

https://github.com/aws/aws-toolkit-vscode/assets/96078566/24e63934-ff4f-45d3-a887-2aa6bf75c244

#### Plaintext

https://github.com/aws/aws-toolkit-vscode/assets/96078566/32d8dee9-124c-4e21-a3cf-788a40f7b601

#### Rust

https://github.com/aws/aws-toolkit-vscode/assets/96078566/4aaeaedc-0908-4f65-b29d-f5699d64816a

#### Scala

https://github.com/aws/aws-toolkit-vscode/assets/96078566/059cb8ab-2016-4428-9a86-1c0d9868233a

#### SQL

https://github.com/aws/aws-toolkit-vscode/assets/96078566/5fb73961-a45b-4591-8f64-315e200057db

#### Go

Uploading golang.mov…

## License


src/codewhisperer/tracker/codewhispererTracker.ts
```diff
@@ -95,13 +95,6 @@ export class CodeWhispererTracker {
         } catch (e) {
             getLogger().verbose(`Exception Thrown from CodeWhispererTracker: ${e}`)
         } finally {
-            let codewhispererRuntimeLanguage: string = suggestion.language
-            if (codewhispererRuntimeLanguage === 'jsx') {
-                codewhispererRuntimeLanguage = 'javascript'
-            } else if (codewhispererRuntimeLanguage === 'tsx') {
-                codewhispererRuntimeLanguage = 'typescript'
-            }
```
Will-ShaoHua, line 103: removed it since userModification is not sent via STE API now

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -14,41 +14,43 @@ export class RuntimeLanguageContext {
...
+                go: 'go',
+                kotlin: 'kotlin',
+                php: 'php',
+                ruby: 'ruby',
+                rust: 'rust',
+                scala: 'scala',
+                sh: 'shell',
+                shellscript: 'shell',
+                sql: 'sql',
+            }
```
Will-ShaoHua, line 51: formatter changes

justinmk3 (approved):

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -47,59 +68,78 @@ export class RuntimeLanguageContext {
...
             kotlin: 'kt',
             php: 'php',
             ruby: 'rb',
             rust: 'rs',
             scala: 'scala',
-            sh: 'sh',
-            shellscript: 'sh',
+            shell: 'sh',
             sql: 'sql',
+            plaintext: 'txt',
```
justinmk3, line 94: nit: alphanumeric order

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -47,59 +68,78 @@ export class RuntimeLanguageContext {
...
-     * @param vscLanguageId : official vscode languageId
+     * To add a new platform language id:
+     * 1. add new platform language ID constant in the file codewhisperer/constant.ts
+     * 2. add corresponding CodeWhispererLanguageId mapping in the constructor of RuntimeLanguageContext
+     * @param languageId : arbitrary string denoting a specific programming language, e.g. CodewhispererLanguageId or PlatformLanguageId
      * @returns corresponding CodewhispererLanguage ID if any, otherwise undefined
      */
-    public mapVscLanguageToCodeWhispererLanguage(vscLanguageId?: string): CodewhispererLanguage | undefined {
-        return this.supportedLanguageMap.get(vscLanguageId) ?? undefined
+    public mapToCodewhispererLanguage(languageId?: string): CodewhispererLanguage | undefined {
```
justinmk3, line 124: is this only for telemetry, or is telemetry one of many uses for this function? 

the names `mapToCodeWhispererRuntimeLanguage` and `mapToCodewhispererLanguage` have no meaningful distinction. 

if this is only for telemetry consider naming it `toTelemetryLanguage`
```suggestion
    public toTelemetryLanguage(languageId?: string): CodewhispererLanguage | undefined {
```

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -47,59 +68,78 @@ export class RuntimeLanguageContext {
...
+    }
 
-        const values = Array.from(this.supportedLanguageMap.values())
-        const keys = Array.from(this.supportedLanguageMap.keys())
-        values.forEach(item => this.supportedLanguageSet.add(item))
-        keys.forEach(item => this.supportedLanguageSet.add(item))
+    /**
+     * Only used when invoking CodeWhisperer service API, for telemetry usage please use mapToCodewhispererLanguage
+     */
+    public mapToCodeWhispererRuntimeLanguage(language: CodewhispererLanguage): CodewhispererLanguage {
```
justinmk3, line 101: why does it accept a `CodewhispererLanguage` and also return a `CodewhispererLanguage`? seems strange.

- this seems like it's doing a "normalize" operation, i.e. it's cleaning up values rather than mapping them
- this is in a CW module, in a CW class, so why repeat "codewhisperer" in the function name?

```suggestion
    public toRuntimeLanguage(language: CodewhispererLanguage): CodewhispererLanguage {
```

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -47,59 +68,78 @@ export class RuntimeLanguageContext {
...
-    public mapVscLanguageToCodeWhispererLanguage(vscLanguageId?: string): CodewhispererLanguage | undefined {
-        return this.supportedLanguageMap.get(vscLanguageId) ?? undefined
+    public mapToCodewhispererLanguage(languageId?: string): CodewhispererLanguage | undefined {
+        return this.supportedLanguageMap.get(languageId)
     }
 
     /**
      * This is for notebook files map to a new filename with the corresponding language extension
-     * @param vscLanguageId : official vscode languageId
+     * @param languageId : arbitrary string denoting a specific programming language, e.g. CodewhispererLanguageId or PlatformLanguageId
```
justinmk3, line 130: it's not really arbitrary so the new docstring adds confusion. is this correct:
```suggestion
     * @param languageId : vscode language id or codewhisperer language name
```

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -112,25 +152,17 @@ export class RuntimeLanguageContext {
...
-            case 'tsx':
-                parentLanguage = { languageName: CodeWhispererConstants.typescript }
-                break
-            case 'jsx':
-                parentLanguage = { languageName: CodeWhispererConstants.javascript }
-                break
-            default:
-                parentLanguage = childLanguage
-                break
+        const runtimeLanguage: codewhispererClient.ProgrammingLanguage = {
```
justinmk3, line 155: This is nuts. How can there be so many names for the same thing in the same subystem?

- "programming language"
- "runtime language"

Use as few names as possible.

Will-ShaoHua: yea agree, it's really crazy to have these different names in cw module.

codewhisperer language is a super set of runtime language(i.e. normalized language)
JSX / TSX will be normalized to JS/TS

ProgammingLanguage is the SDK POJO of the format { languageName: string }

does this more sense to have this?
`type NormalizedLanguageId = Exclude<CodewhispererLanguage, 'jsx' | 'tsx'>`
https://github.com/aws/aws-toolkit-vscode/pull/3940/files#diff-0509671083ef22655748504f51cc4ad12de62ff0ec7a4f317f57d6bf24652838R12

Will-ShaoHua: the reason why I still keep the following 3 types existing in the module
1. type CodewhispererLanguage (which is telemetry.gen type)
2. type PlatformLanguageId
3. NormalizedLanguageId

is to enforce the compiler to do the string check at compile time

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -47,59 +70,78 @@ export class RuntimeLanguageContext {
...
-            sh: 'sh',
-            shellscript: 'sh',
+            shell: 'sh',
             sql: 'sql',
+            plaintext: 'txt',
         })
+    }
+
+    /**
+     * Only used when invoking CodeWhisperer service API, for telemetry usage please use mapToCodewhispererLanguage
```
justinmk3, line 101:
```suggestion
     * Only used when invoking CodeWhisperer service API, for telemetry usage please use toTelemetryLanguage
```

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -47,59 +70,78 @@ export class RuntimeLanguageContext {
...
+     */
+    public ToRuntimeLanguage(language: CodewhispererLanguage): NormalizedLanguageId {
```
justinmk3, line 103:
```suggestion
    public toRuntimeLanguage(language: CodewhispererLanguage): NormalizedLanguageId {
```

Will-ShaoHua: oops..

justinmk3 (approved):

Will-ShaoHua:
fixing compile error

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -3,103 +3,147 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { getLogger } from '../../shared/logger/logger'
 import { CodewhispererLanguage } from '../../shared/telemetry/telemetry.gen'
```
zixlin7, line 7: This `CodewhispererLanguage` is actually telemetry language? can we rename it to be that, otherwise it's still confusing

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -3,103 +3,147 @@
...
-        this.supportedLanguageMap = createConstantMap<CodeWhispererConstants.SupportedLanguage, CodewhispererLanguage>({
-            java: 'java',
-            python: 'python',
-            javascriptreact: 'jsx',
-            javascript: 'javascript',
-            typescript: 'typescript',
-            typescriptreact: 'tsx',
-            csharp: 'csharp',
+        this.supportedLanguageMap = createConstantMap<
+            CodeWhispererConstants.PlatformLanguageId | CodewhispererLanguage,
```
zixlin7, line 51: maybe I'm missing something, is this map for mapping platformId to telemetry id? why do we need `CodewhispererLanguage` in the input type?

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -112,25 +156,17 @@ export class RuntimeLanguageContext {
...
-                break
-            case 'jsx':
-                parentLanguage = { languageName: CodeWhispererConstants.javascript }
-                break
-            default:
-                parentLanguage = childLanguage
-                break
+        const runtimeLanguage: codewhispererClient.ProgrammingLanguage = {
+            languageName: this.toRuntimeLanguage(
+                request.fileContext.programmingLanguage.languageName as CodewhispererLanguage
```
zixlin7, line 161: I think it's confusing that this method is called to mapToRuntimeLanguage when it's in fact mapping to telemetry language, should we rename? and if we are using `normalizedLanguage` as the type for sending to service, let's keep it consistent, and toRuntimeLanguage is essentially toNormalizedLanguage?

src/codewhisperer/util/runtimeLanguageContext.ts
Will-ShaoHua: yes, just need to update the name in common repo, but not sure if rename will cause any concern

src/codewhisperer/util/runtimeLanguageContext.ts
Will-ShaoHua: purpose of this is to make this map, language type agnostic, at least against PlatformLanguage & CodewhispererLanguage. Callers are safe to call as many times as they want without side effect.

Will-ShaoHua: the status quo  is that we are using both PlatformLanguageId and CodewhispererLanguage anywhere without consistency. Take `CodeCoverageTracker.getTracker` for example, https://github.com/aws/aws-toolkit-vscode/blob/master/src/codewhisperer/service/recommendationHandler.ts#L302 is using `session.language` which has type CodewhispererLanguage whereas in activation.ts https://github.com/aws/aws-toolkit-vscode/blob/master/src/codewhisperer/activation.ts#L310, `e.document.languageId` is being used

Will-ShaoHua: Making this map to be `PlatformLanguageId | CodewhispererLanguage` type agnostic is one way to reduce the surface of changes we have to make for this purpose

Will-ShaoHua: The ultimate goal is still to use an unified language interface

src/codewhisperer/util/runtimeLanguageContext.ts
Will-ShaoHua: yes I am open to that, we can name it `normalizeLanguage()`

Will-ShaoHua: > it's in fact mapping to telemetry language
that's why here the return type is `NormalizedLanguage`

Will-ShaoHua: still the best solution and long-term solution is to introduce another language interface which should be used universally in our VSC plugin code base instead of relying on telemetry.gen type. However with that approach, the surface of changes might be huge and it would be better off refactoring component by component

src/codewhisperer/util/runtimeLanguageContext.ts
zixlin7: We can alias the import like `import { CodewhispererLanguage as CodewhispererTelemetryLanguage } from ...` but I don't know how many places we are importing this type so just an idea

src/codewhisperer/util/runtimeLanguageContext.ts
zixlin7: I see it makes sense now thanks

src/codewhisperer/util/runtimeLanguageContext.ts
zixlin7: I see now it is mapping to NormalizedLanguage not the telemetry one I mis-read.If we can just use one of `runtimeLanguage` or `normalizedLanguage` to reduce one additional name that would be great

zixlin7 (approved):

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -3,103 +3,147 @@
...
+            jsx: 'jsx',
             kotlin: 'kt',
+            plaintext: 'txt',
             php: 'php',
+            python: 'py',
             ruby: 'rb',
             rust: 'rs',
             scala: 'scala',
-            sh: 'sh',
-            shellscript: 'sh',
```
andrewyuq, line 69: is there a reason to remove `sh: 'sh'`? IIRC is this added to support a case in C9? Same applies to the above `c_cpp: 'cpp'`

src/codewhisperer/util/runtimeLanguageContext.ts
andrewyuq: For the keys here, do they also need to include C9 platform languageId?

andrewyuq (commented):
How can we test this for all languages for VSCode and C9?

src/codewhisperer/util/runtimeLanguageContext.ts
Will-ShaoHua: here, normalize func is being called https://github.com/aws/aws-toolkit-vscode/pull/3940/files#diff-0509671083ef22655748504f51cc4ad12de62ff0ec7a4f317f57d6bf24652838R138

Will-ShaoHua: removing from this map is to avoid ppl have to manually add so many different [languageId, extension] pair in this map. So I move the whole thing to `this.supportedLanguageMap` at L50.

Will-ShaoHua: also here `isLanguageSupported()` https://github.com/aws/aws-toolkit-vscode/pull/3940/files#diff-0509671083ef22655748504f51cc4ad12de62ff0ec7a4f317f57d6bf24652838R179, will normalize languageId first then check

Will-ShaoHua: basically every methods in this class will normalize the provided language id first then do its work

Will-ShaoHua: yes a functional test would be nice, I will add this and test it on C9

andrewyuq: ah I see.

andrewyuq: How can we prevent people from not using `normalize` first in the future?

Will-ShaoHua: basically "normalized" language is of type `CodewhispererLanguage` so we can change all functions which take language as one of their argument to be of type `CodewhispererLanguage`

Will-ShaoHua: so that we can have compile time check

Will-ShaoHua: which I think it apply to most of our functions but of course something like `getTracker()` is not this case I can update this in this PR as well

src/test/codewhisperer/util/runtimeLanguageContext.test.ts
```diff
@@ -89,39 +91,166 @@ describe('runtimeLanguageContext', function () {
...
+            it(`should return itself if input language is codewhispererLanguageId - ${inputCwsprLanguageId}`, function () {
+                const actual = languageContext.normalizeLanguage(inputCwsprLanguageId)
+                assert.strictEqual(actual, inputCwsprLanguageId)
+            })
+        }
+
+        const platformLanguageIds: [PlatformLanguageId, CodewhispererLanguage][] = [
+            ['cpp', 'cpp'],
+            ['c_cpp', 'cpp'],
+            ['cpp', 'cpp'],
```
justinmk3, line 130: this list seems redundant. is it expected that devs must update this list in the future?

andrewyuq (approved):

Issue #2614, opened 2022-05-03
Cannot find template file when debugging in vs code


ABlackPenny:
**Describe the bug**
I was trying to debug AWS LAMBDA locally in vs code, but I always get the 'Invalid (or missing) template file (path must be workspace-relative, or absolute):' when I run the debug

**To reproduce**

1.Add new file named as template.yml, copy and paste from serverless.yml to template.yml 
 2. Click on 'AWS: Add debug configuration'
 3. open the launch json, add the payload 
 4. See error : Invalid (or missing) template file (path must be workspace-relative, or absolute): e:\companyNameWork\XXX-api\template.yml

**Expected behavior**
The debug function would run and has some output in debug console just like other project's done before.

**Screenshots**
![417adf0871ca9df2da5d085a5dbefda](https://user-images.githubusercontent.com/53291586/166409014-c5436697-23ec-4f75-a495-7cd537e1ed9f.png)

![faa738c770de91d0a46ad5bad2850a0](https://user-images.githubusercontent.com/53291586/166409131-2080b646-0416-44b7-b02b-d93b7138f544.png)

**Your Environment**
- OS: win10 
- AWS Toolkit version: AWS Toolkit v1.37.0
- SAM CLI version: SAM CLI, version 1.40.0

**Additional context**
 1. In same conputer, with same process and IDE(vs code), other project could use AWS Toolkit debug perfectly.
2. I have tried absolute path and relative path, doesn't work. 3.Tried rebuild the template.yml 4. Tried reopen the IDE 5. Tried reopen the computer 6. The API has been deployed and could work

ABlackPenny:
> This means that somehow our file watcher isn't finding the file. Can you check the logs (remove private info)?
> 
> 1. In VSCode settings, ensure that "aws log level" is "verbose".
> 2. Perform the steps to reproduce the issue.
> 3. Use the `AWS: View Toolkit Logs` command to get the logs.
> 
> Is the file path mentioned in the logs?

Hi justinmk3, Thanks for your reply and help.
1. I chenged the log level to 'verbose' 
2. I deleted the current '.vscode' and template.yml 
3. I redo the same process like I did before. 

This is what I got in log. And I think the file path mentioned in the log. 
![微信图片编辑_20220504110134](https://user-images.githubusercontent.com/53291586/166610621-ed77b125-9e1d-4a99-be0f-6b1aa15ff96b.jpg)

justinmk3:
related: #2613

[...]

PR #3946, opened 2023-10-24
fix(sam): "Invalid (or missing) template file"

justinmk3:
# Problem:
SAM debug may fail to find template.yaml even if it is explicitly given by `invokeTarget.templatePath` in the vscode launch config. #2614

# Solution:
- Always call `addItem()` in `validate()`, instead of depending on the  workspace-wide "scan".
- Log a message when `process()` fails to process the file contents.

## License

### DIFF
--- a/src/shared/fs/watchedFiles.ts
+++ b/src/shared/fs/watchedFiles.ts
@@ -172,14 +172,21 @@ export abstract class WatchedFiles<T> implements vscode.Disposable {
     }
 
     /**
-     * Adds an item to registry. Wipes any existing item in its place with new copy of the data
-     * @param uri vscode.Uri containing the item to load in
+     * Adds or updates an item in the registry, and returns the result.
+     *
+     * If the item matches an "exclude" rule, it is not added nor does it update/replace any existing item.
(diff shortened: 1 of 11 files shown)


src/shared/fs/watchedFiles.ts
```diff
@@ -172,31 +172,44 @@ export abstract class WatchedFiles<T> implements vscode.Disposable {
...
-     * @param uri vscode.Uri containing the item to load in
+     * Adds or updates an item in the registry, and returns the result.
+     *
+     * If the item matches an "exclude" rule, it is not added nor does it update/replace any existing item.
+     *
+     * @param uri vscode.Uri containing the item to register.
+     * @param quiet On failure, log a message instead of throwing an exception.
+     * @param contents Optional data to associate with the item, for logical (non-filesystem) URIs.
+     *
+     * @returns Item, or undefined if the name matches an "exclude" rule.
```
bryceitoc9, line 183: Nit: this also returns `undefined` on processing failure

bryceitoc9 (approved):

PR #3947, opened 2023-10-25
feat(codewhisperer): select customization by prefix

wlovo:
## Problem
The `aws.codeWhisperer.connect` command was introduced to allow enterprises to establish a default customization for onboarding developers after establishing an IdC connection.

However, it only selected with a given ARN. Given the current behavior for adding updated customization for CodeWhisperer, an enterprise would need to keep the ARN up-to-date as well when calling this command.

## Solution
Provide the ability to define the customization to select with a prefix for the customization name. This way the logic to select a customization as updated customizations are introduced can be simplified on the administrator side by maintaining a prefix. e.g. MyModel-v1, MyModel-v2, etc.

## License

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -8,11 +8,12 @@ Details about any publicly accessible functionalities exposed through [extension
 
 #### `aws.codeWhisperer.connect`
 
-**Signature**: _async (startUrl?: string, region?: string, customizationArn?: string, customizationName?: string, customizationDescription?: string) => Promise<void>_
+**Signature**: _async (startUrl?: string, region?: string, customizationArn?: string, customizationNamePrefix?: string) => Promise<void>_
 
 Shortcut command to directly connect to Identity Center or prompt start URL entry, as well as set a customization for CodeWhisperer requests.
 
-This command supports two sets of arguments:
+This command supports the following arguments:
 
 -   startUrl and region. If both arguments are provided they will be used, otherwise the command prompts for them interactively.
--   customization{Arn, Name, Description}. If at least customizationArn is provided, the command selects this customization.
+-   customizationArn: select customization by ARN. If provided, `customizationNamePrefix` is ignored.
+-   customizationNamePrefix: select customization by prefix, if `customizationArn` is `undefined`.
--- a/src/codewhisperer/commands/basicCommands.ts
+++ b/src/codewhisperer/commands/basicCommands.ts
@@ -18,6 +18,7 @@ import { ReferenceLogViewProvider } from '../service/referenceLogViewProvider'
 import { AuthUtil } from '../util/authUtil'
 import { isCloud9 } from '../../shared/extensionUtilities'
 import { InlineCompletionService } from '../service/inlineCompletionService'
+import { getLogger } from '../../shared/logger'
 import { openUrl } from '../../shared/utilities/vsCodeUtils'
 import {
     getPersistedCustomizations,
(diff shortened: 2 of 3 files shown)


src/codewhisperer/commands/basicCommands.ts
```diff
@@ -120,37 +121,43 @@ export const showSsoSignIn = Commands.declare('aws.codeWhisperer.sso', () => asy
...
+            match = getPersistedCustomizations().find(c => c.arn == customizationArn)
+        }
+        // Otherwise if only a prefix is given, find an entry that matches it.
+        else if (customizationNamePrefix) {
+            match = getPersistedCustomizations().find(c => c.name?.startsWith(customizationNamePrefix))
+        }
+
+        // If no match is found, nothing to do :)
+        if (!match) {
+            return
```
justinmk3, line 156: should this throw a ToolkitError ? or at least log something.

docs/api.md
```diff
@@ -8,11 +8,11 @@ Details about any publicly accessible functionalities exposed through [extension
...
-**Signature**: _async (startUrl?: string, region?: string, customizationArn?: string, customizationName?: string, customizationDescription?: string) => Promise<void>_
+**Signature**: _async (startUrl?: string, region?: string, customizationArn?: string, customizationNamePrefix?: string) => Promise<void>_
 
 Shortcut command to directly connect to Identity Center or prompt start URL entry, as well as set a customization for CodeWhisperer requests.
 
 This command supports two sets of arguments:
 
 -   startUrl and region. If both arguments are provided they will be used, otherwise the command prompts for them interactively.
--   customization{Arn, Name, Description}. If at least customizationArn is provided, the command selects this customization.
+-   customization{Arn, Name}. If at least customizationArn is provided, the command selects this customization.
```
justinmk3, line 18:
```suggestion
-   customizationArn: select customization by ARN. If provided, `customizationNamePrefix` is ignored.
-   customizationNamePrefix: select customization by prefix, if `customizationArn` is `undefined`.
```

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -120,37 +121,43 @@ export const showSsoSignIn = Commands.declare('aws.codeWhisperer.sso', () => asy
...
+        } else {
+            await getStartUrl()
+        }
+
+        // No customization match information given, exit early.
+        if (!customizationArn && !customizationNamePrefix) {
+            return
         }
+
+        let match: Customization | undefined
```
justinmk3, line 143: this can be const, because of the condition on line 139.

```suggestion
        const match = customizationArn ? ... : ...
```

src/codewhisperer/commands/basicCommands.ts
wlovo: I think a log might be more appropriate here.
We might not want to throw since it's a request from outside of Toolkit, it can easily pass a value that isn't expected.

docs/api.md
wlovo: Good suggestion, adding this change.

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -120,37 +122,41 @@ export const showSsoSignIn = Commands.declare('aws.codeWhisperer.sso', () => asy
...
+        if (!customizationArn && !customizationNamePrefix) {
+            return
+        }
+
+        // If given an ARN, assume a specific customization is desired and find an entry that matches it. Ignores the prefix logic.
+        // Otherwise if only a prefix is given, find an entry that matches it.
+        // Backwards compatible with previous implementation.
+        const match = customizationArn
+            ? getPersistedCustomizations().find(c => c.arn === customizationArn)
+            : getPersistedCustomizations().find(c => c.name?.startsWith(customizationNamePrefix as string))
```
wlovo, line 150: Needed to cast the prefix to a string since the linter can't determine whether it'll be undefined or not.

src/codewhisperer/commands/basicCommands.ts
justinmk3: > since it's a request from outside of Toolkit, it can easily pass a value that isn't expected.

If an external thing makes an invalid request to the Toolkit why would it expect the Toolkit to silently ignore that?

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -120,37 +122,41 @@ export const showSsoSignIn = Commands.declare('aws.codeWhisperer.sso', () => asy
...
+
+        // No customization match information given, exit early.
+        if (!customizationArn && !customizationNamePrefix) {
+            return
+        }
+
+        // If given an ARN, assume a specific customization is desired and find an entry that matches it. Ignores the prefix logic.
+        // Otherwise if only a prefix is given, find an entry that matches it.
+        // Backwards compatible with previous implementation.
+        const match: Customization | undefined = customizationArn
```
justinmk3, line 148: typescript should be able to infer the type
```suggestion
        const match = customizationArn
```

wlovo: How strongly opposed is the team to adding this?
I found it useful for reference on what is being matched, but I agree that on the compiler/linter side it should be able to infer the type.

wlovo: I'll omit it for now.

justinmk3: any LSP-capable editor will show the inferred type

justinmk3 (approved):

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -120,37 +121,41 @@ export const showSsoSignIn = Commands.declare('aws.codeWhisperer.sso', () => asy
...
+            ? getPersistedCustomizations().find(c => c.arn === customizationArn)
+            : getPersistedCustomizations().find(c => c.name?.startsWith(customizationNamePrefix as string))
+
+        // If no match is found, nothing to do :)
+        if (!match) {
+            getLogger().error(`No customization match found: arn=${customizationArn} prefix=${customizationNamePrefix}`)
+            return
+        }
+        // Since we selected based on a match, we'll reuse the persisted values.
+        await selectCustomization(match)
```
andrewyuq, line 157: I think this would actually cause some regression comparing to the current behavior, for the new users.
For the new users, there're no persisted customizations on their machine, so match would return nothing. In this case it will just return, in the previous logic you continued to use the arn the caller specified which actually will work if it matches the arn the service side has.

andrewyuq: To rephrase, the problem now becomes: "How does it auto-select the customizations for new users if they have't used customizations before?"
Could you manually test this case?

wlovo: During testing, we saw that persisted customizations are present after establishing a connection, which is the first action we take before we start attempting to find a match.

In manual testing, whether we gave it just the ARN or just the prefix, we saw it successfully set the correct customization. We also executed `setPersistedCustomizations([])` to clear out any stored entries prior to executing the command. Lastly, we also used a different prefix (`NGDE`) to make sure we used a customization that hasn't been used/set yet.

image

All in all, it seems to be doing what we expected, but I'm happy to follow up on any additional testing to make sure it's as robust as possible.

andrewyuq: Did you call `await setPersistedCustomizations([])` right after `await connectToEnterpriseSso(startUrl, region)`?

The code doesn't look like it will handle this case, so it's suprising to me that this is still passing.

wlovo: We did call `await setPersistedCustomizations([])`, but _before_ the connection logic.

I'm not quite sure why we would want to test the case where we clear the persisted customizations after establishing the connection when the logic is designed to execute _only if_ a connection has been established.

In the worst case, say if a connection is not established, then it will not set a customization as expected.

andrewyuq: The test case is for when a new user has installed Viceroy and CodeWhisperer for the first time, thus their machine does not have any info about any persisted customization, calling `await setPersistedCustomizations([])` before or after connection logic doesn't really matter, as long as it's cleared before we call `getPersistedCustomization()` it would simulate the usecase.

The worst case is when a connection _is_ established but we don't set the customization as expected. And that's beacuse we try to match persisted ones but users don't have such on their machine, comparing to the existing code that prioritizes the passed parameter. This is making the behavior to change.

existing:
```
if (customizationArn) {
                const match = getPersistedCustomizations().find(c => c.arn == customizationArn)
                const customization = {
                    arn: customizationArn, <-- here means even if match doesn't return anything we still favor the given arn
                    name: customizationName ?? match?.name ?? 'unknown',
                    description: customizationDescription ?? match?.description ?? 'unknown',
                }
                await selectCustomization(customization)
            }
```

```
const match = customizationArn
            ? getPersistedCustomizations().find(c => c.arn === customizationArn)
            : getPersistedCustomizations().find(c => c.name?.startsWith(customizationNamePrefix as string))

        // If no match is found, nothing to do :)
        if (!match) { <-- here it means if we don't find a match we will return early without setting any customization.
            getLogger().error(`No customization match found: arn=${customizationArn} prefix=${customizationNamePrefix}`)
            return
```

wlovo: As we discussed offline, it seems the use of persisted customizations might encounter a race condition. The expected behavior is that it shouldn't be set by establishing the connection alone and might only be set as a side-effect.

We'll add a check first to make sure the list isn't empty and call on the notify logic to establish the persisted list.

andrewyuq (approved):

PR #3952, opened 2023-10-25
Fix code-style debugs for Typescript SAM applications

bryceitoc9:
## Problem
Code-style debugs were writing the temp `tsconfig` file incorrectly:
* No `include` specified, so it used `["**/*"]` (good)
* No `exclude` specified, so it used the `outDir` (good, but...)
  * The file was being written with `outDir === sourceDir`, so EVERYTHING was excluded.

## Solution
* Force `outDir` to be the temp `output` directory
* Force the handler written to the temporary `template.yaml` to use an absolute path to the output directory

## License


dangeReis:
Is this something that can be looked int again since it doesn't look like TypeScript debug is working yet.

PR #3956, opened 2023-10-25
cloud9: skip "!isWeb" check

justinmk3:
Problem:
Cloud9 sets `isWeb` even though it has "compute".

Solution:
Add an explicit condition for Cloud9 wherever we check `isWeb`.

ref d2f03d52a221

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -281,7 +281,7 @@
         "debuggers": [
             {
                 "type": "aws-sam",
-                "when": "!isWeb",
+                "when": "isCloud9 || !isWeb",
                 "label": "%AWS.configuration.description.awssam.debug.label%",
                 "configurationAttributes": {
                     "direct-invoke": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -3182,7 +3182,7 @@
                 "command": "aws.ecs.openTaskInTerminal",
                 "title": "%AWS.ecs.openTaskInTerminal%",
                 "category": "%AWS.title%",
-                "enablement": "viewItem == awsEcsContainerNodeExecEnabled && !isWeb",
```
justinmk3, line 3185: This and 2 other cases have non-trivial conditions and cloud9 doesn't support parentheses in the `when` (or `enablement` ?) clause. But these items wouldn't show in a web context anyway (because these explorer nodes won't show), so I just removed the `isWeb` check.

bryceitoc9 (approved):

PR #3958, opened 2023-10-25
codewhisperer: status bar quickpick when clicked

nkomonen-amazon:
When the CodeWhisperer status bar is clicked we will now show a QuickPick with different options for interacting with CodeWhisperer.

- The options in the quick pick are the same as the CodeWhisperer tree nodes in `Developer Tools` with the exception of a `sign out` option in the quick pick
- The CodeWhisperer status bar always exists and is always clickable
- Telemetry is emitted (`vscode_executeCommand`) that includes the name of the `command` that was executed as well as a `source` attribute (`codewhispererQuickPick` & `codewhispererTreeNode`) which indicates what caused it to be executed
- Updates to the CW status will be reflected in both the tree nodes and status bar

## License

### DIFF
--- a/src/codewhisperer/commands/basicCommands.ts
+++ b/src/codewhisperer/commands/basicCommands.ts
@@ -180,9 +180,9 @@ export const refreshStatusBar = Commands.declare(
         if (AuthUtil.instance.isConnectionValid()) {
             InlineCompletionService.instance.setCodeWhispererStatusBarOk()
         } else if (AuthUtil.instance.isConnectionExpired()) {
-            InlineCompletionService.instance.setCodeWhispererStatusBarDisconnected()
+            InlineCompletionService.instance.setCodeWhispererStatusBarExpired()
         } else {
-            InlineCompletionService.instance.hideCodeWhispererStatusBar()
+            InlineCompletionService.instance.setCodeWhispererStatusBarNotConnected()
         }
     }
 )
--- a/src/codewhisperer/commands/basicCommands.ts
+++ b/src/codewhisperer/commands/basicCommands.ts
@@ -202,3 +202,8 @@ export const showCodeWhispererQuickPick = Commands.declare({ id: showCodeWhisper
         ignoreFocusOut: false,
     }).prompt()
 })
+
+export const signoutCodeWhisperer = Commands.declare(
+    'aws.codewhisperer.signout',
+    (auth: AuthUtil) => () => auth.secondaryAuth.deleteConnection()
+)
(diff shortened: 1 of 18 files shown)


src/codewhisperer/commands/basicCommands.ts
```diff
@@ -203,3 +192,17 @@ export const notifyNewCustomizationsCmd = Commands.declare(
         notifyNewCustomizations().then()
     }
 )
+
+export const showCodeWhispererQuickPickCommand = 'aws.codewhisperer.quickpick'
```
justinmk3, line 196: quickpick is an implementation detail, can we name this something more semantic similar to `aws.codecatalyst.openDevEnv` , `aws.codecatalyst.listCommands`, etc.
```suggestion
export const showCodeWhispererActions = 'aws.codewhisperer.listCommands'
```

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -203,3 +192,17 @@ export const notifyNewCustomizationsCmd = Commands.declare(
...
+export const showCodeWhispererQuickPick = Commands.declare({ id: showCodeWhispererQuickPickCommand }, () => () => {
```
justinmk3, line 197:
```suggestion
export const listCommands = Commands.declare({ id: listCommandsCommandId }, () => () => {
```

justinmk3 (approved):

src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -142,42 +150,105 @@ export class InlineCompletionService {
...
-        this.statusBar.hide()
+/** The states that the completion service can be in */
+const states = {
+    loading: 'loading',
+    ok: 'ok',
+    expired: 'expired',
+    notConnected: 'notConnected',
+} as const
+
+export class CodeWhispererStatusBar {
```
justinmk3, line 200: would it be more discoverable for this to live in a module next to `auth/ui/statusBarItem.ts` ? `inlineCompletionService.ts` seems like a random place for it to live.

src/codewhisperer/commands/types.ts
```diff
@@ -0,0 +1,12 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/** Indicates a CodeWhisperer command was executed through a tree node */
+export const cwTreeNodeSource = 'codewhispererTreeNode'
+/** Indicates a CodeWhisperer command was executed through a quick pick item */
+export const cwQuickPickSource = 'codewhispererQuickPick'
```
justinmk3, line 9: seems over-specific. could use the `Component` type from telemetry.gen.ts

Will-ShaoHua (approved):

src/codewhisperer/explorer/codewhispererChildrenNodes.ts
```diff
@@ -3,102 +3,290 @@
...
+                    : {
+                          label: labelResume,
+                          iconPath: iconResume,
+                      }
+            )
+        case 'item':
+            return {
+                data: 'autoSuggestions',
+                label: pause ? codicon`${iconPause} ${labelPause}` : codicon`${iconResume} ${labelResume}`,
+                description: pause ? 'Currently RUNNING' : 'Currently PAUSED',
```
justinmk3, line 58: Is "currently" needed in the message?

PR #3962, opened 2023-10-26
perf(startup): avoid templateRegistry in awsFiletypes.ts

justinmk3:
# Problem:
The AWS Documents handler (`awsFiletypes.ts`) is triggered frequently. It calls `globals.templateRegistry` which triggers an expensive "Scanning CloudFormation templates..." process. #3510

# Solution:
Extract the validation logic out of `CloudFormationTemplateRegistry.process()` so that `awsFiletypes.ts` can use it without requesting the full registry.

## License

### DIFF
--- a/src/shared/cloudformation/cloudformation.ts
+++ b/src/shared/cloudformation/cloudformation.ts
@@ -13,11 +13,22 @@ import { SystemUtilities } from '../systemUtilities'
 import { getLogger } from '../logger'
 import { lambdaPackageTypeImage } from '../constants'
 import { isCloud9 } from '../extensionUtilities'
+import { isUntitledScheme, normalizeVSCodeUri } from '../utilities/vsCodeUtils'
 
 export namespace CloudFormation {
     export const SERVERLESS_API_TYPE = 'AWS::Serverless::Api' // eslint-disable-line @typescript-eslint/naming-convention
     export const SERVERLESS_FUNCTION_TYPE = 'AWS::Serverless::Function' // eslint-disable-line @typescript-eslint/naming-convention
     export const LAMBDA_FUNCTION_TYPE = 'AWS::Lambda::Function' // eslint-disable-line @typescript-eslint/naming-convention
+
+    export const templateFileGlobPattern = '**/*.{yaml,yml}'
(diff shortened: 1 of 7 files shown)


src/shared/cloudformation/cloudformation.ts
```diff
@@ -384,6 +411,47 @@ export namespace CloudFormation {
...
+        const path = normalizeVSCodeUri(uri)
+        try {
+            if (isUntitledScheme(uri)) {
+                if (!contents) {
+                    // this error technically just throw us into the catch so the error message isn't used
+                    throw new Error('Contents must be defined for untitled uris')
+                }
+                rv.template = await CloudFormation.loadByContents(contents, false)
+            } else {
+                rv.template = await CloudFormation.load(path, false)
```
nkomonen-amazon, line 431:
```suggestion
                rv.template = await CloudFormation.load(normalizeVSCodeUri(uri), false)
```

src/shared/cloudformation/cloudformation.ts
```diff
@@ -384,6 +411,47 @@ export namespace CloudFormation {
         return template
     }
 
+    export async function tryLoad(
```
nkomonen-amazon, line 414: Something like this to clarify the conditions

```suggestion
   /** 
   * Returns the parsed the content of the given uri, but if the URI does not exist on disk (untitled scheme) it 
   * parses the given content */
    export async function tryLoad(
```

src/shared/cloudformation/cloudformation.ts
```diff
@@ -384,6 +411,47 @@ export namespace CloudFormation {
...
+                rv.template = await CloudFormation.load(path, false)
+            }
+        } catch (e) {
+            return {
+                template: undefined,
+                kind: undefined,
+            }
+        }
+
+        // same heuristic used by cfn-lint team:
```
nkomonen-amazon, line 440:
```suggestion
        // Determine whether content is 'sam' or 'cfn' using the same heuristic used by cfn-lint team:
```

nkomonen-amazon (approved):

PR #3968, opened 2023-10-27
refactor: eliminate use of "namespace"

justinmk3:
Problem:
namespace is deprecated and unnecessary.

Solution:
remove use of `namespace`.
add a lint rule to prevent it.
clean up old dead code.

## License


src/shared/vscode/commands.ts
```diff
@@ -18,6 +18,7 @@ export interface Commands {
 /**
  * @deprecated Use `./commands2.ts` instead when possible.
  */
+// eslint-disable-next-line @typescript-eslint/no-namespace
 export namespace Commands {
```
justinmk3, line 22: Will migrate this in followup PR.

PR #3974, opened 2023-10-27
Static cache key for builder ID SSO

bryceitoc9:
## Problem
Using Builder ID SSO with an SSH instance would force you to log in anew with every folder you open. This is annoying since users will have to constantly log in, have their cache fill up with multiple entries, and would set different expiries for each folder

## Solution
Use a static Builder ID cache key. This will work as long as we're not expecting users to constantly be logging out of and back into Builder ID SSO.
* From a new folder, the user will have to log in as usual to create the `curr-builder-id` key in our cache.
* Subsequent folders will have you log in with the builder ID, but will not force you to go through the browser login flow afterwards.
* ***BLOCKER FROM PUSHING OUT!!!*** Some simple testing is showing that this won't undo existing folders that already have a cache key, even if you delete the cache entry. We need to find a way to force this to use the `curr-builder-id` ID.
  * If a user signs out in a folder and then signs in again later, this will pick up the correct key on the sign in. However, any "expired connections" and reconnections will continue using the old key. This includes deleting the existing cached files.

## License

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -17,7 +17,7 @@ import { Timeout } from '../shared/utilities/timeoutUtils'
 import { errorCode, isAwsError, isNetworkError, ToolkitError, UnknownError } from '../shared/errors'
 import { getCache } from './sso/cache'
 import { createFactoryFunction, isNonNullable, Mutable } from '../shared/utilities/tsUtils'
-import { builderIdStartUrl, SsoToken, truncateStartUrl } from './sso/model'
+import { builderIdCacheKey, builderIdStartUrl, SsoToken, truncateStartUrl } from './sso/model'
 import { SsoClient } from './sso/clients'
 import { getLogger } from '../shared/logger'
 import { CredentialsProviderManager } from './providers/credentialsProviderManager'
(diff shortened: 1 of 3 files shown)


src/auth/sso/model.ts
```diff
@@ -78,6 +78,7 @@ export interface SsoProfile {
 
 export const builderIdStartUrl = 'https://view.awsapps.com/start'
 export const trustedDomainCancellation = 'TrustedDomainCancellation'
+export const builderIdCacheKey = 'curr-builder-id'
```
justinmk3, line 81:
```suggestion
export const builderIdCacheKey = 'aws-builder-id'
```

src/auth/auth.ts
```diff
@@ -263,13 +268,14 @@ export class Auth implements AuthService, ConnectionManager {
...
-    public async createConnection(profile: Profile): Promise<Connection> {
+    public async createConnection(profile: SsoProfile, id?: string): Promise<SsoConnection>
+    public async createConnection(profile: Profile, id?: string): Promise<Connection> {
         if (profile.type === 'iam') {
             throw new Error('Creating IAM connections is not supported')
         }
 
-        const id = uuid.v4()
+        // builder ID uses a static ID, otherwise use random uuid
+        id = id ?? uuid.v4()
```
justinmk3, line 278: "otherwise" means IdC? should we key by the startUrl+region for IdC?

bryceitoc9:
https://github.com/aws/aws-toolkit-vscode/pull/3983 is better.

PR #3983, opened 2023-10-31
Partition memento by hostname

bryceitoc9:
## Problem
Since we store connections by UUID, we need a consistent mapping per machine so users don't have to continually reauth when using new folders on the same remote machines/WSL.

## Solution
Use global state (the same across all machines connected by the same client), and partition on the hostname. This way, all workspaces connected through a single machine will use the same set of keys per host, and will autoconnect to the last used ID per host.
* One drawback: if a user is using multiple clients to connect to a single instance, they'll have to reauth again (and create new keys).

## License


src/extension.ts
```diff
@@ -73,14 +73,20 @@ import { Logging } from './shared/logger/commands'
...
 
 export async function activate(context: vscode.ExtensionContext) {
     await initializeComputeRegion()
     const activationStartedOn = Date.now()
     localize = nls.loadMessageBundle()
+
     initialize(context)
+    // async init global hostname so it's accessible later to memento
+    const proc = new ChildProcess('hostname', [], { collect: true })
+    globals.hostname = (await proc.run()).stdout.trim() ?? 'unknown-host'
```
nkomonen-amazon, line 88: Should we move this logic in to its own function w/ a different name?
The semantics are something that identifies the "machine", so maybe like "machineId".
And if the `hostname` implementation needs to be changed we only need to update the `getMachineId()` function.
```suggestion
    globals.machineId = getMachineId()
    
    ...

function getMachineId(): string {
  const proc = new ChildProcess('hostname', [], { collect: true })
  await proc.run()).stdout.trim() ?? 'unknown-host'
}
```

src/shared/extensionGlobals.ts
```diff
@@ -115,6 +115,7 @@ interface ToolkitGlobals {
     codelensRootRegistry: CodelensRootRegistry
     resourceManager: AwsResourceManager
     uriHandler: UriHandler
+    hostname: string
```
nkomonen-amazon, line 118: Something like this:

```suggestion
    /** An id to differentiate the current machine being run on. Can help distinguish a remote from a local machine.  */
    machineId: string
```

src/auth/auth.ts
```diff
@@ -766,7 +766,7 @@ export class Auth implements AuthService, ConnectionManager {
                 return partition(globals.context.globalState, devEnvId)
             }
 
-            return globals.context.workspaceState
+            return partition(globals.context.globalState, globals.hostname)
```
nkomonen-amazon, line 769: Can we add a comment for context to future readers:

`// We are connected to a remote backend and want to maintain a separate state from our local state`

src/extension.ts
bryceitoc9: Got a location in mind?

src/extension.ts
```diff
@@ -73,15 +73,17 @@ import { Logging } from './shared/logger/commands'
...
 
 export async function activate(context: vscode.ExtensionContext) {
     await initializeComputeRegion()
     const activationStartedOn = Date.now()
     localize = nls.loadMessageBundle()
-    initialize(context)
-    initializeManifestPaths(context)
+
+    await initialize(context)
+    ;(globals.machineId = await getMachineId()), initializeManifestPaths(context)
```
nkomonen-amazon, line 86: Whats the reason for this syntax? Couldn't it just be
```suggestion
    await initialize(context)
    initializeManifestPaths(context)
    globals.machineId = await getMachineId()
```

nkomonen-amazon (approved):

src/extension.ts
bryceitoc9: I think the linter put this in? Let me try swapping it.

bryceitoc9: Yeah IDK, maybe it was just something weird, it went through this time.

src/shared/extensionGlobals.ts
```diff
@@ -81,7 +81,7 @@ export function checkDidReload(context: ExtensionContext): boolean {
     return !!context.globalState.get<string>('ACTIVATION_LAUNCH_PATH_KEY')
 }
 
-export function initialize(context: ExtensionContext): ToolkitGlobals {
+export async function initialize(context: ExtensionContext): Promise<ToolkitGlobals> {
```
justinmk3, line 84: Does this need to be async?

src/extension.ts
```diff
@@ -416,6 +419,11 @@ async function checkSettingsHealth(settings: Settings): Promise<boolean> {
     return ok
 }
 
+async function getMachineId(): Promise<string> {
+    const proc = new ChildProcess('hostname', [], { collect: true })
```
justinmk3, line 423: should probably pass `logging:false`, else this will be noisy on machines that don't have `hostname`.

src/shared/extensionGlobals.ts
bryceitoc9: Not anymore, I was awaiting the command in init in a past version

src/extension.ts
bryceitoc9: https://github.com/aws/aws-toolkit-vscode/pull/4005

justinmk3:
For reference, here's where vscode internally creates its own concept of a machineId (for an optional feature called "edit sessions" which remotely stores editor state.):

- https://github.com/microsoft/vscode/blob/9050cd75533497f06480488eaa10d393f26b1f49/src/vs/workbench/contrib/editSessions/browser/editSessionsStorageService.ts#L254-L263
- but it looks like it's just an autogenerated UUID: https://github.com/microsoft/vscode/blob/9050cd75533497f06480488eaa10d393f26b1f49/src/vs/platform/externalServices/common/serviceMachineId.ts#L26

PR #3985, opened 2023-11-01
feature/sm-ide - hide CodeCatalyst

JedML:
## Problem
CodeCatalyst is deactivated for sagemaker but is still visible in the developer tools section. This can lead to customer confusion.

## Solution

Hide CodeCatalyst in the developer tools section when in the sagemaker ide platform

## License


src/awsexplorer/activation.ts
```diff
@@ -76,7 +77,9 @@ export async function activate(args: {
     )
 
     const authProvider = CodeCatalystAuthenticationProvider.fromContext(args.context.extensionContext)
-    const codecatalystNode = isCloud9('classic') ? [] : [new CodeCatalystRootNode(authProvider)]
+    const codecatalystNode = hasVendedIamCredentials(isCloud9('classic'))
```
justinmk3, line 80: Thanks for considering the generic case, but I'm not sure that "vended credentials" means we always want to disable codecatalyst. I think this is a case where explicitly using isSageMaker is correct? https://github.com/aws/aws-toolkit-vscode/blob/75d3537de7285d9a40a4ba0fe8af1e4e1b334ba4/src/shared/extensionUtilities.ts#L135

justinmk3:
```suggestion
    const codecatalystNode = isCloud9('classic') || isSageMaker()
```

justinmk3 (approved):

PR #3987, opened 2023-11-01
fix: cannot cancel "Scanning CloudFormation templates..."

bryceitoc9:
## Problem
Cancelling the CFN registry toast would continue registering YAML files in the background, and merely just hid the popup

## Solution
Cancelling through the widget should now stop after the current file finishes processing
* tested best with a breakpoint on the processing--it looks like it stops immediately but haven't tried this on a slow machine with a lot of YAML (we're getting really good at not starting the registry! it doesn't do me any favors trying to test this though...)
* some light refactoring (`registry.rebuild()` is a mandatory call--this makes it easier to pass a cancellation promise)

Note: we don't have a way to restart the load, so right now, cancelling means that you have to restart the toolkit if the registry is busted.

## License

### DIFF
--- a/src/shared/cloudformation/activation.ts
+++ b/src/shared/cloudformation/activation.ts
@@ -56,12 +56,12 @@ export async function activate(extensionContext: vscode.ExtensionContext): Promi
  * and slowing down the extension starting up.
  */
 function setTemplateRegistryInGlobals(registry: CloudFormationTemplateRegistry) {
-    const registrySetupFunc = async (registry: CloudFormationTemplateRegistry) => {
-        await registry.addExcludedPattern(CloudFormation.devfileExcludePattern)
-        await registry.addExcludedPattern(CloudFormation.templateFileExcludePattern)
-        await registry.addWatchPatterns([CloudFormation.templateFileGlobPattern])
-        await registry.watchUntitledFiles()
-
+    const registrySetupFunc = async (registry: CloudFormationTemplateRegistry, cancellationPromise: Promise<void>) => {
+        registry.addExcludedPattern(CloudFormation.devfileExcludePattern)
+        registry.addExcludedPattern(CloudFormation.templateFileExcludePattern)
+        registry.addWatchPatterns([CloudFormation.templateFileGlobPattern])
+        registry.watchUntitledFiles()
+        await registry.rebuild(cancellationPromise)
         return registry
     }
 
--- a/src/shared/cloudformation/activation.ts
+++ b/src/shared/cloudformation/activation.ts
@@ -15,6 +15,7 @@ import * as CloudFormation from './cloudformation'
 import { Commands } from '../vscode/commands2'
 import globals from '../extensionGlobals'
 import { SamCliSettings } from '../sam/cli/samCliSettings'
+import { Timeout } from '../utilities/timeoutUtils'
 
 /**
  * Creates a CloudFormationTemplateRegistry which retains the state of CloudFormation templates in a workspace.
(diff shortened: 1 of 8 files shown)


src/shared/cloudformation/activation.ts
```diff
@@ -56,12 +56,12 @@ export async function activate(extensionContext: vscode.ExtensionContext): Promi
...
 function setTemplateRegistryInGlobals(registry: CloudFormationTemplateRegistry) {
-    const registrySetupFunc = async (registry: CloudFormationTemplateRegistry) => {
-        await registry.addExcludedPattern(CloudFormation.devfileExcludePattern)
-        await registry.addExcludedPattern(CloudFormation.templateFileExcludePattern)
-        await registry.addWatchPatterns([CloudFormation.templateFileGlobPattern])
-        await registry.watchUntitledFiles()
-
+    const registrySetupFunc = async (registry: CloudFormationTemplateRegistry, cancellationPromise: Promise<void>) => {
+        registry.addExcludedPattern(CloudFormation.devfileExcludePattern)
+        registry.addExcludedPattern(CloudFormation.templateFileExcludePattern)
```
justinmk3, line 61: why remove `await`? `addExcludedPattern` calls rebuild() which calls `findFiles`, surely we don't benefit if those are run in a random order.

It might not matter for the current implementation, but if async functions are unawaited that signals that they don't have shared state and are safe to be scheduled however the node event loop or OS chooses to. But if there's no advantage to doing so, then the lack of `await` is a cognitive burden, because it means we need to be sure that this assumption continues to be true.

src/shared/fs/watchedFiles.ts
```diff
@@ -275,26 +281,44 @@ export abstract class WatchedFiles<T> implements vscode.Disposable {
...
+                skips++
                 const err = e as Error
-                if (err.name !== 'Canceled') {
-                    getLogger().error('watchedFiles: findFiles("%s", "%s"): %s', glob, exclude, err.message)
-                }
+                getLogger().error('watchedFiles: findFiles("%s", "%s"): %s', glob, exclude, err.message)
             }
         }
+        getLogger().info(
+            'Registered %s items%s%s',
```
justinmk3, line 317: other log messages in this module have a `${this.name}:` prefix

src/shared/fs/watchedFiles.ts
```diff
@@ -275,26 +281,44 @@ export abstract class WatchedFiles<T> implements vscode.Disposable {
...
 
     /**
-     * Rebuilds registry using current glob and exclusion patterns.
-     * All functionality is currently internal to class, but can be made public if we want a manual "refresh" button
+     * Builds/rebuilds registry using current glob and exclusion patterns. ***Necessary to init registry***.
+     *
+     * @param cancellationPromise Cancels additional registrations when rejected. Otherwise, this
      */
-    public async rebuild(): Promise<void> {
+    public async rebuild(cancellationPromise: Promise<void>): Promise<void> {
```
justinmk3, line 288: In our codebase we tend to use `Timeout` to represent a cancellationtoken: https://github.com/aws/aws-toolkit-vscode/blob/b061a69854a15f627ff1f0d12a61040179f57eac/src/auth/credentials/utils.ts#L74

its promisify() method may be useful: https://github.com/aws/aws-toolkit-vscode/blob/master/src/shared/utilities/timeoutUtils.ts#L160

Related (but off-topic for this PR): https://taskei.amazon.dev/tasks/IDE-6674

src/shared/fs/watchedFiles.ts
bryceitoc9: The original didn't have it. Added.

src/shared/fs/templateRegistry.ts
```diff
@@ -86,15 +87,17 @@ export class AsyncCloudFormationTemplateRegistry {
         }
 
         let perf: PerfLog
+        const cancelSetupPromise = new Timeout(30 * 60 * 1000) // 30 min
```
justinmk3, line 90:
```suggestion
        const cancelSetup = new Timeout(30 * 60 * 1000) // 30 min
```

src/shared/fs/templateRegistry.ts
```diff
@@ -71,7 +71,8 @@ export class AsyncCloudFormationTemplateRegistry {
     constructor(
         private readonly instance: CloudFormationTemplateRegistry,
         private readonly asyncSetupFunc: (
-            instance: CloudFormationTemplateRegistry
+            instance: CloudFormationTemplateRegistry,
+            cancelSetup: Timeout
```
justinmk3, line 75: is this used anywhere?

src/shared/fs/watchedFiles.ts
```diff
@@ -275,26 +282,37 @@ export abstract class WatchedFiles<T> implements vscode.Disposable {
...
-                for (const item of found) {
-                    await this.addItem(item, true)
+                for (let j = 0; j < found.length && !cancellationTimeout?.completed; j++) {
+                    await this.addItem(found[j], true)
                 }
             } catch (e) {
+                // file not processable
+                skips++
                 const err = e as Error
-                if (err.name !== 'Canceled') {
```
justinmk3, line 293: why remove this? I believe I added this because vscode can throw nonsense "Canceled" errors when shutting down the debugger. So we don't want to log those because they are misleading.

justinmk3:
Will do some changes in a followup PR

PR #3988, opened 2023-11-01
Remove deprecated Python configs

bryceitoc9:
Fix broken integ tests. Unknown if this will give us a lot of spam, but it looks like there are no more linting settings for Python.

## License


justinmk3 (approved):
integ logs from last successful integ CI on master is comparable to PR (both ~27k lines)

PR #3989, opened 2023-11-02
CodeWhisperer: Add OptOut field to GenerateCompletions API

andrewyuq:
## License

### DIFF
--- a/src/codewhisperer/util/commonUtil.ts
+++ b/src/codewhisperer/util/commonUtil.ts
@@ -8,6 +8,7 @@ import * as semver from 'semver'
 import { isCloud9 } from '../../shared/extensionUtilities'
 import { getInlineSuggestEnabled } from '../../shared/utilities/editorUtilities'
 import { getLogger } from '../../shared/logger'
+import globals from '../../shared/extensionGlobals'
 
 export function getLocalDatetime() {
     const timezone = Intl.DateTimeFormat().resolvedOptions().timeZone
@@ -59,6 +60,10 @@ export function getPrefixSuffixOverlap(firstString: string, secondString: string
         i--
     }
     return secondString.slice(0, i)
+}
+
+export function getOptOutPreference() {
+    return globals.telemetry.telemetryEnabled ? 'OPTIN' : 'OPTOUT'
 }
 
 export function get(key: string, context: vscode.Memento): any {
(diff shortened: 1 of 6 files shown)


src/test/codewhisperer/util/editorContext.test.ts
```diff
@@ -7,11 +7,21 @@ import assert from 'assert'
...
 describe('editorContext', function () {
+    let telemetryEnabledDefault: boolean
+
     beforeEach(function () {
         resetCodeWhispererGlobalVariables()
+        telemetryEnabledDefault = globals.telemetry.telemetryEnabled
+    })
+
+    afterEach(function () {
+        globals.telemetry.telemetryEnabled = telemetryEnabledDefault
```
Will-ShaoHua, line 22: don't quite understand this change, does this will really change the value of `globals.telemetry.telemetryEnabled` because I dont' see `telemetryEnabledDefault` is assigned a value elsewhere

Will-ShaoHua (approved):

src/codewhisperer/util/commonUtil.ts
```diff
@@ -61,6 +62,10 @@ export function getPrefixSuffixOverlap(firstString: string, secondString: string
     return secondString.slice(0, i)
 }
 
+export function getOptOutPreference() {
+    return globals.telemetry.telemetryEnabled ? 'OPTIN' : 'OPTOUT'
```
justinmk3, line 66: I'm confused about the code organization in /codewhisperer/. 

Isn't this only for telemetry? Why doesn't it live in TelemetryHelper? https://github.com/aws/aws-toolkit-vscode/blob/8ff69bb8d10824260e03afdfdb3ae61ff27eff6d/src/codewhisperer/util/telemetryHelper.ts#L37

justinmk3 (approved):

src/codewhisperer/util/commonUtil.ts
Will-ShaoHua: yea agree, but I think not only to this PR, we do have a lot of files needed to be reorganaized

src/test/codewhisperer/util/editorContext.test.ts
andrewyuq: The tests need to change `globals.telemetry.telemetryEnabled`, the beforeEach and afterEach stores and resets this value before and after every time a test execution.

src/codewhisperer/util/commonUtil.ts
andrewyuq: It's not only for telemetry, the values will apply to the `GenerateCompletions` API and `SendTelemetryEvent` APIs. It's more of a helper util function to convert values.

Issue #3969, opened 2023-10-27
ECS exec command: don't require SimulatePrincipalPolicy permissions


npvisual:
## Problem

Using the "Open Terminal..." option on an ECS resource works great via the SSO session config (for IAM Identity Center) when using `AWSAdministratorAccess` permissions.

However when using the `AWSPowerUserAccess` permissions, I get the following error : 

```
2023-10-27 09:01:43 [ERROR]: aws.ecs.openTaskInTerminal: AccessDenied: User: arn:aws:sts::xxxx:assumed-role/AWSReservedSSO_AWSPowerUserAccess_xxxx/user is not authorized to perform: iam:SimulatePrincipalPolicy on resource: arn:aws:iam::xxxx:role/pod-1-silo-qa-xxxx because no identity-based policy allows the iam:SimulatePrincipalPolicy action (statusCode: 403; requestId: xxxxxx-f8d9-46b1-bab2-bc3e78724113)

```

What's interesting is that, when using the AWS CLI, I have no problem accessing the ECS instance with the `AWSPowerUserAccess` permissions : 

```
aws --profile staging-power ecs execute-command ...
```

with the following config

```
[profile staging-power]
sso_session = staging-power
sso_account_id = xxxxxx
sso_role_name = AWSPowerUserAccess
region = us-east-1
output = json
```

So I am wondering why the `AWSPowerUserAccess` cannot get access to the ECS instance.

## Steps to reproduce the issue

Go to : 
*  <Your Region>
* ECS
* Select your ECS cluster and task
* On your running container right click and select  "Open Terminal..."

You need to make sure your profile is connecting via the AWS CLI SSO using the `AWSPowerUserAccess` permission.

See error in the logs "Failed to run command: aws.ecs.openTaskInTerminal"

## Expected behavior

I would expect the same behavior that when I use `AWSAdministratorAccess` since I can use either permissions via the AWS CLI's `ecs execute-command` to connect to the ECS container.

## System details (run the `AWS: About Toolkit` command)

-   OS: macOS Sonoma 14.0
-   Visual Studio Code version: Version: 1.83.1 (Universal) -- see more below
-   AWS Toolkit version: v1.96.0

From the "About" menu : 
```
Version: 1.83.1 (Universal)
Commit: f1b07bd25dfad64b0167beb15359ae573aecd2cc
Date: 2023-10-10T23:46:55.789Z
Electron: 25.8.4
ElectronBuildId: 24154031
[...]

PR #3993, opened 2023-11-02
fix(ecs): exec command requires SimulatePrincipalPolicy permissions #3969

hayemaxi:
https://github.com/aws/aws-toolkit-vscode/issues/3969

Problem: We attempt `SimulatePrincipalPolicy` to verify the container task role has proper permissions before running exec command. Users may not have access to simulate policies, but may otherwise be able to run exec command without issue.

Solution: Only attempt `SimulatePrincipalPolicy` if exec command fails. Otherwise, there is no need to simulate policies.

Testing: Tested manually with PowerUser case described in #3969. As expected, if `exec command` fails then the failure reason is returned along with the results of the permissions check, if it failed (e.g. from missing permissions or permission denied to simulate policies).

## License

### DIFF
--- a/src/ecs/util.ts
+++ b/src/ecs/util.ts
@@ -9,11 +9,11 @@ import * as nls from 'vscode-nls'
 const localize = nls.loadMessageBundle()
 
 import { EcsClient } from '../shared/clients/ecsClient'
-import { IamClient } from '../shared/clients/iamClient'
+import { DefaultIamClient, IamClient } from '../shared/clients/iamClient'
 import { ToolkitError } from '../shared/errors'
 import { isCloud9 } from '../shared/extensionUtilities'
 import { getOrInstallCli } from '../shared/utilities/cliUtils'
-import { TaskDefinition } from 'aws-sdk/clients/ecs'
+import { Session, TaskDefinition } from 'aws-sdk/clients/ecs'
 import { getLogger } from '../shared/logger'
 import { SSM } from 'aws-sdk'
 import { fromExtensionManifest } from '../shared/settings'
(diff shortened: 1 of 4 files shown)


justinmk3 (approved):
Nice!

src/ecs/util.ts
```diff
@@ -71,14 +71,28 @@ export async function checkPermissionsForSsm(
...
-    const { session } = await client.executeCommand({ ...task, command })
+
+    let session: Session
+    try {
+        session = (await client.executeCommand({ ...task, command })).session!
+    } catch (execErr) {
+        await checkPermissionsForSsm(new DefaultIamClient(globals.regionProvider.defaultRegionId), {
+            taskRoleArn: taskRoleArn,
+        }).catch(permErr => {
+            throw ToolkitError.chain(permErr, `${execErr}`)
```
justinmk3, line 86: Throwing ToolkitError is generally a good idea, so thanks for noticing that.

One thing to note: stringifying errors generally isn't useful (the stacktrace is useless in a non-debug build). Look at existing usages of `ToolkitError.chain` to see the recommended pattern. The "chained" error message can give a more semantic explanation since we are handling it in a place that has more context

    throw ToolkitError.chain(permErr, 'ECS exec failed')

, or if not then perhaps print the `.message`. The stacktrace will just be noise (and this is surfaced in customer-facing messages by the [global handler](https://github.com/justinmk3/aws-toolkit-vscode/blob/106666e0bbac3236b7ef50f45dd81634d6441d05/src/extension.ts#L95-L98))

    throw ToolkitError.chain(permErr, `${(execErr as Error).message}`)

PR #3994, opened 2023-11-02
telemetry: separate execute command events with different args

nkomonen-amazon:
## Problem:

We have a `Commands` class that allows us to register a command with vscode, that when called will execute the callback. But more importantly for this PR, it emits a telemetry event.

When a Command is executed in succession within a short timeframe, the telemetry events emitted are throttled and a debounce counter is instead incremented. Once the throttle period is over, a single metric is emitted with that debounce counter indicating how many times it was called.

The problem is that we throttle a command but if it has different args when executed they all get grouped in to a single debounce counter. If I execute the same command with args "a", and another time with args "b" I do not want them to be merged in to a single debounce counter since the final metric will not include any information about "b".

## Solution:

When deciding if we have already seen a command executed and if we should throttle it, we will now additionally check the `args` of the execution call on top of the already used `id`. This will give us a unique identifier with more granularity and result in separate metrics being emitted when the `args` are different.

Note, we only get a unique identifier if "args" are primitives since it is easier to create a map key. Otherwise we only use the "id" and other executions of the command will be seen as the same thing even if their "args" are different.

## License


hayemaxi:
Are there cases of commands where we don't care about differentiating metrics by the arg values? It looks like this change would effectively just be removing throttling for those cases.

hayemaxi (approved):

src/shared/vscode/commands2.ts
```diff
@@ -367,42 +368,143 @@ interface CommandInfo<T extends Callback> {
...
+    const info = TelemetryDebounceInfo.instance.get(id)
 
+    // to reduce # of events actually emitted, we throttle the same* event for brief period of time
+    // and instead increment a counter for executions while throttled.
     if (!telemetryName && info?.startTime !== undefined && currentTime - info.startTime < threshold) {
         info.debounceCount += 1
-        telemetryInfo.set(id, info)
-        getLogger().debug(`commands: skipped telemetry for "${id}"`)
+        TelemetryDebounceInfo.instance.set(id, info)
+        getLogger().debug(`commands: skipped telemetry for "${id.id}"`)
```
bryceitoc9, line 380: Should this be `id.key`?

src/shared/vscode/commands2.ts
```diff
@@ -367,42 +368,143 @@ interface CommandInfo<T extends Callback> {
...
+    /**
+     * Returns the `source` of a {@link BaseCommandSource} if found
+     * in the given args.
+     */
+    static findSource(args: any[]): string | undefined {
+        const source = args.find(arg => arg instanceof BaseCommandSource)
+        return source ? source.source : undefined
+    }
+
+    toString() {
```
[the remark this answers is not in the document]
nkomonen-amazon: Yeah, when this object is used in something like a log this is shown instead of `object Object`

dhruvigajjar (approved):

src/shared/vscode/commands2.ts
```diff
@@ -367,42 +368,143 @@ interface CommandInfo<T extends Callback> {
...
+ */
+export class BaseCommandSource {
+    constructor(readonly source: string) {}
+
+    /**
+     * Returns the `source` of a {@link BaseCommandSource} if found
+     * in the given args.
+     */
+    static findSource(args: any[]): string | undefined {
+        const source = args.find(arg => arg instanceof BaseCommandSource)
```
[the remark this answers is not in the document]
nkomonen-amazon: This is technically possible, but unlikely since it doesn't make sense in this context to ask for more than one.

> Is there no predictable arg name that we could use instead?
Not sure what you mean by this. Can we determine by a name instead of `instanceof`?

src/shared/vscode/commands2.ts
```diff
@@ -367,42 +368,143 @@ interface CommandInfo<T extends Callback> {
...
+    if (!(source instanceof BaseCommandSource)) {
+        // case where vscode commands can be executed without the expected args
+        return
+    }
+    telemetry.vscode_executeCommand.record({ source: source.source })
+}
+
+/**
+ * Storage of telemetry events that is used for the purposes
+ * of maintaining the count of throttled event executions.
```
justinmk3, line 457: This sounds very similar to `TelemetryLogger` https://github.com/aws/aws-toolkit-vscode/blob/df40fccff2f3d55dad8bf9c581f669c7e7edeacf/src/shared/telemetry/telemetryLogger.ts#L46

If it's too risky to rethink this now, do we have a backlog item for that?

src/shared/vscode/commands2.ts
```diff
@@ -367,42 +368,143 @@ interface CommandInfo<T extends Callback> {
...
 
             return fn(...args)
         })
 }
 
+/**
+ * Adding an instance of this to the execution args of your {@link RegisteredCommand}
+ * will set the "source" attribute in the automatically emitted `vscode_executeCommand` metric.
+ */
+export class BaseCommandSource {
```
justinmk3, line 406: This seems a bit roundabout. We already have a `[RegisteredCommand](https://github.com/aws/aws-toolkit-vscode/blob/df40fccff2f3d55dad8bf9c581f669c7e7edeacf/src/shared/vscode/commands2.ts#L49-L59)` (at least) which is a place where we could define what the "key" of a command is. For expediency, that could be something like:

```diff
diff --git a/src/shared/vscode/commands2.ts b/src/shared/vscode/commands2.ts
index dbb4596cf53d..b8ae76372b08 100644
--- a/src/shared/vscode/commands2.ts
+++ b/src/shared/vscode/commands2.ts
@@ -48,6 +48,7 @@ export interface Command<T extends Callback = Callback> {
  */
 interface RegisteredCommand<T extends Callback = Callback> extends Command<T> {
     dispose(): void
+    setKey(...): void
 }
 
 /**
```

Would this work?

src/shared/vscode/commands2.ts
justinmk3: What is the difference between the two, semantically?

nkomonen-amazon: `TelemetryLogger` maintains a linear array to buffer all logs. If we wanted to we, could internally use a buffer for `TelemetryDebounceInfo` but it doesn't make sense in this case.

The point of `TelemetryDebounceInfo` class is to maintain a counter which would require us to iterate over the entire buffer whenever we want to get the count. Instead we can have a map which we can increment whenever a new metric comes in.

If I'm not seeing what you are lets create a ticket.

justinmk3: It would require some refactoring/redesign. Yet it should be a backlog item.

src/shared/vscode/commands2.ts
nkomonen-amazon: Are you implying we now need to refactor the args of `execute()` to separately allow an object with `argName->value` mappings? This would be a lot of work. Otherwise, I may be lacking in that typescript knowledge

src/shared/vscode/commands2.ts
```diff
@@ -365,44 +370,215 @@ interface CommandInfo<T extends Callback> {
...
+            if (typeof obj === 'boolean') {
+                return obj.toString()
+            }
+            // Add more implementations here as needed
+            throw new ToolkitError(`Cannot create unique key for type: "${typeof obj}"`)
+        })
+
+        const hasher = crypto.createHash('sha256')
+        hashableObjects.forEach(o => hasher.update(o))
+        return hasher.digest('hex')
```
justinmk3, line 566: could use `getStringHash`, though a bit less efficient. https://github.com/aws/aws-toolkit-vscode/blob/bf4c7e446216343d538f8142bdc7c4ad71f6ccd3/src/shared/utilities/textUtilities.ts#L105

src/shared/vscode/commands2.ts
```diff
@@ -365,44 +370,215 @@ interface CommandInfo<T extends Callback> {
...
+
+    /**
+     * Validates the given arg matches the expected type for the attribute.
+     * Throws if the type is unexpected.
+     */
+    validate(arg: any) {
+        switch (this.attribute) {
+            case 'source':
+                if (typeof arg !== 'string') {
+                    throw new ToolkitError(
```
justinmk3, line 405: Why validate? Is this for defensiveness? If so, we can add test coverage for specific commands.

But I'm not sure that validation belongs here. In theory, any arg can be part of a key if the command definition wants it to be.

src/test/shared/vscode/runCommand.test.ts
```diff
@@ -91,5 +91,113 @@ describe('runCommand', function () {
...
+
+                // unsupported
+                await assert.rejects(() => command.execute(1), new ToolkitError(`Cannot create unique key for type: "number"`))
+            })
+
+            it('throws if "source" is not a string', async function () {
+                const command = Commands.register(
+                    {
+                        id: '_test.telemetryName.throwsOnBadSourceType',
+                        uniqueArgIndexes: [new IndexedAttribute(0, 'source')],
```
justinmk3, line 195: - if `IndexedAttribute` were an interface the ergnomics may improve. Classes lose some benefits of typescript's structural typing.
    - maybe we could flip it so that the names are specified as object keys, and the values are the indexes. 
        - that naturally also avoids duplicate indexes.
- `uniqueArgIndexes` is an example of a name that is trying to explain a lot but ends up being opaque anyway. Simple is usually better. Docstrings can reveal details if necessary, in an opt-in fashion.

```suggestion
                        key: {'source': 0, 'foo': 3},
```

src/shared/vscode/commands2.ts
nkomonen-amazon: `IndexAttribute` has string literals for `IndexAttribute.attribute`. The intent is for any arg to **not** be any value it wants to be, instead it must match the type decided in `IndexAttribute.validate()`.

Eg: `source` is meant to be the "source" attribute in `vscode_executeCommand` which is always a string

> Why validate? Is this for defensiveness? If so, we can add test coverage for specific commands.

Yeah for defensiveness. This is intended as a backup just in case the type is wrong and the implementation is not tested, otherwise a mistake can send invalid data

justinmk3: Tests are the way to go here. Or the type system. Keep it simple.

nkomonen-amazon: Deleted

src/test/shared/vscode/runCommand.test.ts
nkomonen-amazon: Good idea. Created interface `ArgIndexes` which does similar to what you mentioned.

The will instead provide a single object similar to `{0: 'source'}` which ensures unique indexes since the key can only be used once

src/shared/vscode/commands2.ts
```diff
@@ -365,44 +369,192 @@ interface CommandInfo<T extends Callback> {
...
+}
 
+/** 
+ * Used by {@link CommandInfo}, will indicate that the arg should be used
+ * as a telemetry attribute.
+ */
+export type ArgIndexes = { [index: number]: ArgIndexType }
+type ArgIndexType = MetricAttribute | undefined
+
+/** Supported `vscode_executeCommand` attributes */
```
justinmk3, line 404: "attributes" is what we currently call "metadata fields"?

src/test/globalSetup.test.ts
```diff
@@ -84,6 +85,12 @@ export const mochaHooks = {
         globals.telemetry.logger.clear()
         ;(globals.context as FakeExtensionContext).globalState = new FakeMemento()
 
+        // During testing if we do not clear this, then `vscode_executeCommand` metrics
+        // will be throttled. Between separate tests `vscode_executeCommand` metrics may not emitted
+        // when expected since their "throttle" status will persist unless we clear the
+        // state after each test.
+        TelemetryDebounceInfo.instance.clear()
```
justinmk3, line 92: this is analogous to `globals.telemetry.logger.clear()` on line 85. could go just below that, then the comment might not be needed.

src/test/shared/vscode/runCommand.test.ts
```diff
@@ -91,5 +91,102 @@ describe('runCommand', function () {
...
+                    (source: any) => {}
+                )
+                await assert.rejects(() => command.execute(1), new ToolkitError(`Unique arg indexes exceed the # of args for: "_test.telemetryName.throwsOnOutOfIndex"`))
+            })
+
+            it('throws if arg type is not supported', async function () {
+                const command = Commands.register(
+                    {
+                        id: '_test.telemetryName.throwsOnUnsupportedType',
+                        argIndexes: {0: undefined},
```
justinmk3, line 178: we want to communicate that this defines a composite key. so let's use a name that makes that clear. there is precedent for simply calling it `key` in e.g. [python](https://docs.python.org/3/howto/sorting.html#key-functions). or we can call it `compositeKey`.
  
```suggestion
                        compositeKey: {0: undefined},
```

src/shared/vscode/commands2.ts
```diff
@@ -365,44 +369,192 @@ interface CommandInfo<T extends Callback> {
...
+     * - By default, `vscode_executeCommand` is throttled for metrics of the
+     *   same `id`.
+     * - Defining arg indexes for this property, the values of the args will be combined
+     *   with the `id`. This ensures throttling only happens for the same `id` + args
+     *   instead.
+     * - Indexes starts at 0.
+     *
+     * @example
+     * const argIndexes: ArgIndexes = {
+     *     0: undefined // the value is not important
```
justinmk3, line 386: why do we need to support `undefined`? can we just require a name always?

src/shared/vscode/commands2.ts
```diff
@@ -365,44 +369,192 @@ interface CommandInfo<T extends Callback> {
...
+/** 
+ * Used by {@link CommandInfo}, will indicate that the arg should be used
+ * as a telemetry attribute.
+ */
+export type ArgIndexes = { [index: number]: ArgIndexType }
+type ArgIndexType = MetricAttribute | undefined
+
+/** Supported `vscode_executeCommand` attributes */
+const MetricAttributes = {
+    source: 'source'
```
justinmk3, line 406: what are we trying to model here? can the createKey and related things use the names provided by the `register()` call?

src/shared/vscode/commands2.ts
```diff
@@ -365,44 +369,192 @@ interface CommandInfo<T extends Callback> {
...
+     * // vscode_executeCommand -> { command: 'myCommand' source: 'SOURCE' } 
+     */
+    readonly argIndexes?: ArgIndexes
+}
 
+/** 
+ * Used by {@link CommandInfo}, will indicate that the arg should be used
+ * as a telemetry attribute.
+ */
+export type ArgIndexes = { [index: number]: ArgIndexType }
```
justinmk3, line 401: don't worry.

```suggestion
export type ArgIndexes = { [index: number]: string }
```

src/shared/vscode/commands2.ts
```diff
@@ -365,44 +369,192 @@ interface CommandInfo<T extends Callback> {
...
+     * The current implementation extracts the indexes from `args` using
+     * the indexes specificed by the {@link IndexedArg} objects.
+     */
+    private createKey(key: { id: string; args: any[]; argIndexes: ArgIndexes }) {
+        const id = key.id
+        const args = key.args
+        const uniqueIndexes: number[] = keysAsInt(key.argIndexes).sort((a, b) => a - b)
+
+        if (uniqueIndexes[uniqueIndexes.length - 1] > args.length - 1) {
+            throw new ToolkitError(`Unique arg indexes exceed the # of args for: "${id}"`)
```
justinmk3, line 517: 👍 good to have fail-fast checks like this.

src/shared/vscode/commands2.ts
```diff
@@ -365,44 +369,192 @@ interface CommandInfo<T extends Callback> {
...
+    const info = TelemetryDebounceInfo.instance.get(id)
+
+    // to reduce # of events actually emitted, we throttle the same* event for brief period of time
+    // and instead increment a counter for executions while throttled.
     if (!telemetryName && info?.startTime !== undefined && currentTime - info.startTime < threshold) {
         info.debounceCount += 1
-        telemetryInfo.set(id, info)
-        getLogger().debug(`commands: skipped telemetry for "${id}"`)
+        TelemetryDebounceInfo.instance.set(id, info)
+        getLogger().debug(`commands: skipped telemetry for "${id.id}"`)
```
justinmk3, line 423: also print the key
```suggestion
        getLogger().debug(`commands: skipped telemetry for "${id.id}" (compositeKey=...)`)
```

src/shared/vscode/commands2.ts
```diff
@@ -365,44 +369,192 @@ interface CommandInfo<T extends Callback> {
...
+    TelemetryDebounceInfo.instance.set(id, { startTime: currentTime, debounceCount: 0 })
+
+    const attributes = findAttributesToAddToMetric(id.args, id.argIndexes)
 
     return <T extends Callback>(fn: T, ...args: Parameters<T>) =>
         span.run(span => {
-            ;(span as Metric<VscodeExecuteCommand>).record({ command: id, debounceCount })
+            ;(span as Metric<VscodeExecuteCommand>).record({
+                command: id.id,
+                debounceCount,
```
justinmk3, line 439: off topic for this PR, but 

- we should probably rename `debounceCount` to `collapsedCount` or something more intuitive.
- we should also include the compositeKey in the metric

src/shared/vscode/commands2.ts
nkomonen-amazon: > what are we trying to model here?

We've defined the field `source` for `vscode_executeCommand`. The design is for a Command to use "source" which results in the value of the arg to be set in the `source` field of the `vscode_executeCommand` metric.

> can the createKey and related things use the names provided by the register() call?

We can't since we need to use the actual value of the arg and not the name to differentiate the key.

src/shared/vscode/commands2.ts
nkomonen-amazon: Current design is that setting a name will add it as a metric field. (eg "source").
When `undefined` we use that arg in the composite key but we do not include it as a metric field.

src/shared/vscode/commands2.ts
justinmk3: The names are provided by the register() call. What I don't understand is why we need this extra, explicit `MetricAttributes` type.

src/shared/vscode/commands2.ts
justinmk3: Ok that's more specific than `value is not important` :)

We can leave this for now. But I'm starting to think that it would simplify a lot if we only defined compositeKey by metrics "metadata fields" instead of args. And with that constraint, if we want to discriminate `vscode_executeCommand` invocations, then yes, the "arg(s)" should (must) be stored _on the metric_. That simplifies everything.

justinmk3: > We could always require a `string` and include all values at all indexes in the fields for `vscode_executeCommand` where the field name is the `string` and the value is the associated `arg

That sounds very good to me. It's one less "mapping" layer.

src/shared/vscode/commands2.ts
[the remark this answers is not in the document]
bryceitoc9: Yes, if we could just always look for `args.source` or something.

justinmk3 (approved):

PR #4002, opened 2023-11-03
feat(dev): display all globalState

bryceitoc9:
## Problem
You can't display all mementos at once
## Solution
Through dark debugger powers, we can display everything.
* ~~Need to try saving, or maybe turn this into a readonly editor.~~
  * Pulling the content into an untitled new editor to prevent it from being processed by the virtual file handler.
* I'm not sure if we can show secret storage
  * ...it also doesn't look like we're using the secret storage???

## License

### DIFF
--- a/src/dev/activation.ts
+++ b/src/dev/activation.ts
@@ -169,7 +169,12 @@ class VirtualObjectFile implements FileProvider {
             const value = (await this.storage.get(key)) ?? ''
             return JSON.stringify(JSON.parse(value), undefined, 4)
         } else {
-            return JSON.stringify(this.storage.get(key, {}), undefined, 4)
+            if (key !== '') {
+                return JSON.stringify(this.storage.get(key, {}), undefined, 4)
+            }
+            // lol hax
+            // as of November 2023, all of a memento's properties are stored as property `f` when minified
+            return JSON.stringify((this.storage as any).f, undefined, 4)
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/dev/activation.ts
```diff
@@ -157,7 +157,12 @@ class VirtualObjectFile implements FileProvider {
             const value = (await this.storage.get(key)) ?? ''
             return JSON.stringify(JSON.parse(value), undefined, 4)
         } else {
-            return JSON.stringify(this.storage.get(key, {}), undefined, 4)
+            if (key !== '') {
+                return JSON.stringify(this.storage.get(key, {}), undefined, 4)
+            }
+            // lol hax
+            // as of November 2023, all of a memento's properties are stored as property `f` when minified
+            return JSON.stringify((this.storage as any).f, undefined, 4)
```
bryceitoc9, line 165: If we want to get smarter about this, we can iterate through the properties and look for an obvious tell, like `object.hasOwnProperty('isExtensionFirstUse')` and return that.

We can also return a READONLY editor for this base case.

I'm not going to think too hard about this now because this is dev mode only.

src/dev/activation.ts
```diff
@@ -252,12 +257,36 @@ async function openStorageFromInput() {
...
+            this.form.key.bindPrompter(({ target }) => {
+                if (target === 'secrets') {
+                    return createInputBox({
+                        title: 'Enter a key',
+                    })
+                } else if (target === 'globals') {
+                    const items = ctx.extensionContext.globalState
+                        .keys()
+                        .map(key => {
+                            if (key === '') {
```
bryceitoc9, line 269: I may have accidentally saved a blank key. Will confirm.

PR #4007, opened 2023-11-06
codewhisperer: telemetry metrics apply fix

ctlai95:
## Problem

We want to track the number of times apply fix command is executed for a security scan issue.

## Solution

- Added `codewhisperer_codeScanIssueApplyFix` metric that emits when command is executed and what the resulting state is.
- Refactored `applySecurityFix` function and tests for readability

## License

### DIFF
--- a/src/shared/telemetry/vscodeTelemetry.json
+++ b/src/shared/telemetry/vscodeTelemetry.json
@@ -47,12 +47,6 @@
             "name": "detectorId",
             "type": "string",
             "description": "The id of the detector which produced the code scan issue"
-        },
-        {
-            "name": "applyFixSource",
-            "type": "string",
-            "allowedValues": ["hover", "webview", "quickfix"],
-            "description": "The location which triggered the apply fix command for a code scan issue"
         }
     ],
     "metrics": [
@@ -276,7 +270,7 @@
         {
             "name": "codewhisperer_codeScanIssueApplyFix",
             "description": "Called when a code scan issue suggested fix is applied",
-            "metadata": [{ "type": "findingId" }, { "type": "detectorId" }, { "type": "applyFixSource" }]
+            "metadata": [{ "type": "findingId" }, { "type": "detectorId" }, { "type": "component" }]
         }
     ]
 }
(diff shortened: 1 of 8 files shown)


src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -47,6 +47,12 @@
             "name": "detectorId",
             "type": "string",
             "description": "The id of the detector which produced the code scan issue"
+        },
+        {
+            "name": "applyFixSource",
+            "type": "string",
+            "allowedValues": ["hover", "webview", "quickfix"],
```
justinmk3, line 54: you could add these values to the existing `component` type: https://github.com/aws/aws-toolkit-common/blob/bbc34c2f5b5f1f0a24befa5c42f4193e3e12032c/telemetry/definitions/commonDefinitions.json#L57

then just use `component` in `codewhisperer_codeScanIssueApplyFix`

justinmk3 (approved):

src/test/codewhisperer/commands/basicCommands.test.ts
```diff
@@ -138,139 +138,160 @@ describe('CodeWhisperer-basicCommands', function () {
...
 
-        sandbox.restore()
+            assert.ok(saveStub.calledOnce)
+            assert.ok(loggerStub.calledOnce)
+            const actual = loggerStub.getCall(0).args[0]
+            assert.strictEqual(actual, 'Apply fix command failed. Error: Writing to file failed.')
+            assertTelemetry('codewhisperer_codeScanIssueApplyFix', {
+                detectorId: codeScanIssue.detectorId,
+                findingId: codeScanIssue.findingId,
+                applyFixSource: 'hover',
```
mj632, line 291: nit: can we use different source values (`hover`, `webview`, or `quickfix`) to verify that `applyFixSource` value is not hardcoded by any chance.

mj632 (approved):

justinmk3:
Is it planned to update the definition of `component` telemetry type, to include the new kinds?

Issue #3368, opened 2023-04-24
use "data validation" library for validating runtime objects (e.g. package.json, settings.json, etc)


justinmk3:
## Problem

For validating unknown input/data at runtime we use either [ad-hoc procedural logic](https://github.com/aws/aws-toolkit-vscode/blob/5439f9527224d8d1c0b98cd9941ee55a0cefb9ae/src/shared/sam/debugger/awsSamDebugConfigurationValidator.ts#L33-L94) or typescript type generators: https://github.com/aws/aws-toolkit-vscode/blob/5439f9527224d8d1c0b98cd9941ee55a0cefb9ae/src/shared/settings.ts#L458-L482

## Solution

TypScript ecosystem now has rather mature libraries like [zod](https://www.youtube.com/watch?v=CughWjaoXZw) which achieve something similar to clojure's [spec](https://clojure.org/guides/spec): rich description of objects with runtime-checking (as opposed to typescript's compile-time-only checking).

- use zod(?) to replace `AwsSamDebugConfigurationValidator`
- zod specs can also be used to _define_ TypeScript types, which avoids duplication for cases where you need both.
    - via `z.infer()`

justinmk3:
> a pre-processor may be even better because then we could just write interfaces/types and have them be rendered as validators at compile time
>
> .e.g instead of
> 
> ```
> const User = z.object({
>   username: z.string(),
> });
> const user = User.parse({ username: "Ludwig" });
> ```
> you do
> 
> ```
> interface User {
>     username: string
> }
> const user = parse<User>({ username: "Ludwig" })
> ```



With zod you can `z.infer()` to get a typescript type, so I think it's equivalent?

> it’s functionally the same but with zod you have to rewrite things with the type combinators whereas a pre-processor creates the combinators for you

the tradeoffs i see are:
- preprocessor would be custom (maintained by us)?
- expressing things in vanilla typescript is preferable if possible in a clear way (compared to something like zod, assuming the expression is clearer there)
- zod can express things that typescript types cannot

PR #4010, opened 2023-11-07
feat(codecatalyst): replace 'cawsStage' dev setting with granular api override setting

hayemaxi:
This allows developers who are testing changes with CodeCatalyst to 1. have flexibility with which endpoint they reach in the event there are additional endpoints other than gamma and more importantly 2. to be able to specify alternative endpoints without touching and building code.

## Testing
Unit tests and manually tested.

## License

### DIFF
--- a/.changes/next-release/Feature-a3aaba22-c8b1-4dd1-9064-f27aad72bc25.json
+++ b/.changes/next-release/Feature-a3aaba22-c8b1-4dd1-9064-f27aad72bc25.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "CodeCatalyst: Add developer setting to override the API configuration."
+}
--- a/.changes/next-release/Feature-a3aaba22-c8b1-4dd1-9064-f27aad72bc25.json
+++ b/.changes/next-release/Feature-a3aaba22-c8b1-4dd1-9064-f27aad72bc25.json
@@ -1,4 +1,0 @@
-{
-	"type": "Feature",
-	"description": "CodeCatalyst: Add developer setting to override the API configuration."
-}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -305,7 +305,7 @@ Example:
 
 ```json
 "aws.dev.codecatalystService": {
-    "region": "us-east-1",
+    "region": "us-west-2",
     "endpoint": "https://public.codecatalyst-gamma.global.api.aws",
     "hostname": "integ.stage.REMOVE.codes",
     "gitHostname": "git.gamma.source.caws.REMOVE",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -299,6 +299,19 @@ Example:
 }
 ```
 
+Overrides specifically for CodeCatalyst can be set using the `aws.dev.codecatalystService` setting. This is a JSON object consisting of keys/values required to override API calls to CodeCatalyst: `region`, `endpoint`, `hostname`, and `gitHostname`. If this setting is present, then all keys need to be explicitly provided.
+
+Example:
+
+```json
+"aws.dev.codecatalystService": {
+    "region": "us-east-1",
+    "endpoint": "https://public.codecatalyst-gamma.global.api.aws",
+    "hostname": "integ.stage.REMOVE.codes",
+    "gitHostname": "git.gamma.source.caws.REMOVE",
+}
+```
+
 ### SAM/CFN ("goformation") JSON schema
 
 See [docs/cfn-schema-support.md](./docs/cfn-schema-support.md) for how to fix
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -455,7 +455,7 @@ There are several ways to make pre-production changes available on a "preview" o
 -   **Beta artifacts:** For a "private beta" launch, `src/dev/beta.ts` contains
     logic to check a hardcoded, stable URL serving the latest `.vsix` build for
     the private beta. The hardcoded URL defined in [`dev/config.ts:betaUrl`](https://github.com/aws/aws-toolkit-vscode/blob/d9c27234c0732b021d07e184a865213d6efde8ec/src/dev/config.ts#L9)
-    also forces the Toolkit to declare version `1.999.0` (since "private beta"
+    also forces the Toolkit to declare version `1.9999.0` (since "private beta"
     has no semver and would conflict with the VSCode marketplace version,
     causing unwanted auto-updating by VSCode). Beta builds of the Toolkit
     automatically query the URL once per session per day.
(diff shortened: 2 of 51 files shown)


CONTRIBUTING.md
```diff
@@ -299,6 +299,19 @@ Example:
 }
 ```
 
+Overrides specifically for CodeCatalyst can be set using the `aws.dev.codecatalystService` setting. This is a JSON object consisting of keys/values required to override API calls to CodeCatalyst: `region`, `endpoint`, `hostname`, and `gitHostname`. If this setting is present, then all keys need to be explicitly provided.
```
nkomonen-amazon, line 302: What was the reason you needed more granularity? In my use cases it was nice and easy since I only had to switch the setting from `prod` to `gamma` but now will need to add in all this info manually.
Will be helpful if you indicate that in the PR description.

CONTRIBUTING.md
```diff
@@ -299,6 +299,19 @@ Example:
...
+
+Example:
+
+```json
+"aws.dev.codecatalystService": {
+    "region": "us-east-1",
```
nkomonen-amazon, line 308: Originally this was `us-west-2`, I'm not sure if this only works with certain regions. @justinmk3 ?

src/shared/clients/codecatalystClient.ts
```diff
@@ -38,24 +38,33 @@ interface CodeCatalystConfig {
     readonly gitHostname: string
 }
 
-export function getCodeCatalystConfig(): CodeCatalystConfig {
-    const stage = DevSettings.instance.get('cawsStage', 'prod')
```
nkomonen-amazon, line 42: Can we keep `cawsStage` for backwards compatibility, but in the case `codeCatalystService` exists then that takes priority?

src/shared/clients/codecatalystClient.ts
```diff
@@ -38,24 +38,33 @@ interface CodeCatalystConfig {
...
+}
+
+export function getCodeCatalystConfigFromSettings(settings: DevSettings): Readonly<CodeCatalystConfig> {
+    const defaultConfig = {
+        region: 'us-east-1',
+        endpoint: 'https://public.codecatalyst.global.api.aws',
+        hostname: 'codecatalyst.aws',
+        gitHostname: 'codecatalyst.aws',
     }
+    const devSetting = 'codecatalystService'
```
nkomonen-amazon, line 52: We may want to export this name so it can be used in tests and maintain a single source

CONTRIBUTING.md
hayemaxi: Will update the PR description. Yes you do have to specify all the values in settings.json, but it is a one time setup. Since settings.json supports comments, you can comment out different configurations as needed.

CONTRIBUTING.md
hayemaxi: Will update.

src/shared/clients/codecatalystClient.ts
hayemaxi: Will update.

src/shared/clients/codecatalystClient.ts
hayemaxi: I agree with backwards compatibility in general but I am wondering if it is needed at this level? `cawsStage` seems used for only very specific stages of development and I imagine it isn't widely used . Would like to keep things cleaner by avoiding redundant settings if the cost is low.

justinmk3: yeah we don't need back-compat for develper settings (usually)

src/shared/clients/codecatalystClient.ts
justinmk3: actually devsettings are "strongly typed" here: https://github.com/aws/aws-toolkit-vscode/blob/a2e3936e84a6c34bc73b0f94c45bdf31700dec13/src/shared/settings.ts#L642-L653

.changes/next-release/Feature-a3aaba22-c8b1-4dd1-9064-f27aad72bc25.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "CodeCatalyst: Add developer setting to override the API configuration."
```
justinmk3, line 3: Thanks for adding a changelog, that's always a good instinct. But in this case we probably shouldn't, because developer features are not supported and not relevant to most customers.

src/shared/clients/codecatalystClient.ts
```diff
@@ -38,24 +38,33 @@ interface CodeCatalystConfig {
...
-            region: 'us-east-1',
-            endpoint: 'https://public.codecatalyst.global.api.aws',
-            hostname: 'codecatalyst.aws',
-            gitHostname: 'codecatalyst.aws',
-        }
+export function getCodeCatalystConfig(): Readonly<CodeCatalystConfig> {
+    return getCodeCatalystConfigFromSettings(DevSettings.instance)
+}
+
+export function getCodeCatalystConfigFromSettings(settings: DevSettings): Readonly<CodeCatalystConfig> {
```
justinmk3, line 45: probably makes sense for this to live in https://github.com/aws/aws-toolkit-vscode/blob/2f7fb0adc426b80ba42207abb93a50a76a90151d/src/shared/settings.ts#L687

that is a common pattern for settings management in the codebase.

src/shared/clients/codecatalystClient.ts
```diff
@@ -38,24 +38,33 @@ interface CodeCatalystConfig {
...
+    const devSetting = 'codecatalystService'
+    const devConfig = settings.get(devSetting, {})
+
+    if (Object.keys(devConfig).length === 0) {
+        return defaultConfig
+    }
+
+    try {
+        // The configuration in dev settings should explicitly override the entire default configuration.
+        assertHasProps(devConfig, ...Object.keys(defaultConfig))
```
justinmk3, line 61: 👍  good idea. 

Related: https://github.com/aws/aws-toolkit-vscode/issues/3368

src/shared/clients/codecatalystClient.ts
```diff
@@ -38,24 +38,33 @@ interface CodeCatalystConfig {
...
+    } catch (err) {
+        throw ToolkitError.chain(err, `Dev setting '${devSetting}' has missing or invalid properties.`)
+    }
+
+    logger.getLogger().debug(`using CodeCatalyst configuration from dev setting '${devSetting}'`)
```
justinmk3, line 66: will this appear every time the config is requested? might be noisy. we have a `once()` function that can be useful

hayemaxi: Added this, nice call out!

src/shared/clients/codecatalystClient.ts
```diff
@@ -30,32 +29,24 @@ import {
...
 
-interface CodeCatalystConfig {
+export interface CodeCatalystConfig {
     readonly region: string
     readonly endpoint: string
     readonly hostname: string
     readonly gitHostname: string
 }
 
+export const defaultServiceConfig: CodeCatalystConfig = {
```
justinmk3, line 41: this could live in `DevSettings` too. after all, it's a default setting value.

justinmk3 (approved):

nkomonen-amazon (approved):

PR #4011, opened 2023-11-07
feat(codewhisperer): add file path in webview

mj632:
## Problem
After opening a webview see detail of the vulnerability, it is hard to map the vulnerability details to the line and file.

## Solution
added file path field in webview details to navigate back to the vulnerable line
image

## License


justinmk3:
> After opening a webview see detail of the vulnerability, it is hard to map the vulnerability details to the line and file.

That's the inverse of the problem the webview was supposed to solve :/

justinmk3 (approved):

ctlai95 (approved):

PR #4012, opened 2023-11-07
fix(telemetry): wrong requestId in UserDecision (Prod data polluted)

laileni-aws:
## Problem
Sending last recommendation request id for all the `userDecision` telemetry events in a session/trigger.
https://issues.amazon.com/issues/ConsolasIssue-12227
## Solution
Send `userDecision` events with respective requestId of each recommendation in a session/trigger.

## License

### DIFF
--- a/src/codewhisperer/service/recommendationHandler.ts
+++ b/src/codewhisperer/service/recommendationHandler.ts
@@ -68,6 +68,7 @@ const lock = new AsyncLock({ maxPending: 1 })
 export class RecommendationHandler {
     public lastInvocationTime: number
     public requestId: string
+    public requestIdList: string[]
     private nextToken: string
     private cancellationToken: vscode.CancellationTokenSource
     public isGenerateRecommendationInProgress: boolean
@@ -83,6 +84,7 @@ export class RecommendationHandler {
 
     constructor() {
         this.requestId = ''
+        this.requestIdList = []
         this.nextToken = ''
         this.lastInvocationTime = performance.now() - CodeWhispererConstants.invocationTimeIntervalThreshold * 1000
         this.cancellationToken = new vscode.CancellationTokenSource()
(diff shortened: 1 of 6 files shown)


justinmk3 (approved):

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -178,7 +178,7 @@ export class TelemetryHelper {
                 recommendationSuggestionState?.set(i, 'Empty')
             }
             const event: CodewhispererUserDecision = {
-                codewhispererRequestId: requestId,
+                codewhispererRequestId: requestIdList[i],
```
andrewyuq, line 181: Ideally it should be a list of recommenationContexts where in each context there's a recommendation, and a requestId(and other things related to that completion). So you don't have to maintain two separate list and worry about the potential problem of inconsistency between these two lists.

laileni-aws: What if there is no recommendation? but still we need to send the requestId? `recordUserDecisionTelemetryForEmptyList`. I think, in this case we can not map recommendation to requestId.
Correct me if I am wrong?

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -83,6 +84,7 @@ export class RecommendationHandler {
 
     constructor() {
         this.requestId = ''
+        this.requestIdList = []
```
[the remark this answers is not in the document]
laileni-aws: Yeah, updated accordingly!

src/codewhisperer/util/telemetryHelper.ts
andrewyuq: for that one you can just use the local requestId and pass that to the event, for the rest, we can add the recommendationContext to the session using similar logic in line 383 of the file recommendationHandler.ts

andrewyuq: This is not a blocker for now.

laileni-aws: Added a TODO for further changes.

andrewyuq (approved):

PR #4019, opened 2023-11-08
codewhisperer: add logs for listCustomization API call

Will-ShaoHua:
## Problem
image

## License

### DIFF
--- a/src/codewhisperer/client/codewhisperer.ts
+++ b/src/codewhisperer/client/codewhisperer.ts
@@ -201,7 +201,7 @@ export class DefaultCodeWhispererClient {
         return pageableToCollection(requester, {}, 'nextToken')
             .promise()
             .then(resps => {
-                let logStr = 'codewhisperer: listAvailableCustomizations API request:'
+                let logStr = 'CodeWhisperer: listAvailableCustomizations API request:'
                 resps.forEach(resp => {
                     const requestId = resp.$response.requestId
                     logStr += `\n${indent('RequestID: ', 4)}${requestId},\n${indent('Customizations:', 4)}`
--- a/src/codewhisperer/client/codewhisperer.ts
+++ b/src/codewhisperer/client/codewhisperer.ts
@@ -204,9 +204,10 @@ export class DefaultCodeWhispererClient {
                 let logStr = 'codewhisperer: listAvailableCustomizations API request:'
                 resps.forEach(resp => {
                     const requestId = resp.$response.requestId
-                    logStr += `\n${indent('RequestID: ')}${requestId},\n${indent('Customizations:')}`
+                    logStr += `\n${indent('RequestID: ', 4)}${requestId},\n${indent('Customizations:', 4)}`
                     resp.customizations.forEach((c, index) => {
-                        logStr += `\n       ${index.toString().padStart(2, '0')}: ${c.name?.trim()}`
+                        const entry = `${index.toString().padStart(2, '0')}: ${c.name?.trim()}`
+                        logStr += `\n${indent(entry, 8)}`
                     })
                 })
                 getLogger().debug(logStr)
--- a/src/codewhisperer/client/codewhisperer.ts
+++ b/src/codewhisperer/client/codewhisperer.ts
@@ -19,6 +19,7 @@ import apiConfig = require('./service-2.json')
 import userApiConfig = require('./user-service-2.json')
 import { session } from '../util/codeWhispererSession'
 import { getLogger } from '../../shared/logger'
+import { indent } from '../../shared/utilities/textUtilities'
 
 export type ProgrammingLanguage = Readonly<
     CodeWhispererClient.ProgrammingLanguage | CodeWhispererUserClient.ProgrammingLanguage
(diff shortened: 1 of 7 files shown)


src/codewhisperer/util/customizationUtil.ts
```diff
@@ -301,12 +301,25 @@ export const selectCustomization = async (customization: Customization) => {
...
+            return listAvailableCustomizationsResponse.customizations
+        })
         .forEach(customizations => {
             items.push(...customizations)
         })
 
+    getLogger().debug(
+        `codewhisperer:listAvailableCustomizations: requestIds: ${requestIds}, customizations: ${items.map(
+            custom => custom.name
+        )}`
```
justinmk3, line 321: what does the log message look like? Ensure that it is readable, using indentionan or possibly the `%O` format string. 

See https://github.com/aws/aws-toolkit-vscode/pull/3888 for examples.

src/codewhisperer/util/customizationUtil.ts
```diff
@@ -301,12 +301,25 @@ export const selectCustomization = async (customization: Customization) => {
 export const getAvailableCustomizationsList = async () => {
     const items: Customization[] = []
     const response = await codeWhispererClient.listAvailableCustomizations()
```
justinmk3, line 303: can the client automatically log all calls, similar to the codecatalyst client? instead of logging at this higher layer in an ad-hoc way.

src/codewhisperer/models/constants.ts
```diff
@@ -7,7 +7,7 @@
  * SDK Client
  */
 
-export const endpoint = 'https://codewhisperer.us-east-1.amazonaws.com/'
+export const endpoint = 'https://rts.gamma-us-east-1.codewhisperer.ai.aws.dev/'
```
justinmk3, line 10: this is targeting mainline?

Will-ShaoHua: oh let me revert it

Will-ShaoHua: accidental change

src/codewhisperer/util/customizationUtil.ts
Will-ShaoHua: sg, will look into how codecatalyst take care of it

src/codewhisperer/util/customizationUtil.ts
Will-ShaoHua: ack

src/codewhisperer/util/customizationUtil.ts
justinmk3: in the CC client all requests go through a `call()` abstraction: https://github.com/aws/aws-toolkit-vscode-staging/blob/a717af3883355d8d3bf69fa444e1841b25dc63dd/src/shared/clients/codecatalystClient.ts#L222-L230

I think the SDKs also support basic logging with a flag, so that could be another way, but it's less customizable.

Ideally we would have a base "client" that does this for all clients, but that is TODO.

Will-ShaoHua: yea I have looked it, but @justinmk3  I think I will need more time to improve our CW client facade to be like CodeCatlyst's. I still moved the logging logic inside CW client facade `codewhisperer.ts` as an interim solution  tho, just the functionality is not as comprehensive as `codecatlystClient.ts` is. Let me cut a backlog item for it

Will-ShaoHua: will share the SIMs(including the codeScan one) in our channel

src/codewhisperer/client/codewhisperer.ts
```diff
@@ -197,7 +198,20 @@ export class DefaultCodeWhispererClient {
...
-        return pageableToCollection(requester, {}, 'nextToken').promise()
+        return pageableToCollection(requester, {}, 'nextToken')
+            .promise()
+            .then(resps => {
+                let logStr = 'codewhisperer: listAvailableCustomizations API request:'
+                resps.forEach(resp => {
+                    const requestId = resp.$response.requestId
+                    logStr += `\n${indent('RequestID: ')}${requestId},\n${indent('Customizations:')}`
+                    resp.customizations.forEach((c, index) => {
+                        logStr += `\n       ${index.toString().padStart(2, '0')}: ${c.name?.trim()}`
```
justinmk3, line 209: could use indent() instead of doing this manually

justinmk3 (approved):

src/codewhisperer/client/codewhisperer.ts
Will-ShaoHua: ack, updated

justinmk3 (approved):
lgtm!

src/codewhisperer/client/codewhisperer.ts
```diff
@@ -197,7 +198,21 @@ export class DefaultCodeWhispererClient {
         const client = await this.createUserSdkClient()
         const requester = async (request: CodeWhispererUserClient.ListAvailableCustomizationsRequest) =>
             client.listAvailableCustomizations(request).promise()
-        return pageableToCollection(requester, {}, 'nextToken').promise()
+        return pageableToCollection(requester, {}, 'nextToken')
+            .promise()
+            .then(resps => {
+                let logStr = 'codewhisperer: listAvailableCustomizations API request:'
```
andrewyuq, line 204: `CodeWhisperer`

andrewyuq (approved):

PR #4021, opened 2023-11-08
codewhisperer: move images to resource folder

ctlai95:
- Moved severity images to `resources/` folder which removes the need for ad-hoc `copyFiles` task.
- Updated security issue hover content

## License


mj632 (approved):

src/codewhisperer/service/securityIssueHoverProvider.ts
```diff
@@ -48,21 +48,12 @@ export class SecurityIssueHoverProvider extends SecurityIssueProvider implements
         markdownString.isTrusted = true
         markdownString.supportHtml = true
         markdownString.supportThemeIcons = true
-        markdownString.baseUri = vscode.Uri.file(
-            path.join(globals.context.extensionPath, 'dist/src/codewhisperer/images/')
-        )
+        markdownString.baseUri = vscode.Uri.file(path.join(globals.context.extensionPath, 'resources/images/'))
```
justinmk3, line 51: If these are icons (I didn't know what "image" referred to in the previous PR), they likely belong somewhere in `resources/icons/`

ctlai95: I was thinking that initially, but it seems like the format of these svgs are different and maybe causing issues in generating the font file.
```
Found a clipPath element in the icon "aws-codewhisperer-severity-critical" the result may be different than expected
```

If I try to use the generated icons in the hover or webview, the icon doesn't render properly. Since I'm not using it this way I figured I'd leave it outside of `resources/icons/`

justinmk3: hmm. please add that explanation to the commit message.

justinmk3 (approved):

PR #4028, opened 2023-11-10
codewhisperer: open file in view column one

ctlai95:
## Problem

Open file from webview is opening a duplicate tab in the same active pane.

## Solution

Open the file in view column one and not the active pane

## License

### DIFF
--- a/src/codewhisperer/views/securityIssue/securityIssueWebview.ts
+++ b/src/codewhisperer/views/securityIssue/securityIssueWebview.ts
@@ -44,10 +44,13 @@ export class SecurityIssueWebview extends VueWebview {
 
     public navigateToFile() {
         if (this.issue && this.filePath) {
-            const position = new vscode.Position(this.issue.startLine, 1)
-            const uri = vscode.Uri.file(this.filePath)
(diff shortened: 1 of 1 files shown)


src/codewhisperer/views/securityIssue/securityIssueWebview.ts
```diff
@@ -44,10 +44,13 @@ export class SecurityIssueWebview extends VueWebview {
 
     public navigateToFile() {
         if (this.issue && this.filePath) {
-            const position = new vscode.Position(this.issue.startLine, 1)
-            const uri = vscode.Uri.file(this.filePath)
-            vscode.commands.executeCommand('vscode.open', uri, {
-                selection: new vscode.Selection(position, position),
+            const range = new vscode.Range(this.issue.startLine, 0, this.issue.endLine, 0)
+            vscode.workspace.openTextDocument(this.filePath).then(doc => {
```
nkomonen-amazon, line 48: You will probably want to return this. In the future if a caller wants to ensure the text document is opened before a subsequent action they will have the ability to do so.

nkomonen-amazon (approved):

mj632 (approved):

PR #4030, opened 2023-11-11
codewhisperer: [refactor] replace sdk pojo with plugin pojo

Will-ShaoHua:
## License

### DIFF
--- a/src/codewhisperer/client/codewhisperer.ts
+++ b/src/codewhisperer/client/codewhisperer.ts
@@ -20,11 +20,14 @@ import userApiConfig = require('./user-service-2.json')
 import { session } from '../util/codeWhispererSession'
 import { getLogger } from '../../shared/logger'
 import { indent } from '../../shared/utilities/textUtilities'
+import { CodewhispererLanguage } from '../../shared/telemetry/telemetry.gen'
 
 export type ProgrammingLanguage = Readonly<
     CodeWhispererClient.ProgrammingLanguage | CodeWhispererUserClient.ProgrammingLanguage
 >
-export type FileContext = Readonly<CodeWhispererClient.FileContext | CodeWhispererUserClient.FileContext>
+export type FileContext = Omit<CodeWhispererUserClient.FileContext, 'programmingLanguage'> & {
+    language: CodewhispererLanguage
+}
 export type ListRecommendationsRequest = Readonly<
     CodeWhispererClient.ListRecommendationsRequest | CodeWhispererUserClient.GenerateCompletionsRequest
 >
--- a/src/codewhisperer/service/classifierTrigger.ts
+++ b/src/codewhisperer/service/classifierTrigger.ts
@@ -5,10 +5,9 @@
 
 import * as os from 'os'
 import * as vscode from 'vscode'
-import { CodewhispererAutomatedTriggerType } from '../../shared/telemetry/telemetry'
+import { CodewhispererAutomatedTriggerType, CodewhispererLanguage } from '../../shared/telemetry/telemetry'
 import { extractContextForCodeWhisperer } from '../util/editorContext'
 import { TelemetryHelper } from '../util/telemetryHelper'
-import { ProgrammingLanguage } from '../client/codewhispereruserclient'
 
 interface normalizedCoefficients {
     readonly lineNum: number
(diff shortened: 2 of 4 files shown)


src/codewhisperer/service/classifierTrigger.ts
```diff
@@ -44,7 +43,7 @@ export class ClassifierTrigger {
         Enter: 0.2853,
     }
 
-    private languageCoefficientMap: Readonly<Record<string, number>> = {
+    private languageCoefficientMap: Readonly<Record<CodewhispererLanguage, number>> = {
```
zixlin7, line 46: so whenever we add a new language we have to also update the language coefficient here? That's not super ideal because new languages usually won't be supported by classifier trigger yet

Will-ShaoHua: but the good side is the compiler will know in advance if you have typo

Will-ShaoHua: which will not be captured if we use plain string

Will-ShaoHua: i don't think it will add much burden when we add a new language tho

zixlin7: can we make it `Partial<CodewhispererLanguage>`?

Will-ShaoHua: yes that sgtm, it's even better in terms of keeping things easy and prevent error

zixlin7: it's not burden it's that adding coefficient needs to be deliberate

Will-ShaoHua: yes, but we should use an unified interface for programming language instead plain string, which makes things error prone

src/codewhisperer/client/codewhisperer.ts
```diff
@@ -20,11 +20,14 @@ import userApiConfig = require('./user-service-2.json')
...
 import { getLogger } from '../../shared/logger'
 import { indent } from '../../shared/utilities/textUtilities'
+import { CodewhispererLanguage } from '../../shared/telemetry/telemetry.gen'
 
 export type ProgrammingLanguage = Readonly<
     CodeWhispererClient.ProgrammingLanguage | CodeWhispererUserClient.ProgrammingLanguage
 >
-export type FileContext = Readonly<CodeWhispererClient.FileContext | CodeWhispererUserClient.FileContext>
+export type FileContext = Omit<CodeWhispererUserClient.FileContext, 'programmingLanguage'> & {
+    language: CodewhispererLanguage
```
Will-ShaoHua, line 29: I will add more fields here in another PR

Will-ShaoHua: for example `startPos`, which currently is sitting in `TelemetryHelper` as a state to maintain

src/test/codewhisperer/util/editorContext.test.ts
```diff
@@ -4,23 +4,63 @@
...
+        }[] = [
+            { editorLanguageId: 'java', expectedFileContextLanguage: 'java', filename: 'test.java' },
+            { editorLanguageId: 'python', expectedFileContextLanguage: 'python', filename: 'test.py' },
+            { editorLanguageId: 'typescriptreact', expectedFileContextLanguage: 'tsx', filename: 'test.tsx' },
+            { editorLanguageId: 'typescript', expectedFileContextLanguage: 'typescript', filename: 'test.ts' },
+            { editorLanguageId: 'javascriptreact', expectedFileContextLanguage: 'jsx', filename: 'test.jsx' },
+            { editorLanguageId: 'shellscript', expectedFileContextLanguage: 'shell', filename: 'test.sh' },
+        ]
+
+        it('shoudl return correct filecontext and language should be normalized', function () {
```
justinmk3, line 48:
```suggestion
        it('returns correct filecontext and language should be normalized', function () {
```

justinmk3 (approved):

Will-ShaoHua:
will wait until post re:invent

src/codewhisperer/service/classifierTrigger.ts
Will-ShaoHua: Zoe I reverted this change for now to make this PR only change 1 aspect FileContext for easier review, let's follow up on the remaining part step by step

Will-ShaoHua:
close it for now, will re-open if needed.

PR #4042, opened 2023-11-13
Auth: Initial add connection page redesign

nkomonen-amazon:
### Changes
- Redesigns the add connection page (lots of changes in `src/auth/ui/vue`
  - The main change is that we show all 3 "Feature Panels" at once instead of needing to click boxes to show the individual ones
- Does refactors around the rest of the code base but no functionality changes.
- An attempt to make the UI responsive to changes in webview dimensions and pixel ration

#### Notes
- Alignment is wonky but will be fixed in a future change
- Design is still not final but working toward it

Screenshot 2023-11-15 at 11 02 19 PM

## License


src/codecatalyst/auth.ts
```diff
@@ -66,7 +66,9 @@ export class CodeCatalystAuthenticationProvider {
         })
 
         // set initial context in case event does not trigger
-        setCodeCatalystConnectedContext(this.isConnectionValid())
+        this.restore().then(() => {
```
nkomonen-amazon, line 69: The existing CC connection wasn't being loaded in on startup, this restores the connection (if it exists) when the instance is created.

src/codewhisperer/activation.ts
```diff
@@ -94,7 +93,6 @@ export async function activate(context: ExtContext): Promise<void> {
 
     context.extensionContext.subscriptions.push(
         signoutCodeWhisperer.register(auth),
-        showManageCwConnections.register(),
```
nkomonen-amazon, line 97: This command called the main `showManageConnections()` underneath but with some pre-set CW args. The issue was that this causes 2 different commands for telemetry instead of having 1. So we will use the original `showManageConnections()` instead.

src/auth/ui/vue/show.ts
```diff
@@ -40,25 +33,33 @@ import {
...
+import { throttle } from 'lodash'
+import { submitFeedback } from '../../../feedback/vue/submitFeedback'
 
 export class AuthWebview extends VueWebview {
     public override id: string = 'authWebview'
     public override source: string = 'src/auth/ui/vue/index.js'
-    public readonly onDidConnectionUpdate = new vscode.EventEmitter<undefined>()
+    public readonly onDidConnectionChangeCodeCatalyst = new vscode.EventEmitter<void>()
+    public readonly onDidConnectionChangeExplorer = new vscode.EventEmitter<void>()
+    public readonly onDidConnectionChangeCodeWhisperer = new vscode.EventEmitter<void>()
```
justinmk3, line 62: This looks less ergonomic than the old single emitter. Why can't the handlers decide based on an event arg like `type='codecatalyst'`? When we spread things out it makes the code harder to follow.

src/auth/ui/vue/show.ts
```diff
@@ -797,11 +726,77 @@ export async function showAuthWebview(
...
+        let result: Result
+
+        if (numConnectionsAdded > 0) {
+            result = 'Succeeded'
+        } else if (authWebview.getTotalAuthAttempts() > 0) {
+            // There were no new auth connections added, but attempts were made
+            result = 'Failed'
+        } else {
+            // No new auth connections added, but no attempts were made
+            result = 'Cancelled'
```
justinmk3, line 785: Can we use the run() pattern instead, and throw a "canceled" error to represent cancel, as usual?

justinmk3 (approved):

src/auth/ui/vue/show.ts
nkomonen-amazon: I had that way initially, but with the current way webviews are setup it caused the event to be emitted 3 times instead of 1. I'll look in to a better solution once I get time.

Will-ShaoHua (approved):

PR #4046, opened 2023-11-15
WIP: codewhisperer telemetry test

Will-ShaoHua:
## License

### DIFF
--- a/src/test/codewhisperer/service/telemetry.test.ts
+++ b/src/test/codewhisperer/service/telemetry.test.ts
@@ -1,0 +1,186 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'assert'
+import * as vscode from 'vscode'
+import * as sinon from 'sinon'
+import { assertTelemetryCurried, createTestWorkspaceFolder, openATextEditorWithText } from '../../testUtil'
+import {
+    DefaultCodeWhispererClient,
+    ListRecommendationsRequest,
+    ListRecommendationsResponse,
+} from '../../../codewhisperer/client/codewhisperer'
+import { invokeRecommendation } from '../../../codewhisperer/commands/invokeRecommendation'
+import { ConfigurationEntry } from '../../../codewhisperer/models/model'
+import { waitUntil } from '../../../shared/utilities/timeoutUtils'
+import { RecommendationHandler } from '../../../codewhisperer/service/recommendationHandler'
+import { resetCodeWhispererGlobalVariables } from '../testUtil'
+import { session } from '../../../codewhisperer/util/codeWhispererSession'
+
+type CodeWhispererResponse = ListRecommendationsResponse & {
+    $response: { requestId: string; httpResponse: { headers: { [key: string]: string } } }
+}
+
+let tempFolder: string
+
+describe('', async function () {
+    let sandbox: sinon.SinonSandbox
+    let client: DefaultCodeWhispererClient
+
+    let config: ConfigurationEntry = {
+        isShowMethodsEnabled: true,
+        isManualTriggerEnabled: true,
+        isAutomatedTriggerEnabled: true,
+        isSuggestionsWithCodeReferencesEnabled: true,
+    }
+
+    beforeEach(async function () {
+        sandbox = sinon.createSandbox()
+        tempFolder = (await createTestWorkspaceFolder()).uri.fsPath
+        client = mockClient(aResponse())
+        await resetStates()
+    })
+
+    afterEach(function () {
+        sandbox.restore()
+    })
+
+    async function resetStates() {
+        await RecommendationHandler.instance.clearInlineCompletionStates()
+        resetCodeWhispererGlobalVariables()
+    }
+
+    function mockClient(response: CodeWhispererResponse): DefaultCodeWhispererClient {
+        const cwClient = new DefaultCodeWhispererClient()
+        sandbox.stub(cwClient, 'listRecommendations').resolves(response)
+        sandbox.stub(cwClient, 'generateRecommendations').resolves(response)
+
+        return cwClient
+    }
+
+    function assertCleanStates() {
+        assert.strictEqual(session.requestIdList.length, 0)
+        assert.strictEqual(session.recommendations.length, 0)
+        assert.strictEqual(session.completionTypes.size, 0)
+        assert.strictEqual(session.completionTypes.size, 0)
+    }
+
+    async function manualTrigger(
+        editor: vscode.TextEditor,
+        client: DefaultCodeWhispererClient,
+        config: ConfigurationEntry
+    ) {
+        await invokeRecommendation(editor, client, config)
+        // await sleep(1000) // no programmatically way to wait until the inline suggestion UI is shown
+    }
+
+    async function waitUntilSuggestionSeen(index: number) {
+        const state = await waitUntil(
+            async () => {
+                const r = session.getSuggestionState(index)
+                if (r) {
+                    return r
+                }
+            },
+            {
+                interval: 50,
+            }
+        )
+
+        assert.ok(state === 'Showed')
+    }
+
+    function acceptByTab() {
+        return vscode.commands.executeCommand('editor.action.inlineSuggest.commit')
+    }
+
+    async function rejectByEsc() {
+        return vscode.commands.executeCommand('editor.action.inlineSuggest.hide')
+    }
+
+    it('simple accept - tab', async function () {
+        assertCleanStates()
+        const editor = await openATextEditorWithText('', 'test.py')
+
+        await manualTrigger(editor, client, config)
+        await waitUntilSuggestionSeen(0)
+        acceptByTab().then(() => {
+            const assertUserTriggerDecision = assertTelemetryCurried('codewhisperer_userTriggerDecision')
+            assertUserTriggerDecision({
+                codewhispererSessionId: 'sessionId',
+                codewhispererFirstRequestId: 'requestId',
+                codewhispererLanguage: 'python',
+                codewhispererTriggerType: 'OnDemand',
+                codewhispererLineNumber: 0,
+                codewhispererCursorOffset: 0,
+                codewhispererSuggestionCount: 1,
+                codewhispererCompletionType: 'Line',
+                codewhispererSuggestionState: 'Accept',
+                codewhispererSuggestionImportCount: 0,
+                codewhispererTypeaheadLength: 0,
+                codewhispererUserGroup: 'Control',
+            })
+        })
+    })
+
+    it('simple reject - esc key', async function () {
+        assertCleanStates()
+        const editor = await openATextEditorWithText('', 'test.py')
+
+        await manualTrigger(editor, client, config)
+        await waitUntilSuggestionSeen(0)
+        await rejectByEsc().then(() => {
+            const s = session
+            console.log()
+            // force codewhisperer to flush telemetry
+            RecommendationHandler.instance.onEditorChange()
+
+            const assertUserTriggerDecision = assertTelemetryCurried('codewhisperer_userTriggerDecision')
+            assertUserTriggerDecision({
+                codewhispererSessionId: 'sessionId',
+                codewhispererFirstRequestId: 'requestId',
+                codewhispererLanguage: 'python',
+                codewhispererTriggerType: 'OnDemand',
+                codewhispererLineNumber: 0,
+                codewhispererCursorOffset: 0,
+                codewhispererSuggestionCount: 1,
+                codewhispererCompletionType: 'Line',
+                codewhispererSuggestionState: 'Reject',
+                codewhispererSuggestionImportCount: 0,
+                codewhispererTypeaheadLength: 0,
+                codewhispererUserGroup: 'Control',
+            })
+        })
+    })
+})
+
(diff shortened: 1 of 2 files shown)


src/test/codewhisperer/service/telemetry.test.ts
```diff
@@ -0,0 +1,655 @@
...
+
+    describe('tab and esc', function () {
+        it('single accept', async function () {
+            assertSessionClean()
+            const editor = await openATextEditorWithText('', 'test.py')
+
+            await manualTrigger(editor, client, config)
+            await waitUntilSuggestionSeen()
+            await acceptByTab()
+            // assert.strictEqual(editor.document.getText(), 'Foo')
```
Will-ShaoHua, line 171: comment out these assertions because there is latency between the editor actually updates the document content. and the latency depends on OS and since the purpose of our tests focus on the telemetry, we don't really need these assertions.

Will-ShaoHua:
might be worth checking with Justin and Nick if it's better to introduce `Robotjs` as dev dependency to simulate keyboard inputs of `tab` `esc` and other keys.

src/test/codewhisperer/service/telemetry.test.ts
```diff
@@ -0,0 +1,656 @@
...
+    return vscode.commands.executeCommand('workbench.action.closeActiveEditor')
+}
+
+async function typing(editor: vscode.TextEditor, s: string) {
+    for (const char of s) {
+        await typeAChar(editor, char)
+    }
+    // waitUntilSuggestionSeen doesn't work for this
+    // see if we can find another way to wait
+    await sleep(100)
```
Will-ShaoHua, line 598: have tried different combination of `waitUntil`, but none of them really worked. If we don't wait some time after typing them directly call `commit` command, there will be no effect because VSCode is still resolving the typeahead and trying to provide the adjusted inline suggestion.

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -247,7 +247,7 @@ export class TelemetryHelper {
         }
         const aggregated: CodewhispererUserTriggerDecision = {
             codewhispererSessionId: sessionId,
-            codewhispererFirstRequestId: this.sessionInvocations[0].codewhispererRequestId ?? requestId,
+            codewhispererFirstRequestId: requestId, // suggested change, not required, otherwise we have to 100% sure this.serviceInvocation is correct
```
Will-ShaoHua, line 250: `requestId` is already from `session.requestIdList[0]`

src/test/codewhisperer/service/telemetry.test.ts
```diff
@@ -0,0 +1,657 @@
...
+import { CodewhispererUserTriggerDecision } from '../../../shared/telemetry/telemetry.gen'
+import { getMinVscodeVersion } from '../../../shared/vscode/env'
+
+type CodeWhispererResponse = ListRecommendationsResponse & {
+    $response: { requestId: string; httpResponse: { headers: { [key: string]: string } } }
+}
+
+let tempFolder: string
+
+describe('', async function () {
```
Will-ShaoHua, line 35: should add a title to this

src/test/codewhisperer/service/telemetry.test.ts
Will-ShaoHua: maybe I will just remove these assertions

src/test/codewhisperer/service/telemetry.test.ts
```diff
@@ -0,0 +1,657 @@
...
+            assertTelemetry('codewhisperer_userTriggerDecision', [
+                session1UserTriggerEvent({ codewhispererSuggestionState: 'Reject' }),
+            ])
+        })
+
+        it('reject: trigger then close editor', async function () {
+            assertSessionClean()
+            const editor = await openATextEditorWithText('', 'test.py')
+
+            await manualTrigger(editor, client, config)
```
zixlin7, line 518: nit: seems like all tests call `manualTrigger` -> `waitUntilSuggestionSeen` -> do something, this can probably be extract to another function?

src/test/codewhisperer/service/telemetry.test.ts
```diff
@@ -0,0 +1,657 @@
...
+            await waitUntilSuggestionSeen()
+            // assert.strictEqual(editor.document.getText(), '')
+
+            await RecommendationHandler.instance.onFocusChange()
+            assertTelemetry('codewhisperer_userTriggerDecision', [
+                session1UserTriggerEvent({ codewhispererSuggestionState: 'Reject' }),
```
zixlin7, line 538: most of tests are only testing sessionId, requestId, and decision, can we cover more fields from the userTriggerDecision event? not a blocker for this PR though, thanks for adding these!

zixlin7 (approved):

src/test/codewhisperer/service/telemetry.test.ts
justinmk3: waitUntil has an `interval` option to avoid a "hot loop": https://github.com/aws/aws-toolkit-vscode-staging/blob/a19a4aaf350ac18ddda0e1a0998afd9f865b0b2c/src/shared/utilities/timeoutUtils.ts#L177-L178

does that help?

src/test/codewhisperer/service/telemetry.test.ts
justinmk3: `assertTextEditorContains` uses waitUntil, doesn't it work? try setting `interval` to a higher number so that waitUntil allows node to do other things. https://github.com/aws/aws-toolkit-vscode-staging/blob/a19a4aaf350ac18ddda0e1a0998afd9f865b0b2c/src/test/testUtil.ts#L264-L271

src/test/codewhisperer/service/telemetry.test.ts
Will-ShaoHua: the problem here is more like we don't have a exact target to wait until since we don't have an API like `isInlineSuggestionVisible` or something.  And if I call `editor.action.inlineSuggestion.commit` before it's ready, there will be nothing happen..

src/test/codewhisperer/service/telemetry.test.ts
Will-ShaoHua: I see, I haven't tried it with a high number, let me try with that

src/test/codewhisperer/service/telemetry.test.ts
Will-ShaoHua: let me try with higher interval value

src/test/codewhisperer/service/telemetry.test.ts
Will-ShaoHua: yea, it's intentional, I was thinking we might need test cases for discard later on.

src/test/codewhisperer/service/telemetry.test.ts
Will-ShaoHua: yea that would be incremental progress, we now only add tests for userTriggerDecision which is considered as p0

src/test/codewhisperer/service/telemetry.test.ts
Will-ShaoHua: but actually let me do this for now, for discard case we can call them seprately

src/test/codewhisperer/service/telemetry.test.ts
Will-ShaoHua: I think I will simply remove these assertions of document content, it's still failing on some machines like mac and windows but succeeded on linux

andrewyuq (approved):

src/test/codewhisperer/service/telemetry.test.ts
```diff
@@ -0,0 +1,613 @@
...
+async function typing(editor: vscode.TextEditor, s: string) {
+    for (const char of s) {
+        await typeAChar(editor, char)
+    }
+    // waitUntilSuggestionSeen doesn't work for this
+    // see if we can find another way to wait
+    await sleep(100)
+}
+
+async function backsapce(editor: vscode.TextEditor) {
```
viktorsaws, line 569: nit: typo

src/test/codewhisperer/service/telemetry.test.ts
Will-ShaoHua: ahh I slightly update the function `acceptByTab()` here https://github.com/aws/aws-toolkit-vscode/pull/4046/files#diff-2dc09778984ac96fa6f8170184bf439825bf4839775915c3c8e5da4d2cc47c25R546

Will-ShaoHua: `assertEditorContains` works perfectly, the only thing left is the waiting until inline suggestion is shown then accept

Will-ShaoHua: so move waiting logic inside `acceptByTab()`, but I felt I can't avoid the very first line of `await sleep(x)` in this `waitUntil` otherwise it just won't work as expected because we don't have a `isInlineSuggestionActive` API to waitUntil

justinmk3 (approved):

Will-ShaoHua:
@justinmk3 this one can wait until post re:invent I think. To reduce the noise

Will-ShaoHua:
also i don't really know why windows CI failed 
```
47%Downloading VS Code [===================-----------] 65%Downloading VS Code [====================----------] 70%Downloading VS Code [=====================---------] 72%Downloading VS Code [======================--------] 75%Downloading VS Code [=======================-------] 79%Downloading VS Code [========================------] 82%Downloading VS Code [==========================----] 88%Downloading VS Code [===========================---] 92%Downloading VS Code [==============================] 100%Downloaded VS Code into D:\a\aws-toolkit-vscode\aws-toolkit-vscode\.vscode-test\vscode-win32-x64-archive-insiders
Downloaded VS Code into D:\a\aws-toolkit-vscode\aws-toolkit-vscode\.vscode-test\vscode-win32-x64-archive-insiders
VS Code test instance location: D:\a\aws-toolkit-vscode\aws-toolkit-vscode\.vscode-test\vscode-win32-x64-archive-insiders\Code - Insiders.exe
Invoking vscode CLI command:
    "D:\a\aws-toolkit-vscode\aws-toolkit-vscode\.vscode-test\vscode-win32-x64-archive-insiders\bin\code-insiders.cmd" ["--extensions-dir=D:\\a\\aws-toolkit-vscode\\aws-toolkit-vscode\\.vscode-test\\extensions","--user-data-dir=D:\\a\\aws-toolkit-vscode\\aws-toolkit-vscode\\.vscode-test\\user-data","--version"]
output: null
Error: VS Code CLI command failed (exit-code: null): D:\a\aws-toolkit-vscode\aws-toolkit-vscode\.vscode-test\vscode-win32-x64-archive-insiders\bin\code-insiders.cmd --extensions-dir=D:\a\aws-toolkit-vscode\aws-toolkit-vscode\.vscode-test\extensions,--user-data-dir=D:\a\aws-toolkit-vscode\aws-toolkit-vscode\.vscode-test\user-data,--version
    at invokeVSCodeCli (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\scripts\test\launchTestUtilities.ts:68:15)
    at setupVSCodeTestInstance (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\scripts\test\launchTestUtilities.ts:45:23)
    at async D:\a\aws-toolkit-vscode\aws-toolkit-vscode\scripts\test\test.ts:13:38
Failed to run tests
Error: Process completed with exit code 1.

```

justinmk3:
> also i don't really know why windows CI failed, it doesn't seem like it ever started the tests

fixed in https://github.com/aws/aws-toolkit-vscode/pull/4075

src/test/codewhisperer/service/telemetry.test.ts
```diff
@@ -581,6 +581,7 @@ async function typing(editor: vscode.TextEditor, s: string) {
     })
 
     assertTextEditorContains(initialContent + s)
+    await sleep(50)
     const positionAfter = editor.document.offsetAt(editor.selection.active)
     assert.strictEqual(positionAfter, positionBefore + s.length)
```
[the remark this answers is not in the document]
Will-ShaoHua: umm @justinmk3  I think here I need to waitUntil Inline suggestion is shown, but currently we don't have that API I think. The case where mac OS is failing is suggestion is `foo` and user typing a `f`, but due to some race condition, accept recommendation `foo` immediately after typing `f` will result in `ffoo`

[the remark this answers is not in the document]
Will-ShaoHua: still seeking solutions to it, i know sleep() for an arbitrary number is brittle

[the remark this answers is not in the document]
Will-ShaoHua: let me debug it, I think we're not sending the correct cursor offset, it's supposed to the one where user triggers the service, but seems it's using a state variable in `telemetryHelper`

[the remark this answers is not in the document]
Will-ShaoHua: guess that's the problem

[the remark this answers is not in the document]
justinmk3: > umm @justinmk3 I think here I need to waitUntil Inline suggestion is shown, but currently we don't have that API I think. 

`waitUntil` can wait for _any_ arbitrary condition. You are already checking a condition : 

```
    const positionAfter = editor.document.offsetAt(editor.selection.active)
    assert.strictEqual(positionAfter, positionBefore + s.length)
```

why can't that condition be re-checked in the `waitUntil` ?

PR #4047, opened 2023-11-15
CodeWhisperer: Add feature fetching component

andrewyuq:
1. Add a component to fetch feature assignments every 30 mins, calling ListFeatureEvaluations API and cache values in memory.

2. Add a field in UserTriggerDecision codewhispererFeatureEvaluations to record all the feature configs in a string, which can be queried on Kibana

respective JB PR: https://github.com/aws/aws-toolkit-jetbrains/pull/3978
## License

### DIFF
--- a/src/codewhisperer/service/featureConfigProvider.ts
+++ b/src/codewhisperer/service/featureConfigProvider.ts
@@ -1,0 +1,87 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { FeatureValue } from '../client/codewhispereruserclient'
+import { codeWhispererClient as client } from '../client/codewhisperer'
+import { AuthUtil } from '../util/authUtil'
+import { getLogger } from '../../shared/logger'
+
+export class FeatureContext {
+    constructor(public name: string, public variation: string, public value: FeatureValue) {}
+}
+
+const testFeatureName = 'testFeature'
+const featureConfigPollIntervalInMs = 30 * 60 * 1000 // 30 mins
+
+// TODO: add real feature later
+export const featureDefinitions = new Map([
+    [testFeatureName, new FeatureContext(testFeatureName, 'CONTROL', { stringValue: 'testValue' })],
+])
+
+export class FeatureConfigProvider {
+    private featureConfigs = new Map<string, FeatureContext>()
+
+    static #instance: FeatureConfigProvider
+
+    constructor() {
+        this.fetchFeatureConfigs()
+
+        setInterval(this.fetchFeatureConfigs.bind(this), featureConfigPollIntervalInMs)
+    }
+
+    public static get instance() {
+        return (this.#instance ??= new this())
+    }
+
+    async fetchFeatureConfigs(): Promise<void> {
+        if (AuthUtil.instance.isConnectionExpired()) {
+            return
+        }
+
+        getLogger().debug('Fetching feature configs')
+        try {
+            const response = await client.listFeatureEvaluations()
+
+            // Overwrite feature configs from server response
+            response.featureEvaluations.forEach(evaluation => {
+                this.featureConfigs.set(
+                    evaluation.feature,
+                    new FeatureContext(evaluation.feature, evaluation.variation, evaluation.value)
+                )
+            })
+        } catch (e) {
+            getLogger().debug('Error when fetching feature configs', e)
+        }
+        getLogger().debug(`Current feature configs: ${this.getFeatureConfigsTelemetry()}`)
+    }
+
+    getFeatureConfigsTelemetry(): string {
+        return `{${Array.from(this.featureConfigs.entries())
+            .map(([name, context]) => `${name}: ${context.variation}`)
+            .join(', ')}}`
+    }
+
+    // TODO: for all feature variations, define a contract that can be enforced upon the implementation of
+    // the business logic.
+    // When we align on a new feature config, client-side will implement specific business logic to utilize
+    // these values by:
+    // 1) Add an entry in featureDefinitions, which is <feature_name> to <feature_context>.
+    // 2) Add a function with name `getXXX`, where XXX refers to the feature name.
+    // 3) Specify the return type: One of the return type string/boolean/Long/Double should be used here.
+    // 4) Specify the key for the `getFeatureValueForKey` helper function which is the feature name.
+    // 5) Specify the corresponding type value getter for the `FeatureValue` class. For example,
+    // if the return type is Long, then the corresponding type value getter is `longValue()`.
+    // 6) Add a test case for this feature.
+    // 7) In case `getXXX()` returns undefined, it should be treated as a default/control group.
+    getTestFeature(): string | undefined {
+        return this.getFeatureValueForKey(testFeatureName).stringValue
+    }
+
+    // Get the feature value for the given key.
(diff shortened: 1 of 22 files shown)


src/codewhisperer/service/featureConfigProvider.ts
```diff
@@ -0,0 +1,87 @@
...
+    // these values by:
+    // 1) Add an entry in featureDefinitions, which is <feature_name> to <feature_context>.
+    // 2) Add a function with name `getXXX`, where XXX refers to the feature name.
+    // 3) Specify the return type: One of the return type string/boolean/Long/Double should be used here.
+    // 4) Specify the key for the `getFeatureValueForKey` helper function which is the feature name.
+    // 5) Specify the corresponding type value getter for the `FeatureValue` class. For example,
+    // if the return type is Long, then the corresponding type value getter is `longValue()`.
+    // 6) Add a test case for this feature.
+    // 7) In case `getXXX()` returns undefined, it should be treated as a default/control group.
+    getTestFeature(): string | undefined {
```
Will-ShaoHua, line 78: we should make it generic instead of adding a getter for every new feature.
maybe like
```
getTestFeature<T>(FeatureContext['name']): T extends FeatureContext
```

src/codewhisperer/util/authUtil.ts
```diff
@@ -104,6 +104,9 @@ export class AuthUtil {
                     vscode.commands.executeCommand('aws.codeWhisperer.gettingStarted')
                     prompts.disablePrompt('codeWhispererNewWelcomeMessage')
                 }
+
+                // start the feature config polling job
+                vscode.commands.executeCommand('aws.codewhisperer.fetchFeatureConfigs')
```
Will-ShaoHua, line 109: if this is should be executed on IDE start, are we able to move it to `activation.ts`?

src/codewhisperer/util/codeWhispererSession.ts
```diff
@@ -70,6 +76,19 @@ class CodeWhispererSession {
...
+        this.sessionId = ''
+        this.requestIdList = []
+        this.startPos = new Position(0, 0)
+        this.startCursorOffset = 0
+        this.leftContextOfCurrentLine = ''
+        this.language = 'python'
+        this.triggerType = 'OnDemand'
+        this.recommendations = []
+        this.suggestionStates.clear()
+        this.completionTypes.clear()
```
Will-ShaoHua, line 90: yea this is easier to read and understand instead of setting different  values across different places. 
for completeness, let's reset `requestContext` as well even though it shouldn't make difference.

src/codewhisperer/service/featureConfigProvider.ts
```diff
@@ -0,0 +1,87 @@
...
+            response.featureEvaluations.forEach(evaluation => {
+                this.featureConfigs.set(
+                    evaluation.feature,
+                    new FeatureContext(evaluation.feature, evaluation.variation, evaluation.value)
+                )
+            })
+        } catch (e) {
+            getLogger().debug('Error when fetching feature configs', e)
+        }
+        getLogger().debug(`Current feature configs: ${this.getFeatureConfigsTelemetry()}`)
```
Will-ShaoHua, line 57: nit: prefix with `CodeWhisperer: `

src/codewhisperer/service/featureConfigProvider.ts
```diff
@@ -0,0 +1,87 @@
...
+    public static get instance() {
+        return (this.#instance ??= new this())
+    }
+
+    async fetchFeatureConfigs(): Promise<void> {
+        if (AuthUtil.instance.isConnectionExpired()) {
+            return
+        }
+
+        getLogger().debug('Fetching feature configs')
```
Will-ShaoHua, line 43: nit: prefix with CodeWhisperer

src/codewhisperer/service/featureConfigProvider.ts
```diff
@@ -0,0 +1,87 @@
...
+            getLogger().debug('Error when fetching feature configs', e)
+        }
+        getLogger().debug(`Current feature configs: ${this.getFeatureConfigsTelemetry()}`)
+    }
+
+    getFeatureConfigsTelemetry(): string {
+        return `{${Array.from(this.featureConfigs.entries())
+            .map(([name, context]) => `${name}: ${context.variation}`)
+            .join(', ')}}`
+    }
```
Will-ShaoHua, line 64: nit: add a test for this, it would be easier for ppl to generate a string to look up on ES

src/codewhisperer/service/featureConfigProvider.ts
```diff
@@ -0,0 +1,87 @@
...
+    // 6) Add a test case for this feature.
+    // 7) In case `getXXX()` returns undefined, it should be treated as a default/control group.
+    getTestFeature(): string | undefined {
+        return this.getFeatureValueForKey(testFeatureName).stringValue
+    }
+
+    // Get the feature value for the given key.
+    // In case of a misconfiguration, it will return a default feature value of Boolean true.
+    private getFeatureValueForKey(name: string): FeatureValue {
+        return this.featureConfigs.get(name)?.value ?? featureDefinitions.get(name)?.value ?? { boolValue: true }
```
Will-ShaoHua, line 85: seems like 
```
type FeatureValue: {
   stringValue?: string
   boolValue?: boolean
   doulbeValue?: boolean
   .....
}
```
(I just check the sdk defnition https://github.com/aws/aws-toolkit-vscode/pull/3989/files#diff-2bd14827c41c712bcaef58803209651343cc08861887376fe80f1285e1b1c5d9R328,  not sure about the exact interface)

a qq, what if the other value is read, for example `stringValue`? so here we just let it to be `undefined` correct?

src/codewhisperer/util/authUtil.ts
andrewyuq: I thought about this but meanwhile wee need to get CodeWhisperer logged in first so we can invoke the api. If we put it in `activation.ts`, CodeWhisperer may not be logged in at that time and we will miss reading this. That's why I chose to do it every time after we get a valid connection.

src/codewhisperer/util/codeWhispererSession.ts
andrewyuq: Will add,
currently this is only used in test, later on we can call this one when the last session finishes, if session is still a global state like this.

src/codewhisperer/service/featureConfigProvider.ts
andrewyuq: It will be undefined.
eventually, the `getXXX()`  for each feature will always return a nullable value, which the caller is responsible for handling.

src/codewhisperer/service/featureConfigProvider.ts
andrewyuq: For me, having a different getter for each feature is slightly easier for the caller and is more controlled. Making it generic makes the caller can pass any parameters, both expected and unexpected. Plus having separate getters can also allow us to define specific return type for the specific feature, for which the contract will be very clear(thus easier to debug in general)(otherwise the caller will need to get either `stringValue` or `boolValue` which is duplicated code and hard to manage). All at the cost of adding a new getter, which I don't think it's a big deal. Anyway the feature usage is a case by case thing.

src/codewhisperer/util/authUtil.ts
Will-ShaoHua: sgtm

Will-ShaoHua (approved):

PR #4048, opened 2023-11-15
fix: Command arg fix

nkomonen-amazon:
Fixes the issue with the args of an executed command being incorrect to the actual function signature.
This is due to the command being either incorrectly called, or that the command is called through a VS Code UI component which gives its own args (we have no control over this).

## License


bryceitoc9 (approved):

src/auth/activation.ts
```diff
@@ -57,5 +53,5 @@ async function showManageConnectionsOnStartup() {
     }
 
     // Show connection management to user
-    AuthCommandDeclarations.instance.declared.showManageConnections.execute('firstStartup')
+    showManageConnections.execute(placeholder, 'firstStartup')
```
justinmk3, line 56: why can't we just pass undefined?

src/auth/ui/vue/show.ts
```diff
@@ -712,20 +721,41 @@ const Panel = VueWebview.compilePanel(AuthWebview)
...
+        // The auth webview page does not make sense to use in C9,
+        // so show the auth quick pick instead.
+        if (isCloud9('any')) {
+            return addConnection.execute()
+        }
 
-export function isAuthSource(obj: any): obj is AuthSource {
-    return Object.values(AuthSources).includes(obj)
-} 
+        if (_ !== placeholder) {
```
justinmk3, line 749: if we are actually using the parameter we should give it a name. `_` usually mean "ignore"

src/shared/vscode/commands2.ts
```diff
@@ -18,6 +18,34 @@ import { keysAsInt } from '../utilities/tsUtils'
...
 
+/**
+ * HACK
+ *
+ * If a command matches the following conditions:
+ * - Is registered with VS Code
+ * - Has args for the execution of the command
+ * - Can be executed through a package.json contribution (eg: ellipsis menu in View)
+ *
+ * Then unexpected args will be passed to the command.
```
justinmk3, line 29: It's not really unexpected. Any command called from a treeview can get a treenode, etc. The commands simply need to be defined with parameters that express that :)

src/auth/activation.ts
nkomonen-amazon: Found out the hard way that in this edge case vscode can either pass an object (tree node command) OR undefined (ellipsis menu item)

src/shared/vscode/commands2.ts
```diff
@@ -18,6 +18,34 @@ import { keysAsInt } from '../utilities/tsUtils'
...
+ * - Currently it sets arg[0] to an object or undefined, depending on the context.
+ * - And the rest of the args will be undefined do not exist.
+ * ---
+ * **So as a workaround**, commands who meet the above criteria should
+ * - have this type as their first arg as a placeholder
```
justinmk3, line 34: Pretty much any command we plan to hook up to a treeview, etc.

src/shared/vscode/commands2.ts
```diff
@@ -18,6 +18,34 @@ import { keysAsInt } from '../utilities/tsUtils'
...
+ */
+export type VsCodeCommandArg = typeof placeholder
+/** A placeholder value for Commands that use {@link VsCodeCommandArg} */
+export const placeholder = 'placeholder**' as const
+
+/**
+ * The command was executed by a VSCode UI component which we
+ * cannot accurately determine a source.
+ */
+export const vscodeComponent = 'vscodeComponent'
```
justinmk3, line 47: When a TreeNode is passed we can guess the source in some cases.

src/auth/ui/vue/show.ts
nkomonen-amazon: This handles the edge case which the docstring on the type `VsCodeCommandArg` should hopefully explain.

Did you have a name in mind?

src/shared/vscode/commands2.ts
```diff
@@ -423,7 +452,7 @@ function getInstrumenter(
     if (!telemetryName && info?.startTime !== undefined && currentTime - info.startTime < threshold) {
         info.debounceCount += 1
         TelemetryDebounceInfo.instance.set(id, info)
-        getLogger().debug(`commands: skipped telemetry for "${id.id}" with key "${id.compositeKey}"`)
+        getLogger().debug(`commands: skipped telemetry for "${id.id}" with key "${JSON.stringify(id.compositeKey)}"`)
```
justinmk3, line 455: `%O` is much more readable in the logs (unless you need to see deeply nested objects)
```suggestion
        getLogger().debug('commands: skipped telemetry for "%s" with key: %O', id.id, id.compositeKey)
```

src/shared/vscode/commands2.ts
```diff
@@ -447,18 +476,47 @@ function getInstrumenter(
...
+        if (field === 'source' && typeof arg !== 'string') {
+            /**
+             * This case happens when either the caller sets the wrong args themselves through
+             * vscode.commands.executeCommand OR if through a VS Code UI component like the ellipsis menu
+             *
+             * TODO: To fix this we insert a value in the `source` arg so that there is something there
+             * for the metric to emit. We need to figure out a better way to handle either incorrectly called
+             * or the VS Code UI component.
+             */
+            getLogger().error(`Commands/Telemetry: "${id}" executed with invalid "source" type: "${args}"`)
```
justinmk3, line 502:
```suggestion
            getLogger().error(`Commands/Telemetry: "${id}" executed with invalid "source" type: "${args}"`)
```
```suggestion
            getLogger().error(`Commands/Telemetry: "%s" executed with invalid "source" type: %O"`, id, args)
```

justinmk3 (approved):

src/auth/activation.ts
justinmk3: > undefined is a false positive that we didn't run in to this edge case.

why does that matter?

src/auth/ui/vue/show.ts
justinmk3: `node` is what we usually call it

PR #4049, opened 2023-11-15
CodeWhisperer: fix various VSC bugs

andrewyuq:
1. Fixed an issue where clientCompoentLatency telemetry event would be sent every time when navigating the suggestions, where it should only be sent once when a first suggestion has been shown.
2. Fixed an issue where the requestIdList is not reset, causing a later check to not pass and stop sending the userTriggerDecisions.
3. Fixed an issue where spamming trigger will cause the invocation start time to reset to current time for each trigger action. This will cause the e2e latency to be recorded unexpectedly.

## License

### DIFF
--- a/src/codewhisperer/service/recommendationHandler.ts
+++ b/src/codewhisperer/service/recommendationHandler.ts
@@ -181,6 +181,7 @@ export class RecommendationHandler {
         let shouldRecordServiceInvocation = true
         session.language = runtimeLanguageContext.getLanguageContext(editor.document.languageId).language
         session.taskType = await this.getTaskTypeFromEditorFileName(editor.document.fileName)
+        session.requestIdList = []
 
         if (pagination) {
             if (page === 0) {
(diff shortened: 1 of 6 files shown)


src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -181,6 +181,7 @@ export class RecommendationHandler {
         let shouldRecordServiceInvocation = true
         session.language = runtimeLanguageContext.getLanguageContext(editor.document.languageId).language
         session.taskType = await this.getTaskTypeFromEditorFileName(editor.document.fileName)
+        session.requestIdList = []
```
Will-ShaoHua, line 184: don't think this place is correct. This method will be called multiple times with paginated call. If the requestIdList field serves to store the request ids within a session, it will reset every API call.

Will-ShaoHua: should move to clearRecommendation() instead.

andrewyuq: emm yes, how about on like 212?

andrewyuq: actually clearRecommendation() is better.

justinmk3 (approved):
CI is failing though.

Please also check cloud9

Will-ShaoHua (approved):

PR #4062, opened 2023-11-17
fix: racecondition in cw auth change

nkomonen-amazon:
### Problem:

In the Add Connection page we listen to auth
events in the backend. When one of these events
is fired the Add Connection page eventually checks if isEnterpriseSsoInUse. The issue is that this
uses a variable that is separately updated by
the event onDidChangeActiveConnection.

There is a race condition for when onDidChangeActiveConnection updates the variable used in isEnterpriseSsoInUse() versus when isEnterpriseSsoInUse() is actually called.

Also when SecondaryAuth.deleteConnection() is called the new state is not updated when the function is complete since we are relying on an event listener to do the final updating.

### Solution:

Get rid of the variable and whenever isEnterpriseSsoInUse() is called it will pull the latest state. Now we don't have to manage it.

Also in deleteConnection() do a final clearing of the saved connection state so that upon completion the user will be able to get the latest state before all the event emitters fire.

## License

### DIFF
--- a/src/codewhisperer/util/authUtil.ts
+++ b/src/codewhisperer/util/authUtil.ts
@@ -45,7 +45,6 @@ export const isValidCodeWhispererConnection = (conn?: Connection): conn is Conne
 export class AuthUtil {
     static #instance: AuthUtil
 
-    private usingEnterpriseSSO: boolean = false
     private reauthenticatePromptShown: boolean = false
     private _isCustomizationFeatureEnabled: boolean = false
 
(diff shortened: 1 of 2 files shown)


src/codewhisperer/util/authUtil.ts
```diff
@@ -79,14 +78,9 @@ export class AuthUtil {
...
-        this.secondaryAuth.onDidChangeActiveConnection(async conn => {
-            if (conn?.type === 'sso') {
-                this.usingEnterpriseSSO = !isBuilderIdConnection(conn)
-                if (!this.isConnectionExpired() && this.usingEnterpriseSSO) {
-                    vscode.commands.executeCommand('aws.codeWhisperer.notifyNewCustomizations')
-                }
-            } else {
-                this.usingEnterpriseSSO = false
+        this.secondaryAuth.onDidChangeActiveConnection(async () => {
+            if (this.isValidEnterpriseSsoInUse()) {
```
nkomonen-amazon, line 82: When `secondaryAuth` emits an event we can assume it already updated all of its state. So if we use `this.conn` it will pull the latest connection from `secondaryAuth`

justinmk3 (approved):

Will-ShaoHua (approved):

Issue #4043, opened 2023-11-14
Disable config validation `isValid()` check


BapRx:
## Problem

A startup check was introduced by #3924 for the issue #3910 which causes errors appearing every time vscode is opened not only if the configuration is valid but also if it's not writable.

As a nix user my configs are read-only on purpose and although the JSON is valid, the error keeps popping up due to insufficient permissions:
```log
2023-11-14 09:28:19 [WARN]: settings: invalid "settings.json" file? failed to update "aws.samcli.lambdaTimeout": Unable to write into user settings because the file has unsaved changes. Please save the user settings file first and then try again.
```

## Expected behavior

No error should appear either by only checking the file structure/syntax or not failing on write permission error.

I guess it would be difficult/impossible to add a setting to disable the setting check but if I'm wrong it could be a 3rd option.

justinmk3:
Interesting, thanks for the report. I did a quick `chmod u-x` on my vscode settings.jsons on macOS, and vscode actually sets it back to writable. I guess that isn't the case on unix?

BapRx:
My settings are managed by [home-manager](https://nixos.wiki/wiki/Home_Manager) which creates a symlink between the actual file (owned by root and meant to be read-only) so the file is not writable because of user permissions.:

```bash
❯ l ~/.config/VSCodium/User/settings.json
lrwxrwxrwx baptiste baptiste 98 B 2 hours ago  /home/baptiste/.config/VSCodium/User/settings.json ⇒ /nix/store/9dg8yz9z8xj3l21kwscasklh97iha9sl-home-manager-files/.config/VSCodium/User/settings.json
❯ l /nix/store/9dg8yz9z8xj3l21kwscasklh97iha9sl-home-manager-files/.config/VSCodium/User/settings.json
lrwxrwxrwx root root 64 B 53 years ago  /nix/store/9dg8yz9z8xj3l21kwscasklh97iha9sl-home-manager-files/.config/VSCodium/User/settings.json ⇒ /nix/store/la0fhvz5lrljypc7ypmi2xj4hnx9mjwg-vscode-user-settings
❯ l /nix/store/la0fhvz5lrljypc7ypmi2xj4hnx9mjwg-vscode-user-settings
.r--r--r-- root root 4.9 KB 53 years ago  /nix/store/la0fhvz5lrljypc7ypmi2xj4hnx9mjwg-vscode-user-settings
```

justinmk3:
[...]

PR #4064, opened 2023-11-18
fix(settings): allow user to skip isValid startup check

BapRx:
## Problem

#4043

## Solution

Add a user setting to skip this check

## License


justinmk3:
See https://github.com/aws/aws-toolkit-vscode/issues/4043#issuecomment-1823008719 

Can you explain the use case? Are you expecting vscode extensions to never update vscode settings?

justinmk3:
I assume this is no longer needed since https://github.com/aws/aws-toolkit-vscode/pull/4474

PR #4072, opened 2023-11-21
fix: update e2e tests

hayemaxi:
## Problem
E2E tests fail when not ran locally.

## Solution
- Update SSO login dialog regex to match latest.
- Update test util to accept 'test' propert of Mocha Context.
- Update data sent to lambda authorizer to match new SSO url dialog.

## License

### DIFF
--- a/src/test/setupUtil.ts
+++ b/src/test/setupUtil.ts
@@ -60,9 +60,9 @@ export function skipTest(testOrCtx: Mocha.Context | Mocha.Test | undefined, reas
     let test
 
     if (testOrCtx?.type === 'test') {
-        test = (testOrCtx as Mocha.Test)
+        test = testOrCtx as Mocha.Test
     } else {
-        const context = (testOrCtx as Mocha.Context | undefined)
+        const context = testOrCtx as Mocha.Context | undefined
         test = context?.currentTest ?? context?.test
     }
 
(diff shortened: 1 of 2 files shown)


src/testE2E/codecatalyst/client.test.ts
```diff
@@ -86,7 +86,7 @@ let projectName: CodeCatalystProject['name']
  *     integ tests, but using the ssh hostname that we get from
  *     {@link prepareDevEnvConnection}.
  */
-describe.skip('Test how this codebase uses the CodeCatalyst API', function () {
+describe('Test how this codebase uses the CodeCatalyst API', function () {
```
justinmk3, line 89: if the throttling error still continues we may need to keep this skip()

hayemaxi: AFAIK this only runs with the `/runIntegrationTests` command right? And they aren't required to merge. Does it matter if it stays skipped until throttling is addressed?

justinmk3: > AFAIK this only runs with the `/runIntegrationTests` command right? 

After a PR is merged, integ and e2e tests are run without `/runIntegrationTests`.

> And they aren't required to merge.

Yes, but a consistently red CI is not useful. Presumably when one runs `/runIntegrationTests` they want to see something that gives a useful signal of whether things are actually broken :)

hayemaxi: makes sense, will update

src/test/setupUtil.ts
```diff
@@ -56,8 +57,14 @@ export function setRunnableTimeout(test: Mocha.Runnable, maxTestDuration: number
...
 export function skipTest(testOrCtx: Mocha.Context | Mocha.Test | undefined, reason?: string) {
-    const test =
-        testOrCtx?.type === 'test' ? (testOrCtx as Mocha.Test) : (testOrCtx as Mocha.Context | undefined)?.currentTest
+    let test
+
+    if (testOrCtx?.type === 'test') {
+        test = testOrCtx as Mocha.Test
+    } else {
+        const context = testOrCtx as Mocha.Context | undefined
+        test = context?.currentTest ?? context?.test
```
justinmk3, line 66: The ternary was nice because it allows `const`, and type inference. But it's probably too messy to add a nested ternary.

justinmk3 (approved):

PR #4074, opened 2023-11-21
codewhisperer: do not skip closing paren handling at any condition

Will-ShaoHua:
In most of auto-suggestion scenario, the handling logic will be skipped.

## License

### DIFF
--- a/src/codewhisperer/commands/onInlineAcceptance.ts
+++ b/src/codewhisperer/commands/onInlineAcceptance.ts
@@ -82,10 +82,7 @@ export async function onInlineAcceptance(
          * Mitigation to right context handling mainly for auto closing bracket use case
          */
         try {
-            // Do not handle extra bracket if there is a right context merge
-            if (acceptanceEntry.recommendation === session.recommendations[acceptanceEntry.acceptIndex].content) {
-                await handleExtraBrackets(acceptanceEntry.editor, acceptanceEntry.recommendation, end, start)
-            }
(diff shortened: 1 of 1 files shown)


justinmk3:
Is this still needed?

Will-ShaoHua:
Yea it's needed, but I just need another round of testing

src/codewhisperer/commands/onInlineAcceptance.ts
```diff
@@ -82,10 +82,7 @@ export async function onInlineAcceptance(
          * Mitigation to right context handling mainly for auto closing bracket use case
          */
         try {
-            // Do not handle extra bracket if there is a right context merge
-            if (acceptanceEntry.recommendation === session.recommendations[acceptanceEntry.acceptIndex].content) {
```
leigaol, line 86: // Do not handle extra bracket if and only if there is a right context merge.

DO handle extra bracket if 
1. There is no right context merge, accepted recommendation is the same from service response.
2. There is no right context merge, accepted recommendation has some typeahead

leigaol: For 1, use condition `acceptanceEntry.recommendation === session.recommendations[acceptanceEntry.acceptIndex].content`

For 2, use session.recommendations[acceptanceEntry.acceptIndex].content.substring(0, typehead.length) === acceptanceEntry.recommendation`

PR #4077, opened 2023-11-22
feat(codecatalyst): Use MDE endpoint set by environment variable

ctidd:
## Problem

MDE requires use of different endpoint in some VPC configurations.

## Solution

Allow MDE to pass different endpoint as environment variable. (Default to current behavior.)

## License

### DIFF
--- a/.changes/next-release/Feature-666cb5e9-eff1-41d9-bcf8-1d65cda2dd2c.json
+++ b/.changes/next-release/Feature-666cb5e9-eff1-41d9-bcf8-1d65cda2dd2c.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Use MDE endpoint set by environment variable"
+}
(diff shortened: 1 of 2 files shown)


.changes/next-release/Feature-666cb5e9-eff1-41d9-bcf8-1d65cda2dd2c.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Use MDE endpoint set by environment variable"
```
justinmk3, line 3: Thanks for including a changelog, but in this case it's not really customer-relevant so we should probably omit it.

src/shared/clients/devenvClient.ts
```diff
@@ -11,7 +11,7 @@ import { getCodeCatalystDevEnvId } from '../vscode/env'
 import { ExtensionUserActivity } from '../extensionUtilities'
 
 const environmentAuthToken = '__MDE_ENV_API_AUTHORIZATION_TOKEN'
-const environmentEndpoint = 'http://127.0.0.1:1339'
+const environmentEndpoint = process.env['__MDE_ENVIRONMENT_API'] ?? 'http://127.0.0.1:1339'
```
justinmk3, line 14: Ideally we would do some sanity checking (and log if necessary). MDE has had regressions where env vars were invalid or went missing.

justinmk3 (approved):

PR #4090, opened 2023-11-28
[Post Re-Invent]: Support for JSON with comments for security scans

laileni-aws:
## Problem
Json with comments `jsonc` is not supported for security scans.
## Solution
- Enabling support for `jsonc` for Codewhisperer security scans

- Telemetry languageid  will be same as json: `json`

## License


justinmk3 (approved):
Waiting for cw team approval

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -96,6 +97,7 @@ export class RuntimeLanguageContext {
             java: 'java',
             javascript: 'js',
             json: 'json',
+            jsonc: 'json',
```
andrewyuq, line 100: can you check if this works with STE passing json(or it's variant) as the payload and didn't throw a validationException?

laileni-aws: You mean to write a test case to validate the json or jsonc support to security scans?

andrewyuq: not only for getDependencyGraph, but also the SendTelemetryEvent with codeScanEvent as the payload.

src/test/codewhisperer/service/codewhisperer.test.ts
```diff
@@ -66,6 +66,18 @@ describe('codewhisperer', async function () {
...
+        const payload = {
+            codeScanEvent: {
+                programmingLanguage: { languageName: 'json' },
+                codeScanJobId: anyString(),
+                timestamp: new Date(),
+            },
+        }
+        await sendTelemetryEventOptoutCheckHelper(payload, true, true)
+        await sendTelemetryEventOptoutCheckHelper(payload, true, false)
+    })
```
justinmk3, line 79: do we seriously need 10 lines of test code for each new filetype?

laileni-aws: > not only for getDependencyGraph, but also the SendTelemetryEvent with codeScanEvent as the payload.

@andrewyuq requested to add test case to send json as payload to sendTelemetryEvent.
If thats ok I can drop this test case ?

andrewyuq: The test I am referring to is not related to "telemetry optout status". We don't need to test "telemetry optout status" for different languages. They are independent fields
I was asking for a test that SendTelemetryEvent with codeScanEvent as the payload for the json language, that the request can succeed without any validation error.

laileni-aws: > I was asking for a test that SendTelemetryEvent with codeScanEvent as the payload for the json language, that the request can succeed without any validation error.

If I understand correctly there is only 1 telemetry event for security scans(`codewhisperer_securityScan`) Previously we tested this event by passing python params. So, correct me if you something like [this](https://github.com/aws/aws-toolkit-vscode/blob/master/src/test/codewhisperer/commands/startSecurityScan.test.ts) to test for json too?

```
        assertTelemetry('codewhisperer_securityScan', {
            codewhispererLanguage: 'json',
            codewhispererCodeScanTotalIssues: 1,
            codewhispererCodeScanIssuesWithFixes: 0,
            codewhispererCodeScanLines: 188,
        } as CodewhispererSecurityScan)
```

andrewyuq: A manual test is actually fine.

laileni-aws: Tested manually and this works fine.

justinmk3 (approved):

andrewyuq (approved):

PR #4093, opened 2023-11-28
fix(transform): do validation checks AFTER user picks project

dhasani23:
## Problem

We run a shell command (`javap`) to determine the Java version of each of the projects open in the user's workspace. Running this command for each open project is totally unnecessary.

## Solution

This fix shows the user as options to transform any of the projects open in their workspace, and then once they have selected a project, the validation checks are run (`javap` etc.) to check that it is a valid project. The solution is mostly a refactoring  -- the logic to do the validation checks was pre-existing, it is simply moved to happen after the user selects a project.

## License

### DIFF
--- a/src/codewhisperer/commands/startTransformByQ.ts
+++ b/src/codewhisperer/commands/startTransformByQ.ts
@@ -20,7 +20,8 @@ import {
     pollTransformationJob,
     convertToTimeString,
     convertDateToTimestamp,
-    getValidModules,
+    getOpenProjects,
+    validateProjectSelection,
 } from '../service/transformByQHandler'
 import { QuickPickItem } from 'vscode'
 import { MultiStepInputFlowController } from '../../shared//multiStepInputFlowController'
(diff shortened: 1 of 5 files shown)


src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -59,51 +60,58 @@ interface UserInputState {
...
         span.record({ codeTransform_SessionId: codeTransformTelemetryState.getSessionId() })
-        let validModules: vscode.QuickPickItem[] | undefined
+
+        let openProjects: vscode.QuickPickItem[] = []
         try {
-            validModules = await getValidModules()
+            openProjects = await getOpenProjects()
         } catch (err) {
-            getLogger().error('Failed to get valid modules: ', err)
+            getLogger().error('Failed to get open projects: ', err)
```
nkomonen-amazon, line 102: In the future you should look to prefix your logs with something like `Gumby: ...` to make it easier to filter through all the logs

nkomonen-amazon (approved):

dhasani23:
In next PR, will add more unit tests to increase code coverage & do general clean-ups

justinmk3 (approved):

justinmk3:
A refactoring is strictly something that doesn't change behavior. I'd call this a "fix" or "perf" (if I understood the description).

PR #4094, opened 2023-11-28
featureDev: Fix /transform to show when /dev shows

jpinkney-aws:
## Problem
- Right now /transform will show/hide depending on the initial auth status when opening up Amazon Q. This doesn't behave correctly because a user can technically open up Amazon Q and then change their auth state after

## Solution
- Make /transform show when /dev shows

## License


justinmk3 (approved):

dogusata (approved):

src/amazonq/webview/ui/main.ts
```diff
@@ -77,7 +87,7 @@ export const createMynahUI = (ideApi: any, featureDevInitEnabled: boolean, gumby
             }
 
             // Unlock every authenticated tab that is now authenticated
-            if (featureDevEnabled) {
+            if (idcEnabled) {
```
gholt-aws, line 90: Why is this changing from if(featureDevEnabled)?
We said earlier that we wanted the IDE to be as ignorant as possible about *why* the feature was visible or not.

jpinkney-aws: I can change the name to amazonQConnected if that makes it more clear? Alternatively I could have changed `onUpdateAuthentication` to take in another field for gumbyEnabled but that doesn't really make sense since featureDev and gumby are enabled in all the same cases (and makes that part of the code harder to maintain)

PR #4096, opened 2023-11-28
featureDev: Move streaming client to shared client

jpinkney-aws:
## Problem
- Other folders were piggybacking off the streaming client

## Solution
- Move to it src/clients so it's in a more central place

## License


justinmk3 (approved):
Thanks!

PR #4099, opened 2023-11-28
featureDev: Simplify use of the messenger

jpinkney-aws:
## Problem
- We currently pass a single messenger around and have to call with it a tabID every time which can be error prone

## Solution
- Allow a messenger factory to take in a tab id and send events directly to it

## License


justinmk3 (approved):
Great to see followup on code ergonomics. 

A singleton could also be fine here, if there is no shared mutable global state.

PR #4101, opened 2023-11-29
fix(transform): various bug fixes

dhasani23:
## Problem

1. QuickPick which pops up to get user input on what project to transform (rarely) freezes / becomes unresponsive.
2. Missing notification to user when job is partially successful
3. /transform causes the QuickPick for user input to open then close immediately as a result of the fix to problem 1
4. We run a `mvn` command which does not create a directory when there are no dependencies to copy over, leading to a directory not found error

## Solution

1. Remove the shouldResume: true (make it false)
2. Add notification in this case
3. Sleep for 1s to give chat time to respond with message, then show input dialog pop-up
4. Include check to ensure directory exists before trying to read from it

## License

### DIFF
--- a/src/amazonqGumby/activation.ts
+++ b/src/amazonqGumby/activation.ts
@@ -32,7 +32,7 @@ export async function activate(context: ExtContext) {
             if (transformByQState.isRunning()) {
                 confirmStopTransformByQ(transformByQState.getJobId())
             } else {
-                vscode.window.showInformationMessage(CodeWhispererConstants.noOngoingJobMessage)
+                vscode.window.showInformationMessage(CodeWhispererConstants.noOngoingJobMessage, { modal: true })
             }
         }),
 
--- a/src/amazonqGumby/activation.ts
+++ b/src/amazonqGumby/activation.ts
@@ -32,7 +32,7 @@ export async function activate(context: ExtContext) {
             if (transformByQState.isRunning()) {
                 confirmStopTransformByQ(transformByQState.getJobId())
             } else {
-                vscode.window.showInformationMessage(CodeWhispererConstants.noOngoingJobMessage, { modal: true })
+                vscode.window.showInformationMessage(CodeWhispererConstants.noOngoingJobMessage)
             }
         }),
 
--- a/src/amazonqGumby/entrypoint.ts
+++ b/src/amazonqGumby/entrypoint.ts
@@ -8,13 +8,15 @@ import { jobInProgressMessage } from '../codewhisperer/models/constants'
 import { transformByQState } from '../codewhisperer/models/model'
 import { AuthUtil } from '../codewhisperer/util/authUtil'
 import vscode from 'vscode'
+import { sleep } from '../shared/utilities/timeoutUtils'
 
-export function processTransformByQ() {
+export async function processTransformByQ() {
     if (!AuthUtil.instance.isEnterpriseSsoInUse()) {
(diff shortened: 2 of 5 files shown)


src/amazonqGumby/activation.ts
```diff
@@ -32,7 +32,7 @@ export async function activate(context: ExtContext) {
             if (transformByQState.isRunning()) {
                 confirmStopTransformByQ(transformByQState.getJobId())
             } else {
-                vscode.window.showInformationMessage(CodeWhispererConstants.noOngoingJobMessage)
+                vscode.window.showInformationMessage(CodeWhispererConstants.noOngoingJobMessage, { modal: true })
```
justinmk3, line 35: I've noticed a lot of use of "modal" in gumby workflow. They should be avoided. Why does this one in particular need to be modal?

dhasani23: It doesn't, and can be changed back. We felt that the modal fit in better, and I wasn't aware that modals should be avoided -- is there a particular reason for that?

dhasani23: Looking at this again, it actually probably makes more sense to _not_make this a modal, so good point. But still curious on why modals should be avoided - I guess because they can be disruptive?

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -296,7 +295,11 @@ export async function startTransformByQ() {
                 )
             }
             if (transformByQState.isSucceeded()) {
-                vscode.window.showInformationMessage(CodeWhispererConstants.transformByQCompleted)
+                vscode.window.showInformationMessage(CodeWhispererConstants.transformByQCompleted, { modal: true })
+            } else if (transformByQState.isPartiallySucceeded()) {
+                vscode.window.showInformationMessage(CodeWhispererConstants.transformByQPartiallyCompletedMessage, {
+                    modal: true,
```
dhasani23, line 301: keeping this as a modal for consistency with the other information message above

src/amazonqGumby/entrypoint.ts
```diff
@@ -8,13 +8,15 @@ import { jobInProgressMessage } from '../codewhisperer/models/constants'
...
+import { sleep } from '../shared/utilities/timeoutUtils'
 
-export function processTransformByQ() {
+export async function processTransformByQ() {
     if (!AuthUtil.instance.isEnterpriseSsoInUse()) {
         vscode.window.showErrorMessage('Transform by Q requires an active IAM Identity Center connection')
         return
     }
     if (transformByQState.isNotStarted()) {
+        await sleep(1000) // sleep so that chat can respond first, then show input prompt
```
andrewyuq, line 19: is there a better way to do this?

dhasani23: I'm not sure of one. The issue is that when I set shouldResume to false, it causes any other event to close the input box, so the chat reply when the user enters `/transform` needs to happen first, and only then can I show the input box so that it doesn't open then immediately close in a glitchy way

andrewyuq (approved):

src/amazonqGumby/activation.ts
justinmk3: > still curious on why modals should be avoided - I guess because they can be disruptive?

Yes. There is a ton of discussion to be found on this via web search, but the short answer is that they demand a user decision. Things that don't actually require a user decision should not demand a user decision.

PR #4103, opened 2023-11-30
telemetry: enforce `result` and `reason` properties for emitted metrics

hayemaxi:
Problem: Different integrations within the toolkit publish metrics differently. The `result` and `reason` fields are meant to be standardized, but some integrations are adding their own fields. 

Solution: Verify that any metrics emitted using `emit()` contain a result property. If the result property is `'Failed'`, then a reason property must also exists. Any tests that violate this code are failed. We do not want to fail in general because it should not create an exception during production toolkit usage.

Includes an example fix for metric: `aws_modifySetting`

## License


src/test/globalSetup.test.ts
```diff
@@ -106,6 +111,12 @@ export const mochaHooks = {
         globals.codelensRootRegistry.dispose()
         globalSandbox.restore()
 
+        // Don't run any validations for tests on telemetry
+        if (!(this.currentTest ?? this.test)?.file?.includes('shared/telemetry/')) {
+            validateTelemetryEmitCalls(this.test as Hook, telemetrySpanSpy)
```
justinmk3, line 116: can we use assertTelemetry: https://github.com/aws/aws-toolkit-vscode/blob/c6abde34bf9ee5b9a87c6ae883e53ce35a889a2c/src/test/testUtil.ts#L207

that avoids mocks, and checks actual metrics emitted by actual code.

hayemaxi: assertTelemetry() takes in a metric name, but we want to just in general check if any published metrics meet the requirements.

While we could add a condition to assertTelemetry() that checks if result/reason exist properly, we would miss any current and future test cases where telemetry calls are not explicitly checked for.

Maybe instead I can add a method that can access all metrics published for the session.

src/test/globalSetup.test.ts
```diff
@@ -148,6 +161,37 @@ function writeLogsToFile(testName: string) {
...
+    const telemetryRunDocsStr =
+        'Consider using `.run()` instead of `.emit()`, which will set these properties automatically. ' +
+        'See https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md#guidelines'
+
+    for (const m of globals.telemetry.logger.getAllMetrics()) {
+        if (m.Metadata) {
+            const metadata = mapMetadata([])(m.Metadata)
+            assert.ok(
+                metadata.result !== undefined,
+                `This test emits metric \`${m.MetricName}\` without the \`result\` property. ` +
```
justinmk3, line 182: Nice. We can clarify a bit: we can drop reference to "This test" because the metric itself is the problem, not any test.
```suggestion
                `Metric \`${m.MetricName}\` was emitted without the \`result\` property. ` +
```

src/test/globalSetup.test.ts
```diff
@@ -148,6 +161,37 @@ function writeLogsToFile(testName: string) {
...
+                    `This property is always required. ${telemetryRunDocsStr}`
+            )
+
+            assert.ok(
+                metadata.result !== failedStr ? true : metadata.reason !== undefined,
+                `This test emits metric \`${m.MetricName}\` without the \`reason\` property. ` +
```
justinmk3, line 188:
```suggestion
                `Metric \`${m.MetricName}\` was emitted without the \`reason\` property. ` +
```

src/test/globalSetup.test.ts
```diff
@@ -106,6 +109,16 @@ export const mochaHooks = {
         globals.codelensRootRegistry.dispose()
         globalSandbox.restore()
 
+        // Don't run any validations for tests on telemetry
+        if (!(this.currentTest ?? this.test)?.file?.includes('shared/telemetry/')) {
+            try {
+                validateTelemetryMetrics()
```
justinmk3, line 115: https://github.com/aws/aws-toolkit-vscode/blob/235a6a67d5ada3e6a66748490d49a575bc13c70c/src/shared/telemetry/telemetryService.ts#L149

There we can skip the check if `!isAutomation()`, see also https://github.com/aws/aws-toolkit-vscode/blob/235a6a67d5ada3e6a66748490d49a575bc13c70c/src/shared/telemetry/activation.ts#L58-L61

docs/TESTPLAN.md
```diff
@@ -81,6 +81,7 @@ modifications/workarounds in `src/test/testRunner.ts`.
 -   Testing AWS SDK client functionality is cumbersome, verbose, and low-yield.
 -   Test code uses multiple “mocking” frameworks, which is confusing, error-prone, hard to onboard, and hard to use.
 -   Coverage not counted for integ tests (because of unresolved tooling issue).
+-   [Backlog](https://github.com/aws/aws-toolkit-vscode/blob/0f3685ab0dc8af3a214136ebfa233829d5d72b2c/src/shared/telemetry/exemptMetrics.ts) of metrics that do not pass validation but are temporarily exempted to not fail CI.
```
justinmk3, line 84: Nice, thank you for updating this doc.

src/shared/telemetry/telemetryService.ts
```diff
@@ -148,6 +148,11 @@ export class DefaultTelemetryService {
 
     public record(event: MetricDatum, awsContext?: AwsContext): void {
         if (this.telemetryEnabled) {
+            if (isAutomation()) {
+                // TODO: While this catches cases in code that is tested, untested code will still release incomplete metrics.
+                // Consider using custom lint rules to address these cases if possible.
+                validateMetricEvent(event)
```
justinmk3, line 154:
```suggestion
                validateMetricEvent(event, !isAutomation())
```

src/shared/telemetry/util.ts
```diff
@@ -85,3 +89,32 @@ export async function getUserAgent(
 
     return pairs.join(' ')
 }
+
+/**
+ * Validates that emitted telemetry metrics
+ * 1. contain a result property and
+ * 2. contain a reason propery if result = 'Failed'.
+ */
+export function validateMetricEvent(event: MetricDatum) {
```
justinmk3, line 98: let's add a `fatal` parameter, then if `!fatal` we only log a warning instead of throwing an error.

src/shared/telemetry/exemptMetrics.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Checks were added to all emitted telemetry metrics to ensure they contain the correct properties.
+ * This check does not exist outside test/automated runtimes because the error is not useful
+ * for the user and doesn't break telemetry.
+ *
+ * These are metrics that fail validation and are emitted by code that is covered in test cases.
+ * This allowlist exists to fix these incrementally and not have CI fail in the meantime.
+ */
+const validationExemptMetrics: Set<string> = new Set([
+    'amazonq_isReviewedChanges',
+    'apigateway_copyUrl',
```
justinmk3, line 19: Can we also fix a couple of these in this PR? That adds some useful context to the changeset.

hayemaxi: Would like to leave the fixes out of scope for this PR, but I can see some value in context. Will fix a metric.

justinmk3 (approved):

src/shared/telemetry/exemptMetrics.ts
```diff
@@ -0,0 +1,61 @@
...
+ * These are metrics that fail validation and are emitted by code that is covered in test cases.
+ * This allowlist exists to fix these incrementally and not have CI fail in the meantime.
+ */
+const validationExemptMetrics: Set<string> = new Set([
+    'amazonq_runCommand',
+    'amazonq_isReviewedChanges',
+    'apigateway_copyUrl',
+    'aws_loadCredentials',
+    'aws_validateCredentials',
+    'cloudwatchlogs_download',
```
dhruvigajjar, line 23: amazonq open/close chat doesnt have result anymore. do we need to include those metrics in this list?

hayemaxi: those metrics didn't fail for me when i was running tests, so it looks like the code that runs those is untested. We can only enforce for metrics that run in test cases at the moment- don't want to throw errors to the user during runtime.

Issue #4105, opened 2023-11-30
codecatalyst: "No handler found" error for vscode://.../connect/codecatalyst URI


justinmk3:
## Problem

- Customer reported this error:
    - ![image](https://github.com/aws/aws-toolkit-vscode/assets/55561878/d5962574-fe40-4258-9387-0d647390db90)
- This implies that this code wasn't run (in time)?: https://github.com/aws/aws-toolkit-vscode/blob/890fd5ec77b81e6a2129dad1175444b8f3ab902d/src/codecatalyst/uriHandlers.ts#L45-L49

## Steps to reproduce

1. in https://codecatalyst.aws web visit your project dev environments
2. clicking on the `Resume` link in codecatalyst to open a URI that AWS Toolkit should handle.
3. The error should never appear for a valid URI, which is why I think this might be an activation (or order-of-operations) bug.

## Expected behavior

- That error should not be shown for valid URIs such as `vscode://amazonwebservices.aws-toolkit-vscode/connect/codecatalyst` . 
- Verify that the Toolkit does not have an order-of-operations bug.
- The Toolkit should _not_ require the user to be logged-in for the URI handler to work.
    - When a URI is received and the user isn't logged in, the login flow should begin, just like if the user had clicked the `Start` node of the codecatalyst view.

## System details (run the `AWS: About Toolkit` command)

-   OS:
-   Visual Studio Code version: 
-   AWS Toolkit version: 2.0

PR #4112, opened 2023-12-01
fix(codecatalyst): "No handler found" error

justinmk3:
Problem:
vscode may invoke our URI handler during startup, before Toolkit activation has finished (and thus before all Toolkit services have registered their path handlers). #4105

    No handler found for: vscode://.../connect/codecatalyst URI

Solution:
- Do not call `vscode.window.registerUriHandler()` until after all Toolkit services have activated.
- Add "onUri" to package.json "activationEvents". It wasn't the root cause of the above issue, but it should be specified anyway. https://code.visualstudio.com/api/references/activation-events#onUri
- Improve visibility via logging.

## License

### DIFF
--- a/src/shared/vscode/uriHandler.ts
+++ b/src/shared/vscode/uriHandler.ts
@@ -22,23 +22,24 @@ interface HandlerWithParser<T> {
     parser?: QueryParser<T>
 }
 
+/**
+ * Note: https://code.visualstudio.com/api/references/vscode-api#window.registerUriHandler
+ * > An extension can only register a single uri handler in its entire activation lifetime.
+ */
 export class UriHandler implements vscode.UriHandler {
     public constructor() {}
 
     private handlers: Map<string, HandlerWithParser<any>> = new Map()
 
     public async handleUri(uri: vscode.Uri): Promise<void> {
-        getLogger().verbose(`UriHandler: received request on path "${uri.path}"`)
-
         const uriNoQuery = uri.with({ query: '' }).toString()
 
         if (!this.handlers.has(uri.path)) {
-            vscode.window.showErrorMessage(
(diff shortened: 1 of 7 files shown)


src/shared/vscode/uriHandler.ts
```diff
@@ -22,23 +22,24 @@ interface HandlerWithParser<T> {
...
-        getLogger().verbose(`UriHandler: received request on path "${uri.path}"`)
-
         const uriNoQuery = uri.with({ query: '' }).toString()
 
         if (!this.handlers.has(uri.path)) {
-            vscode.window.showErrorMessage(
-                localize('AWS.uriHandler.nohandler', 'No handler found for: {0}', uriNoQuery)
-            )
-            getLogger().verbose(`UriHandler: no valid handler found for "${uri.path}"`)
+            showViewLogsMessage(localize('AWS.uriHandler.nohandler', 'No handler for URI: {0}', uriNoQuery))
```
justinmk3, line 38: Before this PR, this would print:

    2023-12-01 09:24:05 [WARN]: UriHandler: no handler for path "/connect/codecatalyst" in handlers: Map(0) {}

After this PR it would only happen for an actually-unhandled path, and in that case would look like:

```
2023-12-01 09:13:47 [WARN]: UriHandler: no found handler for path "/foo/bar" in handlers: Map(3) {
  '/clone' => {
    handler: [AsyncFunction: cloneHandler],
    parser: [Function: parseCloneParams]
  },
  '/connect/codecatalyst' => {
    handler: [AsyncFunction: connectHandler],
    parser: [Function: parseConnectParams]
  },
  '/connect/caws' => {
    handler: [AsyncFunction: connectHandler],
    parser: [Function: parseConnectParamsOld]
  }
}
```

PR #4127, opened 2023-12-04
AmazonQ: Fix for incorrect code selection range values

ppiliaev:
## Problem

Intermittent 5xx errors are being returned from the backend due to invalid request payloads when customers have selected code on the first line of the file.

## Solution

The root cause was identified as incorrect payload formatting that occurred under specific conditions. The payload generation logic has been updated to properly handle these edge cases, and additional unit tests have been written to validate the payload structure in these scenarios

## License

### DIFF
--- a/src/test/codewhispererChat/editor/context/focusArea/focusAreaExtractor.test.ts
+++ b/src/test/codewhispererChat/editor/context/focusArea/focusAreaExtractor.test.ts
@@ -1,0 +1,312 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'assert'
+import { FocusAreaContextExtractor } from '../../../../../codewhispererChat/editor/context/focusArea/focusAreaExtractor'
+import { Range, TextDocument } from 'vscode'
+import { createMockDocument } from '../../../../codewhisperer/testUtil'
+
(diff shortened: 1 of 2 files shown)


dogusata (approved):

hayemaxi:
Can you add to the description of this pr: the details of the edge cases and what the conditions are to reach these edge cases?

src/test/codewhispererChat/editor/context/focusArea/focusAreaExtractor.test.ts
```diff
@@ -0,0 +1,312 @@
...
+                line: 1,
+                character: 9,
+            },
+        }
+        const limit = 10
+
+        // Act
+        const focusAreaContextExtractor = new FocusAreaContextExtractor()
+        const method = Reflect.get(focusAreaContextExtractor, 'getExtendedCodeBlockRange')
+        const result = method.call(focusAreaContextExtractor, document, range, limit)
```
hayemaxi, line 115: nit: is there a reason for this special syntax and not just
```
const focusAreaContextExtractor = new FocusAreaContextExtractor()
const result = focusAreaContextExtractor. getExtendedCodeBlockRange(document, range, limit)
```

hayemaxi (approved):

src/test/codewhispererChat/editor/context/focusArea/focusAreaExtractor.test.ts
ppiliaev: Yes, it's a private method, we cannot call it like this^

PR #4129, opened 2023-12-04
Hide Amazon Q and CodeCatalyst views in Sagemaker

bryceitoc9:
Hides Amazon Q and CodeCatalyst nodes **_AFTER THE EXTENSION IS ACTIVATED_**.
* This will not hide them before load, at the very least there will be a pop-in for the views.
* Depending on how much of the VS Code codebase SageMaker is willing to modify:
  * SageMaker can permanently disable the icon by setting a `aws.isSageMaker` [when clause context](https://code.visualstudio.com/api/references/when-clause-contexts) to `true`
  * SageMaker can also force the extension to load on open, which will slow load time for the IDE but will make the icons and menus pop out faster.
* This also prevents loading their code, which should speed up extension load

## License


hayemaxi (approved):
Nice. Can we open an issue on their end to track interest in adding the context into SageMaker?

justinmk3 (approved):

PR #4130, opened 2023-12-04
fix(transform): check for pom.xml in root dir only

dhasani23:
## Problem

We want to make sure users cannot submit a project unless a pom.xml is present in the root directory, not anywhere in the project.

## Solution

Enforce that pom.xml is in root directory specifically.

## License


andrewyuq (approved):

bryceitoc9 (changes requested):
Please add a changelog entry (`npm run newChange`). If this functionality worked fine with non-root-directory `pom.xml`, please mark this specifically as a breaking change.

Also: does this need to be communicated to doc writers?

dhasani23:
> Please add a changelog entry (`npm run newChange`). If this functionality worked fine with non-root-directory `pom.xml` before, please mark this specifically as a breaking change.

Done.

> Also: does this need to be communicated to doc writers?

Yes, and I've already started this process.

bryceitoc9 (approved):

PR #1099, opened 2020-05-19
Adds visualisation of yaml state machine definitions

alanbo:
## Description
Basic visualization capability for state machines defined in yaml.  

## Motivation and Context
We want to give users basic capability to visualize valid state machines in yaml. This won't be an official release and we will not provide errors or linting when the state machine definition is incorrect. 

## Checklist

- [x] I have read the **README** document
- [x] I have read the **CONTRIBUTING** document
- [x] My code follows the code style of this project
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed
- [x] A short description of the change has been added to the changelog using the script `npm run newChange`

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4134, opened 2023-12-05
refactor: remove yaml dependency, use js-yaml instead.

hayemaxi:
Problem: 'yaml' is outdated and requires human intervention to migrate to V2. It is only used in a few places in the codebase, whereas everywhere else, a similar yaml parser js-yaml is used.

Solution: remove usage of 'yaml' and only use 'js-yaml', as it has some features that we need (ability to write out to a string and skip invalid JSON entries, which 'yaml' does not seem to offer). We do not require all the features of 'yaml' at this time. Also, remove "errors" field from the state machines parser -> webview message. It is not used by anything and was only accessible with 'yaml'.

In one case of 'yaml', we used the 'maxAliasCount' option to prevent users from submitting state machines with aliases in them. Aliases do not prevent publishing normally to step functions if used normally; I believe it is to prevent the user from breaking the toolkit by exponentially copying anchors/aliases in the yaml. This does not affect publishing (it will fail as it is supposed to if the defintion is too large), and slightly slows down toolkit usage while the parse is happening. This validation is not necessary because this sort of behavior would be very specific and intentional from the user, and in my testing still causes slow downs and errors due to the logic centered around how the validation option was used.

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,7 +70,6 @@
                 "winston": "^3.7.1",
                 "winston-transport": "^4.5.0",
                 "xml2js": "^0.6.0",
-                "yaml": "^1.9.2",
                 "yaml-cfn": "^0.3.2"
             },
             "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -4397,7 +4397,6 @@
         "winston": "^3.7.1",
         "winston-transport": "^4.5.0",
         "xml2js": "^0.6.0",
-        "yaml": "^1.9.2",
         "yaml-cfn": "^0.3.2"
     },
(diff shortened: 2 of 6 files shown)


justinmk3 (approved):
Nice!

cc @alanbo 

Original discussion of js-yaml: https://github.com/aws/aws-toolkit-vscode/pull/1099#discussion_r427583441

Since that discussion, it seems that eliminating 'js-yaml' is less feasible than eliminating 'yaml'. The risk of accidental usage of yaml version 1.2, is mitigated because we'll surface errors returned by the service.

PR #4139, opened 2023-12-05
fix: help > open logs may not work after interacting with other options

hayemaxi:
Problem: Interacting with other options in the Amazon Q and Explorer views ellipses menu can cause the view logs option (of the same tree) to open an empty file. Since this command and menu options are maintained in package.json only, vscode may not properly supply arguments which leads to the open logs command breaking.

Solution: Remove unused parameter, only one log file is used for toolkit logging.

## License

### DIFF
--- a/src/shared/logger/commands.ts
+++ b/src/shared/logger/commands.ts
@@ -28,18 +28,16 @@ export class Logging {
         viewLogsAtMessage: Commands.from(this).declareOpenLogId('aws.viewLogsAtMessage'),
     }
 
-    public constructor(private readonly defaultLogUri: vscode.Uri, private readonly logger: Logger) {}
+    public constructor(private readonly logUri: vscode.Uri, private readonly logger: Logger) {}
 
-    public async openLogUri(logUri?: unknown): Promise<vscode.TextEditor | undefined> {
-        telemetry.toolkit_viewLogs.emit() // Perhaps add additional argument to know which log was viewed?
-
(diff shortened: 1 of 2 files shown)


src/shared/logger/commands.ts
```diff
@@ -30,10 +30,11 @@ export class Logging {
 
     public constructor(private readonly defaultLogUri: vscode.Uri, private readonly logger: Logger) {}
 
-    public async openLogUri(logUri = this.defaultLogUri): Promise<vscode.TextEditor | undefined> {
+    public async openLogUri(logUri?: unknown): Promise<vscode.TextEditor | undefined> {
         telemetry.toolkit_viewLogs.emit() // Perhaps add additional argument to know which log was viewed?
 
-        return vscode.window.showTextDocument(logUri)
+        // If the command is called directly from a package.json menu item, then logUri may be a vscode-provided object, typically a TreeNode.
+        return vscode.window.showTextDocument(logUri instanceof vscode.Uri ? logUri : this.defaultLogUri)
```
bryceitoc9, line 37: IIUC, tree nodes might drop a different parameter in? Are we always expecting it to open `this.defaultLogUri`? If so, can we just set that and remove the parameter entirely?

hayemaxi: Code doesn't explicitly use another parameter anywhere, but it seemed implicitly used via some javascript magic/vscode context. Was unable to repro a feature that used another log file. Removing for now, since it is unused.

PR #4140, opened 2023-12-05
CodeCatalystClient throws ToolkitError on call

bryceitoc9:
I think this is the best way to handle errors at the CodeCatalyst command level without forcibly handling errors for all commands the same way. This change should affect any errors coming through CodeCatalyst calls.

**CONSIDER:**
* Is this good enough for error consolidation?
* Should this be a warning or an error? That would have to be handled at the all-command level.

The following examples come from creating a valid Builder ID connection to CodeCatalyst, and modifying the cached credential files to make the access keys invalid, and then clicking "Clone Repository" in the CodeCatalyst view.

Old error message:
image

New error message:
image

## License

### DIFF
--- a/src/shared/clients/codecatalystClient.ts
+++ b/src/shared/clients/codecatalystClient.ts
@@ -277,7 +277,7 @@ class CodeCatalystClientInternal {
                         resolve(defaultVal)
                     } else {
                         const err = e as AWS.AWSError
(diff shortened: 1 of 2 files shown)


justinmk3 (approved):
Great, this is The Way

justinmk3 (approved):

Issue #3339, opened 2023-04-17
CI: failing CodeCatalyst E2E test: updateDevEnvironment causes InternalException


justinmk3:
## Problem

CC e2e tests failing on https://github.com/aws/aws-toolkit-vscode/commit/61bbc248221ac21e30fa74aaec20a12cae59c10a

```
  1) Test how this codebase uses the CodeCatalyst API
       Dev Environment functionality
         connects to a running Dev Environment:
     Error: Dev Environment failed to start (FAILED): test-alias-16817597…
      at /codebuild/output/src753672424/src/github.com/aws/aws-toolkit-vscode/src/shared/clients/codecatalystClient.ts:534:27
      at processTicksAndRejections (node:internal/process/task_queues:96:5)

  2) Test how this codebase uses the CodeCatalyst API
       Dev Environment functionality
         stops a running Dev Environment, then starts it up again:
     Error: Dev env 8c7c9a6c-4ed4-4b5c-8626-33024bc7b708 did not transition to one of the following statuses before timeout: "RUNNING"
      at backend_1 (src/testE2E/codecatalyst/client.test.ts:528:23)
      at Generator.next (<anonymous>)
      at fulfilled (src/testE2E/codecatalyst/client.test.ts:9:58)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)

  3) Test how this codebase uses the CodeCatalyst API
       Dev Environment functionality
         updates the properties of an existing dev environment:
     InternalException: An internal error has occurred while processing your request
```

## Full build log

- [buildlog.txt.gz](https://github.com/aws/aws-toolkit-vscode/files/11255432/buildlog.txt.gz)

## Solution

- ✅  does it need to be reported to the service team?
- is it a false positive? can the test harness make this clearer (by reporting relevant info, but not too much info, in the logs), or guard against it?
- print the related request-id with the failure?

JadenSimon:
> InternalException: An internal error has occurred while processing your request

This is possibly a backend regression with `updateDevEnvironment`

justinmk3:
Also noticed that during _successful_ CC e2e test run, logs show `AccessDeniedException` for `GetDevEnvironment` and `DeleteDevEnvironment`. Looks like this is expected because it's cleanup code in a before/after block. Ideally, we would not log expected errors in the build log.

```
2023-04-17 15:36:17 [ERROR]: API request failed: getDevEnvironment
params: {
[...]

PR #4144, opened 2023-12-06
tests: enable e2e, retries, bug fixes

hayemaxi:
https://github.com/aws/aws-toolkit-vscode/issues/3339
Failures for windows insiders will cancel stable test.

For e2e tests, will we monitor for any throttling.

Also:
- Retry strategy chosen somewhat arbitrarily, but these settings don't seem to delay the test suite by much if at all. In recent runs with these, throttling has stopped.

- Adds retries with jitter by utilizing retries in the SDK, see
https://github.com/aws/aws-sdk-js/blob/3e616251947c73d5239178c167a9d73d985ca581/lib/util.js#L884.

- Also fixes various bugs with the tests that causes errors to be thrown inadvertently, including type checking/verifty the code of some
exceptions. In one case, this threw an AccessDeniedException when checking a dev env that was already deleted previously.

## License


hayemaxi:
throttling with no parallel runs :(

src/testE2E/codecatalyst/client.test.ts
```diff
@@ -354,16 +355,15 @@ describe.skip('Test how this codebase uses the CodeCatalyst API', function () {
             assert.ok(spaces.find(space => space.name === spaceName))
         })
 
-        it('lists all projects for the given user', async function () {
+        // TODO: Re-add this test and delete the projects after the test suite with the appropriate SDK API.
+        it.skip('lists all projects for the given user', async function () {
```
justinmk3, line 359: Assuming the skip() is still intended, I don't quite understand the comment. What's meant by "the appropriate sdk api" ?

justinmk3 (approved):

src/testE2E/codecatalyst/client.test.ts
hayemaxi: updated this comment and one other one

justinmk3 (approved):

nkomonen-amazon:
Nice! @hayemaxi one thing I learned recently is that editing the `/runIntegrationTests` comment and adding some spaces will trigger the integration tests so you don't need to make a new comment each time.

nkomonen-amazon (approved):

PR #4146, opened 2023-12-06
fix/test: improve error handling and add more tests

dhasani23:
## Problem

We had some duplicate code and other minor cleanups to handle. We wanted more test coverage.

## Solution

Remove duplicate code and do minor cleanups. Add test cases.

## License

### DIFF
--- a/src/codewhisperer/commands/startTransformByQ.ts
+++ b/src/codewhisperer/commands/startTransformByQ.ts
@@ -291,18 +291,12 @@ export async function startTransformByQ() {
             )
         }
         if (transformByQState.isSucceeded()) {
-            vscode.window.showInformationMessage(CodeWhispererConstants.transformByQCompleted, { modal: true })
+            vscode.window.showInformationMessage(CodeWhispererConstants.transformByQCompletedMessage, { modal: true })
         } else if (transformByQState.isPartiallySucceeded()) {
             vscode.window.showInformationMessage(CodeWhispererConstants.transformByQPartiallyCompletedMessage, {
                 modal: true,
             })
         }
-        await sleep(2000) // needed as a buffer to allow TransformationHub to update before state is updated
-        clearInterval(intervalId)
-        transformByQState.setToNotStarted() // so that the "Transform by Q" button resets
-        vscode.commands.executeCommand('setContext', 'gumby.isStopButtonAvailable', false)
-        await vscode.commands.executeCommand('aws.amazonq.refresh')
-        vscode.commands.executeCommand('aws.amazonq.showPlanProgressInHub', startTime.getTime())
     }
     await sleep(2000) // needed as a buffer to allow TransformationHub to update before state is updated
     clearInterval(intervalId)
--- a/src/codewhisperer/service/transformByQHandler.ts
+++ b/src/codewhisperer/service/transformByQHandler.ts
@@ -156,12 +156,9 @@ export function getSha256(fileName: string) {
     return hasher.digest('base64')
 }
 
-// TODO: later, consider enhancing the S3 client to include this functionality
-export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrlResponse) {
-    const sha256 = getSha256(fileName)
-
+export function getHeadersObj(sha256: string, kmsKeyArn: string | undefined) {
     let headersObj = {}
-    if (resp.kmsKeyArn === undefined || resp.kmsKeyArn.length === 0) {
+    if (kmsKeyArn === undefined || kmsKeyArn.length === 0) {
         headersObj = {
             'x-amz-checksum-sha256': sha256,
             'Content-Type': 'application/zip',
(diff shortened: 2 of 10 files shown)


src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -560,7 +560,7 @@ export async function pollTransformationJob(jobId: string, validStates: string[]
 
 async function checkIfGradle(projectPath: string) {
     const gradleBuildFile = await vscode.workspace.findFiles(
-        new vscode.RelativePattern(projectPath, '**/build.gradle'),
+        new vscode.RelativePattern(projectPath, 'build.gradle'),
```
dhasani23, line 563: we only want to recognize a project as being gradle if we see the build.gradle in the root directory

src/testInteg/amazonqGumby/transformByQ.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    })
+
+    it('WHEN upload artifact to S3 with expired upload URL THEN fails to upload', async function () {
+        const sha256 = getSha256(zippedCodePath)
+        const createUploadUrlResponse = await codeWhisperer.codeWhispererClient.createUploadUrl({
+            contentChecksum: sha256,
+            contentChecksumType: CodeWhispererConstants.contentChecksumType,
+            uploadIntent: CodeWhispererConstants.uploadIntent,
+        })
+        await sleep(65000) // sleep for 65 seconds since the upload URL expires after 60 seconds
```
nkomonen-amazon, line 47: This will slow down the integ tests significantly. If this is documented S3 behavior we can assume that it expires when expected. instead you should write a test stubs+throws on the request and just ensure you handle it as expected

src/testInteg/amazonqGumby/transformByQ.test.ts
```diff
@@ -0,0 +1,45 @@
...
+        zippedCodePath = await zipCode(tempDir)
+    })
+
+    afterEach(function () {
+        fs.rmSync(tempDir, { recursive: true, force: true })
+    })
+
+    it('WHEN upload payload with valid request THEN succeeds', async function () {
+        await assert.doesNotReject(async () => {
+            await uploadPayload(zippedCodePath)
```
nkomonen-amazon, line 33: Any way to clean this up from S3 after?

dhasani23: Modified this test case so this is now N/A

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -291,18 +291,12 @@ export async function startTransformByQ() {
...
             })
         }
-        await sleep(2000) // needed as a buffer to allow TransformationHub to update before state is updated
-        clearInterval(intervalId)
-        transformByQState.setToNotStarted() // so that the "Transform by Q" button resets
-        vscode.commands.executeCommand('setContext', 'gumby.isStopButtonAvailable', false)
-        await vscode.commands.executeCommand('aws.amazonq.refresh')
-        vscode.commands.executeCommand('aws.amazonq.showPlanProgressInHub', startTime.getTime())
     }
     await sleep(2000) // needed as a buffer to allow TransformationHub to update before state is updated
```
justinmk3, line 301: What happens if the other end (e.g. the network) is slow? When are these sleep() calls going to be revisited?

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -171,9 +168,16 @@ export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrl
...
     }
+    return headersObj
+}
+
+// TODO: later, consider enhancing the S3 client to include this functionality
+export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrlResponse) {
+    const sha256 = getSha256(fileName)
+    const headersObj = getHeadersObj(sha256, resp.kmsKeyArn)
 
     await sleep(2000) // pause to give time to recognize potential cancellation
```
justinmk3, line 182: These are everywhere. I don't understand why gumby is the only feature that requires sleep() calls all over the code.

dhasani23: As mentioned, it's in my plans to re-evaluate the necessity of the sleep() calls

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -488,6 +496,7 @@ export async function getTransformationSteps(jobId: string) {
         return response.transformationPlan.transformationSteps
     } catch (e: any) {
         const errorMessage = e?.message || 'Error in GetTransformationPlan API call'
+        getLogger().error('CodeTransform: GetTransformationPlan error = ', errorMessage)
```
justinmk3, line 499: If this code is called by a command (usually the case), the thrown error (below) is already logged.  https://github.com/aws/aws-toolkit-vscode/blob/af655f604ffc8fa99a49a901636dd270175b5c2e/src/shared/vscode/commands2.ts#L640

Is that not the case for all of these places?

All of this error handling looks very redundant, at least the `telemetry.codeTransform_logApiError` call could be factored out.

dhasani23: It's helpful to have the "CodeTransform: {API}" prefix to our error logging

src/codewhisperer/service/transformByQHandler.ts
justinmk3: is `findFiles` still needed then?

dhasani23: No, `fs.existsSync()` is an acceptable alternative, will make the change now.

nkomonen-amazon (approved):

PR #4149, opened 2023-12-07
feat(amazonqFeatureDev, telemetry): improve generateApproach api error metrics and minor fixes.

sannicm:
## Problem

Currently featureDev errors are returning a generic failure. In the purpose of understanding the specific error with telemetry finding out if it was a 403 or a 409 was hard.

## Solution

Add status code to api errors to track in telemetry.

## License

### DIFF
--- a/src/amazonqFeatureDev/client/featureDev.ts
+++ b/src/amazonqFeatureDev/client/featureDev.ts
@@ -52,6 +52,15 @@ async function createFeatureDevStreamingClient(): Promise<CodeWhispererStreaming
     return streamingClient
 }
 
+const streamResponseErrors: Record<string, number> = {
+    ValidationException: 400,
+    AccessDeniedException: 403,
+    ResourceNotFoundException: 404,
+    ConflictException: 409,
+    ThrottlingException: 429,
+    InternalServerException: 500,
+}
+
 export class FeatureDevClient {
     private async getClient() {
         // Should not be stored for the whole session.
@@ -75,13 +84,9 @@ export class FeatureDevClient {
                 requestId: $response.requestId,
             })
             return conversationId
-        } catch (e) {
-            getLogger().error(
-                `${featureName}: failed to start conversation: ${(e as Error).message} RequestId: ${
-                    (e as any).requestId
-                }`
-            )
-            throw new ApiError((e as Error).message, 'CreateConversation', (e as any).statusCode)
+        } catch (e: any) {
+            getLogger().error(`${featureName}: failed to start conversation: ${e.message} RequestId: ${e.requestId}`)
+            throw new ApiError(e.message, 'CreateConversation', e.statusCode)
         }
     }
 
@@ -109,14 +114,12 @@ export class FeatureDevClient {
             return response
         } catch (e: any) {
             getLogger().error(
-                `${featureName}: failed to generate presigned url: ${(e as Error).message} RequestId: ${
-                    (e as any).requestId
-                }`
+                `${featureName}: failed to generate presigned url: ${e.message} RequestId: ${e.requestId}`
             )
             if (e.code === 'ValidationException' && e.message.includes('Invalid contentLength')) {
                 throw new ContentLengthError()
             }
-            throw new ApiError((e as Error).message, 'CreateUploadUrl', (e as any).statusCode)
+            throw new ApiError(e.message, 'CreateUploadUrl', e.statusCode)
         }
     }
 
@@ -151,11 +154,9 @@ export class FeatureDevClient {
                 assistantResponse.push(responseItem.assistantResponseEvent!.content)
             }
             return assistantResponse.join(' ')
-        } catch (e) {
-            getLogger().error(
-                `${featureName}: failed to execute planning: ${(e as Error).message} RequestId: ${(e as any).requestId}`
-            )
-            throw new ApiError((e as Error).message, 'GeneratePlanFailed', (e as any)?.$metadata?.httpStatusCode ?? 500) // Internal server errors are not returning the http status code
+        } catch (e: any) {
+            getLogger().error(`${featureName}: failed to execute planning: ${e.message} RequestId: ${e.requestId}`)
+            throw new ApiError(e.message, 'GeneratePlan', e.$metadata?.httpStatusCode ?? streamResponseErrors[e.name])
         }
     }
 }
--- a/src/amazonqFeatureDev/client/featureDev.ts
+++ b/src/amazonqFeatureDev/client/featureDev.ts
@@ -86,7 +86,7 @@ export class FeatureDevClient {
             return conversationId
         } catch (e: any) {
             getLogger().error(`${featureName}: failed to start conversation: ${e.message} RequestId: ${e.requestId}`)
-            throw new ApiError(e.message, 'CreateConversation', e.statusCode)
+            throw new ApiError(e.message, 'CreateConversation', e.code, e.statusCode)
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -117,7 +116,7 @@ export class FeatureDevClient {
             if (e.code === 'ValidationException' && e.message.includes('Invalid contentLength')) {
                 throw new ContentLengthError()
             }
-            throw new ToolkitError((e as Error).message, { code: 'CreateUploadUrlFailed' })
+            throw new ApiError((e as Error).message, 'CreateUploadUrl', (e as any).statusCode)
```
jpinkney-aws, line 119: hasn't e already been casted to any on line 110? Why do we need to cast again?

src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -156,7 +155,7 @@ export class FeatureDevClient {
             getLogger().error(
                 `${featureName}: failed to execute planning: ${(e as Error).message} RequestId: ${(e as any).requestId}`
             )
-            throw new ToolkitError((e as Error).message, { code: 'GeneratePlanFailed' })
+            throw new ApiError((e as Error).message, 'GeneratePlanFailed', (e as any)?.$metadata?.httpStatusCode ?? 500) // Internal server errors are not returning the http status code
```
jpinkney-aws, line 158: Is this the only error that doesn't return a status code?

src/amazonqFeatureDev/errors.ts
```diff
@@ -64,6 +70,12 @@ export class ContentLengthError extends ToolkitError {
     }
 }
 
+export class ApiError extends ToolkitError {
+    constructor(message: string, api: string, errorCode: number) {
+        super(message, { code: `${api} - ${errorCode}` })
```
jpinkney-aws, line 75: @nkomonen-amazon or @dhruvigajjar are there any issues with this code having a space in it (just double checking that it doesn't break anything on the backend)

src/amazonqFeatureDev/client/featureDev.ts
sannicm: All midstream errors are at the moment translated as internal server exception by RTS, I'll update this CR to leave a solution that in the future they may at least return a different error name, so we can get the code form that. I do not expect them to update the midstream issue atm.

dianaabv16 (commented):
what do you think about adding httpStatusCode  field instead of appending it to reason?
[link](https://github.com/aws/aws-toolkit-common/blob/a2ab451d26b7dae563985a5b077393ed5cd54e11/telemetry/definitions/commonDefinitions.json#L157)

src/amazonqFeatureDev/client/featureDev.ts
sannicm: No need, I'm removing the casts.

src/amazonqFeatureDev/errors.ts
```diff
@@ -64,6 +70,12 @@ export class ContentLengthError extends ToolkitError {
     }
 }
 
+export class ApiError extends ToolkitError {
+    constructor(message: string, api: string, errorCode: number) {
+        super(message, { code: `${api}-${errorCode}` })
```
dianaabv16, line 75: I would prefer here to have : `${api}-${errorName}-${errorCode}`
In that way it add more readability for the people not familiar with error codes

src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -152,11 +154,14 @@ export class FeatureDevClient {
...
-        } catch (e) {
-            getLogger().error(
-                `${featureName}: failed to execute planning: ${(e as Error).message} RequestId: ${(e as any).requestId}`
+        } catch (e: any) {
+            getLogger().error(`${featureName}: failed to execute planning: ${e.message} RequestId: ${e.requestId}`)
+            throw new ApiError(
+                e.message,
+                'GeneratePlan',
+                e.name,
+                e.$metadata?.httpStatusCode ?? streamResponseErrors[e.name]
```
jpinkney-aws, line 163: I'm kind of confused with what streamResponseErrors does. Shouldn't the incoming errors already have the correct httpStatusCode set? I would think things like AccessDeniedException would already be throwing a 403 and thus it's not really needed in that mapping

sannicm: This is just a map to prepare if in the future RTS would handle midstreaming errors differently. Basically incoming errors thrown before streaming, i.e. validateTaskAssist, will throw an error we can get the status code from. This is no the case for midstream errors which are just translated to an InternalServerError with no extra metadata.

src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -76,13 +84,9 @@ export class FeatureDevClient {
...
             })
             return conversationId
-        } catch (e) {
-            getLogger().error(
-                `${featureName}: failed to start conversation: ${(e as Error).message} RequestId: ${
-                    (e as any).requestId
-                }`
-            )
-            throw new ToolkitError((e as Error).message, { code: 'CreateConversationFailed' })
+        } catch (e: any) {
```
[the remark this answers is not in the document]
sannicm: I basically didn't handle the type of the error as this line here is just about avoiding casting using the `as` typescript operator, if there are wrong metrics, this issue will only impact on the metrics we send. with the metrics it will be evident if our assumptions here are wrong.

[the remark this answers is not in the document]
sannicm: I'll add a check and a different error for cases that it is not being handled currently.

src/amazonqFeatureDev/client/featureDev.ts
sannicm: Not really, mid-stream errors are only thrown with a name. I improved the error parsing in the later commits alluding to a comment about not using e: any, that will make more clear what the shape of the error is.

src/amazonqFeatureDev/errors.ts
```diff
@@ -64,6 +70,18 @@ export class ContentLengthError extends ToolkitError {
...
 
+export class UnknownApiError extends ToolkitError {
+    constructor(message: string, api: string) {
+        super(message, { code: `${api}-Unknown` })
+    }
+}
+
+export class ApiError extends ToolkitError {
+    constructor(message: string, api: string, errorName: string, errorCode: number) {
+        super(message, { code: `${api}-${errorName}-${errorCode}` })
```
justinmk3, line 81: off-topic/future reference: we need to enhance the default handling so that http status codes are automatically captured from the thrown exception and stored in the metric's  https://github.com/aws/aws-toolkit-vscode/blob/af655f604ffc8fa99a49a901636dd270175b5c2e/src/shared/telemetry/spans.ts#L151-L161

Ref: https://sim.amazon.com/issues/IDE-11056

justinmk3 (approved):

justinmk3:
> what do you think about adding httpStatusCode field instead of appending it to reason?
> [link](https://github.com/aws/aws-toolkit-common/blob/a2ab451d26b7dae563985a5b077393ed5cd54e11/telemetry/definitions/commonDefinitions.json#L157)

💯  yes, that will avoid some tech debt here, while also avoiding lots of potential noise in the `reason` (`code:`) field.

sannicm:
> > what do you think about adding httpStatusCode field instead of appending it to reason?
> > [link](https://github.com/aws/aws-toolkit-common/blob/a2ab451d26b7dae563985a5b077393ed5cd54e11/telemetry/definitions/commonDefinitions.json#L157)
> 
> 💯 yes, that will avoid some tech debt here, while also avoiding lots of potential noise in the `reason` (`code:`) field.
> 
> Note that `record({...})` allows any arbitrary fields to be added, so you don't need to update the common definitions.

Oh, wasn't aware of that. we'll take that into account for future changes if needed to improve telemetry. Could you please merge this PR?

PR #4152, opened 2023-12-07
amazonQ: only add missing scopes intentionally

nkomonen-amazon:
### Problem:

Before we would always add missing scopes whenever an existing CW user was missing the new scopes for chat. This would be done during the reauthentication process automatically.

The issue is that we did not want this and only wanted to add missing Amazon Q/chat scopes when the user intentionally wanted them to be added.

### Solution:
- Don't always add missing scopes during `reauthenticate()`. Instead use an arg to determine if we should
- We will add missing scopes in the Chat "reauthenticate" prompt as well as the Amazon Q node.
- Everywhere else will simply reauthenticate without adding new scopes

## License

### DIFF
--- a/src/amazonq/auth/controller.ts
+++ b/src/amazonq/auth/controller.ts
@@ -30,6 +30,6 @@ export class AuthController {
     }
 
     private handleReAuth() {
-        reconnect.execute(placeholder, amazonQChatSource)
+        reconnect.execute(placeholder, amazonQChatSource, true)
     }
 }
--- a/src/amazonq/explorer/amazonQChildrenNodes.ts
+++ b/src/amazonq/explorer/amazonQChildrenNodes.ts
@@ -50,7 +50,7 @@ export const switchToAmazonQNode = () =>
  */
 export const enableAmazonQNode = () =>
     // Simply trigger re-auth to obtain the proper scopes- same functionality as if requested in the chat window.
-    reconnect.build(placeholder, cwTreeNodeSource).asTreeNode({
+    reconnect.build(placeholder, cwTreeNodeSource, true).asTreeNode({
         label: localize('AWS.amazonq.enable', 'Enable'),
         iconPath: getIcon('vscode-debug-start'),
         contextValue: 'awsEnableAmazonQ',
--- a/src/codewhisperer/commands/basicCommands.ts
+++ b/src/codewhisperer/commands/basicCommands.ts
@@ -145,9 +145,10 @@ export const selectCustomizationPrompt = Commands.declare(
 
 export const reconnect = Commands.declare(
     { id: 'aws.codewhisperer.reconnect', compositeKey: { 1: 'source' } },
-    () => async (_: VsCodeCommandArg, source: CodeWhispererSource) => {
-        await AuthUtil.instance.reauthenticate()
-    }
+    () =>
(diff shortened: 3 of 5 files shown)


src/amazonq/auth/controller.ts
```diff
@@ -30,6 +30,6 @@ export class AuthController {
     }
 
     private handleReAuth() {
-        reconnect.execute(placeholder, amazonQChatSource)
+        reconnect.execute(placeholder, amazonQChatSource, true)
```
nkomonen-amazon, line 33: Add scopes in the chat reauth message

src/amazonq/explorer/amazonQChildrenNodes.ts
```diff
@@ -50,7 +50,7 @@ export const switchToAmazonQNode = () =>
  */
 export const enableAmazonQNode = () =>
     // Simply trigger re-auth to obtain the proper scopes- same functionality as if requested in the chat window.
-    reconnect.build(placeholder, cwTreeNodeSource).asTreeNode({
+    reconnect.build(placeholder, cwTreeNodeSource, true).asTreeNode({
```
nkomonen-amazon, line 53: Add scopes in the enable amazon q node

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -145,9 +145,10 @@ export const selectCustomizationPrompt = Commands.declare(
 
 export const reconnect = Commands.declare(
     { id: 'aws.codewhisperer.reconnect', compositeKey: { 1: 'source' } },
-    () => async (_: VsCodeCommandArg, source: CodeWhispererSource) => {
-        await AuthUtil.instance.reauthenticate()
-    }
+    () =>
+        async (_: VsCodeCommandArg, source: CodeWhispererSource, addMissingScopes = false) => {
```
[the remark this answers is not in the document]
nkomonen-amazon: Good find, I thought that TS was implicitly assuming it was a boolean

ppiliaev:
> @ppiliaev can you verify that this change is fine

Changes look fine for me

ppiliaev (approved):

src/codewhisperer/commands/basicCommands.ts
```diff
@@ -145,9 +145,10 @@ export const selectCustomizationPrompt = Commands.declare(
 
 export const reconnect = Commands.declare(
     { id: 'aws.codewhisperer.reconnect', compositeKey: { 1: 'source' } },
-    () => async (_: VsCodeCommandArg, source: CodeWhispererSource) => {
-        await AuthUtil.instance.reauthenticate()
-    }
+    () =>
+        async (_: VsCodeCommandArg, source: CodeWhispererSource, addMissingScopes: boolean = false) => {
```
hayemaxi, line 149: thanks for updating the signature! My concern though is that we might need something like
```
if (typeof addMissingScopes !== 'boolean') {
...
}
```

I see this command is referenced in nodes in package.json. If they are called from a package.json menu item, wouldn't this value have the potential to be something else?

nkomonen-amazon: Ah I see, in the package.json uses the `addMissingScopes` arg doesn't get overwritten in any way and will always use the default value.

But this behavior could change on the vscode side, so I'll add the check you mentioned above.

justinmk3 (approved):

hayemaxi (approved):

PR #4155, opened 2023-12-08
Add timed refresh to Cloud9 CodeWhisperer view after load

bryceitoc9:
## Problem
The nodes in the CodeWhisperer view in Cloud9 sometimes don't work.
* Analyzing the browser console shows that the extension host disposes creates commands for the nodes and disposes them (these have a number tied to them as an "ID").
* Most of the time, the view picks up the most recent copy of the commands (the browser console shows the ID being executed when the node is clicked)
* Sometimes, this is tied to the ID that's disposed. In these cases, the node actions won't work, with no error feedback (outside the browser console)

## Solution
Force a refresh of the CodeWhisperer view in Cloud9 only, 5 seconds after the end of the `extension.ts` `activate()` function.
* This disposes the actions and presumably recreates the nodes in the view
* It's hard to tell if this is truly a workaround/fix: since the error is sporadic, and the node never works _immediately_ during load (it always takes a second or two at least), we can't tell for sure if the view is going to recover on its own or continue to fail.

## License

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -78,6 +78,7 @@ import { registerWebviewErrorHandler } from './webviews/server'
 import { initializeManifestPaths } from './extensionShared'
 import { ChildProcess } from './shared/utilities/childProcess'
 import { initializeNetworkAgent } from './codewhisperer/client/agent'
+import { Timeout } from './shared/utilities/timeoutUtils'
 
 let localize: nls.LocalizeFunc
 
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -283,6 +284,19 @@ export async function activate(context: vscode.ExtensionContext) {
...
+        //       where `getChildren` calls are executed on load but the UI doesn't respond
+        //       (the extension host disposes the commands and recreates them,
+        //       but the nodes remain tied to the old commands).
+        //       This forces a refresh after 5 seconds to ensure a refresh happens at the end of initial extension load.
+        //       If the issue is due to activity on the views,
+        //       this should be fired after planned activities have finished.
+        if (isCloud9()) {
+            const timeout = new Timeout(5000).onCompletion(() => {
+                vscode.commands.executeCommand('aws.codeWhisperer.refresh')
+                timeout.dispose()
```
justinmk3, line 297: could be deleted because onCompletion is triggered on "dispose": https://github.com/aws/aws-toolkit-vscode/blob/d25bc44d5ed2069a3245abef2b6ae4618ce5e61d/src/shared/utilities/timeoutUtils.ts#L133

justinmk3 (approved):

justinmk3 (approved):

PR #4157, opened 2023-12-11
feat(amazonqFeatureDev): add new metrics to api operations

sannicm:
## Problem

featureDev needed to track errors in the metrics in a more detailed way.

## Solution

add new metrics in order to get better error tracking.

## License

This PR needs this dependency to be merged and in this PR the version of the library updated: https://github.com/aws/aws-toolkit-common/pull/645/files

### DIFF
--- a/src/amazonqFeatureDev/errors.ts
+++ b/src/amazonqFeatureDev/errors.ts
@@ -55,6 +55,12 @@ export class PrepareRepoFailedError extends ToolkitError {
     }
 }
 
+export class UploadCodeError extends ToolkitError {
+    constructor(statusCode: string) {
+        super('Unable to upload code', { code: `UploadCode-${statusCode}` })
+    }
+}
+
 export class IllegalStateTransition extends ToolkitError {
     constructor() {
         super('Illegal transition between states, restart the conversation', { code: 'IllegalStateTransition' })
--- a/src/amazonqFeatureDev/session/sessionState.ts
+++ b/src/amazonqFeatureDev/session/sessionState.ts
@@ -34,7 +34,7 @@ export class PrepareRefinementState implements Omit<SessionState, 'uploadId'> {
         this.tokenSource = new vscode.CancellationTokenSource()
     }
     async interact(action: SessionStateAction): Promise<SessionStateInteraction> {
-        const uploadId = await telemetry.amazonq_uploadCode.run(async span => {
+        const uploadId = await telemetry.amazonq_createUpload.run(async span => {
             span.record({ amazonqConversationId: this.config.conversationId })
             const { zipFileBuffer, zipFileChecksum } = await prepareRepoData(
                 this.config.sourceRoot,
--- a/src/amazonqFeatureDev/session/sessionState.ts
+++ b/src/amazonqFeatureDev/session/sessionState.ts
@@ -34,15 +34,24 @@ export class PrepareRefinementState implements Omit<SessionState, 'uploadId'> {
         this.tokenSource = new vscode.CancellationTokenSource()
     }
     async interact(action: SessionStateAction): Promise<SessionStateInteraction> {
-        const { zipFileBuffer, zipFileChecksum } = await prepareRepoData(this.config.sourceRoot, action.telemetry)
+        const uploadId = await telemetry.amazonq_uploadCode.run(async span => {
+            span.record({ amazonqConversationId: this.config.conversationId })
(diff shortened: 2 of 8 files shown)


src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -34,15 +34,24 @@ export class PrepareRefinementState implements Omit<SessionState, 'uploadId'> {
         this.tokenSource = new vscode.CancellationTokenSource()
     }
     async interact(action: SessionStateAction): Promise<SessionStateInteraction> {
-        const { zipFileBuffer, zipFileChecksum } = await prepareRepoData(this.config.sourceRoot, action.telemetry)
+        const uploadId = await telemetry.amazonq_createUpload.run(async span => {
```
dianaabv16, line 37: not sure if uploadId var is needed let' use try/catch here.
example: https://github.com/aws/aws-toolkit-vscode/blob/853b3f8de8d2907f323ea80948324ee6ce4da734/src/amazonqFeatureDev/session/sessionState.ts#L69

src/amazonqFeatureDev/util/upload.ts
```diff
@@ -28,8 +30,8 @@ export async function uploadCode(url: string, buffer: Buffer, checksumSha256: st
                 }),
             },
         })
-    } catch (e) {
+    } catch (e: any) {
         getLogger().error(`${featureName}: failed to upload code to s3: ${(e as Error).message}`)
-        throw e
+        throw new UploadCodeError(e instanceof got.HTTPError ? `${e.response.statusCode}` : '')
```
jpinkney-aws, line 35: When the status code is undefined can we add something like "unknown" here instead of ""?

Right now I think the code that gets send to the telemetry is `UploadCode-` when that piece of code is hit

src/amazonqFeatureDev/errors.ts
```diff
@@ -55,6 +55,12 @@ export class PrepareRepoFailedError extends ToolkitError {
     }
 }
 
+export class UploadCodeError extends ToolkitError {
+    constructor(statusCode: string) {
+        super('Unable to upload code', { code: `UploadCode-${statusCode}` })
```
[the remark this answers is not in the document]
sannicm: In this case, we cannot get error name.

src/amazonqFeatureDev/session/sessionState.ts
sannicm: I need the uploadId as an input in the RefinementState.

src/amazonqFeatureDev/util/upload.ts
sannicm: True, that will make more sense in the metric.

src/amazonqFeatureDev/errors.ts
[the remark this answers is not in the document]
sannicm: I would rather here though just leave the statusCode number, as how I said, is a one to one mapping to http statusCode meanings and this comes from a got call which is an http client.

jpinkney-aws:
The code itself looks fine to me but the tests are now failing

src/amazonqFeatureDev/session/sessionState.ts
dianaabv16: I still would prefer to have try/catch inside iteract

sannicm: I canot wrap the whole interaction, otherwise it means also wrapping the next step constructor which is not needed in this case. It will also then mix metrics with the approachInoke as the RefinementState does that. better to leave them separated so we don't have duplicate metrics.

sannicm:
> The code itself looks fine to me but the tests are now failing

I thought I had fixed it, locally it was passing. I will check again.

jpinkney-aws (approved):

nkomonen-amazon (approved):

justinmk3:
I see that here again we're storing extra data in the `ToolkitError.code` field (which emits as the `reason` field in telemetry). 

    code: `UploadCode-${statusCode}`

As mentioned in  https://github.com/aws/aws-toolkit-vscode/pull/4149#discussion_r1423868511 , this is debt that we're accruing. Storing sub-values in `reason` explodes the kinds of `reason` and defeats the purpose of the `httpsStatusCode` field. Please consider storing the http status code in the `httpStatusCode` metric field.

PR #4162, opened 2023-12-12
Remove moment from getRelativeDate

bryceitoc9:
`moment` is deprecated: https://momentjs.com/docs/ . We have two core `moment` dependencies: one is for localizing timestamps, and one is for providing relative dates. This PR targets the latter since it's a more involved change.

TODO: Localize outside English? FWIW the rest of the toolkit isn't.

## License


src/test/shared/utilities/textUtilities.test.ts
```diff
@@ -92,18 +92,28 @@ describe('getStringHash', async function () {
 })
 
 describe('getRelativeDate', function () {
-    const now = new Date(2020, 2, 2, 2, 2, 2)
+    const now = new Date(2020, 4, 4, 4, 4, 4) // adjusts for clock skew modifier in `getRelativeDate` fn.
```
justinmk3, line 95: is this testing clock skew, or working around it?

bryceitoc9: It works around it for the tests just to make them clearer. The intent with the tests is to drive home some absolutes, while in reality it'll be a lot less likely that 5 seconds will make a huge difference for a user.

src/shared/utilities/textUtilities.ts
```diff
@@ -166,7 +165,73 @@ export function toTitleCase(str: string): string {
...
+ * e.g. "in 1 minute", '1 minute ago'
+ * works on the scales of seconds, minutes, hours, days, weeks, months, years
+ * @param from start Date
+ * @param now end Date (default: current time)
+ * @returns string representation of relative date
+ */
 export function getRelativeDate(from: Date, now: Date = new Date()): string {
-    // Prevent clock skew showing future date
-    return moment(from).subtract(5, 'second').from(now)
+    const rtf = new Intl.RelativeTimeFormat('en', { numeric: 'auto', style: 'long' })
```
justinmk3, line 177: In the future if we need to localize, we can parameterize the `'en'`.

justinmk3 (approved):

PR #4163, opened 2023-12-12
Remove moment dependency

bryceitoc9:
Remove moment dependency for localized timestamps
Example (from the S3 item tooltip)
image

## License


src/shared/logger/activation.ts
```diff
@@ -154,13 +153,17 @@ function getLogLevel(): LogLevel {
...
-
-    return `aws_toolkit_${datetime}.log`
+    const now = new Date()
+    // local to machine: use getMonth/Date instead of UTC equivalent
+    // month is offset by 1
+    const m = now.getMonth() + 1 < 10 ? `0${now.getMonth() + 1}` : now.getMonth() + 1
+    const d = now.getDate() < 10 ? `0${now.getDate()}` : now.getDate()
+    const h = now.getHours() < 10 ? `0${now.getHours()}` : now.getHours()
+    const mn = now.getMinutes() < 10 ? `0${now.getMinutes()}` : now.getMinutes()
+    const s = now.getSeconds() < 10 ? `0${now.getSeconds()}` : now.getSeconds()
```
justinmk3, line 163: `padStart(2, '0')` can be used to ensure 2-digit number.

```suggestion
    const m = (now.getMonth() + 1).toString().padStart(2, '0')
    const d = now.getDate().toString().padStart(2, '0')
    const h = now.getHours().toString().padStart(2, '0')
    const mn = now.getMinutes().toString().padStart(2, '0')
    const s = now.getSeconds().toString().padStart(2, '0')
```

justinmk3: can't we use `toISOString()` or something?

src/shared/utilities/textUtilities.ts
```diff
@@ -170,3 +170,27 @@ export function getRelativeDate(from: Date, now: Date = new Date()): string {
...
+        year: 'numeric',
+        month: 'short',
+        day: 'numeric',
+    })
+    const timeFormat = new Intl.DateTimeFormat('en-US', {
+        hour: 'numeric',
+        minute: 'numeric',
+        second: 'numeric',
+        timeZoneName: 'longOffset',
+    })
```
justinmk3, line 193: these can be combined, then the result looks like:

    "Dec 12, 2023, 5:06:05 PM GMT-08:00"


but i'm not sure if we are intentionally avoiding the extra comma in the middle.

```suggestion
    const dateFormat = new Intl.DateTimeFormat('en-US', {
        year: 'numeric',
        month: 'short',
        day: 'numeric',
        hour: 'numeric',
        minute: 'numeric',
        second: 'numeric',
        timeZoneName: 'longOffset',
    })
```

src/shared/logger/activation.ts
bryceitoc9: Do we want that in UTC? I think this one's typically local.

src/shared/utilities/textUtilities.ts
bryceitoc9: If we want 1:1 with what we had before, this drops the comma between the date and time.

src/shared/logger/activation.ts
justinmk3: UTC sounds completely fine for a logfile name.

src/shared/utilities/textUtilities.ts
```diff
@@ -235,3 +235,51 @@ export function getRelativeDate(from: Date, now: Date = new Date()): string {
...
+ * Format for rendering readable dates.
+ *
+ * Same format used in the S3 console, but it's also locale-aware.
+ * This specifically combines a separate date and time format
+ * in order to avoid a comma between the date and time.
+ *
+ * US: Jan 5, 2020 5:30:20 PM GMT-0700
+ * GB: 5 Jan 2020 17:30:20 GMT+0100
+ */
+export function getFormattedLocalizedDate(d: Date = new Date()): string {
```
justinmk3, line 249: Don't need to explain so much in function names. It's often counterproductive. 

```suggestion
export function toLocalizedDate(d: Date = new Date()): string {
```

src/shared/utilities/textUtilities.ts
```diff
@@ -235,3 +235,51 @@ export function getRelativeDate(from: Date, now: Date = new Date()): string {
...
+        minute: 'numeric',
+        second: 'numeric',
+        timeZoneName: 'longOffset',
+    })
+
+    return `${dateFormat.format(d)} ${timeFormat.format(d)}`
+}
+/**
+ * Matches Insights console timestamp, e.g.: 2019-03-04T11:40:08.781-08:00
+ * TODO: Do we want this this verbose? Log stream just shows HH:mm:ss
```
justinmk3, line 266: We could have separate functions, one for the date+time, the other for only time:

```
// HH:mm:ss
formatTimestamp

// 2019-03-04T11:40:08.781-08:00
formatDateTimestamp()
```

bryceitoc9: This is a copy of the original comment. I have no opinion on the matter.

src/shared/utilities/textUtilities.ts
bryceitoc9: I can apply name changes after integ tests run since these shouldn't change the results.

justinmk3 (approved):

PR #4164, opened 2023-12-12
codewhisperer: update scan polling frequency

ctlai95:
## Problem

Security scans currently poll status every 5 seconds.

## Solution

Increase the frequency to every 1 second.

## License


laileni-aws (approved):

justinmk3:
For future reference: the problem and changelog here do not explain the actual _effect_ that is being fixed. Describing the actual symptom is more useful than "value x is y".

PR #4167, opened 2023-12-13
featureDev: Add tests for ensuring telemetry events are being sent

jpinkney-aws:
## Problem
- We have no test coverage for ensuring telemetry events are being sent when an event happens

## Solution
- Add tests

## License

### DIFF
--- a/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
+++ b/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
@@ -8,27 +8,30 @@ import * as assert from 'assert'
 import * as path from 'path'
 import sinon from 'sinon'
 import { waitUntil } from '../../../../shared/utilities/timeoutUtils'
-import { ControllerSetup, createController } from '../../utils'
-import { ChatControllerEventEmitters } from '../../../../amazonqFeatureDev/controllers/chat/controller'
+import { ControllerSetup, createController, createSession } from '../../utils'
 import { FollowUpTypes, createUri } from '../../../../amazonqFeatureDev/types'
 import { Session } from '../../../../amazonqFeatureDev/session/session'
 import { Prompter } from '../../../../shared/ui/prompter'
-import { toFile } from '../../../testUtil'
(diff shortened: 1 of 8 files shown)


src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
```diff
@@ -158,23 +161,84 @@ describe('Controller', () => {
...
+
+            const voteThumbsUpTelemetry = assertTelemetryCurried('amazonq_approachThumbsUp')
+            voteThumbsUpTelemetry({ amazonqConversationId: conversationID, result: 'Succeeded' })
+        })
+
+        it('incoming downvoted message sends telemetry', async () => {
+            await processChatItemVotedMessage('downvote')
+
+            const voteThumbsDownTelemetry = assertTelemetryCurried('amazonq_approachThumbsDown')
+            voteThumbsDownTelemetry({ amazonqConversationId: conversationID, result: 'Succeeded' })
```
justinmk3, line 216: could use `assertTelemetry` directly instead of the "curried" form, because this code just immediately invoked the curried function.

same for the other cases.

I'd prefer if this change is made so that this doesn't get copy-pasted in many places.

src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
```diff
@@ -8,41 +8,42 @@ import * as assert from 'assert'
...
         sinon.restore()
     })
 
     describe('openDiff', async () => {
-        async function openDiff(
-            controllerEventEmitter: ChatControllerEventEmitters,
-            filePath: string,
-            deleted = false
-        ) {
+        const reviewedChangesTelemetry = assertTelemetryCurried('amazonq_isReviewedChanges')
```
justinmk3, line 42: does this mean we can remove `amazonq_isReviewedChanges` from the exceptions list in https://github.com/aws/aws-toolkit-vscode/blob/170588fcf4b6fb7d2a6eda7eed17262fbeb049f8/src/shared/telemetry/exemptMetrics.ts#L17-L19

PR #3726, opened 2023-08-11
ci: new sam cli update causing CI to fail

nkomonen-amazon:
The python 3.10 sam CI tests were failing and blocking release. This downgrades from the current version causing the issue (1.95.0) to the previous version. Now CI should pass when `/runIntegrationTests` is run.

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4170, opened 2023-12-14
SAM: fix failing SAM tests

nkomonen-amazon:
SAM integ tests are failing.

### Failure 1
Before, we previously created https://github.com/aws/aws-toolkit-vscode/pull/3726/files to use a specific sam cli version in our integration tests since the following version caused CI to fail. The version of sam cli we set is very old now.

Now, this PR removes explicitly setting the sam cli version to download during CI and instead grabs the latest version.
Now when running `/runIntegrationTests` the sam tests should stop failing.

### Failure 2
The python codelens tests were failing. This was due to a bug in a newer python extension version.

Now, we lock to the last working version. More info in IDE-12283

## License

### DIFF
--- a/src/testInteg/sam.test.ts
+++ b/src/testInteg/sam.test.ts
@@ -322,8 +322,7 @@ async function startDebugger(
             try {
                 await vscode.commands.executeCommand('workbench.action.debug.continue')
                 return true
-            }
-            catch {
+            } catch {
                 return false
             }
         },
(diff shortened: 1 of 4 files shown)


src/testInteg/sam.test.ts
```diff
@@ -55,15 +55,6 @@ interface TestScenario {
 // to reduce the chance of automated tests timing out.
 const scenarios: TestScenario[] = [
     // zips
-    {
-        runtime: 'nodejs14.x',
```
nkomonen-amazon, line 59: I see that node14 is deprecated now, this makes sense to delete right?

justinmk3: > Do we need to update anything in `@aws-sdk` in npm?

not directly related

justinmk3 (approved):
Looks great!

PR #4171, opened 2023-12-14
telemetry: add result field to metric

dhasani23:
## Problem

@nkomonen-amazon reported an issue where CI was failing due to failure to include the result field in one of our metric emissions.

## Solution

Add the result field.

## License


justinmk3:
Specifically, the integ test failure was:

```
  1) transformByQ
       "before all" hook for "WHEN upload payload with missing sha256 in headers THEN fails to upload":
     Error: Metric `codeTransform_jobCreateZipEndTime` was emitted without the `result` property. This property is always required. Consider using `.run()` instead of `.emit()`, which will set these properties automatically. See https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md#guidelines
```

PR #4176, opened 2023-12-15
feat(logs): "Watch Logs" command

justinmk3:
Problem:
- Not easy for non-developers to observe the telemetry as it happens.
- Telemetry logs could be much more readable with some formatting.

Solution:
- Use new vscode API [LogOutputChannel](https://github.com/microsoft/vscode/blob/aa835c938d2518832a934a7e3124fa761c6f50a0/src/vscode-dts/vscode.d.ts#L10766)
- Add `AWS (Developer): Watch Logs` command which shows telemetry in the vscode Debug Console.
  - Implementation detail: Debug Console is used because vscode Output channels don't support filtering. https://github.com/microsoft/vscode/issues/52134
- Cleanup the Metadata field before logging it.
- Format with "%O" instead of JSON stringify.

Before:

    telemetry: emitted metric "aws_expandExplorerNode" -> {"Metadata":[{"Key":"serviceType","Value":"ecs"},{"Key":"awsAccount","Value":"541201481031"},{"Key":"awsRegion","Value":"us-west-2"}],"MetricName":"aws_expandExplorerNode","Value":1,"Unit":"None","Passive":false,"EpochTimestamp":1697147194606}

After:

    2023-10-19 14:22:25 [DEBUG]: telemetry: aws_expandExplorerNode {
      Metadata: {
        serviceType: 'logs',
        awsAccount: '541201481031',
        awsRegion: 'us-west-2'
      },
      Value: 1,
      Unit: 'None',
      Passive: false
    }

## License


damntrecky (approved):
So the summary of changes is:

Telemetry logs will now show up in the "Debug output" and the AWS Toolkit output channel when in DevMode (I assume this means localhost instance?)

And the telemetry logs are now formatted nice too.

justinmk3:
> Telemetry logs will now show up in the "Debug output" and the AWS Toolkit output channel when in DevMode

Only if you run the command which was added in this PR.

PR #4180, opened 2023-12-15
telemetry: add code transform data byte size and result field to metrics

tincheng:
## Problem
1. Code Transform artifact upload and download metrics for VSCode did not have totalByteSize.
2. Need to emit dependenciesCopied metric.
3. Several Code Transform metrics did not have `result` field.

## Solution
1. Include totalByteSize where applicable.
2. Emit dependenciesCopied metric.
3. Add `result` field to metrics.

Validated dependenciesCopied emitted:

- `telemetry: emitted metric "codeTransform_dependenciesCopied" -> {"Metadata":[{"Key":"parentMetric","Value":"vscode_executeCommand"},{"Key":"codeTransformSessionId","Value":"<redacted>"},{"Key":"result","Value":"Succeeded"},{"Key":"awsAccount","Value":"not-set"},{"Key":"awsRegion","Value":"us-west-2"}],"MetricName":"codeTransform_dependenciesCopied","Value":1,"Unit":"None","Passive":false,"EpochTimestamp":1702660044892}`

Validated metrics are emitted with expected TotalByteSize:

- `telemetry: emitted metric "codeTransform_jobCreateZipEndTime" -> {"Metadata":[{"Key":"parentMetric","Value":"vscode_executeCommand"},{"Key":"codeTransformSessionId","Value":"<redacted>"},{"Key":"codeTransformTotalByteSize","Value":"2522725"},{"Key":"codeTransformRunTimeLatency","Value":"11208"},{"Key":"result","Value":"Succeeded"},{"Key":"awsAccount","Value":"not-set"},{"Key":"awsRegion","Value":"us-west-2"}],"MetricName":"codeTransform_jobCreateZipEndTime","Value":1,"Unit":"None","Passive":false,"EpochTimestamp":1702660047014}`
- `telemetry: emitted metric "codeTransform_logApiLatency" -> {"Metadata":[{"Key":"parentMetric","Value":"vscode_executeCommand"},{"Key":"codeTransformApiNames","Value":"UploadZip"},{"Key":"codeTransformSessionId","Value":"<redacted>"},
{"Key":"codeTransformUploadId","Value":"<redacted>"},{"Key":"codeTransformRunTimeLatency","Value":"5667"},{"Key":"codeTransformTotalByteSize","Value":"2522725"},{"Key":"result","Value":"Succeeded"},{"Key":"awsAccount","Value":"not-set"},{"Key":"awsRegion","Value":"us-west-2"}],"MetricName":"codeTransform_logApiLatency","Value":1,"Unit":"None","Passive":true,"EpochTimestamp":1702660057297}`
- `telemetry: emitted metric "codeTransform_jobArtifactDownloadAndDeserializeTime" -> {"Metadata":[{"Key":"codeTransformSessionId","Value":"<redacted>"},{"Key":"codeTransformJobId","Value":"<redacted>"},{"Key":"codeTransformRunTimeLatency","Value":"70"},{"Key":"codeTransformTotalByteSize","Value":"152486"},{"Key":"result","Value":"Succeeded"},{"Key":"awsAccount","Value":"not-set"},{"Key":"awsRegion","Value":"us-west-2"}],"MetricName":"codeTransform_jobArtifactDownloadAndDeserializeTime","Value":1,"Unit":"None","Passive":false,"EpochTimestamp":1702660482671}`
- `telemetry: emitted metric "codeTransform_logApiLatency" -> {"Metadata":[{"Key":"codeTransformApiNames","Value":"ExportResultArchive"},{"Key":"codeTransformSessionId","Value":"<redacted>"},{"Key":"codeTransformJobId","Value":"<redacted>"},{"Key":"codeTransformRunTimeLatency","Value":"1133"},{"Key":"codeTransformTotalByteSize","Value":"152486"},{"Key":"codeTransformRequestId","Value":"<redacted>"},{"Key":"awsAccount","Value":"not-set"},{"Key":"awsRegion","Value":"us-west-2"}],"MetricName":"codeTransform_logApiLatency","Value":1,"Unit":"None","Passive":true,"EpochTimestamp":1702660482600}`

## License

### DIFF
--- a/.changes/next-release/Feature-20c6488b-1750-4ca6-9eb3-b509f49f194b.json
+++ b/.changes/next-release/Feature-20c6488b-1750-4ca6-9eb3-b509f49f194b.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Emit additional metadata in telemetry"
+}
--- a/.changes/next-release/Feature-20c6488b-1750-4ca6-9eb3-b509f49f194b.json
+++ b/.changes/next-release/Feature-20c6488b-1750-4ca6-9eb3-b509f49f194b.json
@@ -1,4 +1,4 @@
 {
 	"type": "Feature",
-	"description": "Emit additional metadata in telemetry"
+	"description": "Code Transform Telemetry includes additional metadata"
 }
--- a/src/codewhisperer/commands/startTransformByQ.ts
+++ b/src/codewhisperer/commands/startTransformByQ.ts
@@ -37,6 +37,7 @@ import { codeTransformTelemetryState } from '../../amazonqGumby/telemetry/codeTr
 import { ToolkitError } from '../../shared/errors'
 import { TransformByQUploadArchiveFailed } from '../../amazonqGumby/models/model'
 import { CancelActionPositions, toJDKMetricValue } from '../../amazonqGumby/telemetry/codeTransformTelemetry'
+import { MetadataResult } from '../../shared/telemetry/telemetryClient'
 
 const localize = nls.loadMessageBundle()
 export const stopTransformByQButton = localize('aws.codewhisperer.stop.transform.by.q', 'Stop')
@@ -122,11 +123,13 @@ export async function startTransformByQ() {
     if (selection !== 'Transform') {
         telemetry.codeTransform_jobIsCanceledFromUserPopupClick.emit({
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+            result: MetadataResult.Pass,
         })
         throw new ToolkitError('Transform cancelled', { code: 'DidNotConfirmDisclaimer', cancelled: true })
     } else {
         telemetry.codeTransform_jobIsStartedFromUserPopupClick.emit({
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+            result: MetadataResult.Pass,
         })
     }
 
@@ -146,6 +149,7 @@ export async function startTransformByQ() {
         codeTransformJavaTargetVersionsAllowed: toJDKMetricValue(
             transformByQState.getTargetJDKVersion()
         ) as CodeTransformJavaTargetVersionsAllowed,
+        result: MetadataResult.Pass,
     })
 
     vscode.commands.executeCommand('workbench.view.extension.aws-codewhisperer-transformation-hub')
@@ -347,6 +351,7 @@ export async function confirmStopTransformByQ(
         telemetry.codeTransform_jobIsCancelledByUser.emit({
             codeTransformCancelSrcComponents: cancelSrc,
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+            result: MetadataResult.Pass,
         })
     }
 }
--- a/src/codewhisperer/service/transformByQHandler.ts
+++ b/src/codewhisperer/service/transformByQHandler.ts
@@ -89,71 +89,52 @@ export async function validateProjectSelection(project: vscode.QuickPickItem) {
         '**/node_modules/**',
         1
     )
-    if (compiledJavaFiles.length < 1) {
-        vscode.window.showErrorMessage(CodeWhispererConstants.noSupportedJavaProjectsFoundMessage, { modal: true })
-        telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit({
-            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
-            codeTransformPreValidationError: 'NoJavaProject',
-            result: MetadataResult.Pass,
-        })
-        throw new TransformByQJavaProjectNotFound()
-    }
-    const classFilePath = compiledJavaFiles[0].fsPath
-    const baseCommand = 'javap'
-    const args = ['-v', classFilePath]
-    const spawnResult = spawnSync(baseCommand, args, { shell: false, encoding: 'utf-8' })
-
-    if (spawnResult.error || spawnResult.status !== 0) {
-        vscode.window.showErrorMessage(CodeWhispererConstants.noSupportedJavaProjectsFoundMessage, { modal: true })
-        telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit({
-            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
-            codeTransformPreValidationError: 'NoJavaProject',
-            result: MetadataResult.Pass,
-        })
-        throw new ToolkitError('Unable to determine Java version', {
-            code: 'CannotDetermineJavaVersion',
-            cause: spawnResult.error,
-        })
-    }
-    const majorVersionIndex = spawnResult.stdout.indexOf('major version: ')
-    const javaVersion = spawnResult.stdout.slice(majorVersionIndex + 15, majorVersionIndex + 17).trim()
-    if (javaVersion === CodeWhispererConstants.JDK8VersionNumber) {
-        transformByQState.setSourceJDKVersionToJDK8()
-    } else if (javaVersion === CodeWhispererConstants.JDK11VersionNumber) {
-        transformByQState.setSourceJDKVersionToJDK11()
-    } else {
-        vscode.window.showErrorMessage(CodeWhispererConstants.noSupportedJavaProjectsFoundMessage, { modal: true })
-        telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit({
-            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
-            codeTransformPreValidationError: 'UnsupportedJavaVersion',
-            result: MetadataResult.Pass,
-            reason: javaVersion,
-        })
-        throw new ToolkitError('Project selected is not Java 8 or Java 11', { code: 'UnsupportedJavaVersion' })
-    }
     const buildFile = await vscode.workspace.findFiles(
         new vscode.RelativePattern(projectPath!, 'pom.xml'), // check for pom.xml in root directory only
         '**/node_modules/**',
         1
     )
-    if (buildFile.length < 1) {
-        const buildType = await checkIfGradle(projectPath!)
-        vscode.window.showErrorMessage(CodeWhispererConstants.noPomXmlFoundMessage, { modal: true })
-        if (buildType === 'Gradle') {
-            telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit({
-                codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
-                codeTransformPreValidationError: 'NonMavenProject',
-                result: MetadataResult.Pass,
-                reason: buildType,
+    telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.run(() => {
+        telemetry.record({ codeTransformSessionId: codeTransformTelemetryState.getSessionId() })
+        if (compiledJavaFiles.length < 1) {
+            vscode.window.showErrorMessage(CodeWhispererConstants.noSupportedJavaProjectsFoundMessage, { modal: true })
+            telemetry.record({ codeTransformPreValidationError: 'NoJavaProject' })
+            throw new TransformByQJavaProjectNotFound()
+        }
+        const classFilePath = compiledJavaFiles[0].fsPath
+        const baseCommand = 'javap'
+        const args = ['-v', classFilePath]
+        const spawnResult = spawnSync(baseCommand, args, { shell: false, encoding: 'utf-8' })
+
+        if (spawnResult.error || spawnResult.status !== 0) {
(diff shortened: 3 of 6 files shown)


src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -122,11 +123,13 @@ export async function startTransformByQ() {
     if (selection !== 'Transform') {
         telemetry.codeTransform_jobIsCanceledFromUserPopupClick.emit({
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+            result: MetadataResult.Pass,
```
dhasani23, line 126: I actually have a PR which is about to go out which sets the result and reason fields, so can you omit these changes from this PR? Here, for example, result should be `MetadataResult.Fail`, and we should set a reason field -- like I said my PR does all of this so we can omit these types of changes.

tincheng: I think this particular one should be MetadataResult.Pass since the operation itself is not an error right? User just decided to cancel the job from pop up, which is an acceptable behavior and shouldn't be tracked as a failure (e.g. we shouldn't be alarmed by it). 

I can remove these from the PR.

dhasani23: I see your point! I initially had a different interpretation but I think we can go with this logic. In that case yeah let's just keep this.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -208,6 +208,7 @@ export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrl
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
             codeTransformApiErrorMessage: errorMessage,
             codeTransformRequestId: e?.requestId,
+            result: MetadataResult.Fail,
```
dhasani23, line 211: Need to add `reason` field when result is `Fail`

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -375,6 +380,10 @@ export async function zipCode(modulePath: string) {
             zip.addLocalFile(file, path.dirname(paddedPath))
         }
         zipManifest.dependenciesRoot += `${dependencyFolderName}/`
+        telemetry.codeTransform_dependenciesCopied.emit({
+            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
```
dhasani23, line 384: https://github.com/aws/aws-toolkit-common/blob/main/telemetry/definitions/commonDefinitions.json
Was this added here?

src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -315,9 +320,9 @@ export class ProposedTransformationExplorer {
                     codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
```
dhasani23, line 320: Should move the getLogger() line to be inside the catch block

src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -315,9 +320,9 @@ export class ProposedTransformationExplorer {
...
                     codeTransformJobId: transformByQState.getJobId(),
                     codeTransformRunTimeLatency: calculateTotalLatency(deserializeArchiveStartTime),
-                    // TODO: A nice to have would be getting the zip download size
-                    codeTransformTotalByteSize: 0,
+                    codeTransformTotalByteSize: exportResultsArchiveSize,
                     codeTransformRuntimeError: deserializeErrorMessage,
+                    result: MetadataResult.Pass,
```
dhasani23, line 325: Since this is in the finally block, it will always execute so should make this something like:

                    result: deserializeErrorMessage ? MetadataResult.Fail : MetadataResult.Pass,
                    reason: deserializeErrorMessage ? 'DeserializationFailed' : undefined

src/codewhisperer/service/transformationResultsViewProvider.ts
dhasani23: For some reason it's not letting me comment right next to that line

tincheng: I agree it should be in the catch block. Will update

.changes/next-release/Feature-20c6488b-1750-4ca6-9eb3-b509f49f194b.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Emit additional metadata in telemetry"
```
justinmk3, line 3: please read https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#changelog

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -94,7 +94,7 @@ export async function validateProjectSelection(project: vscode.QuickPickItem) {
         telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit({
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
             codeTransformPreValidationError: 'NoJavaProject',
-            result: MetadataResult.Fail,
+            result: MetadataResult.Pass,
         })
```
justinmk3, line 98: All 5+ of these redundant `codeTransform_isDoubleClickedToTriggerInvalidProject.emit` calls in this function, could be replaced by one `codeTransform_isDoubleClickedToTriggerInvalidProject.run(...)` call. And whereever you need to set a field, simply call record(). E.g. this entire call would instead just be:

    telemetry.record({ codeTransformPreValidationError: 'NoJavaProject' })

.changes/next-release/Feature-20c6488b-1750-4ca6-9eb3-b509f49f194b.json
tincheng: Do you mean changing the description to something like "Code Transform Telemetry includes additional metadata"? Or are you referring to the part "run `npm run newChange` for each change", and that we should include additional lines like the below?
```
{
	"type": "Feature",
	"description": "Code Transform Telemetry includes additional metadata"
},
{
	"type": "Fix",
	"description": "Code Transform Telemetry missing result metadata"
}
```

src/codewhisperer/service/transformByQHandler.ts
tincheng: @justinmk3 I noticed that once I wrapped the function in run(), a "successful" isDoubleClickedToTriggerInvalidProject metric will always be emitted at the end even if project is valid. 

```
telemetry: emitted metric "codeTransform_isDoubleClickedToTriggerInvalidProject" -> {"Metadata":[{"Key":"missingFields","Value":"codeTransformPreValidationError"},{"Key":"result","Value":"Succeeded"},"MetricName":"codeTransform_isDoubleClickedToTriggerInvalidProject","Value":1}
```

I don't think this is the desired behavior, as we only want to receive counts of isDoubleClickedToTriggerInvalidProject metric when the project is actually invalid. Please let me know your thoughts, or I will change it back to independent emits.

tincheng: Reverted. Will use emit() since we do not want to conditionally emit no metric.

.changes/next-release/Feature-20c6488b-1750-4ca6-9eb3-b509f49f194b.json
tincheng: Updated change log message

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -94,7 +94,7 @@ export async function validateProjectSelection(project: vscode.QuickPickItem) {
         telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit({
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
             codeTransformPreValidationError: 'NoJavaProject',
```
damntrecky, line 96: Question: Should be a failure metric because although its a valid state, its something we would not like users have happen to them.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -393,8 +405,7 @@ export async function zipCode(modulePath: string) {
     const mavenStatus = mavenFailed ? MetadataResult.Fail : MetadataResult.Pass
     telemetry.codeTransform_jobCreateZipEndTime.emit({
         codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
-        // TODO: A nice to have would be getting the zipUploadSize
-        codeTransformTotalByteSize: 0,
+        codeTransformTotalByteSize: (await fs.promises.stat(tempFilePath)).size,
```
damntrecky, line 408: Not a blocker: This is the size of the zip folder. I believe for IntelliJ we are logging the byte size pre-zipped.

It will still help us determine overall size of upload/download

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -196,8 +196,8 @@ export async function uploadArtifactToS3(fileName: string, resp: CreateUploadUrl
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
             codeTransformUploadId: resp.uploadId,
             codeTransformRunTimeLatency: calculateTotalLatency(apiStartTime),
-            // TODO: A nice to have would be getting the zipUploadSize
-            codeTransformTotalByteSize: 0,
+            codeTransformTotalByteSize: (await fs.promises.stat(fileName)).size,
```
damntrecky, line 199: Not a blocker: This is the size of the zip folder. I believe for IntelliJ we are logging the byte size pre-zipped.

It will still help us determine overall size of upload/download

src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -286,10 +287,14 @@ export class ProposedTransformationExplorer {
                     codeTransformJobId: transformByQState.getJobId(),
                     codeTransformApiErrorMessage: e?.message || errorMessage,
                     codeTransformRequestId: e?.requestId,
+                    result: MetadataResult.Fail,
+                    reason: 'ExportResultArchiveFailed',
                 })
                 throw new ToolkitError(errorMessage)
             }
 
+            const exportResultsArchiveSize = (await fs.promises.stat(pathToArchive)).size
```
damntrecky, line 296: Not a blocker: This is the size of the zip folder. I believe for IntelliJ we are logging the byte size pre-zipped.

It will still help us determine overall size of upload/download

damntrecky: What we could do is have IntelliJ measure zip size to match VSCode. Since looping through the files and calling `fs.stat` is the reason we did not implement this.

damntrecky (approved):
Approved. Please respond with you opinion on some of the comments.

src/codewhisperer/service/transformByQHandler.ts
tincheng: Aligned offline, will use zip size.

src/codewhisperer/service/transformByQHandler.ts
tincheng: Aligned offline, will use zip size.

src/codewhisperer/service/transformationResultsViewProvider.ts
tincheng: Aligned offline, will use zip size.

src/codewhisperer/service/transformByQHandler.ts
tincheng: Changed back to Fail for all instances of the metric.

justinmk3:
> we only want to receive counts of isDoubleClickedToTriggerInvalidProject metric when the project is actually invalid.

That's why it's generally not recommended to have a metric that is for failures. Metrics should be for tasks--which may succeed or fail. "Invalid project" could have been stored on a higher-level "trigger" action. Oh well.

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -122,11 +123,13 @@ export async function startTransformByQ() {
     if (selection !== 'Transform') {
         telemetry.codeTransform_jobIsCanceledFromUserPopupClick.emit({
```
justinmk3, line 124: Here's another metric that doesn't make sense. `Result=Canceled` is a standard field that could be stored on a higher-level "job attempted" metric. Instead, now we have this special-case "job is canceled" metric :(

justinmk3 (approved):
sigh

src/codewhisperer/commands/startTransformByQ.ts
tincheng: I see what you mean, and it does provide advantages to calculating failure rate. So `jobIsCanceledFromUserPopupClick` will be `UserDialogPopupClicked`, and `isDoubleClickedToTriggerInvalidProject` will be `ProjectValidationTriggered`. We will need to revisit these metrics for both IDEs at the same time.

PR #4181, opened 2023-12-15
fix(codewhisperer): support code generation for IaC languages without external plugin extensions

laileni-aws:
## Problem
CodeWhisperer needs terraform plugin extension to generate code suggestions for `tf` and `hcl` files, as `tf` and `hcl` files are recognized as `plaintext` by VSCode. 
## Solution

- This issue can be resolved by categorizing the files based on file format/ file extensions instead of depending on editor document language.
- Code generation will be independent of external plugin extensions.

## License

### DIFF
--- a/.changes/next-release/Bug Fix-ec18555e-d716-40b5-b8d1-54dad851fdc2.json
+++ b/.changes/next-release/Bug Fix-ec18555e-d716-40b5-b8d1-54dad851fdc2.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "CodeWhisperer supports code generation for Iaac languages without external extensions"
+}
--- a/.changes/next-release/Bug Fix-ec18555e-d716-40b5-b8d1-54dad851fdc2.json
+++ b/.changes/next-release/Bug Fix-ec18555e-d716-40b5-b8d1-54dad851fdc2.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "CodeWhisperer supports code generation for Iaac languages without external extensions"
+	"description": "CodeWhisperer supports code generation for Iac languages without external extensions"
 }
--- a/src/codewhisperer/commands/onAcceptance.ts
+++ b/src/codewhisperer/commands/onAcceptance.ts
@@ -14,6 +14,7 @@ import { handleExtraBrackets } from '../util/closingBracketUtil'
 import { RecommendationHandler } from '../service/recommendationHandler'
 import { ReferenceLogViewProvider } from '../service/referenceLogViewProvider'
 import { ReferenceHoverProvider } from '../service/referenceHoverProvider'
+import { getFileExtension } from '../util/editorContext'
 
 /**
  * This function is called when user accepts a intelliSense suggestion or an inline suggestion
@@ -26,9 +27,7 @@ export async function onAcceptance(acceptanceEntry: OnRecommendationAcceptanceEn
     if (acceptanceEntry.editor) {
         const languageContext = runtimeLanguageContext.getLanguageContext(
             acceptanceEntry.editor.document.languageId,
-            acceptanceEntry.editor.document.fileName.substring(
-                acceptanceEntry.editor.document.fileName.lastIndexOf('.') + 1
-            )
+            getFileExtension(acceptanceEntry.editor.document)
         )
         const start = acceptanceEntry.range.start
         const end = isCloud9() ? acceptanceEntry.editor.selection.active : acceptanceEntry.range.end
--- a/src/codewhisperer/commands/onAcceptance.ts
+++ b/src/codewhisperer/commands/onAcceptance.ts
@@ -24,7 +24,12 @@ export async function onAcceptance(acceptanceEntry: OnRecommendationAcceptanceEn
      * Format document
      */
     if (acceptanceEntry.editor) {
-        const languageContext = runtimeLanguageContext.getLanguageContext(acceptanceEntry.editor.document.languageId)
+        const languageContext = runtimeLanguageContext.getLanguageContext(
+            acceptanceEntry.editor.document.languageId,
+            acceptanceEntry.editor.document.fileName.substring(
+                acceptanceEntry.editor.document.fileName.lastIndexOf('.') + 1
+            )
+        )
         const start = acceptanceEntry.range.start
         const end = isCloud9() ? acceptanceEntry.editor.selection.active : acceptanceEntry.range.end
 
--- a/src/codewhisperer/commands/onAcceptance.ts
+++ b/src/codewhisperer/commands/onAcceptance.ts
@@ -14,7 +14,7 @@ import { handleExtraBrackets } from '../util/closingBracketUtil'
 import { RecommendationHandler } from '../service/recommendationHandler'
 import { ReferenceLogViewProvider } from '../service/referenceLogViewProvider'
 import { ReferenceHoverProvider } from '../service/referenceHoverProvider'
-import { getFileExtension } from '../util/editorContext'
+import path from 'path'
 
 /**
  * This function is called when user accepts a intelliSense suggestion or an inline suggestion
@@ -27,7 +27,7 @@ export async function onAcceptance(acceptanceEntry: OnRecommendationAcceptanceEn
     if (acceptanceEntry.editor) {
         const languageContext = runtimeLanguageContext.getLanguageContext(
             acceptanceEntry.editor.document.languageId,
-            getFileExtension(acceptanceEntry.editor.document)
+            path.extname(acceptanceEntry.editor.document.fileName)
         )
         const start = acceptanceEntry.range.start
         const end = isCloud9() ? acceptanceEntry.editor.selection.active : acceptanceEntry.range.end
--- a/src/codewhisperer/commands/onInlineAcceptance.ts
+++ b/src/codewhisperer/commands/onInlineAcceptance.ts
@@ -27,7 +27,7 @@ import { ReferenceLogViewProvider } from '../service/referenceLogViewProvider'
 import { ReferenceHoverProvider } from '../service/referenceHoverProvider'
 import { ImportAdderProvider } from '../service/importAdderProvider'
 import { session } from '../util/codeWhispererSession'
-import { getFileExtension } from '../util/editorContext'
+import path from 'path'
 
 export const acceptSuggestion = Commands.declare(
     'aws.codeWhisperer.accept',
@@ -76,7 +76,7 @@ export async function onInlineAcceptance(
         await sleep(CodeWhispererConstants.vsCodeCursorUpdateDelay)
         const languageContext = runtimeLanguageContext.getLanguageContext(
             acceptanceEntry.editor.document.languageId,
-            getFileExtension(acceptanceEntry.editor.document)
+            path.extname(acceptanceEntry.editor.document.fileName)
         )
         const start = acceptanceEntry.range.start
         const end = acceptanceEntry.editor.selection.active
--- a/src/codewhisperer/commands/onInlineAcceptance.ts
+++ b/src/codewhisperer/commands/onInlineAcceptance.ts
@@ -73,7 +73,12 @@ export async function onInlineAcceptance(
 
     if (acceptanceEntry.editor) {
         await sleep(CodeWhispererConstants.vsCodeCursorUpdateDelay)
-        const languageContext = runtimeLanguageContext.getLanguageContext(acceptanceEntry.editor.document.languageId)
+        const languageContext = runtimeLanguageContext.getLanguageContext(
+            acceptanceEntry.editor.document.languageId,
+            acceptanceEntry.editor.document.fileName.substring(
+                acceptanceEntry.editor.document.fileName.lastIndexOf('.') + 1
+            )
+        )
         const start = acceptanceEntry.range.start
         const end = acceptanceEntry.editor.selection.active
 
--- a/src/codewhisperer/commands/onInlineAcceptance.ts
+++ b/src/codewhisperer/commands/onInlineAcceptance.ts
@@ -27,6 +27,7 @@ import { ReferenceLogViewProvider } from '../service/referenceLogViewProvider'
 import { ReferenceHoverProvider } from '../service/referenceHoverProvider'
 import { ImportAdderProvider } from '../service/importAdderProvider'
 import { session } from '../util/codeWhispererSession'
+import { getFileExtension } from '../util/editorContext'
 
 export const acceptSuggestion = Commands.declare(
     'aws.codeWhisperer.accept',
@@ -75,9 +76,7 @@ export async function onInlineAcceptance(
         await sleep(CodeWhispererConstants.vsCodeCursorUpdateDelay)
         const languageContext = runtimeLanguageContext.getLanguageContext(
             acceptanceEntry.editor.document.languageId,
-            acceptanceEntry.editor.document.fileName.substring(
-                acceptanceEntry.editor.document.fileName.lastIndexOf('.') + 1
-            )
+            getFileExtension(acceptanceEntry.editor.document)
         )
         const start = acceptanceEntry.range.start
         const end = acceptanceEntry.editor.selection.active
--- a/src/codewhisperer/service/recommendationHandler.ts
+++ b/src/codewhisperer/service/recommendationHandler.ts
@@ -44,6 +44,7 @@ import { application } from '../util/codeWhispererApplication'
 import { openUrl } from '../../shared/utilities/vsCodeUtils'
 import { indent } from '../../shared/utilities/textUtilities'
 import { getFileExtension } from '../util/editorContext'
+import path from 'path'
 
 /**
  * This class is for getRecommendation/listRecommendation API calls and its states
@@ -405,7 +406,7 @@ export class RecommendationHandler {
                     page,
                     runtimeLanguageContext.getLanguageContext(
                         editor.document.languageId,
-                        editor.document.fileName.substring(editor.document.fileName.lastIndexOf('.') + 1)
+                        path.extname(editor.document.fileName)
                     ).language,
                     session.requestContext.supplementalMetadata
                 )
--- a/src/codewhisperer/service/recommendationHandler.ts
+++ b/src/codewhisperer/service/recommendationHandler.ts
@@ -180,7 +180,10 @@ export class RecommendationHandler {
         let latency = 0
         let nextToken = ''
         let shouldRecordServiceInvocation = true
-        session.language = runtimeLanguageContext.getLanguageContext(editor.document.languageId).language
+        session.language = runtimeLanguageContext.getLanguageContext(
+            editor.document.languageId,
+            editor.document.fileName.substring(editor.document.fileName.lastIndexOf('.') + 1)
+        ).language
         session.taskType = await this.getTaskTypeFromEditorFileName(editor.document.fileName)
 
         if (pagination) {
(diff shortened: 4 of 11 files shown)


src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -161,7 +161,22 @@ export class RuntimeLanguageContext {
      * @param languageId : vscode language id or codewhisperer language name
```
ctlai95, line 161: please update this jsdoc with the new parameter

src/codewhisperer/commands/onAcceptance.ts
```diff
@@ -24,7 +24,12 @@ export async function onAcceptance(acceptanceEntry: OnRecommendationAcceptanceEn
      * Format document
      */
     if (acceptanceEntry.editor) {
-        const languageContext = runtimeLanguageContext.getLanguageContext(acceptanceEntry.editor.document.languageId)
+        const languageContext = runtimeLanguageContext.getLanguageContext(
+            acceptanceEntry.editor.document.languageId,
+            acceptanceEntry.editor.document.fileName.substring(
+                acceptanceEntry.editor.document.fileName.lastIndexOf('.') + 1
+            )
```
ctlai95, line 31: Looks like this is repeated in multiple places, we should probably add a utility function, `getFileExtension()` if we don't have one already

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -161,7 +161,22 @@ export class RuntimeLanguageContext {
...
      * @returns An object with a field language: CodewhispererLanguage, if no corresponding CodewhispererLanguage ID, plaintext is returned
      */
-    public getLanguageContext(languageId?: string): { language: CodewhispererLanguage } {
+    public getLanguageContext(languageId?: string, filename?: string): { language: CodewhispererLanguage } {
```
ctlai95, line 164: I think instead of `filename` it should be `extension`?

src/test/codewhisperer/util/runtimeLanguageContext.test.ts
```diff
@@ -281,8 +281,14 @@ describe('runtimeLanguageContext', function () {
...
             ['sql', 'sql'],
+            ['sql', 'sql'],
+            ['tf', 'tf'],
+            ['hcl', 'tf'],
+            ['json', 'json'],
+            ['yaml', 'yaml'],
+            ['yml', 'yaml'],
             ['plaintext', 'plaintext'],
-            ['arbitrary string', 'arbitrary string'],
+            // ['arbitrary string', 'arbitrary string'],
```
ctlai95, line 290: Can this be removed?

src/codewhisperer/util/editorContext.ts
```diff
@@ -185,7 +185,12 @@ export function validateRequest(
         req.fileContext.programmingLanguage.languageName === undefined ||
         req.fileContext.programmingLanguage.languageName.length < 1 ||
         req.fileContext.programmingLanguage.languageName.length > 128 ||
-        !runtimeLanguageContext.isLanguageSupported(req.fileContext.programmingLanguage.languageName)
+        !(
+            runtimeLanguageContext.isLanguageSupported(req.fileContext.programmingLanguage.languageName) ||
+            runtimeLanguageContext.isFileFormatSupported(
+                req.fileContext.filename.substring(req.fileContext.filename.lastIndexOf('.') + 1)
+            )
+        )
```
ctlai95, line 193: Can we split this up? it's a bit hard to read with the nested `!(...)`

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -161,7 +161,22 @@ export class RuntimeLanguageContext {
...
+        const extensionToLanguageMap: Record<string, CodewhispererLanguage> = {
+            tf: 'tf',
+            hcl: 'tf',
+            json: 'json',
+            yaml: 'yaml',
+            yml: 'yaml',
+            // Add more mappings if needed
```
ctlai95, line 171: can we add tsx/jsx also?

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -193,12 +211,21 @@ export class RuntimeLanguageContext {
...
+     * @returns true if the language is supported by CodeWhisperer otherwise false
      */
     public isLanguageSupported(languageId: string): boolean {
         const lang = this.normalizeLanguage(languageId)
         return lang !== undefined && this.normalizeLanguage(languageId) !== 'plaintext'
     }
+    /**
+     *
+     * @param fileFormat : vscode editor filecontext filename extension
+     * @returns  if the fileformat is supported by CodeWhisperer otherwise false
```
ctlai95, line 223:
```suggestion
     * @returns true if the fileformat is supported by CodeWhisperer otherwise false
```

src/codewhisperer/util/runtimeLanguageContext.ts
laileni-aws: We can add `tsx`/`jsx` but the editor is not considering them as `plaintext`, If I am not wrong so if thats the case we do not see any difference even if we add them to this.

If editor recognizes `tsx` and `jsx` as `plaintext` then we can add them as `jsx`to`javascript` and `tsx` to `typescript`

src/codewhisperer/util/runtimeLanguageContext.ts
laileni-aws: Updated !

src/test/codewhisperer/util/runtimeLanguageContext.test.ts
laileni-aws: I need @Will-ShaoHua input on this, Can this be removed from the test case as we are testing for all the other use cases.

src/codewhisperer/util/editorContext.ts
laileni-aws: Yeah sure, I will apply` ! `negation to inside conditions which should clear the complexity.
```
    const isLanguageNameValid = (
        req.fileContext.programmingLanguage.languageName !== undefined &&
        req.fileContext.programmingLanguage.languageName.length >= 1 &&
        req.fileContext.programmingLanguage.languageName.length <= 128 &&
        (
            runtimeLanguageContext.isLanguageSupported(req.fileContext.programmingLanguage.languageName) ||
            runtimeLanguageContext.isFileFormatSupported(
                req.fileContext.filename.substring(req.fileContext.filename.lastIndexOf('.') + 1)
            )
        )
    )
```

Finally applying some logical ops of maths in coding..!

src/codewhisperer/util/runtimeLanguageContext.ts
laileni-aws: Updated!

src/codewhisperer/commands/onAcceptance.ts
laileni-aws: Updated!

ashishrp-aws (approved):

.changes/next-release/Bug Fix-ec18555e-d716-40b5-b8d1-54dad851fdc2.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "CodeWhisperer supports code generation for Iaac languages without external extensions"
```
justinmk3, line 3: should this be IaC? what does Iaac stand for?
```suggestion
	"description": "CodeWhisperer supports code generation for IaC languages without external extensions"
```

src/codewhisperer/commands/onInlineAcceptance.ts
```diff
@@ -73,7 +74,10 @@ export async function onInlineAcceptance(
 
     if (acceptanceEntry.editor) {
         await sleep(CodeWhispererConstants.vsCodeCursorUpdateDelay)
-        const languageContext = runtimeLanguageContext.getLanguageContext(acceptanceEntry.editor.document.languageId)
+        const languageContext = runtimeLanguageContext.getLanguageContext(
+            acceptanceEntry.editor.document.languageId,
+            getFileExtension(acceptanceEntry.editor.document)
+        )
```
justinmk3, line 80: Looks like all of these calls could benefit if `getLanguageContext` accepted a document (could be an optional overload to support both ways)
```suggestion
        const languageContext = runtimeLanguageContext.getLanguageContext(editor.document)
```

src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -399,7 +403,10 @@ export class RecommendationHandler {
                     session.requestIdList,
                     sessionId,
                     page,
-                    editor.document.languageId,
+                    runtimeLanguageContext.getLanguageContext(
+                        editor.document.languageId,
+                        editor.document.fileName.substring(editor.document.fileName.lastIndexOf('.') + 1)
```
justinmk3, line 408: why not use `path.extname()` ? it is more readable, shorter, and potentially more robust.

justinmk3:
```suggestion
                        path.extname(editor.document.fileName)
```

src/codewhisperer/util/editorContext.ts
```diff
@@ -76,6 +76,10 @@ export function getFileName(editor: vscode.TextEditor): string {
     return fileName.substring(0, CodeWhispererConstants.filenameCharsLimit)
 }
 
+export function getFileExtension(document: vscode.TextDocument): string {
+    return document.fileName.substring(document.fileName.lastIndexOf('.') + 1)
```
justinmk3, line 80: why is this needed instead of `path.extname` ?

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -193,12 +211,21 @@ export class RuntimeLanguageContext {
...
         return lang !== undefined && this.normalizeLanguage(languageId) !== 'plaintext'
     }
+    /**
+     *
+     * @param fileFormat : vscode editor filecontext filename extension
+     * @returns  true if the fileformat is supported by CodeWhisperer otherwise false
+     */
+    public isFileFormatSupported(fileFormat: string): boolean {
+        const fileformat = this.supportedLanguageExtensionMap.get(fileFormat)
+        return fileformat !== undefined && this.supportedLanguageExtensionMap.get(fileformat) !== 'txt'
```
justinmk3, line 227: `this.supportedLanguageExtensionMap.get(fileFormat)` is doing a lookup on the thing it just looked up. is that intentional?

src/codewhisperer/service/recommendationHandler.ts
Will-ShaoHua: +1

src/codewhisperer/util/editorContext.ts
```diff
@@ -181,12 +185,14 @@ export async function buildGenerateRecommendationRequest(editor: vscode.TextEdit
...
-        req.fileContext.programmingLanguage.languageName.length > 128 ||
-        !runtimeLanguageContext.isLanguageSupported(req.fileContext.programmingLanguage.languageName)
-    )
+    const isLanguageNameValid =
+        req.fileContext.programmingLanguage.languageName !== undefined &&
+        req.fileContext.programmingLanguage.languageName.length >= 1 &&
+        req.fileContext.programmingLanguage.languageName.length <= 128 &&
+        (runtimeLanguageContext.isLanguageSupported(req.fileContext.programmingLanguage.languageName) ||
+            runtimeLanguageContext.isFileFormatSupported(
+                req.fileContext.filename.substring(req.fileContext.filename.lastIndexOf('.') + 1)
```
Will-ShaoHua, line 194: I think moving all of these logics into single one method `isLanguageSupported()` is easier to use.

src/codewhisperer/util/runtimeLanguageContext.ts
Will-ShaoHua: as Justin suggested, think making the method signature to be `getLanguageContext(document: vscode.TextDocument)` would be nicer and maybe rename it to be `getLanguageForDocument()`

Will-ShaoHua: actually, can we also make this method returning `CodewhispererLanguage` instead of `{ language: CodewhispererLanguage }`, it's our tech debt and meaningless.

src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -158,10 +158,25 @@ export class RuntimeLanguageContext {
...
+            yaml: 'yaml',
+            yml: 'yaml',
+            // Add more mappings if needed
+        }
+
+        if (languageId === 'plaintext' && fileExtension !== undefined) {
+            const languages = extensionToLanguageMap[fileExtension]
+            if (languages) {
+                return { language: languages }
+            }
```
Will-ShaoHua, line 178:
```
public getLanguageContext(doc: vscode.TextDocument): CodewhispererLanguage {
  return this.normalizeLanguage(doc.languageId) ?? extensionToLanguageMap[fileExt] ?? 'plaintext'
}
```

src/test/codewhisperer/util/runtimeLanguageContext.test.ts
Will-ShaoHua: any reason why we need to comment this out?

Will-ShaoHua: if it's not causing compile error or any error I don't get the reason why we have this change here

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -99,10 +99,9 @@ export class TelemetryHelper {
         requestIdList: string[],
         sessionId: string,
         paginationIndex: number,
-        languageId: string,
+        languageContext: CodewhispererLanguage,
```
Will-ShaoHua, line 102: nit: `language: CodewhispererLanguage`

src/test/codewhisperer/util/runtimeLanguageContext.test.ts
laileni-aws: Can I know what it does ?

src/codewhisperer/util/runtimeLanguageContext.ts
Will-ShaoHua: prefer having single one interface `isLanguageSupported` but we can make the signature to be `isLanguageSupported(documetn: vscode.TextDocument)`

src/test/codewhisperer/util/runtimeLanguageContext.test.ts
Will-ShaoHua: it's just a test case. 
```
public mapToRuntimeLanguage<
        T extends codewhispererClient.ListRecommendationsRequest | codewhispererClient.GenerateRecommendationsRequest
    >(request: T): T
```

type of `T.fileContext.programmingLanguage.languageName` is `string`, we can't gurantee it's something we must know in advance

Will-ShaoHua: of course you can remove it IF we update and make our logic more robust by passing type of `CodewhispererLanguage` or something which will show error at compile time.

Will-ShaoHua: although this method `mapToRuntimeLanguage` is another method we should refactor and remove, it's not covered by this PR so let's keep it as is now

src/codewhisperer/util/editorContext.ts
laileni-aws: Yes, but this is taking a cascade effect of changing 6 files, If Yes, I can change in those files too..

src/codewhisperer/util/runtimeLanguageContext.ts
laileni-aws: > this.supportedLanguageExtensionMap.get(fileFormat) is doing a lookup on the thing it just looked up. is that intentional?

I think we did not check for file extension before 

```
    (runtimeLanguageContext.isLanguageSupported(req.fileContext.programmingLanguage.languageName) ||
             runtimeLanguageContext.isFileFormatSupported(
                 req.fileContext.filename.substring(req.fileContext.filename.lastIndexOf('.') + 1)
             ))
```

In this code we check for language and then we check for file extension if anything matches this is treated as a validLanguage name

src/codewhisperer/util/runtimeLanguageContext.ts
laileni-aws: We can not use `(doc: vscode.TextDocument)` as parameter because this `getLanguageContext` is used in different parts of code, example in this we pass `request: codewhispererClient.ListRecommendationsRequest | codewhispererClient.GenerateRecommendationsRequest` so we can not generalize into document as parameter
```      
                runtimeLanguageContext.getLanguageContext(
                    request.fileContext.programmingLanguage.languageName,
                    request.fileContext.filename.substring(request.fileContext.filename.lastIndexOf('.') + 1)
                ).language
```

src/codewhisperer/util/editorContext.ts
```diff
@@ -76,6 +76,10 @@ export function getFileName(editor: vscode.TextEditor): string {
     return fileName.substring(0, CodeWhispererConstants.filenameCharsLimit)
 }
 
+export function getFileExtension(document: vscode.TextDocument): string {
```
justinmk3, line 79: this function is not needed.

src/codewhisperer/util/runtimeLanguageContext.ts
Will-ShaoHua: correct, we didn't check file ext before, I mean can we add these layer of logic into existing `isLanguageSupport()` method instead exposing a new API? 

```
public isLanguageSupport(document: vscode.TextDocument) {
    const Lang = this.normalizedLanguage(document.languageId)
    if (this.supportedLanguageMap.get(normalizedLang)) {
        return true
    } else {
         // check file ext
    }
}
```

Will-ShaoHua: the reasoning is that basically we will always need to 
1. check language id if present
2. if step 1 fails,  then go check file ext.

in this case we don't really split these 2 into 2 API and callers have to explicitly call these 2 APIs. 1 is enough and should be more error prone IMO.

src/codewhisperer/util/runtimeLanguageContext.ts
Will-ShaoHua: I see.

src/codewhisperer/util/editorContext.ts
Will-ShaoHua: ok make sense, it's something should be fixed with a refactor. In that case it's fine here.

Will-ShaoHua:
LGTM. My comments should mostly be take care in CW refactor PRs instead of this one.

Will-ShaoHua (approved):

src/codewhisperer/util/runtimeLanguageContext.ts
laileni-aws: Yes we can add this to `isLanguageSupport` but as I mentioned below we need more refactoring of the code in different files. Let me take up and do this in this PR itself

src/codewhisperer/commands/onInlineAcceptance.ts
laileni-aws: responded below.

justinmk3: Can't that be solved with an overload?

justinmk3 (approved):

src/codewhisperer/commands/onInlineAcceptance.ts
laileni-aws: `getLanguageContext` needs more refactoring along with other functions `mapToRuntimeLanguage` to do this.

PR #4183, opened 2023-12-16
fix: remove deps, reorder input validation, add metadata to feedback

dhasani23:
## Problem

1. Had some dependencies ('he' and 'proxyquire') which were not strictly needed
2. Had some unnecessary sleep statements
3. Missing jobId and clientId when users leave in-IDE feedback [REMOVED, will address in separate PR]
5. Missing jobId to the table shown to users
6. Unnecessary modals shown when displaying notifications
7. Possible bug when zipping dependencies if a dependency is named "target" (no evidence that this has occurred)

## Solution

1. Write a custom HTML encoder function and use sinon, respectively
2. Remove sleep statements (trade off is that now if the user tries to cancel the job within the first ~10 seconds the upload completes first, then the job is cancelled, but this is satisfactory)
3. Add jobId and clientId to metadata when emitting feedback [REMOVED, will address in separate PR]
4. Add jobId to table
5. Remove `modal: true` for notifications which do not require user action
6. Add an extra check when zipping dependencies folder

## License


src/feedback/vue/submitFeedback.vue
```diff
@@ -71,6 +74,7 @@ export default defineComponent({
             isSubmitting: false,
             error: '',
             feedbackName: '',
+            extraCodeTransformData: { jobId: '', clientId: '' },
```
justinmk3, line 77: why is this needed? it's not used anywhere except below, where it could be a local variable.

this should be a separate PR.

src/feedback/vue/submitFeedback.vue
```diff
@@ -82,9 +86,14 @@ export default defineComponent({
             this.isSubmitting = true
             console.log('Submitting feedback...')
             // identifier to help us (internally) know that feedback came from either CodeWhisperer or AWS Toolkit
+            let userComment =
+                this.feedbackName === 'CodeWhisperer' ? 'CodeWhisperer onboarding: ' + this.comment : this.comment
+            // info to help us (internally) know the jobId and clientId for transform by Q job runs
+            this.extraCodeTransformData['jobId'] = transformByQState.getJobId()
```
justinmk3, line 92: `jobId` could mean anything. This feedback is going into a place where all Toolkit feature goes. 

Can we check `feedbackName == 'AmazonQTransform'` or whatever, so that the `transformByQState` data is not included randomly in all feedback.

src/feedback/vue/submitFeedback.vue
```diff
@@ -82,9 +86,14 @@ export default defineComponent({
...
+            this.extraCodeTransformData['clientId'] = await getClientId(globals.context.globalState)
```
justinmk3, line 93: Feedback is supposed to be anonymous...

src/test/codewhisperer/commands/transformByQ.test.ts
```diff
@@ -70,42 +72,31 @@ describe('transformByQ', function () {
...
+                message: 'No valid Maven build file found',
+            }
+        )
     })
 
-    it('WHEN validateProjectSelection called on project with no class files THEN throws error', async function () {
+    it('WHEN validateProjectSelection called on project with pom.xml but no class files THEN throws error', async function () {
+        const tempDir = os.tmpdir()
+        const dummyPomPath = path.join(tempDir, 'pom.xml')
+        fs.writeFileSync(dummyPomPath, '')
```
justinmk3, line 94: could use `toFile` instead https://github.com/aws/aws-toolkit-vscode/blob/1e7da02a5bea08de4f3bd97e7b2f7b938661a5a5/src/test/testUtil.ts#L26

src/test/codewhisperer/commands/transformByQ.test.ts
```diff
@@ -117,6 +108,7 @@ describe('transformByQ', function () {
                 message: 'No Java projects found',
             }
         )
+        fs.rmSync(dummyPomPath)
```
justinmk3, line 111: no needed if `createTestWorkspaceFolder` is used, it cleans itself up after test suite is finished. https://github.com/aws/aws-toolkit-vscode/blob/1e7da02a5bea08de4f3bd97e7b2f7b938661a5a5/src/test/testUtil.ts#L62

src/feedback/vue/submitFeedback.vue
dhasani23: Will fix and do in separate PR

src/feedback/vue/submitFeedback.vue
dhasani23: Makes sense. I wasn't sure where the feedbackName is set though?

dhasani23: Will ask PR author who implemented the feedbackName functionality

dhasani23: Spoke to @laileni and actually https://github.com/aws/aws-toolkit-vscode/blob/f8606ccab4b209aae0cb5d6aa2bc49e8bbf726ae/src/feedback/vue/submitFeedback.vue#L87 is redundant / doesn’t actually do anything. They removed the “submit feedback” button from their welcome / info page so this no longer does anything

It looks like all feedback submitted through IDE goes through the generic AWS Toolkit page. What I can do is only include the jobId if it is not empty i.e. the user must have clicked Transform.

src/feedback/vue/submitFeedback.vue
dhasani23: I've been told that we are permitted to include this by multiple others. Will follow up on this.

src/feedback/vue/submitFeedback.vue
laileni-aws: As the CW Onboarding page has low engagement with users so, @ankit-sharma90 suggested to remove the content from the page. I think @Will-ShaoHua  did the clean up work for VSC pre re invent. 

Previously, we used have a link "`Provide Feedback`" which uses common feedback component but I used above logic to separate CW f/b with toolkit f/b.
As per my understanding this logic is redundant right now after new CW Getting started page.

justinmk3: > this logic is redundant right now after new CW Getting started page.

The idea is that some features have their own "Send Feedback" action which sends this `feedbackName`. That could be implemented for Transform, or more generally for Q. So when "Send Feedback" is selected from the Q panel and other Q contexts, it sends `feedbackName=AmazonQ`, then we can decide extra logic here.

For reference, look for `feedbackName` in https://github.com/aws/aws-toolkit-vscode/commit/398744e7d465f660067cf073b5ba94dc466fb9a3 . You would need to implement something that like for Transform.

> What I can do is only include the jobId if it is not empty i.e. the user must have clicked Transform.

That's an ok compromise for now.

justinmk3 (changes requested):
`package-lock.json` should not have so many changes.

More generally, there are 3 separate topics here. We'll make faster progress if they become 3 separate PRs, this is starting to turn into a dust cloud.

dhasani23:
Agreed that this is getting messy. Will make 1 PR for general Transform by Q fixes, 1 PR for removing the deps, and 1 PR for adding the jobId to the feedback.

src/feedback/vue/submitFeedback.vue
dhasani23: @damntrecky

damntrecky: If we are calling an API, wont the users credentials be logged when calling whatever feedback API is in place?

For the clientside logging, whats the difference to adding `clientId` which is added in every Telemetry event. The id that matches the users IDE download? Not their AWS Account Id. 

There are several emit telemetry events around this feedback call already `telemetry.feedback_result.emit({ result: 'Succeeded' })`. This call is not anonymous right? `clientId` will be present.

Where is the public documentation on this API? It appears codewhisper API's are [not public](https://docs.aws.amazon.com/codewhisperer/latest/userguide/useful-apis.html)? I'm used to having public everything from previous services so my following questions probably don't apply.

Do we legally state that this information is completely anonymous? And why? I only found a small snippet in a readme and the vue file itself.

If feedback is to remain anonymous, we need to present information to our stakeholders so they can make an appropriate decision.

Screenshot 2023-12-18 at 9 16 02 AM
Screenshot 2023-12-18 at 9 23 35 AM

Screenshot 2023-12-18 at 9 19 12 AM

damntrecky: This code snippet is outdated so its hard to place the context of where this was. Not sure if its in the `.emit` call or the submission for the API itself.

damntrecky: I think I found some documentation -> https://docs.aws.amazon.com/toolkit-for-jetbrains/latest/userguide/compliance-validation.html

damntrecky: data protection docs for IntelliJ -> https://docs.aws.amazon.com/toolkit-for-jetbrains/latest/userguide/data-protection.html

Don't mention anonymous and same for VSCode -> https://docs.aws.amazon.com/toolkit-for-vscode/latest/userguide/data-protection.html

PR #4184, opened 2023-12-17
fix: reorder input checks, remove modals, add jobId to table

dhasani23:
## Problem

This PR breaks down https://github.com/aws/aws-toolkit-vscode/pull/4183. This is part 1/3.

Had some unnecessary sleep statements. Missing jobId to the table shown to users. Unnecessary modals shown when displaying notifications. Possible bug when zipping dependencies if a dependency is named "target" (no evidence that this has occurred).

PR 2/3 will add the jobId to the metadata when users submit in-IDE feedback. PR 3/3 will formally remove the "he" and "proxyquire" dependencies.

## Solution

Remove sleep statements. Add jobId to table. Remove modals. Add an extra check when zipping dependencies.

## License

### DIFF
--- a/src/codewhisperer/service/transformByQHandler.ts
+++ b/src/codewhisperer/service/transformByQHandler.ts
@@ -64,7 +64,7 @@ export function throwIfCancelled() {
 export async function getOpenProjects() {
     const folders = vscode.workspace.workspaceFolders
     if (folders === undefined) {
-        vscode.window.showErrorMessage(CodeWhispererConstants.noSupportedJavaProjectsFoundMessage, { modal: true })
+        vscode.window.showErrorMessage(CodeWhispererConstants.noSupportedJavaProjectsFoundMessage)
         throw new ToolkitError('No Java projects found since no projects are open', { code: 'NoOpenProjects' })
     }
     const openProjects: vscode.QuickPickItem[] = []
@@ -84,13 +84,33 @@ export async function getOpenProjects() {
  */
 export async function validateProjectSelection(project: vscode.QuickPickItem) {
     const projectPath = project.description
+    const isMaven = await checkIfMaven(projectPath!)
+    if (!isMaven) {
+        const buildType = await checkIfGradle(projectPath!)
+        vscode.window.showErrorMessage(CodeWhispererConstants.noPomXmlFoundMessage)
+        if (buildType === 'Gradle') {
+            telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit({
+                codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+                codeTransformPreValidationError: 'NonMavenProject',
+                result: MetadataResult.Fail,
+                reason: buildType,
+            })
+        }
+        telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit({
+            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+            codeTransformPreValidationError: 'NoPom',
+            result: MetadataResult.Fail,
+            reason: 'NoPomFileFound',
+        })
+        throw new ToolkitError('No valid Maven build file found', { code: 'CouldNotFindPomXml' })
+    }
     const compiledJavaFiles = await vscode.workspace.findFiles(
         new vscode.RelativePattern(projectPath!, '**/*.class'),
         '**/node_modules/**',
         1
     )
     if (compiledJavaFiles.length < 1) {
-        vscode.window.showErrorMessage(CodeWhispererConstants.noSupportedJavaProjectsFoundMessage, { modal: true })
+        vscode.window.showErrorMessage(CodeWhispererConstants.noSupportedJavaProjectsFoundMessage)
         telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit({
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
             codeTransformPreValidationError: 'NoJavaProject',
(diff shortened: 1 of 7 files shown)


src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -84,13 +84,33 @@ export async function getOpenProjects() {
  */
 export async function validateProjectSelection(project: vscode.QuickPickItem) {
```
dhasani23, line 85: The redundant `telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit()` calls in this function are being addressed in a separate PR by a different author: https://github.com/aws/aws-toolkit-vscode/pull/4180

dhasani23: Here I'm simply changing the order of the input checks that we do: rather than running a shell command then checking for the presence of a `pom.xml` in the root directory, I check for the `pom.xml` check first because it is a simpler and quicker check.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -84,13 +84,33 @@ export async function getOpenProjects() {
...
     const projectPath = project.description
+    const isMaven = await checkIfMaven(projectPath!)
```
JuliaABurch, line 87: Would it make more sense to have a single method that returns the build type (Maven, Gradle, and Unknown)? It seems strange that `checkIfMaven` returns `true/false` and `checkIfGradle` returns a string (which should probably be an enum anyway...)

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -279,14 +274,25 @@ export async function uploadPayload(payloadFileName: string) {
...
+ * "target" is also the name of the folder where .class files, large JARs, etc. are stored after
+ * building, and we do not want these included in the ZIP so we exclude these when calling
+ * getFilesRecursively on the source code folder.
+ */
+function getFilesRecursively(dir: string, isDependenciesFolder: boolean): string[] {
     const entries = fs.readdirSync(dir, { withFileTypes: true })
     const files = entries.flatMap(entry => {
         const res = path.resolve(dir, entry.name)
-        // exclude 'target' directory from ZIP due to issues in backend
+        // exclude 'target' directory from ZIP (except if zipping dependencies) due to issues in backend
```
JuliaABurch, line 292: I look forward to the day this is fixed

src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -276,7 +276,7 @@ export class ProposedTransformationExplorer {
                     pathToArchive
                 )
             } catch (e: any) {
-                // This allows the customer to retry the download
```
JuliaABurch, line 279: Any reason we don't want to keep this comment above the `executeCommand`?

dhasani23: Oh right meant to ask you about that. How do you retry the download?

dhasani23: Is the point that the download button re-appears if the download fails?

src/codewhisperer/service/transformByQHandler.ts
dhasani23: Yep, agreed, fixing now

src/codewhisperer/service/transformationResultsViewProvider.ts
dhasani23: Update: yes, the download button re-appears should there be an issue with the download

damntrecky (approved):

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -84,13 +84,32 @@ export async function getOpenProjects() {
  */
 export async function validateProjectSelection(project: vscode.QuickPickItem) {
     const projectPath = project.description
+    const buildSystem = await checkBuildSystem(projectPath!)
+    if (buildSystem !== BuildSystem.Maven) {
+        vscode.window.showErrorMessage(CodeWhispererConstants.noPomXmlFoundMessage)
+        if (buildSystem === BuildSystem.Gradle) {
```
justinmk3, line 90: For this case the code below will emit `codeTransform_isDoubleClickedToTriggerInvalidProject` twice. Is that intentional? You will have 2x the number of `result=failed` results when buildsystem=gradle...

src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -277,6 +277,7 @@ export class ProposedTransformationExplorer {
                 )
             } catch (e: any) {
                 // This allows the customer to retry the download
+                vscode.window.showErrorMessage('Error downloading the diff')
```
justinmk3, line 280: This error will have no context. Should it mention "Amazon Q Transform" or something like that?

justinmk3 (approved):

src/codewhisperer/service/transformationResultsViewProvider.ts
dhasani23: I see your point, but at this stage of the process our "Transformation Hub" panel is open, as is the "Proposed Changes" panel, and this action is triggered by clicking the "Download Proposed Changes" button, so visually-speaking there is enough context.

justinmk3: I wouldn't make that assumption in general. Many features can be doing things at any time. Sometimes results are delayed, the user may have switched contexts, etc. 

This should not be a general practice.

src/codewhisperer/service/transformByQHandler.ts
dhasani23: Good point, fixed

src/codewhisperer/service/transformationResultsViewProvider.ts
dhasani23: Just added some context to the error message

Will-ShaoHua (approved):

src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -278,6 +278,7 @@ export class ProposedTransformationExplorer {
                 )
             } catch (e: any) {
                 // This allows the customer to retry the download
+                vscode.window.showErrorMessage('Transform by Q experienced an error when downloading the diff')
```
leigaol, line 281: Better if you can put these inline strings into constants file in case you reference them in multiple places.

dhasani23: Agreed, will do so in next PR. There are more than a couple of places where this should be done

leigaol (approved):

bryceitoc9:
Looks like https://github.com/aws/aws-toolkit-vscode/pull/4184#discussion_r1429022502 was merged, can you merge this into the PR so we can see how the duplicate metric was handled?

dhasani23:
The redundant emit() calls were not handled in that PR. It's something we are going to follow up on

bryceitoc9 (approved):

PR #4186, opened 2023-12-18
fix: include jobId in IDE feedback

dhasani23:
## Problem

Users submitting feedback in VS Code had issues with Transform by Q but we had no ways to investigate.

## Solution

Add the jobId to the feedback.

## License

### DIFF
--- a/src/feedback/vue/submitFeedback.vue
+++ b/src/feedback/vue/submitFeedback.vue
@@ -60,6 +60,7 @@ import { defineComponent } from 'vue'
 import { WebviewClientFactory } from '../../webviews/client'
 import saveData from '../../webviews/mixins/saveData'
 import { FeedbackWebview } from './submitFeedback'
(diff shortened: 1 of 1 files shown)


src/feedback/vue/submitFeedback.vue
```diff
@@ -60,6 +60,7 @@ import { defineComponent } from 'vue'
 import { WebviewClientFactory } from '../../webviews/client'
 import saveData from '../../webviews/mixins/saveData'
 import { FeedbackWebview } from './submitFeedback'
+import { transformByQState } from '../../codewhisperer/models/model'
```
justinmk3, line 63: I just noticed this is in a `.vue` (frontend) file. That is probably why the webpack build was failing.

dhasani23:
Moved this to https://github.com/aws/aws-toolkit-vscode/pull/4191

PR #1638, opened 2021-04-13
build: "await" all async function calls

justinmk3:
## Note

_This is a partial batch of changes. The `no-floating-promises` rule will be enabled in a future PR._

## Problem

- code may forget to call async functions with `await`, which can cause bugs or hide failures.
- Example: https://github.com/aws/aws-toolkit-vscode/pull/1636

## Solution

- Enable the eslint `no-floating-promises` rule. [doc](https://typescript-eslint.io/rules/no-floating-promises/)
    - Will be enabled in a future PR, not this one.
- For cases where we _want_ async behavior, use `.then()` or `void`.

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4188, opened 2023-12-19
fix: Fix missing conversationId in /dev

jsalles:
## Problem
There is currently a problem when using `/dev` inside Q which displays the 
```
Sorry, we encountered a problem when processing your request. Amazon Q feature development request failed: Conversation id must exist before continuing
``` 
error message every time a new conversation is started.

Retrying the operation is a workaround for this, and will make the conversation work as expected.

## Solution
This CR adds a missing `await` to the creation of the conversation. The missing `await` caused the error because the toolkit tried reading the `conversationId` before it was populated.

## License

### DIFF
--- a/src/amazonqFeatureDev/session/session.ts
+++ b/src/amazonqFeatureDev/session/session.ts
@@ -70,7 +70,7 @@ export class Session {
         // Store the initial message when setting up the conversation so that if it fails we can retry with this message
         this._latestMessage = msg
 
-        telemetry.amazonq_startConversationInvoke.run(async span => {
+        await telemetry.amazonq_startConversationInvoke.run(async span => {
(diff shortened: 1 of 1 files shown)


dianaabv16 (approved):

gasolima (approved):
Can we ask CWSPR folks to add eslint rule for erroring if you forget the await?
It's not the first time we introduce this error, and it's hard to catch

jpinkney-aws (approved):

justinmk3:
In the future please include the Problem/Solution in the commit message.

> Can we ask CWSPR folks to add eslint rule for erroring if you forget the await?

That is WIP, I'll plan to finish it this sprint. https://github.com/aws/aws-toolkit-vscode/pull/1638

PR #4191, opened 2023-12-19
 fix: error handling, move constants, add test, remove dep

dhasani23:
## Problem

Users submitting feedback in VS Code had issues with Transform by Q but we had no ways to investigate.

EDIT: we are currently trying to determine if the jobId is permitted to be included in the feedback, so ignore the above

Updated problem:

- 'he' dependency used but not necessary: https://sim.amazon.com/issues/EGCP-40
- stopJob error not caught
- not all constants were in the constants file
- no integration test for verifying that the pre-signed upload URL uses HTTPS and sets a 60 second timeout (needed for AppSec)

## Solution

Add the jobId to the feedback (when jobId is not empty, meaning the user must have started a transformation job).

EDIT: we are currently trying to determine if the jobId is permitted to be included in the feedback, so ignore the above

Updated solution:

- use a simple utility function instead to encode the HTML
- catch error and display error message
- move all constants to constants file
- add integration test to check those 2 conditions

Notes:

- the 'he' and 'proxyquire' dependencies will be removed from package.json and package-lock.json in the next PR; had some issues doing this properly last time so want to keep this PR organized and simple.
- this PR should fix the issue described here: https://github.com/aws/aws-toolkit-vscode/pull/4201

## License

### DIFF
--- a/src/codewhisperer/commands/startTransformByQ.ts
+++ b/src/codewhisperer/commands/startTransformByQ.ts
@@ -27,7 +27,7 @@ import { QuickPickItem } from 'vscode'
 import { MultiStepInputFlowController } from '../../shared//multiStepInputFlowController'
 import path from 'path'
 import { sleep } from '../../shared/utilities/timeoutUtils'
-import * as he from 'he'
+import { encodeHTML } from '../../shared/utilities/textUtilities'
 import {
     CodeTransformJavaSourceVersionsAllowed,
     CodeTransformJavaTargetVersionsAllowed,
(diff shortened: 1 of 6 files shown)


tincheng (approved):

bryceitoc9 (commented):
From offline conversation: please verify that adding a job ID doesn't remove anonymity from user feedback. (not blocking in case I'm out for the holidays when that's approved)

Note from David: see PR description, this part of the PR has been removed

justinmk3 (approved):
The pacakge.json item(s) still need to be removed. I assume that will be done in a followup PR.

dhasani23:
Yes as mentioned in PR description follow-up PR (doing that now) will remove them from package.json

PR #4204, opened 2023-12-22
fix(amazonq): "Transformation Hub" is always visible

justinmk3:
Problem:
"Transformation Hub" is always visible, even when Transform feature is not enabled.

image

Solution:
Define a "when" condition.

## License

### DIFF
--- a/.changes/next-release/Bug Fix-d4bf021f-adee-4cba-9737-b371d0be85ff.json
+++ b/.changes/next-release/Bug Fix-d4bf021f-adee-4cba-9737-b371d0be85ff.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Amazon Q: \"Transformation Hub\" is always visible, even if Transform feature is not enabled or used"
+}
--- a/package.json
+++ b/package.json
@@ -769,7 +769,8 @@
                 {
                     "type": "webview",
                     "id": "aws.amazonq.transformationHub",
-                    "name": "Status"
+                    "name": "Status",
+                    "when": "gumby.isTransformAvailable || gumby.isStopButtonAvailable"
                 },
                 {
                     "id": "aws.amazonq.transformationProposedChangesTree",


tincheng (dismissed):

tincheng:
I tested the change locally, and looks like the isTransformAvailable state is changed to false when a job started as well, so the Transformation Hub will disappear when job is running, and show again only after the job has completed.

dhasani23:
We would need to use a different context variable since isTransformAvailable is used for some other things, which is why Ting is seeing that issue. I tested it out a bit and since we also need to handle the case where the user switches b/w IdC and BuilderId (in which case the chat updates so that it knows whether or not to show /transform) this gets a little complicated. Trying to `setContext` (via `vscode.commands.executeCommand`) from any of the chat components causes the chat to error out / not render, so it seems like our only option would be to only show the Transformation Hub when the user has actually started a job -- will check with others to see if we are okay with this behavior. @justinmk3 let me know of your thoughts

justinmk3:
> only show the Transformation Hub when the user has actually started a job

If user has never started a job it should never show. That seems reasonable and doable, no?

dhasani23:
That part is fine, yes. I tested it and am more-so concerned that when you start a job it takes a few seconds for the Transformation Hub to appear. We have a "start" button in the Transformation Hub that allows users to start jobs from there; this button effectively becomes useless with this change, so I want to double check that we are ok with this and will revert to this PR.

damntrecky:
> I tested the change locally, and looks like the isTransformAvailable state is changed to false when a job started as well, so the Transformation Hub will disappear when job is running, and show again only after the job has completed.

I think we need to add to this logic. Users should be able to view Job History as well. If they have already run a job, but a job is not running the transformation hub should appear so users can see this job history table. 

> We also have a "start" button in the Transformation Hub that allows users to start jobs from there; this button becomes partly purposeless with this change (since for the first time a job is started, the Transformation Hub won't be visible)

See the point above about displaying Job History table as well. This doesn't seem like a problem in VSCode because there are multiple buttons outside the panel to trigger job status.


In a short summary this seems like a product level decision and would need to go through UX and our product team because it would diverge the experiences more between IntelliJ and VSCode.

Screenshot 2024-01-04 at 9 01 04 AM
Screenshot 2024-01-04 at 9 00 56 AM
4907d50b">
Screenshot 2024-01-04 at 9 00 44 AM

damntrecky:
I think I have a better grasp of the issue. It seems Code Reference Log is always showing. Can we change the content in our panel to address an action item for the user on how to turn on our feature. This could be a link to docs etc..

But I think telling users how to get the feature turned on is a better product decision than just hiding it.

justinmk3:
> Code Reference Log is always showing. 

That also should be fixed. 

> Can we change the content in our panel to address an action item for the user on how to turn on our feature. This could be a link to docs etc..

If Amazon Q is not being used, especially if Transform is not being used, we should not always show the panel. This has been raised by customers for other panels, previously.

tincheng:
I changed the `when` condition to the below and that persisted the transform panel when a job is running, and hid it when Q was not enabled (or signed out from IdC). 
```
                {
                    "type": "webview",
                    "id": "aws.amazonq.transformationHub",
                    "name": "Status",
                    "when": "gumby.isTransformAvailable || gumby.isStopButtonAvailable"
                },
```
Although it feels slightly buggy if the context `isStopButtonAvailable` may change in the future, but I think it's a viable short term solution while we revamp some experience to support transform through chat (the UX for transform panel may change).

dhasani23:
> I changed the `when` condition to the below and that persisted the transform panel when a job is running, and hid it when Q was not enabled (or signed out from IdC).
> 
> ```
>                 {
>                     "type": "webview",
>                     "id": "aws.amazonq.transformationHub",
>                     "name": "Status",
>                     "when": "gumby.isTransformAvailable || gumby.isStopButtonAvailable"
>                 },
> ```
> 
> Although it feels slightly buggy if the context `isStopButtonAvailable` may change in the future, but I think it's a viable short term solution while we revamp some experience to support transform through chat (the UX for transform panel may change).

Thanks Ting! Tested this and it looks good to me as well. The Transformation Hub disappears when logging out of IdC and re-appears when re-authenticating. For others who may read this: this works because `isTransformAvailable` effectively checks if there is an active IdC connection and `isStopButtonAvailable` will be true while the job is running, allowing the Transformation Hub to be visible for the duration of the job and only when IdC is active.

justinmk3:
Thank you all for the help! Pushed an update with @tincheng 's patch.

dhasani23 (approved):

PR #4209, opened 2023-12-28
feature(browser): enable basic global commands to be shared between browser/non-browser

hayemaxi:
Adds functionality for sharing some basic global commands between browser and desktop versions of the toolkit. Included commands:
- View Toolkit Documentation
- View Source on Github
- About Toolkit
- Report Toolkit Issue
- List Commands (*shared but inaccessible in browser at the moment)
- Send Feedback (*does not load in browser yet until vue/webview support is implemented on our end)

Not included yet:
- View Quick Start

Also removes the use of ExtContext type where used with these commands. Instead, refactor code to only pass what was used in this, which is just the actual vscode extension context.

Send Feedback... was disabled in the browser because vue is not yet supported and this feature may not be necessary for an initial
browser version of the toolkit.

#### Testing
tested both browser and regular(?) toolkit with this change.

## License

### DIFF
--- a/src/extensionShared.ts
+++ b/src/extensionShared.ts
@@ -24,7 +24,7 @@ export function initializeManifestPaths(extensionContext: vscode.ExtensionContex
     )
 }
 
-export function activateGlobalCommands(extensionContext: vscode.ExtensionContext) {
+export function registerCommands(extensionContext: vscode.ExtensionContext) {
     extensionContext.subscriptions.push(
         // No-op command used for decoration-only codelenses.
         vscode.commands.registerCommand('aws.doNothingCommand', () => {}),
--- a/src/extensionShared.ts
+++ b/src/extensionShared.ts
@@ -24,6 +24,9 @@ export function initializeManifestPaths(extensionContext: vscode.ExtensionContex
     )
 }
 
+/**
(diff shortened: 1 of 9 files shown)


justinmk3:
> Not included yet: View Quick Start

Let's remove/hide "View Quick Start" for vscode. It's only needed for Cloud9.

justinmk3 (approved):
Great! Moving as much as possible into `src/extensionShared.ts` is exactly the right direction.

src/extensionShared.ts
```diff
@@ -11,10 +11,42 @@
...
+import { openUrl } from './shared/utilities/vsCodeUtils'
 
 export function initializeManifestPaths(extensionContext: vscode.ExtensionContext) {
     globals.manifestPaths.endpoints = extensionContext.asAbsolutePath(join('resources', 'endpoints.json'))
     globals.manifestPaths.lambdaSampleRequests = extensionContext.asAbsolutePath(
         join('resources', 'vs-lambda-sample-request-manifest.xml')
     )
 }
+
+export function activateGlobalCommands(extensionContext: vscode.ExtensionContext) {
```
justinmk3, line 27: Nit all commands are "global" more or less.
```suggestion
export function registerCommands(extensionContext: vscode.ExtensionContext) {
```

Issue #4203, opened 2023-12-22
codewhisperer securityIssueProvider: "TypeError: Cannot read properties of undefined (reading 'range')"


justinmk3:
CI test logs are showing many of these unhandled rejections:

```
rejected promise not handled within 1 second: TypeError: Cannot read properties of undefined (reading 'range')
stack trace: TypeError: Cannot read properties of undefined (reading 'range')
    at SecurityIssueHoverProvider.handleDocumentChange (/codebuild/output/src3775465363/src/github.com/aws/aws-toolkit-vscode/src/codewhisperer/service/securityIssueProvider.ts:20:54)
    at f.value (/codebuild/output/src3775465363/src/github.com/aws/aws-toolkit-vscode/src/codewhisperer/activation.ts:331:53)
```

In the future, these will fail CI, but currently this particular error is ignored because it's a known issue. https://github.com/aws/aws-toolkit-vscode/blob/7c4e848235ed11d1ec18ac54dfd4a92ac193cbe7/buildspec/linuxTests.yml#L41-L42

Half-baked patch:

```diff
diff --cc src/codewhisperer/service/securityIssueProvider.ts
index 504e07092cad,504e07092cad..0ff03895b4bb
--- a/src/codewhisperer/service/securityIssueProvider.ts
+++ b/src/codewhisperer/service/securityIssueProvider.ts
@@@ -5,6 -5,6 +5,7 @@@
  
  import * as vscode from 'vscode'
  import { AggregatedCodeScanIssue } from '../models/model'
++import { ToolkitError } from '../../shared/errors'
  
  export abstract class SecurityIssueProvider {
      private _issues: AggregatedCodeScanIssue[] = []
@@@ -17,6 -17,6 +18,9 @@@
      }
  
      public handleDocumentChange(event: vscode.TextDocumentChangeEvent) {
++        if (!event.contentChanges || event.contentChanges.length === 0) {
++            throw new ToolkitError(`invalid event.contentChanges: ${JSON.stringify(event.contentChanges)}`)
++        }
          const changedRange = event.contentChanges[0].range
          const changedText = event.contentChanges[0].text
          const lineOffset = this._getLineOffset(changedRange, changedText)
```

justinmk3:
# Steps to reproduce

1. open `src/test/codewhisperer/service/telemetry.test.ts` in vscode
2. run the `Extension Tests (current file)` launch config
[...]

PR #4210, opened 2023-12-28
Fix: codewhisperer securityIssueProvider: "TypeError: Cannot read properties of undefined (reading 'range')"

laileni-aws:
## Problem
https://github.com/aws/aws-toolkit-vscode/issues/4203

## Solution
Add a check for `event.contentChanges` is `undefined` in `handleDocumentChange` function. If yes, `return`

## License

### DIFF
--- a/src/codewhisperer/service/securityIssueProvider.ts
+++ b/src/codewhisperer/service/securityIssueProvider.ts
@@ -16,7 +16,7 @@ export abstract class SecurityIssueProvider {
     }
 
     public handleDocumentChange(event: vscode.TextDocumentChangeEvent) {
-        // handleDocumentChange function may be triggered while testing by our own code generation. Adding a conditional check to return if we do not have event.
+        // handleDocumentChange function may be triggered while testing by our own code generation.
         if (!event.contentChanges || event.contentChanges.length === 0) {
             return
         }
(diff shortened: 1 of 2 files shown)


src/codewhisperer/service/securityIssueProvider.ts
```diff
@@ -17,6 +16,9 @@ export abstract class SecurityIssueProvider {
     }
 
     public handleDocumentChange(event: vscode.TextDocumentChangeEvent) {
+        if (!event.contentChanges || event.contentChanges.length === 0) {
```
justinmk3, line 19: please add a code comment explaining why this is needed, e.g.:

> `handleDocumentChange` is triggered because of change in document due to our own code generation.

justinmk3 (approved):

src/codewhisperer/service/securityIssueProvider.ts
```diff
@@ -17,6 +16,10 @@ export abstract class SecurityIssueProvider {
     }
 
     public handleDocumentChange(event: vscode.TextDocumentChangeEvent) {
+        // handleDocumentChange function may be triggered while testing by our own code generation. Adding a conditional check to return if we do not have event.
```
justinmk3, line 19:
```suggestion
        // handleDocumentChange function may be triggered while testing by our own code generation.
```

PR #4217, opened 2024-01-02
feat(auth): don't display switch connections/add new connections in Q/CW ellipses menus

hayemaxi:
Problem: Ellpises menus for Q/CW will display "Switch Connections", implying that the connection can be switched between other IdC/Builder ID connections on the fly. This isn't supported currently. Also, 'Add new connection' is displayed when already signed in, implying that you can add these new connections to swap to.

Solution: Remove 'Switch Connections' menu item, and only show 'Add new connection' when not signed in.

Also, rename the global context variables to to be more descriptive/not be in all caps.

image

image

## License


nkomonen-amazon (approved):

justinmk3 (approved):

justinmk3:
> Remove 'Switch Connections' menu item, and only show 'Add new connection' when not signed in.

This is worth mentioning in the changelog

davidmerwin (commented):
#2460 @aws-sdk-osds

davidmerwin (commented):
@davidmerwin @aws-sdk-osds

PR #4202, opened 2023-12-22
lint: require "await" or .catch() on async calls

justinmk3:
Continued from #1638

# Problem:

- All promises or async calls should use `await`, or end with `.catch()`.
  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises
- Async calls accidentally forget `await`.
  - Example: 8863ba4a799e
- Test logs show "rejected promise not handled":
  ```
  rejected promise not handled within 1 second: HTTPError: Response code 403 (rate limit exceeded)
  rejected promise not handled within 1 second: Error: command 'aws.codeWhisperer.refreshStatusBar' not found
  rejected promise not handled within 1 second: HTTPError: Response code 404 (Not Found)
  rejected promise not handled within 1 second: CodeExpectedError: cannot open notCloudWatch:hahahaha. Detail: Unable to resolve resource notCloudWatch:hahahaha
  rejected promise not handled within 1 second: AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  rejected promise not handled within 1 second: Error: No access
  rejected promise not handled within 1 second: Error: No access
  rejected promise not handled within 1 second: Error: Timed-out waiting for browser login flow to complete
  rejected promise not handled within 1 second: Error: Download code bindings cancelled
  rejected promise not handled within 1 second: Error: EACCES: permission denied, mkdir '/my'
  ```

# Solution:

- Enable eslint rule: https://typescript-eslint.io/rules/no-floating-promises/

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4218, opened 2024-01-03
lint: require "await" or .catch() on async calls

justinmk3:
Continued from:

- https://github.com/aws/aws-toolkit-vscode/pull/4202
- #1638

# Problem:

- All promises or async calls should use `await`, or end with `.catch()`.
  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises
- Async calls accidentally forget `await`.
  - Example: 8863ba4a799e
- Test logs show "rejected promise not handled":
  ```
  rejected promise not handled within 1 second: HTTPError: Response code 403 (rate limit exceeded)
  rejected promise not handled within 1 second: Error: command 'aws.codeWhisperer.refreshStatusBar' not found
  rejected promise not handled within 1 second: HTTPError: Response code 404 (Not Found)
  rejected promise not handled within 1 second: CodeExpectedError: cannot open notCloudWatch:hahahaha. Detail: Unable to resolve resource notCloudWatch:hahahaha
  rejected promise not handled within 1 second: AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  rejected promise not handled within 1 second: Error: No access
  rejected promise not handled within 1 second: Error: No access
  rejected promise not handled within 1 second: Error: Timed-out waiting for browser login flow to complete
  rejected promise not handled within 1 second: Error: Download code bindings cancelled
  rejected promise not handled within 1 second: Error: EACCES: permission denied, mkdir '/my'
  ```

# Solution:

- Enable eslint rule: https://typescript-eslint.io/rules/no-floating-promises/

## License

### DIFF
--- a/src/test/codewhisperer/tracker/codewhispererTracker.test.ts
+++ b/src/test/codewhisperer/tracker/codewhispererTracker.test.ts
@@ -65,9 +65,11 @@ describe('codewhispererTracker', function () {
         })
 
         it('Should skip if telemetry is disabled', async function () {
+            globals.telemetry.telemetryEnabled = false
             const getTimeSpy = sinon.spy(Date.prototype, 'getTime')
             await CodeWhispererTracker.getTracker().flush()
             assert.ok(!getTimeSpy.called)
+            globals.telemetry.telemetryEnabled = true
         })
     })
 
--- a/src/codewhisperer/activation.ts
+++ b/src/codewhisperer/activation.ts
@@ -111,7 +111,7 @@ export async function activate(context: ExtContext): Promise<void> {
             if (
                 configurationChangeEvent.affectsConfiguration('aws.codeWhisperer.includeSuggestionsWithCodeReferences')
             ) {
-                await ReferenceLogViewProvider.instance.update()
+                ReferenceLogViewProvider.instance.update()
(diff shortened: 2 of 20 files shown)


src/test/codewhisperer/tracker/codewhispererTracker.test.ts
```diff
@@ -65,9 +65,11 @@ describe('codewhispererTracker', function () {
         })
 
         it('Should skip if telemetry is disabled', async function () {
+            globals.telemetry.telemetryEnabled = false
             const getTimeSpy = sinon.spy(Date.prototype, 'getTime')
             await CodeWhispererTracker.getTracker().flush()
             assert.ok(!getTimeSpy.called)
```
justinmk3, line 71: Before the change to `codewhispererTracker.ts`, this test was accidentally passing.

andrewyuq:
So in case when we don't need to await we will just use `void`?

justinmk3:
> So in case when we don't need to await we will just use `void`?

You  always should `await` or `.catch` except when it's not possible.

For this initial effort, I've used `void` in many places simply because manually deciding how to handle 500 different call sites is an undue burden. But that doesn't mean using `void` is an acceptable practice.

PR #4221, opened 2024-01-04
lint: require "await" or .catch() on async calls

justinmk3:
Continued from:

- #1638
- #4202
- #4218
- #4220

Solution:
- Enable eslint rule: https://typescript-eslint.io/rules/no-floating-promises/

## License

### DIFF
--- a/src/lambda/config/templates.ts
+++ b/src/lambda/config/templates.ts
@@ -179,7 +179,7 @@ export function showTemplatesConfigurationError(
 ): void {
     const logger: Logger = getLogger()
 
-    showErrorMessage(
+    void showErrorMessage(
         localize(
             'AWS.lambda.configure.error.fieldtype',
             'Your templates.json file has an issue. {0} was detected as {1} instead of one of the following: [{2}]. Please change or remove this field, and try again.',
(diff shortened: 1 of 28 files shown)


.eslintrc.js
```diff
@@ -93,14 +93,13 @@ module.exports = {
         'no-null/no-null': 'error',
         '@typescript-eslint/no-empty-function': 'off',
         '@typescript-eslint/no-unused-vars': 'off',
-        // New rules --> New TODOs
+        '@typescript-eslint/no-floating-promises': 'error', // Promises must catch errors or be awaited.
```
damntrecky, line 96: Question: Should this rule be applied to only non test files? Maybe excluding files with the word `test` or `Test` in them. I don't think there is the `.spec.` pattern to apply to in this codebase.

justinmk3: That would add a source of accidents to the tests. What would be gained?

src/lambda/config/templates.ts
```diff
@@ -197,7 +197,7 @@ export function showTemplatesConfigurationError(
 }
 
 export async function ensureTemplatesConfigFileExists(path: string): Promise<void> {
-    ensureDir(_path.dirname(path))
+    await ensureDir(_path.dirname(path))
```
justinmk3, line 200: This looks like a potential bug (race).

.eslintrc.js
justinmk3: @damntrecky here's a recent example where tests were not correctly await'd: https://github.com/aws/aws-toolkit-vscode/pull/4240/files#r1444539552

PR #4230, opened 2024-01-04
refactor: remove some uses of `fs-extra`

nkomonen-amazon:
## Problem
`fs-extra` causes issues if a module that is supposed to run in the browser transitively imports `fs-extra` at some point since `fs-extra` executes code on import that breaks in the browser.

## Solution
Remove some uses of `fs-extra` and replace them with the platform agnostic `FileSystemCommon` implementation, using the exported `fsCommon` instance in place of `fs-extra`.

## License


src/srcShared/fs.ts
```diff
@@ -159,3 +179,5 @@ export class FileSystemCommon {
         return vscode.Uri.file(path)
     }
 }
+
+export const fsCommon = FileSystemCommon.instance
```
justinmk3, line 183: Yay :) Could even just name it `fs`, since it will be rare to use multiple "fs" modules at once.
```suggestion
export const fs = FileSystemCommon.instance
```

nkomonen-amazon:
**Lets make sure to not squash this in case we need to debug**

.eslintrc.js
```diff
@@ -116,5 +116,13 @@ module.exports = {
...
         ],
         'aws-toolkits/no-only-in-tests': 'error',
+        // The following will place an error on the `fs-extra` import since we do not want it to be used for browser compatibility reasons.
+        // "no-restricted-imports": [
+        //     "error",
+        //     {
+        //         "name": "fs-extra",
+        //         "message": "Avoid fs-extra, use FileSystemCommon. Notify the Toolkit team if your required functionality is not available."
+        //     }
+        // ],
```
hayemaxi, line 126: When do we plan to enable this? Are we still missing uses of `fs-extra`?

src/codecatalyst/model.ts
```diff
@@ -287,13 +287,13 @@ export async function getDevfileLocation(client: DevEnvClient, root?: vscode.Uri
     async function checkDefaultLocations(rootDirectory: vscode.Uri): Promise<vscode.Uri> {
         // Check the projects root location
         const devfileRoot = vscode.Uri.joinPath(vscode.Uri.parse('/projects'), 'devfile.yaml')
-        if (await fileExists(devfileRoot.fsPath)) {
+        if (await fileOrFolderExists(devfileRoot.fsPath)) {
```
justinmk3, line 290: is the rename needed? "file" is understood to mean any filesystem path, I think. That's presumably why the old name was `fileExists`.

"directory" is used for cases where non-directory files are expected.

src/shared/filesystemUtilities.ts
```diff
@@ -198,18 +196,23 @@ export function getFileDistance(fileA: string, fileB: string): number {
...
+    dir: string,
+    name: string,
+    suffix: string,
+    max: number = 99
+): Promise<string> {
     if (!name) {
         throw new Error(`name is empty`)
     }
-    if (!fs.existsSync(dir)) {
+    if (!(await fsCommon.directoryExists(dir))) {
```
justinmk3, line 208: "dir" is very commonly used in our (and nodejs's) existing filesystem code. No need for the verbose form.
```suggestion
    if (!(await fsCommon.dirExists(dir))) {
```

src/srcShared/fs.ts
```diff
@@ -60,6 +62,10 @@ export class FileSystemCommon {
         return FileSystemCommon.arrayToString(await this.readFile(path))
     }
 
+    /**
+     * The vscode fs implementation does not explicitly provide an append method
```
justinmk3, line 66: would be less ambiguous, and less verbose, to mention the exact name of the vscode api?

src/lambda/wizards/samInitWizard.ts
```diff
@@ -274,12 +274,7 @@ export class CreateNewSamAppWizard extends Wizard<CreateNewSamAppWizardForm> {
 
         this.form.name.bindPrompter(state =>
             createNamePrompter(
-                fsutil.getNonexistentFilename(
-                    state.location!.fsPath,
-                    `lambda-${state.runtimeAndPackage!.runtime}`,
-                    '',
-                    99
-                )
+                `${`lambda-${state.runtimeAndPackage!.runtime}`}-${crypto.randomBytes(4).toString('hex')}`
```
justinmk3, line 277: Why was this change made? New code is a minor UX regression.

src/shared/extensionStartupUtilities.ts
justinmk3, line 1: this file could be named `src/shared/extensionStartup.ts`

src/test/shared/debug/launchConfiguration.test.ts
```diff
@@ -99,14 +99,15 @@ describe('LaunchConfiguration', function () {
     const workspace = vscode.workspace.workspaceFolders![0]
     const testLaunchJson = vscode.Uri.file(path.join(workspace.uri.fsPath, '.vscode/launch.json'))
     /** Object representation of the launch.json test file. */
-    const testLaunchJsonData = JSON.parse(testutil.fromFile(testLaunchJson.fsPath))
+    let testLaunchJsonData: any
```
justinmk3, line 102:
```suggestion
    let testLaunchJsonData: string
```

.eslintrc.js
nkomonen-amazon: I just put this here since we will eventually want it. There are still many uses of `fs-extra`.

I'm wondering if we can just enable even with uses of fs-extra, will this cause CI to fail?

src/srcShared/fs.ts
```diff
@@ -76,6 +82,20 @@ export class FileSystemCommon {
         return this.writeFile(path, finalContent)
     }
 
+    /**
+     * The same as {@link writeFile} except it will also create the parent directories
+     * if missing.
+     */
+    async outputFile(path: Uri | string, content: Uint8Array | string): Promise<void> {
```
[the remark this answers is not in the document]
nkomonen-amazon: Started implementing this, and coincidentally the `vscode.workspace.fs.writeFile()` already does this automatically!

src/test/shared/debug/launchConfiguration.test.ts
nkomonen-amazon: JSON.parse returns an `any` type implicitly, anything else causes the code that uses this to have errors.

src/lambda/wizards/samInitWizard.ts
nkomonen-amazon: This was due to the code needing to be synchronous, I created a duplicate of the async method. See `getNonexistentFilename()`

src/shared/filesystemUtilities.ts
```diff
@@ -198,7 +196,44 @@ export function getFileDistance(fileA: string, fileB: string): number {
...
+ *
+ * To avoid excessive filesystem activity, if all filenames up to `max` exist,
+ * the function instead appends a random string.
+ *
+ * @param dir  Path to a directory
+ * @param name  Filename without extension
+ * @param suffix  Filename suffix, typically an extension (".txt"), may be empty
+ * @param max  Stop searching if all permutations up to this number exist
+ */
+export function getNonexistentFilenameSync(dir: string, name: string, suffix: string, max: number = 99): string {
```
justinmk3, line 236: To avoid code duplication, we could pass in `existsDir` and `existsDirFile` as parameters. 

To unblock this PR, I can do that as a followup

src/lambda/wizards/samInitWizard.ts
```diff
@@ -274,12 +274,7 @@ export class CreateNewSamAppWizard extends Wizard<CreateNewSamAppWizardForm> {
 
         this.form.name.bindPrompter(state =>
             createNamePrompter(
-                fsutil.getNonexistentFilename(
-                    state.location!.fsPath,
-                    `lambda-${state.runtimeAndPackage!.runtime}`,
-                    '',
-                    99
-                )
+                getNonexistentFilenameSync(state.location!.fsPath, `lambda-${state.runtimeAndPackage!.runtime}`, '', 99)
```
justinmk3, line 277: I wonder why `bindPrompter` can't take an async function :/

justinmk3 (approved):
Nicely done

hayemaxi (approved):

src/shared/filesystemUtilities.ts
nkomonen-amazon: I'll merge this for now. I'm not sure `existsDirFile` will work with the current logic since it checks different files in a loop.

justinmk3: Fixed in https://github.com/aws/aws-toolkit-vscode/pull/4576

src/lambda/wizards/samInitWizard.ts
justinmk3: Fixed in https://github.com/aws/aws-toolkit-vscode/pull/4576

PR #4234, opened 2024-01-05
dev: Add dependency diagram utility

nkomonen-amazon:
Adding in this dependency diagram generating utility for toolkit Browser work since we can have transitive dependencies which break the toolkit from running in the Browser.

See the `browser.md` guide for [more details here](https://github.com/aws/aws-toolkit-vscode/blob/81a091e21f81ef845d9f300c068325199b4468f6/docs/browser.md#finding-incompatible-transitive-dependencies)

This will help us to find dependencies more easily instead of manually searching.

## License


hayemaxi (approved):
nice!

docs/browser.md
```diff
@@ -57,3 +57,23 @@ To get to the AWS Toolkit logs:
...
+when I execute `myFile.ts` in the browser it will break due to `fs-extra` not being browser compatible.
+
+It may be difficult to determine which module imported `fs-extra` due to a nested chain of transitive dependencies.
+
+As a solution, we can use [`dependency-cruiser`](https://www.npmjs.com/package/dependency-cruiser) to generate a dependency diagram
+to help us visualize the imports and determine which module is importing a certain module.
+
+### How to use
+
+0. Ensure you have run `npm install --save-dev`
```
justinmk3, line 73: why is `--save-dev` needed? `npm install` should install dev deps by default

package.json
```diff
@@ -4298,6 +4298,7 @@
         "c8": "^9.0.0",
         "circular-dependency-plugin": "^5.2.2",
         "css-loader": "^6.7.3",
+        "dependency-cruiser": "^16.0.0",
```
justinmk3, line 4301: Is this actually needed? The doc mentions `npx`, which should automatically pull stuff dynamically.

docs/browser.md
justinmk3: Since none of our modules are importing `dependency-cruiser`, can we change this instruction to:

    npm install dependency-cruiser

Then our `package-lock.json` won't have the extra deps, and we won't need to maintain that dep.

package.json
nkomonen-amazon: Makes sense, I've removed the dependency

PR #4237, opened 2024-01-06
fix(gumby): add jobId to feedback, delete temp files, add test, typos, minor UI fixes, use shell: true

dhasani23:
## Problem

1. Need to add jobId to in-IDE feedback when a transformation job was run so that we can investigate any issues
2. Need to delete temp files (the ZIP upload and the result archive) once no longer needed
3. Need a test requested by AppSec to ensure zipCode behaves correctly
4. Minor typos
5. Dropdown menu with 3 buttons (Show Changes, Show Plan, Show Summary) was inconsistent: Show Plan was visible but enabled/disabled as necessary while the other 2 buttons toggled their visibility as necessary
6. Used Font Awesome for an icon, but better to use built-in icon
7. When running shell commands setting `shell: false` resulted in the `mvn` command failing for 1 user due to inability to find the Maven executable

## Solution

1. Add jobId
2. Delete temp files
3. Write test
4. Fix typos
5. Align behavior of all 3 buttons via `"enablement"`
6. Use built-in icon
7. Set `shell: true`

## License


package.json
```diff
@@ -2411,15 +2411,18 @@
             },
             {
                 "command": "aws.amazonq.transformationHub.summary.reveal",
-                "title": "%AWS.command.q.transform.showChangeSummary%"
+                "title": "%AWS.command.q.transform.showChangeSummary%",
+                "enablement": "gumby.reviewState == InReview"
             },
```
dhasani23, line 2416: Doing this via "enablement" makes it consistent with the "Show Plan" button in the dropdown, which is visible the entire time but enabled/disabled as appropriate.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -113,7 +113,7 @@ export async function validateProjectSelection(project: vscode.QuickPickItem) {
     const classFilePath = compiledJavaFiles[0].fsPath
     const baseCommand = 'javap'
     const args = ['-v', classFilePath]
-    const spawnResult = spawnSync(baseCommand, args, { shell: false, encoding: 'utf-8' })
+    const spawnResult = spawnSync(baseCommand, args, { shell: true, encoding: 'utf-8' })
```
dhasani23, line 116: Had a customer issue (see below) where setting shell: true for the `mvn` command allowed the shell to find their Maven executable, so will set shell to true here too to pre-empt any issues with the `javap` command. It seems that setting shell to true allows for more permissions so our error rates should drop with these changes.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -273,6 +272,13 @@ export async function uploadPayload(payloadFileName: string) {
         // Pass along error to callee function
         throw new ToolkitError(errorMessage, { cause: e as Error })
     }
+    try {
+        await uploadArtifactToS3(payloadFileName, response)
+    } catch (e: any) {
+        const errorMessage = e?.message || 'Error in uploadArtifactToS3 call'
+        throw new Error(errorMessage)
+    }
+    return response.uploadId
```
dhasani23, line 281: Moved this try-catch down here so that error handling is separate for the CreateUploadUrl and uploadArtifactToS3, rather than bundling it all in one try block which doesn't tell us exactly which method errored out.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -317,7 +323,7 @@ function getProjectDependencies(modulePath: string): string[] {
         '-Dmdep.copyPom=true',
         '-Dmdep.addParentPoms=true',
     ]
-    const spawnResult = spawnSync(baseCommand, args, { cwd: modulePath, shell: false, encoding: 'utf-8' })
+    const spawnResult = spawnSync(baseCommand, args, { cwd: modulePath, shell: true, encoding: 'utf-8' })
```
dhasani23, line 326: Discussed above; needed to resolve a customer issue where their Maven executable was not located unless shell is set to true.

src/codewhisperer/service/transformationHubViewProvider.ts
```diff
@@ -228,7 +244,7 @@ export class TransformationHubViewProvider implements vscode.WebviewViewProvider
         if (stepStatus === StepProgress.Succeeded) {
             return `<span style="color: green"> ✓ </span>`
         } else if (stepStatus === StepProgress.Pending) {
-            return `<span> <i class="fas fa-spinner fa-spin"></i> </span>` // TODO: switch from FA to native VSCode icons
+            return `<span style="color: grey"> <i class="codicon codicon-loading spinner"></i> </span>`
```
dhasani23, line 247: Get rid of Font Awesome CDN and use a built-in icon

src/feedback/vue/submitFeedback.ts
```diff
@@ -35,6 +36,11 @@ export class FeedbackWebview extends VueWebview {
             return 'Choose a reaction (smile/frown)'
         }
 
+        const jobId = transformByQState.getJobId()
+        if (jobId !== '') {
```
dhasani23, line 40: Means a job must have been run

src/codewhisperer/service/transformationHubViewProvider.ts
```diff
@@ -185,11 +185,27 @@ export class TransformationHubViewProvider implements vscode.WebviewViewProvider
             }
         }
         const isJobInProgress = transformByQState.isRunning()
+        const codiconsUri = this._view?.webview.asWebviewUri(
+            vscode.Uri.joinPath(this._extensionUri, 'node_modules', '@vscode/codicons', 'dist', 'codicon.css')
```
justinmk3, line 189: This looks strange. No other webview needs to reach into `node_modules`, why is it needed here? Are there not existing codicon usage examples that can be found in existing webviews in this codebase?

dhasani23: I previously tried https://github.com/aws/aws-toolkit-vscode/blob/f355be267fc59a116a9f697c0163f4f2e8c5e45b/src/codewhisperer/views/securityPanelViewProvider.ts#L52 and also the `getIcon()` used in many other places, neither of which worked, so I took this approach, but regardless the latest work accomplishes the same in a simpler way. I didn't test with the VSIX but I imagine it would fail since `node_modules` is not there; good to know.

src/codewhisperer/service/transformationHubViewProvider.ts
```diff
@@ -237,9 +250,9 @@ export class TransformationHubViewProvider implements vscode.WebviewViewProvider
         if (stepStatus === StepProgress.Succeeded) {
             return `<span style="color: green"> ✓ </span>`
         } else if (stepStatus === StepProgress.Pending) {
-            return `<span> <i class="fas fa-spinner fa-spin"></i> </span>` // TODO: switch from FA to native VSCode icons
+            return `<span style="color: grey" class="spinner"> ↻ </span>`
         } else {
-            return `<span style="color: red"> X </span>`
+            return `<span style="color: grey"> ✓ </span>`
```
dhasani23, line 255: Our PM and UX team decided that they want to display a grey checkmark rather than a red X when a transformation step fails since the transformation was in-progress / partially completed.

dhasani23:
I ran prettier locally and everything seemed to pass fine, I'm not sure why it's failing here. Can try to fix in a future PR if it persists.

justinmk3:
Try rebasing. The lint issue should be fixed in this PR.

package.json
justinmk3: yes, that is a strongly preferred UI practice.

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -212,6 +213,8 @@ export async function startTransformByQ() {
         transformByQState.setJobId(encodeHTML(jobId))
         await vscode.commands.executeCommand('aws.amazonq.refresh')
 
+        fs.rmSync(payloadFileName, { recursive: true, force: true }) // delete ZIP after job has started
```
justinmk3, line 216: what does rmSync do if `payloadFileName` is undefined?

dhasani23: Throws an error but this will only happen if zipCode fails, in which case this line won’t even execute since we throw a TransformByQUploadArchiveFailed error if zipCode fails

justinmk3:
are there customer-facing improvements that should be mentioned in the changelog?

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -273,6 +272,13 @@ export async function uploadPayload(payloadFileName: string) {
         // Pass along error to callee function
         throw new ToolkitError(errorMessage, { cause: e as Error })
     }
+    try {
+        await uploadArtifactToS3(payloadFileName, response)
+    } catch (e: any) {
+        const errorMessage = e?.message || 'Error in uploadArtifactToS3 call'
+        throw new Error(errorMessage)
```
damntrecky, line 279: We throw a ToolkitError above. Should we be doing the same in this catch statement?

justinmk3: `e?` also looks suspicious; when would it be undefined?

src/codewhisperer/service/transformationHubViewProvider.ts
```diff
@@ -237,9 +250,9 @@ export class TransformationHubViewProvider implements vscode.WebviewViewProvider
         if (stepStatus === StepProgress.Succeeded) {
             return `<span style="color: green"> ✓ </span>`
         } else if (stepStatus === StepProgress.Pending) {
-            return `<span> <i class="fas fa-spinner fa-spin"></i> </span>` // TODO: switch from FA to native VSCode icons
+            return `<span style="color: grey" class="spinner"> ↻ </span>`
```
damntrecky, line 253: Question: Is there a better way to show this spinner. It looks like you copy/pasted the unicode character itself from the VSCode website?

Docs -> https://code.visualstudio.com/api/references/icons-in-labels#animation

Can you do something like this example -> https://github.com/microsoft/vscode-extension-samples/blob/22d5639ff5c1d88f144c057fc3d29cc9dfd99d62/webview-codicons-sample/src/extension.ts#L34

Question: Not sure if the "codicon" library is the native icons. Looks like those are not for webviews from the docs say you CANT use the native icons in webview:
Screenshot 2024-01-09 at 10 51 11 AM

damntrecky (approved):
Some questions but I have some tentative follow ups for later this week.

src/codewhisperer/service/transformationHubViewProvider.ts
justinmk3: here's an example of a codicon being used in a webview: https://github.com/aws/aws-toolkit-vscode/blob/a11d5db448b9697adb8e3a0038be248e1e3e1660/src/codewhisperer/views/securityPanelViewProvider.ts#L51-L53

dhasani23: Yes the icon itself is copied/pasted. I initially had it like the example you linked but we don't want to be going into `node_modules` since it doesn't exist in the VSIX.

src/codewhisperer/service/transformByQHandler.ts
dhasani23: We already do, but in the `uploadArtifactToS3` function itself: https://github.com/aws/aws-toolkit-vscode/blob/1b28a401c7866c29bb427f786cb152afb6c2cc14/src/codewhisperer/service/transformByQHandler.ts#L195

tincheng (approved):

src/codewhisperer/service/transformByQHandler.ts
tincheng: Yes! I saw it later on, sorry for the miss

src/codewhisperer/service/transformationHubViewProvider.ts
dhasani23: @justinmk3 I am unable to find an `icons.css` file locally. Also what is the concern with copying/pasting the icon?

src/codewhisperer/commands/startTransformByQ.ts
tincheng: Instead of removing the zip after job has started, what about putting it in a `finally()` after trying to upload to S3, to delete it regardless of whether job starts?

src/codewhisperer/service/transformationHubViewProvider.ts
dhasani23: Screenshot 2024-01-09 at 5 47 09 PM

Copied the code exactly from that example and for some reason only a small fraction of the icon appears. Tried altering the CSS in various ways but unable to resolve it. If this is a significant problem then I can make a SIM for it but I would prefer to get the other important changes in this PR out sooner.

.changes/next-release/Bug Fix-488470d7-db3a-48fe-8c06-2b499e374995.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Make maven command less error prone; delete temp files"
```
justinmk3, line 3: this changelog description is not meaningful to anyone outside of this PR. it needs, at minimum, a prefix indicating which feature it's referring to. 

please review https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#changelog

```suggestion
	"description": "Amazon Q Transform: Make maven command less error prone; delete temp files"
```

justinmk3 (approved):

PR #4246, opened 2024-01-09
ci: fail on "rejected promise not handled"

justinmk3:
# Problem:
rejected promises in test logs are silently ignored:

    rejected promise not handled within 1 second: Error: command 'aws.codeWhisperer.refresh' not found

# Solution:
- Fix various cases (see commits).
- Enable CI failure for `run_and_report`.

## License


hayemaxi:
will this swallow all errors in a command? Shouldn't we surface logic errors in the code? Seems like from a developer perspective this would be an unintended side effect.

PR #4247, opened 2024-01-09
feat(appcomposer): prompt for IAM credentials

JLargent42:
## Problem
 The built-in SAM Sync functionality provides an ambiguous error message when the user is signed in with SSO or Builder ID. The only supported form of authentication is IAM credentials. Otherwise, when the user attempts to sync, they will get this error message:

image

This can be confusing to the user, especially if they are using IAM Identity Center for authentication.

## Solution
This adds a modal to inform the user that their current form of authentication (or lack of authentication) is not supported by SAM Sync and provides a straightforward way to switch to IAM credential authentication.

Screenshot 2024-01-17 at 1 12 25 PM

Under the new flow, if the user chooses to authenticate with IAM credentials, the user will get a quick pick menu to choose from a list of valid credentials (including all IAM credentials, but excluding SSO and Builder ID). Once selected, the sync flow will continue as normal.

Ideally, the `sam sync` workflow should support all forms of authentication. Until this is supported, this new flow helps guide the user to a working form of authentication rather than using an unhelpful error message.

## License

### DIFF
--- a/src/applicationcomposer/messageHandlers/deployMessageHandler.ts
+++ b/src/applicationcomposer/messageHandlers/deployMessageHandler.ts
@@ -7,45 +7,6 @@ import vscode from 'vscode'
 import { Command, DeployRequestMessage, DeployResponseMessage, MessageType, WebviewContext } from '../types'
 import { SamSyncResult } from '../../shared/sam/sync'
 import { telemetry } from '../../shared/telemetry/telemetry'
-import { Auth } from '../../auth/auth'
-import { promptAndUseConnection } from '../../auth/utils'
-import { StatefulConnection } from '../../auth/connection'
-import * as nls from 'vscode-nls'
-import { showMessageWithUrl } from '../../shared/utilities/messages'
-import { credentialHelpUrl } from '../../shared/constants'
-
-const localize = nls.loadMessageBundle()
-
-async function promptReauth(connection?: StatefulConnection) {
-    let errorMessage = localize(
-        'aws.applicationComposer.deploy.authModal.message',
-        'Syncing requires authentication with IAM credentials.'
-    )
-    if (connection?.state === 'valid') {
-        errorMessage =
-            localize(
-                'aws.applicationComposer.deploy.authModal.invalidAuth',
-                'Authentication through Builder ID or IAM Identity Center detected. '
-            ) + errorMessage
-    }
-    const acceptMessage = localize(
-        'aws.applicationComposer.deploy.authModal.accept',
-        'Authenticate with IAM credentials'
-    )
-    const docMessage = localize('aws.applicationComposer.deploy.authModal.docLink', 'Open documentation')
-    const modalResponse = await showMessageWithUrl(
-        errorMessage,
-        credentialHelpUrl,
-        docMessage,
-        'info',
-        [acceptMessage],
-        true
-    )
-    if (modalResponse !== acceptMessage) {
-        return
-    }
-    await promptAndUseConnection(Auth.instance, 'iam-only')
-}
 
 export async function deployMessageHandler(message: DeployRequestMessage, context: WebviewContext) {
     // SAM already handles success/failure, so we only log that the user clicked the deploy button
@@ -58,21 +19,6 @@ export async function deployMessageHandler(message: DeployRequestMessage, contex
      * so we instead check for an undefined response to determine failure. The downside is that
      * we don't get failure reasons.
      */
-    let connection = Auth.instance.activeConnection
-    if (connection?.type !== 'iam' || connection?.state !== 'valid') {
-        await promptReauth(connection)
-        connection = Auth.instance.activeConnection
-        if (connection?.type !== 'iam' || connection?.state !== 'valid') {
-            const response: DeployResponseMessage = {
-                command: Command.DEPLOY,
-                messageType: MessageType.RESPONSE,
-                eventId: message.eventId,
-                isSuccess: false,
-            }
-            await context.panel.webview.postMessage(response)
-            return
-        }
-    }
     const result = (await vscode.commands.executeCommand('aws.samcli.sync', args, false)) as SamSyncResult
     if (result?.isSuccess) {
         void vscode.window.showInformationMessage('SAM Sync succeeded!')
--- a/src/applicationcomposer/messageHandlers/deployMessageHandler.ts
+++ b/src/applicationcomposer/messageHandlers/deployMessageHandler.ts
@@ -7,6 +7,45 @@ import vscode from 'vscode'
 import { Command, DeployRequestMessage, DeployResponseMessage, MessageType, WebviewContext } from '../types'
 import { SamSyncResult } from '../../shared/sam/sync'
 import { telemetry } from '../../shared/telemetry/telemetry'
+import { Auth } from '../../auth/auth'
+import { promptAndUseConnection } from '../../auth/utils'
+import { StatefulConnection } from '../../auth/connection'
+import * as nls from 'vscode-nls'
+import { showMessageWithUrl } from '../../shared/utilities/messages'
+import { credentialHelpUrl } from '../../shared/constants'
+
+const localize = nls.loadMessageBundle()
+
+async function promptReauth(connection?: StatefulConnection) {
+    let errorMessage = localize(
+        'aws.applicationComposer.deploy.authModal.message',
+        'Syncing requires authentication with IAM credentials.'
+    )
+    if (connection?.state === 'valid') {
+        errorMessage =
+            localize(
+                'aws.applicationComposer.deploy.authModal.invalidAuth',
+                'Authentication through Builder ID or IAM Identity Center detected. '
+            ) + errorMessage
+    }
+    const acceptMessage = localize(
+        'aws.applicationComposer.deploy.authModal.accept',
+        'Authenticate with IAM credentials'
+    )
+    const docMessage = localize('aws.applicationComposer.deploy.authModal.docLink', 'Open documentation')
+    const modalResponse = await showMessageWithUrl(
+        errorMessage,
+        credentialHelpUrl,
+        docMessage,
+        'info',
+        [acceptMessage],
+        true
+    )
+    if (modalResponse !== acceptMessage) {
+        return
+    }
+    await promptAndUseConnection(Auth.instance, 'iam-only')
+}
 
 export async function deployMessageHandler(message: DeployRequestMessage, context: WebviewContext) {
     // SAM already handles success/failure, so we only log that the user clicked the deploy button
@@ -19,6 +58,21 @@ export async function deployMessageHandler(message: DeployRequestMessage, contex
      * so we instead check for an undefined response to determine failure. The downside is that
      * we don't get failure reasons.
      */
+    let connection = Auth.instance.activeConnection
+    if (connection?.type !== 'iam' || connection?.state !== 'valid') {
+        await promptReauth(connection)
+        connection = Auth.instance.activeConnection
+        if (connection?.type !== 'iam' || connection?.state !== 'valid') {
+            const response: DeployResponseMessage = {
+                command: Command.DEPLOY,
+                messageType: MessageType.RESPONSE,
+                eventId: message.eventId,
+                isSuccess: false,
+            }
+            await context.panel.webview.postMessage(response)
+            return
+        }
+    }
     const result = (await vscode.commands.executeCommand('aws.samcli.sync', args, false)) as SamSyncResult
     if (result?.isSuccess) {
         void vscode.window.showInformationMessage('SAM Sync succeeded!')
--- a/src/auth/utils.ts
+++ b/src/auth/utils.ts
@@ -53,7 +53,7 @@ import { isValidCodeWhispererCoreConnection } from '../codewhisperer/util/authUt
 // TODO: Look to do some refactoring to handle circular dependency later and move this to ./commands.ts
 export const showConnectionsPageCommand = 'aws.auth.manageConnections'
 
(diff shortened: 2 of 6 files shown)


src/applicationcomposer/messageHandlers/deployMessageHandler.ts
```diff
@@ -3,10 +3,49 @@
...
 import { Command, DeployRequestMessage, DeployResponseMessage, MessageType, WebviewContext } from '../types'
 import { SamSyncResult } from '../../shared/sam/sync'
 import { telemetry } from '../../shared/telemetry/telemetry'
+import { Auth } from '../../auth/auth'
+import { promptAndUseConnection } from '../../auth/utils'
+import { openUrl } from '../../shared/utilities/vsCodeUtils'
+import { localize } from 'vscode-nls'
+import { StatefulConnection } from '../../auth/connection'
+
+const credentialsDocLink = 'https://docs.aws.amazon.com/toolkit-for-vscode/latest/userguide/setup-credentials.html'
```
justinmk3, line 16: any reason not to re-use this constant? https://github.com/aws/aws-toolkit-vscode/blob/2a1b31cf3218c3eea6fa56d69a0fbd7a894eaad1/src/shared/constants.ts#L41-L42

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
```diff
@@ -3,10 +3,49 @@
...
+
+async function promptReauth(connection?: StatefulConnection) {
```
justinmk3, line 18: we have many "reauth" prompts: 
image

can we re-use one of those, or extract one and generalize it a bit, so that AppCompser and e.g. Q and/or CodeWhisperer use the same reauth logic? That helps present a consistent UX, reduces bugs, increases quality, and reduces maintenance cost.

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
```diff
@@ -3,10 +3,49 @@
...
+    const docMessage = localize('aws.applicationComposer.deploy.authModal.docLink', 'Open documentation')
+    const modalResponse = await vscode.window.showInformationMessage(
+        errorMessage,
+        { modal: true },
+        acceptMessage,
+        docMessage
+    )
+    if (modalResponse === docMessage) {
+        await openUrl(Uri.parse(credentialsDocLink))
+    }
```
justinmk3, line 43: use `showMessageWithUrl` https://github.com/aws/aws-toolkit-vscode/blob/2a1b31cf3218c3eea6fa56d69a0fbd7a894eaad1/src/shared/utilities/messages.ts#L62

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
```diff
@@ -3,10 +3,49 @@
...
+            localize(
+                'aws.applicationComposer.deploy.authModal.invalidAuth',
+                'Authentication through Builder ID or IAM Identity Center detected. '
+            ) + errorMessage
+    }
+    const acceptMessage = localize(
+        'aws.applicationComposer.deploy.authModal.accept',
+        'Authenticate with IAM credentials'
+    )
+    const docMessage = localize('aws.applicationComposer.deploy.authModal.docLink', 'Open documentation')
```
justinmk3, line 34: not needed when using `showMessageWithUrl `

src/auth/utils.ts
```diff
@@ -53,8 +53,12 @@ import { isValidCodeWhispererCoreConnection } from '../codewhisperer/util/authUt
 // TODO: Look to do some refactoring to handle circular dependency later and move this to ./commands.ts
 export const showConnectionsPageCommand = 'aws.auth.manageConnections'
 
-export async function promptForConnection(auth: Auth, type?: 'iam' | 'sso'): Promise<Connection | void> {
-    const resp = await createConnectionPrompter(auth, type).prompt()
+export async function promptForConnection(
+    auth: Auth,
+    type?: 'iam' | 'sso',
+    excludeSso?: boolean
```
justinmk3, line 59: why can't `type = 'foo'` be used instead of adding a new `excludeSso` parameter ?

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
JLargent42: I missed that, will update

src/auth/utils.ts
JLargent42: I can add an "iam-only" type, just didn't want to change existing behavior outside of Composer.

justinmk3: something like "iam-only" would be preferred. cc @nkomonen-amazon

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
JLargent42: Our docs writer mentioned that "Open documentation" is preferred over "View documentation". Should this change be made across the entire Toolkit?

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
JLargent42: This makes sense here, just updated.

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
JLargent42: My understanding is that the most similar re-authentication flow to what we want here is the standard flow provided by `promptAndUseConnection`, which is what we're using here. I looked at the Q and CodeWhisperer auth flows, and they seem to be unique enough that I'm not sure a generalized function makes sense.

The main purpose of this function is to construct the error message, send the modal to the user, then trigger the existing auth flow if the user accepts the modal.

Is there something I'm missing here?

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
justinmk3: > Should this change be made across the entire Toolkit?

Not sure, but that should be a separate PR. The Toolkit should use consistent language.

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
justinmk3: > the most similar re-authentication flow to what we want here is the standard flow provided by `promptAndUseConnection`, which is what we're using here.

ok.

src/auth/utils.ts
```diff
@@ -375,7 +375,8 @@ export function createConnectionPrompter(auth: Auth, type?: 'iam' | 'sso') {
         return 2
     }
 
-    const prompter = createQuickPick(loadItems(), {
+    const excludeSso = type === 'iam-only'
```
justinmk3, line 378: Does 'iam' include sso connections? That seems confusing.

(Off-topic for this PR: maybe `type` should be a list instead of only 1 choice.)

JLargent42: Ideally, yes, `sam sync` and App Composer should support SSO. However, running `sam sync` with SSO auth currently fails and gives an ambiguous error message to the user. This is meant to be a quick fix to reduce user confusion as to why their sync isn't working.

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
```diff
@@ -32,19 +33,18 @@ async function promptReauth(connection?: StatefulConnection) {
...
+    const modalResponse = await showMessageWithUrl(
         errorMessage,
-        { modal: true },
-        acceptMessage,
-        docMessage
+        credentialHelpUrl,
+        docMessage,
+        'info',
+        [acceptMessage],
+        true
```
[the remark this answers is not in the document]
JLargent42: We want a modal here because the user has decided that they want to sync their IaC, and we need user action before we can proceed. The sync will not work until the user switches auth. The toast notification feels too easy to miss, as it is in the corner with other notifications and automatically disappears after some time.

From a code logic perspective, the toast notification would also work, but the `await` does not seem to complete execution when the notification disappears from timeout.

src/auth/utils.ts
justinmk3: This quick fix adds a maintenance cost because the code changes are spread out. That isn't easy to reverse, so isn't acceptable as a quick fix.

justinmk3: What does the `createConnectionPrompter` menu look like after this change?

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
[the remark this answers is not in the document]
justinmk3: > The sync will not work until the user switches auth. The toast notification feels too easy to miss, as it is in the corner with other notifications and automatically disappears after some time.

"Too easy to miss" is a feature. Modal dialogs are very rare in vscode, intentionally. The toast behavior is configurable. 

It doesn't sound like this needs to be modal, because the action will be canceled correctly, and this is only informational.

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
```diff
@@ -7,6 +7,45 @@ import vscode from 'vscode'
...
+    const modalResponse = await showMessageWithUrl(
+        errorMessage,
+        credentialHelpUrl,
+        docMessage,
+        'info',
+        [acceptMessage],
+        true
+    )
+    if (modalResponse !== acceptMessage) {
+        return
```
justinmk3, line 45: Instead of showing a modal, maybe a title/descriptive text could be passed to `promptAndUseConnection`

src/auth/utils.ts
justinmk3: Ok I get it now: the filter credentials menu will prompt the user during the sam-sync flow, making it easy for the user to "do the right thing".

This makes sense.

justinmk3 (approved):
Only remaining questions are:

- can we avoid the "modal" behavior (see comments)
- please make the problem/solution clearer
- off-topic for this PR: followup to find out why `createConnectionPrompter('iam')` does not already filter its list.

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
```diff
@@ -7,6 +7,45 @@ import vscode from 'vscode'
...
+    }
+    await promptAndUseConnection(Auth.instance, 'iam-only')
```
justinmk3, line 47: Could we make this change directly in https://github.com/aws/aws-toolkit-vscode/blob/ea875fc9af230a292e7a65b2a0d5f034a9314be2/src/shared/sam/sync.ts#L547-L548 instead of only for AppComposer?

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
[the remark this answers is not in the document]
JLargent42: The way it's currently implemented, it is not only informational. The user is marked as being "in progress of syncing template" until they explicitly cancel, and the re-authentication is considered as a step in the sync process. That's not to say we can't use a toast notification instead, but I think the flow would need to be a bit different to be intuitive.

Current flow for reauthenticating with modal:
Sync -> See modal -> Re-auth -> standard SAM Sync prompter

Given the passive nature of the toast notifications, I don't think it makes sense to include as part of the flow. Instead, I could see something like this being an option:
Sync -> See toast -> Re-auth -> Sync again -> standard SAM Sync prompter

Would the latter approach be preferable?

Galaoaoa (approved):

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
JLargent42: Yes, just moved it inside the SAM sync.

src/applicationcomposer/messageHandlers/deployMessageHandler.ts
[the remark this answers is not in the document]
justinmk3: What about this:

    Sync -> See toast (informational only) -> Prompt to select credentials -> standard SAM Sync prompter

src/shared/sam/sync.ts
```diff
@@ -535,6 +536,34 @@ export type SamSyncResult = {
...
+    )
+    if (connection?.state === 'valid') {
+        errorMessage =
+            localize(
+                'aws.sam.sync.authModal.invalidAuth',
+                'Authentication through Builder ID or IAM Identity Center detected. '
+            ) + errorMessage
+    }
+    const acceptMessage = localize('aws.sam.sync.authModal.accept', 'Authenticate with IAM credentials')
+    const docMessage = localize('aws.sam.sync.authModal.docLink', 'Open documentation')
```
justinmk3, line 552: Different parts of the Toolkit should not use different wording.

src/shared/sam/sync.ts
```diff
@@ -543,9 +572,15 @@ export function registerSync() {
...
+        let connection = Auth.instance.activeConnection
+        if (connection?.type !== 'iam' || connection?.state !== 'valid') {
+            await promptReauth(connection)
+            connection = Auth.instance.activeConnection
+            if (connection?.type !== 'iam' || connection?.state !== 'valid') {
+                throw new ToolkitError('Syncing SAM applications requires IAM credentials', {
+                    code: 'NoIAMCredentials',
+                })
+            }
         }
```
justinmk3, line 584: - If `promptReauth` returned the connection (or returned early if it's already valid), this logic could be greatly simplified.
- Also allows `const`. Removing `const` is usually a signal that logic could be improved.
- Renaming it to `getAuthOrPrompt` could make it clearer.


```suggestion
        const connection = await promptReauth(connection)
        if (!connection) {
            throw new ToolkitError('Syncing SAM applications requires IAM credentials', {
                code: 'NoIAMCredentials',
            })
        }
```

justinmk3 (approved):

PR #4254, opened 2024-01-10
tests: Introduce an e2e testing framework for Amazon Q

jpinkney-aws:
## Problem
- Amazon Q has to be manually tested on each release for everything e2e because we don't have any automation

## Solution
- Introduce a testing framework for Amazon Q that will allow teams to easily test Amazon Q e2e.
- Since VSCode doesn't have a built in way to do e2e tests against a webview, we instead hoist everything that would run inside of the webview into the current test instance, allowing us to test directly against the mynah UI instance programatically. This makes it so that we don't have to maintain any DOM elements + makes writing tests simple/easy to maintain while still testing all the same code flow e2e (minus the code that loads mynah ui into the webview)

## License


package.json
```diff
@@ -4368,6 +4369,7 @@
         "i18n-ts": "^1.0.5",
         "immutable": "^4.3.0",
         "js-yaml": "^4.1.0",
+        "jsdom": "^23.0.1",
```
justinmk3, line 4372: Interesting, jsdom is mentioned at https://github.com/aws/aws-toolkit-vscode/blob/4f9e41a8a13195d04de98df34af47ee8d794dea4/docs/ARCHITECTURE.md?plain=1#L212

justinmk3: can this be a devDependency? doesn't look like it's needed in production.

docs/TESTPLAN.md
```diff
@@ -41,6 +41,7 @@ The test suite has the following categories of tests:
 -   E2E Tests: **slow** tests
     -   Live in `src/testE2E`
     -   These tests are heavier than Integration tests.
+    -   Some E2E tests have a more complicated architecture. The document explaining that architecture can be found in [E2E_TESTING](./E2E_TESTING_ARCHITECTURE.md)
```
justinmk3, line 44:
```suggestion
    -   Some E2E tests have a more complicated architecture, described in [E2E_TESTING](./E2E_TESTING_ARCHITECTURE.md)
```

src/testE2E/amazonq/featureDev.test.ts
```diff
@@ -0,0 +1,127 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as got from 'got'
```
nkomonen-amazon, line 6: Not important for this PR, but for browser compatibility we are using the `fetch` implementation in `src/common/request.ts` instead of `got` (doesn't work in browser).

Just something to keep in mind going forward.

src/testE2E/amazonq/featureDev.test.ts
```diff
@@ -0,0 +1,127 @@
...
+import assert from 'assert'
+import { qTestingFramework } from './framework/framework'
+import { FollowUpTypes } from '../../amazonqFeatureDev/types'
+import sinon from 'sinon'
+import { FeatureDevClient } from '../../amazonqFeatureDev/client/featureDev'
+import { verifyTextOrder } from './framework/text'
+import { examples } from '../../amazonqFeatureDev/text'
+
+describe('Test Amazon Q Feature Dev', function () {
```
justinmk3, line 15: test names never need the word "test", "should", "correct", etc.
```suggestion
describe('Amazon Q Feature Dev', function () {
```

jpinkney-aws:
> Could this be used for non-mynah webviews? How much effort would that be (not in this PR, but curious about future potential)

Nik and I talked about that when I originally told him about the idea for using Mynah UI like this and I think it's possible. For vue its a little different since the toolkit uses SFCs but I think the same principles would apply. If we rendered the vue code in jsdom rather than the webview then any vscode event listeners would be straightforward to use and we should just be able to make some (or all?) calls directly to the UI.

I'd estimate 2 days for figuring out how it's possible for a specific webview and then 2 days for figuring out how to generalize it to all of the vue toolkit code and then many days for writing all the new tests

src/testE2E/amazonq/featureDev.test.ts
```diff
@@ -0,0 +1,135 @@
...
+            const q = framework.createTab()
+            const command = q.findCommand('/dev')
+            if (command.length > 0) {
+                assert.fail('Found command when it should not have been found')
+            }
+        })
+    })
+
+    describe('/dev {msg} entry', async () => {
+        it('Receives chat response', async () => {
```
sannicm, line 71: question: is this a duplicate from the test below?

src/testE2E/amazonq/framework/framework.ts
```diff
@@ -0,0 +1,101 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { JSDOM } from 'jsdom'
+/**
+ * JSDOM is used to help hoist MynahUI to running in a node environment vs in the browser (which is what it's made for)
+ */
+const dom = new JSDOM(undefined, {
```
grimalk, line 10: As a question, I'm not 100% sure.

I think unless you actually have usages, it will work in TS, but are you specifically intending for this to be a "side-effect of an import"?

We could extract this to a function to make this more explicit like calling "injectJSDomforMynah", I think it would be more maintanable

jpinkney-aws: The problem is that it uses some browser apis that aren't available in node. That's why we have to use jsdom

grimalk: yeah, I understand the jsdom part. My main question is about this being a side effect of an import (a.k.a. top level code) as opposed to being explicitly overriding global variables by calling a method.

src/testE2E/amazonq/framework/text.ts
```diff
@@ -0,0 +1,26 @@
...
+import * as assert from 'assert'
+import { ChatItem } from '@aws/mynah-ui'
+
+/**
+ * Verify that the text items in expectedText appears in the exact same order in chatItems
+ * @param chatItems An array of chat items in mynah UI
+ * @param expectedText An array of expected text items in order
+ * @returns true if the items in expectedText are found in the correct order in chatItems otherwise false
+ */
+export function verifyTextOrder(chatItems: ChatItem[], expectedText: string[]) {
```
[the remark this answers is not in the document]
grimalk: I'm not like completely opposed to the idea, but it still raises a few red flags.

E.g. imagine you randomly introduced a commit that adds additional message between the two, you'd never spot it with this method. 
What you're talking about signals more of a `filterOutTechnicalMessages(chatItems: ChatItems[]) -> ChatItems[]` to me.

But for now I don't think we should block on this

src/testE2E/amazonq/featureDev.test.ts
jpinkney-aws: Yes and no, this is a different entry point. Its when you are writing in the chat `/dev implement fibonacci`. The other one is when you first do `/dev` and then the new tab opens and then you do `implement fibonacci`

grimalk (approved):

hayemaxi (changes requested):
use our common `fetch` implementation, going forward we will not use `got` for browser compatibility

see #4291

justinmk3:
> Nik and I talked about that when I originally told him about the idea for using Mynah UI like this and I think it's possible. For vue its a little different

I just noticed that this unmerged feature-branch appears to implement that:
https://github.com/aws/aws-toolkit-vscode/compare/master...feature/vue-testing

We need to see if we can revive that branch.

grimalk (approved):

docs/E2E_TESTING_ARCHITECTURE.md
justinmk3, line 1: Please rename the file to `TEST_E2E.md` so that it is "grouped" next to TESTPLAN.md.

justinmk3 (approved):

src/testE2E/amazonq/featureDev.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    let framework: qTestingFramework
+
+    const samplePlanResponse = 'sample plan response'
+
+    beforeEach(() => {
+        /**
+         * TODO remove these stubs when we know the backend can handle all the test load + when we know the tests
+         * are working without any flakiness
+         */
+        sinon.stub(authUtil, 'getChatAuthState').resolves({
```
nkomonen-amazon, line 26: I'd recommend using `sinon.createSandbox()`, then stubbing with that sandbox, and using sandbox.restore() in the after. Since you are stubbing static instances it can cross in to other tests and cause issues in the future.

PR #4256, opened 2024-01-10
fix(gumby): surround path with double quotes

dhasani23:
## Problem

A PR which was very recently merged () set `shell: true` for a shell command we run to determine the Java version of a user's project (this was necessary to resolve a customer issue). An unintended side effect of this recent change is that paths containing spaces in them are not handled properly.

## Solution

Surround the path with double quotes before running the shell command.

## License

### DIFF
--- a/src/codewhisperer/service/transformByQHandler.ts
+++ b/src/codewhisperer/service/transformByQHandler.ts
@@ -110,7 +110,7 @@ export async function validateProjectSelection(project: vscode.QuickPickItem) {
         })
         throw new TransformByQJavaProjectNotFound()
     }
-    const classFilePath = compiledJavaFiles[0].fsPath
+    const classFilePath = `"${compiledJavaFiles[0].fsPath}"`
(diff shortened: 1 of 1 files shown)


tincheng (approved):

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -110,7 +110,7 @@ export async function validateProjectSelection(project: vscode.QuickPickItem) {
         })
         throw new TransformByQJavaProjectNotFound()
     }
-    const classFilePath = compiledJavaFiles[0].fsPath
+    const classFilePath = `"${compiledJavaFiles[0].fsPath}"`
```
justinmk3, line 113: This is why `shell: true` should be avoided. Shell-specific quoting is very difficult to do reliably. Users may be using fish, zsh, etc. If this `"` delimited path has `\` slashes, `$`, or other special chars, those will be interpreted by the shell.

PR #4272, opened 2024-01-11
build: polyfill crypto module when in the browser

nkomonen-amazon:
polyfills crypto module when in the browser with a browser compatible version

## License


package.json
```diff
@@ -4361,6 +4361,7 @@
         "bytes": "^3.1.2",
         "cross-fetch": "^4.0.0",
         "cross-spawn": "^7.0.3",
+        "crypto-browserify": "^3.12.0",
```
justinmk3, line 4364: I thought the point of `src/browser/package.json` was to avoid needing this in both places (or vice versa)?

nkomonen-amazon: Good find, I must have added this here accidentally. Updated to just src/browser/package.json

justinmk3 (approved):

PR #4273, opened 2024-01-11
refactor: mono function in amazon q code transform

damntrecky:
## Problem
Refactor our mono function as part of our tech debt backlog. The goal is to create a semantically more readable transformation process and help testing out as a secondary step. This PR should NOT introduce new functionality to the users.

- one side effect to update telemetry output as well as VSCode outputs the major version of Java, instead of a human readable version. IntelliJ logs this properly

## License

### DIFF
--- a/src/amazonqGumby/telemetry/codeTransformTelemetry.ts
+++ b/src/amazonqGumby/telemetry/codeTransformTelemetry.ts
@@ -69,7 +69,7 @@ export const JDKToTelemetryValue = (
  * can read the version easily and for VSCode we must exec
  * the javap -v command.
  */
-export const javaPCommandLineToTelemetryValue = (javapCommandLineOutput: string) => {
+export const javapOutputToTelemetryValue = (javapCommandLineOutput: string) => {
     switch (javapCommandLineOutput) {
         case '49':
             return 'JDK_1_5'
--- a/src/amazonqGumby/telemetry/codeTransformTelemetry.ts
+++ b/src/amazonqGumby/telemetry/codeTransformTelemetry.ts
@@ -63,4 +63,54 @@ export const JDKToTelemetryValue = (
     }
 }
 
+/**
+ * @description We want the output of our Java versions found
+ * that are not supported to match IntelliJ output. IntelliJ
+ * can read the version easily and for VSCode we must exec
+ * the javap -v command.
+ */
+export const javaPCommandLineToTelemetryValue = (javapCommandLineOutput: string) => {
+    switch (javapCommandLineOutput) {
+        case '49':
+            return 'JDK_1_5'
+        case '50':
+            return 'JDK_1_6'
+        case '51':
+            return 'JDK_1_7'
+        case '52':
+            return 'JDK_1_8'
+        case '53':
+            return 'JDK_1_9'
+        case '54':
+            return 'JDK_10'
+        case '55':
+            return 'JDK_11'
+        case '56':
+            return 'JDK_12'
+        case '57':
+            return 'JDK_13'
+        case '58':
+            return 'JDK_14'
+        case '59':
+            return 'JDK_15'
+        case '60':
+            return 'JDK_16'
+        case '61':
+            return 'JDK_17'
+        case '62':
+            return 'JDK_18'
+        case '63':
+            return 'JDK_19'
+        case '64':
+            return 'JDK_20'
+        case '65':
+            return 'JDK_21'
+        case '66':
+            return 'JDK_22'
+        default:
+            // If nothing found. Output the number and lookup the java 'major version' numbers online
+            return javapCommandLineOutput
+    }
+}
+
 export const calculateTotalLatency = (startTime: number, endTime: number = Date.now()): number => endTime - startTime
--- a/src/codewhisperer/commands/startTransformByQ.ts
+++ b/src/codewhisperer/commands/startTransformByQ.ts
@@ -100,8 +100,6 @@ async function pickProject(
     transformByQState.setProjectName(encodeHTML(state.project.label)) // encode to avoid HTML injection risk
 }
 
-let payloadFileName = ''
-
 export async function startTransformByQ() {
     let intervalId = undefined
 
@@ -114,7 +112,10 @@ export async function startTransformByQ() {
     try {
         // Set web view UI to poll for progress
         intervalId = setInterval(() => {
-            void vscode.commands.executeCommand('aws.amazonq.showPlanProgressInHub', startTime.getTime())
+            void vscode.commands.executeCommand(
+                'aws.amazonq.showPlanProgressInHub',
+                codeTransformTelemetryState.getStartTime()
+            )
         }, CodeWhispererConstants.progressIntervalMs)
 
         // step 1: CreateCodeUploadUrl and upload code
@@ -130,6 +131,7 @@ export async function startTransformByQ() {
         // step 4: poll until artifacts are ready to download
         const status = await pollTransformationTillComplete(jobId)
 
+        // Set the result state variables for our store and the UI
         await setTransformationJobStatus(status)
     } catch (error) {
         await modernizationJobErrorHandler(error)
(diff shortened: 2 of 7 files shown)


src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -97,18 +97,172 @@ async function pickProject(
     transformByQState.setProjectName(encodeHTML(state.project.label)) // encode to avoid HTML injection risk
 }
 
+let intervalId = undefined
```
damntrecky, line 100: I still have to refactor these variables. I move them as "globals" but deciding on using them as params or adding them to our object/store/singleton state.

Thoughts welcome!

JuliaABurch: It does seem to suggest that we could encapsulate this into a class - `TransformationJob` (for example.) I'd suggest overall moving towards instance variables instead of global/sequence as it would get us closer to letting customers run multiple transformation jobs at once

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -97,18 +97,172 @@ async function pickProject(
...
+let resultStatusMessage = ''
+let payloadFileName = ''
+let startTime = new Date()
+
+export async function preTransformUploadCode(userInputState: UserInputState) {
+    await vscode.commands.executeCommand('aws.amazonq.refresh')
+    await vscode.commands.executeCommand('aws.amazonq.transformationHub.focus')
+
+    let uploadId = ''
+    throwIfCancelled()
```
JuliaABurch, line 111: This isn't necessary for your current PR, but I was thinking about all of these repeated `refresh` and `throwIfCancelled()` commands - perhaps steps can be encapsulated in such a way that we can throw them into a list and, iterating over them, automatically call these update methods after each one. (Basically a simplified state machine.)

src/codewhisperer/commands/startTransformByQ.ts
damntrecky: I moved the variables into the transformByQState.

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -97,18 +97,172 @@ async function pickProject(
...
+        // step 3 (intermediate step): show transformation-plan.md file
+        // TO-DO: on IDE restart, resume here if a job was ongoing
+        await pollTransformationStatusTillPlanReady(jobId)
+
+        // step 4: poll until artifacts are ready to download
+        const status = await pollTransformationTillComplete(jobId)
+
+        await setTransformationJobStatus(status)
+    } catch (error) {
+        await modernizationJobErrorHandler()
```
tincheng: Maybe pass the error into the ErrorHandler and Log the error? I think we have discussed that log is local in the IDE and not sent to the cloud, so should not safe to include original stack trace. If this assumption is not correct please let me know!

damntrecky: Absolutely agree. If we log to users local file system this would be okay.

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -97,18 +97,172 @@ async function pickProject(
...
+    } else {
+        resultStatusMessage = 'JobCompletedSuccessfully'
+    }
+
+    await vscode.commands.executeCommand('aws.amazonq.transformationHub.reviewChanges.reveal')
+    await vscode.commands.executeCommand('aws.amazonq.refresh')
+    sessionPlanProgress['returnCode'] = StepProgress.Succeeded
+}
+
 export async function startTransformByQ() {
```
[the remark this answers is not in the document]
damntrecky: sure.

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -97,18 +97,172 @@ async function pickProject(
...
+    } catch (error) {
+        errorMessage = 'Failed to get transformation job status'
+        getLogger().error(errorMessage, error)
+        throw new ToolkitError(errorMessage, { cause: error as Error })
+    }
+
+    return status
+}
+
+export async function setTransformationJobStatus(status: string) {
```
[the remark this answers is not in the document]
damntrecky: Makes sense. Agree set/get are more reserved for a specific functionality.

src/amazonqGumby/telemetry/codeTransformTelemetry.ts
```diff
@@ -63,4 +63,54 @@ export const JDKToTelemetryValue = (
     }
 }
 
+/**
```
damntrecky, line 66: Tried to just add as much as I can from the oracle docs here -> https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.1-200-B.2

src/codewhisperer/commands/startTransformByQ.ts
damntrecky: Are you suggesting throwing the "steps" into a list as a state machine?

Sounds like an interesting discussion. For this refactor I think its out of scope.

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -153,196 +321,94 @@ export async function startTransformByQ() {
...
+
+export async function cleanupTransformJob(intervalId: NodeJS.Timeout | undefined) {
     clearInterval(intervalId)
-    transformByQState.setToNotStarted() // so that the "Transform by Q" button resets
-    transformByQState.setPolledJobStatus('') // reset polled job status too
+    transformByQState.setJobDefaults()
     await vscode.commands.executeCommand('setContext', 'gumby.isStopButtonAvailable', false)
     await vscode.commands.executeCommand('aws.amazonq.refresh')
-    void vscode.commands.executeCommand('aws.amazonq.showPlanProgressInHub', startTime.getTime())
+    void vscode.commands.executeCommand('aws.amazonq.showPlanProgressInHub', codeTransformTelemetryState.getStartTime())
```
justinmk3, line 411: I see lots of executeCommand() usage in this PR and related code. Commands are for users, except in rare cases where a module can't get a reference to a function. In most cases, it's strongly advised to call functions directly, that results in a more useful stacktrace and type-checking.

damntrecky: From what I'm aware of, some of these are using this execute some updates in our around the bottom panel webview. 
Where is a good example for the best practice on this?

The refresh commands scattered throughout are to update the native VSCode tree UI with status as well. I don't think there is another way to do this?

I think your comments are important and we should definitely continue the discussion here or elsewhere. This refactor is more for just semantic and structural organization of our mono function.

Screenshot 2024-01-12 at 8 08 56 AM

Screenshot 2024-01-12 at 8 10 06 AM

justinmk3: > Where is a good example for the best practice on this?

The best practice is to just call functions instead of using `executeCommand`.

The commands are implemented by functions. Import those functions and call them, instead of using `executeCommand`.

> The refresh commands scattered throughout are to update the native VSCode tree UI with status as well. I don't think there is another way to do this?

For now we can ignore the "refresh" commands since that is a hack that carried over from codewhisperer. But for most other commands there is no reason to use `executeCommand`.

src/amazonqGumby/telemetry/codeTransformTelemetry.ts
```diff
@@ -63,4 +63,54 @@ export const JDKToTelemetryValue = (
...
+ * @description We want the output of our Java versions found
+ * that are not supported to match IntelliJ output. IntelliJ
+ * can read the version easily and for VSCode we must exec
+ * the javap -v command.
+ */
+export const javapOutputToTelemetryValue = (javapCommandLineOutput: string) => {
```
leigaol, line 72: what if  the `javap` is not a valid CLI when running `javap -v`

tincheng (approved):

src/amazonqGumby/telemetry/codeTransformTelemetry.ts
damntrecky: Then it fails. This is already the current behavior. It will report undefined as we cannot identify the users Java version.

This is non blocking and we have an open discussion on identifying Java versions internally.

damntrecky: This code wont execute if the javap command fails. It will go to the catch block.

PR #4274, opened 2024-01-11
refactor: CI test run scripts

nkomonen-amazon:
Just doing some tidying up after learning how our test execution works.

- Behaviour is untouched
- Methods are reordered
- 'export' is removed from methods that do not need it
- runToolkitTests() is generalized to remove duplication in scripts/test/test.ts

## License

### DIFF
--- a/scripts/test/launchTestUtilities.ts
+++ b/scripts/test/launchTestUtilities.ts
@@ -9,6 +9,7 @@ import { downloadAndUnzipVSCode, resolveCliArgsFromVSCodeExecutablePath } from '
 import { join, resolve } from 'path'
 import { runTests } from '@vscode/test-electron'
 import { VSCODE_EXTENSION_ID } from '../../src/shared/extensions'
+import { TestOptions } from '@vscode/test-electron/out/runTest'
 
 const envvarVscodeTestVersion = 'VSCODE_TEST_VERSION'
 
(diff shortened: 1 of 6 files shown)


scripts/test/test.ts
```diff
@@ -4,35 +4,7 @@
...
-        })
-
-        console.log(`Finished running Main test suite with result code: ${result}`)
-        process.exit(result)
-    } catch (err) {
-        console.error(err)
-        console.error('Failed to run tests')
-        process.exit(1)
-    }
+    await runToolkitTests('unit', resolve('dist', 'src', 'test', 'index.js'))
```
nkomonen-amazon, line 9: We can now use the generalized method and de-duplicate code.

scripts/test/test.ts
```diff
@@ -4,35 +4,7 @@
...
-        const rootDir = process.cwd()
-        const testEntrypoint = resolve(rootDir, 'dist/src/test/index.js')
-        const testWorkspace = resolve(rootDir, 'src/testFixtures/workspaceFolder')
-        console.log(`Starting tests: ${testEntrypoint}`)
-
-        const result = await runTests({
-            vscodeExecutablePath: vsCodeExecutablePath,
-            extensionDevelopmentPath: rootDir,
-            extensionTestsPath: testEntrypoint,
-            // For verbose VSCode logs, add "--verbose --log debug". c2165cf48e62c
```
justinmk3, line 23: this comment is non-obvious. would prefer to keep it somewhere

scripts/test/launchTestUtilities.ts
```diff
@@ -75,45 +201,9 @@ export async function invokeVSCodeCli(vsCodeExecutablePath: string, args: string
...
 /**
- * Alternative to the `--disable-extensions` CLI flag which allows us to
- * selectively keep some extensions enabled.
- *
- * @param vsCodeExecutablePath Path to vscode CLI program.
- * @param exceptIds List of extension ids that should *not* be disabled.
- *
- * @returns List of args which the caller is expected to pass to vscode CLI, of the form:
- * ["--disable-extension", "foo.bar.baz", "--disable-extension", ...]
+ * TODO: Look to not export this, but a test needs it currently
```
justinmk3, line 205: the tests should use this instead: https://github.com/aws/aws-toolkit-vscode/blob/02f5ef9f8ee2ac4669c27e5d73d22abc446c18ae/src/shared/vscode/env.ts#L69

and maybe we can eliminate this one.

scripts/test/launchTestUtilities.ts
```diff
@@ -17,8 +18,121 @@ const minimum = 'minimum'
...
+ * Alternative to the `--disable-extensions` CLI flag which allows us to
+ * selectively keep some extensions enabled.
+ *
+ * @param vsCodeExecutablePath Path to vscode CLI program.
+ * @param exceptIds List of extension ids that should *not* be disabled.
+ *
+ * @returns List of args which the caller is expected to pass to vscode CLI, of the form:
+ * ["--disable-extension", "foo.bar.baz", "--disable-extension", ...]
+ */
+async function getCliArgsToDisableExtensions(
```
justinmk3, line 104: this and `setupVSCodeInstance` look almost the same as the old code. is there any way to avoid the diff-churn?

scripts/test/launchTestUtilities.ts
```diff
@@ -17,8 +18,121 @@ const minimum = 'minimum'
...
-export const integrationSuite = 'integration'
-export const e2eSuite = 'e2e'
+type SuiteName = 'integration' | 'e2e' | 'unit'
+
+/**
+ * This is the generalized method that is used by different test suites (unit, integration, ...) in CI to
+ * setup vscode and then run tests. An important thing to note is that in CI we do not have VS Code installed,
+ * so this script needs to do this itself.
+ *
+ * If you want to run the tests manually you should use the `Run & Debug` menu in VS Code instead.
```
justinmk3, line 28: Why? I used `npm run test` heavily while debugging the "unresolved promise" issues. If that is no longer possible then that's a regression. CLI/terminal usage is important, it can really speed up debugging.

scripts/test/test.ts
justinmk3, line 9:
```suggestion
    await runToolkitTests('unit', resolve('dist/src/test/index.js'))
```

justinmk3 (approved):

scripts/test/launchTestUtilities.ts
nkomonen-amazon: `npm run test` still works, you just don't get debugging with it

scripts/test/launchTestUtilities.ts
```diff
@@ -114,67 +196,11 @@ export async function getCliArgsToDisableExtensions(
     return ids
 }
 
-export function getMinVsCodeVersion(): string {
+function getMinVsCodeVersion(): string {
     const vsCodeVersion = packageJson.engines.vscode
 
     // We assume that we specify a minium, so it matches ^<number>, so remove ^'s
     const sanitizedVersion = vsCodeVersion.replace('^', '')
     console.log(`Using minimum VSCode version specified in package.json: ${sanitizedVersion}`)
```
justinmk3, line 204: This log message is the only thing that differs from  `shared/vscode/env.ts:getMinVscodeVersion`. So this function can re-use that one, and only this log message would be added here.

scripts/test/launchTestUtilities.ts
```diff
@@ -114,67 +196,11 @@ export async function getCliArgsToDisableExtensions(
...
     return sanitizedVersion
```
justinmk3, line 205:
```suggestion
    const sanitizedVersion = getMinVscodeVersion()
    console.log(`Using minimum VSCode version specified in package.json: ${sanitizedVersion}`)
    return sanitizedVersion
```

nkomonen-amazon: I was assuming that we do not want to import our extension code in to the scripts. But if there isn't a reason not to then this makes sense.

justinmk3 (approved):

scripts/test/launchTestUtilities.ts
nkomonen-amazon: fixed

scripts/test/launchTestUtilities.ts
justinmk3: There is no reason not to. Except code that requires Toolkit init/activation, perhaps.

PR #4280, opened 2024-01-12
telemetry: appcomposer_initializeProject metric

bildungsroman:
## Problem

We're tracking resource counts in the App Composer console, but not in the VS Code extension. This allows for tracking that telemetry event in VSC as well.

## Solution

This adds the `AppcomposerResourceCount` telemetry event, which records the `resourceCounts` and `totalResourceCount` values coming from `FigAssets`.

No testable changes added, as this uses existing telemetry tracking mechanisms.

## License

### DIFF
--- a/packages/core/src/shared/telemetry/vscodeTelemetry.json
+++ b/packages/core/src/shared/telemetry/vscodeTelemetry.json
@@ -115,6 +115,16 @@
             "description": "Whether the user saved"
         },
         {
+            "name": "resourceCounts",
+            "type": "string",
+            "description": "Per-resource count of each resource in an opened template"
+        },
+        {
+            "name": "totalResourceCount",
+            "type": "int",
+            "description": "Total count of resources in an opened template"
+        },
+        {
             "name": "cwsprChatTriggerInteraction",
             "type": "string",
             "allowedValues": ["hotkeys", "click", "contextMenu"],
@@ -605,6 +615,11 @@
             "name": "appcomposer_closeWfs",
             "description": "Called when Step Functions Workflow Studio is closed",
             "metadata": [{ "type": "didSave" }]
+        },
+        {
+            "name": "appcomposer_resourceCount",
+            "description": "Called to count resources in a template when it is opened",
+            "metadata": [{ "type": "resourceCounts" }, { "type": "totalResourceCount" }]
         },
         {
             "name": "amazonq_openChat",
--- a/packages/core/src/shared/telemetry/vscodeTelemetry.json
+++ b/packages/core/src/shared/telemetry/vscodeTelemetry.json
@@ -617,7 +617,7 @@
             "metadata": [{ "type": "didSave" }]
         },
         {
-            "name": "appcomposer_initializeProjectSucceeded",
+            "name": "appcomposer_initializeProject",
             "description": "Called when a project successfully loads",
             "metadata": [{ "type": "resourceCounts" }, { "type": "totalResourceCount" }]
         },
--- a/packages/core/src/shared/telemetry/vscodeTelemetry.json
+++ b/packages/core/src/shared/telemetry/vscodeTelemetry.json
@@ -617,8 +617,8 @@
             "metadata": [{ "type": "didSave" }]
         },
         {
-            "name": "appcomposer_resourceCount",
-            "description": "Called to count resources in a template when it is opened",
+            "name": "appcomposer_initializeProjectSucceeded",
+            "description": "Called when a project successfully loads",
             "metadata": [{ "type": "resourceCounts" }, { "type": "totalResourceCount" }]
         },
         {
--- a/packages/core/src/applicationcomposer/messageHandlers/emitTelemetryMessageHandler.ts
+++ b/packages/core/src/applicationcomposer/messageHandlers/emitTelemetryMessageHandler.ts
@@ -17,7 +17,7 @@ import {
     AppcomposerOpenWfs,
     AppcomposerPostProcess,
     AppcomposerRegenerateClicked,
-    AppcomposerResourceCount,
+    AppcomposerInitializeProjectSucceeded,
     telemetry,
 } from '../../shared/telemetry/telemetry'
 import { getLogger } from '../../shared/logger'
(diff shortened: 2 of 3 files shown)


src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -114,6 +114,16 @@
             "type": "boolean",
             "description": "Whether the user saved"
         },
+        {
+            "name": "resourceCounts",
+            "type": "string",
```
JLargent42, line 119: Not sure about the best way to do this, but do we have a way to extract this information in Kibana into readable graphs, given that this is a string type?

bildungsroman: That is a Carl question, but I'll follow up with him

src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -601,6 +611,11 @@
             "description": "Called when Step Functions Workflow Studio is closed",
             "metadata": [{ "type": "didSave" }]
         },
+        {
+            "name": "appcomposer_resourceCount",
+            "description": "Called to count resources in a template when it is opened",
+            "metadata": [{ "type": "resourceCounts" }, { "type": "totalResourceCount" }]
```
justinmk3, line 617: Can this be on the `appcomposer_openTemplate` event? That is the recommended practice. There is no need to add a new event for every bit of info that is needed. Events should map to "actions". Events can have arbitrary metadata. Adding a new event just to track metadata should be avoided.

src/applicationcomposer/messageHandlers/emitTelemetryMessageHandler.ts
```diff
@@ -61,6 +62,9 @@ export function emitTelemetryMessageHandler(message: EmitTelemetryMessage) {
             case 'CLOSE_WFS':
                 sendCloseWfs(parsedData as AppcomposerCloseWfs)
                 return
+            case 'TEMPLATE_OPENED':
+                sendResourceCounts(parsedData as AppcomposerResourceCount)
```
justinmk3, line 66: why doesn't this emit the existing `appcomposer_openTemplate` event?

bildungsroman: Despite the naming, they are two separate events. The existing `appcomposer_openTemplate` event requires auth data that this event does not have access to

src/shared/telemetry/vscodeTelemetry.json
bildungsroman: At this point in the code I don't have access to the `hasChatAuth` param required for `appcomposer_openTemplate`. We are getting this event from `FigAssets`, which is environment-agnostic and doesn't have access to VSCode actions

bildungsroman: This PR is mainly to pass that data along. I could rename the `TEMPLATE_OPENED` event to `COUNT_RESOURCES` - the only reason I'm using `TEMPLATE_OPENED` is that's how it's set up in `FigAssets`

src/applicationcomposer/messageHandlers/emitTelemetryMessageHandler.ts
bildungsroman: @justinmk3 Could I get a re-review of this PR?

justinmk3: I still don't understand why a new event is being added. `appcomposer_resourceCount` is not an "event" it's just data. Is there an event this data can be attached to?

JLargent42 (approved):

packages/toolkit/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -601,6 +611,11 @@
             "description": "Called when Step Functions Workflow Studio is closed",
             "metadata": [{ "type": "didSave" }]
         },
+        {
+            "name": "appcomposer_initializeProjectSucceeded",
```
justinmk3, line 615: Note that all metrics can pass/fail. Can this be named `appcomposer_initializeProject` instead. Even if it's not possible to fail currently, that avoids needing another `appcomposer_initializeProjectFailed` metric in the future (which is an antipattern, for at least one big reason: it's harder to detect failures in our monitoring if each "failure" state is bespoke)

packages/toolkit/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -601,6 +611,11 @@
             "description": "Called when Step Functions Workflow Studio is closed",
             "metadata": [{ "type": "didSave" }]
         },
+        {
+            "name": "appcomposer_initializeProject",
```
justinmk3, line 615: As a followup, do you plan to move these local definitions into common? https://github.com/aws/aws-toolkit-common/blob/main/telemetry/definitions/commonDefinitions.json

justinmk3 (approved):

packages/toolkit/src/shared/telemetry/vscodeTelemetry.json
bildungsroman: Ah, I wasn't aware of those. I'd need to ask the team if that's something we want to do

bildungsroman:
Not sure why Windows tests are failing, as I rebased on current `master` and the build passed locally

packages/toolkit/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -601,6 +611,11 @@
...
+            "description": "Called when a project successfully loads",
```
justinmk3, line 616:
```suggestion
            "description": "Called when a project loads",
```

bildungsroman:
Hi, can you please merge this PR @justinmk3

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -114,6 +114,16 @@
             "type": "boolean",
             "description": "Whether the user saved"
         },
+        {
+            "name": "resourceCounts",
+            "type": "string",
+            "description": "Per-resource count of each resource in an opened template"
+        },
+        {
+            "name": "totalResourceCount",
```
justinmk3, line 123: these names are confusing

PR #4285, opened 2024-01-16
feat(sam): improve rendering of SAM output

justinmk3:
## Problem

`vscode.LogOutputChannel` is now available via `createOutputChannel(..., { log: true })`, for improved highlighting and presentation of log-like output in VSCode, but AWS Toolkit still uses the old `vscode.OutputChannel` interface.

image

## Solution

- Use `vscode.LogOutputChannel` when possible.
- Change SAM output to "info" level instead of "error" level.

image

## License

### DIFF
--- a/src/shared/sam/cli/samCliLocalInvoke.ts
+++ b/src/shared/sam/cli/samCliLocalInvoke.ts
@@ -80,7 +80,7 @@ export class DefaultSamLocalInvokeCommand implements SamLocalInvokeCommand {
                         this.logger.verbose('SAM: pid %d: stdout: %s', childProcess.pid(), removeAnsi(text))
                     },
                     onStderr: (text: string): void => {
-                        getLogger('debugConsole').error(text, { raw: true })
+                        getLogger('debugConsole').info(text, { raw: true })
                         // If we have a timeout (as we do on debug) refresh the timeout as we receive text
                         params.timeout?.refresh()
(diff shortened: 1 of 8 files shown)


src/shared/sam/cli/samCliLocalInvoke.ts
```diff
@@ -80,7 +80,7 @@ export class DefaultSamLocalInvokeCommand implements SamLocalInvokeCommand {
                         this.logger.verbose('SAM: pid %d: stdout: %s', childProcess.pid(), removeAnsi(text))
                     },
                     onStderr: (text: string): void => {
-                        getLogger('debugConsole').error(text, { raw: true })
+                        getLogger('debugConsole').info(text, { raw: true })
```
hayemaxi, line 83: why change to info on error?

justinmk3: Previously this had no UI impact. After the change in this PR, vscode prefixes this output with `[error]`, which is misleading in most cases. `stderr` is often used as a "side channel" and for SAM CLI often doesn't indicate error conditions. For example, most of the lines in the screenshot shown above are from stderr, but they aren't errors.

So on balance, I think it's less confusing for customers if we don't log `stderr` as error. However, a future improvement would be to check the output using something like this: https://github.com/aws/aws-toolkit-vscode/blob/a7745ae881f2e3b099f636dc5ab49d7b39e46f2e/src/shared/sam/cli/samCliInvokerUtils.ts#L70-L77

and log any matching messages as `error()`.

PR #4286, opened 2024-01-16
fix(gumby): set shell to false

dhasani23:
## Problem

A previous PR set shell to true for a shell command we run. This can lead to issues since special characters such as $ will be interpreted by the shell.

A smaller issue: the summary.md file was not opening automatically due to the "await" keyword.

## Solution

Set shell to false. Remove "await" keyword.

## License

### DIFF
--- a/src/codewhisperer/service/transformationResultsViewProvider.ts
+++ b/src/codewhisperer/service/transformationResultsViewProvider.ts
@@ -362,7 +362,7 @@ export class ProposedTransformationExplorer {
             })
 
             // Do not await this so that the summary reveals without user needing to close this notification
-            vscode.window.showInformationMessage(CodeWhispererConstants.viewProposedChangesMessage)
+            void vscode.window.showInformationMessage(CodeWhispererConstants.viewProposedChangesMessage)
             await vscode.commands.executeCommand('aws.amazonq.transformationHub.summary.reveal')
         })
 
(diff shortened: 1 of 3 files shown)


src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -302,8 +302,8 @@ export class ProposedTransformationExplorer {
                     codeTransformApiNames: 'ExportResultArchive',
                     codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
                     codeTransformJobId: transformByQState.getJobId(),
-                    codeTransformApiErrorMessage: e?.message || errorMessage,
-                    codeTransformRequestId: e?.requestId,
+                    codeTransformApiErrorMessage: (e as Error).message ?? errorMessage,
```
nkomonen-amazon, line 305: If you put something like this right after the `catch`:

```typescript
if (!(e instanceof Error)) {
  throw e
}
```
then anything after that will have `e` typed as an `Error` and you can skip `e as Error`

nkomonen-amazon (approved):

src/codewhisperer/service/transformationResultsViewProvider.ts
dhasani23: Makes sense, but we also use `e as Error` in many places (search here: https://github.com/aws/aws-toolkit-vscode/blob/master/src/codewhisperer/service/transformByQHandler.ts), so maybe it's better if I leave this as it is for consistency? Either that or change it everywhere, not sure which is better practice

nkomonen-amazon: We can leave it as is for now, don't want to block. But in the future if you need to use it a lot the above snippet will help.

PR #4287, opened 2024-01-16
browser: remove non-complex uses of fs-extra for codewhisperer

hayemaxi:
Continuation of https://github.com/aws/aws-toolkit-vscode/pull/4230

- Adds wrapper for `copy`
- Adds function for deleting something and swalling error if it doesn't exist.
- Update tests cases where immediately relevant.
- Refactors some code from functional style to loops, so that they can be awaited.

Does not update `dependencyGraph.ts` as it contains some complex uses of `fs-extra`, such as sym
links.

#### Testing
Manual testing in progress.

## License

### DIFF
--- a/src/codewhisperer/service/transformByQHandler.ts
+++ b/src/codewhisperer/service/transformByQHandler.ts
@@ -10,7 +10,6 @@ import { getLogger } from '../../shared/logger'
 import { CreateUploadUrlResponse } from '../client/codewhispereruserclient'
 import { sleep } from '../../shared/utilities/timeoutUtils'
 import * as CodeWhispererConstants from '../models/constants'
-import * as fs from 'fs-extra'
 import * as path from 'path'
 import * as os from 'os'
 import * as vscode from 'vscode'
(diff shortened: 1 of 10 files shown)


src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -402,16 +407,16 @@ export async function zipCode(modulePath: string) {
     throwIfCancelled()
 
     const tempFilePath = path.join(os.tmpdir(), 'zipped-code.zip')
-    fs.writeFileSync(tempFilePath, zip.toBuffer())
+    await fsCommon.writeFile(tempFilePath, zip.toBuffer())
```
hayemaxi, line 410: `fsCommon.writeFile` writes in `utf-8` encoding... may need to look back at some of these replacements to ensure that is correct

justinmk3:
Needs conflict resolution

src/codewhisperer/service/transformByQHandler.ts
hayemaxi: `fsCommon.writeFile` does string vs buffer distinction 👍

justinmk3:
> `copy` in VSCode's fs api is not supported in C9

does it require a specific vscode version , and if so, which one?

justinmk3:
Is this still relevant?

justinmk3:
continued in https://github.com/aws/aws-toolkit-vscode/pull/5712

Issue #4238, opened 2024-01-08
unreliable test: codewhisperer "onInlineAcceptance"


justinmk3:
flaky test sometimes fails:

```
  1) onInlineAcceptance
       onInlineAcceptance
         Should report telemetry that records this user decision event:

      AssertionError [ERR_ASSERTION]: telemetry item 1 (of 1) not found (in the expected order) for metric name: "codewhisperer_userDecision" 
      + expected - actual

         "codewhispererRequestId": "test"
         "codewhispererSessionId": "test"
         "codewhispererSuggestionIndex": "0"
         "codewhispererSuggestionReferenceCount": "0"
      -  "codewhispererSuggestionState": "Reject"
      +  "codewhispererSuggestionState": "Accept"
         "codewhispererTriggerType": "OnDemand"
         "codewhispererUserGroup": "Classifier"
         "credentialStartUrl": "testStartUrl"
       }
      
      at partialDeepCompare (src/test/testUtil.ts:196:12)
      at assertTelemetry (src/test/testUtil.ts:237:9)
      at /Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/src/test/testUtil.ts:253:9
      at Context.<anonymous> (src/test/codewhisperer/commands/onInlineAcceptance.test.ts:90:13)
```

justinmk3:
also happens with onAcceptance :

```
  1) onAcceptance
       onAcceptance
         Should report telemetry that records this user decision event:

      AssertionError [ERR_ASSERTION]: telemetry item 1 (of 1) not found (in the expected order) for metric name: "codewhisperer_userDecision" 
      + expected - actual

         "codewhispererRequestId": "test"
         "codewhispererSessionId": "test"
         "codewhispererSuggestionIndex": "0"
         "codewhispererSuggestionReferenceCount": "0"
      -  "codewhispererSuggestionState": "Reject"
      +  "codewhispererSuggestionState": "Accept"
         "codewhispererTriggerType": "OnDemand"
         "codewhispererUserGroup": "Control"
         "credentialStartUrl": "testStartUrl"
       }
      
      at partialDeepCompare (d:\a\aws-toolkit-vscode\aws-toolkit-vscode\src\test\testUtil.ts:196:12)
      at assertTelemetry (d:\a\aws-toolkit-vscode\aws-toolkit-vscode\src\test\testUtil.ts:237:9)
[...]

PR #4288, opened 2024-01-16
fix(tests): unreliable codewhisperer tests

Will-ShaoHua:
## Problem

#4238

## License

### DIFF
--- a/src/test/codewhisperer/commands/onAcceptance.test.ts
+++ b/src/test/codewhisperer/commands/onAcceptance.test.ts
@@ -24,6 +24,7 @@ describe('onAcceptance', function () {
     describe('onAcceptance', function () {
         beforeEach(async function () {
             await resetCodeWhispererGlobalVariables()
+            session.reset()
         })
 
         afterEach(function () {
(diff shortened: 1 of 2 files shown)


src/test/codewhisperer/commands/onAcceptance.test.ts
```diff
@@ -24,6 +24,7 @@ describe('onAcceptance', function () {
     describe('onAcceptance', function () {
         beforeEach(async function () {
             await resetCodeWhispererGlobalVariables()
+            session.reset()
```
Will-ShaoHua, line 27: @justinmk3 I suspect the previous fix PR doesn't work as expected because of async/sync things. Test case itself is async, beforeEach is async and afterEach is sync, any guidance for this thanks!

justinmk3: mocha awaits async `beforeEach`. You can verify this by adding `await sleep(10000)` here. It won't run any `it()` until the sleep() has completed.

Will-ShaoHua: understood, so by saying `mocha awaits async beforeEach/afterEach/etc.`, we can say L31 itself should already do its work correctly and clean the states after each test case, and the root cause for test failure might be something else.

Will-ShaoHua: gotcha, let's add one in beforeEach() then

justinmk3:
unfortunately this didn't fix the issue, see https://github.com/aws/aws-toolkit-vscode/issues/4238#issuecomment-1896476214

PR #4290, opened 2024-01-17
feat(gumby): run mvnw if mvn fails

dhasani23:
## Problem

When `mvn` fails, we want to run `mvnw` because that may succeed, since the user could have the Maven wrapper in their workspace despite not having Maven installed.

## Solution

Run `mvnw` when `mvn` fails. Some smaller things done in this PR: 1) slight renaming and 2) collect input in a simpler way (no need to use `MultiStepInputFlowController` since we only collect 1 piece of input).

## License

### DIFF
--- a/src/codewhisperer/service/transformByQHandler.ts
+++ b/src/codewhisperer/service/transformByQHandler.ts
@@ -18,7 +18,7 @@ import { spawnSync } from 'child_process'
 import AdmZip from 'adm-zip'
 import fetch from '../../common/request'
 import globals from '../../shared/extensionGlobals'
-import { telemetry } from '../../shared/telemetry/telemetry'
+import { CodeTransformMavenBuildCommand, telemetry } from '../../shared/telemetry/telemetry'
 import { ToolkitError } from '../../shared/errors'
 import { codeTransformTelemetryState } from '../../amazonqGumby/telemetry/codeTransformTelemetryState'
 import { calculateTotalLatency, javapOutputToTelemetryValue } from '../../amazonqGumby/telemetry/codeTransformTelemetry'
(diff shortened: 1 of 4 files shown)


src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -326,8 +338,7 @@ function getProjectDependencies(modulePath: string): string[] {
     const spawnResult = spawnSync(baseCommand, args, { cwd: modulePath, shell: true, encoding: 'utf-8' })
```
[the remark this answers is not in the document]
justinmk3: If you are passing a hardcoded, relative path, `shell: true` has less risk. But it still means that the system-configured shell is involved, which means unusual systems may cause problems and will be hard to troubleshoot.

justinmk3 (approved):
This is ever the problem with this kind of design: you are re-inventing a build system / IDE. The edge cases will continue to accumulate. If there is a way to re-use vscode's java plugin instead, that could avoid all of this toil.

justinmk3:
waiting for gumby team approval/review

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -310,12 +310,24 @@ function getFilesRecursively(dir: string, isDependenciesFolder: boolean): string
...
 
-function getProjectDependencies(modulePath: string): string[] {
+function getProjectDependencies(buildCommand: CodeTransformMavenBuildCommand, modulePath: string): string[] {
     // Make temp directory
     const folderName = `${CodeWhispererConstants.dependencyFolderName}${Date.now()}`
     const folderPath = path.join(os.tmpdir(), folderName)
 
-    const baseCommand = 'mvn'
+    let baseCommand = buildCommand as string
+    if (baseCommand === 'mvnw') {
```
damntrecky, line 319: Question: It looks like IntelliJ we run `./mvnw` first and `mvn` second. I think we should align on the order. See IntelliJ Implementation in [CodeModernizerSessionContext.kt](https://github.com/aws/aws-toolkit-jetbrains/blob/main/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/model/CodeModernizerSessionContext.kt#L190)

damntrecky (approved):
Tentative approval. Please review the order in which we run these commands to align with IntelliJ.

src/codewhisperer/service/transformByQHandler.ts
dhasani23: Agreed & fixed

damntrecky:
> waiting for gumby team approval/review

Approved 👍 🟢

PR #4292, opened 2024-01-17
MRU list util implementation

Will-ShaoHua:
## Problem
Not seeing relative util class existing in VSC code base and library

## Solution
Implementation is moved from AWS Toolkit JB https://github.com/aws/aws-toolkit-jetbrains/blob/main/jetbrains-core/src/software/aws/toolkits/jetbrains/utils/MRUList.kt

## License


Will-ShaoHua:
@justinmk3 @nkomonen-amazon let me know if we already have MRU library or implementation in our code base, adding this because I will need one for most recently used text document logics and unfortunately VSC doesn't expose relative API for now.

PR #4293, opened 2024-01-17
fix(cloud9): error when expanding S3 folder

justinmk3:
# Problem
Users on Cloud9 (Node.js 16) cannot view the content of an S3 bucket since Toolkit v2.3.0 (https://github.com/aws/aws-toolkit-vscode/pull/4163).

    2024-01-17 13:59:54 [ERROR]: [RangeError: Value longOffset out of range for Intl.DateTimeFormat options property timeZoneName
    at new DateTimeFormat (<anonymous>)
    at Sv (/…/dist/src/main.js:2653:1432)

# Solution:
Nodejs 16 only supports `timeZoneName: "short" | "long"`.

- Use `timeZoneName="short"` for Cloud9.
- Use `timeZoneName="shortOffset"` for VSCode; the "long" format is unnecessarily verbose.
- Add tests.
- Set a fixed timezone in CI using the $TZ env var. (No other way to force a timezone in Nodejs / javascript.)

## License

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -21,6 +21,8 @@ jobs:
         env:
             VSCODE_TEST_VERSION: ${{ matrix.vscode-version }}
             NODE_OPTIONS: '--max-old-space-size=8192'
+            # Set a fixed timezone for tests. Intentionally _not_ UTC.
+            TZ: 'US/Pacific'
         steps:
             - uses: actions/checkout@v4
             - name: Use Node.js ${{ matrix.node-version }}
@@ -64,6 +66,8 @@ jobs:
         env:
             VSCODE_TEST_VERSION: ${{ matrix.vscode-version }}
             NODE_OPTIONS: '--max-old-space-size=8192'
+            # Set a fixed timezone for tests. Intentionally _not_ UTC.
+            TZ: 'US/Pacific'
         steps:
             - uses: actions/checkout@v4
             - name: Use Node.js ${{ matrix.node-version }}
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -21,8 +21,6 @@ jobs:
         env:
             VSCODE_TEST_VERSION: ${{ matrix.vscode-version }}
             NODE_OPTIONS: '--max-old-space-size=8192'
-            # Set a fixed timezone for tests. Intentionally _not_ UTC.
-            TZ: 'US/Pacific'
         steps:
             - uses: actions/checkout@v4
             - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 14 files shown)


nkomonen-amazon:
I think we need to add a unit test. This change should have required a test to be updated somewhere I would have assumed

src/shared/utilities/textUtilities.ts
```diff
@@ -256,7 +256,7 @@ export function formatLocalized(d: Date = new Date()): string {
         hour: 'numeric',
         minute: 'numeric',
         second: 'numeric',
-        timeZoneName: 'longOffset',
+        timeZoneName: isCloud9() ? 'short' : 'shortOffset',
```
nkomonen-amazon, line 259: Is there a reason to not use `short` for all?

justinmk3: from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

```
// short: 12/16/2021, PST
// long: 12/16/2021, Pacific Standard Time
// shortOffset: 12/16/2021, GMT-8
// longOffset: 12/16/2021, GMT-08:00
// shortGeneric: 12/16/2021, PT
// longGeneric: 12/16/2021, Pacific Time
```

I suspect that the offset format is more common in AWS, and perhaps more accessible: the shortname requires users to know all the timezone abbreviations, whereas the offset format only requires knowing "GMT".

nkomonen-amazon (approved):

.github/workflows/node.js.yml
```diff
@@ -21,6 +21,8 @@ jobs:
         env:
             VSCODE_TEST_VERSION: ${{ matrix.vscode-version }}
             NODE_OPTIONS: '--max-old-space-size=8192'
+            # Set a fixed timezone for tests. Intentionally _not_ UTC.
+            TZ: 'US/Pacific'
```
justinmk3, line 25: No other way (AFAICT) to force a timezone in nodejs/javascript.

TODO: Ideally we could do this in e.g. `launchTestUtilities.ts` but I think it's required at startup, can't change timezone of the running process.

nkomonen-amazon: I was playing around with setting `process.env.tz = "US/Pacific"` in `mochaGlobalSetup()`. Does that work for you?

justinmk3: seems to work!

nkomonen-amazon: Looks like it could be due to the global setup function running too late (`script/test/test.ts` runs before it in CI) and there is a race between defining the env var and when node "locks" it in on the first read
`"you can set the TZ environment variable before calling any date functions"`
https://stackoverflow.com/questions/8083410/how-can-i-set-the-default-timezone-in-node-js

I wonder if setting `process.env.TZ` in the global scope in `test/launchUtilities.ts` (or scripts/test/tests.ts) will get it working since they are the files first executed

justinmk3: > I wonder if setting `process.env.TZ` in the global scope in `test/launchUtilities.ts` (or scripts/test/tests.ts) will get it working since they are the files first executed

I will try that in a followup.

PR #4294, opened 2024-01-17
browser: remove fs-extra uses from systemUtilities.ts

nkomonen-amazon:
Reverts to only using `fs` methods instead of `fs-extra` since `fs` will not break if imported in to the browser.

`fs` code still cannot run in the browser, but it will not break if something imports it.

## License

### DIFF
--- a/src/shared/systemUtilities.ts
+++ b/src/shared/systemUtilities.ts
@@ -3,7 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import * as fs from 'fs-extra'
+import { promises as fsPromises } from 'fs'
+import fs from 'fs'
 import * as vscode from 'vscode'
 import * as os from 'os'
 import * as path from 'path'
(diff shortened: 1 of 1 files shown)


justinmk3:
see also https://github.com/aws/aws-toolkit-vscode/pull/4287

justinmk3 (approved):

hayemaxi:
why not use `fsCommon` ?

nkomonen-amazon:
@hayemaxi this is just a quick fix to remove `fs-extra` since there is more work/nuance to using `fsCommon`

`fsCommon` uses `vscode.workspace.fs` under the hood and when in C9 they stub the object returned by `vscode.workspace.fs` with their custom implementation. This implementation has different behaviors than the implementation in VS Code. One example is that `vscode.workspace.fs.createDirectory()` in C9 does not allow you to to create a dir outside of the workspace folder (eg: cannot mkdir something in `~/.aws`) but with the VS Code implementation we can.

PR #4300, opened 2024-01-18
MRU TextDocument cache implementation

Will-ShaoHua:
## License

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -78,6 +78,7 @@ import { initializeNetworkAgent } from './codewhisperer/client/agent'
 import { Timeout } from './shared/utilities/timeoutUtils'
 import { submitFeedback } from './feedback/vue/submitFeedback'
 import { showQuickStartWebview } from './shared/extensionStartup'
+import { MRUDocuments } from './shared/utilities/editorUtilities'
 
 let localize: nls.LocalizeFunc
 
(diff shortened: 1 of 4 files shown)


Will-ShaoHua:
> what problem is this solving
@justinmk3 
we're planing to run an A/B on our CrossFile context.
The status quo (PROD behavior) is we use opened files and sort them by file distance between the file where users trigger CW.
The next step is to sort them by MRU order instead of file distance.

justinmk3:
Did you evaluate existing MRU packages? Would be ideal if we could find a robust and featureful solution, so that we can use this in many other places that require an MRU.

One particularly important feature is to have "visibility" into the memory cost of the MRU so that we can log warnings if it gets too big, or just log total usage of all MRUs in the Toolkit.

Will-ShaoHua:
> Did you evaluate existing MRU packages? Would be ideal if we could find a robust and featureful solution, so that we can use this in many other places that require an MRU.
> 
> One particularly important feature is to have "visibility" into the memory cost of the MRU so that we can log warnings if it gets too big, or just log total usage of all MRUs in the Toolkit.

yea sounds reasonable to me, I will do some search and see if there is better existing library we can use then.

Will-ShaoHua:
FYI, will not need this PR within short term period.

Will-ShaoHua:
replace it with #5576

PR #4304, opened 2024-01-18
tests: explicitly set TZ for local tests

nkomonen-amazon:
In #4297 we explicitly set the TZ in a part of the code that CI uses, but when running the unit tests locally it does not execute that part of the code.

Due to this the tests still fail when run locally.

Solution:

Add the TZ assignment to a part of the code that occurs a bit later, but will execute for both the CI and local tests Now unit tests will pass when run locally.

The initial reason was that some documentation said if the `Date` object was used before assigning the TZ it would lock in the users actual TZ and overriding the env var would not work anymore. But after testing this seems to work fine.

## License

### DIFF
--- a/src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
+++ b/src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
@@ -210,7 +210,11 @@ describe('LogDataRegistry', async function () {
         it('matches CloudWatch insights timestamps', function () {
             const time = 1624201162222 // 2021-06-20 14:59:22.222 GMT+0
             const timestamp = formatDateTimestamp(true, new Date(time))
-            assert.strictEqual(timestamp, '2021-06-20T14:59:22.222-07:00')
+            assert.strictEqual(
+                timestamp,
+                '2021-06-20T14:59:22.222-07:00',
+                '`process.env.TZ` assignment may have been modified'
+            )
         })
     })
 })
--- a/src/test/testRunner.ts
+++ b/src/test/testRunner.ts
@@ -9,6 +9,12 @@ import Mocha from 'mocha'
 import glob from 'glob'
 import * as fs from 'fs-extra'
 
+// Set explicit timezone to ensure that tests run locally do not use the user's actual timezone, otherwise
+// the test can pass on one persons machine but not anothers.
+// Intentionally _not_ UTC, to increase variation in tests.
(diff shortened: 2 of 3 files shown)


src/test/cloudWatchLogs/registry/logDataRegistry.test.ts
```diff
@@ -210,7 +210,11 @@ describe('LogDataRegistry', async function () {
         it('matches CloudWatch insights timestamps', function () {
             const time = 1624201162222 // 2021-06-20 14:59:22.222 GMT+0
             const timestamp = formatDateTimestamp(true, new Date(time))
-            assert.strictEqual(timestamp, '2021-06-20T14:59:22.222-07:00')
+            assert.strictEqual(
+                timestamp,
+                '2021-06-20T14:59:22.222-07:00',
+                '`process.env.TZ` assignment may have been modified'
```
nkomonen-amazon, line 216: Will hopefully point on how to debug if we change `process.env.TZ` at some point

src/test/testRunner.ts
```diff
@@ -13,6 +13,11 @@ import * as fs from 'fs-extra'
  * @param initTests List of relative paths to test files containing root hooks: https://mochajs.org/#available-root-hooks
  */
 export async function runTests(testFolder: string, initTests: string[] = [], testFiles?: string[]): Promise<void> {
+    // Set explicit timezone to ensure that tests run locally do not use the user's actual timezone, otherwise
+    // the test can pass on one persons machine but not anothers. Also, there is a race condition from when we define this
+    // to the first time a Date is used, so this must be defined before Date is used.
+    process.env.TZ = 'US/Pacific'
```
justinmk3, line 19: can we delete the thing in launchUtilities.ts ?

justinmk3 (approved):

justinmk3:
> but when running the unit tests locally it does not execute that part of the code.

Is there a reason we can't use the same entrypoint for vscode vs `npm run` ?

src/test/testRunner.ts
nkomonen-amazon: Did some more testing, yeah we should be able to safely drop it. From reading around there may be a race condition, but I don't see it here

justinmk3: we could also move it to the module-level in `src/test/testRunner.ts` , to further mitigate a race?

justinmk3 (approved):

PR #4305, opened 2024-01-18
feat(appcomposer): code lens to open cfn template in app composer

CaptainCarpensir:
Alongside this PR I have a couple questions for code testing, as such this is completely untested by code at the moment. 

I couldn't find unit tests for the appcomposer code at all, and I'm unfamiliar with integ/e2e tests here and the contributing doc doesn't answer the question of what does/doesn't need e2e tests. When are integ/e2e tests needed for features? Am I missing unit tests for the appcomposer code? There's no folder for `src/test/appcomposer` so I assume they don't exist. If so, why aren't there unit tests?

## License


src/applicationcomposer/activation.ts
```diff
@@ -4,12 +4,41 @@
...
+                ? await CloudFormation.tryLoad(document.uri)
+                : undefined
+
+        if (cfnTemplate?.template === undefined) {
+            return []
+        }
+
+        const lines = document.getText().split('\n')
+        for (let i = 0; i < lines.length; i++) {
+            if (lines[i].trim() === 'Resources:') {
```
JLargent42, line 30: Nit: could we change this to `if (lines[i].startsWith('Resources:') || lines[i].startsWith('Transform:'))`?

src/applicationcomposer/activation.ts
```diff
@@ -4,12 +4,41 @@
...
 import { AuthUtil, getChatAuthState } from '../codewhisperer/util/authUtil'
 import { telemetry } from '../shared/telemetry/telemetry'
 
+export class OpenApplicationComposerCodeLensProvider implements vscode.CodeLensProvider {
+    public async provideCodeLenses(
+        document: vscode.TextDocument,
+        token: vscode.CancellationToken
+    ): Promise<vscode.CodeLens[]> {
+        const cfnTemplate =
+            CloudFormation.isValidFilename(document.uri) && document.languageId === 'yaml'
```
JLargent42, line 20: Is there a case where a .yml or .yaml file doesn't have the 'yaml' language ID? This seems redundant to me.

src/applicationcomposer/activation.ts
JLargent42: This should ideally cover all of the cases described [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-anatomy.html), though most templates start with Transform, Resources, or AWSTemplateFormatVersion.

src/applicationcomposer/activation.ts
CaptainCarpensir: I don't know why there would be, I copied this code from the other place I found it used [here](https://github.com/aws/aws-toolkit-vscode/blob/144dfccb9b1936c290586777f1233c11233e526b/src/shared/awsFiletypes.ts#L70). My best guess is that this is trying to avoid the case where the TextDocument is on a windows machine, where the file extension could be changed but the `languageId` may still be `yaml`. If this is the case it seems overly restrictive though so I agree with removing the requirement for the `languageId` 👍

JLargent42:
I've opened a separate PR for these changes with a couple tweaks for improved compatibility and better organization: https://github.com/aws/aws-toolkit-vscode/pull/4436

PR #4306, opened 2024-01-18
feat(gumby): use only diff.patch to apply patch

dhasani23:
## Problem

The current implementation of the diff viewer relies on the full changed source code being returned by the ExportResultArchive API. This is unnecessary and led to some customer issues when the download size was too large. We should be able to show the diff viewer using only the diff.patch file returned by the API.

## Solution

Get rid of the "parse-diff" library and instead use "diff" (pre-existing dependency) to apply the patch to the user's project.

## License


src/codewhisperer/models/model.ts
```diff
@@ -248,12 +248,6 @@ export enum BuildSystem {
     Unknown = 'Unknown',
 }
 
-export enum DropdownStep {
```
dhasani23, line 251: This was previously used in the input collection but is no longer used anywhere

src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -6,7 +6,7 @@
 import AdmZip from 'adm-zip'
 import os from 'os'
 import fs from 'fs-extra'
-import parseDiff from 'parse-diff'
```
dhasani23, line 9: I will formally remove "parse-diff" from package.json / package-lock.json in a separate PR

src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -158,8 +209,6 @@ export class DiffModel {
                 return new ModifiedChangeNode(originalPath, tmpChangedPath)
             } else if (!originalFileExists && changedFileExists) {
                 return new AddedChangeNode(originalPath, tmpChangedPath)
-            } else if (originalFileExists && !changedFileExists) {
-                return new RemovedChangeNode(originalPath)
```
dhasani23, line 162: With this implementation, there is really no concept of a RemovedChangeNode, since `changedFileExists` will always be true. If a file is for some reason deleted (which really should not happen; it doesn't really make sense to delete a file when transforming a project) it will show up as a ModifiedChangeNode. As mentioned, the LLM does not delete files as of now as a part of the transformation, so this was probably never necessary. It's also a bit risky to keep it since the `saveFile` method of RemovedChangeNode deletes the file, and if there is some unknown bug and one of the user's project files is accidentally deleted, that would be pretty bad.

dhasani23:
No change log here because there are no user-facing changes

src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -137,19 +112,74 @@ export class DiffModel {
...
     /**
-     *
+     * This function creates a copy of the changed files of the user's project so that the diff.patch can be applied to them
+     * @param pathToWorkspace Path to the project that was transformed
+     * @param changedFiles List of files that were changed
+     * @returns Path to the folder containing the copied files
+     */
+    public copyProject(pathToWorkspace: string, changedFiles: ParsedDiff[]) {
+        const pathToTmpSrcDir = path.join(os.tmpdir(), `project-copy-${Date.now()}`)
+        fs.mkdirSync(pathToTmpSrcDir)
```
justinmk3, line 122: We can punt it in this PR, but we are strictly requiring code to avoid use 'fs-extra' (example: https://github.com/aws/aws-toolkit-vscode/pull/4294 ) in order to be web-compatible.

justinmk3 (approved):
Nice simplification and explanations. Thank you!

tincheng (approved):

PR #4307, opened 2024-01-19
build: flag to zip the packaged VSIX

nkomonen-amazon:
This is useful for development purposes when building a VSIX to test with C9.

Just use `npm run package -- --zip` to create a zipped version of the VSIX. It is the same name but ends with `.zip`

## License

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ node_modules
 .vscode-test/
 coverage/
 *.vsix
+*.vsix.zip
 *.bk
 **/.DS_Store
 .idea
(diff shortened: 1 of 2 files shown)


justinmk3:
> This is useful for development purposes when building a VSIX to test with C9.

Why? [cloud9-toolkit-install.sh](https://github.com/aws/aws-toolkit-vscode/blob/master/cloud9-toolkit-install.sh) supports vsix as well as zip, local files and URLs.

nkomonen-amazon:
@justinmk3 when testing code changes I've been using the method of dragging my VSIX in to a github comment and using that url for the script to pull from. This requires it to be zipped.

PR #4308, opened 2024-01-19
feature(QTransform): Detect and add telemetry for 1p dependencies

carolabadeer:
## Problem
This PR adds detection for 1p dependencies (private dependencies not available on maven central) and creates new telemetry to help the Q Transform team get more granulated information on the types of dependencies in applications submitted through the transformation.

## Solution
This change includes: 
- Running a new maven command that generates a report on dependencies that have newer versions available in maven central. This report is parsed for dependencies that are reported to be on the latest version (which can be available on maven but being used in the application at the latest version available, or be a private dependency). The small subset of dependencies parsed from this report are then checked against maven central. 
    - A few notes on latency: 
        - This approach adds about 3-5 seconds of latency for large multi-module applications tested. The `-DonlyProjectDependencies=true` flag reduced this number from ~1 minute and 40 seconds to only a few seconds. 
        - This logic is done after the transformation is complete to ensure that there is no latency impact on the customer's transformation. 
        
 ### Testing
Tests conducted for this change consisted of end-to-end transformations on the IntelliJ IDE, and tested telemetry generation locally

Sample metric from local testing:
```
2024-01-18 17:34:41 [DEBUG]: telemetry: emitted metric "codeTransform_projectContains1pDependencies" -> 
{"Metadata":[{"Key":"parentMetric","Value":"vscode_executeCommand"},
{"Key":"codeTransformSessionId","Value":"XXXXXX"},
{"Key":"codeTransformJobId","Value":"XXXXXX"},
{"Key":"has1pdependency","Value":"true"},{"Key":"result","Value":"Succeeded"},
{"Key":"awsAccount","Value":"not-set"},{"Key":"awsRegion","Value":"us-west-2"}],"MetricName":"codeTransform_projectContains1pDependencies","Value":1,"Unit":"None","Passive":false,"EpochTimestamp":1705628081678}

```

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,6 +36,7 @@
                 "aws-sdk": "^2.1384.0",
                 "aws-ssm-document-language-service": "^1.0.0",
                 "bytes": "^3.1.2",
+                "cheerio": "^1.0.0-rc.12",
                 "cross-fetch": "^4.0.0",
                 "cross-spawn": "^7.0.3",
                 "fast-json-patch": "^3.1.1",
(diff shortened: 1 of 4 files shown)


justinmk3:
is this still needed?

PR #4312, opened 2024-01-19
fix(gumby): remove unneeded dep, improve error handling, fix button enablement bug

dhasani23:
## Problem

Need to remove "parse-diff" dep from package.json & package-lock.json. Need to handle case where diff is empty. Need to make sure the "Show Transformation Summary" button remains enabled even after a user accepts/rejects changes.

## Solution

`npm uninstall parse-diff` and improve error handling and ensure button remains enabled.

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -4382,7 +4382,6 @@
         "markdown-it": "^13.0.2",
         "mime-types": "^2.1.32",
         "node-fetch": "^2.7.0",
-        "parse-diff": "0.11.1",
         "portfinder": "^1.0.32",
         "sanitize-html": "^2.3.3",
         "semver": "^7.5.4",
--- a/package.json
+++ b/package.json
@@ -2415,7 +2415,7 @@
             {
                 "command": "aws.amazonq.transformationHub.summary.reveal",
                 "title": "%AWS.command.q.transform.showChangeSummary%",
-                "enablement": "gumby.reviewState == InReview"
+                "enablement": "gumby.isSummaryAvailable"
             },
             {
                 "command": "aws.amazonq.transformationHub.reviewChanges.reveal",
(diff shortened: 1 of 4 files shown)


src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -361,6 +364,18 @@ export class ProposedTransformationExplorer {
...
+                // This metric is only emitted when placed before showInformationMessage
+                telemetry.codeTransform_vcsDiffViewerVisible.emit({
+                    codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+                    codeTransformJobId: transformByQState.getJobId(),
+                    result: MetadataResult.Pass,
+                })
+
+                // Do not await this so that the summary reveals without user needing to close this notification
+                void vscode.window.showInformationMessage(CodeWhispererConstants.viewProposedChangesMessage)
+                await vscode.commands.executeCommand('aws.amazonq.transformationHub.summary.reveal')
```
dhasani23, line 378: Moved this code inside the `try` block since before it came after the `finally` block and it would always execute, even when `parseDiff` throws an error -- we only want to show the user the notification that their proposed changes are ready if no errors are experienced.

package.json
```diff
@@ -3805,7 +3805,6 @@
             {
                 "command": "aws.amazonq.showTransformationPlanInHub",
                 "title": "Show Transformation Plan",
-                "icon": "$(book)",
```
dhasani23, line 3808: This button appears in a dropdown menu where the icons are not visible, so removing it.

tincheng (approved):

justinmk3 (approved):

src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -140,6 +140,9 @@ export class DiffModel {
      */
     public parseDiff(pathToDiff: string, pathToWorkspace: string): ProposedChangeNode[] {
         const diffContents = fs.readFileSync(pathToDiff, 'utf8')
+        if (!diffContents) {
```
damntrecky, line 143: Should we be logging this in analytics somewhere? Seems like an edge case, but also worth tracking as a "success" metric of transformations.

PR #4317, opened 2024-01-22
browser: Run Auth Activation in browser

nkomonen-amazon:
This PR does the following for the Toolkit in the Browser:

- Sets up a new `Run & Debug` command `Extension (Chrome)`. This will spin up a Chromium window, load VS Code web + our toolkit extension. Then from here we can set breakpoints in our source code and have it break when executed
- Also this gets auth partially working to the point where we can connect to something like Builder ID
- All storage is done in the browser storage when using a VS Code browser extension
- See `browser.md` for detailed documentation of how to run things

## License


src/extensionWeb.ts
```diff
@@ -24,24 +27,20 @@ export async function activate(context: vscode.ExtensionContext) {
...
         await initializeComputeRegion()
         initialize(context)
         initializeManifestPaths(context)
 
-        const awsContext = new DefaultAwsContext()
-        globals.awsContext = awsContext
+        await activateTelemetry(context, globals.awsContext, Settings.instance)
 
-        globals.sdkClientBuilder = new DefaultAWSClientBuilder(awsContext)
+        await initializeCredentials(context, globals.awsContext, globals.loginManager)
```
justinmk3, line 42: It will help to move as much logic into `extensionShares.ts:activate()` as possible, soon, because if we wait much longer, then it will be less clear what steps are different accidentally vs intentionally.

justinmk3 (approved):

src/extensionWeb.ts
nkomonen-amazon: Good call. I will implement this in my next PR

PR #4322, opened 2024-01-22
fix(tests): unreliable codewhisperer tests

leigaol:
## Problem
Followup of https://github.com/aws/aws-toolkit-vscode/pull/4288.

The reported 'Reject' state comes from `RecommendationHandler.instance.reportUserDecisions(-1)`, which indicates the status of RecommendationHandler is not cleared. 

## Solution

Clear states before these 2 tests. I triggered them 3 times in Github CI and it works fine. 

## License


Will-ShaoHua (approved):

Will-ShaoHua:
@leigaol could you add that into `resetCodeWhispererGlobalVariables`

hayemaxi (approved):

PR #4324, opened 2024-01-23
codewhisperer: Allow endpoint/region overrides for dev mode

jpinkney-aws:
## Problem
- It's not easy to change the codewhisperer endpoint when running tests against different versions of the api

## Solution
- Allow configuration via a setting for dev mode

## License

### DIFF
--- a/src/codewhisperer/models/constants.ts
+++ b/src/codewhisperer/models/constants.ts
@@ -4,13 +4,6 @@
  */
 
 /**
- * SDK Client
- */
-export const endpoint = 'https://codewhisperer.us-east-1.amazonaws.com/'
-
-export const region = 'us-east-1'
-
-/**
  * Automated and manual trigger
  */
 export const invocationTimeIntervalThreshold = 2 // seconds
--- a/src/shared/settings.ts
+++ b/src/shared/settings.ts
@@ -6,6 +6,7 @@
 import * as vscode from 'vscode'
 import * as packageJson from '../../package.json'
 import * as codecatalyst from './clients/codecatalystClient'
+import * as codewhisperer from '../codewhisperer/client/codewhisperer'
 import { getLogger } from './logger'
 import { cast, FromDescriptor, Record, TypeConstructor, TypeDescriptor } from './utilities/typeConstructors'
 import { assertHasProps, ClassToInterfaceType, keys } from './utilities/tsUtils'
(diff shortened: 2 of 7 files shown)


Will-ShaoHua (approved):

src/codewhisperer/models/constants.ts
```diff
@@ -6,9 +6,9 @@
 /**
  * SDK Client
  */
-export const endpoint = 'https://codewhisperer.us-east-1.amazonaws.com/'
+export const endpoint = process.env.CODEWHISPERER_ENDPOINT ?? 'https://codewhisperer.us-east-1.amazonaws.com/'
 
-export const region = 'us-east-1'
+export const region = process.env.CODEWHISPERER_REGION ?? 'us-east-1'
```
justinmk3, line 11: Should be documented in https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#service-endpoints

off-topic: we really should have all env vars in one place. Otherwise it's not clear at all where random env vars can change the behavior anywhere in the codebase.

hayemaxi:
Can we make this a dev setting instead like we have for codecatalyst instead of env var?

https://github.com/aws/aws-toolkit-vscode/blob/6bebba6639862a74234e509903269e8849ae9106/src/shared/settings.ts#L664

used here: https://github.com/aws/aws-toolkit-vscode/blob/6bebba6639862a74234e509903269e8849ae9106/src/shared/settings.ts#L724

And in my VSCode settings JSON, I have:
```
    "aws.dev.codecatalystService": {
        "region": "us-west-2",
        "endpoint": "https://codecatalyst-integ.aws",
        "hostname": "integ.codecatalyst",
        "gitHostname": "git.integ.codecatalyst",
    }
```

The settings in VSCode can be commented out, making switching between prod and test endpoints really easy.

src/codewhisperer/models/constants.ts
andrewyuq: This does mean customers can change them too, is this expected?

Will-ShaoHua: I think the better way is to follow what @hayemaxi suggests to make it a dev setting. 

> This does mean customers can change them too, is this expected?

IMO CX can make these changes anyway by editing the src code and build plugins on their own, thus I don't think it's a concern to make this change?

andrewyuq: that's fair.

hayemaxi (approved):
awesome! thanks for updating the util functions to make them generic

src/shared/settings.ts
```diff
@@ -721,14 +729,14 @@ export class DevSettings extends Settings.define('aws.dev', devSettings) {
         return Object.keys(this.activeSettings).length > 0
     }
 
-    public getCodeCatalystConfig(
-        defaultConfig: codecatalyst.CodeCatalystConfig
-    ): Readonly<codecatalyst.CodeCatalystConfig> {
-        const devSetting = 'codecatalystService'
-        const devConfig = this.get(devSetting, {})
+    public getServiceConfig<T extends ServiceClients>(
```
justinmk3, line 732: Nice, thank you for generalizing this!

PR #4325, opened 2024-01-23
fix(gumby): enhance telemetry to include error info and upload error logs

dhasani23:
## Problem

We run Maven shell commands to `install` and  `copy-dependencies `and a `javap` shell command to determine a project's java version, and although we know the frequency of failures for these commands, we have insufficient insight into why they failed.

## Solution

Add the error message to existing metrics when these commands fail, and add a text file containing the error logs to our upload ZIP so that we can process it in the backend.

## License


src/shared/utilities/download.ts
```diff
@@ -18,7 +18,6 @@ import { fsCommon } from '../../srcShared/fs'
 export class ExportResultArchiveStructure {
     static readonly PathToSummary = path.join('summary', 'summary.md')
     static readonly PathToDiffPatch = path.join('patch', 'diff.patch')
-    static readonly PathToSourceDir = 'sources'
```
dhasani23, line 21: No longer using this anywhere. Also, the `download.ts` above was deleted since that whole file was a duplicate and not used anywhere.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -351,7 +351,7 @@ function getProjectDependencies(buildCommand: CodeTransformMavenBuildCommand, mo
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
             codeTransformMavenBuildCommand: buildCommand,
             result: MetadataResult.Fail,
-            reason: spawnResult.error ? spawnResult.error.message : spawnResult.stdout,
+            reason: 'CannotRunMavenShellCommand',
```
dhasani23, line 354: Changing since `error` and `stdout` may contain PII and we want to log these via CW Logs anyway.

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -87,6 +87,11 @@ async function collectInput(validProjects: vscode.QuickPickItem[]) {
 }
 
 export async function startTransformByQ() {
+    telemetry.codeTransform_jobIsStartedFromUserPopupClick.emit({
+        codeTransformSessionId: `${codeTransformTelemetryState.getSessionId()}-updated`,
+        result: MetadataResult.Pass,
+    })
+
```
dhasani23, line 94: Moved this metric emission to here so that it is logged when a user clicks transform (before any validation occurs, compared to the existing implementation which emits this after all validation occurs). This is now consistent with how the same metric is emitted from the chat entry point. We have a separate metric (`codeTransform_jobStartedCompleteFromPopupDialog`) to know how many transformation attempts pass validation. 

I added the "-updated" so that we have some way to sense that users are using the latest toolkit version, since technically the auto-update can be disabled (https://code.visualstudio.com/docs/supporting/FAQ#_opt-out-of-extension-updates). This is being done because we have a backend change that we need to make, but only want to make it once we know that users are using the latest toolkit version, which contains some breaking changes to the diff viewing experience. Once we are reasonably confident that most or all users are using the latest toolkit version, this "-updated" change will be reverted.

tincheng: I think IntelliJ still emits it (alongside **jobIsCanceledFromUserPopupClick**) at the end of selecting a project: https://github.com/aws/aws-toolkit-jetbrains/blob/bce2facb8170f215f5250b00ff71469b36495239/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/ui/components/PreCodeTransformUserDialog.kt#L127-L139

To summarize, IntelliJ sends **isDoubleClickedToTriggerUserModal** after transform clicked but before project validation starts, **jobIsStartedFromUserPopupClick** after validation passed and user select either OK or cancel on the pop-up, then **jobStartedCompleteFromPopupDialog** after project has been selected. Seems like jobStartedCompleteFromPopupDialog is redundant in IntelliJ.

I think we should align on the usage of these 4 metrics before making the above change.

damntrecky: I agree with Ting, we need to align on this behavior offline and come to a resolution there.

src/amazonqGumby/telemetry/codeTransformTelemetry.ts
```diff
@@ -30,7 +30,7 @@ export enum StartActionPositions {
 
 export const logCodeTransformInitiatedMetric = (source: string): void => {
     const commonMetrics = {
-        codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+        codeTransformSessionId: `${codeTransformTelemetryState.getSessionId()}-updated`,
```
dhasani23, line 33: Appended `-updated` here and above because we need to know that at least most users are using the updated toolkit version (auto-update can be disabled) so that we know that we are safe to make a backend change to our diff viewing experience. If the backend change is made now, users not using the latest toolkit version will experience an error when they go to download the diff. Presence of this `-updated` in metrics will allow us to track how many users have updated the toolkit.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -336,22 +338,20 @@ function getProjectDependencies(buildCommand: CodeTransformMavenBuildCommand, mo
         '-Dmdep.useRepositoryLayout=true',
         '-Dmdep.copyPom=true',
         '-Dmdep.addParentPoms=true',
+        '-X',
+        '-e',
```
dhasani23, line 342: Added these two flags since they will include more error logging in the command output

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -647,3 +647,11 @@ export async function checkBuildSystem(projectPath: string) {
...
     return BuildSystem.Unknown
 }
+
+/*
+ * This function redacts file paths (Windows and Unix) which may contain PII.
+ * It will eventually be extended to handle other types of PII, if needed.
+ */
+function redactPii(errorLog: string) {
+    return errorLog.replace(/[^\s]+[\\/][^\s]*/g, ' <redacted> ')
+}
```
dhasani23, line 657: I know that regex can be inefficient but I'm not aware of any other reasonable way to redact the file paths, and we received a pretty urgent request to get this PR's work out quickly. The only potential place for PII that we have identified is in the file paths. We plan on eventually doing a more thorough analysis with AppSec to ensure that we are not missing any PII, potentially using a more robust PII redacting tool, and sending these error messages to CloudWatch Logs, rather than including them in telemetry.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -647,3 +647,11 @@ export async function checkBuildSystem(projectPath: string) {
     }
     return BuildSystem.Unknown
 }
+
+/*
+ * This function redacts file paths (Windows and Unix) which may contain PII.
+ * It will eventually be extended to handle other types of PII, if needed.
+ */
+function redactPii(errorLog: string) {
+    return errorLog.replace(/[^\s]+[\\/][^\s]*/g, ' <redacted> ')
```
tincheng, line 656: I think there may be an extra "\" in the [\\/] block? Also doesn't work for Windows' path with "\". I tried something like `[^\s]+[\\\/][^\s]*` which seemed to work for both forward and backward slashes.

dhasani23: Sorry don't totally understand. I don't see any "" in the regex. We need the `/g` (global) since that replaces all matching occurrences. The extra `\` in your regex is not strictly needed because you don't need to escape the `/`, just the `\`, which is why I used `\\/`. This is also working fine for me in Windows paths:

Screenshot 2024-01-25 at 3 05 58 PM

tincheng: Gotcha, probably a format issue when I tried to test the above regax using https://regex101.com/. It does make sense the original should have covered both.

tincheng (approved):

src/codewhisperer/service/transformByQHandler.ts
dhasani23: This regex will remove URLs from the logged error message, which could be informative to us, but that is a compromise I think we have to settle for at the moment.

nkomonen-amazon (approved):

dhasani23:
Do not merge

src/codewhisperer/service/transformByQHandler.ts
dhasani23: UPDATE: these comments are now N/A

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -328,11 +329,20 @@ export async function postTransformationJob(userInputState: UserInputState) {
     const durationInMs = calculateTotalLatency(codeTransformTelemetryState.getStartTime())
     const resultStatusMessage = codeTransformTelemetryState.getResultStatus()
 
+    const versionInfoWrapper = await getVersionData('mvnw', transformByQState.getProjectPath())
+    let mavenVersionInfoMessage = versionInfoWrapper[0]
+    let javaVersionInfoMessage = versionInfoWrapper[1]
+    const versionInfo = await getVersionData('mvn', transformByQState.getProjectPath())
+    mavenVersionInfoMessage += ` -- ${versionInfo[0]}`
+    javaVersionInfoMessage += ` -- ${versionInfo[1]}`
```
tincheng, line 337: Should we just have one java version?

dhasani23: Discussed offline, will keep both for now

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -114,18 +114,36 @@ export async function validateProjectSelection(project: vscode.QuickPickItem) {
...
+            if (Buffer.byteLength(spawnResult.stdout, 'utf-8') > CodeWhispererConstants.maxBufferSize) {
+                errorReason += '-BufferOverflow'
+            }
+        } else {
+            errorReason = 'JavapSpawnError'
+        }
+        if (spawnResult.error) {
+            // oddly, the 'code' field is not visible until I stringify the object, but then I have to parse it to access 'code'
+            // 'code' is a high-level symbol representing the error (ex. 'ENOENT', 'ENOBUFS', etc.)
+            const errorCode = JSON.parse(JSON.stringify(spawnResult.error)).code ?? ''
```
dhasani23, line 136: Will change to `?? 'unknown'`

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -312,7 +331,65 @@ function getFilesRecursively(dir: string, isDependenciesFolder: boolean): string
...
+             * which are not being used here (for now), but still keeping this just in case.
+             */
+            if (Buffer.byteLength(spawnResult.stdout, 'utf-8') > CodeWhispererConstants.maxBufferSize) {
+                errorReason += '-BufferOverflow'
+            }
+        } else {
+            errorReason = 'InstallSpawnError'
+        }
+        if (spawnResult.error) {
+            const errorCode = JSON.parse(JSON.stringify(spawnResult.error)).code ?? ''
```
dhasani23, line 377: Will change to `?? 'unknown'`

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -323,13 +400,15 @@ function getProjectDependencies(buildCommand: CodeTransformMavenBuildCommand, mo
         if (os.platform() === 'win32') {
             baseCommand = './mvnw.cmd'
         }
-        const executableName = baseCommand.substring(2) // remove the './' part
+        const executableName = baseCommand.slice(2) // remove the './' part
```
dhasani23, line 403: these are the same, just changing because I use `slice` everywhere else

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -338,22 +417,36 @@ function getProjectDependencies(buildCommand: CodeTransformMavenBuildCommand, mo
...
-
-    if (spawnResult.error || spawnResult.status !== 0) {
-        getLogger().error(`CodeTransform: Error in running Maven command ${baseCommand} = `)
-        // Maven command can still go through and still return an error. Won't be caught in spawnResult.error in this case
-        if (spawnResult.error) {
-            getLogger().error(spawnResult.error)
+    if (spawnResult.status !== 0) {
+        let errorLog = ''
+        errorLog += spawnResult.error ? `${JSON.stringify(spawnResult.error)}` : ''
+        errorLog += `${spawnResult.stderr}\n${spawnResult.stdout}` ?? ''
```
dhasani23, line 423: Will fix the `?? ''` too

src/amazonqGumby/entrypoint.ts
```diff
@@ -21,7 +21,7 @@ export async function processTransformByQ() {
     if (transformByQState.isNotStarted()) {
         await sleep(1000) // sleep so that chat can respond first, then show input prompt
         telemetry.codeTransform_jobIsStartedFromChatPrompt.emit({
-            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+            codeTransformSessionId: `${codeTransformTelemetryState.getSessionId()}-updated`,
```
damntrecky, line 24: Can we add the `reason` field to have `updated` in it or something.

src/amazonqGumby/telemetry/codeTransformTelemetry.ts
damntrecky: Same here.

src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -328,11 +329,20 @@ export async function postTransformationJob(userInputState: UserInputState) {
     const durationInMs = calculateTotalLatency(codeTransformTelemetryState.getStartTime())
     const resultStatusMessage = codeTransformTelemetryState.getResultStatus()
```
damntrecky, line 331: Make sure IntelliJ aligns on this later on.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -114,18 +114,36 @@ export async function validateProjectSelection(project: vscode.QuickPickItem) {
...
+        if (spawnResult.error) {
+            // oddly, the 'code' field is not visible until I stringify the object, but then I have to parse it to access 'code'
+            // 'code' is a high-level symbol representing the error (ex. 'ENOENT', 'ENOBUFS', etc.)
+            const errorCode = JSON.parse(JSON.stringify(spawnResult.error)).code ?? 'UNKNOWN'
+            errorReason += `-${errorCode}`
+        }
         telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit({
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
             codeTransformPreValidationError: 'NoJavaProject',
             result: MetadataResult.Fail,
```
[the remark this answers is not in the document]
damntrecky: Possibly: `message: errorReason` or something more simple than that in message. `if (errorCode) message: "SpawnResultError" : message: "OtherErrorName"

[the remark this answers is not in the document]
damntrecky: Use `codeTransformRuntimeError` here.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -338,22 +417,36 @@ function getProjectDependencies(buildCommand: CodeTransformMavenBuildCommand, mo
...
+        if (spawnResult.error) {
+            const errorCode = JSON.parse(JSON.stringify(spawnResult.error)).code ?? 'UNKNOWN'
+            errorReason += `-${errorCode}`
         }
         telemetry.codeTransform_mvnBuildFailed.emit({
             codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
             codeTransformMavenBuildCommand: buildCommand,
             result: MetadataResult.Fail,
-            reason: spawnResult.error ? spawnResult.error.message : spawnResult.stdout,
+            reason: errorReason,
```
damntrecky, line 445: Just add the `Maven Copy:` prefix here as well.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -647,3 +768,42 @@ export async function checkBuildSystem(projectPath: string) {
     }
     return BuildSystem.Unknown
 }
+
+export async function getVersionData(buildCommand: CodeTransformMavenBuildCommand, modulePath: string) {
```
damntrecky, line 772: This is strictly for telemetry.

src/amazonqGumby/entrypoint.ts
dhasani23: Done. For others: we're doing this to be able to gauge when our customers have all updated their toolkit version. This change will be reverted once we see everyone (or most) are using the most recent version.

src/amazonqGumby/telemetry/codeTransformTelemetry.ts
```diff
@@ -30,20 +30,22 @@ export enum StartActionPositions {
...
             ...commonMetrics,
             result: MetadataResult.Pass,
+            reason: 'Updated',
         })
     } else if (source === StartActionPositions.BottomHubPanel) {
         telemetry.codeTransform_isDoubleClickedToTriggerUserModal.emit({
             codeTransformStartSrcComponents: StartActionPositions.BottomHubPanel,
             ...commonMetrics,
             result: MetadataResult.Pass,
+            reason: 'Updated',
```
justinmk3, line 48: Not sure if there's any precedent for setting `reason` field for `result=success` (non-failure) metrics. Though I suppose it's harmless, but just mentioning for reference. 

As a general guideline I would recommend not setting `reason` for non-failed metrics. But if it's useful to you, then that's also fine.

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -114,18 +114,36 @@ export async function validateProjectSelection(project: vscode.QuickPickItem) {
     const baseCommand = 'javap'
     const args = ['-v', classFilePath]
     const spawnResult = spawnSync(baseCommand, args, { shell: false, encoding: 'utf-8' })
```
justinmk3, line 116: For capturing output and other non-trivial handling, did you consider using `ChildProcess` https://github.com/aws/aws-toolkit-vscode/blob/961ff52e05ed45d25c749a2237de6b2b3237f076/src/shared/utilities/childProcess.ts#L68

src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -114,18 +114,36 @@ export async function validateProjectSelection(project: vscode.QuickPickItem) {
...
+            if (Buffer.byteLength(spawnResult.stdout, 'utf-8') > CodeWhispererConstants.maxBufferSize) {
+                errorReason += '-BufferOverflow'
+            }
+        } else {
+            errorReason = 'JavapSpawnError'
+        }
+        if (spawnResult.error) {
+            // oddly, the 'code' field is not visible until I stringify the object, but then I have to parse it to access 'code'
+            // 'code' is a high-level symbol representing the error (ex. 'ENOENT', 'ENOBUFS', etc.)
+            const errorCode = JSON.parse(JSON.stringify(spawnResult.error)).code ?? 'UNKNOWN'
```
justinmk3, line 136: Did you try `as any` ? `parse(stringify(...))` is a pretty heavy workaround.

```suggestion
            const errorCode = (spawnResult.error as any).code ?? 'UNKNOWN'
```

Though whatever problem you ran into should go away if you switch to using ChildProcess https://github.com/aws/aws-toolkit-vscode/blob/961ff52e05ed45d25c749a2237de6b2b3237f076/src/shared/utilities/childProcess.ts#L68

src/amazonqGumby/telemetry/codeTransformTelemetry.ts
dhasani23: It will be useful to us for probably a couple of weeks, then I plan on reverting this change. Will add a comment that notes this.

src/codewhisperer/service/transformByQHandler.ts
dhasani23: I have not previously considered using `ChildProcess`, added a `TO-DO` comment for this. There will very likely be some changes to the 4 `spawnSync` calls that we have in the next 1-2 weeks (may remove a couple of them, or at least modify them), so I'd like to wait until we finalize that and reserve this consideration for another PR

src/codewhisperer/service/transformByQHandler.ts
dhasani23: Thanks. `as any` worked.

src/codewhisperer/service/transformByQHandler.ts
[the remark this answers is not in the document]
dhasani23: https://github.com/aws/aws-toolkit-common/pull/680

Fixed

src/amazonqGumby/telemetry/codeTransformTelemetry.ts
```diff
@@ -30,20 +30,22 @@ export enum StartActionPositions {
 
 export const logCodeTransformInitiatedMetric = (source: string): void => {
     const commonMetrics = {
-        codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+        codeTransformSessionId: `${codeTransformTelemetryState.getSessionId()}`,
```
damntrecky, line 33: Dump the string literal and the ${. Just log `codeTransformTelemetryState.getSessionId()`

src/amazonqGumby/telemetry/codeTransformTelemetry.ts
damntrecky: +1 to your comment above @dhasani23. Add a code comment explaining how we are using this so the context isn't lost later on.

src/codewhisperer/service/transformByQHandler.ts
damntrecky: +1 To look into this refactor, but out of scope here. @dhasani23 Make sure to capture this to review internally.

src/amazonqGumby/entrypoint.ts
```diff
@@ -21,8 +21,9 @@ export async function processTransformByQ() {
     if (transformByQState.isNotStarted()) {
         await sleep(1000) // sleep so that chat can respond first, then show input prompt
         telemetry.codeTransform_jobIsStartedFromChatPrompt.emit({
-            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+            codeTransformSessionId: `${codeTransformTelemetryState.getSessionId()}`,
```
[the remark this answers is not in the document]
damntrecky:
```suggestion
            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
```

damntrecky (approved):
Approved. Please remove the string literal.

src/amazonqGumby/telemetry/codeTransformTelemetry.ts
justinmk3: > Presence of this `-updated` in metrics will allow us to track how many users have updated the toolkit.

? Toolkit version is attached to each metric.

justinmk3 (approved):

PR #4329, opened 2024-01-25
codewhisperer: update reference setting description

zixlin7:
## Problem
Professional users are not clear that the allow reference setting is controlled by admin
## Solution
Update the description per doc writer

## License


package.nls.json
```diff
@@ -70,7 +70,7 @@
     "AWS.configuration.description.awssam.debug.snippets.api.description": "A new configuration for invoking an AWS Lambda in a CloudFormation template, simulating API Gateway",
     "AWS.configuration.description.awssam.debug.snippets.api.label.cn": "Amazon SAM: API Gateway lambda invoke",
     "AWS.configuration.description.awssam.debug.snippets.api.description.cn": "A new configuration for invoking an Amazon Lambda in a CloudFormation template, simulating API Gateway",
-    "AWS.configuration.description.codewhisperer": "When checked, CodeWhisperer will include suggestions with code references",
+    "AWS.configuration.description.codewhisperer": "When checked, CodeWhisperer will include suggestions with code references. If you authenticate through IAM Identity Center, this setting is controlled by your CodeWhisperer Professional administrator.",
```
justinmk3, line 73: can we link to the docs? markdown hyperlinks are supported here.

zixlin7: This is essentially what the documentation says so I don't see a strong need for linking it here? Seems more straight-forward this way.

leigaol (approved):

package.nls.json
Will-ShaoHua: ahh +1 to link to the doc. IMO even though the string is already reflecting the content of our doc, part of CX will still want to know the details if they're really curious the process. By linking our doc, they don't need to reach out to us to resolve their questions.

Will-ShaoHua (approved):

package.nls.json
zixlin7: ok I will add the doc link

andrewyuq (approved):

package.nls.json
zixlin7: updated

justinmk3 (approved):

PR #4336, opened 2024-01-26
refactor (codewhisperer): unify isCWRunning state into one place instead of two (1)

Will-ShaoHua:
## Problem
There're "2" states maintained in CW which serve the same purpose to represent CW is running/thinking to generate recommendations. They're respectively:
`InlineCompletionService.isPaginatedRunning` https://github.com/aws/aws-toolkit-vscode/blob/master/src/codewhisperer/service/inlineCompletionService.ts#L30

`RecommendationHandler.isGenerateRecommendationInProgress` https://github.com/aws/aws-toolkit-vscode/blob/master/src/codewhisperer/service/recommendationHandler.ts#L75

We can merge these 2 variables into one at `RecommendationService` instead of 2 at their lower level component

## Solution
Move `isGenerateRecommendationInProgress` up to `RecommendationService`. There will be a "followup" PR to move `isPaginatedRunning` to RecommendationService

## License


src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -86,11 +85,7 @@ export class InlineCompletionService {
         autoTriggerType?: CodewhispererAutomatedTriggerType,
         event?: vscode.TextDocumentChangeEvent
     ) {
-        if (
-            vsCodeState.isCodeWhispererEditing ||
-            this._isPaginationRunning ||
```
Will-ShaoHua, line 91: move one level up to `RecommendationService`

justinmk3 (approved):
Having one source of truth definitely helps avoid bugs. Thanks!

PR #4341, opened 2024-01-29
browser: share activate() setup between different entrypoints

nkomonen-amazon:
## Problem
We currently have `extension.ts#activate()` and `extensionWeb.ts#activate()` where the web version is basically a subset of the main `activate()`. This creates code duplication.

## Solution
In `extensionShared.ts` have a shared activate function which both entrypoints will call.
This function will consist of the same setup code regardless of entrypoint, this will de-duplicate the code.

Note in activation order matters, so we may have to do some hacks if we run in to any new issues as
we migrate shared code from `extension.ts` in to `extensionShared.ts`.

## License

### DIFF
--- a/src/extensionShared.ts
+++ b/src/extensionShared.ts
@@ -49,9 +49,12 @@ export async function testActivate(context: vscode.ExtensionContext) {
     globals.loginManager = new LoginManager(globals.awsContext, new CredentialsStore())
     setupGlobalsTempStubs()
 
+    // some "initialize" functions
     await initializeComputeRegion()
     initialize(context)
     initializeManifestPaths(context)
+
+    await activateTelemetry(context, globals.awsContext, Settings.instance)
 }
 
 /**
@@ -62,8 +65,6 @@ export async function testActivate(context: vscode.ExtensionContext) {
 export async function browserActivate(context: vscode.ExtensionContext) {
     try {
         await testActivate(context)
-
-        await activateTelemetry(context, globals.awsContext, Settings.instance)
 
         await initializeCredentials(context, globals.awsContext, globals.loginManager)
         await initializeAwsCredentialsStatusBarItem(globals.awsContext, context)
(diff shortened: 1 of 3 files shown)


hayemaxi (approved):
I wonder if this will result in a lot of friction and hacks as the browser version becomes supported? If so, I'm not sure if its worth the effort to make a shared function now. We can always merge the functions after browser is fully working, or compartmentalize the different activate actions in each to better share them.

justinmk3:
> We can always merge the functions after browser is fully working

The analysis and inspection that we are already doing incrementally for browser support, would then potentially need to be repeated for the shared case, which would be an expensive redundancy. 

If there is risk in starting out with the shared approach then we are doing something wrong.

src/extensionShared.ts
```diff
@@ -9,19 +9,67 @@
...
+import { RegionProvider } from './shared/regions/regionProvider'
+
+/**
+ * This runs activation/setup code that is shared by both the regular extension vs. browser
+ * activation in their respective entrypoints.
+ *
+ * The different entrypoints may do additional setup, but the setup in this function
+ * is shared by both and works for both. So any code that works in both should be
+ * placed here.
+ *
```
justinmk3, line 39: This is very verbose.
```suggestion
 * Activation/setup code that is shared by the regular (nodejs) extension and
 * the browser-compatible extension. Most setup code should live here.
 *
```

src/extensionWeb.ts
```diff
@@ -28,63 +17,21 @@ export async function activate(context: vscode.ExtensionContext) {
...
             stacktrace.length = 40
         }
-        getLogger().error('Failed to activate extension in Browser', error)
-        throw error
+        getLogger().error(`Failed to activate extension`, error)
     }
 }
 
-/** Set up values for the `globals` object */
-function setupGlobals() {
```
justinmk3, line 59: Looks like this change eliminated some unessential variation between the "regular" vs browser case. And that's exactly the point of doing this now: unwinding those non-essential varations _later_ would be difficult.

justinmk3 (approved):

PR #4342, opened 2024-01-29
refactor: move toolkit to monorepo structure

hayemaxi:
These commits restructure our project into a monorepo. Using npm workspaces. The original toolkit extension code now lives under `packages/toolkit/`. Additional extensions and common modules will go under the `packages/` folder. From this point on, you will need to be more specific with how you run npm scripts. I have added passthroughs for many of the npm scripts in the root folder, e.g. `npm run compile` and `npm run test`. But other lesser used commands may need to be accessed by first going to the subproject directory. Additionally, to run VSCode run and debug configs, you will need to set the subproject folder as your root folder in VSCode (or add it as root folder).

Consult https://github.com/aws/aws-toolkit-vscode/blob/21c66e259c88f8c0ca54f8aaff010c301668c683/docs/ARCHITECTURE.md#monorepo-structure for more info on this change.

Remaining work items:
- ~~Get VSCode debugging configurations to work~~
- ~~Fix Github Action-based CI entry points.~~
- ~~Ensure coverage works properly~~
- ~~Merge conflicts :(~~
- ~~(Required after merge?) Fix packaging CI~~
- Documentation
  - Update references
  - Explain new structure/current limitations/future plans
 

## License


hayemaxi:
Looks like VSCode Package Tests will continue to fail with this change until we update the codebuilde spec.

packages/toolkit/.vscodeignore
```diff
@@ -2,6 +2,10 @@
 *
 */**
 
+# Ignore from the root package.json otherwise those dirs will be collected as well.
+../../*
+../../*/**
```
justinmk3, line 7: would this be fixed by invoking `vsce` from this directory as the CWD? I wonder if that would avoid surprises in general.

justinmk3: > Its not entirely clear to me why vsce includes parent directories, but researching has led me to answers that [vsce does not officially support workspaces](https://github.com/microsoft/vscode-vsce/issues/777).

Nice. Would be useful to include that comment + URL in the code comment.

packages/toolkit/src/test/globalSetup.test.ts
```diff
@@ -26,7 +26,7 @@ import { getTestWindow, resetTestWindow } from './shared/vscode/window'
 import { mapTestErrors, normalizeError, setRunnableTimeout } from './setupUtil'
 import { TelemetryDebounceInfo } from '../shared/vscode/commands2'
 
-const testReportDir = join(__dirname, '../../../.test-reports')
+const testReportDir = join(__dirname, '../../../../../.test-reports') // Root project, not subproject
```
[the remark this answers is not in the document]
hayemaxi: yes there should be a shared test setup function/library in a commons folder. Will document.

justinmk3 (approved):

PR #4352, opened 2024-01-31
testing: add testing directory configuration through environment variable

jpinkney-aws:
## Problem
- Theres no external way to set a specific directory to use for tests

## Solution
- Add the TEST_DIR that will be used when running tests

Example:
```
export TEST_DIR=src/testE2E/amazonq
npm run testE2E
```

## License

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -189,6 +189,21 @@ To run a single test in VSCode, do any one of:
     rootTestsPath: __dirname + '/shared/sam/debugger/'
     ```
 
+#### Run all tests in a specific folder
+
+To run tests against a specific folder in VSCode, do any one of:
+
+-   Add the TEST_DIR environment variable to one of the testing launch configs and run it
+-   Run in your terminal
+    -   Unix/macOS/POSIX shell:
+        ```
+        TEST_DIR=src/test/foo npm run test
+        ```
+    -   Powershell:
+        ```
+        $Env:TEST_DIR = "src/test/foo"; npm run test
+        ```
+
 ### Browser Support
 
 Running the extension in the browser (eg: [vscode.dev](https://vscode.dev/)).
@@ -403,6 +418,7 @@ Environment variables can be used to modify the behaviour of VSCode. The followi
 -   `AWS_TOOLKIT_AUTOMATION`: If tests are currently being ran
 -   `DEVELOPMENT_PATH`: The path to the aws toolkit vscode project
 -   `AWS_TOOLKIT_TEST_NO_COLOR`: If the tests should include colour in their output
+-   `TEST_DIR` - The directory where the test runner should find the tests
 
 ### SAM/CFN ("goformation") JSON schema
 
--- a/scripts/test/launchTestUtilities.ts
+++ b/scripts/test/launchTestUtilities.ts
@@ -91,6 +91,7 @@ async function getVSCodeCliArgs(params: {
         extensionTestsEnv: {
             ['DEVELOPMENT_PATH']: projectRootDir,
             ['AWS_TOOLKIT_AUTOMATION']: params.suite,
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -403,6 +403,7 @@ Environment variables can be used to modify the behaviour of VSCode. The followi
 -   `AWS_TOOLKIT_AUTOMATION`: If tests are currently being ran
 -   `DEVELOPMENT_PATH`: The path to the aws toolkit vscode project
 -   `AWS_TOOLKIT_TEST_NO_COLOR`: If the tests should include colour in their output
+-   `E2E_TESTING_DIRECTORY` - Where the e2e tests should find the tests to run. If not specified use `src/testE2E`
```
nkomonen-amazon, line 406: We will be doing some refactoring of the structure so this may become outdated.

```suggestion
-   `E2E_TESTING_DIRECTORY` - Where the e2e tests should find the tests to run. If not specified it uses the default.
```

justinmk3, line 406: why is it specific to "e2e"? setting the test directory seems like a general concept. it would be clumsy to also add  `INTEG_TEST_DIRECTORY` later. 

also please use existing names as precedent. we already have  `TEST_FILE` https://github.com/aws/aws-toolkit-vscode/blob/961ff52e05ed45d25c749a2237de6b2b3237f076/.vscode/launch.json#L87

```suggestion
-   `TEST_DIR` - Directory where tests are found.
```

scripts/test/launchTestUtilities.ts
```diff
@@ -91,6 +91,7 @@ async function getVSCodeCliArgs(params: {
         extensionTestsEnv: {
             ['DEVELOPMENT_PATH']: projectRootDir,
             ['AWS_TOOLKIT_AUTOMATION']: params.suite,
+            ...(params.suite === 'e2e' && { ['E2E_TESTING_DIRECTORY']: process.env.E2E_TESTING_DIRECTORY }),
```
justinmk3, line 94: random special cases buried in the codepath should be avoided. `getVSCodeCliArgs` is a general util. the other env vars here are driven by parameters, yet now we have a random special case in the middle of this function. that makes it harder to reason about the flow.

jpinkney-aws: relativeTestEntryPoint just points to the test runner script, I don't think we want that part configurable since that's the thing that actually loads all the tests

scripts/test/launchTestUtilities.ts
```diff
@@ -91,6 +91,7 @@ async function getVSCodeCliArgs(params: {
         extensionTestsEnv: {
             ['DEVELOPMENT_PATH']: projectRootDir,
             ['AWS_TOOLKIT_AUTOMATION']: params.suite,
+            ['TEST_DIR']: process.env.TEST_DIR,
             ...params.env,
```
justinmk3, line 95: `...params.env` is already included so I guess the `['TEST_DIR']: process.env.TEST_DIR` assignement might not be necessary. But it's good for visibility/explicitness. 

Ideally, this would be passed as a parameter:

    ['TEST_DIR']: testDir,

but I don't know if the codepath is set up for that. Doesn't need to block this PR.

src/testE2E/index.ts
```diff
@@ -6,5 +6,5 @@
 import { runTests } from '../test/testRunner'
 
 export function run(): Promise<void> {
-    return runTests('src/testE2E', ['src/testInteg/globalSetup.test.ts'])
+    return runTests(process.env.TEST_DIR ?? 'src/testE2E', ['src/testInteg/globalSetup.test.ts'])
```
justinmk3, line 9: Seems strange that a test job would have tests located in some other directory, but the path `src/testInteg/globalSetup.test.ts` is still resolved correctly.

justinmk3 (approved):

PR #4350, opened 2024-01-30
fix(codewhisperer): Improve Code percentage reporting

leigaol:
## Problem
Code percentage is not accurate. 

## Solution

1. Only count user keystroke input as user tokens. Do not count code changes from copy and paste, git commands, etc.
2. Report CW tokens for below 2 options
    2.1  Sum of original accepted recommendations 
    2.2 Sum of Unmodified(retained, after user modification) accepted recommendations 
3. Do not emit code percentage if user is not invoking CW.

Tested E2E against dev env. 
Request id: 
6d7004a5-5f5f-4fa1-ad5a-56074a5ebb51 at about 01/30 6pm PST

service returned 200.

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4358, opened 2024-01-31
fix(codewhisperer): Fix edge cases in Code percentage reporting

leigaol:
## Problem

This is a follow up PR of https://github.com/aws/aws-toolkit-vscode/pull/4350.

Edge cases when user has a keystroke input but the IDE modifies it were not captured in previous PR. 

We wanted to capture user keystroke input by only selecting one char input. However, VSC document events somehow merged what user input and what IDE inputs, creating 2 char input or newline + indentation inputs. this PR is to catch these edge cases. 

The document change event in VSC is causing those edge cases to surface while in JetBrains, what user inputs and what IDE changes are in separated events so they do not need this kind of extra handling 

No change log is needed. 

## License

### DIFF
--- a/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
+++ b/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
@@ -24,6 +24,8 @@ interface CodeWhispererToken {
     accepted: number
 }
 
+const autoClosingKeystrokeInputs = ['[]', '{}', '()']
+
 /**
  * This singleton class is mainly used for calculating the code written by codeWhisperer
  */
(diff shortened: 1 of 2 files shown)


src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -238,21 +240,53 @@ export class CodeWhispererCodeCoverageTracker {
         this.addTotalTokens(filename, text.length)
     }
 
+    // For below 2 edge cases
+    // 1. newline character with indentation
+    // 2. 2 character insertion of closing brackets
+    public getCharacterCountFromComplexEvent(e: vscode.TextDocumentChangeEvent) {
+        if (e.document.languageId === 'java' && e.contentChanges.length === 2) {
```
zixlin7, line 247: why do we need `e.document.languageId === 'java'` ?

leigaol: Because this API behaves differently in Java files.

zixlin7 (approved):

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -24,6 +24,8 @@ interface CodeWhispererToken {
     accepted: number
 }
 
+const autoClosingKeystrokeInputs = ['[]', '{}', '()', `\"\"`, `\'\'`]
```
justinmk3, line 27: what is the intention of `\` ? do you want literal `"` or a literal `\"` ? If you only want literal `"` (and `'`), this should work:
```suggestion
const autoClosingKeystrokeInputs = ['[]', '{}', '()', '""', "''"]
```

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
leigaol: I tested it again. 

In other languages, there is a small chance that `e.contentChanges.length` becomes 2, while one is a empty event, another is a merged event. So I am improving the condition in last commit. This is tricky but I am able to reproduce.

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -238,21 +240,63 @@ export class CodeWhispererCodeCoverageTracker {
...
+            const text2 = e.contentChanges[1].text
+            if (text1.length === 0) {
+                if (text2.startsWith('\n') && text2.trim().length === 0) {
+                    return 1
+                }
+                if (autoClosingKeystrokeInputs.includes(text2)) {
+                    return 2
+                }
+            }
+            if (text2.length === 0) {
```
justinmk3, line 258: if `text1 !== 0 && text2 !== 0` , `getCharacterCountFromComplexEvent` will return `0`. is that expected?

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
```diff
@@ -238,21 +240,63 @@ export class CodeWhispererCodeCoverageTracker {
...
+            }
+        } else if (e.contentChanges.length === 1) {
+            const text = e.contentChanges[0].text
+            if (text.startsWith('\n') && text.trim().length === 0) {
+                return 1
+            }
+            if (autoClosingKeystrokeInputs.includes(text)) {
+                return 2
+            }
+        }
```
justinmk3, line 274: Attempt to deduplicate:

```suggestion
        function countChanges(cond: boolean, text: string): number {
            if (text.startsWith('\n') && text.trim().length === 0) {
                return 1
            }
            if (autoClosingKeystrokeInputs.includes(text)) {
                return 2
            }
            return 0
        }

        if (e.contentChanges.length === 2) {
            const text1 = e.contentChanges[0].text
            const text2 = e.contentChanges[1].text
            const text2Count = foo(text1.length === 0, text2)
            const text1Count = foo(text2.length === 0, text1)
            return text2Count > 0 ? text2Count : text1Count
        } else if (e.contentChanges.length === 1) {
            return foo(true, e.contentChanges[0].text)
        }
```

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
leigaol: Yes. When user input one key stroke, either text1 is empty or text2 is empty.

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
leigaol: This works. I have made this change in latest commit.

src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts
justinmk3: but i'm asking about the case where both are _not_ empty

leigaol: >where both are not empty

In this case, it won't be a user keystroke input. I have never seen this case btw.

justinmk3 (approved):

PR #4359, opened 2024-01-31
feat(codecatalyst): add environment variable overrides for service configs

hayemaxi:
Problem: While we have user settings.json overrides for codecatalyst, it is not easy for settings.json to be updated in CI environments.

Solution: Add ability to override on environment variables as well. Also, add generic functions for other services to do the same.

## License

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -329,7 +329,7 @@ Example:
 }
 ```
 
-Overrides specifically for CodeCatalyst can be set using the `aws.dev.codecatalystService` setting. This is a JSON object consisting of keys/values required to override API calls to CodeCatalyst: `region`, `endpoint`, `hostname`, and `gitHostname`. If this setting is present, then all keys need to be explicitly provided.
+<a name="codecatalyst-settings">Overrides specifically for CodeCatalyst can be set using the `aws.dev.codecatalystService` setting. This is a JSON object consisting of keys/values required to override API calls to CodeCatalyst: `region`, `endpoint`, `hostname`, and `gitHostname`. If this setting is present, then all keys need to be explicitly provided.</a>
 
 Example:
 
@@ -382,6 +382,14 @@ Environment variables can be used to modify the behaviour of VSCode. The followi
 -   `__DEV_ENVIRONMENT_PROJECT_NAME`: The project name associated with the running development environment. Automatically set when running the toolkit in Codecatalyst
 -   `__DEV_ENVIRONMENT_SPACE_NAME`: The space name associated with the running development environment. Automatically set when running the toolkit in Codecatalyst
 -   `__DEV_ENVIRONMENT_ORGANIZATION_NAME`: The organization name associated with the running development environment. Automatically set when running the toolkit in Codecatalyst
+
+The following are environment variable versions of the user `settings.json` overrides mentioned [here](#codecatalyst-settings). These will always override the toolkit defaults and those defined in `settings.json`.
+Unlike the user setting overrides, not all of these environment variables have to be set to make use of them.
+
+-   `__CODECATALYST_REGION`: for aws.dev.codecatalystService.region
+-   `__CODECATALYST_ENDPOINT`: for aws.dev.codecatalystService.endpoint
+-   `__CODECATALYST_HOSTNAME`: for aws.dev.codecatalystService.hostname
+-   `__CODECATALYST_GIT_HOSTNAME`: for aws.dev.codecatalystService.gitHostname
 
 #### Lambda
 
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -329,7 +329,7 @@ Example:
 }
 ```
 
-Overrides specifically for CodeCatalyst can be set using the `aws.dev.codecatalystService` setting. This is a JSON object consisting of keys/values required to override API calls to CodeCatalyst: `region`, `endpoint`, `hostname`, and `gitHostname`. If this setting is present, then all keys need to be explicitly provided.
+<a name="codecatalyst-settings">Overrides specifically for CodeCatalyst can be set using the `aws.dev.codecatalystService` setting. This is a JSON object consisting of keys/values required to override API calls to CodeCatalyst: `region`, `endpoint`, `hostname`, and `gitHostname`. If this setting is present, then all keys need to be explicitly provided.</a>
```
justinmk3, line 332: nice :)

src/shared/vscode/env.ts
```diff
@@ -89,3 +91,39 @@ export function getCodeCatalystSpaceName(): string | undefined {
...
+
+    // Log env var overrides, keeping track of which service we are logging for.
+    // This allows us to log only once when env vars for a service change.
+    if (overriden.length > 0) {
+        if (!(service in logConfigsOnce)) {
+            logConfigsOnce[service] = onceChanged(vars => {
+                getLogger().info(`using env vars for ${service} config: ${vars}`)
+            })
+        }
+        logConfigsOnce[service](overriden)
```
justinmk3, line 125: great technique! avoiding redundant noise in the logs is extremely useful because it makes the logs more meaningful.

src/shared/vscode/env.ts
```diff
@@ -89,3 +91,39 @@ export function getCodeCatalystSpaceName(): string | undefined {
...
+type ServiceConfig = Partial<{ [K in keyof ConfigToEnvMap]: any }>
+const logConfigsOnce: { [key: string]: ReturnType<typeof onceChanged> } = {}
+
+/**
+ * Accepts a service name and a {config key -> expected env var name} map.
+ * For each config key, check if the associated env var exists and return
+ * a config map with the found values. Changes are logged once for each
+ * service/found env var combos.
+ */
+export function getServiceEnvVarConfig(service: string, configToEnvMap: ConfigToEnvMap): ServiceConfig {
```
justinmk3, line 105: thanks for generalizing this, will be useful.

justinmk3 (approved):
🚀

src/test/shared/vscode/env.test.ts
```diff
@@ -0,0 +1,23 @@
...
+import { getServiceEnvVarConfig } from '../../../shared/vscode/env'
+
+describe('getServiceEnvVarConfig', function () {
+    it('gets service config', async function () {
+        const service = 'codecatalyst'
+        const configToEnvMap = {
+            region: '__CODECATALYST_REGION',
+            endpoint: '__CODECATALYST_ENDPOINT',
+        }
+        process.env['__CODECATALYST_ENDPOINT'] = 'test.endpoint'
```
justinmk3, line 16: should restore this in afterEach() to avoid polluting other tests.

src/shared/clients/codecatalystClient.ts
```diff
@@ -39,6 +40,13 @@ export interface CodeCatalystConfig {
     readonly gitHostname: string
 }
 
+const configToEnvMap: { [K in keyof CodeCatalystConfig]: string } = {
+    region: '__CODECATALYST_REGION',
+    endpoint: '__CODECATALYST_ENDPOINT',
+    hostname: '__CODECATALYST_HOSTNAME',
+    gitHostname: '__CODECATALYST_GIT_HOSTNAME',
+}
```
justinmk3, line 48: I wonder if it makes sense to generalize the env var override pattern as part of the work done in https://github.com/aws/aws-toolkit-vscode/pull/4324

Just a future idea, doesn't need to be solved here.

src/test/shared/vscode/env.test.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+import assert from 'assert'
+import { getServiceEnvVarConfig } from '../../../shared/vscode/env'
+
+describe('getServiceEnvVarConfig', function () {
+    const envVars: string[] = []
+
+    afterEach(() => {
+        envVars.forEach(v => delete process.env[v])
```
justinmk3, line 13: This assumes that the env var did not already exist. `__CODECATALYST_ENDPOINT` for example would be set by some pipelines that run the E2E tests. This test happens to not be an E2E test, so not an immediate problem.

PR #4367, opened 2024-02-01
fix: use IMDSv2 for all metadata service calls.

hayemaxi:
Problem: We did not attach the token to IMDS calls, resulting in IMDSv1 being used. The SDK does support using tokens but only for calls to the endpoint `/latest/meta-data/iam/security-credentials/`. https://github.com/aws/aws-sdk-js/blob/3333f8b49283f5bbff823ab8a8469acedb7fe3d5/lib/metadata_service.js#L123-L235

Solution: Call the "private" sdk method to get the token so code isn't duplicated, and attach that in the header of our calls just like the sdk does for the above endpoint.

AWS SDK v3 does not support token handling https://github.com/aws/aws-sdk-js-v3/issues/4004

#### Testing
Deployed to C9 since its an ec2 instance, no longer throws error about ec2 creds being unavailable.

## License


justinmk3 (approved):
Thanks for the clear explanation and relevant code comments.

PR #4368, opened 2024-02-02
fix(codewhisperer): Fix race condition when session id is reset in CW Telemetry

leigaol:
## Problem
There is a less than 4% chance that the sessionId in STE API is reset to next session's id. The userTriggerDecision in the toolkit telemetry does not suffer from this issue because its session id is not reset at next invocation.

Analysis of this issue is in an internal ticket. 

## Solution

Use the same session id as aws toolkit telemetry.  

## License

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,11 +49,11 @@ jobs:
                   echo "FEAT_NAME=" >> $GITHUB_ENV
                   echo "TAG_NAME=prerelease" >> $GITHUB_ENV
             - run: npm ci
-            - name: vsix
+            - name: vsix # TODO: For packages/toolkit release only
               run: |
                   npm run createRelease  # Generate CHANGELOG.md
-                  npm run generateNonCodeFiles -w packages/toolkit
                   cp ./README.quickstart.vscode.md ./README.md
+                  cd packages/toolkit
                   npm run package -- --feature "$FEAT_NAME"
             - uses: actions/upload-artifact@v4
               with:
--- a/buildspec/packageTestVsix.yml
+++ b/buildspec/packageTestVsix.yml
@@ -27,7 +27,6 @@ phases:
             - export HOME=/home/codebuild-user
             # Generate CHANGELOG.md
             - npm run createRelease
-            - npm run generateNonCodeFiles -w packages/toolkit
             - cp ./README.quickstart.vscode.md ./README.md
             - npm run package
 
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -31,6 +31,7 @@ Current quirks of the current monorepo status that should be resolved/evaluated 
 -   `packages/toolkit/scripts/` should be generalized and moved to the root of the project as needed.
 -   LICENSE, README.md, and other non-code artifacts that must be packaged into the .vsix are currently
     being copied into the packaging subproject directory from the root project directory as part of the `copyFiles` task.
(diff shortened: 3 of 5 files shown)


src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -317,43 +317,44 @@ export class TelemetryHelper {
...
-                        programmingLanguage: {
-                            languageName: runtimeLanguageContext.toRuntimeLanguage(
-                                this.sessionDecisions[0].codewhispererLanguage
-                            ),
+        if (this.sessionDecisions[0].codewhispererSessionId) {
+            client
+                .sendTelemetryEvent({
+                    telemetryEvent: {
+                        userTriggerDecisionEvent: {
+                            sessionId: this.sessionDecisions[0].codewhispererSessionId,
```
justinmk3, line 325: The `sessionId` function parameter is no longer used, should it be removed?

src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -317,43 +317,44 @@ export class TelemetryHelper {
...
-                telemetryEvent: {
-                    userTriggerDecisionEvent: {
-                        sessionId: sessionId,
-                        requestId: this.sessionDecisions[0].codewhispererFirstRequestId,
-                        customizationArn: selectedCustomization.arn === '' ? undefined : selectedCustomization.arn,
-                        programmingLanguage: {
-                            languageName: runtimeLanguageContext.toRuntimeLanguage(
-                                this.sessionDecisions[0].codewhispererLanguage
-                            ),
+        if (this.sessionDecisions[0].codewhispererSessionId) {
```
justinmk3, line 320: This condition is not checked before line 273. Is that correct?

zixlin7: +1, should we combine this one with the check on ln 256?

src/codewhisperer/util/telemetryHelper.ts
Will-ShaoHua: have a high level q: 
why passing `this.seesionDecisions[0].seesionId` could solve the issue?
IMO, passing here the function argument of `sendUserTriggerDecisionTelemetry(sessionId: string)` would be more robust and if it doesn't work correctly why using a "state" variable could fix this issue.

Will-ShaoHua: would it help if we simply make `sendUserTriggerDecisionTelemetry` an async call and await on the call sites instead of running async

zixlin7: I think the issue here is when next session calls sendUserTriggerDecisionTelemetry, it needs to send the previous session, not the sessionID from current session?

Will-ShaoHua: I see, that make sense

Will-ShaoHua: @leigaol  we should add a comment to note down this

andrewyuq: We should always use sessionId, let's make it to be the correct value but not abandon it. The logic should be the events for the old sessionId should be sent before it gets assigned to a new one.

andrewyuq: I would recommend to discard the usage of this.sessionDecisions, it feels better to manage a single sessionId per session in the `codewhispererSession.ts`, let's discuss on Monday on how to fix this.

PR #4372, opened 2024-02-02
fix(cloud9): use node 'fs' readdir when in cloud9 environment

hayemaxi:
Problem: Cloud9 does not support the readdir function of the vscode fs api. This results in 'not implemented' errors when the fsCommon module's readdir is called.

Solution: In `fsCommon`, use node's readdir instead of the vscode API if we are in cloud9.

#### Testing
Tested in C9, resolves the readdir issue when trying to upload a lambda from a folder.

Unit tests pending.

## License


hayemaxi:
continued at https://github.com/aws/aws-toolkit-vscode/pull/4374

PR #4374, opened 2024-02-02
fix: Use 'fs' methods if in C9

nkomonen-amazon:
## Problem:

We recently made some changes to our file system code that used the vscode.workspace.fs instance. The issue is that some of its methods are not supported in C9

## Solution:

If in C9 use the original "fs" instance methods instead for delete, readdir, and mkdir.

Also added tests.

This builds on top of @hayemaxi work from #4372 but I had to add some extra fixes for other methods

## License


hayemaxi:
Installed on C9, `delete` and `readdir` are working without error.

Let's add a bugfix changelog item

hayemaxi (approved):

PR #4387, opened 2024-02-05
feat(sam): support nodejs20.x Lambda runtime

justinmk3:
Also prepare for deprecation of nodejs 14.

## License


nkomonen-amazon:
Also looks like node14 is deprecated, so if you're still working on this it might be good to remove it
https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

justinmk3:
> Also looks like node14 is deprecated

yes will do in a followup

PR #4388, opened 2024-02-05
fix(ci): "deadsnakes" causes unreliable CI

justinmk3:
Problem:
deadsnakes sometimes fails our CI:

    lazr.restfulclient.errors.ServerError: HTTP Error 504: Gateway Time-out

Solution:
Remove use of deadsnakes. We no longer test against python 3.7/3.8.

## License


hayemaxi (approved):
qq: are we removing any tests against python at all? Or are we keeping latest e.g. 3.10?

justinmk3:
> qq: are we removing any tests against python at all? Or are we keeping latest e.g. 3.10?

Not removing any tests.

Issue #4390, opened 2024-02-05
unreliable test: 'mkdir ... creates folder but uses the "fs" module if in C9'


justinmk3:
test failure sometimes occurs:

```
  1) FileSystem
       mkdir()
         creates folder but uses the "fs" module if in C9: 'a':

      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert_1.default.ok(mkdirSpy.calledOnce)

      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (d:\a\aws-toolkit-vscode\aws-toolkit-vscode\src\test\srcShared\fs.test.ts:177:24)
```

nkomonen-amazon:
Looking

justinmk3:
Asserting `spy.callCount` is fragile. Can we assert the actual result instead?

PR #4391, opened 2024-02-05
fix: flaky fs test

nkomonen-amazon:
Reported by #4390, this test is failing on the mkdirSpy call count assertion.
Not exactly sure why this is flaky, but maybe it is due to the previous folders not being cleaned up.

## Solution

This commit deletes the created folders after each test finishes. Additionally, this commit checks the actual spy call count so that we can better debug the issue

## License

### DIFF
--- a/packages/toolkit/src/test/srcShared/fs.test.ts
+++ b/packages/toolkit/src/test/srcShared/fs.test.ts
@@ -24,6 +24,7 @@ describe('FileSystem', function () {
 
     before(async function () {
         fakeContext = await FakeExtensionContext.create()
+        await deleteTestRoot() // incase a previous test run failed to clean
     })
 
     beforeEach(async function () {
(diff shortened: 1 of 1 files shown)


src/test/srcShared/fs.test.ts
```diff
@@ -169,12 +174,12 @@ describe('FileSystem', function () {
...
                 sandbox.stub(extensionUtilities, 'isCloud9').returns(true)
                 const mkdirSpy = sandbox.spy(fsPromises, 'mkdir')
-                const dirPath = createTestPath(p)
+                dirPath = createTestPath(p)
 
                 await fsCommon.mkdir(dirPath)
 
                 assert.ok(existsSync(dirPath))
-                assert.ok(mkdirSpy.calledOnce)
+                assert.strictEqual(mkdirSpy.callCount, 1)
```
justinmk3, line 182: Yeah, `ok()` should be avoided almost always because it just gives less information, and it's _almost_ never needed.

justinmk3 (approved):

PR #4398, opened 2024-02-06
fix: monorepo script changes

hayemaxi:
Ensure that copying the .vsix to the root folder always works. Also, make sure running npm i at root also runs postinstall scripts for all workspaces.

---

`npm run generateNonCodeFiles` is redundant, it is already called in vscode:prepublish which is called by `vsce` automatically
during `npm run publish`

Also, need to make sure we are in the proper directory to package in CI.

## License

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,11 +49,11 @@ jobs:
                   echo "FEAT_NAME=" >> $GITHUB_ENV
                   echo "TAG_NAME=prerelease" >> $GITHUB_ENV
             - run: npm ci
-            - name: vsix
+            - name: vsix # TODO: For packages/toolkit release only
               run: |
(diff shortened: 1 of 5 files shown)


nkomonen-amazon (approved):

.github/workflows/release.yml
```diff
@@ -49,11 +49,11 @@ jobs:
                   echo "FEAT_NAME=" >> $GITHUB_ENV
                   echo "TAG_NAME=prerelease" >> $GITHUB_ENV
             - run: npm ci
-            - name: vsix
+            - name: vsix # TODO: For packages/toolkit release only
```
justinmk3, line 52: Is the comment describing the current situation or a future situation? This prerelease job can and should start producing multiple artifacts as soon as we can do so--no need to wait on that.

PR #4399, opened 2024-02-06
browser: move more code in to shared activation

nkomonen-amazon:
This moves more code in to the shared activation method so that we should be able
to start enabling CodeWhisperer (it should have all the pre-requisite setup it requires now)

- Shared code was moved out of `extension.ts#activate()` and in to `extensionShared.ts#activateShared()`
- The `ExtContext` object is now created in `activateShared()` and passed to the entrypoints as a return value

## License

### DIFF
--- a/packages/toolkit/src/extension.ts
+++ b/packages/toolkit/src/extension.ts
@@ -55,7 +55,6 @@ import { AwsResourceManager } from './dynamicResources/awsResourceManager'
 import globals from './shared/extensionGlobals'
 import { Experiments, Settings } from './shared/settings'
 import { isReleaseVersion } from './shared/vscode/env'
-import { Commands } from './shared/vscode/commands2'
 import { UriHandler } from './shared/vscode/uriHandler'
 import { telemetry } from './shared/telemetry/telemetry'
 import { Auth } from './auth/auth'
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/extension.ts
```diff
@@ -55,58 +53,31 @@ import { AwsResourceManager } from './dynamicResources/awsResourceManager'
...
-                    telemetry.aws_helpQuickstart.emit({ result: 'Succeeded' })
-                }
-            })
-        )
-    }
-
     try {
         // IMPORTANT: If you are doing setup that should also work in browser, it should be done in the function below
-        await activateShared(context, () => RegionProvider.fromEndpointsProvider(makeEndpointsProvider()))
+        const extContext = await activateShared(context, () =>
```
justinmk3, line 78: If it makes things easier, we could get rid of `ExtContext` . That is in the backlog

packages/toolkit/src/extension.ts
```diff
@@ -174,7 +130,7 @@ export async function activate(context: vscode.ExtensionContext) {
             context: extContext,
             regionProvider: globals.regionProvider,
             toolkitOutputChannel: globals.outputChannel,
-            remoteInvokeOutputChannel,
+            remoteInvokeOutputChannel: globals.invokeOutputChannel,
```
justinmk3, line 133: (off-topic) Another thing that has been in the backlog: there isn't really a strong reason to have a separate output channel for "remote invoke", and even if there were a reason, it should be created on-demand, not on startup...

justinmk3 (approved):
nice!

packages/toolkit/src/extension.ts
nkomonen-amazon: This wasn't a pain yet, so will keep it as-is. Good to know we can drop it though since it is a bit awkward

PR #4400, opened 2024-02-07
fix(codewhisperer): wrong session id reported by race condition when accept/reject during pagination

leigaol:
## Problem

There is a less than 4% chance that the sessionId in STE API is reset to next session's id.  See https://github.com/aws/aws-toolkit-vscode/pull/4368 (Previous PR had incorrect solution to the problem)

Root cause:

When user accept/reject a recommendation before all the pagination API calls finish, the userTriggerDecision of this session is not sent but stored in the IDE session state. When user accept/reject a recommendation next time, the stored state of previous session is sent along with new session, causing wrong session id for some user trigger decision events.

To reproduce it, invoke a recommendation that has multiple pagination calls, managed to hit TAB or escape very fast before the 2nd pagination response arrives at the plugin.  This is tricky to reproduce but possible. 

## Solution

Invoke SendTelemetryEvent API immediately when user hit accept/reject. Remove the extra if condition and extra variables. 

In the above edge case, with this solution, the userTriggerDecision of this session is sent immediately after user hit TAB/REJECT. By the time 2nd pagination response arrives, the userdecision of this response is not be reported to service since the user session does not contain this response. 

Cases I tested:
1. Invoke, type ahead, invoke again. The 2nd invoke will emit the STE of 1st invocation.
2. invoke, accept after pagination done, correct STE are sent.
3.Invoke, accept before pagination done, correct STE are sent. 

## License

### DIFF
--- a/packages/toolkit/src/codewhisperer/service/inlineCompletionService.ts
+++ b/packages/toolkit/src/codewhisperer/service/inlineCompletionService.ts
@@ -128,7 +128,6 @@ export class InlineCompletionService {
                 if (RecommendationHandler.instance.checkAndResetCancellationTokens()) {
                     RecommendationHandler.instance.reportUserDecisions(-1)
                     await vscode.commands.executeCommand('aws.codeWhisperer.refreshStatusBar')
-                    TelemetryHelper.instance.setIsRequestCancelled(true)
                     return
                 }
                 if (!RecommendationHandler.instance.hasNextToken()) {
@@ -136,7 +135,6 @@ export class InlineCompletionService {
                 }
                 page++
             }
-            TelemetryHelper.instance.setNumberOfRequestsInSession(page + 1)
         } catch (error) {
             getLogger().error(`Error ${error} in getPaginatedRecommendation`)
         }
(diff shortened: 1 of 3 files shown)


andrewyuq (approved):
Change looks good, please make sure to spend extra time to test this(dedicated time to play around in sandbox and monitor debug logs).

Will-ShaoHua:
besides functional test, @leigaol could you run telemetry.test.ts manually and make sure they all pass and post here.

Will-ShaoHua:
`telemetry.test.ts` is skipped for now because we need to move it to integration test.

Will-ShaoHua:
changes look to me

justinmk3 (approved):

zixlin7 (approved):

zixlin7:
Can we make toolkit telemetry use the same sessionID as STE?

leigaol:
> Can we make toolkit telemetry use the same sessionID as STE?

In my manual testing, these 2 are always the same string. The variable name is different.

leigaol:
> besides functional test, @leigaol could you run telemetry.test.ts manually and make sure they all pass and post here. note that while you're running test cases there, do not move cursor or change focus or anything) otherwise tests will fail

I am doing lots of manual testing. THe test cases do not cover race conditions

PR #4401, opened 2024-02-07
fix(CodeTransform): pre-select Java version

dhasani23:
## Problem

UX request to change the way we validate input: rather than letting user select a project, then run all validation checks, we want to run all validation checks on all open projects to filter valid projects for the user, plus pre-select the Java version for them if we were able to detect it and it's a supported version (8 or 11).

## Solution

Implement what is described above.

## License

### DIFF
--- a/packages/toolkit/src/codewhisperer/commands/startTransformByQ.ts
+++ b/packages/toolkit/src/codewhisperer/commands/startTransformByQ.ts
@@ -9,7 +9,7 @@ import * as fs from 'fs'
 import * as os from 'os'
 import { getLogger } from '../../shared/logger'
 import * as CodeWhispererConstants from '../models/constants'
-import { transformByQState, StepProgress, TransformByQReviewStatus } from '../models/model'
+import { transformByQState, StepProgress, TransformByQReviewStatus, JDKVersion, DropdownStep } from '../models/model'
 import {
     throwIfCancelled,
     startJob,
@@ -21,10 +21,11 @@ import {
     convertToTimeString,
     convertDateToTimestamp,
     getOpenProjects,
-    validateProjectSelection,
     getVersionData,
+    validateOpenProjects,
 } from '../service/transformByQHandler'
 import { QuickPickItem } from 'vscode'
+import { MultiStepInputFlowController } from '../../shared//multiStepInputFlowController'
 import path from 'path'
 import { sleep } from '../../shared/utilities/timeoutUtils'
 import { encodeHTML } from '../../shared/utilities/textUtilities'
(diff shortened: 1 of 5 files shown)


packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -78,89 +77,123 @@ export async function getOpenProjects() {
...
-        throw new ToolkitError('Project selected is not Java 8 or Java 11', { code: 'UnsupportedJavaVersion' })
+        throw new ToolkitError('No valid Maven build file found', { code: 'CouldNotFindPomXml' })
+    }
+
+    /*
+     * these projects we know must contain a pom.xml and a .java file
+     * here we try to get the Java version of each project so that we
+     * can pre-select a default version in the QuickPick for them
+     */
+    const projectsValidToTransform = new Map<vscode.QuickPickItem, JDKVersion | undefined>()
```
damntrecky, line 135: Can we make this a function call like `const projectsValidToTransform = getValidProjectsToTransform()` and just take the following 50 lines or so and make it a function.

packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -78,89 +77,123 @@ export async function getOpenProjects() {
...
-            errorReason = 'JavapExecutionError'
-            // should never happen -- stdout from javap has always been much, much smaller than the default buffer limit of 1MB
-            if (Buffer.byteLength(spawnResult.stdout, 'utf-8') > CodeWhispererConstants.maxBufferSize) {
-                errorReason += '-BufferOverflow'
-            }
-        } else {
-            errorReason = 'JavapSpawnError'
+        throw new ToolkitError('No Java projects found', { code: 'CouldNotFindJavaProject' })
+    }
+    const mavenJavaProjects = []
```
damntrecky, line 108: Possibly the same here to make more modular code. Something like `const mavenJavaProjects = getMavenJavaProjects()`

packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
damntrecky: Blocker

damntrecky (approved):
Tentative approval. Please make some of the code more modular and re-usable by making helper functions.

packages/toolkit/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -65,26 +66,82 @@ export async function startTransformByQWithProgress() {
...
-    totalSteps: number
-    targetLanguage: QuickPickItem
-    targetVersion: QuickPickItem
-    project: QuickPickItem
+    project: QuickPickItem | undefined
+    sourceJavaVersion: QuickPickItem | undefined
 }
 
-async function collectInput(validProjects: vscode.QuickPickItem[]) {
+async function collectInput(validProjects: Map<vscode.QuickPickItem, JDKVersion | undefined>) {
```
nkomonen-amazon, line 73: Does this work?

```suggestion
async function collectInput(validProjects: { [item: vscode.QuickPickItem]: JDKVersion | undefined }) {
```

I think we want to avoid using the actual Map type since it more explicit in the required implementation

packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -78,89 +77,123 @@ export async function getOpenProjects() {
...
-        telemetry.codeTransform_isDoubleClickedToTriggerInvalidProject.emit({
-            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
-            codeTransformPreValidationError: 'NonMavenProject',
-            result: MetadataResult.Fail,
-            reason: buildSystem === BuildSystem.Gradle ? buildSystem : 'NoPomFileFound',
-        })
-        throw new ToolkitError('No valid Maven build file found', { code: 'CouldNotFindPomXml' })
+export async function validateOpenProjects(projects: vscode.QuickPickItem[]) {
+    const javaProjects = []
+    for (const project of projects) {
```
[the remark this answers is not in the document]
dhasani23: I think what you're referring to is the code we've had since November (https://github.com/aws/aws-toolkit-vscode/pull/4401/files/66be7c4d834748ff81141fb5caebb59d2c5d11d1#r1482102566) which checks for .class files to be present. I guess I could generalize this `vscode.workspace.findFiles` in a separate method and have it accept the pattern we are searching for (ex. `**/*.java` or `**/*.class`) but that doesn't seem necessary to me.

packages/toolkit/src/codewhisperer/commands/startTransformByQ.ts
dhasani23: The literal `{}` will not let me use the QuickPickItem as a key, so unfortunately that doesn't work

packages/toolkit/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -65,26 +66,82 @@ export async function startTransformByQWithProgress() {
...
+
+async function pickSourceVersion(
+    input: MultiStepInputFlowController,
+    state: Partial<UserInputState>,
+    validSourceVersions: vscode.QuickPickItem[]
+) {
+    let detectedJavaVersion = undefined
+    const sourceJDKVersion = transformByQState.getSourceJDKVersion()
+    if (sourceJDKVersion === JDKVersion.JDK8) {
+        detectedJavaVersion = validSourceVersions[0]
```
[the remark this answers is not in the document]
dhasani23: We're simply using `detectedJavaVersion` as the `activeItem` in the QuickPick

packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
dhasani23: Fixed

packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
dhasani23: Fixed

packages/toolkit/src/codewhisperer/models/constants.ts
```diff
@@ -297,7 +293,7 @@ export const transformByQCompletedMessage = 'Transformation completed'
 export const transformByQPartiallyCompletedMessage = 'Transformation partially completed'
 
 export const noPomXmlFoundMessage =
-    'We could not find a valid configuration file. We currently support Maven build tool and require a POM.xml in the root directory to identify build configurations. Be sure to also build your project.'
+    'None of your open Java projects are supported by Q Code Transformation. We were unable to find a pom.xml in any of your Java projects. We only support Java projects built on Maven at the moment.'
```
justinmk3, line 296: Should this be "Amazon Q", I thought that marketing decision was that "Q" should never be presented without "Amazon".
```suggestion
    'None of your open Java projects are supported by Amazon Q Code Transformation. We were unable to find a pom.xml in any of your Java projects. We only support Java projects built on Maven at the moment.'
```

same for the other cases in this PR and all other code/messages.

dhasani23: That's correct. Fixed.

nkomonen-amazon (approved):

PR #4405, opened 2024-02-07
tests: Allow Amazon Q e2e tests to call the backend

jpinkney-aws:
_pending creation of amazonq-test-account secret_

## Problem
- Amazon Q e2e tests currently use mocked responses instead of calling the backend directly

## Solution
- Allow Amazon Q e2e tests to call the backend directly

## License

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -420,7 +420,7 @@ Unlike the user setting overrides, not all of these environment variables have t
 
 #### Lambda
 
--   `AUTH_UTIL_LAMBDA_ARN`: The Amazon Resource Name (ARN) of the lambda function
+-   `AUTH_UTIL_LAMBDA_ARN`: The Auth Util Lambda is used to log into using Builder ID/IdC automatically when running e2e tests. This is the arn that points to the auth util lambda.
 
 #### ECS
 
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -441,8 +441,10 @@ Unlike the user setting overrides, not all of these environment variables have t
 -   `GITHUB_ACTION`: The name of the current GitHub Action workflow step that is running
 -   `CODEBUILD_BUILD_ID`: The unique ID of the current CodeBuild build that is executing
 -   `AWS_TOOLKIT_AUTOMATION`: If tests are currently being ran
--   `DEVELOPMENT_PATH`: The path to the aws toolkit vscode project
+-   `AWS_TOOLKIT_TEST_START_URL`: The start url you want to use on E2E tests
+-   `AWS_TOOLKIT_TEST_START_URL_REGION`: The region for the start url you want to use on E2E tests
```
justinmk3, line 445:
```suggestion
-   `TEST_AUTH_SSO_STARTURL`: The start url you want to use on E2E tests
-   `TEST_AUTH_SSO_STARTURL_REGION`: The region for the start url you want to use on E2E tests
```

CONTRIBUTING.md
```diff
@@ -441,8 +441,10 @@ Unlike the user setting overrides, not all of these environment variables have t
...
 -   `AWS_TOOLKIT_TEST_NO_COLOR`: If the tests should include colour in their output
+-   `DEVELOPMENT_PATH`: The path to the aws toolkit vscode project
 -   `TEST_DIR` - The directory where the test runner should find the tests
```
justinmk3, line 448: Parameters are starting to sprawl. Why is `DEVELOPMENT_PATH` separate from `TEST_DIR`? Why can't CWD be used by the scripts? Why do we need to parameterize that?

jpinkney-aws: Development path points to the script that loads in the tests through mocha: e.g. https://github.com/aws/aws-toolkit-vscode/blob/master/packages/toolkit/src/testE2E/index.ts. TEST_DIR (optionally) tells the mocha test loader which subset of the tests should be loaded. I'm not sure how CWD can be used here because we can't know which test directories we want to use unless it's manually specified. Development path is also required by VSCode to know how to load and report the tests. I think they're serving two different purposes

jpinkney-aws: Development path points to the root of the project directory. Test dir points to the subset of tests that you want to run for the given test execution. How would cwd be used in this case? You want the ability to select the subset of tests either through some env var or some other way

hayemaxi: it seems that DEVELOPMENT_PATH is always cwd, which should always be `packages/toolkit`(since that is where you initiate the tests from). Doesn't seem like we need this to be configurable, can we just remove it?

jpinkney-aws: We might be able to. But I think that needs more research in a different PR then this one

justinmk3 (approved):
Please remove DEVELOPMENT_PATH in a followup PR

PR #4406, opened 2024-02-07
refactor: remove 'glob' from CW

nkomonen-amazon:
## Problem:
We cannot use 'glob' in browser mode since it breaks if not in nodejs

## Solution:
Use `vscode.workspace.findFiles` which simply searches the entire workspace (or current folder
if no workspace was selected)

Note the previous comments about it not working, which was due to us needing to use a relative path since it searches the workspace/root folder only

## License

### DIFF
--- a/packages/toolkit/src/codewhisperer/util/supplementalContext/utgUtils.ts
+++ b/packages/toolkit/src/codewhisperer/util/supplementalContext/utgUtils.ts
@@ -3,7 +3,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { glob } from 'glob'
 import * as fs from 'fs-extra'
 import * as path from 'path'
 import * as vscode from 'vscode'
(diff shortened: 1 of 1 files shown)


Will-ShaoHua:
Nick could you do a quick functional test by

1. have a file Main.java and MainTest.java
2. trigger CW in Main.java
3. inspect logs (there should be something like
```
CodeWhispererSupplementalContext:
    isUtg: false,
    isProcessTimeout: false,
    contentsLength: 1291,
    latency: 3.1642799377441406,
Chunk 0:
        Path: **/MainTest.java        
Content: aaaaaaaaaaaaaaaa

```
then it should be fine

Will-ShaoHua (approved):

nkomonen-amazon:
Followed the steps from @Will-ShaoHua and can confirm that it is showing the expected information (but had to test the `MainTest.java` file instead)

PR #4408, opened 2024-02-07
Initial tailing implementation

jsicheng:
## License

### DIFF
--- a/packages/toolkit/src/cloudWatchLogs/commands/tailLogGroup.ts
+++ b/packages/toolkit/src/cloudWatchLogs/commands/tailLogGroup.ts
@@ -1,0 +1,354 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+import * as nls from 'vscode-nls'
+import {
+    CloudWatchLogsGroupInfo,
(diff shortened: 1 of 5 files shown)


packages/toolkit/src/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -0,0 +1,354 @@
...
+        // Pre-stream exceptions are captured here
+        console.log(err)
+    }
+}
+
+function formatGroupArn(groupArn: string): string {
+    return groupArn.endsWith(':*') ? groupArn.substring(0, groupArn.length - 2) : groupArn
+}
+
+export async function handleResponseAsync(response: StartLiveTailCommandOutput, uri: vscode.Uri) {
```
justinmk3, line 137: This is the main new code. The other code looks pretty redundant, right? 

To start with, I think the "tail" feature can just be an option/button on the existing Log Stream view. If we also add a new "Tail" command, it should use the same codepath with that option enabled. That should avoid most of the redundant code.

jsicheng: Sorry, this PR was created by mistake. We were testing stuff out for a quick POC.

justinmk3: Nevertheless, please consider the above comments in the non-POC case.

PR #4413, opened 2024-02-09
refactor: reorder initialize since breaks in C9

nkomonen-amazon:
The order of the intialize methods caused issues in C9 due to the order of execution.

## License


hayemaxi:
works in c9, nice

hayemaxi (approved):

PR #4424, opened 2024-02-12
feature(q): add Amazon Q + CodeWhisperer subproject files

hayemaxi:
Initial commit to add the subproject to packages/. This includes:

- Extending tsconfigs for subprojects
- .vscode settings to start the extension sandbox in vscode
- barebones activation functions and package.json, package.nls.json

Coming work:
- webpack configs
- more npm scripts (`serve` is a placeholder)
- Get the extension to show up in the side bar.
- Package to .vsix

## License


packages/.tsconfig.packages.json
```diff
@@ -0,0 +1,22 @@
+{
```
nkomonen-amazon, line 1: What is the difference between this and the tsconfig in the root of the project? Should we just have the root tsconfig and the sub-tsconfigs point to that?

nkomonen-amazon (approved):

PR #4425, opened 2024-02-12
refactor: move some gumby out of CW files

nkomonen-amazon:
This moves gumby commands to their own file.
- We are doing this since CW unintentionally imports gumby, but gumby code uses fs-extra which breaks browser mode.
- In the future we need make sure to keep CW and Gumby code separate where possible

## License

### DIFF
--- a/packages/toolkit/src/amazonq/explorer/amazonQChildrenNodes.ts
+++ b/packages/toolkit/src/amazonq/explorer/amazonQChildrenNodes.ts
@@ -7,13 +7,14 @@ import * as vscode from 'vscode'
 import * as nls from 'vscode-nls'
 import { Commands, placeholder } from '../../shared/vscode/commands2'
 import { getIcon } from '../../shared/icons'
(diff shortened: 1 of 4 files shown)


dhasani23 (approved):

packages/toolkit/src/codewhisperer/commands/gumbyCommands.ts
```diff
@@ -0,0 +1,50 @@
+/*!
```
justinmk3, line 1: should this file live in `packages/toolkit/src/amazonqGumby/` ?

dhasani23: That should also work fine, and probably makes more sense since it will be away from `codewhisperer` files as we're technically a separate service

justinmk3: that seems like it would be more intuitive

nkomonen-amazon: this makes sense, I missed that

PR #4428, opened 2024-02-13
refactor: extract codewhisperer tree view setup

nkomonen-amazon:
## Problem:
We want to use the CodeWhisperer tree view during browser mode. The issue is when enabling it the current code will try and enable the other tree views such as the Resource Explorer.

Additionally, web mode breaks due to imports in `activation.ts`.

## Solution:
Extract the logic for the tree view that we want to work in both node and browser environments in to its own shared activate function. We then have more control over what we enable and they are guaranteed to work in the browser.

Now we can use that shared activation for both environments.

## License

### DIFF
--- a/packages/toolkit/src/awsexplorer/activationShared.ts
+++ b/packages/toolkit/src/awsexplorer/activationShared.ts
@@ -1,0 +1,51 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+import { createToolView, ToolView } from './toolView'
+import { telemetry } from '../shared/telemetry/telemetry'
+import { cdkNode, CdkRootNode } from '../cdk/explorer/rootNode'
+import {
+    CodeWhispererNode,
+    getCodewhispererNode,
+    refreshCodeWhisperer,
+    refreshCodeWhispererRootNode,
+} from '../codewhisperer/explorer/codewhispererNode'
+import { once } from '../shared/utilities/functionUtils'
+
+/**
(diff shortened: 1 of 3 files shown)


packages/toolkit/src/awsexplorer/activationShared.ts
```diff
@@ -0,0 +1,51 @@
...
+    getCodewhispererNode,
+    refreshCodeWhisperer,
+    refreshCodeWhispererRootNode,
+} from '../codewhisperer/explorer/codewhispererNode'
+import { once } from '../shared/utilities/functionUtils'
+
+/**
+ * Activates vscode Views (eg tree view) that work in any vscode environment (nodejs or browser).
+ */
+export async function activateViewsShared(context: vscode.ExtensionContext): Promise<void> {
```
justinmk3, line 21: nit: it's already in a separate file so don't need a new name. this could be called `activate` just like all the other activate() cases.

justinmk3: alternatively this could live in `packages/toolkit/src/awsexplorer/activation.ts` , then it might need a new name. 

that is probably better, since the direction we are going is that (almost) all features would be web-capable. So separating things into new files would be throwaway work, thus should be avoided unless there is a technical need for it (e.g. to avoid import issues)

nkomonen-amazon: This file was created to avoid import issues from `packages/toolkit/src/awsexplorer/activation.ts` like you mentioned.

The plan would be to incrementally move features in to `activationShared` that work in both.

hayemaxi: yeah, non-agnostic code should eventually be the exception (and require its own file(s))

hayemaxi (approved):

justinmk3 (approved):

PR #4430, opened 2024-02-13
webpack, packaging generalizations for subprojects

hayemaxi:
## Problem
`package.ts` and webpack configs are not generalizable or extendable in their current state.

## Solution
`package.ts` now works on any subproject in `packages/`, and webpack configs can be inherited from base configs now.

Also includes:
- other small changes for npm scripts run in package.jsons
- browser entry point for amazonq

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4431, opened 2024-02-13
web: web tests in github CI

nkomonen-amazon:
Will run the web unit tests in github CI

## License


packages/toolkit/scripts/test/launchTestUtilities.ts
```diff
@@ -64,7 +64,7 @@ async function getVSCodeCliArgs(params: {
     let disableExtensionsArgs: string[] = []
     let disableWorkspaceTrustArg: string[] = []
 
-    if (params.suite !== 'unit') {
+    if (params.suite === 'integration' || params.suite === 'e2e') {
         disableExtensionsArgs = await getCliArgsToDisableExtensions(params.vsCodeExecutablePath, {
```
justinmk3, line 68: Should we pass `--disable-extensions` (i.e disable _all_ extensions) for the unit/web cases?

.github/workflows/node.js.yml
```diff
@@ -53,6 +53,29 @@ jobs:
                   file: ./coverage/lcov.info
                   flags: codewhisperer
 
+    web:
+        name: test Web
+        runs-on: macos-latest
```
justinmk3, line 58: Not a blocker, but we should probably do Linux for this case. Because it's faster than macOS. Unless there's a reason not to.

justinmk3 (approved):

.github/workflows/node.js.yml
nkomonen-amazon: Nope no reason for mac. Good to know linux is faster, will updated.

packages/toolkit/package.json
```diff
@@ -4247,11 +4247,13 @@
         "copyFiles": "ts-node ./scripts/build/copyFiles.ts",
         "buildScripts": "npm run generateClients && npm run generatePackage && npm run generateNonCodeFiles && npm run copyFiles",
         "browserWatch": "npm run clean && npm run buildScripts && webpack --config webpack.browser.config.js --watch",
+        "browserCompile": "npm run clean && npm run buildScripts && webpack --config webpack.browser.config.js",
```
hayemaxi, line 4250: note: conflicts with https://github.com/aws/aws-toolkit-vscode/pull/4430

can we reference by --config-name instead of filepath, for future reference?

packages/toolkit/scripts/test/launchTestUtilities.ts
```diff
@@ -18,7 +18,7 @@ const minimum = 'minimum'
 
 const disableWorkspaceTrust = '--disable-workspace-trust'
 
-type SuiteName = 'integration' | 'e2e' | 'unit'
+type SuiteName = 'integration' | 'e2e' | 'unit' | 'web'
```
hayemaxi, line 21: should we be consistence with `browser` vs `web`? applicable throughout cr

nkomonen-amazon: Justin recently mentioned we should start using `web` instead since that seems to be the branding VS Code is going with. So we'll be using `web` for now and swapping uses of `browser`

packages/toolkit/package.json
nkomonen-amazon: Updated to do this

hayemaxi (approved):

packages/toolkit/scripts/test/launchTestUtilities.ts
nkomonen-amazon: Makes sense, this was defined in the launch.json but not here. Added it in below

justinmk3:
E2E tests failing on master:

```
  1) transformByQ
       "before all" hook for "WHEN upload payload with missing sha256 in headers THEN fails to upload":
     TypeError [ERR_INVALID_ARG_VALUE]: The argument 'file' cannot be empty. Received ''
      at new NodeError (node:internal/errors:405:5)
      at normalizeSpawnArguments (node:child_process:561:11)
      at spawnSync (node:child_process:879:8)
      at installProjectDependencies (src/codewhisperer/service/transformByQHandler.ts:392:34)
      at zipCode (src/codewhisperer/service/transformByQHandler.ts:498:9)
      at Context.<anonymous> (src/testE2E/amazonqGumby/transformByQ.test.ts:33:39)
```

PR #4433, opened 2024-02-14
fix(CodeTransform): prompt user for JAVA_HOME

dhasani23:
## Problem

Our `Maven` shell commands sometimes fail and we suspect it is because we are not running those commands with the proper JAVA_HOME (needs to match the project JDK).

## Solution

If the Maven wrapper executable exists in the project directory, use that for both `Maven` shell commands, else use regular `mvn` -- and if the Java version used by Maven differs from the project's Java version, prompt the user for JAVA_HOME and use that when running the shell commands.

## License

### DIFF
--- a/packages/toolkit/src/codewhisperer/commands/startTransformByQ.ts
+++ b/packages/toolkit/src/codewhisperer/commands/startTransformByQ.ts
@@ -81,9 +81,25 @@ async function collectInput(validProjects: Map<vscode.QuickPickItem, JDKVersion 
     const validSourceVersions: vscode.QuickPickItem[] = versionsArray.map(version => ({
         label: version,
     }))
+    validSourceVersions.push({ label: 'Other' }) // if user selects 'Other', terminate execution
     await MultiStepInputFlowController.run(input => pickSourceVersion(input, state, validSourceVersions))
     if (!state.sourceJavaVersion) {
         throw new ToolkitError('No version selected', { code: 'NoVersionSelected' })
+    } else if (state.sourceJavaVersion.label === 'Other') {
+        telemetry.codeTransform_jobStartedCompleteFromPopupDialog.emit({
+            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+            codeTransformJavaSourceVersionsAllowed: JDKToTelemetryValue(
+                transformByQState.getSourceJDKVersion()!
+            ) as CodeTransformJavaSourceVersionsAllowed, // will be 'undefined'
+            codeTransformJavaTargetVersionsAllowed: JDKToTelemetryValue(
+                transformByQState.getTargetJDKVersion()
+            ) as CodeTransformJavaTargetVersionsAllowed,
+            result: MetadataResult.Fail,
+        })
+        await vscode.window.showErrorMessage(CodeWhispererConstants.unsupportedJavaVersionSelectedMessage, {
+            modal: true,
+        })
+        throw new ToolkitError('', { code: 'OtherVersionSelected' })
     }
     return state as UserInputState
 }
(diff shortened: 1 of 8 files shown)


packages/toolkit/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -81,9 +81,25 @@ async function collectInput(validProjects: Map<vscode.QuickPickItem, JDKVersion
     const validSourceVersions: vscode.QuickPickItem[] = versionsArray.map(version => ({
         label: version,
     }))
+    validSourceVersions.push({ label: 'Other' }) // if user selects 'Other', terminate execution
     await MultiStepInputFlowController.run(input => pickSourceVersion(input, state, validSourceVersions))
     if (!state.sourceJavaVersion) {
         throw new ToolkitError('No version selected', { code: 'NoVersionSelected' })
+    } else if (state.sourceJavaVersion.label === 'Other') {
```
dhasani23, line 88: Small change: allow user to select 'Other' in Java version dropdown

packages/toolkit/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -141,6 +157,47 @@ async function pickSourceVersion(
...
+        javaVersionUsedByMaven = javaVersionUsedByMaven.slice(0, 3)
+        if (javaVersionUsedByMaven === '1.8') {
+            javaVersionUsedByMaven = JDKVersion.JDK8
+        } else if (javaVersionUsedByMaven === '11.') {
+            javaVersionUsedByMaven = JDKVersion.JDK11
+        }
+    }
+    if (javaVersionUsedByMaven !== transformByQState.getSourceJDKVersion()) {
+        // means either javaVersionUsedByMaven is undefined or it does not match the project JDK
+        // TO-DO: give user help on how to find JAVA_HOME for corresponding project JDK
```
dhasani23, line 192: Most likely this help will come in the form of a link to our documentation or some TBD messaging right in the IDE.

packages/toolkit/src/codewhisperer/models/constants.ts
```diff
@@ -357,10 +362,6 @@ export const transformationJobTimeoutSeconds = 72000
 
 export const progressIntervalMs = 1000
 
-export const targetLanguages = ['Java']
-
-export const targetVersions = new Map<string, string[]>([['Java', ['JDK17']]])
-
```
dhasani23, line 363: Both of these were unused.

packages/toolkit/src/codewhisperer/models/model.ts
```diff
@@ -252,6 +252,7 @@ export enum BuildSystem {
 export class ZipManifest {
     sourcesRoot: string = 'sources/'
     dependenciesRoot: string | undefined = 'dependencies/'
+    buildLogs: string = 'build-logs.txt'
```
dhasani23, line 255: Small change: add this entry to our `manifest.json` since backend is expecting it to be there.

packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -375,26 +375,19 @@ function getFilesRecursively(dir: string, isDependenciesFolder: boolean): string
...
+    let baseCommand = transformByQState.getMavenName()
+    // if JAVA_HOME could not be found or did not match the project JDK, baseCommand is prefixed with JAVA_HOME={user-input}
+    if (transformByQState.getJavaHome() !== undefined) {
+        baseCommand = `JAVA_HOME=${transformByQState.getJavaHome()} ${baseCommand}`
     }
+    const modulePath = transformByQState.getProjectPath()
 
     transformByQState.appendToErrorLog(`Running command ${baseCommand} clean install`)
 
+    // Note: IntelliJ runs 'clean' separately from 'install'. Evaluate benefits (if any) of this.
```
dhasani23, line 390: I personally don't think there are any benefits to doing this. Can change later if needed but all `clean` does is delete the "target" directory so doesn't seem necessary to do it separately.

packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -497,61 +487,32 @@ function copyProjectDependencies(buildCommand: CodeTransformMavenBuildCommand, m
...
-            mavenCopyDepsFailed = true
-        }
-    }
-
-    const copyDependenciesFailed = mavenCopyDepsFailed && mavenWrapperCopyDepsFailed
-
-    if (copyDependenciesFailed) {
         void vscode.window.showErrorMessage(CodeWhispererConstants.dependencyErrorMessage)
+        throw err
     }
```
dhasani23, line 512: In the two places above, we simply `throw err` because if either `Maven` shell command fails we simply stop the transformation process.

packages/toolkit/src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -426,6 +427,7 @@ export class ProposedTransformationExplorer {
                 codeTransformPatchViewerCancelSrcComponents: 'cancelButton',
                 codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
                 codeTransformJobId: transformByQState.getJobId(),
+                codeTransformStatus: transformByQState.getStatus(),
```
dhasani23, line 430: Small change: updated telemetry so that we know the status of the job when someone either accepts or rejects our proposed changes.

nkomonen-amazon (approved):

packages/toolkit/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -141,6 +157,47 @@ async function pickSourceVersion(
...
+        const javaHome = await vscode.window.showInputBox({
```
JuliaABurch, line 193: Do we know what we want this experience to look like when we switch to chat invocation for project validation? Is it still ok to prompt the user with a pop-up input box then?

dhasani23: I don't know the details of how exactly it's going to work in chat, but we definitely will need to somehow prompt the user with the input box, either as a QuickPick or maybe directly in the chat.

JuliaABurch (approved):

PR #4442, opened 2024-02-16
browser: remove fs-extra usage in CodeWhisperer logic

nkomonen-amazon:
Adds additional changes on top of #4421 from @andrewyuq (I wasn't able to add my commit and github autoclosed his PR).

This PR removes fs-extra logic from CodeWhisperer logic so that it can run in web mode.

## License

### DIFF
--- a/packages/toolkit/src/codewhisperer/util/dependencyGraph/dependencyGraph.ts
+++ b/packages/toolkit/src/codewhisperer/util/dependencyGraph/dependencyGraph.ts
@@ -152,12 +152,6 @@ export abstract class DependencyGraph {
         return dir + extension
     }
 
-    protected async removeZip(zipFilePath: string) {
-        if (await fsCommon.exists(zipFilePath)) {
-            await fsCommon.unlink(zipFilePath)
-        }
-    }
-
     protected getTruncDirPath(uri: vscode.Uri) {
         if (this._truncDir === '') {
             this._truncDir = path.join(
@@ -182,7 +176,7 @@ export abstract class DependencyGraph {
 
     public async removeTmpFiles(truncation: Truncation) {
         getLogger().verbose(`Cleaning up temporary files...`)
-        await this.removeZip(truncation.zipFilePath)
+        await fsCommon.delete(truncation.zipFilePath)
         await fsCommon.delete(truncation.rootDir)
         getLogger().verbose(`Complete cleaning up temporary files.`)
     }
--- a/packages/toolkit/src/codewhisperer/util/dependencyGraph/dependencyGraph.ts
+++ b/packages/toolkit/src/codewhisperer/util/dependencyGraph/dependencyGraph.ts
@@ -3,10 +3,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import * as fs from 'fs-extra'
+import { fsCommon } from '../../../srcShared/fs'
 import * as vscode from 'vscode'
 import admZip from 'adm-zip'
-import { existsSync, statSync } from 'fs'
 import { asyncCallWithTimeout } from '../commonUtil'
 import * as path from 'path'
 import { tempDirPath } from '../../../shared/filesystemUtilities'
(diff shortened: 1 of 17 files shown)


packages/toolkit/src/srcShared/fs.ts
```diff
@@ -152,12 +152,7 @@ export class FileSystemCommon {
     async copy(source: vscode.Uri | string, target: vscode.Uri | string): Promise<void> {
         const sourcePath = FileSystemCommon.getUri(source)
         const targetPath = FileSystemCommon.getUri(target)
-        return await fs.copy(sourcePath, targetPath)
-    }
-
-    async unlink(uri: vscode.Uri | string): Promise<void> {
```
nkomonen-amazon: We don't need `unlink()`, can just use `fsCommon.delete()`

packages/toolkit/src/codewhisperer/util/dependencyGraph/dependencyGraph.ts
```diff
@@ -153,18 +152,6 @@ export abstract class DependencyGraph {
...
 
-    protected removeDir(dir: string) {
-        if (existsSync(dir)) {
-            fs.removeSync(dir)
-        }
-    }
-
-    protected removeZip(zipFilePath: string) {
-        if (existsSync(zipFilePath)) {
-            fs.unlinkSync(zipFilePath)
```
justinmk3, line 164: Just for future reference: checking for existence is a race condition anyway. So it's better to just handle the error instead of checking first.

justinmk3 (approved):

packages/toolkit/src/codewhisperer/util/dependencyGraph/dependencyGraph.ts
nkomonen-amazon: Good point, ended up having to update this PR to do this

packages/toolkit/src/srcShared/fs.ts
```diff
@@ -132,7 +132,15 @@ export class FileSystemCommon {
             return
         }
 
-        return fs.delete(path, { recursive: true })
+        try {
+            await fs.delete(path, { recursive: true })
+        } catch (e) {
+            // The latest vscode version will never throw this error, but the current min version can
+            if (e instanceof vscode.FileSystemError && vscode.FileSystemError.FileNotFound().code === e.code) {
```
justinmk3, line 139: the `instanceof` check may be counterproductive. if the code looks like a "not found" error, it doesn't really matter what its type is, we want to skip it.
```suggestion
            if (vscode.FileSystemError.FileNotFound().code === e.code) {
```

packages/toolkit/src/test/srcShared/fs.test.ts
```diff
@@ -6,13 +6,14 @@
...
 import * as path from 'path'
-import { existsSync, mkdirSync, promises as fsPromises, readFileSync, rmSync, writeFileSync } from 'fs'
+import { existsSync, mkdirSync, promises as fsPromises, readFileSync, rmSync } from 'fs'
 import { FakeExtensionContext } from '../fakeExtensionContext'
 import { fsCommon } from '../../srcShared/fs'
 import * as os from 'os'
 import { isMinimumVersion } from '../../shared/vscode/env'
 import Sinon from 'sinon'
 import * as extensionUtilities from '../../shared/extensionUtilities'
+import { outputFile } from 'fs-extra'
```
justinmk3, line 16: can avoid this. we have `toFile` already. https://github.com/aws/aws-toolkit-vscode/blob/66cc926ab7bae2397f64396bb59720f6f38f877d/packages/toolkit/src/test/testUtil.ts#L26

nkomonen-amazon: Thanks for pointing out. It will require a good amount of changes to add a `mode` option.
But I've updated `makeFile()` to now use this function + `chmod` and gotten rid of the `fs-extra` import.

andrewyuq (approved):

PR #4444, opened 2024-02-16
web: Active CodeWhisperer in Web Mode

nkomonen-amazon:
With this PR you can now:
- Start up the Chromium test window with the toolkit installed
- Connect to auth + use CodeWhisperer code suggestions

Note errors may still pop up, but these are being worked on.

## License

### DIFF
--- a/packages/toolkit/src/extensionWeb.ts
+++ b/packages/toolkit/src/extensionWeb.ts
@@ -8,6 +8,7 @@ import { setInBrowser } from './common/browserUtils'
 import { getLogger } from './shared/logger'
 import { activateShared, deactivateShared } from './extensionShared'
 import { RegionProvider, defaultRegion } from './shared/regions/regionProvider'
+import os from 'os'
 
 export async function activate(context: vscode.ExtensionContext) {
     setInBrowser(true) // THIS MUST ALWAYS BE FIRST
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/extensionWeb.ts
```diff
@@ -34,6 +37,14 @@ export async function activate(context: vscode.ExtensionContext) {
     }
 }
 
+/**
+ * The browserfied version of os does not have a `version()` method,
+ * so we patch it.
+ */
+function patchOsVersion() {
+    ;(os.version as any) = () => '1.0.0'
```
justinmk3, line 45: I assume this is for library code. Is there a particular dependency that fails if this doesn't exist? Does the browserified "os" return something actually meaningful for the other fields?

packages/webpack.browser.config.js
```diff
@@ -47,6 +47,7 @@ const webConfig = {
             fs: false,
             crypto: require.resolve('crypto-browserify'),
             'fs-extra': false,
+            perf_hooks: false, // should be using globalThis.performance instead
```
justinmk3, line 50: 👍

justinmk3 (approved):

packages/webpack.browser.config.js
nkomonen-amazon: Yup created a ticket to start linting these dependencies

packages/toolkit/src/extensionWeb.ts
justinmk3: > `shouldTriggerFromClassifier()` using `os.version()`

That's code that we own, so perhaps we should abstract `os` and use the abstraction from there, instead of shimming. That will avoid surprising situations where `os.version` is used to make a wrong decision.

PR #4451, opened 2024-02-19
Add project details telemetry

damntrecky:
## Problem

1. Start transforming our metrics
2. Add project details to metrics to all users

## License


justinmk3:
always use `npm install --save-dev ...` to install deps (or `--save` for non-dev deps).

CI is failing:

```
npm ERR! `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm ERR! 
npm ERR! Invalid: lock file's @aws-toolkits/telemetry@1.0.186 does not satisfy @aws-toolkits/telemetry@1.0.189
```

packages/toolkit/src/amazonqGumby/activation.ts
```diff
@@ -13,14 +13,43 @@ import { transformByQState } from '../codewhisperer/models/model'
...
+            // Only fire closed during running/active job status
+            if (transformByQState.isRunning()) {
+                telemetry.codeTransform_jobIsClosedDuringIdeRun.emit({
+                    codeTransformJobId: transformByQState.getJobId(),
+                    codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+                    codeTransformStatus: transformByQState.getStatus(),
+                })
+            }
+        } else {
+            telemetry.codeTransform_jobIsResumedAfterIdeClose.emit({
```
justinmk3, line 45: gumby's activate() is always called, which means this will always emit even for users that have never used gumby. does that make sense?

packages/toolkit/src/amazonqGumby/activation.ts
```diff
@@ -13,14 +13,43 @@ import { transformByQState } from '../codewhisperer/models/model'
...
+import { getOpenProjects, validateOpenProjects } from '../codewhisperer/service/transformByQHandler'
+import { MetadataResult } from '../shared/telemetry/telemetryClient'
 
 export async function activate(context: ExtContext) {
     const transformationHubViewProvider = new TransformationHubViewProvider()
     new ProposedTransformationExplorer(context.extensionContext)
 
+    // Register an activation event listener to determine when the IDE opens, closes or users
+    // select to open a new workspace
+    const workspaceChangeEvent = vscode.workspace.onDidChangeWorkspaceFolders(event => {
```
justinmk3, line 32: Is there a way to avoid setting this up if gumby has not been used? (Same general question for all gumby features, but that's a backlog item...)

packages/toolkit/src/amazonqGumby/activation.ts
```diff
@@ -54,29 +83,28 @@ export async function activate(context: ExtContext) {
...
+        const validProjects = await validateOpenProjects(openProjects)
+        const firstKey = validProjects.keys().next().value
+        const firstModuleEntry = validProjects.get(firstKey)
+        telemetry.codeTransform_projectDetails.emit({
+            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+            codeTransformLocalJavaVersion:
+                (JDKToTelemetryValue(firstModuleEntry) as CodeTransformJavaSourceVersionsAllowed) || undefined,
+        })
+    } catch (err: any) {
+        telemetry.codeTransform_projectDetails.emit({
```
justinmk3, line 103: This will always be emitted for every user. Why?

damntrecky: This is a P0 requirement from leadership.

packages/toolkit/src/amazonqGumby/activation.ts
damntrecky: Maybe we can reduce the logic here to fire less. To just codewhisperer users only or some form of eligible users.

dhasani23: Won't this fire when a user opens a new folder or removes one?

dhasani23: Basically I don't know that we even need this `else` block

packages/toolkit/src/amazonqGumby/activation.ts
```diff
@@ -54,29 +83,28 @@ export async function activate(context: ExtContext) {
...
+            codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+            codeTransformPreValidationError: err?.name || undefined,
```
dhasani23, line 105: Maybe hardcode a value for this field so that it's never undefined?

dhasani23: Same with `reason` below and in `startTransformByQ.ts`

packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -175,36 +179,41 @@ async function getProjectsValidToTransform(mavenJavaProjects: vscode.QuickPickIt
  * As long as the project contains a .java file and a pom.xml file, the project is still considered valid for transformation,
  * and we allow the user to specify the Java version.
  */
-export async function validateOpenProjects(projects: vscode.QuickPickItem[]) {
+export async function validateOpenProjects(projects: vscode.QuickPickItem[], projectOnLoad = false) {
```
dhasani23, line 182: I don't totally understand the purpose of `projectOnLoad` and is it supposed to always be `false`?

packages/toolkit/src/amazonqGumby/activation.ts
justinmk3: > Maybe we can reduce this to just be Codewhisperer users, or eligible users of Gumby.

Does the `transformByQState` global have something that tells you "this user has touched gumby"? Most users don't use gumby, so it is not correct to emit gumby metrics for them (that makes telemetry very noise, which makes dashboards harder to reason about).

packages/toolkit/src/amazonqGumby/activation.ts
dhasani23: We will already have the info we need if the user actually clicks on "Transform", so what we are more so looking for is info on the users who do not click Transform (specifically, do they have a project valid for transformation open in their workspace?). One thing we could do to somewhat narrow things down is at least add this condition (https://github.com/aws/aws-toolkit-vscode/blob/5faec4ac40181605e6e8611e2a381310265873f7/packages/toolkit/src/amazonq/explorer/amazonQNode.ts#L87) before emitting the metric since that tells us the "Transform" button is visible to the user.

dhasani23: If we really wanted to restrict metric emission to just users who have clicked "Transform", we can simply emit a metric here if `validProjects.length > 0`:
https://github.com/aws/aws-toolkit-vscode/blob/5faec4ac40181605e6e8611e2a381310265873f7/packages/toolkit/src/codewhisperer/commands/startTransformByQ.ts#L409

packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
damntrecky: Its a default optional param value of false. Its intended for onload we are triggering a validation check for reporting.

damntrecky: The validation should be silent and NOT throw normal errors/toast notifications.

packages/toolkit/src/amazonqGumby/activation.ts
```diff
@@ -13,70 +13,101 @@ import { transformByQState } from '../codewhisperer/models/model'
...
+        // Try to validate project silently
+        try {
+            const openProjects = await getOpenProjects()
+            const validProjects = await validateOpenProjects(openProjects)
+            const firstKey = validProjects.keys().next().value
+            const firstModuleEntry = validProjects.get(firstKey)
+            telemetry.codeTransform_projectDetails.emit({
+                codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+                codeTransformLocalJavaVersion:
+                    (JDKToTelemetryValue(firstModuleEntry) as CodeTransformJavaSourceVersionsAllowed) || undefined,
```
dhasani23, line 102: I think you want `??` instead of `||`

packages/toolkit/src/amazonqGumby/activation.ts
```diff
@@ -13,70 +13,101 @@ import { transformByQState } from '../codewhisperer/models/model'
...
+            })
+        } catch (err: any) {
+            telemetry.codeTransform_projectDetails.emit({
                 codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
-                codeTransformStatus: transformByQState.getStatus(),
+                codeTransformPreValidationError: err?.name || undefined,
+                result: MetadataResult.Fail,
+                reason: err?.code || undefined,
```
dhasani23, line 109: Same here we should use `??`

packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
dhasani23: I get that and maybe I'm missing something, but do you ever set it to `true`?

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -17,14 +17,59 @@ import * as vscode from 'vscode'
...
+    // Try to validate project silently
+    let reason,
+        result,
+        codeTransformLocalJavaVersion,
+        codeTransformPreValidationError = undefined
+    try {
+        const openProjects = await getOpenProjects()
+        const validProjects = await validateOpenProjects(openProjects, true)
+        if (validProjects.size > 0) {
+            const firstKey = validProjects.keys().next().value
```
mk-fan, line 49: Add `result = MetadataResult.Pass`?

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -175,36 +224,41 @@ async function getProjectsValidToTransform(mavenJavaProjects: vscode.QuickPickIt
  * As long as the project contains a .java file and a pom.xml file, the project is still considered valid for transformation,
  * and we allow the user to specify the Java version.
  */
-export async function validateOpenProjects(projects: vscode.QuickPickItem[]) {
+export async function validateOpenProjects(projects: vscode.QuickPickItem[], onProjectFirstOpen = false) {
```
mk-fan, line 227: Is `validateOpenProjects` called anywhere else? Trying to understand why we only log error when `onProjectFirstOpen` is false, but the only call to `validateOpenProjects` set `onProjectFirstOpen` to true.

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -17,13 +17,59 @@ import * as vscode from 'vscode'
...
+            const firstKey = validProjects.keys().next().value
+            const firstModuleEntry = validProjects.get(firstKey)
+            codeTransformLocalJavaVersion = JDKToTelemetryValue(
+                firstModuleEntry
+            ) as CodeTransformJavaSourceVersionsAllowed
+        }
+    } catch (err: any) {
+        result = MetadataResult.Fail
+        reason = err?.code
+        codeTransformPreValidationError = err?.name
```
dhasani23, line 58: `.name` or `.message`?

packages/core/src/codewhisperer/service/transformByQHandler.ts
dhasani23: I'm also wondering this @damntrecky

mk-fan (approved):

dhasani23 (approved):

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -17,13 +17,59 @@ import * as vscode from 'vscode'
...
+                codeTransformLocalJavaVersion,
+                codeTransformPreValidationError,
+                result,
+                reason,
+            })
+        }
+    }
+}
+
+/* TODO: once supported in all browsers and past "experimental" mode, use Intl DurationFormat: */
```
justinmk3, line 72: comment looks out of place. same comment is on line 78

justinmk3 (approved):

damntrecky:
The CI failure seems to be unrelated to these changes. Here is another PR with the same CI issue -> https://github.com/aws/aws-toolkit-vscode/pull/4509

damntrecky:
I believe this PR fixes the failing CI -> https://github.com/aws/aws-toolkit-vscode/pull/4531/files

PR #4452, opened 2024-02-20
fix(CodeTransform): add check for zip size, show logs, get JAVA_HOME

dhasani23:
## Problem

- Want to show user build error logs when `Maven` commands fail. 
- Want to block upload if project size > 1GB. Want to help user find value of `JAVA_HOME` to increase chances of `Maven` commands succeeding. 
- Other minor UX fixes: show request ID in API error notifications, include link to FAQ doc in error notifications, and improve logging.

## Solution

Implement the above.

## License


packages/toolkit/src/amazonqFeatureDev/session/session.ts
```diff
@@ -51,7 +51,7 @@ export class Session {
             await this.setupConversation(msg)
             this.preloaderFinished = true
 
-            telemetry.amazonq_startChat.emit({
+            telemetry.amazonq_startConversationInvoke.emit({
```
dhasani23, line 54: See: https://github.com/aws/aws-toolkit-common/pull/693. Looks like another team removed the `amazonq_startChat` metric in favor of this one, but didn't update it here so I'm doing it to avoid a compilation error. I updated the telemetry package version because I made my own updates to our metrics.

packages/toolkit/src/test/amazonqFeatureDev/session/session.test.ts
```diff
@@ -26,8 +26,6 @@ describe('session', () => {
         it('emits start chat telemetry', async () => {
             await session.preloader('implement twosum in typescript')
 
-            assertTelemetry('amazonq_startChat', { amazonqConversationId: conversationID })
```
dhasani23, line 29: See https://github.com/aws/aws-toolkit-vscode/pull/4452/files#r1495272642

packages/toolkit/src/amazonqFeatureDev/session/session.ts
dianaabv16: blocker.
this metrics already emitted on L3.
You can fully delete this metric here as in this CR: 
https://github.com/aws/aws-toolkit-vscode-staging/pull/1578/files

dhasani23: Done

packages/toolkit/src/amazonqFeatureDev/session/session.ts
```diff
@@ -51,12 +51,6 @@ export class Session {
             await this.setupConversation(msg)
             this.preloaderFinished = true
 
-            telemetry.amazonq_startChat.emit({
```
damntrecky, line 54: Were we explicitly asked to remove this? The Q team might be using this metric in their dashboards.

dhasani23: Yes. See above. Was asked to remove it.

packages/toolkit/src/codewhisperer/models/constants.ts
```diff
@@ -316,32 +317,46 @@ export const viewProposedChangesMessage =
...
+export const windowsJavaHomeHelpMessage =
+    'In a terminal, try running "cd C:\\Program Files\\Java" and then "dir" to see a list of installed JDKs. Then "cd" into JDK JAVA_VERSION_HERE (if present) and run "cd" once more to get the JAVA_HOME. It should look something like "C:\\Program Files\\Java\\jdk-11.0.1"'
+
+export const nonWindowsJava8HomeHelpMessage =
+    'Try running "/usr/libexec/java_home -v 1.8" in a terminal to find this. It should look something like "/Library/Java/JavaVirtualMachines/jdk-8.jdk/Contents/Home"'
+
+export const nonWindowsJava11HomeHelpMessage =
+    'Try running "/usr/libexec/java_home -v 11" in a terminal to find this. It should look something like "/Library/Java/JavaVirtualMachines/jdk-11.jdk/Contents/Home"'
+
+export const projectSizeTooLargeMessage =
```
damntrecky, line 349: Are these approved text? It seems like this error message is vague. What limit? Whos limit?

Eg) `Your project size exceeds the Amazon Q Code Transform upload limit of 1GB`

packages/toolkit/src/codewhisperer/models/model.ts
```diff
@@ -267,6 +267,10 @@ export class TransformByQState {
     private projectName: string = ''
     private projectPath: string = ''
 
+    private totalSourceCodeBytes: number = 0
```
damntrecky, line 270: Is any of this code used in this PR, or just left over from another change?

packages/toolkit/src/codewhisperer/models/constants.ts
```diff
@@ -316,32 +317,46 @@ export const viewProposedChangesMessage =
...
+    'Your project size exceeds the limit of 1GB. Please see: https://docs.aws.amazon.com/amazonq/latest/aws-builder-use-ug/code-transformation.html'
 
 export const JDK8VersionNumber = '52'
 
 export const JDK11VersionNumber = '55'
 
 export const numMillisecondsPerSecond = 1000
 
+export const uploadZipSizeLimitInBytes = 1000000000 // 1GB
```
damntrecky, line 358: Is this variable used in this PR?

packages/toolkit/src/codewhisperer/models/constants.ts
dhasani23: Will change the text to match that error message you suggested. I think we can fine-tune the text later but there's other important work that shouldn't be blocked by perfecting the strings.

And yes it is used in `transformByQHandler.ts` -- you have to click 'View Diff' for that since it's not shown by default as it's too large.

packages/toolkit/src/codewhisperer/models/constants.ts
dhasani23: Used in `transformByQHandler.ts`

packages/toolkit/src/codewhisperer/models/model.ts
dhasani23: Update: removed some of this code so N/A now

damntrecky (approved):

PR #4456, opened 2024-02-20
feat(cwspr): Add new login UX as a library

leigaol:
## Problem

The current login UX is not ideal. 

## Solution

Build a new login UX shared between toolkit, CodeWhisperer and Amazon Q. 

This PR only adds a library to existing code base, it is not used in production code path and it has no functional impact on the build artifact. It has 0 deletions to original code. 

## License


packages/toolkit/src/webviews/main.ts
```diff
@@ -216,6 +217,13 @@ export abstract class VueWebview {
                 return this.instance
             }
 
+            public async setup(webview: vscode.Webview) {
```
hayemaxi, line 220: Why call this in advance? Code in this function looks like its handled in the following show() function

packages/toolkit/src/login/webview/vue/backend.ts
```diff
@@ -0,0 +1,166 @@
...
+        } else {
+            return this.ssoSetup('createIdentityCenterConnection', async () => {
+                const ssoProfile = createSsoProfile(startUrl, region)
+                const conn = await Auth.instance.createConnection(ssoProfile)
+                await Auth.instance.useConnection(conn)
+                void vscode.window.showInformationMessage('Toolkit: Successfully connected to AWS IAM Identity Center')
+                void this.showResourceExplorer()
+            })
+        }
+    }
```
hayemaxi, line 126: this class and especially these functions have hardcoded options that you can pass in. this will become messy and difficult to maintain if we ever add new functionality that requires its own auth.

for an initial prototype, this is fine. But we should think about how we can make this more dynamic later for additional usecases without a bunch of specialized if statements.

packages/toolkit/src/webviews/main.ts
leigaol: I cannot call the show() function because that function was specifically design to show a webview in a WebViewPanel as a custom editor on the right hand side. I need to build a function that only takes care of setting up the backend of the vue webview.

packages/toolkit/src/login/webview/vue/backend.ts
leigaol: I am aware of this. I have yet found an very effective way to make it very clean.

leigaol: This is fixed in latest commit

packages/toolkit/src/login/webview/vue/backend.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * This wraps the execution of the given setupFunc() and handles common errors from the SSO setup process.
+     *
+     * @param methodName A value that will help identify which high level function called this method.
+     * @param setupFunc The function which will be executed in a try/catch so that we can handle common errors.
+     * @returns
+     */
+    async ssoSetup(methodName: string, setupFunc: () => Promise<any>): Promise<AuthError | undefined> {
```
nkomonen-amazon, line 37: Can we make this static and have `show.ts` for the current auth page use this instead? It will help to have this in a single spot

leigaol: Can you elaborate? There is no `show.ts` in the new login code path. 

I can import this function from current auth page to avoid duplicating code.

packages/toolkit/src/login/webview/commonAuthViewProvider.ts
```diff
@@ -0,0 +1,114 @@
...
+            enableCommandUris: true,
+            localResourceRoots: [dist, resources],
+        }
+        webviewView.webview.html = this._getHtmlForWebview(this.extensionContext.extensionUri, webviewView.webview)
+        // register the webview server
+        await this.webView?.setup(webviewView.webview)
+    }
+
+    private _getHtmlForWebview(extensionURI: Uri, webview: vscode.Webview) {
+        const source = path.join('src', 'login', 'webview', 'vue', 'index.js')
```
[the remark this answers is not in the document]
leigaol: Having the source in VueWebview is a convention in this repository. It is used for `VueWebview.compilePanel`

Example

`
export class CodeCatalystConfigureWebview extends VueWebview {
    public readonly id = 'configureCodeCatalyst'
    public readonly source = 'src/codecatalyst/vue/configure/index.js'`
    
    
    
  The path join here is also used in AmazonQ
  
  `const source = path.join('src', 'amazonq', 'webview', 'ui', 'amazonq-ui.js')`
  
  Here the webviewProvider needs to get the absolute path of the source js file, while the VueWebView requires a relative path. Hence there are 2.

packages/toolkit/src/login/webview/vue/backend.ts
leigaol: I can further improve this in next PR.

nkomonen-amazon (approved):

packages/toolkit/src/login/webview/vue/backend_amazonq.ts
```diff
@@ -0,0 +1,51 @@
...
+import { SsoConnection } from '../../../auth/connection'
+import { AuthUtil } from '../../../codewhisperer/util/authUtil'
+import { AuthError, CommonAuthWebview } from './backend'
+import { awsIdSignIn } from '../../../codewhisperer/util/showSsoPrompt'
+import { connectToEnterpriseSso } from '../../../codewhisperer/util/getStartUrl'
+
+export class AmazonQLoginWebview extends CommonAuthWebview {
+    // TODO: update source, id
+    public override id: string = 'aws.AmazonCommonAuth'
+    public override source: string = 'src/login/webview/vue/index.js'
```
hayemaxi, line 15: do the child classes need to define the source? can we just pull from the abstract class?

hayemaxi (approved):

packages/toolkit/src/login/webview/vue/index.ts
justinmk3, line 1: `webview/vue/` seems redundant. Avoid hyper-nested directory structures. Can the files live in `webview/` or `vue/` instead of `webview/vue/`

PR #4458, opened 2024-02-21
rename: "browser" to "web"

nkomonen-amazon:
## Problem

We've decided to use the "web" term for the toolkit running in the browser instead of "browser"
https://code.visualstudio.com/api/extension-guides/web-extensions

## Solution
Replace uses of "browser" with "web"

## License


justinmk3 (approved):
Thank you!!

PR #4460, opened 2024-02-21
webpack: minor changes

nkomonen-amazon:
Minor webpack changes, see commit message for description of each change


package.json
```diff
@@ -44,29 +44,30 @@
...
         "pretty-quick": "^3.1.3",
-        "husky": "^9.0.7",
-        "typescript": "^5.0.4",
         "ts-node": "^10.9.1",
+        "typescript": "^5.0.4",
         "webpack": "^5.83.0",
         "webpack-cli": "^5.1.4",
-        "webpack-dev-server": "^4.15.1"
+        "webpack-dev-server": "^4.15.1",
+        "webpack-merge": "^5.10.0"
```
nkomonen-amazon, line 70: We just added this new package, the other changes are just things being automatically reordered.

packages/webpack.browser.config.js
```diff
@@ -10,19 +10,26 @@
...
-const webConfig = {
-    ...baseConfig,
+const webConfig = merge(baseConfig, {
     name: 'web',
     target: 'webworker',
+    /**
+     * We are explicitly using the following value due to breakpoints being offset during debugging w/ Chromium if we use `source-map`.
+     * From this thread, depending on the value, breakpoints can be buggy: https://github.com/webpack/webpack/issues/1487
+     *
+     * TODO: We do not want to use the following for Production, so look to add logic to switch depending on production vs development.
```
justinmk3, line 25: > We are explicitly using the following value 

what is the default value? Or why are we mentioning this?

> We do not want to use the following for Production

why?

packages/webpack.browser.config.js
```diff
@@ -10,19 +10,26 @@
...
 
 /** @type WebpackConfig */
-const webConfig = {
-    ...baseConfig,
+const webConfig = merge(baseConfig, {
     name: 'web',
     target: 'webworker',
+    /**
+     * We are explicitly using the following value due to breakpoints being offset during debugging w/ Chromium if we use `source-map`.
+     * From this thread, depending on the value, breakpoints can be buggy: https://github.com/webpack/webpack/issues/1487
```
justinmk3, line 23:
```suggestion
     * The choice of "devtool" can cause breakpoints can be buggy: https://github.com/webpack/webpack/issues/1487
```

packages/webpack.vue.config.js
```diff
@@ -37,16 +38,14 @@ const createVueEntries = (targetPattern = 'index.ts') => {
...
+const vueConfig = merge(baseConfig, {
     name: 'vue',
     target: 'web',
     output: {
-        ...baseConfig.output,
         libraryTarget: 'this',
     },
     module: {
-        rules: (baseConfig.module?.rules ?? []).concat(
+        rules: [
```
justinmk3, line 48: webpack-merge has its own semantics and "magic", so it's less obvious than the old code. If we suspect that the old `(baseConfig.module?.rules ?? []` approach could have subtle bugs then it's probably worth taking on a dependency.

justinmk3 (approved):
Nice choices

PR #4464, opened 2024-02-22
fix(appcomposer): add null checks to parsed JSON files

JLargent42:
## Problem

`js-yaml` can parse JSON, but a JSON string without braces (such as an empty JSON file) will return undefined. 

## Solution

This adds null checks so that we don't try to access members of a `null` or `undefined` parsed template.

## License

### DIFF
--- a/packages/toolkit/src/shared/cloudformation/cloudformation.ts
+++ b/packages/toolkit/src/shared/cloudformation/cloudformation.ts
@@ -439,9 +439,9 @@ export async function tryLoad(
 
     // Check if the template is a SAM template, using the same heuristic as the cfn-lint team:
     // https://github.com/aws-cloudformation/aws-cfn-lint-visual-studio-code/blob/629de0bac4f36cfc6534e409a6f6766a2240992f/client/src/yaml-support/yaml-schema.ts#L39-L51
-    if (rv.template.AWSTemplateFormatVersion || rv.template.Resources) {
+    if (rv.template?.AWSTemplateFormatVersion || rv.template?.Resources) {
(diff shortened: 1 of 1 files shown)


justinmk3:
> `js-yaml` can parse JSON, but a JSON string without braces (such as an empty JSON file) will return undefined.

Thanks for mentioning that, that's useful to know.

packages/toolkit/src/shared/cloudformation/cloudformation.ts
```diff
@@ -439,9 +439,9 @@ export async function tryLoad(
 
     // Check if the template is a SAM template, using the same heuristic as the cfn-lint team:
     // https://github.com/aws-cloudformation/aws-cfn-lint-visual-studio-code/blob/629de0bac4f36cfc6534e409a6f6766a2240992f/client/src/yaml-support/yaml-schema.ts#L39-L51
-    if (rv.template.AWSTemplateFormatVersion || rv.template.Resources) {
+    if (rv.template?.AWSTemplateFormatVersion || rv.template?.Resources) {
         rv.kind =
-            rv.template.Transform && rv.template.Transform.toString().startsWith('AWS::Serverless') ? 'sam' : 'cfn'
+            rv.template?.Transform && rv.template.Transform.toString().startsWith('AWS::Serverless') ? 'sam' : 'cfn'
```
justinmk3, line 444: not needed since the condition already checks it

JLargent42:
Duplicate of https://github.com/aws/aws-toolkit-vscode/pull/4465

PR #4467, opened 2024-02-23
web: various fixes/improvements

nkomonen-amazon:
This PR:
- removes some uses of `fs-extra`
- Gets the region provider working in web. This enables the user to sign in w/ IdC and be able to select a region listed in  our `endpoints.json` file.
- Fixes an issue w/ webpack + web extension in chrome + breakpoints.

See each commit for specifics.

## License

### DIFF
--- a/packages/toolkit/src/extensionShared.ts
+++ b/packages/toolkit/src/extensionShared.ts
@@ -14,7 +14,7 @@ import * as nls from 'vscode-nls'
 import globals, { initialize } from './shared/extensionGlobals'
 import { join } from 'path'
 import { Commands } from './shared/vscode/commands2'
-import { documentationUrl, githubCreateIssueUrl, githubUrl } from './shared/constants'
+import { documentationUrl, endpointsFileUrl, githubCreateIssueUrl, githubUrl } from './shared/constants'
 import { getIdeProperties, aboutToolkit, isCloud9 } from './shared/extensionUtilities'
 import { telemetry } from './shared/telemetry/telemetry'
 import { openUrl } from './shared/utilities/vsCodeUtils'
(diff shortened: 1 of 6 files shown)


packages/toolkit/src/extensionShared.ts
```diff
@@ -238,6 +234,36 @@ function logAndShowWebviewError(err: unknown, webviewId: string, command: string
...
+ */
+function makeEndpointsProvider() {
+    let localManifestFetcher: ResourceFetcher
+    let remoteManifestFetcher: ResourceFetcher
+    if (isWeb()) {
+        // In web mode everything must be in a single file, so things like the endpoints file will not be available.
+        // The following imports the endpoints file, which causes webpack to bundle it in the final output file
+        const endpoints = require('../resources/endpoints.json')
+        localManifestFetcher = { get: async () => JSON.stringify(endpoints) }
+        // Cannot use HttpResourceFetcher due to web mode breaking on import
```
justinmk3, line 250: we plan to fix that, right?
```suggestion
        // TODO: HttpResourceFetcher is currently broken in web mode.
```

packages/webpack.web.config.js
```diff
@@ -44,6 +36,22 @@ const webConfig = merge(baseConfig, {
             NODE_DEBUG: 'development',
             READABLE_STREAM: 'disable',
         }),
+        /** 
+         * HACK: the http resource fetcher breaks Web mode if imported, BUT we still dynamically import this module for non web mode
```
justinmk3, line 40: why not use the exact name?

```suggestion
         * HACK: HttpResourceFetcher breaks Web mode if imported, BUT we still dynamically import this module for non web mode
```

justinmk3 (approved):

packages/webpack.web.config.js
nkomonen-amazon: Good point, updated

PR #4475, opened 2024-02-27
refactor: separate toolkit into core library

hayemaxi:
In order to share code between multiple extensions we need to create a core library. Rather than create a core lib package, we decided to rename the current toolkit as a core library and create a barebones replacement that will generate the extension by importing from the core lib. This is easier than searching for common components, unwiring them, and hoisting them into an empty core lib package.

Summary of changes:
- previous `packages/toolkit` has been renamed to packages/core. The core lib npm name is `aws-core-vscode`.
- current `packages/toolkit` is now a barebones extension that imports all of the code from the core lib. It still functions the same.
- Extension tests run from the core lib. Since some of the tests require an extension context/sandbox, we initiate a "fake" extension to run these tests. This is also why there are vscode extension properties in the package.json
- Extension runs from the toolkit/
- Small changes to some of the lib code to export certain types, make fields truly private (with `#`), so that the package can be imported. **Anonymous classes do not support private/protected types.**
- Update `moduleResolution` to `node16` for tsconfigs. This is required to import from the library, otherwise there is an error in VSCode.
- Webpack updates/generalizations + updates to some of the scripts.

This approach is not without some disadvantages. The biggest "hack" with this change comes from that fact that toolkit code wasn't built to be a library. So, some of the code depends on the package.json (especially the contributes section). Since this section is very large AND needs to be present in both the core lib and toolkit extension package.jsons, there is a build script to copy necessary fields over to the toolkit extension. In the future, we should implement a simpler way of inheriting or removing the need for both packages to depend on this data.

Remaining work for this PR:
- documentation

## License


packages/core/scripts/build/copyFiles.ts
```diff
@@ -0,0 +1,71 @@
+/*!
```
hayemaxi, line 1: will generalize this script in the future.

package-lock.json
```diff
@@ -18903,6 +18896,25 @@
...
+                "npm": "^10.1.0",
+                "vscode": "^1.68.0"
+            }
+        },
+        "packages/toolkit": {
+            "name": "aws-toolkit-vscode",
+            "version": "2.12.0-SNAPSHOT",
+            "license": "Apache-2.0",
+            "devDependencies": {
+                "aws-core-vscode": "file:../core/"
```
justinmk3, line 18915: Should it be in regular "dependencies" instead of "devDependencies" ?

package-lock.json
```diff
@@ -18903,6 +18896,25 @@
                 "assert": "^2.1.0"
             }
         },
+        "packages/q": {
```
justinmk3, line 18899: (not a blocker) I assume we can rename this to packages/amazonq in a future PR

packages/core/.vscodeignore
justinmk3, line 1: (not a blocker) .vscodeignore is probably a file that can live at the top level

justinmk3 (approved):
Fantastic! Nicely explained.

package-lock.json
hayemaxi: yes 👍

package-lock.json
hayemaxi: seems to be a leftover artifact, nice catch. It is already amazonq/

packages/core/scripts/test/launchTestUtilities.ts
```diff
@@ -122,7 +123,14 @@ async function setupVSCodeTestInstance(suite: SuiteName): Promise<string> {
 
     console.log(`Setting up VS Code test instance, version: ${vsCodeVersion}`)
     const platform = process.platform === 'win32' ? 'win32-x64-archive' : undefined
-    const vsCodeExecutablePath = await downloadAndUnzipVSCode(vsCodeVersion, platform)
+
+    const downloadOptions = {
+        platform,
+        version: vsCodeVersion,
+        cachePath: process.env.AWS_TOOLKIT_TEST_CACHE_DIR ?? defaultCachePath,
```
justinmk3, line 130: Nice, would be great if this works out.

leigaol:
Great!

Have you figured out how to keep 2 separate change logs for 2 extensions?

hayemaxi:
> Great!
> 
> Have you figured out how to keep 2 separate change logs for 2 extensions?

Will tackle this problem after we have two functioning extensions- gonna be part of a wider array of infra changes to accommodate multiple extensions.

PR #4477, opened 2024-02-27
fix(CodeTransform): show link to docs after errors

dhasani23:
## Problem

Wanted to clean up our logging / messages, include link to docs when error occurs, and include request ID in error notifications when APIs fail.

## Solution

Implement the above.

## License

### DIFF
--- a/packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
+++ b/packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
@@ -14,7 +14,7 @@ import * as fs from 'fs-extra'
 import * as path from 'path'
 import * as os from 'os'
 import * as vscode from 'vscode'
-import { spawnSync } from 'child_process' // TO-DO: consider using ChildProcess once we finalize all spawnSync calls
+import { spawnSync } from 'child_process' // Consider using ChildProcess once we finalize all spawnSync calls
 import AdmZip from 'adm-zip'
 import globals from '../../shared/extensionGlobals'
 import { CodeTransformMavenBuildCommand, telemetry } from '../../shared/telemetry/telemetry'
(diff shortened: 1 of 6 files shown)


packages/toolkit/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -24,7 +24,7 @@ import { calculateTotalLatency, javapOutputToTelemetryValue } from '../../amazon
 import { MetadataResult } from '../../shared/telemetry/telemetryClient'
 import request from '../../common/request'
 
-/* TODO: once supported in all browsers and past "experimental" mode, use Intl DurationFormat:
+/* Once supported in all browsers and past "experimental" mode, use Intl DurationFormat:
```
dhasani23, line 27: Removed `TODO` but keeping the content of the comment. Got feedback from a security engineer that apparently having `TODO`s scattered is not good.

nkomonen-amazon: Did they say why TODOs were bad?

dhasani23: Apparently customers looking through the code? Makes it seem incomplete. I disagreed but that's what I was told. I'm not going to remove the comments themselves so that I can still see them.

justinmk3: I suggest pushing back on that.

dhasani23: I'll do that. I think his suggestion was to instead create tickets for these items but I prefer a TO-DO over a ticket as it's easier for me to track / remember.

dhasani23: In case he doesn't get back to me soon enough, here's a ticket to track this: https://sim.amazon.com/issues/EGCP-97

damntrecky (approved):

PR #4478, opened 2024-02-28
feat(cwspr): Improve percentage code written metrics to include some user multi character input

leigaol:
## Problem

Previously when we count user written code we only count user keystroke input, which contributes to a higher than expected % code written metric.

However, if we include all multi character user inputs from copy & paste or other sources, the % code written metric will be very low, providing little value to customers.

We discovered that if a user copy a large file, it will generate huge total user input, which is not what we want to track. Instead, small user multi character input are indeed considered as user written code. For example, copy code snippet from Stackoverflow, copy another function signature, accepting IntelliSense suggestion, etc.

## Solution

1. Add multi char user input less than 500 as total user written code. We believe 500 is a good middle ground to begin with. 
2. Add temporary telemetry field to also report the user multi char input for different thresholds, we will use this data to adjust the cut off threshold. (these code is meant to be removed in the future once we finalize the threshold)

This one-off data collection is VS Code only, once the threshold is finalized, JetBrains toolkit will adopt same threshold. Therefore the telemetry changes are checked in to the override.json

## License

### DIFF
--- a/.changes/next-release/Feature-742cb7ff-2139-4c0f-9b9e-22d72c960f60.json
+++ b/.changes/next-release/Feature-742cb7ff-2139-4c0f-9b9e-22d72c960f60.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Improve percentage code written metrics to include user multi character input"
+}
--- a/packages/toolkit/src/codewhisperer/service/inlineCompletionItemProvider.ts
+++ b/packages/toolkit/src/codewhisperer/service/inlineCompletionItemProvider.ts
@@ -104,6 +104,7 @@ export class CWInlineCompletionItemProvider implements vscode.InlineCompletionIt
             }
             return undefined
         }
+        TelemetryHelper.instance.lastSuggestionInDisplay = truncatedSuggestion
         return {
             insertText: truncatedSuggestion,
             range: new vscode.Range(start, end),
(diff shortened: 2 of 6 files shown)


.changes/next-release/Feature-742cb7ff-2139-4c0f-9b9e-22d72c960f60.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Improve percentage code written metrics to include user multi character input"
```
justinmk3, line 3:
```suggestion
	"description": "CodeWhisperer: Improve percentage code written metrics to include user multi character input"
```

andrewyuq (approved):

justinmk3:
Test is failing on Windows:

```
  1 failing
  1) codewhispererCodecoverageTracker
       countTotalTokens
         Should not skip when content change size is less than 500:

undefined !== 300

      at Context.<anonymous> (d:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\toolkit\src\test\codewhisperer\tracker\codewhispererCodeCoverageTracker.test.ts:270:20)
      at Context.fn (d:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\toolkit\src\test\setupUtil.ts:33:24)
```

justinmk3:
@leigaol Please squash-merge if the individual commits are not meaningful (as in this PR).

PR #4479, opened 2024-02-28
feat(toolkit): Switch aws toolkit to new auth flow under feature flag

leigaol:
## Problem

This PR is part of task "moving aws toolkit to new login UX". 

## Solution

This is a PR that utilizes the new login library in https://github.com/aws/aws-toolkit-vscode/pull/4456 in aws toolkit plugin.  There is a feature flag in DevSettings that controls whether the new login flow is enabled or not. When the feature flag is turned off, the extension falls back to existing login flow. 

## License


packages/toolkit/src/auth/ui/vue/show.ts
```diff
@@ -744,6 +745,12 @@ export const showManageConnections = Commands.declare(
         if (!isServiceItemId(serviceToShow)) {
             serviceToShow = undefined
         }
+
+        if (DevSettings.instance.isNewLoginEnabled()) {
+            void vscode.commands.executeCommand('setContext', 'aws.explorer.showAuthView', true)
```
justinmk3, line 750: why not `await`? Then purpose of that lint rule is not to ignore it, but to avoid race conditions.

packages/toolkit/src/dev/activation.ts
```diff
@@ -152,6 +153,18 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
     if (!isCloud9() && !isReleaseVersion() && config.betaUrl) {
         ctx.subscriptions.push(watchBetaVSIX(config.betaUrl))
     }
+
+    if (DevSettings.instance.isNewLoginEnabled()) {
+        void vscode.commands.executeCommand('setContext', 'aws.explorer.newLoginEnabled', true)
```
justinmk3, line 158: Would be cool if DevSettings automatically set `aws.dev.xx` by default (for all "dev settings"), then we wouldn't need this.

packages/toolkit/src/auth/ui/vue/show.ts
leigaol: Fixed in newer commits

packages/toolkit/src/dev/activation.ts
leigaol: I did not find an easy way to do this. My next priority is to understand how auth can be shared across extensions. I can get back to this later.

PR #4482, opened 2024-02-28
feat(appcomposer): load and save from VS Code buffer

JLargent42:
## Problem
- Application Composer works directly off of the workspace files. Changes in VS Code will only be detected by App Composer when the user saves, and changes in Composer will immediately be saved to disk. This does not follow the standard VS Code pattern.

## Solution
- This updates Application Composer to read from and write to VS Code text documents when possible, ensuring that Composer is synced with the latest user changes and giving the user more control over what changes are saved.

## License


ThaiuLinh (approved):

packages/toolkit/package.json
```diff
@@ -4352,6 +4352,7 @@
         "bytes": "^3.1.2",
         "cross-fetch": "^4.0.0",
         "cross-spawn": "^7.0.3",
+        "fast-deep-equal": "^3.1.3",
```
justinmk3, line 4355: Why is this needed?

packages/toolkit/src/applicationcomposer/utils/templateShouldBeUpdated.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { loadByContents } from '../../shared/cloudformation/cloudformation'
+import equal from 'fast-deep-equal/es6'
+
+/**
+ * Checks whether a template needs to be updated. This is eiter when the template is out of sync, or
+ * when at least one of the templates cannot be parsed. Comments and whitespace should not result
+ * in a template update.
+ */
+export async function templateShouldBeUpdated(oldTemplate: string, newTemplate: string) {
```
justinmk3, line 14: Do not create one-function-per file. Use modules to group related functionality.

packages/toolkit/src/applicationcomposer/messageHandlers/addFileWatchMessageHandler.ts
```diff
@@ -13,31 +13,47 @@ import {
...
-                    fileContents: fileContents,
-                }
-
-                await context.panel.webview.postMessage(fileChangedMessage)
-                context.fileWatches[filePath] = { fileContents: fileContents }
+        vscode.workspace.onDidChangeTextDocument(async event => {
+            if (event.document.fileName !== context.textDocument.fileName || event.contentChanges.length === 0) {
+                return
             }
+            bufferFileChange(event, context)
```
justinmk3, line 56: Avoid jargon-forking. vscode calls it a textDocument so let's be consistent, since we already have many functions with that name.
```suggestion
            textDocumentChange(event, context)
```

packages/toolkit/package.json
justinmk3: > and is only 13kB unpacked

All dependencies must be maintained, and that becomes a big problem over time. :) 13kB is a lot of code which may have bugs and CVEs...

PR #4484, opened 2024-02-29
AmazonQ: Result field to chat telemetry events

ppiliaev:
## Problem

Chat events verification fails

## Solution

Result field is added to chat events to prevent verification errors

## License


hayemaxi (approved):

justinmk3 (approved):
This is fine, but ideally you would record when some of these operations fail. E.g. if an error occurs when the user clicks a "followup", presumably you would want to emit `amazonq_interactWithMessage` with `result=Failed`. Currently, you won't have visibility into such failures.

PR #4494, opened 2024-03-01
added wsl computeEnv

dhruvigajjar:
## Problem
adding wsl as computeEnv

### DIFF
--- a/packages/core/src/shared/telemetry/util.ts
+++ b/packages/core/src/shared/telemetry/util.ts
@@ -92,7 +92,7 @@ export async function getUserAgent(
     return pairs.join(' ')
 }
 
-type EnvType = 'cloud9' | 'cloud9-codecatalyst' | 'codecatalyst' | 'local' | 'ec2' | 'sagemaker' | 'test' | 'unknown'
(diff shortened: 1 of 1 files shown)


packages/core/src/shared/telemetry/util.ts
```diff
@@ -92,7 +92,7 @@ export async function getUserAgent(
     return pairs.join(' ')
 }
 
-type EnvType = 'cloud9' | 'cloud9-codecatalyst' | 'codecatalyst' | 'local' | 'ec2' | 'sagemaker' | 'test' | 'unknown'
+type EnvType = 'cloud9' | 'cloud9-codecatalyst' | 'codecatalyst' | 'local' | 'ec2' | 'sagemaker' | 'test' | 'wsl' |'unknown'
```
nkomonen-amazon, line 95: you'll want to remove the trailing space. I think this will cause linting to fail

nkomonen-amazon (approved):

packages/core/src/shared/telemetry/util.ts
hayemaxi: shouldnt formatting remove this automatically?

PR #4496, opened 2024-03-01
fix(CodeTransform): Updating commands for copying dependencies

sagdutt:
## Problem
The current mechanism to copy dependencies is faulty and often results in cases where dependencies are only partially copied or the command execution fails, both of which results in users being unable to transform their code.

## Solution
Using `mvn install` as the primary mechanism to copy dependencies. The order of execution is as follows:
1. Run `mvn dependency:copy-dependencies` to copy any locally persisted dependencies into an empty dependencies directory. We will continue execution even if this command fails as this is just an optimization to avoid downloading any locally available dependencies which can be copied with this command.
2. Run` mvn clean install` with the `-Dmaven.repo.local` arg pointing to the dependencies directory created in step 1. This is the primary command we are going to depend on to gather dependencies. If this command fails, fail the submission.

## Testing
Created a debug build with `npm run package -- --debug`. Used this to verify by running a test job that dependencies were generated and uploaded.

Job screenshot

## License

### DIFF
--- a/packages/core/src/codewhisperer/service/transformByQHandler.ts
+++ b/packages/core/src/codewhisperer/service/transformByQHandler.ts
@@ -24,6 +24,11 @@ import { calculateTotalLatency, javapOutputToTelemetryValue } from '../../amazon
 import { MetadataResult } from '../../shared/telemetry/telemetryClient'
 import request from '../../common/request'
 
+interface FolderInfo {
+    path: string
+    name: string
+}
+
 /* Once supported in all browsers and past "experimental" mode, use Intl DurationFormat:
  * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat#browser_compatibility
  * Current functionality: given number of milliseconds elapsed (ex. 4,500,000) return hr / min / sec it represents (ex. 1 hr 15 min)
@@ -383,7 +388,7 @@ function getFilesRecursively(dir: string, isDependenciesFolder: boolean): string
 }
 
 // run 'install' with either 'mvnw.cmd', './mvnw', or 'mvn' (if wrapper exists, we use that, otherwise we use regular 'mvn')
-function installProjectDependencies() {
+function installProjectDependencies(dependenciesFolder: FolderInfo) {
     // baseCommand will be one of: '.\mvnw.cmd', './mvnw', 'mvn'
     const baseCommand = transformByQState.getMavenName()
     const modulePath = transformByQState.getProjectPath()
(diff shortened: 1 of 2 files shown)


packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -383,15 +388,15 @@ function getFilesRecursively(dir: string, isDependenciesFolder: boolean): string
...
-function installProjectDependencies() {
+function installProjectDependencies(dependenciesFolder: FolderInfo) {
     // baseCommand will be one of: '.\mvnw.cmd', './mvnw', 'mvn'
     const baseCommand = transformByQState.getMavenName()
     const modulePath = transformByQState.getProjectPath()
 
     transformByQState.appendToErrorLog(`Running command ${baseCommand} clean install`)
 
     // Note: IntelliJ runs 'clean' separately from 'install'. Evaluate benefits (if any) of this.
-    const args = ['clean', 'install', '-q']
```
dhasani23, line 394: I've found `-q` to be sufficient in terms of giving the user enough info to debug. My only concern with removing it is that I've noticed (it's pretty rare), the buffer that `spawnSync` stores `stdout` in can overflow, so if we really must remove `-q` that's going to make `stdout` larger and so we should override the `maxBufferSize` option in the `spawnSync` call (default is 1024*1024 bytes): https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options

sagdutt: I see, I think for vscode I can add the flag back for now because we don't really show the log to the user. 
For intellij, I think removing it makes sense because the logs can represent a sense of progression for the user because `maven install` can take a quite some time to complete depending on the app. We could consider a similar experience for vscode in the future and we could revisit this then.

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -531,35 +539,27 @@ export async function zipCode() {
...
+        copyProjectDependencies(dependenciesFolder)
     } catch (err) {
-        void vscode.window.showErrorMessage(CodeWhispererConstants.installErrorMessage)
-        // open build-logs.txt file to show user error logs
-        const logFilePath = await writeLogs()
-        const doc = await vscode.workspace.openTextDocument(logFilePath)
-        await vscode.window.showTextDocument(doc)
-        throw err
+        // continue in case of errors
     }
```
arangatang, line 548: We probably want to emit telemetry also here? That would allow us to catch issues with this approach.

arangatang: On another note, do we not have any preexisting tests for zipping?

sagdutt: Telemetry is already being emitted inside the `copyProjectDependencies` method on line 508

dhasani23 (approved):

hayemaxi (approved):

PR #4507, opened 2024-03-06
fix: Temporarily disable amazon q tests

jpinkney-aws:
## Problem
- The test infra isn't fully setup so these tests will always fail

## Solution
- Wait until the test infra is fully setup again before unskipping these tests

## License


hayemaxi (approved):
Should we add a reminder here? https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/test/techdebt.test.ts

PR #4508, opened 2024-03-07
feat(cwspr): create 2 vue app instance from one root.vue with parameters

leigaol:
## Problem
Previously the login UX `app` option is hard-coded.  It is assigned in root.vue, which makes not suitable to be used as a library for 2 extensions. 

## Solution
Make the app name a parameter in root.vue.  
Create 2 vue apps from root.vue using different parameters.

For consumer extensions,  just create a `CommonAuthViewProvider` with app name and subscribes it.

`const authProvider = new CommonAuthViewProvider(context, undefined, 'TOOLKIT')`

## License


packages/core/src/login/webview/commonAuthViewProvider.ts
```diff
@@ -59,7 +59,15 @@ export class CommonAuthViewProvider implements WebviewViewProvider {
             app === 'TOOLKIT'
                 ? VueWebview.compilePanel(ToolkitLoginWebview)
                 : VueWebview.compilePanel(AmazonQLoginWebview)
-        this.source = path.join('vue', 'src', 'login', 'webview', 'vue', 'index.js') // Sent to dist/vue folder in webpack.
+        this.source = path.join(
+            'vue',
+            'src',
+            'login',
+            'webview',
+            'vue',
```
justinmk3, line 67:
```suggestion
            'vue/src/login/webview/vue',
```

justinmk3 (approved):

justinmk3:
Might be worth rebasing this, then let's merge it?

PR #4521, opened 2024-03-11
tests: Add feature dev codegen e2e tests

jpinkney-aws:
## Problem
- We don't have any e2e tests to make sure codegen is working

## Solution
- Add tests for all codegen states
- Add tests for all possible ending states including iterations

List of test cases:
```
Amazon Q Feature Dev
    quick action availability
      - Shows /dev when feature dev is enabled
      - Does NOT show /dev when feature dev is NOT enabled
    /dev {msg} entry
      - Should receive chat response
      Moves directly from approach to codegen
        Clicks accept code
          - clicks new task
          - click close session
        Iterates on codegen
          Clicks accept code
            - clicks new task
            - click close session
      Iterates on approach
        - Should iterate successfully
        Moves to codegen after iteration
          Clicks accept code
            - clicks new task
            - click close session
          Iterates on codegen
            Clicks accept code
              - clicks new task
              - click close session
    /dev entry
      - Clicks examples
      - Should receive chat response
      Moves directly from approach to codegen
        Clicks accept code
          - clicks new task
          - click close session
        Iterates on codegen
          Clicks accept code
            - clicks new task
            - click close session
      Iterates on approach
        - Should iterate successfully
        Moves to codegen after iteration
          Clicks accept code
            - clicks new task
            - click close session
          Iterates on codegen
            Clicks accept code
              - clicks new task
              - click close session
```

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
+++ b/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
@@ -26,7 +26,7 @@ import { placeholder } from '../../../shared/vscode/commands2'
 import { EditorContentController } from '../../../amazonq/commons/controllers/contentController'
 import { openUrl } from '../../../shared/utilities/vsCodeUtils'
 import { getPathsFromZipFilePath, getWorkspaceFoldersByPrefixes } from '../../util/files'
-import { userGuideURL } from '../../../amazonq/webview/ui/texts/constants'
+import { examples, newTaskChanges, sessionClosed, updateCode } from '../../userFacingText'
 
 export interface ChatControllerEventEmitters {
     readonly processHumanChatMessage: EventEmitter<any>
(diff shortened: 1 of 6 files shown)


packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -246,8 +246,14 @@ export class FeatureDevController {
     private async onApproachGeneration(session: Session, message: string, tabID: string) {
         await session.preloader(message)
 
+        this.messenger.sendAnswer({
+            type: 'answer',
+            tabID,
+            message: 'Ok, let me create a plan. This may take a few minutes.',
```
grimalk, line 252: This is also user facing text, right?


Let's extract this also to the shared place

packages/core/src/testE2E/amazonq/framework/messenger.ts
```diff
@@ -73,29 +74,48 @@ export class Messenger {
...
 
+    getPlaceholder() {
+        return this.getStore().promptInputPlaceholder
+    }
+
+    hasButton(type: FollowUpTypes) {
+        return this.getChatItems()
+            .pop()
+            ?.followUp?.options?.map(opt => opt.type)
+            .includes(type)
```
grimalk, line 85: Am I missing something? This code either just throws TypeError on `undefined` or returns `undefined`, probably the former, because the `.includes` isn't chained

Can we make this method always return boolean? I don't really see the point of using this method as below with `this.hasButton(FollowUpTypes.Retry) ?? false`

grimalk (changes requested):

justinmk3 (approved):

PR #4523, opened 2024-03-12
feat: Introduce file by file acceptance

jpinkney-aws:
## Problem
- We don't have a way for users to individually select which files should be accepted after code generation

## Solution
- Allow users to reject files from being accepted after code generation

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18819,7 +18819,7 @@
             },
             "devDependencies": {
                 "@aws-sdk/types": "^3.13.1",
-                "@aws-toolkits/telemetry": "^1.0.192",
+                "@aws-toolkits/telemetry": "^1.0.193",
                 "@aws/fully-qualified-names": "^2.1.4",
                 "@cspotcode/source-map-support": "^0.8.1",
                 "@sinonjs/fake-timers": "^10.0.2",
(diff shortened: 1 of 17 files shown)


package-lock.json
```diff
@@ -3556,9 +3556,9 @@
             }
         },
         "node_modules/@aws-toolkits/telemetry": {
-            "version": "1.0.192",
-            "resolved": "https://registry.npmjs.org/@aws-toolkits/telemetry/-/telemetry-1.0.192.tgz",
-            "integrity": "sha512-EasZTUEtSbYDo7AdpxAmRdW3YxL0C67fG65ZagtwiCacT8P3XB6jH1x99EW21Q4bM+t0YQBmnCQ0qFhlGDYCXA==",
+            "version": "1.0.193",
```
justinmk3, line 3559: are these package-lock changes intentional? does this match package.json?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -600,6 +612,27 @@ To learn more, visit the _[Amazon Q User Guide](${userGuideURL})_.
         return { left, right }
     }
 
+    private async fileClicked(message: fileClickedMessage) {
+        // TODO: add Telemetry here
```
justinmk3, line 616: could use the existing `ui_click` metric

justinmk3 (approved):

PR #4526, opened 2024-03-12
fix(cwspr): add more log info for auth code path

leigaol:
## Problem

The current auth code path does not have sufficient logging, which makes it hard to debug auth related issues.

## Solution

Add log.info in aws toolkit auth code path. 

## License

### DIFF
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -506,12 +506,12 @@ export class Auth implements AuthService, ConnectionManager {
     private async validateConnection<T extends Profile>(id: Connection['id'], profile: StoredProfile<T>) {
         const runCheck = async () => {
             if (profile.type === 'sso') {
-                getLogger().info(`auth: Validating SSO connection: ${id}`)
                 const provider = this.getTokenProvider(id, profile)
                 if ((await provider.getToken()) === undefined) {
                     getLogger().info(`auth: Connection is not valid: ${id} `)
                     return this.updateConnectionState(id, 'invalid')
                 } else {
+                    getLogger().info(`auth: Connection is valid: ${id} `)
                     return this.updateConnectionState(id, 'valid')
                 }
             } else {
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -188,7 +188,7 @@ export class Auth implements AuthService, ConnectionManager {
         if (profile === undefined) {
             throw new Error(`Connection does not exist: ${id}`)
         }
-        getLogger().info(`validation connection ${id} before using it`)
+        getLogger().info(`auth: validating connection before using it: ${id}`)
         const validated = await this.validateConnection(id, profile)
         const conn =
             validated.type === 'sso' ? this.getSsoConnection(id, validated) : this.getIamConnection(id, validated)
@@ -339,7 +339,7 @@ export class Auth implements AuthService, ConnectionManager {
      * Put the SSO connection in to an expired state
      */
     public async expireConnection(conn: Pick<SsoConnection, 'id'>): Promise<void> {
-        getLogger().info(`Expiring connection ${conn.id}`)
+        getLogger().info(`auth: Expiring connection ${conn.id}`)
         const profile = this.store.getProfileOrThrow(conn.id)
         if (profile.type === 'iam') {
             throw new ToolkitError('Auth: Cannot force expire an IAM connection')
@@ -375,7 +375,7 @@ export class Auth implements AuthService, ConnectionManager {
 
     public async updateConnection(connection: Pick<SsoConnection, 'id'>, profile: SsoProfile): Promise<SsoConnection>
     public async updateConnection(connection: Pick<Connection, 'id'>, profile: Profile): Promise<Connection> {
-        getLogger().info(`Updating connection ${connection.id}`)
+        getLogger().info(`auth: Updating connection ${connection.id}`)
         if (profile.type === 'iam') {
             throw new Error('Updating IAM connections is not supported')
         }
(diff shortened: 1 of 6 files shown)


packages/core/src/auth/auth.ts
```diff
@@ -502,8 +506,10 @@ export class Auth implements AuthService, ConnectionManager {
     private async validateConnection<T extends Profile>(id: Connection['id'], profile: StoredProfile<T>) {
         const runCheck = async () => {
             if (profile.type === 'sso') {
+                getLogger().info(`Validating SSO connection ${id}`)
                 const provider = this.getTokenProvider(id, profile)
                 if ((await provider.getToken()) === undefined) {
+                    getLogger().info(`Connection ${id} is not valid`)
```
justinmk3, line 512: Almost always, the variable part should go at the end of the message. That helps readability.

```suggestion
                    getLogger().info(`Connection is not valid: ${id}`)
```

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -84,7 +85,9 @@ export class SsoAccessTokenProvider {
 
     public async getToken(): Promise<SsoToken | undefined> {
         const data = await this.cache.token.load(this.tokenCacheKey)
-
+        getLogger().info(`current client registration id=${data?.registration?.clientId}, 
+                            expires at ${data?.registration?.expiresAt}, 
+                            key = ${this.tokenCacheKey}`)
```
justinmk3, line 90: use the indent() function from textUtilities.ts

packages/core/src/auth/secondaryAuth.ts
```diff
@@ -156,6 +156,10 @@ export class SecondaryAuth<T extends Connection = Connection> {
     }
 
     public get isConnectionExpired() {
+        if (this.activeConnection) {
+            getLogger().info(`secondaryAuth connection id = ${this.activeConnection.id}`)
+            getLogger().info(`secondaryAuth connection status = ${this.auth.getConnectionState(this.activeConnection)}`)
```
justinmk3, line 161: put these in one message. use the indent() function from textUtilities.ts

packages/core/src/auth/auth.ts
```diff
@@ -502,8 +506,10 @@ export class Auth implements AuthService, ConnectionManager {
     private async validateConnection<T extends Profile>(id: Connection['id'], profile: StoredProfile<T>) {
         const runCheck = async () => {
             if (profile.type === 'sso') {
+                getLogger().info(`Validating SSO connection ${id}`)
```
justinmk3, line 509: why are these info instead of debug? this will cause a lot of messages for all users.

leigaol: Because level info is by default. Many users do not adjust log level before providing log to us, making it hard for us to get debug level logs. 

Making them info level is a temporary solution(actually proposed by Anil). We can downgrade them to debug once this auth issue is resolved.

packages/core/src/auth/auth.ts
```diff
@@ -502,8 +506,10 @@ export class Auth implements AuthService, ConnectionManager {
     private async validateConnection<T extends Profile>(id: Connection['id'], profile: StoredProfile<T>) {
         const runCheck = async () => {
             if (profile.type === 'sso') {
+                getLogger().info(`auth: Validating SSO connection: ${id}`)
                 const provider = this.getTokenProvider(id, profile)
                 if ((await provider.getToken()) === undefined) {
+                    getLogger().info(`auth: Connection is not valid: ${id} `)
                     return this.updateConnectionState(id, 'invalid')
                 } else {
                     return this.updateConnectionState(id, 'valid')
```
justinmk3, line 515: Let's add a "Connect is valid" log here. And remove the "Validating..." log on line 509. That gives us something explicit to look for in both cases.

packages/core/src/auth/secondaryAuth.ts
```diff
@@ -156,6 +157,12 @@ export class SecondaryAuth<T extends Connection = Connection> {
     }
 
     public get isConnectionExpired() {
+        if (this.activeConnection) {
+            getLogger().info(
+                indent(`secondaryAuth connection id = ${this.activeConnection.id}
+            secondaryAuth connection status = ${this.auth.getConnectionState(this.activeConnection)}`)
```
justinmk3, line 163: we probably want to use the `clear` parameter of `indent()`.
```suggestion
            secondaryAuth connection status = ${this.auth.getConnectionState(this.activeConnection)}`,
            4, true)
```

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -84,7 +86,11 @@ export class SsoAccessTokenProvider {
 
     public async getToken(): Promise<SsoToken | undefined> {
         const data = await this.cache.token.load(this.tokenCacheKey)
-
+        getLogger().info(
+            indent(`current client registration id=${data?.registration?.clientId}, 
+                            expires at ${data?.registration?.expiresAt}, 
+                            key = ${this.tokenCacheKey}`)
```
justinmk3, line 92:
```suggestion
                            key = ${this.tokenCacheKey}`, 4, true)
```

packages/core/src/auth/auth.ts
leigaol: DONE

justinmk3 (approved):

leigaol:
Restarted many times but the CI in win still fails.

Will-ShaoHua (approved):

justinmk3:
Lint has been failing since this PR (could be a coincidence though)?

PR #4527, opened 2024-03-12
fix: Toolkit creates too many Output channels

justinmk3:
## Problem

- UI cost: Security Scan feature creates vscode Output channel on startup. This is unnecessary for users that aren't currently using that feature.
    - image

## Solution

- create Security Scan channel on-demand, instead of in module-scope.
    - image

## License

### DIFF
--- a/packages/core/src/codewhisperer/commands/basicCommands.ts
+++ b/packages/core/src/codewhisperer/commands/basicCommands.ts
@@ -33,6 +33,7 @@ import { CodeWhispererSource } from './types'
 import { showManageConnections } from '../../auth/ui/vue/show'
 import { FeatureConfigProvider } from '../service/featureConfigProvider'
 import { TelemetryHelper } from '../util/telemetryHelper'
+import { isTextEditor } from '../../shared/utilities/editorUtilities'
 
 export const toggleCodeSuggestions = Commands.declare(
     { id: 'aws.codeWhisperer.toggleCodeSuggestion', compositeKey: { 1: 'source' } },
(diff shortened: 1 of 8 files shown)


jpinkney-aws (approved):

hayemaxi (approved):

packages/core/src/codewhisperer/commands/startSecurityScan.ts
```diff
@@ -34,15 +34,27 @@ import { openUrl } from '../../shared/utilities/vsCodeUtils'
 import { AuthUtil } from '../util/authUtil'
 import { DependencyGraphConstants } from '../util/dependencyGraph/dependencyGraph'
 import path from 'path'
+import { once } from '../../shared/utilities/functionUtils'
 
 const performance = globalThis.performance ?? require('perf_hooks').performance
-const securityScanOutputChannel = vscode.window.createOutputChannel('CodeWhisperer Security Scan Logs')
-const codeScanLogger = makeLogger({
-    outputChannels: [securityScanOutputChannel],
-})
```
justinmk3, line 42: Using memoization (`once()`) eliminates these top-level variables, and improves their type signatures (eliminates nullability).

packages/core/src/codewhisperer/commands/startSecurityScan.ts
```diff
@@ -276,6 +289,7 @@ function populateCodeScanLogStream(scannedFiles: Set<string>) {
         const uri = vscode.Uri.file(file)
         codeScanLogger.info(`File scanned: ${uri.fsPath}`)
     }
+    codeScanOutpuChan.show()
```
justinmk3, line 292: This is reliable, unlike `codeScanLogsOutputChannelId`

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -86,7 +87,7 @@ export const showSecurityScan = Commands.declare(
                 await AuthUtil.instance.notifyReauthenticate()
             }
             const editor = vscode.window.activeTextEditor
-            if (editor) {
+            if (editor && isTextEditor(editor)) {
```
justinmk3, line 90: This was a bug? @leigaol @Will-ShaoHua

Will-ShaoHua: it seems to be a bug.

PR #4532, opened 2024-03-13
feat(codewhisperer): Line Tracker implementation

Will-ShaoHua:
This PR is a split of the original all-in-one PR #4554  

Please refer to the PR listed above for project demos and use cases

## Problem
LineTracker is a component which will track users' line selection changes and fire `activeLineChanged` event when there is a change and notify subscribers to update their UI.

## Solution

`LineTracker` will subscribe the following events fired by VSCode IDE, update `LineTracker.selections` and fire `LineChangedEvent`
(1) contentChanged
(2) editorSelectionChanged
(3) activeEditorChanged

## License


packages/core/src/codewhisperer/tracker/lineTracker.ts
```diff
@@ -0,0 +1,166 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
hayemaxi, line 5: can you add some documentation about what this class is and what its intended to be used for?

Will-ShaoHua: sgtm, Max fyi I merged #4534 #4535 into this one.

packages/core/src/codewhisperer/views/activeStateController.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    // TODO: uncomment
+    // private async onSuggestionActionEvent(e: SuggestionActionEvent) {
+    //     if (!this._isReady) {
+    //         return
+    //     }
+
+    //     this.clear(e.editor) // do we need this?
+    //     await this.refreshDebounced.promise(e.editor)
+    // }
```
Will-ShaoHua, line 70: waiting for https://github.com/aws/aws-toolkit-vscode/pull/4533 to be merged

Will-ShaoHua:
Note that this new UI component is not used yet in the PR

packages/core/src/codewhisperer/service/serviceContainer.ts
```diff
@@ -0,0 +1,33 @@
...
+        if (Container.#instance) {
+            throw new Error('Container already exists')
+        }
+
+        Container.#instance = new Container(authProvider)
+        return Container.#instance
+    }
+
+    static get instance(): Container {
+        return Container.#instance ?? Container.create(AuthUtil.instance)
```
nkomonen-amazon, line 23: With the following you can get rid of the `create` method

```suggestion
        return Container.#instance ??= new Container(AuthUtil.instance)
```

packages/core/src/codewhisperer/service/serviceContainer.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    readonly lineTracker: LineTracker
+    readonly activeStateController: ActiveStateController
+
+    constructor(readonly auth: AuthUtil) {
```
nkomonen-amazon, line 29: If you make this `protected` then you do not have to worry about anyone accidentally not using `instance`

packages/core/src/codewhisperer/service/serviceContainer.ts
```diff
@@ -0,0 +1,33 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { AuthUtil } from '../util/authUtil'
+import { ActiveStateController } from '../views/activeStateController'
+import { LineTracker } from '../tracker/lineTracker'
+
+export class Container {
```
justinmk3, line 10: How are other developers expected to know what the purpose of this class is?

packages/core/src/codewhisperer/service/serviceContainer.ts
```diff
@@ -0,0 +1,33 @@
...
+        Container.#instance = new Container(authProvider)
+        return Container.#instance
+    }
+
+    static get instance(): Container {
+        return Container.#instance ?? Container.create(AuthUtil.instance)
+    }
+
+    readonly lineTracker: LineTracker
+    readonly activeStateController: ActiveStateController
```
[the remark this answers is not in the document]
Will-ShaoHua: yea, agree, that's the reason why I add this container class, afterward, I want to bring in other components into this container so that they can have reference to each other easily and don't need to make them singleton

packages/core/src/codewhisperer/tracker/lineTracker.ts
```diff
@@ -0,0 +1,169 @@
...
+    readonly selections: LineSelection[] | undefined
+
+    readonly reason: 'editor' | 'selection' | 'content'
+}
+
+/**
+ * This class providees a single entry point interface to manage and access users' "line" selections
+ * Callers could use it by subscribing onDidChangeActiveLines to do UI updates or logic needed to be executed when line selections get changed
+ */
+export class LineTracker implements vscode.Disposable {
```
justinmk3, line 25: why is this not part of `CodeWhispererTracker` ? should `CodeWhispererTracker` be renamed to `ModificationsTracker` ?

packages/core/src/shared/utilities/editorUtilities.ts
```diff
@@ -9,6 +9,11 @@ import { Settings } from '../settings'
 
 const defaultTabSize = 4
 
+export function isTextEditor(editor: vscode.TextEditor): boolean {
```
justinmk3, line 12: Thanks for adding this to the shared utils

packages/core/src/codewhisperer/service/serviceContainer.ts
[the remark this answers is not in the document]
Will-ShaoHua: so by passing the container into its children components' constructors, we can just `this.container.otherCompoenent` instead of `otherCompoenent.getInstance`

packages/core/src/codewhisperer/views/activeStateController.ts
```diff
@@ -0,0 +1,145 @@
...
+import * as vscode from 'vscode'
+import { Container } from '../service/serviceContainer'
+import { RecommendationHandler } from '../service/recommendationHandler'
+import { cancellableDebounce } from '../../shared/utilities/functionUtils'
+import { LineSelection, LinesChangeEvent } from '../tracker/lineTracker'
+import { isTextEditor } from '../../shared/utilities/editorUtilities'
+import { RecommendationService } from '../service/recommendationService'
+import { subscribeOnce } from '../../shared/utilities/vsCodeUtils'
+
+export class ActiveStateController implements vscode.Disposable {
```
justinmk3, line 15: needs a docstring. what is the purpose of this, how does it differ from the numerous other "state" related classes in CW?

packages/core/src/codewhisperer/service/serviceContainer.ts
Will-ShaoHua: ack, will add it. So my expectation is we need to move other objects into it as well,, e.g. `RecommendationService`, `RecommendationHandler` etc. But that's outside of the scope of this PR and project. We shouldn't add more singleton object if it's not needed.

packages/core/src/codewhisperer/views/activeStateController.ts
Will-ShaoHua: Gotit, this one is responsible for  updating UI specifically on suggestion action event or line change event. So currently we don't have a single source of truth component who is responsible for this. For example, status bar is updated within `InlineCompletionService.ts`, and I lean to have a dedicated component to handle this instead of spreading them every where in the code base.

Will-ShaoHua: So it's a good idea to move status bar update logic into this component afterward.

Will-ShaoHua: maybe I should rename this class and make it more self-explanatory it's an UI updater for suggestion action.

packages/core/src/codewhisperer/service/serviceContainer.ts
Will-ShaoHua: ackk

packages/core/src/codewhisperer/tracker/lineTracker.ts
Will-ShaoHua: > should CodeWhispererTracker be renamed to ModificationsTracker ?

yea it should be renamed to `ModificationTracker`.

Will-ShaoHua: > why is this not part of CodeWhispererTracker

I think it makes more sense `LineTracker` should combine with `CodeCoverageTracker` as they both 2 are subscribing DocumentChangedEvent, however I would prefer extracting this part to another later `DocumentChangeTracker` given the fact that "LineTracker" sounds more reasonable to only fire events when there is "line" change but not content change.

Will-ShaoHua:
I will make PRs respectively for `Active State Indicator`(codewhisperer is generating...) and `Inline Tutorial controller` so that this PR has its own source of changes LineTracker and some utils.

andrewyuq (approved):

packages/core/src/codewhisperer/tracker/lineTracker.ts
justinmk3: > I think it makes more sense `LineTracker` should combine with `CodeCoverageTracker` as they both 2 are subscribing DocumentChangedEvent, however I would prefer extracting this part to another component maybe `DocumentChangeTracker`.

ok. Hoping to see a followup PR for that.

justinmk3 (approved):
Approving contigent on backlog items being addressed in the near future.

PR #4537, opened 2024-03-14
AmazonQFeatureDev: Handle canned errors during code generation

jsalles:
## Problem

FeatureDev's code generation will sometimes return canned errors with detailed information about the problems that happened in the process. These errors are not retryable and should be worked around by the customer, using the information provided.

## Solution

This CR adds the support for these canned errors, and adds dynamic logic for showing or hiding the `Retry` button whenever the code generation step fails.

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/client/codewhispererruntime-2022-11-11.json
+++ b/packages/core/src/amazonqFeatureDev/client/codewhispererruntime-2022-11-11.json
@@ -246,6 +246,9 @@
                     "shape": "ThrottlingException"
                 },
                 {
+                    "shape": "ResourceNotFoundException"
+                },
+                {
                     "shape": "InternalServerException"
                 },
                 {
@@ -272,6 +275,9 @@
             "errors": [
                 {
                     "shape": "ThrottlingException"
+                },
+                {
+                    "shape": "ResourceNotFoundException"
                 },
                 {
                     "shape": "InternalServerException"
@@ -514,6 +520,9 @@
                     "shape": "ThrottlingException"
                 },
                 {
+                    "shape": "ResourceNotFoundException"
+                },
+                {
                     "shape": "InternalServerException"
                 },
                 {
@@ -533,10 +542,18 @@
             "members": {
                 "message": {
                     "shape": "String"
+                },
+                "reason": {
+                    "shape": "AccessDeniedExceptionReason"
                 }
             },
             "documentation": "<p>This exception is thrown when the user does not have sufficient access to perform this action.</p>",
             "exception": true
+        },
+        "AccessDeniedExceptionReason": {
+            "type": "string",
+            "documentation": "<p>Reason for AccessDeniedException</p>",
+            "enum": ["UNAUTHORIZED_CUSTOMIZATION_RESOURCE_ACCESS"]
         },
         "ArtifactMap": {
             "type": "map",
@@ -558,8 +575,7 @@
             "required": ["content"],
             "members": {
                 "messageId": {
-                    "shape": "AssistantResponseMessageMessageIdString",
-                    "documentation": "<p>Unique identifier for the chat message</p>"
+                    "shape": "MessageId"
                 },
                 "content": {
                     "shape": "AssistantResponseMessageContentString",
@@ -586,11 +602,6 @@
             "min": 0,
             "sensitive": true
         },
-        "AssistantResponseMessageMessageIdString": {
-            "type": "string",
-            "max": 128,
-            "min": 0
-        },
         "Base64EncodedPaginationToken": {
             "type": "string",
             "max": 2048,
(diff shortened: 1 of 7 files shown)


packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -266,6 +265,15 @@ abstract class CodeGenBase {
...
+                    // canned errors are not retryable
+                    if (codegenResult.codeGenerationStatusDetail) {
+                        return {
+                            result: 'failed',
+                            message: codegenResult.codeGenerationStatusDetail,
+                            retryable: false,
+                        }
+                    }
+
                     throw new ToolkitError('Code generation failed', { code: 'CodeGenFailed' })
```
jpinkney-aws, line 277: Won't we also want to know that code generation failed? I think this code will make it so the telemetry failure event never gets reported when we get a canned error

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -411,9 +435,13 @@ export class FeatureDevController {
         let session
         try {
             session = await this.sessionStorage.getSession(message.tabID)
+            if (session.state.codeGenerationResult?.result !== 'success') {
+                throw new ToolkitError('Invalid state in code generation')
```
jpinkney-aws, line 439: Is it worth extending a toolkiterror and defining a code just in case these cases are being hit and sending this data to telemetry? I see a couple places like these in the changes that might give us some potential insights if something is going very wrong on the users end

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -144,7 +145,7 @@ export class FeatureDevController {
         })
     }
 
-    private async processChatItemVotedMessage(tabId: string, messageId: string, vote: string) {
+    private async processChatItemVotedMessage(tabId: string, _messageId: string, vote: string) {
```
jpinkney-aws, line 148: probably just makes sense to remove it rather then underscore it if its not being used, but that can be a different PR

packages/core/src/amazonqFeatureDev/session/sessionState.ts
jsalles: This function is invoked from inside a `amazonq_codeGenerationInvoke` telemetry block, so I assume we log the failure if we throw inside the block.

For the canned error case that we don't throw, we set 
```
            telemetry.setCodeGenerationResult(codegenResult.codeGenerationStatus.status)
```
so we will have the information of generations that failed.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
jsalles: Yeah good point! Those pieces of code should normally not be reachable, so it makes sense for us to get special telemetry on them.

I'll extend the ToolkitError for this.

jpinkney-aws (approved):

packages/core/src/amazonqFeatureDev/client/codewhispererruntime-2022-11-11.json
sannicm, line 1: question: Why so many model changes are needed for this handling?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -300,6 +306,29 @@ export class FeatureDevController {
         this.messenger.sendAsyncEventProgress(tabID, false, undefined)
     }
 
+    private printFailedCodeGenMessage(session: Session, message: string, tabID: string, retryable: boolean) {
```
sannicm, line 309: question: isn't this function missing the chat prompt blocking?

packages/core/src/amazonqFeatureDev/client/codewhispererruntime-2022-11-11.json
jsalles: They're not all related to this change. 

This json is auto generated, so this here is the latest version of the models.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
jsalles: The prompt is blocked on line 397. It's the last step of the function that calls this one.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -364,13 +391,15 @@ export class FeatureDevController {
...
-            this.messenger.sendUpdatePlaceholder(tabID, 'Select an option above to proceed')
         } finally {
             // Finish processing the event
             this.messenger.sendAsyncEventProgress(tabID, false, undefined)
-
-            // Lock the chat input until they explicitly click one of the follow ups
-            this.messenger.sendChatInputEnabled(tabID, false)
+            this.messenger.sendChatInputEnabled(tabID, enableInput)
+            if (enableInput) {
+                this.messenger.sendUpdatePlaceholder(tabID, 'How can this plan be improved?')
```
sannicm, line 399: question: has this message been approved by johanna?

I think we need to synchronize to understand what message can be a follow up for retryable errors.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -319,30 +349,27 @@ export class FeatureDevController {
             })
             this.messenger.sendUpdatePlaceholder(tabID, 'Generating code ...')
             await session.send(message)
-            const filePaths = session.state.filePaths ?? []
-            const deletedFiles = session.state.deletedFiles ?? []
+
+            if (session.state.codeGenerationResult?.result === 'pending') {
+                throw new ToolkitError('Unexpected state in code generation')
```
sannicm, line 354: question: shouldn't here be thrown InvalidCodeGenStateError?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -319,30 +349,27 @@ export class FeatureDevController {
...
+            }
+            if (session.state.codeGenerationResult?.result === 'failed') {
+                // enable the input if the error is not retryable.
+                // we're expecting a new prompt in this case
+                enableInput = !session.state.codeGenerationResult.retryable
```
sannicm, line 359: question: will there be failed code generation results that are non retryable? 

Only canned errors are at the moment handled and they are all non retryable.

PR #4540, opened 2024-03-14
New Transform Q Feature: Chat Invocation

JuliaABurch:
## Problem
Rather than using assorted and disparate IDE UX features to drive the Q Transformation process, we would like to have the customer be able to do the majority of it through Q Chat.

## Solution
This PR moves the logic governing the `/transform` quick action to its own chat handler, which then takes user input to determine which module to transform and tracks the job submission status. 

Screenshot 2024-03-14 at 10 10 49 AM

## License


packages/core/src/amazonq/webview/ui/tabs/generator.ts
```diff
@@ -51,6 +53,12 @@ Here I can provide code suggestions across files in your current project.
 Before I begin generating code, let's agree on an implementation plan. What change would you like to make?
 `,
         ],
+        [
+            'gumby',
+            `Welcome to Code Transformation!
+
+I can help you upgrade your Java 8 and 11 codebases to Java 17.`,
```
dhasani23, line 60: Make sure Allie is okay with this text and the welcome message too

packages/core/src/amazonq/webview/ui/tabs/generator.ts
```diff
@@ -81,5 +95,14 @@ Before I begin generating code, let's agree on an implementation plan. What chan
                   ]
                 : [],
         }
+
+        if (needConfirmationClose) {
+            tabData.tabCloseConfirmationMessage =
+                'Are you sure want to close the tab? Closing the tab would mean that your running job will stop.'
```
dhasani23, line 101: Same here, check with Allie

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    private async prepareProjectForSubmission(message: any | undefined = undefined): Promise<void> {
+        if (message !== undefined && message.formSelectedValues !== undefined) {
+            const javaHome: string = message.formSelectedValues['JavaHomeFormInput'].trim()
+
+            if (!javaHome) {
+                const errorMessage = 'No JDK path provided'
+                this.messenger.sendErrorMessage(errorMessage, message.tabID)
+                throw new JavaHomeNotSetError()
```
dhasani23, line 231: Do we need a custom error class?

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -0,0 +1,358 @@
...
+            if (candidateModule.JDKVersion !== undefined) {
+                uniqueJavaOptions.add(candidateModule.JDKVersion)
+            }
+        })
+
+        const jdkFormOptions: { value: string; label: string }[] = []
+        uniqueJavaOptions.forEach(jdkVersion => {
+            jdkFormOptions.push({
+                value: jdkVersion.toString(),
+                label: jdkVersion.toString(),
```
dhasani23, line 116: `value` and `label` are the same?

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -0,0 +1,358 @@
...
+                    value: JDKVersion.JDK17.toString(),
+                    label: JDKVersion.JDK17.toString(),
+                },
+            ],
+        })
+
+        this.dispatcher.sendAsyncEventProgress(
+            new AsyncEventProgressMessage(tabID, {
+                inProgress: true,
+                message: `I can upgrade your Java ${jdkFormOptions[0].label} project. To start the transformation, I need some information from you. Choose the module you want to upgrade and the target code version to upgrade to, and then choose Transform. It can take 10-30 minutes to upgrade your code, depending on the size of your module.`,
```
dhasani23, line 153: Should this be hardcoded as `[0]` when the user hasn't selected a project yet? Also as a general note I think we should choose between `module` and `project` and use that everywhere. I've been doing `project` so far.

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -0,0 +1,358 @@
...
+            case 'no-java-project-found':
+                message =
+                    'None of your open projects are supported by Amazon Q Code Transformation. Currently, Amazon Q can only upgrade Java projects built on Maven. \n For more information, see the Amazon Q documentation.'
+                break
+            case 'no-maven-java-project-found':
+                message =
+                    'None of your open Java projects are supported by Amazon Q Code Transformation. Currently, Amazon Q can only upgrade Java projects built on Maven. A pom.xml must be present in the root of your project to upgrade it. \n For more information, see the Amazon Q documentation.'
+                break
+            case 'could-not-compile-project':
+                message = `Amazon Q couldn't execute the Maven install or Maven copy-dependencies commands. To troubleshoot, see the [Amazon Q documentation.](https://docs.aws.amazon.com/amazonq/latest/aws-builder-use-ug/troubleshooting-code-transformation.html#maven-commands-failing)`
```
dhasani23, line 289: Maybe just generalize this to `Maven commands`

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -94,13 +94,24 @@ export const createMynahUI = (ideApi: any, featureDevInitEnabled: boolean, gumby
...
+                        mynahUI.updateStore(tabID, {
+                            promptInputDisabledState: false,
+                        })
+                    }
                 }
             }
         },
         onFileActionClick: (tabID: string, messageId: string, filePath: string, actionName: string): void => {},
+        onQuickHandlerCommand: (tabID: string, command?: string, eventId?: string) => {
+            console.log(`onQuickHandlerCommandMessage!! command: ${command} tabID: ${tabID}`)
```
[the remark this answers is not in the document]
damntrecky: Blocker: Agree with @dhasani23.

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -299,13 +341,19 @@ export const createMynahUI = (ideApi: any, featureDevInitEnabled: boolean, gumby
...
-                quickActionHandler.handle(prompt, tabID)
+                quickActionHandler.handle(prompt, tabID, eventId)
                 return
             }
 
             textMessageHandler.handle(prompt, tabID)
         },
         onVote: connector.onChatItemVoted,
+        onInBodyButtonClicked: (tabId, messageId, action, eventId) => {
+            connector.onCustomFormAction(tabId, messageId, action, undefined)
```
damntrecky, line 352: NIT: Probably don't have to pass in undefined to the function if `undefined` is a type for that argument. Would only have to pass in `undefined` if there were more params after. Eg) `            connector.onCustomFormAction(tabId, messageId, action, undefined, anotherArgument)`

packages/core/src/amazonq/webview/ui/quickActions/handler.ts
```diff
@@ -45,16 +47,71 @@ export class QuickActionHandler {
...
-    private handleGumbyCommand(tabID: string) {
-        this.connector.transform(tabID)
+    private handleGumbyCommand(tabID: string, eventId: string | undefined) {
+        if (!this.isGumbyEnabled) {
+            return
+        }
+
+        let gumbyTabId: string | undefined = undefined
+
+        this.tabsStorage.getTabs().forEach(tab => {
```
damntrecky, line 65: Question: I assume `tabType` is a typed field and `gumby` is a recognized string literal? Would this throw a syntax error during development if I did `tab.type === 'gumbyBadString`? 

Possible suggestions: Verify string literal type or enum for `gumby` string.

packages/core/src/amazonq/webview/ui/quickActions/handler.ts
```diff
@@ -45,16 +47,71 @@ export class QuickActionHandler {
...
+        }
+
+        if (affectedTabId === undefined) {
+            this.mynahUI.notify({
+                content: uiComponentsTexts.noMoreTabsTooltip,
+                type: NotificationType.WARNING,
+            })
+            return
+        } else {
+            this.tabsStorage.updateTabTypeFromUnknown(affectedTabId, 'gumby')
```
damntrecky, line 93: Question: I see this `gumby` string a lot. Is it typed to a string literal in all instances or could we provide a string constant or enum value for all these instances?

packages/core/src/amazonq/webview/ui/storages/tabsStorage.ts
```diff
@@ -4,7 +4,7 @@
  */
 
 export type TabStatus = 'free' | 'busy' | 'dead'
-export type TabType = 'cwc' | 'featuredev' | 'unknown'
+export type TabType = 'cwc' | 'featuredev' | 'gumby' | 'unknown'
```
damntrecky, line 7: Note: Here `TabType` is a string literal. Ignore previous comments on string literal.

packages/core/src/amazonqGumby/app.ts
```diff
@@ -0,0 +1,77 @@
...
+            const authenticatingSessions = sessionStorage.getAuthenticatingSessions()
+
+            authenticatingSessionIDs = authenticatingSessions.map(session => session.tabID)
+
+            // We've already authenticated these sessions
+            authenticatingSessions.forEach(session => (session.isAuthenticating = false))
+        }
+
+        messenger.sendAuthenticationUpdate(authenticated, authenticatingSessionIDs)
+    }, 500)
```
damntrecky, line 67: NIT: Is there a global debounce constant to reference?

Eg)
```
const DEBOUNCE = {
   USER_TYPE_INPUT: 200,
   AUTH_REAVALIDATION: 500,
   ...
   }
   ```

File: packages/core/src/amazonqGumby/chat/controller/controller.ts (hunk 3 above)
Comment (damntrecky) [reply packages/core/src/amazonqGumby/chat/controller/controller.ts:231]: Question: Can we use `AWSToolkitError here`?

File: packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts [hunk 13]
```diff
@@ -0,0 +1,358 @@
...
+        uniqueJavaOptions.forEach(jdkVersion => {
+            jdkFormOptions.push({
+                value: jdkVersion.toString(),
+                label: jdkVersion.toString(),
+            })
+        })
+
+        const formItems: ChatItemFormItem[] = []
+        formItems.push({
+            id: 'GumbyTransformModuleForm',
```
damntrecky, line 122: Syntax NIT: Put these ID's into string literal or enum.

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -0,0 +1,358 @@
...
+            })
+        )
+
+        this.dispatcher.sendChatPrompt(
+            new ChatPrompt(
+                {
+                    message: 'Q Code Transformation',
+                    formItems: formItems,
+                },
+                `TransformForm`,
```
damntrecky, line 163: Nit: No need for backtick to denote a string literal, could just be normal string syntax `"Transform"` or `'Transform'`

packages/core/src/amazonqGumby/chat/controller/messenger/stringConstants.ts
```diff
@@ -0,0 +1,23 @@
...
+export const nonWindowsJava8HomeHelpMessage =
+    'To find your JAVA_HOME path, run this command in a new IDE terminal:  `/usr/libexec/java_home -v 1.8`. The path should look similar to this: `/Library/Java/JavaVirtualMachines/jdk-8.jdk/Contents/Home`.'
+
+export const nonWindowsJava11HomeHelpMessage =
+    'To find your JAVA_HOME path, run this command in a new IDE terminal:  `/usr/libexec/java_home -v 11`. The path should look similar to this: `/Library/Java/JavaVirtualMachines/jdk-11.jdk/Contents/Home`.'
+
+export const projectSizeTooLargeMessage =
+    'Your project size exceeds the Amazon Q Code Transformation upload limit of 1GB. Please see: https://docs.aws.amazon.com/amazonq/latest/aws-builder-use-ug/code-transformation.html#prerequisites'
+
+export const JDK8VersionNumber = '52'
```
damntrecky, line 21: We have these all well defined in `javapOutputToTelemetryValue()` function in `codeTransformTelemetry.ts`

packages/core/src/amazonqGumby/commands.ts
```diff
@@ -5,28 +5,41 @@
...
         }
 
         if (transformByQState.isNotStarted()) {
             logCodeTransformInitiatedMetric(source)
-            await startTransformByQWithProgress()
+            await focusAmazonQPanel().then(async () => {
+                // non blocking wait for a tenth of a second
+                // note: we need this because the webview has to be loaded
+                // before it can listen to events we fire
+                await sleep(100)
```
damntrecky, line 32: Learning: There is some data from Nielsons Heuristics and 100-200ms is perceived as almost instant for humans. 300-500ms is an "easing" motion. This principles are for UX animations, but this is similar.

Screenshot 2024-03-14 at 10 58 38 AM

packages/core/src/amazonqGumby/chat/controller/controller.ts
damntrecky: Ignore: I see `JavaHomeNotSetError` extends `ToolkitError`

packages/core/src/amazonqGumby/commands.ts
dhasani23, line 32: Maybe increase to 250 or something to be safe

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -445,9 +471,19 @@ function installProjectDependencies(dependenciesFolder: FolderInfo) {
...
         })
-        throw new ToolkitError('Maven install error', { code: 'MavenInstallError' })
+        throw new ToolkitError(`Maven ${argString} error`, { code: 'MavenExecutionError' })
     } else {
-        transformByQState.appendToErrorLog(`${baseCommand} clean install succeeded`)
+        transformByQState.appendToErrorLog(`${baseCommand} ${argString} succeeded`)
+    }
+}
+
+function getMavenErrorCode(args: string[]) {
```
dhasani23, line 480: What's the benefit of making this a separate method?

packages/core/src/amazonqGumby/chat/controller/controller.ts
dhasani23: Yeah but still wondering in general why not just use ToolkitError directly. Something like `new ToolkitError("message", { code: "some code"})` or something

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -278,15 +278,19 @@ export class Messenger {
...
+                message = `None of your open projects are supported by Amazon Q Code Transformation. Currently, Amazon Q can only upgrade Java projects built on Maven.
+
+For more information, see the Amazon Q documentation.`
                 break
             case 'no-maven-java-project-found':
-                message =
-                    'None of your open Java projects are supported by Amazon Q Code Transformation. Currently, Amazon Q can only upgrade Java projects built on Maven. A pom.xml must be present in the root of your project to upgrade it. \n For more information, see the Amazon Q documentation.'
+                message = `None of your open Java projects are supported by Amazon Q Code Transformation. Currently, Amazon Q can only upgrade Java projects built on Maven. A pom.xml must be present in the root of your project to upgrade it.
+                    
+For more information, see the Amazon Q documentation.`
```
dhasani23, line 288: Link to a webpage here and above? Like you did below.

packages/core/src/amazonq/webview/ui/apps/gumbyChatConnector.ts
```diff
@@ -0,0 +1,195 @@
...
+            tabOpenInteractionType,
+        })
+    }
+
+    onTabRemove(tabID: string) {
+        this.sendMessageToExtension({
+            tabID: tabID,
+            command: 'tab-was-removed',
+            tabType: 'gumby',
+        })
```
arangatang, line 62: Are there constants to use, will lessen chances for typos down the line?

packages/core/src/amazonq/webview/ui/apps/gumbyChatConnector.ts
```diff
@@ -0,0 +1,195 @@
...
+            return
+        }
+
+        const answer: ChatItem = {
+            type: ChatItemType.AI_PROMPT,
+            body: messageData.message,
+            formItems: messageData.formItems,
+            buttons: messageData.formButtons,
+            followUp: undefined,
+            status: 'info',
```
arangatang, line 76: Constant?

packages/core/src/amazonq/webview/ui/apps/gumbyChatConnector.ts
```diff
@@ -0,0 +1,195 @@
...
+            await this.processChatMessage(messageData)
+            return
+        }
+
+        if (messageData.type === 'authNeededException') {
+            await this.processAuthNeededException(messageData)
+            return
+        }
+
+        if (messageData.type === 'chatInputEnabledMessage') {
```
arangatang, line 145: These `messageData.type` must all be defined somewhere, can we use those?

packages/core/src/amazonq/webview/ui/apps/gumbyChatConnector.ts
```diff
@@ -0,0 +1,195 @@
...
+            body: messageData.message,
+        })
+    }
+
+    // This handles messages received from the extension, to be forwarded to the webview
+    handleMessageReceive = async (messageData: any): Promise<void> => {
+        if (messageData.type === 'errorMessage') {
+            this.onError(messageData.tabID, messageData.message, messageData.title)
+            return
+        }
```
arangatang, line 133: This looks like a perfect place to use a switch

packages/core/src/amazonq/webview/ui/connector.ts
```diff
@@ -130,8 +135,8 @@ export class Connector {
 
     transform = (tabID: string): void => {
         switch (this.tabsStorage.getTab(tabID)?.type) {
-            default:
-                this.cwChatConnector.transform(tabID)
+            case 'gumby':
+                this.gumbyChatConnector.transform(tabID)
                 break
         }
```
arangatang, line 141: Is this a switch with one entry? Could we make it into an if?

packages/core/src/amazonq/webview/ui/connector.ts
```diff
@@ -348,4 +361,17 @@ export class Connector {
...
+        tabId: string,
+        messageId: string | undefined,
+        action: any,
+        eventId: string | undefined = undefined
+    ): void | undefined => {
+        switch (this.tabsStorage.getTab(tabId)?.type) {
+            case 'gumby':
+                this.gumbyChatConnector.onCustomFormAction(tabId, action)
+                break
+        }
```
arangatang, line 375: nit: Will we have >1 entry in this switch going forward? Otherwise maybe convert to if?

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -94,13 +94,24 @@ export const createMynahUI = (ideApi: any, featureDevInitEnabled: boolean, gumby
                         type: ChatItemType.ANSWER,
                         body: 'Authentication successful. Connected to Amazon Q.',
                     })
-                    mynahUI.updateStore(tabID, {
-                        promptInputDisabledState: false,
-                    })
+
+                    if (tabsStorage.getTab(tabID)?.type === 'gumby') {
```
arangatang, line 98: Constant for gumby.

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -94,13 +94,24 @@ export const createMynahUI = (ideApi: any, featureDevInitEnabled: boolean, gumby
...
             }
         },
         onFileActionClick: (tabID: string, messageId: string, filePath: string, actionName: string): void => {},
+        onQuickHandlerCommand: (tabID: string, command?: string, eventId?: string) => {
+            console.log(`onQuickHandlerCommandMessage!! command: ${command} tabID: ${tabID}`)
+            if (command === 'aws.awsq.transform') {
+                quickActionHandler.handle({ command: '/transform' }, tabID, eventId)
+            } else if (command === 'aws.awsq.clearchat') {
+                quickActionHandler.handle({ command: '/clear' }, tabID)
+            }
```
arangatang, line 113: Constants?

packages/core/src/amazonqGumby/chat/controller/controller.ts
JuliaABurch: I check the type of errors in a `try catch` block so that I can show relevant error messages in the chat. This is more personal preference - I find custom errors easier to read / handle

packages/core/src/amazonq/webview/ui/main.ts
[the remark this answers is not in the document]
JuliaABurch: Thought I caught all of my silly debug statements -- removed! Thanks for catching it

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
JuliaABurch: Removed an unnecessary transformation here

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
JuliaABurch: That's a good point about module/project -- I'll run this by Allie.

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
JuliaABurch: I'll run this by Allie!

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -162,7 +198,13 @@ export const createMynahUI = (ideApi: any, featureDevInitEnabled: boolean, gumby
                 return
             }
 
-            if (item.body !== undefined || item.relatedContent !== undefined || item.followUp !== undefined) {
+            if (
+                item.body !== undefined ||
```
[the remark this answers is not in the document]
JuliaABurch: I have a lot of muscle memory around dealing with the weird/unexpected way JS deals with truthiness - the issue with just doing `if (object)` _in general_ is that a lot of things can be coalesced into booleans -- for example, empty strings, 0, `null` and `undefined` all equate to false.

In this particular case, there actually are reasons why you may want to render an empty string if `item.body` is set to `''` -- this is often done to create a "blank" chat bubble that then gets an answer streamed character-by-character to it. So the _specificity_ of checking for undefined actually is relevant here. 

It's true that this isn't always the case -- the other properties you point out won't bring up any weird edge cases or behavior (they're arrays, not primitives) -- but I always lean towards being consistent and explicit

packages/core/src/amazonq/webview/ui/main.ts
JuliaABurch: Thanks for catching this - there's actually a different bug here where `eventId` was meant to be passed here but I failed to do so

packages/core/src/amazonq/webview/ui/quickActions/handler.ts
JuliaABurch: This is a good callout, and it's actually already done here! `tab.type` is a `type` called `TabType` (gotta love descriptive class names) which explicitly defines valid strings. If you did `tab.type === 'gumbyBadString'` it would fail at compile time.

I prefer the readability of enums myself

packages/core/src/amazonq/webview/ui/quickActions/handler.ts
JuliaABurch: Yes, it's a [literal type](https://www.typescriptlang.org/docs/handbook/literal-types.html). Like in the previous comment, I prefer enums, but I'm following an already established convention here, and the IDE will show you 'valid' values for `TabType` as you develop so there's very little difference between it and an enum beyond preference

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
JuliaABurch: Good callout - this is something I wanted to do anyway, will do in the next revision

packages/core/src/amazonqGumby/app.ts
JuliaABurch: Will check!

packages/core/src/amazonqGumby/commands.ts
JuliaABurch: Oh interesting, thank you for providing this!

packages/core/src/codewhisperer/service/transformByQHandler.ts
JuliaABurch: I'll re-review this / remove this -- in a previous version of the chat flow, `maven clean` was run separately from `maven install` so that it was easier to determine which specific step failed and indicate this to the user in chat. After discussions / maven command changes, we're no longer running them separately, so this method is something of a vestigial organ I think

packages/core/src/amazonq/webview/ui/apps/gumbyChatConnector.ts
JuliaABurch: Another literal type!

packages/core/src/amazonq/webview/ui/apps/gumbyChatConnector.ts
JuliaABurch: THESE are not -- I will make them so -- thank you for catching

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -0,0 +1,362 @@
...
+                message = 'To begin, please open a workspace.'
+                break
+            case 'no-project-found':
+            case 'no-java-project-found':
+                message = `None of your open projects are supported by Amazon Q Code Transformation. Currently, Amazon Q can only upgrade Java projects built on Maven.
+
+For more information, see the Amazon Q documentation.`
+                break
+            case 'no-maven-java-project-found':
+                message = `None of your open Java projects are supported by Amazon Q Code Transformation. Currently, Amazon Q can only upgrade Java projects built on Maven. A pom.xml must be present in the root of your project to upgrade it.
```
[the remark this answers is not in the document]
JuliaABurch: This error message is what shows when we've determined that you have valid Java projects (8 or 11) but none of these projects are built with Maven -- a different one shows if we've failed to detect appropriate Java projects

packages/core/src/amazonqGumby/chat/controller/messenger/stringConstants.ts
```diff
@@ -0,0 +1,23 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+export const enterJavaHomeMessage = 'Enter the path to JDK '
```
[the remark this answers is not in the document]
JuliaABurch: I'll run this by Allie!

packages/core/src/amazonqGumby/chat/views/actions/uiMessageListener.ts
```diff
@@ -0,0 +1,79 @@
...
+            case 'auth-follow-up-was-clicked':
+                this.authClicked(msg)
+                break
+            case 'form-action-click':
+                this.formActionClicked(msg)
+                break
+        }
+    }
+
+    private transform(msg: any) {
```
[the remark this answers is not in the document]
JuliaABurch: I have a general "go back and use explicit types everywhere I missed them" to-do, I will go through and do this

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -445,9 +471,19 @@ function installProjectDependencies(dependenciesFolder: FolderInfo) {
...
+    if (args.includes('install')) {
+        return 'Install'
+    } else if (args.includes('clean')) {
+        return 'Clean'
+    } else {
+        return ''
```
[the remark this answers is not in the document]
JuliaABurch: I'm pretty sure this method is no longer needed -- but you're right, this should have had a more explicit general/default case. Will either remove or fix

packages/core/src/amazonq/webview/ui/tabs/generator.ts
JuliaABurch: All of these strings are provided by Allie, but we'll have a final walkthrough with her so she can approve them in the context of the feature

packages/core/src/amazonq/webview/ui/connector.ts
JuliaABurch: I think this is a good suggestion, it may better convey that `/transform` isn't an action we ever expect another app to handle

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -0,0 +1,283 @@
...
+                session.isAuthenticating = true
+                return
+            }
+        } catch (err: any) {
+            this.messenger.sendErrorMessage(err.message, message.tabID)
+        }
+    }
+
+    private async tabClosed(data: any) {
+        transformByQState.setGumbyChatTabID(undefined)
```
mk-fan, line 104: We should delete the session as well?

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -0,0 +1,283 @@
...
+        const workspaceFolders = vscode.workspace.workspaceFolders
+        if (workspaceFolders === undefined || workspaceFolders.length === 0) {
+            this.messenger.sendStaticTextResponse('no-project-found', message.tabID)
+            return
+        }
+
+        // check that the session is authenticated
+        let session
+        try {
+            session = await this.sessionStorage.getSession(message.tabID)
```
mk-fan, line 137: Nit: session can be a const inside try scope.

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
damntrecky: Discussion: Modules are Maven terminology. Project is our terminology, similar to VSCode defines "project" as the 
root of a "workspace" -> https://code.visualstudio.com/docs/editor/workspaces

packages/core/src/amazonqGumby/commands.ts
damntrecky: Note: Looking at Nielsons Heuristics linked above, a "large move" or "complex easing" can be 200-500ms.

I would generalize a panel display is a "large" ux move.

packages/core/src/amazonqGumby/chat/storages/chatSession.ts
```diff
@@ -0,0 +1,38 @@
...
+        if (currentSession === undefined) {
+            // If a session doesn't already exist just create it
+            return this.createSession(tabID)
+        }
+        return currentSession
+    }
+
+    // Find all sessions that are currently waiting to be authenticated
+    public getAuthenticatingSessions(): Session[] {
+        return Array.from(this.sessions.values()).filter(session => session.isAuthenticating)
```
justinmk3, line 32: This entire file was copy-pasted from featureDev? https://github.com/aws/aws-toolkit-vscode/blob/46560e70b7dd29d83ff557b1db246db5091dd7fe/packages/core/src/amazonqFeatureDev/storages/chatSession.ts#L33

I don't see how it is sustainable that we keep copy-pasting modules everywhere.

packages/core/src/amazonqGumby/chat/views/actions/uiMessageListener.ts
```diff
@@ -0,0 +1,79 @@
...
+import { MessageListener } from '../../../../amazonq/messages/messageListener'
+import { ExtensionMessage } from '../../../../amazonq/webview/ui/commands'
+import { ChatControllerEventEmitters } from '../../controller/controller'
+
+export interface UIMessageListenerProps {
+    readonly chatControllerEventEmitters: ChatControllerEventEmitters
+    readonly webViewMessageListener: MessageListener<any>
+}
+
+export class UIMessageListener {
```
justinmk3, line 15: This is another thing that was copy-pasted from featureDev? https://github.com/aws/aws-toolkit-vscode/blob/46560e70b7dd29d83ff557b1db246db5091dd7fe/packages/core/src/amazonqFeatureDev/views/actions/uiMessageListener.ts#L15

2000 lines of code. It's very hard to review PRs when so much code is copy-pasted. And then it's hard in the future to maintain the project.

packages/core/src/amazonqGumby/chat/session/session.ts
```diff
@@ -0,0 +1,15 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Messenger } from '../controller/messenger/messenger'
+
+export class Session {
+    // Used to keep track of whether or not the current session is currently authenticating/needs authenticating
+    public isAuthenticating: boolean = false
```
justinmk3, line 10: How is this different than the other similar flags ?  E.g. https://github.com/aws/aws-toolkit-vscode/blob/46560e70b7dd29d83ff557b1db246db5091dd7fe/packages/core/src/amazonqFeatureDev/session/session.ts#L37

Can we use auth.ts or secondaryAuth.ts as the source of truth, instead of storing all of this shared state redundantly in many places? https://github.com/aws/aws-toolkit-vscode/blob/46560e70b7dd29d83ff557b1db246db5091dd7fe/packages/core/src/auth/auth.ts

packages/core/src/amazonqGumby/chat/storages/chatSession.ts
JuliaABurch: This file has gone under significant change, as a code transformation session does conceptually need to track different things than FeatureDev, which should alleviate your concern about copy/pasting code. Nothing in this file impacts global state in a way that should impact other features.

I agree that there is a better way to handle authentication -- there are patterns introduced in CodeWhisperer that were replicated by FeatureDev and now by Gumby, where everyone would benefit by having common classes to use. There's no base "chatTab" module that handles this sort of thing, thus the code duplication.

Right now, all 3 use a pattern where they poll for the current global state at specific points in the chat flow (when the chat opens, when a user interaction would cause an interaction with the backend, etc.) _If_ they see that the global state is 'unauthenticated',  all 3 show an error message to the user, update their local state to `isAuthenticating`, and then lock the tab. They then wait for an event to be fired by the global auth handler so that they can unlock the tab -- in the meantime, they require that local 'isAuthenticating' state to prevent certain user interactions.

Unfortunately I don't have much insight into this pattern beyond that ): I'd like to see if this entire pattern could be event-driven rather than partially so, but it would require introducing further complexity, and making changes to other teams' code -- this would be the sort of thing I'd want to design and get buy-in ahead of time for rather than trying to rush it in to launch a feature.

In the meantime, I will see if I can extract common behaviors to a utility module that can be shared between the apps!

packages/core/src/amazonqGumby/chat/views/actions/uiMessageListener.ts
JuliaABurch: Like I said in the previous comment, I'll see if I can create a base class that both FeatureDev and Gumby can inherit from. This is one of those issues where FeatureDev needs to build a yellow house with 3 walls and 2 railings, and Gumby needs to build a red house with a garage and 6 railings, so there is going to be very similar looking code regardless.

packages/core/src/amazonqGumby/chat/session/session.ts
JuliaABurch: A gumby tab has no way of accessing a featureDev tab's session state. It only queries the global state through `authUtil.ts` (which in turn uses `auth.ts` and `secondaryAuth.ts`).  There may be a way to extract this to a common base class in the meantime though, which I'll puzzle over.

packages/core/src/amazonq/webview/ui/connector.ts
JuliaABurch: I could see other apps needing to use buttons in the future, so i'll keep this as a switch

packages/core/src/amazonq/webview/ui/quickActions/handler.ts
```diff
@@ -45,16 +47,71 @@ export class QuickActionHandler {
...
                 break
             case '/clear':
                 this.handleClearCommand(tabID)
                 break
         }
     }
 
-    private handleGumbyCommand(tabID: string) {
-        this.connector.transform(tabID)
+    private handleGumbyCommand(tabID: string, eventId: string | undefined) {
```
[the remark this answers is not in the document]
JuliaABurch: No, both of these are essentially randomly generated UUIDs

packages/core/src/amazonqGumby/chat/storages/chatSession.ts
justinmk3: > I agree that there is a better way to handle authentication -- there are patterns introduced in CodeWhisperer that were replicated by FeatureDev and now by Gumby, where everyone would benefit by having common classes to use. 

Yes, exactly. 

> this would be the sort of thing I'd want to design and get buy-in ahead of time for rather than 

Ok, can you create a back log issue?

> trying to rush it in to launch a feature.

Sure, but we made similar compromises for re:Invent. We need to pay down this tech debt soon.

> I will see if I can extract common behaviors to a utility module that can be shared between the apps!

Thank you!

packages/core/src/amazonqGumby/app.ts
JuliaABurch: This specifically will be addressed in the implementation of a backlog item that looks at authentication strategy as a whole

packages/core/src/amazonqGumby/errors.ts
```diff
@@ -0,0 +1,37 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+import { ToolkitError } from '../shared/errors'
+
+export class JavaHomeNotSetError extends ToolkitError {
```
damntrecky, line 9: Possible merge blocker: Based on some of @dhasani23 changes, we might need to extend `Error` Instead of `ToolkitError`

damntrecky (approved):

packages/core/src/amazonqGumby/errors.ts
dhasani23: This should be fine

dhasani23 (approved):

.changes/next-release/Feature-f0522fe2-1aa9-44fb-aa35-81aa9e78b52c.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Code Transformation runs through Q Chat"
```
justinmk3, line 3: In any public materials, "Q" is always "Amazon Q".

```suggestion
	"description": "Code Transformation runs through Amazon Q Chat"
```

PR #4541, opened 2024-03-14
fix(debugging): watch core lib during debugging, support hot reloads

hayemaxi:
Problem: Due to the project refactor into a core library and separate extension packages, debugging no longer works as expected. Starting the extension in debug mode does not pick up changes in the core lib. You also cannot hot reload the debugger, so you have to kill the debugger and all the tasks manually.

Solution: Implement a fast recompile for the core lib and watch on changes during debugging. Also, adjust the tasks so that you can restart the debugging window instantly with your changes.

Notes:
- Do not use the reload button in the VSCode debug bar, it seems to not properly reconnect to the window. Either close out the test instance and re-run debugging (should be instant) or reload the window via command palette in the debug instance.
- Changes do not apply to web debugging yet.
- You don't need to kill the tasks manually anymore, you can just start debugging up again.
- We up the memory to 5gb for running serve as we are running out of memory when the window is reloaded during debugging.

## License


packages/toolkit/.vscode/tasks.json
```diff
@@ -101,12 +145,18 @@
...
+            "presentation": {
+                "close": true
             }
         },
         {
-            "label": "terminate",
+            "label": "extensionTerminate",
             "command": "echo run terminate",
             "type": "shell",
+            "presentation": {
```
nkomonen-amazon, line 157: Question: what does this do?

hayemaxi: closes the terminal after it's done instead of leaving it open for other processes.

nkomonen-amazon (approved):

PR #4544, opened 2024-03-15
feat(toolkit): Expose ListConnections API in AWS Toolkit

leigaol:
## Problem

Other extensions(Especially Amazon Q) need a way to fetch the connections of AWS Toolkit. 

## Solution

Expose the listConnections API. 

By fetching the connections, Amazon Q can do a `AuthUtil.instance.useConnection()` to use one of the returned connection. 

If there is a need for other extensions to use the toolkit connection, the API for getCredentials can be exposed later. 

## License


packages/toolkit/src/api.ts
```diff
@@ -0,0 +1,15 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Auth, Connection } from 'aws-core-vscode/auth'
+/*!
```
hayemaxi, line 7:
```suggestion
/**
```

jpinkney-aws (approved):

.changes/next-release/Feature-b17cee45-65a7-4dea-a358-6d4a4c9ecf67.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Expose listConnections API in AWS Toolkit"
```
justinmk3, line 3: Please mention it in the docs: https://github.com/aws/aws-toolkit-vscode/blob/master/docs/api.md

packages/toolkit/src/api.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+import { Auth, Connection } from 'aws-core-vscode/auth'
+/**
+ * Exposing listConnections API for other extension to read or re-use
+ * the available connections in aws toolkit.
+ */
+export const awsToolkitApi = {
+    async listConnections(): Promise<Connection[]> {
+        return Auth.instance.listConnections()
```
justinmk3, line 13: Returning the internal `Connection` object is not how this should be exposed. There should be an explicit interface, to avoid accidentally exposing internal things. It may also need to be serializable (not sure)?

> If there is a need for other extensions to use the toolkit connection, the API for getCredentials can be exposed later.

`Connection` already exposes `getCredentials`: https://github.com/aws/aws-toolkit-vscode/blob/7c77f0ec164c906e4a298f09604a77e7e6f26434/packages/core/src/auth/connection.ts#L109-L112

We should not be returning internal objects directly. The public API should only expose things we explicitly want to expose. And in particular, the reason for separating `listConnections` from `getCredentials` is so that we _don't_ have to build an authorization model around `listConnections`, while leaving the door open for `getCredentials` in the future (which may require an authorization model, e.g. asking the user to "confirm that extension X is allowed to use credentials Y").

PR #4545, opened 2024-03-15
fix(cwspr): Fix cw api call failure in sm code editor

leigaol:
## Problem

In Sage Maker Code Editor, CWSPR should create request in type GenerateRecommendation for SigV4 API invocation. Otherwise there will be a Unexpected key 'optOutPreference' found in params

## Solution

Add extra if else for Sage Maker Code Editor to let the recommendation Handler use GenerateRecommendation Request payload instead of using the ListRecommendation. 

Tested in SageMaker Code Editor and it works. 

## License

### DIFF
--- a/.changes/next-release/Bug Fix-d70cbb8e-491a-439e-9cbb-b155431e3743.json
+++ b/.changes/next-release/Bug Fix-d70cbb8e-491a-439e-9cbb-b155431e3743.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Fix CodeWhisperer API call failure in SageMaker Code Editor"
+}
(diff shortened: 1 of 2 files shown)


Will-ShaoHua:
@leigaol can we move the logic to `RecommendationService` instead?

leigaol:
> @leigaol can we move the logic to `RecommendationService` instead? From the description, it seems we just need SM case to run C9 code path, isnt' it the case?

No. C9 uses intelliSense UX while SM CodeEditor uses inline completion. They are in different code path. 

THis is why you see isSM condition has been defined in `recommendationHandler` in previous PRs.

justinmk3 (approved):

andrewyuq (approved):

PR #4550, opened 2024-03-18
fix(CodeTransformation): edit notifications, show feedback prompt

dhasani23:
## Problem

- Needed to review exact strings with our technical writer
- Needed show users a feedback prompt once job is complete
- Needed to increase `maxBufferSize` since it occasionally overflows according to our telemetry.

## Solution

Implement the above. I increased the `maxBufferSize` by a factor of 8. There's a package here (https://www.npmjs.com/package/execa) that increases it by a factor of 100 (from 1MB to 100MB), but going to 8MB should be sufficient for us.

## License


packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -76,7 +77,7 @@ async function collectInput(validProjects: Map<vscode.QuickPickItem, JDKVersion
     transformByQState.setTargetJDKVersion(JDKVersion.JDK17)
     await MultiStepInputFlowController.run(input => pickProject(input, state, validProjects))
     if (!state.project) {
-        throw new ToolkitError('No project selected', { code: 'NoProjectSelected' })
+        throw new ToolkitError('', { code: 'NoProjectSelected' })
```
dhasani23, line 80: You'll see in a lot of places I deleted the message here, because it was showing a notification, but we already have notifications handled, so these were duplicates. Should I be throwing a regular Error instead perhaps?

packages/core/src/codewhisperer/service/transformationHubViewProvider.ts
```diff
@@ -178,11 +176,6 @@ export class TransformationHubViewProvider implements vscode.WebviewViewProvider
                     }
                 }
             }
-            if (planProgress['transformCode'] === StepProgress.Succeeded) {
-                progressHtml += `<p> ${this.getProgressIconMarkup(
-                    planProgress['returnCode']
-                )} Validate and prepare proposed changes</p>`
-            }
```
dhasani23, line 185: IntelliJ does not show this, so removed it to match.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
nkomonen-amazon: These shouldn't be empty strings. The handler should be catching these errors and then propagating the message to the user. Is there something I'm missing with how it works?

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -69,30 +69,25 @@ export async function validateAndLogProjectDetails() {
...
     path: string
     name: string
 }
 
-/* Once supported in all browsers and past "experimental" mode, use Intl DurationFormat:
- * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat#browser_compatibility
- * Current functionality: given number of milliseconds elapsed (ex. 4,500,000) return hr / min / sec it represents (ex. 1 hr 15 min)
- */
+// Given number of milliseconds elapsed (ex. 4,500,000) return hr / min / sec it represents (ex. 1 hr 15 min)
 export function convertToTimeString(durationInMs: number) {
```
nkomonen-amazon, line 78: nit: Time stuff should be in a shared module if it does not exist already

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -702,7 +725,7 @@ export async function startJob(uploadId: string) {
             },
         })
         if (response.$response.requestId) {
-            transformByQState.setJobFailureMetadata(`(request ID: ${response.$response.requestId})`)
+            transformByQState.setJobFailureMetadata(` (request ID: ${response.$response.requestId})`)
```
nkomonen-amazon, line 728: Did you want a space?

dhasani23: Yes

packages/core/src/codewhisperer/commands/startTransformByQ.ts
dhasani23: Fixed to use `Error`

packages/core/src/codewhisperer/models/constants.ts
```diff
@@ -280,66 +280,97 @@ export const newCustomizationMessage = 'You have access to new CodeWhisperer cus
...
+export const transformByQCompletedMessage =
+    'Amazon Q successfully transformed your code. You can download a summary of the transformation and a diff with proposed changes in the Transformation Hub.'
 
-export const transformByQPartiallyCompletedMessage = 'Transformation partially completed'
+export const transformByQPartiallyCompletedMessage =
+    'Amazon Q partially transformed your code. You can download a summary of the transformation and a diff with proposed changes in the Transformation Hub.'
 
 export const noPomXmlFoundMessage =
-    'None of your open Java projects are supported by Amazon Q Code Transformation. We were unable to find a pom.xml in any of your Java projects. We only support Java projects built on Maven at the moment. Please see: https://docs.aws.amazon.com/amazonq/latest/aws-builder-use-ug/code-transformation.html#prerequisites'
+    'None of your open Java projects are supported by Amazon Q Code Transformation. Currently, Amazon Q can only upgrade Java projects built on Maven. A pom.xml must be present in the root of your project to upgrade it. For more information, see the [Amazon Q documentation](LINK_HERE).'
```
damntrecky, line 320: Not a blocker: I don't see the find/replace as a problem. Just wondering if there are examples of "delimiters" for this in the repo. Eg) `<replaceMe>` or `__link_here__`. Just something to help identify the find/replace better.

I think if this was dynamic it may be important. Thoughts?

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -362,22 +365,22 @@ export async function pollTransformationStatusUntilComplete(jobId: string) {
...
 }
 
 export async function finalizeTransformationJob(status: string) {
     if (!(status === 'COMPLETED' || status === 'PARTIALLY_COMPLETED')) {
-        const errorMessage = 'Failed to complete transformation'
+        const errorMessage = CodeWhispererConstants.failedToCompleteJobMessage
         getLogger().error(`CodeTransformation: ${errorMessage}`)
         sessionPlanProgress['transformCode'] = StepProgress.Failed
         transformByQState.setJobFailureErrorMessage(errorMessage)
-        throw new ToolkitError(errorMessage, { code: 'JobDidNotSucceed' })
```
damntrecky, line 380: Question: We are ditching the ToolkitError because that forces a notification to pop up?

What other side effects are there from removing ToolkitError? Any logging concerns?

dhasani23: Correct that forces a notification to pop up. Not aware of any other side effects, we already handle notifications, telemetry, and logging ourselves everywhere.

packages/core/src/shared/utilities/textUtilities.ts
```diff
@@ -310,3 +310,30 @@ export function sanitizeFilename(input: string, replaceString = '_'): string {
             )
     )
 }
+
+// Given number of milliseconds elapsed (ex. 4,500,000) return hr / min / sec it represents (ex. "1 hr 15 min")
+export function convertToTimeString(durationInMs: number) {
```
nkomonen-amazon, line 315: ty for putting in this module

nkomonen-amazon (approved):

packages/core/src/codewhisperer/commands/startTransformByQ.ts
nkomonen-amazon: No major concerns with ditching the ToolkitError. The message the user sees will not be formatted as nicely but it will still be logged. For reference, [this handles displaying Errors](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/extensionShared.ts#L203)

damntrecky: I think a side effect is `telemetry_1.telemetry.ui_click.run(() => {` the `.run` telemetry event picks up `ToolkitError` and sets fields automatically.

We currently don't use `.run` in our product.

damntrecky (approved):

packages/core/src/shared/utilities/textUtilities.ts
```diff
@@ -310,3 +310,30 @@ export function sanitizeFilename(input: string, replaceString = '_'): string {
...
+        timeString = `${minutes} min ${timeString}`
+    }
+    if (hours > 0) {
+        timeString = `${hours} hr ${timeString}`
+    }
+    return timeString
+}
+
+// Given Date object, return timestamp it represents (ex. "01/01/23, 12:00 AM")
+export function convertDateToTimestamp(date: Date) {
```
justinmk3, line 331: Any reason not to use the existing `formatDateTimestamp` in this same file? Ideally we would show consistent formatting in most features in Q/Toolkit.

PR #4551, opened 2024-03-18
feat(logging): log command args

justinmk3:
Problem:
Log message "command: running with arguments …" does not print useful argument info.

Solution:
Print objects at depth of 1.

before:

    2024-03-18 18:10:21 [DEBUG]: command: running "aws.cdk.renderStateMachineGraph" with arguments [[object Object], undefined]

after:

    2024-03-18 13:15:20 [DEBUG]: command: running "_aws.auth.reauthenticate" with arguments: {
      '0': Auth {
        ...
      },
      '1': {
        ...
      }
    }

### DIFF
--- a/packages/core/src/auth/sso/clients.ts
+++ b/packages/core/src/auth/sso/clients.ts
@@ -24,7 +24,7 @@ import {
     StartDeviceAuthorizationRequest,
 } from '@aws-sdk/client-sso-oidc'
 import { AsyncCollection } from '../../shared/utilities/asyncCollection'
-import { pageableToCollection } from '../../shared/utilities/collectionUtils'
+import { pageableToCollection, partialClone } from '../../shared/utilities/collectionUtils'
 import { assertHasProps, isNonNullable, RequiredProps, selectFrom } from '../../shared/utilities/tsUtils'
 import { getLogger } from '../../shared/logger'
 import { SsoAccessTokenProvider } from './ssoAccessTokenProvider'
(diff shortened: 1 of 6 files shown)


packages/core/src/ssmDocument/ssm/ssmClient.ts
```diff
@@ -102,7 +102,7 @@ export async function activate(extensionContext: ExtensionContext) {
     // Create the language client and start the client.
     const client = new LanguageClient(
         'ssm',
-        localize('ssm.server.name', 'SSM Document Language Server'),
+        localize('ssm.server.name', 'AWS: SSM Document Language Server'),
```
justinmk3, line 105: Use a common prefix for all Toolkit channels...

image

nkomonen-amazon (approved):

packages/core/src/shared/vscode/commands2.ts
```diff
@@ -622,13 +622,15 @@ export const defaultTelemetryThrottleMs = 300_000 // 5 minutes
...
     const logger = logging ? getLogger() : new NullLogger()
-    const withArgs = args.length > 0 ? ` with arguments [${args.map(String).join(', ')}]` : ''
     const threshold = info.telemetryThrottleMs ?? defaultTelemetryThrottleMs
     const instrumenter = logging
         ? getInstrumenter({ id, args, compositeKey: compositeKey ?? {} }, threshold || 0, info.telemetryName)
         : undefined
 
-    logger.debug(`command: running ${label}${withArgs}`)
+    logger.debug(
+        `command: running ${label} with arguments: %O`,
```
nkomonen-amazon, line 631: question: is there somewhere that shows the available formatting symbols (%s, %O)?

justinmk3: It's part of `console.log` and node's `util.inspect()`. This seems like the best reference: https://console.spec.whatwg.org/#formatting-specifiers

packages/core/src/auth/sso/clients.ts
```diff
@@ -199,27 +199,17 @@ export class SsoClient {
...
             if (HttpRequest.isInstance(args.request)) {
                 const { hostname, path } = args.request
-                const input = omitIfPresent(
+                const input = partialClone(
                     // TODO: Fix
                     args.input as unknown as Record<string, unknown>,
-                    'clientSecret',
-                    'accessToken',
-                    'refreshToken'
+                    3,
```
justinmk3, line 210: Replaced `omitIfPresent` with a more generic `partialClone` function. It has some nice features which are often needed.

Ideally we would just use node's `util.inspect()`, but
- don't know a good browser-compatible polyfill
- only produces strings, not "partial objects"

PR #4553, opened 2024-03-18
auth: log errors when we go from valid to invalid state

nkomonen-amazon:
## Problem:
When errors occur somewhere in the getToken() process + the token was previously valid, if it fails we do not log the actual underlying error anywhere. We simply tell them to reauth.

This caused issues when debugging premature auth expiration problems since we couldn't see the actual error.

## Solution
If we switch from valid to invalid connection state, then log the error as a warning. This should give us the reason as to why it is failing, though we will also see the expected case where it actually expired.

## License

### DIFF
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -754,6 +754,9 @@ export class Auth implements AuthService, ConnectionManager {
             })
         }
         if (previousState === 'valid') {
+            // Non-token expiration errors can happen. We must log it here, otherwise they are lost.
+            getLogger().warn(`auth: valid connection became invalid. Last error: %s`, this.#validationErrors.get(id))
+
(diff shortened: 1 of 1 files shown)


packages/core/src/auth/auth.ts
```diff
@@ -754,6 +754,12 @@ export class Auth implements AuthService, ConnectionManager {
             })
         }
         if (previousState === 'valid') {
+            // Non-token-expiration errors can happen, we want to show this in the logs
+            getLogger().warn(
+                `auth: valid connection became invalid. If this was not expected there may be a problem: %s`,
```
justinmk3, line 759:
```suggestion
                `auth: valid connection became invalid. Last error: %s`,
```

justinmk3 (approved):

PR #4556, opened 2024-03-19
fix: featureDev partial code acceptance

dianaabv16:
## Problem
Bug:
Rejected files were ignored when generated code was accepted.

Fix:
Rejected files are excluded when generated code was accepted.

Testing:
I added a test to ensure the `fileClick` action correctly updates the state of the file, including rejecting files and reverting rejections. 
Additionally, I manually tested the most common scenarios, such as accepting all files, rejecting all files, and accepting a subset of them. All cases functioned as expected.

## License

### DIFF
--- a/.changes/next-release/Bug Fix-c76a701b-a106-46ef-99d9-5bb0760c469a.json
+++ b/.changes/next-release/Bug Fix-c76a701b-a106-46ef-99d9-5bb0760c469a.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "FeatureDev: Rejected files are excluded when generated code is accepted by the customer"
+}
--- a/packages/core/src/amazonqFeatureDev/session/session.ts
+++ b/packages/core/src/amazonqFeatureDev/session/session.ts
@@ -166,7 +166,7 @@ export class Session {
     }
 
     public async insertChanges() {
-        for (const filePath of this.state.filePaths ?? []) {
+        for (const filePath of this.state.filePaths?.filter(i => !i.rejected) ?? []) {
             const absolutePath = path.join(filePath.workspaceFolder.uri.fsPath, filePath.relativePath)
 
             const uri = filePath.virtualMemoryUri
--- a/packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
+++ b/packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
@@ -9,14 +9,15 @@ import * as path from 'path'
 import sinon from 'sinon'
 import { waitUntil } from '../../../../shared/utilities/timeoutUtils'
 import { ControllerSetup, createController, createSession } from '../../utils'
-import { FollowUpTypes, createUri } from '../../../../amazonqFeatureDev/types'
+import { CurrentWsFolders, FollowUpTypes, createUri } from '../../../../amazonqFeatureDev/types'
 import { Session } from '../../../../amazonqFeatureDev/session/session'
 import { Prompter } from '../../../../shared/ui/prompter'
 import { assertTelemetry, toFile } from '../../../testUtil'
 import { SelectedFolderNotInWorkspaceFolderError } from '../../../../amazonqFeatureDev/errors'
-import { PrepareRefinementState } from '../../../../amazonqFeatureDev/session/sessionState'
+import { CodeGenState, PrepareRefinementState } from '../../../../amazonqFeatureDev/session/sessionState'
 import { FeatureDevClient } from '../../../../amazonqFeatureDev/client/featureDev'
 
+let mockGetCodeGeneration: sinon.SinonStub
 describe('Controller', () => {
     const tabID = '123'
     const conversationID = '456'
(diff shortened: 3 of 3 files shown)


jpinkney-aws (approved):

dianaabv16:
> Can you add a changelog item

Isn't it add automatically?

jpinkney-aws (approved):

packages/core/src/amazonqFeatureDev/session/session.ts
```diff
@@ -166,7 +166,7 @@ export class Session {
     }
 
     public async insertChanges() {
-        for (const filePath of this.state.filePaths ?? []) {
+        for (const filePath of this.state.filePaths?.filter(i => !i.rejected) ?? []) {
```
vishaalmehrishi, line 169: Can we add a test for this specific scenario as well? A test will catch any future regressions for something that is important functionality

dianaabv16: I will add it as a follow up, I want to merge it as soon as possible to mitigate the bug

vishaalmehrishi: The tests are supposed to provide confidence that the bug was mitigated. Have we done some manual testing to ensure the fix works? I don't see any testing details in the PR description.

dianaabv16: my bad, updated description. Yes, I manually tested. I will add a test in this PR

dianaabv16: updated descriptions, added tests

packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
```diff
@@ -242,4 +243,84 @@ describe('Controller', () => {
...
+                action,
+            })
+
+            // Wait until the controller has time to process the event
+            await waitUntil(() => {
+                return Promise.resolve(getSessionStub.callCount > 0)
+            }, {})
+            return getSessionStub.getCall(0).returnValue
+        }
+        it('This test case verifies that when a customer clicks on the "Reject File" button, the state of the file is updated correctly to "rejected: true".', async () => {
```
vishaalmehrishi, line 310: These tests verify that the file state is correctly set to rejected, but this alone is not enough for the feature to work correctly - it is crucial that the rejected files are *actually ignored* when code is inserted.

This key logic is being updated in `session.ts/insertChanges()`

```
for (const filePath of this.state.filePaths?.filter(i => !i.rejected) ?? []) {
```

the tests should cover this updated logic as well

dianaabv16: I am working on this test.
I will add it a follow up

vishaalmehrishi: As discussed at standup, it is important to get this fix out to customers in the weekly VSCode Toolkit release today. Approving this with the understanding that the test will be added in a follow-up PR.

vishaalmehrishi (approved):

jpinkney-aws (approved):

.changes/next-release/Bug Fix-c76a701b-a106-46ef-99d9-5bb0760c469a.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "FeatureDev: Rejected files are excluded when generated code is accepted by the customer"
```
justinmk3, line 3: does "FeatureDev" mean anything to customers? In the future I would word it as "Amazon Q FeatureDev", or something like that.

packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
```diff
@@ -242,4 +243,84 @@ describe('Controller', () => {
...
+            const session = await createCodeGenState()
+            const getSessionStub = sinon.stub(controllerSetup.sessionStorage, 'getSession').resolves(session)
+
+            const rejectFile = await fileClicked(getSessionStub, 'reject-change')
+            assert.strictEqual(rejectFile.state.filePaths?.find(i => i.relativePath === filePath)?.rejected, true)
+        })
+        it('This test case verifies that when a customer clicks on the "Reject File" button and then clicks on the "Revert Reject File" button the state of the file is updated correctly to "rejected: false".', async () => {
```
justinmk3, line 317: This is an unnecessarily verbose test name. `it()` is already named so that the test case can just describe what's being tested. Test names do not need to be paragraphs long. Also "correctly" is redundant, because tests are assumed to test for "correct" behavior.

Here is a shorter name that has the same information:

    it('clicking the "Reject File" button and then "Revert Reject File", updates the file state to "rejected: false".', async () => {

.changes/next-release/Bug Fix-c76a701b-a106-46ef-99d9-5bb0760c469a.json
dianaabv16: got it thanks

PR #4557, opened 2024-03-19
feat(CodeCatalyst): handle partial scope expiration

aarolieb:
## Problem
Extended session duration (90 days) for CodeWhisperer is coming soon, but other scopes on the same SSO connection will not be affected, so there must be handling for "partial" expiration. Partial expiration is relevant when an SSO connection is authorized for both Q/CodeWhisperer and CodeCatalyst.

## Solution
When an AccessDeniedException (ADE) is encountered when making a CodeCatalyst call, it will clear the bearer token. When a bearer token is refreshed, it will call `CodeCatalystClientInternal.verifySession()` before making further calls. If the token was able to refresh, this means that the entire token was not expired, and that specific scopes are expired.

1. If `verifySession` fails due to an ADE, fire an EventEmitter. EventEmitters were used to prevent circular dependencies between `CodeCatalystAuthenticationProvider` and `CodeCatalystClientInternal`. I think it might be better long-term to have the client consume the auth provider, but it would have required a decent amount of refactoring.
2.  `CodeCatalystAuthenticationProvider` subscribes to this event, handling the ADE. It will determine if the scope is partially expired, or if all scopes in the connection are expired. It will make a call to CodeWhisperer to determine this.
3. The CodeCatalyst context will be updated, displaying a "Re-Authorize" node in the explorer tree that matches the existing behavior in CodeWhisperer.
4. An information box will display informing the user that CodeCatalyst is expired. If the connection is partially expired, then it will also mention that CodeWhisperer is still useable. This notification will only display at most once per session, and has a "Don't show again" option.
6. On CC activation, check if the connection is shared with CodeWhisperer, and trigger a `verifySession` if it is. This shouldn't impact activation time of users that are not using CC and CW on the same connection.

### Additional Changes
1. Moved `isInDevEnv` from `codecatalyst/utils` to `shared/vscode/env` to prevent circular dependencies.
2. Call `verifySession` during CodeCatalyst activation to detect scope expiration on startup
3. Created `showReauthenticateMessage` in `utils/messages.ts` to share CodeWhisperer and CodeCatalyst reauth behavior

### UX Impact
When the user is not using CodeWhisperer and CodeCatalyst on the same SSO profile, there will not be a UX impact. When the connection is still "valid" (able to refresh the access token), but code catalyst receives an access denied exception while attempting to call `verifySession`, the CC scope is considered expired. Anytime CC scope is expired, it will update the CodeCatalyst explorer to show this. If a user action triggered the CC call, a notification will appear telling the user to authenticate. If the connection was previously expired when starting VSCode, there will not be a notification.

### Testing

1. Added debug commands to expire and un-expire tokens instantly by changing the bearerToken to an invalid token during request signing.
2. Performed testing on a feature-enabled IdC service with the upcoming extended duration changes.
2. Verified that all behavior is expected for just CodeWhisperer, CodeWhisperer & CodeCatalyst, and just CodeCatalyst

### Screenshots

CodeCatalyst explorer after expiration
explorer

Partial expiration notification
notification

## License


packages/core/src/codecatalyst/activation.ts
```diff
@@ -37,6 +37,17 @@ export async function activate(ctx: ExtContext): Promise<void> {
...
     await authProvider.restore()
 
+    // if connection is shared with CW, check if CC scopes are expired
+    // set suppressNextReauthPrompt to true to prevent notification on startup
+    const conn = authProvider.activeConnection
+    if (conn && authProvider.getConnectionState(conn).shared) {
+        try {
+            authProvider.suppressNextReauthPrompt = true
+            await createClient(conn)
+            authProvider.suppressNextReauthPrompt = false
```
justinmk3, line 47: setting globals is fragile and is a less obvious interface for the operation of the auth subsystem. can this be plumbed through `createClient` instead?

```suggestion
            await createClient(conn, { suppressReauthPrompt = true })
```

justinmk3:
Rebasing should fix some of the lint errors. Other lint errors are from this PR.

packages/core/src/shared/clients/codecatalystClient.ts
```diff
@@ -175,7 +178,16 @@ export async function createClient(
...
-    await c.verifySession()
+    try {
+        await c.verifySession()
+    } catch (e) {
+        if (!(e instanceof ToolkitError) || e.code !== 'AccessDeniedException') throw e
+        onAccessDeniedExceptionEmitter.fire()
+
+        // We "cancel" this error so that it prevents further execution,
+        // but doesn't display the error
+        throw new ToolkitError('CodeCatalyst scope is expired', { code: 'ScopeExpiration', cancelled: true })
```
justinmk3, line 189: Is there a reason not to throw a `CancellationError`? https://github.com/aws/aws-toolkit-vscode/blob/2c4ea6516c5249c048b301d5826f4b0cfe16bf0e/packages/core/src/auth/sso/ssoAccessTokenProvider.ts#L188

packages/core/src/shared/clients/codecatalystClient.ts
```diff
@@ -175,7 +178,16 @@ export async function createClient(
 ): Promise<CodeCatalystClient> {
     const sdkClient = await createCodeCatalystClient(connection, regionCode, endpoint, retryOptions)
     const c = new CodeCatalystClientInternal(connection, sdkClient)
-    await c.verifySession()
+    try {
+        await c.verifySession()
+    } catch (e) {
+        if (!(e instanceof ToolkitError) || e.code !== 'AccessDeniedException') throw e
```
justinmk3, line 184: should we also check `isAwsError`? https://github.com/aws/aws-toolkit-vscode/blob/e5fdf4109207f801a9b91a4d162ef1fb8d1bffd1/packages/core/src/codecatalyst/auth.ts#L314

packages/core/src/shared/clients/codecatalystClient.ts
```diff
@@ -231,13 +243,14 @@ class CodeCatalystClientInternal {
     private async call<T>(req: AWS.Request<T, AWS.AWSError>, silent: boolean, defaultVal?: T): Promise<T> {
         const log = this.log
         const bearerToken = (await this.connection.getToken()).accessToken
-        req.httpRequest.headers['Authorization'] = `Bearer ${bearerToken}`
+
```
justinmk3, line 246: > 3\. Removed the redundant bearerToken header in `CodeCatalystClientInternal.call()` since the token is added during the signing of the request in the aws-cdk

do you mean "aws-sdk"? perhaps this change should be a separate commmit, it's a bit subtle.

packages/core/src/shared/clients/codecatalystClient.ts
```diff
@@ -175,7 +178,16 @@ export async function createClient(
...
     const c = new CodeCatalystClientInternal(connection, sdkClient)
-    await c.verifySession()
+    try {
+        await c.verifySession()
+    } catch (e) {
+        if (!(e instanceof ToolkitError) || e.code !== 'AccessDeniedException') throw e
+        onAccessDeniedExceptionEmitter.fire()
+
+        // We "cancel" this error so that it prevents further execution,
+        // but doesn't display the error
```
justinmk3, line 188: The wording is confusing. We aren't cancelling an error, but cancelling execution, right?
```suggestion
        // Throw a "cancel" error to prevent further execution.
```

packages/core/src/shared/clients/codecatalystClient.ts
```diff
@@ -191,7 +203,7 @@ function fixAliasInRequest<
     return request
 }
 
-class CodeCatalystClientInternal {
+export class CodeCatalystClientInternal {
```
justinmk3, line 206: does this need to be exported? I don't see a use in the rest of the PR
```suggestion
class CodeCatalystClientInternal {
```

packages/core/src/codecatalyst/explorer.ts
```diff
@@ -44,7 +44,18 @@ async function getLocalCommands(auth: CodeCatalystAuthenticationProvider) {
         iconPath: getIcon('vscode-question'),
     })
 
-    if (!auth.activeConnection || !auth.isConnectionValid()) {
+    // There is a connection, but it is expired, or code catalyst scopes are expired
```
justinmk3, line 47: use consistent spellings everywhere.
```suggestion
    // There is a connection, but it is expired, or CodeCatalyst scopes are expired.
```

packages/core/src/codecatalyst/auth.ts
```diff
@@ -118,6 +155,54 @@ export class CodeCatalystAuthenticationProvider {
...
+        await setCodeCatalystConnectedContext(this.isConnectionValid())
+
+        this.showReauthenticationPrompt(this.activeConnection!, isPartialExpiration)
+    }
+
+    public async showReauthenticationPrompt(conn: Connection, isPartialExpiration?: boolean): Promise<void> {
+        const settings = PromptSettings.instance
+        const shouldShow = await settings.isPromptEnabled('codeCatalystConnectionExpired')
+        if (this.suppressNextReauthPrompt) {
+            this.suppressNextReauthPrompt = false
```
justinmk3, line 185: this logic looks pretty fragile

packages/core/src/codecatalyst/auth.ts
```diff
@@ -118,6 +155,54 @@ export class CodeCatalystAuthenticationProvider {
...
+
+        // Check if CodeWhisper and CodeCatalyst share the same connection
+        const isSharedConn =
+            this.secondaryAuth.activeConnection?.id == CodeWhispererAuth.instance.secondaryAuth.activeConnection?.id
+
+        if (isSharedConn) {
+            await codeWhispererClient.listAvailableCustomizations()
+        }
+
+        this.updateConnectionState(this.activeConnection!, { shared: isSharedConn })
```
justinmk3, line 169: Why is an extra `shared` flag needed? Can we just calculate it every time, instead (could be abstracted in a `isSharedConn()` function if necessary?

packages/core/src/codecatalyst/auth.ts
```diff
@@ -81,8 +105,21 @@ export class CodeCatalystAuthenticationProvider {
         return this.secondaryAuth.hasSavedConnection
     }
 
+    public setScopeExpired(isExpired: boolean) {
+        if (!this.activeConnection) return
```
justinmk3, line 109: This is hard to follow. `setScopeExpired` just randomly will do nothing if there isn't an active connection? Can the caller make this decision instead?

packages/core/src/codecatalyst/auth.ts
```diff
@@ -118,6 +155,54 @@ export class CodeCatalystAuthenticationProvider {
...
+            'CodeCatalyst connection has expired. Amazon Q/CodeWhisperer is still connected.'
+        const message = isPartialExpiration ? partiallyExpiredMessage : expiredMessage
+        const connect = 'Connect with AWS'
+        const doNotShow = "Don't Show Again"
+        const resp = await vscode.window.showInformationMessage(message, connect, doNotShow)
+        if (resp === connect) {
+            await this.auth.reauthenticate(conn)
+        } else if (resp === doNotShow) {
+            await settings.disablePrompt('codeCatalystConnectionExpired')
+        }
```
justinmk3, line 203: This looks like a copy-paste from https://github.com/aws/aws-toolkit-vscode/blob/e5fdf4109207f801a9b91a4d162ef1fb8d1bffd1/packages/core/src/codewhisperer/util/authUtil.ts#L358-L370

can we abstract it to a common function in [messages.ts](https://github.com/aws/aws-toolkit-vscode/blob/e5fdf4109207f801a9b91a4d162ef1fb8d1bffd1/packages/core/src/shared/utilities/messages.ts)

justinmk3 (changes requested):
Wonderfully explained! Please includes the PR description in the commit message. Git (not Github) is the source of truth.

Can you also put a brief hint(s) in the relevant place(s) in the code? It's not obvious from the code how/why/when partial expiration is happening.

packages/core/src/shared/clients/codecatalystClient.ts
aarolieb: Leftover from debugging. Good catch

packages/core/src/shared/clients/codecatalystClient.ts
aarolieb: Yes, I meant "aws-sdk", and you're right, this change is a bit unrelated so I'll put it into a separate commit.

packages/core/src/shared/clients/codecatalystClient.ts
aarolieb: isAwsError ensures that the error contains a `time` attribute. `ToolKitError` doesn't contain a time. Now that you mention it, I don't think the `isOnboardingError` should even work at all. I'll test and confirm the behavior and update `isOnboardingError` if that is the case.

packages/core/src/codecatalyst/auth.ts
aarolieb: Yes, you're totally right. Passing the value down is a much better idea

packages/core/src/shared/clients/codecatalystClient.ts
aarolieb: Moved to https://github.com/aws/aws-toolkit-vscode/pull/4616

packages/core/src/codecatalyst/auth.ts
justinmk3: Should be done in this PR. Copy-paste is not an acceptable practice and the expected result is that this PR will be smaller, since factoring out common code reduces duplication.

packages/core/src/codecatalyst/activation.ts
```diff
@@ -37,6 +37,15 @@ export async function activate(ctx: ExtContext): Promise<void> {
 
     await authProvider.restore()
 
+    // if connection is shared with CodeWhisperer, check if CodeCatalyst scopes are expired
+    if (authProvider.activeConnection && authProvider.isSharedConn()) {
+        try {
+            await createClient(authProvider.activeConnection, undefined, undefined, undefined, {
+                showReauthPrompt: false,
+            })
+        } catch {}
```
justinmk3, line 46: why is the exception swallowed without logging? if it is normal for createClient to throw an exception, there should be a code comment explaining why we don't want to handle (or even log, apparently) the exception.

justinmk3 (approved):
Needs a rebase / fix conflicts.

LGTM with these assumptions:

- https://github.com/aws/aws-toolkit-vscode/pull/4557#discussion_r1557945272 will be addressed
- https://github.com/aws/aws-toolkit-vscode/pull/4557#discussion_r1538230082 (copy-pasted code) will be addressed in a fast-follow PR very soon

packages/core/src/codecatalyst/activation.ts
aarolieb: An exception is expected, but you're right, there should be a log. I'll add one

packages/core/src/codecatalyst/auth.ts
aarolieb: Added to this PR

packages/core/src/shared/utilities/messages.ts
```diff
@@ -139,6 +140,43 @@ export async function showConfirmationMessage({
...
+export async function showReauthenticateMessage({
+    message,
+    connect,
+    doNotShow,
+    suppressId,
+    reauthFunc,
+}: {
+    message: string
+    connect: string
+    doNotShow: string
```
justinmk3, line 161: This should be the same throughout the extension. CW doesn't need its own "do not show" text.

I'll do this and possibly some other things in a followup PR.

justinmk3 (approved):

packages/core/src/shared/utilities/messages.ts
justinmk3: Followup: https://github.com/aws/aws-toolkit-vscode/pull/4679

PR #4560, opened 2024-03-19
fix(cwspr): Amend exposing listConnection api

leigaol:
## Problem
This is a follow up PR of https://github.com/aws/aws-toolkit-vscode/pull/4544.

## Solution

It no longer exposes internal data structure `Connection` . Now it exposes a new type `AwsToolkitSsoConnection`  that does not have `getCredentials`. 

For AmazonQ extension, I have tested that it can cast the new type `AwsToolkitSsoConnection` into `Connection` and 
it can further getCredentials from this connection. This is possible only for AmazonQ because it is re-using the same auth library as toolkit. For other 3rd party extensions, they won't be able to perform `getCredentials` from `AwsToolkitSsoConnection`

## License

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -20,15 +20,15 @@ This command supports the following arguments:
 
 ### Extension API
 
-#### `listSsoConnections`
+#### `listConnections`
 
-**Signature**: _async () => Promise<AwsToolkitSsoConnection>_
+**Signature**: _async () => Promise<AwsConnection>_
 
-This is an API that exposes the metadata of SSO connections of AWS Toolkit. It returns a list of `AwsToolkitSsoConnection` which contains below fields:
+This is an API that exposes the metadata of SSO connections of AWS Toolkit. It returns a list of `AwsConnection` which contains below fields:
 
 -   id: string that presents the id of the connection
 -   label: label of the connection
--   type: type of the connection, can only be 'sso'
+-   type: type of the connection, currently only 'sso' is returned
 -   ssoRegion: region of the connection, e.g: us-west-2
 -   startUrl: start url of the connection
 -   scopes?: list of the scopes of the connection
--- a/docs/api.md
+++ b/docs/api.md
@@ -17,3 +17,18 @@ This command supports the following arguments:
 -   startUrl and region. If both arguments are provided they will be used, otherwise the command prompts for them interactively.
 -   customizationArn: select customization by ARN. If provided, `customizationNamePrefix` is ignored.
 -   customizationNamePrefix: select customization by prefix, if `customizationArn` is `undefined`.
+
+### Extension API
+
+#### `listSsoConnections`
+
+**Signature**: _async () => Promise<AwsToolkitSsoConnection>_
+
+This is an API that exposes the metadata of SSO connections of AWS Toolkit. It returns a list of `AwsToolkitSsoConnection` which contains below fields:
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/api.ts
```diff
@@ -4,12 +4,36 @@
...
+}
+
 /**
  * Exposing listConnections API for other extension to read or re-use
  * the available connections in aws toolkit.
  */
 export const awsToolkitApi = {
-    async listConnections(): Promise<Connection[]> {
-        return Auth.instance.listConnections()
+    async listSsoConnections(): Promise<AwsToolkitSsoConnection[]> {
```
justinmk3, line 22: - `listConnections` can return all kinds of connections. It might only return SSO for now, but we should leave the door open for other kind sin the future.
- we can name our public interfaces `AwsXX`, we don't need to repeat "Toolkit" everywhere. It will be in a namespaced module.

```suggestion
    async listSsoConnections(): Promise<AwsConnection[]> {
```

packages/toolkit/src/api.ts
```diff
@@ -4,12 +4,36 @@
  */
 
 import { Auth, Connection } from 'aws-core-vscode/auth'
+
+export interface AwsToolkitSsoConnection {
+    readonly id: string
```
justinmk3, line 9:
```suggestion
export interface AwsConnection {
    readonly kind: 'sso' | 'iam'
    readonly id: string
```

packages/toolkit/src/api.ts
```diff
@@ -4,12 +4,36 @@
...
 export const awsToolkitApi = {
-    async listConnections(): Promise<Connection[]> {
-        return Auth.instance.listConnections()
+    async listSsoConnections(): Promise<AwsToolkitSsoConnection[]> {
+        const connections = await Auth.instance.listConnections()
+        const exposedConnections: AwsToolkitSsoConnection[] = []
+        connections.forEach((x: Connection) => {
+            if ('ssoRegion' in x) {
+                exposedConnections.push({
+                    id: x.id,
```
justinmk3, line 28:
```suggestion
                    id: x.id,
                    kind: 'sso'
```

docs/api.md
```diff
@@ -17,3 +17,18 @@ This command supports the following arguments:
...
+
+#### `listSsoConnections`
+
+**Signature**: _async () => Promise<AwsToolkitSsoConnection>_
+
+This is an API that exposes the metadata of SSO connections of AWS Toolkit. It returns a list of `AwsToolkitSsoConnection` which contains below fields:
+
+-   id: string that presents the id of the connection
+-   label: label of the connection
+-   type: type of the connection, can only be 'sso'
```
justinmk3, line 31:
```suggestion
-   type: type of the connection, currently only 'sso' is returned
```

packages/toolkit/src/api.ts
leigaol: Fixed in latest commit. Thanks.

leigaol:
> why are we going through this effort exactly? What's stopping another extension from using our library code and accessing the credentials?

https://github.com/aws/aws-toolkit-vscode/pull/4544#discussion_r1531023184

It makes more sense to use a dedicated AwsConnection interface than using internal data structures.

hayemaxi (approved):

PR #4565, opened 2024-03-20
lint

Will-ShaoHua:
## License

### DIFF
--- a/packages/core/src/codewhisperer/tracker/lineTracker.ts
+++ b/packages/core/src/codewhisperer/tracker/lineTracker.ts
@@ -48,16 +48,18 @@ export class LineTracker implements vscode.Disposable {
 
     constructor() {
         this._disposable = vscode.Disposable.from(
-            vscode.window.onDidChangeActiveTextEditor(e => {
-                this.onActiveTextEditorChanged(undefined)
+            vscode.window.onDidChangeActiveTextEditor(async e => {
+                await this.onActiveTextEditorChanged(undefined)
             }),
-            vscode.window.onDidChangeTextEditorSelection(e => {
-                this.onTextEditorSelectionChanged(e)
+            vscode.window.onDidChangeTextEditorSelection(async e => {
+                await this.onTextEditorSelectionChanged(e)
             }),
-            vscode.workspace.onDidChangeTextDocument(this.onContentChanged, this)
+            vscode.workspace.onDidChangeTextDocument(e => {
+                this.onContentChanged(e)
+            })
         )
 
-        queueMicrotask(() => this.onActiveTextEditorChanged(vscode.window.activeTextEditor))
+        queueMicrotask(async () => await this.onActiveTextEditorChanged(vscode.window.activeTextEditor))
     }
 
     dispose() {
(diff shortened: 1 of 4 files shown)


Will-ShaoHua:
```
(base) ➜  core git:(linter) npm run lint  

> aws-core-vscode@1.0.0 lint
> ts-node ./scripts/lint/testLint.ts && npm run format

Running linting tests...


  git-secrets
INFO: git-secrets already installed
    1) ensures no git secrets are found
    ✔ sanity check it finds secrets (246ms)

  eslint
    ✔ passes eslint (46959ms)
```

leigaol (approved):

packages/core/src/codewhisperer/tracker/lineTracker.ts
```diff
@@ -48,16 +48,18 @@ export class LineTracker implements vscode.Disposable {
 
     constructor() {
         this._disposable = vscode.Disposable.from(
-            vscode.window.onDidChangeActiveTextEditor(e => {
-                this.onActiveTextEditorChanged(undefined)
+            vscode.window.onDidChangeActiveTextEditor(async e => {
+                await this.onActiveTextEditorChanged(undefined)
             }),
```
justinmk3, line 53: for future reference: return the promise should also work
```suggestion
            vscode.window.onDidChangeActiveTextEditor(e => {
                return this.onActiveTextEditorChanged(undefined)
            }),
```

packages/core/src/codewhisperer/views/activeStateController.ts
```diff
@@ -53,9 +61,9 @@ export class ActiveStateController implements vscode.Disposable {
 
     private _isReady: boolean = false
 
-    private onReady(): void {
+    private async onReady(): Promise<void> {
         this._isReady = true
-        this._refresh(vscode.window.activeTextEditor)
+        await this._refresh(vscode.window.activeTextEditor)
```
justinmk3, line 66: If you _intentionally_ want this to be async (i.e. "don't care if it fails"), an alternative is to `.catch()` without `await`.
```suggestion
       this._refresh(vscode.window.activeTextEditor).catch(... /* log the error */)
```

packages/core/src/codewhisperer/views/activeStateController.ts
```diff
@@ -32,17 +32,25 @@ export class ActiveStateController implements vscode.Disposable {
...
+                await this.onSuggestionActionEvent(e)
+            }),
+            RecommendationHandler.instance.onDidReceiveRecommendation(async _ => {
+                await this.onDidReceiveRecommendation()
+            }),
+            this.container.lineTracker.onDidChangeActiveLines(async e => {
+                await this.onActiveLinesChanged(e)
+            }),
+            subscribeOnce(this.container.lineTracker.onReady)(async _ => {
+                await this.onReady()
```
justinmk3, line 45: FYI: If the eventemitter doesn't `await` the subscribed listeners, this will still silently fail. 

That's why sometimes it's better to `.catch()` and log the error instead of sending a `async` function to these eventemitters.

packages/core/src/codewhisperer/tracker/lineTracker.ts
Will-ShaoHua: ack, knowledge learned. 
Are these 2 equivalent?

packages/core/src/codewhisperer/views/activeStateController.ts
Will-ShaoHua: understood, yea will play around with the approach you suggests Justin

packages/core/src/codewhisperer/tracker/lineTracker.ts
justinmk3: yes. but `async` is useful if you are calling mutliple functions.

PR #4568, opened 2024-03-21
Bugfix project details loggin

dhasani23:
## Problem

Late discovery of an issue where upon opening VS Code with recently merged changes we get an error message about the "projectDetails" metric not being passive.

Also reverting the "Error" to "ToolkitError" (how it was originally) because I get a generic VS Code error about not being able to execute `aws.q.transform` when you open VS Code with no open projects.

## Solution

Implement the above.

## License


nkomonen-amazon (approved):

justinmk3:
> we get an error message about the "projectDetails" metric not being passive.

That means the projectDetails metric is _always_ emitted on startup. This is the issue I raised here: https://github.com/aws/aws-toolkit-vscode/pull/4451#discussion_r1495076168

PR #4569, opened 2024-03-21
fix: malformed SSO cache

nkomonen-amazon:
### Problem:
With some good debugging from Lei, it was discovered that we were getting errors during auth but that they were being lost.

The errors we now see indicate that during a read of the SSO cache, the JSON is malformed.

### Solution:
A guess is that multiple writes to a file can happen at the same time.

To mitigate this, when writing cache we will make it more atomic by writing all the content to a temp file, then renaming the temp file to the actual target.

Now, if there is a race condition each write will either fully make it in or be overwritten by the following write. We should now avoid the malformed JSON.

#### Other findings
- If we do multiple token refreshes in rapid succession, both refreshed access tokens will be valid by the looks of it. I guess there isn't a max of 1 active token. So if we have a race condition for the last access token to be saved to cache it doesn't really matter.

## License

### DIFF
--- a/packages/core/src/shared/utilities/cacheUtils.ts
+++ b/packages/core/src/shared/utilities/cacheUtils.ts
@@ -7,6 +7,9 @@ import * as vscode from 'vscode'
 import { dirname } from 'path'
 import { ToolkitError, isFileNotFoundError } from '../errors'
 import { SystemUtilities } from '../systemUtilities'
+import { promises as fsPromises } from 'fs'
+import { isWeb } from '../../common/webUtils'
+import crypto from 'crypto'
 
 // TODO(sijaden): further generalize this concept over async references (maybe create a library?)
 // It's pretty clear that this interface (and VSC's `Memento`) reduce down to what is essentially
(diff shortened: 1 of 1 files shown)


justinmk3:
>  both refreshed access tokens will be valid by the looks of it. I guess there isn't a max of 1 active token. So if we have a race condition for the last access token to be saved to cache it doesn't really matter.

looks like the commit message needs to reflect that?

packages/core/src/shared/utilities/cacheUtils.ts
```diff
@@ -7,6 +7,9 @@ import * as vscode from 'vscode'
 import { dirname } from 'path'
 import { ToolkitError, isFileNotFoundError } from '../errors'
 import { SystemUtilities } from '../systemUtilities'
+import { promises as fsPromises } from 'fs'
+import { isWeb } from '../../common/webUtils'
+import crypto from 'crypto'
```
justinmk3, line 12: Not a blocker, but potential followup: does this import cause any problem for web?

Also: could we use `getNonexistentFilename(..., max=0)` instead?  https://github.com/aws/aws-toolkit-vscode/blob/49828d79371bdf440f616b2ae9ce9aa9e90a05c3/packages/core/src/shared/filesystemUtilities.ts#L200

justinmk3 (approved):

PR #4571, opened 2024-03-21
fix(QCT): unnecessary notifications

dhasani23:
## Problem

When opening VS Code, all CWSPR users were getting a notification that says if they have any valid projects open for transformation. Also, the `Error`s we were throwing was causing an error notification to popup saying "Cannot execute command aws.q.transform". Changing this back to ToolkitError (with an empty string, since we already show a notification and I don't want duplicate notifications), fixes this issue too.

## Solution

Implement the above.

## License

### DIFF
--- a/packages/core/src/codewhisperer/service/transformByQHandler.ts
+++ b/packages/core/src/codewhisperer/service/transformByQHandler.ts
@@ -39,7 +39,7 @@ export async function validateAndLogProjectDetails() {
         codeTransformLocalJavaVersion,
         codeTransformPreValidationError = undefined
     try {
-        const openProjects = await getOpenProjects()
+        const openProjects = await getOpenProjects(true)
         const validProjects = await validateOpenProjects(openProjects, true)
         if (validProjects.size > 0) {
             const firstKey = validProjects.keys().next().value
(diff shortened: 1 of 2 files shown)


nkomonen-amazon (approved):

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -77,16 +77,19 @@ export function throwIfCancelled() {
...
+            void vscode.window.showErrorMessage(
+                CodeWhispererConstants.noSupportedJavaProjectsFoundMessage.replace(
+                    'LINK_HERE',
+                    CodeWhispererConstants.linkToPrerequisites
+                )
             )
-        )
-        throw new ToolkitError('No Java projects found since no projects are open', { code: 'NoOpenProjects' })
+        }
+        throw new ToolkitError('', { name: 'NoProjectsOpen' })
```
justinmk3, line 92: Throwing an error with an empty message is pretty suspicious. Would be preferred to remove the other, redundant error, and use ToolkitError consistently.

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -232,7 +235,7 @@ export async function validateOpenProjects(projects: vscode.QuickPickItem[], onP
                 reason: 'NoPomFileFound',
             })
         }
-        throw new ToolkitError('No valid Maven build file found', { code: 'NoPomFileFound', name: 'NonMavenProject' })
+        throw new ToolkitError('', { name: 'NonMavenProject' })
```
justinmk3, line 238: This is not an acceptable pattern to continue. It's a sign of a design issue.

PR #4575, opened 2024-03-21
fix(QCT): add progressUpdates details to bottom panel

dhasani23:
## Problem

Want to show users more details on what is happening at each step of their transformation. Also, want to match step names and colors between both IDEs.

## Solution

Display some info from an API response through the Transformation Hub. Previously the inner-most, indented red/green information was not being displayed ("Updated 1 dependency", "Step finished successfully", etc.).

image

## License

### DIFF
--- a/packages/core/src/codewhisperer/commands/startTransformByQ.ts
+++ b/packages/core/src/codewhisperer/commands/startTransformByQ.ts
@@ -50,7 +50,6 @@ import {
 import { MetadataResult } from '../../shared/telemetry/telemetryClient'
 import { submitFeedback } from '../../feedback/vue/submitFeedback'
 import { placeholder } from '../../shared/vscode/commands2'
-import { ToolkitError } from '../../shared/errors'
 
 const localize = nls.loadMessageBundle()
 export const stopTransformByQButton = localize('aws.codewhisperer.stop.transform.by.q', 'Stop')
@@ -58,7 +57,11 @@ export const stopTransformByQButton = localize('aws.codewhisperer.stop.transform
 let sessionJobHistory: { timestamp: string; module: string; status: string; duration: string; id: string }[] = []
 
 export async function startTransformByQWithProgress() {
-    await startTransformByQ()
+    try {
+        await startTransformByQ()
+    } catch (error: any) {
+        // swallow error; appropriate notifications have already been shown
+    }
 }
 
 interface UserInputState {
@@ -237,7 +240,7 @@ export async function startTransformByQ() {
         await setMaven()
         await validateJavaHome()
     } catch (error: any) {
-        throw new ToolkitError('', { code: 'InputValidationFailed' })
+        throw new Error('Input validation failed')
     }
 
     // Set the default state variables for our store and the UI
--- a/packages/core/src/codewhisperer/commands/startTransformByQ.ts
+++ b/packages/core/src/codewhisperer/commands/startTransformByQ.ts
@@ -50,6 +50,7 @@ import {
 import { MetadataResult } from '../../shared/telemetry/telemetryClient'
 import { submitFeedback } from '../../feedback/vue/submitFeedback'
 import { placeholder } from '../../shared/vscode/commands2'
+import { ToolkitError } from '../../shared/errors'
 
 const localize = nls.loadMessageBundle()
 export const stopTransformByQButton = localize('aws.codewhisperer.stop.transform.by.q', 'Stop')
@@ -135,7 +136,7 @@ async function pickSourceVersion(
     let placeholderText = ''
     if (sourceJDKVersion === JDKVersion.JDK8 || sourceJDKVersion === JDKVersion.JDK11) {
         placeholderText = `We found Java ${sourceJDKVersion}. Select a different version if incorrect.`
-    } else if (sourceJDKVersion === JDKVersion.UNSUPPORTED) {
+    } else if (sourceJDKVersion === JDKVersion.Other) {
         placeholderText = 'We found an unsupported Java version. Select your version here if incorrect.'
     } else {
         placeholderText = "Choose your project's Java version here." // if no .class files found or if javap fails
@@ -156,7 +157,7 @@ async function pickSourceVersion(
         transformByQState.setSourceJDKVersion(JDKVersion.JDK11)
     } else if (pick === validSourceVersions[2]) {
         // corresponds with the 'Other' option
-        transformByQState.setSourceJDKVersion(JDKVersion.UNSUPPORTED)
+        transformByQState.setSourceJDKVersion(JDKVersion.Other)
     }
 }
 
@@ -228,16 +229,19 @@ async function validateJavaHome() {
 
 export async function startTransformByQ() {
     let intervalId = undefined
+    let userInputState = undefined
 
-    // Validate inputs. If failed, Error will be thrown and execution stops
-    const userInputState = await validateTransformationJob()
-
-    await setMaven()
-
-    await validateJavaHome()
+    try {
+        // Validate inputs. If failed, Error will be thrown and execution stops
+        userInputState = await validateTransformationJob()
+        await setMaven()
+        await validateJavaHome()
+    } catch (error: any) {
+        throw new ToolkitError('', { code: 'InputValidationFailed' })
+    }
 
     // Set the default state variables for our store and the UI
-    await setTransformationToRunningState(userInputState)
+    await setTransformationToRunningState(userInputState!)
 
     try {
         // Set web view UI to poll for progress
@@ -246,10 +250,10 @@ export async function startTransformByQ() {
                 'aws.amazonq.showPlanProgressInHub',
                 codeTransformTelemetryState.getStartTime()
             )
-        }, CodeWhispererConstants.progressIntervalMs)
+        }, CodeWhispererConstants.transformationJobPollingIntervalSeconds)
 
         // step 1: CreateCodeUploadUrl and upload code
-        const uploadId = await preTransformationUploadCode(userInputState)
+        const uploadId = await preTransformationUploadCode(userInputState!)
 
         // step 2: StartJob and store the returned jobId in TransformByQState
         const jobId = await startTransformationJob(uploadId)
@@ -265,7 +269,7 @@ export async function startTransformByQ() {
     } catch (error: any) {
         await transformationJobErrorHandler(error)
     } finally {
-        await postTransformationJob(userInputState)
+        await postTransformationJob(userInputState!)
         await cleanupTransformationJob(intervalId)
     }
 }
--- a/packages/core/src/codewhisperer/commands/startTransformByQ.ts
+++ b/packages/core/src/codewhisperer/commands/startTransformByQ.ts
@@ -241,7 +241,7 @@ export async function startTransformByQ() {
     }
 
     // Set the default state variables for our store and the UI
-    await setTransformationToRunningState(userInputState!)
+    await setTransformationToRunningState(userInputState)
 
     try {
         // Set web view UI to poll for progress
@@ -253,7 +253,7 @@ export async function startTransformByQ() {
         }, CodeWhispererConstants.transformationJobPollingIntervalSeconds)
 
         // step 1: CreateCodeUploadUrl and upload code
-        const uploadId = await preTransformationUploadCode(userInputState!)
+        const uploadId = await preTransformationUploadCode(userInputState)
 
         // step 2: StartJob and store the returned jobId in TransformByQState
         const jobId = await startTransformationJob(uploadId)
@@ -269,7 +269,7 @@ export async function startTransformByQ() {
     } catch (error: any) {
         await transformationJobErrorHandler(error)
     } finally {
-        await postTransformationJob(userInputState!)
+        await postTransformationJob(userInputState)
         await cleanupTransformationJob(intervalId)
     }
 }
--- a/packages/core/src/codewhisperer/commands/startTransformByQ.ts
+++ b/packages/core/src/codewhisperer/commands/startTransformByQ.ts
@@ -9,7 +9,14 @@ import * as fs from 'fs'
 import * as os from 'os'
 import { getLogger } from '../../shared/logger'
 import * as CodeWhispererConstants from '../models/constants'
-import { transformByQState, StepProgress, TransformByQReviewStatus, JDKVersion, DropdownStep } from '../models/model'
+import {
+    transformByQState,
+    StepProgress,
+    TransformByQReviewStatus,
+    JDKVersion,
+    DropdownStep,
+    sessionPlanProgress,
+} from '../models/model'
 import { convertToTimeString, convertDateToTimestamp } from '../../shared/utilities/textUtilities'
 import {
     throwIfCancelled,
@@ -48,18 +55,6 @@ const localize = nls.loadMessageBundle()
 export const stopTransformByQButton = localize('aws.codewhisperer.stop.transform.by.q', 'Stop')
 
 let sessionJobHistory: { timestamp: string; module: string; status: string; duration: string; id: string }[] = []
-
-const sessionPlanProgress: {
-    uploadCode: StepProgress
-    buildCode: StepProgress
-    transformCode: StepProgress
-    returnCode: StepProgress
-} = {
-    uploadCode: StepProgress.NotStarted,
-    buildCode: StepProgress.NotStarted,
-    transformCode: StepProgress.NotStarted,
-    returnCode: StepProgress.NotStarted,
-}
 
 export async function startTransformByQWithProgress() {
     await startTransformByQ()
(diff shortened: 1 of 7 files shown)


packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -329,6 +329,7 @@ export async function startTransformationJob(uploadId: string) {
 
     await sleep(2000) // sleep before polling job to prevent ThrottlingException
     throwIfCancelled()
+    sessionPlanProgress['uploadCode'] = StepProgress.Succeeded
```
damntrecky, line 332: Question: What is this line doing/changing?

damntrecky (approved):
Tentative approval: Should reach out to UX and verify any other related changes to have feature parity with IntelliJ.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
dhasani23: Oh this is used to control when the green check mark appears next to each overall step (i.e. "Waiting for job to start", etc.) -- I'll rename this to 'startJob' instead of 'uploadCode' to be more clear -- here I set it to Succeeded right after the startJob API call was successful, so that next to "Waiting for job to start" we can show a green check mark.

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -827,6 +835,18 @@ export async function pollTransformationJob(jobId: string, validStates: string[]
                 result: MetadataResult.Pass,
             })
             status = response.transformationJob.status!
+            console.log('STATUS = ', status)
```
damntrecky, line 838: NIT: Remove console.log once PR is ready.

packages/core/src/codewhisperer/service/transformationHubViewProvider.ts
```diff
@@ -248,4 +258,14 @@ export class TransformationHubViewProvider implements vscode.WebviewViewProvider
             return `<span style="color: grey"> ✓ </span>`
         }
     }
+
+    private getFontColorForSubStep(status: string) {
```
damntrecky, line 262: Question: For IntelliJ we use native colors. Those native colors depend on the dark/light theme displaying. 

Do we need a tertiary statement here like the following pseudo code:
```
return isLightTheme() ? VSCode.colors.light.green : VSCode.colors.dark.green
```

packages/core/src/codewhisperer/service/transformationHubViewProvider.ts
```diff
@@ -120,15 +120,20 @@ export class TransformationHubViewProvider implements vscode.WebviewViewProvider
             planSteps = await getTransformationSteps(transformByQState.getJobId())
         }
         let progressHtml = `<p><b>Transformation Status</b></p><p>No job ongoing</p>`
```
damntrecky, line 122: Question: If its in scope for your timeline, it would be a "nice to have" to refactor the `showPlanProgress` to use string literals to build the progress panel markup.

This would be a significant refactor of this function, into several smaller functions, but would help readability and extensibility in the future.

See example refactor:
```
    
    private renderTransformationHeader() {
        return "<p><b>Transformation Status</b></p><p>No job ongoing</p>"

    }

    private renderStartJobSucceededStep(planProgress) {
        const startJobSucceededFlag = planProgress['startJob'] === StepProgress.Succeeded
        if (!startJobSucceededFlag) return ''

        return `<p> 
            ${this.getProgressIconMarkup(planProgress['buildCode'])} 
            Build uploaded code in secure build environment
        </p>
        `
    }

    private renderBuildJobSucceededStep(planProgress) {
        const buildCodeSucceededFlag = planProgress['buildCode'] === StepProgress.Succeeded
        if (!buildCodeSucceededFlag) return ''

        return `<p> 
            ${this.getProgressIconMarkup(planProgress['generatePlan'])} 
            Generate transformation plan
        </p>
        `
    }

    private renderPlanStepsSucceededStep(planProgress, planSteps) {
        const generatePlanSucceededFlag = planProgress['generatePlan'] === StepProgress.Succeeded
        if (!generatePlanSucceededFlag) return ''

        return `<p> 
            ${this.getProgressIconMarkup(planProgress['transformCode'])} 
            Transform your code to Java 17 using transformation plan
        </p>
        `
            // TODO Convert/refactor below if statement into string literal
            if () {
                progressHtml += `<p> 
            ${this.getProgressIconMarkup(planProgress['transformCode'])} 
            Transform your code to Java 17 using transformation plan
        </p>`
                // now get the details of each sub-step of the "transformCode" step
                if (planSteps !== undefined) {
                    const stepStatuses = []
                    for (const step of planSteps) {
                        stepStatuses.push(step.status)
                    }
                    for (let i = 0; i < planSteps.length; i++) {
                        const step = planSteps[i]
                        const stepStatus = step.status
                        let stepProgress = undefined
                        if (stepStatus === 'COMPLETED' || stepStatus === 'PARTIALLY_COMPLETED') {
                            stepProgress = StepProgress.Succeeded
                        } else if (
                            stepStatus === 'STOPPED' ||
                            stepStatus === 'FAILED' ||
                            planProgress['transformCode'] === StepProgress.Failed
                        ) {
                            stepProgress = StepProgress.Failed
                        } else {
                            stepProgress = StepProgress.Pending
                        }
                        // include check for the previous step not being CREATED, as this means it has finished, so we can display the next step
                        if (step.startTime && step.endTime && (i === 0 || stepStatuses[i - 1] !== 'CREATED')) {
                            const stepTime = step.endTime.toLocaleDateString('en-US', {
                                hour: '2-digit',
                                minute: '2-digit',
                                second: '2-digit',
                            })
                            const stepDuration = convertToTimeString(step.endTime.getTime() - step.startTime.getTime())
                            progressHtml += `<p style="margin-left: 20px">${this.getProgressIconMarkup(stepProgress)} ${
                                step.name
                            } [finished on ${stepTime}] <span style="color:grey">${stepDuration}</span></p>`
                        } else if (stepStatuses[i - 1] !== 'CREATED') {
                            progressHtml += `<p style="margin-left: 20px">${this.getProgressIconMarkup(stepProgress)} ${
                                step.name
                            }</p>`
                        }
                        if (step.progressUpdates) {
                            for (const subStep of step.progressUpdates) {
                                progressHtml += `<p style="margin-left: 40px">- ${subStep.name}</p>`
                                if (subStep.description) {
                                    progressHtml += `<p style="margin-left: 60px; color:${this.getFontColorForSubStep(
                                        subStep.status
                                    )}">- ${subStep.description}</p>`
                                }
                            }
                        }
                    }
                }
            }
    }

    private renderProgressStepDetails(planProgress, planSteps) {

        return `
            ${renderStartJobSucceededStep(planProgress)}
            ${renderBuildJobSucceededStep(planProgress)}
            ${renderPlanStepsSucceededStep(planProgress, planSteps)}
        `
    }

    private renderTransformationStepBody(planProgress, planSteps) {
        if (planProgress['transformCode'] === StepProgress.NotStarted) return ''

        return `
            <section>
                <p><b>Transformation Status</b></p>
                <p> ${this.getProgressIconMarkup(planProgress['startJob'])} Waiting for job to start</p>
                ${renderProgressStepDetails(planProgress, planSteps)}
            </section>
        `
    }

    private renderProgressHtml(planProgress, planSteps) {
        return `
            ${renderTransformationHeader()}
            ${renderTransformationStepBody(planProgress, planSteps)}
        `
    }
    
    private async showPlanProgress(startTime: number): Promise<string> {
        const planProgress = getPlanProgress()
        let planSteps = undefined
        if (planProgress['buildCode'] === StepProgress.Succeeded) {
            planSteps = await getTransformationSteps(transformByQState.getJobId())
        }
        const progressHtml = renderProgressHtml(planProgress, planSteps)
    }
```

packages/core/src/codewhisperer/models/constants.ts
```diff
@@ -419,7 +419,7 @@ export const transformByQStatePartialSuccessMessage = 'partially succeeded'
 
 export const transformByQStoppedState = 'STOPPED'
 
-export const transformationJobPollingIntervalSeconds = 10
+export const transformationJobPollingIntervalSeconds = 5
```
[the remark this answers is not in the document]
dhasani23: This matches what IntelliJ has been doing since November, so no.

packages/core/src/codewhisperer/service/transformationHubViewProvider.ts
dhasani23: image
image

dhasani23: Looks like there are dozens of different theme options, each of which may use their own shade of red/green, so using generic red/green as I've done here is probably best. Don't think it's worth extracting the theme the user is using, determining if light/dark/other, then making a color decision, that's kind of overkill IMO.

packages/core/src/codewhisperer/service/transformationHubViewProvider.ts
damntrecky: I see some uses like this in the VSCode repository:
```
        if (vscode.window.activeColorTheme.kind === vscode.ColorThemeKind.Light) {
            panel.iconPath = vscode.Uri.file(
                context.asAbsolutePath(join('resources', 'icons', 'aws', 'applicationcomposer', 'icon.svg'))
            )
        } else {
            panel.iconPath = vscode.Uri.file(
                context.asAbsolutePath(join('resources', 'icons', 'aws', 'applicationcomposer', 'icon-dark.svg'))
            )
        }
        ```

Comment (damntrecky) [reply packages/core/src/codewhisperer/service/transformationHubViewProvider.ts:122]: For IntelliJ you can check the `CodeModernizerUIConstants.kt` [file](https://github.com/aws/aws-toolkit-jetbrains/blob/main/plugins/toolkit/jetbrains-core/src/software/aws/toolkits/jetbrains/services/codemodernizer/constants/CodeModernizerUIConstants.kt) and see we have constants to set a consistent, IDE specific RED/Green we use throughout our UI:

```
        fun getGreenThemeFontColor(): Color = if (JBColor.isBright()) JBColor.GREEN.darker() else JBColor.GREEN
        fun getRedThemeFontColor(): Color = JBColor.RED
        ```

damntrecky: We also have the mynah-ui CSS variables available -> https://github.com/aws/mynah-ui/blob/5d128f1e4f0e7931cfc8ff6c774bd770d0a49a81/docs/STYLING.md

I'm torn here because we are integrating with their product/design system and should look seamless with them. On the other hand we should also look seamless with the IDE and if their brand colors are different, we will look different.

dhasani23: I can access the color theme kind (options are Light, High Contrast Light, Dark, High Contrast Dark) but the problem is which exact shade of red/green do I use for each kind? The mynah-ui CSS doesn't define exact shades of each color to use; it lets you customize the shade with a hex code. So yes we can do this but I need the specific color we want for each circumstance. Elsewhere in the code (for the green/grey checkmarks), I just use standard green/gray, so currently the implementation at least matches that.

damntrecky: I think for IntelliJ we just covered:
- light themes get a darker red/green
- dark themes get a light red/green

We did not cover individual themes, because there are too many and custom ones. We went with a "good enough" approach covering the base dark/light themes. I think for VSCode we can do the same.

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -17,52 +25,35 @@ import * as vscode from 'vscode'
...
-            const firstModuleEntry = validProjects.get(firstKey)
-            codeTransformLocalJavaVersion = JDKToTelemetryValue(
-                firstModuleEntry
-            ) as CodeTransformJavaSourceVersionsAllowed
+            javaVersion = validProjects.get(firstKey)
         }
     } catch (err: any) {
-        result = MetadataResult.Fail
-        reason = err?.code
-        codeTransformPreValidationError = err?.name
```
dhasani23, line 54: Since I changed the way we do Error handling, `err.message` is really what we want to access

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -233,16 +229,19 @@ async function validateJavaHome() {
...
-    await setMaven()
-
-    await validateJavaHome()
+    try {
+        // Validate inputs. If failed, Error will be thrown and execution stops
+        userInputState = await validateTransformationJob()
+        await setMaven()
+        await validateJavaHome()
+    } catch (error: any) {
+        throw new ToolkitError('', { code: 'InputValidationFailed' })
```
dhasani23, line 240: Fixed error handling so that we always use regular `Error` (`ToolkitError` not needed since we manually show our own/custom notifications) -- but this means users get a notification saying "Could not execute command aws.q.transform" when `Error` is thrown -- so I handle this by catching all of those and using a regular `ToolkitError` just once (here) with an empty string, rather than using `ToolkitError` with an empty string in 10+ places.

packages/toolkit/package.json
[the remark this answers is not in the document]
justinmk3: gitignore doesn't work for tracked files. 

Some things are in flux because of the "split toolkit" work. It's just something we have to work around.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
justinmk3: This seems like hacks upon hacks.

> since we manually show our own/custom notifications)

Why?

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -233,16 +229,19 @@ async function validateJavaHome() {
...
+    }
 
     // Set the default state variables for our store and the UI
-    await setTransformationToRunningState(userInputState)
+    await setTransformationToRunningState(userInputState!)
```
justinmk3, line 244: Usually a red flag. Why not put this call in the try{} ?

packages/core/src/codewhisperer/commands/startTransformByQ.ts
dhasani23: I wouldn't say hack necessarily but more so just that I've been doing: showing notifications via `vscode.window.showError/InfoMessage` + throwing regular `Error` versus bundling the notification with the `ToolkitError`. We previously had both types of errors throughout which was confusing at times, so wanted to pick one and stick with it.

Also, there are some places such as https://github.com/aws/aws-toolkit-vscode/blob/cf3f0fe020370ed80f55c7f4aba8f3691e8ce73c/packages/core/src/codewhisperer/commands/startTransformByQ.ts#L566 where we allow the user to click a "Submit feedback" button through the notification, which I don't believe I can implement with the `ToolkitError`. So from this for consistency I figured it's better to use `vscode.window` to do notification handling everywhere, and would like to avoid duplicate notifications being shown, hence why I got rid of the many `ToolkitError`s.

justinmk3: Throwing an error with an empty message could lead to some strange situations and just isn't an expected or recommended practice. That's not done anywhere else in the codebase (except a few tests).

The fact that some accidental situations have lead to this workaround doesn't change the fact that there is a design issue here.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -49,20 +56,12 @@ export const stopTransformByQButton = localize('aws.codewhisperer.stop.transform
...
-    transformCode: StepProgress.NotStarted,
-    returnCode: StepProgress.NotStarted,
-}
-
 export async function startTransformByQWithProgress() {
-    await startTransformByQ()
+    try {
+        await startTransformByQ()
+    } catch (error: any) {
+        // swallow error; appropriate notifications have already been shown
```
damntrecky, line 63: Question: Linter doesn't complain about a void or empty catch statement?

packages/core/src/codewhisperer/service/transformationHubViewProvider.ts
damntrecky: Not a blocker, just a comment. I don't think its overkill. We should definitely have a mechanism in the future for getting a certain color palette based on light/dark themes.

For IntelliJ we just covered:

    light themes get a darker red/green
    dark themes get a light red/green

We did not cover individual themes, because there are too many and custom ones. We went with a "good enough" approach covering the base dark/light themes. I think for VSCode we can do the same.

damntrecky (approved):

dhasani23:
The recently merged chat invocation PR should address the notification showing issues I presented here, and since I don't want to resolve the conflicts incorrectly, going to open a new PR which just makes the Transformation Hub changes and the other small changes not related to notifications.

PR #4576, opened 2024-03-22
feat(wizard): support async prompter functions

justinmk3:
Problem:
- `bindPrompter` does not support binding to async functions. This can be very awkward, and limits how prompts can be implemented.
- `getNonexistentFilename` has a bug: it checks `existsFile` which returns false for directories.

Solution:
- Change the definition of `PrompterProvider` to allow returning a Promise.
- Fix the check in `getNonexistentFilename`.

## License

### DIFF
--- a/packages/core/src/test/shared/filesystemUtilities.test.ts
+++ b/packages/core/src/test/shared/filesystemUtilities.test.ts
@@ -11,7 +11,6 @@ import {
     fileExists,
     getFileDistance,
     getNonexistentFilename,
-    getNonexistentFilenameSync,
     isInDirectory,
     makeTemporaryToolkitFolder,
     tempDirPath,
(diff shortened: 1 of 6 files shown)


packages/core/src/test/shared/filesystemUtilities.test.ts
```diff
@@ -40,29 +39,25 @@ describe('filesystemUtilities', function () {
...
-                assert.strictEqual(await methodUnderTest(dir, 'foo', '', 99), 'foo')
-            })
-            it(`${methodUnderTest.name}: returns "foo-RANDOM.txt" if max is reached`, async function () {
-                const dir = tempFolder
-                await writeFile(path.join(dir, 'foo.txt'), '', 'utf8')
-                await writeFile(path.join(dir, 'foo-1.txt'), '', 'utf8')
-                // Looks like "foo-75446d5d.txt".
-                assert.ok(/^foo-[a-fA-F0-9]{8}.txt$/.test(await methodUnderTest(dir, 'foo', '.txt', 1)))
-            })
+        it(`getNonexistentFilename: failure modes`, async function () {
```
jpinkney-aws, line 42: since you removed getNonexistentFilenameSync should we remove 'getNonexistentFilename' at the start of the test names since it's already in the describe?

jpinkney-aws (approved):

PR #4579, opened 2024-03-22
codewhisperer: zip util and triggers for file scans

ctlai95:
## Problem

Add support for auto scanning of the current active file.

## Solution

- Add `ZipUtil` helper to prepare current file's contents as a zip file (tweaked from `DependencyGraph` implementation)
- Update `startSecurityScan` to run a scan on the current active file
- Add new activation subscriptions for auto scanning:
  - when the extension first loads
  - when contents of the text editor changes
  - when the active text editor changes
  - when the auto scan toggle has been enabled
- Add new `isLanguageSupported` check for security scans

Todo next: Update the existing security scan trigger to scan the whole project without dependency graph logic

## License

### DIFF
--- a/packages/core/src/codewhisperer/activation.ts
+++ b/packages/core/src/codewhisperer/activation.ts
@@ -47,7 +47,7 @@ import { ReferenceLogViewProvider } from './service/referenceLogViewProvider'
 import { ReferenceHoverProvider } from './service/referenceHoverProvider'
 import { ReferenceInlineProvider } from './service/referenceInlineProvider'
 import { SecurityPanelViewProvider } from './views/securityPanelViewProvider'
-import { disposeSecurityDiagnostic } from './service/diagnosticsProvider'
+import { disposeSecurityDiagnostic, securityScanRender } from './service/diagnosticsProvider'
 import { RecommendationHandler } from './service/recommendationHandler'
 import { Commands, registerCommandsWithVSCode } from '../shared/vscode/commands2'
 import { InlineCompletionService, refreshStatusBar } from './service/inlineCompletionService'
@@ -64,6 +64,8 @@ import { SecurityIssueCodeActionProvider } from './service/securityIssueCodeActi
 import { listCodeWhispererCommands } from './ui/statusBarMenu'
 import { updateUserProxyUrl } from './client/agent'
 import { Container } from './service/serviceContainer'
+import { debounceStartSecurityScan } from './commands/startSecurityScan'
+import { securityScanLanguageContext } from './util/securityScanLanguageContext'
 const performance = globalThis.performance ?? require('perf_hooks').performance
 
 export async function activate(context: ExtContext): Promise<void> {
@@ -94,6 +96,11 @@ export async function activate(context: ExtContext): Promise<void> {
      * Service control
      */
     const client = new codewhispererClient.DefaultCodeWhispererClient()
+
+    /**
+     * CodeWhisperer auto scans
+     */
+    setSubscriptionsForAutoScans()
 
     // Service initialization
     const container = Container.instance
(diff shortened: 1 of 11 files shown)


mj632 (approved):

packages/core/src/codewhisperer/util/zipUtil.ts
```diff
@@ -0,0 +1,129 @@
...
+                buildPayloadSizeInBytes: 0,
+                lines: this._totalLines,
+            }
+        } catch (error) {
+            getLogger().error('Zip error caused by:', error)
+            throw error
+        }
+    }
+
+    public async removeTmpFiles(zipMetadata: ZipMetadata) {
```
justinmk3, line 123: Have you considered using a "streaming" approach for zipping files? Then temp files wouldn't be needed. That also avoids memory limits. And this would be useful for several other features in the codebase (featureDev, etc.).

- https://stackoverflow.com/a/71279922/152142
- https://github.com/archiverjs/node-zip-stream

packages/core/src/codewhisperer/util/zipUtil.ts
```diff
@@ -0,0 +1,129 @@
...
+    buildPayloadSizeInBytes: number
+    zipFileSizeInBytes: number
+    lines: number
+}
+
+export const ZipConstants = {
+    newlineRegex: /\r?\n/,
+}
+
+export class ZipUtil {
```
justinmk3, line 28: Is this really doing much that is codewhisperer specific? Things like `fileScanPayloadSizeLimitBytes` could be parameterized. If a "zip util" is needed then presumably that would be useful to other parts of the codebase, not only CW.

packages/core/src/codewhisperer/activation.ts
```diff
@@ -267,6 +274,93 @@ export async function activate(context: ExtContext): Promise<void> {
...
+            }),
+            // Trigger scan if the file contents change
+            vscode.workspace.onDidChangeTextDocument(async event => {
+                const editor = vscode.window.activeTextEditor
+                if (
+                    CodeScansState.instance.isScansEnabled() &&
+                    editor &&
+                    event.document === editor.document &&
+                    securityScanLanguageContext.isLanguageSupported(editor.document.languageId) &&
+                    event.contentChanges.length > 0
```
ashishrp-aws, line 331: Are we going to trigger scans even for 1 character change? even though they are just file scans, wouldn't that be expensive?

packages/core/src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -168,7 +225,26 @@ function getMd5(fileName: string) {
...
+                throw new CodeScanStoppedError()
+            }
+            break
+        default:
+            getLogger().warn(`Unknown scan type: ${scanType}`)
+            break
+    }
+}
+
+// TODO: Remove this
```
ashishrp-aws, line 246: Are we removing all the old functions?

ashishrp-aws (approved):

ashishrp-aws (commented):
Code Test failure for ListCodeWhispererCommands: https://github.com/aws/aws-toolkit-vscode/actions/runs/8396824421/job/22999007540?pr=4579#step:5:4464

packages/core/src/codewhisperer/activation.ts
ctlai95: For now yes, we don't know whether the 1 character change will complete the function or not. We can adjust later once we get actual usage data

packages/core/src/codewhisperer/service/securityScanHandler.ts
ctlai95: Yes they will be removed in a future PR to this branch

packages/core/src/codewhisperer/util/zipUtil.ts
ctlai95: Will look into this in a separate PR

packages/core/src/codewhisperer/util/zipUtil.ts
ctlai95: There will be more changes included in the next PR for project scans. Will look into making it generic in a separate PR

packages/core/src/test/codewhisperer/commands/startSecurityScanOld.test.ts
```diff
@@ -0,0 +1,271 @@
...
+        )
+        await startSecurityScan.confirmStopSecurityScan()
+        await scanPromise
+        assert.ok(securityScanRenderSpy.calledOnce)
+        assert.ok(securityScanStoppedErrorSpy.notCalled)
+        const warnings = testWindow.shownMessages.filter(m => m.severity === SeverityLevel.Warning)
+        assert.ok(warnings.map(m => m.message).includes(stopScanMessage))
+    })
+
+    it('Should highlight files after scan is completed', async function () {
```
justinmk3, line 245: This test failed on macOS:

```
  1) startSecurityScan
       Should highlight files after scan is completed:

      AssertionError [ERR_ASSERTION]: telemetry item 1 (of 1) not found (in the expected order) for metric name: "codewhisperer_securityScan" 
      + expected - actual

       {
         "codewhispererCodeScanIssuesWithFixes": "0"
      -  "codewhispererCodeScanLines": "0"
      -  "codewhispererCodeScanTotalIssues": "0"
      +  "codewhispererCodeScanLines": "66"
      +  "codewhispererCodeScanTotalIssues": "1"
         "codewhispererLanguage": "python"
       }
      
      at partialDeepCompare (src/test/testUtil.ts:218:12)
      at assertTelemetry (src/test/testUtil.ts:259:9)
      at Context.<anonymous> (src/test/codewhisperer/commands/startSecurityScan.test.ts:295:24)
```

packages/core/src/test/codewhisperer/commands/startSecurityScanOld.test.ts
```diff
@@ -0,0 +1,271 @@
...
+            }
+        })
+        await startSecurityScan.startSecurityScanOld(
+            mockSecurityPanelViewProvider,
+            editor,
+            createClient(),
+            extensionContext
+        )
+        assert.ok(commandSpy.calledWith(codeScanLogsOutputChannelId))
+        assertTelemetry('codewhisperer_securityScan', {
```
justinmk3, line 264: you could try wrapping this in a `waitUntil` so that it retries a few times, if there is some async work happening above

packages/core/src/test/codewhisperer/commands/startSecurityScanOld.test.ts
```diff
@@ -0,0 +1,271 @@
...
+        mockSecurityPanelViewProvider = new SecurityPanelViewProvider(extensionContext)
+        appRoot = join(workspaceFolder, 'python3.7-plain-sam-app')
+        appCodePath = join(appRoot, 'hello_world', 'app.py')
+        editor = await openTestFile(appCodePath)
+    })
+    afterEach(function () {
+        sinon.restore()
+    })
+    after(async function () {
+        await closeAllEditors()
```
justinmk3, line 145: should this be in afterEach? because beforeEach is opening a new editor for each test...

packages/core/src/test/codewhisperer/commands/startSecurityScanOld.test.ts
```diff
@@ -0,0 +1,271 @@
...
+let editor: vscode.TextEditor
+
+describe('startSecurityScanOld', function () {
+    const workspaceFolder = getTestWorkspaceFolder()
+
+    beforeEach(async function () {
+        extensionContext = await FakeExtensionContext.create()
+        mockSecurityPanelViewProvider = new SecurityPanelViewProvider(extensionContext)
+        appRoot = join(workspaceFolder, 'python3.7-plain-sam-app')
+        appCodePath = join(appRoot, 'hello_world', 'app.py')
```
justinmk3, line 138: there is never a reason to use join() on constant strings. and it's less readable.

```suggestion
        appCodePath = join(appRoot, 'hello_world/app.py')
```

PR #4582, opened 2024-03-25
fix(auth): network error may invalidate SSO token

nkomonen-amazon:
## Problem:

In Auth we were having premature SSO connection expiration issues due to network problems such as no internet connection. If the extension attempted a token refresh, or the extension was started without internet connected, the existing auth token would be invalidated.

## Solution:

We had some code that validated the token, but if it failed
due to network issues it would invalidate it.

Now, when we validate the token and there is a network error,
we will keep the state as-is instead of setting the connection
to invalid.

The user will instead see an error message indicating network issues or a log message depending on the action
performed.

## Additional
Some renaming in the auth code to improve readability. See each commit for specifics.

## License


justinmk3 (approved):
Nice! I think the commits can be squashed, it's not too hard to follow as one unit.

Can a test be added? See https://github.com/aws/aws-toolkit-vscode/commit/ecd0eeab61a2a56325ec3bf5660078ba4b752a82 for a similar test case.

justinmk3 (approved):

PR #4585, opened 2024-03-25
feature: create initial standalone Q/CW and Toolkit extensions

hayemaxi:
Summary of changes:
- Q/CW extension builds and packages into a functioning independent extension.
- Toolkit extension does not contain any Q/CW functionality, but does connect and share auth with Q/CW.
- "dynamic" commands. For example, to init auth and its commands, you have to pass a prefix that identifies the extension. E.g., auth commands for amazonq will register as `aws.amazonq.auth.logout` instead of `aws.auth.logout`
- Disabled type checking for settings at this time. It is not compatible with having disjoint package.json files :(
- Disabled many cases where commands would be registered on import. Instead, they will be registered if activation code is executed.
- New auth UI!

## License


packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -431,6 +432,23 @@ export async function getChatAuthState(cwAuth = AuthUtil.instance): Promise<Feat
     return state
 }
 
+export function isPreviousQUser() {
```
justinmk3, line 435: Probably worth a docstring that says what this is for. This eventually won't be needed, right?

    /** Returns true if the user is an existing Toolkit user who has newly installed the Q standalone extension. */

packages/core/src/codewhisperer/util/codewhispererSettings.ts
```diff
@@ -11,10 +11,10 @@ const description = {
 }
 export class CodeWhispererSettings extends fromExtensionManifest('aws.codeWhisperer', description) {
     public isSuggestionsWithCodeReferencesEnabled(): boolean {
-        return this.get(`includeSuggestionsWithCodeReferences`, false)
+        return this.get(`includeSuggestionsWithCodeReferences`, false) as any
```
justinmk3, line 14: Seems suspicious. Is this because package.json is in flux?

packages/core/src/login/webview/commonAuthViewProvider.ts
```diff
@@ -8,26 +8,25 @@
...
  * 1. Create a view in package.json
  * {
-        "type": "webview",
-        "id": "aws.AmazonCommonAuth",
-        "name": "%AWS.amazonq.login%",
-        "when": "!isCloud9 && !aws.isSageMaker && !aws.amazonq.showView"
-    },
+"type": "webview",
+"id": "aws.AmazonCommonAuth",
+"name": "%AWS.amazonq.login%",
```
justinmk3, line 13: this (and below) looks accidental

packages/core/src/shared/extensionGlobals.ts
```diff
@@ -121,8 +121,10 @@ export default globals
  */
 interface ToolkitGlobals {
     readonly context: ExtensionContext
+    contextPrefix: string
```
justinmk3, line 124:
```suggestion
    /** Decides the prefix for vscode "setContext" keys. */
    contextPrefix: string
```

packages/core/src/shared/settings.ts
```diff
@@ -319,7 +320,7 @@ function createSettingsClass<T extends TypeDescriptor>(section: string, descript
          * @param key Setting name
          * @param defaultValue Value returned if setting is missing or invalid
          */
-        public get<K extends keyof Inner>(key: K & string, defaultValue?: Inner[K]) {
+        public get<K extends keyof Inner>(key: K & string, defaultValue?: Inner[K]): any {
```
justinmk3, line 323: temporary while package.json's are in flux?

packages/core/src/shared/settings.ts
```diff
@@ -477,40 +478,40 @@ export interface ResetableMemento extends vscode.Memento {
...
-            ? T[L]
-            : Select<T[L], R>
-        : never
-    : never
-
-type Sections = { [P in keyof SettingsProps as Join<Pop<Split<P, '.'>>, '.'>]: Select<Config, Pop<Split<P, '.'>>> }
+// type SettingsProps = typeof settingsProps
+
+// type Split<T, S extends string> = T extends `${infer L}${S}${infer R}` ? [L, ...Split<R, S>] : [T]
+// type Pop<T> = T extends [...infer R, infer _] ? R : never
```
justinmk3, line 484: oh :)

packages/core/src/shared/settings.ts
```diff
@@ -567,21 +565,23 @@ export function fromExtensionManifest<T extends TypeDescriptor & Partial<Section
     //
     // As long as the above holds true, throwing an error here will always be caught by CI
 
-    const resolved = keys(descriptor).map(k => `${section}.${k}`)
-    const missing = resolved.filter(k => (settingsProps as Record<string, any>)[k] === undefined)
+    // TODO: restore/remove checking
+    // const resolved = keys(descriptor).map(k => `${section}.${k}`)
+    // const missing = resolved.filter(k => (settingsProps as Record<string, any>)[k] === undefined)
```
justinmk3, line 570: Hope we can restore it. Unchecked settings usage was a common source of bugs.

justinmk3 (approved):

packages/core/src/shared/settings.ts
hayemaxi: Yeah this is an unfortunate loss, and I will remove this code instead of commenting it shortly. This code relies on the core libs package.json (that is copied to the toolkit extension). Since the standalone extensions will have divergent package.jsons, this is no longer viable.
Ideally, the core lib can supply settings classes and extensions can provide their own package.json to get type checking, but the code that access these custom settings classes would need to live in their individual packages.
This will have to be tech debt

hayemaxi:
will have to fix CI in future PR. Let's get this stuff in first.

PR #4586, opened 2024-03-26
feat(codewhisperer): Add back arrow at login start screen;move code catalyst login option

leigaol:
## Problem
This is a PR moved from https://github.com/hayemaxi/aws-toolkit-vscode/pull/7.

    1. User cannot easily quit the login screen
    2. Code catalyst login should be under Workforce button

## Solution

1 Add back arrow at login start screen ONLY for aws toolkit
Screenshot 2024-03-25 at 2 45 16 PM

2. Move code catalyst under Workforce button login screen
image (2)

## License

### DIFF
--- a/packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
+++ b/packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
@@ -114,5 +114,6 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
         throw new Error('Method not implemented.')
     }
 
+    /** If users are unauthenticated in Q/CW, we should always display the auth screen. */
     async quitLoginScreen() {}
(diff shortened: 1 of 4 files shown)


packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
```diff
@@ -113,4 +113,6 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
     ): Promise<AuthError | undefined> {
         throw new Error('Method not implemented.')
     }
+
```
hayemaxi, line 116:
```suggestion

/** If users are unauthenticated in Q/CW, we should always display the auth screen. */
```

hayemaxi (approved):

PR #4591, opened 2024-03-26
fix(codewhisperer): line tracker stuck in an infinite loop while user's editor focus is on OUTPUT/LOGS and documentChangedListener will not stop

Will-ShaoHua:
## Problem
`LineTracker` is subscribing `vscode.TextDocumentChangeEvent` and VSCode will also fire the above event when the plugin is adding new entry of logs. 
Therefore when user's focus is at log file, the following events will happen in order
1. VSCode `onContentChanged`
2. `LineTracker` emits events to update CWSPR UI components
3. `ActiveStateController.refresh()` & `LineAnnotationController.refresh()` 
4. `refresh()` invoke `isConnectionValid()`, which will add logs
5. back to step (1)

## Before

https://github.com/aws/aws-toolkit-vscode/assets/96078566/aa499070-1953-4126-9ab2-505bda778bcd

## After

https://github.com/aws/aws-toolkit-vscode/assets/96078566/b34f75c9-aafb-4556-9bc4-c40c64dc476f

## License

### DIFF
--- a/packages/core/src/codewhisperer/views/lineAnnotationController.ts
+++ b/packages/core/src/codewhisperer/views/lineAnnotationController.ts
@@ -327,11 +327,6 @@ export class LineAnnotationController implements vscode.Disposable {
             return
         }
 
-        if (!this.container.auth.isConnectionValid()) {
-            this.clear()
-            return
-        }
-
         if (this.isTutorialDone()) {
             this.clear()
             return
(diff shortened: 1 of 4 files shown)


packages/core/src/codewhisperer/views/lineAnnotationController.ts
```diff
@@ -360,6 +355,11 @@ export class LineAnnotationController implements vscode.Disposable {
             return
         }
 
+        if (!this.container.auth.isConnectionValid()) {
+            this.clear()
+            return
+        }
+
```
Will-ShaoHua, line 362: same here, L341 is checking `isTextEditor`

justinmk3 (approved):
waiting for CW team member to review

leigaol (approved):

andrewyuq (approved):

PR #4593, opened 2024-03-26
fix(codewhisperer): fix tutorial state not persisted when user press esc to skip tutorial

Will-ShaoHua:
## Problem
Missing `set(inlinehintKey, this._currentState.id, globals.context.globalState)` call

## Solution
Add `set(inlinehintKey, this._currentState.id, globals.context.globalState)` call

## License

### DIFF
--- a/packages/core/src/codewhisperer/views/lineAnnotationController.ts
+++ b/packages/core/src/codewhisperer/views/lineAnnotationController.ts
@@ -262,8 +262,7 @@ export class LineAnnotationController implements vscode.Disposable {
                     try {
                         telemetry.ui_click.emit({ elementId: `dismiss_${this._currentState.id}` })
                     } catch (_) {}
-                    this._currentState = new EndState()
-                    await vscode.commands.executeCommand('setContext', inlinehintWipKey, false)
+                    await this.markTutorialDone()
                     getLogger().debug(`codewhisperer: user dismiss tutorial.`)
                 }
             })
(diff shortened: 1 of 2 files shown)


packages/core/src/codewhisperer/views/lineAnnotationController.ts
```diff
@@ -262,8 +262,7 @@ export class LineAnnotationController implements vscode.Disposable {
                     try {
                         telemetry.ui_click.emit({ elementId: `dismiss_${this._currentState.id}` })
                     } catch (_) {}
-                    this._currentState = new EndState()
-                    await vscode.commands.executeCommand('setContext', inlinehintWipKey, false)
+                    await this.markTutorialDone()
```
Will-ShaoHua, line 265: for reference 
https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/views/lineAnnotationController.ts#L288

andrewyuq (approved):

packages/core/src/codewhisperer/views/lineAnnotationController.ts
Will-ShaoHua:
```
    async markTutorialDone() {
        this._currentState = new EndState()
        await vscode.commands.executeCommand('setContext', inlinehintWipKey, false)
        await set(inlinehintKey, this._currentState.id, globals.context.globalState)
    }
```

PR #4594, opened 2024-03-26
feat(Amazon Q FeatureDev): add test for insertChanges

dianaabv16:
## Problem
Follow up for the this PR:
https://github.com/aws/aws-toolkit-vscode/pull/4556

why?
This test makes sure only non rejected files are being inserted when customer accepting changes.

## License

### DIFF
--- a/packages/core/src/test/amazonqFeatureDev/session/session.test.ts
+++ b/packages/core/src/test/amazonqFeatureDev/session/session.test.ts
@@ -3,19 +3,34 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import * as vscode from 'vscode'
+import * as assert from 'assert'
+
 import sinon from 'sinon'
-import { createMessenger, createSession } from '../utils'
-import { Session } from '../../../amazonqFeatureDev/session/session'
+
+import {
+    ControllerSetup,
+    createController,
+    createMessenger,
+    createSession,
+    generateVirtualMemoryUri,
+    sessionRegisterProvider,
+    sessionWriteFile,
+} from '../utils'
+import { CurrentWsFolders } from '../../../amazonqFeatureDev/types'
+import { CodeGenState } from '../../../amazonqFeatureDev/session/sessionState'
+import { FeatureDevClient } from '../../../amazonqFeatureDev/client/featureDev'
+import path from 'path'
 import { assertTelemetry } from '../../testUtil'
+import { Messenger } from '../../../amazonqFeatureDev/controllers/chat/messenger/messenger'
+import { FileSystemCommon } from '../../../srcShared/fs'
 
 describe('session', () => {
     const conversationID = '12345'
+    let messenger: Messenger
 
-    let session: Session
-
-    beforeEach(async () => {
-        const messenger = createMessenger()
-        session = await createSession({ messenger, conversationID })
+    beforeEach(() => {
+        messenger = createMessenger()
     })
 
     afterEach(() => {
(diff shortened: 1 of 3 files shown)


packages/core/src/test/amazonqFeatureDev/session/session.test.ts
```diff
@@ -31,4 +40,77 @@ describe('session', () => {
...
+                    getCodeGeneration: () => sinon.stub(),
+                    exportResultArchive: () => sinon.stub(),
+                } as unknown as FeatureDevClient,
+                sourceRoots: [''],
+                uploadId: uploadID,
+                workspaceFolders,
+            }
+            const testApproach = 'test-approach'
+
+            const codeGenState = new CodeGenState(
```
jpinkney-aws, line 71: I think in the future we should definitely start creating some easier ways to create codegen state (and states in general). So many defaults store between here and the actual session stuff that we could just extract away when they aren't being manually set

packages/core/src/test/amazonqFeatureDev/session/session.test.ts
```diff
@@ -31,4 +40,77 @@ describe('session', () => {
...
+
+            await session.insertChanges()
+
+            await waitUntil(() => {
+                return Promise.resolve(fsStub.callCount > 0)
+            }, {})
+
+            const absolutePath = path.join(workspaceFolder.uri.fsPath, 'myfile.js')
+            assert.ok(fsStub.calledOnce)
+            assert.ok(fsStub.calledWith(absolutePath, ''))
```
jsalles, line 113: Is this empty string here the file content?

It'd make it clearer if we added some value to the file content on the files above.
`myFile.js` could have a content of `Contents of myFile.js` for example and this could be tested here. Better than testing for an empty string :)

dianaabv16: added!

packages/core/src/amazonqFeatureDev/session/session.ts
```diff
@@ -173,13 +171,13 @@ export class Session {
...
-            await fs.mkdir(path.dirname(absolutePath))
-            await fs.writeFile(absolutePath, decodedContent)
+            await FileSystemCommon.instance.mkdir(path.dirname(absolutePath))
+            await FileSystemCommon.instance.writeFile(absolutePath, decodedContent)
         }
 
         for (const filePath of this.state.deletedFiles?.filter(i => !i.rejected) ?? []) {
             const absolutePath = path.join(filePath.workspaceFolder.uri.fsPath, filePath.relativePath)
-            await fs.delete(absolutePath)
+            await FileSystemCommon.instance.delete(absolutePath)
```
justinmk3, line 180: ? why

dianaabv16: otherwise the mock of the test doesn't work

jsalles (approved):

packages/core/src/amazonqFeatureDev/session/session.ts
justinmk3: why are you using mocks instead of creating actual files (which tests real codepaths instead of mocks)?

packages/core/src/test/amazonqFeatureDev/session/session.test.ts
```diff
@@ -31,4 +39,84 @@ describe('session', () => {
...
+                .resolves(new TextEncoder().encode(notRejectedFileContent))
+
+            await session.insertChanges()
+
+            await waitUntil(() => {
+                return Promise.resolve(fsStubWriteFile.callCount > 0)
+            }, {})
+            await waitUntil(() => {
+                return Promise.resolve(fsStubReadFile.callCount > 0)
+            }, {})
```
justinmk3, line 114: this could be in a single waitUntil (use `&&` or something)

packages/core/src/test/amazonqFeatureDev/session/session.test.ts
```diff
@@ -3,32 +3,110 @@
...
 describe('session', () => {
     const conversationID = '12345'
-
-    let session: Session
-
-    beforeEach(async () => {
-        const messenger = createMessenger()
-        session = await createSession({ messenger, conversationID })
-    })
+    const messenger = createMessenger()
```
justinmk3, line 21: Non-trivial code must not be called in describe() scope. use a `before()` (not `beforeEach()`) if you want to do one-time setup.

dianaabv16: moved to beforeEach

jsalles (approved):

packages/core/src/test/amazonqFeatureDev/utils.ts
```diff
@@ -70,6 +73,23 @@ export async function createSession({
...
+    const uri = vscode.Uri.from({ scheme: featureDevScheme, path: generationFilePath })
+    return uri
+}
+
+export async function sessionWriteFile(session: Session, uri: vscode.Uri, encodedContent: Uint8Array) {
+    await session.config.fs.writeFile(uri, encodedContent, {
+        create: true,
+        overwrite: true,
+    })
+}
```
justinmk3, line 91: This is needless indirection.

packages/core/src/test/amazonqFeatureDev/utils.ts
```diff
@@ -70,6 +73,23 @@ export async function createSession({
     return session
 }
 
+export async function sessionRegisterProvider(session: Session, uri: vscode.Uri, fileContents: Uint8Array) {
+    session.config.fs.registerProvider(uri, new VirtualMemoryFile(fileContents))
+}
+
```
justinmk3, line 79: unnecessary indirection

justinmk3 (approved):
Merging with the expectation that these tests will be simplifed, as discussed offline:

- avoid stubs/mocks for everything except the low-level service api client
- test against an actual workspace, files, session, etc, so that the real codepaths are exercised instead of stubbing almost everything.

PR #4595, opened 2024-03-26
fix(q): login state context simplification

hayemaxi:
- Move auth hook intializations to a dedicated function in AuthUtil, so that both extensions don't attach the hooks.
- Remove extra instances of context setting for the login page.

## License

### DIFF
--- a/packages/core/src/auth/ui/vue/show.ts
+++ b/packages/core/src/auth/ui/vue/show.ts
@@ -718,7 +718,7 @@ export function getShowManageConnections(): RegisteredCommand<any> {
 export function registerCommands(context: vscode.ExtensionContext, prefix: string) {
     showManageConnections = Commands.register(
         { id: `aws.${prefix}.auth.manageConnections`, compositeKey: { 1: 'source' } },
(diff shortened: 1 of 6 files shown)


packages/core/src/auth/ui/vue/show.ts
```diff
@@ -718,7 +718,7 @@ export function getShowManageConnections(): RegisteredCommand<any> {
 export function registerCommands(context: vscode.ExtensionContext, prefix: string) {
     showManageConnections = Commands.register(
         { id: `aws.${prefix}.auth.manageConnections`, compositeKey: { 1: 'source' } },
-        (_: VsCodeCommandArg, source: AuthSource, serviceToShow?: ServiceItemId) => {
+        async (_: VsCodeCommandArg, source: AuthSource, serviceToShow?: ServiceItemId) => {
```
leigaol, line 721: you can add await to where this command is executed

hayemaxi: command executions are always async, so no change to callers.

leigaol (approved):

PR #4598, opened 2024-03-26
fix(amazonq): Rename settings in Amazonq

leigaol:
## Problem
The settings are still called CodeWhisperer, they should be renamed into Amazon Q. 
## License


packages/amazonq/package.json
```diff
@@ -128,23 +128,23 @@
                     },
                     "additionalProperties": false
                 },
-                "aws.codeWhisperer.includeSuggestionsWithCodeReferences": {
+                "amazonQ.developer.includeSuggestionsWithCodeReferences": {
```
justinmk3, line 131:
```suggestion
                "amazonQ.includeSuggestionsWithCodeReferences": {
```

packages/amazonq/package.json
```diff
@@ -70,7 +70,7 @@
                     "deprecationMessage": "The current profile is now stored internally by the Toolkit.",
                     "description": "%AWS.configuration.profileDescription%"
                 },
-                "aws.amazonq.logLevel": {
+                "amazonQ.logLevel": {
```
hayemaxi, line 73: this will break some of the dynamic setting/context code i implemented. elsewhere in the code, we will do `aws.${prefix}.logLevel`. same for telemetry.

by the way, shouldn't we leave `aws.` in the title, since AWS is the publisher

hayemaxi: on second thought, maybe we can/should share this config value between the extensions

leigaol: >since AWS is the publisher

Amazon Q is not branded under AWS, it is called Amazon Q.

justinmk3: > Amazon Q is not branded under AWS, it is called Amazon Q.

Yep. But if there's a technical reason why we need to keep the `aws.` prefix then we will need to compromise.

PR #4601, opened 2024-03-27
fix(cwspr): Add 2 way connection sharing between toolkit & amazonq

leigaol:
## Problem
Connection needs to be shared between toolkit & q.

## Solution
Add 2 way connection sharing between toolkit & amazonq.

* AWS Toolkit exposes a public API which Amazon Q calls and subscribes-to.
* In order to avoid confusion about which is the “leader”, Amazon Q does not expose an API. 
* Amazon Q gets/lists connections from AWS Toolkit.
* Amazon Q listens to Toolkit connection events via onDidUpdateConnection.
* Amazon Q updates (updateConnection, deleteConnection) the state of AWS Toolkit whenever Amazon Q connection state is updated.
    * Avoid infinite loop: updateConnection, deleteConnection do not trigger

## License


hayemaxi:
Tests will fail until we can address them, but lets make sure the lint test passes at least

leigaol:
> Tests will fail until we can address them, but lets make sure the lint test passes at least

Linter error is fixed.

packages/core/src/login/webview/commonAuthViewProvider.ts
```diff
@@ -76,8 +76,10 @@ export class CommonAuthViewProvider implements WebviewViewProvider {
         context: WebviewViewResolveContext,
         _token: CancellationToken
     ) {
-        webviewView.onDidChangeVisibility(() => {
+        webviewView.onDidChangeVisibility(async () => {
             this.onDidChangeVisibility?.fire(webviewView.visible)
+            // force webview to reload
+            await vscode.commands.executeCommand('workbench.action.webview.reloadWebviewAction')
```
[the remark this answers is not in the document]
leigaol: Right. When you sign in/ sign out in toolkit, this Q login page needs to be updated.

packages/core/src/codewhisperer/activation.ts
```diff
@@ -271,6 +272,10 @@ export async function activate(context: ExtContext): Promise<void> {
     // we need to do it manually here because the Toolkit would have been unable to call
     // this API if the Q/CW extension started afterwards (and this code block is running).
     if (isExtensionInstalled(VSCODE_EXTENSION_ID.awstoolkit)) {
+        await activateExtension(VSCODE_EXTENSION_ID.awstoolkit)
```
hayemaxi, line 275: deadlock with https://github.com/aws/aws-toolkit-vscode/blob/feature/standalone/packages/core/src/amazonq/explorer/amazonQTreeNode.ts#L147

?

packages/core/src/auth/auth.ts
```diff
@@ -890,15 +891,81 @@ export class Auth implements AuthService, ConnectionManager {
...
+        const id = connection.id
+        const profile = this.store.getProfile(connection.id)
+        const newProfile = {
+            type: connection.type,
+            ssoRegion: connection.ssoRegion,
+            scopes: connection.scopes,
+            startUrl: connection.startUrl,
+        } as SsoProfile
+        if (profile === undefined) {
+            await this.store.addProfile(id, newProfile)
```
hayemaxi, line 923: so update is also upsert here? if Q adds a new connection, it will be registered in the toolkit?

if so, can we rename this function to upsert to be more clear

justinmk3: could name it "set" instead of "update". Reason I suggested "update" is because that's the convention used by `memento` / globalStorage

packages/core/src/codewhisperer/activation.ts
```diff
@@ -286,7 +291,46 @@ export async function activate(context: ExtContext): Promise<void> {
                     await getChatAuthState()
                 ).codewhispererChat
             )
+            // when changing connection state in Q, also change connection state in toolkit
```
hayemaxi, line 294: we need a way to register these callbacks if the toolkit is installed after Q. As discussed, we should put these registrations behind a command and a `once()`. Q can call its own command on startup if toolkit is already installed. Toolkit can `tryExecute()` this command on its activation, covering the case if it is installed later on.
I think this would also eliminate deadlock on line 275, and we can change line 274 to be `isExtensionActive()` instead.

packages/core/src/auth/auth.ts
hayemaxi: "set" is good

packages/core/src/codewhisperer/activation.ts
leigaol: This is fixed in latest commit

packages/core/src/auth/auth.ts
leigaol: Renamed into `setConnections`

packages/core/src/codewhisperer/activation.ts
leigaol: This is removed in latest commit.

packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
```diff
@@ -94,6 +92,24 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
...
+                    await AuthUtil.instance.secondaryAuth.useNewConnection(conn)
+                    await AuthUtil.instance.secondaryAuth.addScopes(conn, amazonQScopes)
+                } catch (e) {
+                    throw ToolkitError.chain(e, 'Failed to add Amazon Q scope', {
+                        code: 'FailedToConnect',
+                    })
+                }
+            } finally {
+                this.notifyToolkit()
+            }
```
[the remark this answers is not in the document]
leigaol: Fixed in latest commit

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -326,3 +328,79 @@ export const signoutCodeWhisperer = Commands.declare(
...
+    () => async () => {
+        // While the Q/CW exposes an API for the Toolkit to register callbacks on auth changes,
+        // we need to do it manually here because the Toolkit would have been unable to call
+        // this API if the Q/CW extension started afterwards (and this code block is running).
+        if (isExtensionInstalled(VSCODE_EXTENSION_ID.awstoolkit)) {
+            const auth = Auth.instance
+            const toolkitExt = vscode.extensions.getExtension(VSCODE_EXTENSION_ID.awstoolkit)
+            const toolkitApi = toolkitExt?.exports
+            getLogger().info(`Trying to register toolkit callback. Toolkit is installed.`)
+            const registerCallback = once(async () => {
```
hayemaxi, line 343: not 100% sure but `once()` may have to be in an outer scope. It looks like it is getting created each time the command is executed, which would make a new instance of a one-off function each time... which means it runs multiple times.

packages/core/src/extension.ts
```diff
@@ -224,6 +224,10 @@ export async function activate(context: vscode.ExtensionContext) {
         if (!isReleaseVersion()) {
             globals.telemetry.assertPassiveTelemetry(globals.didReload)
         }
+        // after toolkit is activated, ask Amazon Q to register toolkit api callbacks
+        setTimeout(async () => {
+            await Commands.tryExecute('aws.amazonq.refreshConnectionCallback')
+        }, 2000)
```
hayemaxi, line 230: may not be necessary (and could create a small but noticeable delay in the UI).

Can't we just call it immediately (maybe with void instead of await)? If Q is already activated, then we don't have to wait. If it isn't activated, then toolkit will finish activating first and Q will handle running this command. Maybe there could be a race condition if they are both activating at the same time?

leigaol: If I call it immediately, at that point toolkit is not activated, toolkit API is not ready. So Q will fail to register the connection Api hooks.

hayemaxi (approved):

packages/core/src/extension.ts
hayemaxi: is it enough to do a void instead of await without a timeout? then it is in the back of the event queue

leigaol: > is it enough to do a void instead of await without a timeout? then it is in the back of the event queue

No. I tried but this won't work.

hayemaxi: hmm stuff like this is the source of some subtle/rare issues. I would like to see if there is an alternative way to do this. Some ideas:
1. call register on auth connections hooks in each plugin. this will spam the command, but once should take care of it.
2. have `aws.amazonq.refreshConnectionCallback` accept an optional argument, which is the toolkit API. That way, if Q calls the command, then we can get the api from toolkit. If tookit calls the command, it can pass the API to Q without having to worry about having itselfs finish activation, e.g.


packages/core/src/extension.ts
```
import {toolkitApi} from '...'
...
await Commands.tryExecute('aws.amazonq.refreshConnectionCallback', toolkitApi)
``` 

packages/core/src/codewhisperer/commands/basicCommands.ts
```
export const registerToolkitApiCallback = Commands.declare(
    { id: 'aws.amazonq.refreshConnectionCallback' },
    () => async (toolkitApi?: any) => {
       if (tookitApi === undefined) {
          toolkitApi = extensions.get('aws-toolkit-vscode').api
       }  //  else use provided toolkitApi from arguments

      // call callbacks
       ...
    })
```

As I was typing I started to favor #2. I tested locally and you are able to pass complex objects (functions) through commands. Thoughts?

leigaol: This change is made in the latest commit. Tested for  below cases:
1. Install Q first, then install toolkit.
2. Install Toolkit, then install Q.

Both can share states.

hayemaxi:
Nice!

PR #4602, opened 2024-03-27
fix(amazonq): add progressUpdates to TransformationHub

dhasani23:
## Problem

Needed to add progress updates to the Transformation Hub. Builds off of this old PR (https://github.com/aws/aws-toolkit-vscode/pull/4575) which I closed because a lot of that work became unneeded & there were many merge conflicts; was easier to start over and add my desired changes on top of the latest work.

## Solution

Implement the above.

## License


packages/core/package.json
```diff
@@ -3753,12 +3753,6 @@
                     }
                 }
             },
-            {
-                "command": "aws.amazonq.startTransformationInHub",
-                "title": "Start Transformation",
-                "icon": "$(play)",
-                "enablement": "gumby.isTransformAvailable"
-            },
```
dhasani23, line 3761: The play button was removed a while ago; this was dead code.

packages/core/package.json
```diff
@@ -743,7 +743,7 @@
                     "type": "webview",
                     "id": "aws.amazonq.transformationHub",
                     "name": "Status",
-                    "when": "gumby.isTransformAvailable || gumby.isStopButtonAvailable"
+                    "when": "gumby.wasQCodeTransformationUsed"
```
dhasani23, line 746: Simplified this logic a lot; Transformation Hub simply shows once the user clicks on Transform or does /transform action from chat.

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -794,7 +802,6 @@ export async function getTransformationPlan(jobId: string) {
 
 export async function getTransformationSteps(jobId: string) {
     try {
-        await sleep(2000) // prevent ThrottlingException
```
dhasani23, line 797: When we poll, we do so every 5 seconds, so this was unnecessary.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -343,13 +333,14 @@ export async function setTransformationToRunningState() {
...
-
     await vscode.commands.executeCommand('aws.amazonq.refresh')
 }
 
 export async function postTransformationJob() {
-    await vscode.commands.executeCommand('setContext', 'gumby.isTransformAvailable', true)
+    transformByQState.getChatControllers()?.transformationFinished.fire({
+        jobStatus: transformByQState.getPolledJobStatus(),
+        tabID: ChatSessionManager.Instance.getSession().tabID,
+    })
```
dhasani23, line 343: Moved this in `postTransformationJob()` which is in a `finally` block above -- it always runs so that we can reset the job status in the chat, even after a job is stopped/cancelled, allowing users to start a new transformation right after stopping one. Previously, there was a bug which didn't allow you to start a new job.

packages/core/src/codewhisperer/service/transformByQHandler.ts
arangatang: Are you sure, I remember we added in an extra delay at one point when we transitioned to the next step (i.e. prepare -> planned or something like that)

Don't know if this is that one but just ensure that there is no way that another API is called prior to this within 1 second :)

packages/core/src/codewhisperer/models/constants.ts
```diff
@@ -415,13 +413,26 @@ export const uploadIntent = 'TRANSFORMATION'
...
+    'PLANNED',
+    'TRANSFORMING',
+    'TRANSFORMED',
+]
+
+// initial build succeeded
+export const validStatesForBuildSucceeded = ['PREPARED', 'PLANNING', 'PLANNED', 'TRANSFORMING', 'TRANSFORMED']
+
+// plan must be available
+export const validStatesForPlanGenerated = ['PLANNED', 'TRANSFORMING', 'TRANSFORMED']
```
arangatang, line 431: Should this not also contain Completed and partially completed?

packages/toolkit/package.json
```diff
@@ -52,8 +52,4185 @@
     ],
     "main": "./dist/src/main",
     "browser": "./dist/src/mainWeb",
-    "engines": "This field will be autopopulated from the core module during debugging and packaging.",
-    "contributes": "This field will be autopopulated from the core module during debugging and packaging.",
+    "engines": {
```
damntrecky, line 55: Question: This section is "re-generated" because of above `package.json` change?

justinmk3:
- Is there a user-facing change that should be mentioned in the changelog?
- Are the package.json changes intentional?

otherwise LGTM

packages/toolkit/package.json
dhasani23: No, I'm not sure why it's regenerated, but does not have to do with my `package.json` changes.

packages/core/src/codewhisperer/service/transformByQHandler.ts
dhasani23: Ok, in that case, `GetTransformationJob` may be running at same time, so adding back this 2s sleep for safety.

packages/core/src/codewhisperer/models/constants.ts
dhasani23: Not necessary, the TRANSFORMING stage will have been hit by then, we poll every 5s so there's no way that we would skip over TRANSFORMING, which lasts multiple to several+ minutes.

packages/toolkit/package.json
dhasani23: I will remove this `package.json` from this PR though.

JuliaABurch (approved):

.changes/next-release/Feature-c6272bd8-5538-4478-be87-490ada29be8e.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Q Code Transformation: show detailed progress updates in Transformation Hub"
```
justinmk3, line 3:
```suggestion
	"description": "Amazon Q Code Transformation: show detailed progress updates in Transformation Hub"
```

PR #4605, opened 2024-03-27
telemetry: aws_loginWithBrowser includes user polling

nkomonen-amazon:
## Problem:

Before, `aws_loginWithBrowser` metric only represented the opening of the browser window and not the part where users would login and confirm scopes. So if they failed somewhere in the browser, `aws_loginWithBrowser` would still result in `Succeeded`.

We did not have good insight in to the success of the part where we poll until the user completes the browser login.

## Solution:

Now, `aws_loginWithBrowser` will include the part where the user must login and accept scopes. Only once that is done do we emit Succeeded.

## License

### DIFF
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -60,6 +60,7 @@ import {
     scopesSsoAccountAccess,
 } from './connection'
 import { isSageMaker, isCloud9 } from '../shared/extensionUtilities'
+import { telemetry } from '../shared/telemetry/telemetry'
 
 interface AuthService {
     /**
@@ -170,7 +171,13 @@ export class Auth implements AuthService, ConnectionManager {
         const profile = this.store.getProfileOrThrow(id)
         if (profile.type === 'sso') {
             const provider = this.getSsoTokenProvider(id, profile)
-            await this.authenticate(id, () => provider.createToken())
+
+            // We create our own span with run() since we want to
+            // set the isReAuth attribute
+            await telemetry.aws_loginWithBrowser.run(async span => {
+                span.record({ isReAuth: true })
+                await this.authenticate(id, () => provider.createToken())
+            })
 
             return this.getSsoConnection(id, profile)
         } else {
--- a/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
+++ b/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
@@ -184,19 +184,31 @@ export class SsoAccessTokenProvider {
             clientSecret: registration.clientSecret,
         })
 
-        if (!(await openSsoPortalLink(this.profile.startUrl, authorization))) {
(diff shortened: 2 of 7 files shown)


justinmk3 (approved):
Nice

packages/core/src/auth/auth.ts
```diff
@@ -170,7 +171,13 @@ export class Auth implements AuthService, ConnectionManager {
...
         if (profile.type === 'sso') {
             const provider = this.getSsoTokenProvider(id, profile)
-            await this.authenticate(id, () => provider.createToken())
+
+            // We know createToken() will eventually emit for aws_loginWithBrowser, so we pre-emptively set isReAuth
+            // using the following:
+            await telemetry.aws_loginWithBrowser.run(async span => {
+                span.record({ isReAuth: true })
+                await this.authenticate(id, () => provider.createToken())
+            })
```
justinmk3, line 180: I think this means the `aws_loginWithBrowser.run()` in `packages/core/src/auth/sso/ssoAccessTokenProvider.ts` will be redundant, or "nested". Not a big deal, but ideally in `ssoAccessTokenProvider.ts` we would skip the redundant `aws_loginWithBrowser.run()` if the current scope/execution already is tagged as `aws_loginWithBrowser`.

justinmk3: Also because you are doing `span.record()` instead of `telemetry.record()`, I think the nested/redundant `aws_loginWithBrowser` won't have `isReAuth`.

It's easy to confirm/disconfirm this using the logs (e.g. `AWS (Developer): Watch Logs`).

nkomonen-amazon: Good catch. When I was testing before I missed the nested metric was also being emitted which is redundant.

I'm going with the `telemetry.activeSpan` check in SsoAccessTokenProvider to check if doing a `run()` is required

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -184,19 +184,31 @@ export class SsoAccessTokenProvider {
...
 
-        const tokenRequest = {
-            clientId: registration.clientId,
-            clientSecret: registration.clientSecret,
-            deviceCode: authorization.deviceCode,
-            grantType: deviceGrantType,
+        let token: ReturnType<typeof openBrowserAndWaitUntilComplete>
+        if (telemetry.spans.map(s => s.name).includes('aws_loginWithBrowser')) {
+            // During certain flows, eg reauthentication, we are already running within a span (run())
+            // so we don't need to create a new one.
```
justinmk3, line 205: Extending your dedup work in https://github.com/aws/aws-toolkit-vscode/pull/3994 , we could make this happen automatically for various metrics (such as `aws_loginWithBrowser`).

Actually, it's probably rare that we would want to nest a metric in *itself*.

PR #4610, opened 2024-03-28
fix(codecatalyst): use code catalyst sso login in new login view

leigaol:
## Problem

When signing in to Code Catalyst using IdC, Code Catalyst scope is missing.

## Solution

Add  Code Catalyst  scope by invoking
`await this.codeCatalystAuth.connectToEnterpriseSso(startUrl, region)`

## License

### DIFF
--- a/packages/core/src/awsexplorer/activation.ts
+++ b/packages/core/src/awsexplorer/activation.ts
@@ -129,6 +129,7 @@ export async function activate(args: {
                 retainContextWhenHidden: true,
             },
         }),
+        // Hacky way for a webview to call setLoginService().
         vscode.commands.registerCommand('aws.explorer.setLoginService', (serviceToShow?: string) => {
             if (toolkitAuthProvider.webView && 'setLoginService' in toolkitAuthProvider.webView.server) {
                 toolkitAuthProvider.webView.server.setLoginService(serviceToShow)
--- a/packages/core/src/awsexplorer/activation.ts
+++ b/packages/core/src/awsexplorer/activation.ts
@@ -30,6 +30,7 @@ import { S3FolderNode } from '../s3/explorer/s3FolderNode'
 import { amazonQNode, refreshAmazonQ, refreshAmazonQRootNode } from '../amazonq/explorer/amazonQTreeNode'
 import { GlobalState } from '../shared/globalState'
 import { activateViewsShared, registerToolView } from './activationShared'
+import { CommonAuthViewProvider } from '../login/webview'
 
 /**
  * Activates the AWS Explorer UI and related functionality.
@@ -120,6 +121,20 @@ export async function activate(args: {
     for (const viewNode of viewNodes) {
         registerToolView(viewNode, args.context.extensionContext)
     }
+
+    const toolkitAuthProvider = new CommonAuthViewProvider(args.context.extensionContext, 'toolkit')
+    args.context.extensionContext.subscriptions.push(
+        vscode.window.registerWebviewViewProvider(toolkitAuthProvider.viewType, toolkitAuthProvider, {
+            webviewOptions: {
+                retainContextWhenHidden: true,
+            },
+        }),
+        vscode.commands.registerCommand('aws.explorer.setLoginService', (serviceToShow?: string) => {
+            if (toolkitAuthProvider.webView && 'setLoginService' in toolkitAuthProvider.webView.server) {
+                toolkitAuthProvider.webView.server.setLoginService(serviceToShow)
+            }
+        })
+    )
 }
 
 async function registerAwsExplorerCommands(
--- a/packages/core/src/dev/activation.ts
+++ b/packages/core/src/dev/activation.ts
@@ -21,7 +21,6 @@ import { isReleaseVersion } from '../shared/vscode/env'
 import { isAnySsoConnection } from '../auth/connection'
 import { Auth } from '../auth/auth'
 import { getLogger } from '../shared/logger'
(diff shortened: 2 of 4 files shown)


packages/core/src/dev/activation.ts
```diff
@@ -160,6 +160,11 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
             webviewOptions: {
                 retainContextWhenHidden: true,
             },
+        }),
+        vscode.commands.registerCommand('aws.explorer.setLoginService', (serviceToShow?: string) => {
```
justinmk3, line 164: this file is for dev tools, it isn't activated if the user does not have `aws.dev.*` settings

packages/core/src/login/webview/vue/toolkit/backend_toolkit.ts
```diff
@@ -14,12 +14,28 @@ import { CodeCatalystAuthenticationProvider } from '../../../../codecatalyst/aut
...
     constructor(private readonly codeCatalystAuth: CodeCatalystAuthenticationProvider) {
         super(ToolkitLoginWebview.sourcePath)
     }
 
+    setLoginService(serviceToShow?: string) {
+        if (serviceToShow === 'codecatalyst') {
+            this.isCodeCatalystEnterpriseLogin = true
+        } else {
+            this.isCodeCatalystEnterpriseLogin = false
+        }
```
hayemaxi, line 28:
```suggestion
        this.isCodeCatalystEnterpriseLogin = serviceToShow === 'codecatalyst'
```

packages/core/src/login/webview/vue/toolkit/backend_toolkit.ts
```diff
@@ -14,12 +14,28 @@ import { CodeCatalystAuthenticationProvider } from '../../../../codecatalyst/aut
...
 
     constructor(private readonly codeCatalystAuth: CodeCatalystAuthenticationProvider) {
         super(ToolkitLoginWebview.sourcePath)
     }
 
+    setLoginService(serviceToShow?: string) {
+        if (serviceToShow === 'codecatalyst') {
+            this.isCodeCatalystEnterpriseLogin = true
+        } else {
+            this.isCodeCatalystEnterpriseLogin = false
```
justinmk3, line 27: the names are confusing. `setLoginService('codecatalyst')` sets a "enterprise" flag. why?

should `this.isCodeCatalystEnterpriseLogin` be named `this.isCodeCatalystLogin` ?

packages/core/src/dev/activation.ts
```diff
@@ -160,6 +160,11 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
+            if (toolkitAuthProvider.webView && 'setLoginService' in toolkitAuthProvider.webView.server) {
+                toolkitAuthProvider.webView.server.setLoginService(serviceToShow)
```
hayemaxi, line 166: how will this be reset from code catalyst, if the user starts using Q IdC for example

leigaol: This is reset everytime command `aws.${prefix}.auth.manageConnections` is triggered. This command runs whenever use clicks Sign-in button in toolkit or Add New Connection in Toolkit, which renders the LoginView.

leigaol: Ideally I should pass the code catalyst as a param into the webview but I did not find out an easy way.

packages/core/src/awsexplorer/activation.ts
```diff
@@ -120,6 +121,20 @@ export async function activate(args: {
...
     }
+
+    const toolkitAuthProvider = new CommonAuthViewProvider(args.context.extensionContext, 'toolkit')
+    args.context.extensionContext.subscriptions.push(
+        vscode.window.registerWebviewViewProvider(toolkitAuthProvider.viewType, toolkitAuthProvider, {
+            webviewOptions: {
+                retainContextWhenHidden: true,
+            },
+        }),
+        vscode.commands.registerCommand('aws.explorer.setLoginService', (serviceToShow?: string) => {
```
justinmk3, line 132: The purpose of this command is not clear. If it's a hacky way for a webview to call `setLoginService`, then a comment should mention that.
```suggestion
        // Hacky way for a webview to call setLoginService().
        vscode.commands.registerCommand('aws.explorer.setLoginService', (serviceToShow?: string) => {
```

justinmk3 (approved):

packages/core/src/awsexplorer/activation.ts
leigaol: I will try to implement a non-hacky way in next commit

packages/core/src/dev/activation.ts
leigaol: In the current UX, I have yet found out a way to do below flow

>the user starts using Q IdC 
> the user re-use this connection for CodeCatalyst

This flow currently does not exist. We can revisit this once we got more feedback.

packages/core/src/awsexplorer/activation.ts
leigaol: >If it's a hacky way for a webview to call setLoginService, then a comment should mention that.

Command `aws.${prefix}.auth.manageConnections` is the only caller which invokes `await vscode.commands.executeCommand('aws.explorer.setLoginService', serviceToShow)`. It is not a webview.

leigaol:
The best practice for solving this issue is to design specific buttons for CC login so that it's onClick method does not require any state manipulation. However, this is not aligned with the current design.

PR #4611, opened 2024-03-28
fix(cwspr): properly handle deletion from toolkit to q

leigaol:
## Problem

Deletion events from Toolkit to Amazon Q does not refresh Amazon Q UX.

## Solution

Let deletion events from Toolkit to Amazon Q refresh Amazon Q UX.

When Toolkit delete a connection, if that connection also exists in Q, then Q only removes this connection from momento DB and active connection state.  Q do not perform logout API call. Q also updates login UX and status bar.

If that connection is not in Q, do nothing. 

## License

### DIFF
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -946,18 +946,13 @@ export class Auth implements AuthService, ConnectionManager {
 
     // Used by Amazon Q to delete connection status & scope when this deletion is made by AWS Toolkit
     // NO event should be emitted from this deletion
-    public async deletionConnectionCallback(id: string) {
+    // Do not actually perform the delete because toolkit has done the deletion
+    // Delete the momento states only.
+    public async onDeleteConnection(id: string) {
         const profile = this.store.getProfile(id)
-        // it is possible the connection was already deleted
-        // but was still requested to be deleted. We pretend
-        // we deleted it and continue as normal
         if (profile) {
-            if (id === this.#activeConnection?.id) {
-                await this.store.setCurrentProfileId(undefined)
-            } else {
-                await this.invalidateConnection(id)
-            }
             await this.store.deleteProfile(id)
+            await this.store.setCurrentProfileId(undefined)
         }
     }
 }
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -929,7 +929,7 @@ export class Auth implements AuthService, ConnectionManager {
 
     // Used by Amazon Q to update connection status & scope when this connection is updated by AWS Toolkit
     // do not create connection in Q for each change event from Toolkit
-    public async updateConnectionCallback(connection: AwsConnection) {
+    public async onConnectionUpdate(connection: AwsConnection) {
         const conn = await this.getConnection({ id: connection.id })
         if (conn) {
             const profile = {
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -953,17 +953,11 @@ export class Auth implements AuthService, ConnectionManager {
         // we deleted it and continue as normal
         if (profile) {
             if (id === this.#activeConnection?.id) {
-                // Server-side invalidation.
-                await this.logout()
+                await this.store.setCurrentProfileId(undefined)
             } else {
                 await this.invalidateConnection(id)
             }
             await this.store.deleteProfile(id)
-            if (profile.type === 'sso') {
-                // There may have been linked IAM credentials attached to this
-                // so we will want to clear them.
-                await this.clearStaleLinkedIamConnections()
-            }
         }
     }
 }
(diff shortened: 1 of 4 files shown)


packages/core/src/auth/secondaryAuth.ts
```diff
@@ -274,4 +274,13 @@ export class SecondaryAuth<T extends Connection = Connection> {
             return conn
         }
     }
+
+    // Used by Amazon Q to delete connection status & scope when this deletion is made by AWS Toolkit
+    // NO event should be emitted from this deletion
+    public async deletionConnectionCallback(id: string) {
```
justinmk3, line 280:
```suggestion
    public async onDeleteConnection(id: string) {
```

packages/core/src/auth/secondaryAuth.ts
```diff
@@ -274,4 +274,13 @@ export class SecondaryAuth<T extends Connection = Connection> {
...
+        await this.auth.deletionConnectionCallback(id)
```
justinmk3, line 281:
```suggestion
        await this.auth.onDeleteConnection(id)
```

justinmk3 (approved):

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -174,6 +175,26 @@ export class AuthUtil {
         }
     }
 
+    /* Callback used by Amazon Q to delete connection status & scope when this deletion is made by AWS Toolkit
+     ** 1. NO event should be emitted from this deletion
+     ** 2. Should update the context key to update UX
+     */
+    public async deletionConnectionCallback(id: string) {
```
hayemaxi, line 182:
```suggestion
    public async onDeleteConnection(id: string) {
```

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -174,6 +175,26 @@ export class AuthUtil {
...
+        await this.secondaryAuth.onDeleteConnection(id)
+        await this.setVscodeContextProps()
+        await vscode.commands.executeCommand('aws.codeWhisperer.refreshStatusBar')
+    }
+
+    /* Callback used by Amazon Q to delete connection status & scope when this deletion is made by AWS Toolkit
+     ** 1. NO event should be emitted from this deletion
+     ** 2. Should update the context key to update UX
+     */
+    public async updateConnectionCallback(connection: AwsConnection) {
```
hayemaxi, line 192:
```suggestion
    public async onUpdateConnection(connection: AwsConnection) {
```

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -370,12 +370,12 @@ const registerToolkitApiCallbackOnce = once(async () => {
         _toolkitApi.onDidChangeConnection(
             async (connection: AwsConnection) => {
                 getLogger().info(`toolkitApi: connection change callback ${connection.id}`)
-                await auth.updateConnectionCallback(connection)
+                await AuthUtil.instance.updateConnectionCallback(connection)
```
hayemaxi, line 373:
```suggestion
                await AuthUtil.instance.onUpdateConnection(connection)
```

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -370,12 +370,12 @@ const registerToolkitApiCallbackOnce = once(async () => {
...
             },
 
             async (id: string) => {
                 getLogger().info(`toolkitApi: connection delete callback ${id}`)
-                await auth.deletionConnectionCallback(id)
+                await AuthUtil.instance.deletionConnectionCallback(id)
```
hayemaxi, line 378:
```suggestion
                await AuthUtil.instance. onDeleteConnection(id)
```

packages/core/src/auth/auth.ts
```diff
@@ -946,24 +946,13 @@ export class Auth implements AuthService, ConnectionManager {
...
     // NO event should be emitted from this deletion
-    public async deletionConnectionCallback(id: string) {
+    // Do not actually perform the delete because toolkit has done the deletion
+    // Delete the momento states only.
+    public async onDeleteConnection(id: string) {
         const profile = this.store.getProfile(id)
-        // it is possible the connection was already deleted
-        // but was still requested to be deleted. We pretend
-        // we deleted it and continue as normal
         if (profile) {
```
hayemaxi, line 953: when debugging this, I noticed that this if statement never caught because in Q's `this.store`, the connection was already gone. It's like the toolkit is updating the store for Q, which it shouldn't have access to. Haven't narrowed down why this is the case, but I am surprised that this works given that.

leigaol: I ran a debugger and did not encounter the "connection already gone" issue after I restarted my VSC.

hayemaxi: :o

hayemaxi (approved):

packages/core/src/auth/auth.ts
```diff
@@ -946,24 +946,13 @@ export class Auth implements AuthService, ConnectionManager {
 
     // Used by Amazon Q to delete connection status & scope when this deletion is made by AWS Toolkit
```
justinmk3, line 947: Ideally, any Q-only logic would be in a Q-only module. Not sure if that's possible currently.

PR #4612, opened 2024-03-29
fix(q): toolkit explorer updates if toolkit pushes connection changes

hayemaxi:
**Squash and merge with this commit message**

- Stop using an API to communicated with Q about the status of its connection (this is not the same API that is used to shared connections)
  - Communication is now done via commands.
- Remove calls to refreshing the Q explorer tree- they were unnecessary
- Refactor Amazon Q tree node class, use instance instead
- Add some fault tolerance to switch to the Q login/chat window from the explorer tree
- Refactor AuthUtil to correctly house the function `getChatAuthState`. Also allows us to add an optional parameter to it that lets us avoid doing connection state changes.

## License


packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -371,11 +371,23 @@ const registerToolkitApiCallbackOnce = once(async () => {
             async (connection: AwsConnection) => {
                 getLogger().info(`toolkitApi: connection change callback ${connection.id}`)
                 await AuthUtil.instance.onUpdateConnection(connection)
+                await vscode.commands.executeCommand(
+                    'aws.amazonq.refresh',
+                    (
+                        await getChatAuthState()
```
leigaol, line 377: Q always pushes connection status to toolkit using `setConnection`, therefore, without this command, toolkit still knows the status of Q connection.

hayemaxi: I am trying it locally without this change and it is not updating for me

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -418,55 +418,59 @@ export class AuthUtil {
...
+        }
 
-    if (isBuilderIdConnection(currentConnection)) {
-        // Regardless, if using Builder ID, Amazon Q is unsupported
-        state[Features.amazonQ] = AuthStates.unsupported
-    }
+        // The state of the connection may not have been properly validated
+        // and the current state we see may be stale, so refresh for latest state.
+        if (shouldRefresh) {
+            await this.auth.refreshConnectionState(currentConnection)
```
leigaol, line 441: If this refresh emit connection state change event, can it run into a infinite loop?

leigaol: Example: If it refreshConnectionState ---> emit connection changed event ---> connection change hook ---> `refreshConnectionState`

packages/core/src/amazonq/explorer/amazonQChildrenNodes.ts
```diff
@@ -46,10 +46,13 @@ export const switchToAmazonQCommand = Commands.declare(
             })
             if (signIn) {
                 await vscode.commands.executeCommand('setContext', 'aws.amazonq.showLoginView', true)
-                await vscode.commands.executeCommand('aws.amazonq.AmazonCommonAuth.focus')
-            } else {
-                await vscode.commands.executeCommand('aws.AmazonQChatView.focus')
             }
+
+            // Attempt to show both, in case something is wrong with the state of the tree. Only the active
```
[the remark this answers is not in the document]
hayemaxi: there are now more changes, listed in desc.

packages/core/src/codewhisperer/util/authUtil.ts
hayemaxi: no it should converge, because each time it is called it compares to old values.

packages/amazonq/src/extension.ts
```diff
@@ -5,11 +5,9 @@
 
 import * as vscode from 'vscode'
 import { activateShared, deactivateShared } from './extensionShared'
-import { amazonQApi } from './api'
```
leigaol, line 8: Good.

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -542,3 +550,25 @@ function getEnvType(): keyof HasAlreadySeenQWelcome {
...
+/**
+ * Refreshes toolkit's Amazon Q tree view with the current Amazon Q connection state.
+ *
+ * `getChatAuthState()` has the ability to update the active connection/state. If this
+ * is called in a connection update callback, we could potentially be in an infinite loop.
+ * However, the callbacks only trigger if there is a change to the active connection/state.
+ * This means that our loop would converge immediately, or within a few iterations of the
+ * state is being updated rapidly due to race conditions.
+ */
+export const refreshToolkitQState = Commands.declare(
```
leigaol, line 563: This authUtil is used by both plugins, that means the commands.declare can only happen in one extension. If both extension activates, they may compete and one will fail to declare this command.

The intention is to let toolkit own this command. Can you make sure this command is owned by toolkit?

hayemaxi: this is actually intentional. multiple extensions can declare the same command, but it will only be an issue if both try to register it. registration will only happen in CW activation.

packages/core/src/amazonq/explorer/amazonQTreeNode.ts
```diff
@@ -34,7 +34,13 @@ export class AmazonQNode implements TreeNode {
 
     public static amazonQState: AuthState
 
-    constructor() {}
+    private constructor() {
+        if (isExtensionActive(VSCODE_EXTENSION_ID.amazonq)) {
+            // 'void' instead of await, so that the command call doesn't trigger an infinite loop
+            // on constructing these instances.
+            void vscode.commands.executeCommand('aws.amazonq.refreshToolkitQState')
```
leigaol, line 41: nit: consider rename 'aws.amazonq.refreshToolkitQState' into 'aws.toolkit.refreshToolkitQState' 

Because this command is owned by toolkit and it updates states in aws toolkit. 

Also double check isExtensionActive does not throw if Q is not installed.

packages/core/src/amazonq/explorer/amazonQTreeNode.ts
```diff
@@ -117,33 +124,25 @@ function createNewMenuButton(): TreeNode<Command> {
     })
 }
 
-export const amazonQNode = new AmazonQNode()
 export const refreshAmazonQ = (provider?: ResourceTreeDataProvider) =>
-    Commands.register({ id: 'aws.amazonq.refresh', logging: false }, (state?: AuthState) => {
+    Commands.register({ id: 'aws.toolkit.amazonq.refresh', logging: false }, (state?: AuthState) => {
         if (state) {
             AmazonQNode.amazonQState = state
+            getLogger().debug(`aws.amazonq.refresh called, updating state to ${state}.`)
```
leigaol, line 131: The git logger says ``aws.amazonq.refresh` but the command is actually 'aws.toolkit.amazonq.refresh'

packages/core/src/amazonq/explorer/amazonQTreeNode.ts
hayemaxi: verified isExtensionActive will not throw.

Will leave the command name though, because Q owns it and registers it. Toolkit only owns and registers `aws.toolkit.amazonq.refresh`

packages/core/src/amazonq/explorer/amazonQTreeNode.ts
leigaol: Consider rename 'aws.toolkit.amazonq.refresh' into 'aws.toolkit.refreshQTreeNode' or something similar

packages/core/src/codewhisperer/util/authUtil.ts
leigaol: >registration will only happen in CW activation.
That's good.

packages/core/src/amazonq/explorer/amazonQTreeNode.ts
hayemaxi: done

leigaol (approved):

PR #4616, opened 2024-03-29
fix(CodeCatalyst): remove redundant bearer token header

aarolieb:
## Problem
aws-sdk automatically adds the bearer token to requests during the signing process. The CodeCatalystClientInternal `call()` function manually adds the bearer token to the header which is unnecessary and may cause confusion.

## Solution
Remove the header assignment

## License

### DIFF
--- a/packages/core/src/shared/clients/codecatalystClient.ts
+++ b/packages/core/src/shared/clients/codecatalystClient.ts
@@ -231,7 +231,6 @@ class CodeCatalystClientInternal {
     private async call<T>(req: AWS.Request<T, AWS.AWSError>, silent: boolean, defaultVal?: T): Promise<T> {
         const log = this.log
         const bearerToken = (await this.connection.getToken()).accessToken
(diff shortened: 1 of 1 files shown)


justinmk3:
Can you confirm that you tested this?

/runintegrationtests

aarolieb:
> Can you confirm that you tested this?
> 
> /runintegrationtests

@justinmk3 Yes, I tested it. Used a branch without it for multiple weeks. I confirmed that the bearer token is added during the signing process, removed it, and stayed authenticated. The integration tests look good too. Can you confirm that the one integration test that failed is not a result of the change? It seems that at least one integration fails in all of the recent PRs. I'm not in the loop regarding this consistent time out failure in the integ tests

justinmk3:
Thanks for the attention to detail!

justinmk3:
> aws-sdk automatically adds the bearer token to requests during the signing process

Where does that actually happen? I see lots of code related to "caching", but I don't see where the token is given to the SDK.

When making changes like this, it's worth being explicit.

PR #4619, opened 2024-03-29
fix(amazonq): allow 'start a new transformation' button to work and p…

dhasani23:
…revent duplicate chat messages

## Problem

"Start a new transformation" button in our chat was not working. Also, duplicate chat messages were being sent when a job is stopped.

## Solution

Reset ConversationState to fix first issue. Get rid of duplicate messages to fix second issue.

## License

### DIFF
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -220,7 +220,6 @@ export class GumbyController {
                 break
             case ButtonActions.STOP_TRANSFORMATION_JOB:
                 await stopTransformByQ(transformByQState.getJobId(), CancelActionPositions.Chat)
-                this.messenger.sendJobFinishedMessage(message.tabId, true)
                 break
             case ButtonActions.CONFIRM_START_TRANSFORMATION_FLOW:
                 this.messenger.sendCommandMessage({ ...message, command: GumbyCommands.CLEAR_CHAT })
@@ -262,6 +261,8 @@ export class GumbyController {
             await compileProject()
         } catch (err: any) {
             this.messenger.sendRetryableErrorResponse('could-not-compile-project', message.tabID)
+            // reset state to allow "Start a new transformation" button to work
+            this.sessionStorage.getSession().conversationState = ConversationState.IDLE
             throw err
         }
 
@@ -304,7 +305,7 @@ export class GumbyController {
 
     private async transformationFinished(message: { tabID: string; jobStatus: string }) {
         this.sessionStorage.getSession().conversationState = ConversationState.IDLE
-        this.messenger.sendJobSubmittedMessage(message.tabID, true)
+        // at this point job is either completed, partially_completed, cancelled, or failed
         this.messenger.sendJobFinishedMessage(message.tabID, false, message.jobStatus)
     }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -220,7 +220,6 @@ export class GumbyController {
                 break
             case ButtonActions.STOP_TRANSFORMATION_JOB:
                 await stopTransformByQ(transformByQState.getJobId(), CancelActionPositions.Chat)
-                this.messenger.sendJobFinishedMessage(message.tabId, true)
```
dhasani23, line 223: Was causing an extra message to show when you stop a job. "transformFinished" gets fired at end of job which already handles this.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -304,7 +305,7 @@ export class GumbyController {
 
     private async transformationFinished(message: { tabID: string; jobStatus: string }) {
         this.sessionStorage.getSession().conversationState = ConversationState.IDLE
-        this.messenger.sendJobSubmittedMessage(message.tabID, true)
```
dhasani23, line 307: This was a bug. It wasn't showing in the chat because the message immediately below gets sent right after which masks the first one, but this shouldn't be showing in `transformationFinished` -- this already shows when you start a job.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -325,7 +326,6 @@ export class GumbyController {
                     })
                 } else {
                     this.messenger.sendRetryableErrorResponse('invalid-java-home', data.tabID)
-                    this.messenger.sendJobFinishedMessage(data.tabID, true, undefined)
```
dhasani23, line 328: Line right above already sends a chat message, not necessary for 2.

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -348,7 +348,7 @@ For more information, see the [Amazon Q documentation.](https://docs.aws.amazon.
 
         if (!cancelled) {
             message =
-                'The transformation job has been completed. If you want to start another transformation, choose **Start a new transformation.**'
+                'The transformation job is over. If you want to start another transformation, choose **Start a new transformation.**'
```
dhasani23, line 351: We can follow up on exact wording later. Right now, this text is shown when a job fails (ex. network error when polling) and also at end of a successful job, so need to make it a bit more generic so as to not be misleading. Long-term fix, we will use the job status to customize the exact message shown.

damntrecky (approved):

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -262,6 +261,8 @@ export class GumbyController {
             await compileProject()
         } catch (err: any) {
             this.messenger.sendRetryableErrorResponse('could-not-compile-project', message.tabID)
```
damntrecky, line 263: Note: This line below is the main fix to allow chat to work again.

dhasani23:
The current behavior / bug is this when the Maven commands fail and you click the "Start a new transformation" button:

image

PR #4621, opened 2024-04-01
auth: add more network related errors to check

nkomonen-amazon:
## Problem:

We have a function `isNetworkError()` that determines if a an error is network related.

The issue is that we can see other network related errors in our telemetry metric `aws_refreshCredentials` that are not caught by `isNetworkError()`.

## Solution:

Add some of the network errors that we see in telemetry to the isNetworkError check.

This will improve false token invalidations during the sso token refresh process since we were previously invalidating incorrectly on certain network issues

## License

### DIFF
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -550,7 +550,6 @@ export class Auth implements AuthService, ConnectionManager {
     }
 
     private async handleSsoTokenError(id: Connection['id'], err: unknown) {
-        // Bubble-up networking issues so we don't treat the session as invalid
         this.throwOnNetworkError(err)
 
         this.#validationErrors.set(id, UnknownError.cast(err))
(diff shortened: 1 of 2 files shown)


packages/core/src/auth/auth.ts
```diff
@@ -728,13 +726,13 @@ export class Auth implements AuthService, ConnectionManager {
...
+     * Auth processes can fail due to reasons outside of authentication actually being expired.
+     * We do not want to intepret these failures as invalid/expired auth tokens.
      *
-     * We use this to check if the given error is network related and then
-     * throw, expecting the caller to not change the state of the connection.
+     * We use this to check if the given error is unanticipated. If it is we throw,
+     * expecting the caller to not change the state of the connection.
      */
-    private throwOnNetworkError(e: unknown) {
+    private throwOnUnanticipatedError(e: unknown) {
```
hayemaxi, line 735: `throwOnNetworkError` was a good name, since that is the only thing its checking for.

nkomonen-amazon: Makes sense. Was going to add other types of errors but ended up not.

hayemaxi (approved):

packages/core/src/auth/auth.ts
```diff
@@ -725,7 +724,6 @@ export class Auth implements AuthService, ConnectionManager {
     private readonly getToken = keyedDebounce(this._getToken.bind(this))
     private async _getToken(id: Connection['id'], provider: SsoAccessTokenProvider): Promise<SsoToken> {
         const token = await provider.getToken().catch(err => {
-            // Bubble-up networking issues so we don't treat the session as invalid
             this.throwOnNetworkError(err)
```
justinmk3, line 727: Is this comment no longer valid? Seems like a helpful hint.

PR #4623, opened 2024-04-01
fix(commands): update command handling and ellipses menus for q, toolkit explorer

hayemaxi:
- Hide unnecessary auth commands in amazon q (we should evaluate and remove them in the future)
- Add Q versions of generic toolkit commands for: feedback, going to github, about extension, submit issue, etc.
  - This means the status bar menu github button works now.
- Ellipses for Q menu now independent of the toolkit ext.
- Toolkit explorer buttons no longer throw missing command errors.
- Cleanup package.json to remove unneeded or unused context values.

## License

### DIFF
--- a/packages/amazonq/package.json
+++ b/packages/amazonq/package.json
@@ -258,15 +258,20 @@
                     "when": "false"
                 },
                 {
+                    "command": "aws.amazonq.auth.switchConnections",
+                    "when": "false"
+                },
+                {
+                    "command": "aws.amazonq.auth.manageConnections",
+                    "when": "false"
+                },
+                {
                     "command": "aws.amazonq.auth.signout",
                     "when": "false"
                 },
                 {
                     "command": "aws.amazonq.auth.help",
                     "when": "false"
-                },
-                {
-                    "command": "aws.amazonq.auth.manageConnections"
                 },
                 {
                     "command": "aws.dev.openMenu",
@@ -329,11 +334,6 @@
                     "command": "aws.amazonq.learnMore",
                     "when": "view == aws.AmazonQChatView || view == aws.amazonq.AmazonCommonAuth",
                     "group": "1_amazonQ@1"
-                },
-                {
-                    "command": "aws.codewhisperer.manageConnections",
-                    "when": "(view == aws.AmazonQChatView) && !aws.codewhisperer.connected",
-                    "group": "2_amazonQ@2"
                 },
                 {
                     "command": "aws.codewhisperer.signout",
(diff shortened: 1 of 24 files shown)


packages/amazonq/package.json
```diff
@@ -454,13 +441,13 @@
                 "command": "aws.amazonq.login",
                 "title": "%AWS.command.login%",
                 "category": "%AWS.title%",
-                "enablement": "isCloud9 || !aws.isWebExtHost",
```
leigaol, line 457: Can you double check whether Q standalone should be supported in C9 and web mode? If yes, then this PR needs to be tested against C9 and web

packages/core/src/feedback/vue/submitFeedback.ts
```diff
@@ -69,16 +69,33 @@ export class FeedbackWebview extends VueWebview {
 
 type FeedbackId = 'AWS Toolkit' | 'CodeWhisperer' | 'Amazon Q' | 'Application Composer'
```
leigaol, line 70: Is 'CodeWhisperer' | 'Amazon Q'  supposed to be merged?

packages/core/package.json
```diff
@@ -1245,13 +1246,13 @@
                     "group": "3_lambda@3"
                 },
                 {
-                    "submenu": "aws.submenu.feedback",
-                    "when": "view =~ /^aws\\./",
+                    "submenu": "aws.toolkit.submenu.feedback",
+                    "when": "view =~ /^aws\\./ && view != aws.AmazonQChatView && view != aws.amazonq.AmazonCommonAuth",
```
leigaol, line 1250: Is it necessary to add `view != aws.AmazonQChatView && view != aws.amazonq.AmazonCommonAuth`? This command belongs to Toolkit, it should not depend on Q states.

hayemaxi: Yes actually. Any extension can add views, submenus, options, etc to any other extension- which seems silly at first but makes sense in certain scenarios. If we don't add this, then toolkit will put a duplicate Feedback... submenu item in the ellipses menu for Q.

packages/core/src/feedback/vue/submitFeedback.ts
hayemaxi: No, out of scope for now.

packages/amazonq/package.json
hayemaxi: Q does not need to be supported in C9. Webmode is uncertain, but out of scope for the immediate future. We will revisit these commands.

leigaol (approved):

PR #4625, opened 2024-04-02
auth: add more info when connecting to auth

nkomonen-amazon:
This PR does the following:
- Requires a `source` argument for partner teams using the VS Code command `aws.codeWhisperer.connect`
  - We will use this to help identify which external extension is using the toolkit to connect
- In the metric `aws_loginWithBrowser` we are adding in the attribute `credentialStartUrl`
  - This will help us determine which type of SSO the user is using when they go through the browser login flow for SSO

## License

### DIFF
--- a/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
+++ b/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
@@ -199,16 +199,25 @@ export class SsoAccessTokenProvider {
             return await pollForTokenWithProgress(() => this.oidc.createToken(tokenRequest), authorization)
         }
 
-        let token: ReturnType<typeof openBrowserAndWaitUntilComplete>
-        if (telemetry.spans.map(s => s.name).includes('aws_loginWithBrowser')) {
-            // During certain flows, eg reauthentication, we are already running within a span (run())
-            // so we don't need to create a new one.
-            token = openBrowserAndWaitUntilComplete()
-        } else {
-            token = telemetry.aws_loginWithBrowser.run(async () => openBrowserAndWaitUntilComplete())
-        }
+        const token = this.withBrowserLoginTelemetry(() => openBrowserAndWaitUntilComplete())
 
         return this.formatToken(await token, registration)
+    }
+
+    /**
+     * Wraps the given function with telemetry related to the browser login.
+     */
+    private withBrowserLoginTelemetry<T extends (...args: any[]) => any>(func: T): ReturnType<T> {
+        if (telemetry.spans.some(s => s.name === 'aws_loginWithBrowser')) {
+            // During certain flows, eg reauthentication, we are already running within a span (run())
+            // so we don't need to create a new one.
+            return func()
+        }
+
+        return telemetry.aws_loginWithBrowser.run(span => {
+            span.record({ credentialStartUrl: this.profile.startUrl })
+            return func()
+        })
     }
 
     /**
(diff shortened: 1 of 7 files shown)


packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -143,48 +143,57 @@ export const showSsoSignIn = Commands.declare('aws.codeWhisperer.sso', () => asy
 // Shortcut command to directly connect to Identity Center or prompt start URL entry
 // It can optionally set a customization too based on given values to match on
 export const connectWithCustomization = Commands.declare(
-    'aws.codeWhisperer.connect',
```
nkomonen-amazon, line 146: This changed. The other changes are formatting related

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -143,48 +143,57 @@ export const showSsoSignIn = Commands.declare('aws.codeWhisperer.sso', () => asy
...
-        //  * customizationNamePrefix: select customization by prefix, if `customizationArn` is `undefined`.
-        if (startUrl && region) {
-            await connectToEnterpriseSso(startUrl, region)
-        } else {
-            await getStartUrl()
-        }
+    { id: 'aws.codeWhisperer.connect', compositeKey: { 1: 'source' } },
+    () =>
+        async (
+            source: CodeWhispererSource,
```
nkomonen-amazon, line 149: This arg was added in. The other changes are formatting related.

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -143,48 +143,60 @@ export const showSsoSignIn = Commands.declare('aws.codeWhisperer.sso', () => asy
...
+     * This command supports the following arguments:
+     * @param source - an identifier for who used this command. This value is not explicitly used in the function, but is used elsewhere.
+     * startUrl and region. If both arguments are provided they will be used, otherwise
+     *  the command prompts for them interactively.
+     * customizationArn: select customization by ARN. If provided, `customizationNamePrefix` is ignored.
+     * customizationNamePrefix: select customization by prefix, if `customizationArn` is `undefined`.
+     */
+    () =>
+        async (
+            source: string,
```
nkomonen-amazon, line 157: This was added in. The other changes are formatting related

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -199,16 +199,25 @@ export class SsoAccessTokenProvider {
...
+        const token = this.withBrowserLoginTelemetry(() => openBrowserAndWaitUntilComplete())
+
+        return this.formatToken(await token, registration)
+    }
+
+    /**
+     * Wraps the given function with telemetry related to the browser login.
+     */
+    private withBrowserLoginTelemetry<T extends (...args: any[]) => any>(func: T): ReturnType<T> {
+        if (telemetry.spans.some(s => s.name === 'aws_loginWithBrowser')) {
```
justinmk3, line 211: If this is true, should we still call

    telemetry.record({ credentialStartUrl: this.profile.startUrl })

? That will update/ovewrite the `credentialStartUrl` field for the current span(s).

justinmk3 (approved):

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
nkomonen-amazon: We're making the assumption that the parent `aws_loginWithBrowser` span (in `auth.ts`) has already set it.

I wanted to keep all values set in the parent one unless we had a reason not to.

PR #4627, opened 2024-04-02
ci(amazonq): update for beta publishing

hayemaxi:
## License

### DIFF
--- a/packages/amazonq/package.json
+++ b/packages/amazonq/package.json
@@ -1,37 +1,21 @@
 {
     "name": "amazon-q-vscode",
-    "displayName": "Amazon Q (Preview) + CodeWhisperer",
-    "description": "Amazon Q (Preview) + CodeWhisperer",
-    "version": "1.0.0-SNAPSHOT",
+    "displayName": "Standalone Test",
+    "description": "",
+    "version": "0.1.0-SNAPSHOT",
     "extensionKind": [
         "workspace"
     ],
     "publisher": "amazonwebservices",
     "license": "Apache-2.0",
-    "repository": {
-        "type": "git",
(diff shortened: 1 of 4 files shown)


packages/amazonq/package.json
```diff
@@ -1,37 +1,21 @@
 {
     "name": "amazon-q-vscode",
-    "displayName": "Amazon Q (Preview) + CodeWhisperer",
-    "description": "Amazon Q (Preview) + CodeWhisperer",
-    "version": "1.0.0-SNAPSHOT",
+    "displayName": "Standalone Test",
+    "description": "",
+    "version": "0.1.0-SNAPSHOT",
```
justinmk3, line 5: Good idea. I noticed that MS marketplace doesn't allow you to re-publish an existing version, so if we were to publish 1.0.0 we could not overwrite it.

packages/amazonq/package.json
```diff
@@ -1,37 +1,21 @@
...
     "extensionKind": [
         "workspace"
     ],
-    "publisher": "amazonwebservices",
+    "publisher": "amazonwebservices-internal",
```
justinmk3, line 9: Hmm. I was expecting that we would publish under `amazonwebservices`.

I guess we could try this.  Then we need to update https://github.com/aws/aws-toolkit-vscode/blob/c4767fce31e7e914b8287ebf46e27bbe3375cad5/packages/core/src/shared/extensions.ts#L17 , and possibly other things

justinmk3 (approved):

packages/amazonq/package.json
hayemaxi: This is what I saw in the mock page, I'll just change it back.

PR #4629, opened 2024-04-03
feat: add util class to create in memory zip streams

ctlai95:
## Problem

A generic utility class to create in-memory zip streams.

## Solution

Example:
```ts
const zipStream = new ZipStream()
zipStream.writeString('some text', 'file.txt')
const { buffer } = zipStream.finalize();

...
await requests.fetch('PUT', 'some-upload-url' , {
  ...
  body: buffer
})
```

Uses `archiver` library internally since it is designed for streaming and has 6M+ weekly installs.

## License


packages/core/src/shared/utilities/zipStream.ts
```diff
@@ -0,0 +1,77 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import archiver from 'archiver'
+import { WritableStreamBuffer } from 'stream-buffers'
```
justinmk3, line 7: Is stream-buffers usable in web browser (not nodejs)?

packages/core/src/shared/utilities/zipStream.ts
```diff
@@ -0,0 +1,77 @@
...
+import { getLogger } from '../logger'
+
+export interface ZipStreamResult {
+    sizeInBytes: number
+    md5: string
+    buffer: Buffer
+}
+
+/**
+ * Utility class to create in-memory zip archives that outputs to a stream buffer.
```
justinmk3, line 18:
```suggestion
 * Creates in-memory zip archives that output to a stream buffer.
```

packages/core/src/shared/utilities/zipStream.ts
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * Example usage:
```
justinmk3, line 20: Always helpful, thanks!

packages/core/src/shared/utilities/zipStream.ts
```diff
@@ -0,0 +1,77 @@
...
+ * console.log(result) // { sizeInBytes: ..., md5: ..., buffer: ... }
+ * ```
+ */
+export class ZipStream {
+    private _archive: archiver.Archiver
+    private _outputBuffer: WritableStreamBuffer
+    private _hasher: crypto.Hash
+
+    constructor() {
+        this._archive = archiver('zip')
```
justinmk3, line 35: Can you mention in the PR description why you chose this vs other libraries?

packages/core/src/shared/utilities/zipStream.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    public writeFile(file: string, path: string) {
+        this._archive.file(file, { name: path })
+    }
+
+    public finalize(): Promise<ZipStreamResult> {
+        return new Promise((resolve, reject) => {
+            void this._archive.finalize()
+            this._archive.on('finish', () => {
+                const buffer = this._outputBuffer.getContents()
```
justinmk3, line 63: Looks like this loads the contents fully in memory. It's nice that we avoid the need for a temp directory, but ideally, the caller could choose to stream the zip file itself to the filesystem, so that the full contents (input or output) are never in memory. That allows arbitrarily large zip results without using lots of memory.

Is this possibe as a followup?

justinmk3 (approved):
This is a good start! My comments aren't blockers, but they are important

packages/core/src/shared/utilities/zipStream.ts
ctlai95: I couldn't find much info, but I think it won't work natively in the browser without some additional configuration. Let me see if anything else is available as a follow up

packages/core/src/shared/utilities/zipStream.ts
ctlai95: That's true, moving the `getContents()` part out of this class and leaving it to the caller to do something with the stream

PR #4632, opened 2024-04-04
fix(codewhisperer): inline tutorial telemetry

Will-ShaoHua:
## Problem
telemetry ui_click `codewhisperer_learnmore_case*` should ONLY be sent if users complete the corresponding case/step. However

- `codewhisperer_learnmore_case_1` is sent as soon as user has seen it
  - root cause: `seen.has(this.currentState)` will allow case_1's telemetry sent and not blocking it because either case_1a or case_2 is not in the set of `seen` https://github.com/aws/aws-toolkit-vscode/pull/4632/files#diff-9c8a244cafa64940c3c9f141252d08c61e9e6e79cf5e84773548bd1961f3591eL432

- `codewhisperer_learnmore_case_3` is *not* sent if the step is fulfilled by 30s timeout (case_3 has two exit criteria: one is to click the statusBar and the other is it will disappear itself after 30 seconds)
  - root cause: invocation of `markTutorialDone` will immediately stop any logic being executed within `updateDecoration()` https://github.com/aws/aws-toolkit-vscode/pull/4632/files#diff-9c8a244cafa64940c3c9f141252d08c61e9e6e79cf5e84773548bd1961f3591eL393

## Before
Case 1: telemetry will be sent out once case_1 is rendered (which is supposed to be sent out ONLY if case_1 is "completed" by accepting a recommendation)

https://github.com/aws/aws-toolkit-vscode/assets/96078566/6f82cc25-ddf9-40b2-abab-04f1eca45311

Case 3: timeout will not send telemetry

https://github.com/aws/aws-toolkit-vscode/assets/96078566/bd2bee23-6dff-4537-9a27-aecd429d5e9d

## After

Case 1: accept a recommendation

https://github.com/aws/aws-toolkit-vscode/assets/96078566/6953a687-1f85-4a09-981c-acc826b97680

Case 2: manual trigger

https://github.com/aws/aws-toolkit-vscode/assets/96078566/a15fe68d-6ffb-42a8-a037-c07f04e8f819

Case 3: 30s timeout

https://github.com/aws/aws-toolkit-vscode/assets/96078566/1bb49e6f-6f62-4c5f-ae94-9689aab001d1

Case 3: click status bar

https://github.com/aws/aws-toolkit-vscode/assets/96078566/4e76839d-d244-4d67-af2b-d8f1595c3d71

## License

### DIFF
--- a/packages/core/src/codewhisperer/views/lineAnnotationController.ts
+++ b/packages/core/src/codewhisperer/views/lineAnnotationController.ts
@@ -47,7 +47,7 @@ interface AnnotationState {
 
     text: () => string
     updateState(changeSource: AnnotationChangeSource, force: boolean): AnnotationState | undefined
-    isNextState(state: AnnotationState): boolean
+    isNextState(state: AnnotationState | undefined): boolean
 }
 
 /**
@@ -79,7 +79,7 @@ class AutotriggerState implements AnnotationState {
         }
     }
 
-    isNextState(state: AnnotationState): boolean {
+    isNextState(state: AnnotationState | undefined): boolean {
         return state instanceof ManualtriggerState
     }
 }
@@ -105,7 +105,7 @@ class PressTabState implements AnnotationState {
         return new AutotriggerState().updateState(changeSource, force)
     }
 
-    isNextState(state: AnnotationState): boolean {
+    isNextState(state: AnnotationState | undefined): boolean {
         return state instanceof ManualtriggerState
     }
 }
@@ -147,7 +147,7 @@ class ManualtriggerState implements AnnotationState {
         }
     }
 
-    isNextState(state: AnnotationState): boolean {
+    isNextState(state: AnnotationState | undefined): boolean {
         return state instanceof TryMoreExState
     }
 }
@@ -175,7 +175,7 @@ class TryMoreExState implements AnnotationState {
         return this
     }
 
-    isNextState(state: AnnotationState): boolean {
+    isNextState(state: AnnotationState | undefined): boolean {
         return state instanceof EndState
     }
 
@@ -451,10 +451,8 @@ export class LineAnnotationController implements vscode.Disposable {
             return undefined
         }
 
-        if (!this._seen.has(updatedState.id)) {
-            try {
-                telemetry.ui_click.emit({ elementId: this._currentState.id, passive: true })
-            } catch (e) {}
+        if (this._currentState.isNextState(updatedState)) {
+            telemetry.ui_click.emit({ elementId: this._currentState.id, passive: true })
         }
 
         // update state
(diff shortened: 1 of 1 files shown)


packages/core/src/codewhisperer/views/lineAnnotationController.ts
```diff
@@ -44,8 +44,10 @@ function fromId(id: string | undefined): AnnotationState | undefined {
 interface AnnotationState {
     id: string
     suppressWhileRunning: boolean
+
     text: () => string
-    nextState(changeSource: AnnotationChangeSource, force: boolean): AnnotationState | undefined
```
Will-ShaoHua, line 48: rename to `updateState`

packages/core/src/codewhisperer/views/lineAnnotationController.ts
```diff
@@ -92,10 +98,15 @@ class PressTabState implements AnnotationState {
...
     text = () => 'CodeWhisperer Tip 1/3: Press [TAB] to accept ([ESC] to exit)'
 
-    nextState(changeSource: AnnotationChangeSource, force: boolean): AnnotationState | undefined {
-        return new AutotriggerState().nextState(changeSource, force)
+    updateState(changeSource: AnnotationChangeSource, force: boolean): AnnotationState | undefined {
+        return new AutotriggerState().updateState(changeSource, force)
+    }
+
+    isNextState(state: AnnotationState | undefined): boolean {
+        return state instanceof ManualtriggerState
```
Will-ShaoHua, line 109: Note that `case_1` and `case_1a` are both `case_1`, therefore their "nextState" is always `case_2` (ManualTriggerState)

packages/core/src/codewhisperer/views/lineAnnotationController.ts
```diff
@@ -284,7 +304,7 @@ export class LineAnnotationController implements vscode.Disposable {
         return this._currentState.id === new EndState().id
     }
 
-    async markTutorialDone() {
```
Will-ShaoHua, line 287: rename this method as it should only be called if it's "dismiss" use cases

packages/core/src/codewhisperer/views/lineAnnotationController.ts
```diff
@@ -384,14 +404,13 @@ export class LineAnnotationController implements vscode.Disposable {
...
             // Endstate is meaningless and doesnt need to be rendered
             this.clear()
-            await this.markTutorialDone()
+            await this.dismissTutorial()
             return
         } else if (decorationOptions.renderOptions?.after?.contentText === new TryMoreExState().text()) {
             // special case
             // case 3 exit criteria is to fade away in 30s
             setTimeout(async () => {
-                await this.markTutorialDone()
```
Will-ShaoHua, line 393: this function should only be called when user dismiss the tutorial because the entire `lineAnnotationController` will simply skip logic afterward

andrewyuq (approved):

PR #4633, opened 2024-04-04
Amazon Q Code Transform: Fix issues handling whitespace in user provided JAVA_HOME

arangatang:
## Problem
When customers provide a path to JDK installation that contain whitespaces we fail to use it when executing Maven to build the customers project. 
i.e. passing:

```
/Library/Java/JavaVirtualMachines/jdk 8 with spaces/amazon-corretto-8.jdk/Contents/Home
```

fails with

```
I'm sorry, I could not locate your Java installation. To troubleshoot, see the [Amazon Q documentation.](https://docs.aws.amazon.com/amazonq/latest/aws-builder-use-ug/troubleshooting-code-transformation.html#maven-commands-failing)
```

Screenshot 2024-04-04 at 16 03 42

## Solution
1. Use `path` to parse input instead of string manipulation (previously split on whitespace)
2. ~~Wrap `JAVA_HOME` variable value in quotes~~ 
3. Fixed some faulty messaging on windows of how to check for Java installation.

## Testing
Tested this by:
1. Ran it on mac using `/Library/Java/JavaVirtualMachines/jdk 8 with spaces/amazon-corretto-8.jdk/Contents/Home`
Screenshot 2024-04-04 at 16 10 21
2. Ran it on Windows

image

## License


justinmk3:
Looks like you committed all files in your workspace. Please don't do that. Please review your changes before taking a PR out of "draft". The package.json changes need to be reverted.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -336,10 +336,6 @@ export class GumbyController {
 }
 
 function extractPath(text: string): string | undefined {
```
justinmk3, line 338: Not obvious what this function does. The docstring should give an input/output example, that is a very concise way to make its behavior very clear:

```
"..." => "..."
```

justinmk3: This also apparently doesn't have test coverage, since I don't see any test changes in this PR?

packages/core/src/amazonqGumby/chat/controller/messenger/stringConstants.ts
```diff
@@ -7,7 +7,7 @@
 export const enterJavaHomeMessage = 'Enter the path to JDK '
 
 export const windowsJavaHomeHelpMessage =
-    'To find the JAVA_HOME path, run the following command in a new IDE terminal: `cd C:ProgramFilesJava` and then `dir`. If you see your JDK version, run `cd JDK<version>` and then `cd` to show the path.'
+    'To find the JAVA_HOME path, run the following command in a new IDE terminal: `cd "C:\\Program Files\\Java" && dir`. If you see your JDK version, run `cd <version>` and then `cd` to show the path.'
```
justinmk3, line 10: This is never a reason to use `\` path separators on Windows. Just use "/" everywhere.
```suggestion
    'To find the JAVA_HOME path, run the following command in a new IDE terminal: `cd "C:/Program Files/Java" && dir`. If you see your JDK version, run `cd <version>` and then `cd` to show the path.'
```

justinmk3, line 10: There is (almost) never a reason to use `\` path separators on Windows. Just use "/" everywhere.

`cd c:/foo/bar/` works just fine in both cmd.exe and powershell.

```suggestion
    'To find the JAVA_HOME path, run the following command in a new IDE terminal: `cd "C:/Program Files/Java" && dir`. If you see your JDK version, run `cd <version>` and then `cd` to show the path.'
```

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -336,10 +336,6 @@ export class GumbyController {
...
-    const words = text.split(/\s+/) // Split text into words by whitespace
```
justinmk3, line 339: Maybe this is the only fix that was needed? Are you sure that the [JAVA_HOME change](https://github.com/aws/aws-toolkit-vscode/pull/4633/files#r1553957216) is also needed?

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -538,7 +538,7 @@ function copyProjectDependencies(dependenciesFolder: FolderInfo) {
     let environment = process.env
     // if JAVA_HOME not found or not matching project JDK, get user input for it and set here
     if (transformByQState.getJavaHome() !== undefined) {
-        environment = { ...process.env, JAVA_HOME: transformByQState.getJavaHome() }
+        environment = { ...process.env, JAVA_HOME: `"${transformByQState.getJavaHome()}"` }
```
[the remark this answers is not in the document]
arangatang: Yep, you are right I just tested on my windows machine as well, this change seem to be not needed, I added this more as a precaution to be sure. Can remove it.

packages/core/src/amazonqGumby/chat/controller/messenger/stringConstants.ts
arangatang: While I agree its easier to type, the actual paths shown in the cmd are with the `\` and not `/` so to not add confusion for windows users that are used to the default `\` I prefer `\` in the text.

packages/core/src/amazonqGumby/chat/controller/controller.ts
arangatang: Yep, think you are right. Will add some.

packages/core/src/codewhisperer/service/transformByQHandler.ts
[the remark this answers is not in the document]
justinmk3: It should be removed/reverted.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -335,11 +335,15 @@ export class GumbyController {
     }
 }
 
+/**
+ * extractPath("./some/path/here") => "C:some/root/some/path/here"
+ * extractPath("C:some/nonexistent/path/here") => undefined
+ * extractPath("C:some/filepath/.txt") => undefined
```
justinmk3, line 341: is there a particular reason it uses windows paths? isn't a `/` required after `c:` ?
```suggestion
 * extractPath("./some/path/here") => "C:/some/root/some/path/here"
 * extractPath("C:/some/nonexistent/path/here") => undefined
 * extractPath("C:/some/filepath/.txt") => undefined
```

PR #4631, opened 2024-04-03
feat(codewhisperer): only register cw.connect command in Q

leigaol:
## Problem

Command `aws.codeWhisperer.connect` should only be declared in Amazon Q. 

## Solution

1. Move the declaration & registration of command `aws.codeWhisperer.connect` inside Amazon Q activation.
2. Output log information when the declaration of command `aws.codeWhisperer.connect` fails. 

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4637, opened 2024-04-04
feat(amazonq): Add extension id in toolkit APIs

leigaol:
## Problem

current credentials API doesn't know what called it.

## Solution

1. Add extension id in toolkit APIs
2. Log which extension called the APIs.

## License


packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
```diff
@@ -31,7 +31,7 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
         const importedApi = toolkitExt?.exports
```
justinmk3, line 31: I wonder if we can do something like this:

```suggestion
        const importedApi = toolkitExt?.exports.getApi(VSCODE_EXTENSION_ID.amazonq)
```

That would avoid the need for an `extensionId` parameter on every function.

packages/toolkit/src/api.ts
```diff
@@ -4,13 +4,15 @@
...
 
 export const awsToolkitApi = {
     /**
      * Exposing listConnections API for other extension to read or re-use
      * the available connections in aws toolkit.
      */
-    async listConnections(): Promise<AwsConnection[]> {
+    async listConnections(extensionId: string): Promise<AwsConnection[]> {
+        getLogger().debug(`listConnections: extension ${extensionId}`)
         const connections = await Auth.instance.listConnections()
```
justinmk3, line 16: For many/most of these, we also want to plumb through the extensionId and include it on related telemetry.

```suggestion
        const connections = await Auth.instance.listConnections(extensionId)
```

We could store it as a `source` field on related metrics. https://github.com/aws/aws-toolkit-common/blob/b7f05546af2d084ebf36b98c837ddf3641e3a63f/telemetry/definitions/commonDefinitions.json#L1226-L1230

Need to think about this. Doesn't need to be done in this PR.

leigaol: In that case, I would recommend separating the Auth.instance into 3 parts:

1. A AuthService that provides APIs which are intentionally to be invoked by both toolkit, amazonq and even other 3rd party plugins. As the Auth backend, it requires extension id as the source of all requests. It should be near stateless.

2. A stateful ToolkitAuth that implements auth features only for Toolkit or AmazonQ. It holds states. It do not provide any external API.

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -417,11 +417,11 @@ export const registerToolkitApiCallback = Commands.declare(
...
-                _toolkitApi = toolkitApi
+                _toolkitApi = toolkitApi.getApi(VSCODE_EXTENSION_ID.amazonq)
             } else {
                 // when this command is executed by Amazon Q activation
                 const toolkitExt = vscode.extensions.getExtension(VSCODE_EXTENSION_ID.awstoolkit)
-                _toolkitApi = toolkitExt?.exports
+                _toolkitApi = toolkitExt?.exports.getApi(VSCODE_EXTENSION_ID.amazonq)
             }
             if (_toolkitApi) {
                 await registerToolkitApiCallbackOnce()
```
justinmk3, line 427: (off topic): not sure Toolkit should be calling its own public API (that would also lead to some weird & unnecessary telemetry)

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -417,11 +417,11 @@ export const registerToolkitApiCallback = Commands.declare(
             getLogger().info(`Trying to register toolkit callback. Toolkit is installed.`)
             if (toolkitApi) {
                 // when this command is executed by AWS Toolkit activation
-                _toolkitApi = toolkitApi
+                _toolkitApi = toolkitApi.getApi(VSCODE_EXTENSION_ID.amazonq)
```
justinmk3, line 420: not sure Toolkit should be calling its own public API (that would also lead to some weird & unnecessary telemetry).

but for now, the extension id should be corrected at least.
```suggestion
                _toolkitApi = toolkitApi.getApi(VSCODE_EXTENSION_ID.toolkit)
```

packages/toolkit/src/api.ts
```diff
@@ -4,71 +4,83 @@
  */
 
 import { Auth, Connection, AwsConnection } from 'aws-core-vscode/auth'
+import { getLogger } from 'aws-core-vscode/shared'
 
 export const awsToolkitApi = {
     /**
-     * Exposing listConnections API for other extension to read or re-use
-     * the available connections in aws toolkit.
+     * Creating API object for external extensions.
```
justinmk3, line 11: * @param extensionId Extension id that identifies the caller.

packages/core/src/codewhisperer/commands/basicCommands.ts
leigaol: Justin, this command is actually invoked by Amazon Q. You are reviewing Amazon Q code. 

it would be great if we can move Amazon Q code from core/ to amazonq folder to make it clear.

justinmk3: but the comment on line 419 says:

> // when this command is executed by AWS Toolkit activation

justinmk3 (approved):

leigaol:
Found possible** issues after rebasing this PR. Marking it in draft until I fully tested it again.

packages/toolkit/src/api.ts
justinmk3: Why "stateful"? We have too much stateful things, and it's  a constant source of problems.

justinmk3:
> 2024-04-05 15:51:59.639 [error] Error: ToolkitGlobals.contextPrefix accessed before initialize()

https://github.com/aws/aws-toolkit-vscode/pull/4631#issuecomment-2041121015

leigaol:
> > 2024-04-05 15:51:59.639 [error] Error: ToolkitGlobals.contextPrefix accessed before initialize()
> 
> [#4631 (comment)](https://github.com/aws/aws-toolkit-vscode/pull/4631#issuecomment-2041121015)

https://github.com/aws/aws-toolkit-vscode/pull/4653

packages/toolkit/src/api.ts
leigaol: >A stateful ToolkitAuth that implements auth features only for Toolkit or AmazonQ. It holds states. It do not provide any external API.

I mean: This ToolkitAuth manages state of toolkit connections. It does not provide APIs for external extensions.

packages/core/src/codewhisperer/commands/basicCommands.ts
leigaol: Here this piece of code is invoked by Amazon Q. Toolkit won't be making API calls against itself.

leigaol:
> > > 2024-04-05 15:51:59.639 [error] Error: ToolkitGlobals.contextPrefix accessed before initialize()
> > 
> > 
> > [#4631 (comment)](https://github.com/aws/aws-toolkit-vscode/pull/4631#issuecomment-2041121015)
> 
> #4653

Rebased and tested.

justinmk3 (approved):

PR #4642, opened 2024-04-05
feat(amazonq): Auto install AmazonQ when logging into aws toolkit

leigaol:
## Problem
Toolkit Q users should be migrated to Q standalone.

Screenshot 2024-04-04 at 5 28 50 PM

Screenshot 2024-04-04 at 5 29 31 PM

## Solution
Auto install AmazonQ when logging into aws toolkit for existing Q users. Prompt for Q install for new users. 

## License

### DIFF
--- a/packages/core/src/amazonq/explorer/amazonQChildrenNodes.ts
+++ b/packages/core/src/amazonq/explorer/amazonQChildrenNodes.ts
@@ -24,7 +24,7 @@ export const learnMoreAmazonQCommand = Commands.declare('aws.toolkit.amazonq.lea
 })
 
 export const qExtensionPageCommand = Commands.declare('aws.toolkit.amazonq.extensionpage', () => () => {
-    void vscode.env.openExternal(vscode.Uri.parse(`vscode:extension/${VSCODE_EXTENSION_ID.awstoolkit}`))
+    void vscode.env.openExternal(vscode.Uri.parse(`vscode:extension/${VSCODE_EXTENSION_ID.amazonq}`))
 })
 
 export const dismissQTree = Commands.declare('aws.toolkit.amazonq.dismiss', () => async () => {
--- a/packages/core/src/amazonq/explorer/amazonQChildrenNodes.ts
+++ b/packages/core/src/amazonq/explorer/amazonQChildrenNodes.ts
@@ -7,7 +7,7 @@ import * as vscode from 'vscode'
 import * as nls from 'vscode-nls'
 import { Command, Commands, placeholder } from '../../shared/vscode/commands2'
 import { codicon, getIcon } from '../../shared/icons'
-import { reconnect } from '../../codewhisperer/commands/basicCommands'
+import { installAmazonQExtension, reconnect } from '../../codewhisperer/commands/basicCommands'
 import { amazonQHelpUrl } from '../../shared/constants'
 import { cwTreeNodeSource } from '../../codewhisperer/commands/types'
 import { telemetry } from '../../shared/telemetry/telemetry'
@@ -31,9 +31,9 @@ export const dismissQTree = Commands.declare('aws.toolkit.amazonq.dismiss', () =
     await globals.context.globalState.update(amazonQDismissedKey, true)
     await vscode.commands.executeCommand('setContext', amazonQDismissedKey, true)
 })
-
+// Learn more button of Amazon Q now opens the Amazon Q marketplace page.
 export const createLearnMoreNode = () =>
-    learnMoreAmazonQCommand.build().asTreeNode({
+    qExtensionPageCommand.build().asTreeNode({
         label: localize('AWS.amazonq.learnMore', 'Learn More About Amazon Q (Preview)'),
         iconPath: getIcon('vscode-question'),
         contextValue: 'awsAmazonQLearnMoreNode',
@@ -103,7 +103,7 @@ export function createSignIn(type: 'item' | 'tree'): any {
 }
 
 export function createInstallQNode() {
-    return qExtensionPageCommand.build().asTreeNode({
+    return installAmazonQExtension.build().asTreeNode({
         label: 'Install the Amazon Q Extension', // TODO: localize
         iconPath: getIcon('vscode-extensions'),
     })
--- a/packages/core/src/extension.ts
+++ b/packages/core/src/extension.ts
@@ -175,7 +175,7 @@ export async function activate(context: vscode.ExtensionContext) {
                     if (isPreviousQUser()) {
                         await installAmazonQExtension.execute()
                         await vscode.window.showInformationMessage(
-                            'Amazon Q has moved to its own VSCode extension, which has been automatically installed',
+                            'Amazon Q has moved to its own VSCode extension, which has been automatically installed.',
                             'OK'
                         )
                     } else {
--- a/packages/core/src/extension.ts
+++ b/packages/core/src/extension.ts
@@ -175,9 +175,12 @@ export async function activate(context: vscode.ExtensionContext) {
                     if (isPreviousQUser()) {
                         await installAmazonQExtension.execute()
                         void vscode.window
-                            .showInformationMessage('We have installed Amazon Q extension for you', 'Restart')
+                            .showInformationMessage(
+                                'Amazon Q has moved to its own VSCode extension, which has been automatically installed',
+                                'Reload Now'
+                            )
                             .then(async resp => {
-                                if (resp === 'Restart') {
+                                if (resp === 'Reload Now') {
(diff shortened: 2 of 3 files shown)


packages/core/src/extension.ts
```diff
@@ -171,22 +171,32 @@ export async function activate(context: vscode.ExtensionContext) {
...
-                                void learnMoreAmazonQCommand.execute()
-                            } else if (resp === 'Install') {
-                                void installAmazonQExtension.execute()
-                            }
-                        })
+                if (!isExtensionInstalled(VSCODE_EXTENSION_ID.amazonq)) {
+                    if (isPreviousQUser()) {
+                        await installAmazonQExtension.execute()
+                        void vscode.window
+                            .showInformationMessage('We have installed Amazon Q extension for you', 'Restart')
```
justinmk3, line 178:
```suggestion
                            .showInformationMessage('Amazon Q has moved to its own VSCode extension, which has been automatically installed.', 'Restart')
```

packages/core/src/extension.ts
```diff
@@ -171,22 +171,32 @@ export async function activate(context: vscode.ExtensionContext) {
...
+                            .then(async resp => {
+                                if (resp === 'Restart') {
```
justinmk3, line 180: If we are only reloading the window, then let's use the same language that vscode does.

image

```suggestion
                                if (resp === 'Reload Now') {
```

packages/core/src/extension.ts
```diff
@@ -171,22 +171,32 @@ export async function activate(context: vscode.ExtensionContext) {
...
+                            .showInformationMessage(
+                                'Amazon Q has moved to its own VSCode extension.' +
+                                    '\nInstall to work with Amazon Q, a generative AI assistant, with chat and code suggestions.',
+                                'Install',
+                                'Learn More'
+                            )
+                            .then(async resp => {
+                                if (resp === 'Learn More') {
+                                    void learnMoreAmazonQCommand.execute()
+                                } else if (resp === 'Install') {
```
justinmk3, line 195: does vscode show a "reload" prompt after this?

packages/core/src/extension.ts
```diff
@@ -171,22 +171,32 @@ export async function activate(context: vscode.ExtensionContext) {
...
-                        .then(async resp => {
-                            if (resp === 'Learn More') {
-                                void learnMoreAmazonQCommand.execute()
-                            } else if (resp === 'Install') {
-                                void installAmazonQExtension.execute()
-                            }
-                        })
+                if (!isExtensionInstalled(VSCODE_EXTENSION_ID.amazonq)) {
+                    if (isPreviousQUser()) {
+                        await installAmazonQExtension.execute()
```
hayemaxi, line 176: Can this fail for any reason? If so, we should fall back to the dialogue box for manual install.

packages/core/src/extension.ts
```diff
@@ -171,22 +171,32 @@ export async function activate(context: vscode.ExtensionContext) {
...
+                                'Amazon Q has moved to its own VSCode extension.' +
+                                    '\nInstall to work with Amazon Q, a generative AI assistant, with chat and code suggestions.',
+                                'Install',
+                                'Learn More'
+                            )
+                            .then(async resp => {
+                                if (resp === 'Learn More') {
+                                    void learnMoreAmazonQCommand.execute()
+                                } else if (resp === 'Install') {
+                                    void installAmazonQExtension.execute()
```
hayemaxi, line 196: for non-previous-users, we should confirm if we should install the extension or take users to marketplace page.

packages/core/src/extension.ts
hayemaxi, line 180: I don't think we need to actually reload the window. The extension works fine when activated mid-session.

packages/core/src/extension.ts
leigaol: No. I think Max is right, we don't need to reload.

packages/core/src/extension.ts
leigaol: Agree. I can replace the button to OK.

packages/core/src/extension.ts
leigaol: The design says provide a Install button for user to click and then install.

packages/core/src/extension.ts
leigaol: >Can this fail for any reason?

I tried to turn-off wifi but the installation is still successful. Cannot find any case when this fails.

packages/core/src/extension.ts
leigaol: Only uninstall requires a reload to remove the registered commands(etc) from that extension

packages/core/src/extension.ts
```diff
@@ -171,22 +171,30 @@ export async function activate(context: vscode.ExtensionContext) {
...
-                            }
-                        })
+                    } else {
+                        void vscode.window
+                            .showInformationMessage(
+                                'Amazon Q has moved to its own VSCode extension.' +
+                                    '\nInstall to work with Amazon Q, a generative AI assistant, with chat and code suggestions.',
+                                'Install',
+                                'Learn More'
+                            )
```
hayemaxi, line 188: Don't we want a dismiss button as well?

packages/core/src/extension.ts
```diff
@@ -171,22 +171,30 @@ export async function activate(context: vscode.ExtensionContext) {
...
+                            .then(async resp => {
+                                if (resp === 'Learn More') {
+                                    // Clicking learn more will open the q extension page
+                                    void qExtensionPageCommand.execute()
+                                } else if (resp === 'Install') {
+                                    void installAmazonQExtension.execute()
```
hayemaxi, line 194: let's await to catch errors (unless this keeps the dialogue box open for some reason)

packages/core/src/extension.ts
```diff
@@ -171,22 +171,30 @@ export async function activate(context: vscode.ExtensionContext) {
...
-                        .showInformationMessage(
-                            'Amazon Q has moved to its own VSCode extension.' +
-                                '\nInstall to work with Amazon Q, a generative AI assistant, with chat and code suggestions.',
-                            'Install',
-                            'Learn More',
-                            'Dismiss'
+                if (!isExtensionInstalled(VSCODE_EXTENSION_ID.amazonq)) {
+                    if (isPreviousQUser()) {
+                        await installAmazonQExtension.execute()
+                        void vscode.window.showInformationMessage(
```
hayemaxi, line 177: also await

packages/core/src/amazonq/explorer/amazonQChildrenNodes.ts
hayemaxi, line 1: awesome!

packages/core/src/extension.ts
hayemaxi, line 177:
```suggestion
                        await vscode.window.showInformationMessage(
```

packages/core/src/extension.ts
```diff
@@ -171,22 +171,30 @@ export async function activate(context: vscode.ExtensionContext) {
...
                         )
-                        .then(async resp => {
-                            if (resp === 'Learn More') {
-                                void learnMoreAmazonQCommand.execute()
-                            } else if (resp === 'Install') {
-                                void installAmazonQExtension.execute()
-                            }
-                        })
+                    } else {
+                        void vscode.window
```
hayemaxi, line 182:
```suggestion
                        await vscode.window
```

packages/core/src/extension.ts
justinmk3: All messages have an "X", so "Dismiss" doesn't seem necessary. And it's unusual.

hayemaxi: makes sense. let's make sure we record that the user closed this notification, otherwise it will display every time they open vsc

packages/core/src/extension.ts
```diff
@@ -171,22 +171,30 @@ export async function activate(context: vscode.ExtensionContext) {
...
-                            'Amazon Q has moved to its own VSCode extension.' +
-                                '\nInstall to work with Amazon Q, a generative AI assistant, with chat and code suggestions.',
-                            'Install',
-                            'Learn More',
-                            'Dismiss'
+                if (!isExtensionInstalled(VSCODE_EXTENSION_ID.amazonq)) {
+                    if (isPreviousQUser()) {
+                        await installAmazonQExtension.execute()
+                        await vscode.window.showInformationMessage(
+                            'Amazon Q has moved to its own VSCode extension, which has been automatically installed',
```
hayemaxi, line 178:
```suggestion
                            'Amazon Q has moved to its own VSCode extension, which has been automatically installed.',
```

hayemaxi (approved):

Issue #4643, opened 2024-04-05
S3 Tree View Not Displaying Bucket Contents with Specific IAM Role Policies


avsthiago:
## Problem

The S3 tree view does not show the contents of the bucket when the IAM role being used has a policy that restricts the list_objects_v2 action to specific prefixes (including the root prefix). 

## Steps to reproduce the issue

1. Create two s3 buckets e.g. `aws-toolkit-list-experiment-01` and `aws-toolkit-list-experiment-02` 

2. Apply the following policy to the IAM role being used by the AWS Toolkit.

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "AllowListAllBuckets",
            "Effect": "Allow",
            "Action": [
                "s3:ListAllMyBuckets",
                "s3:GetBucketLocation"
            ],
            "Resource": "*"
        },
        {
            "Sid": "AllowListAccessEx1",
            "Effect": "Allow",
            "Action": "s3:ListBucket*",
            "Resource": "arn:aws:s3:::aws-toolkit-list-experiment-01"
        },
        {
            "Sid": "AllowListAccessEx2",
            "Effect": "Allow",
            "Action": "s3:ListBucket*",
            "Resource": "arn:aws:s3:::aws-toolkit-list-experiment-02",
            "Condition": {
                "StringLike": {
                    "s3:prefix": [
                        ""
                    ]
                }
            }
        }
    ]
}
```

3. Open the extension's Explorer view and expand the S3 bucket tree view.

You will notice that the contents of `aws-toolkit-list-experiment-01` are displayed but the contents of `aws-toolkit-list-experiment-02` are not displayed.

![issue-aws-toolkit](https://github.com/aws/aws-toolkit-vscode/assets/13870565/9ad60149-f3d1-411c-a260-37a4310e0846)

## Expected behavior

I would expect the contents of the bucket to be displayed in the tree view if at least the root prefix is allowed by the policy.

## Additional context

<details>
<summary>Extension's error log</summary>

2024-04-05 07:44:25 [ERROR]: [AccessDenied: Access Denied
	at constructor.extractError (/home/ec2-user/SageMaker/.xdg/data/code-server/extensions/amazonwebservices.aws-toolkit-vscode-2.18.0-448bb26/dist/src/main.js:62:10373)
[...]

PR #4644, opened 2024-04-05
fix(S3 View): Display bucket contents with specific IAM role policies

avsthiago:
## Problem: 
S3 tree view doesn't show contents with specific IAM role policies. Prefix parameter was null.

## Solution: 
Set Prefix parameter to an empty string as default. Allows bucket contents display if root prefix is allowed.

## Related issue:
#4643 S3 Tree View Not Displaying Bucket Contents with Specific IAM Role Policies

## License

### DIFF
--- a/packages/core/src/shared/clients/s3Client.ts
+++ b/packages/core/src/shared/clients/s3Client.ts
@@ -18,7 +18,7 @@ import { toStream } from '../utilities/collectionUtils'
 
 export const DEFAULT_MAX_KEYS = 300 // eslint-disable-line @typescript-eslint/naming-convention
 export const DEFAULT_DELIMITER = '/' // eslint-disable-line @typescript-eslint/naming-convention
-export const DEFAULT_PREFIX = '' // eslint-disable-line @typescript-eslint/naming-convention
+export const defaultPrefix = ''
 
 export type Bucket = InterfaceNoSymbol<DefaultBucket>
 export type Folder = InterfaceNoSymbol<DefaultFolder>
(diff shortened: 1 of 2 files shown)


packages/core/src/shared/clients/s3Client.ts
```diff
@@ -450,7 +451,7 @@ export class DefaultS3Client {
                 Bucket: bucket.name,
                 Delimiter: DEFAULT_DELIMITER,
                 MaxKeys: request.maxResults ?? DEFAULT_MAX_KEYS,
-                Prefix: request.folderPath,
+                Prefix: request.folderPath ?? DEFAULT_PREFIX,
```
justinmk3, line 454: can you add a brief comment that explains what's going on here, and link to https://github.com/aws/aws-toolkit-vscode/issues/4643

justinmk3 (approved):
Thanks for the clear root-cause analysis in https://github.com/aws/aws-toolkit-vscode/issues/4643 , that is very helpful.

packages/core/src/shared/clients/s3Client.ts
avsthiago: Addressed on 63b7bed.

packages/core/src/shared/clients/s3Client.ts
```diff
@@ -18,6 +18,7 @@ import { toStream } from '../utilities/collectionUtils'
 
 export const DEFAULT_MAX_KEYS = 300 // eslint-disable-line @typescript-eslint/naming-convention
 export const DEFAULT_DELIMITER = '/' // eslint-disable-line @typescript-eslint/naming-convention
+export const DEFAULT_PREFIX = '' // eslint-disable-line @typescript-eslint/naming-convention
```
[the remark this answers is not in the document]
avsthiago: Addressed on 63b7bed.

packages/core/src/shared/clients/s3Client.ts
```diff
@@ -450,7 +451,12 @@ export class DefaultS3Client {
                 Bucket: bucket.name,
                 Delimiter: DEFAULT_DELIMITER,
                 MaxKeys: request.maxResults ?? DEFAULT_MAX_KEYS,
-                Prefix: request.folderPath,
+                /**
+                 * Set '' as the default prefix to ensure that the bucket's content will be displayed
```
ivikash, line 455: JSDoc has a property `@default`. We can probably use that here. This way, IDEs can directly display the default values. 

```
/**
* Set '' as the default prefix to ensure that the bucket's content will be displayed
* when the user has at least list access to the root of the bucket.
* https://github.com/aws/aws-toolkit-vscode/issues/4643
* @default '' 
```

avsthiago: Adjusted on 468270c.

ivikash (approved):

kozer9:
I am also having trouble with the same bug. Can anyone please let me know when this PR will be implemented?

PR #4646, opened 2024-04-05
fix(tests): amazon q featureDev e2e test prompts

aemada-aws:
## Problem
 The current e2e prompts cause false positive rejections and other internal quality issues which we are working on resolving.
## Solution
The new prompts are more stable. It used to always fail now it passed twice in a row.

## License

### DIFF
--- a/packages/core/src/testE2E/amazonq/featureDev.test.ts
+++ b/packages/core/src/testE2E/amazonq/featureDev.test.ts
@@ -20,9 +20,9 @@ describe('Amazon Q Feature Dev', function () {
     let tab: Messenger
 
     const maxTestDuration = 600000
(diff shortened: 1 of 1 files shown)


.changes/next-release/Test-148a692f-ec5a-4d58-b451-f04334e0779b.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Test",
+	"description": "change prompts for q feature dev e2e tests"
```
jpinkney-aws, line 3: Since this change isn't customer impacting I don't think we need a changelog item

jpinkney-aws (approved):

PR #4650, opened 2024-04-05
fix(release): add branch env var, protect against release loops

hayemaxi:
## License

### DIFF
--- a/buildspec/release/00clonerepo.yml
+++ b/buildspec/release/00clonerepo.yml
@@ -13,12 +13,23 @@ phases:
         commands:
             # Check for implicit env vars passed from the release pipeline.
             - test -n "${TOOLKITS_GITHUB_REPO_OWNER}"
+            - test -n "${TARGET_BRANCH}"
 
     build:
         commands:
             - git clone https://github.com/${TOOLKITS_GITHUB_REPO_OWNER}/aws-toolkit-vscode.git aws-toolkit-vscode
-            # checkout master as we want to commit to it later
-            - cd aws-toolkit-vscode && git checkout master
+            # checkout the target branch as we want to commit to it later to update versions
+            - cd aws-toolkit-vscode && git checkout ${TARGET_BRANCH}
+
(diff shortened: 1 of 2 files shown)


buildspec/release/00clonerepo.yml
```diff
@@ -13,12 +13,23 @@ phases:
...
-            - cd aws-toolkit-vscode && git checkout master
+            # checkout the target branch as we want to commit to it later to update versions
+            - cd aws-toolkit-vscode && git checkout ${TARGET_BRANCH}
+
+            # Since we don't have a manual gate on the amazon q release pipeline, we must make sure we aren't
+            # releasing off of a commit from another release, i.e. infinite loop.
+            # This will be removed prior to standalone extension release.
+            - |
+                AUTHOR_AUTOMATION="aws-toolkit-automation"
+                if [[ $(git log -1 --pretty=%an) = $AUTHOR_AUTOMATION && $(git log -1 --pretty=%B) != "Merge master into feature/standalone" ]]; then
```
leigaol, line 29: "Merge master into feature/standalone" 

How about commit in the format of "Release 2.18.0"

leigaol, line 29: When there is a merge conflict, the merge message may not be "Merge master into feature/standalone". 

Is there a more robust way to detect if the last commit is from pipeline?

justinmk3: could use `grep` to check for `[mM]erge`. Following is untested:
```suggestion
                if [[ $(git log -1 --pretty=%an) = $AUTHOR_AUTOMATION && git log -1 --pretty=%B | >/dev/null grep '[mM]erge' ]]; then
```

hayemaxi: I just removed this block. Releasing like this is just too risky, will have to deal with race conflicts. Will update the release pipeline to have an alternative method.

justinmk3 (approved):

PR #4661, opened 2024-04-08
deps: require node 16+

justinmk3:
Problem:
cloud9 now has minimum of node 16, so we can require node 16.

Solution:
- bump minimum to node 16
- remove "performance", "uuid" workarounds because the equivalent features are now available in node 16 + web browser.

## License


packages/core/src/shared/telemetry/spans.ts
```diff
@@ -252,7 +252,7 @@ export class TelemetryTracer extends TelemetryBase {
         const frame = this.switchContext(span)
 
         try {
-            const result = this.#context.run(frame, fn, span)
+            const result = this.#context.run(frame, fn, span as any)
```
justinmk3, line 255: 😬

jpinkney-aws: Out of curiosity which change made casting to any required?

nkomonen-amazon (approved):

packages/core/src/shared/telemetry/spans.ts
justinmk3: After bumping `@types/node` tsc gave this error... I don't get it.

```
npm ERR! src/shared/telemetry/spans.ts(255,57): error TS2345: Argument of type 'TelemetrySpan<MetricBase>' is not assignable to parameter of type 'Metric<MetricShapes[U]>'.
npm ERR!   Property 'run' is missing in type 'TelemetrySpan<MetricBase>' but required in type 'Metric<MetricShapes[U]>'.
```

PR #4663, opened 2024-04-08
fix(amazonq): projectDetails not always emitted, add field to hold if user on Builder Id or IdC

dhasani23:
## Problem

Wanted to add telemetry to capture if user is on Builder ID or IdC. Also, had an issue where projectDetails metric was not being emitted for non-Java 8/11 projects.

## Solution

Add field to existing metric to capture Builder ID vs IdC. Fix projectDetails issue.

## License


packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -191,12 +191,6 @@ export async function preTransformationUploadCode() {
     } catch (err) {
         const errorMessage = `Failed to upload code due to ${(err as Error).message}`
         getLogger().error(errorMessage)
-        telemetry.codeTransform_logGeneralError.emit({
```
dhasani23, line 194: We were already emitted a metric for this, these were duplicate and unnecessary.

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -65,34 +65,26 @@ export interface FolderInfo {
 
 // log project details silently
 export async function validateAndLogProjectDetails() {
```
dhasani23, line 67: Simplified this. There was also an issue where metric not being emitted for non-Java 8/11 projects due to the conditional in the finally block.

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -322,7 +314,7 @@ export async function uploadArtifactToS3(
         })
         getLogger().info(`CodeTransformation: Status from S3 Upload = ${response.status}`)
     } catch (e: any) {
-        const errorMessage = (e as Error).message ?? 'Error in S3 UploadZip API call'
```
dhasani23, line 325: `message` field is required so `??` wasn't doing anything

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -629,57 +621,69 @@ export async function prepareProjectDependencies(dependenciesFolder: FolderInfo)
...
-    if (fs.existsSync(dependenciesFolder.path)) {
-        fs.rmSync(dependenciesFolder.path, { recursive: true, force: true })
+        tempFilePath = path.join(os.tmpdir(), 'zipped-code.zip')
+        fs.writeFileSync(tempFilePath, zip.toBuffer())
+        if (fs.existsSync(dependenciesFolder.path)) {
+            fs.rmSync(dependenciesFolder.path, { recursive: true, force: true })
+        }
+        fs.rmSync(logFilePath) // will always exist here
+    } catch (e: any) {
+        telemetry.codeTransform_logGeneralError.emit({
```
dhasani23, line 679: Wrapped this in a try-catch and emitted a metric to match IntelliJ behavior

packages/core/src/amazonqGumby/telemetry/codeTransformTelemetry.ts
```diff
@@ -26,37 +22,6 @@ export enum CancelActionPositions {
     Chat = 'qChatPanel',
 }
 
-export enum StartActionPositions {
-    DevToolsSidePanel = 'devToolsStartButton',
-    BottomHubPanel = 'bottomPanelSideNavButton',
-    ChatPrompt = 'chatPrompt',
-}
-
-export const logCodeTransformInitiatedMetric = (source: string): void => {
```
dhasani23, line 35: Unused

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -227,9 +231,19 @@ export class GumbyController {
         }
     }
 
-    // Any given project could have multiple candidate projects to transform --
-    // The user gets prompted to pick a specific one
+    // prompt user to pick project and specify source JDK version
     private async initiateTransformationOnProject(message: any) {
+        let authType: CredentialSourceId | undefined = undefined
```
nkomonen-amazon, line 236: If you think this `authType` code may be used again in the future, it might be good to separate in to its own function for now

packages/core/src/codewhisperer/service/transformationResultsViewProvider.ts
```diff
@@ -330,18 +326,26 @@ export class ProposedTransformationExplorer {
...
-                    codeTransformApiErrorMessage: (e as Error).message ?? errorMessage,
+                    codeTransformApiErrorMessage: downloadErrorMessage,
                     codeTransformRequestId: e.requestId ?? '',
                     result: MetadataResult.Fail,
                     reason: 'ExportResultArchiveFailed',
                 })
                 throw new Error('Error downloading diff')
+            } finally {
+                // This metric is emitted when user clicks Download Proposed Changes button
+                telemetry.codeTransform_vcsViewerClicked.emit({
```
nkomonen-amazon, line 342: Is there a reason for not using `telemetry.codeTransform_vscViewerClicked.run()`?

dhasani23: We have work in progress that intends to consolidate all of the custom metrics we've added into one large metric to hold anything/everything we need. I'd say as a part of that work we can transition from the `emit` calls we have everywhere to `run`.

Right above here in the `catch` we have a different metric being emitted, so when we complete that ^ work that metric's info could be captured via `telemetry.consolidated_Metric.record(...)`

packages/core/src/amazonqGumby/chat/controller/controller.ts
dhasani23: Done

nkomonen-amazon (approved):

packages/core/src/codewhisperer/service/transformByQHandler.ts
```diff
@@ -63,36 +65,38 @@ export interface FolderInfo {
     name: string
 }
 
+export async function getAuthType() {
+    let authType: CredentialSourceId | undefined = undefined
+    if (AuthUtil.instance.isEnterpriseSsoInUse() && AuthUtil.instance.isConnectionValid()) {
+        authType = 'iamIdentityCenter'
```
mk-fan, line 71: nit: convert to enum

mk-fan (approved):

PR #4664, opened 2024-04-08
auth: Implement authorization grant with pkce

jpinkney-aws:
## Problem
- The current authorization flow is moving towards authorization grant with pkce instead of device code

## Solution
- Move to pkce authorization which decreases friction for authorization
- Current device code rfc: https://datatracker.ietf.org/doc/html/rfc8628
- New rfc for authorization grant: https://datatracker.ietf.org/doc/html/rfc6749 and pkce: https://datatracker.ietf.org/doc/html/rfc7636

## Testing
If you are interested in testing this PR you need to set `"aws.dev.pkceAuth": true` in the vscode settings

## TODOs:
- emit telemetry when a user gets redirected with /sso/authenticated
- browser based authentication
- CORS
- missing test for cors
- missing test for new auth flow
- missing support for web
     

## License


packages/core/src/auth/sso/clients.ts
```diff
@@ -33,6 +33,12 @@ import { DevSettings } from '../../shared/settings'
 import { SdkError } from '@aws-sdk/types'
 import { HttpRequest, HttpResponse } from '@aws-sdk/protocol-http'
 import { StandardRetryStrategy, defaultRetryDecider } from '@aws-sdk/middleware-retry'
+import OidcClientPKCE from './oidcclientpkce'
+import { toSnakeCase } from '../../shared/utilities/textUtilities'
+import { Credentials, Service } from 'aws-sdk'
+import apiConfig = require('./service-2.json')
```
nkomonen-amazon, line 39: I see we import package.json like the following, maybe this works:
`import packageJson from '../../package.json'`

packages/core/src/test/credentials/sso/server.test.ts
```diff
@@ -0,0 +1,108 @@
...
+        await server.close()
+    })
+
+    function createURL(baseUrl: string, params: Record<string, string>) {
+        const url = new URL(baseUrl)
+        url.search = new URLSearchParams(params).toString()
+        return url.toString()
+    }
+
+    async function createRequest(params: Record<string, string>, expectedErrorMsg?: string) {
```
nkomonen-amazon, line 40: Should we rename this to something like `assertRequestError()`. Also does it make sense for `expectedErrorMsg` to be required?

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -164,90 +179,42 @@ export class SsoAccessTokenProvider {
...
-        const cacheKey = this.registrationCacheKey
-        const cachedRegistration = await this.cache.registration.load(cacheKey)
-
-        // Clear cached if registration is expired
-        if (cachedRegistration && isExpired(cachedRegistration)) {
-            await this.invalidate()
+    protected abstract getValidatedClientRegistration(): Promise<ClientRegistration>
+    protected abstract registerClient(): Promise<ClientRegistration>
+
+    public static create(
```
nkomonen-amazon, line 225: Nice separation of implementations

packages/core/scripts/build/generateServiceClient.ts
```diff
@@ -245,6 +245,10 @@ void (async () => {
             serviceJsonPath: 'src/amazonqFeatureDev/client/codewhispererruntime-2022-11-11.json',
             serviceName: 'FeatureDevProxyClient',
         },
+        {
```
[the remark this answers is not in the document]
nkomonen-amazon: Nvm I see now that a different client is used

packages/core/src/auth/sso/clients.ts
jpinkney-aws: Yeah that should work as well, I was mostly just copying how other teams imported their json files for apiConfig

justinmk3:
CI is timing out because this is based on an old commit. Rebasing will fix the CI timeouts.

packages/core/src/test/shared/utilities/textUtilities.test.ts
```diff
@@ -107,6 +108,44 @@ describe('getStringHash', async function () {
     })
 })
 
+describe('toSnakeCase', function () {
```
justinmk3, line 111: Related: https://github.com/aws/aws-toolkit-vscode/blob/1bc938bbd9396a665f43b9358d531b70c9741405/packages/core/src/shared/vscode/env.ts#L100-L110

does it make sense to use `toSnakeCase` in  `getEnvVars` ?

packages/core/src/auth/sso/server.ts
```diff
@@ -0,0 +1,209 @@
...
+        return Promise.race([
+            this.authenticationPromise,
+            new Promise<string>((_, reject) => {
+                globals.clock.setTimeout(() => {
+                    reject(
+                        new ToolkitError('Timed-out waiting for browser login flow to complete', {
+                            code: 'TimedOut',
+                        })
+                    )
+                }, this.authenticationTimeoutInMs)
```
justinmk3, line 205: After ~1 minute we should log a WARN-level message.

packages/core/src/auth/sso/server.ts
```diff
@@ -0,0 +1,209 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import http from 'http'
+import { getLogger } from '../../shared/logger'
+import { ToolkitError } from '../../shared/errors'
+import { Socket } from 'net'
```
justinmk3, line 9: nit: for cases where we know a module is definitely node-only, we should start using the `node:` prefix on imports. 
```suggestion
import { Socket } from 'node:net'
```

justinmk3, line 9: nit: for cases where we know a module is definitely node-only, we should start using the `node:` prefix on imports.


```suggestion
import http from 'node:http'
import { getLogger } from '../../shared/logger'
import { ToolkitError } from '../../shared/errors'
import { Socket } from 'node:net'
```

justinmk3 (approved):
Beautiful

packages/core/src/test/shared/utilities/textUtilities.test.ts
jpinkney-aws: Yeah, I don't see any reason not to add it there

justinmk3:
Build is failing, I guess we can't use `node:` prefix for now. 😅

jpinkney-aws:
Just removed the node commit, I'll wait for the ci to pass and then i'll merge

justinmk3:
CI will pass after rebasing since https://github.com/aws/aws-toolkit-vscode/pull/4669 was merged. But I think it's fine to just merge, because the current failing tests are the same that were failing before that PR.

justinmk3:
> But I think it's fine to just merge, because the current failing tests are the same that were failing before that PR.

Oops, I didn't notice that there were "rejected promises" issues, besides the failing tests:

```
rejected promise not handled within 1 second: Error: AuthSSOServer: foo: foo
rejected promise not handled within 1 second: Error: AuthSSOServer: invalid state
rejected promise not handled within 1 second: Error: AuthSSOServer: missing code
rejected promise not handled within 1 second: Error: AuthSSOServer: missing state

ERROR: Test output matched this pattern 5 times:
```

jsamuel1:
Question - is it worth collecting metrics on failed vs succeeded login flows? (with a baseline equivalent in the original flow?)

It would be great to prove the decrease in user drop-offs due to auth flow.  (And potentially find/fix any future friction in this space)

PR #4665, opened 2024-04-08
feat(amazonq): Skip login screen for migration use case

leigaol:
## Problem

For users who migrate from toolkit to Q, we need to let them use Q without doing any login. The migration should be seamless.

## Solution

For users who installed Q, if this user

1. Has aws toolkit installed
2. has a Amazon Q connection in toolkit 
3. Do not have any connection in Q.

Then we know this user is doing the migration from toolkit to Q, in this case, we skip the login screen for the user and immediately use the toolkit connection in Q. 

## License

### DIFF
--- a/packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
+++ b/packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
@@ -35,15 +35,22 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
         }
         return connections
     }
-
-    // Prioritize reusing a valid connection that has Amazon Q scopes
+    /**
+     * Select a connection for Amazon Q to use
+     * @param connections List of AWS Toolkit Connections
+     * @returns A connection that has Amazon Q scopes. It will return a valid connection if exists.
+     * Returns undefined if no connection has Q scopes.
+     */
     selectConnection(connections: AwsConnection[]): AwsConnection | undefined {
-        const score = (c: AwsConnection) =>
-            Number(c.scopes?.includes(scopesCodeWhispererChat[0])) * 10 + Number(c.state === 'valid')
+        const hasQScopes = (c: AwsConnection) => amazonQScopes.every(s => c.scopes?.includes(s))
+        const score = (c: AwsConnection) => Number(hasQScopes(c)) * 10 + Number(c.state === 'valid')
         connections.sort(function (a, b) {
             return score(b) - score(a)
         })
-        return connections[0]
+        if (hasQScopes(connections[0])) {
+            return connections[0]
+        }
+        return undefined
     }
 
     async useConnection(connectionId: string): Promise<AuthError | undefined> {
--- a/packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
+++ b/packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import * as vscode from 'vscode'
-import { SsoConnection, scopesCodeWhispererChat, AwsConnection } from '../../../../auth/connection'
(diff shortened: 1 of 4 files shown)


packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
```diff
@@ -22,19 +22,36 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
...
         }
         return connections
     }
+    /**
+     * Select a connection for Amazon Q to use
+     * @param connections List of AWS Toolkit Connections
+     * @returns A connection that has Amazon Q scopes. It will return a valid connection if exists.
+     * Returns undefined if no connection has Q scopes.
+     */
+    selectConnection(connections: AwsConnection[]): AwsConnection | undefined {
```
justinmk3, line 44: "select" sounds like it's choosing a menu item. But the logic seems to be choosing the most-appropriate connection from multiple connections.  The current docstring also doesn't make this clear.
```suggestion
    /**
     * Gets a connection that is usable by Amazon Q.
     *
     * @param connections List of AWS Toolkit Connections
     * @returns Amazon Q connection, or undefined if none of the given connections have scopes required for Amazon Q.
     */
    getConnection(connections: AwsConnection[]): AwsConnection | undefined {
```

leigaol: >Why does this function live in backend_amazonq.ts instead of one of the "auth" util modules?

Because the functions of re-use the connections are defined in the backend. If I move it to somewhere else, I still need to call `useConnection()`

justinmk3: `findConnection` is a pure function. although the "client" might need to define a pass-through so that it can be called by the vue frontend, its impl could still live in a shared module.

leigaol: This findConnection function takes a AwsConnection class, which is not defined within the scope of the Vue app. If I make it a util function in Vue app it will fail.

leigaol: `Reusing the connection`  involves UI updates, which should be handled in the Vue app. If reusing one of the existing expired connection failed, the UI should update.

packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
```diff
@@ -22,19 +22,36 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
...
+        const hasQScopes = (c: AwsConnection) => amazonQScopes.every(s => c.scopes?.includes(s))
+        const score = (c: AwsConnection) => Number(hasQScopes(c)) * 10 + Number(c.state === 'valid')
+        connections.sort(function (a, b) {
+            return score(b) - score(a)
+        })
+        if (hasQScopes(connections[0])) {
+            return connections[0]
+        }
+        return undefined
+    }
```
justinmk3, line 54: Here is what I am suggesting:

```suggestion
    findConnection(connections: AwsConnection[]): AwsConnection | undefined {
        return someSharedModule.findConnection(connections)
    }
```

packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
justinmk3: > `Reusing the connection` involves UI updates

findConnection doesn't do that. It's just a filter function. See also https://github.com/aws/aws-toolkit-vscode/pull/4665#discussion_r1558480853

packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
justinmk3: `backend_amazonq.ts` currently looks pretty small so this isn't much of a problem right now.

But something to consider for the future.

PR #4669, opened 2024-04-09
fix(ci): restore settings type checking, fix unit tests

hayemaxi:
Squash and merge with commit message:

- Fix all unit tests.
- Enable the amazon q extension artifact for packaging tests
- Update tech debt test to only fail for master base branch
- Fix launch config for debugging so that it always executes the proper default task
- Restore settings type checking at compile time
- Sync settings to subprojects

## License

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -156,7 +156,7 @@
                     "description": "%AWS.configuration.enableCodeLenses%",
                     "default": false
                 },
-                "aws.toolkit.suppressPrompts": {
+                "aws.suppressPrompts": {
                     "type": "object",
                     "description": "%AWS.configuration.description.suppressPrompts%",
                     "default": {},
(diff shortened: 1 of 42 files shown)


packages/core/src/test/testUtil.ts
```diff
@@ -473,3 +474,22 @@ export function shuffleList<T>(list: T[]): T[] {
 
     return shuffledList
 }
+
+/**
+ * Try to register a command for tests since some commands are only registered during
+ * extension activation. These commands will need to be manually activated here.
```
justinmk3, line 480: The tests globalSetup already should (does, on master branch) activate the extension. Lots of code depends on that (we previously tried to avoid it but ended up giving up). Is that not happening on feature/standalone? Or I'm misunderstanding?

hayemaxi: CodeWhisperer commands are only activated in the Amazon Q extension, which is a separate entry point than the toolkit activation.
Let me see if both can be activated at once. Otherwise, I will update the docstring to mention this.

justinmk3: > Otherwise, we will need to drop tests down into their respective subprojects later (but I will update the docstring to mention this now).

We definitely want to do that eventually. If we can't do it now, let's add `tryRegister` to our followup checklist.

packages/core/package.json
```diff
@@ -233,6 +233,85 @@
                     "type": "object",
                     "description": "%AWS.configuration.description.lambda.recentlyUploaded%",
                     "default": []
+                },
+                "aws.amazonq.logLevel": {
```
leigaol, line 237: If amazonq commands are declared here in core package.json, is it OK to remove the amazonq command in /amazonq/package.json ?

leigaol: If they are duplicated in both q and core, when I change Q settings, should I update both duplicate or ?

packages/core/package.json
```diff
@@ -233,6 +233,85 @@
...
+                    "type": "string",
+                    "default": "debug",
+                    "enum": [
+                        "error",
+                        "warn",
+                        "info",
+                        "verbose",
+                        "debug"
```
justinmk3, line 245: side-note: we won't need a "loglevel" setting in the future, we can use the OutputChannel to control that instead.

image

packages/core/package.json
justinmk3: @leigaol see `docs/arch_develop.md`: https://github.com/aws/aws-toolkit-vscode/pull/4669/commits/964cc471f2a91e20ce94052c9638f8cd5b6495bb

PR #4670, opened 2024-04-09
ignore me

nkomonen-amazon:
## License


packages/core/package.json
```diff
@@ -4218,23 +4218,31 @@
         "clean": "ts-node ../../scripts/clean.ts dist/",
         "copyFiles": "ts-node ./scripts/build/copyFiles.ts",
         "buildScripts": "npm run generateClients && npm run generatePackage && npm run copyFiles",
-        "compile": "npm run clean && npm run buildScripts && tsc -p ./ && webpack --mode development",
-        "compileLite": "tsc -p ./",
-        "watch": "npm run clean && npm run buildScripts && tsc -watch -p ./",
-        "testCompile": "npm run clean && npm run buildScripts && tsc -p ./",
+        "compile": "npm run testCompile && webpack",
```
justinmk3, line 4221: Not a blocker for this PR,  but it's weird that `compile` depends on `testCompile`. And previously `compile` was doing a development build. All of this is confusing. But it's tech debt we've had for a long time.

packages/core/package.json
```diff
@@ -4218,23 +4218,31 @@
...
+        "webpackDevelopment": "webpack --mode development",
+        "serveVue": "webpack serve --config-name vue --mode development",
+        "watch": "npm run testCompile -- -- --watch",
+        "testsWebWatch": "npm run clean && npm run buildScripts && npm run webpackDevelopment -- --watch",
+        "testCompile": "npm run clean && npm run buildScripts && npm run compileOnly",
         "test": "npm run testCompile && c8 ts-node ./scripts/test/test.ts",
-        "webCompile": "npm run clean && npm run buildScripts && webpack --config-name web",
-        "webWatch": "npm run webCompile -- --watch",
-        "testWeb": "npm run webCompile && c8 ts-node ./scripts/test/testWeb.ts",
+        "webDevelopmentCompile": "npm run compile -- --mode development",
```
justinmk3, line 4229: Current impl doesn't look related to "web". It's just doing a regular `compile`.

nit: could be named `webDevCompile`, that matches the NPM convention (`devDependencies)`.

packages/core/package.json
```diff
@@ -4218,23 +4218,31 @@
...
         "format": "prettier --ignore-path ../../.prettierignore --check src scripts",
         "formatfix": "prettier --ignore-path ../../.prettierignore --write src scripts",
         "lint": "ts-node ./scripts/lint/testLint.ts && npm run format",
         "generateClients": "ts-node ./scripts/build/generateServiceClient.ts ",
         "generatePackage": "ts-node ./scripts/build/generateIcons.ts",
-        "generateTelemetry": "node ../../node_modules/@aws-toolkits/telemetry/lib/generateTelemetry.js --extraInput=src/shared/telemetry/vscodeTelemetry.json --output=src/shared/telemetry/telemetry.gen.ts",
-        "serve": "webpack serve --config-name vue-hmr --mode development"
+        "generateTelemetry": "node ../../node_modules/@aws-toolkits/telemetry/lib/generateTelemetry.js --extraInput=src/shared/telemetry/vscodeTelemetry.json --output=src/shared/telemetry/telemetry.gen.ts"
+    },
+    "scriptsComments": {
```
justinmk3, line 4240: is this an NPM convention (i.e. does it feed into tools like vscode) or just something we are doing?

justinmk3 (approved):
Excellent

justinmk3:
> Now, if you edit a Vue just file do `Command Palette` -> `Reload Webviews` and the webview will update with the new changes.

Is that mentioned in CONTRIBUTING.md?

packages/core/package.json
nkomonen-amazon: My rationale is that a `testCompile` is enough to get testing working, but a full `compile` is `testCompile` plus anything extra needed for a production compile. I still feel we could simplify things for sure so open to any suggestions.

packages/core/package.json
nkomonen-amazon: When I made this I wasn't sure if Web mode would need different configs. But looks like YAGNI. Switching it to `compileDev`

packages/core/package.json
nkomonen-amazon: I saw this convention on stackoverflow. My understanding is that VSCode knows which package.json keys it cares about and ignores the rest. So adding `*Comments` shouldn't match any existing keys and wont be an issue

justinmk3: > VSCode knows which package.json keys it cares about and ignores the rest.

That's actually part of the package.json spec. I was just curious if this particular key was actually used by any tooling.

justinmk3 (approved):

PR #4673, opened 2024-04-09
fix(amazonq): Fix various auto-installation issue

leigaol:
## Problem

1. [NEW FEATURE] Installing Amazon Q from toolkit should immediately open Q extension view.
2. [BUG FIX] aws.codewhisperer.connect command conflict when user first install toolkit then install Q.
3. [BUG FIX] 2 installation notification displayed at the same time
4. [NEW FEATURE] User can click dismiss (X) button when seeing the ask user to install Q notification.   If user does not click that button, there will be no resp. It is not considered as dismiss.

## License

### DIFF
--- a/packages/core/src/extension.ts
+++ b/packages/core/src/extension.ts
@@ -178,7 +178,6 @@ export async function activate(context: vscode.ExtensionContext) {
                             'Amazon Q has moved to its own VSCode extension, which has been automatically installed.',
                             'OK'
                         )
-                        void vscode.commands.executeCommand(`workbench.view.extension.${VSCODE_EXTENSION_ID.amazonq}`)
                     } else {
                         const dismissedInstall = globals.context.globalState.get<boolean>(amazonQInstallDismissedKey)
                         if (!dismissedInstall) {
@@ -195,10 +194,7 @@ export async function activate(context: vscode.ExtensionContext) {
                                         await qExtensionPageCommand.execute()
                                     } else if (resp === 'Install') {
                                         await installAmazonQExtension.execute()
-                                        void vscode.commands.executeCommand(
-                                            `workbench.view.extension.${VSCODE_EXTENSION_ID.amazonq}`
-                                        )
-                                    } else {
+                                    } else if (resp === undefined) {
                                         // when user click x button to dismiss it
                                         await globals.context.globalState.update(amazonQInstallDismissedKey, true)
                                     }
--- a/packages/core/src/extension.ts
+++ b/packages/core/src/extension.ts
@@ -59,6 +59,7 @@ import {
 import { AuthUtil, isPreviousQUser } from './codewhisperer/util/authUtil'
 import { installAmazonQExtension } from './codewhisperer/commands/basicCommands'
 import { isExtensionInstalled, VSCODE_EXTENSION_ID } from './shared/utilities'
+import { amazonQInstallDismissedKey } from './codewhisperer/models/constants'
 
 let localize: nls.LocalizeFunc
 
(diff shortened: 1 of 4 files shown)


hayemaxi:
Seems like this command no longer needs a wrapper and try block on it:
https://github.com/leigaol/aws-toolkit-vscode/blob/897ef7d03adf117ea5873d3cbc396e394b25c800/packages/core/src/codewhisperer/commands/basicCommands.ts#L142

Can we make this in scope to remove it and update the docstrings? Declaring a command twice does not cause an error anyway, so I'm not sure what this was trying to achieve.

packages/core/src/extension.ts
```diff
@@ -178,21 +179,27 @@ export async function activate(context: vscode.ExtensionContext) {
...
-                            )
-                            .then(async resp => {
-                                if (resp === 'Learn More') {
-                                    // Clicking learn more will open the q extension page
-                                    await qExtensionPageCommand.execute()
-                                } else if (resp === 'Install') {
-                                    await installAmazonQExtension.execute()
-                                }
-                            })
+                        const dismissedInstall = globals.context.globalState.get<boolean>(amazonQInstallDismissedKey)
```
hayemaxi, line 182: nice!

leigaol:
> With this PR it seems like this command no longer needs a wrapper and try block on it: https://github.com/leigaol/aws-toolkit-vscode/blob/897ef7d03adf117ea5873d3cbc396e394b25c800/packages/core/src/codewhisperer/commands/basicCommands.ts#L142
> 
> Can we make this in scope to remove it and update the docstrings? Declaring a command twice does not cause an error anyway, so I'm not sure what this was trying to achieve.

That change was to make Viceroy compatible.

justinmk3 (approved):

packages/core/src/extension.ts
```diff
@@ -178,21 +179,27 @@ export async function activate(context: vscode.ExtensionContext) {
...
+                                .then(async resp => {
+                                    if (resp === 'Learn More') {
+                                        // Clicking learn more will open the q extension page
+                                        await qExtensionPageCommand.execute()
+                                    } else if (resp === 'Install') {
+                                        await installAmazonQExtension.execute()
+                                    } else if (resp === undefined) {
+                                        // when user click x button to dismiss it
+                                        await globals.context.globalState.update(amazonQInstallDismissedKey, true)
+                                    }
```
hayemaxi, line 200: maybe we shouldn't display the message again if users uninstall Q
```suggestion
                                    } else {
                                        await installAmazonQExtension.execute()
                                        await globals.context.globalState.update(amazonQInstallDismissedKey, true)
                                    }
```

hayemaxi, line 200: maybe we shouldn't display it again if the user uninstalls Q later.

```suggestion
                                    if (resp === 'Learn More') {
                                        // Clicking learn more will open the q extension page
                                        await qExtensionPageCommand.execute()
                                        return
                                    } else if (resp === 'Install') {
                                        await installAmazonQExtension.execute()
                                    }
                                    // If user dismisses, then we jump to here.
                                    await globals.context.globalState.update(amazonQInstallDismissedKey, true)
```

leigaol:
> With this PR it seems like this command no longer needs a wrapper and try block on it: https://github.com/leigaol/aws-toolkit-vscode/blob/897ef7d03adf117ea5873d3cbc396e394b25c800/packages/core/src/codewhisperer/commands/basicCommands.ts#L142
> 
> Can we make this in scope to remove it and update the docstrings? Declaring a command twice does not cause an error anyway (only registering does), so I'm not sure what this was trying to achieve.

Yes. 

Viceroy should work fine when invoking the aws.CodeWhisperer.connect command because we have the auto install amazon q implemented.

hayemaxi (approved):

PR #4683, opened 2024-04-10
feat(amazonq): use same telemetry client id between Q and toolkit

leigaol:
## Problem
use same telemetry client id between Amazon Q and aws toolkit

## Solution

1. Use hash of machine id + user name (OS) to unique identify the user. In this case, all new IDs from Toolkit and Amazon Q will always be the same.

2. For toolkit with old random UUID, let Amazon Q adopt its UUID

The `env.machineId` is used in VS Code telemetry client id https://github.com/microsoft/vscode/blob/6a06de0033d052521f03079aabb356ca7916145c/src/vs/workbench/api/common/extHost.api.impl.ts#L357.

It comes from hardware Mac addr https://github.com/microsoft/vscode/blob/6a06de0033d052521f03079aabb356ca7916145c/src/vs/base/node/id.ts#L93

The Mac address of a hardware is unique. 

## License

### DIFF
--- a/packages/core/src/codewhisperer/commands/basicCommands.ts
+++ b/packages/core/src/codewhisperer/commands/basicCommands.ts
@@ -410,6 +410,7 @@ const registerToolkitApiCallbackOnce = once(async () => {
     if (_toolkitApi && 'getTelemetryClientId' in _toolkitApi) {
         const clientId = await _toolkitApi.getTelemetryClientId()
         if (clientId) {
+            getLogger().debug(`Adopting telemetry client id ${clientId}`)
             await globals.context.globalState.update('telemetryClientId', clientId)
         }
     }
--- a/packages/core/src/codewhisperer/commands/basicCommands.ts
+++ b/packages/core/src/codewhisperer/commands/basicCommands.ts
@@ -35,6 +35,7 @@ import { TelemetryHelper } from '../util/telemetryHelper'
 import { Auth, AwsConnection } from '../../auth'
 import { once } from '../../shared/utilities/functionUtils'
 import { isTextEditor } from '../../shared/utilities/editorUtilities'
+import { globals } from '../../shared'
 
 export const toggleCodeSuggestions = Commands.declare(
     { id: 'aws.codeWhisperer.toggleCodeSuggestion', compositeKey: { 1: 'source' } },
@@ -404,6 +405,13 @@ const registerToolkitApiCallbackOnce = once(async () => {
                 await refreshToolkitQState.execute(false)
             }
         )
+    }
+    // use the same client id as aws toolkit if it is available
+    if (_toolkitApi && 'getTelemetryClientId' in _toolkitApi) {
+        const clientId = await _toolkitApi.getTelemetryClientId()
+        if (clientId) {
+            await globals.context.globalState.update('telemetryClientId', clientId)
+        }
     }
 })
 export const registerToolkitApiCallback = Commands.declare(
--- a/packages/core/src/shared/telemetry/util.ts
+++ b/packages/core/src/shared/telemetry/util.ts
@@ -3,12 +3,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import * as os from 'os'
 import { env, Memento, version } from 'vscode'
 import { getLogger } from '../logger'
 import { fromExtensionManifest } from '../settings'
 import { shared } from '../utilities/functionUtils'
 import { extensionVersion, isAutomation } from '../vscode/env'
-import { randomUUID } from 'crypto'
+import * as crypto from 'crypto'
 import { addTypeName } from '../utilities/typeConstructors'
 import globals from '../extensionGlobals'
 import { mapMetadata } from './telemetryLogger'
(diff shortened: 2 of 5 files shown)


packages/core/src/shared/telemetry/util.ts
```diff
@@ -43,6 +44,20 @@ export function convertLegacy(value: unknown): boolean {
...
+ */
+export function generateTelemetryClientId(): string {
+    const userId = `${env.machineId}-${os.userInfo({ encoding: 'utf-8' }).username}`
+    const shasum = crypto.createHash('sha256')
+    const mid = shasum.update(userId).digest('hex')
+    const uid = `${mid.substring(0, 8)}-${mid.substring(8, 12)}-${mid.substring(12, 16)}-${mid.substring(
+        16,
+        20
+    )}-${mid.substring(20, 32)}`
+    return uid
```
justinmk3, line 59: where did this logic come from? why are we changing it?

`globals.machineId` should be used, assuming we actually need this logic. 
https://github.com/aws/aws-toolkit-vscode/blob/3d4c65e158ff2a5a92586bad96e5229df90191ce/packages/core/src/extensionShared.ts#L98

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -405,6 +406,14 @@ const registerToolkitApiCallbackOnce = once(async () => {
             }
         )
     }
+    // use the same client id as aws toolkit if it is available
+    if (_toolkitApi && 'getTelemetryClientId' in _toolkitApi) {
+        const clientId = await _toolkitApi.getTelemetryClientId()
+        if (clientId) {
+            getLogger().debug(`Adopting telemetry client id ${clientId}`)
+            await globals.context.globalState.update('telemetryClientId', clientId)
```
justinmk3, line 414: This will change all of client-ids on launch day. Did we think about the implications of this?

justinmk3:
> Use hash of machine id + user name (OS) ... all new IDs from Toolkit and Amazon Q will always be the same.

What does `env.machineId` return?  Does it return something different in a codecatalyst dev env (or other remote env)?

Our `globals.machineId` is based on `hostname`, which may be something generic like "mymachine". Username also may be generic ("user1", "justin", etc.). So these aren't enough to generate uniqueness.

packages/core/src/shared/telemetry/util.ts
leigaol: >where did this logic come from?

This is a unique `machineId` used by VS Code telemetry. This is not ` globals.machineId`.  These are completely different thing.

leigaol: By VS Code API definition: 
		```
/**
		 * A unique identifier for the computer.
		 */
		export const machineId: string;
```

packages/core/src/codewhisperer/commands/basicCommands.ts
leigaol: >This will change all of client-ids on launch day. 

Not true. It will not change all client ids.

leigaol:
> > Use hash of machine id + user name (OS) ... all new IDs from Toolkit and Amazon Q will always be the same.
> 
> What does `env.machineId` return? Does it return something different in a codecatalyst dev env (or other remote env)?
> 
> Our `globals.machineId` is based on `hostname`, which may be something generic like "mymachine". Username also may be generic ("user1", "justin", etc.). So these aren't enough to generate uniqueness.



>What does `env.machineId` return? 

It is used in VS Code telemetry client id https://github.com/microsoft/vscode/blob/6a06de0033d052521f03079aabb356ca7916145c/src/vs/workbench/api/common/extHost.api.impl.ts#L357.

It comes from hardware Mac addr.

packages/core/src/codewhisperer/commands/basicCommands.ts
justinmk3: > This API is to solve one edge case: When toolkit use the old random UUID, Q will adopt the random old UUID of toolkit.

Why is that a problem?



> It will not change all client ids. Both toolkit and Q will adopt the client id from a hash of (hardware macAddr, OS user name).

I don't understand how that confirms that the client id won't change on launch day (compared to the current, existing client id).

leigaol: > > This API is to solve one edge case: When toolkit use the old random UUID, Q will adopt the random old UUID of toolkit.
> 
> Why is that a problem?
> 
> > It will not change all client ids. Both toolkit and Q will adopt the client id from a hash of (hardware macAddr, OS user name).
> 
> I don't understand how that confirms that the client id won't change on launch day (compared to the current, existing client id).


The old toolkit, once upgraded to new toolkit, still has its random UUID as the client id. It won't upgrade to this new MacAddr based Id. 

The new Amazon Q, once installed, will attempt to query toolkit to use its client id, if not available, will fallback to the new MacAddr based Id. 

If toolkit becomes available after Q initialize, then Q will adopt the client id from toolkit. Regardless of whether the toolkit client id is old random UUID or MacAddr based id.

leigaol:
> > This is not ` globals.machineId`. These are completely different thing.
> 
> I know. But in the past we decided not to use `env.machineId` . I don't remember why (possibly: (1) Cloud9, (2) PII?).
> 
> I know the docstring says "unique identifer", but we need to check what the actual value is on:
> 
>     1. desktop
> 
>     2. Windows
> 
>     3. codecatalyst dev env
> 
>     4. does the desktop value differ from the remote env value
>        
>        * when we share globalState, the remote shares the local value. so currently, telemetry client-id is (I believe) the same on cc dev env and local desktop.
> 
> 
> > It comes from hardware Mac addr https://github.com/microsoft/vscode/blob/6a06de0033d052521f03079aabb356ca7916145c/src/vs/base/node/id.ts#L93
> 
> Ok, that seems good. Just confirm how it works on cc dev env (and Windows).


If `env.machineId` was ruled out in a discussion years ago, then there might be a reason for that.  Verifying this id across multiple use cases and environment can be a lot of work. 

I tried to adopt the Settings based approach but encountered some issues, now since Max has pushed changes to settings, I might be able to try the shared Settings based approach again.

packages/core/src/codewhisperer/commands/basicCommands.ts
justinmk3: > The new Amazon Q, once installed, will attempt to query toolkit to use its client id, if not available, will fallback to the new MacAddr based Id.

👍

PR #4689, opened 2024-04-11
fix(amazonFeatureDev): accepted files telemetry

sannicm:
## Problem

File acceptance telemetry was not counting deleted files, i.e. the files that code generation marked to be deleted, even if they were rejected or not.

## Solution

Count deleted files for the metric and add tests to assure this.

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
+++ b/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
@@ -449,9 +449,15 @@ export class FeatureDevController {
         let session
         try {
             session = await this.sessionStorage.getSession(message.tabID)
+
+            const acceptedFiles = (paths?: { rejected: boolean }[]) => (paths || []).filter(i => !i.rejected).length
+
+            const amazonqNumberOfFilesAccepted =
+                acceptedFiles(session.state.filePaths) + acceptedFiles(session.state.deletedFiles)
+
             telemetry.amazonq_isAcceptedCodeChanges.emit({
                 amazonqConversationId: session.conversationId,
-                amazonqNumberOfFilesAccepted: session.state.filePaths?.filter(i => !i.rejected).length ?? -1,
+                amazonqNumberOfFilesAccepted,
                 enabled: true,
                 result: 'Succeeded',
             })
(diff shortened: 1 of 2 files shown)


packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -449,9 +449,16 @@ export class FeatureDevController {
         let session
         try {
             session = await this.sessionStorage.getSession(message.tabID)
+
+            const acceptedFiles = (paths?: { rejected: boolean }[]) =>
+                !paths ? -1 : paths.filter(i => !i.rejected).length
```
dianaabv16, line 454: Why -1?
if file paths are not defined for some reason let's make it 0

packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
```diff
@@ -323,4 +329,87 @@ describe('Controller', () => {
...
+                            rejected: false,
+                            virtualMemoryUri: '' as unknown as vscode.Uri,
+                            workspaceFolder: controllerSetup.workspaceFolder,
+                        },
+                        {
+                            zipFilePath: 'myfile2.js',
+                            relativePath: 'myfile2.js',
+                            fileContent: '',
+                            rejected: true,
+                            virtualMemoryUri: '' as unknown as vscode.Uri,
```
dianaabv16, line 359: I don't recommend setting vritualMemoryUri like this.

you can use   generateVirtualMemoryUri() function like in this test [link](https://github.com/aws/aws-toolkit-vscode/pull/4594/files)

Note: I doesn't effect the test functionality, but still better to keep code as close to "actual" implementation as possible, but cool that you added a test!

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
ivikash: Looking at the code in here, I was curious, can we don something like 

```typescript
const acceptedFilesLength = (paths?: { rejected: boolean }[]) => (paths || []).filter(i => !i.rejected).length
```

dianaabv16 (approved):

ivikash (approved):

PR #4693, opened 2024-04-11
web: make shared crypto.randomUUID() function

nkomonen-amazon:
## Problem
`crypto.randomUUID()` is not provided by `crypto-browserify` so it fails when in web mode even though we polyfilled the `crypto` module.

## Solution
Create a `common` implementation of randomUUID() that works in both node and web.

Node.js provides `crypto.randomUUID()` and web provides `globalThis.crypto.randomUUID()`. There is not an out-of-the-box way to get a single implementation that works for both environments so this commit creates it.

We will want to look to setup an eslint rule which points developers to use this implementation once we get that set up.

## License


packages/core/src/common/crypto.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { isWeb } from './webUtils'
+
+export function randomUUID(): `${string}-${string}-${string}-${string}-${string}` {
+    if (isWeb()) {
+        return globalThis.crypto.randomUUID()
+    }
```
justinmk3, line 11: Can we just use `globalThis.crypto` everywhere, and only "polyfill" the nodejs case. That avoids the need for this `isWeb()` check. We only need something like:

```
if (!isWeb()) {
    globalThis.crypto = ...
}

File: packages/core/src/common/crypto.ts [hunk 2]
```diff
@@ -0,0 +1,14 @@
...
+
+    return require('crypto').randomUUID()
+}
```
justinmk3, line 14:
```suggestion
if (!isWeb()) {
    import * as crypto from 'crypto'
    globalThis.crypto = ...
}
```

justinmk3: 👍  That info would be worth mentioning in the module docstring, perhaps.

justinmk3 (approved):

packages/core/src/testWeb/common/crypto.test.ts
justinmk3, line 1: Are these tests identical to the non-web tests? We should have a test tree that contains tests which run in both web and non-web. Then we don't need to duplicate in the web-only and non-web-only trees.

PR #4697, opened 2024-04-12
deps: Update Mynah UI to 4.5.5 

jpinkney-aws:
## Problem
- we are seeing multiple "Generating your answer..." cards in featureDev

## Solution
- update mynah ui to 4.5.5

The bug hasn't been released to customers yet, that why theres no changelog entry

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3607,9 +3607,9 @@
             }
         },
         "node_modules/@aws/mynah-ui": {
-            "version": "4.5.4",
-            "resolved": "https://registry.npmjs.org/@aws/mynah-ui/-/mynah-ui-4.5.4.tgz",
-            "integrity": "sha512-ZFJLLxybwzfBKhO5NKDzt++ntHbtf7bUqyAX//9pcvYVLHBHegGgqJcDECA1inYMUb2rybeVRIv7jpmjX7ks2g==",
+            "version": "4.5.5",
+            "resolved": "https://registry.npmjs.org/@aws/mynah-ui/-/mynah-ui-4.5.5.tgz",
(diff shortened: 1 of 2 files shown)


ivikash (approved):

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -333,11 +333,7 @@ export const createMynahUI = (ideApi: any, amazonQEnabled: boolean) => {
                 return
             }
 
-            if (tabsStorage.getTab(tabID)?.type === 'featuredev') {
-                mynahUI.addChatItem(tabID, {
-                    type: ChatItemType.ANSWER_STREAM,
-                })
```
vishaalmehrishi, line 339: I'm curious about how removing this fixes the issue. We still need one "Generating your answer" message - does that come from somewhere else?

justinmk3 (approved):

packages/core/src/amazonq/webview/ui/main.ts
dogusata: I suspect this can be the issue, since adding the empty streaming card (generating your answer preloader) usually gets done on the initial request, but as i see also on async process we're adding it. And in the general block not /transform or /dev or cwc specific. @JuliaABurch Can you please check?

dogusata (approved):

packages/core/src/amazonq/webview/ui/main.ts
JuliaABurch: Following up - I reverted the section of code @dogusata pointed to and tested, and while it fixed the multiple `Generating your answer` messages, other issues arose in both `featuredev` and `gumby`. We have varying ways of creating blank cards, cards in progress, etc, that isn't consistent across controllers. This change instead bumps the version of `mynah-ui` to one that preventatively checks to make sure multiple blank cards aren't being pushed to chat messages.

JuliaABurch (approved):

PR #4698, opened 2024-04-12
feat(amazonq): Force telemetry client id to be the same between Amazon Q and AWS Toolkit

leigaol:
## Problem

The elemetry client id has to to be the same between Amazon Q and AWS Toolkit

## Solution

Use `process.env` to pass the telemetry client id. Each extension attempt to set its telemetry client id at time of activation. 

There can be edge cases when Q activates but toolkit is uninstalled (still has old client id), in this case, the 2nd time vscode window reloads, the amazon Q will sync its client id with AWS toolkit. 

## License


leigaol:
Package test failed due to 
>error: changelog data file already exists: /codebuild/output/src225367328/src/github.com/aws/aws-toolkit-vscode/.changes/2.19.0-SNAPSHOT.json

Which is not relevant tot his commit.

packages/core/src/shared/telemetry/activation.ts
```diff
@@ -120,3 +122,42 @@ export async function handleTelemetryNoticeResponse(
         getLogger().error('Error while handling response from telemetry notice: %O', err as Error)
     }
 }
+
+/**
+ * Setup the telemetry client id at extension activation.
+ * This function is designed to let AWS Toolkit and Amazon Q share
+ * the same telemetry client id.
+ */
+
```
justinmk3, line 131:
```suggestion
```

packages/core/src/shared/telemetry/activation.ts
```diff
@@ -120,3 +122,41 @@ export async function handleTelemetryNoticeResponse(
...
+ * Setup the telemetry client id at extension activation.
+ * This function is designed to let AWS Toolkit and Amazon Q share
+ * the same telemetry client id.
+ */
+export async function setupTelemetryId(extensionContext: vscode.ExtensionContext) {
+    const globalStatekey = 'telemetryClientId'
+    const envKey = '__TELEMETRY_CLIENT_ID'
+    try {
+        const currentClientId = globals.context.globalState.get<string>(globalStatekey)
+        const storedClientId = process.env[envKey]
```
justinmk3, line 136:
```suggestion
        // Use env var because vscode doesn't have "getContext": https://github.com/microsoft/vscode/issues/10471
        const storedClientId = process.env[envKey]
```

justinmk3:
I assume `process.env` won't work in a web browser (not nodejs) context. Will we need to use a setting there?

packages/core/src/shared/telemetry/activation.ts
```diff
@@ -120,3 +122,42 @@ export async function handleTelemetryNoticeResponse(
         getLogger().error('Error while handling response from telemetry notice: %O', err as Error)
     }
 }
+
+/**
+ * Setup the telemetry client id at extension activation.
+ * This function is designed to let AWS Toolkit and Amazon Q share
+ * the same telemetry client id.
+ */
+export async function setupTelemetryId(extensionContext: vscode.ExtensionContext) {
```
justinmk3, line 131: can this live in `packages/core/src/shared/telemetry/util.ts` ? that is where `getClientId` lives. seems like a natural place for this to live. they probably have overlapping logic...

leigaol:
> I assume `process.env` won't work in a web browser (not nodejs) context. Will we need to use a setting there?

I cannot get the settings approach to work. How can I test this in web browser?

justinmk3:
> How can I test this in web browser?

Let's use `env.machineId` in web browser, for now. You can check browser mode via `isWeb()`: https://github.com/aws/aws-toolkit-vscode/blob/26af5a6ff465957e4060d3d09fc51a8b88aedd09/packages/core/src/auth/activation.ts#L53

packages/core/src/shared/telemetry/util.ts
```diff
@@ -154,3 +157,46 @@ export function validateMetricEvent(event: MetricDatum, fatal: boolean) {
         getLogger().warn(msg)
     }
 }
+
+/**
+ * Setup the telemetry client id at extension activation.
+ * This function is designed to let AWS Toolkit and Amazon Q share
+ * the same telemetry client id.
+ */
+
```
justinmk3, line 166:
```suggestion
```

packages/core/src/shared/telemetry/util.ts
```diff
@@ -154,3 +157,45 @@ export function validateMetricEvent(event: MetricDatum, fatal: boolean) {
...
     }
 }
+
+/**
+ * Setup the telemetry client id at extension activation.
+ * This function is designed to let AWS Toolkit and Amazon Q share
+ * the same telemetry client id.
+ */
+export async function setupTelemetryId(extensionContext: vscode.ExtensionContext) {
+    const globalStatekey = 'telemetryClientId'
```
justinmk3, line 167: now that the code is in the same module let's move this to a module-scoped variable and re-use it from `getClientId`

justinmk3 (approved):

hayemaxi (approved):

PR #4701, opened 2024-04-12
feat(ci): support newChange, createRelease/changelogs, pre-release changelogs for multiple ext

hayemaxi:
- Pre-releases on GitHub with both extension artifacts will now have both recent changelogs in the description.
- newChange and createRelease work on the individual extensions now, and must be accessed with e.g. -w packages/toolkit instead of running at root.
- Release pipeline will call the target extension's createRelease script.

TODO:
- When we release, the "latest" will just refer the last extension that was released. Maybe we should have custom tags if possible.

## License


.github/workflows/release.yml
```diff
@@ -62,11 +62,18 @@ jobs:
             - name: Export outputs
               id: build
               run: |
+                  write_package_info() {
+                      PKG_NAME=$1
+                      PKG_DISPLAY_NAME=$(grep -m 1 displayName packages/${PKG_NAME}/package.json | grep -o '[a-zA-z][^\"]\+' | tail -n1)
+                      echo "version=$(grep -m 1 version packages/${PKG_NAME}/package.json | grep -o '[0-9][^\"]\+' | sed 's/-SNAPSHOT//')" >> $GITHUB_OUTPUT
+                      echo "${PKG_NAME}_changes<<EOF" >> $GITHUB_OUTPUT
+                      # The following line adds extension display name to the topmost changelog section.
```
justinmk3, line 70:
```suggestion
                      # Add extension display name to the topmost changelog section.
```

buildspec/release/10changeversion.yml
```diff
@@ -26,9 +26,8 @@ phases:
                 npm version --no-git-tag-version "$VERSION" -w packages/${TARGET_EXTENSION}
                 # Call npm ci because 'createRelease' uses ts-node
                 npm ci
-            # TODO: fix createRelease (generates CHANGELOG) for "amazonq".
             - |
-                echo "$CODEBUILD_WEBHOOK_HEAD_REF" | >/dev/null grep master && npm run createRelease || true
+                npm run createRelease -w packages/${TARGET_EXTENSION}
```
justinmk3, line 30: if we are doing this on feature/standalone, won't we run into https://github.com/aws/aws-toolkit-vscode/pull/4700 again?

package.json
```diff
@@ -30,8 +30,8 @@
         "testE2E": "npm run testE2E -w packages/ --if-present",
         "testInteg": "npm run testInteg -w packages/ --if-present",
         "package": "npm run package -w packages/toolkit",
-        "newChange": "ts-node ./scripts/newChange.ts",
-        "createRelease": "ts-node ./scripts/createRelease.ts",
+        "newChange": "echo 'Must specify subproject/workspace with -w packages/<subproject>' && false",
+        "createRelease": "echo 'Must specify subproject/workspace with -w packages/<subproject>' && false",
```
justinmk3, line 34: helpful!

justinmk3 (approved):

buildspec/release/10changeversion.yml
hayemaxi: No unless there is some edge case I can't think of. feature/standalone only releases amazonq, so only changelogs for amazonq can be updated and committed.

I also think we should cherry-pick this (and other buildspec changes) into mainline to reduce merge-conflicts.

PR #4704, opened 2024-04-12
Remove certain maven metadata from uploaded dependencies.

dhasani23:
## Problem

Same issue as in IntelliJ -- when we include certain Maven-related metadata in the upload ZIP, our backend sometimes fails to build the project.

## Solution

Skip metadata files during zipping process.

## License


damntrecky (approved):

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -195,6 +195,18 @@ export async function uploadPayload(payloadFileName: string) {
     return response.uploadId
 }
 
+/**
+ * Maven adds certain metadata to the dependencies in the local repository which are checked during compile time.
+ * If any `*.repositories` or `*.sha1` files are found in the local repository, maven will verify if the source repository is available from the machine.
+ * If it is not available, then Maven will attempt to redownload dependencies again even if dependencies are available locally.
+ * This behaviour cannot be disabled when invoking maven except for by removing these metadata files.
+ * @param fileEntry
```
ivikash, line 203: nit: Should it `path` instead of `fileEntry`?

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -195,6 +195,18 @@ export async function uploadPayload(payloadFileName: string) {
...
+ * @returns
+ */
+function isExcludedDependencyFile(path: string): boolean {
+    return path.endsWith('.repositories') || path.endsWith('.sha1')
```
ivikash, line 207: Can we create a module scoped array for `extensionsToBeExcluded` and then use that in the `isExcludedDependencyFile` function? Something like

```suggestion
/** 
 * Array of file extensions used by Maven as metadata in the local repository. 
 * Files with these extensions influence Maven's behavior during compile time, 
 * particularly in checking the availability of source repositories and potentially 
 * re-downloading dependencies if the source is not accessible. Removing these 
 * files can prevent Maven from attempting to download dependencies again.
 */
const extensionsToBeExcluded = ['.repositories', '.sha1'];

/**
 * Determines if the specified file path corresponds to a Maven metadata file
 * by checking against known metadata file extensions. This is used to identify
 * files that might trigger Maven to recheck or redownload dependencies from source repositories.
 *
 * @param path The file path to evaluate for exclusion based on its extension.
 * @returns {boolean} Returns true if the path ends with an extension associated with Maven metadata files; otherwise, false.
 */
function isExcludedDependencyFile(path: string): boolean {
    return extensionsToBeExcluded.some(extension => path.endsWith(extension));
}
```

arangatang: Changed it to this.

arangatang:
I am not sure about the failing CI task... there are no logs and no overview available. 
@ivikash is it possible to rerun failing tests?

damntrecky (approved):

nkomonen-amazon (approved):

PR #4705, opened 2024-04-12
fix: handle rejected promise in server tests

jpinkney-aws:
## Problem
- We get:
rejected promise not handled within 1 second: Error: AuthSSOServer: foo: foo
rejected promise not handled within 1 second: Error: AuthSSOServer: invalid state
rejected promise not handled within 1 second: Error: AuthSSOServer: missing code
rejected promise not handled within 1 second: Error: AuthSSOServer: missing state

## Solution
- Catch and test the authorization promise

## License

### DIFF
--- a/packages/core/src/auth/sso/server.ts
+++ b/packages/core/src/auth/sso/server.ts
@@ -204,11 +204,9 @@ export class AuthSSOServer {
             new Promise<Result<string>>((_, reject) => {
                 globals.clock.setTimeout(() => {
                     reject(
-                        Result.err(
-                            new ToolkitError('Timed-out waiting for browser login flow to complete', {
-                                code: 'TimedOut',
-                            })
-                        )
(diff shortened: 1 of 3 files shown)


packages/core/src/auth/sso/server.ts
```diff
@@ -190,18 +191,20 @@ export class AuthSSOServer {
...
         return Promise.race([
             this.authenticationPromise,
-            new Promise<string>((_, reject) => {
+            new Promise<Result<string>>((_, reject) => {
                 globals.clock.setTimeout(() => {
                     reject(
-                        new ToolkitError('Timed-out waiting for browser login flow to complete', {
-                            code: 'TimedOut',
-                        })
+                        Result.err(
```
justinmk3, line 203: should it be `resolve()` now? or maybe `reject(new ToolkitError(...))` still "works" for this case?

PR #4706, opened 2024-04-12
feat(amazonq): Rename public facing commands from CodeWhisperer to amazon q: Part 3

leigaol:
## Problem

Adopt Amazon Q as the new product name.

## Solution

Rename public facing commands from CodeWhisperer to amazon q. 

Screenshot 2024-04-12 at 2 29 03 PM
ae156-c516-4f4b-9510-07f98317bb45">

Note:
1. Only commands are renamed.
2. For view ids and other resources, as long as they are not visible to customers, they are not renamed.

## License


packages/core/package.nls.json
```diff
@@ -241,7 +241,7 @@
     "AWS.walkthrough.gettingStarted.connect": "Connect to AWS",
     "AWS.walkthrough.gettingStarted.changeRegions": "Change AWS Regions",
     "AWS.walkthrough.gettingStarted.setupToolchain": "Configure your toolchain",
-    "AWS.command.codewhisperer.title": "CodeWhisperer Invoke Service",
+    "AWS.command.codewhisperer.title": "Manual trigger inline suggestion",
```
ivikash, line 244: It may be well beyond the scope of this PR, though I am curious if should replace all instances of `AWS.command.codewhisperer` to `Amazon.command.amazonQ`??

leigaol: >Only commands are renamed.

only customer facing commands are renamed. 

>all instances of AWS.command.codewhisperer

These are the keys for strings that are not public facing. They are not blocker for new release. 

Eventually they should be renamed

ivikash: Thank you @leigaol for that. It is super helpful :)

PR #4711, opened 2024-04-12
fix(amazonq): Prevent Q activation if detected old aws toolkit; Q gain focus at first use

leigaol:
## Problem
Q is not compatible with old toolkit. 

## Solution

Prevent Q activation if detected old aws toolkit from the marketplace. 

It will not prevent Q activation if detected toolkit with a dev version. 

Auto installs new toolkit from marketplace and then prompt user to reload.

Let Q gain focus at first activation. 

## License


packages/amazonq/src/extensionShared.ts
```diff
@@ -36,6 +37,20 @@ export async function activateShared(context: vscode.ExtensionContext) {
     const contextPrefix = 'amazonq'
     globals.contextPrefix = 'amazonq.' //todo: disconnect from above line
 
+    // Avoid activation if older toolkit is installed
+    // Amazon Q is only compatible with AWS Toolkit >= 3.0.0
+    const toolkit = vscode.extensions.getExtension(VSCODE_EXTENSION_ID.awstoolkit)
+    if (toolkit) {
+        const toolkitVersion = semver.coerce(toolkit.packageJSON.version)
+        if (toolkitVersion && toolkitVersion.major < 3) {
```
ivikash, line 45: nit: We can make this shorter by using optional chaining operator.

```suggestion
        if (toolkitVersion?.major < 3) {
```

packages/amazonq/src/extensionShared.ts
```diff
@@ -36,6 +37,20 @@ export async function activateShared(context: vscode.ExtensionContext) {
...
+            await vscode.window.showInformationMessage(
+                `Amazon Q is not compatible with AWS Toolkit ${toolkit.packageJSON.version}. Please upgrade your AWS Toolkit extension to the latest version and then restart VS Code.`,
+                'OK'
```
hayemaxi, line 48: We should link to the extension page for toolkit in vscode.

packages/amazonq/src/extensionShared.ts
leigaol: Screenshot 2024-04-15 at 3 42 49 PM

ivikash: The optional chaining should've work and something feels a little weird here.

Verified om typescript lang - Typescript compiles `toolkitVersion?.major < 3` to `(toolkitVersion === null || toolkitVersion === void 0 ? void 0 : toolkitVersion.major) < 3;`

https://www.typescriptlang.org/play?#code/C4exBsGsEtgNQKYCcDO0QDsD8A6AtgIYBWISABADxkDMQA

packages/amazonq/src/extensionShared.ts
leigaol: Added in latest commit

packages/core/src/codewhisperer/activation.ts
```diff
@@ -263,7 +263,8 @@ export async function activate(context: ExtContext): Promise<void> {
         vscode.languages.registerCodeActionsProvider(
             [...CodeWhispererConstants.platformLanguageIds],
             SecurityIssueCodeActionProvider.instance
-        )
+        ),
+        vscode.commands.registerCommand('aws.codeWhisperer.openEditorAtRange', openEditorAtRange)
```
leigaol, line 267: This is to let the Amazon Q show the message when it is installed with old aws toolkit. Otherwise, it will fail before even showing that message.

hayemaxi: Cool. Out of scope for this PR but we should probably wrap it with our Commands.register interface instead.

packages/amazonq/src/extensionShared.ts
```diff
@@ -36,6 +37,22 @@ export async function activateShared(context: vscode.ExtensionContext) {
...
+    // Avoid activation if older toolkit is installed
+    // Amazon Q is only compatible with AWS Toolkit >= 3.0.0
+    // Or AWS Toolkit with a development version. Example: 2.19.0-3413gv
+    const toolkit = vscode.extensions.getExtension(VSCODE_EXTENSION_ID.awstoolkit)
+    if (toolkit) {
+        const toolkitVersion = semver.coerce(toolkit.packageJSON.version)
+        const isDevVersion = toolkit.packageJSON.version.toString().includes('-')
+        if (toolkitVersion && toolkitVersion.major < 3 && !isDevVersion) {
+            await vscode.window.showInformationMessage(
+                `The Amazon Q extension is incompatible with AWS Toolkit versions 2.9 and below. Your AWS Toolkit will be updated to version 3.0.`,
```
hayemaxi, line 49: will this string need to be updated to the latest toolkit version prior to release?

hayemaxi: Also, this makes it sound like we are installing the extension update for them. Is this possible?

packages/amazonq/src/extensionShared.ts
hayemaxi: I am not seeing the change

leigaol: It was in https://github.com/aws/aws-toolkit-vscode/pull/4711/commits/d651f8f743cd3504d0c31ff556a04aa969ed59b0. 

However, I reverted it because the PM document does not mention a extension page.

packages/amazonq/src/extensionShared.ts
```diff
@@ -36,6 +37,32 @@ export async function activateShared(context: vscode.ExtensionContext) {
...
+    // Avoid activation if older toolkit is installed
+    // Amazon Q is only compatible with AWS Toolkit >= 3.0.0
+    // Or AWS Toolkit with a development version. Example: 2.19.0-3413gv
+    const toolkit = vscode.extensions.getExtension(VSCODE_EXTENSION_ID.awstoolkit)
+    if (toolkit) {
+        const toolkitVersion = semver.coerce(toolkit.packageJSON.version)
+        const isDevVersion = toolkit.packageJSON.version.toString().includes('-')
+        if (toolkitVersion && toolkitVersion.major < 3 && !isDevVersion) {
+            await vscode.commands
+                .executeCommand('workbench.extensions.installExtension', VSCODE_EXTENSION_ID.awstoolkit)
```
leigaol, line 49: this installs latest version in marketplace

justinmk3:
followup https://github.com/aws/aws-toolkit-vscode/pull/4737

PR #4713, opened 2024-04-15
feat(q): dynamic menu/tree nodes for extensions

hayemaxi:
Problem: Tree/menu nodes were being re-used across extensions. This created issues for the underlying commands. For example, if toolkit wasn't installed then the switch to Q chat button would not work in Q's status bar menu.

Solution: Move these nodes to a common module and allow dynamic registration of the underlying commands.

Note: Should make the isWeb mode check unnecessary, but this wasn't tested.

## License


packages/core/src/amazonq/explorer/amazonQChildrenNodes.ts
```diff
@@ -31,6 +29,9 @@ export const dismissQTree = Commands.declare('aws.toolkit.amazonq.dismiss', () =
     await globals.context.globalState.update(amazonQDismissedKey, true)
     await vscode.commands.executeCommand('setContext', amazonQDismissedKey, true)
 })
+
+export const toolkitSwitchToAmazonQCommand = Commands.declare('_aws.toolkit.amazonq.focusView', () => switchToAmazonQ)
```
nkomonen-amazon, line 33: We will want to add a `source` argument to this, see `basicCommands.ts` for examples. You'll need to update the `switchToAmazonQ` function to have a `source` parameter. Then whenever you execute it will look something like `toolkitSwitchToAmazonQCommand.execute('mySource')`.

The "show amazon q chat" command will be able to be triggered by many different places (I am doing some work which will also call this), and we will want to know who called it.

To verify your work, whenever the command is executed look in the logs for `vscode_executeCommand`, with the name of the `command` equal to this command ID. Here is an example:

```
2024-04-16 11:31:43.162 [info] telemetry: vscode_executeCommand {
  Metadata: {
    parentMetric: 'vscode_executeCommand',
    command: 'aws.auth.manageConnections',
    source: 'codewhispererQuickPick',
    duration: '4',
    result: 'Succeeded',
    awsAccount: 'not-set',
    awsRegion: 'us-east-1'
  },
  Value: 1,
  Unit: 'None',
  Passive: true
}
```
Notice how the `source` is `codewhispererQuickPick`

packages/core/src/amazonq/explorer/commonNodes.ts
```diff
@@ -0,0 +1,90 @@
...
+import * as nls from 'vscode-nls'
+import { Command, DeclaredCommand } from '../../shared/vscode/commands2'
+import { codicon, getIcon } from '../../shared/icons'
+import { telemetry } from '../../shared/telemetry/telemetry'
+import { DataQuickPickItem } from '../../shared/ui/pickerPrompter'
+import { TreeNode } from '../../shared/treeview/resourceTreeDataProvider'
+
+const localize = nls.loadMessageBundle()
+
+export async function switchToAmazonQ(signIn: boolean = false) {
```
nkomonen-amazon, line 16: We want developers to only use the commands which consume this function and not this function itself.
We should put a warning here to not use this directly

packages/core/src/amazonq/explorer/amazonQChildrenNodes.ts
hayemaxi: I am working on follow up PRs that will be adding telemetry to these commands, so this will be OOS for this PR

nkomonen-amazon (approved):

PR #4715, opened 2024-04-16
fix(amazonq): show error messages in chat and match with IntelliJ

dhasani23:
## Problem

Errors were only being shown to the user via notifications, with a generic error (Something like "Amazon Q couldn't complete the transformation. Try again later.") shown in the chat in most cases. Also, error messages between VS Code and IntelliJ differed by a fair amount.

## Solution

Make the error messages between both IDEs more similar, and show the same message, which already shows via a notification, via the chat too.

## License


packages/core/src/amazonqGumby/activation.ts
```diff
@@ -52,8 +51,6 @@ export async function activate(context: ExtContext) {
             Commands.register('aws.amazonq.stopTransformationInHub', async (cancelSrc: CancelActionPositions) => {
                 if (transformByQState.isRunning()) {
                     void stopTransformByQ(transformByQState.getJobId(), cancelSrc)
-                } else {
-                    void vscode.window.showInformationMessage(CodeWhispererConstants.noOngoingJobMessage)
```
dhasani23, line 56: Not needed b/c the Stop button is disabled if no job is running

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -35,7 +36,6 @@ export type StaticTextResponseType =
 
 export type ErrorTextResponseType =
     | 'no-project-found'
-    | 'no-workspace-open'
```
dhasani23, line 38: Unused

packages/core/src/codewhisperer/models/model.ts
```diff
@@ -303,7 +303,9 @@ export class TransformByQState {
 
     private payloadFilePath: string = ''
 
-    private jobFailureErrorMessage: string = ''
+    private jobFailureErrorNotification: string | undefined = undefined
+
+    private jobFailureErrorChatMessage: string | undefined = undefined
```
dhasani23, line 309: Need separate messages for notifications and for chat messages since they differ slightly

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -138,10 +139,6 @@ export class DiffModel {
      */
     public parseDiff(pathToDiff: string, pathToWorkspace: string): ProposedChangeNode[] {
         const diffContents = fs.readFileSync(pathToDiff, 'utf8')
-        if (!diffContents) {
```
dhasani23, line 141: Won't happen, already handled by backend

nkomonen-amazon (approved):

packages/core/src/test/codewhisperer/commands/transformByQ.test.ts
```diff
@@ -67,12 +65,6 @@ describe('transformByQ', function () {
     })
 
     it('WHEN job is stopped THEN status is updated to cancelled', async function () {
-        const testWindow = getTestWindow()
-        testWindow.onDidShowMessage(message => {
-            if (message.message === stopTransformByQMessage) {
-                message.selectItem(startTransformByQ.stopTransformByQButton)
```
dhasani23, line 73: Deleted since we no longer have this pop-up asking users to confirm to stop their job.

justinmk3 (approved):
Thanks for the attention to cleaning up the code!

Waiting for transform team member to review

damntrecky (approved):

PR #4717, opened 2024-04-16
feat(amazonq): Notification string updates for standalone. Part 4 string changes

leigaol:
## Problem

Update strings in notifications to align with product requirement. This is part 4 of all string changes.

## License

### DIFF
--- a/packages/core/src/shared/telemetry/activation.ts
+++ b/packages/core/src/shared/telemetry/activation.ts
@@ -87,7 +87,7 @@ function showTelemetryNotice(extensionContext: vscode.ExtensionContext) {
 
     const telemetryNoticeText: string = localize(
         'AWS.telemetry.notificationMessage',
-        '{0} Toolkit collects anonymous usage metrics to help drive toolkit improvements. This can be changed in the settings.',
+        'AWS IDE Extensions collects anonymous usage metrics to improve the product. You can opt-out in settings.',
         getIdeProperties().company
     )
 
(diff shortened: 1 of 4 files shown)


packages/core/src/shared/telemetry/activation.ts
```diff
@@ -87,7 +87,7 @@ function showTelemetryNotice(extensionContext: vscode.ExtensionContext) {
 
     const telemetryNoticeText: string = localize(
         'AWS.telemetry.notificationMessage',
-        '{0} Toolkit collects anonymous usage metrics to help drive toolkit improvements. This can be changed in the settings.',
+        'AWS IDE Extensions collects anonymous usage metrics to improve the product. You can opt-out in settings.',
```
hayemaxi, line 90: I think this notice appears once for each extension. We will need to revisit that to only display once between extensions.

hayemaxi (approved):

packages/core/src/shared/telemetry/activation.ts
leigaol: It's in hasUserSeenTelemetryNotice. Stored in a global state key. Therefore, new Q users will see it.

justinmk3:
Try rebasing,  https://github.com/aws/aws-toolkit-vscode/pull/4706 was merged.

leigaol:
> Try rebasing, #4706 was merged.

DONE.

PR #4719, opened 2024-04-16
feat(amazonq): initial standalone telemetry

hayemaxi:
Implemented telemetry:
- auth_userState
  - Describes the state of auth after activation.
- toolkit_showNotification
  - The user was shown a notification about Amazon Q being auto-installed or about the extension and if they'd like to install it.
- toolkit_invokeAction
  - The user installed, learned more, or dismissed the previously mentioned notification.
- ui_click
  - amazonq_switchToQSignIn
- sources on various codewhisperer commands

Other stuff:
- Collapse amazon q commands into one set (registered from amazon q)
- ExtensionUse.wasUpdated util func.

Follow up items:
- Moving the definitions to common is a prioritized work item and will occur after this PR
- Login page telemetry

## License


packages/amazonq/src/extensionShared.ts
```diff
@@ -93,6 +95,34 @@ export async function activateShared(context: vscode.ExtensionContext) {
...
+
+        if (firstUse) {
+            telemetry.record({ source: ExtStartUpSource.FirstStartUp })
+        } else if (wasUpdated) {
+            telemetry.record({ source: ExtStartUpSource.Update })
+        } else {
+            telemetry.record({ source: ExtStartUpSource.Reload })
+        }
+
+        const authState = (await AuthUtil.instance.getChatAuthState()).codewhispererChat
```
nkomonen-amazon, line 113: BiD and IdC will now be at parity in terms of the Q features they offer, so I think we will only need to use `amazonQ` instead of `codewhispererChat`. We only had this discrepancy before since only some of Q worked with BiD

packages/core/src/amazonq/explorer/amazonQChildrenNodes.ts
```diff
@@ -25,12 +27,21 @@ export const qExtensionPageCommand = Commands.declare('aws.toolkit.amazonq.exten
     void vscode.env.openExternal(vscode.Uri.parse(`vscode:extension/${VSCODE_EXTENSION_ID.amazonq}`))
 })
 
-export const dismissQTree = Commands.declare('aws.toolkit.amazonq.dismiss', () => async () => {
-    await globals.context.globalState.update(amazonQDismissedKey, true)
-    await vscode.commands.executeCommand('setContext', amazonQDismissedKey, true)
-})
+export const dismissQTree = Commands.declare(
+    { id: '_aws.toolkit.amazonq.dismiss', compositeKey: { 1: 'source' } },
```
nkomonen-amazon, line 31: Update: I see you used it correctly in the other uses.

---

The `1` in `{ 1: 'source'}` means the arg at index `1` of the below `(source: string`, which is incorrect. I'm assuming that the source is not properly set in the final metric.

You can verify in the Toolkit Logs, once you execute this command search for something like `command: _aws.toolkit.amazonq.dismiss` in a `vscode_executeCommand` metric. Then see if the `source` field is what you expect.

packages/core/src/amazonq/explorer/amazonQChildrenNodes.ts
```diff
@@ -25,12 +27,21 @@ export const qExtensionPageCommand = Commands.declare('aws.toolkit.amazonq.exten
...
+    () => async (source: string) => {
+        await telemetry.toolkit_invokeAction.run(async () => {
+            telemetry.record({
+                source: ExtensionUse.instance.isFirstUse() ? ExtStartUpSource.FirstStartUp : ExtStartUpSource.None,
```
nkomonen-amazon, line 35: We've been avoiding using Enums (`ExtStartUpSource`) in favor of an object with `const`. There are some quirks with it, and it is better to use the basic types when possible.

Example of how we do it in our code:
```
export const AuthSources = {
    addConnectionQuickPick: 'addConnectionQuickPick',
    firstStartup: 'firstStartup',
    codecatalystDeveloperTools: 'codecatalystDeveloperTools',
    vscodeComponent: vscodeComponent,
    cwQuickPick: cwQuickPickSource,
    cwTreeNode: cwTreeNodeSource,
    amazonQChat: amazonQChatSource,
    authNode: 'authNode',
} as const

export type AuthSource = (typeof AuthSources)[keyof typeof AuthSources]
```

Isn't a blocker for this PR, but is probably worth to update at some point. Your code won't have to change much though. `ExtStartUpSource` can just become the `AuthSources` equivalent, and then make a type from that.

packages/core/src/auth/utils.ts
```diff
@@ -634,28 +642,63 @@ export class ExtensionUse {
...
 
+    /**
+     * Checks if the extension has been updated since the last time it was used.
+     * If this is the first used version of the extension, it returns true.
+     * If the extension has been downgraded, it returns true.
+     *
+     * Caveat: This may return true even if an update hadn't occurred IF
+     * this function hasn't been called.
+     */
+    wasUpdated(state: vscode.Memento = globals.context.globalState) {
```
nkomonen-amazon, line 658: Since we'll know the specific version that this item is released, maybe we can use that to our advantage to get a more accurate result? I'm not sure exactly how we intend to use this so maybe my recommendation is not useful.

packages/amazonq/src/extensionShared.ts
hayemaxi: Good call out. I am going to add researching this as a follow up item to our checklist. Last time I tested this, Q was working fine despite returning false for .amazonq but not .codewhispererChat.

packages/core/src/auth/utils.ts
hayemaxi: I think you are asking about checking updating from the regular extension to split versions. This check is just to see if the extension was updated since the last load in general, it is not specific to the split.

packages/core/src/amazonq/explorer/amazonQChildrenNodes.ts
hayemaxi: missed this one, fixed

nkomonen-amazon (approved):

packages/core/src/codewhisperer/commands/types.ts
```diff
@@ -11,6 +12,12 @@ export const cwTreeNodeSource = 'codewhispererTreeNode'
 export const cwQuickPickSource = 'codewhispererQuickPick'
 /** Indicates a CodeWhisperer command was executed through the Amazon Q chat pane */
 export const amazonQChatSource = 'amazonQChat'
+/** Indicates a CodeWhisperer command was executed during the first start of the extension */
+export const firstStartUpSource = ExtStartUpSources.firstStartUp
+/** Indicates a CodeWhisperer command was executed as a result of signing out */
+export const cwSignOut = 'codewhispererSignOut'
+/** Indicates a CodeWhisperer command was executed as a result of selecting an ellipses menu item */
+export const cwEllipsesMenu = 'ellipsesMenu'
```
justinmk3, line 20: Why are these extra variables needed? Using literals works the same (including IDE completion, goto definition, etc), and avoids this extra boilerplate.

packages/core/src/shared/telemetry/util.ts
```diff
@@ -203,3 +203,15 @@ export async function setupTelemetryId(extensionContext: vscode.ExtensionContext
...
+ * Potentially helpful values for the 'source' field in telemetry.
+ */
+export const ExtStartUpSources = {
+    firstStartUp: 'firstStartUp',
+    update: 'update',
+    reload: 'reload',
+    none: 'none',
+} as const
+
+export type ExtStartUpSource = (typeof ExtStartUpSources)[keyof typeof ExtStartUpSources]
```
justinmk3, line 217: This pattern is useful if any code needs to *iterate* these values. Otherwise it's not necessary.

PR #4721, opened 2024-04-16
feat(PolicyCheck): create Webview HTML, connect Webview to command handler

kevluu-aws:
## Problem
Add a new feature to use IAM Access Analyzer policy checks(validatePolicy, checkNoNewAccess, checkAccessNotGranted)

## Solution
Setup Webview and command handler as the basis of the feature

Screenshot of Webview window that shows after invocation from command palette: 
Screenshot 2024-04-16 at 4 30 38 PM

## License


packages/core/src/accessanalyzer/vue/policyChecks.ts
```diff
@@ -0,0 +1,128 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+import * as fs from 'fs'
+import { getLogger, Logger } from '../../shared/logger'
+import { localize } from '../../shared/utilities/vsCodeUtils'
+import { VueWebview } from '../../webviews/main'
+import { ExtContext } from '../../shared/extensions'
+//import { telemetry } from '../../shared/telemetry/telemetry'
```
kajalrj-aws, line 12: Remove this commented out unused import.

packages/core/src/accessanalyzer/vue/policyChecks.ts
```diff
@@ -0,0 +1,128 @@
...
+    tfConfigPath: string
+    cfnParameterPath: string
+    referenceDocument: string
+}
+
+export class PolicyChecksWebview extends VueWebview {
+    public static readonly sourcePath: string = 'src/accessanalyzer/vue/index.js'
+    public readonly id = 'policyChecks'
+
+    //private readonly logger = getLogger()
```
kajalrj-aws, line 26: Same comment as above.

packages/core/src/accessanalyzer/vue/policyChecks.ts
kevluu-aws: We will be implementing Telemetry soon, kept it there as a reminder

packages/core/src/accessanalyzer/activation.ts
```diff
@@ -0,0 +1,25 @@
...
+import * as vscode from 'vscode'
+import { ExtContext } from '../shared/extensions'
+import { DevSettings } from '../shared/settings'
+import { Commands } from '../shared/vscode/commands2'
+import { renderPolicyChecks } from './vue/policyChecks'
+
+/**
+ * Activate Policy Checks functionality for the extension.
+ */
+export async function activate(extContext: ExtContext): Promise<void> {
```
hayemaxi, line 15: you will need to call this from the main extension activation. `packages/core/src/extension.ts`

packages/core/package.json
```diff
@@ -263,6 +263,24 @@
                     "type": "object",
                     "description": "%AWS.configuration.description.lambda.recentlyUploaded%",
                     "default": []
+                },
```
mponaws, line 266: On another note - please confirm the other parameters that you could pass to the CLI and if that is something we would like customers to define their config file for this extension.

packages/core/package.json
```diff
@@ -2190,6 +2208,12 @@
             ]
         },
         "commands": [
+            {
```
mponaws, line 2211: Is this the name for the task/command? We should confirm the name of it - I was wondering if we should call it `Policy Validator` to keep it aligned with our branding with other tools?

packages/core/package.json
```diff
@@ -263,6 +263,24 @@
...
+                    "description": "File path or S3 path to a control policy for Custom Policy Checks.",
+                    "scope": "window"
+                },
+                "aws.accessAnalyzer.policyChecks.cloudFormationParameterFilePath": {
+                    "type": "string",
+                    "default": "",
+                    "description": "Keys and values for CloudFormation template parameters. Only parameters that are referenced by IAM policies in the template are required.",
+                    "scope": "machine-overridable"
+                },
+                "aws.accessAnalyzer.policyChecks.terraformConfigFilePath": {
```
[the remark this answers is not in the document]
kevluu-aws: I will double check! I am leaning towards not allowing the customers to override it. And if we do, probably make it optional and provide the default if not inputted.

packages/core/package.json
kevluu-aws: We are only targeting the necessary parameters for this launch.

packages/core/package.json
kevluu-aws: The name will be `IAM Policy Checks`

packages/core/package.json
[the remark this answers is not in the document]
kevluu-aws: I checked in with Matt, we will enforce the standard default config. Thanks for raising this!

packages/core/src/accessanalyzer/vue/policyChecks.ts
```diff
@@ -0,0 +1,128 @@
...
+                referenceFilePath: referencePolicyFilePath ? referencePolicyFilePath : '',
+                tfConfigPath: tfConfigPath ? tfConfigPath : '',
+                cfnParameterPath: cfnParameterPath ? cfnParameterPath : '',
+                referenceDocument: _getReferenceDocument(referencePolicyFilePath),
+            },
+            client
+        )
+        await wv.show({
+            viewColumn: vscode.ViewColumn.Beside,
+            title: localize('AWS.policyChecks.title', 'IAM Policy Checks'),
```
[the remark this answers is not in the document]
kevluu-aws: We can! Not at the top of my list right now; will be sure to try and include it where I can. Not sure if we can sanitize our responses to be localized or not though. And that might be a jarring experience for users.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -0,0 +1,146 @@
...
+async function _getReferenceDocument(input: string): Promise<string> {
+    if (fs.existsSync(input)) {
+        return fs.readFileSync(input).toString()
+    } else {
+        try {
+            const { region, bucket, key } = AmazonS3URI(input)
+            const s3Client = new S3({ region })
+            const resp = await s3Client.getObject({ Bucket: bucket, Key: key }).promise()
+            return resp.Body!.toString()
+        } catch (e: any) {}
```
mponaws, line 143: We should surface the error here for customers to understand the issue with s3 URI passed

packages/core/package.json
```diff
@@ -263,6 +263,18 @@
                     "type": "object",
                     "description": "%AWS.configuration.description.lambda.recentlyUploaded%",
                     "default": []
+                },
+                "aws.accessAnalyzer.policyChecks.referencePolicyFilePath": {
+                    "type": "string",
+                    "default": "",
+                    "description": "File path or S3 path to a control policy for Custom Policy Checks.",
+                    "scope": "window"
```
mponaws, line 271: Q: What's the difference between scope `window` & `machine-overridable`

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -0,0 +1,146 @@
...
+            .getConfiguration()
+            .get('aws.accessAnalyzer.policyChecks.referencePolicyFilePath')!
+        const cfnParameterPath: string = vscode.workspace
+            .getConfiguration()
+            .get('aws.accessAnalyzer.policyChecks.cloudFormationParameterFilePath')!
+
+        //Check if Cfn and Tf tools are installed
+        var cfnToolInstalled = true
+        var tfToolInstalled = true
+        exec('tf-policy-validator', err => {
```
mponaws, line 99: Q: do we plan to surface other errors from the CLI tool to VS code customers?

packages/core/package.json
```diff
@@ -263,6 +263,18 @@
                     "type": "object",
                     "description": "%AWS.configuration.description.lambda.recentlyUploaded%",
                     "default": []
+                },
+                "aws.accessAnalyzer.policyChecks.referencePolicyFilePath": {
+                    "type": "string",
+                    "default": "",
+                    "description": "File path or S3 path to a control policy for Custom Policy Checks.",
```
weibenz1, line 270: Q: does the user need to have access to this s3 object? or is it a preregistered url?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -0,0 +1,146 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+import * as fs from 'fs'
+import { getLogger, Logger } from '../../shared/logger'
+import { localize } from '../../shared/utilities/vsCodeUtils'
+import { VueWebview } from '../../webviews/main'
+import { ExtContext } from '../../shared/extensions'
+//import { telemetry } from '../../shared/telemetry/telemetry'
```
weibenz1, line 12: Q: is this placeholder?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -0,0 +1,146 @@
...
+import { AccessAnalyzer, S3 } from 'aws-sdk'
+import { exec } from 'child_process'
+const AmazonS3URI = require('amazon-s3-uri')
+
+export interface IamPolicyChecksInitialData {
+    referenceFilePath: string
+    cfnParameterPath: string
+    referenceDocument: string
```
weibenz1, line 20: If I read it correct, referenceFilePath is from customer input whereas referenceDocument is from reading the file. Can we add some comments to explain this?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -0,0 +1,146 @@
...
+                if (err.message.includes('command not found')) {
+                    cfnToolInstalled = false
+                }
+            }
+        })
+
+        const wv = new Panel(
+            context.extensionContext,
+            {
+                referenceFilePath: referencePolicyFilePath ? referencePolicyFilePath : '',
```
weibenz1, line 117: Q: Isn't this a required parameter?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -0,0 +1,146 @@
...
+        return fs.readFileSync(input).toString()
+    } else {
+        try {
+            const { region, bucket, key } = AmazonS3URI(input)
+            const s3Client = new S3({ region })
+            const resp = await s3Client.getObject({ Bucket: bucket, Key: key }).promise()
+            return resp.Body!.toString()
+        } catch (e: any) {}
+    }
+    return ''
```
weibenz1, line 145: Q: is this not considered to be an error?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
weibenz1: should we mark this as readonly?

packages/core/package.json
mponaws: User should have access to this S3 object. This object is owned by customers and use for storing their reference policies

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: Agreed, I will be updating a way to surface the error alongside some UI updates.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: Yes, that will be included when I add the CLI implementations.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: Yes it is a placeholder

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: Yes, I think we can mark as readonly and I will add comments to denote that.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: Yes, but it is not always provided at the start. It is inputted by the user through a setting or by inputting it directly in the Webview.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: It is, and in this version I decided not to surface these errors because I was afraid it was too noisy if a user were to change their reference path frequently. I will be updating this to surface errors in a more user-friendly way.

packages/core/package.json
kevluu-aws: `machine-overridable`can only be overridden by workspace or folder settings, whereas `window` can be overriden by user, workspace, or remote settings.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -0,0 +1,146 @@
...
+async function _getReferenceDocument(input: string): Promise<string> {
+    if (fs.existsSync(input)) {
+        return fs.readFileSync(input).toString()
+    } else {
+        try {
+            const { region, bucket, key } = AmazonS3URI(input)
+            const s3Client = new S3({ region })
+            const resp = await s3Client.getObject({ Bucket: bucket, Key: key }).promise()
+            return resp.Body!.toString()
+        } catch (e: any) {}
```
hayemaxi, line 143: instead of taking a new dependency on a package for one usage, can we just use or add to util functions from https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/clients/s3Client.ts ?

We should also use this to get the s3Client rather than constructing it here.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
hayemaxi, line 12: instead of a comment, maybe you can track this in a checklist or add // TODO: add telemetry here?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
weibenz1: `because I was afraid it was too noisy if a user were to change their reference path frequently.`, if the changes cannot be resolved to a correct reference document, the plugin would simply not work, and user should be aware that this is an error right?

Or maybe you can elaborate a bit on what's your thought on `surface errors in a more user-friendly way.` mean?

kevluu-aws: I was planning to surface all errors with notifications, which would be pretty annoying if you're typing out the path and notifications keep popping up. In the above, we display a "preview" of the file; if it is empty then I assumed users would be able to tell that it is not being read.

I am going to surface errors in the Webview itself, as an error message next to the input path.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: I agree that it would be nice to not have a new dependency for one thing. Unfortunately, S3 as a service does not have a built-in way of extracting, or pulling, from an S3 URI(outside of Java SDK). This new dependency is an adoption of that Java implementation into Typescript. I can add a new helper function, but it will just be using the same algorithm as AmazonS3URI.

kevluu-aws: I also construct the S3Client here because the region is variable depending on the provided link. I supposed that a new client with the appropriate region would need to be constructed every time.

hayemaxi:
This needs an approval from your team.

mponaws (approved):

mponaws (approved):

hayemaxi:
The dependency is minimal, I think its fine if you address that and the S3 Client in another PR, given that this is an experimental feature.

hayemaxi (approved):

packages/core/package.json
```diff
@@ -2001,6 +2013,37 @@
...
+            {
+                "command": "aws.amazonq.optimizeCode",
+                "title": "%AWS.command.amazonq.optimizeCode%",
+                "category": "%AWS.title%"
+            },
+            {
+                "command": "aws.amazonq.sendToPrompt",
+                "title": "%AWS.command.amazonq.sendToPrompt%",
+                "category": "%AWS.title%"
+            },
```
hayemaxi, line 2046: Looks like a merge error, these shouldn't be in core.

```suggestion
```

hayemaxi (approved):

mponaws (approved):

.github/CODEOWNERS
```diff
@@ -3,3 +3,4 @@ packages/core/src/codewhisperer/ @aws/codewhisperer-team
 packages/core/src/amazonqFeatureDev/ @aws/earlybird
 packages/core/src/codewhispererChat/ @aws/dexp
 packages/core/src/amazonq/ @aws/dexp
+packages/core/src/accessanalyzer/ @aws/access-analyzer
```
justinmk3, line 6: @kevluu-aws please consider moving your code to `packages/toolkit/src/accessanalyzer` . That is where Toolkit-only code should live. /core/ is for code that may be useful for both Toolkit and Amazon Q.

packages/core/src/shared/settings.ts
```diff
@@ -727,6 +727,7 @@ const devSettings = {
     codecatalystService: Record(String, String),
     codewhispererService: Record(String, String),
     ssoCacheDirectory: String,
+    enableIamPolicyChecksFeature: Boolean,
```
justinmk3, line 730: You may want to enable this is an "experiment" instead of a dev-only feature. See https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#preview-releases-and-experiments

.github/CODEOWNERS
kevluu-aws: I will move it to the Toolkit folder in a future PR.

packages/core/src/shared/settings.ts
kevluu-aws: I will enable it as an experiment. Thanks for the tip!

PR #4722, opened 2024-04-17
fix(amazonq): Improve login CSS. Part 1.

leigaol:
## Problem

Login CSS does not match the design

## Solution
Screenshot 2024-04-16 at 8 38 56 PM
Screenshot 2024-04-16 at 8 39 09 PM
Screenshot 2024-04-16 at 8 40 46 PM
Screenshot 2024-04-16 at 8 50 58 PM

1. Support VS Code light mode and dark mode
2. Match CSS with the design ( as close as it is). But for some font size I cannot make it as large as the design. 
3. Fix a bug when icon is not correctly showed.
4. Fix some strings in the login UX. 

This is the first part of CSS update. I will make another PR after fit & finish review. There are still some not finalized strings in the login UX. 

Q: Why font family change to Amazon Ember is not in this PR? 
A: I will do it in another PR. I am trying to reduce the scope of this PR.

## License


justinmk3 (approved):

leigaol:
> quick question: why can't we match the font sizes?
> 
> Also,
> 
> > Q: Why font family change to Amazon Ember is not in this PR?
> > A: I will do it in another PR. I am trying to reduce the scope of this PR.
> 
> This is fine, but is there a technical limitation or some large increase in effort that prevents this in this PR?

I was wrong. I did match all known font sizes in this PR.

PR #4724, opened 2024-04-17
fix(amazonqFeatureDev): mitigate race conditions on session creation

sannicm:
## Problem

When a customer starts a /dev chat with a message, e.g. /dev <do something>, the connector logic calls two event handlers in amazonqFeatureDev logic parallely that try to create a session, this creates a race condition where the customer could get to code generation with the wrong session, i.e. one with no conversationId.

## Solution

Add a lock to the getSession function in order to mitigate concurrent calls to getSession.

This is a similar approach to the one taken in JetBrain's toolkit: [ChatSessionStorage.kt](https://github.com/aws/aws-toolkit-jetbrains/blob/ec168fc6e9306f22f07945fd516d6acdc7fe396e/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/storage/ChatSessionStorage.kt#L18)

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/storages/chatSession.ts
+++ b/packages/core/src/amazonqFeatureDev/storages/chatSession.ts
@@ -3,11 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import AsyncLock from 'async-lock'
 import { Messenger } from '../controllers/chat/messenger/messenger'
 import { Session } from '../session/session'
 import { createSessionConfig } from '../session/sessionConfigFactory'
-
+import { getLogger } from '../../shared/logger'
 export class ChatSessionStorage {
+    private lock = new AsyncLock()
+
     private sessions: Map<string, Session> = new Map()
 
     constructor(private readonly messenger: Messenger) {}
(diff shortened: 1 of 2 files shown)


jpinkney-aws (approved):
@sannicm Just to link everything together can you point a link to the jetbrains code where we had to do the same thing

justinmk3:
Looks like yet another reason why the "session" concept shouldn't be copy-pasted in each Q subfeature. Will other Q features (transform, refactor-assist, etc) potentially encounter the same bug?

Is there a customer-facing improvement that can be mentioned in the changelog? E.g. "Amazon Q [subfeature]: potential duplicate sessions"

packages/core/src/amazonqFeatureDev/storages/chatSession.ts
```diff
@@ -20,12 +23,14 @@ export class ChatSessionStorage {
     }
 
     public async getSession(tabID: string): Promise<Session> {
-        const sessionFromStorage = this.sessions.get(tabID)
-        if (sessionFromStorage === undefined) {
-            // If a session doesn't already exist just create it
-            return this.createSession(tabID)
-        }
-        return sessionFromStorage
+        return this.lock.acquire(tabID, async () => {
```
ivikash, line 26: I'm curious if we can have `this.lock.acquire` as decorator?

sannicm:
> @sannicm Just to link everything together can you point a link to the jetbrains code where we had to do the same thing

https://github.com/aws/aws-toolkit-jetbrains/blob/main/plugins/amazonq/chat/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonqFeatureDev/storage/ChatSessionStorage.kt#L18

sannicm:
> Looks like yet another reason why the "session" concept shouldn't be copy-pasted in each Q subfeature. Will other Q features (transform, refactor-assist, etc) potentially encounter the same bug?
> 
> Is there a customer-facing improvement that can be mentioned in the changelog? E.g. "Amazon Q [subfeature]: potential duplicate sessions"

I cannot estimate the chances they will encounter the same issue as every feature has its own integration with mynah-ui. It depends if their implementation of the integration with mynah-ui/amazonq chat fires multiple events together and if they handle that correctly.

For amazonqFeatureDev, the best solution would be to update the `handleDevFeatureCommand` in the quick actions: `src/amazonq/webview/ui/quickActions/handler.ts` file making it only send one event (e.g. just using the `onKnownTabOpened`) even if there is a command message, i.e. the text in `/dev <text>` which currently we do not have capacity, it will require a parallel effort in JetBrains to update the same and test. 

This is not though reducing the code duplication you notice.

packages/core/src/amazonqFeatureDev/storages/chatSession.ts
sannicm: That is still a typescript's experimental feature, correct?

justinmk3: Would that make the code clearer, or less obvious?

PR #4726, opened 2024-04-17
fix: only bind auth server to 127.0.0.1

jpinkney-aws:
## Problem
- Auth server was being bound to all interfaces

## Solution
- Bind it only to 127.0.0.1

## License

### DIFF
--- a/packages/core/src/test/credentials/sso/server.test.ts
+++ b/packages/core/src/test/credentials/sso/server.test.ts
@@ -105,4 +105,13 @@ describe('AuthSSOServer', function () {
         const token = await server.waitForAuthorization()
         assert.deepStrictEqual(code, token)
     })
+
+    it('address is bound to localhost', function () {
+        const address = server.getAddress()
+        if (address instanceof Object) {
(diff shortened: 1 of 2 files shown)


packages/core/src/test/credentials/sso/server.test.ts
```diff
@@ -105,4 +105,13 @@ describe('AuthSSOServer', function () {
...
         assert.deepStrictEqual(code, token)
     })
+
+    it('address is bound to localhost', function () {
+        const address = server.getAddress()
+        if (address instanceof Object) {
+            assert.deepStrictEqual(address.address, '127.0.0.1')
+            return
+        }
+        assert.fail('Expected address 127.0.0.1')
```
justinmk3, line 115: somewhat verbose. just mentioning for future reference (also easier to follow):

```suggestion
        const address = server.getAddress()
        assert.deepStrictEqual((address as any).address, '127.0.0.1')
```

justinmk3 (approved):

PR #4730, opened 2024-04-17
fix(amazonq): Let CodeCatalyst reuse connections in Amazon Q

leigaol:
## Problem

 CodeCatalyst cannot reuse connections in Amazon Q

## Solution

Let CodeCatalyst reuse connections in Amazon Q

## License

### DIFF
--- a/packages/core/src/login/webview/vue/toolkit/backend_toolkit.ts
+++ b/packages/core/src/login/webview/vue/toolkit/backend_toolkit.ts
@@ -7,7 +7,7 @@ import * as vscode from 'vscode'
 import { tryAddCredentials } from '../../../../auth/utils'
 import { getLogger } from '../../../../shared/logger'
 import { AuthError, CommonAuthWebview } from '../backend'
-import { AwsConnection, createSsoProfile } from '../../../../auth/connection'
+import { AwsConnection, createSsoProfile, scopesCodeCatalyst } from '../../../../auth/connection'
 import { Auth } from '../../../../auth/auth'
 import { CodeCatalystAuthenticationProvider } from '../../../../codecatalyst/auth'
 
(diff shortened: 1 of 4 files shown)


packages/core/src/login/webview/vue/toolkit/backend_toolkit.ts
```diff
@@ -75,14 +75,54 @@ export class ToolkitLoginWebview extends CommonAuthWebview {
...
+     */
     async useConnection(connectionId: string): Promise<AuthError | undefined> {
-        return undefined
+        return this.ssoSetup('useConnection', async () => {
+            const conn = await Auth.instance.getConnection({ id: connectionId })
+            if (conn === undefined || conn.type !== 'sso') {
+                return
+            }
+            if (this.isCodeCatalystLogin) {
+                if (conn.scopes?.includes(scopesCodeCatalyst[0])) {
```
nkomonen-amazon, line 110: - This codecatalyst code is a better fit in the class `CodeCatalystAuthenticationProvider`. The methods `connectToAwsBuilderID()` and `connectToEnterpriseSso()` are very similar to what you are doing, so you probably just need to do some refactorings to create a `useConnection()` method in `CodeCatalystAuthenticationProvider` that reuses some of their code
- The function `isValidCodeCatalystConnection()` is also the correct way to determine if the connection is usable, you can see this used in the methods I mentioned above.

leigaol: >This codecatalyst code is a better fit in the class CodeCatalystAuthenticationProvider

Not all of these code. Most code are specific to the new login webview and it contains cross extension interaction that should only exists in the webview backend code. These code utilizes CodeCatalystAuthenticationProvider to login to CC however it also updates the states of the login webview. CodeCatalystAuthenticationProvider should never be able to update login webview states. 


>The function isValidCodeCatalystConnection() is also the correct way to determine if the connection is usable, you can see this used in the methods I mentioned above.

These are exactly the same as `(conn.scopes?.includes(scopesCodeCatalyst[0]))` since the scopes only has one element in the list. I can make the change in case future scopes changes.

nkomonen-amazon (approved):

PR #4732, opened 2024-04-18
amazonq: habit building features

nkomonen-amazon:
## Problem

We want more ways for users to use Amazon Q chat.

## Solution

This adds the following features:
- For first time users will show a CodeLens(small grey text above the current line) that prompts users to try Q Chat, and if clicked will open it
  - We only show this codelens 10 times (can adjust if necessary). Then never show it again. We do not want to annoy the user.
- A new keyboard shortcut `cmd` + `i` which will open Q chat
- The right click `Send to Amazon Q` menu will always show now, before it only showed when a line was highlighted.
  - If the user does a right click but no highlight, the entire line will be used in the command

Screenshot 2024-04-22 at 12 06 37 PM

## License

### DIFF
--- a/packages/core/src/codewhispererChat/editor/codelens.ts
+++ b/packages/core/src/codewhispererChat/editor/codelens.ts
@@ -1,0 +1,143 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import vscode from 'vscode'
+import globals from '../../shared/extensionGlobals'
+import { ToolkitError } from '../../shared/errors'
+import { Commands, placeholder } from '../../shared/vscode/commands2'
+import { platform } from 'os'
+import { focusAmazonQPanel } from '../commands/registerCommands'
+
+/** When the user clicks the CodeLens that prompts user to try Amazon Q chat */
+export const tryChatCodeLensCommand = Commands.declare(`_aws.amazonq.tryChatCodeLens`, () => async () => {
+    await focusAmazonQPanel.execute(placeholder, 'codeLens')
+})
+
+/**
+ * As part of hinting at users to use Amazon Q Chat, we will show codelenses
+ * prompting them a certain amount of time/uses. Then after
+ * a certain amount of times we will never show it again.
+ *
+ * This codelens appears above every clicked line.
+ */
+export class TryChatCodeLensProvider implements vscode.CodeLensProvider {
+    private _onDidChangeCodeLenses = new vscode.EventEmitter<void>()
+    onDidChangeCodeLenses: vscode.Event<void> = this._onDidChangeCodeLenses.event
+
+    /** How many times we've shown the CodeLens */
+    private count: number = 0
+    /** How many times we want to show the CodeLens */
+    static readonly maxCount = 10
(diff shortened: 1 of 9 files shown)


packages/core/src/codewhispererChat/editor/codelens.ts
```diff
@@ -0,0 +1,149 @@
...
+        this.disposables.push(
+            vscode.window.onDidChangeActiveTextEditor(() => this._onDidChangeCodeLenses.fire()),
+            vscode.window.onDidChangeTextEditorSelection(() => this._onDidChangeCodeLenses.fire())
+        )
+    }
+
+    static async register(): Promise<boolean> {
+        const shouldShow = globals.context.globalState.get(this.showCodeLensId, true)
+        if (!shouldShow) {
+            // We do not want to show the codelens. The stub disposable to makes the caller happy.
```
jpinkney-aws, line 53: Is this supposed to be for line 57?

packages/core/src/test/codewhispererChat/editor/codelens.test.ts
```diff
@@ -0,0 +1,72 @@
...
+            ])
+
+            codeLensCount++
+        }
+        const emptyResult = await instance.provideCodeLenses({} as any, new vscode.CancellationTokenSource().token)
+        assert.deepStrictEqual(emptyResult, [])
+    })
+
+    it('does not register the provider if we do not want to show the code lens', async function () {
+        await globals.context.globalState.update(TryChatCodeLensProvider.showCodeLensId, true)
```
jpinkney-aws, line 63: Should globals be mocked so that it doesn't effect the actual runtime global state? (or does the test harness already do this?)

packages/core/src/test/codewhispererChat/editor/codeAction.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import assert from 'assert'
+import { ChatExplainCodeActionProvider } from '../../../codewhispererChat/editor/codeAction'
+import vscode from 'vscode'
+import { explainCodeCommandId } from '../../../codewhispererChat/commands/registerCommands'
+
+describe('Code Action', () => {
+    const instance: ChatExplainCodeActionProvider = new ChatExplainCodeActionProvider()
+
+    it('should have a test', async () => {
```
jpinkney-aws, line 14: I can't tell if this is a default test name or not 😄. Should it be should show code action?

packages/core/src/codewhispererChat/editor/context/focusArea/focusAreaExtractor.ts
```diff
@@ -36,7 +36,8 @@ export class FocusAreaContextExtractor {
 
         // It means we don't really have a selection, but cursor position only
         if (!this.isCodeBlockSelected(editor)) {
-            importantRange = editor.visibleRanges[0]
+            // Select the whole line
+            importantRange = editor.document.lineAt(importantRange.start.line).range
```
jpinkney-aws, line 40: does cw know this will change?

packages/core/src/codewhispererChat/editor/codelens.ts
nkomonen-amazon: I think this was for some old code, removed

jpinkney-aws (approved):

PR #4733, opened 2024-04-18
fix(amazonFeatureDev): change Timeouts values

dianaabv16:
## Problem

Why those lines of code are removed:
```
describe('Moves directly from approach to codegen', () => {
  codegenTests()
})
```
Current set up:
approach phase -> moves Moves directly from approach to codegen (generate code -> insert changes -> generate code-> work on the new task -> generate code -> CloseSession) -> iterate on approach => when trying iterate on approach when task is finished, task cannot be continued and the following error is shown ->  `Amazon Q feature development: failed to generate presigned url: Resource is in an invalid state`

Set up after this fix:
approach phase ->  iterate on approach->code gen started -> code gen iterated -> code gen accepted -> coden gen new task -> session closed

----
Why those lines of code are removed:
```
describe('Clicks accept code', () => {
   insertCodeTests()
})
```
Current setup:
```
function codegenTests(){
  ...
   insertCodeTests()
   describe('Iterates on codegen', () => {
      ... 
       insertCodeTests()
   })
}
```
 insertCodeTests() calls  `tab.clickButton(FollowUpTypes.CloseSession)` => which means Iterate on codegen on the code gen won't be possible with the same error:
 
`Amazon Q feature development: failed to generate presigned url: Resource is in an invalid state`

Tested:
```
11 passing (33m)
globalSetup: after()
No test coverage found
```

## License

### DIFF
--- a/packages/core/src/testE2E/amazonq/featureDev.test.ts
+++ b/packages/core/src/testE2E/amazonq/featureDev.test.ts
@@ -19,11 +19,11 @@ describe('Amazon Q Feature Dev', function () {
     let framework: qTestingFramework
     let tab: Messenger
 
-    const maxTestDuration = 300000
+    const maxTestDuration = 900000
     const prompt = 'Implement fibonacci in python'
     const iterateApproachPrompt = prompt + ' and add a unit test'
     const codegenApproachPrompt = prompt + ' and add a readme that describes the changes'
-    const tooManyRequestsWaitTime = 600000
+    const tooManyRequestsWaitTime = 900000
 
     before(async function () {
         /**
--- a/packages/core/src/testE2E/amazonq/featureDev.test.ts
+++ b/packages/core/src/testE2E/amazonq/featureDev.test.ts
@@ -19,11 +19,11 @@ describe('Amazon Q Feature Dev', function () {
     let framework: qTestingFramework
     let tab: Messenger
 
-    const maxTestDuration = 600000
+    const maxTestDuration = 300000
     const prompt = 'Implement fibonacci in python'
     const iterateApproachPrompt = prompt + ' and add a unit test'
     const codegenApproachPrompt = prompt + ' and add a readme that describes the changes'
-    const tooManyRequestsWaitTime = 100000
+    const tooManyRequestsWaitTime = 600000
 
     before(async function () {
         /**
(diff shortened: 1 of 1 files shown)


packages/core/src/testE2E/amazonq/featureDev.test.ts
```diff
@@ -212,10 +208,6 @@ describe('Amazon Q Feature Dev', function () {
                 })
             })
 
-            describe('Clicks accept code', () => {
-                insertCodeTests()
```
sannicm, line 216: question: even if they are being removed is the flow still tested?

jpinkney-aws: No, I think this change also removes about half the tests

jpinkney-aws:
They be though, these tests create ~16 tests with ~16 different conversation id's that follow all the different flows. Everything operating in this block:
```
Moves directly from approach to codegen
  Clicks accept code
    - clicks new task
    - click close session
  Iterates on codegen
    Clicks accept code
      - clicks new task
      - click close session
```
is ~4 separate conversation ids (or at least should be). So its like:
- `/dev ${msg}` -> click generate code -> click new task
- `/dev ${msg}` -> click generate code -> click close session
- `/dev ${msg}` -> click generate code -> iterates -> click new task
- `/dev ${msg}` -> click generate code -> iterates -> click close session

Click close session is always a terminal node, we shouldn't be doing anything after that

packages/core/src/testE2E/amazonq/featureDev.test.ts
sannicm: Could we instead of removing, fix the tests somehow?

dianaabv16: I added them back as they started to pass once I increased  `waitIntervalInMs`, `waitTimeoutInMs` times

dianaabv16:
@jpinkney-aws makes sense now. 
I added tests back, but they still were failing with the error in the descriptions. Increasing interval time helped:
``
✔ click close session
  23 passing (1h)
globalSetup: after()
No test coverage found
````

File: packages/core/src/testE2E/amazonq/featureDev.test.ts [hunk 2]
```diff
@@ -19,11 +19,11 @@ describe('Amazon Q Feature Dev', function () {
     let framework: qTestingFramework
     let tab: Messenger
 
-    const maxTestDuration = 600000
+    const maxTestDuration = 900000
     const prompt = 'Implement fibonacci in python'
     const iterateApproachPrompt = prompt + ' and add a unit test'
     const codegenApproachPrompt = prompt + ' and add a readme that describes the changes'
-    const tooManyRequestsWaitTime = 100000
+    const tooManyRequestsWaitTime = 900000
```
justinmk3, line 26: so each test may wait up to 15 minutes if we get throttled?

PR #4735, opened 2024-04-18
Q Code Transformation - Reset state between sequential job runs 

arangatang:
## Problem
Customers who ran multiple jobs after each other without accepting / rejecting changes could not see proposed changes for the subsequent job.
(have videos with behaviour, but too large for github upload)

## Solution
Reset state when new job is submitted + streamline reset logic into one piece of code (previously scattered in two places)

### Testing

1. Ran three jobs in a row and verified the proposed changes could be downloaded each time.
(have videos with behaviour, but too large for github upload)

## License

### DIFF
--- a/packages/core/src/codewhisperer/commands/startTransformByQ.ts
+++ b/packages/core/src/codewhisperer/commands/startTransformByQ.ts
@@ -292,9 +292,7 @@ export async function getValidCandidateProjects(): Promise<TransformationCandida
 
 export async function setTransformationToRunningState() {
     await setContextVariables()
-    getLogger().info(`In startTransformByQ about to reset project`)
     await vscode.commands.executeCommand('aws.amazonq.transformationHub.reviewChanges.reset')
-    getLogger().info(`In startTransformByQ project reset`)
     transformByQState.setToRunning()
     sessionPlanProgress['startJob'] = StepProgress.Pending
     sessionPlanProgress['buildCode'] = StepProgress.Pending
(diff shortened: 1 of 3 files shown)


packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -54,10 +54,6 @@ import { sleep } from '../../shared/utilities/timeoutUtils'
 
 let sessionJobHistory: { timestamp: string; module: string; status: string; duration: string; id: string }[] = []
 
-export async function startTransformByQWithProgress() {
-    await startTransformByQ()
-}
-
```
arangatang, line 60: Did a cmd+shift+f and did not find any usages of this one.

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -261,9 +261,30 @@ export class ProposedTransformationExplorer {
             treeDataProvider: transformDataProvider,
         })
 
+        const reset = async () => {
+            await vscode.commands.executeCommand('setContext', 'gumby.transformationProposalReviewInProgress', false)
+            await vscode.commands.executeCommand('setContext', 'gumby.reviewState', TransformByQReviewStatus.NotStarted)
+
+            // delete result archive after changes cleared
+            // Summary is under ResultArchiveFilePath
```
dhasani23, line 269: nit: maybe add a comment saying that `force: true` prevents an error from being thrown if the file path does not exist, since this `reset()` is called at beginning of every job where those file paths won't always exist

dhasani23 (approved):

justinmk3 (approved):

PR #4736, opened 2024-04-18
chore(amazonq): remove command - Amazon Q (Preview) Welcome Page

ivikash:
## Problem

Remove Command - `Amazon Q (Preview) Welcome Page`

This is an intrusive change and removes onboarding page in favor of walkthrough #4739

## Solution
Screenshot 2024-04-18 at 11 27 43 AM

## License


packages/core/src/amazonq/activation.ts
```diff
@@ -53,14 +46,3 @@ function registerApps(appInitContext: AmazonQAppInitContext) {
     featureDevChatAppInit(appInitContext)
     gumbyChatAppInit(appInitContext)
 }
-
-export const amazonQWelcomeCommand = Commands.declare(
-    { id: 'aws.amazonq.welcome', compositeKey: { 1: 'source' } },
```
nkomonen-amazon, line 58: There is still a use of `aws.amazonq.welcome` in `authUtil.ts` I think. `ctrl + shift + f` to search globally in our codebase. The string is unique so any uses of it are probably related to this change.

The command stuff in VS Code is kind of confusing at first. But think of a "command" as a global api that our extension provides. VS Code provides an API to create a command `vscode.commands.registerCommand/executeCommand()`. But to get some extra functionality we wrap that vscode API with the `Commands` class. Some example uses are `Commands.declare()` or `Commands.register()`. So any uses of these are related in some way.

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -515,20 +491,6 @@ function buildFeatureAuthState(state: AuthState): FeatureAuthState {
     }
 }
 
-function getEnvType(): keyof HasAlreadySeenQWelcome {
```
justinmk3, line 518: For reference: the replacement is `getComputeEnvType()` : https://github.com/aws/aws-toolkit-vscode/blob/7ec63ca3607e5d307f7f6fe83a2a1a51878e844c/packages/core/src/shared/telemetry/util.ts#L113

packages/core/src/amazonq/activation.ts
```diff
@@ -53,14 +46,3 @@ function registerApps(appInitContext: AmazonQAppInitContext) {
...
-    (context: ExtensionContext, publisher: MessagePublisher<any>) => (_: VsCodeCommandArg, source: string) => {
-        telemetry.ui_click.run(() => {
-            void focusAmazonQPanel()
-            welcome(context, publisher)
```
justinmk3, line 62: The welcome() function is no longer needed. Should it be deleted?

Also, the the welcome page involved a bunch of resources (svg files, etc), generally everything in `src/amazonq/onboardingPage/` . So, is any of that still needed, and why?

PR #4739, opened 2024-04-18
amazonq: Add Walkthrough page

nkomonen-amazon:
## Problem
We want a walkthrough for Amazon Q. Also some separate changes.

## Other Changes

- Fix Amazon Q debug scripts. A `serve` task was failing constantly before this.
- Enable Dev mode in Amazon Q

### Walkthrough Changes

- Creates the Amazon Q walkthrough
- GIFs still need to be created, we are waiting on those but we can merge for now
- When the user logs in to Amazon Q the first time they will be shown the walkthrough page. It will never show again on subsequent logins
- We have telemetry for when the walkthrough is shown through our "first login" automation, and when certain events are triggered in the walkthrough such as opening the chat

Icon and GIFs will be added later:
Screenshot 2024-04-18 at 5 58 08 PM

## License


packages/amazonq/src/extensionShared.ts
```diff
@@ -61,6 +62,8 @@ export async function activateShared(context: vscode.ExtensionContext) {
     await activateCWChat(context)
     await activateQGumby(extContext as ExtContext)
 
+    await activateDev(context)
```
justinmk3, line 65: I think activateDev needs to be owned by Toolkit. won' this create duplicate commands?

packages/core/src/amazonq/onboardingPage/walkthrough.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export async function showAmazonQWalkthroughOnce(
+    state = globals.context.globalState,
+    showWalkthrough = () => openAmazonQWalkthrough.execute()
+) {
+    const hasShownWalkthroughId = 'aws.amazonq.hasShownWalkthrough'
+    const hasShownWalkthrough = state.get(hasShownWalkthroughId, false)
+    if (hasShownWalkthrough) {
+        return
+    }
```
justinmk3, line 23: Perhaps we need something like `PromptSettings` , but for globalState? Or can we use PromptSettings for this?

justinmk3 (approved):

packages/amazonq/src/extensionShared.ts
nkomonen-amazon: Good catch. activateDev has been removed from this PR, I'll properly implement it in a different one.

packages/core/src/amazonq/onboardingPage/walkthrough.ts
nkomonen-amazon: Yeah, I think we should create something similar for globalState. Theres probably a lot of duplicate stuff we can remove.

justinmk3 (approved):

PR #4751, opened 2024-04-19
fix: About extension message always shows AWS Toolkit

jpinkney-aws:
## Problem
- If you are using the about from the amazon q extension it shows AWS Toolkit. e.g. "AWS Toolkit: 0.10.0"

## Solution
- If you are using the q extension then show Amazon Q Toolkit

## License


packages/core/src/shared/extensionUtilities.ts
```diff
@@ -264,6 +265,12 @@ export async function aboutExtension(): Promise<void> {
     }
 }
 
+function getProductName(): string {
+    return globals.context.extension.id === VSCODE_EXTENSION_ID.amazonq
+        ? 'Amazon Q Toolkit'
```
hayemaxi, line 270:
```suggestion
        ? 'Amazon Q'
```

justinmk3: Oops I merged too fast :(

PR #4754, opened 2024-04-19
fix(globals): use 'local' globals if not running web mode

hayemaxi:
Problem: To get web tests working we used globalThis for globals. This is shared across extensions in VS Code, resulting in problems if another extension declares the same global twice. This is occurs when we run standalone and toolkit at the same time.

Solution: Use local globals if we are not in web mode. Otherwise, use full globals. This allows the web tests to pass. We will have to revisit this if we have 2 extensions running in web mode.

- Also add slightly more informative globals error handling.
- Refactor CW class so that globals doesn't get called during imports.

## License


packages/core/src/extensionShared.ts
```diff
@@ -59,13 +58,18 @@ let localize: nls.LocalizeFunc
  * Activation/setup code that is shared by the regular (nodejs) extension AND web mode extension.
  * Most setup code should live here, unless there is a reason not to.
  */
-export async function activateShared(context: vscode.ExtensionContext, contextPrefix: string): Promise<ExtContext> {
+export async function activateShared(
+    context: vscode.ExtensionContext,
+    contextPrefix: string,
```
nkomonen-amazon, line 63: In the future, if we have to add more parameters we'll want to create a single object with all these params instead.

nkomonen-amazon (approved):

packages/core/src/extensionShared.ts
hayemaxi: like an args: {} of sorts? Agreed

packages/core/src/shared/extensionGlobals.ts
```diff
@@ -96,39 +98,61 @@ function resolveGlobalsObject(): ToolkitGlobals {
 function proxyGlobals(globals_: ToolkitGlobals): ToolkitGlobals {
     return new Proxy(globals_, {
         get: (target, prop) => {
+            // Test for initialize()
+            if (Object.keys(target).length === 0) {
+                throw new Error(`ToolkitGlobals accessed before initialize()`)
+            }
+
+            // Test that the property was set before access.
+            // Tradeoff: not being able to do something like `globals.myValue ??= ...` without a try/catch
```
justinmk3, line 107: Good point, useful comment. But for globals, assigning with `??=` is probably a sign of buggy startup logic...

PR #4761, opened 2024-04-20
telemetry(amazonq chat): numberOfCodeBlocks 

laileni-aws:
## Problem
- The Q chat telemetry event `recordAddMessage` is missing the required parameter `cwsprChatResponseCodeSnippetCount` for AB testing.
## Solution
- Counting the three back ticks(```) to find out totalN number of code blocks in the generated message response.

## License


packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -88,6 +88,16 @@ export class Messenger {
             )
         )
     }
+
+    public countTotalNumberOfCodeBlocks(message: string): number {
```
dogusata, line 92: We've discussed several times that this method will not work properly especially in case there are code block markdowns inside code blocks. Please use `mynahUI.endMessageStream` instead.

dogusata: Send message to webview from IDE, grab the latest count by ending the stream on presentation layer with `mynahUI.endStream` and send it back to IDE level from webview.

dogusata (changes requested):
Please replace the code block calculation method with ui endstream method return.

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
laileni-aws: I tried this approach and not able to get the `noOfCountBlocks` from mynahUi and even with this we can not send this param to the Telemetry event `recordAddTelemetry` event as this will be sent by then. So the margin of error is same for both Claude V2 and V3. Discussed with Chandan and Vladimir regarding this and went with this approach.

dogusata: As a temporary solution (until backend response contains the number of code blocks with the last chunk) i'm approving it. But definitely needs to be updated asap.

dogusata (approved):

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -88,6 +89,21 @@ export class Messenger {
...
+    public async countTotalNumberOfCodeBlocks(message: string): Promise<number> {
+        if (message === undefined) {
+            return 0
+        }
+
+        // To Convert Markdown text to HTML using marked library
+        const html = await marked(message)
+
+        // Count the number of <pre> tags in the HTML to find the total number of code blocks
+        const totalNumberOfCodeBlocks = (html.match(/<pre>/g) || []).length
```
mitroviv, line 102: This is not specific enough - it could match unrelated `<pre>` tags.

ptrucks: My suggestion would be [jsdom](https://www.npmjs.com/package/jsdom) and the method [querySelectorAll](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll) with a query of `pre > code` to find all code blocks

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -88,6 +90,24 @@ export class Messenger {
...
+        }
+
+        // // To Convert Markdown text to HTML using marked library
+        const html = await marked(message)
+
+        const dom = new JSDOM(html)
+        const document = dom.window.document
+
+        // Search for <pre> elements containing <code> elements
+        const codeBlocks = document.querySelectorAll('pre > code')
```
mitroviv, line 106: Please test this with some messages that already have content that looks like a proper html snippet that contains `<pre>` and within it nested `<code>`. 

In other words you might need to be more specific here and possibly provide a bit longer chain of descendent (preceding `pre` tag) to make sure not to pick out some part of the message that would turn into a similar looking element. 
The `<code>` within `<pre>` comes from UI and if we use marked library separately, it is not clear how the things will get rendered and if you can have additional metadata (such as class/id names etc.) to increase the probability of positive identification.

Are there any unit tests for this?

dogusata (approved):

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -88,6 +90,24 @@ export class Messenger {
             )
         )
     }
+
+    public async countTotalNumberOfCodeBlocks(message: string): Promise<number> {
```
dogusata, line 94: This makes much more sense than counting triple backticks. However, for efficiency, we still need to get it directly from the response. But for a client implementation good work👍🏻

PR #4771, opened 2024-04-22
fix: User is able to create a dev environment inside of a dev environment

jpinkney-aws:
## Problem
- Currently a user is able to create a dev environment inside a dev environment

## Solution
- Throw an error is a user is trying to create a dev environment inside of a dev environment. This is because the token used is derived from whatever token initiated the "Start" API causing issues in a downstream service

## License

### DIFF
--- a/packages/core/src/codecatalyst/commands.ts
+++ b/packages/core/src/codecatalyst/commands.ts
@@ -205,6 +205,14 @@ type Inject<T, U> = T extends (...args: infer P) => infer R
 
 type WithClient<T> = Parameters<Inject<T, CodeCatalystClient>>
 
+class RemoteContextError extends ToolkitError {
+    constructor() {
+        super('Cannot connect from a remote context. Try again from a local VS Code instance.', {
+            code: 'ConnectedToRemote',
+        })
+    }
+}
+
 export class CodeCatalystCommands {
     public readonly withClient: ClientInjector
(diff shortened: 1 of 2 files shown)


packages/core/src/codecatalyst/commands.ts
```diff
@@ -222,6 +230,9 @@ export class CodeCatalystCommands {
     }
 
     public createDevEnv(): Promise<void> {
+        if (vscode.env.remoteName === 'ssh-remote') {
```
hayemaxi, line 233: will this prevent users from creating dev environments from remote machines that aren't dev environments?

justinmk3 (approved):

packages/core/src/codecatalyst/commands.ts
```diff
@@ -222,6 +231,9 @@ export class CodeCatalystCommands {
     }
 
     public createDevEnv(): Promise<void> {
+        if (vscode.env.remoteName === 'ssh-remote' && isInDevEnv()) {
```
nkomonen-amazon, line 234: Do we only need `isInDevEnv()`?

jpinkney-aws: I think we also want to disable remote environments as well since create tries to open a devEnv and that's not possible on remote envs (as far as I know)

PR #4780, opened 2024-04-23
fix(auth): delay PKCE server close

jpinkney-aws:
## Problem
- Right now we are relying on the auth server page load time to take less than the create token request. In the case where the auth server page load takes more time then the create token request the auth server will already be closed and won't be able to serve the ui

**note:** in practice this should actually never happen. This is just extra insurance that if it _does_ happen we are protected

## Solution
- delay for 5000ms

## License

### DIFF
--- a/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
+++ b/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
@@ -449,7 +449,11 @@ class AuthFlowAuthorization extends SsoAccessTokenProvider {
 
             return this.formatToken(token, registration)
         } finally {
-            await authServer.close()
+            // Temporary delay to make sure the auth ui was displayed to the user before closing
+            // inspired by https://github.com/microsoft/vscode/blob/a49c81edea6647684eee87d204e50feed9c455f6/extensions/github-authentication/src/flows.ts#L262
+            setTimeout(() => {
+                void authServer.close()
+            }, 5000)
         }
     }
 


packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -449,7 +449,11 @@ class AuthFlowAuthorization extends SsoAccessTokenProvider {
 
             return this.formatToken(token, registration)
         } finally {
-            await authServer.close()
+            // Temporary delay to make sure the auth ui was displayed to the user before closing
+            // inspired by https://github.com/microsoft/vscode/blob/a49c81edea6647684eee87d204e50feed9c455f6/extensions/github-authentication/src/flows.ts#L262
+            setTimeout(() => {
```
hayemaxi, line 454: is there a way to use callbacks instead of timed delays?

jpinkney-aws: I originally wanted to try for that but I don't think it's possible. The auth server has no context of whats the final resource it has to serve before its finished and as far as I see we don't have a way for the the javascript running in the browser to tell the toolkit to continue on (unless we explicitly force them to click the vscode:// link but I don't think thats a good idea)

FWIW I don't really expect this piece of code to ever be hit I just added it in because I saw the github authentication plugin had it. I've never run into this problem in the 50+ times I've tested the flow

justinmk3 (approved):

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -449,7 +449,11 @@ class AuthFlowAuthorization extends SsoAccessTokenProvider {
...
+                void authServer.close()
```
justinmk3, line 455: If this fails, it won't be logged anywhere. Usually good to at least log failures, especially in timers which otherwise may cause mysterious behavior.

PR #4781, opened 2024-04-23
fix: copy auth page to amazon q extension

jpinkney-aws:
## Problem
- Amazon q extension doesn't have the resources for the auth page

## Solution
- Copy them in

## License

### DIFF
--- a/packages/amazonq/scripts/build/copyFiles.ts
+++ b/packages/amazonq/scripts/build/copyFiles.ts
@@ -32,6 +32,10 @@ const tasks: CopyTask[] = [
 
     { target: path.join('../core', 'resources'), destination: path.join('..', 'resources') },
     { target: path.join('../core', 'package.nls.json'), destination: path.join('..', 'package.nls.json') },
+    {
+        target: '../core/src/auth/sso/vue',
+        destination: 'src/auth/sso/vue',
+    },
 
     // Vue
     {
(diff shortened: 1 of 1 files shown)


packages/amazonq/scripts/build/copyFiles.ts
```diff
@@ -32,6 +32,10 @@ const tasks: CopyTask[] = [
 
     { target: path.join('../core', 'resources'), destination: path.join('..', 'resources') },
     { target: path.join('../core', 'package.nls.json'), destination: path.join('..', 'package.nls.json') },
+    {
+        target: '../core/src/auth/sso/vue',
+        destination: 'src/auth/sso/vue',
+    },
 
     // Vue
     {
```
hayemaxi, line 41: nit:
```suggestion
    { target: path.join('../core', 'package.nls.json'), destination: path.join('..', 'package.nls.json') },

    // Vue
    {
        target: '../core/src/auth/sso/vue',
        destination: 'src/auth/sso/vue',
    },
    {
```

hayemaxi (approved):
qq: why isn't this copied during the vue copy operation on line 51?

jpinkney-aws:
It's because the code in that folder isn't actually vue, its html (we just kept the folder structure that way for standards). When the project is compiled only the actual .vue files are compiled and added to /dist

PR #4783, opened 2024-04-23
feat(amazonq): reword settings + import old CW settings

hayemaxi:
Problem: Some of the wording of the settings was off. Also, changing the name of the settings = a brand new setting in VS Code.

Solution: In order to preserve trust, we will try to import old settings users may have had. Extensions leave settings behind when they are uinstalled, so we can detect the old CW settings and import them into the new ones named after Amazon Q. This creates a seamless setting experience for migrating users.

## License


packages/core/src/codewhisperer/activation.ts
```diff
@@ -67,6 +67,10 @@ import { switchToAmazonQCommand, switchToAmazonQSignInCommand } from '../amazonq
 
 export async function activate(context: ExtContext): Promise<void> {
     const codewhispererSettings = CodeWhispererSettings.instance
+
+    // Import old CodeWhisperer settings into Amazon Q
```
leigaol, line 71: Please also update code in activation of codewhisperer


```


            if (configurationChangeEvent.affectsConfiguration('aws.amazonQ.shareCodeWhispererContentWithAWS')) {
                if (auth.isEnterpriseSsoInUse()) {
```

packages/core/package.json
```diff
@@ -261,19 +261,19 @@
                     },
                     "additionalProperties": false
                 },
-                "aws.amazonQ.includeSuggestionsWithCodeReferences": {
+                "aws.amazonQ.showInlineCodeSuggestionsWithCodeReferences": {
```
jpinkney-aws, line 264: Out of curiosity why are we still declaring these in core when they are only relevant to q now? Is it for tests?

hayemaxi: [https://github.com/aws/aws-toolkit-vscode/blob/master/docs/arch_develop.md?plain=1#L60
and
](https://github.com/aws/aws-toolkit-vscode/blob/feature/standalone/docs/arch_develop.md#contributes-and-settings)

packages/core/src/shared/settings.ts
```diff
@@ -898,3 +898,47 @@ export async function migrateSetting<T, U = T>(
...
+
+export type DeprecatedSetting = (typeof deprecatedSettings)[number]
+
+/**
+ * Import the setting value of an old setting into a new setting, if the old setting exists.
+ * Useful for setting renames.
+ *
+ * newKey must be defined in the extension, e.g. keys of {@link settingsProps}.
+ */
+export async function tryImportSetting(oldKey: DeprecatedSetting, newKey: keyof SettingsProps) {
```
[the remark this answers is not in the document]
hayemaxi: Hmm interesting. It looks like yes, except it only handles global/application level settings (we need both for this), and its typing is a little looser.

justinmk3 (approved):

PR #4796, opened 2024-04-24
fix(lint): ignore codewhisperer setting name too long

hayemaxi:
## License


packages/core/src/codewhisperer/util/codewhispererSettings.ts
```diff
@@ -8,7 +8,7 @@ import { codewhispererSettingsImportedKey } from '../models/constants'
 
 const description = {
     showInlineCodeSuggestionsWithCodeReferences: Boolean, // eslint-disable-line id-length
-    importRecommendationForInlineCodeSuggestions: Boolean,
+    importRecommendationForInlineCodeSuggestions: Boolean, // eslint-disable-line id-length
```
justinmk3, line 11: There is no reason this name needs to be so long. But we can ignore it for now, to unblock CI.

PR #4798, opened 2024-04-24
feat(dev): enable dev mode for multiple extensions

hayemaxi:
wip

todo: 'aws-dev' scheme is duplicate and causes error

## License

### DIFF
--- a/packages/amazonq/package.json
+++ b/packages/amazonq/package.json
@@ -259,7 +259,7 @@
                     "when": "false"
                 },
                 {
-                    "command": "aws.dev.openMenu",
+                    "command": "aws.amazonq.dev.openMenu",
                     "when": "aws.isDevMode || isCloud9"
                 },
                 {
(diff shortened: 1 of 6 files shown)


justinmk3:
Can we instead decide that Toolkit "owns" dev-mode?

hayemaxi:
> Can we instead decide that Toolkit "owns" dev-mode? Is there a reason that Q needs its own dev-mode, or can the developer just install Toolkit?

Each extension has individual states, e.g. global state, auth connections, etc. Even though the auth states are tied with the standalone work, there may be some unexpected behavior with expiring/deleting connections this way. It seemed easier to just set up 2 individual menus rather than try to create communication from Q -> toolkit for the dev menu.

I will take a look again at this approach. Maybe expiring/deleting states this way is a good way to catch if the auth communication is not working properly.

justinmk3:
> Each extension has individual global states.

True. If that's the only such feature, then serializing global state should be easy. Not sure about "editing" the state though.

If there are more features like that, then yeah, maybe it doesn't make sense.

hayemaxi:
closed in favor of https://github.com/aws/aws-toolkit-vscode/pull/4839

PR #4800, opened 2024-04-24
feat(amazonq): Update quick action styles and grouping

dogusata:
## Problem
- As requested from UX team quick actions list should not have a title anymore and needs a separator between `/transform, /dev` commands and the `/help, /clear` commands.
- Webview content is not fitting to its parent panel
## Solution
- Updated quick action commands group look&feel from mynah-ui with version update to **[`v4.6.4`](https://github.com/aws/mynah-ui/releases/tag/v4.6.4)**
- Updated quick action commands list. Removed the `groupName` from /dev group. `/transform` and `/dev` combined in one group.
- Applied necessary body and html width and height styles to the webview's content which were not coming from mynah-ui. 

Updated Quick action commands view:
image

## License


packages/core/src/amazonq/webview/ui/styles.css
```diff
@@ -0,0 +1,10 @@
+body,
```
jpinkney-aws, line 1: I think we generally store css files in the `resources/css` folder

jpinkney-aws (approved):

viktorsaws (approved):

PR #4805, opened 2024-04-24
feat(amazonq): plan explainability support

dhasani23:
## Problem

Our `GetPlan` API is being enriched with more data about the transformation job and this needs to be displayed to the user.

## Solution

Display the additional data with icons and tables in the existing markdown file that we create for the transformation plan.

image
image

## License


packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -385,40 +385,182 @@ export function getImageAsBase64(filePath: string) {
...
+        // }
+        // telemetry.codeTransform_logApiLatency.emit({
+        //     codeTransformApiNames: 'GetTransformationPlan',
+        //     codeTransformSessionId: codeTransformTelemetryState.getSessionId(),
+        //     codeTransformJobId: jobId,
+        //     codeTransformRunTimeLatency: calculateTotalLatency(apiStartTime),
+        //     codeTransformRequestId: response.$response.requestId,
+        //     result: MetadataResult.Pass,
+        // })
+
```
dhasani23, line 448: Using a mocked response until backend work is done.

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -488,6 +636,7 @@ export async function pollTransformationJob(jobId: string, validStates: string[]
                 result: MetadataResult.Pass,
             })
             status = response.transformationJob.status!
+            console.log('status = ' + status)
```
nkomonen-amazon, line 639: Is temporary for debugging? If not use getLogger()

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -385,40 +385,182 @@ export function getImageAsBase64(filePath: string) {
     return `data:image/svg+xml;base64,${fileContents}`
 }
 
+/*
+ * Given the icon name from core/resources/icons/aws/amazonq, get the appropriate icon according to the user's theme.
+ * ex. getIcon('transform-file') returns the 'transform-file-light.svg' icon if user has a light theme enabled,
+ * otherwise 'transform-file-dark.svg' is returned.
+ */
+export function getIcon(iconName: string) {
```
nkomonen-amazon, line 393: Maybe rename this to `getAmazonQIcon()` since we already have `getIcon()`

nkomonen-amazon (approved):

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
dhasani23: Yes just temporary

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
justinmk3: why doesn't this use the existing getIcon util ?

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -385,40 +385,182 @@ export function getImageAsBase64(filePath: string) {
...
+    let iconPath = iconName
+    const themeColor = vscode.window.activeColorTheme.kind
+    if (themeColor === vscode.ColorThemeKind.Light || themeColor === vscode.ColorThemeKind.HighContrastLight) {
+        iconPath += '-light.svg'
+    } else {
+        iconPath += '-dark.svg'
+    }
+    return getImageAsBase64(globals.context.asAbsolutePath(path.join('resources', 'icons', 'aws', 'amazonq', iconPath)))
```
justinmk3, line 401: literals never need to be joined
```suggestion
    return getImageAsBase64(globals.context.asAbsolutePath(path.join('resources/icons/aws/amazonq', iconPath)))
```

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
justinmk3: > so are light/dark versions of the icon somehow generated for you and if so how?

yes. explained in:

- https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#font-generation
- https://github.com/aws/aws-toolkit-vscode/blob/master/docs/icons.md , which is linked from

dhasani23: Just spent some time looking into this. For some reason when I add my icons under `resources/icons/aws/amazonq/dark/` and `resources/icons/aws/amazonq/light/` (following instructions here: https://github.com/aws/aws-toolkit-vscode/blob/master/docs/icons.md#theme-overrides) then run `npm run generatePackage` the `package.json` is not updated with the icons I just added. Some changes appear to be made but my new icons are not added. When I move the `dark/` and `light/` folders to be directly under `aws/` that also does not work. I have to add the icons directly under the existing `amazonq/` folder for `npm run generatePackage` to update the `package.json` properly. I also noticed that other folders (`applicationcomposer/`, `codewhisperer/`, and `mynah/`) simply put a dark and light version of the icon directly in their folder without a `dark/` or `light/` sub-folder, which matches my original implementation in this PR.

Not sure what I'm doing wrong. I would like to follow up on this but that will have to be post-GA because this PR likely needs to be merged by EOD tomorrow and I still have a couple of other small changes to make, so I'll make a SIM for this.

dhasani23: SIM ID is EGCP-130

dhasani23:
Don't merge yet; this needs some more work

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -385,12 +385,135 @@ export function getImageAsBase64(filePath: string) {
...
+    switch (s) {
+        case 'linesOfCode':
+            return CodeWhispererConstants.linesOfCodeMessage
+        case 'plannedDependencyChanges':
+            return CodeWhispererConstants.dependenciesToBeReplacedMessage
+        case 'plannedDeprecatedApiChanges':
+            return CodeWhispererConstants.deprecatedCodeToBeReplacedMessage
+        case 'plannedFileChanges':
+            return CodeWhispererConstants.filesToBeChangedMessage
+        case 'dependencyName':
```
justinmk3, line 436: not a blocker (future followup): can this be modeled as a map? instead of a bunch of loose constants mapped by a switch.

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -402,21 +525,37 @@ export async function getTransformationPlan(jobId: string) {
...
+        const tableMapping = getTableMapping(stepZeroProgressUpdates)
+
+        const jobStatistics = JSON.parse(tableMapping['0']).rows // ID of '0' reserved for job statistics table
+
+        // get logo directly since we only use one logo regardless of color theme
+        const logoIcon = getTransformationIcon('transformLogo')
+
+        const arrowIcon = getTransformationIcon('upArrow')
+
+        let plan = `<style>table {border: 1px solid #424750;}</style>\n\n<a id="top"></a><br><p style="font-size: 24px; color: white;"><img src="${logoIcon}" style="margin-right: 15px; vertical-align: middle;"></img><b>${CodeWhispererConstants.planTitle}</b></p><br>`
```
damntrecky, line 546: NOT a blocker: For better readability, I Would like to follow up with a more string literal/template approach instead of `plan +=`.

damntrecky (approved):

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -237,17 +237,20 @@ export async function pollTransformationStatusUntilPlanReady(jobId: string) {
...
-    transformByQState.setPlanFilePath(planFilePath)
-    await vscode.commands.executeCommand('setContext', 'gumby.isPlanAvailable', true)
+    if (plan !== undefined) {
+        const planFilePath = path.join(transformByQState.getProjectPath(), 'transformation-plan.md')
+        fs.writeFileSync(planFilePath, plan)
+        await vscode.commands.executeCommand('markdown.showPreview', vscode.Uri.file(planFilePath))
+        transformByQState.setPlanFilePath(planFilePath)
+        await vscode.commands.executeCommand('setContext', 'gumby.isPlanAvailable', true)
+    }
     sessionPlanProgress['generatePlan'] = StepProgress.Succeeded
```
JuliaABurch, line 254: Should this be moved into the `if` block above? If `plan` is undefined, this step didn't succeed

dhasani23: No this is fine. We need to set it to succeeded so that the Transformation Hub steps continue to render properly, and technically the plan did "generate" properly, it just didn't get displayed to user due to some parsing issue. As a followup, we should technically show a yellow icon instead of a green checkmark when this happens, but that would be the only change to make.

Issue #4758, opened 2024-04-20
CodeCatalyst dev environment sessions timing out


sahilrc:
## Problem

With timeout in CodeCatalyst dev environments set to 15 minutes, I'm kicked out of the SSH session after 15 minutes of use, while still having VSCode active on my machine. 

## Steps to reproduce the issue

1. Connect to a CodeCatalyst dev environment with timeout set
2. Wait till specified timeout
3. VSCode states: "Reconnecting to SSH: aws-devenv"

## Expected behavior

Not kicked out.

I was able to workaround this by specifying "No timeout" in the CodeCatalyst console.

## System details 

-   OS: Mac Sonoma 14.3.1
-   Visual Studio Code version: 1.87.2
-   AWS Toolkit version: v2.20.0

mrpackethead:
Its 'SSH session' rather than RDP, session. 

I'm getting the same thing.  THe development system will shut down after its timeout period irrispective of activity or not. 

THis issue only started mid this week.  IT was working just fine.



Version: 1.88.1 (user setup)
Commit: e170252f762678dec6ca2cc69aba1570769a5d39
Date: 2024-04-10T17:41:02.734Z
Electron: 28.2.8
ElectronBuildId: 27744544
Chromium: 120.0.6099.291
Node.js: 18.18.2
V8: 12.0.267.19-electron.0
OS: Windows_NT x64 10.0.22621

AWS Toolkit v2.20.0

boomclickthump:
I can confirm that reverting to 2.18.0 resovles the issue as well.

mrpackethead:
Looks like aws are making a fix

https://github.com/aws/aws-toolkit-vscode/pull/4811

justinmk3:
Fixed in the latest AWS Toolkit.

mrpackethead:
@justinmk3 , I've installed and will try it out and come back to you and let you know after using it for a day.

PR #4811, opened 2024-04-24
codecatalyst: fix extension activation in new dev environments

jpinkney-aws:
## Problem
- The extension fails to activate in new dev environments: `TypeError: Cannot read properties of undefined (reading 'scopeExpired')
	at pe.isScopeExpired`
- The issue also cascades to existing dev environment where things like activity heartbeat fails to activate because thisDevenv on codecatalyst/activation #L80 ends up being undefined

## Solution
- Provide a default for connection states if they aren't found. The reason this is neccessary is on #L96 in codecatalyst/auth.ts we are making a call to this.isConnectionValid(), which then checks if the scopes are expired. This happens BEFORE the scopesExpired is available in the global state for 'codecatalyst.connections' causing the issue

## License

### DIFF
--- a/.changes/next-release/Bug Fix-5d2dc465-bf49-4808-94a1-bf8006af3b96.json
+++ b/.changes/next-release/Bug Fix-5d2dc465-bf49-4808-94a1-bf8006af3b96.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "codecatalyst: fix extension activation in new dev environments"
+}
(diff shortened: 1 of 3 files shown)


mrpackethead:
this looks like it will fix this issue 
https://github.com/aws/aws-toolkit-vscode/issues/4758

hayemaxi (approved):

aarolieb (approved):

justinmk3 (approved):
🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀

PR #4812, opened 2024-04-25
fix(amazonq): Various css updates to login page

leigaol:
## Problem
VSCode CSS issues for new auth page:

1. Q icon should be 71px, not 100px
2. First page: Looks like the width of buttons is wrapping to fit the contents - can we set the width of the sign in buttons and continue button 260px?
3. First page: Each of the elements should have a margin of 12px, buttons should have a padding of 12px
4. SSO/IAM Creds page: Back button is center aligned now for some reason
5. SSO/IAM Creds page: Padding and size of input fields are off (width: 244px; padding: 6px 8px;)
6. SSO/IAM Creds page: Margin should be 12px between groups
7. SSO/IAM Creds page: Field labels (ex. 'Start URL') should be 11px, with a weight of 500
8. In progress page: 12px margin between elements
9. In progress page: Cancel button should be secondary grey color
10. change to 'Sign in with AWS IAM Identity Center'
11. Styling of the region dropdown input text should match mocks
12. Should have max width for the input boxes (see mocks and JB implementation)

## License

### DIFF
--- a/packages/core/src/login/webview/vue/login.vue
+++ b/packages/core/src/login/webview/vue/login.vue
@@ -521,6 +521,7 @@ export default defineComponent({
 
     /* Stretches our overall container to the whole screen */
     height: 100%;
+    width: 100%;
     /* Centers all content in to middle of page since the height is the whole screen*/
     justify-content: center;
 }
--- a/packages/core/src/login/webview/vue/login.vue
+++ b/packages/core/src/login/webview/vue/login.vue
@@ -72,7 +72,7 @@
             </svg>
         </div>
         <template v-if="stage === 'START'">
-            <button class="back-button" v-if="app === 'TOOLKIT'" @click="handleBackButtonClick">
+            <button class="back-button bottomMargin" v-if="app === 'TOOLKIT'" @click="handleBackButtonClick">
                 <svg width="13" height="11" viewBox="0 0 13 11" fill="none" xmlns="http://www.w3.org/2000/svg">
                     <path
                         d="M4.98667 0.0933332L5.73333 0.786666L1.57333 4.94667H12.0267V5.96H1.57333L5.73333 10.0667L4.98667 10.8133L0.0266666 5.8V5.10667L4.98667 0.0933332Z"
@@ -151,7 +151,7 @@
                 </svg>
             </button>
             <div class="auth-container-section">
-                <div class="header">Sign in with SSO:</div>
+                <div class="header">Sign in with AWS IAM Identity Center:</div>
                 <div class="code-catalyst-login" v-if="app === 'TOOLKIT'">
                     <div style="margin-bottom: 4px"></div>
                     <div class="h4">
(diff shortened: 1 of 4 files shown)


packages/core/src/login/webview/vue/login.vue
```diff
@@ -71,20 +71,20 @@
...
+                    <path
+                        d="M4.98667 0.0933332L5.73333 0.786666L1.57333 4.94667H12.0267V5.96H1.57333L5.73333 10.0667L4.98667 10.8133L0.0266666 5.8V5.10667L4.98667 0.0933332Z"
+                        fill="#21A2FF"
+                    />
+                </svg>
+            </button>
             <div class="auth-container-section">
                 <div class="existing-logins" v-if="existingLogins.length > 0">
                     <div class="header">Connect with an existing account:</div>
+                    <div style="margin-bottom: 12px"></div>
```
nkomonen-amazon, line 87: Make a class instead named something like `bottomMargin` and set this class to the element you care about. Then set the margin in that class and if you need to update, it will apply to all, like a variable

packages/core/src/login/webview/vue/login.vue
```diff
@@ -94,10 +94,12 @@
                             :itemTitle="existingLogin.title"
                             class="selectable-item"
                         ></SelectableItem>
+                        <div style="margin-bottom: 12px"></div>
```
nkomonen-amazon, line 97: I don't think you need a new element, just apply the class I mentioned in my above comment to the element itself. `SelectableItem` in this case:

```
class="selectable-item bottomMargin"
```

packages/core/src/login/webview/vue/login.vue
```diff
@@ -6,8 +6,8 @@
 
             <svg
                 v-if="app === 'AMAZONQ' && stage !== 'CONNECTED'"
-                width="100"
```
nkomonen-amazon, line 9: Does the logo in `reauthenticate.vue` also need to be updated then?

packages/core/src/login/webview/vue/login.vue
leigaol: done

packages/core/src/login/webview/vue/login.vue
leigaol: done

packages/core/src/login/webview/vue/login.vue
```diff
@@ -581,44 +593,59 @@ export default defineComponent({
...
-}
-.vscode-light .regionSelect {
+    padding-left: 8px;
+    padding-right: 8px;
+    padding-top: 6px;
+    padding-bottom: 6px;
+    font-size: 13px;
+    font-weight: 400;
+}
+body.vscode-light .regionSelect {
```
nkomonen-amazon, line 644: Not a blocker/relevant to this PR. But our root element has a **class** `body`, but then there is the actual `body` element which this is targeting. If you have time later we should probably rename the class `body` to something else like `appRoot` so that there isn't confusion

nkomonen-amazon (approved):

PR #4818, opened 2024-04-25
fix(q): Handle monthly conversation limit error for Q Feature development

kumsmrit:
## Problem
Currently, on reaching monthly conversation limit for QFeature development, Retry button was incorrectly shown which would degrade customer experience as clicking on Retry would render them the same error again and again.

## Solution
- Handled ServiceQuotaExceeded error correctly to not show Retry button on reaching monthly conversation limit.

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
+++ b/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
@@ -14,6 +14,7 @@ import { featureDevScheme } from '../../constants'
 import {
     CodeIterationLimitError,
     ContentLengthError,
+    MonthlyConversationLimitError,
     PlanIterationLimitError,
     SelectedFolderNotInWorkspaceFolderError,
     createUserFacingErrorMessage,
@@ -242,6 +243,8 @@ export class FeatureDevController {
                         },
                     ],
                 })
+            } else if (err instanceof MonthlyConversationLimitError) {
+                this.messenger.sendErrorMessage(err.message, message.tabID, 0, undefined, true)
             } else if (err instanceof PlanIterationLimitError) {
                 this.messenger.sendErrorMessage(err.message, message.tabID, this.retriesRemaining(session))
                 this.messenger.sendAnswer({
--- a/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
+++ b/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
@@ -244,7 +244,7 @@ export class FeatureDevController {
                     ],
                 })
             } else if (err instanceof MonthlyConversationLimitError) {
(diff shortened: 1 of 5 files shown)


packages/core/src/amazonqFeatureDev/controllers/chat/messenger/messenger.ts
```diff
@@ -46,7 +46,20 @@ export class Messenger {
...
 
-    public sendErrorMessage(errorMessage: string, tabID: string, retries: number, phase?: SessionStatePhase) {
+    public sendErrorMessage(
+        errorMessage: string,
+        tabID: string,
+        retries: number,
+        phase?: SessionStatePhase,
+        monthlyLimitError: boolean = false
+    ) {
+        if (monthlyLimitError) {
```
jpinkney-aws, line 56: Instead of adding another code path to this function I would probably create a separate function that just handles taking in a string (+ other arguments as required) and then sending the error message itself. That way we don't have to complicate this function signature more and it reduces potential code paths

packages/core/src/amazonqFeatureDev/controllers/chat/messenger/messenger.ts
```diff
@@ -46,7 +46,20 @@ export class Messenger {
...
+            this.dispatcher.sendErrorMessage(
+                new ErrorMessage(`Sorry, we encountered a problem when processing your request.`, errorMessage, tabID)
```
jpinkney-aws, line 58: I feel like we can make this error a bit more descriptive, at this point we know that they've hit a monthly limit error so we can probably notify them that they can't continue

packages/core/src/amazonqFeatureDev/controllers/chat/messenger/messenger.ts
kumsmrit: As discussed offline, creating separate method explicitly for rendering monthlyQuota limit error; we can later update method name if required to handle for other use-cases.

packages/core/src/amazonqFeatureDev/controllers/chat/messenger/messenger.ts
kumsmrit: Updated error message based on inputs from UX.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -242,6 +243,9 @@ export class FeatureDevController {
                         },
                     ],
                 })
+            } else if (err instanceof MonthlyConversationLimitError) {
+                this.messenger.sendMonthlyLimitError(message.tabID)
+                this.messenger.sendChatInputEnabled(message.tabID, false)
```
jpinkney-aws, line 248: Doesn't this happen anyways on line 290?

jpinkney-aws (approved):

PR #4822, opened 2024-04-25
fix Amazon Q import

nkomonen-amazon:
Import was directly from the `core` module, but needed to first be exported by it, then consumed as a `core` module export.

## License

### DIFF
--- a/packages/amazonq/src/extensionShared.ts
+++ b/packages/amazonq/src/extensionShared.ts
@@ -32,11 +32,10 @@ import { initializeAuth, CredentialsStore, LoginManager, AuthUtils } from 'aws-c
 import { makeEndpointsProvider, registerCommands } from 'aws-core-vscode'
 import { activate as activateCWChat } from 'aws-core-vscode/amazonq'
 import { activate as activateQGumby } from 'aws-core-vscode/amazonqGumby'
-import { CommonAuthViewProvider } from 'aws-core-vscode/login'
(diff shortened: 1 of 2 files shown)


justinmk3 (approved):

packages/amazonq/src/extensionShared.ts
```diff
@@ -28,15 +28,14 @@ import {
     getLogger,
     getMachineId,
 } from 'aws-core-vscode/shared'
-import { initializeAuth, CredentialsStore, LoginManager, AuthUtils } from 'aws-core-vscode/auth'
+import { initializeAuth, CredentialsStore, LoginManager, AuthUtils, CommonAuthWebview } from 'aws-core-vscode/auth'
 import { makeEndpointsProvider, registerCommands } from 'aws-core-vscode'
 import { activate as activateCWChat } from 'aws-core-vscode/amazonq'
 import { activate as activateQGumby } from 'aws-core-vscode/amazonqGumby'
 import { CommonAuthViewProvider } from 'aws-core-vscode/login'
```
hayemaxi, line 35: there is a login export already, we should probably move it to /auth or move CommonAuthWebview to /login

hayemaxi (approved):

packages/amazonq/src/extensionShared.ts
nkomonen-amazon: Good call, updated

PR #4827, opened 2024-04-25
refactor: unify "show Amazon Q panel" function

nkomonen-amazon:
### Problem:

We had different areas using different implementations to show
the Amazon Q panel. This is bad since there is dupliction and
eventually drift.

### Solution:

Centralize all focusing of Amazon Q in to a single Command, this
also gives us telemetry.

### Additional Change:
See each commit for an atomic change

## License


justinmk3 (approved):
This looks great! There is a failing test though

PR #4833, opened 2024-04-26
feat(telemetry): improve login page telemetry

hayemaxi:
- refactor to allow for better telemetry state handling
- reauthenticate page keeps more informative state + more informative cancellation
- cancellation keeps better state about current auth status
- properly check scopes for what features are enabled
- add telemetry for using an existing connection (not autoconnect)
- fixed bug that prevented logging in with builder id emitting telemetry

todo:
- toolkit needs auth_userState emitted
- auto-connect for Q from toolkit needs to be reworked, its not captured in the initial auth_userState because it occurs on UI side

## License


justinmk3:
> * reauthenticate page keeps more informative state + more informative cancellation
> * cancellation keeps better state about current auth status

Nice. I wonder if that can be abstracted in the future, so that we can have some analogous to `telemetry.xx.run()` which works for forms/webviews (which don't have an "execution context").

justinmk3 (approved):

PR #4843, opened 2024-04-26
fix(amazonq): Switch UI when CC reused Q connection

leigaol:
## Problem
`IsOnboarded` status of CC connection is not checked when CC reuses Q connection

## Solution

Force Toolkit check `IsOnboarded` status of CC connection. 

This is a fix that has minimal blast radius. It only affects the flow when CC reuses Q connection. However, the connection state management of CC should be able to intelligently forcing the `IsOnboarded` status update using the CC API when necessary and do not perform the update when not necessary ( like UI updates). 

https://github.com/aws/aws-toolkit-vscode/assets/97199248/234c44ca-95be-473b-9959-79b9322b29cf

## License

### DIFF
--- a/packages/core/src/codecatalyst/auth.ts
+++ b/packages/core/src/codecatalyst/auth.ts
@@ -379,10 +379,12 @@ export class CodeCatalystAuthenticationProvider {
         if (isValidCodeCatalystConnection(conn)) {
             getLogger().info(`auth: re-use connection from existing connection id ${connId}`)
             await this.secondaryAuth.useNewConnection(conn)
+            await this.isConnectionOnboarded(conn, true)
         } else {
             getLogger().info(`auth: re-use(new scope) to connection from existing connection id ${connId}`)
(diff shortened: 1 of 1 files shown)


nkomonen-amazon (approved):

packages/core/src/codecatalyst/auth.ts
```diff
@@ -379,10 +379,12 @@ export class CodeCatalystAuthenticationProvider {
         if (isValidCodeCatalystConnection(conn)) {
             getLogger().info(`auth: re-use connection from existing connection id ${connId}`)
             await this.secondaryAuth.useNewConnection(conn)
+            await this.isConnectionOnboarded(conn, true)
```
justinmk3, line 382: It seems non-obvious that `tryUseConnection` is doing this? Is there a 1-line comment that can give a hint . Also for the one below. 

```suggestion
            // HACK: 
            await this.isConnectionOnboarded(conn, true)
```

PR #4846, opened 2024-04-26
telemetry: Amazon Q indicate if it can auto connect

nkomonen-amazon:
## Problem

In the telemetry event auth_userState we need to indicate
our authStatus field as 'connected' if the Amazon Q extension does
not have a connection, but we detect one exists already in the
AWS Toolkit extension.

## Solution

This commit adds extra logic which checks if an autoconnectable
connection can be found in AWS Toolkit and then indicates that we
are connected. This assumes that the separate Login auto-connect logic
will eventually do the auto connection.

## License


packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -429,6 +429,24 @@ export class AuthUtil {
...
         return state
     }
+
+    /**
+     * Returns a connection that will work with Amazon Q from the given connections.
+     *
+     * TODO: This method seems like it can be generalized inside the Q AuthUtil. We should eventually
+     * look to add this in to there if possible so that Auth operations are centralized.
+     */
+    findUsableQConnection(connections: AwsConnection[]): AwsConnection | undefined {
```
nkomonen-amazon, line 439: This was moved from `backend_amazonq.ts` since we needed it to be shareable outside of just the UI

packages/amazonq/src/extensionShared.ts
```diff
@@ -140,21 +142,72 @@ export async function activateShared(context: vscode.ExtensionContext, isWeb: bo
...
+/**
+ * Returns a connection from the standalone toolkit extension that
+ * the Amazon Q extension can use. Otherwise it returns undefined.
+ *
+ * HACK: Our frontend Login ui for Amazon Q will auto connect if required/possible,
+ * but we cannot detect this at the end of Amazon Q extension activation.
+ * So we reuse the same {@link findUsableQConnection}()
+ * and assume that the frontend will have the same result and auto connect.
+ */
+async function getAutoConnectableConnection(): Promise<AwsConnection | undefined> {
```
justinmk3, line 194: should this and getAuthStatus live in a new `packages/amazonq/src/auth/util.ts` file?

justinmk3 (approved):

packages/amazonq/src/auth/util.ts
```diff
@@ -0,0 +1,61 @@
...
+import { activateExtension, getLogger } from 'aws-core-vscode/shared'
+import { VSCODE_EXTENSION_ID } from 'aws-core-vscode/utils'
+
+/** Provides the status of the Auth connection for Amazon Q, specifically for telemetry purposes. */
+export async function getAuthStatus(): Promise<AuthStatus> {
+    // Get auth state from the Amazon Q extension
+    const authState = (await AuthUtil.instance.getChatAuthState()).codewhispererChat
+    let authStatus: AuthStatus = authState === 'connected' || authState === 'expired' ? authState : 'notConnected'
+
+    // If the Q extension does not have its own extension, it will fallback and check
```
hayemaxi, line 18:
```suggestion
    // If the Q extension does not have its own connection, it will fallback and check
```

packages/amazonq/src/auth/util.ts
```diff
@@ -0,0 +1,61 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { AuthUtil } from 'aws-core-vscode/codewhisperer'
+import { AuthStatus } from '../../../core/dist/src/shared/telemetry/telemetry'
```
hayemaxi, line 7: this is a /dist/ import

packages/amazonq/src/extensionShared.ts
```diff
@@ -156,16 +156,18 @@ export async function activateShared(context: vscode.ExtensionContext, isWeb: bo
...
         }
 
-        const authState = (await AuthUtil.instance.getChatAuthState()).codewhispererChat
         const authKinds: AuthUtils.AuthSimpleId[] = []
         if (await AuthUtils.hasBuilderId('codewhisperer')) {
             authKinds.push('builderIdCodeWhisperer')
         }
         if (await AuthUtils.hasSso('codewhisperer')) {
             authKinds.push('identityCenterCodeWhisperer')
         }
```
hayemaxi, line 165: this section is only relevant if we did not use the authStatus of the autoconnection. Maybe getAuthStatus can also do this work for both the AuthUtils (if it has it) or the autoconnection. This function will do the work for you, so it should be an easy change: https://github.com/aws/aws-toolkit-vscode/blob/feature/standalone/packages/core/src/login/webview/vue/backend.ts#L280

hayemaxi: nevermind, this isn't quite the same metric as what the function provides. we will have to come back to this.

hayemaxi: This PR will add an appropriate util function. We can follow up. https://github.com/aws/aws-toolkit-vscode/pull/4860

hayemaxi (approved):

packages/amazonq/src/auth/util.ts
nkomonen-amazon: Import types is safe. During compilation this import is dropped since JS does not have types

PR #4607, opened 2024-03-27
feat(amazon q transform): Human In The Loop

damntrecky:
## Problem

Users builds fail sometimes do to 1p dependencies and those resources are not available at the transformation level in the backend. We are adding an "human in the loop" intervention to allow users to choose a version and resume a transformation

## Solution

Display chat input fields, from maven output, to allow users to choose a 1p dependency upgrade version and continue on.

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4851, opened 2024-04-27
fix(gumby): resume transformation job to get partial success when error occurs in HIL

JuliaABurch:
## Problem
If we exited the [Human in the Loop](https://github.com/aws/aws-toolkit-vscode/pull/4607) feature early - due to user cancellation or an error - the chat was getting stuck in a state where the user could neither get the partial transformation of their flow or start a new transformation. 

## Solution
We enter the status polling loop when we terminate HIL early, so that the backend can continue to give us progress on the transformation job. 

## License

### DIFF
--- a/packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
@@ -36,7 +36,7 @@ export type StaticTextResponseType =
     | 'job-transmitted'
     | 'end-HIL-early'
 
-export type ErrorTextResponseType =
+export type UnrecoverableErrorType =
     | 'no-project-found'
     | 'no-java-project-found'
     | 'no-maven-java-project-found'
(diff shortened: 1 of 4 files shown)


packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -337,12 +343,15 @@ export class Messenger {
...
     }
 
-    public sendUnrecoverableError(type: UnrecoverableErrorResponseType, tabID: string) {
+    public sendKnownErrorResponse(type: ErrorResponseType, tabID: string) {
         let message = '...'
 
         switch (type) {
             case 'no-alternate-dependencies-found':
-                message = `I couldn't find any other versions of this dependency in your local Maven repository. Try transforming the depedency to make it compatible with Java 17, and then try transforming this module again.`
+                message = `I could not find any other versions of this dependency in your local Maven repository. Try transforming the depedency to make it compatible with Java 17, and then try transforming this module again.`
```
dhasani23, line 351: nit: I believe we're calling it `project` for VS Code and `module` for IntelliJ

PR #4853, opened 2024-04-27
fix(amazonq): register webview error handler in Amazon Q

leigaol:
## Problem

When login view of Q goes wrong,  for example: When attempting to login with a user id whose access was revoked, the login failure error message was not properly displayed to user. 

Root cause: 

The webview error handler was registered in `extensionShared.ts`. However, `extensionShared.ts`is not shared, it only actives in AWS Toolkit. 

## Solution

register webview error handler in Amazon Q

## License

### DIFF
--- a/packages/core/src/codewhisperer/activation.ts
+++ b/packages/core/src/codewhisperer/activation.ts
@@ -4,6 +4,7 @@
  */
 
 import * as vscode from 'vscode'
+import * as nls from 'vscode-nls'
 import { getTabSizeSetting } from '../shared/utilities/editorUtilities'
 import { KeyStrokeHandler } from './service/keyStrokeHandler'
 import * as EditorContext from './util/editorContext'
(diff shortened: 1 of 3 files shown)


packages/core/src/codewhisperer/activation.ts
```diff
@@ -66,6 +66,8 @@ import { updateUserProxyUrl } from './client/agent'
 import { Container } from './service/serviceContainer'
 import { debounceStartSecurityScan } from './commands/startSecurityScan'
 import { securityScanLanguageContext } from './util/securityScanLanguageContext'
+import { registerWebviewErrorHandler } from '../webviews/server'
+import { logAndShowWebviewError } from '../extensionShared'
```
hayemaxi, line 70: can we move the function to a shared util file somewhere? We are basically importing the entire toolkit extension by doing this

leigaol: Fixed in latest commit.

hayemaxi (approved):

PR #4857, opened 2024-04-27
fix(telemetry): notice mentions extension name

hayemaxi:
## License

### DIFF
--- a/packages/core/src/shared/telemetry/activation.ts
+++ b/packages/core/src/shared/telemetry/activation.ts
@@ -12,7 +12,7 @@ import * as vscode from 'vscode'
 import { AwsContext } from '../awsContext'
 import { DefaultTelemetryService } from './telemetryService'
 import { getLogger } from '../logger'
-import { getComputeRegion, getIdeProperties, isAmazonQ, isCloud9 } from '../extensionUtilities'
+import { getComputeRegion, isAmazonQ, isCloud9, productName } from '../extensionUtilities'
 import { openSettings, Settings } from '../settings'
 import { TelemetryConfig, setupTelemetryId } from './util'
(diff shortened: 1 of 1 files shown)


packages/core/src/shared/telemetry/activation.ts
```diff
@@ -110,7 +110,9 @@ function showTelemetryNotice(extensionContext: vscode.ExtensionContext) {
 
     const telemetryNoticeText: string = localize(
         'AWS.telemetry.notificationMessage',
-        'AWS IDE Extensions collects anonymous usage metrics to improve the product. You can opt-out in settings.',
+        `${
+            isAmazonQ() ? 'Amazon Q' : 'AWS Toolkit'
+        } collects anonymous usage metrics to improve the product. You can opt-out in settings.`,
         getIdeProperties().company
```
justinmk3, line 116: nit: could use `productName` (which takes care of China branding) https://github.com/aws/aws-toolkit-vscode/blob/1b217e978fb269424738c821d91f70d0b9ebaa5d/packages/core/src/shared/extensionUtilities.ts#L40

Also, `getIdeProperties().company` doesn't look like it's used here anymore

justinmk3: p.s. technically, `localize()` needs a constant string. to string-replace stuff in a `localize()` string, use a `{0}` token and pass a parameter.

This would matter if/when we get translations for strings.

hayemaxi: fixed, interesting point about the constant string. Didn't know about that 👍

justinmk3 (approved):

justinmk3 (approved):

PR #4860, opened 2024-04-27
feat(telemetry): add auth_userState to toolkit ext

hayemaxi:
- also, to match the spec we rename metric field from 'enabledAuthConnections' to 'authEnabledConnections'

## License

### DIFF
--- a/packages/core/src/auth/ui/vue/types.ts
+++ b/packages/core/src/auth/ui/vue/types.ts
@@ -35,4 +35,5 @@ export const authFormTelemetryMapping: {
     identityCenterCodeCatalyst: { featureType: 'codecatalyst', authType: 'iamIdentityCenter' },
     identityCenterExplorer: { featureType: 'awsExplorer', authType: 'iamIdentityCenter' },
     aggregateExplorer: { featureType: 'awsExplorer', authType: 'other' }, // this should never actually be used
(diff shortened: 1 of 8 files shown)


packages/core/src/auth/ui/vue/types.ts
```diff
@@ -35,4 +35,5 @@ export const authFormTelemetryMapping: {
     identityCenterCodeCatalyst: { featureType: 'codecatalyst', authType: 'iamIdentityCenter' },
     identityCenterExplorer: { featureType: 'awsExplorer', authType: 'iamIdentityCenter' },
     aggregateExplorer: { featureType: 'awsExplorer', authType: 'other' }, // this should never actually be used
+    unknown: { featureType: 'unknown', authType: 'other' }, // this should never actually be used
```
hayemaxi, line 38: post 4/30 we will delete the old auth page stuff, just adding this to unblock compilation.

packages/core/src/auth/utils.ts
```diff
@@ -685,7 +685,9 @@ export class ExtensionUse {
 }
 
 /**
- *  Simple readable ID for telemetry reporting
+ * @deprecated
```
hayemaxi, line 688: this should be removed as part of https://github.com/aws/aws-toolkit-vscode/pull/4846 or as a follow up

justinmk3 (approved):
Nicely done

PR #4862, opened 2024-04-28
fix(amazonq): get most occurring language from project

ctlai95:
## Problem

Project scans fail server side validation if the active file is an unsupported codewhisperer language, even if there are other supported files in the project.

## Solution

Get the highest occurring language from file extension when preparing the zip file and use that as the language

## License


ashishrp-aws (approved):

justinmk3 (approved):
The Problem/Solution description is great!

PR #4865, opened 2024-04-28
fix(amazonq): CodeCatalyst login text too small

leigaol:
## Solution

Fix font size of CC login. Also make the gap between start url and CC login to 12px.

## License

### DIFF
--- a/packages/core/src/login/webview/vue/login.vue
+++ b/packages/core/src/login/webview/vue/login.vue
@@ -154,7 +154,7 @@
                 <div class="header bottomMargin">Sign in with AWS IAM Identity Center:</div>
                 <div class="code-catalyst-login" v-if="app === 'TOOLKIT'">
                     <div style="margin-bottom: 4px"></div>
-                    <div class="h4">
+                    <div class="subHeader">
                         Using AWS Builder ID?
                         <a href="#" @click="handleCodeCatalystSignin()">Skip to sign-in</a>
                     </div>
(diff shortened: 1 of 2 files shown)


packages/core/src/login/webview/vue/login.vue
```diff
@@ -566,8 +566,7 @@ export default defineComponent({
 }
 
 .h4 {
-    font-size: 8px;
-    margin-bottom: 8px;
```
leigaol, line 570: this margin-bottom is unnecessary. It does not exist in design.

packages/core/src/login/webview/vue/login.vue
```diff
@@ -151,11 +151,11 @@
                 </svg>
             </button>
             <div class="auth-container-section">
-                <div class="header">Sign in with AWS IAM Identity Center:</div>
+                <div class="header bottomMargin">Sign in with AWS IAM Identity Center:</div>
```
leigaol, line 154: 12px between Sign in with AWS IAM Identity Center and CC login

packages/core/src/login/webview/vue/login.vue
```diff
@@ -566,8 +566,7 @@ export default defineComponent({
 }
 
 .h4 {
```
nkomonen-amazon, line 568: We shouldn't call this `.h4` since this is easily confused with the `h4` element. Can we call it `subHeader` or something.

nkomonen-amazon (approved):

packages/core/src/login/webview/vue/login.vue
leigaol: renamed

PR #4871, opened 2024-04-29
Various telemetry adjustments

hayemaxi:
**Do not squash merge**

- fix authEnabledConnections for amazon Q auth_userState reporting
- fix circular dependency + minor renames, imports
- allow us to identify internal users by keeping a source variable. This only covers the connection via command case. We can add this to other cases as well, but trying to reduce changes before launch.

## License


packages/amazonq/src/auth/util.ts
```diff
@@ -28,10 +29,14 @@ export async function getAuthStatus(): Promise<AuthStatus> {
         // Determine the status of the Toolkit connection we will autoconnect to
         if (autoConnectConn) {
             authStatus = autoConnectConn.state === 'valid' ? 'connected' : 'expired'
+
+            // Though TS won't say it, AwsConnection sufficiently overlaps with Connection for the purposes
+            // of `getAuthFormIdsFromConnection`
+            authEnabledConnections = AuthUtils.getAuthFormIdsFromConnection(autoConnectConn as unknown as Connection)
```
justinmk3, line 35: future / not a blocker: would `Partial<>` work instead?

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -195,15 +205,28 @@ export abstract class SsoAccessTokenProvider {
...
+            SsoAccessTokenProvider.authSource = 'unknown'
+
             return func()
         }
 
+        // During certain flows, eg reauthentication, we are already running within a span (run())
+        // so we don't need to create a new one.
+        const span = telemetry.spans.find(s => s.name === 'aws_loginWithBrowser')
+        if (span !== undefined) {
+            return run(span as unknown as Metric<AwsLoginWithBrowser>)
```
justinmk3, line 225: not a blocker: does this not work (or whatever the right syntax is?

```suggestion
        if (span !== undefined && span instanceof Metric) {
            return run(span as Metric<AwsLoginWithBrowser>)
```

justinmk3 (approved):

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -183,23 +184,13 @@ export const connectWithCustomization = Commands.declare(
             customizationArn?: string,
             customizationNamePrefix?: string
         ) => {
+            SsoAccessTokenProvider.authSource = source
```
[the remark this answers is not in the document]
justinmk3: future idea: could we not grab the current metric here, and annotate it with telemetry.record()? that could avoid the `static source`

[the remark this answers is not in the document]
hayemaxi: > is this not during a telemetry.run() context

Bingo. We _could_ wrap the following calls in a telemetry.run() event, and the other changes would detect this and attach the metrics to the existing run event. But, this code path may not result in a browser auth event in certain circumstances. It would be wrong to emit the metric in those cases.

packages/amazonq/src/auth/util.ts
hayemaxi: Yes, but we would need to adjust the types of a bunch of utility functions :(

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
hayemaxi: could not get anything like this to work... maybe my brain is fried on type checking right now :(

PR #4872, opened 2024-04-29
fix(amazonq): Use correct URL for Q telemetry settings

leigaol:
## Problem
Use correct URL for Q telemetry settings
## License


hayemaxi:
We should double check this link... seems to not go to the intended page?

leigaol:
> We should double check this link... seems to not go to the intended page?

I have same feeling but I got this from the requirement doc.

hayemaxi (approved):

PR #4874, opened 2024-04-29
fix(amazonq): acquire Toolkit api on-demand

justinmk3:
(This can wait until after release)

## Problem:
The logic to get the Toolkit API is spread around and involves:
- a "aws.amazonq.refreshConnectionCallback" command (silent failures, not strongly typed, etc)
- numerous attempts to "eagerly" activate Toolkit
- Toolkit's own startup passes the toolkit API as an object via the "aws.amazonq.refreshConnectionCallback" command
  - this might not work if the object doesn't serialize correctly, e.g. for web workers or when the vscode "extensions.experimental.affinity" behavior is enabled.

## Solution:
Instead of trying to eagerly activate Toolkit, just get the Toolkit API on-demand. This avoids complexity and is more reliable.

## License

### DIFF
--- a/packages/core/src/codewhisperer/commands/basicCommands.ts
+++ b/packages/core/src/codewhisperer/commands/basicCommands.ts
@@ -18,7 +18,7 @@ import { ReferenceLogViewProvider } from '../service/referenceLogViewProvider'
 import { AuthUtil } from '../util/authUtil'
 import { isCloud9 } from '../../shared/extensionUtilities'
 import { getLogger } from '../../shared/logger'
-import { isExtensionActive, isExtensionInstalled, openUrl } from '../../shared/utilities/vsCodeUtils'
+import { openUrl } from '../../shared/utilities/vsCodeUtils'
 import {
     getPersistedCustomizations,
     notifyNewCustomizations,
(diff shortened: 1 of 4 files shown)


packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
```diff
@@ -36,15 +35,13 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
     }
 
     async fetchConnections(): Promise<AwsConnection[] | undefined> {
-        if (!isExtensionInstalled(VSCODE_EXTENSION_ID.awstoolkit)) {
+        const toolkitApi = getToolkitApi()
+        if (!toolkitApi) {
             return undefined
         }
-        await activateExtension(VSCODE_EXTENSION_ID.awstoolkit)
```
justinmk3, line 42: Could add this back, but it's kind of pointless because Toolkit activates on `onStartupFinished`.

leigaol:
Can you also update how toolkit api is acquired in `getAutoConnectableConnection`?

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -430,27 +459,6 @@ const registerToolkitApiCallbackOnce = once(async () => {
             }
         )
     }
-})
-export const registerToolkitApiCallback = Commands.declare(
-    { id: 'aws.amazonq.refreshConnectionCallback' },
```
leigaol, line 435: When this command is removed, this change has to go through both toolkit and Q. Assume it is released in 1.0.1 Q and 3.0.1 Toolkit, then there can be command not found issue when using 1.0.1 Q with 3.0.0 toolkit or using 1.0.0 Q with 3.0.1 toolkit.   We do wrap it under tryExecute so user should not see a command not found popup. This is not a blocker issue since user can upgrade both (but both needs to go out at same time)

justinmk3:
Not relevant since this logic was removed in https://github.com/aws/aws-toolkit-vscode/pull/5103

PR #4878, opened 2024-04-29
fix(amazonq): file scan support for c, cpp, php

ctlai95:
## Problem

Code scans support new languages

## Solution

Add c, cpp, php to support languages list

## License


justinmk3:
this was closed because it needs to target master

PR #4888, opened 2024-04-30
fix(featureDev): Prevent crash on large repos or unsupported encoding during file collection

jsalles:
## Problem
There are currently 2 problems in featureDev when doing collection of repository files:
- Trying to decode binary files might crash the extension host, and 
- too large repos might make the extension run out of available memory.

## Solution
This CR fixes both problems by:
- Adding a maximum allowed repo size of 200mb for collection
- Changing the behavior of the TextDecoder to replace the unsupported characters instead of throwing TypeError

Users who try to use featureDev on a codebase larger than 200mb will see a message asking for a different workspace to be selected.

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/constants.ts
+++ b/packages/core/src/amazonqFeatureDev/constants.ts
@@ -14,6 +14,9 @@ export const featureDevChat = 'featureDevChat'
 
 export const featureName = 'Amazon Q feature development'
 
+// Max allowed size for file collection
+export const maxRepoSizeBytes = 200 * 1024 * 1024
+
 // License text that's used in codewhisperer reference log
 export const referenceLogText = (reference: CodeReference) =>
     `[${new Date().toLocaleString()}] Accepted recommendation from Amazon Q. Code provided with reference under <a href="${LicenseUtil.getLicenseHtml(
--- a/packages/core/src/amazonqFeatureDev/util/files.ts
+++ b/packages/core/src/amazonqFeatureDev/util/files.ts
@@ -12,7 +12,7 @@ import { Uri } from 'vscode'
 import { GitIgnoreFilter } from './gitignore'
 
 import AdmZip from 'adm-zip'
-import { PrepareRepoFailedError } from '../errors'
+import { ContentLengthError, PrepareRepoFailedError } from '../errors'
 import { getLogger } from '../../shared/logger/logger'
 import { maxFileSizeBytes } from '../limits'
 import { createHash } from 'crypto'
@@ -21,6 +21,7 @@ import { ToolkitError } from '../../shared/errors'
 import { AmazonqCreateUpload, Metric } from '../../shared/telemetry/telemetry'
 import { TelemetryHelper } from './telemetryHelper'
 import { sanitizeFilename } from '../../shared/utilities/textUtilities'
+import { maxRepoSizeBytes } from '../constants'
 
 export function getExcludePattern(additionalPatterns: string[] = []) {
     const globAlwaysExcludedDirs = getGlobDirExcludedPatterns().map(pattern => `**/${pattern}/*`)
@@ -94,6 +95,7 @@ export async function collectFiles(
         return prefix === '' ? path : `${prefix}/${path}`
     }
 
+    let totalSizeBytes = 0
     for (const rootPath of sourcePaths) {
         const allFiles = await vscode.workspace.findFiles(
             new vscode.RelativePattern(rootPath, '**'),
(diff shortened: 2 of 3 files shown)


packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -115,9 +119,13 @@ export async function collectFiles(
                         zipFilePath: prefixWithFolderPrefix(relativePath.workspaceFolder, relativePath.relativePath),
                     })
                 }
+                totalSize += fileStat.size
+                if (totalSize > maxRepoSize) {
+                    throw new ContentLengthError()
```
vishaalmehrishi, line 124: What does the customer see when this happens?

jsalles: "The project you have selected for source code is too large to use as context. Please select a different folder to use for this conversation"

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -115,9 +119,13 @@ export async function collectFiles(
                         zipFilePath: prefixWithFolderPrefix(relativePath.workspaceFolder, relativePath.relativePath),
                     })
                 }
+                totalSize += fileStat.size
+                if (totalSize > maxRepoSize) {
```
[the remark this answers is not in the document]
dianaabv16: Do we need to push potentially not needed file to storage? we could smth like this:
```
   if (relativePath) {

            totalSize += fileSize;


            if (totalSize > maxRepoSize) {

                throw new ContentLengthError(`Total file collection size (${totalSize} bytes) exceeds the maximum allowed size (${maxRepoSize} bytes).`);

            }


            fileList.push({

                content: fileContent,

                zipFilePath: prefixWithFolderPrefix(relativePath.workspaceFolder, relativePath.relativePath),

            });

        }
 ```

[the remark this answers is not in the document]
Comment (jsalles) [reply packages/core/src/amazonqFeatureDev/util/files.ts]: The error is already logged
```
2024-04-30 13:02:06.014 [debug] featureDev: Failed to prepare repo: Error: The project you have selected for source code is too large to use as context. Please select a different folder to use for this conversation
```

Changed the order to only account for files that we have a relativePath

File: packages/core/src/amazonqFeatureDev/util/files.ts (hunk 1 above)
Comment (jsalles) [reply packages/core/src/amazonqFeatureDev/util/files.ts]: For reference
image

File: packages/core/src/amazonqFeatureDev/util/files.ts (hunk 2 above)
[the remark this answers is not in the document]
Comment (vishaalmehrishi) [reply packages/core/src/amazonqFeatureDev/util/files.ts]: This is a great callout @dianaabv16

**Review by vishaalmehrishi — APPROVED** [review]

File: packages/core/src/amazonqFeatureDev/util/files.ts [hunk 3]
```diff
@@ -102,29 +104,43 @@ export async function collectFiles(
         const files = respectGitIgnore ? await filterOutGitignoredFiles(rootPath, allFiles) : allFiles
 
         for (const file of files) {
-            try {
-                const fileContent = await SystemUtilities.readFile(file, new TextDecoder('utf8', { fatal: true }))
-                const relativePath = getWorkspaceRelativePath(file.fsPath, { workspaceFolders })
+            const fileStat = await vscode.workspace.fs.stat(file)
```
jpinkney-aws, line 107: does it make sense to move this right above L113? I don't think we need to check the file stat if we didn't end up reading the file (if it was a binary or whatever)

dianaabv16 (approved):

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -102,29 +104,43 @@ export async function collectFiles(
...
 }
 
+const readFile = async (file: vscode.Uri, workspaceFolders: CurrentWsFolders) => {
+    try {
+        const fileContent = await SystemUtilities.readFile(file, new TextDecoder('utf8', { fatal: false }))
+        const relativePath = getWorkspaceRelativePath(file.fsPath, { workspaceFolders })
+        return { fileContent, relativePath }
+    } catch (error) {
+        getLogger().debug(
+            `featureDev: Failed to read file ${file.fsPath} when collecting repository. Skipping the file`
```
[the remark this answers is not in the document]
jsalles: We only log error when unexpected exceptions happen. I didn't want to change this logic.

packages/core/src/amazonqFeatureDev/util/files.ts
jsalles: Yep, makes sense! Pushed the change.

jpinkney-aws:
Just one more thing, are you able to add a changelog? I'm sure theres some users that would like to know this is fixed 😄

jsalles:
> Just one more thing, are you able to add a changelog? I'm sure theres some users that would like to know this is fixed 😄

Yes, definitely! Just added a changelog with the bugfix :)

vishaalmehrishi (approved):

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -102,29 +104,43 @@ export async function collectFiles(
         const files = respectGitIgnore ? await filterOutGitignoredFiles(rootPath, allFiles) : allFiles
 
         for (const file of files) {
-            try {
-                const fileContent = await SystemUtilities.readFile(file, new TextDecoder('utf8', { fatal: true }))
-                const relativePath = getWorkspaceRelativePath(file.fsPath, { workspaceFolders })
+            const { relativePath, fileContent } = await readFile(file, workspaceFolders)
```
grimalk, line 107: That call is already reading the full file content into memory.

If the file is 1TB long, will it work?
Can we move the `const fileStat = await vscode.workspace.fs.stat(file)` branch higher?

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -137,9 +153,9 @@ export async function prepareRepoData(
     span: Metric<AmazonqCreateUpload>
 ) {
     try {
+        const files = await collectFiles(repoRootPaths, workspaceFolders, true)
         const zip = new AdmZip()
```
justinmk3, line 157: the current approach used to zip files does not "stream" the zip creation so will also run into memory constraints.

please look at https://github.com/aws/aws-toolkit-vscode/pull/4769 , can you rewrite this feature on top of that (after it lands)?

grimalk: What's the ETA of this? Presume we want both these in the same wave of the toolkit?

packages/core/src/amazonqFeatureDev/constants.ts
```diff
@@ -14,6 +14,9 @@ export const featureDevChat = 'featureDevChat'
 
 export const featureName = 'Amazon Q feature development'
 
+// Max allowed size for file collection
+export const maxRepoSize = 200 * 1024 * 1024
```
grimalk, line 18: nit: please indicate the units used.

e.g. `maxRepoSizeBytes`
Who knows what you mean, unicode chars weigh more than 1 byte

grimalk (approved):

packages/core/src/amazonqFeatureDev/util/files.ts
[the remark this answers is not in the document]
jsalles: Good point. Made the changes to calculate the allowed size before reading the file.

packages/core/src/amazonqFeatureDev/util/files.ts
jsalles: Yeah I'm happy to change our implementation to use the stream after it's merged.

jpinkney-aws (approved):

jpinkney-aws (approved):

grimalk (approved):

PR #4901, opened 2024-05-01
fix: invalid $HOME is not detected

ivikash:
## Problem
- Users with misconfigured env variables -`HOME` or `USERPROFILE` or `HOMEPATH` `AWS_SHARED_CREDENTIALS_FILE` or `AWS_CONFIG_FILE` are unable to login into Amazon Q and AWS Toolkits extensions.

## Solution
- Validate $HOME and related env vars.
- if $HOME is invalid, try other env vars or `os.homedir()`.
- Show an error on startup if a valid home dir couldn't be resolved.

## License


justinmk3:
>  Users on windows are unable to login to AWS Toolkits and Amazon Q extension on `Windows` because env variable: `HOME`  is set incorrectly

It's not really a Windows specific issue. Any system with a misconfigured $HOME env var could have this issue.

packages/core/src/auth/credentials/sharedCredentialsFile.ts
```diff
@@ -8,15 +8,24 @@
...
 
 export function getCredentialsFilename(): string {
     const env = process.env as EnvironmentVariables
 
-    return env.AWS_SHARED_CREDENTIALS_FILE || join(SystemUtilities.getHomeDirectory(), '.aws', 'credentials')
+    if (env.AWS_SHARED_CREDENTIALS_FILE && isValidPath(env.AWS_SHARED_CREDENTIALS_FILE)) {
+        return env.AWS_SHARED_CREDENTIALS_FILE
+    }
+
+    return join(SystemUtilities.getHomeDirectory(), '.aws', 'credentials')
```
justinmk3, line 20: join() is never needed for literals, and reduces readability. (same comment for all other cases in this PR, and in the future)

```suggestion
    return join(SystemUtilities.getHomeDirectory(), '.aws/credentials')
```

ivikash: NIce find :). Will update in the next revision.

packages/core/src/shared/utilities/pathUtils.ts
```diff
@@ -126,3 +128,24 @@ export function getDriveLetter(path: string): string {
...
+ *
+ * @param {string} [path] - The path to be checked. If not provided, defaults to an empty string.
+ * @returns {boolean} Returns true if the path is valid and exists, otherwise returns false.
+ */
+export function isValidPath(path?: string): boolean {
+    if (isWeb()) {
+        return !!path
+    }
+
+    const fs = require('fs-extra')
```
justinmk3, line 144: `existsSync` is in the normal 'fs' library. why is fs-extra needed? using plain 'fs' avoids the problem with web-mode, and then the isWeb() check on line 140 is no longer needed.
```suggestion
    const fs = require('fs')
```

packages/core/src/shared/utilities/pathUtils.ts
```diff
@@ -126,3 +128,24 @@ export function getDriveLetter(path: string): string {
...
 }
+
+/**
+ * Checks if the provided path is valid and exists.
+ * @todo: Migrate fs.existsSync to fsCommon.exists
+ *
+ * @param {string} [path] - The path to be checked. If not provided, defaults to an empty string.
+ * @returns {boolean} Returns true if the path is valid and exists, otherwise returns false.
+ */
+export function isValidPath(path?: string): boolean {
```
justinmk3, line 139: this would never be hit in web-mode because `getHomeDirectory` returns early for web-mode

packages/core/src/shared/utilities/pathUtils.ts
```diff
@@ -126,3 +128,24 @@ export function getDriveLetter(path: string): string {
...
+    if (isWeb()) {
+        return !!path
+    }
+
+    const fs = require('fs')
+
+    if (path && path.length !== 0 && fs.existsSync(_path.resolve(path))) {
```
justinmk3, line 146: why are we using `resolve()` here but not in the result returned by `getHomeDirectory()`. that means this is checking something different than would be returned. it probably is a signal that `getHomeDirectory()` should return a `resolve() result.

more generally, please think about how code can make decisions once, which are then used for the rest of the codepath, instead of making many decisions in multiple different places.

packages/core/src/test/credentials/sharedCredentials.test.ts
```diff
@@ -17,6 +19,7 @@ describe('sharedCredentials', function () {
         // Make a temp folder for all these tests
         // Stick some temp credentials files in there to load from
         tempFolder = await makeTemporaryToolkitFolder()
+        sinon.stub(pathUtils, 'isValidPath').returns(true)
```
justinmk3, line 22: stubbing isValidPath in various tests that happen to current hit the `getHomeDirectory()` codepath, means that future tests might run into the same issue.

justinmk3 (changes requested):

packages/core/src/shared/utilities/pathUtils.ts
ivikash: `fs` and `fsCommon` is not supported in `web` where as `fs-extra` only has a promise based API.

Thus, if we don't return early for web-mode, we will have update with a `Promise` based API.

packages/core/src/shared/utilities/pathUtils.ts
ivikash: Would you recommend something like this?

`packages/core/src/auth/credentials/sharedCredentialsFile.ts`
```diff
-    if (env.AWS_SHARED_CREDENTIALS_FILE && isValidPath(env.AWS_SHARED_CREDENTIALS_FILE)) {
-        return env.AWS_SHARED_CREDENTIALS_FILE
+    if (env.AWS_SHARED_CREDENTIALS_FILE && isValidPath(resolve(env.AWS_SHARED_CREDENTIALS_FILE))) {
+        return resolve(env.AWS_SHARED_CREDENTIALS_FILE)
     }

-    if (env.AWS_CONFIG_FILE && isValidPath(env.AWS_CONFIG_FILE)) {
-        return env.AWS_CONFIG_FILE
+    if (env.AWS_CONFIG_FILE && isValidPath(resolve(env.AWS_CONFIG_FILE))) {
+        return resolve(env.AWS_CONFIG_FILE)
     }
```

`packages/core/src/shared/systemUtilities.ts`
```diff
-        if (env.HOME && isValidPath(env.HOME)) {
-            return env.HOME
+        if (env.HOME && isValidPath(path.resolve(env.HOME))) {
+            return path.resolve(env.HOME)
         }
-        if (env.USERPROFILE && isValidPath(env.USERPROFILE)) {
-            return env.USERPROFILE
+        if (env.USERPROFILE && isValidPath(path.resolve(env.USERPROFILE))) {
+            return path.resolve(env.USERPROFILE)
         }
-        if (env.HOMEPATH && isValidPath(env.HOMEPATH)) {
+        if (env.HOMEPATH && isValidPath(path.resolve(env.HOMEPATH))) {
             const homeDrive: string = env.HOMEDRIVE || 'C:'
 
-            return path.join(homeDrive, env.HOMEPATH)
+            return path.join(homeDrive, path.resolve(env.HOMEPATH))
         }
 ```

Comment (ivikash) [reply packages/core/src/shared/utilities/pathUtils.ts:146]: Or skipping the `resolve` like

```diff
-    if (path && path.length !== 0 && fs.existsSync(_path.resolve(path))) {
+    if (path && path.length !== 0 && fs.existsSync(path)) {
```

I'd prefer the former where we check `isValidPath(path.resolve(env.HOME))` and return `path.resolve(env.HOME)`

packages/core/src/test/credentials/sharedCredentials.test.ts
ivikash: Given that in unit tests we are validating an actual path, indeed it will require stubbing in future as well. What would be your recommendation?

packages/core/src/shared/utilities/pathUtils.ts
justinmk3: I pushed a change to introduce `initUserHomeDir()`, which is a one-time routine done at startup that validates $HOME and other env vars. That allows `initUserHomeDir()` to be async while `getUserHomeDir()` remains "sync". As a minor perf bonus, `getUserHomeDir()` now uses a cached value.

justinmk3 (approved):

packages/core/src/test/credentials/sharedCredentials.test.ts
justinmk3: The [change](https://github.com/aws/aws-toolkit-vscode/pull/4901#discussion_r1667731918) that I pushed avoids this problem by only validating at startup (when `initUserHomeDir()` is called).

PR #4904, opened 2024-05-01
Rebase with master

damntrecky:
## License


justinmk3:
Looks like this is just a merge, not a rebase, which is good. *Rebasing* master onto a feature branch, would cause problems.

damntrecky:
> Looks like this is just a merge, not a rebase, which is good. _Rebasing_ master onto a feature branch, would cause problems. However rebasing the feature branch onto master would be ok (but unnecessary; a plain "merge" will have fewer conflicts).

@justinmk3 I'm not sure what to do here. We have a big chunk of work that went through a PR on this q-transform-staging branch. We have a second PR coming to finalize the project we are working on.

Should I:
1. Create a temporary branch locally to merge our feature and master and that's what this PR is?
2. Open the PR for the second chunk of work?

Is this why we don't like feature branches? Maintaining the upstream with mainline is extra steps?

PR #4911, opened 2024-05-02
feat(IamPolicyChecks): implement backend functionality, update UI

kevluu-aws:
## Problem
Implement three policy checks using Access Analyzer's APIs and Python CLI tools.
Update UI, moving away from notifications.

## License


packages/core/src/accessanalyzer/terraformConfig/default.yaml
```diff
@@ -0,0 +1,121 @@
+# generate fake ARN
+# default can be specified using the following format:
+# {<key>?<default>}
+arnServiceMap:
```
mponaws, line 4: Q: How do we plan to keep this updated with the file in open source repo. We should define a process to fetch it from the repo or some central store for a given version, as this file gets updated very often

mponaws: please create a ticket to handle this process more gracefully to reduce the ops effort.

packages/toolkit/scripts/build/copyFiles.ts
```diff
@@ -94,6 +94,12 @@ const tasks: CopyTask[] = [
         target: path.join('../../node_modules', 'web-tree-sitter', 'tree-sitter.wasm'),
         destination: path.join('src', 'tree-sitter.wasm'),
     },
+
+    // AccessAnalyzer
```
mponaws, line 98: nit: worth elaborating it to indicate that it is used for PolicyChecks for Terraform templates

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -5,55 +5,90 @@
...
+import { ToolkitError } from '../../shared/errors'
 const AmazonS3URI = require('amazon-s3-uri')
 
+const defaultTerraformConfigPath = path.join(__dirname, '..', 'terraformConfig', 'default.yaml')
+const tempFilePath = path.join(__dirname, 'policyDocument.json')
+const customPolicyCheckDiagnosticCollection = vscode.languages.createDiagnosticCollection('customPolicyCheck')
+const validatePolicyDiagnosticCollection = vscode.languages.createDiagnosticCollection('validatePolicy')
+
 export interface IamPolicyChecksInitialData {
     referenceFilePath: string
```
mponaws, line 25: Q: Why do we not have fields related to the CheckNoNewAccess here?

mponaws: I see that we're reusing these member variables for other check, I don't think we should be doing that. We should initialize the inputs for other API separately for `actions` in CheckNoNewAccess

kevluu-aws: CheckNoNewAccess and CheckAccessNotGranted both go under the same diagnostic collection, as they share the same view in the UI.

mponaws: Why can't we have a separate diagnostic collection? What happens when new APIs are added to the custom policy checks

kevluu-aws: We can have a separate diagnostic collection, but since they share the same view it may be difficult for users to determine which problems belong to which check. I imagine that users may execute the wrong checks when trying to resolve a different issue altogether. That will cause confusion as to why X warning will not go away, even though they are "running" the check again.

They will reuse the same collection.

mponaws: As discussed offline, we will share the diagnostic collection for now. Let's rename the variables and add documentation (through comments) to indicate that they are being shared across the checks

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -5,55 +5,90 @@
...
-    cfnParameterPath: string
     referenceDocument: string
+    referenceFileErrorMessage: string
+    cfnParameterPath: string
     pythonToolsInstalled: boolean
 }
 
 export class IamPolicyChecksWebview extends VueWebview {
     public static readonly sourcePath: string = 'src/accessanalyzer/vue/index.js'
     public readonly id = 'iamPolicyChecks'
```
mponaws, line 34: is it worth moving these static strings to constants?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -5,55 +5,90 @@
...
-
-    //private readonly logger = getLogger()
+    private static editedDocumentUri: vscode.Uri = vscode.window.activeTextEditor?.document.uri!
+    private static editedDocumentFileName: string = vscode.window.activeTextEditor?.document.uri.path!
+    private static editedDocument: string = ''
 
     public constructor(
         private readonly data: IamPolicyChecksInitialData,
         private readonly client: AccessAnalyzer,
+        private readonly region: string,
```
mponaws, line 42: Q: does the region here has to match with the region used in the credentials?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: These are just used to pass the initial states(from settings by the user) to initialize fields in the Webview on startup.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -71,12 +106,259 @@ export class IamPolicyChecksWebview extends VueWebview {
...
+            fs.unlinkSync(tempFilePath)
+        }
+    }
+
+    public async checkAccessNotGranted(documentType: string, referenceDocument: string, cfnParameterPath?: string) {
+        const document = IamPolicyChecksWebview.editedDocumentFileName
+        customPolicyCheckDiagnosticCollection.clear()
+        if (referenceDocument !== '') {
+            // Remove spaces, line breaks, carriage returns, and tabs
+            referenceDocument = referenceDocument.replace(/\s*|\t|\r|\n/gm, '')
```
mponaws, line 234: Any specific reason why we are doing this?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.vue
```diff
@@ -225,19 +242,24 @@ const client = WebviewClientFactory.create<IamPolicyChecksWebview>()
 export default defineComponent({
     mixins: [saveData],
     data: () => ({
-        documentType: 'JSON Policy Language',
+        documentType: 'CloudFormation',
         policyType: 'Identity',
         resourcePolicyType: 'Identity',
```
mponaws, line 247: I would recommend renaming this to `referencePolicyType` (as it is used for referencePolicy in CheckNoNewAccess)

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: CLI tool expects a comma-separated list of actions.
Users can give any text blob, so this sanitizes it.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.vue
```diff
@@ -225,19 +242,24 @@ const client = WebviewClientFactory.create<IamPolicyChecksWebview>()
 export default defineComponent({
     mixins: [saveData],
     data: () => ({
-        documentType: 'JSON Policy Language',
+        documentType: 'CloudFormation',
         policyType: 'Identity',
```
mponaws, line 246: worth adding comments and/or renaming it to indicate this is used for validatePolicy API

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
mponaws: yeah, but would that lead to different results when they use the API directly? (if we are not sanitizing them in the API, I wonder if we should do it here - don't want to effect the response given from the API v/s using the toolkit

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -5,55 +5,90 @@
 
 import * as vscode from 'vscode'
 import * as fs from 'fs'
+import * as path from 'path'
 import { getLogger, Logger } from '../../shared/logger'
 import { localize } from '../../shared/utilities/vsCodeUtils'
 import { VueWebview } from '../../webviews/main'
 import { ExtContext } from '../../shared/extensions'
-//import { telemetry } from '../../shared/telemetry/telemetry'
+// TODO: Implement Telemetry ... import { telemetry } from '../../shared/telemetry/telemetry'
```
mponaws, line 13: Ticket for telemetry if we don't have it already

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: Do you mean to give the same input for the same text blob? The tool should extract the action elements from the text blob if possible. The API expects an array of 'access' elements, we are just making it easier for users to provide that list here.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
mponaws: Can you confirm this? I think this should be the destination region where the resources are being deployed to -https://github.com/awslabs/aws-cloudformation-iam-policy-validator. I think it may give incorrect results if we fetch region from the AWS credentials when they are trying to deploy to a different region

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -5,55 +5,90 @@
...
 
     public init(): typeof this.data {
         return this.data
     }
 
     public async getReferenceDocument(path: string): Promise<string> {
-        return _getReferenceDocument(path)
+        try {
+            const referenceDocument = await _getReferenceDocument(path)
+            this.onFileReadError.fire('') // Reset error message display if no error is found
```
mponaws, line 67: What is the error message that we plan to surface ?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -71,12 +106,266 @@ export class IamPolicyChecksWebview extends VueWebview {
...
+                                            )
+                                        )
+                                        validatePolicyDiagnosticCollection.set(
+                                            IamPolicyChecksWebview.editedDocumentUri,
+                                            diagnostics
+                                        )
+                                    }
+                                })
+                                this.onValidatePolicyResponse.fire([
+                                    'Please view the problems panel to review the issues with your policy document. Policy checks should be run until issues are no longer found in your policy document.',
```
mponaws, line 168: nit: recommend moving the static texts to a constants file. 

Also, we should review with the doc writer for the final custom text messages for errors

mponaws: (we can finalize the error strings later, please track it in your ticket)

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: These are not static, but I can move some other things that can be constants.

packages/core/src/accessanalyzer/terraformConfig/default.yaml
```diff
@@ -0,0 +1,121 @@
...
+    aws_sagemaker_model_package_group_policy: model_package_group_name?fakeModelPackageGroupName
+    aws_secretsmanager_secret: name?fakename
+    aws_secretsmanager_secret_policy: secret_arn?fakeSecretArn
+    aws_transfer_access: server_id?fakeServerId
+    aws_transfer_user: user_name?fakeUserName
+    aws_vpc_endpoint: fakeName
+# iamChecks:
+#   - AccessAnalyzer
+
+# iamExceptions:
```
weibenz1, line 57: Q: what are these?

packages/core/src/accessanalyzer/terraformConfig/default.yaml
```diff
@@ -0,0 +1,121 @@
...
+    aws_elasticsearch_domain_policy: access_policies
+    aws_glacier_vault: access_policy
+    aws_glacier_vault_lock: access_policy
+    aws_glue_resource_policy: policy
+    aws_iot_policy: policy
+    aws_kms_external_key: policy
+    aws_kms_key: policy
+    aws_kms_replica_external_key: policy
+    aws_kms_replica_key: policy
+    # aws_lambda_layer_version_permission: policy
```
weibenz1, line 92: why commented out?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -5,55 +5,90 @@
...
     public constructor(
         private readonly data: IamPolicyChecksInitialData,
         private readonly client: AccessAnalyzer,
+        private readonly region: string,
         public readonly onChangeInputPath = new vscode.EventEmitter<string>(),
         public readonly onChangeReferenceFilePath = new vscode.EventEmitter<string>(),
-        public readonly onChangeCloudformationParameterFilePath = new vscode.EventEmitter<string>()
+        public readonly onChangeCloudformationParameterFilePath = new vscode.EventEmitter<string>(),
+        public readonly onValidatePolicyResponse = new vscode.EventEmitter<[string, string]>(),
+        public readonly onCustomPolicyCheckResponse = new vscode.EventEmitter<[string, string]>(),
```
weibenz1, line 47: Q: guess it goes back to design - does this mean customers click on different button or use different views for linter/treehouse checks?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -71,12 +106,266 @@ export class IamPolicyChecksWebview extends VueWebview {
...
+     * Responses are exposed in the Webview, including errors
+     * Diagnostic objects are created to expose findings, which appear in the problems panel
+     */
+    public async validatePolicy(documentType: string, policyType: string, cfnParameterPath?: string) {
+        const document = IamPolicyChecksWebview.editedDocumentFileName
+        validatePolicyDiagnosticCollection.clear()
+        const diagnostics: vscode.Diagnostic[] = []
+        switch (documentType) {
+            case PolicyChecksDocumentType.JSON_POLICY_LANGUAGE: {
+                this.client.validatePolicy(
```
weibenz1, line 119: Q: do we do a basic JSON linting before sending the request?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -71,12 +106,266 @@ export class IamPolicyChecksWebview extends VueWebview {
...
+                        policyType: policyType === 'Identity' ? 'IDENTITY_POLICY' : 'RESOURCE_POLICY',
+                    },
+                    (err, data) => {
+                        if (err) {
+                            this.onValidatePolicyResponse.fire([err.message, 'red'])
+                        } else {
+                            if (data.findings.length > 0) {
+                                data.findings.forEach((finding: AccessAnalyzer.ValidatePolicyFinding) => {
+                                    const message = `${finding.findingType}: ${finding.issueCode} - ${finding.findingDetails} Learn more: ${finding.learnMoreLink}`
+                                    if (finding.findingType === 'ERROR') {
```
weibenz1, line 131: Is `ERROR` not an enum or something similar?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -71,12 +106,266 @@ export class IamPolicyChecksWebview extends VueWebview {
...
+        }
+
+        switch (documentType) {
+            case PolicyChecksDocumentType.TERRAFORM_PLAN: {
+                const tfCommand = `tf-policy-validator check-no-new-access --template-path ${document} --region ${this.region} --config ${defaultTerraformConfigPath} --reference-policy ${tempFilePath} --reference-policy-type ${policyType}`
+                this.executeCustomPolicyChecksCommand(tfCommand)
+                return
+            }
+            case PolicyChecksDocumentType.CLOUDFORMATION: {
+                const cfnCommand =
```
weibenz1, line 221: better to make these logic a helper method so it's unit testable and easier to extend in future.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -71,12 +106,266 @@ export class IamPolicyChecksWebview extends VueWebview {
...
+                        ),
+                        `${finding.findingType}: ${finding.code} - ${finding.details.findingDetails} Learn more: ${finding.details.learnMoreLink}`,
+                        vscode.DiagnosticSeverity.Error
+                    )
+                )
+                validatePolicyDiagnosticCollection.set(IamPolicyChecksWebview.editedDocumentUri, diagnostics)
+            })
+            jsonOutput.NonBlockingFindings.forEach((finding: any) => {
+                diagnostics.push(
+                    new vscode.Diagnostic(
```
weibenz1, line 295: i see this logic being repeated at multiple places, should it be a helper method instead?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -71,12 +106,266 @@ export class IamPolicyChecksWebview extends VueWebview {
...
+                            new vscode.Range(0, 0, 0, 0),
+                            `${finding.findingType}: ${finding.message} - Resource name: ${finding.resourceName}, Policy name: ${finding.policyName} - ${reason.description}`,
+                            vscode.DiagnosticSeverity.Warning
+                        )
+                    )
+                    customPolicyCheckDiagnosticCollection.set(IamPolicyChecksWebview.editedDocumentUri, diagnostics)
+                })
+            })
+            this.onCustomPolicyCheckResponse.fire([
+                'Result: FAIL. Please view the problems panel to review the issues with your policy document. Policy checks should be run until issues are no longer found in your policy document.',
```
weibenz1, line 363: lets reuse these messages as const

packages/core/src/accessanalyzer/vue/iamPolicyChecks.vue
```diff
@@ -289,8 +322,25 @@ export default defineComponent({
...
+        },
+        runCustomPolicyCheck: function () {
+            if (this.checkType == 'CheckNoNewAccess') {
+                client.checkNoNewAccess(
+                    this.documentType,
+                    this.resourcePolicyType,
+                    this.initialData.referenceDocument,
+                    this.initialData.cfnParameterPath
+                )
+            } else {
```
weibenz1, line 336: i'd not assume there is only two checks

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -140,7 +423,65 @@ async function _getReferenceDocument(input: string): Promise<string> {
...
+    } catch (err: any) {
+        if (err.message.includes('command not found')) {
+            cfnToolInstalled = false
+        }
+    }
+    return cfnToolInstalled && tfToolInstalled
+}
+
+// Since TypeScript can only get the CLI tool's error output as a string, we have to parse and sanitize it ourselves
+function getCliErrorMessage(message: string): string {
```
weibenz1, line 458: is there no exit code/error code that we can more reliably depending on?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: The error message is S3 SDK response and `Invalid file path or S3 URI` where file path and S3 URI are invalid.

packages/core/src/accessanalyzer/terraformConfig/default.yaml
kevluu-aws: This file is needed for the Terraform CLI tool as a parameter, and is copied from the repo as well. We are copying it and providing it in the extension by default.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: There are two sections, one for ValidatePolicy and one for CustomChecks

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: VS Code provides it out of the box, but ValidatePolicy does very basic linting as well.

packages/core/src/accessanalyzer/terraformConfig/default.yaml
weibenz1: i meant what are the commented ones

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: I can create an enum, but since the type is a string from the API response, it will not be strictly respected.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.vue
mponaws: Yup, I agree with this. We should explicitly have `else if` for checking with other API

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
weibenz1: yeah, my point is we do not want to copy this string around for comparisons. easy to make typo, and imagine one day that findingType is becoming something else, how many places you'd need to update?

packages/core/src/accessanalyzer/terraformConfig/default.yaml
kevluu-aws: I am not exactly sure, as it is pulled directly from the CLI tool.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: Will move it in the future. Adding to ticket.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: Adding constants in future. Adding to ticket.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: The CLI error message varies and is not exactly consistent. As of now there are cracks of which the error output is simply the CLI error message, but we will revisit error messages. Adding to ticket.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -71,12 +106,259 @@ export class IamPolicyChecksWebview extends VueWebview {
...
+        if (referenceDocument !== '') {
+            fs.writeFileSync(tempFilePath, referenceDocument)
+        } else {
+            this.onCustomPolicyCheckResponse.fire(['Reference policy document is missing.', 'red'])
+            return
+        }
+
+        switch (documentType) {
+            case 'Terraform Plan':
+                const tfCommand = `tf-policy-validator check-no-new-access --template-path ${document} --region ${this.region} --config ${defaultTerraformConfigPath} --reference-policy ${tempFilePath} --reference-policy-type ${policyType}`
```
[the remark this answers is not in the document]
kevluu-aws: Added to ticket.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -5,61 +5,97 @@
...
-//import { telemetry } from '../../shared/telemetry/telemetry'
+// TODO: Implement Telemetry ... import { telemetry } from '../../shared/telemetry/telemetry'
 import { AccessAnalyzer, S3 } from 'aws-sdk'
-import { exec } from 'child_process'
+import { execSync } from 'child_process'
+import { ToolkitError } from '../../shared/errors'
 const AmazonS3URI = require('amazon-s3-uri')
 
+const defaultTerraformConfigPath = path.join(__dirname, '..', 'terraformConfig', 'default.yaml')
+const tempFilePath = path.join(__dirname, 'policyDocument.json')
```
justinmk3, line 20: `__dirname` should be avoided, it's not compatible in a browser (non-node) context.

static assets should be stored in the `resources/` directory and loaded as resources. there are various examples of this in the codebase. https://github.com/aws/aws-toolkit-vscode/blob/31728d1b3c0c225bb26d4ccbfc85155d23e38110/packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts#L412

packages/toolkit/scripts/build/copyFiles.ts
```diff
@@ -94,6 +94,13 @@ const tasks: CopyTask[] = [
...
+    {
+        // Used for Terraform Policy Validator CLI tool in IAM Policy Checks
+        target: path.join('../core', 'src', 'accessanalyzer', 'terraformConfig', 'default.yaml'),
+        destination: path.join('terraformConfig', 'default.yaml'),
+    },
 ]
```
justinmk3, line 105: not necessary. just put the file in `resources/`

packages/core/src/accessanalyzer/vue/iamPolicyChecks.vue
```diff
@@ -20,6 +20,9 @@
                     <li>
                         <code> pip install tf-policy-validator </code>
                     </li>
+                    <li>
+                        <p>Provide IAM Roles Credentials</p>
```
hayemaxi, line 24: To check and make use of this, do you plan on using the built in auth system provided by the toolkit?

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -140,7 +423,72 @@ async function _getReferenceDocument(input: string): Promise<string> {
...
+            } else {
+                throw new PolicyChecksError(e.message, PolicyChecksErrorCode.FileReadError)
+            }
+        }
     }
-    return ''
+}
+
+//Check if Cfn and Tf tools are installed
+function arePythonToolsInstalled(): boolean {
```
hayemaxi, line 437: Some logging about which tool isn't found may be helpful.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
hayemaxi: maybe try `makeTemporaryToolkitFolder` or `globals.context.storageUri`

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: Added logging for which tool isnt found.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.vue
kevluu-aws: Currently it only uses the shared credentials defined by the SDK. I think it will eventually use the other ways of auth in Toolkit. AFAIK, Toolkits provides a few ways to sign-in aside from IAM Roles Credentials, but if a user set their credentials using Toolkit auth system, it should pick up the same credentials since it is written to the same file.

mponaws (approved):

justinmk3 (approved):

hayemaxi (approved):

PR #4916, opened 2024-05-02
feat(auth): request necessary scopes only

hayemaxi:
Applies to both IdC and BuilderID, Amazon Q and Toolkit.

- Don't request AWS scopes if we are signing into Amazon Q
- Request AWS scopes for explorer sign in. Request AWS + CodeCatalyst scopes for CC sign in.
- Request scope difference if re-using a connection. Cancelling or otherwise failing to get the new scopes does NOT invalidate the connection.
  - Updating a connection profile now no longer invalidates the connection. I couldn't find a use for it anymore.
- Works via login page AND quickpick menu

## License

### DIFF
--- a/packages/core/src/auth/utils.ts
+++ b/packages/core/src/auth/utils.ts
@@ -102,7 +102,11 @@ export async function promptAndUseConnection(...[auth, type]: Parameters<typeof 
         // HACK: We assume that if we are toolkit we want AWS account scopes.
         // Although, Q shouldn't enter this codepath anyways.
         if (!isAmazonQ() && isSsoConnection(conn) && !hasScopes(conn, scopesSsoAccountAccess)) {
-            conn = await addScopes(conn, scopesSsoAccountAccess, { invalidate: false })
+            try {
+                conn = await addScopes(conn, scopesSsoAccountAccess)
+            } catch (e: any) {
+                throw new ToolkitError(`Failed to use connection${conn.label ? `: ${conn.label}` : '.'}`, e)
+            }
         }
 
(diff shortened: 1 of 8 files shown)


packages/core/src/test/credentials/auth.test.ts
```diff
@@ -154,13 +154,6 @@ describe('Auth', function () {
             assert.deepStrictEqual(updated.scopes, updatedProfile.scopes)
         })
 
-        it('invalidates the connection', async function () {
-            const conn = await auth.createConnection(ssoProfile)
-            const updated = await auth.updateConnection(conn, updatedProfile)
-
-            assert.strictEqual(auth.getConnectionState(updated), 'invalid')
```
justinmk3, line 161: should we assert "valid" instead of removing the test?

packages/core/src/auth/utils.ts
```diff
@@ -93,12 +94,22 @@ export async function promptForConnection(auth: Auth, type?: 'iam' | 'iam-only'
 
 export async function promptAndUseConnection(...[auth, type]: Parameters<typeof promptForConnection>) {
```
justinmk3, line 95: The PR description is a beautifully succinct outline of the "spec". Would this be the best place to add that "spec" as a docstring?

justinmk3 (approved):

packages/core/src/auth/utils.ts
hayemaxi: added to addScopes instead, it is a common point for the prompter in toolkit and the login pages

PR #4924, opened 2024-05-03
fix(amazonq): incorrect project scan size limit

ctlai95:
## Problem

`collectFiles` method uses a different max payload size value than security scans.

## Solution

- Pass the desired max value to `collectFiles`
- Use 500MB as the max size

## License


laileni-aws (approved):

andrewyuq (approved):

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -69,7 +69,8 @@ async function filterOutGitignoredFiles(rootPath: string, files: Uri[]): Promise
 export async function collectFiles(
     sourcePaths: string[],
     workspaceFolders: CurrentWsFolders,
-    respectGitIgnore: boolean = true
+    respectGitIgnore: boolean = true,
+    maxSize = maxRepoSizeBytes
```
jpinkney-aws, line 73: I think we have an opportunity to improve some quality in the codebase which was probably left over from re-invent. It looks like codewhisperer is importing from the featureDev part of the codebase. If this is the case, can we pull `collectFiles` out to a different place?

Generic utils (like collectFiles) used by multiple q products shouldn't be in individual products part of the codebase.

Also, was this issue caused by a lack of tests? Because we should be adding more tests to know that this issue is fixed

jpinkney-aws (changes requested):

packages/core/src/amazonqFeatureDev/util/files.ts
ctlai95: Hey @jpinkney-aws , I agree with your points and I think we should make those changes, but I think we should make those changes separately. This PR addresses the incorrect file size limit being used for customers so it's a little more urgent than the refactor.

justinmk3 (approved):
Thank you for lifting `collectFiles` into a shared module!

PR #4925, opened 2024-05-03
auth: remove pkce feature flag and use device code on remote

jpinkney-aws:
_Still pending final decision on what we will do before releasing_

## Problem
- auth server ports are hard to forward (and unforward) making the device code a better option for now when running in a remote workspace

## Solution
- remove the feature flag and use device code authorization when in a remote workspace

## License


jpinkney-aws:
This can be closed since we've released pkce

PR #4926, opened 2024-05-03
fix: update session state config when modifying root folder

sannicm:
## Problem

When the repo is larger that 200MB amazonqFeatureDev doesn't allow the customer to upload the repo and it prompts the customer to select a folder that could be used for the plan. This modification was not updating accordingly the config of the session state so it kept using the same root.

## Solution

Update the config of the state on the fly and the session config to have the root folder correctly selected.

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/session/sessionState.ts
+++ b/packages/core/src/amazonqFeatureDev/session/sessionState.ts
@@ -55,6 +55,11 @@ export class PrepareRefinementState implements Omit<SessionState, 'uploadId'> {
     constructor(private config: Omit<SessionStateConfig, 'uploadId'>, public approach: string, public tabID: string) {
         this.tokenSource = new vscode.CancellationTokenSource()
     }
+
+    updateWorkspaceRoot(workspaceRoot: string) {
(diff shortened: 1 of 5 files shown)


packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -55,6 +55,11 @@ export class PrepareRefinementState implements Omit<SessionState, 'uploadId'> {
     constructor(private config: Omit<SessionStateConfig, 'uploadId'>, public approach: string, public tabID: string) {
         this.tokenSource = new vscode.CancellationTokenSource()
     }
+
+    updateWorkspaceRoot(workspaceRoot: string) {
+        this.config.workspaceRoots = [workspaceRoot]
```
jsalles, line 60: How will this work if the previous value of workspaceRoots had more than one entry?

sannicm: It will override them, given the circumstances I think it makes sense. The customer is also informed of the situation that needs to reduce the context.

jsalles (approved):

nkomonen-amazon (approved):

PR #4930, opened 2024-05-03
fix(build): invalid resource paths on Windows

ivikash:
## Problem
Development on `Windows` for `aws-toolkit-vscode` is not working because inaccurate path

## Solution
glob.sync requires posix style path. The code changes updates path.resolve to replace \\ to / from Windows to posix style.

## License

### DIFF
--- a/packages/core/scripts/build/generateIcons.ts
+++ b/packages/core/scripts/build/generateIcons.ts
@@ -112,7 +112,7 @@ async function generate(mappings: Record<string, number | undefined> = {}) {
         template: 'css',
         templateClassName: 'icon',
         descent: 200, // Icons were negatively offset on the y-axes, this fixes it
-        templateFontPath: path.relative(stylesheetsDir, fontsDir),
+        templateFontPath: path.relative(stylesheetsDir, fontsDir).replace(/\\/g, '/'),
         glyphTransformFn: obj => {
             const filePath = (obj as { path?: string }).path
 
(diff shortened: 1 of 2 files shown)


packages/core/scripts/build/generateIcons.ts
```diff
@@ -112,7 +112,7 @@ async function generate(mappings: Record<string, number | undefined> = {}) {
         template: 'css',
         templateClassName: 'icon',
         descent: 200, // Icons were negatively offset on the y-axes, this fixes it
-        templateFontPath: path.relative(stylesheetsDir, fontsDir),
+        templateFontPath: path.relative(stylesheetsDir, fontsDir).replace(/\\/g, '/'),
```
justinmk3, line 115: does this also affect the production build? i.e. this might have been a problem not only during development?

packages/webpack.vue.config.js
```diff
@@ -32,7 +32,7 @@ const createVueBundleName = file => {
  */
 const createVueEntries = (targetPattern = 'index.ts') => {
     return glob
-        .sync(path.resolve(currentDir, 'src', '**', 'vue', '**', targetPattern))
+        .sync(path.resolve(currentDir, 'src', '**', 'vue', '**', targetPattern).replace(/\\/g, '/'))
```
justinmk3, line 35: NB: `replaceAll()` is also available on the platforms we target.

packages/core/scripts/build/generateIcons.ts
justinmk3, line 115: There is related logic on line 144. Ideally that would be extracted and de-duplicated, which would help avoid similar bugs in the future.

PR #4937, opened 2024-05-06
placeholder changelog item

nkomonen-amazon:
We just need something so the CHANGELOG.md does not look weird and has no items for the Toolkit release.

We will want a better mechanism for this in the future.

## License


justinmk3 (approved):
Since https://github.com/aws/aws-toolkit-vscode/pull/4930 was merged, we could perhaps mention:

    fix: some icons may be missing on Windows

In the future, the script can auto-generate something when there are no changes.

PR #4942, opened 2024-05-07
feat: add a conversationId as an info log to ease customer feedback

sannicm:
## Problem

Searching for the conversationId in logs could be cumbersome for some customers. This aims to make it a little bit easier.

## Solution

Add an info level log with the conversation id in approach and code generation, including their iterations for easy conversation Id finding.

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
+++ b/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
@@ -300,6 +300,8 @@ export class FeatureDevController {
     private async onApproachGeneration(session: Session, message: string, tabID: string) {
         await session.preloader(message)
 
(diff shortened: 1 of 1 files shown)


jsalles:
Can we also add this log message after we receive the `conversationId` from `StartConversation`?
That way we'll have this info for flows that fail before starting planning.

sannicm:
> Can we also add this log message after we receive the `conversationId` from `StartConversation`? That way we'll have this info for flows that fail before starting planning

Printing effectively after the preloader runs is doing what you suggest, printing before any approach interaction is done. This way we would have conversationId in every approach plan or plan iteration as well before any upload or generation.

jsalles (approved):

nkomonen-amazon (approved):

PR #4945, opened 2024-05-07
[WIP] - Amazon Q Code Transform: Enforce valid JDK for user provided JAVA_HOME

arangatang:
## Problem
As part of the Code Transformation Process, users can provide a custom JAVA_HOME. Previously we only verified if the default JAVA_HOME was correctly detected and not this user provided JAVA_HOME.
This meant customers could pass an invalid JDK which could lead to the following problems:

1. Pass a path thats not even a JDK -> Causing build errors
2. Pass a path thats the wrong JDK version -> I.e. 17 and leveraging backwards compatibllity when build requires 8/11. This obfuscates certain issues and makes them only detectable during the first backend build which builds using the JDK 8 or 11.
3. Customers have to provide a custom JAVA_HOME multiple times upon retrying submission.

### Secondary problem - Bugfix
Extracted to #4949 

## Solution
1. If JAVA_HOME recorded into state use that when verifying mavens detected JAVA version
2. Run validation also after JAVA_HOME is provided and ask customers to retry on failure.
3. Change all occurences in Q Code Transform of `tabId` to `tabID` - Extracted to #4949 

#### Screenshots for main problem
1. When running JDK 8 and user provides JDK 17 -> Ask customer to try again
Screenshot 2024-05-07 at 17 54 11

2. When user retries with valid JDK, job continues
Screenshot 2024-05-07 at 17 55 18

3. When user started a successfull job and provided a valid JDK version then subsequent jobs of same language version does not require users to provide JAVA_HOME again.
Screenshot 2024-05-07 at 17 56 24

4. When user changes project to a JDK 11 project, we request a valid JDK version for that (even when JDK 8 version was successful for previous project)
Screenshot 2024-05-07 at 17 59 46

## License


packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -327,6 +320,29 @@ export class GumbyController {
...
+                        await validateCanCompileProject()
+                    } catch (err: any) {
+                        if (err instanceof JavaHomeNotSetError) {
+                            transformByQState.unsetJavaHome()
+                            this.sessionStorage.getSession().conversationState = ConversationState.PROMPT_JAVA_HOME
+                            this.messenger.sendStaticTextResponse('java-home-not-set', data.tabID)
+                            this.messenger.sendChatInputEnabled(data.tabID, true)
+                            this.messenger.sendUpdatePlaceholder(
+                                data.tabID,
+                                `Incompatible JDK, Enter the path to your Java ${transformByQState.getSourceJDKVersion()} installation.`
```
JuliaABurch, line 339: 1. This is an update to the string and should be reviewed 
2. Was this interaction reviewed with Ishaani? I'm unsure if updating the placeholder to indicate the issue is the best way to do this, if only because it's not a pattern we repeat anywhere else with errors. 
3. Can we extract any of this to a common method shared between `processHumanChatMessage` and `validateBuildWithPromptOnError`? Can the former call the latter?
4. Does this retry need to be limited in any way, or is it ok that it's effectively an infinite retry loop?

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -327,6 +320,29 @@ export class GumbyController {
                 const pathToJavaHome = extractPath(data.message)
 
                 if (pathToJavaHome) {
+                    transformByQState.setJavaHome(pathToJavaHome)
+                    getLogger().info(
+                        `CodeTransformation: using JAVA_HOME = ${transformByQState.getJavaHome()} since source JDK does not match Maven JDK`
+                    )
+
+                    try {
+                        // When user provided a
```
JuliaABurch, line 329: Incomplete comment?

damntrecky (approved):

packages/core/src/amazonqGumby/chat/controller/controller.ts
arangatang: Great points, ill sync this with Ishaani, think that will answer 1,2,4. 

3. Can look into that

packages/core/src/amazonqGumby/chat/controller/controller.ts
arangatang: True, removed it!

packages/core/src/codewhisperer/models/model.ts
```diff
@@ -598,6 +598,10 @@ export class TransformByQState {
         this.javaHome = javaHome
     }
 
+    public unsetJavaHome() {
```
dhasani23, line 601: nit: `reset`?

packages/core/src/amazonqGumby/chat/controller/controller.ts
damntrecky: Should add a retry limit of 3 and review with UX after.

arangatang:
Force push due to rebase against master.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -327,6 +320,39 @@ export class GumbyController {
...
+                        await validateCanCompileProject()
+                    } catch (err: any) {
+                        if (err instanceof JavaHomeNotSetError) {
+                            const providedJdkVersion =
+                                (await getJavaVersionStringUsedByMaven()) ?? JDKVersion.UNSUPPORTED
+                            const expectedJdkVersion = transformByQState.getSourceJDKVersion() ?? JDKVersion.UNSUPPORTED
+                            getLogger().warn(
+                                `CodeTransformation: non matching JAVA_HOME provided: ${providedJdkVersion} expected: ${expectedJdkVersion} JDK release must match`
+                            )
+                            if (transformByQState.incrementAndGetJavaHomeAttempts() > 3) {
```
dhasani23, line 338: Move `3` to constants file

arangatang:
Failing test seem to be flaky and is unrelated.

arangatang:
Having some discussion internally if this is the way we want to go forward with this, so moving to draft mode to avoid accidental merges.

PR #4952, opened 2024-05-07
fix(featureDev): placeholder message

kumsmrit:
## Problem
Updated placeholder message for feature dev in planning phase which was misleading previously due to "Briefly" in message.
## Solution
Updated placeholder message for feature dev in planning phase to new UX message

## License

### DIFF
--- a/packages/core/src/amazonq/webview/ui/quickActions/generator.ts
+++ b/packages/core/src/amazonq/webview/ui/quickActions/generator.ts
@@ -28,7 +28,7 @@ export class QuickActionGenerator {
                         ? [
                               {
                                   command: '/dev',
-                                  placeholder: 'Briefly describe a task or issue',
+                                  placeholder: 'Describe your task or issue in as much detail as possible',
                                   description:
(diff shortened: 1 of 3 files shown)


sarayev (approved):

packages/core/src/amazonq/webview/ui/quickActions/generator.ts
```diff
@@ -28,7 +28,7 @@ export class QuickActionGenerator {
                         ? [
                               {
                                   command: '/dev',
-                                  placeholder: 'Briefly describe a task or issue',
+                                  placeholder: 'Describe your task or issue in as much detail as possible',
```
sarayev, line 31: non-blocker: Should we store this text in a constant and refer to it from all 3 places we need?

kumsmrit: I don't see any of the messages being shared for webview ui across, as all messages are explicitly mentioned within the generators.

PR #4963, opened 2024-05-08
web: get Amazon Q debug mode running

nkomonen-amazon:
Amazon Q is not fully working in web, but this gets the Debug setup working so we can start developing

## License

### DIFF
--- a/packages/amazonq/src/extensionWeb.ts
+++ b/packages/amazonq/src/extensionWeb.ts
@@ -1,0 +1,15 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { ExtensionContext } from 'vscode'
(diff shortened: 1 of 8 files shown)


packages/amazonq/src/extensionWeb.ts
```diff
@@ -0,0 +1,15 @@
+/*!
```
justinmk3, line 1: off topic / not a blocker: should we rename this to main.ts/mainWeb.ts as in toolkit: https://github.com/aws/aws-toolkit-vscode/blob/d27031ea5f26ece189a0921616e4676edbc84027/packages/toolkit/src/main.ts

or vice versa?

justinmk3 (approved):

packages/amazonq/src/extensionWeb.ts
nkomonen-amazon: I agree, I was thinking that we should go back to using `extension`. I'll look to update this

PR #4964, opened 2024-05-08
feat(IamPolicyChecks): Refactor code(constants, functions, errors), move from DevSettings to Experiments, change S3Client

kevluu-aws:
Refactor code by moving static strings to a constants file, creating help functions, and updating error messages.

Moving from DevSettings to Experiments to allow access.

S3Client changed to use `DefaultS3Client` and adapted URI parser [dependency](https://www.npmjs.com/package/amazon-s3-uri) to a helper function.

## License


packages/core/src/accessanalyzer/vue/constants.ts
```diff
@@ -0,0 +1,37 @@
...
+    CLOUDFORMATION = 'CloudFormation',
+    JSON_POLICY_LANGUAGE = 'JSON Policy Language',
+}
+
+export enum ValidatePolicyFindingType {
+    ERROR = 'ERROR',
+    SECURITY_WARNING = 'SECURITY_WARNING',
+    SUGGESTION = 'SUGGESTION',
+    WARNING = 'WARNING',
+}
```
justinmk3, line 37: enums really aren't needed in modern typescript. instead it's preferred to use this pattern: https://github.com/aws/aws-toolkit-vscode/blob/1ba1e0427f4e797c1124bc12c5bd6addfec5138f/packages/core/src/login/webview/vue/reauthenticate.vue#L85-L89

packages/core/src/shared/clients/s3Client.ts
```diff
@@ -709,3 +741,89 @@ async function createSdkClient(regionCode: string): Promise<S3> {
 function clearInternalBucketCache(): void {
     ;(S3.prototype as any).bucketRegionCache = {}
 }
+
+/**
+ * A URI parser that can parse out information about an S3 URI
+ * Adapted from
+ * @see https://github.com/frantz/amazon-s3-uri/
+ */
+export function parseS3Uri(uri: string): [region: string, bucket: string, key: string] {
```
justinmk3, line 750: It's very important to avoid dependencies, but I'm not sure I understand the motivation for "inlining" this particular case. Was it pulling in outdated transitive packages, or some other reason?

packages/core/package.json
```diff
@@ -2017,7 +2022,7 @@
                 "command": "aws.accessanalyzer.iamPolicyChecks",
                 "title": "%AWS.command.accessanalyzer.iamPolicyChecks%",
                 "category": "%AWS.title%",
-                "enablement": "aws.iamPolicyChecks.enabled"
+                "enablement": "config.aws.experiments.iamPolicyChecks"
```
mponaws, line 2025: Do we plan to create another PR to enable this by default before the launch and release that version through the toolkits pipeline?

packages/core/src/accessanalyzer/vue/constants.ts
```diff
@@ -0,0 +1,37 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export abstract class IamPolicyChecksConstants {
```
mponaws, line 6: We should review these error messages with the doc writer

packages/core/src/accessanalyzer/vue/constants.ts
```diff
@@ -0,0 +1,37 @@
...
+    static readonly ValidatePolicySuccessWithFindings =
+        'Please view the problems panel to review the issues with your policy document. Policy checks should be run until issues are no longer found in your policy document.'
+    static readonly ValidatePolicySuccessNoFindings = 'Policy checks did not discover any problems with your policy.'
+    static readonly CustomCheckSuccessWithFindings =
+        'Result: FAIL. Please view the problems panel to review the issues with your policy document. Policy checks should be run until issues are no longer found in your policy document.'
+    static readonly CustomCheckSuccessNoFindings =
+        'Result: PASS. Policy checks did not discover any problems with your policy.'
+    static readonly CustomCheckFilePathSetting = 'aws.accessAnalyzer.policyChecks.customChecksFilePath'
+    static readonly CfnParameterFilePathSetting = 'aws.accessAnalyzer.policyChecks.cloudFormationParameterFilePath'
+    static readonly MissingReferenceDocError = 'Reference policy document is missing.'
```
mponaws, line 16: I don't think this is appropriate. We use the same for actions - the reference doc in that scenario is a not a policy document

packages/core/src/shared/clients/s3Client.ts
```diff
@@ -626,6 +636,28 @@ export class DefaultS3Client {
             }
         }
     }
+
+    /**
+     * Gets an object from a bucket.
+     *
+     * The bucket should reside in the same region as the one configured for the client.
```
mponaws, line 643: Why do we plan to enforce limitation? Can't we do cross region calls? 
https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/s3-cross-region.html

packages/core/src/shared/clients/s3Client.ts
```diff
@@ -709,3 +741,89 @@ async function createSdkClient(regionCode: string): Promise<S3> {
...
+    const defaultRegion = 'us-east-1' // Default region for URI parsing, if region is not found
+    const parsedUri = url.parse(uri)
+    let bucket: string | undefined = undefined
+    let region: string = defaultRegion
+    let key: string | undefined = undefined
+
+    if (parsedUri.protocol === 's3:') {
+        bucket = parsedUri.host ?? undefined
+        if (!bucket) {
+            throw new Error(`Invalid S3 URI: no bucket: ${uri}`)
```
mponaws, line 761: Why do we want to lint the URI given? Can't we rely on S3 for that?

kevluu-aws: S3 does not support GetObject using URI out of the box. The idea of this is to get the region, bucket, and key from the URI and use those parameters to perform GetObject.

packages/core/src/shared/clients/s3Client.ts
kevluu-aws: I don't think it was pulling in anything transitive. @hayemaxi suggested that we move away from the dependency just as a general standard I believe.

packages/core/package.json
kevluu-aws: I am assuming that is the process, but this will now make it available through Experiments, which is a checkbox in the settings.

packages/core/src/accessanalyzer/vue/constants.ts
kevluu-aws: You're right, it should just be "reference document," will update.

packages/core/src/shared/clients/s3Client.ts
kevluu-aws: The comment was taken from similar functions. GetObject should be able to get it cross-region.

kevluu-aws: Removed the comment.

mponaws (approved):

PR #4968, opened 2024-05-09
ux(auth): change IdC roles message to warn instead of error

hayemaxi:
image

## License


nkomonen-amazon (approved):

justinmk3:
Previously this was avoided for IdC (such as in the screenshot) that only had codewhisperer scopes. I wonder if that heuristic needs to be updated.

PR #4973, opened 2024-05-10
fix(auth): re-use existing connections if provided manually

hayemaxi:
- If the user enters start url and region of an already existing connection that was displayed, we will just proceed with re-using that connection.
- If the user enters start url but region differs from what exists, we will display the error.

Other misc fixes:
- fix(toolkit): return to explorer if login page is exited
- fix(codecatalyst): use default scopes (includes aws account scopes) for sign in

## License


justinmk3 (approved):
if some of these logs are noisy, may want to use a `logOnce` pattern

PR #4977, opened 2024-05-13
fix(featureDev): use max repo size from featuredev constants

gasolima:
Problem: using default maxSize in util function that used by multiple team can make bugs in the future
Solution: use featureDev constants and override the default one, since they're now matching the limits will not be affected

## License


jsalles (approved):

nkomonen-amazon (approved):

justinmk3:
Does this conflict with https://github.com/aws/aws-toolkit-vscode/pull/4985 (which also has a test)?

gasolima:
> Does this conflict with #4985 (which also has a test)?

Looks like he's doing the same thing as i do, so i can close this PR

PR #4980, opened 2024-05-13
devEnv: Fix Dev Env timeout + logs

nkomonen-amazon:
## Problem:

We only tried to start the activity heartbeat on extension startup. So if
that failed due to something like an invalid connection, the heartbeat mechanism
would not start

## Solution:

- Try and start the heartbeat mechanism when the CC connection changes as well.
But if it is already running then we just skip it.
- Also added better logs to help debug this issue. The user should see `codecatalyst: DevEnvActivity: heartbeat sent at ...` in the logs, which indicates we called the Dev Env `/activity` api

## License

### DIFF
--- a/packages/core/src/auth/ui/vue/show.ts
+++ b/packages/core/src/auth/ui/vue/show.ts
@@ -5,7 +5,7 @@
  * This module sets up the necessary components
  * for the webview to be shown.
  */
-import globals, { isWeb } from '../../../shared/extensionGlobals'
+import globals from '../../../shared/extensionGlobals'
 import { getIdeProperties, isCloud9 } from '../../../shared/extensionUtilities'
 import { VueWebview } from '../../../webviews/main'
 import * as vscode from 'vscode'
@@ -35,26 +35,24 @@ import {
     hasSso,
     BuilderIdKind,
     findSsoConnections,
-    authCommands,
 } from '../../utils'
 import { Region } from '../../../shared/regions/endpoints'
 import { CancellationError } from '../../../shared/utilities/timeoutUtils'
 import { validateSsoUrl, validateSsoUrlFormat } from '../../sso/validation'
-import { awsIdSignIn, showCodeWhispererConnectionPrompt } from '../../../codewhisperer/util/showSsoPrompt'
-import { AuthError, ServiceItemId, isServiceItemId, authFormTelemetryMapping, userCancelled } from './types'
+import { awsIdSignIn } from '../../../codewhisperer/util/showSsoPrompt'
+import { AuthError, ServiceItemId, authFormTelemetryMapping, userCancelled } from './types'
 import { connectToEnterpriseSso } from '../../../codewhisperer/util/getStartUrl'
 import { trustedDomainCancellation } from '../../sso/model'
 import { CredentialSourceId, Result, telemetry } from '../../../shared/telemetry/telemetry'
 import { AuthFormId } from './authForms/types'
 import { handleWebviewError } from '../../../webviews/server'
-import { Commands, RegisteredCommand, VsCodeCommandArg, placeholder } from '../../../shared/vscode/commands2'
+import { placeholder } from '../../../shared/vscode/commands2'
 import { ClassToInterfaceType } from '../../../shared/utilities/tsUtils'
 import { debounce } from 'lodash'
 import { submitFeedback } from '../../../feedback/vue/submitFeedback'
 import { InvalidGrantException } from '@aws-sdk/client-sso-oidc'
 import { ExtStartUpSources } from '../../../shared/telemetry'
-import { CommonAuthWebview } from '../../../login/webview/vue/backend'
-import { AuthSource, AuthSources } from '../../../login/webview/util'
+import { AuthSource } from '../../../login/webview/util'
 import { focusAmazonQPanel } from '../../../codewhispererChat/commands/registerCommands'
 
 // This file has some used functions, but most of it should be removed soon. We have a new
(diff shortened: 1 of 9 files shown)


packages/core/src/codecatalyst/devEnv.ts
```diff
@@ -11,6 +11,45 @@ import { Timeout } from '../shared/utilities/timeoutUtils'
...
+
+    if (!thisDevenv) {
+        getLogger().verbose('codecatalyst: no dev env to start heartbeat')
+        return
+    }
+
+    const maxInactivityMinutes = thisDevenv.summary.inactivityTimeoutMinutes
+    const devEnvClient = thisDevenv.devenvClient
+    const devEnvActivity = await DevEnvActivity.instanceIfActivityTrackingEnabled(devEnvClient)
+    if (shouldTrackUserActivity(maxInactivityMinutes) && devEnvActivity) {
```
justinmk3, line 45: In theory, any `await` call may park the current execution and re-enter `tryStartDevEnvActivityHeartbeat` from e.g. `onDidChangeActiveConnection` during startup.  Even if that's unlikely or currently impossible, it's always a good idea to make the expectations as clear as possible. Thus we should check `didStartDevEnvActivityHeartbeat` just before we create the resource(s).


```suggestion
    if (shouldTrackUserActivity(maxInactivityMinutes) && devEnvActivity && !didStartDevEnvActivityHeartbeat) {
```

packages/core/src/codecatalyst/devEnv.ts
```diff
@@ -11,6 +11,45 @@ import { Timeout } from '../shared/utilities/timeoutUtils'
...
+let didStartDevEnvActivityHeartbeat: boolean = false
+/**
+ * Trys to start the Dev Env Activity Heartbeat to keep the Dev Env from timing out.
+ * If this successfully starts, subsequent calls to this function will do nothing.
+ *
+ * ---
+ * Why do we need this?
+ *
+ * We need this function since there are times the heartbeat mechanism cannot be started. Eg auth
+ * is not connected in a Dev Env. So we need to try again elsewhere.
```
justinmk3, line 26: is there a way to express the same info in a more concise (yet still accurate and complete) wording?

packages/core/src/auth/ui/vue/show.ts
```diff
@@ -714,47 +712,6 @@ const Panel = VueWebview.compilePanel(AuthWebview)
 let activePanel: InstanceType<typeof Panel> | undefined
 let subscriptions: vscode.Disposable[] | undefined
 
-let showManageConnections: RegisteredCommand<any> | undefined
```
nkomonen-amazon, line 717: Moved this code to `login/commands.ts` due to circular dependency issues

packages/core/src/codecatalyst/devEnv.ts
nkomonen-amazon: Good find, I noticed this was getting called before fully completing. I've refactored a bit, but updated to use the `shared()` helper function to essentially debounce any subsequent calls that can occur.

packages/core/src/codecatalyst/devEnv.ts
nkomonen-amazon: Reduced doc string

nkomonen-amazon:
@justinmk3 could you give this a final look when you have time

packages/core/src/codecatalyst/devEnv.ts
```diff
@@ -11,6 +11,122 @@ import { Timeout } from '../shared/utilities/timeoutUtils'
...
+        ))
+    }
+
+    /** Keeps executing {@link DevEnvActivityHeartbeatStarter.tryStartDevEnvActivityHeartbeat} on an interval until it succeeds*/
+    private async tryStartDevEnvActivityHeartbeatWithRetry() {
+        // try immediately since interval below delays
+        await this.tryStartDevEnvActivityHeartbeat()
+
+        this.startHeartbeatInterval ??= globals.clock.setInterval(async () => {
+            await this.tryStartDevEnvActivityHeartbeat()
```
justinmk3, line 54: why not use waitUntil?

packages/core/src/codecatalyst/devEnv.ts
```diff
@@ -11,6 +11,122 @@ import { Timeout } from '../shared/utilities/timeoutUtils'
 import { showMessageWithCancel } from '../shared/utilities/messages'
 import { isCloud9 } from '../shared/extensionUtilities'
 import { getLogger } from '../shared/logger'
+import { CodeCatalystAuthenticationProvider } from './auth'
+import { getThisDevEnv } from './model'
+import { isInDevEnv } from '../shared/vscode/env'
+import { shared } from '../shared/utilities/functionUtils'
+
+/** This class exists due to complexities with starting the Dev Env Activity Hearbeat mechanism */
+export class DevEnvActivityHeartbeatStarter {
```
justinmk3, line 20: long names really can be counterproductive. shorter names help others by mentioning the most important aspects of a thing. repeating "DevEnv" in something that lives in devEnv.ts is probably avoidable. and "Starter" is over-explanation, because there is no other "ActivityHeartbear" class, so there is no ambiguity without "Starter".

```suggestion
export class ActivityHeartbeat {
```

packages/core/src/codecatalyst/devEnv.ts
```diff
@@ -11,6 +11,122 @@ import { Timeout } from '../shared/utilities/timeoutUtils'
...
+    /**
+     * Trys to start the Dev Env Activity Heartbeat mechanism.
+     * This mechanism keeps the Dev Env from timing out.
+     */
+    public static registerDevEnvActivityHeartbeatStarter(authProvider: CodeCatalystAuthenticationProvider) {
```
justinmk3, line 25: (usually) no need to repeat the name of the class in its functions
```suggestion
    public static register(authProvider: CodeCatalystAuthenticationProvider) {
```

packages/core/src/codecatalyst/devEnv.ts
```diff
@@ -11,6 +11,122 @@ import { Timeout } from '../shared/utilities/timeoutUtils'
...
+            getLogger().debug('codecatalyst: not in a devenv, not registering DevEnvActivityHeartbeatStarter')
+            return
+        }
+
+        DevEnvActivityHeartbeatStarter.authProvider = authProvider
+        void DevEnvActivityHeartbeatStarter.instance.tryStartDevEnvActivityHeartbeatWithRetry()
+    }
+
+    /** If true, the Activity Heartbeat is running */
+    private didStartDevEnvActivityHeartbeat: boolean = false
```
justinmk3, line 36: this is a local variable. no need to repeat so much info in its name. it has a small scope and there is little risk of ambiguity. a shorter name like `didStart` is visually easy to notice and draws attention to the most important aspects.
```suggestion
    private didStart: boolean = false
```

packages/core/src/codecatalyst/devEnv.ts
```diff
@@ -11,6 +11,122 @@ import { Timeout } from '../shared/utilities/timeoutUtils'
...
+                // on success stop the interval
+                this.startHeartbeatInterval?.unref()
+                this.startHeartbeatInterval = undefined
+            }
+        }, 30_000)
+    }
+
+    /** Trys to start the Activity Heartbeat mechanism */
+    private tryStartDevEnvActivityHeartbeat = shared(this._tryStartDevEnvActivityHeartbeat.bind(this)) // prevents race conditions
+    private async _tryStartDevEnvActivityHeartbeat() {
```
justinmk3, line 66: probably can avoid the extra function by using arrow function (which binds to `this`)


```suggestion
    private tryStartDevEnvActivityHeartbeat = shared(async () => {
```

nkomonen-amazon: nice

packages/core/src/codecatalyst/devEnv.ts
nkomonen-amazon: This makes sense, but this class only handles starting of the `DevEnvActivity` class.
I made this class to contain all the "starting" code in a single place. I've renamed to `DevEnvActivityStarter`, and update the docstring to better reflect this.

PR #4981, opened 2024-05-13
fix(amazonq): Autofocus to prompt input when panel opens

dogusata:
## Problem
When window gets focus or there is a tab addition/change on Q Chat, even though the `autoFocus` property in the config is set to `true`, input field doesn't get focus.

## Solution
Added autofocus to initialization, window focus and tab add/change events through [mynah-ui fix](https://github.com/aws/mynah-ui/pull/48) and with the [version v4.8.1](https://github.com/aws/mynah-ui/releases/tag/v4.8.1)

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,6 +47,61 @@
                 "webpack-merge": "^5.10.0"
             }
         },
+        "../mynah-ui": {
+            "name": "@aws/mynah-ui",
+            "version": "4.8.1",
+            "hasInstallScript": true,
+            "license": "Apache License 2.0",
+            "dependencies": {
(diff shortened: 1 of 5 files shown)


hayemaxi (approved):

package-lock.json
```diff
@@ -47,6 +47,61 @@
                 "webpack-merge": "^5.10.0"
             }
         },
+        "../mynah-ui": {
```
hayemaxi, line 50: This looks like it shouldn't be here. Might be causing the build to fail?

dogusata: 100% right, thanks

PR #4985, opened 2024-05-14
fix(/dev): logic to allow customer to modify root source folder

sannicm:
## Problem

A recent refactor (https://github.com/aws/aws-toolkit-vscode/pull/4924) moved `collectFiles` to a shared file but also updated the logic to throw another error that the `/dev` controller was not expecting. This effectively broke the feature that allows customer to modify root source folder when the folder exceeds 200MB

## Solution

- Updated the if-else logic to check for the correct error
- Added a unit test to make it clearer that `/dev` controller expects the error to be specific to avoid breaking the logic.

## License

### DIFF
--- a/packages/amazonq/.changes/next-release/Bug Fix-7034adfe-fc09-44d2-8328-d908ab99f1ac.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-7034adfe-fc09-44d2-8328-d908ab99f1ac.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Fixes an issue where the /dev chat wouldn't let customers modify the source folder when exceeding the size limit"
+}
--- a/packages/core/src/amazonqFeatureDev/util/files.ts
+++ b/packages/core/src/amazonqFeatureDev/util/files.ts
@@ -16,6 +16,7 @@ import { CurrentWsFolders } from '../types'
 import { ToolkitError } from '../../shared/errors'
 import { AmazonqCreateUpload, Metric } from '../../shared/telemetry/telemetry'
 import { TelemetryHelper } from './telemetryHelper'
+import { maxRepoSizeBytes } from '../constants'
 
 const getSha256 = (file: Buffer) => createHash('sha256').update(file).digest('base64')
 
@@ -29,7 +30,7 @@ export async function prepareRepoData(
     span: Metric<AmazonqCreateUpload>
 ) {
     try {
-        const files = await collectFiles(repoRootPaths, workspaceFolders, true)
+        const files = await collectFiles(repoRootPaths, workspaceFolders, true, maxRepoSizeBytes)
         const zip = new AdmZip()
 
         let totalBytes = 0
@@ -55,8 +56,8 @@ export async function prepareRepoData(
         }
     } catch (error) {
         getLogger().debug(`featureDev: Failed to prepare repo: ${error}`)
-        if (error instanceof ContentLengthError) {
-            throw error
+        if (error instanceof ToolkitError && error.code === 'ContentLengthError') {
+            throw new ContentLengthError()
         }
         throw new PrepareRepoFailedError()
     }
--- a/packages/core/src/test/amazonqFeatureDev/util/files.test.ts
+++ b/packages/core/src/test/amazonqFeatureDev/util/files.test.ts
@@ -2,12 +2,16 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
-
+import * as vscode from 'vscode'
 import assert from 'assert'
 import { prepareRepoData } from '../../../amazonqFeatureDev/util/files'
 import { createTestWorkspace } from '../../testUtil'
 import { TelemetryHelper } from '../../../amazonqFeatureDev/util/telemetryHelper'
 import { AmazonqCreateUpload, Metric } from '../../../shared/telemetry/telemetry'
+import { ContentLengthError } from '../../../amazonqFeatureDev/errors'
+import { maxRepoSizeBytes } from '../../../amazonqFeatureDev/constants'
+import sinon from 'sinon'
+import { fsCommon } from '../../../srcShared/fs'
 
 describe('file utils', () => {
     describe('prepareRepoData', function () {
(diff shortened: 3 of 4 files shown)


packages/core/package.json
```diff
@@ -3984,7 +3984,7 @@
     },
     "devDependencies": {
         "@aws-sdk/types": "^3.13.1",
-        "@aws-toolkits/telemetry": "^1.0.205",
```
sannicm, line 3987: note to reviewers: 

I've updated telemetry and added a new event emission to the /dev logic as the cost was very low and it was just one line in the logic. Let me know if you'd rather split this.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -231,7 +232,7 @@ export class FeatureDevController {
                     break
             }
         } catch (err: any) {
-            if (err instanceof ContentLengthError) {
+            if (err instanceof ToolkitError && err.code === 'ContentLengthError') {
```
kumsmrit, line 235: If we modify this check, how will the `ContentLengthError` be handled that is being thrown explicitly [here](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/amazonqFeatureDev/client/featureDev.ts#L120) ?

sannicm: Good catch, I need to handle both.

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -19,6 +19,8 @@ import { TelemetryHelper } from './telemetryHelper'
 
 const getSha256 = (file: Buffer) => createHash('sha256').update(file).digest('base64')
 
+export const amazonqFeatureDevMaxRepoSize = 200 * 1024 * 1024 // 200 MB
```
justinmk3, line 22: can the existing constant be used? https://github.com/aws/aws-toolkit-vscode/blob/e52e000a9116a1a718bb1b54c1c4172c9cd38efe/packages/core/src/amazonqFeatureDev/constants.ts#L18

I see a lot of global constants being created, but rarely used more than once.

packages/amazonq/.changes/next-release/Bug Fix-7034adfe-fc09-44d2-8328-d908ab99f1ac.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "fix logic to allow customer to modify root source folder"
```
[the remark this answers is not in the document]
sannicm: I understand, the issue is that is hard to put in words because the root cause is a refactor that changed an exception that broke our logic, so basically this is fixing the logic again as said in the description.

packages/core/src/amazonqFeatureDev/util/files.ts
sannicm: Yes! good catch.

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -55,8 +56,8 @@ export async function prepareRepoData(
         }
     } catch (error) {
         getLogger().debug(`featureDev: Failed to prepare repo: ${error}`)
-        if (error instanceof ContentLengthError) {
-            throw error
+        if (error instanceof ToolkitError && error.code === 'ContentLengthError') {
```
kumsmrit, line 59: Its better to keep existing check for `ContentLengthError` as well here.

kumsmrit (approved):

nkomonen-amazon (approved):

packages/core/src/test/amazonqFeatureDev/util/files.test.ts
```diff
@@ -28,5 +30,32 @@ describe('file utils', () => {
...
+            const telemetry = new TelemetryHelper()
+
+            try {
+                await prepareRepoData([workspace.uri.fsPath], [workspace], telemetry, {
+                    record: () => {},
+                } as unknown as Metric<AmazonqCreateUpload>)
+            } catch (err) {
+                assert.strictEqual(err instanceof ContentLengthError, true)
+            }
+        }).timeout(120 * 1000) // Allow test to run longer than 30 seconds
```
hayemaxi, line 59: Why does the test need to run longer than 30 seconds?

sannicm: Becuase I'm trying to create files that will break the 200MB limit for repo size, this file creation and all takes more time than 30s

sannicm: I'm adding the test specifically to avoid any future regressions as a refactor made by another team broke our feature.

hayemaxi: This seems like a convoluted way of creating a large set of files. Can't you just try something like?
```
fs.writeFile('file', new Buffer(<some size>));
```
That way you can just make one big file, and it might be faster.

If so, make sure to use the fs implementation if `srcShared/fs.ts`

hayemaxi: Alternatively, maybe we can stub `vscode.workspace.fs.stat(file.fileUri))` or something. I think that would be better than actually allocating a bunch of space on the testing machine (which would be our machines much of the time)

packages/core/src/test/amazonqFeatureDev/util/files.test.ts
```diff
@@ -28,5 +30,32 @@ describe('file utils', () => {
...
+
+            const workspace = await createTestWorkspace(files, { fileNamePrefix, fileContent })
+            const telemetry = new TelemetryHelper()
+
+            try {
+                await prepareRepoData([workspace.uri.fsPath], [workspace], telemetry, {
+                    record: () => {},
+                } as unknown as Metric<AmazonqCreateUpload>)
+            } catch (err) {
+                assert.strictEqual(err instanceof ContentLengthError, true)
```
justinmk3, line 57: if `prepareRepoData` doesn't throw, the test will pass. is that intended? why not use `assert.throws` or `assert.rejects` (look for examples in the codebase)

packages/core/src/test/amazonqFeatureDev/util/files.test.ts
justinmk3: > That way you can just make one big file, and it might be faster. Would like to avoid extending timeouts since tests already take a long time to run.

Yes, absolutely. In fact, this should be in `testInteg/`, that is where slow tests live.

sannicm:
> #4977 also changed the collectFiles() call in `packages/core/src/amazonqFeatureDev/session/sessionState.ts` , should that be done in this PR?

I noticed that PR far later I did this PR so it is ok the change was merged on that one. I will rebase here and remove my change regarding that.

packages/core/src/test/amazonqFeatureDev/util/files.test.ts
sannicm: I've resorted to stubbing the `fsCommon.stat` function. As there is no need in that case to make the test longer, by just stubbing the function I can return a value that will make the logic throw the error.

hayemaxi (approved):

justinmk3:
Does this collectFiles call need to be updated? https://github.com/aws/aws-toolkit-vscode/blob/df730861b1c3febdfc1e0df12659ac83f17dd054/packages/core/src/amazonqFeatureDev/session/sessionState.ts#L380-L384

PR #4994, opened 2024-05-14
fix(amazonq): Resume Builder ID connection if scope of feature dev

leigaol:
## Problem

If a builder id connection only has 3 scopes, it cannot be transferred to Amazon Q without user clicking a button to re-authenticate to add 2 new scopes: feature dev and transform. 

## Solution

For users who have yet migrated, let Q pick up connections even if it does not have 2 new scopes. Q will use connection that only has cwspr core scopes.

For users who was not successfully migrated, they will be in below states:
1. Q is not logged in, Q has no connection. Toolkit has old 3 scope builder id connection. Q login screen displays an option for user to click to sign in using toolkit Q connection.  For this case, a re-install of this new Amazon Q extension will automatically reconnect them in Q with 3 scope builder id connection.

2. Q has expired 3 scope builder id connection marked as invalid (pushed from toolkit). User sees a re-authentication needed screen in Q login view, which is incorrectly marked as invalid. For this case, we let the Q extension grab the toolkit connection and correct its state.

This PR should be reverted in the future.

## License

### DIFF
--- a/packages/core/src/codewhisperer/util/authUtil.ts
+++ b/packages/core/src/codewhisperer/util/authUtil.ts
@@ -434,7 +434,7 @@ export class AuthUtil {
      * From the given connections, returns a connection that works with Amazon Q.
      */
     findUsableQConnection(connections: AwsConnection[]): AwsConnection | undefined {
-        const hasQScopes = (c: AwsConnection) => amazonQScopes.every(s => c.scopes?.includes(s))
+        const hasQScopes = (c: AwsConnection) => codeWhispererChatScopes.every(s => c.scopes?.includes(s))
         const score = (c: AwsConnection) => Number(hasQScopes(c)) * 10 + Number(c.state === 'valid')
         connections.sort(function (a, b) {
             return score(b) - score(a)
--- a/packages/core/src/codewhisperer/util/authUtil.ts
+++ b/packages/core/src/codewhisperer/util/authUtil.ts
@@ -444,6 +444,22 @@ export class AuthUtil {
         }
         return undefined
     }
+
+    /**
+     * For builder id connections with 3 Q scopes ( prev 2.20.0 aws toolkit)
+     * Refresh its state
+     */
+    async refreshOldAwsBuilderIdConnection() {
+        if (
+            this.conn &&
+            isBuilderIdConnection(this.conn) &&
+            !this.conn.scopes?.includes(scopesFeatureDev[0]) &&
+            this.conn.scopes?.includes(scopesCodeWhispererChat[0]) &&
+            this.isConnectionExpired()
+        ) {
+            await this.auth.refreshConnectionState(this.conn)
+        }
+    }
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
```diff
@@ -58,6 +60,49 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
...
+        const currentConnection = AuthUtil.instance.conn
+
+        // Only when Amazon Q has no valid builder id connections while Toolkit has connections
+        // Auto connect Q using toolkit connection.
+        const shouldReuseToolkitConnection = () => {
+            if (currentConnection === undefined) {
+                return true
+            }
+            if (
+                currentConnection.type === 'sso' &&
```
leigaol, line 86: This massive if-else is trying to make the blast radius much smaller to reduce the risk.

andrewyuq (approved):

packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
```diff
@@ -58,6 +60,49 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
...
     }
 
+    /**
+     * Attempts to re-use a valid connection in AWS Toolkit.
+     * This logic should only apply to below users:
+     * 1. They have a invalid builder id connection in Amazon Q that do not have all 5 scopes
+     * For these users, if there is a valid builder id connection in AWS Toolkit,
+     * We attempt to re-use this valid builder id connection.
+     */
+    async tryReuseConnectionFromAwsToolkit() {
```
nkomonen-amazon, line 70: Not related to this specifically but can you add a test in `techdebt.test.ts` to revert this work at a certain date.

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -434,7 +434,7 @@ export class AuthUtil {
      * From the given connections, returns a connection that works with Amazon Q.
      */
     findUsableQConnection(connections: AwsConnection[]): AwsConnection | undefined {
-        const hasQScopes = (c: AwsConnection) => amazonQScopes.every(s => c.scopes?.includes(s))
+        const hasQScopes = (c: AwsConnection) => codeWhispererChatScopes.every(s => c.scopes?.includes(s))
```
nkomonen-amazon, line 437: If this function returns a connection, do we know if it is working (non-expired). It is possible that we return a connection with the correct scopes but it is expired?

nkomonen-amazon: Looks like `useConnection` (which this flow eventually calls) will validate the connection (see if it can retrieve a token) and then allow the user to proceed. So this doesn't look to be an issue

justinmk3:
We decided to improve various parts of the UI instead of this short-term fix.

PR #4995, opened 2024-05-14
dev: Edit Auth connections + fix stale auths

nkomonen-amazon:
## Problem:

It is common for want to view/modify the Auth Profile store. The Profile Store is our source of truth for auth connections that the extension is aware of and using.
We sometimes want to read and modify this source of truth for debugging purposes.

The issue is that this option is hard to find since you need to dig through multiple menus. Also once you get to it, the content you see will become stale if it ever changes and there is no way to refresh it to get the latest changes due to a bug.

## Solution:

- Add a new item to our Developer Menu to open the auth profile store editor.
- Fix issue where the text document editor was not refreshing with the latest changes.
  - The root of the issue was that we never signaled that the file content changed.
  - So we need to call `refresh()` to force the file to signal its content change, which then propagates to the FileSystemProvider.

#### New option
Screenshot 2024-05-14 at 4 13 16 PM

#### It opens this which you can edit
Screenshot 2024-05-14 at 4 13 30 PM

#### Example of editing
1. I highlight the Builder ID entry
2. I delete the Builder ID connection
3. I reopen the auth text editor which refreshes the content to the latest state
  i. We must do this manual refresh since we don't have an event listener setup to know when to refresh. 

https://github.com/aws/aws-toolkit-vscode/assets/118216176/6ddf10e5-01ef-4755-a410-30655a4226df

## License

### DIFF
--- a/packages/core/src/dev/activation.ts
+++ b/packages/core/src/dev/activation.ts
@@ -21,6 +21,7 @@ import { isAnySsoConnection } from '../auth/connection'
 import { Auth } from '../auth/auth'
 import { getLogger } from '../shared/logger'
 import { entries } from '../shared/utilities/tsUtils'
+import { getEnvironmentSpecificMemento } from '../shared/utilities/mementos'
 
 interface MenuOption {
     readonly label: string
@@ -38,6 +39,7 @@ export type DevFunction =
     | 'showEnvVars'
     | 'deleteSsoConnections'
     | 'expireSsoConnections'
+    | 'editAuthConnections'
 
 let targetContext: vscode.ExtensionContext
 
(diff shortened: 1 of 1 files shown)


packages/core/src/dev/activation.ts
```diff
@@ -204,10 +211,21 @@ class VirtualObjectFile implements FileProvider {
...
+        this.mTime++
+        this.onDidChangeEmitter.fire()
+    }
+
     public stat(): { ctime: number; mtime: number; size: number } {
         // This would need to be filled out to track conflicts
-        return { ctime: 0, mtime: 0, size: 0 }
+        return { ctime: 0, mtime: this.mTime, size: 0 }
     }
+    private mTime = 0
```
justinmk3, line 228: nice approach. nit: should this variable live near the other object state (line 210)

PR #4996, opened 2024-05-14
feat(IamPolicyChecks): implement Telemetry, update error colors, fix problems line number

kevluu-aws:
Telemetry implementation, updating response colors to be shown in the Webview, and return intended line numbers in problems panel.

Corresponding PR for new metrics in `aws-toolkit-common`: https://github.com/aws/aws-toolkit-common/pull/735
## License

### DIFF
--- a/packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
+++ b/packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
@@ -10,19 +10,24 @@ import { getLogger, Logger } from '../../shared/logger'
 import { localize } from '../../shared/utilities/vsCodeUtils'
 import { VueWebview } from '../../webviews/main'
 import { ExtContext } from '../../shared/extensions'
-// TODO: Implement Telemetry ... import { telemetry } from '../../shared/telemetry/telemetry'
-import { AccessAnalyzer } from 'aws-sdk'
+import { telemetry } from '../../shared/telemetry/telemetry'
+import { AccessAnalyzer, SharedIniFileCredentials } from 'aws-sdk'
 import { execSync } from 'child_process'
 import { ToolkitError } from '../../shared/errors'
 import { makeTemporaryToolkitFolder, tryRemoveFolder } from '../../shared/filesystemUtilities'
 import { globals } from '../../shared'
 import {
     IamPolicyChecksConstants,
+    PolicyChecksCheckType,
     PolicyChecksDocumentType,
     PolicyChecksErrorCode,
+    PolicyChecksPolicyType,
+    PolicyChecksResult,
+    PolicyChecksUiClick,
     ValidatePolicyFindingType,
 } from './constants'
 import { DefaultS3Client, parseS3Uri } from '../../shared/clients/s3Client'
+import { ExpiredTokenException } from '@aws-sdk/client-sso-oidc'
 
 const defaultTerraformConfigPath = 'resources/policychecks-tf-default.yaml'
 // Diagnostics for Custom checks are shared
@@ -46,7 +51,7 @@ export class IamPolicyChecksWebview extends VueWebview {
 
     public constructor(
         private readonly data: IamPolicyChecksInitialData,
-        private readonly client: AccessAnalyzer,
+        private client: AccessAnalyzer,
         private readonly region: string,
         public readonly onChangeInputPath = new vscode.EventEmitter<string>(),
         public readonly onChangeCustomChecksFilePath = new vscode.EventEmitter<string>(),
@@ -74,7 +79,7 @@ export class IamPolicyChecksWebview extends VueWebview {
             const customChecksTextArea = await _readCustomChecksFile(path)
             this.onFileReadError.fire('') // Reset error message display if no error is found
             return customChecksTextArea
-        } catch (err) {
+        } catch (err: any) {
             if (err instanceof PolicyChecksError && (err.code as PolicyChecksErrorCode) === 'FileReadError') {
                 this.onFileReadError.fire(err.message)
             }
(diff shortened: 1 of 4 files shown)


packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -74,10 +77,13 @@ export class IamPolicyChecksWebview extends VueWebview {
             const customChecksTextArea = await _readCustomChecksFile(path)
             this.onFileReadError.fire('') // Reset error message display if no error is found
             return customChecksTextArea
-        } catch (err) {
+        } catch (err: any) {
             if (err instanceof PolicyChecksError && (err.code as PolicyChecksErrorCode) === 'FileReadError') {
                 this.onFileReadError.fire(err.message)
             }
+            telemetry.accessanalyzer_iamPolicyChecksError.emit({
```
hayemaxi, line 84: instead of using `.emit()`, use `.record()`. This will handle errors automatically. There are many examples in the codebase and it is also explained in `telemetry.md`.

Applicable throughout PR.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -112,11 +118,19 @@ export class IamPolicyChecksWebview extends VueWebview {
         })
     }
 
+    public emitUiClick(elementId: string) {
```
hayemaxi, line 121: Type your elementIds.

Example:
https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/login/webview/vue/types.ts#L48

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -130,13 +144,17 @@ export class IamPolicyChecksWebview extends VueWebview {
                         },
                         (err, data) => {
                             if (err) {
-                                if (err.message.includes('The security token included in the request is invalid')) {
+                                if (
+                                    err.message.includes('The security token included in the request is invalid') ||
+                                    err.message.includes('The security token included in the request is expired')
```
hayemaxi, line 149: Can you maybe check for instances of `ExpiredTokenException` or similar?

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1230,6 +1264,59 @@
...
+                    "type": "accessanalyzerDocumentType"
+                },
+                {
+                    "type": "accessanalyzerFindingsCount"
+                },
+                {
+                    "type": "accessanalyzerInputPolicyType"
+                },
+                {
+                    "type": "accessanalyzerReferencePolicyType"
```
hayemaxi, line 1288: The `accessanalyzer` prefixes for the metadata are unnecessary

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: One of the reasons I strayed away from `.record()` is because the tool we use returns an "error" when we discover findings, under a different exit code. We check for the exit code to handle it gracefully, but if I used `.run()` and `.record()` there is no way for me to specify that it actually `Succeeded`.
I found that I am unable to set the `result` or `reason` when using `.record()`

kevluu-aws: Using `.record()` now

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: Yeah, that should do that trick for this instance. Unfortunately, I will have to parse it for the CLI output as we get it back as a blob.

kevluu-aws: This does the trick for this instance, though I will still need to parse for the message for the CLI errors below, since they only give back string blobs.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -308,90 +386,163 @@ export class IamPolicyChecksWebview extends VueWebview {
...
+                if (err.status === 2) {
+                    //CLI responds with a status code of 2 when findings are discovered
+                    const findingsCount = this.handleValidatePolicyCliResponse(err.stdout.toString())
+                    span.record({
+                        findingsCount: findingsCount,
+                    })
+                } else {
+                    span.record({
+                        findingsCount: 0,
+                    })
```
hayemaxi, line 435: nit: you can add this to your first record on line 415. Fields can be re-recorded as much as you want, and the latest recording will be the one that makes it into the metric.

packages/core/src/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -514,15 +664,17 @@ function arePythonToolsInstalled(): boolean {
...
-    const errorMatch = message.match(/ERROR: .*/)
-    const botoMatch = message.match(/(?<=botocore\.exceptions\.).*/) // Boto errors have a different match
+    const cfnMatch = message.match(/ERROR: .*/)
+    const botoMatch = message.match(/(?<=botocore\.exceptions\.).*/) // Boto errors have a special match
     const terraformMatch = message.match(/AttributeError:.*/) // Terraform CLI responds with a different error schema... this catches invalid .json plans
-    if (errorMatch?.[0]) {
-        return errorMatch[0]
+    if (
+        message.includes('The security token included in the request is invalid') ||
+        message.includes('The security token included in the request is expired')
```
hayemaxi, line 672: nit: same comment as before, try to prefer comparing to an exception type if possible rather than the message.

hayemaxi:
Lgtm, waiting for an approval from your team.

hayemaxi (approved):

mponaws (approved):

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1,5 +1,39 @@
 {
     "types": [
+        {
+            "name": "cfnParameterFileUsed",
+            "type": "boolean",
+            "description": "Boolean value of whether or not a Cfn parameter file is provided."
```
dhruvigajjar, line 6: can you try and put all metrics in common only. makes it easier to keep track

kevluu-aws: I created these to have the metrics available in my dev environment. I am still waiting for the common [changes](https://github.com/aws/aws-toolkit-common/pull/735) to be merged, which has the same metrics.

PR #4998, opened 2024-05-15
fix: remove unused arg

nkomonen-amazon:
This function does not accept a 4th arg anymore, but somehow the previous PR (#4987) did not fail CI.

## License


hayemaxi:
does this mean our CI has a bug?

hayemaxi (approved):

nkomonen-amazon:
@hayemaxi yeah, we need to update our `npm run lint` since it only runs in `packages/core` atm, looking at it now. This PR fixes an issue so other devs don't run in to it

PR #5004, opened 2024-05-15
Updated Q feature dev name

kumsmrit:
## Problem
QFeature Dev was renamed to Amazon Q Developer Agent for software development; so updated the feature name for chat messages and logs.
## Solution
Updated the Q feature dev name for chat messages and logs.

## License


sarayev (approved):

sannicm (approved):

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -726,12 +727,21 @@ export class FeatureDevController {
...
             }
         } catch (err: any) {
-            this.messenger.sendErrorMessage(
-                createUserFacingErrorMessage(err.message),
-                message.tabID,
-                this.retriesRemaining(session),
-                session?.state.phase
-            )
+            if (err instanceof WorkspaceFolderNotFoundError) {
+                this.messenger.sendAnswer({
```
justinmk3, line 731: how is this change related?

kumsmrit: This has been updated based on UX feedback [here](https://quip-amazon.com/fA3JAlEEK3ab/QFeatureDev-Errors-string-review#temp:C:SAH3f48f3e68c914276b0c7517c6) as the older message contained feature name which was required to be removed.

sarayev (approved):

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
sannicm: Is the WorkspaceFolderNotFoundError actually possible here?

kumsmrit: Yes, this is where `WorkspaceFolderNotFoundError` will be thrown for case when workspace is not open.

sannicm: I think is a valid remark then the one from justin, it feels like it is being combined two features into one. The rename and the handling of this new error. I approve anyway

kumsmrit: This was changed as we do not want to show feature name in case of this error message.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -440,7 +441,7 @@ export class FeatureDevController {
             if (!this.isAmazonQVisible) {
                 const open = 'Open chat'
                 const resp = await vscode.window.showInformationMessage(
-                    'Your code suggestions from Amazon Q are ready to review',
+                    'The Amazon Q Developer Agent for software development has generated code for you to review',
```
hayemaxi, line 444: IMO: this is quite verbose

kumsmrit: This text is suggested by AWS documentation team [here](https://quip-amazon.com/fA3JAlEEK3ab/QFeatureDev-Errors-string-review#temp:C:SAH3f48f3e68c914276b0c7517c6)

hayemaxi (approved):

PR #5014, opened 2024-05-17
[WIP] - Amazon Q Code Transform: Fix job loss on restart

arangatang:
## Problem
Currently we only store the latest job details and this is lost on restart. This means customer job results are lost irrevocably. 

## Solution
1. Store relevant job details in vsCodes workspace state
2. Add buttons to show diff / summary in job history panel
3. Make job history panel display multiple entries
4. Add expireOn field to indicate when job details should be removed
5. Remove expired job details and delete associated job artifacts from disk on extension start

## TODOs
1. Changelog
2. Avoid deleting on accept / reject (so that job data stays)

## License


justinmk3:
Do you plan to continue this?

PR #5020, opened 2024-05-17
fix(amazonq): duplicate findings for multi-folder workspaces

ctlai95:
## Problem

Finding sometimes get duplicated for multi-folder workspaces if two folders have the same parent folder.

## Solution

Drop the first part of the key instead of using `..` in the path.

## License

### DIFF
--- a/packages/core/src/codewhisperer/service/securityScanHandler.ts
+++ b/packages/core/src/codewhisperer/service/securityScanHandler.ts
@@ -59,7 +59,7 @@ export async function listScanResults(
     })
     codeScanIssueMap.forEach((issues, key) => {
         projectPaths.forEach(projectPath => {
-            const filePath = path.join(projectPath, '..', key)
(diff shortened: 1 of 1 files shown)


packages/core/src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -59,7 +59,7 @@ export async function listScanResults(
     })
     codeScanIssueMap.forEach((issues, key) => {
         projectPaths.forEach(projectPath => {
-            const filePath = path.join(projectPath, '..', key)
+            const filePath = path.join(projectPath, key.split('/').slice(1).join('/'))
```
justinmk3, line 62: This seems non-obvious, might be worth a one-line comment so that others in the future know about this quirk if they are changing the code

justinmk3 (approved):

PR #5029, opened 2024-05-20
Amazon Q 3 scopes to 5 scopes migration

hayemaxi:
Continuation of https://github.com/aws/aws-toolkit-vscode/pull/5026

Changes:
- Force the flow by focusing the amazon Q panel specifically for users with a 3 scope connection.
- Additional commit to turn the status bar item red when not connected.

## License

### DIFF
--- a/packages/core/src/codewhisperer/commands/basicCommands.ts
+++ b/packages/core/src/codewhisperer/commands/basicCommands.ts
@@ -42,6 +42,7 @@ import { SystemUtilities } from '../../shared/systemUtilities'
 import { ToolkitError } from '../../shared/errors'
 import { isRemoteWorkspace } from '../../shared/vscode/env'
 import { hasScopes } from '../../auth/connection'
+import globals from '../../shared/extensionGlobals'
 
 export const toggleCodeSuggestions = Commands.declare(
     { id: 'aws.amazonq.toggleCodeSuggestion', compositeKey: { 1: 'source' } },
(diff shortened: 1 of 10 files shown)


packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -270,6 +271,13 @@ export const notifyNewCustomizationsCmd = Commands.declare(
     }
 )
 
+function focusQAfterDelay() {
+    // this command won't work without a small delay after install
+    setTimeout(() => {
```
nkomonen-amazon, line 276: Use `globals.clock.setTimeout()` instead. All time related things should be done with `globals.clock`. This allows us to stub `globals.clock` in our tests and be able to control the time there.

packages/core/src/codewhisperer/service/inlineCompletionService.ts
```diff
@@ -248,6 +248,7 @@ export class CodeWhispererStatusBar {
             }
             case 'notConnected':
                 statusBar.text = codicon` ${getIcon('vscode-chrome-close')} ${title}`
+                statusBar.backgroundColor = new vscode.ThemeColor('statusBarItem.errorBackground')
```
nkomonen-amazon, line 251: I'm assuming `errorBackground` is red. Do we really want to make it red when they haven't added a connection yet? I don't think we do this for the Toolkit extension, only showing red once expired.

hayemaxi: Yeah this is intended

packages/core/src/codewhisperer/commands/basicCommands.ts
hayemaxi: done

nkomonen-amazon (approved):

PR #5035, opened 2024-05-21
codewhisperer: send customizationArn via telemetry clientComponentLatency, perceivedLatency

Will-ShaoHua:
Telemetry Common repo PR: https://github.com/aws/aws-toolkit-common/pull/738

## Solution

Add customizationArn field when sending `clientComponentLatency`, `perceivedLatency`

## BuilderId

vsc_perceivedLatency_builderId

vsc_componentLatency_builderId

## IdC - default foundation

vsc_perceivedLatency_idc_foundation

vsc_componentLatency_idc_foundation

## IdC - using customization

vsc_clientComponentLatency

vsc_perceivedLatency

## License


Will-ShaoHua:
will revert telemetry override after telemetry version bump

Will-ShaoHua:
irrelevant to this PR(testing with market place Q plugin and was able to repro), notice there is a log entry which we might need to fix 
```
2024-05-21 13:13:23.506 [warning] telemetry: invalid Metric: "codewhisperer_clientComponentLatency" emitted without the `result` property, which is always required. Consider using `.run()` instead of `.emit()`, which will set these properties automatically. See https://github.com/aws/aws-toolkit-vscode/blob/master/docs/
```

hayemaxi:
> will revert telemetry override after telemetry version bump

Can't we remove it in this PR since the version is being bumped here?

Will-ShaoHua:
> Can't we remove it in this PR since the version is being bumped here?
@hayemaxi  yea it's been removed , https://github.com/aws/aws-toolkit-vscode/pull/5035/commits/a883008b409f1cfc5baeb82de9a23048f545406b

andrewyuq (approved):

justinmk3:
> > Can't we remove it in this PR since the version is being bumped here?
> 
> yea it's been removed , [a883008](https://github.com/aws/aws-toolkit-vscode/commit/a883008b409f1cfc5baeb82de9a23048f545406b)

that commit isn't in this PR. should it be?

PR #5036, opened 2024-05-22
Fix/overlapping tips

ivikash:
## Problem
1. Tips are displayed on programming languages not supported by Amazon Q
2. First time users experience multiple onboarding tips

## Solution
| Before | After | Notes |
|  --- | --- | ---- |
| Tip - 1 | Tip - 2 | Disable Tips for file types not supported by Amazon Q |
| Screenshot 2024-05-21 at 3 35 06 PM | Screenshot 2024-05-21 at 2 06 02 PM | Disable Codelens if Amazon Q Chat is open |
| Screenshot 2024-05-21 at 2 05 31 PM | Screenshot 2024-05-21 at 5 33 52 PM | Disable Codelens is tip is being displayed | 

## License


ankityyz:
Tested, lgtm!

packages/core/src/codewhisperer/models/constants.ts
```diff
@@ -132,6 +132,8 @@ export const contextPreviewLen = 20
  */
 export const unsupportedLanguagesCacheTTL = 10 * 60 * 60 * 1000
 
+export const isAmazonQVisibleKey = 'AMAZON_Q_CHAT_VISIBLE'
```
justinmk3, line 135: there is no reason for key names to be uppercaese. the ones below are legacy, that isn't the convention going forward. and it doesn't match most of the existing keys.
```suggestion
export const isAmazonQVisibleKey = 'amazonq.chat.visible'
```

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -111,8 +113,10 @@ export class ChatController {
 
         onDidChangeAmazonQVisibility(visible => {
             if (visible) {
+                void globals.context.globalState.update(isAmazonQVisibleKey, true)
                 this.telemetryHelper.recordOpenChat()
             } else {
+                void globals.context.globalState.update(isAmazonQVisibleKey, false)
```
justinmk3, line 119: We have too many globals. This is yet another. Why is this needed? It looks like the state could be stored in this class, instead of globalState. Is the intention to communicate with other instances of vscode at the same time? If so, that should be called out in a comment.

packages/core/src/codewhispererChat/editor/codelens.ts
```diff
@@ -67,6 +69,18 @@ export class TryChatCodeLensProvider implements vscode.CodeLensProvider {
         return new Promise(resolve => {
             token.onCancellationRequested(() => resolve([]))
 
+            /**
+             * Disable Codelens if
+             * - lineAnnotationController is visible (i.e not in end state)
+             * - Amazon Q chat is visible
+             */
+            const isLineAnnotationVisible = globals.context.globalState.get<string>(inlinehintKey)
+            const isAmazonQVisible = globals.context.globalState.get<boolean>(isAmazonQVisibleKey)
```
justinmk3, line 78: instead of another global, this could read `ChatController.isAmazonQVisible()` .

but again, this assumes we don't already have a global for this. Please look for existing similar state. There are so many of these.

packages/core/src/codewhisperer/models/constants.ts
ivikash: Will update in the next revision :)

justinmk3 (approved):

PR #5043, opened 2024-05-22
chore(auth): add login tests

hayemaxi:
Just some basic tests. We could use some more tests that cover the connection re-use functionality.

## License

### DIFF
--- a/packages/core/src/test/login/webview/vue/backend_toolkit.test.ts
+++ b/packages/core/src/test/login/webview/vue/backend_toolkit.test.ts
@@ -55,7 +55,7 @@ describe('Toolkit Login', function () {
         sandbox.restore()
     })
 
-    it('successfully signs into builder ID and emits telemetry', async function () {
+    it('signs into builder ID and emits telemetry', async function () {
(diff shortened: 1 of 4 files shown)


packages/core/src/test/login/webview/vue/backend_toolkit.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+        assertTelemetry('auth_addConnection', {
+            result: 'Succeeded',
+            credentialSourceId: 'awsId',
+            authEnabledFeatures: 'codecatalyst',
+            isReAuth: false,
+        })
+    })
+
+    it('successfully signs into account IdC and emits telemetry', async function () {
```
justinmk3, line 75: "successfully" is usually redundant in test names. It would be unusual to test that something failed (and then we would say "fails" in the name)
```suggestion
    it('signs into account IdC and emits telemetry', async function () {
```

justinmk3 (approved):
Clearly structured with minimal indirection. Nice!

PR #5044, opened 2024-05-23
Amazon Q Code Transform: Add telemetry helper functions for common API telemetry.

arangatang:
## Problem
1. Telemetry emitted for api calls + uploads and downloads differ between different api calls despite all using the `codeTransform_logApiLatency` and `codeTransform_logApiError` metrics. 
2. `codeTransform_logApiLatency` only emitted on success but not also on failure
3. downloadExportResultArchive - Inconsistent telemetry depending on who invoked
4. Lots of duplicate code for telemetry emission

## Solution

Add `TelemetryHelper` which offers `callApi` function that wraps an API call with relevant telemetry for CodeTransform APIs.  Download / Upload calls are not supported as the structure of their responses differ. 
`TelemetryHelper` is also a first step towards:

1. Avoiding direct access to TelemetryState from core business logic (seperation of concerns)
2. Code cleanup - gets rid of code duplication due to telemetry code

## License

### DIFF
--- a/packages/core/src/shared/utilities/download.ts
+++ b/packages/core/src/shared/utilities/download.ts
@@ -7,7 +7,6 @@ import path from 'path'
 import { CodeWhispererStreaming, ExportResultArchiveCommandInput } from '@amzn/codewhisperer-streaming'
 import { ToolkitError } from '../errors'
 import { CodeTransformTelemetryState } from '../../amazonqGumby/telemetry/codeTransformTelemetryState'
-import { transformByQState } from '../../codewhisperer/models/model'
 import { calculateTotalLatency } from '../../amazonqGumby/telemetry/codeTransformTelemetry'
 import { telemetry } from '../telemetry/telemetry'
 import { fsCommon } from '../../srcShared/fs'
@@ -55,7 +54,7 @@ export async function downloadExportResultArchive(
         telemetry.codeTransform_logApiError.emit({
             codeTransformApiNames: 'ExportResultArchive',
             codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId(),
-            codeTransformJobId: transformByQState.getJobId(),
+            codeTransformJobId: exportResultArchiveArgs.exportId,
             codeTransformApiErrorMessage: downloadErrorMessage,
             codeTransformRequestId: e.requestId ?? '',
             result: MetadataResult.Fail,
@@ -66,7 +65,7 @@ export async function downloadExportResultArchive(
         telemetry.codeTransform_logApiLatency.emit({
             codeTransformApiNames: 'ExportResultArchive',
             codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId(),
-            codeTransformJobId: transformByQState.getJobId(),
+            codeTransformJobId: exportResultArchiveArgs.exportId,
             codeTransformRunTimeLatency: calculateTotalLatency(apiStartTime),
             codeTransformTotalByteSize: totalDownloadBytes,
             codeTransformRequestId: result?.$metadata.requestId,


packages/core/src/shared/utilities/download.ts
```diff
@@ -28,31 +30,46 @@ export async function downloadExportResultArchive(
...
+        throw e
+    } finally {
+        telemetry.codeTransform_logApiLatency.emit({
+            codeTransformApiNames: 'ExportResultArchive',
+            codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId(),
+            codeTransformJobId: transformByQState.getJobId(),
+            codeTransformRunTimeLatency: calculateTotalLatency(apiStartTime),
+            codeTransformTotalByteSize: totalDownloadBytes,
+            codeTransformRequestId: result?.$metadata.requestId,
+        })
```
arangatang, line 73: Note: Ideally this should be part of the TelemetryHelper through some function like
`callStreamingApi` however, this will be added in a seperate PR to avoid making this larger.

packages/core/src/amazonqGumby/telemetry/codeTransformTelemetry.ts
```diff
@@ -90,3 +105,79 @@ export const javapOutputToTelemetryValue = (javapCommandLineOutput: string) => {
...
+    }
+
+    static async callApi<T extends CodeTransformAPIresponse>(parameters: {
+        apiCall: () => Promise<PromiseResult<T, AWSError>>
+        apiName: CodeTransformApiNames
+        errorReason: string
+        uploadId?: string
+        jobId?: string
+        uploadFileByteSize?: number
+        setJobFailureMetadata: boolean
```
tincheng, line 129: We can default this to `true`? I'm not sure why GetTransformation was the only operation with setJobFailureMetadata set to false, but I think it can also benefit from the metadata (requestId)

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -760,7 +756,6 @@ export async function stopTransformByQ(
                         void submitFeedback(placeholder, CodeWhispererConstants.amazonQFeedbackKey)
                     }
                 })
-            getLogger().error(`CodeTransformation: Error stopping transformation ${err}`)
```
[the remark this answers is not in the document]
arangatang: Removed it since we will be logging any api error in the `stopJob` call as part of the changes in codeTransformTelemetry line 142.

packages/core/src/amazonqGumby/telemetry/codeTransformTelemetry.ts
```diff
@@ -90,3 +105,79 @@ export const javapOutputToTelemetryValue = (javapCommandLineOutput: string) => {
...
+            if (parameters.setJobFailureMetadata && result?.$response.requestId !== undefined) {
+                transformByQState.setJobFailureMetadata(` (request ID: ${result.$response.requestId})`)
+            }
+            return result
+        } catch (e: any) {
+            let errorMessage = (e as Error).message
+            getLogger().error(`CodeTransformation: ${parameters.apiName} error: = ${errorMessage}`)
+            if (parameters.shouldAttachJobFailureMetadataOnError) {
+                errorMessage += ` -- ${transformByQState.getJobFailureMetadata()}`
+            }
```
tincheng, line 145: Follow up with the above comment regarding "setJobFailureMetadata". Non-blocking, but looks like the failureMetadata is only used for appending the requestId to the errorMessage for telemetry (not even logging). However, the telemetry already have `codeTransformRequestId` field, so I don't think failureMetadata is needed and can be removed. For reference, IntelliJ doesn't append requestId to the end of error message.

tincheng (approved):

packages/core/src/shared/utilities/download.ts
```diff
@@ -28,31 +30,46 @@ export async function downloadExportResultArchive(
...
             }
         }
+        await fsCommon.writeFile(toPath, Buffer.concat(buffer))
+    } catch (e: any) {
+        const downloadErrorMessage = (e as Error).message
+        getLogger().error(`CodeTransformation: ExportResultArchive error = ${downloadErrorMessage}`)
+        telemetry.codeTransform_logApiError.emit({
+            codeTransformApiNames: 'ExportResultArchive',
+            codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId(),
+            codeTransformJobId: transformByQState.getJobId(),
```
JuliaABurch, line 58: This gets passed in as part of `exportResultArchiveArgs` (we set `exportId` to `jobId`) -- can we use this? Anywhere we can try to move away from globals would be great, especially as we want to support simultaneous job submissions in the future

arangatang: Yes, done.

packages/core/src/amazonqGumby/telemetry/codeTransformTelemetry.ts
arangatang: Yeah you are probably right, if we get this in I can remove this in a follow up pr tomorrow?

arangatang:
Failures are unrelated to my changes:

#### Unix failures
```
2462 passing (2m)
  10 pending
  1 failing
  1) Auth
       promptForConnection
         reauthenticates a connection if the user selects an expired one:

      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert_1.default.ok(!refreshedConnItem.description?.match(/expired/i))

      + expected - actual

      -false
      +true
      
      at /Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/core/src/test/credentials/auth.test.ts:512:24
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async /Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/core/src/test/shared/vscode/window.ts:501:29
```

#### Windows failures
```
rejected promise not handled within 1 second: Error: ToolkitGlobals accessed before initialize()
stack trace: Error: ToolkitGlobals accessed before initialize()
    at Object.get (d:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\core\src\shared\extensionGlobals.ts:107:23)
```

packages/core/src/amazonqGumby/telemetry/codeTransformTelemetry.ts
```diff
@@ -90,3 +105,79 @@ export const javapOutputToTelemetryValue = (javapCommandLineOutput: string) => {
...
+                reason: parameters.errorReason,
+            })
+            throw e // pass along error to callee
+        } finally {
+            let uploadId = parameters.uploadId
+            try {
+                uploadId = (result as CreateUploadUrlResponse).uploadId
+            } catch (_) {
+                //noop
+            }
```
justinmk3, line 162: try/catch is not necessary here. Just use `.?` operator.
that also allows use of `const`.

```suggestion
            const uploadId = (result as CreateUploadUrlResponse)?.uploadId ?? parameters.uploadId
```

packages/core/src/amazonqGumby/telemetry/codeTransformTelemetry.ts
```diff
@@ -90,3 +105,79 @@ export const javapOutputToTelemetryValue = (javapCommandLineOutput: string) => {
...
+
+            let requestId = telemetryUndefined
+            try {
+                requestId = (result as PromiseResult<CodeTransformAPIresponse, AWSError>).$response.requestId
+            } catch (_) {
+                // noop
+            }
```
justinmk3, line 169: try/catch not necessary, and discouraged. use `getRequestId` from errors.ts.

packages/core/src/amazonqGumby/telemetry/codeTransformTelemetry.ts
```diff
@@ -90,3 +105,79 @@ export const javapOutputToTelemetryValue = (javapCommandLineOutput: string) => {
...
+
+            telemetry.codeTransform_logApiLatency.emit({
+                codeTransformApiNames: parameters.apiName,
+                codeTransformSessionId: this.sessionId,
+                codeTransformUploadId: uploadId,
+                codeTransformRunTimeLatency: calculateTotalLatency(apiStartTime),
+                codeTransformJobId: parameters.jobId,
+                codeTransformTotalByteSize: parameters.uploadFileByteSize,
+                codeTransformRequestId: requestId,
```
justinmk3, line 178: there is a standard `requestId` field already. any reason not to use it?

justinmk3:
Glad to see this effort to reduce code duplication, but it's really unfortunate that gumby telemetry is designed this way. This could be avoided by wrapping your logic in ` telemetry.codeTransform_xx.run()`, which automatically sets many standard metric fields (and more will be added soon, such as HTTP status code): https://github.com/aws/aws-toolkit-vscode/blob/93396ccf07732ab5fc1e60822d6097fdcb98513d/packages/core/src/shared/telemetry/spans.ts#L151-L164

Unfortunately, because gumby has not modeled telemetry to represent actions, and instead has "latency" and "error" metrics, this doesn't quite fit semantically.

PR #5046, opened 2024-05-23
web: setup web mode for Q inline suggestions

nkomonen-amazon:
## Problem
With inline suggestions moved to the Q extension we need to get the Q extension working in Web mode, including the debug mode to run Web locally.

## Solution
This PR is mainly refactors, see individual commits for each specific change.

The refactors to the code were so that Amazon Q can now be run in Web mode. This PR will have the following limitations:
- You can only use Builder ID to sign in
- It will not work in vscode.dev due to auth/cors limitations (temporary issue)
- You must run things locally with `Extension (web) (amazonq` in the Run + Debug menu
- Local storage is not setup, so your login session will not be remembered when testing.

The Timeout class was modified to support it in both Web and Node runtimes. The inline suggestion feature indirectly relied on this so this was necessary to get working in Web mode.

### Additional
As part of the solution this PR creates a doc in `arch_development.md` that explains the new naming structure for code that can only run in certain environments. See that doc for more explanation as this will influence the project structure moving forward.

https://github.com/aws/aws-toolkit-vscode/assets/118216176/bdd3e9a0-e920-476d-a904-941b288e1236

## License

### DIFF
--- a/docs/arch_develop.md
+++ b/docs/arch_develop.md
@@ -45,8 +45,8 @@ Current quirks of the current monorepo status that should be resolved/evaluated 
         an issue if more subprojects are added and the contents of the root and local modules folders differ.
 -   [`globalSetup.test.ts`](../packages/core/src/test/globalSetup.test.ts) should be configured to work as a library/run tests for all subprojects.
 -   Subproject `tsconfig.json`s should extend a root `tsconfig.packages.json`.
--   Linting tests should run at the root level, not subproject level.
 -   `packages/*/scripts/` should be generalized and moved to the root of the project as needed.
+-   Linting tests should run at the root level, not subproject level.
 -   LICENSE, README.md, and other non-code artifacts that must be packaged into the .vsix are currently
     being copied into the packaging subproject directory from the root project directory as part of the `copyFiles` task.
 -   Pre-release only publishes packages/toolkit extension directly. It should be extended to other added extensions. See [`release.yml`](../.github/workflows/release.yml)
@@ -56,7 +56,7 @@ Additional quirks introduced by creating a core library from the original extens
 
 -   Tests are ran from `packages/core/`
 -   Extension runs from `packages/toolkit`
--   Extension tests run from the core lib. Since some of the tests require an extension context/sandbox, we initiate a "fake" extension to run these tests. This is also why there are vscode extensionproperties in the package.json
+-   Extension tests run from the core lib. Since some of the tests require an extension context/sandbox, we initiate a "fake" extension to run these tests. This is also why there are vscode extension properties in the package.json
 -   Some of original extension code (that now lives in `packages/core`) depends on the package.json, specifically the contributes section. This section is very large AND needs to be present in both the core library and toolkit extension package.jsons. The core library code needs access to this section to create types, set up SAM debuggers, etc. The toolkit needs this section during packaging/debugging so that the extension can run in vscode. The short term solution was to creat a [build script](../packages/toolkit/scripts/build/handlePackageJson.ts) to copy necessary fields over to the toolkit extension during packaging and debugging.
 
 ### Contributes and Settings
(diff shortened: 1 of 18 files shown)


docs/arch_develop.md
```diff
@@ -72,6 +72,39 @@ Some components of the core library depend on the `package.json`s of the extensi
     -   Does not restore, it is a superset of what exists in `packages/core` for `configuration.properties`.
     -   To develop for the Amazon Q extension: add all changes to `packages/amazonq/package.json`, EXCEPT for settings that are references by code in the core library, or settings that already exist in the core `package.json`
 
+## Shared vs Common names
+
+In this repo, the keywords **"shared"** and **"common"** have specific meanings in the context of file/folder names.
+
+### "common"
+
+Code within a folder/file that has the "common" keyword implies that it can run in any environment. Examples of environments are: Web mode, or Node.js (local desktop)
```
justinmk3, line 81: node-only or web-only code would live in a specific place (`.../node/...` or `.../web/...`), right? So everything else can be assumed to be "common". Thus we may not need the "common" name at all.

The tricky part is the transitional phase. Currently we don't have a clear dilineation between anything. So names like `extensionCommon.ts` are an attempt to bridge that transition. But because 80% of all code should be "common", adding "common" everywhere seems much more noisy than adding "node" or "web" to specific parts that are intended to be platform-specific.

nkomonen-amazon: Mentioned below:
> If there is no "shared" or "common" naming used for the file/folder, then assume it only works in Node.js.

I think it is easiest to consider everything as Node.js specific in this transitional phase. Then as we verify code is `common` we can move that code in to the appropriate `common` folder. This will allow us to defer verifying all our code is common, and instead do it incrementally.

So for now this would essentially look like:
- create a `common` subfolder in the folder you are working in and any existing or new code that you know is `common` should be moved in to there as part of your PR
- Anything outside the `common` subfolder is assumed Node.js only

---

Eventually most of our code will be `common`, and by then we can update the structure to be the reverse (anything not defined is assumed as `common`).

justinmk3: > This will allow us to defer verifying all our code is common, and instead do it incrementally.

But that implies we would (eventually) move, say 80% of all code a "common" folder. That's a big lift.

> Eventually most of our code will be common, and by then we can update the structure to be the reverse

And that would again be a big (inverse) lift.

It's a much smaller lift to instead identify things that are truly node-only or web-only. This means we assume that everything is "universal" until proven otherwise. As we discover non-universal pockets, we move only those particular parts to "node" or "web" folders.

packages/core/src/codewhisperer/ui/codeWhispererNodes.ts
```diff
@@ -216,9 +217,19 @@ export function createSignIn(): DataQuickPickItem<'signIn'> {
...
     const icon = getIcon('vscode-account')
 
+    let onClick = () => {
+        void focusAmazonQPanel.execute(placeholder, 'codewhispererQuickPick')
+    }
+    if (isWeb()) {
+        // TODO: nkomonen, call a Command instead
+        onClick = () => {
+            void AuthUtil.instance.connectToAwsBuilderId()
+        }
```
justinmk3, line 227: could be a ternary, which would allow `const` (and less code)

packages/amazonq/src/extension.ts
```diff
@@ -4,12 +4,78 @@
...
+    await activateAmazonQCommon(context, false)
+    await activateAmazonQNonCommon(context)
+}
+
+/**
+ * The code in this function is not common, implying it only works in Node.js and not web.
+ * The goal should be for this to not exist and that all code is "common". So if possible make
+ * the code compatible with web and move it to {@link activateAmazonQCommon}.
+ */
+async function activateAmazonQNonCommon(context: vscode.ExtensionContext) {
```
justinmk3, line 28: "NonCommon" means "node only" right? "Nodejs" is probably a less awkward name
```suggestion
async function activateAmazonQNodejs(context: vscode.ExtensionContext) {
```

justinmk3 (approved):

PR #5049, opened 2024-05-23
lint: catch incorrect usages of once(), fix current usages

hayemaxi:
Problem: `once()` can be used incorrectly such that code can be executed multiple times, even if it seems that that the developer just wanted it to run a single time.

Solution: Custom lint rule to try to catch these cases. It is not foolproof, but it should catch most issues.

## License

### DIFF
--- a/packages/core/src/awsexplorer/activationShared.ts
+++ b/packages/core/src/awsexplorer/activationShared.ts
@@ -7,7 +7,6 @@ import * as vscode from 'vscode'
 import { createToolView, ToolView } from './toolView'
 import { telemetry } from '../shared/telemetry/telemetry'
 import { CdkRootNode } from '../cdk/explorer/rootNode'
-import { once } from '../shared/utilities/functionUtils'
 
 /**
  * Activates vscode Views (eg tree view) that work in any vscode environment (nodejs or browser).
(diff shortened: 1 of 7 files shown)


packages/core/src/awsexplorer/activationShared.ts
```diff
@@ -25,8 +24,7 @@ export function registerToolView(viewNode: ToolView, context: vscode.ExtensionCo
     toolView.onDidExpandElement(e => {
         if (e.element.resource instanceof CdkRootNode) {
             // Legacy CDK metric, remove this when we add something generic
-            const recordExpandCdkOnce = once(() => telemetry.cdk_appExpanded.emit())
```
justinmk3, line 28: Future idea: I wonder if we could use a [weakmap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) or similar so that once() would be easier to use.

plugins/eslint-plugin-aws-toolkits/lib/rules/no-incorrect-once-usage.ts
```diff
@@ -0,0 +1,171 @@
...
+            // Check for calls of once (e.g. once()()) and assignment of once (e.g. const x = once())
+            CallExpression(node) {
+                if (
+                    node.callee.type !== AST_NODE_TYPES.Identifier || // check for any func()
+                    node.callee.name !== 'once' // check for once() and not myFunc()
+                ) {
+                    return
+                }
+
+                // Check for once()()
```
justinmk3, line 89: Very helpful comment :)

justinmk3 (approved):
Nice!

PR #5054, opened 2024-05-24
lint: update @typescript-eslint, fix warnings

justinmk3:
closes https://github.com/aws/aws-toolkit-vscode/pull/5034

## License


plugins/eslint-plugin-aws-toolkits/lib/rules/no-await-on-vscode-msg.ts
```diff
@@ -37,7 +36,7 @@ export default ESLintUtils.RuleCreator.withoutDocs({
     meta: {
         docs: {
             description: 'disallow awaiting on vscode notifications if the response is unused',
-            recommended: 'error',
+            recommended: 'recommended',
```
justinmk3, line 39: 'error' is not a valid value here, apparently

packages/core/src/applicationcomposer/messageHandlers/emitTelemetryMessageHandler.ts
```diff
@@ -26,43 +26,43 @@ export function emitTelemetryMessageHandler(message: EmitTelemetryMessage) {
     try {
         const parsedData = message.metadata ? JSON.parse(message.metadata) : {}
         switch (message.eventType) {
-            case 'GENERATE_CLICKED':
+            case TelemetryType.GENERATE_CLICKED:
```
justinmk3, line 29: This is another small disadvantage of enums. If a `type` was used, the literal value could be used instead.

nkomonen-amazon: Would it be better now to switch it from `enum` in to a `type` instead? It looks safe to do

nkomonen-amazon (approved):

PR #5055, opened 2024-05-24
fix(featureDev): File Rejection stopped working

dianaabv16:
## Problem
when clicking on the X to reject change nothing happens

Fix was introduced to Mynah in 4.9.1 version.

On our side we needed to update mynah chat card update by messageId.

Tested:
Screenshot 2024-05-24 at 14 47 51

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/session/session.ts
+++ b/packages/core/src/amazonqFeatureDev/session/session.ts
@@ -166,8 +166,13 @@ export class Session {
         return resp.interaction
     }
 
-    public async updateFilesPaths(tabID: string, filePaths: NewFileInfo[], deletedFiles: DeletedFileInfo[]) {
-        this.messenger.updateFileComponent(tabID, filePaths, deletedFiles)
+    public async updateFilesPaths(
+        tabID: string,
+        filePaths: NewFileInfo[],
+        deletedFiles: DeletedFileInfo[],
+        messageId: string
+    ) {
+        this.messenger.updateFileComponent(tabID, filePaths, deletedFiles, messageId)
     }
 
     public async insertChanges() {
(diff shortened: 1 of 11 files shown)


dogusata (approved):

sannicm (approved):

packages/core/src/amazonqFeatureDev/session/session.ts
```diff
@@ -166,8 +166,13 @@ export class Session {
         return resp.interaction
     }
 
-    public async updateFilesPaths(tabID: string, filePaths: NewFileInfo[], deletedFiles: DeletedFileInfo[]) {
-        this.messenger.updateFileComponent(tabID, filePaths, deletedFiles)
+    public async updateFilesPaths(
+        tabID: string,
+        filePaths: NewFileInfo[],
+        deletedFiles: DeletedFileInfo[],
+        messageId: string
```
hayemaxi, line 173: There's now 4 complex args for this function, and multiple other functions in this file use the same set of args.
To help with readability, consider passing a typed args object to each function with the same set of arguments instead, e.g.

```
type UpdateFileOpts {
  tabId: string;
  filePaths: NewFileInfo[];
  deletedFiles: DeletedFileInfo[];
  messageId: string;
}

public async updateFilesPaths(opts: UpdateFileOpts) {
   ...
}

updateFilesPaths({
  tabId: this.tabId,
  filePaths,
  deletedFiles,
  messageId: this.messageId
})
```

dianaabv16: if its ok I will add it as a follow up.

jpinkney-aws:
Can you add a changelog item for this

dianaabv16:
Latest commit was to only update changelog.
Before that all checks were green

justinmk3:
Rebasing on latest master will fix CI.

justinmk3:
Ah, I see the problem. Unrelated to this PR

PR #5061, opened 2024-05-28
fix(amazonq): show more specific error messages

dhasani23:
## Problem

Need to show the user more specific error messages for various failure cases during a transformation.

## Solution

Handle specific errors with specific messaging, text already approved by our technical writer.

## License


packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -187,10 +187,8 @@ export async function startTransformByQ() {
     } catch (error: any) {
         await transformationJobErrorHandler(error)
     } finally {
-        if (transformByQState.isCancelled()) {
-            await postTransformationJob()
-            await cleanupTransformationJob()
-        }
+        await postTransformationJob()
+        await cleanupTransformationJob()
```
dhasani23, line 191: These should always be running since they reset the Transformation Hub and other things to prepare for a subsequent transformation run.

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -384,11 +384,7 @@ export class Messenger {
         this.dispatcher.sendCommandMessage(new SendCommandMessage(message.command, message.tabID, message.eventId))
     }
 
-    public sendJobFinishedMessage(tabID: string, message: string = '') {
-        if (message === '') {
```
dhasani23, line 388: This never happened.

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -178,16 +182,12 @@ export async function resumeTransformationJob(jobId: string, userActionStatus: T
...
+        throw new Error(errorMessage)
     }
 }
 
 export async function stopJob(jobId: string) {
     if (!jobId) {
-        throw new Error('Job ID is empty')
-    }
-
-    if (transformByQState.isNotStarted()) {
```
dhasani23, line 190: Duplicate check; already handled where we call `stopJob`

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -178,16 +182,12 @@ export async function resumeTransformationJob(jobId: string, userActionStatus: T
...
             reason: 'ResumeTransformationFailed',
         })
-        throw new Error('Resume transformation job failed')
+        throw new Error(errorMessage)
     }
 }
 
 export async function stopJob(jobId: string) {
     if (!jobId) {
-        throw new Error('Job ID is empty')
```
dhasani23, line 187: Don't throw an error since that gets surfaced to customer; just avoid calling `stopJob` if no `jobId` has been generated yet.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -479,16 +484,28 @@ export async function startTransformationJob(uploadId: string) {
...
-        if ((error as Error).message.includes('too many active running jobs')) {
+        const errorMessage = (error as Error).message
+        if (errorMessage.includes('too many active running jobs')) {
             transformByQState.setJobFailureErrorNotification(
                 CodeWhispererConstants.failedToStartJobTooManyJobsNotification
             )
             transformByQState.setJobFailureErrorChatMessage(
                 CodeWhispererConstants.failedToStartJobTooManyJobsChatMessage
             )
+        } else if (errorMessage.includes('Lines of Code limit breached')) {
```
JuliaABurch, line 495: Would it be better to create custom Error types and switch on the type rather than looking at strings? If strings get updated in one place but not everywhere, then these lines of code break

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -479,16 +484,28 @@ export async function startTransformationJob(uploadId: string) {
         getLogger().info(`CodeTransformation: jobId: ${jobId}`)
     } catch (error) {
         getLogger().error(`CodeTransformation: ${CodeWhispererConstants.failedToStartJobNotification}`, error)
-        if ((error as Error).message.includes('too many active running jobs')) {
+        const errorMessage = (error as Error).message
+        if (errorMessage.includes('too many active running jobs')) {
```
JuliaABurch, line 488: It would be nice if we could centralize all of this error handling (setting the notification / chat messages) in one class, as right now this functionality gets scattered across several different files / classes

packages/core/src/test/codewhisperer/commands/transformByQ.test.ts
```diff
@@ -131,25 +131,18 @@ describe('transformByQ', function () {
...
     })
 
     it('WHEN stop job called with valid jobId THEN stop API called', async function () {
-        transformByQState.setToRunning()
         const stopJobStub = sinon.stub(codeWhisperer.codeWhispererClient, 'codeModernizerStopCodeTransformation')
         await stopJob('dummyId')
         sinon.assert.calledWithExactly(stopJobStub, { transformationJobId: 'dummyId' })
     })
 
-    it('WHEN stop job that has not been started THEN stop API not called', async function () {
```
JuliaABurch, line 147: Do we not need this edge case / is it tested elsewhere?

dhasani23: No longer needed since in `stopJob` I removed a check that checks if the job is marked as "running" or not, since that was a redundant check as it's handled elsewhere.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
dhasani23: We'd still have to check the Strings though right? This String comes directly from backend.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
dhasani23: Yeah agreed, from UX we soon will have to modify this so that error messages only show in notification if the chat is not visible to user, when that work is done I think will be a good time to refactor how we set the error messages.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
tincheng: finalizeTransformByQ (down the chain of humanInTheLoopRetryLogic for happy path) will make duplicated call to postTransformationJob and cleanupTransformationJob -> can this cause extra messages to be published to chat or extra telemetry to be emitted? 

```
export async function finalizeTransformByQ(status: string) {
    try {
        // Set the result state variables for our store and the UI
        await finalizeTransformationJob(status)
    } catch (error: any) {
        await transformationJobErrorHandler(error)
    } finally {
        await postTransformationJob()
        await cleanupTransformationJob()
    }
}
```
I think we can remove the postTransformationJob() and cleanupTransformationJob() from finalizeTransformByQ

packages/core/src/codewhisperer/models/constants.ts
```diff
@@ -451,37 +454,47 @@ export const unsupportedJavaVersionChatMessage =
...
 
 export const failedToStartJobTooManyJobsChatMessage =
     'Sorry, I couldn’t begin the transformation. You have too many active transformations running. Please try again after your other transformations have completed.'
 
 export const failedToStartJobTooManyJobsNotification =
-    "Amazon Q couldn't begin the transformation. You have too many active transformations running. Please try again after your other transformations have completed."
+    'Amazon Q could not begin the transformation. You have too many active transformations running. Please try again after your other transformations have completed.'
+
+export const failedToStartJobLinesLimitNotification =
+    'Amazon Q could not begin the transformation. Your code exceeds the lines of code limit. Please try again with a smaller project.'
```
tincheng, line 469: There are currently two scenarios for failedToStartJobLinesLimitNotification:
1. "Job Lines of Code limit breached" - single project exceeded line limit.
2. "Monthly aggregated Lines of Code limit breached" - completed jobs in the current month exceeded aggregated line limit across jobs.

This message always assume case 1.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
justinmk3: there are some useful patterns (and util functions) in `packages/core/src/shared/errors.ts`

packages/core/src/codewhisperer/commands/startTransformByQ.ts
damntrecky: Not a blocker: I think @JuliaABurch is asking about semantic readability and maybe a small separation of concerns. The code might read better for developers if it was something like

`else if (error === LinesOfCodeLimitBreachedError)` and you abstract the "logic" of the string checking in the error types we create.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -479,16 +481,35 @@ export async function startTransformationJob(uploadId: string) {
...
             transformByQState.setJobFailureErrorChatMessage(
                 CodeWhispererConstants.failedToStartJobTooManyJobsChatMessage
             )
+        } else if (errorMessage.includes('Monthly aggregated Lines of Code limit breached')) {
+            transformByQState.setJobFailureErrorNotification(
+                CodeWhispererConstants.failedToStartJobMonthlyLimitNotification
+            )
+            transformByQState.setJobFailureErrorChatMessage(
+                CodeWhispererConstants.failedToStartJobMonthlyLimitChatMessage
+            )
```
tincheng, line 498: Sorry didn't catch this last time, but Lines of Code limit is checked during the transformation build phase and will be returned as a transformation failure reason, instead of a validation error from start job. These should be handled during transformation polling.

PR #5063, opened 2024-05-28
refactor(commands): clean up unnecessary commands

hayemaxi:
This commit aims to reduce ambiguity with commands. It is hard to search for commands and see how they are used because they are dynamically created, e.g. `aws.${toolkit | amazonq}.auth.reauthenticate`. This is a leftover artifact of splitting the extension into packages/toolkit and packages/amazonq.

- Remove dynamic auth commands, Amazon Q did not use most of them.
  - Remove unused commands on Amazon Q side.
  - Hardcode 'toolkit' in the used commands, and move them to another file for later use.
- The few commands that amazon Q used are now registered in packages/amazonq
- Remove 'aws.toolkit.logout' because it is the same as 'aws.toolkit.auth.signout'
- Add `getOrThrow` util function to `Commands`.

Follow-up:
- We should register the toolkit commands in packages/toolkit, but the core lib needs them to run tests. So once we drop tests down to the subpackages we can also move command registrations.

## License


justinmk3 (approved):
Nice approach

packages/amazonq/src/extensionShared.ts
```diff
@@ -103,7 +103,10 @@ export async function activateShared(context: vscode.ExtensionContext, isWeb: bo
     await activateQGumby(extContext as ExtContext)
 
     // Generic extension commands
-    registerCommands(context, contextPrefix)
+    registerGenericCommands(context, contextPrefix)
+
+    // Amazon Q specific commands
+    registerCommands(context)
```
ivikash, line 109: @hayemaxi, I see you are removing `registerCommands` from the imports in line number 30 and adding registerGenericCommands. Is this expected? cc: @justinmk3

hayemaxi: Addressed in https://github.com/aws/aws-toolkit-vscode/pull/5080

Looks like we have identified a gap... our CI should be catching this. Nice catch, thanks!

PR #5067, opened 2024-05-29
tests: move feature dev e2e tests to amazon q extension folder

jpinkney-aws:
## Problem
- feature dev e2e tests don't work anymore because the required registrations happen in the amazon q extension and not the core/toolkit extensions

## Solution
- move feature dev e2e tests to the amazon q folder
- parameterize globalSetup so that it can be used by multiple extensions
- export test and featureDev modules

## Other notes
- this doesn't fix the `Error: DialogService: refused to show dialog in tests. Contents: Do you want Code to open the external website?` issues in the e2e tests
- this makes it so that the tests run locally again

## License


packages/amazonq/src/testE2E/index.ts
justinmk3, line 1: If it's not much trouble, maybe `test*/` dirs should be sibling-to `src/`.

justinmk3 (approved):

jpinkney-aws:
Depends on https://github.com/aws/aws-toolkit-vscode/pull/5069 since the packaging fails to find the amazonq-webview.css

PR #5069, opened 2024-05-29
tech debt: import q css inside of the webview directly

jpinkney-aws:
## Problem
- the current approach required us to override the module loader which isn't ideal

## Solution
- add the css into the webview directly rather than inside of the javascript. That way we seperate where style sheets are loaded for q and where javascript is loaded

## License


hayemaxi (approved):

justinmk3 (approved):
Nice

justinmk3:
Is this something we need to document somewhere?

jpinkney-aws:
I'm in the process of adding a couple lines to the arch_develop docs. Turns out it was actually a solved issue in another place in the codebase: https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/views/securityPanelViewProvider.ts#L52 that I didn't know about

PR #5072, opened 2024-05-29
Revert "fix(featureDev): File Rejection stopped working #5055"

dianaabv16:
This reverts commit 03e5e294360b56e4cf9cd2c112120336f3c99230.

## Problem
- After bumping mynah UI and making changes to onFileComponentUpdate. The plan disappears after clicking generate Code.
This is a critical functionality for us to have.

## Tested
Yes, after reverting this change plan stays, but it re-introduces bug when reverting file changes.
Screenshot 2024-05-29 at 19 37 02

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3335,9 +3335,9 @@
             }
         },
         "node_modules/@aws/mynah-ui": {
-            "version": "4.9.1",
-            "resolved": "https://registry.npmjs.org/@aws/mynah-ui/-/mynah-ui-4.9.1.tgz",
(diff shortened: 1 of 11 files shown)


vishaalmehrishi (approved):

jpinkney-aws (approved):
I think ideally we would try to fix the bug rather than reverting, especially since these changes aren't releases yet AFAIK

justinmk3 (changes requested):
As discussed offline, the reverted commit wasn't released yet. Next step is to try to fix potentially incorrect usage of mynah-ui instead of rolling back.

PR #5076, opened 2024-05-30
fix(/dev): conversation id log message is now homogeneous and similar to the customer facing text

sannicm:
## Problem

The log lines for helping the customer get its conversation ID where not homogeneous between the plan and code generation phase, as well as with the ui chat message.

## Solution

Now all logs will have the same format `featureName Conversation ID: conversationId`. Note that Conversation ID text is now similar to the ui chat message.

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/userFacingText.ts
+++ b/packages/core/src/amazonqFeatureDev/userFacingText.ts
@@ -5,11 +5,12 @@
 
 import { manageAccessGuideURL } from '../amazonq/webview/ui/texts/constants'
 import { userGuideURL } from '../amazonq/webview/ui/texts/constants'
+import { featureName } from './constants'
 
 export const examples = `
 You can use /dev to:
 - Add a new feature or logic
-- Write tests 
+- Write tests
 - Fix a bug in your project
 - Generate a README for a file, folder, or project
 
@@ -21,3 +22,8 @@ export const updateCode = 'Code has been updated. Would you like to work on anot
 export const sessionClosed = 'Your session is now closed.'
 export const newTaskChanges = 'What change would you like to make?'
 export const uploadCodeError = `Amazon Q is unable to upload workspace artifacts to S3 for feature development. For more information, see the [Amazon Q documentation](${manageAccessGuideURL}) or contact your network or organization administrator.`
+
+// Utils for logging and showing customer facing conversation id text
+export const messageWithConversationId = (conversationId?: string) =>
+    conversationId ? `\n\nConversation ID: **${conversationId}**` : ''
+export const logWithConversationId = (conversationId: string) => `${featureName} Conversation ID: ${conversationId}`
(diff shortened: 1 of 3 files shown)


packages/core/src/amazonqFeatureDev/userFacingText.ts
```diff
@@ -21,3 +21,9 @@ export const updateCode = 'Code has been updated. Would you like to work on anot
 export const sessionClosed = 'Your session is now closed.'
 export const newTaskChanges = 'What change would you like to make?'
 export const uploadCodeError = `Amazon Q is unable to upload workspace artifacts to S3 for feature development. For more information, see the [Amazon Q documentation](${manageAccessGuideURL}) or contact your network or organization administrator.`
+
+// Utils for logging and showing customer facing conversation id text
+export const conversationIdChat = (conversationId?: string) =>
+    conversationId ? `\n\nConversation ID: **${conversationId}**` : ''
+export const conversationIdLog = (featureName: string = '', conversationId: string) =>
```
jpinkney-aws, line 28: Is there ever a case where we would pass in featureName? Isn't there a constant that just defines featureName for us?

packages/core/src/amazonqFeatureDev/userFacingText.ts
```diff
@@ -21,3 +21,9 @@ export const updateCode = 'Code has been updated. Would you like to work on anot
 export const sessionClosed = 'Your session is now closed.'
 export const newTaskChanges = 'What change would you like to make?'
 export const uploadCodeError = `Amazon Q is unable to upload workspace artifacts to S3 for feature development. For more information, see the [Amazon Q documentation](${manageAccessGuideURL}) or contact your network or organization administrator.`
+
+// Utils for logging and showing customer facing conversation id text
+export const conversationIdChat = (conversationId?: string) =>
```
jpinkney-aws, line 26: I'm trying to think if theres a slightly better way we can name these functions. Maybe `conversationIdChat -> conversationIdLog` (since its just logging the conversation id) and then `conversationIdLog -> featureDevLog` WDYT?

sannicm: mmh, it is a good question, to me it make sense because they point to the exact place they will be loged, one in the log the other one in the chat. Maybe something like: `conversationIdToChatMessage` and `conversationIdToToolkitLogs`

packages/core/src/amazonqFeatureDev/userFacingText.ts
sannicm: Good point, we can avoid the featureName passing as it is just a constant, I can hardcode that.

packages/core/src/amazonqFeatureDev/userFacingText.ts
jpinkney-aws: The hard part for me reading it is i wouldn't exactly when to use them. It didn't click for me that one was in the chat message and one was in the output log
maybe:
`logWithConversationId` and `messageWithConversationId` ?

sannicm: That would work, sure, I will update.

jpinkney-aws (approved):

PR #5080, opened 2024-05-30
fix: registerCommands not imported in amazon q

hayemaxi:
## License


jpinkney-aws (approved):
I'm curious whats causing our CI to not show this build error. Are we not using ci for amazon q?

hayemaxi:
> I'm curious whats causing our CI to not show this build error. Are we not using ci for amazon q?

At the very least packaging should have caught this. Some simple research shows that its the loader we are using in webpack, which is `esbuild-loader`. I tested with `ts-loader` and it correctly exits on this error.

PR #1586, opened 2021-03-19
build: Migrate ts-loader to esbuild-loader

JadenSimon:
## Description
Replaced ts-loader and Terser with esbuild-loader to see if we get any sort of increase in build-speed. This PR requires the extension to be transpiled to es2018 instead of es2015 (ES6) because esbuild does not support async generators and async iterators. Changing the build target shouldn't affect anything since VSCode appears to support it. 

Note that webpack is still being used. Esbuild-loader merely acts as a wrapper for esbuild so it can be used with webpack. Webpack could be removed entirely but it would require a much larger change in the build process.

Three unique warnings currently appear (not sure if they appeared before, will have to look into this...)

`WARNING in ./src/lambda/config/templates.ts 114:9-34
"export 'ParseErrorCode' (imported as 'jsonParser') was not found in 'jsonc-parser'
 @ ./src/shared/sam/debugger/commands/addSamDebugConfiguration.ts
 @ ./src/shared/sam/activation.ts
 @ ./src/extension.ts
`

`WARNING in ./node_modules/typescript/lib/typescript.js 2710:20-42
Critical dependency: the request of a dependency is an expression
 @ ./src/shared/typescriptLambdaHandlerSearch.ts
 @ ./src/shared/codelens/typescriptCodeLensProvider.ts
 @ ./src/shared/sam/activation.ts
 @ ./src/extension.ts
`

`
WARNING in ./node_modules/adm-zip/util/fileSystem.js
Module not found: Error: Can't resolve 'original-fs' in '/Users/sijaden/VSCode/aws-toolkit-vscode/node_modules/adm-zip/util'
 @ ./node_modules/adm-zip/util/fileSystem.js
 @ ./node_modules/adm-zip/util/index.js
 @ ./node_modules/adm-zip/adm-zip.js
 @ ./src/eventSchemas/commands/downloadSchemaItemCode.ts
 @ ./src/eventSchemas/activation.ts
 @ ./src/extension.ts
`

## Checklist

- [x] I have read the **README** document
- [x] I have read the **CONTRIBUTING** document
- [x] My code follows the code style of this project
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed
- [ ] A short description of the change has been added to the changelog using the script `npm run newChange`

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5084, opened 2024-05-30
ci: add compile checking to subprojects

hayemaxi:
Problem: Subprojects (not core/) are not compiled normally, and webpack uses `esbuild-loader`. This results in no typescript compilation checks during build and packaging.
Note: by design `esbuild-loader` does not offer this, see https://github.com/privatenumber/esbuild-loader?tab=readme-ov-file#type-checking

Solution: Add a type check to buildScripts.

I did not notice any significant slow down, but if this is noticed later, we can instead use this recommended dependency https://github.com/TypeStrong/fork-ts-checker-webpack-plugin

## License

### DIFF
--- a/packages/amazonq/package.json
+++ b/packages/amazonq/package.json
@@ -46,7 +46,7 @@
     "browser": "./dist/src/extensionWeb",
     "scripts": {
         "vscode:prepublish": "npm run clean && npm run buildScripts && webpack --mode production",
-        "buildScripts": "npm run generateNonCodeFiles && npm run copyFiles && npm run syncPackageJson",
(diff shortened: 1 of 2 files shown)


jpinkney-aws (approved):

nkomonen-amazon (approved):

justinmk3:
>  why are we using esbuild-loader as opposed to just ts-loader ? https://github.com/aws/aws-toolkit-vscode/blob/5a713ad8b7685601461cec98b25b59508acb6e56/packages/webpack.base.config.js#L77
>
> esbuild-loader doesnt have typescript compile checks, so CI will never catch compile errors


see https://github.com/aws/aws-toolkit-vscode/pull/1586 for context.
IIRC we also wanted to eliminate webpack and use only esbuild, but that was blocked at the time

justinmk3 (approved):

PR #5086, opened 2024-05-31
Enabled start, stop, and reboot EC2 instance commands

bwfinley:
## Problem
Certain EC2 commands were only available in development mode.

## Solution
Make whether the commands appear or not, not dependent on development mode.

## License


justinmk3:
> not dependent on development mode.

EC2 is currently intentionally hidden behind dev mode.

PR #5087, opened 2024-05-31
lint: disallow console.log and similar

hayemaxi:
- Add lint rule to detect logging with console
- Fix cases where this occurs
- Add exception for cases where required, e.g.
    - logger.ts
    - any front end code
    - any language servers
    - tests
    - build scripts
- Refactor logger to take only a dependency on `vscode` and nothing else. This is done by moving perfLogger to its own file.
    - This also adds a dependency to globals to have logger
 

## License


jpinkney-aws (approved):

packages/core/src/shared/logger/logger.ts
```diff
@@ -3,8 +3,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+// Disable because this is the main logger file.
+/* eslint-disable aws-toolkits/no-console-log */
```
justinmk3, line 7: can we put this in the `ConsoleLogger` scope? shouldn't be needed for the rest of this file.

packages/core/src/shared/sam/sync.ts
```diff
@@ -686,6 +686,8 @@ class ProcessTerminal implements vscode.Pseudoterminal {
     public constructor(private readonly process: ChildProcess) {
         // Used in integration tests
         if (isAutomation()) {
+            // Disable because it is a test.
+            // eslint-disable-next-line aws-toolkits/no-console-log
             this.onDidWrite(text => console.log(text.trim()))
```
justinmk3, line 691: This is fine for specific, intentionally cases. But it's good to have the rule generally enabled for tests.

packages/core/src/webviews/client.ts
```diff
@@ -187,6 +187,8 @@ export class WebviewClientFactory {
                 },
                 get: (_, prop) => {
                     if (typeof prop !== 'string') {
+                        // Disable because we cannot log to extension in a webview.
+                        // eslint-disable-next-line aws-toolkits/no-console-log
                         console.warn(`Tried to index webview client with non-string property: ${String(prop)}`)
```
justinmk3, line 192: Can we forward logs to the backend? Do we have a webview base class that could have a `log()` function?

PR #5090, opened 2024-05-31
fix(login): pre-select auth option for the user

ivikash:
## Problem
Users have to manually select one of the auth options for Login

## Solution
In order to reduce the friction and one click, we pre-select the first option by default. 

**Note**
- In case of existing logins, we still select the default option as we are in process of separating out auth.

## License

### DIFF
--- a/packages/core/src/login/webview/vue/login.vue
+++ b/packages/core/src/login/webview/vue/login.vue
@@ -366,13 +366,16 @@ export default defineComponent({
         await this.emitUpdate('created')
     },
 
-    mounted() {
+    async mounted() {
         this.fetchRegions()
-        void this.updateExistingConnections()
+        await this.updateExistingConnections()
 
         // Reset gathered telemetry data each time we view the login page.
         // The webview panel is reset on each view of the login page by design.
-        void client.resetStoredMetricMetadata()
+        client.resetStoredMetricMetadata()
+
+        // Pre-select the first available login option
+        await this.preselectLoginOption()
     },
     methods: {
         toggleItemSelection(itemId: number) {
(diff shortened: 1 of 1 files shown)


ivikash:
| Toolkit (Default) | Q (Default) | Q (Re-use connection) |
| -- | -- | --- |
| Screenshot 2024-05-31 at 10 11 36 AM | Screenshot 2024-05-31 at 10 11 27 AM | Screenshot 2024-05-31 at 10 12 28 AM |

jpinkney-aws:
Just out of curiosity, wouldn't it make more sense to pre-select the existing connection in q if it already exists? rather than the free tier one?

hayemaxi:
> Just out of curiosity, wouldn't it make more sense to pre-select the existing connection in q if it already exists? rather than the free tier one?

Yes, but since we are splitting sessions this no longer applies

hayemaxi (approved):

hayemaxi:
> > Just out of curiosity, wouldn't it make more sense to pre-select the existing connection in q if it already exists? rather than the free tier one?
> 
> Yes, but since we are splitting sessions this no longer applies

Actually, we should wait until these details are finalized.

ivikash:
Screenshot 2024-06-11 at 11 19 00 AM
Screenshot 2024-06-11 at 11 18 29 AM
Screenshot 2024-06-11 at 11 17 53 AM
Screenshot 2024-06-11 at 11 15 26 AM
Screenshot 2024-06-11 at 11 13 46 AM
Screenshot 2024-06-11 at 11 13 28 AM

packages/core/src/login/webview/vue/login.vue
```diff
@@ -366,13 +366,16 @@ export default defineComponent({
...
 
-    mounted() {
+    async mounted() {
         this.fetchRegions()
-        void this.updateExistingConnections()
+        await this.updateExistingConnections()
 
         // Reset gathered telemetry data each time we view the login page.
         // The webview panel is reset on each view of the login page by design.
         void client.resetStoredMetricMetadata()
```
justinmk3, line 375: should this also be await'd ?

justinmk3 (approved):

hayemaxi (approved):

packages/core/src/login/webview/vue/login.vue
ivikash: `resetStoredMetricMetadata` is not an async function so await won't be required.

https://github.com/ivikash/aws-toolkit-vscode/blob/45ae5d97bd8c6f5aa876f94ce394048e5be25797/packages/core/src/login/webview/vue/backend.ts#L224-L226

justinmk3: then why are we calling it with `void` ?

ivikash: Thats right, it shouldn't have void. I am updating it now :).

PR #5097, opened 2024-06-03
feat(amazonq): display notification for sign in

ivikash:
## Problem
- Prompt user to sign-in into Amazon Q if they close the chat and haven't logged in. Display the notification only once. 
- Prompt the user with help link if they close the chat and were still authenticating. 

## Solution
Display a notification to the user prompting for Sign In int Amazon Q.

User will see the notification if both the following conditions are met (i.e AND condition)
- User has manually closed the Amazon Q Sidepanel (this is necessary because by default we assume the sidebar is open).
- User has never attempted auth (aka clicked on the final continue button that opened the browser either for builder id or for idc connection).

## License

### DIFF
--- a/packages/core/src/amazonq/activation.ts
+++ b/packages/core/src/amazonq/activation.ts
@@ -14,9 +14,15 @@ import { activateBadge } from './util/viewBadgeHandler'
 import { amazonQHelpUrl } from '../shared/constants'
 import { focusAmazonQChatWalkthrough, openAmazonQWalkthrough } from './onboardingPage/walkthrough'
 import { listCodeWhispererCommandsWalkthrough } from '../codewhisperer/ui/statusBarMenu'
-import { Commands } from '../shared/vscode/commands2'
+import { Commands, placeholder } from '../shared/vscode/commands2'
 import { focusAmazonQPanel, focusAmazonQPanelKeybinding } from '../codewhispererChat/commands/registerCommands'
 import { TryChatCodeLensProvider, tryChatCodeLensCommand } from '../codewhispererChat/editor/codelens'
+import { Auth } from '../auth'
+import { learnMoreUri } from '../codewhisperer/models/constants'
+import { openUrl } from '../shared/utilities/vsCodeUtils'
+import { AuthUtil } from '../codewhisperer'
+import { ConnectionStateChangeEvent } from '../auth/auth'
+import { telemetry } from '../shared/telemetry'
 
 export async function activate(context: ExtensionContext) {
     const appInitContext = DefaultAmazonQAppInitContext.instance
@@ -51,10 +57,101 @@ export async function activate(context: ExtensionContext) {
     })
 
     await activateBadge()
+    void setupAuthNotification(
+        appInitContext.onDidChangeAmazonQVisibility.event,
+        Auth.instance.onDidChangeConnectionState
+    )
 }
 
 function registerApps(appInitContext: AmazonQAppInitContext) {
     cwChatAppInit(appInitContext)
     featureDevChatAppInit(appInitContext)
     gumbyChatAppInit(appInitContext)
+}
+
+/**
+ * Display a notification to user for Log In.
+ *
+ * Authentication Notification is displayed when:
+ * - The user closes the Amazon Q chat panel, and
+ * - The user has not performed any authentication action.
+ *
+ * Error Notification is displayed when:
+ * - The user closes the Amazon Q chat panel, and
+ * - The user attempts an authentication action but is not logged in.
+ *
+ * @param {Event} onAmazonQChatVisibility - Event indicating the visibility status of the Amazon Q chat.
+ * @param {Event} onDidUpdateConnection - Event indicating the authentication connection update.
+ */
+async function setupAuthNotification(
+    onAmazonQChatVisibility: vscode.Event<boolean>,
+    onDidUpdateConnection: vscode.Event<ConnectionStateChangeEvent | undefined>
+) {
+    let isAmazonQVisible = true // Assume Chat is open by default.
+    let notificationDisplayed = false // Auth Notification should be displayed only once.
+    let authConnection: ConnectionStateChangeEvent
+
+    // Updates the visibility state of the Amazon Q chat.
+    const updateVisibility = async (visible: boolean) => {
+        isAmazonQVisible = visible
+        await tryShowNotification()
+    }
+
+    // Updates the source of the connection for Amazon Q sign in.
+    const updateConnection = async (connection: ConnectionStateChangeEvent | undefined) => {
+        if (connection) {
+            authConnection = connection
+            await tryShowNotification()
+        }
+    }
+
+    const disposables: vscode.Disposable[] = [
+        onAmazonQChatVisibility(updateVisibility),
+        onDidUpdateConnection(updateConnection),
+    ]
+
+    async function tryShowNotification() {
+        if (notificationDisplayed || Auth.instance.activeConnection) {
+            return
+        }
+
+        const source = 'authNotification'
+
+        if (!isAmazonQVisible && !authConnection && !AuthUtil.instance.isConnectionExpired()) {
+            const buttonAction = 'Sign In'
+            notificationDisplayed = true
+            disposables.forEach(item => item.dispose())
+
+            telemetry.toolkit_showNotification.emit({
+                component: 'editor',
+                id: source,
+                reason: 'notLoggedIn',
+                result: 'Succeeded',
+            })
+            const selection = await vscode.window.showWarningMessage('Start using Amazon Q', buttonAction)
+
+            if (selection === buttonAction) {
+                void focusAmazonQPanel.execute(placeholder, source)
+            }
+        } else if (!isAmazonQVisible && authConnection.state === 'authenticating') {
+            const buttonAction = 'Open documentation'
+            notificationDisplayed = true
+            disposables.forEach(item => item.dispose())
+
+            telemetry.toolkit_showNotification.emit({
+                component: 'editor',
+                id: source,
+                reason: 'authenticating',
+                result: 'Succeeded',
+            })
+            const selection = await vscode.window.showWarningMessage(
+                'See Amazon Q documentation for help on signing in',
+                buttonAction
+            )
+
+            if (selection === buttonAction) {
+                void openUrl(vscode.Uri.parse(`${learnMoreUri}#q-in-IDE-setup-bid`), source)
+            }
+        }
+    }
 }
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -114,7 +114,7 @@ function keyedDebounce<T, U extends any[], K extends string = string>(
     }
 }
 
-interface ConnectionStateChangeEvent {
+export interface ConnectionStateChangeEvent {
     readonly id: Connection['id']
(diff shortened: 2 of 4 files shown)


packages/core/src/login/webview/commonAuthViewProvider.ts
```diff
@@ -87,6 +87,8 @@ export class CommonAuthViewProvider implements WebviewViewProvider {
         // This will fire whenever the user opens or closes the login page from 'somewhere else'
         // i.e. NOT when switching from/to the chat window, which uses the same view area.
         webviewView.onDidChangeVisibility(async () => {
+            this.onDidChangeVisibility?.fire(webviewView.visible)
```
justinmk3, line 90: there is another case on line 117, which means this would fire twice. is line 117 still needed?

packages/core/src/auth/auth.ts
```diff
@@ -135,6 +135,9 @@ export class Auth implements AuthService, ConnectionManager {
     /** Fired when a connection and its metadata has been completely deleted */
     public readonly onDidDeleteConnection = this.#onDidDeleteConnection.event
 
+    // User has attempted auth
+    public readonly onDidInitiateConnection = new vscode.EventEmitter<string | undefined>()
```
justinmk3, line 139: This doesn't look right. This is adding a special-case without considering the semantics of this interface. This adds a new event but it's being controlled by other (amazonq) code . So listening to this from toolkit code would do nothing, which is very misleading.

Why can't this be captured by this existing `onDidChangeConnectionState` event, or one of the other (3!) existing "change" or "update" events?

justinmk3: Also it doesn't look like there are actually any listeners to this event, so... what is its purpose?

packages/core/src/auth/auth.ts
```diff
@@ -282,6 +282,8 @@ export class Auth implements AuthService, ConnectionManager {
         }
 
         const id = randomUUID()
+        void this.updateConnectionState(id, 'authenticating')
```
justinmk3, line 285: 'authenticating' is also the state in `authenticate()`:
https://github.com/aws/aws-toolkit-vscode/blob/0bbfa1320b7a073ad688f246058fe171c5a51be4/packages/core/src/auth/auth.ts#L723

are the semantics the same. is it correct to always set 'authenticating'  whenever `createConnection()` is called?


also directly below we're passing 'unauthenticated' ...

packages/core/src/amazonq/activation.ts
```diff
@@ -17,6 +17,72 @@ import { listCodeWhispererCommandsWalkthrough } from '../codewhisperer/ui/status
 import { Commands } from '../shared/vscode/commands2'
 import { focusAmazonQPanel, focusAmazonQPanelKeybinding } from '../codewhispererChat/commands/registerCommands'
 import { TryChatCodeLensProvider, tryChatCodeLensCommand } from '../codewhispererChat/editor/codelens'
+import { Auth } from '../auth'
+import { customLearnMoreUri } from '../codewhisperer/models/constants'
+import { openUrl } from '../shared/utilities/vsCodeUtils'
+import { AuthUtil } from '../codewhisperer'
+import { ConnectionStateChangeEvent } from '../auth/auth'
+
+async function showAuthNotification(
```
justinmk3, line 26: a brief docstring explaining the purpose of this will be helpful. there are tons of "show auth notification" places in the codebase, what's different about this one? not obvious from the code

packages/core/src/amazonq/activation.ts
```diff
@@ -17,6 +17,72 @@ import { listCodeWhispererCommandsWalkthrough } from '../codewhisperer/ui/status
...
+        if (notificationDisplayed || Auth.instance.activeConnection) {
+            return
+        }
+
+        const source = 'authNotification'
+
+        if (!isAmazonQVisible && !authConnection && !AuthUtil.instance.isConnectionExpired()) {
+            const buttonAction = 'Sign In'
+            notificationDisplayed = true
+            disposables.map(item => item.dispose())
```
justinmk3, line 63: forEach or each() is clearer if this is only done for side effects. same for the other case(s) in this PR
```suggestion
            disposables.each(item => item.dispose())
```

packages/core/src/auth/auth.ts
ivikash: > is it correct to always set 'authenticating' whenever createConnection() is called?
tbh, I think this is the right approach as we should store intermediate state of `unauthenticated`.

The `unauthenticated` state is stored in `tokenProvider` though is not updated in state.

Looking at the function `updateConnectionState` --
https://github.com/aws/aws-toolkit-vscode/blob/75d0d159c8597559763cb97a61a9e8e36768336c/packages/core/src/auth/auth.ts#L507C5-L512C10

was storing the `unauthenticated` value because `this.store.getProfileOrThrow(id)` throw error and have an early return.

packages/core/src/auth/auth.ts
```diff
@@ -506,6 +508,12 @@ export class Auth implements AuthService, ConnectionManager {
 
     private async updateConnectionState(id: Connection['id'], connectionState: ProfileMetadata['connectionState']) {
         getLogger().info(`auth: Updating connection state of ${id} to ${connectionState}`)
+
+        if (connectionState === 'authenticating') {
+            this.#onDidChangeConnectionState.fire({ id, state: connectionState })
+            return
+        }
```
justinmk3, line 515: This seems like a special case but perhaps it's harmless.

nkomonen-amazon: This looks like we are changing behavior. Now if the state is set as `authenticating` we will not store it in the metadata using `this.store.updateMetadata()` below.

What is the reason?

packages/core/src/amazonq/activation.ts
```diff
@@ -17,6 +17,86 @@ import { listCodeWhispererCommandsWalkthrough } from '../codewhisperer/ui/status
...
+ *
+ * Error Notification is displayed when:
+ * - The user closes the Amazon Q chat panel, and
+ * - The user attempts an authentication action but is not logged in.
+ *
+ * @param {Event} isAmazonQVisibleEvent - Event indicating the visibility status of the Amazon Q chat.
+ * @param {Event} onDidUpdateConnectionEvent - Event indicating the authentication connection update.
+ */
+async function showAuthNotification(
+    isAmazonQVisibleEvent: vscode.Event<boolean>,
```
nkomonen-amazon, line 41: We should use the `on***` variable naming for events to keep things consistent.

Possible solution:
```suggestion
    onAmazonQChatVisibility: vscode.Event<boolean>,
```

packages/core/src/amazonq/activation.ts
```diff
@@ -17,6 +17,86 @@ import { listCodeWhispererCommandsWalkthrough } from '../codewhisperer/ui/status
...
+    onDidUpdateConnectionEvent: vscode.Event<ConnectionStateChangeEvent | undefined>
```
nkomonen-amazon, line 42: The `on***` implies it is an event instance in our code base

```suggestion
    onDidUpdateConnection: vscode.Event<ConnectionStateChangeEvent | undefined>
```

packages/core/src/amazonq/activation.ts
```diff
@@ -17,6 +17,86 @@ import { listCodeWhispererCommandsWalkthrough } from '../codewhisperer/ui/status
...
+            await showNotification()
+        }
+    }
+
+    const disposables: vscode.Disposable[] = [
+        isAmazonQVisibleEvent(updateVisibility),
+        onDidUpdateConnectionEvent(updateConnection),
+    ]
+
+    async function showNotification() {
```
nkomonen-amazon, line 67: Since this only shows the notification on certain conditions we should use `try***` to imply that

```suggestion
    async function tryShowNotification() {
```

packages/core/src/amazonq/activation.ts
```diff
@@ -17,6 +17,86 @@ import { listCodeWhispererCommandsWalkthrough } from '../codewhisperer/ui/status
...
+ * - The user has not performed any authentication action.
+ *
+ * Error Notification is displayed when:
+ * - The user closes the Amazon Q chat panel, and
+ * - The user attempts an authentication action but is not logged in.
+ *
+ * @param {Event} isAmazonQVisibleEvent - Event indicating the visibility status of the Amazon Q chat.
+ * @param {Event} onDidUpdateConnectionEvent - Event indicating the authentication connection update.
+ */
+async function showAuthNotification(
```
nkomonen-amazon, line 40: Since we are only setting things up to show the notification, but conditions must be met for it to actually show

```suggestion
async function setupAuthNotification(
```

packages/core/src/amazonq/activation.ts
```diff
@@ -17,6 +17,86 @@ import { listCodeWhispererCommandsWalkthrough } from '../codewhisperer/ui/status
...
+
+        if (!isAmazonQVisible && !authConnection && !AuthUtil.instance.isConnectionExpired()) {
+            const buttonAction = 'Sign In'
+            notificationDisplayed = true
+            disposables.forEach(item => item.dispose())
+
+            const selection = await vscode.window.showWarningMessage('Start using Amazon Q', buttonAction)
+
+            if (selection === buttonAction) {
+                void vscode.commands.executeCommand('aws.amazonq.focusChat', 'placeholder**', source)
```
nkomonen-amazon, line 82: Can you use the actual export instead of the `placeholder**` string

nkomonen-amazon: Can you instead import `focusAmazonQPanel` instead of `vscode.commands.executeCommand`. We do this since `focusAmazonQPanel` will provide the actual typing for arguments while `vscode.commands.executeCommand` does not

packages/core/src/amazonq/activation.ts
```diff
@@ -17,6 +17,86 @@ import { listCodeWhispererCommandsWalkthrough } from '../codewhisperer/ui/status
...
+        }
+    }
+
+    const disposables: vscode.Disposable[] = [
+        isAmazonQVisibleEvent(updateVisibility),
+        onDidUpdateConnectionEvent(updateConnection),
+    ]
+
+    async function showNotification() {
+        if (notificationDisplayed || Auth.instance.activeConnection) {
```
nkomonen-amazon, line 68: This looks like after the first time we show a notification `notificationDisplayed` is true, and we will not show the message again.

But when the user reloads VS Code do we intended to show the message again, essentially once every session?

packages/core/src/auth/auth.ts
nkomonen-amazon: We should avoid touching `auth.ts` unless required. Some more info on why this was needed will help.

Since the connection does not exist yet, I'm wondering if it makes sense to update the state. No one knows about this new connection at this point.

packages/core/src/amazonq/activation.ts
ivikash: Yes, we intend to display the notification once every session. cc: @ankityyz

packages/core/src/amazonq/activation.ts
ivikash: Done.

packages/core/src/auth/auth.ts
ivikash: The idea is to store the intermediate step so that we can display a notification if they attempt to do an auth and still close the chat window.

nkomonen-amazon (approved):

PR #5098, opened 2024-06-04
feat(featureDev): generated plan being shown from top

dianaabv16:
## Problem 
feat(featureDev): add snapToTop prop for geneterated plan

when a WB response comes out, we don’t auto scroll all the way to the bottom or at all.
So once the Plan is generated the customer will see the the top of the plan instead of the end of it
## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
+++ b/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
@@ -348,6 +348,7 @@ export class FeatureDevController {
             type: 'answer',
             tabID: tabID,
             canBeVoted: true,
+            snapToTop: true,
         })
 
         if (interactions.responseType === 'VALID') {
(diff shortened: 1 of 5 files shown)


packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -348,6 +348,7 @@ export class FeatureDevController {
             type: 'answer',
             tabID: tabID,
             canBeVoted: true,
+            snapToTop: true,
```
kumsmrit, line 351: Is this only required for plan generation?

dianaabv16: yes, plan is the only "long" card we have. With code generation the card is normally "fits" the scrren

kumsmrit (approved):

jpinkney-aws (approved):

PR #5100, opened 2024-06-04
tests: add script for running amazon q e2e tests

jpinkney-aws:
## Problem
- Running `npm run testE2E` from the root doesn't run the amazon q e2e folders

## Solution
- Add a `testE2E` script

## License


nkomonen-amazon (approved):

packages/amazonq/package.json
```diff
@@ -52,11 +52,14 @@
         "syncPackageJson": "ts-node ./scripts/build/syncPackageJson.ts",
         "clean": "ts-node ../../scripts/clean.ts dist/ LICENSE NOTICE",
         "compile": "npm run clean && npm run buildScripts && webpack --mode development",
+        "compileOnly": "tsc -p ./",
         "package": "ts-node ../../scripts/package.ts",
         "format": "prettier --ignore-path ../../.prettierignore --check src scripts",
         "formatfix": "prettier --ignore-path ../../.prettierignore --write src scripts",
         "lint": "echo 'Nothing to lint yet!'",
         "watch": "npm run clean && npm run buildScripts && tsc -watch -p ./",
+        "testCompile": "npm run clean && npm run buildScripts && npm run compileOnly",
```
justinmk3, line 61: Why does testCompile not use "compile", why does it want "compileOnly"?

packages/core/scripts/test/testE2E.ts
```diff
@@ -5,5 +5,9 @@
 
 import { runToolkitTests } from './launchTestUtilities'
 void (async () => {
-    await runToolkitTests('e2e', 'dist/src/testE2E/index.js')
+    const relativeEntrypoint = process.argv[2]
+    if (!relativeEntrypoint) {
+        throw new Error('Relative entrypoint is required')
```
justinmk3, line 10: Is "relative" mentioned as in "relative path"? An example in the error would also be a good hint.

    throw new Error('relative path to entrypoint is required (example: dist/src/testE2E/index.js)')

PR #5104, opened 2024-06-04
build: remove --vueHr flag

hayemaxi:
Problem: arg should go to `npm run copyFiles` in `buildScripts`, but it isn't and hasn't been. Now with the addition it of `tsc -p ./ --noEmit` it is going to that instead and it's an unrecognized arg.

Solution: remove it

Not sure if the arg ever worked. We can revisit this if the feature is desirable.

## License

### DIFF
--- a/packages/amazonq/package.json
+++ b/packages/amazonq/package.json
@@ -56,7 +56,7 @@
         "format": "prettier --ignore-path ../../.prettierignore --check src scripts",
         "formatfix": "prettier --ignore-path ../../.prettierignore --write src scripts",
         "lint": "echo 'Nothing to lint yet!'",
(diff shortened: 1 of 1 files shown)


ivikash:
Should we also remove the following?

https://github.com/aws/aws-toolkit-vscode/blob/780c42719341ec047f2c064ed6877cfed5bf2972/packages/amazonq/scripts/build/copyFiles.ts#L88-L91

hayemaxi:
> Should we also remove the following?
> 
> https://github.com/aws/aws-toolkit-vscode/blob/780c42719341ec047f2c064ed6877cfed5bf2972/packages/amazonq/scripts/build/copyFiles.ts#L88-L91

No, if we ever need to add hot reloads back then that code would be a good starting point.

PR #5107, opened 2024-06-05
build(core): kill 8080 port before serving vue files

ivikash:
## Problem
AWS Toolkits and Amazon Q extension expects the web view to be served at `8080` port. When there is a existing process, `vue` process starts at `8081` and so on ports. When this happens, webview doesn't always serve the latest code hampering developer productivity.

## Solution
- Kill the process and free up `8080` port before starting `serveVueCore` process. This ensures `webview` is always served on port 8080.
- `kill-port` in cross platform.
## License


packages/core/package.json
```diff
@@ -4069,6 +4069,7 @@
         "immutable": "^4.3.0",
         "js-yaml": "^4.1.0",
         "jsonc-parser": "^3.2.0",
+        "kill-port": "^2.0.1",
```
justinmk3, line 4072: This should be a devDependency, it's not necessary for the production build.

But not sure this is worth adding a dependency for. Can we choose a different port instead of 8080? Killing the "port" seems wrong. Why isn't the task that we started being killed properly?

PR #5118, opened 2024-06-07
fix(build): package task fails if commit id starts with zero

justinmk3:
# Problem:
package task fails if commit id starts with zero.

    ERROR  Invalid extension version '3.9.0-0238109'

semver [disallows](https://semver.org/#spec-item-9) leading zero for prerelease. This is why `git describe` always adds a "g" prefix, for example.

# Solution:
Prefix the prerelease string with "g", similar to `git describe`.

## License

### DIFF
--- a/scripts/package.ts
+++ b/scripts/package.ts
@@ -99,7 +99,8 @@ function getVersionSuffix(feature: string, debug: boolean): string {
     const debugSuffix = debug ? '-debug' : ''
     const featureSuffix = feature === '' ? '' : `-${feature}`
     const commitId = child_process.execFileSync('git', ['rev-parse', '--short=7', 'HEAD']).toString().trim()
-    const commitSuffix = commitId ? `-${commitId}` : ''
+    // Commit id is prefixed with "g" because "-0abc123" is not a valid semver prerelease, and will cause vsce to fail.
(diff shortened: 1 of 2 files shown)


scripts/package.ts
```diff
@@ -99,7 +99,8 @@ function getVersionSuffix(feature: string, debug: boolean): string {
     const debugSuffix = debug ? '-debug' : ''
     const featureSuffix = feature === '' ? '' : `-${feature}`
     const commitId = child_process.execFileSync('git', ['rev-parse', '--short=7', 'HEAD']).toString().trim()
-    const commitSuffix = commitId ? `-${commitId}` : ''
+    // Commit id is prefixed with "g" because "-0abc123" is not a valid semver prerelease, and will cause vsce to fail.
+    const commitSuffix = commitId ? `-g${commitId}` : ''
```
ivikash, line 103: Should we do `-v` in favor of `version`?

justinmk3: I don't think "v" makes sense because this is a (g)it sha. The (v)ersion is different.

ivikash (approved):

jpinkney-aws (approved):

PR #5126, opened 2024-06-10
fix(amazonq): remember "Pause Auto-Suggestions" after IDE restart

leigaol:
## Problem
IDE restart will auto enable auto trigger.

## Solution
Do not auto turn on auto trigger whenever IDE restarts.

Tested with local builds. 

## License

### DIFF
--- a/packages/amazonq/.changes/next-release/Bug Fix-17572ed1-4601-4554-9144-9eb5143bf0ff.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-17572ed1-4601-4554-9144-9eb5143bf0ff.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Do not turn on auto trigger when IDE reloads"
+	"description": "Amazon Q inline suggestions: remember `Pause Auto-Suggestions` after IDE restart"
 }
--- a/packages/amazonq/.changes/next-release/Bug Fix-17572ed1-4601-4554-9144-9eb5143bf0ff.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-17572ed1-4601-4554-9144-9eb5143bf0ff.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Do not turn on auto trigger when IDE reloads"
+}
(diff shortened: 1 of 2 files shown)


packages/amazonq/src/extensionCommon.ts
```diff
@@ -113,9 +112,6 @@ export async function activateAmazonQCommon(context: vscode.ExtensionContext, is
     // reload webviews
     await vscode.commands.executeCommand('workbench.action.webview.reloadWebviewAction')
 
-    // enable auto suggestions on activation
-    await CodeSuggestionsState.instance.setSuggestionsEnabled(true)
```
Will-ShaoHua, line 117: don't we have cache key-value to use here?
```
const value = get(autoTriggerEnabledKey, extContext)
await CodeSuggestionsState.instance.setSuggestionsEnabled(value)
```

packages/amazonq/.changes/next-release/Bug Fix-17572ed1-4601-4554-9144-9eb5143bf0ff.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Do not turn on auto trigger when IDE reloads"
```
justinmk3, line 3:
```suggestion
	"description": "Amazon Q inline suggestions: remember `Pause Auto-Suggestions` after IDE restart"
```

packages/amazonq/src/extensionCommon.ts
leigaol: There was a logic to reset the AutoTrigger enabled status, was it removed after some refactoring?

leigaol: Based on my testing, without your proposed code, the extension memorizes the previous auto trigger settings without an issue.

leigaol: When the statusBarMenu is rendered, it calls `getAmazonQCodeWhispererNodes`, which calls `CodeSuggestionsState.instance.isSuggestionsEnabled()`, which reads from ` get(autoTriggerEnabledKey, this.#context)`. 

Therefore, below extra code is not needed.
```
const value = get(autoTriggerEnabledKey, extContext)
await CodeSuggestionsState.instance.setSuggestionsEnabled(value)

```

Will-ShaoHua: understood, as long as we are memorizing and using users' previous config then it's fine. we can close this conversation.

PR #5128, opened 2024-06-10
techdebt: remove almost all of the tryRegister commands     

jpinkney-aws:
### Problem:
- tryRegister is a hack to allow settings that get activated in the amazonq extension to run in core
    
### Solution:
- setup the amazonq project to run unit tests
- Add "test" script to amazonq package.json
- Move amazonq, feature dev, gumby, codewhisperer, codewhispererChat unit tests from core to amazonq. This requires us to export any required dependencies from their respective source folders
- remove tryRegister in as many places as possible. See "Other notes" for the reason why we couldn't get rid of them all
    
### Other notes:
- A few test files need to override vscode api in order to function correctly. Those test files have been left in core because the overriding wasn't working from a seperate package. tryRegister remains for a few of those files that are left behind and still need it
- A fallback for timer was needed when moving tests from core -> amazonq. It's unclear why these changes forced that fallback

## License


packages/amazonq/scripts/build/copyFiles.ts
```diff
@@ -32,6 +32,7 @@ const tasks: CopyTask[] = [
 
     { target: path.join('../core', 'resources'), destination: path.join('..', 'resources') },
     { target: path.join('../core', 'package.nls.json'), destination: path.join('..', 'package.nls.json') },
+    { target: path.join('test', 'unit', 'amazonqGumby', 'resources') },
```
justinmk3, line 35:
```suggestion
    { target: 'test/'unit/amazonqGumby/resources' },
```

justinmk3 (approved):

PR #5129, opened 2024-06-10
Security scans: VSC triggers to calculate codeScanProjectBytes only for Project scans

laileni-aws:
## Problem
- VSC triggers `getDirSize()` function to calculate `codeScanProjectBytes` for telemetry for both `Auto-scans` and `Project scans`.
## Solution

- Disable the `codeScanProjectBytes` calculation for `Auto-scans`: Since the `codeScanProjectBytes`
 parameter is optional for `Auto-scans` and is affecting the performance, it makes sense to disable the calculation for this scenario. This will help improve the overall performance of the Auto-scans.

- Enable the `codeScanProjectBytes` calculation for `Project scans`: For `Project scans`, you should continue to calculate the `codeScanProjectBytes` using the `getDirSize()` function, as this information is likely valuable for the telemetry.

## License

### DIFF
--- a/packages/core/src/codewhisperer/commands/startSecurityScan.ts
+++ b/packages/core/src/codewhisperer/commands/startSecurityScan.ts
@@ -255,7 +255,7 @@ export async function startSecurityScan(
         codeScanState.setToNotStarted()
         codeScanTelemetryEntry.duration = performance.now() - codeScanStartTime
         codeScanTelemetryEntry.codeScanServiceInvocationsDuration = performance.now() - serviceInvocationStartTime
-        await emitCodeScanTelemetry(codeScanTelemetryEntry)
+        await emitCodeScanTelemetry(codeScanTelemetryEntry, scope)
     }
 }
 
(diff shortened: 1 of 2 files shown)


leigaol:
In the PR statement, please elaborate on `a bit laggy` of `getDirSize`. 

Run it in a repository of 1k files with 10 to 20 folder levels, explain the time it takes to run.

leigaol:
Can you rename this PR ?

packages/core/src/codewhisperer/commands/startSecurityScan.ts
```diff
@@ -255,7 +255,7 @@ export async function startSecurityScan(
         codeScanState.setToNotStarted()
         codeScanTelemetryEntry.duration = performance.now() - codeScanStartTime
         codeScanTelemetryEntry.codeScanServiceInvocationsDuration = performance.now() - serviceInvocationStartTime
-        await emitCodeScanTelemetry(codeScanTelemetryEntry)
+        await emitCodeScanTelemetry(codeScanTelemetryEntry, scope)
```
leigaol, line 258: Does this prevent gitDirSize invocation when it comes to auto security scan? Then how about manual project level security scan?

manmartgarc:
Bumping, really need this 👍

packages/core/src/codewhisperer/commands/startSecurityScan.ts
laileni-aws: Manual scan is triggered once so this should be executed only once, which is exact functionality before 4/30

justinmk3 (approved):
Needs a changelog item. And approval from a CW team member

leigaol:
I can approve once you add a change log item

leigaol (approved):
please add one more change log item

PR #5131, opened 2024-06-10
fix(telemetry): guard for string overflow

damntrecky:
## Problem

These changes are to prevent emitting telemetry items that are too long and is a defensive coding strategy to prevent overflows in the backend.

## Solution

Internal change for telemetry. Does not need to go in release notes.

## License


packages/toolkit/README.quickstart.vscode.md
justinmk3, line 1: this file looks like an accidental commit

packages/core/src/amazonqGumby/telemetry/codeTransformMetadata.ts
```diff
@@ -0,0 +1,35 @@
...
+export const codeTransformMaxMetaDataLength = 65536
+
+export interface ICodeTransformMetaData {
+    dependencyVersionSelected?: string
+    canceledFromChat?: boolean
+    retryCount?: number
+    errorMessage?: string
+}
+type ICodeTransformMetaDataKeys = keyof ICodeTransformMetaData
+export const toJsonString = (metaData: ICodeTransformMetaData) => {
```
justinmk3, line 15: codebase doesn't use arrow functions for most declarations

packages/core/src/amazonqGumby/telemetry/codeTransformMetadata.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    let currentLength = 0
+    const objectKeys = Object.keys(metaData) as ICodeTransformMetaDataKeys[]
+    for (const key of objectKeys) {
+        const value = metaData[key]
+        // add 5 for quotes and comma around key-value pairs
+        const elementLength = key.length + JSON.stringify(value).length + 5
+        if (currentLength + elementLength <= codeTransformMaxMetaDataLength) {
+            delete metaData[key]
+            currentLength += elementLength
```
justinmk3, line 30: the `truncateProps()` function may be useful https://github.com/aws/aws-toolkit-vscode/blob/c8c57b1961df9bf982d466b888549a4fc8459f56/packages/core/src/shared/utilities/textUtilities.ts#L59

damntrecky: Used `partialClone` as we are removing the key for now. Not truncating it.

If the behavior changes in the future I'm leaving a comment for someone to follow up on.

damntrecky: I actually kept the delete behavior. The omission of the key entirely will tell us if something was deleted and overflowed. Adding logging to users client as well.

packages/core/src/amazonqGumby/telemetry/codeTransformMetadata.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    let currentLength = 0
+    const metaDataCopy = { ...metaData }
+    const objectKeys = Object.keys(metaData) as ICodeTransformMetaDataKeys[]
+    for (const key of objectKeys) {
+        const value = metaData[key]
+        // add 5 for quotes and comma around key-value pairs
+        const elementLength = key.length + JSON.stringify(value).length + 5
+        if (currentLength + elementLength >= codeTransformMaxMetaDataLength) {
+            delete metaDataCopy[key]
```
justinmk3, line 41: This could be an optional behavior of `truncateProps`

packages/core/src/amazonqGumby/telemetry/codeTransformMetadata.ts
justinmk3: > The omission of the key entirely will tell us if something was deleted and overflowed.

setting the value to something like "overflow" might be useful

PR #5132, opened 2024-06-11
fix(amazonq): handle proxy-related error cases

dhasani23:
## Problem

We want to give users more helpful error message for proxy-related errors they may experience.

## Solution

Show users pre-approved text when these proxy-related errors happen.

## License

### DIFF
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -267,7 +267,7 @@ export class GumbyController {
                 this.messenger.sendJobFinishedMessage(message.tabID, CodeWhispererConstants.jobCancelledChatMessage)
                 break
             case ButtonActions.VIEW_TRANSFORMATION_HUB:
-                await vscode.commands.executeCommand(GumbyCommands.FOCUS_TRANSFORMATION_HUB)
+                await vscode.commands.executeCommand(GumbyCommands.FOCUS_TRANSFORMATION_HUB, CancelActionPositions.Chat)
                 this.messenger.sendJobSubmittedMessage(message.tabID)
                 break
             case ButtonActions.STOP_TRANSFORMATION_JOB:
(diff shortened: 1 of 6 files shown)


packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -435,10 +435,8 @@ export class GumbyController {
             this.messenger.sendKnownErrorResponse('no-alternate-dependencies-found', message.tabID)
             await this.continueTransformationWithoutHIL(message)
         } else if (message.error instanceof ModuleUploadError) {
-            this.messenger.sendUnrecoverableErrorResponse('upload-to-s3-failed', message.tabID)
```
dhasani23, line 438: This and the one below were causing duplicate chat messages to show.

tincheng (approved):

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -136,6 +136,10 @@ export async function uploadArtifactToS3(
         let errorMessage = `The upload failed due to: ${(e as Error).message}`
         if (errorMessage.includes('Request has expired')) {
             errorMessage = CodeWhispererConstants.errorUploadingWithExpiredUrl
+        } else if (errorMessage.includes('Failed to establish a socket connection')) {
```
justinmk3, line 139: is there a `(e as Error).code` that can be checked ?

damntrecky: +1 to this.

damntrecky (approved):

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
dhasani23: image
The `code` field doesn't appear to be part of the `Error` interface, and `name` would be too generic.

PR #5140, opened 2024-06-12
Amazon Q Code Transform: Failure Build Log

JuliaABurch:
This change enables customers to view the server-side build logs that are generated for a code transform after a project has been submitted. The build log will be downloaded locally, and the option is given to view it in the chat.

Screenshot 2024-06-12 at 4 28 50 PM

## License

### DIFF
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -460,9 +460,9 @@ export class GumbyController {
             this.messenger.sendKnownErrorResponse('no-alternate-dependencies-found', message.tabID)
             await this.continueTransformationWithoutHIL(message)
         } else if (message.error instanceof ModuleUploadError) {
-            this.messenger.sendUnrecoverableErrorResponse('upload-to-s3-failed', message.tabID)
+            this.resetTransformationChatFlow()
         } else if (message.error instanceof JobStartError) {
-            this.messenger.sendUnrecoverableErrorResponse('job-start-failed', message.tabID)
+            this.resetTransformationChatFlow()
         } else if (message.error instanceof TransformationPreBuildError) {
             this.messenger.sendJobSubmittedMessage(message.tabID, true)
             this.messenger.sendAsyncEventProgress(
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -20,9 +20,11 @@ import {
     compileProject,
     finishHumanInTheLoop,
     getValidCandidateProjects,
+    openBuildLogFile,
     openHilPomFile,
     postTransformationJob,
     processTransformFormInput,
+    resetDebugArtifacts,
     startTransformByQ,
     stopTransformByQ,
     validateCanCompileProject,
@@ -35,6 +37,7 @@ import {
     ModuleUploadError,
     NoJavaProjectsFoundError,
     NoMavenJavaProjectsFoundError,
+    TransformationPreBuildError,
 } from '../../errors'
 import * as CodeWhispererConstants from '../../../codewhisperer/models/constants'
 import MessengerUtils, { ButtonActions, GumbyCommands } from './messenger/messengerUtils'
@@ -263,6 +266,7 @@ export class GumbyController {
 
     private async formActionClicked(message: any) {
         const typedAction = MessengerUtils.stringToEnumValue(ButtonActions, message.action as any)
+        console.log(`typedAction: ${typedAction}`)
         switch (typedAction) {
             case ButtonActions.CONFIRM_TRANSFORMATION_FORM:
                 await this.initiateTransformationOnProject(message)
@@ -280,6 +284,8 @@ export class GumbyController {
                 await cleanupTransformationJob()
                 break
             case ButtonActions.CONFIRM_START_TRANSFORMATION_FLOW:
+                this.resetTransformationChatFlow()
+                await resetDebugArtifacts()
                 this.messenger.sendCommandMessage({ ...message, command: GumbyCommands.CLEAR_CHAT })
                 await this.transformInitiated(message)
                 break
@@ -292,6 +298,10 @@ export class GumbyController {
                 break
             case ButtonActions.OPEN_FILE:
                 await openHilPomFile()
+                break
+            case ButtonActions.OPEN_BUILD_LOG:
+                await openBuildLogFile()
+                this.messenger.sendViewBuildLog(message.tabID)
                 break
         }
     }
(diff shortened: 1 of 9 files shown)


packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -763,6 +763,9 @@ export async function pollTransformationJob(jobId: string, validStates: string[]
                     transformByQState.setJobFailureErrorChatMessage(
                         CodeWhispererConstants.failedToStartJobLinesLimitChatMessage
                     )
+                }
+                if (errorMessage.includes('The requested module could not be built')) {
+                    throw new TransformationPreBuildError(response.transformationJob.reason || '')
```
tincheng, line 768: Ideally we don't want to rely on comparing strings, since this is a one-way door decision that will break if service side changes the error message, or constrain service side from updating the string. Instead, as a short term mitigation, we can attempt to pre-fetch the artifact for all PreBuild failures, then only show the UI (i.e. throw TransformationPreBuildError) if artifact was available from the download attempt.

tincheng (approved):

packages/core/src/test/codewhisperer/commands/transformByQ.test.ts
```diff
@@ -297,3 +297,43 @@ describe('transformByQ', function () {
...
+        await toFile('', preBuildLogFilePath)
+        transformByQState.setPreBuildLogFilePath(preBuildLogFilePath)
+
+        await resetDebugArtifacts()
+
+        assert.strictEqual(fs.existsSync(preBuildLogFilePath), false)
+        assert.strictEqual(transformByQState.getPreBuildLogFilePath(), '')
+    })
+
+    it.skip('should not remove any directory if the pre-build log file path is not set', async () => {
```
dhasani23, line 314: Why skip this one and the below?

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -451,9 +460,18 @@ export class GumbyController {
             this.messenger.sendKnownErrorResponse('no-alternate-dependencies-found', message.tabID)
             await this.continueTransformationWithoutHIL(message)
         } else if (message.error instanceof ModuleUploadError) {
-            this.resetTransformationChatFlow()
+            this.messenger.sendUnrecoverableErrorResponse('upload-to-s3-failed', message.tabID)
```
dhasani23, line 463: In my earlier PR (https://github.com/aws/aws-toolkit-vscode/pull/5132/files#r1633998488), I kept the `resetTransformationChatFlow()` but deleted the `sendUnrecoverableErrorResponse()` calls as they were causing duplicate chat messages to show. What is the behavior with this change?

justinmk3 (approved):

packages/core/src/test/codewhisperer/commands/transformByQ.test.ts
```diff
@@ -297,3 +297,43 @@ describe('transformByQ', function () {
...
+        sinon.assert.notCalled(fsSpy)
+        assert.strictEqual(transformByQState.getPreBuildLogFilePath(), '')
+        sinon.restore()
+    })
+
+    it.skip('should not remove any directory if the pre-build log file does not exist', async () => {
+        const preBuildLogFilePath = 'non/existent/path/to/pre-build.log'
+        transformByQState.setPreBuildLogFilePath(preBuildLogFilePath)
+
+        const fsSpy = sinon.spy(fs, 'rmSync')
```
justinmk3, line 330: This is a brittle test. If gumby changes its implementation (e.g. to use [fsCommon](https://github.com/aws/aws-toolkit-vscode/blob/696823d36fa374928731976c1b7cd1eaba28da82/packages/core/src/srcShared/fs.ts#L32), as is strongly recommended) then these tests fail, even if the behavior is correct. Instead the test should check that the directory is actually deleted, without mocking fs functions.

packages/core/src/amazonqGumby/chat/controller/controller.ts
JuliaABurch: Oof, I see that now. I'll revert this behavior to match your implementation - the intention was that the `controller` and the `messenger` class controls what is displayed in the chat, it's not ideal to have that functionality scattered across multiple classes, but this is refactoring better done in a separate PR with testing

justinmk3 (approved):

packages/core/src/test/codewhisperer/commands/transformByQ.test.ts
JuliaABurch: It's fair to not want tests to check whether specific unwrapped library calls are being made. 

Checking to see that the directory is actually deleted is the implementation of the test I've made (above this one) -- these two tests are checking that a non-existent directory isn't deleted, which is tricky when there's no side-effects to validate. 

After thinking about it, I decided to simplify this test, removing the mock

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -848,36 +847,26 @@ export function findDownloadArtifactStep(transformationSteps: TransformationStep
...
     let downloadErrorMessage = undefined
     const cwStreamingClient = await createCodeWhispererChatStreamingClient()
+
     try {
         await downloadExportResultArchive(
             cwStreamingClient,
             {
                 exportId: jobId,
                 exportIntent: ExportIntent.TRANSFORMATION,
-                exportContext: {
```
dhasani23, line 869: Just curious, why remove this?

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -848,36 +847,26 @@ export function findDownloadArtifactStep(transformationSteps: TransformationStep
...
-                        downloadArtifactId,
-                        downloadArtifactType: TransformationDownloadArtifactType.CLIENT_INSTRUCTIONS,
-                    },
-                },
             },
             pathToArchive
         )
     } catch (e: any) {
         downloadErrorMessage = (e as Error).message
-        // This allows the customer to retry the download
```
dhasani23, line 880: Can users not retry now by clicking the blue download button / is that why we `throw e` below?

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
JuliaABurch: There's an issue in the proxy service - even though the values inside exportContext are marked nullable, if they're both null (which is the case when there's a prebuild error as `downloadArtifactId` is null), there's a `ValidationException` thrown. Right now, these values aren't needed, so I removed them. 

An alternative would be to optionally add `transformationExportContext` if `downloadArtifactId` isn't null -- I'm fixing up some things, so this will go into my next PR

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
JuliaABurch: The comment wasn't true (the code it was commenting wasn't responsible for allowing the customer to retry the download)

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -763,14 +763,13 @@ export async function pollTransformationJob(jobId: string, validStates: string[]
...
                         CodeWhispererConstants.failedToStartJobLinesLimitChatMessage
                     )
-                } else {
-                    transformByQState.setJobFailureErrorChatMessage(
-                        `${CodeWhispererConstants.failedToCompleteJobGenericChatMessage} ${errorMessage}`
-                    )
-                    transformByQState.setJobFailureErrorNotification(
-                        `${CodeWhispererConstants.failedToCompleteJobGenericNotification} ${errorMessage}`
-                    )
                 }
```
dhasani23, line 766: Why remove the `else`?

PR #5147, opened 2024-06-12
chore: upgrade codewhisperer-streaming client

JuliaABurch:
## Problem
The streaming client needs to be updated to access recent API changes.

## Solution
I followed the instructions [here](https://github.com/aws/aws-toolkit-vscode/blob/master/docs/build.md) to regenerate the client source files. I ran a quick test of basic chat features (Q, Featuredev, Gumby) to ensure stability.

~~Note: the client is 2 months out of date at this point, so quite a few changes have been generated.~~

## License

### DIFF
--- a/packages/core/src/codewhisperer/client/user-service-2.json
+++ b/packages/core/src/codewhisperer/client/user-service-2.json
@@ -1920,7 +1920,7 @@
         },
         "TransformationDownloadArtifactType": {
             "type": "string",
-            "enum": ["ClientInstructions"]
+            "enum": ["ClientInstructions", "Logs"]
         },
         "TransformationDownloadArtifacts": {
             "type": "list",
--- a/src.gen/@amzn/codewhisperer-streaming/README.md
+++ b/src.gen/@amzn/codewhisperer-streaming/README.md
@@ -485,20 +485,6 @@ CreateResolution
 </details>
 <details>
 <summary>
-CreateSession
-</summary>
-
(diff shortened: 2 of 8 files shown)


tincheng (approved):

justinmk3:
1 million added lines and almost no deletions seems wrong. Please look at previous PRs to see which files actually need to be added.

JuliaABurch:
> 1 million added lines and almost no deletions seems wrong.
> 
> Probably the first step, is to delete the existing directory and replace it with the new generated result. That should result in a much more reasonable diff.
> 
> If that's not mentioned in the instructions, please update the instructions.

I've updated the instructions with some error handling so that the clean / deletion happens much more reliably.

src.gen/@amzn/codewhisperer-streaming/package-lock.json
justinmk3, line 1: I don't think `src.gen/@amzn/codewhisperer-streaming/package-lock.json` should be committed? None of our other sub projects have package-lock.json , and it wasn't needed before. 

The locked packages for this sub-project are already reflected in the top-level package-lock.json. It's possible that some changes are missing from that file, in this PR.

justinmk3: I don't see any dependency changes in `src.gen/@amzn/codewhisperer-streaming/package.json`, so the top-level package-lock.json probably doesn't have any changes.

So all that's needed is to remove `src.gen/@amzn/codewhisperer-streaming/package-lock.json` from this PR.

PR #5150, opened 2024-06-12
[split sessions] feat(auth): update api for split sessions

hayemaxi:
**Separate sessions commit**

- Add new API to Toolkit to "declare" and "undeclare" a connection. Unlike the other API methods, this will inform the toolkit of existing connections without accepting them into its auth environment. This allows it to use params from other connections, e.g. start url region. It does not persist. 
- Retain original API because it may be beneficial in the future for other extensions/use cases
- Change Amazon Q's usage of the API to declare instead of direct session sharing.

## License

### DIFF
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -121,12 +121,8 @@ export interface ConnectionStateChangeEvent {
 
 export type AuthType = Auth
 
-type DeclaredConnection = {
-    startUrl: string
-    region: string
-    source: string
-    id?: string
-}
+export type DeletedConnection = { connId: Connection['id']; storedProfile?: StoredProfile }
+type DeclaredConnection = Pick<SsoProfile, 'ssoRegion' | 'startUrl'> & { source: string }
 
 export class Auth implements AuthService, ConnectionManager {
     readonly #ssoCache = getCache()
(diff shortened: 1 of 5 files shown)


nkomonen-amazon:
What I gather from this PR is that Amazon Q can look at the connections from Toolkits, and use their metadata to create a separate connection for itself? Does this mean that the webview which shows being able to "reuse" an existing connection will still remain, but if they select that we will create a new connection instead of adding scopes to it?

hayemaxi:
> What I gather from this PR is that Amazon Q can look at the connections from Toolkits, and use their metadata to create a separate connection for itself? Does this mean that the webview which shows being able to "reuse" an existing connection will still remain, but if they select that we will create a new connection instead of adding scopes to it?

That's correct. The implementation for that will be in the next PR.

packages/core/src/auth/auth.ts
```diff
@@ -121,6 +121,13 @@ export interface ConnectionStateChangeEvent {
 
 export type AuthType = Auth
 
+type DeclaredConnection = {
```
nkomonen-amazon, line 124: Is it possible to use the `SsoProfile` interface instead? Or make a new interface which extends `SsoProfile`. This type and SsoProfile are semantically related. If we do not even need the `id` then maybe we can use `SsoProfile` itself.

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -424,64 +424,34 @@ export const signoutCodeWhisperer = Commands.declare(
...
-                    label: conn.label,
-                } as AwsConnection)
+            if (conn?.type === 'sso') {
+                getLogger().info(`toolkitApi: declare connection ${id}`)
+                _toolkitApi.declareConnection(
+                    {
+                        type: conn.type,
+                        ssoRegion: conn.ssoRegion,
+                        startUrl: conn.startUrl,
+                        id: id,
```
nkomonen-amazon, line 442: Do we need the `id` of a connection still? Did this make sense when we shared connections, but isn't required anymore

nkomonen-amazon (approved):

justinmk3 (approved):

packages/core/src/auth/auth.ts
hayemaxi: done

packages/core/src/codewhisperer/commands/basicCommands.ts
hayemaxi: The delete connection callback only supplied id, so that was the main hash. I also didn't want to create conflicts with starturls of varying regions. But you can't have the same starturl in multiple regions, and most users will not have multiple sso logins to choose from anyway.

So instead I updated the delete connection callback to also return the stored profile so we can identify and hash on the start url.

nkomonen-amazon (approved):

PR #5152, opened 2024-06-13
fix(codecatalyst): avoid self-triggered UserActivity

justinmk3:
Problem:
UserActivity may fire when the extension writes to its own logfile.

Solution:
Special-case the `onDidOpenTextDocument` event.

## License

### DIFF
--- a/packages/core/src/shared/extensionUtilities.ts
+++ b/packages/core/src/shared/extensionUtilities.ts
@@ -321,17 +321,19 @@ export function getComputeRegion(): string | undefined {
 }
 
 /**
- * Provides an {@link vscode.Event} that is triggered when the user interacts with the extension.
- * This will help to be notified of user activity.
+ * Provides a generic {@link vscode.Event} representing "user activity".
  */
-export class ExtensionUserActivity implements vscode.Disposable {
+export class UserActivity implements vscode.Disposable {
     private disposables: vscode.Disposable[] = []
     private activityEvent = new vscode.EventEmitter<void>()
-    /** The event that is triggered when the user interacts with the extension */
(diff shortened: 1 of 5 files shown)


packages/core/src/shared/extensionUtilities.ts
```diff
@@ -321,17 +321,19 @@ export function getComputeRegion(): string | undefined {
 }
 
 /**
- * Provides an {@link vscode.Event} that is triggered when the user interacts with the extension.
- * This will help to be notified of user activity.
+ * Provides a generic {@link vscode.Event} representing "user activity".
  */
-export class ExtensionUserActivity implements vscode.Disposable {
+export class UserActivity implements vscode.Disposable {
```
justinmk3, line 326: Renamed. This event is intended to track any kind of "user active" signal, not only "extension" activity.

packages/core/src/shared/extensionUtilities.ts
```diff
@@ -385,17 +402,21 @@ export class ExtensionUserActivity implements vscode.Disposable {
...
-                    // File changed was not the one the user was actively working on
+                throttledEmit(vscode.workspace.onDidChangeTextDocument)
+            })
+        )
+
+        this.register(
+            vscode.workspace.onDidOpenTextDocument(e => {
+                if (vscode.window.activeTextEditor?.document.uri.toString() !== e.uri.toString()) {
+                    // User is not editing this document.
                     return
```
justinmk3, line 417: changed onDidOpenTextDocument to a special case

nkomonen-amazon (approved):

jpinkney-aws (approved):

PR #5154, opened 2024-06-13
ci: fix windows global activation

jpinkney-aws:
### Problem:
- Windows ci is no longer working after we moved amazonq tests to the
  amazonq package

### Solution:
- Utilize upper case drive letters for getting the root in
  amazonq/toolkit. That way imports will resolve correctly.

### Additional info:
In multi root npm workspaces on windows it looks like imports into other npm workspace packages makes the loaded module id an uppercase drive letter in the node require cache.

E.g. when we import a file from core the module ids inside of amazonq/toolkits node require cache are something like:
- C:\${pathToWorkspace}\packages\core\myfile.js

However, internal workspace package imports are lower case drive letters. That means when core imports a module inside of core we see this as:
- c:\${pathToWorkspace}\packages\core\myfile.js

This can cause things like globals to be undefined, since tests inside of amazonq/toolkit are looking for upper case module ids, whereas tests inside of core are always looking
for lower case module ids (since the tests live inside of core itself)

## License

### DIFF
--- a/packages/core/src/test/shared/vscode/runCommand.test.ts
+++ b/packages/core/src/test/shared/vscode/runCommand.test.ts
@@ -143,11 +143,17 @@ describe('runCommand', function () {
                 this.skip()
             }
 
-            const pat =
-                os.platform() === 'linux'
-                    ? // vscode error not raised on linux? 💩
-                      /EISDIR: illegal operation on a directory/
-                    : /EEXIST: file already exists/
+            const pat = (() => {
+                switch (os.platform()) {
+                    case 'linux':
+                        // vscode error not raised on linux? 💩
+                        return /EISDIR: illegal operation on a directory/
+                    case 'win32':
+                        return /EPERM: operation not permitted/
+                    default:
+                        return /EEXIST: file already exists/
+                }
+            })()
             await runAndWaitForMessage(pat, async () => {
                 // Try to write to the current directory. 💩
                 const err = await SystemUtilities.writeFile('.', 'foo').catch(e => e)
@@ -174,7 +180,12 @@ describe('runCommand', function () {
         })
 
         it('toolkit `PermissionsError`', async function () {
-            const viewLogsDialog = getTestWindow().waitForMessage(/incorrect permissions. Expected rw-, found r--/)
+            const expectedMsg =
+                os.platform() === 'win32'
+                    ? /EPERM: operation not permitted/
+                    : /incorrect permissions. Expected rw-, found r--/
+
+            const viewLogsDialog = getTestWindow().waitForMessage(expectedMsg)
 
             await Promise.all([
                 viewLogsDialog.then(dialog => dialog.close()),
@@ -190,7 +201,9 @@ describe('runCommand', function () {
         })
 
         it('nodejs EACCES (not wrapped by toolkit `PermissionsError`)', async function () {
-            const viewLogsDialog = getTestWindow().waitForMessage(/EACCES: permission denied/)
+            const expectedMsg =
+                os.platform() === 'win32' ? /EPERM: operation not permitted/ : /EACCES: permission denied/
+            const viewLogsDialog = getTestWindow().waitForMessage(expectedMsg)
 
             await Promise.all([
                 viewLogsDialog.then(dialog => dialog.close()),
--- a/packages/core/src/test/testRunner.ts
+++ b/packages/core/src/test/testRunner.ts
@@ -31,20 +31,40 @@ export async function runTests(
 
     function getRoot(): string {
         const abs = process.env['DEVELOPMENT_PATH'] ?? process.cwd()
-
         if (process.platform !== 'win32') {
             return abs
         }
 
-        // Force all drive letters (or whatever else is before the first colon) to be lowercase.
-        //
-        // Node's `require` will always cache modules based off case-sensitive paths, regardless
-        // of the underlying file system. This is normally not a problem, but VS Code also happens
-        // to normalize paths on Windows to use lowercase drive letters when using its bootstrap loader.
-        // This means that each module ends up getting loaded twice, once by the extension and once
-        // by any test code, causing all sorts of bizarre behavior during tests.
+        /**
+         * Node's `require` will always cache modules based off case-sensitive paths, regardless
+         * of the underlying file system. This is normally not a problem, but VS Code also happens
+         * to normalize paths on Windows to use lowercase drive letters when using its bootstrap loader.
+         * This means that each module ends up getting loaded twice, once by the extension and once
+         * by any test code, causing all sorts of bizarre behavior during tests unless you normalize
+         * paths like below.
+         *
+         * In multi root npm workspaces on windows it looks like imports into other npm workspace packages
+         * makes the loaded module id an uppercase drive letter in the node require cache.
+         *
+         * E.g. when we import a file from core the module ids inside of amazonq/toolkits node require
+         * cache are something like:
+         *  - C:\${pathToWorkspace}\packages\core\myfile.js
+         *
+         * However, internal workspace package imports are lower case drive letters. That means when
+         * core imports a module inside of core we see this as:
+         *  - c:\${pathToWorkspace}\packages\core\myfile.js
+         *
+         * This can cause things like globals to be undefined, since tests inside of amazonq/toolkit
+         * are looking for upper case module ids, whereas tests inside of core are always looking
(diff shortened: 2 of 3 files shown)


packages/core/src/test/globalSetup.test.ts
```diff
@@ -51,8 +51,7 @@ export async function mochaGlobalSetup(extensionId: string) {
 
         // Extension activation has many side-effects such as changing globals
         // For stability in tests we will wait until the extension has activated prior to injecting mocks
-        const activationLogger = (msg: string, ...meta: any[]) => console.log(format(msg, ...meta))
-        await activateExtension(extensionId, false, activationLogger)
+        await vscode.extensions.getExtension(extensionId)?.activate()
```
justinmk3, line 54: is the logger issue not relevant?

justinmk3 (approved):

justinmk3:
For reference, here's a stacktrace when this happens locally (macOS, can't repro consistently):

```
Activating extension: amazonwebservices.aws-core-vscode
extensionHostProcess.js:147
Extension failed to activate: amazonwebservices.aws-core-vscode: Error: ToolkitGlobals accessed before initialize()
    at Object.get (/…/packages/core/src/shared/extensionGlobals.ts:107:23)
    at new Timeout (/…/packages/core/src/shared/utilities/timeoutUtils.ts:57:35)
    at activateExtension (/…/packages/core/src/shared/utilities/vsCodeUtils.ts:149:48)
    at Runner.<anonymous> (/…/packages/core/src/test/globalSetup.test.ts:55:32)
    at async Mocha._runGlobalFixtures (/…/node_modules/mocha/lib/mocha.js:1214:5)
    at async Mocha.runGlobalSetup (/…/node_modules/mocha/lib/mocha.js:1174:5)
    at async runAsync (/…/node_modules/mocha/lib/mocha.js:1016:11)
extensionHostProcess.js:147
rejected promise not handled within 1 second: Error: ToolkitGlobals accessed before initialize()
extensionHostProcess.js:147
stack trace: Error: ToolkitGlobals accessed before initialize()
    at Object.get (/…/packages/core/src/shared/extensionGlobals.ts:107:23)
    at new Timeout (/…/packages/core/src/shared/utilities/timeoutUtils.ts:57:35)
    at activateExtension (/…/packages/core/src/shared/utilities/vsCodeUtils.ts:149:48)
    at Runner.<anonymous> (/…/packages/core/src/test/globalSetup.test.ts:55:32)
    at async Mocha._runGlobalFixtures (/…/node_modules/mocha/lib/mocha.js:1214:5)
    at async Mocha.runGlobalSetup (/…/node_modules/mocha/lib/mocha.js:1174:5)
    at async runAsync (/…/node_modules/mocha/lib/mocha.js:1016:11)
```

justinmk3:
These failing windows tests are unrelated, can be fixed by special-casing Windows for those tests (or just skip them for now): 

```
  3 failing
  1) runCommand
       shows error
         vscode ISDIR:
     Error: Timed out waiting for message: /EEXIST: file already exists/, displayed:
[
    "Failed to run command: _test.throwable: Error: EPERM: operation not permitted, open 'D:\\'"
]

  2) runCommand
       shows error
         toolkit `PermissionsError`:
     Error: Timed out waiting for message: /incorrect permissions. Expected rw-, found r--/, displayed:
[
    "Failed to run command: _test.throwable: Error: EPERM: operation not permitted, open 'c:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\aws-toolkit-vscode\\vsctka53892c2\\unwritableFile'"
]

  3) runCommand
       shows error
         nodejs EACCES (not wrapped by toolkit `PermissionsError`):
     Error: Timed out waiting for message: /EACCES: permission denied/, displayed:
[
    "Failed to run command: _test.throwable: EPERM: operation not permitted, open 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\aws-toolkit-vscode\\vsctka53892c2\\unwritableFile'"
```

packages/core/src/test/testRunner.ts
```diff
@@ -31,20 +31,40 @@ export async function runTests(
...
 
-        // Force all drive letters (or whatever else is before the first colon) to be lowercase.
-        //
-        // Node's `require` will always cache modules based off case-sensitive paths, regardless
-        // of the underlying file system. This is normally not a problem, but VS Code also happens
-        // to normalize paths on Windows to use lowercase drive letters when using its bootstrap loader.
-        // This means that each module ends up getting loaded twice, once by the extension and once
-        // by any test code, causing all sorts of bizarre behavior during tests.
+        /**
+         * Node's `require` will always cache modules based off case-sensitive paths, regardless
```
justinmk3, line 39: you mentioned upstream vscode issues, are there any good ones we can reference here?

justinmk3:
Is this primarily relevant to the require() call here: https://github.com/aws/aws-toolkit-vscode/blob/6f263c81f58d6d50f5cbc6ae223460a28f6f3146/packages/core/src/test/testRunner.ts#L96-L105

- should we add a comment there like `see comment in getRoot()` ?
- maybe also mention `NODE_DEBUG=module` https://stackoverflow.com/a/21057355/152142 
- The existing comments there suggest that this may be a mocha issue or at least we may want to (in the future) look more closely at how we're using mocha, since that code basically hasn't change for 5 years :)

Is there any way we can *detect* this problem to make it less painful in the future? For example, can we iterate node's resolved paths and make a guess about the proper casing? Or can we check the result of `require(fullPath)` ?

packages/core/src/test/shared/vscode/runCommand.test.ts
```diff
@@ -174,7 +180,11 @@ describe('runCommand', function () {
         })
 
         it('toolkit `PermissionsError`', async function () {
-            const viewLogsDialog = getTestWindow().waitForMessage(/incorrect permissions. Expected rw-, found r--/)
+            const expectedMsg = os.platform() === 'win32'
+            ?
+              /EPERM: operation not permitted/
+            : /EACCES: permission denied/
```
justinmk3, line 186:
```suggestion
            : /incorrect permissions. Expected rw-, found r--/
```

jpinkney-aws:
>  should we add a comment there like see comment in getRoot() ?

I'm not sure I understand how these things are related, I think mocha require https://mochajs.org/#-require-module-r-module and what's happening here are two different things. The require.cache from everything coming from `aws-core-vscode` is already populated before we do anything with mocha

> The existing comments there suggest that this may be a mocha issue or at least we may want to (in the future) look more closely at how we're using mocha, since that code basically hasn't change for 5 years :)

It's worth checking to see if we can update this, but I think its a node/vscode issue rather than a mocha issue. I think their requires won't help us here

packages/core/src/test/testRunner.ts
jpinkney-aws: I couldn't find any that were relevant to this specific comment. I'm not sure if Jaden looked through the code to figure that out or not. The ones I found just referenced issues with breakpoints and debug paths and a couple other random internal things

justinmk3:
> I'm not sure I understand how these things are related, I think mocha require https://mochajs.org/#-require-module-r-module and what's happening here are two different things.

We are modifying `getRoot` to fix this issue. Which code using the result of `getRoot` is affected by this issue? I thought it was the mocha require(), no?

jpinkney-aws:
> We are modifying getRoot to fix this issue. Which code using the result of getRoot is affected by this issue?

Everything that uses getRoot is affected. AFAICT when a new module is loaded all of its imports are put into the require cache depending on the case of their drive letter. So if globalSetup was required with an upper case drive letter and all the mocha files [are loaded](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/test/testRunner.ts#L98) with an uppercase drive letter then everything in the require.cache that's loaded inside of toolkit/amazonq will be with an upper case drive letter. Everything in core will always use lower case drive letters though

justinmk3:
> when a new module is loaded all of its imports are put into the require cache depending on the case of their parent modules drive letter. So if globalSetup was required with an upper case drive letter and all the mocha files [are loaded](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/test/testRunner.ts#L98) with an uppercase drive letter then everything in the require.cache that's loaded inside of toolkit/amazonq will be with an upper case drive letter. Everything in core will always use lower case drive letters though

This is a much clearer explanation. Could replace the entire comment in `getRoot`.

justinmk3 (approved):

PR #5157, opened 2024-06-13
feature: replace lodash dependencies with native

32teeth:
## Problem

most of the used lodash features can be replaced with modern JavaScript

## Solution

* removed `lodash` from dependencies `npm uninstall @types/lodash lodash`
* `_.cloneDeep` replaced with `_CloneDeep`
* `_.set` replaced with `_Set`
* `_.get` replaced with `_Get`
* `_.isEmpty` replaced with `_IsEmpty`

## License


justinmk3:
Thanks for looking into this! We have a backlog issue that proposed this. 

For debounce/throttle, maybe it makes sense to only depend on those specific lodash packages? That would avoid accidental and presumably unnecessary use of most other lodash features.

32teeth:
Closed this one. Will have another soon

package-lock.json
```diff
@@ -971,108 +971,6 @@
                 "node": ">=16.0.0"
             }
         },
-        "node_modules/@aws-sdk/client-sso-oidc/node_modules/@aws-sdk/credential-provider-node/node_modules/@aws-sdk/client-sso-oidc": {
```
justinmk3, line 974: this seems unexpected.

btw, should there be a package.json change?

PR #5179, opened 2024-06-18
feat: add featureDevTelemetryEvent

sannicm:
## Problem

For testing ABTesting configuration and infrastructure, is needed to start sending telemetry events so RTS. This will create the basic implementation to track ABTests for featureDev.

## Solution

Add the new event model to the service schema, Add a sendTelemetryEvent request when a conversation is created. 

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/client/featureDev.ts
+++ b/packages/core/src/amazonqFeatureDev/client/featureDev.ts
@@ -308,4 +308,34 @@ export class FeatureDevClient {
             throw new ToolkitError((e as Error).message, { code: 'ExportResultArchiveFailed' })
         }
     }
+
+    /**
+     * This event is specific to ABTesting purposes.
+     *
+     * No need to fail currently if the event fails in the request. In addition, currently there is no need for a return value.
+     *
+     * @param conversationId
+     */
+    public async sendFeatureDevTelemetryEvent(conversationId: string) {
+        try {
+            const client = await this.getClient()
+            const params: FeatureDevProxyClient.SendTelemetryEventRequest = {
+                telemetryEvent: {
+                    featureDevEvent: {
+                        conversationId,
+                    },
+                },
+            }
+            const response = await client.sendTelemetryEvent(params).promise()
+            getLogger().debug(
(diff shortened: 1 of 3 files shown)


packages/core/src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -308,4 +308,34 @@ export class FeatureDevClient {
...
+            const client = await this.getClient()
+            const params: FeatureDevProxyClient.SendTelemetryEventRequest = {
+                telemetryEvent: {
+                    featureDevEvent: {
+                        conversationId,
+                    },
+                },
+            }
+            const response = await client.sendTelemetryEvent(params).promise()
+            getLogger().info(
```
justinmk3, line 330: do users normally need to see this? if not, prefer debug()

packages/core/src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -308,4 +308,34 @@ export class FeatureDevClient {
...
+                `${featureName}: successfully sent featureDevEvent: ConversationId: ${conversationId} RequestId: ${response.$response.requestId}`
+            )
+        } catch (e) {
+            getLogger().error(
+                `${featureName}: failed to send faeture dev telemetry event: ${(e as Error).message} RequestId: ${
```
justinmk3, line 335: typo. also, usually a good idea to include e.name
```suggestion
                `${featureName}: failed to send feature dev telemetry: ${(e as Error).name}: ${(e as Error).message}: RequestId: ${
```

justinmk3 (approved):
Needs review from someone on your team.

packages/core/src/amazonqFeatureDev/client/featureDev.ts
sannicm: thanks for the suggestion

packages/core/src/amazonqFeatureDev/client/featureDev.ts
sannicm: Good point, they probably would not.

PR #5181, opened 2024-06-18
fix(amazonq): filesystem errors are not surfaced

justinmk3:
## Problem
- amazonq main does not use activateCommon, so it is missing some common logic.
    - does not define a handler for bubbled-up errors, so errors are not displayed in a user-facing message.
- `findBestErrorInChain()` discards non-AWSError, so filesystem errors are not surfaced.

## Solution
- call `registerCommandErrorHandler` in amazonq init.
    - TODO: can amazonq init share common init w/ toolkit?
- improve `findBestErrorInChain()` so that it finds filesystem errors.

## Examples

- image
- image

## License


packages/core/src/codewhisperer/activation.ts
```diff
@@ -293,7 +298,8 @@ export async function activate(context: ExtContext): Promise<void> {
 
     if (auth.isConnectionExpired()) {
         auth.showReauthenticatePrompt().catch(e => {
-            getLogger().error('showReauthenticatePrompt failed: %s', (e as Error).message)
+            const defaulMsg = localize('AWS.generic.message.error', 'Failed to reauth:')
+            void logAndShowError(localize, e, 'showReauthenticatePrompt', defaulMsg)
```
justinmk3, line 302: If `showReauthenticatePrompt` failed, no message was shown here. Fixed now, but need to revisit this. Not great to special-case it here.

PR #5182, opened 2024-06-18
feat(ui): show login prompt for split sessions

hayemaxi:
**Separate sessions commit**

- If user is logged out of toolkit due to extension session splitting, display a prompt that opens the sign in page when clicked.
- Dismissing or clicking the button dismisses it permanently.
- The prompt will continue to show across restarts until it is dismissed, the button is pressed, or the user signs into toolkit.
- The prompt will display only once, regardless of how many connections are logged out (forgotten).
- The sign in button will go to the login page with the proper scopes, i.e.g if codecatalyst was signed out, go to code catalyst login. Otherwise go to explorer only login.

## License


justinmk3 (approved):

packages/core/src/auth/auth.ts
```diff
@@ -999,6 +1004,67 @@ export class Auth implements AuthService, ConnectionManager {
...
+    /**
+     * Displays a sign in prompt to the user if they have been logged out of the Toolkit as part of
+     * separating auth sessions between extensions. Pass a command name to open that login
+     * (e.g. codecatalyst vs explorer login). Passing nothing will open a prompt for that last used command name
+     * (or do nothing if no command name has ever been passed), which is useful to redisplay the prompt after
+     * reloads in case a user misses it.
+     *
+     * TODO: Remove after some time.
+     */
+    public async displaySeparationPrompt(cmd?: string) {
```
nkomonen-amazon, line 1024: not a blocker: To keep this code more isolated, and to reduce the amount of code in the Auth class, would putting this in its own class be cleaner? It would still live in `auth.ts`

nkomonen-amazon (approved):

jpinkney-aws (approved):

packages/core/src/auth/auth.ts
hayemaxi: updated

PR #5194, opened 2024-06-19
telemetry(amazonq security scan): set reasonDesc

laileni-aws:
## Problem
- Telemetry error messages are not consistent in VSC and JB
## Solution
-  Improving telemetry error messages to keep consistent in both IDE's

## License

### DIFF
--- a/packages/core/src/codewhisperer/commands/startSecurityScan.ts
+++ b/packages/core/src/codewhisperer/commands/startSecurityScan.ts
@@ -31,7 +31,7 @@ import {
 import { cancel, ok } from '../../shared/localizedText'
 import { getDirSize } from '../../shared/filesystemUtilities'
 import { telemetry } from '../../shared/telemetry/telemetry'
-import { isAwsError } from '../../shared/errors'
+import { ToolkitError, isAwsError } from '../../shared/errors'
 import { openUrl } from '../../shared/utilities/vsCodeUtils'
 import { AuthUtil } from '../util/authUtil'
 import path from 'path'
@@ -39,7 +39,7 @@ import { ZipMetadata, ZipUtil } from '../util/zipUtil'
 import { debounce } from 'lodash'
 import { once } from '../../shared/utilities/functionUtils'
 import { randomUUID } from '../../common/crypto'
-import { CodeAnalysisScope } from '../models/constants'
+import { CodeAnalysisScope, ProjectSizeExceededErrorMessage } from '../models/constants'
 import { CodeScanJobFailedError, CreateCodeScanFailedError, SecurityScanError } from '../models/errors'
 
 const localize = nls.loadMessageBundle()
@@ -186,7 +186,7 @@ export async function startSecurityScan(
         throwIfCancelled(scope, codeScanStartTime)
         const jobStatus = await pollScanJobStatus(client, scanJob.jobId, scope, codeScanStartTime)
         if (jobStatus === 'Failed') {
-            logger.verbose(`Security scan job failed.`)
+            logger.verbose(`Security scan failed.`)
             throw new CodeScanJobFailedError()
         }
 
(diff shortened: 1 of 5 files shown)


packages/core/src/codewhisperer/commands/startSecurityScan.ts
```diff
@@ -250,7 +250,11 @@ export async function startSecurityScan(
                 CodeScansState.instance.setMonthlyQuotaExceeded()
             }
         }
-        codeScanTelemetryEntry.reason = (error as SecurityScanError).message
+        if ((error as ToolkitError).code == 'ContentLengthError') {
+            codeScanTelemetryEntry.reason = 'Payload size limit reached'
+        } else {
+            codeScanTelemetryEntry.reason = (error as SecurityScanError).message
```
justinmk3, line 256: `reason` is for code-like strings. Use `reasonDesc` for free-form text. use `getTelemetryReasonDesc` to get it: https://github.com/aws/aws-toolkit-vscode/blob/cb603edfe2d2429ff4ee080ae4094a7de2806989/packages/core/src/shared/errors.ts#L310

```suggestion
            codeScanTelemetryEntry.reasonDesc = getTelemetryReasonDesc(error)
```

packages/core/src/codewhisperer/commands/startSecurityScan.ts
```diff
@@ -250,7 +250,11 @@ export async function startSecurityScan(
                 CodeScansState.instance.setMonthlyQuotaExceeded()
             }
         }
-        codeScanTelemetryEntry.reason = (error as SecurityScanError).message
+        if ((error as ToolkitError).code == 'ContentLengthError') {
+            codeScanTelemetryEntry.reason = 'Payload size limit reached'
```
justinmk3, line 254: reasonDesc, not reason

laileni-aws: `Security scans` was using `reason` from the start, Let me discuss with @ctlai95 as this will impact metrics in kibana accordingly.

laileni-aws: Discussed and I will update `reason` with `reasonDesc` for security scans.

packages/core/src/codewhisperer/commands/startSecurityScan.ts
laileni-aws: Updated accordingly.

packages/core/src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -289,7 +290,10 @@ export async function uploadArtifactToS3(
         getLogger().error(
             `Amazon Q is unable to upload workspace artifacts to Amazon S3 for security scans. For more information, see the Amazon Q documentation or contact your network or organization administrator.`
         )
-        const errorMessage = (error as Error).message
+        const errorMessage = getTelemetryReasonDesc(error)?.includes(`"PUT" request failed with code "403"`)
+            ? `UploadArtifactToS3Exception: "PUT" request failed with code "403"`
```
ashishrp-aws, line 294: Don't we need add the backend errors here as well just like JetBrains. like 407s etc.., event though they're not seen here, they can happen right.

laileni-aws: Yes they may happen but we dont know the error message etc so didn't handle it. Thanks to Justin, by using this approach of `reasonDesc` as we can diff the error from where they are originating from.

packages/core/src/codewhisperer/commands/startSecurityScan.ts
```diff
@@ -250,7 +250,11 @@ export async function startSecurityScan(
                 CodeScansState.instance.setMonthlyQuotaExceeded()
             }
         }
-        codeScanTelemetryEntry.reason = (error as SecurityScanError).message
+        codeScanTelemetryEntry.reasonDesc =
+            (error as ToolkitError)?.code === 'ContentLengthError'
+                ? 'Payload size limit reached'
+                : getTelemetryReasonDesc(error)
+        codeScanTelemetryEntry.reason = (error as ToolkitError)?.code ?? 'DefaultError'
```
justinmk3, line 257: you may want to use `getTelemetryReason`  https://github.com/aws/aws-toolkit-vscode/blob/6bc6cf13cddafe23d48e5340f16efc55b6e00f3a/packages/core/src/shared/errors.ts#L317

justinmk3 (approved):

ashishrp-aws (approved):

PR #5196, opened 2024-06-20
fix(codecatalyst): dev env logged out on load

hayemaxi:
**Separate sessions commit**

Problem: We forget all connections without all codecatalyst scopes only (aws account + coca scopes). However, the stored credentials in dev environments do not have the account scopes, so we are forgetting it.

Solution: Check for Q scopes explicitly.

## License

### DIFF
--- a/packages/core/src/extension.ts
+++ b/packages/core/src/extension.ts
@@ -128,6 +128,9 @@ export async function activate(context: vscode.ExtensionContext) {
         // TODO: Remove after some time?
         for (const conn of await Auth.instance.listConnections()) {
             if (isSsoConnection(conn) && hasScopes(conn, codeWhispererCoreScopes)) {
+                getLogger().debug(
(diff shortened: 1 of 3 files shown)


packages/core/src/extension.ts
```diff
@@ -128,6 +128,9 @@ export async function activate(context: vscode.ExtensionContext) {
         // TODO: Remove after some time?
         for (const conn of await Auth.instance.listConnections()) {
             if (isSsoConnection(conn) && hasScopes(conn, codeWhispererCoreScopes)) {
+                getLogger().debug(
+                    `forgetting connection: ${conn.id} with starturl/scopes: ${conn.startUrl} / ${conn.scopes}`
```
justinmk3, line 132: I assume this already works good enough, but for reference, `%O` works better than toString() for printing objects
```suggestion
                getLogger().debug(
                    `forgetting connection: ${conn.id} with starturl/scopes: ${conn.startUrl} / %O`,
                    conn.scopes
```

justinmk3 (approved):

PR #5198, opened 2024-06-20
auth: ignore vscode proxy errors on token refresh

nkomonen-amazon:
## Problem

We were seeing an error in `aws_refreshCredentials` metric. This error was networking related when the user has setup a proxy in vscode, but they were not connected to it. Maybe they didn't connect to VPN.

We typically catch network errors during the SSO token refresh process so that the session does not become invalidated on it. We do not catch all errors, but will add them in incrementally as we see them in telemetry. This proxy error was not caught and caused sessions to become invalidated prematurely.

## Solution

Treat this VSCode Proxy Error as a network error so that we ignore it during token refresh, keeping the users session valid.

## License

### DIFF
--- a/packages/core/src/shared/errors.ts
+++ b/packages/core/src/shared/errors.ts
@@ -565,6 +565,10 @@ export class PermissionsError extends ToolkitError {
 }
 
 export function isNetworkError(err?: unknown): err is Error & { code: string } {
+    if (isVSCodeProxyError(err)) {
+        return true
+    }
+
     if (!hasCode(err)) {
         return false
     }
@@ -583,4 +587,18 @@ export function isNetworkError(err?: unknown): err is Error & { code: string } {
         'ENOBUFS', // client side memory issue during http request?
         'EADDRNOTAVAIL', // port not available/allowed?
     ].includes(err.code)
+}
+
+/**
+ * This error occurs on a network call if the user has set up a proxy in the
+ * VS Code settings but the proxy is not reachable.
+ *
+ * Setting ID: http.proxy
+ */
+function isVSCodeProxyError(err?: unknown): boolean {
+    if (!(err instanceof Error)) {
+        return false
+    }
+
+    return err.name === 'Error' && err.message.startsWith('Failed to establish a socket connection to proxies')
 }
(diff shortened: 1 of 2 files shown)


packages/core/src/shared/errors.ts
```diff
@@ -584,3 +588,17 @@ export function isNetworkError(err?: unknown): err is Error & { code: string } {
...
     ].includes(err.code)
 }
+
+/**
+ * This error occurs on a network call if the user has set up a proxy in the
+ * VS Code settings but the proxy is not reachable.
+ *
+ * Setting ID: http.proxy
+ */
+function isVSCodeProxyError(err?: unknown): boolean {
```
justinmk3, line 598: Not sure why `unknown` is needed in some of these functions, but at least for most of these we can force the caller to make that decision.
```suggestion
function isVSCodeProxyError(err?: Error): boolean {
```

packages/core/src/shared/errors.ts
```diff
@@ -584,3 +588,17 @@ export function isNetworkError(err?: unknown): err is Error & { code: string } {
...
+    if (!(err instanceof Error)) {
+        return false
+    }
+
+    return err.name === 'Error' && err.message.startsWith('Failed to establish a socket connection to proxies')
```
justinmk3, line 603: what other fields are set on the object? is there any other (non-standard) field we can use as a more reliable hint?

packages/core/src/test/shared/errors.test.ts
```diff
@@ -417,3 +418,16 @@ describe('util', function () {
         assert.deepStrictEqual(getErrorMsg(err), 'aws error desc 1')
     })
 })
+
+describe('isNetworkError()', function () {
+    it('returns true when it is a VS Code Proxy error', function () {
```
justinmk3, line 423: this would fit in the "util" describe() just above here. and can be a single it()

packages/core/src/shared/errors.ts
nkomonen-amazon: the `err` comes from a `catch` and is unknown, so this just makes it easier to throw it in and not worry about it

packages/core/src/shared/errors.ts
nkomonen-amazon: This is the best we can do with the error that they give us. Nothing else useful I could see in it

justinmk3 (approved):

jpinkney-aws (approved):

PR #5213, opened 2024-06-24
auth: TimeoutError OIDC client fix attempt

nkomonen-amazon:
## Problem

During API calls with the OIDC client the request would time out and we would get TimeoutError.

## Solution:

See if adding an explicit request timeout amount will avoid this. We'll need to observe telemetry to see if this
specific type of error goes down.

## License


justinmk3:
This could be a SDK bug: https://github.com/denoland/deno/issues/18932
Which may be fixed if we upgrade aws sdk?

justinmk3 (approved):

jpinkney-aws (approved):

Issue #5115, opened 2024-06-07
unreliable test: Auth promptForConnection


justinmk3:
```
  1) Auth
       promptForConnection
         reauthenticates a connection if the user selects an expired one:

      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert_1.default.ok(!refreshedConnItem.description?.match(/expired/i))

      + expected - actual

      -false
      +true
      
      at /codebuild/output/src3389505232/src/github.com/aws/aws-toolkit-vscode/packages/core/src/test/credentials/auth.test.ts:512:24
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async /codebuild/output/src3389505232/src/github.com/aws/aws-toolkit-vscode/packages/core/src/test/shared/vscode/window.ts:496:29
```

justinmk3:
resolved by https://github.com/aws/aws-toolkit-vscode/pull/5228

PR #5214, opened 2024-06-24
fix: added branch main to yml files because new forking styles

32teeth:
## Problem

[GitHub changed it's naming many years ago](https://github.com/github/renaming?tab=readme-ov-file). As should we.

## Solution

Added `main` where needed in workflows

## License

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
             #   run: echo 'TAG_NAME=prerelease' >> $GITHUB_ENV
             - if: github.event_name == 'workflow_dispatch'
               run: echo "TAG_NAME=${{ github.event.inputs.tag_name }}" >> $GITHUB_ENV
-            - if: github.ref_name != 'master' || github.ref_name != 'main'
+            - if: github.ref_name != 'master' && github.ref_name != 'main'
               run: |
                   TAG_NAME=${{ github.ref_name }}
                   FEAT_NAME=$(echo $TAG_NAME | sed 's/feature\///')
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ on:
                 required: false
                 default: prerelease
     push:
-        branches: [master, feature/*]
+        branches: [master, main, feature/*]
         # tags:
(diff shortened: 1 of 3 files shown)


32teeth:
Wondering how `yml` change impacts Linux build?

```
Build Name
Linux Unit Tests - minimum
```

maybe flaky retries on with assertion?

justinmk3:
the failing promptForConnection test is a known issue https://github.com/aws/aws-toolkit-vscode/issues/5115

.github/workflows/release.yml
```diff
@@ -40,13 +40,13 @@ jobs:
...
               run: echo "TAG_NAME=${{ github.event.inputs.tag_name }}" >> $GITHUB_ENV
-            - if: github.ref_name != 'master'
+            - if: github.ref_name != 'master' || github.ref_name != 'main'
               run: |
                   TAG_NAME=${{ github.ref_name }}
                   FEAT_NAME=$(echo $TAG_NAME | sed 's/feature\///')
                   echo "FEAT_NAME=$FEAT_NAME" >> $GITHUB_ENV
                   echo "TAG_NAME=pre-$FEAT_NAME" >> $GITHUB_ENV
-            - if: github.ref_name == 'master'
+            - if: github.ref_name == 'master' || github.ref_name == 'main'
```
justinmk3, line 49: these should be `&&` ?

justinmk3:
is this for forks? 

if/when this repo changes its default branch, then it would make sense to replace 'master' with 'main'. until then, what is the purpose of adding `|| 'main'` to various conditions?

.github/workflows/release.yml
32teeth: Thanks for the corrections.
Yes, for the `!==` we should have the `&&` condition

32teeth: as per

```
- if: github.ref_name == 'master' || github.ref_name == 'main'
   run: |
       echo "FEAT_NAME=" >> $GITHUB_ENV
       echo "TAG_NAME=prerelease" >> $GITHUB_ENV
```

should be fine since it's either

justinmk3: i'm more concerned about the semantics, rather than the fact that it happens to work ok

PR #5217, opened 2024-06-24
Fix[gumby]: include exportContext when downloadArtifactId is not undefined

JuliaABurch:
## Problem
Some gumby workflows need a specific artifact ID, but if one is not defined, we expect `exportContext` to not be defined

## Solution
This change uses nullish coalition to optinally add the `exportContext` property when `downloadArtifactId` is defined.

Note: in adding test coverage, I brought the test from the `amazonq` package to the `core` package, as it tests the internals of the `transformApiHandler` class, which lives in the `core` package. 

## Testing done
- Successful transformation case
- HIL case
- Build failure test case

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3321,9 +3321,9 @@
             }
         },
         "node_modules/@aws/mynah-ui": {
-            "version": "4.14.1",
-            "resolved": "https://registry.npmjs.org/@aws/mynah-ui/-/mynah-ui-4.14.1.tgz",
-            "integrity": "sha512-jqTqYc3prKpPiWoXV+MjlgUOinQVwp3sSFEFM+Lrsg2YFv1lmD9X0Ug+wYVrkoxCWu3XVjUxTq32b6rmUvc28g==",
+            "version": "4.15.1",
+            "resolved": "https://registry.npmjs.org/@aws/mynah-ui/-/mynah-ui-4.15.1.tgz",
+            "integrity": "sha512-ATptfkyv7L1SirkYuUPWL5OtCJZLtnkfyPC4mzBSOl/2xNh0HQtwzdRMfFt/0Gvl1x5RELaqtxMEFvtvOKqqlA==",
             "hasInstallScript": true,
             "dependencies": {
                 "escape-html": "^1.0.3",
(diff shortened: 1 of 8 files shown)


dhasani23 (approved):

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -862,16 +861,19 @@ export async function downloadResultArchive(
...
+        const downloadErrorMessage = (e as Error).message
         getLogger().error(`CodeTransformation: ExportResultArchive error = ${downloadErrorMessage}`)
         telemetry.codeTransform_logApiError.emit({
             codeTransformApiNames: 'ExportResultArchive',
             codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId(),
-            codeTransformJobId: transformByQState.getJobId(),
+            codeTransformJobId: jobId,
             codeTransformApiErrorMessage: downloadErrorMessage,
             codeTransformRequestId: e.requestId ?? '',
             result: MetadataResult.Fail,
```
arangatang, line 879: Pretty sure this will cause conflicts with this one: https://github.com/aws/aws-toolkit-vscode/pull/5044

tincheng (approved):

packages/core/src/test/amazonqGumby/transformApiHandler.test.ts
```diff
@@ -0,0 +1,228 @@
+/*!
```
damntrecky, line 1: Nice tests!

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
JuliaABurch: I left a comment on your PR as to what can be changed there to make essentially the same change I'm doing here -- otherwise I'm fine with this block of code being completely overwritten by your change when it comes time to merge

justinmk3:
> Note: in adding test coverage, I brought the test from the `amazonq` package to the `core` package, as it tests the internals of the `transformApiHandler` class, which lives in the `core` package.

Thanks for mentioning that. Can that be avoided? This is reversing the direction we want to go.

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
arangatang: Yeah I made the change your requested! Tbh we should never use globals that is not constants imo. And, if strictly required limit them to be project specific / internal to the class they are used in to scope them down. Great you are driving this.

PR #5221, opened 2024-06-25
feat: update sendFeatureDevTelemetry to include user context and opt in/out telemetry options

sannicm:
## Problem

User context and opt in/out telemetry options are needed for the sendTelemetryEvent activity to parse the event. This were missing from the initial implemention.

## Solution

It is being added the above fields to the request. It has been tested locally sending these events to the server and checking for their correct processing.

## License

### DIFF
--- a/packages/amazonq/test/unit/codewhisperer/service/codewhisperer.test.ts
+++ b/packages/amazonq/test/unit/codewhisperer/service/codewhisperer.test.ts
@@ -11,10 +11,9 @@ import {
     AuthUtil,
     codeWhispererClient,
 } from 'aws-core-vscode/codewhisperer'
-import { globals, getClientId } from 'aws-core-vscode/shared'
+import { globals, getClientId, getOperatingSystem } from 'aws-core-vscode/shared'
 import { AWSError, Request } from 'aws-sdk'
 import { createSpyClient } from 'aws-core-vscode/test'
-import * as os from 'os'
 
 describe('codewhisperer', async function () {
     let clientSpy: CodeWhispererUserClient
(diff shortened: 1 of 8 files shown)


kumsmrit (approved):

packages/core/src/shared/systemUtilities.ts
```diff
@@ -49,6 +49,17 @@ export class SystemUtilities {
     private static gitPath: string
     private static bashPath: string
 
+    public static getOperatingSystem(): 'MAC' | 'WINDOWS' | 'LINUX' {
```
jpinkney-aws, line 52: A quick search shows that this exact code is in `core/src/codewhisperer/client/codewhisperer.ts`. Can you update that class as well to use this util

packages/core/src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -64,6 +67,10 @@ export class FeatureDevClient {
         return await createFeatureDevProxyClient()
     }
 
+    public getOptOutPreference() {
```
jpinkney-aws, line 70: A quick search shows that this exact code is present in `core/src/codewhisperer/util/commonUtil.ts`. If this is shared code can we extract this.

sarayev (approved):

packages/core/src/shared/systemUtilities.ts
sannicm: Yes, I know. I specially omitted that as it is not my vertical but still thought of having this in the shared folder.

packages/core/src/amazonqFeatureDev/client/featureDev.ts
sannicm: I know as well, this one I didn't see the issue to leave it here private, also I could have just left it as a one liner in the function that makes the request to `sendTelemetryEvent`.

Anyway, for both, sure I can make the changes.

packages/amazonq/test/unit/codewhisperer/service/codewhisperer.test.ts
```diff
@@ -14,7 +14,7 @@ import {
 import { globals, getClientId } from 'aws-core-vscode/shared'
 import { AWSError, Request } from 'aws-sdk'
 import { createSpyClient } from 'aws-core-vscode/test'
-import * as os from 'os'
+import { getOperatingSystem } from '../../../../../core/dist/src/shared/telemetry/util'
```
jpinkney-aws, line 17: This test file lives in core so you'll have to export getOperatingSystem from `src/shard/index.ts` and add getOperatingSystem to the imports of `aws-core-vscode/shared`

sannicm: oh, I see

jpinkney-aws (approved):

PR #5223, opened 2024-06-25
fix(featureDev): progress UX during backend calls

kumsmrit:
## Problem
- There was some mynah update which caused the in progress message to be hidden leading to bad experience for customers as they would not be aware that their calls are in progress.

Screenshot 2024-06-24 at 10 30 02

## Solution
- Mynah changes were fixed with this PR: https://github.com/aws/mynah-ui/pull/67
- Bumped mynah-version to use latest fix

Screenshot 2024-06-25 at 13 59 51

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3336,9 +3336,9 @@
             }
         },
         "node_modules/@aws/mynah-ui": {
-            "version": "4.13.0",
-            "resolved": "https://registry.npmjs.org/@aws/mynah-ui/-/mynah-ui-4.13.0.tgz",
-            "integrity": "sha512-ZOl4hggf1M7HuHipnNxf4lnh8Dt38hFD/SZPpmL1fw2pJbT+hnIAkz6nru7wUtfUSh990PshuzLRNmVrZDPpag==",
+            "version": "4.14.0",
+            "resolved": "https://registry.npmjs.org/@aws/mynah-ui/-/mynah-ui-4.14.0.tgz",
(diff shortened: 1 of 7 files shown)


dogusata:
This version update of mynah-ui also contains following fixes

* Authenticate button and suggested follow up answer is expected to be rendered as button in the Chat UI, instead they are appearing inline as text
* Clicking a different button (i.e. "Explain Selected Code") before the "Authenticate" button appears will show the Authenticate button along with "Generating answer prompt". This "Generating answer" prompt will not disappear until the next user input
* Scrolling through the User Input message textbox in the Chat UI doesn't scroll the text, it scrolls the additional UI
* (non-blocking, nice to have) Some clipping on the left hand side when code is shown in the chat response. Toolkit does not control the rendering piece, so has to be handled MynahUI side
* Generating your answer card appears right before the help message called from `/help` command.

sannicm (approved):

jpinkney-aws:
Should we be adding changelog items for these?
>Authenticate button and suggested follow up answer is expected to be rendered as button in the Chat UI, instead they are appearing inline as text
    Clicking a different button (i.e. "Explain Selected Code") before the "Authenticate" button appears will show the Authenticate button along with "Generating answer prompt". This "Generating answer" prompt will not disappear until the next user input
    Scrolling through the User Input message textbox in the Chat UI doesn't scroll the text, it scrolls the additional UI
    (non-blocking, nice to have) Some clipping on the left hand side when code is shown in the chat response. Toolkit does not control the rendering piece, so has to be handled MynahUI side
    Generating your answer card appears right before the help message called from /help command.

Some of those seem very customer facing and would be good for users to know its fixed

dogusata (approved):

jpinkney-aws (approved):

Issue #5110, opened 2024-06-06
StepFunctions: Error checking not working as per the demo video


Elizabethbranion:
## System details (run `AWS: About` and/or `Amazon Q: About`)

-   OS: MacOS
-   Visual Studio Code version: Version: 1.89.1
-   AWS Toolkit version: v3.8.0
-   Amazon Q version: v1.7.0

## Question

Error checking not working as per the demo video:

Where Rob makes a deliberate mistake 13 min 59 seconds in - he demos the error checking functionality.

https://youtu.be/1RJtCKpdELQ?si=6zc6T_Pw8hgfWmwF&t=839

This no longer seems to work. Wondering why not.

Haven't raised an issue, as maybe that functionality has been removed i.e. this is intentional?

justinmk3:
It should work. Can you provide the AWS Toolkit logs? There may also be logs in the "ASL" channel, if you see one.

1. In the vscode [command palette](https://code.visualstudio.com/docs/getstarted/userinterface#_command-palette), run `AWS: View Logs`. This reveals the Output panel with `AWS Toolkit Logs` selected.
2. Click the gear icon and choose `Debug`, then click it again and choose `Set As Default`.
3. Perform the steps to reproduce the issue.
4. Remove any private info from the logs.

Amazon Q: View Logs

Elizabethbranion:
Sorry - I added the logging on here a while ago  and unfortunately must have neglected to click save.

I'll gather it again in a while.

In the mean time - I _have_ re-produced this on a different machine (i.e. a separated environment).

Here's a screen shot (attached). Note there is no "red underline" as there should be.

<img width="1722" alt="Screenshot 2024-06-07 at 17 43 55" src="https://github.com/aws/aws-toolkit-vscode/assets/69082436/40b951a0-b8be-4049-a1e0-74538ebf75b9">

justinmk3:
Thanks! We have narrowed down the code related to this issue and are working on a fix.

justinmk3:
Fixed in AWS Toolkit 3.12.0.

PR #5226, opened 2024-06-25
fix: step function language server activation

jpinkney-aws:
## Problem
- step function language server failed to activate

## Solution
- remove globals from the language server

## Additional investigation
- Still investigating but it looks like something between version 2.20 and 3.0.0 broke the step functions language server activation. AFAICT the error occurs because the language server was starting in another process and that process doesn't have access to vscode which globals seems to be importing. It's not exactly clear why this is caused in 3.0.0 and not early though

## License

### DIFF
--- a/packages/core/src/shared/languageServer/languageModelCache.ts
+++ b/packages/core/src/shared/languageServer/languageModelCache.ts
@@ -9,7 +9,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { TextDocument } from 'vscode-languageserver-textdocument'
-import globals from '../extensionGlobals'
 
 export interface LanguageModelCache<T> {
     get(document: TextDocument): T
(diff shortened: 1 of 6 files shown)


justinmk3:
>  looks like something between version 2.20 and 3.0.0 broke the step functions language server activation

In https://github.com/aws/aws-toolkit-vscode/commit/fa54d7c0a07a4acb8675ffe89178a1c8c146cbd5 (April 2024) the globals check became more strict. In this case, it appears to have found a legitimate issue, but somehow the failing ASL server wasn't noticed in CI.

VaidSaraswat (approved):

VaidSaraswat:
Just to be clear, this will mean customers won't be able to see the error checking if they have an `asl.json` file open from a previous session. If they were to close the file and reopen it the language server starts (which is the behaviour I verified). I will be looking into the [second issue](https://github.com/aws/aws-toolkit-vscode/issues/5110) reported where the server doesn't start from when a file is already opened.

jpinkney-aws:
> Just to be clear, this will mean customers won't be able to see the error checking if they have an asl.json file open from a previous session

I can't reproduce that anymore with this PR. This step functions activation issue has been present for ~2.5 months apparently so any step functions intellisense between toolkit 2.20.0 and now shouldn't have been working at all

nkomonen-amazon (approved):

PR #5231, opened 2024-06-26
feat(amazonqFeatureDev): improve Error messages 

tverney:
## Problem
- Currently we don't have granular error handling for some specific scenarios shown by the technical writer.

## Solution
- This PR updates those messages.
- In addition:
   - Refactor the use of simple strings to describe the "phases" and use enum instead
   - Extract the error handling in controller state to a switch/case instead if/else

## License

### DIFF
--- a/packages/amazonq/.changes/next-release/Bug Fix-b1dd8c2c-cc3d-4986-be5b-20be5ebc3925.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-b1dd8c2c-cc3d-4986-be5b-20be5ebc3925.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Update client user error messages and refactor controller error handling"
+}
--- a/packages/amazonq/.changes/next-release/Bug Fix-b1dd8c2c-cc3d-4986-be5b-20be5ebc3925.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-b1dd8c2c-cc3d-4986-be5b-20be5ebc3925.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Update client user error messages and refactor controller error handling"
+	"description": "Amazon Q /dev command: improve user error messages"
 }
--- a/packages/core/src/amazonq/webview/ui/tabs/constants.ts
+++ b/packages/core/src/amazonq/webview/ui/tabs/constants.ts
@@ -24,7 +24,7 @@ export const TabTypeDataMap: Record<TabType, TabTypeData> = {
     featuredev: {
         title: 'Q - Dev',
         placeholder: 'Describe your task or issue in as much detail as possible',
-        welcome: `Hi, I'm the Amazon Q Developer Agent for software development.
+        welcome: `Welcome to the agent for software development.
 
     I can generate code to implement new functionality across your workspace. We'll start by discussing an implementation plan, and then we can review and regenerate code based on your feedback.
 
--- a/packages/core/src/amazonq/webview/ui/tabs/constants.ts
+++ b/packages/core/src/amazonq/webview/ui/tabs/constants.ts
@@ -24,7 +24,7 @@ export const TabTypeDataMap: Record<TabType, TabTypeData> = {
     featuredev: {
         title: 'Q - Dev',
         placeholder: 'Describe your task or issue in as much detail as possible',
-        welcome: `Welcome to the agent for software development.
+        welcome: `Hi, I'm the Amazon Q Developer Agent for software development.
 
     I can generate code to implement new functionality across your workspace. We'll start by discussing an implementation plan, and then we can review and regenerate code based on your feedback.
 
--- a/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
+++ b/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
@@ -49,6 +49,7 @@ import {
     messageWithConversationId,
 } from '../../userFacingText'
 import { getWorkspaceFoldersByPrefixes } from '../../../shared/utilities/workspaceUtils'
+import { ErrorMessages } from './messenger/constants'
 
 export interface ChatControllerEventEmitters {
     readonly processHumanChatMessage: EventEmitter<any>
@@ -297,17 +298,17 @@ export class FeatureDevController {
                 switch (session?.state.phase) {
                     case DevPhase.APPROACH:
                         if (isDenyListedError) {
-                            defaultMessage = `I'm sorry, I'm having trouble generating a plan. Please try again.`
+                            defaultMessage = ErrorMessages.approachPhase.denyListedError
                         } else {
-                            defaultMessage = `I'm sorry, I ran into an issue while trying to approach the problem. Please try again.`
+                            defaultMessage = ErrorMessages.approachPhase.default
                         }
                         break
                     case DevPhase.CODEGEN:
                         if (isDenyListedError) {
-                            defaultMessage = `I'm sorry, I'm having trouble generating your code and can't continue at the moment. Please try again later, and share feedback to help me improve.`
+                            defaultMessage = ErrorMessages.codeGen.denyListedError
                             noRetries = true
                         } else {
-                            defaultMessage = `I'm sorry, I ran into an issue while trying to generate your code. Please try again.`
+                            defaultMessage = ErrorMessages.codeGen.default
                         }
                         break
                 }
--- a/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
+++ b/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
@@ -213,7 +213,7 @@ export class FeatureDevController {
         )
 
         let defaultMessage
-        const noRetries = false
+        let noRetries = false
         const isDenyListedError = denyListedErrors.some(err => errorMessage.includes(err))
 
         switch (err.code) {
@@ -305,6 +305,7 @@ export class FeatureDevController {
                     case DevPhase.CODEGEN:
                         if (isDenyListedError) {
                             defaultMessage = `I'm sorry, I'm having trouble generating your code and can't continue at the moment. Please try again later, and share feedback to help me improve.`
+                            noRetries = true
                         } else {
                             defaultMessage = `I'm sorry, I ran into an issue while trying to generate your code. Please try again.`
                         }
--- a/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
+++ b/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
@@ -46,6 +46,7 @@ import {
     sessionClosed,
     updateCode,
     logWithConversationId,
+    messageWithConversationId,
 } from '../../userFacingText'
 import { getWorkspaceFoldersByPrefixes } from '../../../shared/utilities/workspaceUtils'
 
@@ -212,8 +213,8 @@ export class FeatureDevController {
             `${featureName} request failed: ${err.cause?.message ?? err.message}`
         )
 
-        let isSendErrorMessage = true
-        let defaultMessage = ''
+        let isSendErrorMessageWithRetries = true
+        let defaultMessage
         const isDenyListedError = denyListedErrors.some(err => errorMessage.includes(err))
 
         switch (err.code) {
@@ -245,7 +246,7 @@ export class FeatureDevController {
                 this.messenger.sendAnswer({
                     type: 'answer',
                     tabID: message.tabID,
-                    message: err.message,
+                    message: err.message + messageWithConversationId(session?.conversationIdUnsafe),
                 })
                 this.messenger.sendAnswer({
                     type: 'system-prompt',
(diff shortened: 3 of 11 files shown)


packages/amazonq/.changes/next-release/Bug Fix-b1dd8c2c-cc3d-4986-be5b-20be5ebc3925.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Update client user error messages and refactor controller error handling"
```
justinmk3, line 3: Thanks for including a changelog. Please review the [changelog guidelines](https://github.com/aws/aws-toolkit-vscode/blob/915e150f50bc7a3c54d6fb2de9f4b49092c486d5/CONTRIBUTING.md#changelog). 

At least one improvement would be to  mention the feature this refers to. Another note is that refactoring probably isn't relevant to the user, thus not worth mentioning. 
```suggestion
	"description": "Amazon Q /dev command: improve user error messages"
```

packages/core/src/amazonqFeatureDev/controllers/chat/messenger/messenger.ts
```diff
@@ -85,25 +86,42 @@ export class Messenger {
...
+                if (skipTitle) {
+                    this.sendAnswer({
+                        type: 'answer',
+                        tabID: tabID,
+                        message: errorMessage,
+                    })
+                } else {
+                    this.dispatcher.sendErrorMessage(
+                        new ErrorMessage(
+                            `Sorry, we're experiencing an issue on our side. Would you like to try again?`,
```
justinmk3, line 118: Why is there so much apologizing ("Sorry") in these messages? And why are these messages so verbose? That doesn't actually help users, it gets annoying.
```suggestion
                            `We're experiencing an issue on our side. Try again?`,
```

justinmk3 (approved):
someone on your team should also review/approve this

packages/amazonq/.changes/next-release/Bug Fix-b1dd8c2c-cc3d-4986-be5b-20be5ebc3925.json
tverney: Agreed and thanks for sharing the guidelines within this changelog. I'll push the proposed updates.

packages/core/src/amazonqFeatureDev/controllers/chat/messenger/messenger.ts
tverney: Actually this is the original message we currently have, most of the scenarios this one is removed. Here we're adding just as a fallback.

tverney:
> someone on your team should also review/approve this

Yes, and thanks for this review. I'll add some to the loop.

packages/core/src/amazonq/webview/ui/tabs/constants.ts
```diff
@@ -24,7 +24,7 @@ export const TabTypeDataMap: Record<TabType, TabTypeData> = {
     featuredev: {
         title: 'Q - Dev',
         placeholder: 'Describe your task or issue in as much detail as possible',
-        welcome: `Hi, I'm the Amazon Q Developer Agent for software development.
+        welcome: `Welcome to the agent for software development.
```
chengoramazon, line 27: Why we change this welcome message back? Did you rebase the origin/main?

tverney: Added it back your version, sounds like its more updated than mine lol

packages/core/src/amazonqFeatureDev/types.ts
```diff
@@ -36,7 +42,7 @@ export enum FollowUpTypes {
     SendFeedback = 'SendFeedback',
 }
 
-export type SessionStatePhase = 'Init' | 'Approach' | 'Codegen'
+export type SessionStatePhase = DevPhase.INIT | DevPhase.APPROACH | DevPhase.CODEGEN
```
chengoramazon, line 45: nit: can we simply make the SessionStatePhase enum.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -306,19 +302,86 @@ export class FeatureDevController {
...
+                        if (isDenyListedError) {
+                            defaultMessage = `I'm sorry, I'm having trouble generating a plan. Please try again.`
+                        } else {
+                            defaultMessage = `I'm sorry, I ran into an issue while trying to approach the problem. Please try again.`
+                        }
+                        break
+                    case DevPhase.CODEGEN:
+                        if (isDenyListedError) {
+                            defaultMessage = `I'm sorry, I'm having trouble generating your code and can't continue at the moment. Please try again later, and share feedback to help me improve.`
+                            sendErrorMessage = false
```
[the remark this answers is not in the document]
tverney: Good one, it might be confusing with the sendErrorMessage in the messenger. Let me update that.

packages/core/src/amazonqFeatureDev/types.ts
tverney: I'll keep a note on that. The intent here is to have type and enums separated. For example, if we want include a new DevPhase that isn't necessarily part of SessionStatePhase. 

However if I notice that this would be exact the same through the weeks, I'll change in the next review. The idea here is to have at least the DevPhase mapped as enums and not hardcoded strings for the moment.

justinmk3: > : can we simply make the SessionStatePhase enum.

enums are discouraged in this codebase (they serve no purpose, and are basically deprecated in typescript). Algebraic types as done here are preferred.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -201,47 +206,18 @@ export class FeatureDevController {
...
-        try {
-            getLogger().debug(`${featureName}: Processing message: ${message.message}`)
+    private processErrorChatMessage = (err: any, message: any, session: Session | undefined) => {
+        const skipTitle = true
+        const errorMessage = createUserFacingErrorMessage(
+            `${featureName} request failed: ${err.cause?.message ?? err.message}`
+        )
 
-            session = await this.sessionStorage.getSession(message.tabID)
+        let isSendErrorMessage = true
```
kumsmrit, line 215: Why do we need this `isSendErrorMessage` boolean ?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -306,19 +302,86 @@ export class FeatureDevController {
...
+                            defaultMessage = `I'm sorry, I'm having trouble generating a plan. Please try again.`
+                        } else {
+                            defaultMessage = `I'm sorry, I ran into an issue while trying to approach the problem. Please try again.`
+                        }
+                        break
+                    case DevPhase.CODEGEN:
+                        if (isDenyListedError) {
+                            defaultMessage = `I'm sorry, I'm having trouble generating your code and can't continue at the moment. Please try again later, and share feedback to help me improve.`
+                            isSendErrorMessage = false
+                            this.messenger.sendAnswer({
```
kumsmrit, line 319: If we directly invoke `messenger.sendAnswer` here, we will not be able to show the sendFeedback button, as that logic exists inside `messenger.sendErrorMessage` [here](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/amazonqFeatureDev/controllers/chat/messenger/messenger.ts#L68-L86)

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
tverney: There is one scenario in the switch that we don't wanna throw the error with retries, but a final message. Maybe I can come up with a better naming for this.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -201,47 +206,18 @@ export class FeatureDevController {
...
-                    await this.onApproachGeneration(session, message.message, message.tabID)
-                    break
-                case 'Codegen':
-                    await this.onCodeGeneration(session, message.message, message.tabID)
-                    break
-            }
-        } catch (err: any) {
-            if (err instanceof ContentLengthError) {
+        switch (err.code) {
+            case ContentLengthError.name:
```
kumsmrit, line 220: Have we verified if the name property has correct Error name ?

tverney: Yes, it was tested in some few different scenarios, all return the error name.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
tverney: I've just added the sendFeedback for those scenarios.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -286,14 +264,34 @@ export class FeatureDevController {
                         },
                     ],
                 })
-            } else if (err instanceof CodeIterationLimitError) {
+                break
+            case UploadCodeError.name:
```
kumsmrit, line 268: Will this not lead to have no retries available in case of `UploadCodeError`

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -306,19 +293,78 @@ export class FeatureDevController {
...
                 this.messenger.sendErrorMessage(
-                    errorMessage,
+                    defaultMessage ? defaultMessage : errorMessage,
                     message.tabID,
-                    this.retriesRemaining(session),
-                    session?.state.phase,
-                    session?.conversationIdUnsafe
+                    noRetries ? 0 : this.retriesRemaining(session),
+                    session?.conversationIdUnsafe,
+                    !!defaultMessage
```
kumsmrit, line 321: Should this not be `session?.conversationIdUnsafe` being passed here?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -475,7 +521,6 @@ export class FeatureDevController {
                 errorMessage,
                 message.tabID,
                 this.retriesRemaining(session),
-                session?.state.phase,
```
kumsmrit, line 478: Why are we removing this param `session?.state.phase,` from being passed ?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -532,7 +577,6 @@ export class FeatureDevController {
                 createUserFacingErrorMessage(`Failed to insert code changes: ${err.message}`),
```
kumsmrit, line 577: Same as above

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
tverney: public sendErrorMessage in messenger have the following attributes:
```
errorMessage: string,
tabID: string,
retries: number,
conversationId?: string,
showDefaultMessage?: boolean
```
session?.conversationIdUnsafe is being passed as the conversationId, which match the practice being used in other places. 

Ex: 
```
private async generateCodeClicked(message: any) {...
this.messenger.sendErrorMessage(
                errorMessage,
                message.tabID,
                this.retriesRemaining(session),
                session?.conversationIdUnsafe
            )
```

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -306,19 +293,78 @@ export class FeatureDevController {
...
+                        if (isDenyListedError) {
+                            defaultMessage = ErrorMessages.approachPhase.denyListedError
+                        } else {
+                            defaultMessage = ErrorMessages.approachPhase.default
+                        }
+                        break
+                    case DevPhase.CODEGEN:
+                        if (isDenyListedError) {
+                            defaultMessage = ErrorMessages.codeGen.denyListedError
+                            noRetries = true
```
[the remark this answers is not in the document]
tverney: denyListedError in code gen state for the following below:

"I'm sorry, I'm having trouble generating your code and can't continue at the moment. Please try again later, and share feedback to help me improve."

Try again later, in the string, conflicts with a immediate option for a Retry button. Therefore, by the message, indicates that its best for the user start a new conversation instead later on.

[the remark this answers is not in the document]
kumsmrit: I think that should not be true, we should allow retries for such errors as well; we can review back the text and get it updated accordingly.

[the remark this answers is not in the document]
tverney: Synced offline, sounds like we have 2 scenarios for this one - no retries and with retry. I'll update to include both cases.

kumsmrit (approved):

Issue #5163, opened 2024-06-14
Step Functions: Intellisense does not work on state addition


marcinbelczewski:
## Steps to reproduce the issue

1. Go to 'Comand Palette' and choose 'AWS: Create new Step Functions
2. Choose 'Hello world' basic example in JSON
3. Position the cursor anywhere in the map of states as if intending to add a new state
4. Press Ctrl+Space to invoke intellisense
5. See no state proposals but rather keywords (prefixed with abc) indicating extension didn't return anything for the intellisense

**AWS Toolkit Logs:**

```
2024-06-13 16:07:16.150 [info] Log level: info
2024-06-13 16:07:16.156 [info] endpoints: retrieving AWS endpoints data
2024-06-13 16:07:16.156 [info] Telemetry product: AWS Toolkit For VS Code
2024-06-13 16:07:16.158 [info] telemetry cache not found: '/Users/marcin.belczewski/Library/Application Support/Code/User/globalStorage/amazonwebservices.aws-toolkit-vscode/telemetryCache'
2024-06-13 16:07:16.163 [info] OS: Darwin arm64 23.5.0
2024-06-13 16:07:16.163 [info] Visual Studio Code extension host:  1.90.0
2024-06-13 16:07:16.163 [info] AWS Toolkit:  3.1.0
2024-06-13 16:07:16.163 [info] node: 20.9.0
2024-06-13 16:07:16.163 [info] electron: 29.4.0
2024-06-13 16:07:16.164 [info] initialized 'submitFeedback' command with default feedback id: AWS Toolkit
2024-06-13 16:07:16.180 [info] auth: Updating connection state of profile:octo1 to authenticating
2024-06-13 16:07:16.493 [info] auth: Updating connection state of profile:octo1 to valid
2024-06-13 16:07:17.035 [error] Ec2MetadataClient failed to fetch token: TimeoutError: Socket timed out without establishing a connection ()
2024-06-13 16:07:17.544 [warning] command failed (not registered?): "aws.amazonq.refreshConnectionCallback"
2024-06-13 16:07:18.817 [warning] telemetry: invalid Metric: "stepfunctions_createStateMachineFromTemplate" emitted without the `result` property, which is always required. Consider using `.run()` instead of `.emit()`, which will set these properties automatically. See https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md#guidelines
2024-06-13 16:12:17.099 [info] telemetry: sent batch (size=12)
```

[...]

PR #5234, opened 2024-06-27
StepFunctions LSP server activates when users re-open vscode with an ASL document from a previous editing session

VaidSaraswat:
### Problem
Users who open vscode after previously editing an ASL document are not seeing error checking activate

- https://github.com/aws/aws-toolkit-vscode/issues/5163
- https://github.com/aws/aws-toolkit-vscode/issues/5110

### Solution
This fixes that behaviour where users will now see error checking activate when re-opening vscode and editing an ASL document from a previous session

### Testing
- I elected to keep the two new functions private as they only pertain to activation. Thus there are no unit tests with this particular change.

- I verified this fix by opening up a new vscode session from which I had previously been editing an ASL document. I've attached a video for proof.

- I wasn't sure if we can mock this behaviour somehow in integration tests or E2E tests so if anyone has any ideas or recommendations, I'm happy to implement them.

- [Demo](https://github.com/aws/aws-toolkit-vscode/assets/41310238/b2563bb1-8f27-4e69-9474-80a057c8bf1a)

## License

### DIFF
--- a/packages/core/src/stepFunctions/activation.ts
+++ b/packages/core/src/stepFunctions/activation.ts
@@ -42,13 +42,16 @@ export async function activate(
     // eslint-disable-next-line prefer-const
     onDidOpenAslDoc = vscode.window.onDidChangeActiveTextEditor(async e => {
         if (e?.document && ASL_FORMATS.includes(e.document.languageId)) {
-            const perflog = new PerfLog('stepFunctions: start LSP client/server')
-            await ASLLanguageClient.create(extensionContext)
-            perflog.done()
             onDidOpenAslDoc?.dispose() // Handler should only run once.
+            await startLspServer(extensionContext)
         }
     }, undefined)
     extensionContext.subscriptions.push(onDidOpenAslDoc)
+
+    if (isASLFileOpen()) {
+        onDidOpenAslDoc?.dispose()
+        await startLspServer(extensionContext)
+    }
 }
 
 /*
@@ -122,6 +125,18 @@ export function initalizeWebviewPaths(
             context.asAbsolutePath(join('resources', 'css', 'stateMachineRender.css'))
         ),
     }
+}
+
+async function startLspServer(extensionContext: vscode.ExtensionContext) {
+    const perflog = new PerfLog('stepFunctions: start LSP client/server')
+    await ASLLanguageClient.create(extensionContext)
+    perflog.done()
+}
+
+function isASLFileOpen() {
+    return vscode.window.visibleTextEditors.some(
+        textEditor => textEditor?.document && ASL_FORMATS.includes(textEditor.document.languageId)
+    )
 }
 
 function initializeCodeLens(context: vscode.ExtensionContext) {
--- a/packages/toolkit/.changes/next-release/Bug Fix-9d578109-b17a-40e9-a116-8c738b19e727.json
+++ b/packages/toolkit/.changes/next-release/Bug Fix-9d578109-b17a-40e9-a116-8c738b19e727.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Step functions language server activates when editing ASL documents from previous session"
+}


packages/core/src/stepFunctions/activation.ts
```diff
@@ -124,6 +125,19 @@ export function initalizeWebviewPaths(
     }
 }
 
+async function startLspServer(extensionContext: vscode.ExtensionContext) {
+    const perflog = new PerfLog('stepFunctions: start LSP client/server')
+    await ASLLanguageClient.create(extensionContext)
+    perflog.done()
+}
+
+function editingASLFromPrevSession() {
```
jpinkney-aws, line 134: would it make sense for this to be isASLFileOpen? (Or something like that?) The fact that its from a previous session shouldn't really matter. Technically the only thing that matters is if its open during the execution of the flow

VaidSaraswat: Yeah sure I can make that change, I think `isASLFileOpen()` makes sense.

packages/core/src/stepFunctions/activation.ts
```diff
@@ -124,6 +125,19 @@ export function initalizeWebviewPaths(
...
+    const aslTextEditors = vscode.window.visibleTextEditors.filter(
+        textEditor => textEditor?.document && ASL_FORMATS.includes(textEditor.document.languageId)
+    )
+    return aslTextEditors.length > 0
```
justinmk3, line 138: can use `some()` instead of `filter().length > 0`.

the linter will probably flag this if you rebase

justinmk3 (approved):

packages/core/src/stepFunctions/activation.ts
VaidSaraswat: Updated, thanks!

packages/core/src/stepFunctions/activation.ts
```diff
@@ -124,6 +126,19 @@ export function initalizeWebviewPaths(
...
 }
 
+async function startLspServer(extensionContext: vscode.ExtensionContext) {
+    const perflog = new PerfLog('stepFunctions: start LSP client/server')
+    await ASLLanguageClient.create(extensionContext)
+    perflog.done()
+}
+
+function isASLFileOpen() {
+    return some(
```
jpinkney-aws, line 136: I think we're trying? to remove lodash as a dependency. Are you able to do this without using it?

VaidSaraswat: I misunderstood, I believe I need to use this `some()`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some. Will update.

packages/core/src/stepFunctions/activation.ts
```diff
@@ -37,14 +37,15 @@ export async function activate(
     await registerStepFunctionCommands(extensionContext, awsContext, outputChannel)
     initializeCodeLens(extensionContext)
 
+    if (isASLFileOpen()) {
+        await startLspServer(extensionContext)
```
justinmk3, line 41: this should also call `onDidOpenAslDoc?.dispose()`, else a duplicate LSP server will start. 
```suggestion
        await startLspServer(extensionContext)
        onDidOpenAslDoc.dispose()
```

this needs to line 52

VaidSaraswat: I see, would something like this make sense then:

```
let onDidOpenAslDoc: vscode.Disposable // eslint-disable-line prefer-const
    // PERFORMANCE: Start the LSP client/server _only_ when the first ASL document is opened.
    // eslint-disable-next-line prefer-const
    onDidOpenAslDoc = vscode.window.onDidChangeActiveTextEditor(async e => {
        if (e?.document && ASL_FORMATS.includes(e.document.languageId)) {
            await startLspServer(extensionContext, onDidOpenAslDoc)
        }
    }, undefined)
    extensionContext.subscriptions.push(onDidOpenAslDoc)

    if (isASLFileOpen()) {
        await startLspServer(extensionContext, onDidOpenAslDoc)
    }
```

And we then update the `startLspServer` method to accept an `vscode.Disposable` type

```
async function startLspServer(extensionContext: vscode.ExtensionContext, onDidOpenAslDoc: vscode.Disposable) {
    const perflog = new PerfLog('stepFunctions: start LSP client/server')
    await ASLLanguageClient.create(extensionContext)
    perflog.done()
    onDidOpenAslDoc?.dispose() // Handler should only run once.
}
```

jpinkney-aws: Can't you just do something like:
```typescript
// PERFORMANCE: Start the LSP client/server _only_ when the first ASL document is opened. // eslint-disable-next-line prefer-const
const onDidOpenAslDoc = vscode.window.onDidChangeActiveTextEditor(async e => {
    if (e?.document && ASL_FORMATS.includes(e.document.languageId)) {
        await startLspServer(extensionContext, onDidOpenAslDoc)
    }
}, undefined)
extensionContext.subscriptions.push(onDidOpenAslDoc)

if (isASLFileOpen()) {
    onDidOpenAslDoc.dispose()
    await startLspServer(extensionContext)
}
```

then if a step functions file is open on activation onDidOpenASLDoc is instantly disposed (since we don't need it anymore, the server is started). If a file is not open at start then the if condition never gets hit and it waits for an open file. Rather than passing onDidOpenAslDoc through to the server start

VaidSaraswat: `await startLspServer(extensionContext, onDidOpenAslDoc)`

Did you meant to do this?:

```
await startLspServer(extensionContext)
onDidOpenAslDoc.dispose()
```

Everything else makes sense to me.

jpinkney-aws (approved):

PR #5237, opened 2024-06-27
Amazon Q : Fixing the inappropriate yellow lines shown in the editor file using CodeSnippet.

laileni-aws:
## Problem
- If user triggers a scan and modifies the code in the editor, the security issue shows yellow squiggle lines at different place in the editor.

## Solution:
- To avoid this we compare the existing code diff from editor with the codeSnippet from response. If both matches we show the issue to the user if not we drop the suggestion.
- Implemented only for Auto-Scans.

## License


packages/core/src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -109,19 +111,34 @@ function mapRawToCodeScanIssue(issue: RawCodeScanIssue): CodeScanIssue {
...
     const codeScanIssues: RawCodeScanIssue[] = JSON.parse(json)
     codeScanIssues.forEach(issue => {
-        if (codeScanIssueMap.has(issue.filePath)) {
-            const list = codeScanIssueMap.get(issue.filePath)
-            if (list === undefined) {
-                codeScanIssueMap.set(issue.filePath, [issue])
+        let shouldShow = true
+
+        if (scope === CodeWhispererConstants.CodeAnalysisScope.FILE && editor) {
+            shouldShow = issue.codeSnippet.every(code => {
```
ctlai95, line 125: IIRC doesn't the code snippet include a few lines before and after? If so we need to check only the impacted line numbers

packages/core/src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -109,19 +111,34 @@ function mapRawToCodeScanIssue(issue: RawCodeScanIssue): CodeScanIssue {
...
+        }
+
+        if (shouldShow) {
+            const filePath = issue.filePath
+            if (codeScanIssueMap.has(filePath)) {
+                const issueList = codeScanIssueMap.get(filePath)
+                if (issueList) {
+                    issueList.push(issue)
+                    codeScanIssueMap.set(filePath, issueList)
+                }
```
ctlai95, line 138: There's an else block missing, is that intentional?

```typescript
codeScanIssueMap.set(filePath, [issue])
```

packages/core/src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -109,19 +111,34 @@ function mapRawToCodeScanIssue(issue: RawCodeScanIssue): CodeScanIssue {
     }
 }
 
-function mapToAggregatedList(codeScanIssueMap: Map<string, RawCodeScanIssue[]>, json: string) {
+function mapToAggregatedList(
```
ctlai95, line 114: please add some unit tests for the new logic

packages/core/src/codewhisperer/service/securityScanHandler.ts
laileni-aws: Right now,
I am checking all the text provided in the code snippet. Should I restrict this to only the impacted line numbers ?

packages/core/src/codewhisperer/service/securityScanHandler.ts
laileni-aws: I think Both code snippet does the same functionality

packages/core/src/codewhisperer/models/errors.ts
```diff
@@ -30,8 +30,8 @@ export class ProjectSizeExceededError extends SecurityScanError {
 }
 
 export class DefaultError extends SecurityScanError {
-    constructor() {
-        super('Security scan failed.', 'DefaultError', DefaultCodeScanErrorMessage)
+    constructor(error: string | undefined) {
+        super(`Security scan failed. ${error}`, 'DefaultError', DefaultCodeScanErrorMessage)
```
justinmk3, line 34: why not pass the error as the `cause` field? also consider `ToolkitError.chain()` .

strinigifying errors randomly in messages completely defeats the builtin mechanisms in the extension such as : https://github.com/aws/aws-toolkit-vscode/blob/6300d520ac347b7596d98c7cec312b67f98ee528/packages/core/src/shared/errors.ts#L467

and makes it much harder to show useful errors to users, in reporting, etc. Please, use care when designing errors.

laileni-aws: These errors are not shown to the user. This is for telemetry purpose only.

laileni-aws: This is redundant change. Reverted acc in draft CR

packages/core/src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -109,19 +111,32 @@ function mapRawToCodeScanIssue(issue: RawCodeScanIssue): CodeScanIssue {
...
+                }
             }
+        }
+        return true
+    })
+
+    filteredIssues.forEach(issue => {
+        const filePath = issue.filePath
+        if (codeScanIssueMap.has(filePath)) {
+            codeScanIssueMap.get(filePath)!.push(issue)
```
ctlai95, line 137: Can we avoid using `!`? I think it is not guaranteed that the map entry contains an array type

ctlai95 (approved):

packages/core/src/codewhisperer/models/errors.ts
justinmk3: > These errors are not shown to the user

This cannot be assumed. Codepaths can change. Errors should use best practices.

justinmk3:
Does this have a customer-facing impact that can be described in the changelog?

1. Review the [changelog guidelines](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#changelog).
2. Add a new changelog entry: `npm run newChange -w packages/amazonq` or `npm run newChange -w packages/toolkit`

PR #5243, opened 2024-06-27
auth: add telemetry for why the user needed to reauth

nkomonen-amazon:
## Problem:
    
In the reauth metric (`aws_loginWithBrowser` + `isReAuth: true`) we didn't know why
the user needed to reauth. It could have been due to regular expiration or some unexpected
error.

## Solution:

**See each commit for specific change**

Add a new field in `aws_loginWithBrowser` named `reAuthReason` which will have some sort
of identifier for why reauth was needed.

What was actually changed in this PR:
- `SsoAccessTokenProvider` has an object `ReAuthState` (name may change) which maps an SSO connection -> reason why we need to reauth
  - In certain flows such as SSO token refresh, if it fails we will update the `ReAuthState` with the reason why.
  - Later on the user may want to reauthenticate, and when they start the flow `SsoAccessTokenProvider` will know the reason and record it in `aws_loginWithBrowser`
- `isReAuth` is now an argument in `createToken()`
  - This allows us to get rid of some hacky telemetry code where we tried to essentially set `isReAuth: true` globally (through `record()`)
  - Now that it is an argument we pass `isReAuth` throughout the execution, this is more readable. 
- A generalized Map class was created
  - This maps a key to some data
  - We used this implementation with `ReAuthState`
  - It can probably be reused in other places of our code and reduce duplication of code

## License

### DIFF
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -183,11 +183,11 @@ export class Auth implements AuthService, ConnectionManager {
         if (profile.type === 'sso') {
             const provider = this.getSsoTokenProvider(id, profile)
 
-            // We cannot easily set isReAuth inside the createToken() call,
+            // We cannot easily set reauth related fields inside the createToken() call,
             // so we need to set it here.
             await telemetry.aws_loginWithBrowser.run(async span => {
                 span.record({ credentialStartUrl: profile.startUrl })
-                await this.authenticate(id, () => provider.createToken({isReAuth: true}), shouldInvalidate)
+                await this.authenticate(id, () => provider.createToken({ isReAuth: true }), shouldInvalidate)
             })
 
             return this.getSsoConnection(id, profile)
@@ -386,7 +386,7 @@ export class Auth implements AuthService, ConnectionManager {
             throw new ToolkitError('Auth: Cannot force expire an IAM connection')
         }
         const provider = this.getSsoTokenProvider(conn.id, profile)
-        await provider.invalidate()
+        await provider.invalidate('devModeManualExpiration')
         // updates the state of the connection
         await this.refreshConnectionState(conn)
     }
@@ -530,7 +530,8 @@ export class Auth implements AuthService, ConnectionManager {
 
             // XXX: never drop tokens in a dev environment, unless you are Amazon Q!
             if (getCodeCatalystDevEnvId() === undefined || isAmazonQ()) {
-                await provider.invalidate()
+                // TODO: Require invalidateConnection() to require a reason and then pass it in to the following instead
+                await provider.invalidate('explicitInvalidation')
             }
         } else if (profile.type === 'iam') {
             globals.loginManager.store.invalidateCredentials(fromString(id))
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -185,9 +185,9 @@ export class Auth implements AuthService, ConnectionManager {
 
             // We cannot easily set isReAuth inside the createToken() call,
             // so we need to set it here.
-            await telemetry.aws_loginWithBrowser.run(async (span) => {
-                span.record({ isReAuth: true, credentialStartUrl: profile.startUrl })
-                await this.authenticate(id, () => provider.createToken(), shouldInvalidate)
+            await telemetry.aws_loginWithBrowser.run(async span => {
+                span.record({ credentialStartUrl: profile.startUrl })
+                await this.authenticate(id, () => provider.createToken({isReAuth: true}), shouldInvalidate)
             })
 
             return this.getSsoConnection(id, profile)
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -182,13 +182,7 @@ export class Auth implements AuthService, ConnectionManager {
         const profile = this.store.getProfileOrThrow(id)
         if (profile.type === 'sso') {
             const provider = this.getSsoTokenProvider(id, profile)
-
-            // We cannot easily set reauth related fields inside the createToken() call,
-            // so we need to set it here.
-            await telemetry.aws_loginWithBrowser.run(async span => {
-                span.record({ credentialStartUrl: profile.startUrl })
-                await this.authenticate(id, () => provider.createToken({ isReAuth: true }), shouldInvalidate)
-            })
+            await this.authenticate(id, () => provider.createToken({ isReAuth: true }), shouldInvalidate)
 
             return this.getSsoConnection(id, profile)
         } else {
--- a/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
+++ b/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
@@ -204,7 +204,7 @@ export abstract class SsoAccessTokenProvider {
         func: T,
         args?: CreateTokenArgs
     ): ReturnType<T> {
-        return telemetry.aws_loginWithBrowser.run(span => {
+        return telemetry.aws_loginWithBrowser.run((span) => {
             span.record({
                 credentialStartUrl: this.profile.startUrl,
                 source: SsoAccessTokenProvider._authSource,
--- a/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
+++ b/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
@@ -103,8 +103,8 @@ export abstract class SsoAccessTokenProvider {
         }
     }
 
-    public async createToken(): Promise<SsoToken> {
-        const access = await this.runFlow()
+    public async createToken(args?: CreateTokenArgs): Promise<SsoToken> {
+        const access = await this.runFlow(args)
         const identity = this.tokenCacheKey
         await this.cache.token.save(identity, access)
         await globals.globalState.setSsoSessionCreationDate(this.tokenCacheKey, new globals.clock.Date())
@@ -112,10 +112,10 @@ export abstract class SsoAccessTokenProvider {
         return { ...access.token, identity }
     }
 
-    private async runFlow() {
+    private async runFlow(args?: CreateTokenArgs) {
         const registration = await this.getValidatedClientRegistration()
         try {
-            return await this.authorize(registration)
+            return await this.authorize(registration, args)
         } catch (err) {
             if (err instanceof SSOOIDCServiceException && isClientFault(err)) {
                 await this.cache.registration.clear(
@@ -185,11 +185,15 @@ export abstract class SsoAccessTokenProvider {
     /**
      * Wraps the given function with telemetry related to the browser login.
      */
-    protected withBrowserLoginTelemetry<T extends (...args: any[]) => any>(func: T): ReturnType<T> {
+    protected withBrowserLoginTelemetry<T extends (...args: any[]) => any>(
+        func: T,
+        args?: CreateTokenArgs
+    ): ReturnType<T> {
         const run = (span: Metric<AwsLoginWithBrowser>) => {
             span.record({
                 credentialStartUrl: this.profile.startUrl,
                 source: SsoAccessTokenProvider._authSource,
+                isReAuth: args?.isReAuth,
             })
 
             // Reset source in case there is a case where browser login was called but we forgot to set the source.
@@ -211,7 +215,10 @@ export abstract class SsoAccessTokenProvider {
         })
     }
 
-    protected abstract authorize(registration: ClientRegistration): Promise<{
+    protected abstract authorize(
+        registration: ClientRegistration,
+        args?: CreateTokenArgs
+    ): Promise<{
         token: SsoToken
         registration: ClientRegistration
         region: string
@@ -249,6 +256,15 @@ export abstract class SsoAccessTokenProvider {
         }
         return new AuthFlowAuthorization(profile, cache, oidc)
     }
+}
+
+/**
+ * Supplementary arguments for the create token flow. This data can be used
+ * for things like telemetry.
+ */
+type CreateTokenArgs = {
+    /** true if the create token flow is for reauthentication */
+    isReAuth?: boolean
 }
 
 const backoffDelayMs = 5000
@@ -377,7 +393,8 @@ export class DeviceFlowAuthorization extends SsoAccessTokenProvider {
     }
 
     override async authorize(
-        registration: ClientRegistration
+        registration: ClientRegistration,
+        args?: CreateTokenArgs
     ): Promise<{ token: SsoToken; registration: ClientRegistration; region: string; startUrl: string }> {
         // This will NOT throw on expired clientId/Secret, but WILL throw on invalid clientId/Secret
         const authorization = await this.oidc.startDeviceAuthorization({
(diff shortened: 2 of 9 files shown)


packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -663,9 +674,52 @@ class WebAuthorization extends SsoAccessTokenProvider {
...
+/**
+ * Remembers the reason an SSO session was put in to a "needs reauthentication" state.
+ * The current use is for telemetry. When the user reauths, we want {@link AwsLoginWithBrowser}
+ * to know why it needed to be reauthed.
+ *
+ * The flow is to use `setReason()` to remember why the user was put in to a reauth state,
+ * then upon the next reauth use `getReason()`. Finally, use `clearReason()` if the reauth is
+ * successful.
+ */
+export class ReAuthReasonState {
```
justinmk3, line 693: is a class the right abstraction for this, or is "auth state" something that can/should be generalized in auth.ts?

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -196,7 +202,12 @@ export abstract class SsoAccessTokenProvider {
             // We don't want to attribute the wrong source.
             SsoAccessTokenProvider.authSource = 'unknown'
 
-            return func()
+            const result = await func()
+
+            // The browser login process has succeeded, the reason is now stale
+            await ReAuthReasonState.instance.clearReason(this.profile)
```
justinmk3, line 208: is it possible this doesn't get called? then the "wrong" state may be used in the future

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -663,9 +674,52 @@ class WebAuthorization extends SsoAccessTokenProvider {
...
+    private readonly memento: vscode.Memento
+    static readonly reasonNotSetState = 'reasonNotSet'
+
+    protected constructor(memento: vscode.Memento) {
+        this.memento = memento
+    }
+
+    static #instance: ReAuthReasonState
+    static get instance() {
+        return (this.#instance ??= new ReAuthReasonState(globals.context.globalState))
```
justinmk3, line 703: why use globalState instead of just a variable? what scenarios are we anticipating here?

packages/core/src/auth/auth.ts
```diff
@@ -186,7 +186,14 @@ export class Auth implements AuthService, ConnectionManager {
             // We cannot easily set isReAuth inside the createToken() call,
             // so we need to set it here.
             await telemetry.aws_loginWithBrowser.run(async span => {
-                span.record({ isReAuth: true, credentialStartUrl: profile.startUrl })
+                span.record({
+                    isReAuth: true,
+                    credentialStartUrl: profile.startUrl,
+                    reAuthReason: await ReAuthReasonState.instance.getReason({
```
justinmk3, line 192: this has the same risk as other global state, which e.g. codewhisperer has suffered from. 

are we certain this can't be achieved with a more direct plumbing-through of the "reauth reason", instead of setting global mutable state?

justinmk3:
Thought experiement: in the places where this PR currently calls, `ReAuthReasonState.setReason`, why can't we use `telemetry.aws_loginWithBrowser.record({reAuthReason: ...})` instead?

If the answer is "because the execution context might be lost by the time we emit `telemetry.aws_loginWithBrowser`, then next questions are:

- can we change it so that the `telemetry.xx.record()` state that is built-up, is preserved until `telemetry.xx.run()` is called?
- or in the worst case, we could add a new function `telemetry.xx.recordNext()` which maintains state that is used for whenever the next `xx` metric is emitted.

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
nkomonen-amazon: If they are put in to a reauth state we do not know when they will reauth next. An IDE reload or extension update would wipe the state if it was in a variable.

justinmk3: If the IDE reloads or restarts, is it actually valid to say that the "reauth reason" is something that happened minutes or hours ago?

nkomonen-amazon: Yes, I would say so. It is the reason their connection was put in to a reauth state. Days or weeks later the cause is still the same, only once they reauth do we consider it stale.

justinmk3: How common is the scenario that a vscode restart would happen after a reauth? It seems like an accident, and not connected to the reauth. Normally, the reauth will happen in the same session, so storing it in memory will capture enough data to be representative of the problems that we want to notice. 

If that's correct, then we don't need to store this in globalState, and that opens the door for using telemetry.record() as described in my other comments in this PR.

nkomonen-amazon: If we are fine with assumption, then yes we can ignore storing it in globalState.

As for the implementation I will look in to something you mentioned like `telemetry.xx.recordNext()` as the telemetry context will be different and `record()` will not work as it currently is

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -372,14 +389,6 @@ function getSessionDuration(id: string) {
  *         - RefreshToken (optional)
  */
 export class DeviceFlowAuthorization extends SsoAccessTokenProvider {
-    constructor(
```
nkomonen-amazon, line 359: We can inherit the parent constructor

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -67,16 +68,18 @@ export abstract class SsoAccessTokenProvider {
     public constructor(
         protected readonly profile: Pick<SsoProfile, 'startUrl' | 'region' | 'scopes' | 'identifier'>,
         protected readonly cache = getCache(),
-        protected readonly oidc: OidcClient = OidcClient.create(profile.region)
+        protected readonly oidc: OidcClient = OidcClient.create(profile.region),
+        protected readonly reAuthState: ReAuthState = ReAuthState.instance
     ) {}
 
-    public async invalidate(): Promise<void> {
+    public async invalidate(reason: string): Promise<void> {
```
jpinkney-aws, line 75: would it be possible to pass in a type that contains all the possible options? I think it might help to see all the possible invalidation reasons enumerated

nkomonen-amazon: After looking at this again, some places prepend data to the string to give more context. Because of this it doesn't work nicely.

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -651,9 +652,46 @@ class WebAuthorization extends SsoAccessTokenProvider {
...
+ * The current use is for telemetry. When the user reauths, we want {@link AwsLoginWithBrowser}
+ * to know why it needed to be reauthed.
+ *
+ * The flow is to use `set()` to remember why the user was put in to a reauth state,
+ * then upon the next reauth use `get()`. Finally, use `clear()` if the reauth is
+ * successful.
+ */
+export class ReAuthState extends MemoryMap<ReAuthStateKey, ReAuthStateValue> {
+    static #instance: ReAuthState
+    static get instance() {
```
justinmk3, line 673: since ReAuthState lives as `this.reauthState` on the sso objects, can we eliminate this singleton to avoid accidental usage
```suggestion
```

packages/core/src/shared/utilities/map.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Simply maps a Key to an Object. Everything is done in memory, allowing for atomic updates.
+ *
+ * There is a common use case for this class, so look to reuse this when possible.
+ * Benefits, instead of making your own:
+ * - A key does not have to be a string, {@link MemoryMap.asKey} enables this
+ * - Implements most of the boilerplate
+ */
+export abstract class MemoryMap<Key, Value = { [key: string]: any }> implements MapSync<Key, Value> {
```
justinmk3, line 16: most data is in-memory, so it seems confusing to call that out. if we add a non-memory variant then its name could call that out, since it would be less common.

```suggestion
export abstract class NestedMap<Key, Value = { [key: string]: any }> implements MapSync<Key, Value> {
```

packages/core/src/shared/utilities/map.ts
```diff
@@ -0,0 +1,98 @@
...
+    protected abstract get default(): Value
+}
+
+/**
+ * A synchronous version of a Map of Maps.
+ * - This allows for atomic updates as it is not async.
+ * - They Key does not need to be a string
+ */
+export interface MapSync<Key, ReturnObject = { [key: string]: any }> {
+    /**
```
justinmk3, line 66: What is this abstraction communicating? "not async" implies there is an async variant (where?).

Is there a reason this doesn't derive from the ES `Map` interface?

packages/core/src/shared/utilities/map.ts
```diff
@@ -0,0 +1,98 @@
...
+    clear(key: Key, reason: string): void {
+        delete this.data[this.asKey(key)]
+        getLogger().debug(`${this.name}: cleared cache, key: ${JSON.stringify(key)}, reason: ${reason}`)
+    }
+
+    /**
+     * Converts the user-provided key to a string so that it can be used
+     * as an object key.
+     */
+    protected abstract asKey(key: Key): string
```
justinmk3, line 47: seems to overlap with `TelemetryDebounceInfo` https://github.com/aws/aws-toolkit-vscode/blob/d49eee404d1d65332c955c55d907116ae870e378/packages/core/src/shared/vscode/commands2.ts#L605

can we reconcile them? maybe we should start with the interface defined by `TelemetryDebounceInfo` and see what the delta is.

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
nkomonen-amazon: Nope, since the SsoAccessTokenProvider instance is created at the moment of use in different flows: https://github.com/aws/aws-toolkit-vscode/blob/b0965b7b768c957d740511ec2fd76a658794807a/packages/core/src/auth/auth.ts#L701.

We could move ReAuthState in to `auth.ts`, but I'd like to avoid that for now to keep ReAuthState contained to a single file.

packages/core/src/shared/utilities/map.ts
nkomonen-amazon: This makes sense. I don't want it to block this PR anymore. I'll look to do this after we merge.

nkomonen-amazon:
@justinmk3 can you take a final look when you have time

packages/core/src/shared/utilities/map.ts
justinmk3: should this live in `cacheUtils.ts` ? it seems very related.

justinmk3: > There are some differences with the Map interface, 

is that difference big enough to lose the discoverability/intuitiveness gained by implementing `Map` ? 

> such as not allowing partial values in `set()` that wouldn't allow it to be used.

I am wondering about this. Is it just a typing issue? I think we can work around it.

packages/core/src/shared/utilities/cacheUtils.ts
```diff
@@ -10,13 +10,16 @@ import fs from '../../shared/fs/fs'
...
+import type { MapSync } from './map'
 
 // TODO(sijaden): further generalize this concept over async references (maybe create a library?)
 // It's pretty clear that this interface (and VSC's `Memento`) reduce down to what is essentially
 // a "pointer" with RW operations that return futures.
 
 /**
  * A general, basic interface for a cache with key associativity.
+ *
+ * Look to use {@link MapSync} instead if you need atomicity.
```
justinmk3, line 22: what indicates that `KeyedCache` doesn't have atomicity? can be a followup discussion, but... this isn't clear.

packages/core/src/shared/utilities/map.ts
```diff
@@ -0,0 +1,104 @@
...
+
+import { getLogger } from '../logger'
+import type { KeyedCache } from './cacheUtils'
+
+/**
+ * Simply maps a Key to an Object. Everything is done in memory, allowing for atomic updates.
+ *
+ * There is a common use case for this class, so look to reuse this when possible.
+ * Benefits, instead of making your own:
+ * - A key does not have to be a string, {@link NestedMap.asKey} enables this
```
justinmk3, line 14: (followup) this is called "hash" usually, isn't it? we can look for ways to give clues to other developers, by using common terms. 

```suggestion
 * - A key does not have to be a string, {@link NestedMap.hash} enables this
```

packages/core/src/test/credentials/auth.test.ts
```diff
@@ -256,15 +256,10 @@ describe('Auth', function () {
...
+        it('reauthentication flag is set at start of reauth process', async function () {
             const conn = await auth.createInvalidSsoConnection(ssoProfile)
             await auth.reauthenticate(conn)
-            assertTelemetry('aws_loginWithBrowser', {
-                result: 'Succeeded',
-                isReAuth: true,
-                credentialStartUrl: ssoProfile.startUrl,
-            })
-            assert.strictEqual(getMetrics('aws_loginWithBrowser').length, 1)
+            auth.getTestTokenProvider(conn).createToken.calledWith({ isReAuth: true })
```
justinmk3, line 262: can't we test the actual metric, as before?

packages/core/src/test/credentials/sso/ssoAccessTokenProvider.test.ts
```diff
@@ -233,6 +235,65 @@ describe('SsoAccessTokenProvider', function () {
...
         })
 
+        describe('telemetry: ReAuthReason during SSO flow', () => {
+            // Note in the future we should use an `identifier` instead of `startUrl` for this test
+            beforeEach(() => {
+                setupFlow()
+                stubOpen()
+            })
+
+            it('does not clear the reAuthReason on failed login', async () => {
```
justinmk3, line 245: we have existing tests for the various auth cases. can we just add asserts about reAuth in those existing tests? reauth itself is not a topic, it's just a field. 

It helps connect concepts if auth fields are all tested in the auth tests, instead of having separate test groups for every piece of auth state.

justinmk3 (approved):
approving to include in today's release, comments can be a fast-follow

packages/core/src/test/credentials/auth.test.ts
nkomonen-amazon: This suite already stubs the underlying sso token provider `createToken()` in the test so no telemetry calls are actually made in this case.

This test only existed due to the `aws_loginWithBrowser` hack we previously used in `auth.ts`. Now that all telemetry is isolated to `SsoAccessTokenProvider.ts` telemetry is asserted in the SsoTokenProvider tests.

packages/core/src/test/credentials/sso/ssoAccessTokenProvider.test.ts
nkomonen-amazon: Updated

packages/core/src/shared/utilities/cacheUtils.ts
nkomonen-amazon: Making updates requires async read() then write(). So by the time you write the data may have changed due to async. But since MapSync does not use promises the user can read the latest data, then update it atomically.

PR #5245, opened 2024-06-28
fix: handling validation errors in approach/code generation

gasolima:
## Problem
When a validation errors(repo size exceeds limit or corrupted zip file) thrown from approach/code generation it will show generic internal error to the user
## Solution
Catch those errors and show proper errors

Screenshot 2024-06-28 at 08 44 13
Screenshot 2024-06-28 at 08 42 34

    REMINDER:
    - Read CONTRIBUTING.md first.
    - Add test coverage for your changes.
    - Update the changelog using `npm run newChange`.
    - Link to related issues/commits.
    - Testing: how did you test your changes?
    - Screenshots (if the pull request is related to UI/UX then please include light and dark theme screenshots)
-->

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/client/featureDev.ts
+++ b/packages/core/src/amazonqFeatureDev/client/featureDev.ts
@@ -20,6 +20,7 @@ import {
     MonthlyConversationLimitError,
     PlanIterationLimitError,
     UnknownApiError,
+    ZipFileError,
 } from '../errors'
 import { ToolkitError, isAwsError, isCodeWhispererStreamingServiceException } from '../../shared/errors'
 import { getCodewhispererConfig } from '../../codewhisperer/client/codewhisperer'
@@ -27,6 +28,8 @@ import { LLMResponseType } from '../types'
 import { createCodeWhispererChatStreamingClient } from '../../shared/clients/codewhispererChatClient'
 
 // Create a client for featureDev proxy client based off of aws sdk v2
+const zippedFileError = 'zipped file is corrupted';
+const repoSizeError = 'repo size is exceeding the limits';
 export async function createFeatureDevProxyClient(): Promise<FeatureDevProxyClient> {
     const bearerToken = await AuthUtil.instance.getBearerToken()
     const cwsprConfig = getCodewhispererConfig()
@@ -184,6 +187,13 @@ export class FeatureDevClient {
                     e.name === 'ServiceQuotaExceededException'
                 ) {
                     throw new PlanIterationLimitError()
+                } else if (
+                    e.name === 'ValidationException' &&
+                    e.message.includes(repoSizeError)
+                ) {
+                    throw new ContentLengthError()
+                } else if (e.name === 'ValidationException' && e.message.includes(zippedFileError)) {
+                    throw new ZipFileError()
                 }
                 throw new ApiError(
                     e.message,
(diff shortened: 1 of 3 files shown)


packages/core/src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -184,6 +185,11 @@ export class FeatureDevClient {
                     e.name === 'ServiceQuotaExceededException'
                 ) {
                     throw new PlanIterationLimitError()
+                } else if (e.message.includes('The repo size is exceeding the limits.')) {
+                    throw new ContentLengthError()
+                } else if (e.message.includes('The zipped file is corrupted.')) {
+                    getLogger().error('throw the zip error')
```
kumsmrit, line 191: Do we need this logger here ?

packages/core/src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -184,6 +185,11 @@ export class FeatureDevClient {
                     e.name === 'ServiceQuotaExceededException'
                 ) {
                     throw new PlanIterationLimitError()
+                } else if (e.message.includes('The repo size is exceeding the limits.')) {
```
kumsmrit, line 188: Should we only rely on error message or add check for exception name as well ?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -306,6 +307,14 @@ export class FeatureDevController {
                         },
                     ],
                 })
+            } else if (err instanceof ZipFileError) {
+                this.messenger.sendErrorMessage(
+                    '',
+                    message.tabID,
+                    0,
```
kumsmrit, line 314: Is it good to not provide retry for this, as this conversation will be counted ?

packages/core/src/amazonqFeatureDev/constants.ts
```diff
@@ -15,7 +15,7 @@ export const featureDevChat = 'featureDevChat'
 export const featureName = 'Amazon Q Developer Agent for software development'
 
 // Max allowed size for file collection
-export const maxRepoSizeBytes = 200 * 1024 * 1024
+export const maxRepoSizeBytes = 250 * 1024 * 1024
```
kumsmrit, line 18: Why are we changing this ?

packages/core/src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -230,6 +238,18 @@ export class FeatureDevClient {
                     e.code === 'ServiceQuotaExceededException')
             ) {
                 throw new CodeIterationLimitError()
+            } else if (
+                isAwsError(e) &&
+                e.name === 'ValidationException' &&
```
kumsmrit, line 243: For `CodewhispererRuntimeException`, we will have to use `e.code` for verifying exception type.

packages/core/src/amazonqFeatureDev/client/featureDev.ts
gasolima: Nope removing it.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
gasolima: As many retries the user will do will not help, since this only can happen when our toolkit has a bug, then we need to have another version of the toolkit

packages/core/src/amazonqFeatureDev/constants.ts
gasolima: This shouldn't be committed, removing it from the CR

packages/core/src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -184,6 +185,13 @@ export class FeatureDevClient {
                     e.name === 'ServiceQuotaExceededException'
                 ) {
                     throw new PlanIterationLimitError()
+                } else if (
+                    e.name === 'ValidationException' &&
+                    e.message.includes('repo size is exceeding the limits')
```
jpinkney-aws, line 190: it looks like this same message is being used twice in two different places. Can we extract it somewhere?

Same with `zipped file is corrupted`

gasolima: sure good idea

kumsmrit (approved):

justinmk3:
CI failing. Also needs a rebase.

PR #5248, opened 2024-06-28
web: Fix failing web mode due to `fs-extra` import

nkomonen-amazon:
## Problem:

When importing the codewhisperer exports module from index.ts in web mode
it was breaking due to a transitive fs-extra import.

## Solution:

Export a separate module from indexCommon.ts for code that works in web
or node. This file is a subset of index.ts, so behavior remains the same.

## License


packages/core/src/codewhisperer/indexCommon.ts
```diff
@@ -0,0 +1,14 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * This module is a subset of `./index.js` except it is for 'common' code only.
+ *
+ * See "Shared vs Common" in our docs/ folder for the meaning of 'common'
```
justinmk3, line 9: I don't think we should continue this pattern. The fact that it requires explanation is one signal that it's not the right pattern. 

If we need node-specific or web-specific code, create a `fooNode.ts` module or `node/` or `web/` directory. The "common" pattern will be hard to untangle later. We should not continue it.

nkomonen-amazon:
@justinmk3 can we merge this now to unblock people while I make the changes we previously discussed

justinmk3 (approved):

PR #5252, opened 2024-07-01
fix(gumby): Do not check auth when initializing transformation hub UX

JuliaABurch:
## Problem
There was a bug that happened when VSCode is opened without having first been signed into `Amazon Q` and the user attempts a code transformation. The Transformation Hub window would open, but would not update, and Code Transform would appear to hang.

## Solution

There is some initialization code that registers commands that other parts of Gumby uses to update the Transformation Hub, but this code was locked behind an auth check; if it fails, the commands are simply never registered, even if the user signs in again later.

This portion of the code is not necessary to lock behind an auth check (it doesn't save memory or make anything run faster to not have it run), so the check has been removed.

Testing:
1. User opens VSCode without already being signed in; the user signs in and then transforms a project successfully
2. User opens VSCode without already being signed in; the user signs in and then transforms a project unsuccessfully (partial success)
3. User opens VSCode without already being signed in; the user signs in and then is not able to submit a project to be transformed (fails to build locally)
4. User opens VSCode without already being signed in; the user signs in and then transforms a project that triggers HIL
5. User opens VSCode already having been signed in; the user transforms a project successfully

## License

### DIFF
--- a/packages/core/src/amazonqGumby/activation.ts
+++ b/packages/core/src/amazonqGumby/activation.ts
@@ -13,62 +13,59 @@ import { ProposedTransformationExplorer } from '../codewhisperer/service/transfo
 import { CodeTransformTelemetryState } from './telemetry/codeTransformTelemetryState'
 import { telemetry } from '../shared/telemetry/telemetry'
 import { CancelActionPositions } from './telemetry/codeTransformTelemetry'
-import { AuthUtil } from '../codewhisperer/util/authUtil'
 
 export async function activate(context: ExtContext) {
     void vscode.commands.executeCommand('setContext', 'gumby.wasQCodeTransformationUsed', false)
-    // If the user is codewhisperer eligible, activate the plugin
-    if (AuthUtil.instance.isValidCodeTransformationAuthUser()) {
-        const transformationHubViewProvider = new TransformationHubViewProvider()
-        new ProposedTransformationExplorer(context.extensionContext)
-        // Register an activation event listener to determine when the IDE opens, closes or users
-        // select to open a new workspace
-        const workspaceChangeEvent = vscode.workspace.onDidChangeWorkspaceFolders(event => {
-            // A loophole to register the IDE closed. This is when no folders were added nor
-            // removed, but the event still fired. This assumes the user closed the workspace
-            if (event.added.length === 0 && event.removed.length === 0) {
-                // Only fire closed during running/active job status
-                if (transformByQState.isRunning()) {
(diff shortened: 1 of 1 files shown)


packages/core/src/amazonqGumby/activation.ts
```diff
@@ -13,62 +13,59 @@ import { ProposedTransformationExplorer } from '../codewhisperer/service/transfo
 import { CodeTransformTelemetryState } from './telemetry/codeTransformTelemetryState'
 import { telemetry } from '../shared/telemetry/telemetry'
 import { CancelActionPositions } from './telemetry/codeTransformTelemetry'
-import { AuthUtil } from '../codewhisperer/util/authUtil'
 
 export async function activate(context: ExtContext) {
     void vscode.commands.executeCommand('setContext', 'gumby.wasQCodeTransformationUsed', false)
-    // If the user is codewhisperer eligible, activate the plugin
-    if (AuthUtil.instance.isValidCodeTransformationAuthUser()) {
```
JuliaABurch, line 21: The only change is that this line has been removed; the rest of the changes are formatting / whitespace related

dhasani23: Does this mean that the Transformation Hub will always show up, even for users who haven't interacted with QCT? I remember adding this line a while ago because we received some feedback that the Transformation Hub shouldn't show up unless a user has interacted with QCT.

dhasani23:
Can you also test the case where you open VS Code already signed in, then sign out, then sign in again and run `/transform`?

packages/core/src/amazonqGumby/activation.ts
JuliaABurch: I tested this scenario, let me know if it covers the edge case you're concerned about:

1. Open VSCode, sign into Amazon Q
2. I go to the lower windows (where terminal, etc) are located. `Transformation Hub` is not visible
3. I click `...` but `Transformation Hub` is not an option there
4. It is only when I successfully submit a project to be transformed that `Transformation Hub` becomes a tab in this collection of lower windows
5. If I close `Transformation Hub` after it is open, I can only open it from the Chat (selecting `Show Transformation Hub`)

dhasani23 (approved):

JuliaABurch:
> Can you also test the case where you open VS Code already signed in, then sign out, then sign in again and run `/transform`?

Yes, I just tested this and this works as well.

justinmk3:
Seems like this has user impact, do we need a changelog item?

PR #5256, opened 2024-07-02
fix: app composer generate suggestions unclickable when logged in

jpinkney-aws:
## Problem
- app composers generate suggestions became unclickable when we switched back to seperate auth since they were relying on auth session sharing

## Solution
- allow extensions to call into auth/chat for amazon q. This allows toolkit to have features that depend on q auth without exposing all of auth itself

## License

### DIFF
--- a/packages/amazonq/src/api.ts
+++ b/packages/amazonq/src/api.ts
@@ -1,0 +1,28 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { GenerateAssistantResponseCommandOutput, GenerateAssistantResponseRequest } from '@amzn/codewhisperer-streaming'
+import { AuthUtil } from 'aws-core-vscode/codewhisperer'
+import { ChatSession } from 'aws-core-vscode/codewhispererChat'
+import { api } from 'aws-core-vscode/amazonq'
+
+export default {
+    chatApi: {
+        async chat(request: GenerateAssistantResponseRequest): Promise<GenerateAssistantResponseCommandOutput> {
+            const chatSession = new ChatSession()
+            return chatSession.chat(request)
+        },
+    },
+    authApi: {
+        async reauthIfNeeded() {
+            if (AuthUtil.instance.isConnectionExpired()) {
+                await AuthUtil.instance.showReauthenticatePrompt()
+            }
+        },
+        async getChatAuthState() {
+            return AuthUtil.instance.getChatAuthState()
+        },
+    },
+} satisfies api
--- a/packages/core/src/amazonq/extApi.ts
+++ b/packages/core/src/amazonq/extApi.ts
@@ -1,0 +1,43 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import vscode from 'vscode'
+import { VSCODE_EXTENSION_ID } from '../shared/utilities'
+import { GenerateAssistantResponseCommandOutput, GenerateAssistantResponseRequest } from '@amzn/codewhisperer-streaming'
+import { FeatureAuthState } from '../codewhisperer/util/authUtil'
+import { ToolkitError } from '../shared'
+
+/**
+ * This interface is used and exported by the amazon q extension. If you make a change here then
+ * update the corresponding api implementation in packages/amazonq/src/api.ts
+ */
+export interface api {
+    chatApi: {
+        chat(request: GenerateAssistantResponseRequest): Promise<GenerateAssistantResponseCommandOutput>
+    }
+    authApi: {
+        reauthIfNeeded(): Promise<void>
+        getChatAuthState(): Promise<FeatureAuthState>
+    }
+}
+
+export class AmazonqNotFoundError extends ToolkitError {
+    constructor() {
+        super(`${VSCODE_EXTENSION_ID.amazonq} is not installed`, { code: 'AmazonQNotInstalled' })
+    }
+}
+
+/**
+ * Get the extension API for Amazon q
+ *
+ * @returns The extension API for Amazon q, or undefined if the extension is not installed
+ */
+export async function getAmazonqApi(): Promise<api | undefined> {
+    const ext = vscode.extensions.getExtension(VSCODE_EXTENSION_ID.amazonq)
+    if (!ext) {
+        return undefined
+    }
+    return ext.activate()
+}
--- a/packages/core/src/applicationcomposer/messageHandlers/generateResourceHandler.ts
+++ b/packages/core/src/applicationcomposer/messageHandlers/generateResourceHandler.ts
@@ -11,10 +11,9 @@ import {
     Command,
     MessageType,
 } from '../types'
-import { ChatSession } from '../../codewhispererChat/clients/chat/v0/chat'
-import { AuthUtil } from '../../codewhisperer/util/authUtil'
 import globals from '../../shared/extensionGlobals'
 import { getLogger } from '../../shared/logger/logger'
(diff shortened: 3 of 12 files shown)


packages/amazonq/src/api.ts
```diff
@@ -0,0 +1,28 @@
...
+        async reauthIfNeeded() {
+            if (AuthUtil.instance.isConnectionExpired()) {
+                await AuthUtil.instance.showReauthenticatePrompt()
+            }
+        },
+        async getChatAuthState() {
+            return AuthUtil.instance.getChatAuthState()
+        },
+    },
+} as qAPI
```
justinmk3, line 28:
```suggestion
} as api
```

Also, `amazonq` is the name for "Amazon Q" in most of the codebase, let's stay consistent with that.

packages/core/src/amazonq/extApi.ts
```diff
@@ -0,0 +1,31 @@
...
+    authApi: {
+        reauthIfNeeded(): Promise<void>
+        getChatAuthState(): Promise<FeatureAuthState>
+    }
+}
+
+/**
+ * Get the extension API for Q
+ */
+export async function getQAPI(): Promise<qAPI> {
```
justinmk3, line 25: "amazonq" is the name for Amazon Q in the codebase. Acronyms are not case differently that other words ("HTML" => "Html", etc.).

```suggestion
export async function getAmazonqApi(): Promise<amazonqApi> {
```

packages/core/src/amazonq/extApi.ts
```diff
@@ -0,0 +1,31 @@
...
+    const ext = vscode.extensions.getExtension(VSCODE_EXTENSION_ID.amazonq)
+    if (!ext) {
+        throw new ToolkitError('Amazon Q is not installed', { code: 'AmazonQNotInstalled' })
```
JLargent42, line 28: Can we avoid this? Application Composer's integration with Q should be optional. Composer should still be able to open when Q is not installed (though the Q integration will be unavailable).

JLargent42 (changes requested):

JLargent42 (approved):

packages/core/src/amazonq/extApi.ts
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import vscode from 'vscode'
+import { VSCODE_EXTENSION_ID } from '../shared/utilities'
+import { GenerateAssistantResponseCommandOutput, GenerateAssistantResponseRequest } from '@amzn/codewhisperer-streaming'
+import { FeatureAuthState } from '../codewhisperer/util/authUtil'
+import { ToolkitError } from '../shared'
+
+export interface api {
```
nkomonen-amazon, line 12: Nice work to add types. Do we need to keep this interface 1:1 with `packages/amazonq/src/api.ts`? If so we should make a comment here and there for future devs.

packages/core/src/applicationcomposer/messageHandlers/generateResourceHandler.ts
```diff
@@ -72,13 +74,11 @@ async function generateResource(prompt: string) {
         let supplementaryWebLinks: SupplementaryWebLink[] = []
         let references: Reference[] = []
 
-        if (AuthUtil.instance.isConnectionExpired()) {
-            await AuthUtil.instance.showReauthenticatePrompt()
-        }
+        await amazonqApi.authApi.reauthIfNeeded()
```
nkomonen-amazon, line 77: Under the hood this shows the prompt to reauth. If the prompt is shown, should this function (`generateResource()`) return early instead of continuing since it doesn't actually wait until the user is properly authenticated.

packages/core/src/amazonq/extApi.ts
jpinkney-aws: `packages/amazonq/src/api.ts` consumes the interface so if we update the interface and don't update `packages/amazonq/src/api.ts` the build will fail. I can still add a comment though

packages/core/src/amazonq/extApi.ts
```diff
@@ -0,0 +1,43 @@
...
+        super(`${VSCODE_EXTENSION_ID.amazonq} is not installed`, { code: 'AmazonQNotInstalled' })
+    }
+}
+
+/**
+ * Get the extension API for Amazon q
+ *
+ * @returns The extension API for Amazon q, or undefined if the extension is not installed
+ */
+export async function getAmazonqAPI(): Promise<api | undefined> {
```
justinmk3, line 37:
```suggestion
export async function getAmazonqApi(): Promise<api | undefined> {
```

packages/amazonq/src/api.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { GenerateAssistantResponseCommandOutput, GenerateAssistantResponseRequest } from '@amzn/codewhisperer-streaming'
+import { AuthUtil } from 'aws-core-vscode/codewhisperer'
+import { ChatSession } from 'aws-core-vscode/codewhispererChat'
+import { api } from 'aws-core-vscode/amazonq'
+
+export default {
```
justinmk3, line 11: If it wasn't a circular dependency, this would `extend` the interface defined in `packages/core/src/amazonq/extApi.ts` , right?

This is kind of strange. Somewhat unavoidable since most of amazonq lives in "core" currently.

packages/amazonq/src/api.ts
```diff
@@ -0,0 +1,28 @@
...
+        async reauthIfNeeded() {
+            if (AuthUtil.instance.isConnectionExpired()) {
+                await AuthUtil.instance.showReauthenticatePrompt()
+            }
+        },
+        async getChatAuthState() {
+            return AuthUtil.instance.getChatAuthState()
+        },
+    },
+} as api
```
justinmk3, line 28: does `as` force (cast) the object to the interface even if it doesn't conform? is there a different syntax we can use which will add stricter type checking?

packages/core/src/applicationcomposer/messageHandlers/generateResourceHandler.ts
jpinkney-aws: It waits until the user is properly auth'd. If you open app composer, expire the connection, then hit generate suggestions it will prompt you to authenticate then it just waits until the user is authenticated before continuing

packages/amazonq/src/api.ts
nkomonen-amazon: Just tested and it does error if the incorrect properties are given. You would have to do `as unknown as api` to truly override it.

jpinkney-aws: Just switched it to satisfies instead of as which is an even stronger check

nkomonen-amazon (approved):

packages/core/src/applicationcomposer/messageHandlers/initMessageHandler.ts
```diff
@@ -5,19 +5,24 @@
...
 export async function initMessageHandler(context: WebviewContext) {
     const filePath = context.defaultTemplatePath
-    const authState = await AuthUtil.instance.getChatAuthState()
+    const amazonqApi = await getAmazonqApi()
+    let isConnectedToCodeWhisperer = false
+    if (amazonqApi) {
+        const authState = await amazonqApi.authApi.getChatAuthState()
+        isConnectedToCodeWhisperer =
+            authState.codewhispererChat === 'connected' || authState.codewhispererChat === 'expired'
+    }
```
justinmk3, line 18: since this pattern is appearing in a few places i will suggest:

```suggestion
    const amazonqApi = await getAmazonqApi()
    const authState = await amazonqApi?.authApi.getChatAuthState()
    const isConnectedToAmazonQ = authState?.codewhispererChat === 'connected' || authState?.codewhispererChat === 'expired'
```

note: `await undefined` is legal

PR #5258, opened 2024-07-02
web: minor Timer/Timeout change for CW

nkomonen-amazon:
## Problem

When in Web mode, we were using the Node.js Timeout `refresh()` method. This does not exist
in web mode.

## Solution

- Update the use with our environment agnostic `Timeout` class, which works in web or node.
- Copy the timeout unit tests from our regular `test/` in to the web mode tests. Fixed a minor bug when refreshing in web mode, and now they all pass without any modifications to the tests.

## License

### DIFF
--- a/packages/core/src/testWeb/shared/utilities/timeoutUtils.test.ts
+++ b/packages/core/src/testWeb/shared/utilities/timeoutUtils.test.ts
@@ -1,0 +1,13 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * IMPORTANT: We are importing the same test from the node timeoutUtils
+ * since the behavior should be the exact same.
+ *
+ * Any web specific tests should be made within their own `describe()`.
+ */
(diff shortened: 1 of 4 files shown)


packages/core/src/testWeb/shared/utilities/timeoutUtils.test.ts
```diff
@@ -0,0 +1,394 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * IMPORTANT: This test is a copy of of the Node timeoutUtils.test.ts file, with
+ * all tests in the `describe('timeoutUtils')` block.
+ * We should look to keep the tests 1:1 so that updates to the Node test file
+ * can be easily copied here.
```
justinmk3, line 10: can this file import the non-web test instead?

nkomonen-amazon: Good point, was able to export the `describe` and it all works

jpinkney-aws (approved):

PR #5259, opened 2024-07-02
feat(telemetry): emit on opt-out and opt-in 

hayemaxi:
- Emits 1 final metric on opt-out that will let us know the user opted out.
- Also, on opt out it will emit any telemetry that was previously recorded and not sent.

## License

### DIFF
--- a/packages/core/src/shared/telemetry/activation.ts
+++ b/packages/core/src/shared/telemetry/activation.ts
@@ -18,6 +18,7 @@ import { TelemetryConfig, setupTelemetryId } from './util'
 import { isAutomation, isReleaseVersion } from '../vscode/env'
 import { AWSProduct } from './clienttelemetry'
 import { DefaultTelemetryClient } from './telemetryClient'
+import { telemetry } from './telemetry'
 import { Commands } from '../vscode/commands2'
 
 export const noticeResponseViewSettings = localize('AWS.telemetry.notificationViewSettings', 'Settings')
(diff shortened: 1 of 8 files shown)


packages/core/src/shared/telemetry/activation.ts
```diff
@@ -49,12 +50,23 @@ export async function activate(
...
+                    if (!val) {
+                        // Record 'disabled' right before its turned off, so we can send this + the batch we have already.
+                        telemetry.ui_click.emit({ elementId: 'aws_disableTelemetry' })
+                    }
+
+                    await globals.telemetry.setTelemetryEnabled(val)
+
+                    if (val) {
+                        // Record 'enabled' after its turned on, otherwise this is ignored.
+                        telemetry.ui_click.emit({ elementId: 'aws_enableTelemetry' })
```
justinmk3, line 68: there is an `enabled` field that could be set: https://github.com/aws/aws-toolkit-common/blob/9fdefa93c1741bf2cc3046a01f70fd7cc3eed540/telemetry/definitions/commonDefinitions.json#L1064-L1068

then we can have a single metric (or in this case, "ui_click element") that has multiple states. Instead of representing every state with a new "metric"

packages/core/src/shared/telemetry/activation.ts
```diff
@@ -49,12 +50,23 @@ export async function activate(
...
 
         extensionContext.subscriptions.push(
-            (isAmazonQ() ? config.amazonQConfig : config.toolkitConfig).onDidChange(event => {
+            (isAmazonQ() ? config.amazonQConfig : config.toolkitConfig).onDidChange(async event => {
                 if (event.key === 'telemetry') {
-                    globals.telemetry.telemetryEnabled = config.isEnabled()
+                    const val = config.isEnabled()
+                    if (!val) {
+                        // Record 'disabled' right before its turned off, so we can send this + the batch we have already.
+                        telemetry.ui_click.emit({ elementId: 'aws_disableTelemetry' })
```
justinmk3, line 61: `aws_modifySetting` is an alternative to `ui_click` : https://github.com/aws/aws-toolkit-common/blob/9fdefa93c1741bf2cc3046a01f70fd7cc3eed540/telemetry/definitions/commonDefinitions.json#L2343

dhruvigajjar: Ohh I missed it. Thanks Justin! let's change it to 
`aws_modifySetting.emit({ settingId: 'telemetryEnabled', settingState: true/false })`

when settingState is `false` it means telemetry is disabled 
when settingState is `true` it means telemetry is enabled 
image

packages/core/src/shared/telemetry/activation.ts
```diff
@@ -49,12 +50,26 @@ export async function activate(
...
+        await globals.telemetry.setTelemetryEnabled(config.isEnabled())
 
         extensionContext.subscriptions.push(
-            (isAmazonQ() ? config.amazonQConfig : config.toolkitConfig).onDidChange(event => {
+            (isAmazonQ() ? config.amazonQConfig : config.toolkitConfig).onDidChange(async event => {
                 if (event.key === 'telemetry') {
-                    globals.telemetry.telemetryEnabled = config.isEnabled()
+                    const val = config.isEnabled()
+                    await telemetry.aws_modifySetting.run(async () => {
+                        telemetry.record({ settingId: 'telemetryEnabled' })
```
justinmk3, line 60: Typically the settingId would be the actual setting id, either `amazonQ.telemetry` or `aws.telemetry`

justinmk3 (approved):

justinmk3 (approved):

PR #5260, opened 2024-07-02
fix(codecatalyst): simplify InactivityMessage

justinmk3:
## Problem

- The "progress" message tests are flaky because of the complex logic, and sometimes fail:
  ```
    1) InactivityMessages
         shows expected messages 5 minutes before shutdown on a 60 minute inactivity timeout:
       AssertionError [ERR_ASSERTION]: Expected 3 messages, but got 2
        at assertMessagesShown (src/testInteg/codecatalyst/devEnv.test.ts:182:20)
        at async Context.<anonymous> (src/testInteg/codecatalyst/devEnv.test.ts:85:9)
  ```
- Toolkit has complex logic for showing a "progress" message when the codecatalyst dev env timeout is approaching, followed by different logic for showing the "final" message. This isn't worth the complexity because if the user sees the message, clicking it is zero-cost and will happen immediately, otherwise the user isn't active and won't see any of the messages anyway.
- The non-modal progress message may be *hidden* if the user enabled vscode's "do not disturb" feature.

## Solution

- Remove the non-modal "progress" message. Only show the final, modal message **starting 5 min before shutdown**. Display it until user clicks it.

Followup to https://taskei.amazon.dev/tasks/IDE-13892

## License

### DIFF
--- a/packages/core/src/codecatalyst/devEnv.ts
+++ b/packages/core/src/codecatalyst/devEnv.ts
@@ -7,9 +7,7 @@ import { DevEnvironment } from '../shared/clients/codecatalystClient'
 import { DevEnvActivity, DevEnvClient } from '../shared/clients/devenvClient'
 import globals from '../shared/extensionGlobals'
 import * as vscode from 'vscode'
-import { Timeout, waitUntil } from '../shared/utilities/timeoutUtils'
-import { showMessageWithCancel } from '../shared/utilities/messages'
-import { isCloud9 } from '../shared/extensionUtilities'
+import { waitUntil } from '../shared/utilities/timeoutUtils'
 import { getLogger } from '../shared/logger'
 import { CodeCatalystAuthenticationProvider } from './auth'
 import { getThisDevEnv } from './model'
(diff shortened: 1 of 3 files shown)


packages/core/src/codecatalyst/devEnv.ts
```diff
@@ -246,84 +241,39 @@ export class InactivityMessage implements vscode.Disposable {
...
+                `Your CodeCatalyst Dev Environment has been inactive for ${minutesUserWasInactive} minutes, and will stop soon.`,
+                { modal: true },
+                imHere
             )
-        }
+            .then(res => {
+                if (res === imHere) {
+                    userIsActive()
+                }
+            })
```
hayemaxi, line 272: what if the user X'es out the notification? That means they are active. Other than closing VSC, not sure if there are other ways that that a notification can enter this block without being explicitly interacted with.
```suggestion
            .then(userIsActive)
```

justinmk3: > what if the user X'es out the notification? That means they are active.

Yep. Presumably if they hit Cancel or X, they are intentionally choosing not to click `I'm Here` because they don't want to trigger activity.

PR #5264, opened 2024-07-04
feature: get size of workspace utility

32teeth:
## Problem

Sometimes it takes long for the `collectFiles` to exit with an error if more than 200MB, depending on number of files

## Solution

* implemented `getWorkspaceSize` return accumulated size in bytes
* used a preflight in `collectFiles` to 'eject' faster that trying to collect and zip and use storage

> [!TIP]
> It now collects and returns the files to reduce redundancy in collectFiles
```
export async function getWorkspaceSize(sourcePaths: string[], respectGitIgnore: boolean = true): Promise<vscode.Uri[]> {
    let totalSizeBytes = 0
    let collection: vscode.Uri[] = []

    for (const rootPath of sourcePaths) {
        const allFiles = await vscode.workspace.findFiles(
            new vscode.RelativePattern(rootPath, '**'),
            getExcludePattern()
        )

        const files = respectGitIgnore ? await filterOutGitignoredFiles(rootPath, allFiles) : allFiles
        const fileStats = await Promise.all(files.map(file => fsCommon.stat(file)))

        totalSizeBytes += fileStats.reduce((acc, fileStat) => acc + fileStat.size, 0)
        collection = collection.concat(files)
    }

    if (totalSizeBytes > maxRepoSizeBytes) {
        throw new ToolkitError(
            'The project you have selected for source code is too large to use as context. Please select a different folder to use',
            { code: 'ContentLengthError' }
        )
    }

    return collection
}
```

```
// collect files excerpt
    const collection = await getWorkspaceSize(sourcePaths, respectGitIgnore)
    for (const file of collection) {
        const relativePath = getWorkspaceRelativePath(file.fsPath, { workspaceFolders })
        if (!relativePath) {
            continue
        }

        const fileContent = await readFile(file)
        if (fileContent === undefined) {
            continue
        }

        storage.push({
            workspaceFolder: relativePath.workspaceFolder,
            relativeFilePath: relativePath.relativePath,
            fileUri: file,
            fileContent: fileContent,
            zipFilePath: prefixWithFolderPrefix(relativePath.workspaceFolder, relativePath.relativePath),
        })
    }
```

## License

### DIFF
--- a/packages/core/src/shared/utilities/workspaceUtils.ts
+++ b/packages/core/src/shared/utilities/workspaceUtils.ts
@@ -17,6 +17,7 @@ import { sanitizeFilename } from './textUtilities'
 import { GitIgnoreAcceptor } from '@gerhobbelt/gitignore-parser'
 import * as parser from '@gerhobbelt/gitignore-parser'
 import { fsCommon } from '../../srcShared/fs'
+import { maxRepoSizeBytes } from '../../amazonqFeatureDev/constants'
 
 type GitIgnoreRelativeAcceptor = {
     folderPath: string
@@ -295,6 +296,35 @@ async function filterOutGitignoredFiles(rootPath: string, files: vscode.Uri[]): 
 }
 
 /**
+ * Get the size of the currently sleected workspace and return it in bytes
+ * @param sourcePaths the paths where collection starts
+ * @param workspaceFolders the current workspace folders opened
+ * @param respectGitIgnore whether to respect gitignore file
+ * @returns the size of the workspace in bytes
+ *
+ */
+export async function getWorkspaceSize(
+    sourcePaths: string[],
+    workspaceFolders: CurrentWsFolders,
+    respectGitIgnore: boolean = true
+): Promise<number> {
+    let totalSizeBytes = 0
+    for (const rootPath of sourcePaths) {
+        const allFiles = await vscode.workspace.findFiles(
+            new vscode.RelativePattern(rootPath, '**'),
+            getExcludePattern()
+        )
+        const files = respectGitIgnore ? await filterOutGitignoredFiles(rootPath, allFiles) : allFiles
+
+        for (const file of files) {
+            const fileStat = await fsCommon.stat(file)
+            totalSizeBytes += fileStat.size
+        }
+    }
+    return totalSizeBytes
+}
+
+/**
  * collects all files that are marked as source
  * @param sourcePaths the paths where collection starts
  * @param workspaceFolders the current workspace folders opened
@@ -305,7 +335,7 @@ export async function collectFiles(
     sourcePaths: string[],
     workspaceFolders: CurrentWsFolders,
     respectGitIgnore: boolean = true,
-    maxSize = 200 * 1024 * 1024 // 200 MB
+    maxSize = maxRepoSizeBytes
 ): Promise<
     {
         workspaceFolder: vscode.WorkspaceFolder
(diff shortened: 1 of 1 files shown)


32teeth:
Woot 14/14 checks passed #firsttry

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -315,6 +345,14 @@ export async function collectFiles(
         zipFilePath: string
     }[]
 > {
+    const workspaceSize = await getWorkspaceSize(sourcePaths, workspaceFolders, respectGitIgnore)
+    if (workspaceSize > maxSize) {
+        throw new ToolkitError(
+            'The project you have selected for source code is too large to use as context. Please select a different folder to use',
```
jpinkney-aws, line 351: This util is generic but this comment seems very specific to q

32teeth: If this can work elsewhere, we can explore abstracting it in the future

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -294,6 +295,35 @@ async function filterOutGitignoredFiles(rootPath: string, files: vscode.Uri[]):
     return gitIgnoreFilter.filterFiles(files)
 }
 
+/**
+ * Get the size of the currently sleected workspace and return it in bytes
```
justinmk3, line 299: typo. please also use punctuation, it's a cheap way to reduce ambiguity.
```suggestion
 * Gets the size of the currently selected workspace and return it in bytes.
 *
```

packages/core/src/shared/utilities/workspaceUtils.ts
justinmk3: It's already abstracted, that's why it lives in shared/util/.

The caller can pass an `maxSizeMsg` string.

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -315,6 +345,14 @@ export async function collectFiles(
         zipFilePath: string
     }[]
 > {
+    const workspaceSize = await getWorkspaceSize(sourcePaths, workspaceFolders, respectGitIgnore)
```
justinmk3, line 348: Did you look for existing code that calculates workspace size? I seem to recall a few of those,  e.g. for "security scan". Please check for those and see if they can be updated to use this shared logic.

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -294,6 +295,35 @@ async function filterOutGitignoredFiles(rootPath: string, files: vscode.Uri[]):
...
+    let totalSizeBytes = 0
+    for (const rootPath of sourcePaths) {
+        const allFiles = await vscode.workspace.findFiles(
+            new vscode.RelativePattern(rootPath, '**'),
+            getExcludePattern()
+        )
+        const files = respectGitIgnore ? await filterOutGitignoredFiles(rootPath, allFiles) : allFiles
+
+        for (const file of files) {
+            const fileStat = await fsCommon.stat(file)
```
justinmk3, line 320: collectFiles already calls `fsCommon.stat` on each file, but if it calls this then that doubles the cost. Is it possible to couple this with the logic in collectFiles ?

32teeth:
Thanks for all the intake comments.

I did observe there was an unused utility for something similar, will investigate on how to implement

As per the double cost. Absolutely, worth a look on how to reduce cost on the `fs.stat`

Ideally I can retrofit it to
1. Accept a max size
2. If max size is exceeded, return error
3. If within threshold, return the files to reduce cost of next call

Great feedback, thank you

justinmk3 (changes requested):
There is already a `getDirSize` function: https://github.com/aws/aws-toolkit-vscode/blob/bbec3563d1c78ed3b3462120a1da4ec5b5fefe94/packages/core/src/shared/filesystemUtilities.ts#L22

Why is this new function needed? `getDirSize` also has a time limit, which is a useful and necessary feature.

justinmk3:
Set to "draft" since this still has lots of work needed. Un-draft it when it's ready for review.

32teeth:
Closing for now

PR #5271, opened 2024-07-08
feat(amazonq): Chat Project Context With LSP 

leigaol:
## Problem

Amazon Q chat is not aware of files opened in user's workspace. User should be able to ask Q any question regarding current open workspace. 

## Solution

By performing local indexing with CodeSage model in LSP, we will retrieve the most relevant files for user chat input, and utilize that input to enhance Q chat experience when user asks  Q any question regarding current open workspace. 

### Scopes:

1. Download and installation of LSP server, follow DEXP download manifest. We enforce the LSP version in the IDE so that when we release new LSP, there is no production impact. A new extension is required to run new LSP.
2. Start the LSP server with the stdin encryption spec from DEXP.
3. Call the LSP server to index project,
4. Call the LSP server to retrieve project context
5. New settings for the LSP server. Enable/Disable, Threads, Use GPU or not.
6. New metrics.

## License


packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,361 @@
...
+    artifactId: string
+    artifactDescription: string
+    isManifestDeprecated: boolean
+    versions: {
+        serverVersion: string
+        isDelisted: boolean
+        targets: Target[]
+    }[]
+}
+const manifestUrl = 'https://aws-toolkit-language-servers.amazonaws.com/temp/manifest.json'
```
justinmk3, line 66: So the "temp" URL is not temporary? Can this tech debt be avoided, why not rename the URL?

packages/amazonq/.changes/next-release/Feature-69f0d3bf-3e57-4ad6-93ea-48de3c890331.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Add LSP for Amazon Q Chat Local Indexing"
```
justinmk3, line 3: The changelog is a "mini blog" for advertising new features. It also allows markdown, so you can link to external URLs.

 Is there a way to describe this so that customers can discover the actual usefulness of the feature?

packages/amazonq/package.json
```diff
@@ -125,6 +125,30 @@
                     "markdownDescription": "%AWS.configuration.description.amazonq.shareContentWithAWS%",
                     "default": true,
                     "scope": "application"
+                },
+                "amazonQ.localWorkspaceIndex": {
```
justinmk3, line 129: why is `local` prefixing these settings? when would it be "non local" ? 

Local is the normal case. If there is a "remote" case, then it would make sense to name settings like `remoteFoo`. But in the normal case, it's redundant and confusing to mention "local".

packages/amazonq/test/unit/amazonq/lsp/lspClient.test.ts
```diff
@@ -0,0 +1,37 @@
+import * as sinon from 'sinon'
+import assert from 'assert'
+import { LspClient } from 'aws-core-vscode/amazonq'
+
+describe('Test LSP client', function () {
```
justinmk3, line 5: It is never necessarily to mention "Test" in a test name. Nor "should", "correct", etc. 
```suggestion
describe('Amazon Q LSP', function () {
```

packages/amazonq/test/unit/amazonq/lsp/lspClient.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    beforeEach(function () {
+        lspClient = new LspClient()
+        encryptFunc = sinon.spy(lspClient, 'encrypt')
+    })
+
+    it('should encrypt payload of query ', async () => {
+        await lspClient.query('mock_input')
+        assert.ok(encryptFunc.calledOnce)
+        assert.ok(encryptFunc.calledWith(JSON.stringify({ query: 'mock_input' })))
```
justinmk3, line 17: can the actual encrypted result be tested instead of using a spy?

packages/amazonq/test/unit/amazonq/lsp/lspClient.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    })
+
+    it('should encrypt payload of index files ', async () => {
```
justinmk3, line 20:
```suggestion
    it('encrypts payload of index files ', async () => {
```

packages/amazonq/test/unit/amazonq/lsp/lspClient.test.ts
```diff
@@ -0,0 +1,37 @@
...
+describe('Test LSP client', function () {
+    let lspClient: LspClient
+    let encryptFunc: sinon.SinonSpy
+
+    beforeEach(function () {
+        lspClient = new LspClient()
+        encryptFunc = sinon.spy(lspClient, 'encrypt')
+    })
+
+    it('should encrypt payload of query ', async () => {
```
justinmk3, line 14:
```suggestion
    it('encrypts payload of query ', async () => {
```

packages/amazonq/test/unit/amazonq/lsp/lspController.test.ts
```diff
@@ -0,0 +1,51 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import assert from 'assert'
+import sinon from 'sinon'
+import { Content, LspController } from 'aws-core-vscode/amazonq'
+import { createTestFile } from 'aws-core-vscode/test'
+import { fsCommon } from 'aws-core-vscode/srcShared'
+
+describe('Test LSP controller', function () {
+    it('Verify downloaded mechanism checks against hash, when hash matches', async function () {
```
justinmk3, line 12:
```suggestion
    it('downloaded mechanism checks against hash, when hash matches', async function () {
```

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,361 @@
...
+// this LSP client in Q extension is only going to work with these LSP server versions
+const supportedLspServerVersions = ['0.1.0']
+
+const nodeBinName = process.platform === 'win32' ? 'node.exe' : 'node'
+
+export class LspController {
```
justinmk3, line 72: what is the purpose of this class? Needs a docstring.  This "controller" object stores no state, so why is it a singleton?

I see "controller" added to many class names without any purpose. It often is just a collection of util functions. If that's the case, why not just have a `lspUtil.ts` module?

packages/core/src/codewhispererChat/view/messages/messageListener.ts
```diff
@@ -95,9 +95,15 @@ export class UIMessageListener {
             case 'footer-info-link-click':
                 this.processFooterInfoLinkClick(msg)
                 break
+            case 'open-settings':
+                this.processOpenSettings(msg)
         }
     }
 
+    private processOpenSettings(msg: any) {
+        void vscode.commands.executeCommand('workbench.action.openSettings', `@id:amazonQ.localWorkspaceIndex`)
```
justinmk3, line 104: https://github.com/aws/aws-toolkit-vscode/blob/b15135d64f9eec6d9475f133bb48af376801f775/packages/core/src/shared/settings.ts#L919

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -154,6 +154,36 @@
             "type": "boolean",
             "description": "true if user has selected code snippet, false otherwise."
         },
+        {
+            "name": "cwsprChatHasProjectLevelContext",
+            "type": "boolean",
+            "description": "true if query has project level context."
+        },
+        {
+            "name": "cwsprChatProjectLevelContextQueryLatencyInMs",
```
justinmk3, line 163: can say the same thing with less verbosity:

```suggestion
            "name": "cwsprChatProjectContextQueryMs",
```

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -154,6 +154,36 @@
             "type": "boolean",
             "description": "true if user has selected code snippet, false otherwise."
         },
+        {
+            "name": "cwsprChatHasProjectLevelContext",
```
justinmk3, line 158:
```suggestion
            "name": "cwsprChatHasProjectContext",
```

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -154,6 +154,36 @@
...
+        },
+        {
+            "name": "amazonqIndexMemoryUsageInMB",
+            "type": "int",
+            "description": "Memory usage of LSP server"
+        },
+        {
+            "name": "amazonqIndexCpuUsagePercentage",
+            "type": "int",
+            "description": "Cpu usage of LSP server"
```
justinmk3, line 185:
```suggestion
            "description": "CPU used by LSP server as a percentage of all available CPUs on the system"
```

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -511,3 +513,87 @@ export function getWorkspaceFoldersByPrefixes(
...
 }
+
+/**
+ * collects all files that are marked as source
+ * @param sourcePaths the paths where collection starts
+ * @param workspaceFolders the current workspace folders opened
+ * @param respectGitIgnore whether to respect gitignore file
+ * @returns all matched files
+ */
+export async function collectFilesForIndex(
```
justinmk3, line 524: Why can't this re-use collectFiles? https://github.com/aws/aws-toolkit-vscode/blob/b15135d64f9eec6d9475f133bb48af376801f775/packages/core/src/shared/utilities/workspaceUtils.ts#L304

Could a small adjustment be made to `collectFiles` so that it can be re-used? E.g., add an optional callback parameter that allows the caller to do extra processing on each item?

Code re-use is very, very, very important. It is very important that we avoid copy-pasting code. That has created a lot of tech debt.

packages/core/src/amazonq/webview/ui/tabs/generator.ts
```diff
@@ -33,6 +33,17 @@ export class TabDataGenerator {
                 'Use of Amazon Q is subject to the [AWS Responsible AI Policy](https://aws.amazon.com/machine-learning/responsible-ai/policy/).',
             quickActionCommands: this.quickActionsGenerator.generateForTab(tabType),
             promptInputPlaceholder: TabTypeDataMap[tabType].placeholder,
+            contextCommands: [
+                {
+                    groupName: 'Metion code',
```
justinmk3, line 38: typo

packages/core/src/amazonq/lsp/lspClient.ts
```diff
@@ -0,0 +1,208 @@
...
+import { GetUsageRequestType, IndexRequestType, QueryRequestType, UpdateIndexRequestType, Usage } from './types'
+import { Writable } from 'stream'
+import { CodeWhispererSettings } from '../../codewhisperer/util/codewhispererSettings'
+import { getLogger } from '../../shared'
+
+const localize = nls.loadMessageBundle()
+
+const key = crypto.randomBytes(32)
+
+export function writeEncryptionInit(stream: Writable): void {
```
justinmk3, line 29: docstring should explain the requirements of this function. does it need to match a specification somewhere? if so, provide a URL reference.

packages/core/src/amazonq/activation.ts
```diff
@@ -52,14 +55,25 @@ export async function activate(context: ExtensionContext) {
...
+            }
+        })
     )
 
     Commands.register('aws.amazonq.learnMore', () => {
         void vscode.env.openExternal(vscode.Uri.parse(amazonQHelpUrl))
     })
 
     await activateBadge()
+    void LspController.instance.trySetupLsp(context)
```
justinmk3, line 76: This one isn't debounced. Perhaps the debounced function should be declared so that this can also call the debounced function?

justinmk3:
were the comments in https://github.com/aws/aws-toolkit-vscode-staging/pull/1621 addressed?

packages/core/src/amazonq/lsp/lspController.ts
leigaol: DEXP team told me to use this url.

packages/amazonq/package.json
leigaol: We are planning to do a remote indexing in the near future, this is why we name it local here.

packages/core/src/amazonq/lsp/lspController.ts
leigaol: 1. In the latest commit, it does have one state, is Indexing in progress. 

2. Making it a object makes it easier to use sinon.stub for unit testing.

3. It is named as controller because it manages some state ( not having lsp --> install lsp --> lsp running -> indexing in progress)

I will add doc string

packages/amazonq/.changes/next-release/Feature-69f0d3bf-3e57-4ad6-93ea-48de3c890331.json
leigaol: Added more description in the change log.

packages/core/src/amazonq/lsp/lspClient.ts
leigaol: Added URL reference

packages/core/src/shared/utilities/workspaceUtils.ts
leigaol: No for this PR because 
1. collectFiles has hardcoded security scan features like compresses these filles into a zip, it also returns an object with zipFilePath, it also performs file read.  while collect file for index do not do any zip file, do not actually performs the file read operation.
2. collect file for index has hardcoded indexing features like filtering files, ranking them by index priority, etc, which is not there in the collectFiles. And it is not trivial to merge these two.

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -154,6 +154,36 @@
...
+        },
+        {
+            "name": "cwsprChatProjectLevelContextQueryLatencyInMs",
+            "type": "int",
+            "description": "query latency in ms for local project context"
+        },
+        {
+            "name": "amazonqIndexFileSizeInMB",
+            "type": "int",
+            "description": "Index size in MB"
```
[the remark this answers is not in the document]
leigaol: This is actually the total number of MB of these indexed files, not the size of the index. I made it clear in the latest commit.

packages/amazonq/test/unit/amazonq/lsp/lspClient.test.ts
leigaol: This is done in latest commit

leigaol:
> were the comments in [aws/aws-toolkit-vscode-staging#1621](https://github.com/aws/aws-toolkit-vscode-staging/pull/1621) addressed?

That temp url is still pending. Asked DEXP for this

packages/core/src/amazonq/index.ts
```diff
@@ -11,7 +11,8 @@ export { MessageListener } from './messages/messageListener'
 export { AuthController } from './auth/controller'
 export { showAmazonQWalkthroughOnce } from './onboardingPage/walkthrough'
 export { openAmazonQWalkthrough } from './onboardingPage/walkthrough'
-
+export { LspController, Content } from './lsp/lspController'
+export { LspClient } from './lsp/lspClient'
```
Will-ShaoHua, line 15: uh are they being used?

leigaol: Yes being used in unit tests

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+import AdmZip from 'adm-zip'
+import { RelevantTextDocument } from '@amzn/codewhisperer-streaming'
+import { makeTemporaryToolkitFolder } from '../../shared/filesystemUtilities'
+import { CodeWhispererSettings } from '../../codewhisperer/util/codewhispererSettings'
+import { activate as activateLsp } from './lspClient'
+import { telemetry } from '../../shared/telemetry'
+import { isCloud9 } from '../../shared/extensionUtilities'
+import { globals } from '../../shared'
+import { AuthUtil } from '../../codewhisperer'
+import { isWeb } from '../../shared/extensionGlobals'
```
Will-ShaoHua, line 23: should have an empty line

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+            const files = await collectFilesForIndex(
+                projPaths,
+                vscode.workspace.workspaceFolders as CurrentWsFolders,
+                true,
+                CodeWhispererSettings.instance.getMaxIndexSize() * 1024 * 1024
+            )
+            let totalSizeBytes = 0
+            for (let i = 0; i < files.length; i += 1) {
+                totalSizeBytes += files[i].fileSizeBytes
+            }
```
Will-ShaoHua, line 298:
```suggestion
            const totalSizeBytes = files.reduce((accumulator, currentFile) => accumulator + currentFile.fileSizeBytes, 0);
```

packages/core/src/amazonq/lsp/lspClient.ts
```diff
@@ -0,0 +1,219 @@
...
+    }
+    stream.write(JSON.stringify(request))
+    stream.write('\n')
+}
+/**
+ * LspClient manages the API call between VS Code extension and LSP server
+ * It encryptes the payload of API call.
+ */
+export class LspClient {
+    static #instance: LspClient
```
Will-ShaoHua, line 47: don't think we need this LspClient to be a singleton?
maybe we should just manage it within LspController

packages/core/src/amazonq/lsp/lspClient.ts
```diff
@@ -0,0 +1,219 @@
...
+    client: LanguageClient | undefined
```
Will-ShaoHua, line 48: why don't we just make it a non-nullable `LanguageClient`

Will-ShaoHua: what's the case when LanguageClient is undefined

packages/core/src/amazonq/lsp/lspClient.ts
```diff
@@ -0,0 +1,219 @@
...
+        },
+    }
+
+    // Create the language client and start the client.
+    LspClient.instance.client = new LanguageClient(
+        'codewhisperer',
+        localize('codewhisperer.server.name', 'Amazon Q Language Server'),
+        serverOptions,
+        clientOptions
+    )
```
Will-ShaoHua, line 185: prefer injecting `LanguageClient` with LspClient's ctor

Will-ShaoHua: also not seeing the value to make LspClient singleton

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+        setImmediate(async () => {
+            if (!CodeWhispererSettings.instance.isLocalIndexEnabled()) {
+                // only download LSP for users who did not turn on this feature
+                // do not start LSP server
+                await LspController.instance.tryInstallLsp(context)
+                return
+            }
+            const ok = await LspController.instance.tryInstallLsp(context)
+            if (!ok) {
+                return
```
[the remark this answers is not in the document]
leigaol: the try install itself will print error logs

packages/core/src/amazonq/lsp/lspClient.ts
```diff
@@ -0,0 +1,219 @@
...
+        } catch (e) {
+            getLogger().error(`LspClient: updateIndex error: ${e}`)
+            return undefined
+        }
+    }
+}
+/**
+ * Activates the language server, this will start LSP server running over IPC protocol.
+ * It will create a output channel named Amazon Q Language Server.
+ * This function assumes the LSP server has already been downloaded.
```
Will-ShaoHua, line 120: `This function assumes the LSP server has already been downloaded.`, should we add a sanity check on this?

packages/core/src/amazonq/lsp/lspClient.ts
```diff
@@ -0,0 +1,219 @@
...
+        toDispose.push(disposableFunc)
+    })
+}
+
+export async function deactivate(): Promise<any> {
+    if (!LspClient.instance.client) {
+        return undefined
+    }
+    return LspClient.instance.client.stop()
+}
```
Will-ShaoHua, line 219: is this func being used? i didn't see other reference

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+import { activate as activateLsp } from './lspClient'
+import { telemetry } from '../../shared/telemetry'
+import { isCloud9 } from '../../shared/extensionUtilities'
+import { globals } from '../../shared'
+import { AuthUtil } from '../../codewhisperer'
+import { isWeb } from '../../shared/extensionGlobals'
+function getProjectPaths() {
+    const workspaceFolders = vscode.workspace.workspaceFolders
+    if (!workspaceFolders || workspaceFolders.length === 0) {
+        throw Error('No workspace folders found')
```
Will-ShaoHua, line 27: nit: use ToolkitError maybe? https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/errors.ts#L135

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+    }
+    return workspaceFolders.map(folder => folder.uri.fsPath)
+}
+
+export interface Chunk {
+    readonly filePath: string
+    readonly content: string
+    readonly context?: string
+    readonly relativePath?: string
+    readonly programmingLanguage?: string
```
Will-ShaoHua, line 37: nit: use `CodewhispererLanguage` instead of vanilla string

Will-ShaoHua: we should rename it to QLanguage tho

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    async _download(localFile: string, remoteUrl: string) {
+        const res = await fetch(remoteUrl, {
+            headers: {
+                'User-Agent': 'curl/7.68.0',
+            },
+        })
+        if (!res.ok) {
+            throw new Error(`Failed to download. Error: ${JSON.stringify(res)}`)
```
Will-ShaoHua, line 101: nit: ToolkitError

packages/amazonq/test/unit/amazonq/lsp/lspClient.test.ts
justinmk3: Is the spy still needed ?

packages/amazonq/test/unit/amazonq/lsp/lspController.test.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import assert from 'assert'
+import sinon from 'sinon'
+import { Content, LspController } from 'aws-core-vscode/amazonq'
+import { createTestFile } from 'aws-core-vscode/test'
+import { fs } from 'aws-core-vscode/shared'
+
+describe('Test LSP controller', function () {
```
justinmk3, line 11:
```suggestion
describe('Amazon Q LSP controller', function () {
```

packages/amazonq/test/unit/amazonq/lsp/lspController.test.ts
```diff
@@ -0,0 +1,51 @@
...
+        const lspController = new LspController()
+        sinon.stub(lspController, '_download')
+        const mockFileName = 'test_case_1.zip'
+        const mockDownloadFile = await createTestFile(mockFileName)
+        await fs.writeFile(mockDownloadFile.fsPath, 'test')
+        const result = await lspController.downloadAndCheckHash(mockDownloadFile.fsPath, content)
+        assert.strictEqual(result, true)
+    })
+
+    it('Verify downloaded mechanism checks against hash, when hash does not match', async function () {
```
justinmk3, line 30:
```suggestion
    it('downloaded mechanism checks against hash, when hash does not match', async function () {
```

packages/amazonq/test/unit/amazonq/lsp/lspController.test.ts
```diff
@@ -0,0 +1,51 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import assert from 'assert'
+import sinon from 'sinon'
+import { Content, LspController } from 'aws-core-vscode/amazonq'
+import { createTestFile } from 'aws-core-vscode/test'
+import { fs } from 'aws-core-vscode/shared'
+
+describe('Test LSP controller', function () {
+    it('Downloaded mechanism checks against hash, when hash matches', async function () {
```
justinmk3, line 12:
```suggestion
    it('downloaded mechanism checks against hash, when hash matches', async function () {
```

packages/amazonq/package.json
justinmk3: But again, "local" is the normal, common case. All settings are assumed to be "local". 

In the "remote" case, a "remote" qualifier makes sense. But that doesn't mean every setting needs a "local" qualifier.

packages/core/src/shared/utilities/workspaceUtils.ts
justinmk3: There is no reason that `collectFiles` needs to be coupled with zipping. Please share a backlog item to generalize these "collect" routines and de-duplicate this logic.

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -717,6 +747,44 @@
                 },
                 {
                     "type": "cwsprChatConversationType"
+                },
+                {
+                    "type": "cwsprChatHasProjectContext",
+                    "required": false
+                },
+                {
+                    "type": "cwsprChatProjectContextQueryMs",
```
justinmk3, line 756: can the `duration` field be used instead?

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+                telemetry.amazonq_indexWorkspace.emit({
+                    duration: performance.now() - start,
+                    result: 'Failed',
+                    amazonqIndexFileCount: 0,
+                    amazonqIndexFileSizeInMB: 0,
+                })
+            }
+        } catch (e) {
+            getLogger().error(`LspController: Failed to build vector index of project`)
+            telemetry.amazonq_indexWorkspace.emit({
```
justinmk3, line 328: can the logic be wrapped in `amazonq_indexWorkspace.run(...)`? that would avoid all of these try/catch cases and redundant `amazonq_indexWorkspace.emit()` calls. And fields like `duration` and `result`, `reason`, and `reasonDesc` would be automatically set.

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+import * as path from 'path'
+import * as fs from 'fs-extra'
```
justinmk3, line 8: do not use fs-extra. we have a `fs.ts` module that has most of its features.

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+            }
+            const manifest: Manifest = (await this.downloadManifest()) as Manifest
+            const qserver = this.getQserverFromManifest(manifest)
+            const noderuntime = this.getNodeRuntimeFromManifest(manifest)
+            if (!qserver || !noderuntime) {
+                getLogger().info(`LspController: Did not find LSP URL for ${process.platform} ${process.arch}`)
+                return false
+            }
+
+            const tempFolder = await makeTemporaryToolkitFolder()
```
justinmk3, line 231: should this tempFolder be deleted in a finally{} block

justinmk3, line 231: should this tempFolder be deleted in in a finally{] block?

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+        return (this.#instance ??= new this())
+    }
+    constructor() {}
+
+    isIndexingInProgress() {
+        return this._isIndexingInProgress
+    }
+
+    async _download(localFile: string, remoteUrl: string) {
+        const res = await fetch(remoteUrl, {
```
justinmk3, line 95: can the shared `packages/core/src/common/request.ts` be used instead?

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+            headers: {
+                'User-Agent': 'curl/7.68.0',
```
justinmk3, line 97: Q user agent is here: https://github.com/aws/aws-toolkit-vscode/blob/10dacd932e04f5b0d377a1d516f2eef7f6ffa2a1/packages/core/src/shared/telemetry/util.ts#L123

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+            if (!supportedLspServerVersions.includes(version.serverVersion)) {
+                continue
+            }
+            for (const t of version.targets) {
+                if (
+                    (t.platform === process.platform || (t.platform === 'windows' && process.platform === 'win32')) &&
+                    t.arch === process.arch
+                ) {
+                    for (const content of t.contents) {
+                        if (content.filename.startsWith('qserver') && content.hashes.length > 0) {
```
Will-ShaoHua, line 157: nit: localize 'qserver'

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+            let downloadOk = await this.downloadAndCheckHash(zipFilePath, qserver)
+            if (!downloadOk) {
+                return false
+            }
+            const zip = new AdmZip(zipFilePath)
+            zip.extractAllTo(context.asAbsolutePath(path.join('resources')))
+            fs.removeSync(zipFilePath)
+
+            const noderuntimeFilePath = path.join(tempFolder, nodeBinName)
+            downloadOk = await this.downloadAndCheckHash(noderuntimeFilePath, noderuntime)
```
Will-ShaoHua, line 242: prefer making this a different variable rather than sharing the same var with L233

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+                        const usage = await LspClient.instance.getLspServerUsage()
+                        if (usage) {
+                            getLogger().info(
+                                `LspController: LSP server CPU ${usage.cpuUsage}%, LSP server Memory ${
+                                    usage.memoryUsage / (1024 * 1024)
+                                }MB  `
+                            )
+                        }
+                    },
+                    20 * 60 * 1000
```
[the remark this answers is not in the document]
leigaol: periodically log the LSP CPU and Memory usage. I will add more doc string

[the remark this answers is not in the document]
Will-ShaoHua: i guess it's == 20min?

packages/core/src/shared/telemetry/vscodeTelemetry.json
leigaol: No. Duration stands for the entire event duration while this new field only applies to the LSP query API latency

packages/core/src/amazonq/lsp/lspController.ts
leigaol: This is a filename in the CDN

packages/core/src/amazonq/lsp/lspController.ts
leigaol: updated in latest commit

packages/core/src/amazonq/lsp/lspController.ts
leigaol: added in latest commit

packages/core/src/amazonq/lsp/lspController.ts
leigaol: The language name and language runtime convertion is a tech debt . If we move these inside LSP then they can be reduced.

packages/core/src/amazonq/lsp/lspClient.ts
leigaol: the client has to be activated to become not null

packages/core/src/amazonq/lsp/lspClient.ts
leigaol: this activate function is only called after the download is complete.

packages/core/src/amazonq/lsp/lspController.ts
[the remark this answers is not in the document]
leigaol: I changed it to 30 min. Updated doc string.

packages/core/src/amazonq/lsp/lspClient.ts
Will-ShaoHua: could we activate the underlying LanguageClient before initialize this LspClient?

Will-ShaoHua (approved):

packages/core/src/shared/utilities/workspaceUtils.ts
leigaol: That `collectFiles` is owned by security scan team. I have created a backlog item.

packages/core/src/amazonq/lsp/lspClient.ts
leigaol: no because the language client creates a local IPC connection. 

I had to create a lsp client class to. make generating unit test easier.

packages/amazonq/test/unit/amazonq/lsp/lspClient.test.ts
leigaol: I still need the spy to verify what was encrypted.

packages/core/src/amazonq/lsp/lspClient.ts
leigaol: 1. There are strange dependency issue if I add this in deactivate of extension
2. Without calling this function, lsp process is still shutdown if I quit VS Code.

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -0,0 +1,378 @@
...
+
+import * as vscode from 'vscode'
+import * as path from 'path'
+import * as fs from 'fs-extra'
+import * as crypto from 'crypto'
+import { getLogger } from '../../shared/logger/logger'
+import { CurrentWsFolders, collectFilesForIndex } from '../../shared/utilities/workspaceUtils'
+import fetch from 'node-fetch'
+import { LspClient } from './lspClient'
+import AdmZip from 'adm-zip'
```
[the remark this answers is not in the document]
leigaol: done.

packages/amazonq/package.json
leigaol: removed prefix "local"

packages/core/src/amazonq/lsp/lspController.ts
leigaol: 1. I need to use `const file = fs.createWriteStream(localFile)`. This is not supported in current fs.ts
2. `fs-extra` is widely used across this code base.
3. This feature does not support Web

packages/core/src/amazonq/lsp/lspController.ts
leigaol: No it does not pipe to a file writestream. 

I got resp.body?.pipeTo is not a function

leigaol: But I am using request.fetch when getting the manifest.

packages/core/src/amazonq/lsp/lspClient.ts
leigaol: it must be a singleton. There should be only one Q LSP client for this server.

packages/core/src/amazonq/lsp/lspController.ts
leigaol: Yes. 

However, I don't understand how this run works with try catch. If I catch the exception and do not throw it, does it set the result to failed with reason?

justinmk3: > If I catch the exception and do not throw it, does it set the result to failed with reason?

No. Let run() handle the exception. You can always follow it with a `.catch()` if you don't want the exception to bubble up, but if this is part of a vscode command or webview, then Toolkit automatically handles the exception by showing a user message.

packages/core/src/amazonq/lsp/lspController.ts
justinmk3: > need to use const file = fs.createWriteStream(localFile)

ok,  fs.ts doesn't have an answer for that yet.

> fs-extra is widely used across this code base

we are in the process of removing it. using it more adds work :)

> This feature does not support Web

that might be the case now, but it can change in the future. and web support in general is more feasible when most modules are using fs.ts instead of node-only things.

justinmk3 (approved):

juntylpro1 (commented):
async function lastAsyncTask() {
  await null;
  throw new Error("failed");
}

async function withAwait() {
  return await lastAsyncTask();
}

withAwait();

// Error: failed
//    at lastAsyncTask
//    at async withAwait

PR #5278, opened 2024-07-08
fix(appcomposer): add warning message to version 1.91.0 about drag an…

JLargent42:
## Problem:
VS Code version 1.91.0 has a bug causing the dragEnd action to not trigger in WebViews unless Shift is held down. This breaks App Composer's ability to add new resources.

## Solution:
This has already been [reverted](https://github.com/microsoft/vscode/pull/219896) and should be fixed in the next VS Code version. In the meantime, this adds a warning message (upon opening Composer with version 1.91.0) informing users of the workaround.

image

## License


hayemaxi (approved):

ssibrahimpur:
Kindly make an immediate release with the solution as this is blocking our production release for a plugin which uses drag drop feature.

justinmk3:
@ssibrahimpur AWS Toolkit was released with this fix. Did you mean to comment on the vscode repo?

Issue #5277, opened 2024-07-08
uv_os_get_passwd returned ENOENT if user uses AD (ldap or similar)


memorsolutions:
## Problem

Both the AWS Toolkit and Amazon Q extensions produce the following error when the extension is activated and the effective user account is one that is defined externally (e.g., AD, LDAP).
`[error] SystemError: A system error occurred: uv_os_get_passwd returned ENOENT (no such file or directory)`

This appears to have been introduced by #5215 with the addition of the call to os.userInfo in getUsername (env.ts).

In such cases, the effective user ID does not have a matching entry in /etc/passwd and userInfo throw an exception. This error is discussed in several issues in other repos such as:
- https://github.com/microsoft/vscode-remote-release/issues/9649
- https://github.com/cyjake/ssh-config/issues/77

Might a fallback to process.env.USER be an acceptable approach for these extensions?

NOTE: Recent restructuring in https://github.com/aws/aws-toolkit-vscode/commit/c21dce155b5b3d3cd70094c843db6e8ae5b32417 appears to retain this bug.

## Steps to reproduce the issue

1. Login as a user that is synced to AD/LDAP.
2. Install AWS Toolkit and Amazon Q
3. See errors in `~/.config/Code/logs/<date and time>/window1/exthost/exthost.log`
```
2024-07-08 12:09:53.596 [info] ExtensionService#_doActivateExtension amazonwebservices.amazon-q-vscode, startup: false, activationEvent: 'onStartupFinished'
2024-07-08 12:09:54.179 [error] Activating extension amazonwebservices.amazon-q-vscode failed due to an error:
2024-07-08 12:09:54.179 [error] SystemError: A system error occurred: uv_os_get_passwd returned ENOENT (no such file or directory)
        at new SystemError (node:internal/errors:257:5)
        at new NodeError (node:internal/errors:368:7)
        at Object.userInfo (node:os:365:11)
        at B (/home/jcarswell@GatorFam5.local/.vscode/extensions/amazonwebservices.amazon-q-vscode-1.11.0/dist/src/extension.js:6527:34920)
        at h (/home/jcarswell@GatorFam5.local/.vscode/extensions/amazonwebservices.amazon-q-vscode-1.11.0/dist/src/extension.js:6464:1883)
        at N.stop (/home/jcarswell@GatorFam5.local/.vscode/extensions/amazonwebservices.amazon-q-vscode-1.11.0/dist/src/extension.js:6505:14214)
[...]

PR #5215, opened 2024-06-24
telemetry: scrub reasonDesc field

justinmk3:
followup to f010123ef2dd

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5279, opened 2024-07-09
fix(env): getUserInfo() may fail on AD/LDAP

justinmk3:
## Problem

Both AWS Toolkit and Amazon Q fail to start when the effective user account is defined externally (e.g. by AD/LDAP). #5277

    [error] SystemError: A system error occurred: uv_os_get_passwd returned ENOENT (no such file or directory)

Bisected to #5215 which calls os.userInfo in getUsername (env.ts).

When the effective user ID does not have a matching entry in /etc/passwd, userInfo throws an exception. Discussed in:
- https://github.com/microsoft/vscode-remote-release/issues/9649
- https://github.com/cyjake/ssh-config/issues/77

### Steps to reproduce

1. Login as a user that is synced to AD/LDAP.
2. Install AWS Toolkit and Amazon Q
3. Errors in `~/.config/Code/logs/<date and time>/window1/exthost/exthost.log`
   ```
   2024-07-08 12:09:54.179 [error] Activating extension amazonwebservices.amazon-q-vscode failed due to an error:
   2024-07-08 12:09:54.179 [error] SystemError: A system error occurred: uv_os_get_passwd returned ENOENT (no such file or directory)
           at new SystemError (node:internal/errors:257:5)
           at new NodeError (node:internal/errors:368:7)
           at Object.userInfo (node:os:365:11)
   ```

## Solution
- Move getUserInfo/getUsername into fs.ts since they depend on the filesystem.
- If userInfo fails, fallback to (in order of precedence):
  - process.env.USER
  - getUserHomeDir() directory name
  - "unknown-user"

## License

### DIFF
--- a/packages/core/src/shared/vscode/env.ts
+++ b/packages/core/src/shared/vscode/env.ts
@@ -5,7 +5,6 @@
 
 import * as semver from 'semver'
 import * as vscode from 'vscode'
-import * as os from 'os'
 import * as packageJson from '../../../package.json'
 import { getLogger } from '../logger'
 import { onceChanged } from '../utilities/functionUtils'
@@ -104,9 +103,9 @@ export function isRemoteWorkspace(): boolean {
 
 /** Returns true if OS is Windows. */
 export function isWin(): boolean {
-    if (isWeb()) {
-        return false
-    }
+    // if (isWeb()) {
+    //     return false
+    // }
 
     return process.platform === 'win32'
 }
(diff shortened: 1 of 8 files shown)


packages/core/src/shared/vscode/env.ts
```diff
@@ -104,9 +103,9 @@ export function isRemoteWorkspace(): boolean {
 
 /** Returns true if OS is Windows. */
 export function isWin(): boolean {
-    if (isWeb()) {
-        return false
-    }
+    // if (isWeb()) {
+    //     return false
+    // }
```
hayemaxi, line 108: why is this commented out?

justinmk3: I added that recently to avoid using `process` on web, but it might be more trouble than benefit. Because `isWin` is called before globals are initialized (thus `isWeb()` fails).

If `process` is polyfilled for web-mode, then this check can be dropped.

justinmk3:
changelog: https://github.com/aws/aws-toolkit-vscode/commit/10dacd932e04f5b0d377a1d516f2eef7f6ffa2a1

PR #5280, opened 2024-07-09
refactor(amazonqFeatureDev): include exceptions treated in code generation status detail

tverney:
## Problem

Currently we don't handle specific Guardrails type exception for code generation status detail, which leads to throw a general error and not help user self diagnose the prompt.

## Solution

This PR proposes:

- Specific error handling for  GuardrailsException, PromptRefusalException, EmptyPatchException, ThrottlingException   
- Small refactor to use CodeGenerationStatus enum instead of single string

## License


packages/core/src/amazonqFeatureDev/errors.ts
```diff
@@ -57,6 +57,44 @@ export class SelectedFolderNotInWorkspaceFolderError extends ToolkitError {
...
+        super("I'm sorry, I'm having trouble generating your code. Please try again.", {
+            code: 'EmptyPatchException',
+        })
+    }
+}
+
+export class ThrottlingException extends ToolkitError {
+    constructor() {
+        super(
+            "I'm sorry, I'm experiencing high demand at the moment and can't generate your code. This attempt won't count toward usage limits. Please try again.",
```
justinmk3, line 90: Do these new errors map to service-side errors? If so, can we define a `FeatureDevServiceError extends ToolkitError` instead of defining a new error class for every single failure mode?

packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -275,24 +283,31 @@ abstract class CodeGenBase {
...
-                        Code generation EmptyPatchException
-                        Code generation ThrottlingException
-                     * 
-                     * */
-                    throw new ToolkitError('Code generation failed', { code: 'CodeGenFailed' })
+                case CodeGenerationStatus.PREDICT_FAILED:
+                case CodeGenerationStatus.DEBATE_FAILED:
+                case CodeGenerationStatus.FAILED: {
+                    switch (true) {
+                        case codegenResult.codeGenerationStatusDetail?.includes('Guardrails'): {
```
justinmk3, line 295: what is the source of `codeGenerationStatusDetail`? is it an exception caught by mynah, or LSP, or some other component? If it's an exception, does it have a `code` that can be checked instead of string-matching the message?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -262,7 +267,13 @@ export class FeatureDevController {
                     ],
                 })
                 break
+            case PromptRefusalException.name:
+                this.messenger.sendErrorMessage(errorMessage, message.tabID, 0, session?.conversationIdUnsafe, true)
```
justinmk3, line 271: where is the code that sets `errorMessage` ? Does it use `getTelemetryReasonDesc` ? If the error message is sent to telemetry, `getTelemetryReasonDesc` should be used somewhere in the codepath to ensure that PII is not sent to telemetry. https://github.com/aws/aws-toolkit-vscode/blob/7740f3af6872d2d9b6ea98653ef46061e0b3bdbe/packages/core/src/shared/errors.ts#L397

packages/core/src/amazonqFeatureDev/session/sessionState.ts
tverney: Our BE its returning the exception in this format:
```
reason: ${error_type.slice(0, error_type.indexOf('Exception'))}
```
Which leads to a combination of exception plus the error code like this:
```
Mocked error message reason: Guardrails
```
So answering your question, don't have a code because the identifier comes with the message itself so requires string-match those.

justinmk3: Why doesn't the backend model the errors with a code? Surely this is possible with smithy/coral? Do you have plans to change this?

This is resulting in a lot of low-value boilerplate code in all of your clients (such as this repo).

packages/core/src/amazonqFeatureDev/errors.ts
tverney: The reason that we have the mapping is to use and respond differently for those if we need. Some require retries, other don't, so have them separately help map those internally. But I get your point. One thing that we can make is a FeatureDevServiceError and send over the message we want when we're creating. We use this in all that have the same behaviour expected (message with retries).

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
tverney: No, as far I've checked in this latest version, we're not sending this errorMessage to telemetry. We can, but then I agree with you that we need sanitize PII before doing that.

packages/core/src/amazonqFeatureDev/session/sessionState.ts
tverney: Yes, we've had a brief discussion about this internally. BE team said that we couldn't afford an API shape model change at this point because would require an extra bar raising process. So, for now, as far I've understood from their response, it was decided to leverage the existing details field mapping the exception code within the message. Those strings are extracted from the exception code, the only difference is that its combined with the details string, which is a details message coming from the BE which wasn't being filled until now.

jeffpyj: Yes we have plans to improve the API shape for error propagation. In Q4 there is a planned UX/UI revamp work of our service and we plan to improve the backend models as part of that effort. This is more of a short term improvement to reduce some pain points that our customers have encountered.

packages/core/src/amazonqFeatureDev/errors.ts
```diff
@@ -57,6 +57,23 @@ export class SelectedFolderNotInWorkspaceFolderError extends ToolkitError {
     }
 }
 
+export class PromptRefusalException extends ToolkitError {
+    constructor() {
+        super(
+            'I\'m sorry, I can\'t generate code for your request. Please make sure your message and code files comply with the <a href="https://aws.amazon.com/machine-learning/responsible-ai/policy/">AWS Responsible AI Policy.</a>',
```
linyuxi0511, line 63: Just to confirm if these error msgs got reviewed by doc writers.

tverney: Yup, I got this extract from the doc writer errors string review doc. I can DM you with the link,

linyuxi0511 (approved):

packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -282,24 +288,40 @@ abstract class CodeGenBase {
...
+                            throw new FeatureDevServiceError(
+                                "I'm sorry, I'm having trouble generating your code. Please try again.",
```
jpinkney-aws, line 302: Has there been any thought on de-duping some of the strings? I'm seeing the same string on line 311

jpinkney-aws (approved):

packages/core/src/amazonqFeatureDev/session/sessionState.ts
tverney: That's a good one. We don't have in VSCode toolkit a .properties file like in JB, but we somehow can avoid these duplications. I'll let this for we dedupe in a specific task to handle i18n, if you're ok with it.

hayemaxi: `package.nls.json` is a good usecase for this, but even a simple top-level variable is better than duplication here.

tverney: Sounds good, I'll push a rev removing this duplication. Thanks

tverney: Done. @hayemaxi are we good to merge this one?

nkomonen-amazon (approved):

PR #5282, opened 2024-07-09
feat: proactively show code generation iterations

linyuxi0511:
## Problem
Currently, there is no information shown to customers in the chat regarding the number of iterations they could run during code generation.
They only get to know this on hitting the iteration limits and seeing the error like you have reached the limit for number of iterations.
## Solution
Hence, with this change, we are showing the remaining iterations proactively by exposing the remaining iteration numbers and total iteration numbers to the end of each code generation. 

## License

### DIFF
--- a/packages/amazonq/.changes/next-release/Feature-4a1c951a-2289-4808-b106-86fe4f03528e.json
+++ b/packages/amazonq/.changes/next-release/Feature-4a1c951a-2289-4808-b106-86fe4f03528e.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Amazon Q/dev proactively show code generation iterations"
+}
--- a/packages/core/src/amazonqFeatureDev/session/sessionState.ts
+++ b/packages/core/src/amazonqFeatureDev/session/sessionState.ts
@@ -254,6 +254,8 @@ abstract class CodeGenBase {
         newFiles: NewFileInfo[]
         deletedFiles: DeletedFileInfo[]
         references: CodeReference[]
+        codeGenerationRemainingIterationCount?: number
+        codeGenerationTotalIterationCount?: number
     }> {
         for (
             let pollingIteration = 0;
@@ -261,6 +263,9 @@ abstract class CodeGenBase {
             ++pollingIteration
         ) {
             const codegenResult = await this.config.proxyClient.getCodeGeneration(this.conversationId, codeGenerationId)
+            const codeGenerationRemainingIterationCount = codegenResult.codeGenerationRemainingIterationCount
+            const codeGenerationTotalIterationCount = codegenResult.codeGenerationTotalIterationCount
+
             getLogger().debug(`Codegen response: %O`, codegenResult)
             telemetry.setCodeGenerationResult(codegenResult.codeGenerationStatus.status)
             switch (codegenResult.codeGenerationStatus.status) {
@@ -273,6 +278,8 @@ abstract class CodeGenBase {
                         newFiles: newFileInfo,
                         deletedFiles: getDeletedFileInfos(deletedFiles, workspaceFolders),
                         references,
+                        codeGenerationRemainingIterationCount: codeGenerationRemainingIterationCount,
+                        codeGenerationTotalIterationCount: codeGenerationTotalIterationCount,
                     }
                 }
                 case 'predict-ready':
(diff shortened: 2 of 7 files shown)


linyuxi0511:
Screenshot 2024-07-09 at 5 54 34 PM
Screenshot 2024-07-09 at 5 55 07 PM

linyuxi0511:
Original task: https://taskei.amazon.dev/tasks/D138685414
Back-end changes are already in prod.

packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -321,7 +328,9 @@ export class CodeGenState extends CodeGenBase implements SessionState {
         public deletedFiles: DeletedFileInfo[],
         public references: CodeReference[],
         tabID: string,
-        private currentIteration: number
+        private currentIteration: number,
+        public codeGenerationRemainingIterationCount?: number,
```
jeffpyj, line 332: Consider putting public fields before private fields

packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -321,7 +328,9 @@ export class CodeGenState extends CodeGenBase implements SessionState {
...
+        public codeGenerationTotalIterationCount?: number
```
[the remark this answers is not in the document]
linyuxi0511: You are right, initially made them optional to accommodate the current behaviour. Since the backend changes are in now, they can be required.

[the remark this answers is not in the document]
linyuxi0511: Actually, we do. need it to be optional, since at the very first time when we asking for code generation, we don't know this iteration counts.

packages/core/src/amazonqFeatureDev/session/sessionState.ts
linyuxi0511: We couldn't because "A required parameter cannot follow an optional parameter."

packages/amazonq/.changes/next-release/Feature-4a1c951a-2289-4808-b106-86fe4f03528e.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "proactively show code generation iterations"
```
tverney, line 3: One suggestion to be more specific to our vertical here and talk about the counts - "Amazon Q /dev: proactively show code generation iterations count"

packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -367,7 +379,9 @@ export class CodeGenState extends CodeGenBase implements SessionState {
                     this.deletedFiles,
                     this.references,
                     this.tabID,
-                    this.currentIteration + 1
+                    this.currentIteration + 1,
```
tverney, line 382: curious: there is any risk that currentIteration be different from those counts? And if its, have we tested what is shown to the user?

packages/amazonq/.changes/next-release/Feature-4a1c951a-2289-4808-b106-86fe4f03528e.json
linyuxi0511: Updated.

packages/core/src/amazonqFeatureDev/session/sessionState.ts
linyuxi0511: Ideally the currentIteration would be simply "totaliteration-remainingIteration". This is calculated in the front end. 
However we decided to expose and use the API carried values since they are got from the source of truth - our ddb table.

tverney: OK, so do we need remove currentIteration since we now have new counts?

linyuxi0511: And this field `currentIteration` is not actually being used for now. I think the idea is to use iterationCounts instead.

tverney (approved):

packages/core/src/amazonqFeatureDev/session/sessionState.ts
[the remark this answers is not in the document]
jeffpyj: Ah I see, this is because this is getting CodeGenResult instead of CodeGenStatus. Sounds good

jeffpyj (approved):

jeffpyj:
CI / test Windows (16.x, stable) (pull_request) seems to be failing. Is that expected?

jpinkney-aws:
This kind of feels like it should be inside of that diff view panel rather then a chat message that happens after

linyuxi0511:
Yes this is reviewed with UX. 
Also this is considered to be a short-term solution as we are going to improve the UI/UX shotly.

linyuxi0511:
Confirmed with @jpinkney-aws  that the tests are known to be flaky.

jpinkney-aws (approved):

PR #5283, opened 2024-07-09
Add postMessage embedded APIs; set isEmbedded global

AwsBrettBailey:
## Problem

`sendMessage` and `receiveMessage` VSCode proposed APIs, which are made available on VSCode for Web via a fork of the core code-oss project, need to be accessible to the AWS Toolkit. Additionally, Toolkit should be aware of when it is being run in an embedded context.

## Solution

Detect if `sendMessage` and `receiveMessage` APIs are available on extension activate. If they are available, VSCode is running in a context which supports these two APIs.

In a future commit, we will add support for a unique auth strategy when embedded mode is detected. Sending this up as an initial PR, so I can learn the process and ensure that this is an acceptable base from which to start.

Changelog: I think this should not be required as there are not yet any user facing changes
Test coverage: I added a minimal test in the style of `web.test.ts`
Testing: I have some local changes that enable codewhisperer with an "embedded" credentials strategy, which is working as expected.

## License


packages/core/src/shared/extensionGlobals.ts
```diff
@@ -160,6 +160,10 @@ export function isWeb() {
     return globals.isWeb
 }
 
+export function isEmbedded() {
+    return globals.isEmbedded
```
justinmk3, line 164: can callers use `globals.isEmbedded` instead? It's not really a goal to have module-level aliases to all of the `globals` fields.

packages/core/src/shared/extensionGlobals.ts
```diff
@@ -217,4 +221,6 @@ interface ToolkitGlobals {
     }
     /** If this extension is running in Web mode (the browser), compared to running on the desktop (node) */
     isWeb: boolean
+    /** If this extensions is hosted in a parent browser context that supports postMessage between frames */
```
justinmk3, line 224:
```suggestion
    /** If this extension is hosted in a parent browser context that supports postMessage between frames */
```

packages/core/src/extensionCommon.ts
```diff
@@ -81,6 +81,7 @@ export async function activateCommon(
     await initializeComputeRegion()
 
     globals.contextPrefix = '' //todo: disconnect supplied argument
+    globals.isEmbedded = !!(vscode.window.sendMessage && vscode.window.receiveMessage)
```
justinmk3, line 84: This should live in `initialize` https://github.com/aws/aws-toolkit-vscode/blob/2ef5d5e888fe634d2e04d0d1d2238b99a2b1b824/packages/core/src/shared/extensionGlobals.ts#L139

packages/core/src/test/embedded.test.ts
```diff
@@ -0,0 +1,13 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as assert from 'assert'
+import { isEmbedded } from '../shared/extensionGlobals'
+
+describe('isEmbedded', function () {
```
justinmk3, line 9: I assume more tests for "embedded mode" will be added here. Otherwise it might make more sense to create a `globals.test.ts` or `startup.test.ts` for testing various init state.

packages/core/src/shared/extensionGlobals.ts
AwsBrettBailey: Sure. I was just following the pattern for `isWeb`, but if it's preferred to access globals directly, I will make that change.

packages/core/src/test/embedded.test.ts
AwsBrettBailey: OK. If I add one of these files and include only tests relevant to the `embedded` global, will that be sufficient?

packages/core/src/shared/extensionGlobals.ts
justinmk3: isWeb is a special case rather than a pattern

packages/core/src/test/embedded.test.ts
justinmk3: yes

AwsBrettBailey:
Fixing issue with the most recent commit.

AwsBrettBailey:
Remaining test failure appear to be caused by replacement of the `globals` object after `initialize` is called. Copying my message from slack here for explanation of my blockers here:

> Would someone from the toolkit team be available (hopefully today) to walk me through some challenges I'm facing testing the globals initialize function. The tests do some setup and make some assumptions in globalSetup.test.ts about the globals object. Here's the specific issues I'm facing as I understand them:
> * Tests are using this [module scope globals object](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/extensionGlobals.ts#L134). This is intended for web mode where globals is shared. Calling initialize [replaces](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/extensionGlobals.ts#L144) this object, meaning I have to duplicate some setup of the globals object from global mocha [beforeEach and afterEach hooks](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/test/globalSetup.test.ts#L64-L113). I can work around this via that duplication, but it seems like the tests should be using the globals object from initialize in the first place, unless they are a web mode specific test.
> * As requested by comments on my above CR, I have moved detection for isEmbedded into initialize. Detection works by checkin the vscode.window object for sendMessage and receiveMessage methods. I would like to test cases where the methods are present and not present. This means that I need to be able to provide a different vscode.window object, or have the ability to modify the mocked window with mocks for those methods. vscode.window is configured and set to the result of getTestWindow in the [global mocha hooks here](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/test/globalSetup.test.ts#L73).

> Basically, what I'm saying is that the global setup for the tests seems to assume that globals has already been initialized and testing initialize in this framework is difficult. It doesn't seem like there are sufficiently different test setups for web vs non-web, or the ability to opt-in or out of setup for the two modes.

> I am still new to the test setup, so maybe I am overthinking this and there is a simpler solution, but I don't think it is in scope of my PR to modify the test setup substantially in order to test globals initialize and isEmbedded functionality.

nkomonen-amazon:
@AwsBrettBailey regarding your latest comment. I think a simple solution would be to pass the `vscode` object as an argument to [`initialize()`](https://github.com/aws/aws-toolkit-vscode/blob/469fa91eb79922bfc1acb43eceb3e5acaaf7fb66/packages/core/src/shared/extensionGlobals.ts#L141), similar to how we do it with `isWeb`. Additionally you could have it default to the `vscode` instance. Then in your tests you could pass in your modified `vscode`.

packages/core/src/shared/extensionGlobals.ts
```diff
@@ -151,6 +151,7 @@ export function initialize(context: ExtensionContext, isWeb: boolean = false): T
         manifestPaths: {} as ToolkitGlobals['manifestPaths'],
         visualizationResourcePaths: {} as ToolkitGlobals['visualizationResourcePaths'],
         isWeb,
+        isEmbedded: !!(vscode.window.sendMessage && vscode.window.receiveMessage),
```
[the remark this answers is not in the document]
AwsBrettBailey: Yeah, I was considering this also. Makes sense.

justinmk3:
CI is failing. Do you plan to continue working on this?

AwsBrettBailey:
Yes, but I'm not sure when the next opportunity will be. If you would prefer to close, that is OK and we can reopen in the future. Otherwise, we will resume work here when we have a bit more bandwidth. Apologies for the delay.

PR #5287, opened 2024-07-10
fix: eslint DeprecationWarning

justinmk3:
## Problem

    (node:99880) DeprecationWarning: "no-incorrect-once-usage" rule is
    using `context.getScope()`, which is deprecated and will be removed
    in ESLint v9. Please use `sourceCode.getScope()` instead. (Use `node
    --trace-deprecation ...` to show where the warning was created)

## Solution

update rule

## License

### DIFF
--- a/plugins/eslint-plugin-aws-toolkits/lib/rules/no-incorrect-once-usage.ts
+++ b/plugins/eslint-plugin-aws-toolkits/lib/rules/no-incorrect-once-usage.ts
@@ -122,7 +122,7 @@ export default ESLintUtils.RuleCreator.withoutDocs({
                         declaration.init.callee?.type === AST_NODE_TYPES.Identifier &&
                         declaration.init.callee.name === 'once'
                     ) {
-                        const scope = context.getScope()
+                        const scope = context.sourceCode.getScope(declaration) // TODO: should this be `getScope(node)`?
                         const variable = scope.variables.find(
                             v => v.name === (declaration.id as TSESTree.Identifier).name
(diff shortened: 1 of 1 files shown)


plugins/eslint-plugin-aws-toolkits/lib/rules/no-incorrect-once-usage.ts
```diff
@@ -122,7 +122,7 @@ export default ESLintUtils.RuleCreator.withoutDocs({
                         declaration.init.callee?.type === AST_NODE_TYPES.Identifier &&
                         declaration.init.callee.name === 'once'
                     ) {
-                        const scope = context.getScope()
+                        const scope = context.sourceCode.getScope(declaration) // TODO: should this be `getScope(node)`?
```
justinmk3, line 125: @hayemaxi does this look right?

hayemaxi: Looks fine to me, AFAIK they are identical.
Consider:
```
L1 const x = 1
L2 const y = 2, z = 3
```
There are two VariableDeclarations here (L1, L2). L1 node.declarations is `[x = 1]`. L2 node.declarations is `[y = 2, z = 3]`.

Whether you take the root node level (L1/L2) or the individual declarations in each root, the scope is the same.

```suggestion
                        const scope = context.sourceCode.getScope(declaration)
```

hayemaxi (approved):
Thanks!

PR #5289, opened 2024-07-10
Timeout error

nkomonen-amazon:
## Problem:

When we attempt to refresh an SSO session, we sometimes saw
a TimeoutError in our telemetry as the reason for refresh failure.

## Solution:

We treat the TimeoutError as a network error.
See `isSocketTimeoutError()` docstring for more details.

## License

### DIFF
--- a/packages/core/src/auth/sso/clients.ts
+++ b/packages/core/src/auth/sso/clients.ts
@@ -102,17 +102,9 @@ export class OidcClient {
                 return true
             }
 
-            // Sessions may "expire" earlier than expected due to network faults.
-            // TODO: add more cases from node_modules/@aws-sdk/client-sso-oidc/dist-types/models/models_0.d.ts
-            // ExpiredTokenException
-            // InternalServerException
-            // InvalidClientException
-            // InvalidRequestException
-            // SlowDownException
-            // UnsupportedGrantTypeException
-            // InvalidRequestRegionException
-            // InvalidRedirectUriException
-            // InvalidRedirectUriException
+            // As part of SIM IDE-10703, there was an assumption that retrying on InvalidGrantException
+            // may be useful. This may not be the case anymore and if more research is done, this may not be needed.
+            // TODO: setup some telemetry to see if there are any successes on a subsequent retry for this case.
             return err.name === 'InvalidGrantException'
         }
         const client = new SSOOIDC({
@@ -124,7 +116,9 @@ export class OidcClient {
             ),
             customUserAgent: getUserAgent({ includePlatform: true, includeClientId: true }),
             requestHandler: {
-                requestTimeout: 30_000,
+                // This field may have a bug: https://github.com/aws/aws-sdk-js-v3/issues/6271
+                // If the bug is real but is fixed, then we can probably remove this field and just have no timeout by default
+                requestTimeout: 5000,
             },
         })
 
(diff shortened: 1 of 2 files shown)


packages/core/src/shared/errors.ts
```diff
@@ -760,7 +760,11 @@ export class PermissionsError extends ToolkitError {
 }
 
 export function isNetworkError(err?: unknown): err is Error & { code: string } {
-    if (isVSCodeProxyError(err)) {
+    if (!(err instanceof Error)) {
+        return false
+    }
+
+    if (isVSCodeProxyError(err) || isTimeoutError(err)) {
```
justinmk3, line 767: is `isTimeoutError` always a network error? if we want to keep the semantics of `isNetworkError` , perhaps the caller should check `isNetworkError || isTimeoutError` instead of embedding it here

nkomonen-amazon: In the aws-sdk pr linked above `TimeoutError` represents: `"ECONNRESET", "EPIPE", "ETIMEDOUT"`. Which all look to be network errors

justinmk3: Seems likely that `err.name === 'TimeoutError'` could have false positives. Can we (also) check for `"ECONNRESET", "EPIPE", "ETIMEDOUT"` somehow? Even `message.includes()` would be better than nothing, to avoid false positives.

`isNetworkError` is an interface that claims to do something very specific. If it silently lies about certain errors, that can cause hidden bugs.

nkomonen-amazon: Looked at the aws-sdk code and the `E...` errors are part of the `code` field of the error. So we can actually just add the missing error codes to our existing list and that should indirectly handle the TimeoutError.

packages/core/src/auth/sso/clients.ts
```diff
@@ -113,7 +113,7 @@ export class OidcClient {
             // InvalidRequestRegionException
             // InvalidRedirectUriException
             // InvalidRedirectUriException
-            return err.name === 'InvalidGrantException'
+            return false
```
justinmk3, line 116: should the comment above this be updated?

> Do not retry on InvalidGrantException since there is no point. When we get that error, there is no scenario where things will work again. This was indicated to us by someone on the identity team.

That's surprising. This code was added in https://github.com/aws/aws-toolkit-vscode/pull/3498 to fix a sev2. Reviewing that commit (and related, see also https://sim.amazon.com/issues/IDE-10703 ), can we form an explanation that gives confidence that this won't cause a regression? Did this change not actually fix the sev2?

nkomonen-amazon: Getting final verification again, but updated the ticket with references to previous conversations: https://sim.amazon.com/issues/IDE-10703?selectedConversation=e611a3c1-6924-477f-8b65-37638ee2a063

justinmk3: Nice. This would be a great comment to add in the source code here:

> retry on invalid_grant would not be recommended as it's an indication that refresh token is invalid now.

nkomonen-amazon: Updated the comments. It will now give more context and points to our SIM ticket

nkomonen-amazon:
@justinmk3 should be ready to merge now

PR #5238, opened 2024-06-27
feat(feature dev): Filter out non-code files at preparation state

willyyhuang:
## Problem

Feature dev compiles a compressed file with relevant local files to prepare response. This process would allow some irrelevant files to be added into the zip without many filters.

The current filtering pattern: ignore the gitignores and files that are big in size (> 1024000 kb).

## Solution

Implement an allowlist behaviour checking file extensions, we will be proactive to see which other files should be added in this list.

Filter out any files that aren't in the allowlist.

Tested by building extension locally, verified the s3 artifact size, also added unit test to reflect this behaviour.

Follow up: emit telemetry event, once it's defined.

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

---

Tested using the same repository, but 2 different extension version

Before, `master` branch behaviour
Screenshot 2024-06-27 at 12 37 09 PM

After, `wlsun/filter-non-code-files-feature-dev` branch behaviour
Screenshot 2024-06-27 at 12 37 04 PM

PR #5296, opened 2024-07-11
telemetry(feature dev): add ignore file metric while bundling repository

willyyhuang:
## Problem

We (feature dev team) want to understand if there are any other relevant code file extension that should be added to the allowlist.

## Solution

To obtain such information, we want to emit telemetry event to see the most common file extensions being ignored while bundling.

Example log: 
```
2024-07-11 10:35:18.127 [info] Amazon Q Developer Agent for software development Conversation ID: dea452f8-e685-4285-b350-900a06a40d51
2024-07-11 10:35:19.146 [info] ignoring file extension name: graphql, count: 82
2024-07-11 10:35:19.146 [info] ignoring file extension name: eslintignore, count: 1
2024-07-11 10:35:19.146 [info] ignoring file extension name: nvmrc, count: 1
```

This is a follow up of this [pull request](https://github.com/aws/aws-toolkit-vscode/pull/5238)

## License

### DIFF
--- a/packages/core/src/amazonqFeatureDev/util/files.ts
+++ b/packages/core/src/amazonqFeatureDev/util/files.ts
@@ -14,7 +14,7 @@ import { maxFileSizeBytes } from '../limits'
 import { createHash } from 'crypto'
 import { CurrentWsFolders } from '../types'
 import { ToolkitError } from '../../shared/errors'
-import { AmazonqCreateUpload, Metric } from '../../shared/telemetry/telemetry'
+import { AmazonqCreateUpload, Metric, telemetry as amznTelemetry } from '../../shared/telemetry/telemetry'
 import { TelemetryHelper } from './telemetryHelper'
 import { maxRepoSizeBytes } from '../constants'
 import { isCodeRelatedFile } from './fileExtension'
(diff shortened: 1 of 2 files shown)


packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -368,6 +368,16 @@
             "type": "string",
```
chengoramazon, line 368: Shall we add the definition in the aws-toolkiet-common pkg?

willyyhuang: Yes, will do

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -35,19 +35,40 @@ export async function prepareRepoData(
...
+            ignoredExtensionMap.forEach((value, key) => {
+                const event = {
+                    filenameExt: key,
+                    count: value,
+                }
 
-            const zipFolderPath = path.dirname(file.zipFilePath)
-            zip.addLocalFile(file.fileUri.fsPath, zipFolderPath)
-        }
+                bundleSpan.record(event)
```
justinmk3, line 69: FYI: `amazonq_bundleExtensionIgnored.run()` emits one metric. This `bundleSpan.record()` call will overwrite the fields for the "current" metric.

If you want to emit one `amazonq_bundleExtensionIgnored` metric for each filetype, call `amazonq_bundleExtensionIgnored.run()` multiple times (and thus probably want to move the for-loop out of this `run()`).

chengoramazon (approved):

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -49,6 +60,17 @@ export async function prepareRepoData(
             zip.addLocalFile(file.fileUri.fsPath, zipFolderPath)
         }
 
+        ignoredExtensionMap.forEach(async (value, key) => {
```
justinmk3, line 63: why not use a regular for-loop? if this async callback fails, will it correctly bubble up an exception? that is one reason why forEach() can lead to hidden bugs.

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -49,6 +60,17 @@ export async function prepareRepoData(
...
+            await amznTelemetry.amazonq_bundleExtensionIgnored.run(async (bundleSpan) => {
```
justinmk3, line 64: in this existing test for `prepareRepoData`, you can easily test the change in this PR using the `assertTelemetry` function.

the changes in this PR could easily have bugs, so that is worth doing

willyyhuang: alright, let me add tests

packages/core/src/amazonqFeatureDev/util/files.ts
willyyhuang: ok, will make a change

PR #5299, opened 2024-07-11
fix: refactor broke webview paths

nkomonen-amazon:
## Problem:

A recent refactor did not update certain static paths which prevent webviews from working.

## Solution:

Fix the incorrect paths.

## Testing it works

- Create a REST API Gateway on my AWS account
- In the AWS Explorer right click the REST API and select `Invoke on AWS`, this will open the webview.

## License

### DIFF
--- a/packages/core/src/awsService/apigateway/vue/invokeRemoteRestApi.ts
+++ b/packages/core/src/awsService/apigateway/vue/invokeRemoteRestApi.ts
@@ -42,7 +42,7 @@ export interface InvokeRemoteRestApiInitialData {
 }
 
 export class RemoteRestInvokeWebview extends VueWebview {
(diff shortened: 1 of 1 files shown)


hayemaxi (approved):

jpinkney-aws (approved):

justinmk3:
What about this one: https://github.com/aws/aws-toolkit-vscode/blob/87a1b76f7b14ad1446edcc241fa5cca3b44d8b61/packages/core/src/codecatalyst/vue/configure/backend.ts#L33 

and others?

nkomonen-amazon:
The only paths impacts were the folders moved in to `src/awsService`. I haven't moved codecatalyst, q, and some other folders in to it so they are fine. We will probably not touch Q, and the remaining we will look at individually. @justinmk3

PR #5306, opened 2024-07-11
Amazon Q: multiple issues fixes

andrewyuq:
1. Updated wording of data option settings to say " Amazon Q Developer Pro Tier instead of Business Tier"
2. Fixed an issue where SendTelemetryEvent call would fail due to undefined ProgrammingLanguage field
3. Fixed a chat hang issue from an infinite loop when trying to truncate chat context below limit.
Updated wording of data optin settings
4. Make auto trigger by default true even if enableCodeSuggestion is not called

## License


leigaol:
re-run CI?

packages/core/src/codewhisperer/models/model.ts
```diff
@@ -88,7 +88,7 @@ export class CodeSuggestionsState {
         return (this.#instance ??= new this())
     }
 
-    protected constructor(context: vscode.Memento = globals.context.globalState, fallback: boolean = false) {
+    protected constructor(context: vscode.Memento = globals.context.globalState, fallback: boolean = true) {
```
justinmk3, line 91: ci failure might be related to this change. tests may need to be updated

andrewyuq: tests fixed.

leigaol (approved):

zixlin7 (approved):

jpinkney-aws (approved):

Issue #3662, opened 2023-07-17
build: enable eslint rules for arrow functions


justinmk3:
# Problem

javascript arrow functions allow braces `{ ... }` to surround an expression without a `return`, which leads to bugs such as https://github.com/aws/aws-toolkit-vscode/pull/3659

Example:

```javascript
return supplementalContexts.filter(item => { item.content.trim().length !== 0 })
```

should be:

```javascript
return supplementalContexts.filter(item => item.content.trim().length !== 0)
```

# Solution

Enable these rules:

- https://eslint.org/docs/latest/rules/no-confusing-arrow
- https://eslint.org/docs/latest/rules/arrow-body-style
- https://eslint.org/docs/latest/rules/no-arrow-condition

Evaluate other arrow-function related estlint rules and potentially enable them.

PR #5307, opened 2024-07-12
lint: use gitignore, enforce prettier rules

justinmk3:
## Problem

- .eslintignore is mostly redundant with .gitignore and is missing something specified by .gitignore.
- some prettier rules are not enforced by our eslint rules. For example, if a developer doesn't have prettier configured, they may commit files without EOL and this isn't caught by our lint CI.

## Solution

- use `eslint --ignore-path .gitignore` and remove .eslintignore
- install `eslint-plugin-prettier` which automatically enforces prettier rules as eslint rules.

Note: the formatting changes are from `arrowParens=always`. See commit for details.

## License


package.json
```diff
@@ -16,7 +16,7 @@
         "singleQuote": true,
         "semi": false,
         "bracketSpacing": true,
-        "arrowParens": "avoid",
+        "arrowParens": "always",
```
justinmk3, line 19: Partial solution for https://github.com/aws/aws-toolkit-vscode/issues/3662

package.json
```diff
@@ -32,10 +32,8 @@
         "package": "npm run package -w packages/toolkit -w packages/amazonq",
         "newChange": "echo 'Must specify subproject/workspace with -w packages/<subproject>' && false",
         "createRelease": "echo 'Must specify subproject/workspace with -w packages/<subproject>' && false",
-        "format": "prettier --check plugins && npm run format -w packages/ --if-present",
-        "formatfix": "prettier --write plugins && npm run formatfix -w packages/ --if-present",
```
justinmk3, line 36: Removed all `format` and `formatfix` tasks. They're redundant now that eslint enforces prettier rules.

package.json
```diff
@@ -32,10 +32,8 @@
...
-        "lint": "npm run lint -w packages/ --if-present && npm run format",
-        "lintfix": "eslint -c .eslintrc.js --fix --ext .ts packages plugins && npm run formatfix",
+        "lint": "npm run lint -w packages/ --if-present",
+        "lintfix": "eslint -c .eslintrc.js --ignore-path .gitignore --ignore-pattern '**/*.json' --ignore-pattern '**/*.gen.ts' --ignore-pattern '**/types/*.d.ts' --ignore-pattern '**/src/testFixtures/**' --ignore-pattern '**/resources/js/graphStateMachine.js' --fix --ext .ts packages plugins",
```
justinmk3, line 36: Future: need to extract the eslint invocation to something like `scripts/lint.ts`.

packages/core/src/testLint/eslint.test.ts
```diff
@@ -15,16 +15,32 @@ describe('eslint', function () {
...
+                '../toolkit',
+                // TODO: fix lint issues in scripts/
+                // '../../scripts',
             ],
             {
                 throws: false,
             }
         )
-        assert.strictEqual(result.status, 0, result.stdout.toString())
+        assert.strictEqual(result.status, 0, result.output.toString())
```
justinmk3, line 44: stderr wasn't shown

packages/core/src/testLint/eslint.test.ts
```diff
@@ -15,16 +15,32 @@ describe('eslint', function () {
...
+                '--ignore-pattern',
+                '**/resources/js/graphStateMachine.js',
                 '--ext',
                 '.ts',
-                '.',
+                '../amazonq',
+                '../core',
+                '../toolkit',
+                // TODO: fix lint issues in scripts/
+                // '../../scripts',
```
justinmk3, line 38: run eslint on all packages. 

Future: `testLint/` should live outside of core/

packages/core/src/testLint/eslint.test.ts
```diff
@@ -15,16 +15,32 @@ describe('eslint', function () {
                 '../../node_modules/.bin/eslint',
                 '-c',
                 '../../.eslintrc.js',
+                // Note: eslint currently does not support multiple  --ignore-path args.
+                // Use --ignore-pattern as a workaround.
                 '--ignore-path',
-                '../../.eslintignore',
+                '../../.gitignore',
+                '--ignore-pattern',
+                '**/*.json',
```
justinmk3, line 23: re-use gitignore instead of eslintignore

package.json
hayemaxi: How does this partially solves the issue? Seems like `arrowParens` only adds parenthesis around the arguments, but the issue is relating to how arrow funcs structure their returns.

justinmk3: 😅  yeah, it doesn't directly solve https://github.com/aws/aws-toolkit-vscode/issues/3662 . I should have said "this is something I discovered while working on https://github.com/aws/aws-toolkit-vscode/issues/3662 , and since `arrowParens:always` is now the default for what appears to be a solid rationale, I enabled it".

hayemaxi (approved):

PR #5309, opened 2024-07-12
refactor: use globalState abstraction

justinmk3:
WIP

followup to https://github.com/aws/aws-toolkit-vscode/pull/5293

## License


packages/core/src/test/codewhisperer/commands/basicCommands.test.ts
```diff
@@ -75,35 +74,10 @@ describe('CodeWhisperer-basicCommands', function () {
         sinon.restore()
     })
 
-    it('test get()', async function () {
-        const fakeMemeto = new FakeMemento()
-        await fakeMemeto.update(CodeWhispererConstants.autoTriggerEnabledKey, true)
```
justinmk3, line 80: These cw get/set functions don't do anything except call globalState. So I removed them and these tests (which aren't testing anything useful).

packages/core/src/test/codewhisperer/testUtil.ts
```diff
@@ -32,7 +32,8 @@ export async function resetCodeWhispererGlobalVariables() {
     CodeWhispererCodeCoverageTracker.instances.clear()
     globals.telemetry.logger.clear()
     session.reset()
-    await CodeSuggestionsState.instance.setSuggestionsEnabled(false)
+    await globals.globalState.clear()
+    await CodeSuggestionsState.instance.setSuggestionsEnabled(true)
```
justinmk3, line 36: `true` is the default since https://github.com/aws/aws-toolkit-vscode/commit/2fed6a62288c291d22b6555abd4b23569fc31c3e . This wasn't caught because the tests use stubs/mocks. This is an example of why stubs/mocks are fragile and don't represent the actual application code.

packages/core/src/test/codewhisperer/commands/basicCommands.test.ts
```diff
@@ -477,7 +451,7 @@ describe('CodeWhisperer-basicCommands', function () {
             getTestWindow().onDidShowQuickPick(async (e) => {
                 e.assertItems([
                     createSeparator('Inline Suggestions'),
-                    createAutoSuggestions(false),
+                    createAutoSuggestions(true),
```
justinmk3, line 454: `true` is the default since https://github.com/aws/aws-toolkit-vscode/commit/2fed6a62288c291d22b6555abd4b23569fc31c3e . This wasn't caught because the tests use stubs/mocks. This is an example of why stubs/mocks are fragile and don't represent the actual application code.

packages/core/src/test/codewhisperer/commands/basicCommands.test.ts
```diff
@@ -426,10 +400,10 @@ describe('CodeWhisperer-basicCommands', function () {
         it('shows expected quick pick items when connected', async function () {
             sinon.stub(AuthUtil.instance, 'isConnectionExpired').returns(false)
             sinon.stub(AuthUtil.instance, 'isConnected').returns(true)
-            sinon.stub(CodeScansState.instance, 'isScansEnabled').returns(false)
+            await CodeScansState.instance.setScansEnabled(false)
             getTestWindow().onDidShowQuickPick((e) => {
                 e.assertContainsItems(
-                    createAutoSuggestions(false),
+                    createAutoSuggestions(true),
```
justinmk3, line 406: `true` is the default since https://github.com/aws/aws-toolkit-vscode/commit/1bec3c3eeba9b2d962db8abbd701bda28dae103e . This wasn't caught because the tests use stubs/mocks. This is an example of why stubs/mocks are fragile and don't represent the actual application code.

packages/core/src/awsService/s3/commands/uploadFile.ts
```diff
@@ -437,7 +437,7 @@ export async function promptUserForBucket(
         }
     })
 
-    const lastTouchedFolder = globals.context.globalState.get<SavedFolder | undefined>('aws.lastTouchedS3Folder')
+    const lastTouchedFolder = globals.globalState.tryGet<SavedFolder>('aws.lastTouchedS3Folder', Object)
```
justinmk3, line 440: `tryGet` checks that the stored value is actually an object, unlike vscode's `get()`. This avoids bugs like https://github.com/aws/aws-toolkit-vscode/pull/3060#discussion_r1050240283

packages/core/src/codewhisperer/ui/codeWhispererNodes.ts
```diff
@@ -29,21 +29,21 @@ import { submitFeedback } from '../../feedback/vue/submitFeedback'
...
     const iconResume = getIcon('vscode-debug-start')
     const labelPause = localize('AWS.codewhisperer.pauseCodeWhispererNode.label', 'Pause Auto-Suggestions')
     const iconPause = getIcon('vscode-debug-pause')
 
     return {
         data: 'autoSuggestions',
-        label: pause ? codicon`${iconPause} ${labelPause}` : codicon`${iconResume} ${labelResume}`,
-        description: pause ? 'Currently RUNNING' : 'Currently PAUSED',
+        label: running ? codicon`${iconPause} ${labelPause}` : codicon`${iconResume} ${labelResume}`,
+        description: running ? 'Currently RUNNING' : 'Currently PAUSED',
```
justinmk3, line 41: It looks like the logic was "swapped", i.e. it treated `pause` as if it were `running`. So I renamed the param to `running`. @hayemaxi

leigaol: what does `running` mean? Why when running is true, the label is pause?

justinmk3: It's the same as the previously-existing `pause` parameter, which controls what this function returns.

> what does running mean? Why when running is true, the label is pause?

)

- The "play" button shows when it's paused. 
- The "pause" button shows when it's running.
- 
![image](https://github.com/user-attachments/assets/1c4441aa-95f6-4051-a306-fe416ff8fc2f

ctlai95 (approved):

PR #5313, opened 2024-07-12
telemetry: upstream metrics to aws-toolkit-common

justinmk3:
depends on https://github.com/aws/aws-toolkit-common/pull/785

## License


packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -310,7 +310,7 @@ export class GumbyController {
     // prompt user to pick project and specify source JDK version
     private async initiateTransformationOnProject(message: any) {
         const authType = await getAuthType()
-        telemetry.codeTransform_jobIsStartedFromChatPrompt.emit({
+        telemetry.codeTransform_jobStart.emit({
```
justinmk3, line 313: @sunwoo-ms @dhasani23 this is from https://github.com/aws/aws-toolkit-common/pull/770

Are we ok to start using this in the client?

PR #5314, opened 2024-07-12
[WIP] DocumentDB Initial changes

mvcninja:
## Problem

- Users should be able to see the DocumentDB in the Explorer tab, it should list all the clusters available.

## Solution

- Able to create and delete clusters and instances

## License


packages/core/src/test/docdb/utils.test.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    it('Validates instance name starts with a lowercase letter', function () {
+        const message = validateInstanceName('404')
+        assert.strictEqual(message, 'Instance name must start with a lowercase letter')
+    })
+
+    it('Validates instance name does not contain uppercase characters', function () {
+        const message = validateInstanceName('abcDEF')
+        assert.strictEqual(message, 'Instance name must only contain lowercase letters, numbers, and hyphens')
+    })
```
justinmk3, line 122: there is no reason for these tests to have separate it() cases, especially when the assert() has an explanatory message.

please look for ways to be less verbose in all parts of this PR.

packages/core/src/docdb/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+ */
+
+import { localize } from '../shared/utilities/vsCodeUtils'
+
+/**
+ * Validates a cluster name for the CreateCluster API.
+ * @see https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/docdb/command/CreateDBClusterCommand/
+ * @returns undefined if the name passes validation. Otherwise, an error message is returned.
+ */
+export function validateClusterName(name: string): string | undefined {
```
justinmk3, line 13: Does the service not return validation errors that we can simply pass along?

packages/core/src/docdb/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+    const year = date.getFullYear()
+    const month = (date.getMonth() + 1).toString().padStart(2, '0')
+    const day = date.getDate().toString().padStart(2, '0')
+    return `${year}-${month}-${day}`
+}
+
+/**
+ * Format for rendering the time as HH:mm
+ */
+export function formatTime(date: Date = new Date()): string {
```
justinmk3, line 145: please look at https://github.com/aws/aws-toolkit-vscode/blob/56b1c4b785dfec0e4949776f39fa104e2fbca4aa/packages/core/src/shared/utilities/textUtilities.ts#L340-L356

and see if you can use those instead. if not, why not?

packages/core/src/shared/clients/docdbClient.ts
```diff
@@ -0,0 +1,278 @@
...
+const DocDBEngine = 'docdb'
+
+function isElasticCluster(clusterId: string): boolean {
+    return clusterId?.includes(':docdb-elastic:')
+}
+
+export const DBStorageType = { Standard: 'standard', IOpt1: 'iopt1' } as const
+
+/**
+ * <p>A list of Amazon DocumentDB clusters.</p>
```
justinmk3, line 24: where did the html come from?

packages/core/src/shared/clients/docdbClient.ts
```diff
@@ -0,0 +1,278 @@
...
+                Filters: [{ Name: 'engine', Values: [DocDBEngine] }],
+                DBClusterIdentifier: clusterId,
+            }
+            const command = new DocDB.DescribeDBClustersCommand(input)
+            const response = await client.send(command)
+            return response.DBClusters ?? []
+        } catch (e) {
+            throw ToolkitError.chain(e, 'Failed to get DocumentDB clusters')
+        } finally {
+            client.destroy()
```
justinmk3, line 114: most of these client functions seem similar. is there a way to deduplicate them

packages/core/src/docdb/utils.ts
mvcninja: It does, but we want to be proactive where possible especially as this is part of a multi-step wizard.

packages/core/src/docdb/utils.ts
mvcninja: I looked at those, but the format doesn't meet the requirements for this scenario.

packages/core/src/shared/clients/docdbClient.ts
mvcninja: Removed

PR #5316, opened 2024-07-13
fix(amazonq): always show latest build logs

dhasani23:
## Problem

When users download the pre-build error logs, then run another transformation without clicking the "Start a new transformation" button, if another pre-build error occurs, the **old** logs will show up. This could happen if they close the Q Chat tab and re-run a transformation, or if they close the IDE between job runs. If they do click the "Start a new transformation" button, everything works fine, but we need to handle these other situations.

## Solution

Use `makeTemporaryToolkitFolder()` which creates a new folder with a random name, so that the logs are downloaded there, meaning we don't have to deal with deleting the old logs that would otherwise be in the same folder as the new logs.

## License

### DIFF
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -270,6 +270,7 @@ export class GumbyController {
         const typedAction = MessengerUtils.stringToEnumValue(ButtonActions, message.action as any)
         switch (typedAction) {
             case ButtonActions.CONFIRM_TRANSFORMATION_FORM:
+                await resetDebugArtifacts()
                 await this.initiateTransformationOnProject(message)
                 break
             case ButtonActions.CANCEL_TRANSFORMATION_FORM:
@@ -473,6 +474,7 @@ export class GumbyController {
                 undefined,
                 GumbyNamedMessages.JOB_FAILED_IN_PRE_BUILD
             )
+            await openBuildLogFile()
             this.messenger.sendViewBuildLog(message.tabID)
         }
     }
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -24,7 +24,6 @@ import {
     openHilPomFile,
     postTransformationJob,
     processTransformFormInput,
-    resetDebugArtifacts,
     startTransformByQ,
     stopTransformByQ,
     validateCanCompileProject,
@@ -270,7 +269,6 @@ export class GumbyController {
         const typedAction = MessengerUtils.stringToEnumValue(ButtonActions, message.action as any)
         switch (typedAction) {
             case ButtonActions.CONFIRM_TRANSFORMATION_FORM:
-                await resetDebugArtifacts()
                 await this.initiateTransformationOnProject(message)
                 break
             case ButtonActions.CANCEL_TRANSFORMATION_FORM:
@@ -287,7 +285,6 @@ export class GumbyController {
                 break
             case ButtonActions.CONFIRM_START_TRANSFORMATION_FLOW:
                 this.resetTransformationChatFlow()
-                await resetDebugArtifacts()
                 this.messenger.sendCommandMessage({ ...message, command: GumbyCommands.CLEAR_CHAT })
                 await this.transformInitiated(message)
                 break
--- a/packages/core/src/codewhisperer/commands/startTransformByQ.ts
+++ b/packages/core/src/codewhisperer/commands/startTransformByQ.ts
@@ -71,6 +71,7 @@ import DependencyVersions from '../../amazonqGumby/models/dependencies'
 import { dependencyNoAvailableVersions } from '../../amazonqGumby/models/constants'
 import { HumanInTheLoopManager } from '../service/transformByQ/humanInTheLoopManager'
 import { setContext } from '../../shared/vscode/setContext'
+import { makeTemporaryToolkitFolder } from '../../shared'
 
 function getFeedbackCommentData() {
     const jobId = transformByQState.getJobId()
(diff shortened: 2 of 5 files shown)


packages/core/src/test/codewhisperer/commands/transformByQ.test.ts
```diff
@@ -310,21 +310,4 @@ describe('resetDebugArtifacts', () => {
         assert.strictEqual(fs.existsSync(preBuildLogFilePath), false)
         assert.strictEqual(transformByQState.getPreBuildLogFilePath(), '')
     })
-
-    it('should not remove any directory if the pre-build log file path is not set', async () => {
```
dhasani23, line 314: See the unit case above — it covers the required case. These two test cases are N/A now due to my changes to `resetBuildArtifacts()`

packages/core/src/test/codewhisperer/commands/transformByQ.test.ts
```diff
@@ -300,31 +301,14 @@ describe('transformByQ', function () {
 
 describe('resetDebugArtifacts', () => {
     it('should remove the directory containing the pre-build log file if it exists', async () => {
-        const dirPath = await createTestWorkspaceFolder()
-        const preBuildLogFilePath = path.join(dirPath.uri.fsPath, 'DummyLog.log')
-        await toFile('', preBuildLogFilePath)
+        const dirPath = path.join(os.tmpdir(), 'q-transformation-build-logs')
```
dhasani23, line 304: Have to use `os.tmpdir()` since that's what `resetDebugArtifacts()` uses.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -532,7 +532,7 @@ export async function pollTransformationStatusUntilPlanReady(jobId: string) {
         const pathToLog = path.join(tmpDir, 'buildCommandOutput.log')
         transformByQState.setPreBuildLogFilePath(pathToLog)
 
-        if (fs.existsSync(pathToLog)) {
+        if (fs.existsSync(pathToLog) && !transformByQState.isCancelled()) {
```
dhasani23, line 535: If the user cancelled the job, don't bother trying to download anything.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -473,6 +474,7 @@ export class GumbyController {
                 undefined,
                 GumbyNamedMessages.JOB_FAILED_IN_PRE_BUILD
             )
+            await openBuildLogFile()
```
dhasani23, line 477: Added this so that the build logs open up automatically once downloaded; the `View Logs` button can be difficult to see.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -270,6 +270,7 @@ export class GumbyController {
         const typedAction = MessengerUtils.stringToEnumValue(ButtonActions, message.action as any)
         switch (typedAction) {
             case ButtonActions.CONFIRM_TRANSFORMATION_FORM:
+                await resetDebugArtifacts()
```
dhasani23, line 273: This was missing. The first time you run `/transform`, this case is executed — need to delete the log file if it exists because it's from a previous job run.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -736,11 +736,11 @@ export async function cleanupTransformationJob() {
 }
 
 export async function resetDebugArtifacts() {
-    const buildLogPath = transformByQState.getPreBuildLogFilePath()
-    if (buildLogPath && fs.existsSync(buildLogPath)) {
```
dhasani23, line 740: The `preBuildLogFilePath` was reset to an empty string each time after the "Start a new transformation" button is clicked — the first part of this condition was always evaluating to `False`, meaning the old job's error logs were never being deleted. All we need to check is if the directory exists.

tincheng (approved):

packages/core/src/test/codewhisperer/commands/transformByQ.test.ts
justinmk3: why not add an optional `tmpdir` arg to `resetDebugArtifacts` ? Then this becomes less fragile, and more explicit, thus easier to follow.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -736,11 +736,11 @@ export async function cleanupTransformationJob() {
...
-        fs.rmSync(path.dirname(transformByQState.getPreBuildLogFilePath()), { recursive: true, force: true })
+    const buildLogDir = path.join(os.tmpdir(), 'q-transformation-build-logs')
```
justinmk3, line 739: should this use `makeTemporaryToolkitFolder` ? `os.tmpdir()` can be problematic, as hinted in the comment in `tempDirPath`. https://github.com/aws/aws-toolkit-vscode/blob/56b1c4b785dfec0e4949776f39fa104e2fbca4aa/packages/core/src/shared/filesystemUtilities.ts#L15-L18

packages/core/src/test/codewhisperer/commands/transformByQ.test.ts
```diff
@@ -6,6 +6,7 @@
 import assert from 'assert'
 import * as vscode from 'vscode'
 import * as fs from 'fs-extra'
+import * as os from 'os'
 import * as sinon from 'sinon'
 import { makeTemporaryToolkitFolder } from '../../../shared/filesystemUtilities'
```
justinmk3, line 11: tests should use `createTestWorkspaceFolder`
```suggestion
import { createTestWorkspaceFolder } from '...'
```

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -527,14 +528,26 @@ export async function pollTransformationStatusUntilPlanReady(jobId: string) {
         // Since we don't yet have a good way of knowing what the error was,
         // we try to fetch any build failure artifacts that may exist so that we can optionally
         // show them to the user if they exist.
-        const tmpDir = path.join(os.tmpdir(), 'q-transformation-build-logs')
-        await downloadAndExtractResultArchive(jobId, undefined, tmpDir, 'Logs')
-        const pathToLog = path.join(tmpDir, 'buildCommandOutput.log')
-        transformByQState.setPreBuildLogFilePath(pathToLog)
+        let pathToLog = ''
+        try {
+            const tempToolkitFolder = await makeTemporaryToolkitFolder()
```
dhasani23, line 533: Here's a simpler solution to solve this problem: use `makeTemporaryToolkitFolder()` instead of `os.tmpdir()` so that a new folder with a random name is used to download the logs, which removes the requirement to find and delete the old logs.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
dhasani23: Update: don't need to reset / delete anything with latest implementation.

packages/core/src/test/codewhisperer/commands/transformByQ.test.ts
dhasani23: Update: removed `resetDebugArtifacts` and its associated unit test.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
justinmk3: > which removes the requirement to find and delete the old logs.

is that causing problems? ideally we should not leave things in /tmp, even though in theory the OS cleans it up (in practicae not really)

dhasani23: We were previously using `os.tmpdir()` which appears to stay constant, so we had to delete the old logs. When we downloaded the new logs, for some reason that ZIP was not overwriting the old ZIP, so the old logs were showing up. We need to be able to store the logs indefinitely after a transformation job so the user can click on them anytime until they run a new transformation.

dhasani23: Anytime `makeTemporaryToolkitFolder()` is used there are going to be things left there undeleted, unless I store the random names of the all of the temporary folders created for an arbitrary number of transformations, and clean them up myself in place of the OS.

justinmk3: > We need to be able to store the logs indefinitely after a transformation job so the user can click on them anytime until they run a new transformation.

I see. Usually I'd suggest a workspace-local hidden folder for this, that is often what users expect. But then the user will need to add that to .gitignore.

PR #5322, opened 2024-07-16
fix: Code coverage report for amazon q

jpinkney-aws:
## Problem
- Codecoverage for amazon q wasn't actually generating any lcov.info information since the c8rc.json lived in core and amazonq couldn't find it

## Solution
- Move c8rc.json to the root of the project
- Merge the resulting lcov.info files

## License

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -32,6 +32,17 @@ jobs:
               uses: coactions/setup-xvfb@v1
               with:
                   run: npm test
+            - name: Code coverage (Core)
+              env:
+                  # Unset NODE_OPTIONS because of https://github.com/codecov/uploader/issues/475
+                  NODE_OPTIONS: ''
+              if: ${{ github.repository == 'aws/aws-toolkit-vscode' && ( github.ref == 'master' || github.event_name == 'pull_request' ) }}
+              uses: codecov/codecov-action@v4
+              with:
+                  flags: macos-core-unittests
+                  verbose: true
+                  file: ./coverage/core/lcov.info
+                  token: ${{ secrets.CODECOV_TOKEN }}
             - name: Code coverage (Toolkit)
               env:
                   # Unset NODE_OPTIONS because of https://github.com/codecov/uploader/issues/475
@@ -41,7 +52,7 @@ jobs:
               with:
                   flags: macos-toolkit-unittests
                   verbose: true
-                  file: ./coverage/core/lcov.info
+                  file: ./coverage/toolkit/lcov.info
                   token: ${{ secrets.CODECOV_TOKEN }}
             - name: Code coverage (Amazon Q)
               env:
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -32,16 +32,27 @@ jobs:
               uses: coactions/setup-xvfb@v1
               with:
                   run: npm test
-            - name: Code coverage
+            - name: Toolkit code coverage
               env:
                   # Unset NODE_OPTIONS because of https://github.com/codecov/uploader/issues/475
                   NODE_OPTIONS: ''
               if: ${{ github.repository == 'aws/aws-toolkit-vscode' && ( github.ref == 'master' || github.event_name == 'pull_request' ) }}
               uses: codecov/codecov-action@v4
               with:
-                  flags: macos-unittests
+                  flags: macos-toolkit-unittests
                   verbose: true
-                  file: ./coverage/lcov.info
+                  file: ./coverage/core/lcov.info
+                  token: ${{ secrets.CODECOV_TOKEN }}
+            - name: Amazon Q code coverage
+              env:
+                  # Unset NODE_OPTIONS because of https://github.com/codecov/uploader/issues/475
+                  NODE_OPTIONS: ''
+              if: ${{ github.repository == 'aws/aws-toolkit-vscode' && ( github.ref == 'master' || github.event_name == 'pull_request' ) }}
+              uses: codecov/codecov-action@v4
+              with:
+                  flags: macos-amazonq-unittests
+                  verbose: true
+                  file: ./coverage/amazonq/lcov.info
                   token: ${{ secrets.CODECOV_TOKEN }}
             - name: Code coverage (CodeWhisperer)
               env:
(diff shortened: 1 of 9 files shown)


docs/arch_develop.md
```diff
@@ -293,22 +292,22 @@ Commands and events are defined on the backend via sub-classes of `VueWebview`.
     ```ts
     client
         .foo()
-        .then(response => console.log(response))
-        .catch(err => console.log(err))
+        .then((response) => console.log(response))
```
jpinkney-aws, line 295: format on save 🙃

package.json
```diff
@@ -57,6 +58,7 @@
         "eslint-plugin-security-node": "^1.1.4",
         "eslint-plugin-unicorn": "^54.0.0",
         "husky": "^9.0.7",
+        "lcov-result-merger": "^5.0.1",
```
justinmk3, line 61: This might not be needed, codecov should automatically merge the results as long as we upload all the different lcov.info files. https://about.codecov.io/product/feature/report-merging/#:~:text=Codecov%20will%20merge%20the%20reports,or%20across%20different%20CI%2FCDs.

jpinkney-aws: I was just hoping we could just do a single call rather than multiple and do everything with globs so we wouldn't have to revisit this if a new package was added. It looks like the result merge might be mangling the results and codecov doesn't seem to like that though

buildspec/linuxTests.yml
```diff
@@ -48,7 +48,8 @@ phases:
             - VCS_COMMIT_ID="${CODEBUILD_RESOLVED_SOURCE_VERSION}"
             - CI_BUILD_URL=$(echo $CODEBUILD_BUILD_URL | sed 's/#/%23/g') # Encode `#` in the URL because otherwise the url is clipped in the Codecov.io site
             - CI_BUILD_ID="${CODEBUILD_BUILD_ID}"
-            - test -n "${CODECOV_TOKEN}" && ./codecov --token=${CODECOV_TOKEN} --branch=${CODEBUILD_RESOLVED_SOURCE_VERSION} --repository=${CODEBUILD_SOURCE_REPO_URL} --file=./coverage/lcov.info
+            - test -n "${CODECOV_TOKEN}" && ./codecov --token=${CODECOV_TOKEN} --branch=${CODEBUILD_RESOLVED_SOURCE_VERSION} --repository=${CODEBUILD_SOURCE_REPO_URL} --file=./coverage/amazonq/lcov.info
```
justinmk3, line 51: `--file` accepts multiple files according to https://docs.codecov.com/docs/cli-options

although that might mean either:

    --file=./coverage/amazonq/lcov.info --file=./coverage/core/lcov.info --file=./coverage/toolkit/lcov.info

or:

    --file ./coverage/*/lcov.info

I'm guessing it's the long one :( 

(Note: `=` is usually optional for most CLI args parsers, that's why the short one is at least syntactically valid)

.github/workflows/node.js.yml
```diff
@@ -32,16 +32,27 @@ jobs:
               uses: coactions/setup-xvfb@v1
               with:
                   run: npm test
-            - name: Code coverage
+            - name: Toolkit code coverage
```
justinmk3, line 35: should we use the same naming pattern as the existing `Code coverage (CodeWhisperer)` item, or change that one? it's probably visually helpful to have a common prefix

```suggestion
            - name: Code coverage (Toolkit)
```

.github/workflows/node.js.yml
```diff
@@ -32,16 +32,27 @@ jobs:
...
+                  file: ./coverage/core/lcov.info
+                  token: ${{ secrets.CODECOV_TOKEN }}
+            - name: Amazon Q code coverage
+              env:
+                  # Unset NODE_OPTIONS because of https://github.com/codecov/uploader/issues/475
+                  NODE_OPTIONS: ''
+              if: ${{ github.repository == 'aws/aws-toolkit-vscode' && ( github.ref == 'master' || github.event_name == 'pull_request' ) }}
+              uses: codecov/codecov-action@v4
+              with:
+                  flags: macos-amazonq-unittests
```
justinmk3, line 53: i would guess that Q/CW wants this in the `codewhisperer` category so that all of their coverage is under one umbrella. I assume that's what `flags` does
```suggestion
                  flags: codewhisperer
```

buildspec/linuxE2ETests.yml
```diff
@@ -41,7 +41,7 @@ phases:
             - VCS_COMMIT_ID="${CODEBUILD_RESOLVED_SOURCE_VERSION}"
             - CI_BUILD_URL=$(echo $CODEBUILD_BUILD_URL | sed 's/#/%23/g')
             - CI_BUILD_ID="${CODEBUILD_BUILD_ID}"
-            - test -n "${CODECOV_TOKEN}" && ./codecov --token=${CODECOV_TOKEN} --branch=${CODEBUILD_RESOLVED_SOURCE_VERSION} --repository=${CODEBUILD_SOURCE_REPO_URL} --file=./coverage/lcov.info
+            - test -n "${CODECOV_TOKEN}" && ./codecov --token=${CODECOV_TOKEN} --branch=${CODEBUILD_RESOLVED_SOURCE_VERSION} --repository=${CODEBUILD_SOURCE_REPO_URL} --file=./coverage/core/lcov.info --file=./coverage/amazonq/lcov.info
```
justinmk3, line 44: should we add `--file=.../toolkit/...` in anticipation of future toolkit-only tests?

codecov.yml
```diff
@@ -18,13 +18,13 @@ coverage:
             codewhisperer:
                 target: 70%
                 paths:
-                    - src/codewhisperer/*
+                    - packages/core/src/codewhisperer/*
                 flags:
                     - 'codewhisperer'
             amazonqFeatureDev:
```
justinmk3, line 24: does this correspond to the `flags` thing?

.github/workflows/node.js.yml
jpinkney-aws: If I understand codecov correct right now its setup so that amazon q will report all the tests in `macos-amazonq-unittests`, but the `codewhisperer` flag only includes reports that are in `packages/core/src/codewhisperer/` so that you can filter down of a subset of the coverage thats only provided by the codewhisperer module. If we switched the flag to codewhisperer this codecoverage would only report the coverage of `packages/core/src/codewhisperer/`

codecov.yml
jpinkney-aws: Flags scope down coverage reports for different tests/features. You can use codecov projects to report test targets and include flags

justinmk3 (approved):

PR #5330, opened 2024-07-17
feat(amazonq): Enable workspace index for amzn users

leigaol:
## Problem

This should be merged after https://github.com/aws/aws-toolkit-vscode/pull/5328 is merged!

As required, by default enable workspace index for amzn users and always send relevantDocument. 

## License


packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -373,6 +373,14 @@ export class LspController {
                 // only download LSP for users who did not turn on this feature
                 // do not start LSP server
                 await LspController.instance.tryInstallLsp(context)
+                // enable local workspace index ONCE for Amzn users
+                if (
+                    AuthUtil.instance.isAmznUser() &&
+                    globals.globalState.tryGet('aws.amazonq.workspaceIndexToggleOn', Boolean) !== true
```
justinmk3, line 379: why not use the existing `workspaceIndex` setting? if the `workspaceIndex` setting is undefined, then do this "one-time" step. Then set `workspaceIndex` to true. Then the user can set it to false if they want.

leigaol: that setting has default value false. Let me double check if it is undefined

justinmk3: try `isSet` https://github.com/aws/aws-toolkit-vscode/blob/246548b94e9ea688263e92884dca54ee27daff9c/packages/core/src/shared/settings.ts#L160-L170

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -373,6 +373,10 @@ export class LspController {
                 // only download LSP for users who did not turn on this feature
                 // do not start LSP server
                 await LspController.instance.tryInstallLsp(context)
+                // enable local workspace index ONCE for Amzn users
```
justinmk3, line 376:
```suggestion
                // Enable local workspace index by default, for Amzn users.
```

packages/core/src/amazonq/lsp/lspController.ts
leigaol: `isSet` does not work. `isSet` always return false,

zixlin7 (approved):

jpinkney-aws (approved):

packages/core/src/codewhisperer/service/featureConfigProvider.ts
```diff
@@ -96,6 +104,17 @@ export class FeatureConfigProvider {
...
             }
+
+            const dataCollectionValue = this.featureConfigs.get(dataCollectionFeatureName)?.value.stringValue
+            if (dataCollectionValue === 'data-collection') {
+                this._isDataCollectionGroup = true
+                // Enable local workspace index by default, for Amzn users.
+                const isSet = globals.globalState.get<boolean>('aws.amazonq.workspaceIndexToggleOn') || false
+                if (!isSet) {
+                    await CodeWhispererSettings.instance.enableLocalIndex()
+                    globals.globalState.tryUpdate('aws.amazonq.workspaceIndexToggleOn', true)
```
justinmk3, line 115: why is this using globalState instead of a vscode setting?

PR #5332, opened 2024-07-17
fix: scrubNames() drops file extension

nkomonen-amazon:
## Problem:

The scrubNames() function removes PII from file paths in a given string. The issue was that in certain cases where the file path extension was composed of multiple extensions, no file extension would be used.

So instead of 'x.ext1.ext2' -> 'x.ext2', we would just get 'x'.

## Solution:

Preserve the last file extension.

Also, this indirectly fixed some of the unit tests that were returning unexpected results.

## License

### DIFF
--- a/packages/core/src/shared/errors.ts
+++ b/packages/core/src/shared/errors.ts
@@ -333,7 +333,7 @@ export function getTelemetryResult(error: unknown | undefined): Result {
  */
 export function scrubNames(s: string, username?: string) {
     let r = ''
-    const fileExtRe = /\.[^.]{1,4}$/
+    const fileExtRe = /\.[^.\/]{1,4}$/
     const slashdot = /^[~.]*[\/\\]*/
 
     /** Allowlisted filepath segments. */
@@ -375,7 +375,6 @@ export function scrubNames(s: string, username?: string) {
         let scrubbed = ''
         // Get the frontmatter ("/", "../", "~/", or "./").
         const start = word.trimStart().match(slashdot)?.[0] ?? ''
-        const fileExt = word.trimEnd().match(fileExtRe)?.[0] ?? ''
         pathSegments[0] = pathSegments[0].trimStart().replace(slashdot, '')
         for (const seg of pathSegments) {
             if (driveLetterRegex.test(seg)) {
@@ -391,6 +390,8 @@ export function scrubNames(s: string, username?: string) {
             }
         }
 
+        // includes leading '.', eg: '.json'
+        const fileExt = pathSegments[pathSegments.length - 1].match(fileExtRe) ?? ''
         r += ` ${start.replace(/\\/g, '/')}${scrubbed.replace(/^[\/\\]+/, '')}${fileExt}`
     }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/test/shared/errors.test.ts
```diff
@@ -477,10 +477,10 @@ describe('util', function () {
     it('scrubNames()', async function () {
         const fakeUser = 'jdoe123'
         assert.deepStrictEqual(scrubNames('', fakeUser), '')
-        assert.deepStrictEqual(scrubNames('a ./ b', fakeUser), 'a ././ b') // TODO: fix this
-        assert.deepStrictEqual(scrubNames('a ../ b', fakeUser), 'a .././ b') // TODO: fix this
+        assert.deepStrictEqual(scrubNames('a ./ b', fakeUser), 'a ./ b')
+        assert.deepStrictEqual(scrubNames('a ../ b', fakeUser), 'a ../ b')
         assert.deepStrictEqual(scrubNames('a /.. b', fakeUser), 'a /.. b') // TODO: fix this
```
justinmk3, line 482:
```suggestion
        assert.deepStrictEqual(scrubNames('a /.. b', fakeUser), 'a /.. b')
```

packages/core/src/shared/errors.ts
```diff
@@ -391,7 +390,9 @@ export function scrubNames(s: string, username?: string) {
             }
         }
 
-        r += ` ${start.replace(/\\/g, '/')}${scrubbed.replace(/^[\/\\]+/, '')}${fileExt}`
+        // includes leading '.', eg: '.json'
+        const fileExt = path.extname(pathSegments[pathSegments.length - 1]) ?? ''
+        r += ` ${start.replace(/\\/g, '/')}${scrubbed.replace(/^[\/\\]+/, '')}${fileExt.length < 6 ? fileExt : ''}`
```
justinmk3, line 395: do we have a test case for the `length>=6` case?

packages/core/src/shared/errors.ts
```diff
@@ -333,7 +334,6 @@ export function getTelemetryResult(error: unknown | undefined): Result {
  */
 export function scrubNames(s: string, username?: string) {
     let r = ''
-    const fileExtRe = /\.[^.]{1,4}$/
```
justinmk3, line 336: probably doesn't matter, but would keeping this allow us to avoid the 'path' import?

justinmk3 (approved):
Thanks! This was definitely part of the original goal :)

packages/core/src/shared/errors.ts
nkomonen-amazon: Looked back in to the pattern as it was the root for the other failing tests. A small fix and now we can avoid path import.

packages/core/src/shared/errors.ts
nkomonen-amazon: Added in

packages/core/src/shared/errors.ts
```diff
@@ -391,7 +390,9 @@ export function scrubNames(s: string, username?: string) {
             }
         }
 
-        r += ` ${start.replace(/\\/g, '/')}${scrubbed.replace(/^[\/\\]+/, '')}${fileExt}`
+        // includes leading '.', eg: '.json'
+        const fileExt = pathSegments[pathSegments.length - 1].match(fileExtRe) ?? ''
+        r += ` ${start.replace(/\\/g, '/')}${scrubbed.replace(/^[\/\\]+/, '')}${fileExt.length < 6 ? fileExt : ''}`
```
justinmk3, line 395: since `fileExtRe` is defined with `{1,4}` , do we still need the `fileExt.length < 6` check?

PR #5334, opened 2024-07-17
test codewhisperer coverage changes

jpinkney-aws:
## Problem
Testing to see if we can trigger codewhisperer coverage changes

## License


packages/amazonq/.c8rc.json
```diff
@@ -2,6 +2,5 @@
     "report-dir": "../../coverage/amazonq",
     "reporter": ["lcov"],
     "all": true,
-    "include": ["src/**", "dist/src/**"],
```
jpinkney-aws, line 5: As far as I can tell/understand include isn't really necessary since we're using the "all" flag which considers all files in the cwd, --allowExternal which allows for src files to be outside of the amazonq directory (so it can see and report on the amazonq/codewhisperer test files), and exclude which specifically excludes certain files from the report

packages/amazonq/package.json
```diff
@@ -57,8 +57,8 @@
         "lint": "true",
         "watch": "npm run clean && npm run buildScripts && tsc -watch -p ./",
         "testCompile": "npm run clean && npm run buildScripts && npm run compileOnly",
-        "test": "npm run testCompile && c8 ts-node ../core/scripts/test/test.ts dist/test/unit/index.js ../core/dist/src/testFixtures/workspaceFolder",
-        "testE2E": "npm run testCompile && c8 ts-node ../core/scripts/test/testE2E.ts dist/test/e2e/index.js ../core/dist/src/testFixtures/workspaceFolder",
+        "test": "npm run testCompile && c8 --allowExternal ts-node ../core/scripts/test/test.ts dist/test/unit/index.js ../core/dist/src/testFixtures/workspaceFolder",
```
jpinkney-aws, line 60: the --allowExternal flag doesn't seem to be allowed in the c8rc.json, I guess it was only hooked up for the cli

packages/amazonq/.c8rc.json
```diff
@@ -2,6 +2,5 @@
...
-    "exclude": ["dist/test/**", "src/test/**"]
+    "exclude": ["test/**", "dist/test/**", "**/core/dist/test*/**", "**/core/src/test*/**", "**/node_modules/**"]
```
justinmk3, line 5: any reason not to do `**/test*/**` ?
```suggestion
    "exclude": ["**/test*/**", "**/node_modules/**"]
```

justinmk3 (approved):

PR #5335, opened 2024-07-17
fix: EPERM error during SSO token refresh

nkomonen-amazon:
## Problem:

During token refresh we were seeing unexpected EPERM errors in the `aws_refreshCredentials` metric, specifically in Windows.

## Solution:

A node filesystem rename() call can fail due to a racecondition on Windows. Looking in to the VS Code Filesystem implementation they account for this: https://github.com/microsoft/vscode/blob/09d5f4efc5089ce2fc5c8f6aeb51d728d7f4e758/src/vs/base/node/pfs.ts#L496
So we just need to use this rename instead.

Additional info here from graceful-fs which inspired the VSCode implementation: https://github.com/isaacs/node-graceful-fs/blob/234379906b7d2f4c9cfeb412d2516f42b0fb4953/polyfills.js#L87

## License


justinmk3 (approved):
Definitely think we can say something in the changelog

nkomonen-amazon:
@justinmk3 good point, added changelog items

PR #5336, opened 2024-07-17
refactor: node/web/common file structure

nkomonen-amazon:
## Problem

The AWS extensions can run in different environments like Node or the Web (in a browser). The problem is that some of
our code is not environment agnostic, eg: Node filesystem functions do no work in Web mode.

We need to structure a code in a way so that it is easy for developers to know if they can safely use the module. I.e can I use this module in Web mode.

## Solution

- Document the new project structure to separate environment dependent code, [see new documentation by this PR](https://github.com/aws/aws-toolkit-vscode/blob/1bd0112f3dcc329e54e3fd13a286c3143f174b89/docs/arch_develop.md#web-node-common-shared-naming-conventions)
- We get rid of the `src/common/` folder, moving the existing files in to `src/shared/`
- Refactor our `src/extension.ts` files
  - Now it will look like `extension.ts` (common code), `extensionNode.ts` (node specific/vscode desktop), `extensionWeb.ts` (Web mode)
  - Each subproject in `packages/` has received this updated.

## License


Will-ShaoHua (approved):

packages/amazonq/src/extensionNode.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    return api
+}
+
+/**
+ * The code in this function is not common, implying it only works in Node.js and not web.
+ * The goal should be for this to not exist and that all code is "common". So if possible make
+ * the code compatible with web and move it to {@link activateAmazonQCommon}.
+ */
+async function activateAmazonQNonCommon(context: vscode.ExtensionContext) {
```
jpinkney-aws, line 33: Maybe this should be something like:
```suggestion
async function activateAmazonQNode(context: vscode.ExtensionContext) {
```
?

jpinkney-aws (approved):

docs/arch_develop.md
```diff
@@ -71,39 +71,35 @@ Some components of the core library depend on the `package.json`s of the extensi
     -   Does not restore, it is a superset of what exists in `packages/core` for `configuration.properties`.
     -   To develop for the Amazon Q extension: add all changes to `packages/amazonq/package.json`, EXCEPT for settings that are references by code in the core library, or settings that already exist in the core `package.json`
 
-## Shared vs Common names
+## `web`, `node`, `common`, `shared` naming conventions
```
justinmk3, line 74: Maybe not in this PR, but "common" should go away, is it needed still?

docs/arch_develop.md
```diff
@@ -71,39 +71,35 @@ Some components of the core library depend on the `package.json`s of the extensi
...
-
-> NOTE: Something that is "shared" does not mean it is "common", as it could be reused in different places but only work in Node.js for example.
+```
+src/
+├── myTopic/
+│   ├── commonCode.ts
+│   ├── node/
+│   │   └── nodeCode.ts
+│   └── web/
+│       └── webCode.ts
```
justinmk3, line 90: for cases where a `node/` folder is needed, we wouldn't add "node" in the file names.

```suggestion
│   ├── subtopic.ts
│   ├── node/
│   │   └── subtopic.ts
│   └── web/
│       └── subtopic.ts
```

packages/core/package.json
```diff
@@ -24,17 +24,16 @@
...
         ".": "./dist/src/extension.js",
+        "./node": "./dist/src/extensionNode.js",
         "./web": "./dist/src/extensionWeb.js",
         "./webShared": "./dist/src/extensionWebShared.js",
-        "./extensionCommon": "./dist/src/extensionCommon.js",
-        "./common": "./dist/src/common/index.js",
         "./amazonq": "./dist/src/amazonq/index.js",
         "./codewhisperer": "./dist/src/codewhisperer/index.js",
-        "./codewhispererCommon": "./dist/src/codewhisperer/indexCommon.js",
+        "./codewhisperer/node": "./dist/src/codewhisperer/indexNode.js",
```
justinmk3, line 36: can we avoid exposing `/node` cases in the library interface (i.e. this "exports" list)? It gives partner teams the wrong idea, and may start to proliferate. 

*Internally*, `node/` and `fooNode.ts` are ways to route to proper implementations. But the interface itself (this "exports" list) shouldn't need to expose that.

Since gumby doesn't support web, that just means gumby won't be called from `extensionWeb.ts`.

packages/core/src/amazonqGumby/telemetry/codeTransformTelemetryState.ts
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { randomUUID } from '../../common/crypto'
+import { randomUUID } from '../../shared/crypto'
```
justinmk3, line 6: Nice

justinmk3 (approved):
Nice progress.

docs/arch_develop.md
nkomonen-amazon: My intention was that the actual file name is up to the user. Will update with your changes as it is more clear

docs/arch_develop.md
nkomonen-amazon: "common" is the overarching name we use to refer to code that works in any evironment. So while we wont be explicitly using it in file/folder names, it will be used in comments and such

packages/core/package.json
nkomonen-amazon: This separation for node specific was made since indirectly importing gumby code broke web mode. We need to isolate gumby code to `codewhisperer/node` so that we can import `codewhisperer` in to web.

I'll revisit this in the future, openend ticket: IDE-14397

justinmk3 (approved):

PR #5339, opened 2024-07-18
fix(codewhisperer): respect telemetryEnabled field

jpinkney-aws:
## Problem
- When enterprise sso is in use the telemetryEnabled field isn't properly respected

## Solution
- Gate codewhisperer telemetry only on the telemetryEnabled field

## License

### DIFF
--- a/packages/core/src/codewhisperer/client/codewhisperer.ts
+++ b/packages/core/src/codewhisperer/client/codewhisperer.ts
@@ -260,7 +260,7 @@ export class DefaultCodeWhispererClient {
                 ideVersion: extensionVersion,
             },
         }
-        if (!AuthUtil.instance.isValidEnterpriseSsoInUse() && !globals.telemetry.telemetryEnabled) {
+        if (!globals.telemetry.telemetryEnabled) {
             return
         }
         const response = await (await this.createUserSdkClient()).sendTelemetryEvent(requestWithCommonFields).promise()


packages/core/src/codewhisperer/client/codewhisperer.ts
```diff
@@ -260,7 +260,7 @@ export class DefaultCodeWhispererClient {
                 ideVersion: extensionVersion,
             },
         }
-        if (!AuthUtil.instance.isValidEnterpriseSsoInUse() && !globals.telemetry.telemetryEnabled) {
+        if (!globals.telemetry.telemetryEnabled) {
```
justinmk3, line 263: IIRC, this was like this because service telemetry (as opposed to Toolkit telemetry, which isn't emitted by this function) is not considered optional. This function (presumably) is invoked only if/when the user is actually using CW API.

https://github.com/aws/aws-toolkit-vscode/commit/35f1c4ca1c99c7d287dc4db9f5b9ad5f1079baa0

https://github.com/aws/aws-toolkit-vscode/commit/e583d501da44fe58732455574888c10fd3d85660

@andrewyuq @leigaol

zixlin7: Yes the service telemetry is required for all enterprise users so admin have visibility on metrics in productivity dashboard in console. Please don't change this.

PR #5340, opened 2024-07-18
chore(settings): remove core/package.json dependency

hayemaxi:
Settings are instead auto-generated from each extension package.json. This allows us to keep compile-time checks.
Follow up work:
- Settings validation (core settings shared between extensions)
- Update docs (requires removing syncing from packages completely)

## License


nkomonen-amazon (approved):
Nice idea

packages/core/src/shared/settings-amazonq.gen.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+/**
+ * This file was autogenerated by generateSettings.ts. Do NOT modify by hand.
+ * To update this file, update the settings in the extension package.json
+ * and run `npm run generateSettings` for that extension.
+ */
+
+export const amazonqSettings = {
+    "amazonQ.telemetry": {},
```
justinmk3, line 13: nit: should we include these specific .gen.ts files in our linter/prettier?

packages/core/src/shared/settings.ts
```diff
@@ -617,7 +618,7 @@ export function fromExtensionManifest<T extends TypeDescriptor & Partial<Section
  * TODO: Settings should be defined in individual extensions, and passed to the
  * core lib as necessary.
```
justinmk3, line 619: is the TODO still relevant or feasible, given our new strategy?

justinmk3 (approved):

scripts/generateSettings.ts
```diff
@@ -0,0 +1,56 @@
...
+            if (v.properties !== undefined) {
+                for (const inner of Object.keys(v.properties)) {
+                    settings[k][inner] = {}
+                }
+            }
+        }
+    )
+
+    const contents = `/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jpinkney-aws, line 38: I guess this is another place where we could re-use that shared header

jpinkney-aws (approved):

packages/core/src/shared/settings.ts
hayemaxi: The work required to refactor for this method far outweighs the benefit we would get from it. I'd say its no longer relevant to anything other than a design decision for a new repo or library.

scripts/generateSettings.ts
hayemaxi: will add in a follow-up

packages/core/src/shared/settings-amazonq.gen.ts
hayemaxi: no, there would be some extra effort to achieve proper formatting on gen/commit (albeit quite small) but I don't see any value here because its not meant for humans

PR #5341, opened 2024-07-18
chore(icons): generate icons in subprojects

hayemaxi:
- Icon generation still happens in core for compile time type checks.
  - Each extension has the same icons available anyways.
- Validate that the icon sections in subprojects match whats in core/.
  - Prevents hand modifying subproject icon sections which will bypass type checks.

## License


nkomonen-amazon:
So from my understanding we are now permanently adding the icons to `toolkit` and `amazonq` package.json as part of `npm run buildScripts`? We now don't need to copy the icons from `core/package.json` since they will additionally be generated in the subproject?

nkomonen-amazon (approved):

hayemaxi:
> So from my understanding we are now permanently adding the icons to `toolkit` and `amazonq` package.json as part of `npm run buildScripts`? We now don't need to copy the icons from `core/package.json` since they will additionally be generated in the subproject?

The icons are still copied via `scripts/copyFiles.ts`. This just enables creating `contributes.icons` from those copied icons in the local subprojects without having to sync that field from `core/`. Any icons added to `core/resources/` will be enabled for each subproject automatically.

justinmk3 (approved):

jpinkney-aws (approved):

PR #5342, opened 2024-07-18
chore: remove debuggers dependency on core/package.json

hayemaxi:
Generate debuggers definitions file from toolkit. We will now store this in the repo.

There is a lot of code that relies on the definitions, which is why it is is still generated into core. Future work: drop this code down into packages/toolkit

## License


packages/toolkit/scripts/build/generateConfigurationAttributes.ts
```diff
@@ -49,6 +38,31 @@ async function generateConfigurationAttributes(): Promise<void> {
...
             continue
         }
+
+        if (!fs.existsSync(path.dirname(debuggerConfig.outputFile))) {
+            throw new Error(`Output folder of ${debuggerConfig.outputFile} does not exist.`)
+        }
+
+        const header = `
+/*!
+* Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
nkomonen-amazon, line 48: Should we make this copyright blob a global constant? We'll probably create other generated files in the future. Maybe in a `scripts/constants.ts`

hayemaxi: good idea 👍

justinmk3 (approved):
Great choices.

packages/toolkit/scripts/build/generateConfigurationAttributes.ts
hayemaxi: will create in a follow-up

PR #5356, opened 2024-07-22
Refactor diff functionality to make it more usable

jpinkney-aws:
## Problem
- A lot of the diffing functionality is hidden in feature dev

## Solution
- Pull it out into amazon q commons

## License


packages/core/src/amazonq/commons/diff.ts
justinmk3, line 1: off-topic / future: `commons/` is a new name that doesn't have any other counterpart in the code base. It probably should be renamed to `util/` or `shared/`.

justinmk3 (approved):
great idea

packages/core/src/amazonq/commons/diff.ts
justinmk3: Also, the test is in `common/` (singular)

PR #5361, opened 2024-07-22
bug: premature SSO expiration due to failed file `rename()`

nkomonen-amazon:
## Background

When writing content to our SSO cache file we noticed the json content was malformed.
As a solution we did an "atomic" write where we wrote the contents to a temp
file then renamed to the actual file. This prevented the content from being malformed due to
by separate write.

## Problem:

The combination of writing a temp file, then renaming it started failing recently:
- Failures were indicated by our metric `aws_refreshCredentials` 
  - its properties are `reason: 'FileNotFound'` + `reasonDesc: 'Failed to save "x:/Users/x/.aws/sso/cache/x"'`
- The problem was introduce in #5335 version of Amazon Q 1.16.0/AWS Toolkit 3.15.0
  - We were using Node.js `rename()` and then switched to the VS Code Filesystem `rename()`
  - [This is the specific code change that started causing the issue](https://github.com/aws/aws-toolkit-vscode/pull/5335/files#diff-b6c2a56b25754285c889ce3f70925b9ec01d508a244128466193f4066bff7133R138), Node.js' `rename()` to the VS Code Filesystem `rename()`

A guess to why we are getting the failure:
- The underlying VS Code Filesystem `rename()` method first runs a `stat()` to check if the file exists. ([high level call](https://github.com/microsoft/vscode/blob/09d5f4efc5089ce2fc5c8f6aeb51d728d7f4e758/src/vs/platform/files/node/diskFileSystemProvider.ts#L673) which eventually does an [actual stat call which throws the `FileNotFound` error](https://github.com/microsoft/vscode/blob/09d5f4efc5089ce2fc5c8f6aeb51d728d7f4e758/src/vs/platform/files/node/diskFileSystemProvider.ts#L747)

## Solution:
- Update our FileSystem `rename()` method so that it attempts to validate that the file exists by retrying multiple times.
  - If this retry fails we will report the issue in a telemetry metric to maybe get more data seeing when the file starts to exist.
- Move the atomic write logic in to `FileSystem.writeFile()` itself. It is triggered as part of an argument to the function, we can maybe consider making making this the default if it makes sense for the performance tradeoff.

## License

### DIFF
--- a/packages/core/src/shared/fs/fs.ts
+++ b/packages/core/src/shared/fs/fs.ts
@@ -7,10 +7,20 @@ import os from 'os'
 import { promises as nodefs, constants as nodeConstants, WriteFileOptions } from 'fs'
 import { isCloud9 } from '../extensionUtilities'
 import _path from 'path'
-import { PermissionsError, PermissionsTriplet, ToolkitError, isFileNotFoundError, isPermissionsError } from '../errors'
+import {
+    PermissionsError,
+    PermissionsTriplet,
+    ToolkitError,
+    isFileNotFoundError,
+    isPermissionsError,
+    scrubNames,
+} from '../errors'
 import globals from '../extensionGlobals'
 import { isWin } from '../vscode/env'
 import { resolvePath } from '../utilities/pathUtils'
+import crypto from 'crypto'
+import { waitUntil } from '../utilities/timeoutUtils'
+import { telemetry } from '../telemetry/telemetry'
 
 const vfs = vscode.workspace.fs
 type Uri = vscode.Uri
@@ -182,24 +192,57 @@ export class FileSystem {
      *
      * @param path File location
      * @param data File content
-     * @param opt File permissions/flags. Only works in a non-web (nodejs) context. If provided,
+     * @param opts File permissions/flags. Only works in a non-web (nodejs) context. If provided,
      * nodejs filesystem interface is used instead of routing through vscode VFS.
+     * @param opts.atomic If true, ensures content is not corrupted from a concurrent write.
+     *                    This is not truly atomic as an async write can override the final result,
+     *                    but if the content is structured like JSON it will still be parseable.
+     *
+     *                    Eg: we were seeing a JSON file with an unexpected extra '}' and when parsing it
+     *                    failed. We suspected a race condition from separate write, and all we wanted was
+     *                    a parseable JSON.
+     *
+     *                    This optional has an uncalcuated performance impact as there is an additional
+     *                    rename() operation.
      */
-    async writeFile(path: Uri | string, data: string | Uint8Array, opt?: WriteFileOptions): Promise<void> {
+    async writeFile(
+        path: Uri | string,
+        data: string | Uint8Array,
+        opts?: WriteFileOptions & { atomic?: boolean }
+    ): Promise<void> {
         const uri = this.#toUri(path)
         const errHandler = createPermissionsErrorHandler(this.isWeb, uri, '*w*')
         const content = this.#toBytes(data)
-        // - Special case: if `opt` is given, use nodejs directly. This isn't ideal, but is the only
-        //   way (unless you know better) we can let callers specify permissions.
-        // - Cloud9 vscode.workspace.writeFile has limited functionality, e.g. cannot write outside
-        //   of open workspace.
-        const useNodejs = (opt && !this.isWeb) || isCloud9()
 
-        if (useNodejs) {
-            return nodefs.writeFile(uri.fsPath, content, opt).catch(errHandler)
+        if (this.isWeb) {
+            return vfs.writeFile(uri, content).then(undefined, errHandler)
         }
 
-        return vfs.writeFile(uri, content).then(undefined, errHandler)
+        // Node writeFile is the only way to set `writeOpts`, such as the `mode`, on a file .
+        // When not in web we will use Node's writeFile() for all other scenarios.
+        // It also has better error messages than VS Code's writeFile().
+        let write = (u: Uri) => nodefs.writeFile(u.fsPath, content, opts).then(undefined, errHandler)
+
+        if (isCloud9()) {
+            // In Cloud9 vscode.workspace.writeFile has limited functionality, e.g. cannot write outside
+            // of open workspace.
+            //
+            // This is future proofing in the scenario we switch the initial implementation of `write()`
+            // to something else, C9 will still use node fs.
+            write = (u: Uri) => nodefs.writeFile(u.fsPath, content, opts).then(undefined, errHandler)
+        }
+
+        // Node writeFile does NOT create parent folders by default, unlike VS Code FS writeFile()
+        await fs.mkdir(_path.dirname(uri.fsPath))
+
+        if (opts?.atomic) {
+            const tempFile = this.#toUri(`${uri.fsPath}.${crypto.randomBytes(8).toString('hex')}.tmp`)
+            await write(tempFile)
+            await fs.rename(tempFile, uri)
+            return
+        } else {
+            await write(uri)
+        }
     }
 
     async rename(oldPath: vscode.Uri | string, newPath: vscode.Uri | string) {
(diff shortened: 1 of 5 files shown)


packages/core/src/shared/fs/fs.ts
```diff
@@ -211,6 +222,37 @@ export class FileSystem {
...
+        const isExists = await waitUntil(async () => {
+            const result = await fs.exists(oldUri)
+            attempts += 1
+            return result
+        }, renameTimeoutOpts)
+        // TODO: Move the `aws_fileSystem` or some variation of this metric in to common telemetry
+        if (!isExists) {
+            // source path never existed after multiple attempts.
+            // Either the file never existed, or had we waited longer it would have. We won't know.
+            telemetry.aws_fileSystem.emit({
```
justinmk3, line 240: will this event be noisy? can we at least leave a comment like:

    // TODO: deduplicate

or add a field or something so that we can find these metrics later, when deduplication is generalized

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -273,6 +273,16 @@
         }
     ],
     "metrics": [
+        {
+            "name": "aws_fileSystem",
```
justinmk3, line 277: in https://github.com/aws/aws-toolkit-common/pull/793 I proposed a new `ide_` prefix. maybe it fits better? filesystem operations are not really related to "aws"

packages/core/src/shared/utilities/cacheUtils.ts
```diff
@@ -131,17 +130,16 @@ export function createDiskCache<T, K>(
                 await fs.mkdir(dirname(target))
                 if (isWeb()) {
                     // There is no web-compatible rename() method. So do a regular write.
-                    await fs.writeFile(target, JSON.stringify(data), { mode: 0o600 })
```
justinmk3, line 134: why are we removing the `mode:` feature?

packages/core/src/shared/fs/fs.ts
nkomonen-amazon: This should not be noisy as we only emit on the edge case where the source file does not exist or took time to appear. In the average case where it runs as normal, no telemetry is executed

packages/core/src/shared/utilities/cacheUtils.ts
nkomonen-amazon: Explained in the second commit of this PR

packages/core/src/shared/utilities/cacheUtils.ts
```diff
@@ -131,17 +130,16 @@ export function createDiskCache<T, K>(
...
                     // write, though there can still be race conditions with which version remains after overwrites.
-                    const tempFile = `${target}.tmp-${crypto.randomBytes(4).toString('hex')}`
-                    await fs.writeFile(tempFile, JSON.stringify(data), { mode: 0o600 })
-                    await fs.rename(tempFile, target)
+                    await fs.writeFile(target, JSON.stringify(data), { atomic: true })
                 }
             } catch (error) {
-                throw ToolkitError.chain(error, `Failed to save "${target}"`, {
+                const message = error instanceof Error ? error.message : 'NoMessage'
+                throw ToolkitError.chain(error, `Failed to save "${target}" with message: ${message}`, {
```
justinmk3, line 142: message is contained in the `cause` already, which is done by chain(). why do we need to repeat it here?

if nested messages should be shown to the user, then we should improve our error handling rather than duplicating messages in each part of the chain.

nkomonen-amazon: In telemetry we do not show the nested error in `reasonDesc`. So in this example before my change we only saw: `Failed to save "${target}"`, the message from the wrapped error was lost.

packages/core/src/shared/fs/fs.ts
```diff
@@ -182,24 +185,32 @@ export class FileSystem {
...
         const errHandler = createPermissionsErrorHandler(this.isWeb, uri, '*w*')
         const content = this.#toBytes(data)
-        // - Special case: if `opt` is given, use nodejs directly. This isn't ideal, but is the only
-        //   way (unless you know better) we can let callers specify permissions.
-        // - Cloud9 vscode.workspace.writeFile has limited functionality, e.g. cannot write outside
-        //   of open workspace.
-        const useNodejs = (opt && !this.isWeb) || isCloud9()
-
-        if (useNodejs) {
-            return nodefs.writeFile(uri.fsPath, content, opt).catch(errHandler)
```
justinmk3, line 199: for the non-atomic case can we still use nodefs.writeFile? that gives better error messages.

packages/core/src/shared/fs/fs.ts
```diff
@@ -182,24 +185,32 @@ export class FileSystem {
...
+
+        if (isCloud9()) {
+            return nodefs.writeFile(uri.fsPath, content).catch(errHandler)
         }
 
-        return vfs.writeFile(uri, content).then(undefined, errHandler)
+        if (opts.atomic) {
+            const tempFile = this.#toUri(`${uri.fsPath}.${crypto.randomBytes(4).toString('hex')}.tmp`)
```
justinmk3, line 208: probably worth using more than 4 bytes

packages/core/src/shared/fs/fs.ts
nkomonen-amazon: After running in to the Windows handling fix that VS Code FS does for `rename()` I figured we are better off sticking with them as there may be more edge cases that `nodefs` may not handle.

packages/core/src/shared/fs/fs.ts
```diff
@@ -211,6 +222,37 @@ export class FileSystem {
...
+                reason: 'SourceNotExists',
+                reasonDesc: `After ${renameTimeoutOpts.timeout}ms the source path did not exist: ${oldUri.fsPath}`,
+            })
+        } else if (attempts > 1) {
+            // Indicates that rename() would have failed if we had not waited for it to exist.
+            telemetry.aws_fileSystem.emit({
+                result: 'Succeeded',
+                action: 'rename',
+                reason: 'RenameRaceCondition',
+                reasonDesc: `After ${attempts} attempts the source path existed: ${oldUri.fsPath}`,
```
justinmk3, line 252: there is an `attempts` field: https://github.com/aws/aws-toolkit-common/blob/2ca7102873d5c9a1ad166cbcc6fa127d22b461c7/telemetry/definitions/commonDefinitions.json#L101

should we use fields instead of using a human-readable message to store data

packages/core/src/shared/fs/fs.ts
```diff
@@ -211,6 +222,37 @@ export class FileSystem {
...
+        }, renameTimeoutOpts)
+        // TODO: Move the `aws_fileSystem` or some variation of this metric in to common telemetry
+        if (!isExists) {
+            // source path never existed after multiple attempts.
+            // Either the file never existed, or had we waited longer it would have. We won't know.
+            telemetry.aws_fileSystem.emit({
+                result: 'Failed',
+                action: 'rename',
+                reason: 'SourceNotExists',
+                reasonDesc: `After ${renameTimeoutOpts.timeout}ms the source path did not exist: ${oldUri.fsPath}`,
```
justinmk3, line 244: does `oldUri.fsPath` need to be scrubbed https://github.com/aws/aws-toolkit-vscode/blob/5dfc8c59a1ea268f0bedb02f0fae5cf8356e01dd/packages/core/src/shared/errors.ts#L334

packages/core/src/shared/telemetry/vscodeTelemetry.json
nkomonen-amazon: This metric is temporary as I'm not 100% sure this PR fixes the underlying issue and we need something quick.
I'll update to `ide_`, but we should probably discuss a better overall metric like the one you suggested

packages/core/src/shared/fs/fs.ts
```diff
@@ -182,24 +185,32 @@ export class FileSystem {
      *
      * @param path File location
      * @param data File content
-     * @param opt File permissions/flags. Only works in a non-web (nodejs) context. If provided,
-     * nodejs filesystem interface is used instead of routing through vscode VFS.
+     * @param opts.atomic If true, ensures content is not corrupted from a concurrent write.
+     *                    This is not truly atomic as an async write can override the final result,
+     *                    but if the content is structured like JSON it will still be parseable.
```
justinmk3, line 190: can this be clarified? why would it matter what the contents (JSON or otherwise) are? either it can be overwritten or not. and if it's "atomic", then sure, the "peer" write will fully overwrite the file.

packages/core/src/shared/utilities/cacheUtils.ts
justinmk3: > We do not need these opts since there is only 1 place that uses them, and after re-evaluating they are not needed.

Why aren't they needed? When we create ~/.aws files they need to have restricted permissions. If we aren't doing that, it's a bug.

packages/core/src/shared/fs/fs.ts
justinmk3: that might be true for now, but still should be a TODO since `aws_fileSystem` presumably could track other fs ops in the future

packages/core/src/shared/fs/fs.ts
justinmk3: we are already (in this PR) re-implementing some vscode's fs operations, so it's not clear at all that we can depend on it fully.

packages/core/src/shared/fs/fs.ts
nkomonen-amazon: The original reason for this "atomic" code was that our SSO cache file (JSON) was unparseable since it was malformed, eg there was an extra `}`. We guesses the case was that there were competing writes. We are fine if an overwrite happens, but wanted to avoid malformed content.

Does this clarify things?

packages/core/src/shared/fs/fs.ts
nkomonen-amazon: Does it make sense for telemetry to process any `reasonDesc` through scrub names automatically?
We could search for `reasonDesc` [somewhere here](https://github.com/nkomonen-amazon/aws-toolkit-vscode/blob/8d88a04d1ebf4e14a7e87b94200d299278034a10/packages/core/src/shared/telemetry/spans.ts#L129) and process it through `scrubNames()` always.

But for this current code I'll just use scrubNames inline and we can revisit in a future PR

packages/core/src/shared/utilities/cacheUtils.ts
nkomonen-amazon: Opened a PR to address the underlying issue: https://github.com/aws/aws-toolkit-vscode/pull/5373

packages/core/src/shared/fs/fs.ts
nkomonen-amazon: Added a todo comment. We can probably have an underlying telemetry wrapper around all filesystem calls and then we report errors only. This will help uncover any unexpected failures.

packages/core/src/shared/utilities/cacheUtils.ts
```diff
@@ -131,14 +130,12 @@ export function createDiskCache<T, K>(
                 await fs.mkdir(dirname(target))
                 if (isWeb()) {
                     // There is no web-compatible rename() method. So do a regular write.
-                    await fs.writeFile(target, JSON.stringify(data), { mode: 0o600 })
+                    await fs.writeFile(target, JSON.stringify(data))
```
nkomonen-amazon, line 133: Since we are using the VS Code File system implementation which does not support setting mode/permissions, we do not need this anymore

jpinkney-aws (approved):

packages/core/src/shared/utilities/cacheUtils.ts
justinmk3: We do need it. See https://github.com/aws/aws-toolkit-vscode/pull/5361/files#r1688216497

This PR removed functionality. We need to be able to specify permissions where possible. That may require special-casing fs.writeFile internally, when nodejs is available.

justinmk3: Because this is the `isWeb` case. But not for the "nodejs" case.

Issue #5311, opened 2024-07-12
test/CI performance: vscode downloaded multiple times


justinmk3:
# Problem

Since we now run the tests in each `package/x/` separately, vscode is downloaded redundantly in our tests:

```
Running unit test suite...
Setting up VS Code test instance, version: stable
Downloading VS Code 1.91.1 from https://update.code.visualstudio.com/1.91.1/darwin-arm64/stable
Downloading VS Code (130067222B)
```

This adds time to each CI run.

# Proposal

vscode-test has a flag that allows pointing to a directory where it should be downloaded (or skipped if it already exists). Update our test runner script to specify a directory at the repo top level.

PR #5364, opened 2024-07-23
fix: VSCode is downloaded for every test package

jpinkney-aws:
## Problem
- In github ci vscode is downloaded for every test package (core/q)

## Solution
- Use AWS_TOOLKIT_TEST_CACHE_DIR and AWS_TOOLKIT_TEST_USER_DIR to cache the test and user directories
- By default use `../../.vscode-test` directory

Related issue: https://github.com/aws/aws-toolkit-vscode/issues/5311

## License

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -21,6 +21,8 @@ jobs:
         env:
             VSCODE_TEST_VERSION: ${{ matrix.vscode-version }}
             NODE_OPTIONS: '--max-old-space-size=8192'
+            AWS_TOOLKIT_TEST_CACHE_DIR: '/tmp/.vscode-test/'
+            AWS_TOOLKIT_TEST_USER_DIR: '/tmp/.vscode-test/user-data/'
         steps:
             - uses: actions/checkout@v4
             - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/node.js.yml
```diff
@@ -21,6 +21,8 @@ jobs:
         env:
             VSCODE_TEST_VERSION: ${{ matrix.vscode-version }}
             NODE_OPTIONS: '--max-old-space-size=8192'
+            AWS_TOOLKIT_TEST_CACHE_DIR: '/tmp/.vscode-test/'
```
jpinkney-aws, line 24: We _may_ be able to drop this environment variable I think and instead just use `../../.vscode-test`

jpinkney-aws: Ah this is why it was needed in codebuild: https://github.com/aws/aws-toolkit-vscode/blob/master/buildspec/linuxE2ETests.yml#L11

jpinkney-aws: The warning still exists in github ci as well (taken from the latest commits ci):
```
WARNING: IPC handle "/Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/amazonq/.vscode-test/user-data/1.91-main.sock" is longer than 103 chars, try a shorter --user-data-dir
```
so at the very least using this env variable disables the warning and any potential repercussions from it

justinmk3 (approved):

PR #5371, opened 2024-07-23
refactor(amazonqFeatureDev): add i18n support for messaging from package.nls.json

tverney:
## Problem

Currently /dev relies on hardcoded strings to provide user messaging. Those are all written in english and don't use wrapper to translate in different VS Code supported idiom languages.

## Solution

This PR reuses the package.nls.json which is the standard approach provided by the toolkit team. All our strings were updated to import from there.

## License


packages/core/src/amazonqFeatureDev/util/i18n-helper.ts
```diff
@@ -0,0 +1,17 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import translations from '../../../package.nls.json'
```
jpinkney-aws, line 6: this file feels like it can live in a more general place. Maybe core/src/shared?

packages/core/src/amazonqFeatureDev/errors.ts
```diff
@@ -159,7 +142,10 @@ export class ApiError extends ToolkitError {
     }
 }
 
-export const denyListedErrors: string[] = ['Deserialization error', 'Inaccessible host']
+export const denyListedErrors: string[] = [
+    i18n('AWS.amazonq.featureDev.error.deserialization'),
```
jpinkney-aws, line 146: Are these strings i18n'd on the backend?

tverney: Good question. Theoretically should have, but I would not rely on those unless we have some uniformity (or make sure they match both toolkit and service). I'll update to keep those static.

tverney: Good question. Theoretically should have, but I would not rely on those unless we have some uniformity (or make sure they match both toolkit and service). I'll update to keep those static.

packages/core/src/amazonqFeatureDev/util/i18n-helper.ts
tverney: I've reused the default file for Q  extension, but perhaps we can change it to something more sharable. I'll double check on that.

packages/core/src/amazonqFeatureDev/util/i18n-helper.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+import translations from '../../../src/shared/i18n/package.nls.json'
+import { localize } from '../../shared/utilities/vsCodeUtils'
+
+export const i18n = (code: string) => {
+    const translationsTyped = translations as { [key: string]: string }
+
+    if (translationsTyped[code]) {
+        return localize(code, translationsTyped[code])
```
chengoramazon, line 13: Why not just use localize(code,  "fallback")?

tverney: Because we will keep the same strings in the code and its easy just fallback to the default english file without to duplicate the same tag in a fallback. At least makes more sense to me.

chengoramazon: Have you tried `vscode.l10n.t()` ?

tverney: I agree that the new VS Code API have a more solid support using l10n. However, @chengoramazon the current Amazon Q implementation uses https://github.com/microsoft/vscode-nls, which is the same I'm reusing it. Use l10n now api would make we differ the current implementation and require we have another approach in all the other 175 files that its using nls.

chengoramazon: > [vscode-nls](https://github.com/microsoft/vscode-nls) ⚠️ This package is no longer receiving new features in favor of the new localization library, [vscode-l10n](https://github.com/microsoft/vscode-l10n). Please use that collection of libraries instead.

We don't need the fallback if we go with l10n, which would be proper.

tverney: Obviously, but you're suggesting we change the vscode-nls and upgrade to l10n right now in all packages?

jpinkney-aws: We would have to update our vscode minimum version if we wanted to do this: https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/package.json#L12. vscode-l10n only supports 1.73.0 and above I think

tverney: This PR suggests to add support for amazonqFeatureDev, not replace the library for all packages.

chengoramazon: Yea, otherwise, we need do it again once l10n is available for us.

packages/core/src/shared/i18n/package.nls.json
```diff
@@ -265,5 +265,51 @@
     "AWS.amazonq.learnMore": "Learn More About Amazon Q",
```
jpinkney-aws, line 265: Why is this file being moved? This file should remain at the root beside package.json

tverney: I see. I just changed because what I've saw from the previous conversation the suggestion was to include it in a shared folder? Anyway, I can add it back.

jpinkney-aws: Ah my bad, I just meant you could look into using your i18n utils into the shared folder and then keep the package.nls.json in the same place

tverney: Oh I see 😅 This makes sense. I'll include over there.

packages/core/src/amazonqFeatureDev/util/i18n-helper.ts
chengoramazon: I think we looked at latest l10n version. It is already in our current vscode version. https://github.com/aws/aws-toolkit-vscode/blob/master/package-lock.json#L17394

v0.0.13

jpinkney-aws:
Looks like theres a conflict now

tverney:
> Looks like theres a conflict now

Yep, I'll rebase and push the conflict fix.

packages/core/src/amazonqFeatureDev/util/i18n-helper.ts
tverney: This link sounds a 2nd level dependency. We're not using it directly.

tverney:
@jpinkney-aws Let me know if there is anything more required here to merge/approve this one.

PR #5373, opened 2024-07-23
telemetry: include chained error message in reasonDesc

nkomonen-amazon:
## Problem:

When we use the function getTelemetryReasonDesc() we would lose information about the underlying error in the case the error was a ToolkitError with an underlying chained error. Only the message of the ToolkitError was used.

So in telemetry we would be missing some information about the real cause of the error.

## Solution:

getTelemetryReasonDesc() will return a  message which consists of all messages in the chain. It recursively calls in to the next chained error of a ToolkitError and appends it to the final output.

This will give us some context about the higher level call as well as the root of the issue.

The final message could look something like:
"Message A::Message B::Message C"

## License

### DIFF
--- a/packages/core/src/shared/errors.ts
+++ b/packages/core/src/shared/errors.ts
@@ -243,9 +243,22 @@ export class ToolkitError extends Error implements ErrorInformation {
     }
 }
 
-export function getErrorMsg(err: Error | undefined): string | undefined {
+/**
+ * Derives an error message from the given error object.
+ * Depending on the Error, the property used to derive the message can vary.
+ *
+ * @param withCause If the error is a ToolkitError, the message of the cause will
+ *                  be appended and delimited by a '::'. Eg: msg1::causeMsg1:causeMsg2
+ */
+export function getErrorMsg(err: Error | undefined, withCause = false): string | undefined {
     if (err === undefined) {
         return undefined
+    }
+
+    if (withCause && err instanceof ToolkitError) {
+        // append the message of the cause, recursively
+        // output eg: "ToolkitError Message::Cause Message::Nested Cause Message::Nested Nested Cause Message"
+        return `${err.message}${err.cause ? '::' + getErrorMsg(err.cause, true) : ''}`
     }
 
     // Non-standard SDK fields added by the OIDC service, to conform to the OAuth spec
@@ -404,7 +417,7 @@ export function scrubNames(s: string, username?: string) {
  * @param err Error object, or message text
  */
 export function getTelemetryReasonDesc(err: unknown | undefined): string | undefined {
-    const m = typeof err === 'string' ? err : getErrorMsg(err as Error) ?? ''
+    const m = typeof err === 'string' ? err : getErrorMsg(err as Error, true) ?? ''
     const msg = scrubNames(m, _username)
 
     // Truncate to 200 chars.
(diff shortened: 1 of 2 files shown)


packages/core/src/shared/errors.ts
```diff
@@ -243,11 +243,24 @@ export class ToolkitError extends Error implements ErrorInformation {
...
+ */
+export function getErrorMsg(err: Error | undefined, withCause = false): string | undefined {
     if (err === undefined) {
         return undefined
     }
 
+    if (withCause && err instanceof ToolkitError) {
+        // append the message of the cause, recursively
+        // output eg: "ToolkitError Message::Cause Message::Nested Cause Message::Nested Nested Cause Message"
+        return `${err.message}${err.cause ? '::' + getErrorMsg(err.cause, true) : ''}`
```
justinmk3, line 261:
```suggestion
        return `${err.message}${err.cause ? ' / ' + getErrorMsg(err.cause, true) : ''}`
```

packages/core/src/shared/errors.ts
```diff
@@ -243,11 +243,24 @@ export class ToolkitError extends Error implements ErrorInformation {
...
+ *
+ * @param withCause If the error is a ToolkitError, the message of the cause will
+ *                  be appended and delimited by a '::'. Eg: msg1::causeMsg1:causeMsg2
+ */
+export function getErrorMsg(err: Error | undefined, withCause = false): string | undefined {
     if (err === undefined) {
         return undefined
     }
 
+    if (withCause && err instanceof ToolkitError) {
```
justinmk3, line 258: javascript errors can have `cause`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

I think our ES type definitions are just out of date (may change after we update our min vscode).

```suggestion
    if (withCause && (err as any).code) {
```

packages/core/src/shared/errors.ts
```diff
@@ -243,11 +243,24 @@ export class ToolkitError extends Error implements ErrorInformation {
     }
 }
 
-export function getErrorMsg(err: Error | undefined): string | undefined {
+/**
+ * Derives an error message from the given error object.
+ * Depending on the Error, the property used to derive the message can vary.
+ *
+ * @param withCause If the error is a ToolkitError, the message of the cause will
+ *                  be appended and delimited by a '::'. Eg: msg1::causeMsg1:causeMsg2
```
justinmk3, line 251:
```suggestion
 * @param withCause Append the message(s) from the cause chain, recursively.
```

packages/core/src/test/shared/errors.test.ts
```diff
@@ -451,14 +452,33 @@ describe('util', function () {
...
+        const awsErr = new TestAwsError('ValidationException', 'aws validation msg 1', new Date())
+        assert.deepStrictEqual(getErrorMsg(awsErr), 'aws validation msg 1')
+        ;(awsErr as any).error_description = ''
+        assert.deepStrictEqual(getErrorMsg(awsErr), 'aws validation msg 1')
+        ;(awsErr as any).error_description = {}
+        assert.deepStrictEqual(getErrorMsg(awsErr), 'aws validation msg 1')
+        ;(awsErr as any).error_description = 'aws error desc 1'
+        assert.deepStrictEqual(getErrorMsg(awsErr), 'aws error desc 1')
+
+        // -- start -- With 'composite' arg as as true
```
justinmk3, line 464:
```suggestion
        // withCause=true
```

justinmk3 (approved):
Good idea. Since `getTelemetryReasonDesc` limits (truncates) the result, there isn't much reason not to do this.

However, there's still an open question about also presenting parts of the cause-chain in our default user-facing error handler.

PR #5375, opened 2024-07-24
feature: added pick string to enable running without any other extensions

32teeth:
## Problem

When constantly restarting the debugger in watch mode, the  [Extension Development Host] sometimes takes an unruly time to load Amazon Q since i might have dozens of other extensions, themes, plugins running.

## Solution

Option
Screenshot 2024-07-24 at 11 45 37 AM

Results
Screenshot 2024-07-24 at 12 14 24 PM

When either restarting of `F5` in the primary window with [Extension Development Host]  already open i am presented with a select option of wether or not i want to run it with other extensions

>[!NOTE]
> Why is it in `launch.json`?

Attempted to solution by placing the "inputs" "pickString" in the `tasks.json` file, this will cause issues and ammend the argument to the end of the command string

ex: `'npm run compileOnly -- --watch' --disable-extensions` 

`tasks.json` is really meant for all the `preLaunchTask` mechanisms anyway

## License

### DIFF
--- a/packages/amazonq/.vscode/launch.json
+++ b/packages/amazonq/.vscode/launch.json
@@ -15,7 +15,7 @@
                 "webRoot": "${workspaceFolder}"
             },
             "runtimeExecutable": "${execPath}",
-            "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
+            "args": ["--extensionDevelopmentPath=${workspaceFolder}", "${input:disableExtensions}"],
             "env": {
                 "SSMDOCUMENT_LANGUAGESERVER_PORT": "6010",
                 "WEBPACK_DEVELOPER_SERVER": "http://localhost:8080"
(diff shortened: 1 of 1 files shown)


jpinkney-aws:
Is there a way to disable extensions except for a select few? With this change every time someone wants to debug the q extension they have to select whether or not they want to disable or enable extensions. It's feels a tad intrusive

32teeth:
> Is there a way to disable extensions except for a select few? With this change every time someone wants to debug the q extension they have to select whether or not they want to disable or enable extensions. It's feels a tad intrusive

This is only on first time build. And only consecutively if you have closed or detached the Host

If you simple refresh,you do not get asked

We can look into potentially selectively disabling but this may be a big ask considering how many a user has

Wondering if there was a specific question

nkomonen-amazon:
@32teeth can you provide a video of what the new flow looks like?

packages/amazonq/.vscode/launch.json
```diff
@@ -102,5 +102,23 @@
...
+                {
+                    "value": "--disable-extensions",
+                    "label": "Yes"
+                },
+                {
+                    "value": "",
+                    "label": "No"
+                }
+            ],
+            "default": "No"
```
nkomonen-amazon, line 121: Just want to verify. If I run Debug mode and the quick pick appears. If I click enter without selecting anything it will automatically choose "No"?

32teeth:
Closing due to inactivity

PR #5377, opened 2024-07-24
fix: Auth Server is not found when user closes webview before finishing browser auth

jpinkney-aws:
## Problem
- If a user opens the authorization url and then before authentication is finished closes the amazonq/toolkit webview the auth server shuts down, giving them a 400 once they try to complete login

## Solution
- Have one auth server per extension that stays open until the extension is deactivated. This simplifies starting/stopping the auth server and sidesteps any of these issues

## License


justinmk3:
Is this still relevant?

PR #5379, opened 2024-07-25
bug: Amazon Q sign in blocked by stale connection

nkomonen-amazon:
## Problem:

Previously the AWS Toolkit and Amazon Q extensions shared Auth connections. To do this they communicated their Auth connections to each other, with each extension storing this in their auth connection cache.

Later we decided to stop sharing the Auth connections between the extensions, instead having each extension create its own connection.

The problem is that when we created this separate auth connection change, we had stale connections remaining in Amazon Q. This caused issues like users being unable to sign in to a new connection since the cached connections showed it already existed.

## Solution:

On every startup clear the stale connections.
We have existing code that does this, but it only ran in a CodeCatalyst dev environment. But it looks like we always want to run this check as it seems there are other cases where this happens.

## License

### DIFF
--- a/packages/core/src/codewhisperer/util/authUtil.ts
+++ b/packages/core/src/codewhisperer/util/authUtil.ts
@@ -34,6 +34,7 @@ import { indent } from '../../shared/utilities/textUtilities'
 import { showReauthenticateMessage } from '../../shared/utilities/messages'
 import { showAmazonQWalkthroughOnce } from '../../amazonq/onboardingPage/walkthrough'
 import { setContext } from '../../shared/vscode/setContext'
+import { isInDevEnv } from '../../shared/vscode/env'
 
 /** Backwards compatibility for connections w pre-chat scopes */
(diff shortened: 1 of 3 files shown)


jpinkney-aws (approved):

ivikash (approved):

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -409,6 +410,23 @@ export class AuthUtil {
 
         return state
     }
+
+    /**
+     * Edge Case: Due to a change in behaviour/functionality, there are potential extra
+     * auth connections that the Amazon Q extension has cached. We need to remove these
+     * as they are irrelevant to the Q extension and can cause issues.
+     */
+    public async clearExtraConnections(): Promise<void> {
```
justinmk3, line 419: Is this concept actually specific to Q/CW or could it live in the top-level auth.ts module (or similar)?

Although it's only called by Q currently, if it's needed for non-Q in the future, it's more discoverable if it lives in the base auth.ts module.

PR #5383, opened 2024-07-26
Code documentation and Unit test cases

Lokesh1302:
## License


packages/core/src/dynamoDb/utils/dynamodb.ts
```diff
@@ -33,7 +42,10 @@ export async function getTableContent(
     return tableData
 }
 
-export function getTableColumnsNames(items: DynamoDB.Types.ScanOutput): {
+/**
```
jpinkney-aws, line 45: same comment as above, simple functions like this are fine to leave uncommented because the function name already tells you what information you'll get. I think that applies for a lot of comments in this PR. My rule of thumb is to only comment if its complex enough that you can't capture good details in the function name. If the function name more or less describes the information you're trying to convey with this function you're good to go

This is mostly because people change function implementations often and rarely go back to change function comments, so by naming the function correctly you sort of kill 2 birds with 1 stone

Lokesh1302: Yes, I have removed code documentation on private methods. I did not change the documentation on public functions as they can be used from a different location and documentation might help.

jpinkney-aws (approved):

Lokesh1302:
Changes are already covered in https://github.com/aws/aws-toolkit-vscode/pull/5399

PR #5386, opened 2024-07-26
fix: add class files from local file

ctlai95:
## Problem

When zipping the project workspace, binary files are getting decoded/added to the zip artifacts as a utf-8 string which results in .class files being invalid and causing scans to fail.

## Solution

Process binary files by adding them to the zip using the local file path instead of a buffer.

## License


jpinkney-aws:
Can you add a changelog item for this?

laileni-aws (approved):

jpinkney-aws (approved):

PR #5392, opened 2024-07-29
fix: FileNotFound error during SSO token refresh

nkomonen-amazon:
## Problem:

In our `aws_refreshCredentials` telemetry we saw a spike in `FileNotFound`
errors when writing the refreshed token to our filesystem.
This is caused by something in the `vscode.workspace.fs.rename()`
implementation that is not obvious and does not seem like an obvious error
on our end. Most of the time things work, so most users are not running
in to this problem.

Originally, to fix some errors happening during writing to the filesystem
we created an "atomic" write, but there are edge cases where this did
not work. So there are different implementations of writing a file that we used
to rememdy this.

## Solution:

Try the next write file method as a fallback when one fails.
See the comment in the code for the explanation of the implementation.

We also report telemetry on the specific failed cases to hopefully get a better understanding
of the specific failures. We can use this

## Additional

- A change was made to `scrubNames()` since the telemetry we were looking at for `aws_refreshCredentials` was missing the file extension. A guess is that the length of the file extension was exceeding 4 chars. So we instead do not factor in the length of the file ext anymore.

## License

### DIFF
--- a/packages/core/src/shared/errors.ts
+++ b/packages/core/src/shared/errors.ts
@@ -345,7 +345,7 @@ export function getTelemetryResult(error: unknown | undefined): Result {
  */
 export function scrubNames(s: string, username?: string) {
     let r = ''
-    const fileExtRe = /\.[^.\/]{1,4}$/
+    const fileExtRe = /\.[^.\/]+$/
     const slashdot = /^[~.]*[\/\\]*/
 
     /** Allowlisted filepath segments. */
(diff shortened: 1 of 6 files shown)


jpinkney-aws (approved):

ivikash (approved):

packages/core/src/shared/errors.ts
```diff
@@ -345,7 +345,7 @@ export function getTelemetryResult(error: unknown | undefined): Result {
  */
 export function scrubNames(s: string, username?: string) {
     let r = ''
-    const fileExtRe = /\.[^.\/]{1,4}$/
+    const fileExtRe = /\.[^.\/]+$/
```
justinmk3, line 348: If the name is really long it's probably not an extension. Do we actually care about names >6 chars?

PR #5398, opened 2024-07-30
config(toolkits): check for presence of test files in PRs

jpinkney-aws:
## Problem
Tests aren't being written

## Solution
Consider adding an action that checks for the presence of test files in a PR

## License

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -32,6 +32,12 @@ jobs:
             - uses: actions/setup-node@v4
               with:
                   node-version: '20'
+            - name: Check for tests
+              uses: actions/github-script@v7
+              with:
+                  script: |
+                      const notify = require('.github/workflows/notify.js')
+                      await notify({github, context})
             - name: Check PR title
               run: |
                   node "$GITHUB_WORKSPACE/.github/workflows/lintcommit.js"
--- a/.github/workflows/notify.js
+++ b/.github/workflows/notify.js
@@ -4,7 +4,7 @@
  */
 
 const needsTestFiles =
-    'We have detected that your changes do not add or modify test files. Please add or update tests or indicate why tests are not needed.'
+    'This pull request modifies files in src/ but no tests were added/updated. Confirm whether tests should be added or ensure the PR description explains why tests are not required.'
 
 /**
  * Remind partner teams that tests are required. We don't need to remind them if:
--- a/.github/workflows/notify.js
+++ b/.github/workflows/notify.js
@@ -1,0 +1,60 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const needsTestFiles =
+    'We have detected that your changes do not add or modify test files. Please add or update tests or indicate why tests are not needed.'
+
+/**
+ * Remind partner teams that tests are required. We don't need to remind them if:
+ *  1. They did not change anything in a src directory
+ *  2. They already have test files in the PR
+ *  3. We've already told them in a previous PR comment
+ * fooo
+ */
+module.exports = async ({ github, context }) => {
+    const owner = context.repo.owner
+    const repo = context.repo.repo
+
+    const response = await github.rest.repos.compareCommitsWithBasehead({
+        owner,
+        repo,
+        basehead: `${context.payload.pull_request.base.ref}...${context.payload.pull_request.head.ref}`,
+    })
+
+    const filenames = response.data.files.map((file) => file.filename)
+
+    // Check if src directory changed
+    const srcFiles = filenames.filter((file) => file.includes('src/'))
+    if (srcFiles.length === 0) {
+        console.log('Did not find src files in the code changes')
+        return
+    }
+
+    // Check if test files were added or modified
+    const testFiles = filenames.filter((file) => file.endsWith('.test.ts'))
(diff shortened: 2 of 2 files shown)


.github/workflows/node.js.yml
```diff
@@ -10,6 +10,34 @@ on:
...
+    check-test-files:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  fetch-depth: 2
+            - name: Check for the presence test files
+              id: check_file
+              run: |
+                  testFiles=$(git diff --name-only ${{ github.event.pull_request.base.sha }} | { grep '\.test\.ts$' || true; })
```
hayemaxi, line 22: I wonder if we can make this smarter? E.g. if source files modified + no test files modified. Getting this for package.json changes or other infra changes would get old quickly I think, but maybe a worthy tradeoff if its not possible.

.github/workflows/node.js.yml
```diff
@@ -10,6 +10,34 @@ on:
...
+              if: steps.check_file.outputs.TEST_FILES_FOUND == 'true'
+              uses: actions/github-script@v7
+              with:
+                  github-token: ${{ secrets.GITHUB_TOKEN }}
+                  script: |
+                      github.rest.issues.createComment({
+                          issue_number: context.issue.number,
+                          owner: context.repo.owner,
+                          repo: context.repo.repo,
+                          body: 'We have detected that your changes do not add or modify test files'
```
hayemaxi, line 38:
```suggestion
                          body: 'We have detected that your changes do not add or modify test files. Please add or update tests, or indicate why tests are not needed.'
```

.github/workflows/node.js.yml
jpinkney-aws: > if source files modified + no test files modified

:+1: It should behave like that now

.github/workflows/node.js.yml
```diff
@@ -10,6 +10,42 @@ on:
...
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  fetch-depth: 2
+            - name: Check for the presence test files
+              id: check_file
+              run: |
+                  # Check if there are any changes in the src folder
+                  sourceFiles=$(git diff --name-only ${{ github.event.pull_request.base.sha }} | { grep 'src' || true; })
+                  if [ -z "$sourceFiles" ]; then
```
justinmk3, line 24: Just for reference: if this script gets bigger, we can move it to a .js file and call it from this `actions/github-script` job as described in: https://github.com/actions/github-script?tab=readme-ov-file#run-a-separate-file

.github/workflows/node.js.yml
```diff
@@ -10,6 +10,42 @@ on:
...
+              if: steps.check_file.outputs.TEST_FILES_FOUND == 'true'
+              uses: actions/github-script@v7
+              with:
+                  github-token: ${{ secrets.GITHUB_TOKEN }}
+                  script: |
+                      github.rest.issues.createComment({
+                          issue_number: context.issue.number,
+                          owner: context.repo.owner,
+                          repo: context.repo.repo,
+                          body: 'We have detected that your changes do not add or modify test files. Please add or update tests, or indicate why tests are not needed.'
```
justinmk3, line 46: Now that conventional commits are checked https://github.com/aws/aws-toolkit-vscode/pull/5458 , we could also leave a note requesting a changelog item for `feat` and `fix` PRs.

.github/workflows/node.js.yml
```diff
@@ -10,6 +10,42 @@ on:
...
+                      exit 0
+                  fi
+                  echo "TEST_FILES_FOUND=true" >> $GITHUB_OUTPUT
+            - name: Notify contributor
+              if: steps.check_file.outputs.TEST_FILES_FOUND == 'true'
+              uses: actions/github-script@v7
+              with:
+                  github-token: ${{ secrets.GITHUB_TOKEN }}
+                  script: |
+                      github.rest.issues.createComment({
```
justinmk3, line 42: Is there a way to update the existing comment instead of always creating a new one?

.github/workflows/node.js.yml
jpinkney-aws: Just changed to a script, I think it's going to be easier to maintain

.github/workflows/node.js.yml
jpinkney-aws: In the newest changes the commits are now de-dupped

.github/workflows/node.js.yml
jpinkney-aws: Should we do that separately inside of the lintcommit file? It would actually be reasonably easy to get that working

justinmk3: Not sure. We can try either way and see what works best :)

.github/workflows/notify.js
```diff
@@ -0,0 +1,60 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const needsTestFiles =
+    'We have detected that your changes do not add or modify test files. Please add or update tests or indicate why tests are not needed.'
```
justinmk3, line 7:
```suggestion
    'This pull request modifies files in src/ but no tests were added/updated. Confirm whether tests should be added or ensure the PR description explains why tests are not required.'
```

Ideally this comment would be skipped if the PR description already mentions "test"?

justinmk3 (approved):

PR #5402, opened 2024-07-31
feat(amazonq): parse pom.xml for absolute paths

dhasani23:
## Problem

Some customers have absolute path references in their `pom.xml` files, which causes issues during our backend build as we can't access those things.

## Solution

Try to find absolute paths and warn users if we detect any.

image

## License


dhasani23:
Opened this PR, but double checking to investigate how necessary this really is, will update here shortly.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -231,6 +234,22 @@ export async function finalizeTransformByQ(status: string) {
     }
 }
 
+export async function parseBuildFile() {
+    const absolutePaths = ['Users/', 'System/', 'Volumes/', 'C:', 'D:']
```
hayemaxi, line 238: is it possible that these could also exist in pom.xml as lowercase? e.g. vscode on windows and node normalize drive paths to lowercase sometimes.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -231,6 +234,22 @@ export async function finalizeTransformByQ(status: string) {
...
+export async function parseBuildFile() {
+    const absolutePaths = ['users/', 'system/', 'volumes/', 'c:', 'd:']
+    const alias = path.basename(os.homedir())
+    absolutePaths.push(alias)
+    const buildFilePath = path.join(transformByQState.getProjectPath(), 'pom.xml')
+    if (fs.existsSync(buildFilePath)) {
+        const buildFileContents = fs.readFileSync(buildFilePath).toString().toLowerCase()
+        if (absolutePaths.some((path) => buildFileContents.includes(path))) {
+            void vscode.window.showWarningMessage(
+                `We detected what may be an absolute path in this file: ${path.basename(buildFilePath)}, which may cause issues during our backend build. You will see error logs open if this happens.`
```
dhasani23, line 246: Intentionally keeping this generic with `path.basename()` because soon `build.gradle` will be parsed here too.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -232,6 +235,34 @@ export async function finalizeTransformByQ(status: string) {
...
+            const detectedPaths = []
+            for (const absolutePath of absolutePaths) {
+                if (buildFileContents.includes(absolutePath)) {
+                    detectedPaths.push(absolutePath)
+                }
+            }
+            if (detectedPaths.length > 0) {
+                const warningMessage = `We detected ${detectedPaths.length} absolute ${detectedPaths.length === 1 ? 'path' : 'paths'} (${detectedPaths.join(', ')}) in this file: ${path.basename(buildFilePath)}, which may cause issues during our backend build. You will see error logs open if this happens.`
+                void vscode.window.showWarningMessage(warningMessage)
+                getLogger().info('CodeTransformation: absolute path potentially in build file')
```
tincheng, line 256: Can we show the warning in Chat instead?

dhasani23: Good point; let me check with PM.

dhasani23: Updated to show warning in Chat.

tincheng (approved):

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -457,7 +462,7 @@ export class GumbyController {
 
     private async handleError(message: { error: Error; tabID: string }) {
         if (message.error instanceof AlternateDependencyVersionsNotFoundError) {
-            this.messenger.sendKnownErrorResponse('no-alternate-dependencies-found', message.tabID)
+            this.messenger.sendKnownErrorResponse(message.tabID, CodeWhispererConstants.dependencyVersionsErrorMessage)
```
jpinkney-aws, line 465: Out of curiosity why is this dependency version error message stored in codewhisperer constants when it has to do with gumby code?

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -232,6 +233,39 @@ export async function finalizeTransformByQ(status: string) {
...
+        if (fs.existsSync(buildFilePath)) {
+            const buildFileContents = fs.readFileSync(buildFilePath).toString().toLowerCase()
+            const detectedPaths = []
+            for (const absolutePath of absolutePaths) {
+                if (buildFileContents.includes(absolutePath)) {
+                    detectedPaths.push(absolutePath)
+                }
+            }
+            if (detectedPaths.length > 0) {
+                const warningMessage = CodeWhispererConstants.absolutePathDetectedMessage
```
jpinkney-aws, line 251: Please use string interpolation here: https://stackoverflow.com/questions/45399951/how-to-perform-string-interpolation-in-typescript. We shouldn't need to be replacing magic strings in text

packages/amazonq/.changes/next-release/Feature-adb7e3c8-9c3e-4090-a283-2d29fc396f04.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "feat(Amazon Q Code Transformation): warn user if absolute path found in pom.xml"
```
hayemaxi, line 3:
```suggestion
	"description": "Code Transformation: Warn user if absolute paths are found in pom.xml"
```
Changelog entries are user-facing and listed right on the marketplace pages, so they need to be more friendly than commit style messages

packages/core/src/codewhisperer/commands/startTransformByQ.ts
hayemaxi: Yes. Additionally, you can get rid of this new constant `CodeWhispererConstants .absolutePathDetectedMessage` and add the message directly to the `AbsolutePathDetectedError` class. There, you can pass your params and autofill the message with string interpolation

`new AbsolutePathDetectedError({numPaths: 3, paths: ['my/path'], buildFile: 'my/buildfile.file']})`

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -232,6 +233,39 @@ export async function finalizeTransformByQ(status: string) {
...
+                transformByQState.getChatControllers()?.errorThrown.fire({
+                    error: new AbsolutePathDetectedError(warningMessage),
+                    tabID: ChatSessionManager.Instance.getSession().tabID,
+                })
+                getLogger().info('CodeTransformation: absolute path potentially in build file')
+                return warningMessage
+            }
+        }
+    } catch (err: any) {
+        // swallow error
```
hayemaxi, line 264: Should we log the error at least?

packages/core/src/amazonqGumby/chat/controller/controller.ts
dhasani23: All of our constants are in CodeWhisperer constants because that's how QCT was originally intended. Now that we're entirely separate, a future PR is needed to move those constants.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
dhasani23: If I use string interpolation, my String constant is just going to be floating around in `startTransformByQ.ts` rather than being stored in the constants file, so I'd prefer to keep it in the constants file unless you think this is highly problematic. And I already do add the message directly to the `AbsolutePathDetectedError` class if you see a few lines below.

jpinkney-aws: Why cant it live in constants anymore if you use string interpolation? It should be fine to keep it there. Or alternatively what max said works as well

dhasani23: I need to write out the String in-line no? Something like `We detected {numPaths} file paths in file {fileName}` which won't allow me to keep the String as a constant in that file

jpinkney-aws: Not sure if i'm understanding 😅 but why not? There are several instances of doing this in the constants file:
e.g. https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/models/constants.ts#L167, https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/models/constants.ts#L169, https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/models/constants.ts#L179

jpinkney-aws:
```typescript
export const absolutePathDetectedMessage = (numPaths: number, buildFile: string, listOfPaths: string[]) =>
    'I detected ${numPaths} absolute file path(s) in your ${buildFile} file: **${listOfPaths}**. Absolute file paths might cause issues when I build your code. Any errors will show up in the build log.'
```

dhasani23: Ah now I see. I didn't realize string interpolation could be done that way too, fixing now.

jpinkney-aws (approved):

PR #5403, opened 2024-07-31
Amazon Q Code Transform: Improve telemetry to capture transform journey

tincheng:
## Problem
Emit new metrics to better capture user journey for Amazon Q Code Transform.

## Solution
New metrics published:
```
codeTransform_initiateTransform
codeTransform_validateProject
codeTransform_submitSelection
codeTransform_localBuildProject
codeTransform_uploadProject
codeTransform_jobStart
codeTransform_viewArtifact
codeTransform_downloadArtifact
```
Marked deprecated metrics as To-be-deleted for BI backward compatibility.

Also changed usage of `Date.now()` to `globals.clock.Date.now()` in transform as requested.

## License


packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -261,6 +280,18 @@ export class GumbyController {
                 reason: err?.code,
                 reasonDesc: getTelemetryReasonDesc(err),
             })
+
+            // TODO: if build system version is available at this stage, include buildSystemVersion field.
```
dhasani23, line 284: Local build system version is not available here, but we do already emit a metric for this when that logic runs: https://github.com/aws/aws-toolkit-vscode/blob/3c8c3a12a16042d5a8eb61afee3f8afc849efe5a/packages/core/src/codewhisperer/commands/startTransformByQ.ts#L120-L126

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -216,10 +222,23 @@ export class GumbyController {
...
-            CodeTransformTelemetryState.instance.setSessionId()
             this.messenger.sendTransformationIntroduction(message.tabID)
+        } catch (e: any) {
+            // if there was an issue getting the list of valid projects, the error message will be shown here
+            this.messenger.sendErrorMessage(e.message, message.tabID)
+            errorMessage = e.code
+        } finally {
+            // Emit initial metric for /transform
+            const authType = await getAuthType()
+            telemetry.codeTransform_initiateTransform.emit({
```
hayemaxi, line 233: Please see [telemetry.md](https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md).
You can remove a lot of this boilerplate code by doing using `telemetry.run()` instead of emit. **Applicable throughout entire PR.**
```
try {
    telemetry.codeTransform_initiateTransform.emit(async () => {
        // logic
        // ...

        telemetry.record({codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId()})
        
        // some more logic
        // ...
        
        // once you know more information, you can add it. If you knew all information at the start, the one `record` is sufficient
        telemetry.record({credentialSourceId: authType})

        // if something is wrong, instead of recording a string, just throw an exception
        if (authState.amazonQ !== 'connected') {
                void this.messenger.sendAuthNeededExceptionMessage(authState, message.tabID)
                throw new Error('auth-failed') // prefer a more detailed exception, this is just an example.
        }
    })
} catch (e) {
    this.messenger.sendErrorMessage(e.message, message.tabID)
}
```
Result, reason, and reasonDesc will be filled automatically from the appropriate fields of the exception.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -338,6 +400,7 @@ export class GumbyController {
         }
 
         const projectPath = transformByQState.getProjectPath()
+        // TODO: remove deprecated metric once BI started using new metrics
```
hayemaxi, line 403: Please add an appropriate test in `core/src/test/techdebt.test.ts`

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -168,6 +169,7 @@ export function startInterval() {
 
 export async function startTransformByQ() {
     // Set the default state variables for our store and the UI
+    const transformStartTime = Date.now()
```
dhasani23, line 172: This is passed to startJob and eventually we use it to calculate the latency of the API call, but here it's set long before we call startJob? I think we should just be using `Date.now()` right before we call the API.

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -320,8 +320,9 @@ export class ProposedTransformationExplorer {
                 transformByQState.getJobId(),
                 'ExportResultsArchive.zip'
             )
-
+            let exportResultsArchiveSize = 0
             let downloadErrorMessage = undefined
+            let downloadStartTime = Date.now()
```
dhasani23, line 325: nit: use `const`, as suggested below

packages/core/src/codewhisperer/commands/startTransformByQ.ts
hayemaxi, line 172: Do not use `Date` object directly. Use `globals.clock.Date`.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -238,18 +240,27 @@ export async function preTransformationUploadCode() {
     void vscode.window.showInformationMessage(CodeWhispererConstants.jobStartedNotification)
 
     let uploadId = ''
-    let payloadFilePath = ''
+    let payloadFilePath = undefined
+    let zipSize = 0
+    let dependenciesCopied = false
+    let errorMessage = undefined
+    const zipStartTime = Date.now()
```
hayemaxi, line 247: Use `globals.clock.Date`.

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
hayemaxi, line 325: Use `globals.clock.Date`. Also, you will have to fix the lint suggestion to get that test to pass.

packages/core/src/amazonqGumby/chat/controller/controller.ts
tincheng: I'm planning to deprecate `isDoubleClickedToTriggerInvalidProject` with `validateProject`, but thanks for sharing! I'll see if I can get the information earlier.

packages/core/src/amazonqGumby/chat/controller/controller.ts
tincheng: This is super useful! Although at the moment our BI team is still discussing on a date, so this PR will be delayed a bit until I get a committed date.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
tincheng: Since we are removing metrics based on per API call, I want to capture the actual latency from user submitting the form to transformation started (as part of user journey). We can still find latency to the API call on codewhisperer service side.

tincheng: Sure, I can switch them.

packages/core/src/amazonqGumby/chat/controller/controller.ts
tincheng: After taking a look at telemetry.md, I see there are benefit of using run() if the telemetry span across multiple functions or contains a large number of fields that need to be updated. However, looks like for it to set the correct failure result, it would need a `ToolkitError`, and I don't think transform is using ToolkitError everywhere (which may be something to fix, but I'll leave it out of scope for the PR).

I found it easier to read when emitting the whole metric through emit(). It makes it easy to see which fields had been populated, and follows normal coding practices where a metric is emitted at the end of some operation (e.g. try/catch/finally). Since we are not emitting redundant metrics, I'm more inclined to keep using emit() the way it is.

packages/core/src/amazonqGumby/chat/controller/controller.ts
tincheng: I tried fetching maven version in an earlier step, but got `'undefined (mvn)'`. I need more investigation here, but since we weren't publishing maven version in project validation metrics before, I'll skip it in this PR.

dhasani23 (approved):

packages/core/src/test/techdebt.test.ts
```diff
@@ -45,4 +45,10 @@ describe('tech debt', function () {
         // forgetConnection() function and calls
         fixByDate('2024-08-30', 'Remove the edge case code from the commit that this test is a part of.')
     })
+
+    it('remove deprecated code transform metrics', async function () {
```
hayemaxi, line 49: Nice, thanks!

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -179,47 +182,64 @@ export class GumbyController {
...
+                telemetry.record({
+                    codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId(),
+                    credentialSourceId: authType,
+                })
+
+                // check that a project is open
+                const workspaceFolders = vscode.workspace.workspaceFolders
+                if (workspaceFolders === undefined || workspaceFolders.length === 0) {
+                    this.messenger.sendUnrecoverableErrorResponse('no-project-found', message.tabID)
+                    telemetry.record({ result: MetadataResult.Fail, reason: 'no-project-found' })
```
hayemaxi, line 201:
```suggestion
                    throw { name: 'no-project-found', message: 'Q Code Transform could not find a project in this workspace.' }
```
Result and reason will be set automatically if you throw an error. If you do it this way, then the function will continue to run since no error is thrown (seems unintended) and when it ends, telemetry will record a success since no error was thrown.

Ideally you throw a custom error implementation:
```
class NoProjectFoundError extends Error {
    override name: string = 'no-project-found' // name will be gathered by telemetry automatically in .run()
}

...

throw new NoProjectFoundError("")
```

packages/core/src/codewhisperer/service/transformByQ/transformMavenHandler.ts
```diff
@@ -16,71 +20,83 @@ import { throwIfCancelled } from './transformApiHandler'
...
+            // TODO: remove deprecated metric once BI started using new metrics
+            telemetry.codeTransform_mvnBuildFailed.emit({
+                codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId(),
+                codeTransformMavenBuildCommand: mavenBuildCommand as CodeTransformMavenBuildCommand,
+                result: MetadataResult.Fail,
+                reason: errorReason,
+            })
+
+            // Explicitly set metric as failed since no exception was caught
+            telemetry.record({ result: MetadataResult.Fail, reason: errorReason })
```
hayemaxi, line 93: This will be set as a result of the next line. Make sure to throw the error with the reason that you want
```
throw new ToolkitError(`Maven ${argString} error: ${errorReason}`, { code: 'MavenExecutionError' })
```

hayemaxi (approved):
Original exception handling logic isn't ideal (more info in comments), but out of scope for this PR

PR #5407, opened 2024-07-31
feat(amazonq): emit ide_editCodeFile

hayemaxi:
4294f8d1522511308651d3f82992c1489f7c2e4c only emits for toolkit. This adds to amazon q as well.

This will mean telemetry events are duplicated if both extensions are installed. With some additional complexity and over-engineering, we can avoid this. But de-duplication should not be a big deal across products.

## License

### DIFF
--- a/packages/amazonq/src/extensionNode.ts
+++ b/packages/amazonq/src/extensionNode.ts
@@ -7,7 +7,8 @@ import * as vscode from 'vscode'
 import { activateAmazonQCommon, amazonQContextPrefix, deactivateCommon } from './extension'
 import { DefaultAmazonQAppInitContext } from 'aws-core-vscode/amazonq'
 import { activate as activateQGumby } from 'aws-core-vscode/amazonqGumby'
-import { ExtContext } from 'aws-core-vscode/shared'
+import { ExtContext, globals } from 'aws-core-vscode/shared'
+import { filetypes, SchemaService } from 'aws-core-vscode/sharedNode'
 import { updateDevMode } from 'aws-core-vscode/dev'
 import { CommonAuthViewProvider } from 'aws-core-vscode/login'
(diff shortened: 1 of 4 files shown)


dhruvigajjar (approved):

packages/amazonq/src/extensionNode.ts
```diff
@@ -51,6 +52,9 @@ async function activateAmazonQNode(context: vscode.ExtensionContext) {
         registerSubmitFeedback(context, 'Amazon Q', amazonQContextPrefix)
     )
 
+    globals.schemaService = new SchemaService()
+    filetypes.activate()
```
justinmk3, line 56: this looks like another case that was missed because we don't yet have a startup "base" that is used by both Q and Toolkit : https://taskei.amazon.dev/tasks/IDE-14189

packages/core/src/shared/indexNode.ts
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Contains exports that only work for node, NOT web.
+ * Attempting to import from this file in web will throw an error,
+ * likely `TypeError: Cannot read properties of undefined (reading 'native')`
+ */
+
+export * as filetypes from './filetypes'
```
justinmk3, line 12: what is node-only about filetypes.ts ? we should try to avoid exposing node-only or web-only in our index.ts files. that's an implementation detail

hayemaxi: Something imported in `SchemaService` is node only and I haven't looked into what it was yet. `filetypes` uses `SchemaService`, and adding the schema service to the standard `index.ts` caused web to fail even if it doesn't use that import.
I'm not sure what the alternative is besides finding and eliminating the node dependency in `SchemaService`

PR #5410, opened 2024-08-01
feat(amazonq): handle extension uninstallation

ivikash:
Open a Survey Link when Amazon Q is uninstalled.

TODO
- [ ] Update survey link. The link in the PR is a placeholder link for now
- [x] Telemetry and Logger needs to be verified. 

## License


packages/core/src/shared/handleUninstall.ts
```diff
@@ -0,0 +1,79 @@
...
+    obsoleteFilePath: string,
+    extensionsDirPath: string
+): Promise<void> {
+    try {
+        const [obsoleteFileContent, extensionsDirContent] = await Promise.all([
+            vscode.workspace.fs.readFile(vscode.Uri.file(obsoleteFilePath)),
+            vscode.workspace.fs.readDirectory(vscode.Uri.file(extensionsDirPath)),
+        ])
+
+        const extensionId = context.extension.id
```
hayemaxi, line 31: Let's pass only the `extension.id` (maybe typed to `VSCODE_EXTENSION_ID`) to this function since its not using anything else in `context`

hayemaxi (approved):

jpinkney-aws (approved):

packages/core/src/shared/handleUninstall.ts
```diff
@@ -0,0 +1,95 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+import { join } from 'path'
+import { getLogger } from './logger/logger'
+import { telemetry } from './telemetry'
+import { VSCODE_EXTENSION_ID } from './extensions'
+import { extensionVersion } from './vscode/env'
+
```
nkomonen-amazon, line 12: Can we add documentation which links to the original implementation and note how this is a hack.
If this implementation stops working, future devs will be able to look at the other project's work and hopefully they will have come up with another solution.

packages/core/src/shared/handleUninstall.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Sets up a file system watcher to monitor the .obsolete file for changes and handle
+ * extension un-installation if the extension is marked as obsolete.
+ *
+ * @param {string} extensionName - The name of the extension.
+ * @param {vscode.ExtensionContext} context - The extension context.
+ * @param {function} callback - Action performed when extension is uninstalled.
+ * @returns {void}
+ */
+export function handleUninstall(
```
nkomonen-amazon, line 69: optional nit: rename to something like `start/setupUninstallHandler`

Issue #5412, opened 2024-08-01
Sign in to AWS Toolkit in Visual Studio Code with AWS IAM Identity Center - Start URL not accepted


dominiklanger:
## Problem

I cannot log in to the AWS Toolkit with AWS IAM Identity Center. The Start URL is not accepted. When I enter my Start URL, the following message is shown in red font below the "Start URL" text field: "_URLs must start with http:// or https://. Example: https://d-xxxxxxxxxx.awsapps.com/start_". The "Continue" button remains disabled (grayed out) and I cannot continue. See screenshot below:

![image](https://github.com/user-attachments/assets/ccdf9ccb-2c26-4777-91e5-b9b603fd8a1d)

Note: I don't seem to be the only user with this problem. See the review from Chris Weis entered on 6/10/2024: https://marketplace.visualstudio.com/items?itemName=AmazonWebServices.aws-toolkit-vscode&ssr=false#review-details

## Steps to reproduce the issue

1. Install Visual Studio Code.
2. Install AWS Toolkit version via the Extensions Marketplace.
3. Create an AWS user via IAM Identity Center.
4. Open the AWS Toolkit extension.
5. Choose sign-in option "Workforce - Sign in to AWS with single sign-on".
6. Click "Continue" button.
7. On the the screen "Sign in with AWS IAM Identity Center", enter the Start URL associated with the AWS user created in step 2 (which is of the following format: https://d-xxxxxxxxxx.awsapps.com/start/) and select AWS region from the "Region" drop-down.

## Expected behavior

- After filling the Start URL and the Region, the "Continue" button should become enabled but does not. Instead, the "Start URL" error message persists, as described above.

## System details (run `AWS: About` and/or `Amazon Q: About`)

-   OS: Windows 11 Enterprise
-   Visual Studio Code version: 1.91.1
-   AWS Toolkit version: v3.17.0
-   Amazon Q version: N/A

hayemaxi:
Hi, thanks for reporting this. Two quick questions:
1. Does the url you are trying to enter contain a forward slash as the last character? E.g. `.../start/` instead of `.../start```
    - There seems to be a bug where the URL is not accepted if it ends with a forward slash.
2. In the input box, are you typing any portion of the start url manually or do you get this error from directly copy pasting?

dominiklanger:
1. Yes my url contained a forward slash as the last character. I now removed this forward slashed and it worked properly. Thank you!
2. I copy-pasted the URL from the AWS console.

PR #5418, opened 2024-08-02
Misc small changes (allow trailing slash in start url, updating log level)

hayemaxi:
Contains 2 commits (don't squash merge)

[chore: change ec2 token fetch error to warn instead](https://github.com/aws/aws-toolkit-vscode/commit/f247601818a5949aab279c44242d60d1dc115b8c) 

It isn't really an error unless you are using the ext in EC2. It's just cluttering the logs.
`warn` may even be a stretch.

[fix(auth): allow trailing slash for start urls](https://github.com/aws/aws-toolkit-vscode/commit/717062f91cb98ae5346b49c12744b5caebd6b525) 

Might fix cases for some users reporting that the auth screen is not accepting their start url. ~~May or may not~~ Does fix https://github.com/aws/aws-toolkit-vscode/issues/5412

## License

### DIFF
--- a/packages/core/src/login/webview/vue/login.vue
+++ b/packages/core/src/login/webview/vue/login.vue
@@ -286,7 +286,7 @@ type Stage = 'START' | 'SSO_FORM' | 'CONNECTED' | 'AUTHENTICATING' | 'AWS_PROFIL
 
 function validateSsoUrlFormat(url: string) {
     const regex =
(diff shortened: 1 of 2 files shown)


jpinkney-aws (approved):

nkomonen-amazon:
The original code looks to be very similar to `auth/sso/validation.ts`. We should look to unify this if possible in a future PR if we do not want to block this.

PR #5420, opened 2024-08-03
[Security Scans] Adding performance test for auto scans

laileni-aws:
## Problem
- Security scans does not have performance tests for file scan.
## Solution
- Added performance test for file scans with payload size of 200KB(Max Payload size).
- Observations in local machine(Mac)

**Input:**

> Payload Size is 198.99 KB

**Result:**

| Machine  | CPU Usage % | Memory Usage in MB  | 
| ------------- | ------------- | ------------- | 
| Local Machine(Mac OS)  | 25.34  | 239.0625  |
| MacOS 16.x minimum  | 23.08  | 228.32813  |
| MacOS 16.x stable  | 19.97  | 209.23438  |
| MacOS 16.x insiders  | 22.45  | 216.48438 |
| Windows 16.x stable  | 39.98  | 206.50391  |
| Windows 16.x insiders  | -  | 214.00391  |

> Memory Usage for performing a file scan : 203.75390625 MB
> Start of a Scan:
> User CPU Usage: 21.446874 sec and System CPU Usage: 7.044477 sec
> Difference of CPU Usage at End of a Scan:
> User CPU Usage: 0.058255 sec and System CPU Usage: 0.004867 sec

This depends on machine, os and other factors, currently gave a limit of 400MB max memory usage but its volatile.

## License


packages/core/src/test/codewhisperer/commands/startSecurityScan.test.ts
```diff
@@ -449,3 +452,107 @@ describe('startSecurityScan', function () {
...
     })
 })
+
+describe('startSecurityScanPerformanceTest', function () {
+    beforeEach(async function () {
+        extensionContext = await FakeExtensionContext.create()
+        mockSecurityPanelViewProvider = new SecurityPanelViewProvider(extensionContext)
+
+        // Create a temp directory
+        const tempDirectory = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'temp-'))
```
justinmk3, line 462: `createTestWorkspaceFolder` https://github.com/aws/aws-toolkit-vscode/blob/408befebd22af28e9759bcc66c50f49f5fbb6c47/packages/core/src/test/testUtil.ts#L64

packages/core/src/test/codewhisperer/commands/startSecurityScan.test.ts
```diff
@@ -449,3 +449,104 @@ describe('startSecurityScan', function () {
...
+
+        assert(
+            EndScanCpuUsageByUser > 0,
+            `User CPU usage difference should be greater than 0 but actual is ${EndScanCpuUsageByUser}`
+        )
+
+        // These limits are considered from local mac but may vary with machine, OS etc factors
+
+        assert(
+            EndScanMemoryUsageInMB < 400,
```
hayemaxi, line 539: How was 400 determined exactly? It is not clear by the above comment.
Does the size of the input impact this? Shouldn't this be sampled while the scan is in progress vs. after it has finished?

laileni-aws: Yes, Size of the input, machine os and other factors impact this IIRC. 
400MB is the range in which machine did not slow in my observations(made a doc and posted in channel).

packages/core/src/test/codewhisperer/commands/startSecurityScan.test.ts
```diff
@@ -449,3 +449,104 @@ describe('startSecurityScan', function () {
...
+            Result:-
+            Scan Memory Usage at end: 203.75390625 MB
+            Start of a Scan:
+                User CPU Usage: 21.446874 sec and System CPU Usage: 7.044477 sec
+            Difference at End of a Scan:
+                User CPU Usage: 0.058255 sec and System CPU Usage: 0.004867 sec
+         */
+
+        assert(
+            EndScanCpuUsageByUser > 0,
```
hayemaxi, line 532: Maybe I don't fully understand what is being captured here but this can never be 0 or less right? Since you can't have "negative" CPU usage.
Shouldn't this be converting a percentage of CPU usage and validating it against some baseline, e.g. cpuUsage < 50%

packages/core/src/test/codewhisperer/commands/startSecurityScan.test.ts
hayemaxi: Makes sense. It would be good if we can quickly validate with others that 400MB is the upper limit.

packages/core/src/test/codewhisperer/commands/startSecurityScan.test.ts
laileni-aws: calculated and tested CPU usage in 6 different machines and made note in the summary.

hayemaxi:
Interesting, windows seems to use much more CPU (but we would need a larger sample size to be more confident about that). Maybe we should consider having separate test threholds for each OS, e.g. windows: 50%, mac: 35%, linux: ?%

hayemaxi (approved):

PR #5422, opened 2024-08-05
Delete Item from Context Menu

Lokesh1302:
## Problem

User should be able to delete item from context menu of table row.

## Solution

image

## License

### DIFF
--- a/packages/core/src/dynamoDb/utils/dynamodb.ts
+++ b/packages/core/src/dynamoDb/utils/dynamodb.ts
@@ -4,6 +4,7 @@
  */
 
 import { DynamoDB } from 'aws-sdk'
+import { getLogger } from 'aws-core-vscode/shared'
 import { copyToClipboard } from '../../shared/utilities/messages'
 import { DynamoDbTableNode } from '../explorer/dynamoDbTableNode'
 import { DynamoDbClient } from '../../shared/clients/dynamoDbClient'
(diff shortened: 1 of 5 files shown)


packages/core/src/dynamoDb/utils/dynamodb.ts
```diff
@@ -4,6 +4,7 @@
  */
 
 import { DynamoDB } from 'aws-sdk'
+import { getLogger } from 'aws-core-vscode/shared'
```
jpinkney-aws, line 7: For dynamodb we should use the relative import rather than the library import

packages/core/src/dynamoDb/utils/dynamodb.ts
```diff
@@ -216,3 +217,36 @@ function getAttributeValue(attribute: AttributeValue): { key: string; value: any
...
+) {
+    const partitionKeyName = tableSchema.partitionKey.name
+    const partitionKeyValue = selectedRow[partitionKeyName]
+
+    const deleteRequest: DynamoDB.DocumentClient.DeleteItemInput = {
+        TableName: tableName,
+        Key: {
+            partitionKeyName: {
+                S: partitionKeyValue,
+            } as any,
```
jpinkney-aws, line 236: Does this have to be any?

packages/core/src/dynamoDb/utils/dynamodb.ts
```diff
@@ -216,3 +217,36 @@ function getAttributeValue(attribute: AttributeValue): { key: string; value: any
...
+        },
+    }
+    if (tableSchema.sortKey) {
+        const sortKeyName = tableSchema.sortKey.name
+        const sortKeyValue = selectedRow[sortKeyName]
+        deleteRequest.Key[sortKeyName] = { S: sortKeyValue } as any
+    }
+
+    try {
+        const response = await client.deleteItem(deleteRequest)
```
jpinkney-aws, line 246: I think you could just `return client.deleteItem(deleteRequest)` instead of reassigning. Or alternativelty `return await client.deleteItem(deleteRequest)`.

https://stackoverflow.com/questions/38708550/difference-between-return-await-promise-and-return-promise has more info on the differences

PR #5423, opened 2024-08-05
feat(amazonq): show users estimated cost of /transform

dhasani23:
## Problem

Users of large projects on our pro tier don't have an estimate of how much /transform will cost on that project.

## Solution

Show the max /transform may cost in the transformation plan.

image

## License


tincheng (approved):

packages/amazonq/.changes/next-release/Feature-9dd7b119-abfe-4372-aec9-eeb9a938e417.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "feat(Amazon Q Code Transformation): show pro tier users estimated cost of /transform on projects over 100K lines\""
```
hayemaxi, line 3:
```suggestion
	"description": "Code Transform: Show pro tier users estimated cost of `/transform` on projects over 100K lines"
```

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -640,6 +640,13 @@ export async function getTransformationPlan(jobId: string) {
         const arrowIcon = getTransformationIcon('upArrow')
 
         let plan = `<style>table {border: 1px solid #424750;}</style>\n\n<a id="top"></a><br><p style="font-size: 24px;"><img src="${logoIcon}" style="margin-right: 15px; vertical-align: middle;"></img><b>${CodeWhispererConstants.planTitle}</b></p><br>`
+        const authType = await getAuthType()
+        const linesOfCode = jobStatistics.find(
+            (stat: { name: string; value: number }) => stat.name === 'linesOfCode'
+        ).value
+        if (authType === 'iamIdentityCenter' && linesOfCode > 100000) {
+            plan += `<p>${linesOfCode} lines of code submitted for transformation, maximum charge of this transformation is $${(linesOfCode * 0.003).toFixed(2)} (this charge applies only after the free limit in your organization's subscriptions is exhausted). To prevent the charge, you can stop the job before the transformation completes.</p>`
```
hayemaxi, line 648: Can we pull this price calculation out into a utility function so its more obvious? It would also make it easier to update and keep track of later.
Is there a risk with the price updating and this calculation being stale, which would show  the user the wrong price?

dhasani23: Good point about the price updating! We decided to link to our docs instead, will update PR.

packages/core/src/codewhisperer/models/constants.ts
```diff
@@ -410,6 +410,11 @@ export const codeTransformTroubleshootDownloadError =
 export const codeTransformPrereqDoc =
     'https://docs.aws.amazon.com/amazonq/latest/aws-builder-use-ug/code-transformation.html#prerequisites'
 
+export const codeTransformBillingText =
```
jpinkney-aws, line 413: Why don't you make this a function that takes in the number of lines of code and then does the calculations inline using string interpolation?

e.g.

```typescript
export const codeTransformBillingText = (linesOfCode: number) {
    return `<p>${linesOfCode} lines of code were submitted for transformation. If you reach the quota for lines of code included in your subscription, you will be charged ${CodeWhispererConstants.codeTransformBillingRate.toString()} for each additional line of code. You might be charged up to ${(linesOfCode * CodeWhispererConstants.codeTransformBillingRate).toFixed(2)} for this transformation. To avoid being charged, stop the transformation job before it completes. For more information on pricing and quotas, see [Amazon Q Developer pricing](https://aws.amazon.com/q/developer/pricing/).</p>`
}
```

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -603,6 +603,14 @@ export function getJobStatisticsHtml(jobStatistics: any) {
     return htmlString
 }
 
+export function getBillingString(linesOfCode: number) {
```
jpinkney-aws, line 606: See above comment, you shouldn't need to manually replace magic strings in your text. String interpolation can do this for you

dhasani23: See my reply on other PR; not opposed to String interpolation but then my String constants will get disorganized.

packages/core/src/codewhisperer/models/constants.ts
dhasani23: Fixed

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
dhasani23: Edit: fixed

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -603,6 +603,9 @@ export function getJobStatisticsHtml(jobStatistics: any) {
     return htmlString
 }
 
+export const getBillingString = (linesOfCode: number): string =>
```
jpinkney-aws, line 606: Why do we need this function that just wraps the same contents as codeTransformBillingText? Can't we just call it directly?

dhasani23: That's true; fixed

jpinkney-aws (approved):

PR #5428, opened 2024-08-07
Delete Item from Context Menu 

Lokesh1302:
## Problem

User should be able to delete item from context menu of table row.

## Solution

image

## License

### DIFF
--- a/packages/core/src/dynamoDb/utils/dynamodb.ts
+++ b/packages/core/src/dynamoDb/utils/dynamodb.ts
@@ -231,7 +231,7 @@ export async function deleteItem(
     const deleteRequest: DynamoDB.DocumentClient.DeleteItemInput = {
         TableName: tableName,
         Key: {
-            partitionKeyName: {
+            [partitionKeyName]: {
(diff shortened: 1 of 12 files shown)


packages/core/src/dynamoDb/utils/dynamodb.ts
```diff
@@ -216,3 +216,33 @@ function getAttributeValue(attribute: AttributeValue): { key: string; value: any
...
+            } as any,
+        },
+    }
+    if (tableSchema.sortKey) {
+        const sortKeyName = tableSchema.sortKey.name
+        const sortKeyValue = selectedRow[sortKeyName]
+        deleteRequest.Key[sortKeyName] = { S: sortKeyValue } as any
+    }
+    const response = await client.deleteItem(deleteRequest)
+    if (response.$response.error) {
```
jpinkney-aws, line 244: Does it make sense to throw an error here? and then handle that error accordingly inside of the tableView?

jpinkney-aws (approved):

PR #5432, opened 2024-08-08
auth: Add more network errors

nkomonen-amazon:
## Problem

During SSO token refresh (`aws_refreshCredentials`) we were seeing certain transitive Network specific errors causing the refresh to fail and the connection to show as expired.

## Solution

Add the specific errors to the `isNetworkError()` function.

What this ends up doing:
- when these errors appear during token refresh the SSO connection remains valid, but the `getToken()` call will throw

## License

### DIFF
--- a/packages/core/src/test/shared/errors.test.ts
+++ b/packages/core/src/test/shared/errors.test.ts
@@ -492,6 +492,9 @@ describe('util', function () {
             false,
             'Incorrectly indicated as network error'
         )
+        const err = new Error("Unexpected token '<'")
+        err.name = 'SyntaxError'
+        assert.deepStrictEqual(isNetworkError(err), false, 'Incorrectly indicated as network error')
     })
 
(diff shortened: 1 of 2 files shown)


packages/core/src/test/shared/errors.test.ts
```diff
@@ -485,13 +485,20 @@ describe('util', function () {
...
+            'Did not VS Code Proxy error as network error'
         )
         assert.deepStrictEqual(
             isNetworkError(new Error('I am NOT a network error')),
             false,
             'Incorrectly indicated as network error'
         )
+        let err = new Error("Unexpected token '<'")
+        err.name = 'SyntaxError'
+        assert.deepStrictEqual(isNetworkError(err), true, 'Did not indicate SyntaxError as network error')
```
justinmk3, line 497: Is the extra "Did not indicate..." message giving any insight beyond what is already expressed by the assert?

Issue #5431, opened 2024-08-08
@workspace attempts to index python virtual envs (.venv, venv)


wilmatos:
## Problem
When triggering the indexing of a project, if the .venv or venv directories from a python virtual environment is not listed in the .gitignore file, the workspace feature tries to index all the files in those directories causing it to take a really long time OR fail due to size.

## Steps to reproduce the issue
In a workspace in vscode:
1. create a python virtual environment using virtualenv or venv with the dir name of .venv or venv.  
2. Confirm that the folder or directory is not listed in the .gitignore file.
3. Open Amazon Q Chat by clicking on Q icon or clicking on Amazon Q in status bar and selecting 'Open Chat'
4. Type `@workspace` in chat and hit enter (or send the prompt)
5. It will trigger indexing.
6. Open the Output panel
7. Select 'Amazon Q Language Server'
8. Observe it attempting to index the virtual env directory.

## Expected behavior
The indexing process should ignore the venv or .venv directory by default.

## System details (run `AWS: About` and/or `Amazon Q: About`)
-   OS: Sonoma 14.6
-   Visual Studio Code version: 1.87.2 (Universal)
-   AWS Toolkit version: v3.18.0
-   Amazon Q version: v1.19.0

jpinkney-aws:
In the latest release we are ignoring venv files during local indexing!

PR #5433, opened 2024-08-08
fix(amazonq): ignore venv when local indexing

leigaol:
## Problem
https://github.com/aws/aws-toolkit-vscode/issues/5431

## Solution

Ignore .venv and venv at local indexing. We will build more robust ignore mechanism in the LSP soon. This is a stopgap for now.

## License

### DIFF
--- a/packages/core/src/shared/utilities/workspaceUtils.ts
+++ b/packages/core/src/shared/utilities/workspaceUtils.ts
@@ -552,7 +552,7 @@ export async function collectFilesForIndex(
     }
 
     const isBuildOrBin = (filePath: string) => {
-        const k = /[/\\](bin|build|node_modules|env|\.idea)[/\\]/i
+        const k = /[/\\](bin|build|node_modules|env|\.idea|\.venv|venv)[/\\]/i
         return k.test(filePath)
     }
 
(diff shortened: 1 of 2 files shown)


jpinkney-aws (approved):

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -552,7 +552,7 @@ export async function collectFilesForIndex(
     }
 
     const isBuildOrBin = (filePath: string) => {
-        const k = /[/\\](bin|build|node_modules|env|\.idea)[/\\]/i
+        const k = /[/\\](bin|build|node_modules|env|\.idea|\.venv|venv)[/\\]/i
```
rli, line 555: do we need to worry about cases where the user has a non-conventional virtualenv path?

justinmk3, line 555: @leigaol why don't we use vscode's `'files.exclude'` and/or `'search.exclude'` settings? That has these benefits:

1. allows customer to control the excludes
2. benefits from vscode default exclude patterns (which include what you have here plus more)

PR #5438, opened 2024-08-09
DynamoDB: Pagination in Query Panel

Lokesh1302:
## Problem

- Currently there is no pagination for the items in query panel

## Solution

- Added pagination for the items in query panel
image

## License

### DIFF
--- a/packages/core/package.nls.json
+++ b/packages/core/package.nls.json
@@ -267,7 +267,7 @@
     "AWS.amazonq.codewhisperer.title": "Amazon Q",
     "AWS.amazonq.toggleCodeSuggestion": "Toggle Auto-Suggestions",
     "AWS.amazonq.toggleCodeScan": "Toggle Auto-Scans",
-    "AWS.command.dynamoDb.searchTables": "Search Tables",
+    "AWS.command.dynamoDb.searchTables": "Search DynamoDB Tables",
     "AWS.command.dynamoDb.viewTable": "View Table",
     "AWS.command.dynamoDb.deleteTable": "Delete Table",
     "AWS.command.dynamoDb.openTableInConsole": "Open Table in Console"
--- a/packages/core/package.nls.json
+++ b/packages/core/package.nls.json
@@ -9,7 +9,7 @@
     "AWS.configuration.description.lambda.recentlyUploaded": "Recently selected Lambda upload targets.",
     "AWS.configuration.description.ecs.openTerminalCommand": "The command to run when starting a new interactive terminal session.",
     "AWS.configuration.description.iot.maxItemsPerPage": "Controls how many IoT Things, Certificates, or Policies are listed before showing a node to `Load More...`.",
-    "AWS.configuration.description.dynamodb.maxItemsPerPage": "Controls how many items to show per page in dynamoDb table view.",
+    "AWS.configuration.description.dynamodb.maxItemsPerPage": "Controls how many items to show per page in DynamoDB table view.",
     "AWS.configuration.description.s3.maxItemsPerPage": "Controls how many S3 items are listed before showing a node to `Load More...`.\nThis corresponds to the `MaxKeys` requested in a single call to S3. [Learn More](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#AmazonS3-ListObjectsV2-response-MaxKeys)",
     "AWS.configuration.description.samcli.lambdaTimeout": "Maximum time (in milliseconds) to wait for SAM output while starting a Local Lambda session",
     "AWS.configuration.description.samcli.location": "Location of SAM CLI. SAM CLI is used to create, build, package, and deploy Serverless Applications. [Learn More](https://aws.amazon.com/serverless/sam/)",
@@ -270,5 +270,5 @@
     "AWS.command.dynamoDb.searchTables": "Search DynamoDB Tables",
     "AWS.command.dynamoDb.viewTable": "View Table",
     "AWS.command.dynamoDb.deleteTable": "Delete Table",
-    "AWS.command.dynamoDb.openTableInConsole": "Open Table in Console"
+    "AWS.command.dynamoDb.openTableInConsole": "Open table in browser"
 }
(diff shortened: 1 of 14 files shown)


packages/core/src/dynamoDb/utils/dynamodb.ts
```diff
@@ -242,3 +248,13 @@ export async function deleteItem(
...
 }
+
+async function getMaxItemsPerPage(): Promise<number> {
+    return Settings.instance.getSection('aws').get<number>('dynamodb.maxItemsPerPage', 100)
+}
+
+export function dynamoDbConsoleUrl(node: DynamoDbTableNode): vscode.Uri {
+    const service = 'dynamodb'
+    const resourcePath = `tables:selected=${node.dynamoDbtable}`
+    return vscode.Uri.parse(`https://console.aws.amazon.com/${service}/home?region=${node.regionCode}#/${resourcePath}`)
```
jpinkney-aws, line 259: I'm not sure if this will work in regions like china btw. I could be wrong but I don't think aws exists there

packages/core/src/dynamoDb/utils/editItem.ts
```diff
@@ -0,0 +1,158 @@
...
+export async function editItem(
+    selectedRow: RowData,
+    { tableName, regionCode, tableSchema }: { tableName: string; regionCode: string; tableSchema: TableSchema }
+) {
+    const currentContent = JSON.stringify(selectedRow, undefined, 4)
+    let filename = `${selectedRow[tableSchema.partitionKey.name]}`
+    if (tableSchema.sortKey) {
+        filename += `-${selectedRow[tableSchema.sortKey.name]}`
+    }
+    const tempFilePath = path.join(os.tmpdir(), `${filename}.json`)
```
jpinkney-aws, line 24: If we have time I wonder if we should re-visit if we can figure out a virtual file

packages/core/src/dynamoDb/utils/editItem.ts
```diff
@@ -0,0 +1,158 @@
...
+
+async function saveItem(
+    tempFilePath: string,
+    selectedRow: RowData,
+    { tableName, regionCode, tableSchema }: { tableName: string; regionCode: string; tableSchema: TableSchema }
+) {
+    try {
+        const updatedContent = JSON.parse(fs.readFileSync(tempFilePath, 'utf-8'))
+        const response = await updateDynamoDbItem(updatedContent, selectedRow, { tableName, regionCode, tableSchema })
+        if (response) {
```
jpinkney-aws, line 56: We probably don't need to notify users if the item was successfully updated because thats already their expectation + we can see it updated in the ui

jpinkney-aws (commented):
I have a cou

packages/core/package.nls.json
```diff
@@ -9,6 +9,7 @@
     "AWS.configuration.description.lambda.recentlyUploaded": "Recently selected Lambda upload targets.",
     "AWS.configuration.description.ecs.openTerminalCommand": "The command to run when starting a new interactive terminal session.",
     "AWS.configuration.description.iot.maxItemsPerPage": "Controls how many IoT Things, Certificates, or Policies are listed before showing a node to `Load More...`.",
+    "AWS.configuration.description.dynamodb.maxItemsPerPage": "Controls how many items to show per page in dynamoDb table view.",
```
jpinkney-aws, line 12: I think this should be:
```suggestion
    "AWS.configuration.description.dynamodb.maxItemsPerPage": "Controls how many items to show per page in dynamoDB table view.",
```

or 
```suggestion
    "AWS.configuration.description.dynamodb.maxItemsPerPage": "Controls how many items to show per page in DynamoDB table view.",
```

packages/core/package.nls.json
```diff
@@ -266,7 +267,8 @@
     "AWS.amazonq.codewhisperer.title": "Amazon Q",
     "AWS.amazonq.toggleCodeSuggestion": "Toggle Auto-Suggestions",
     "AWS.amazonq.toggleCodeScan": "Toggle Auto-Scans",
-    "AWS.command.dynamoDb.searchTables": "Search Tables",
+    "AWS.command.dynamoDb.searchTables": "Search DynamoDB Tables",
     "AWS.command.dynamoDb.viewTable": "View Table",
-    "AWS.command.dynamoDb.deleteTable": "Delete Table"
+    "AWS.command.dynamoDb.deleteTable": "Delete Table",
+    "AWS.command.dynamoDb.openTableInConsole": "Open Table in Console"
```
jpinkney-aws, line 273: I think someone mentioned this in the bug bash, but I wonder if it makes sense to have this be "Open Table in AWS Console"

packages/core/src/dynamoDb/utils/editItem.ts
```diff
@@ -0,0 +1,158 @@
...
+                Value: {
+                    [attributeMap.key]: updatedItem[key],
+                },
+            }
+        }
+    }
+    return expressionAttributeValues
+}
+
+function compareRowDataItems(updatedItem: RowData, currentItem: RowData) {
```
jpinkney-aws, line 125: I wonder if we updated this function name we could make this more clear. Maybe something like areRowDataKeysEqual? Because I think its just comparing the keys

packages/core/src/dynamoDb/utils/editItem.ts
```diff
@@ -0,0 +1,158 @@
...
+    updatedItem: RowData,
+    currentItem: RowData,
+    { tableName, regionCode, tableSchema }: { tableName: string; regionCode: string; tableSchema: TableSchema },
+    client: DynamoDbClient = new DynamoDbClient(regionCode)
+) {
+    const partitionKey = tableSchema.partitionKey.name
+    const sortKey = tableSchema.sortKey?.name
+
+    if (updatedItem[partitionKey] !== currentItem[partitionKey]) {
+        void vscode.window.showErrorMessage(`Partition key '${partitionKey}' cannot be updated.`)
```
jpinkney-aws, line 73: Can we change this error message to explain more of why they cannot update their partition key? I was confused yesterday when I saw this error. I didn't realize I couldn't change it (though it totally makes sense why I couldn't)

Maybe something like: Partition keys are unable to be updated? Or DynamoDB does not allow updating partition keys. I think the same applies to the error message below as well

packages/core/src/dynamoDb/utils/editItem.ts
```diff
@@ -0,0 +1,158 @@
...
+        return
+    }
+
+    if (sortKey && updatedItem[sortKey] !== currentItem[sortKey]) {
+        void vscode.window.showErrorMessage(`Sort key '${sortKey}' cannot be updated.`)
+        return
+    }
+
+    if (!compareRowDataItems(updatedItem, currentItem)) {
```
jpinkney-aws, line 82: We might be able to throw an error here to tell the users that you can't add new keys in

packages/core/package.nls.json
Lokesh1302: Changed to "Open table in browser"

PR #5439, opened 2024-08-09
feat(telemetry): add registration info to auth events

hayemaxi:
**Depends on https://github.com/aws/aws-toolkit-common/pull/810**
- For easier analysis, auth events (auth_userState, auth_addConnection, aws_refreshCredentials, aws_loginWithBrowser) will now contain start url, region, registration client id, and registration expiration time.

## License


package.json
```diff
@@ -39,6 +39,7 @@
         "generateNonCodeFiles": "npm run generateNonCodeFiles -w packages/ --if-present"
     },
     "devDependencies": {
+        "@aws-toolkits/telemetry": "^1.0.239",
```
nkomonen-amazon, line 42: I think telemetry is fine as a dev dependency, but was this intentional

packages/amazonq/src/extension.ts
```diff
@@ -149,10 +149,16 @@ export async function activateAmazonQCommon(context: vscode.ExtensionContext, is
         }
 
         const authState = (await AuthUtil.instance.getChatAuthState()).codewhispererChat
+        const currConn = AuthUtil.instance.conn as SsoConnection
```
nkomonen-amazon, line 152: I'm not sure in what cases but `.conn` can be an IAM connection, see `AuthUtil.getCredentials()`.

Look to return early or throw if `isAnySsoConnection()` is false. This should also automatically cast the conn to an SsoConnection for you

packages/core/src/codecatalyst/auth.ts
```diff
@@ -294,7 +294,8 @@ export class CodeCatalystAuthenticationProvider {
             await this.promptOnboarding()
         }
 
-        return this.secondaryAuth.useNewConnection(conn)
+        await this.secondaryAuth.useNewConnection(conn)
```
nkomonen-amazon, line 297: Can we change `useNewConnection` to return the given conn instead?

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -201,7 +201,8 @@ export class AuthUtil {
             conn = await this.auth.reauthenticate(conn)
         }
 
-        return this.secondaryAuth.useNewConnection(conn)
+        await this.secondaryAuth.useNewConnection(conn)
+        return conn
     }
 
     public async connectToEnterpriseSso(startUrl: string, region: string) {
```
nkomonen-amazon, line 208: Might be worth adding an explicit return type since it has changed from `void` to `SsoConnection`

package.json
hayemaxi: Yes, its only needed to generate definitions right? So only for building.

packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
```diff
@@ -56,7 +56,13 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
                 isReAuth: false,
             })
 
-            await awsIdSignIn()
+            const conn = await awsIdSignIn()
+            const registration = await conn.getRegistration()
+            this.storeMetricMetadata({
```
nkomonen-amazon, line 61: Does this make sense to do in `AuthUtil` instead? Under the hood this eventuall calls `AuthUtil.connectToAwsBuilderId`.

If true, then the same would be for `connectToEnterpriseSso`

packages/core/src/login/webview/vue/amazonq/backend_amazonq.ts
```diff
@@ -56,7 +56,13 @@ export class AmazonQLoginWebview extends CommonAuthWebview {
...
+                ssoRegistrationExpiresAt: registration?.expiresAt.toISOString(),
```
nkomonen-amazon, line 62: There is some duplication of storing registration in the metric metadata. Would it make sense for a generalized function that takes in the registration as an arg and that adds it to the metadata? This file and backend_toolkit could reuse this

PR #5442, opened 2024-08-12
feat: telemetry stack trace - runWithCallEntry()

nkomonen-amazon:
See the doc added in this PR for explanation on Problem + Solution

The gist is that you can now add "options" to a `run()`, allowing configurations.
In our case:
- we can signal for run() to not emit, but it will still emit by default.
- Add a `FunctionEntry` which will add to the function execution stack that is derived from the span stack

## License

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -139,3 +139,124 @@ Finally, if `setupStep2()` was the thing that failed we would see a metric like:
     ...
 }
 ```
+
+## Adding a "Stack Trace" to your metric
+
+### Problem
+
+Common example: _"I have a function, `thisFailsSometimes()` that is called in multiple places. The function sometimes fails, I know from telemetry, but I do not know if it is failing when it is a specific caller. If I knew the call stack/trace that it took to call my function that would help me debug."_
+
+```typescript
+function outerA() {
+    thisFailsSometimes(1) // this succeeds
+}
+
+function outerB() {
+    thisFailsSometimes(0) // this fails
+}
+
+function thisFailsSometimes(num: number) {
+    return telemetry.my_Metric.run(() => {
+        if (number === 0) {
+            throw Error('Cannot be 0')
+        }
+        ...
+    })
+}
+```
+
+### Solution
+
+Add a value to `function` in the options of a `run()`. This will result in a stack of functions identifiers that were previously called
+before `thisFailsSometimes()` was run. You can then retrieve the stack in the `run()` of your final metric using `getFunctionStack()`.
+
+```typescript
+function outerA() {
+    telemetry.my_Metric.run(() => thisFailsSometimes(1), { functionId: { name: 'outerA' }})
+}
+
+function outerB() {
+    telemetry.my_Metric.run(() => thisFailsSometimes(0), { functionId: { source: 'outerB' }})
+}
+
+function thisFailsSometimes(num: number) {
+    return telemetry.my_Metric.run(() => {
+        telemetry.record({ theCallStack: asStringifiedStack(telemetry.getFunctionStack())})
+        if (number === 0) {
+            throw Error('Cannot be 0')
+        }
+        ...
+    }, { functionId: { name: 'thisFailsSometimes' }})
+}
+
+// Results in a metric: { theCallStack: 'outerB:thisFailsSometimes', result: 'Failed' }
+// { theCallStack: 'outerB:thisFailsSometimes' } implies 'outerB' was run first, then 'thisFailsSometimes'. See docstrings for more info.
+outerB()
+```
+
+### Important Notes
+
+-   If a nested function does not use a `run()` then it will not be part of the call stack.
+
+    ```typescript
+    function a() {
+        return telemetry.my_Metric.run(() => b(), { functionId: { name: 'a' } })
+    }
+
+    function b() {
+        return c()
+    }
+
+    function c() {
+        return telemetry.my_Metric.run(() => asStringifiedStack(telemetry.getFunctionStack()), {
+            functionId: { name: 'c' },
+        })
+    }
+
+    c() // result: 'a:c', note that 'b' is not included
+    ```
+
+-   If you are using `run()` with a class method, you can also add the class to the entry for more context
+
+    ```typescript
+    class A {
+        a() {
+            return telemetry.my_Metric.run(() => this.b(), { functionId: { name: 'a', class: 'A' } })
+        }
+
+        b() {
+            return telemetry.my_Metric.run(() => asStringifiedStack(telemetry.getFunctionStack()), {
+                functionId: { name: 'b', class: 'A' },
+            })
+        }
+    }
+
+    const inst = new A()
+    inst.a() // 'A#a,b'
+    ```
+
+-   If you do not want your `run()` to emit telemetry, set `emit: false` in the options
+
+    ```typescript
+    function a() {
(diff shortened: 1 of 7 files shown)


packages/core/src/shared/context.ts
```diff
@@ -0,0 +1,89 @@
...
+ *                 to be wrapped in this.
+ */
+export const withExecutionContext = (context: { source: string; class?: string }, callback: () => any) =>
+    executionContext.run(createStore(context), callback)
+
+type ExecutionContextStore = {
+    /**
+     * The stringified representation of the call stack.
+     *
+     * Example: "ClassA#methodA,methodB:ClassB#methodY,methodZ"
```
hayemaxi, line 39: How is a method without a class represented?

nkomonen-amazon: I've updated the description again. But it will just not have a `ClassX#` in front of it.

Eg: `ClassA#methodA:myFunctionWithNoClass`

nkomonen-amazon:
@hayemaxi There is potential reusability for this to add something like a `startUrl` to the `ExtensionContext` so that it can be retrieved for telemetry purposes in a later function call

hayemaxi:
So does each function need to call its login within `withExecutionContext` in order to record itself in the call stack?

nkomonen-amazon:
@hayemaxi 

> So does each function need to call its login within withExecutionContext in order to record itself in the call stack?

Yeah exactly. If it doesn't call `withExecutionContext` then it will not be in the call stack, but if a nested call does, that one will be. We will probably miss some functions initially, but over time we can add it where missing.

We previously discussed using something like `arguments.callee` but this is deprecated: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/caller

Also there were considerations for using an `Error.stack`, but there were some complexities with that such as it not working with code obfuscation and the parsing of an error stack would require an external dependency that hasn't been updated in a while.

The other above options would have required less manual work to enable the stack trace, but they have some gaps.

hayemaxi:
Makes sense. I suppose this more of a nit but I wonder if we could simplify this by borrowing from telemetry? Not sure if there is much to re-use there, but it would be nice if we could just "record" in various functions. I think it would be a lot easier to follow.

Something like this:
```
function login() {
    executionContext.record(this.functionName) // call a record function. If there is a an active context, record the stack. Otherwise do nothing, maybe log an error?
    createProfile()
}

function createProfile() {
     executionContext.record(this.functionName)
     ...
}

function startLoginFlow() {
    return withExecutionContext(async () => {
        executionContext.record(this.functionName)
        login()
    })
}
```
I suppose one downside is that you have to make sure you call `record` before doing anything else or it might record in the wrong order. The original method would make it easier to keep track of that.

hayemaxi (approved):

justinmk3:
> I wonder if we could simplify this by borrowing from telemetry? Not sure if there is much to re-use there, but it would be nice if we could just "record" in various functions.

Yes. That was the original (and still current) intention of `TelemetryTracer` https://github.com/aws/aws-toolkit-vscode/blob/408befebd22af28e9759bcc66c50f49f5fbb6c47/packages/core/src/shared/telemetry/spans.ts#L203-L205

why can't we re-use that? I don't consider this a nit, rather it's a blocker.

nkomonen-amazon:
@justinmk3 @jpinkney-aws 

I've synced with Josh and the latest solution works for his case.
Justin please give a look when you have time.

docs/telemetry.md
```diff
@@ -139,3 +139,124 @@ Finally, if `setupStep2()` was the thing that failed we would see a metric like:
...
+                function: { name: 'b', class: 'A' },
+            })
+        }
+    }
+
+    const inst = new A()
+    inst.a() // 'A#a,b'
+    ```
+
+-   If you do not want your `run()` to emit, set `emit: false` in the options
```
jpinkney-aws, line 238: nit:
```suggestion
-   If you do not want your `run()` to emit telemetry, set `emit: false` in the options
```

docs/telemetry.md
```diff
@@ -139,3 +139,124 @@ Finally, if `setupStep2()` was the thing that failed we would see a metric like:
...
+
+    ```typescript
+    function a() {
+        return telemetry.my_Metric.run(() => b(), { function: { name: 'a' }, emit: false })
+    }
+    ```
+
+-   If you want to add to the function stack, but don't have a specific metric to use,
+    use the metric named `function_call`. Also look to set `emit: false` if you do not
+    want it to emit.
```
jpinkney-aws, line 248: nitpick:
```suggestion
    want it to emit telemetry.
```

packages/core/src/shared/telemetry/spans.ts
```diff
@@ -90,6 +90,40 @@ function getValidatedState(state: Partial<MetricBase>, definition: MetricDefinit
...
+     *
+     * function c() {
+     *   telemetry.your_Metric.run(() => {
+     *     const stack = telemetry.getFunctionStack()
+     *     console.log(stack) // [ {source: 'a' }, { source: 'b' }, { source: 'c' }]
+     *   }, { function: { name: 'c'} })
+     * }
+     * ```
+     */
+    function?: FunctionEntry
```
jpinkney-aws, line 124: I wonder if this could be functionId? So a function ids is a a single entry in a stack of function ids? naming is hard 🙃

packages/core/src/shared/telemetry/spans.ts
```diff
@@ -113,7 +148,17 @@ export class TelemetrySpan<T extends MetricBase = MetricBase> {
      */
     static readonly #excludedFields = ['passive', 'value']
 
-    public constructor(public readonly name: string) {}
+    public constructor(
+        public readonly name: string,
+        options?: SpanOptions
+    ) {
+        // set defaults on undefined options
+        this.#options = {
```
jpinkney-aws, line 156: can we add some tests for the emit option as well

nkomonen-amazon: done

packages/core/src/shared/telemetry/spans.ts
nkomonen-amazon: done

jpinkney-aws (approved):

packages/core/src/shared/telemetry/spans.ts
```diff
@@ -90,6 +90,40 @@ function getValidatedState(state: Partial<MetricBase>, definition: MetricDefinit
...
+     *
+     * function c() {
+     *   telemetry.your_Metric.run(() => {
+     *     const stack = telemetry.getFunctionStack()
+     *     console.log(stack) // [ {source: 'a' }, { source: 'b' }, { source: 'c' }]
+     *   }, { functionId: { name: 'c'} })
+     * }
+     * ```
+     */
+    functionId?: FunctionEntry
```
justinmk3, line 124: why can't this be a field on the `function_call` metric?

justinmk3, line 124: I don't quite get why SpanOptions is needed. Why can't its fields live on the `function_call` metric itself?

We have a story to better support "nested" metrics: https://sim.amazon.com/issues/IDE-10178

With that support, each `function_call` could be a child of the containing (parent) `function_call`. And then the "stack" could be derived that way. 

And then we avoid this extra alternative concept.

PR #5443, opened 2024-08-13
feat(amazonq): support ab testing by reusing CodeWhisperer

ivikash:
## Problem

Enable A/B testing for AWS Toolkits and Amazon Q by reusing CodeWhisperer's A/B testing platform

## Solution
- Abstract out Feature Config from CodeWhisperer to Shared folder. 

## License


justinmk3 (approved):
Nice choices!

ivikash:
Remove Telemetry in the PR after https://github.com/aws/aws-toolkit-common/pull/820 is merged.

packages/core/src/extension.ts
```diff
@@ -194,6 +195,9 @@ export function registerGenericCommands(extensionContext: vscode.ExtensionContex
         }),
         Commands.register(`aws.${contextPrefix}.aboutExtension`, async () => {
             await aboutExtension()
+        }),
+        Commands.register(`aws.${contextPrefix}.fetchFeatureConfigs`, async () => {
+            await FeatureConfigProvider.instance.fetchFeatureConfigs()
```
justinmk3, line 200: why is this a command instead of a plain old function? I don't see any cases in the codebase that require it to be a command. Is this intended for users to invoke from the cmd palette?

ivikash: The intention is to allow customer to do `await vscode.commands.executeCommand('aws.amazonq.fetchFeatureConfigs')` or `await vscode.commands.executeCommand('aws.toolkit.fetchFeatureConfigs')`. 

May be we could just make it a function instead.

packages/core/src/test/mockFeatureConfigData.ts
justinmk3, line 1: we have the `packages/core/src/test/fake/` dir which is intended for this kind of thing.

packages/amazonq/test/unit/codewhisperer/service/featureConfigProvider.test.ts
justinmk3, line 1: should this file be moved to `packages/core/test/unit/shared/featureConfig.test.ts` ?

packages/core/src/codewhisperer/activation.ts
```diff
@@ -594,6 +592,8 @@ export async function activate(context: ExtContext): Promise<void> {
         )
     }
 
+    void FeatureConfigProvider.instance.fetchFeatureConfigs()
```
justinmk3, line 595: why not await this? if the intention is to do the work asynchronously, use `.catch()` so that failures are at least logged

ivikash: I will add a catch so that even if the API is down the extension is not impacted in the critical path.

packages/amazonq/test/unit/codewhisperer/service/featureConfigProvider.test.ts
ivikash: Moved to `packages/core/src/test/shared/featureConfig.test.ts`

justinmk3 (approved):

justinmk3:
ci failure:

```
Error: Cannot find module '.../aws-toolkit-vscode/packages/core/dist/dist/src/codewhisperer/index.js'
```

ivikash:
> ci failure. is there a bad `import` somewhere?
> 
> ```
> Error: Cannot find module '.../aws-toolkit-vscode/packages/core/dist/dist/src/codewhisperer/index.js'
> ```

Fixed here - https://github.com/aws/aws-toolkit-vscode/pull/5506

justinmk3 (approved):

justinmk3:
ci still failing

PR #5446, opened 2024-08-13
performance: Create a starter set of tools for measuring the CPU/memory impact of code

jpinkney-aws:
## Problem
- We are sometimes running into problems where functions are using a ton of memory/cpu but we have no way to figure that out ahead of time without manually debugging

## Solution
- Create a starter set of tools that allows us to measure the impact of different parts of code. These tools should be used as needed and surround heavy computational parts of the code i.e. collecting all files in the workspace, reading them in and putting them in a zip

See:
- [Security Scans] Adding performance test for auto scans: https://github.com/aws/aws-toolkit-vscode/pull/5420
- [Amazon Q Security Scans] Refactoring code to improve the performance: https://github.com/aws/aws-toolkit-vscode/pull/5346#

## License

### DIFF
--- a/packages/core/src/shared/performance/performance.ts
+++ b/packages/core/src/shared/performance/performance.ts
@@ -1,0 +1,118 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'assert'
+import { getLogger } from '../logger'
+import { isWeb } from '../extensionGlobals'
+
+interface PerformanceMetrics {
+    cpuUsage: number
+    heapTotal: number
+    duration: number
+}
+
+interface TestOptions {
+    darwin?: Partial<PerformanceMetrics>
+    win32?: Partial<PerformanceMetrics>
+    linux?: Partial<PerformanceMetrics>
+}
+
+export interface PerformanceSpan<T> {
+    value: T
+    performance: PerformanceMetrics
+}
+
+export class PerformanceTracker {
+    #startPerformance:
+        | {
+              cpuUsage: NodeJS.CpuUsage
+              memory: number
+              duration: [number, number]
+          }
+        | undefined
+
+    constructor(private readonly name: string) {}
+
+    static enabled(name: string, trackPerformance?: boolean): boolean {
+        return name === 'function_call' && (trackPerformance ?? false) && !isWeb()
+    }
+
+    start() {
+        this.#startPerformance = {
+            cpuUsage: process.cpuUsage(),
+            memory: process.memoryUsage().heapTotal,
+            duration: process.hrtime(),
+        }
+    }
+
+    stop(): PerformanceMetrics | undefined {
+        if (this.#startPerformance) {
+            const endCpuUsage = process.cpuUsage(this.#startPerformance?.cpuUsage)
+            const userCpuUsage = endCpuUsage.user / 1000000
+            const systemCpuUsage = endCpuUsage.system / 1000000
+
+            const elapsedTime = process.hrtime(this.#startPerformance.duration)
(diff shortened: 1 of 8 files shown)


justinmk3:
see also https://github.com/aws/aws-toolkit-vscode/pull/3315

packages/core/src/shared/telemetry/spans.ts
```diff
@@ -195,6 +205,10 @@ export class TelemetrySpan<T extends MetricBase = MetricBase> {
      */
     public start(): this {
         this.#startTime = new globals.clock.Date()
+        if (this.#options.trackPerformance) {
+            this.#performance?.start()
```
hayemaxi, line 209: With other suggestion:
```suggestion
            (this.#performance ??= new PerformanceTracker(this.name)).start()
```

packages/core/src/shared/telemetry/spans.ts
```diff
@@ -157,6 +162,11 @@ export class TelemetrySpan<T extends MetricBase = MetricBase> {
             // do emit by default
             emit: options?.emit === undefined ? true : options.emit,
             functionId: options?.functionId,
+            trackPerformance: PerformanceTracker.enabled(this.name, options?.trackPerformance),
+        }
+
+        if (this.#options.trackPerformance) {
+            this.#performance = new PerformanceTracker(name)
         }
```
hayemaxi, line 170:
```suggestion
```

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -275,6 +275,21 @@
             "name": "connectionState",
             "type": "string",
             "description": "A detailed state of a specific auth connection. Use `authStatus` for a higher level view of an extension's general connection."
+        },
+        {
+            "name": "callStack",
```
hayemaxi, line 280: How will this be used? Seems compatible (or should be compatible) with https://github.com/aws/aws-toolkit-vscode/pull/5442

jpinkney-aws: I've removed it for now but ideally we would be able to use it to help find the call path. IIRC Nik's changes require the functions to manually set their function name/class name so we can't use it for free

packages/core/src/shared/performance/performance.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export function performanceTest(options: TestOptions, name: string, fn: () => Promise<void>): Mocha.Test
+export function performanceTest(options: TestOptions, name: string, fn: () => void): Mocha.Test
+export function performanceTest(options: TestOptions, name: string, fn: () => void | Promise<void>) {
+    const testOption = options[process.platform as 'linux' | 'darwin' | 'win32']
+
+    const performanceTracker = new PerformanceTracker(name)
+
+    return it(name, async () => {
```
justinmk3, line 81: Just to double-check... we didn't (and won't) promote mocha to a "production" dependency, right?

packages/core/src/shared/telemetry/spans.ts
```diff
@@ -97,6 +98,9 @@ export type SpanOptions = {
     /** True if this span should emit its telemetry events. Defaults to true if undefined. */
     emit?: boolean
 
+    /** True if this span should emit performance metrics acquired when running the function. Defaults to false if undefined */
+    trackPerformance?: boolean
```
justinmk3, line 102: Is there a reason this can't be a field on the metric itself (or MetricBase or whatever)? I don't quite understand why we need SpanOptions. We treat `passive` like an "option", so why not have other fields that work like that?

justinmk3 (approved):

packages/core/src/shared/performance/performance.ts
jpinkney-aws: no, this is only for tests. Ideally telemetry tells us the rough runtime of functions then we can use that data to reduce the options in the performance test. That way we can see regressions as well

PR #5447, opened 2024-08-13
fix(amazonq): Correct indent when insertAtCursor

leigaol:
## Problem

https://github.com/user-attachments/assets/2bba03c8-e833-4b75-9d51-57b883df2544

When inserting code from Q chat, the indentation is wrong.

## Solution

Pad recommendation with correct indentation

https://github.com/user-attachments/assets/1ed0024d-bb47-43dd-8796-0f1eee5d114a

Note: This solution cannot solve cases when code suggestion contains a new line raw string in the string variable but it should be an improvement over existing experience.

## License

### DIFF
--- a/packages/core/src/amazonq/commons/controllers/contentController.ts
+++ b/packages/core/src/amazonq/commons/controllers/contentController.ts
@@ -28,17 +28,17 @@ export class EditorContentController {
             if (indent.trim().length !== 0) {
                 indent = ' '.repeat(indent.length - indent.trimStart().length)
             }
-            let textWithIntent = ''
+            let textWithIndent = ''
             text.split('\n').forEach((line, index) => {
                 if (index === 0) {
-                    textWithIntent += line
+                    textWithIndent += line
                 } else {
-                    textWithIntent += '\n' + indent + line
+                    textWithIndent += '\n' + indent + line
                 }
             })
             editor
                 .edit((editBuilder) => {
-                    editBuilder.insert(cursorStart, textWithIntent)
+                    editBuilder.insert(cursorStart, textWithIndent)
                 })
                 .then(
                     (appliedEdits) => {
--- a/packages/core/src/amazonq/commons/controllers/contentController.ts
+++ b/packages/core/src/amazonq/commons/controllers/contentController.ts
@@ -51,7 +51,6 @@ export class EditorContentController {
                         getLogger().error('TextEditor.edit failed: %s', (e as Error).message)
                     }
                 )
-            await Commands.tryExecute('editor.action.format')
         }
     }
 }
--- a/packages/core/src/amazonq/commons/controllers/contentController.ts
+++ b/packages/core/src/amazonq/commons/controllers/contentController.ts
@@ -3,20 +3,43 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import * as vscode from 'vscode'
 import { Position, TextEditor, window } from 'vscode'
 import { getLogger } from '../../../shared/logger'
+import { Commands } from '../../../shared'
 
 export class EditorContentController {
-    public insertTextAtCursorPosition(
+    /* *
+     *  Insert the Amazon Q chat written code to the cursor position
+     *  Add current intentation to the next few lines of the recommendation
+     * @param text the raw text from Amazon Q chat
+     * @param trackCodeEdit callback to track user edits
+     */
+    public async insertTextAtCursorPosition(
         text: string,
         trackCodeEdit: (editor: TextEditor, cursorStart: Position) => void
     ) {
         const editor = window.activeTextEditor
         if (editor) {
             const cursorStart = editor.selection.active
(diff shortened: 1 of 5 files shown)


packages/core/src/amazonq/commons/controllers/contentController.ts
```diff
@@ -3,20 +3,43 @@
...
         trackCodeEdit: (editor: TextEditor, cursorStart: Position) => void
     ) {
         const editor = window.activeTextEditor
         if (editor) {
             const cursorStart = editor.selection.active
+            const indentRange = new vscode.Range(new vscode.Position(cursorStart.line, 0), cursorStart)
+            // use the user editor intent if the position to the left of cursor is just space or tab
+            // otherwise indent with empty space equal to the intent at this position
+            let indent = editor.document.getText(indentRange)
+            if (indent.trim().length !== 0) {
```
andrewyuq, line 29: also do trimStart() here?

leigaol: If string.trimStart() is non-empty, string.trim() will also be non-empty. Same .

leigaol:
Unit test failure comes from `startSecurityScanPerformanceTest`

packages/core/src/amazonq/commons/controllers/contentController.ts
leigaol: The intention is to detect if the indent is purely space or tab, if it is, then we use the same indent. Otherwise we fill with space of same length

packages/core/src/amazonq/commons/controllers/contentController.ts
```diff
@@ -3,20 +3,42 @@
...
+                indent = ' '.repeat(indent.length - indent.trimStart().length)
+            }
+            let textWithIntent = ''
```
zixlin7, line 31: nit: `textWithIndent`?

zixlin7 (approved):

packages/core/src/amazonq/commons/controllers/contentController.ts
andrewyuq: There are cases when this range has trailing spaces, is it working expected in this logic?

leigaol: https://github.com/user-attachments/assets/e977871a-47e3-4eae-86dd-b980e1307ba3

Yes here is one example when the range has trailing spaces.

jpinkney-aws:
Is it possible to add a test for this?

leigaol:
> Is it possible to add a test for this?

Done

packages/core/src/test/amazonq/common/contentController.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    afterEach(async function () {
+        await fs.remove(tempFolder)
+    })
+
+    describe('insertTextAtCursorPosition', () => {
+        it('insert code when left hand size has no non empty character', async () => {
+            const programFile = path.join(tempFolder, 'program.python')
+            await fs.writeFile(programFile, 'def hello_world():\n    ')
+            const doc = await vscode.workspace.openTextDocument(programFile)
+            const editor = await vscode.window.showTextDocument(doc)
```
jpinkney-aws, line 27: You can replace these first few lines with `openATextEditorWithText` from `core/src/test/testUtil`

See editorUtilities.test.ts, it has a lot of examples of using it

packages/core/src/test/amazonq/common/contentController.test.ts
```diff
@@ -0,0 +1,44 @@
...
+import { openATextEditorWithText } from '../../testUtil'
+
+describe('contentController', () => {
+    let controller: EditorContentController
+
+    beforeEach(async () => {
+        controller = new EditorContentController()
+    })
+
+    afterEach(async function () {})
```
jpinkney-aws, line 13: nit: you don't need to define this if you aren't using it

packages/core/src/test/amazonq/common/contentController.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    describe('insertTextAtCursorPosition', () => {
+        it('insert code when left hand size has no non empty character', async () => {
+            const editor = await openATextEditorWithText('def hello_world():\n    ', 'test.py')
+            if (editor) {
```
jpinkney-aws, line 18: if editor isn't found for some reason can we fail the test, since we won't actually know if function execution happened successfully

jpinkney-aws (approved):

justinmk3:
linux CI is failing : 

(note: "rejected promise" is not checked in mac/win CI so that's why only linux catches it)


```
rejected promise not handled within 1 second: AssertionError [ERR_ASSERTION]: 'def abc\n   defhello_world():\n    ' == 'def abc\n    defhello_world():\n'
rejected promise not handled within 1 second: AssertionError [ERR_ASSERTION]: 'def hello_world():\n    abc\n       def' == 'def hello_world():\n    abc\n    def'

ERROR: Test output matched this pattern 3 times:
```

PR #5448, opened 2024-08-14
feat(amazonq): add session configuration notification

zixlin7:
## Problem
need to add a notification to IdC users to contact their admin to configure extended session
## Solution
show the notification once to IdC users when their connection expired

## License

### DIFF
--- a/packages/core/src/codewhisperer/util/authUtil.ts
+++ b/packages/core/src/codewhisperer/util/authUtil.ts
@@ -349,7 +349,7 @@ export class AuthUtil {
     public async notifySessionConfiguration() {
         const suppressId = 'amazonQSessionConfigurationMessage'
         const settings = AmazonQPromptSettings.instance
-        const shouldShow = await settings.isPromptEnabled(suppressId as any)
+        const shouldShow = await settings.isPromptEnabled(suppressId)
         if (!shouldShow) {
             return
         }
@@ -367,7 +367,7 @@ export class AuthUtil {
                 await openUrl(learnMoreUrl)
             }
         })
-        await settings.disablePrompt(suppressId as any)
+        await settings.disablePrompt(suppressId)
     }
 
     public async notifyReauthenticate(isAutoTrigger?: boolean) {
--- a/packages/core/src/codewhisperer/util/authUtil.ts
+++ b/packages/core/src/codewhisperer/util/authUtil.ts
@@ -35,6 +35,9 @@ import { showReauthenticateMessage } from '../../shared/utilities/messages'
 import { showAmazonQWalkthroughOnce } from '../../amazonq/onboardingPage/walkthrough'
 import { setContext } from '../../shared/vscode/setContext'
 import { isInDevEnv } from '../../shared/vscode/env'
+import { openUrl } from '../../shared/utilities/vsCodeUtils'
+import * as nls from 'vscode-nls'
+const localize = nls.loadMessageBundle()
 
 /** Backwards compatibility for connections w pre-chat scopes */
 export const codeWhispererCoreScopes = [...scopesCodeWhispererCore]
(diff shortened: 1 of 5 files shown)


packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -343,6 +346,30 @@ export class AuthUtil {
         }
     }
 
+    public async notifySessionConfiguration() {
+        const suppressId = 'amazonQSessionConfigurationMessage'
```
leigaol, line 350: This should be under typeof amazonQPromptName, see `type amazonQPromptName = keyof typeof amazonQPrompts`

leigaol: By doing so you don't have to do supressId as any

zixlin7: right I forgot to remove any type, done

leigaol (approved):

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -345,6 +348,30 @@ export class AuthUtil {
...
+        const settings = AmazonQPromptSettings.instance
+        const shouldShow = await settings.isPromptEnabled(suppressId)
+        if (!shouldShow) {
+            return
+        }
+
+        const message = localize(
```
jpinkney-aws, line 359: This is probably a stupid question but will this show up to users who already have extended sessions on their IdC enabled?

zixlin7: yes it will, so we will only show it once as a FYI

jpinkney-aws: Can we detect if an IdC already has extended session enabled?

I'm guessing the team that owns the IdC might not be the team necessarily using this extension so having a dev (or multiple devs) message the IdC team to ask if extended sessions it enabled could annoy them

zixlin7: the session duration is controlled by the Q admin, product's idea is that users can contact their admin to configure the extended session if this is something that they want as end users

jpinkney-aws: Right, I was more concerned with notifying customers who already have extended session duration enabled. AFAICT we will also be notifying them and saying "hey tell your admin to enable extended session duration" even though its already enabled for them

zixlin7: agreed, but I don't think we need to get into checking end user's session configuration just for the notification purpose, it's something that they can easily check with their admin. This is what we aligned with product.

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -345,6 +348,30 @@ export class AuthUtil {
...
+            'aws.amazonq.sessionConfiguration.message',
+            'Your maximum session length for Amazon Q can be extended to 90 days by your administrator. For more information, refer to How to extend the session duration for Amazon Q in the IDE in the IAM Identity Center User Guide.'
+        )
+
+        const learnMoreUrl = vscode.Uri.parse(
+            'https://docs.aws.amazon.com/singlesignon/latest/userguide/configure-user-session.html#90-day-extended-session-duration'
+        )
+        await vscode.window.showInformationMessage(message, localizedText.learnMore).then(async (resp) => {
```
hayemaxi, line 367:
```suggestion
        void vscode.window.showInformationMessage(message, localizedText.learnMore).then(async (resp) => {
```

Q activation will be blocked on the user interacting with this message (close, click a button). If they just ignore it or minimize the notifications then this await will never resolve.

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -345,6 +348,30 @@ export class AuthUtil {
         }
     }
 
+    public async notifySessionConfiguration() {
```
hayemaxi, line 351: add telemetry?
[tookit_showNotification](https://github.com/aws/aws-toolkit-vscode/blob/2f5ff3a1e6304398437436819d6d16e21335eac6/packages/core/src/extensionNode.ts#L258)
[toolkit_invokeAction](https://github.com/aws/aws-toolkit-vscode/blob/2f5ff3a1e6304398437436819d6d16e21335eac6/packages/core/src/extensionNode.ts#L276)

jpinkney-aws (approved):

PR #5456, opened 2024-08-15
Add DocumentDB support

mvcninja:
## Problem

Users should be able to utilize DocumentDB features in the Explorer tab, it should list all the clusters available.
This will make DocumentDB more visible to developers.

## Solution

Users can now view, create, and delete regional clusters and instances.  Feedback is welcome.

- View clusters and instances

DocumentDB-Clusters

- Create a new cluster in a region

DocumentDB-CreateCluster

- Add an instance to a cluster

DocumentDB-CreateInstance

- Rename clusters & instances
- Start & stop a cluster
- Delete a cluster
- Delete an instance

DocumentDB-ClusterMenu

- Modify an instance class

DocumentDB-ModifyInstance

- Reboot an instance

DocumentDB-RebootInstance

## License


mvcninja:
4800 lines added are changes to `package-lock.json`.  Net code increase minus tests = 1330

packages/core/src/docdb/commands/addRegion.ts
```diff
@@ -0,0 +1,167 @@
...
+        const regionCode = response.RegionCode
+        let input: CreateDBClusterMessage
+        let clusterName = response.GlobalClusterName
+
+        try {
+            if (node instanceof DBClusterNode) {
+                // Create new global cluster from regional cluster
+                const primaryCluster = node.cluster
+
+                getLogger().info(`docdb:Creating global cluster: ${clusterName}`)
```
jpinkney-aws, line 107: I'd add a space between all these:
```suggestion
                getLogger().info(`docdb: Creating global cluster: ${clusterName}`)
```

packages/core/src/docdb/commands/addRegion.ts
```diff
@@ -0,0 +1,167 @@
...
+                    DBClusterIdentifier: response.Cluster.DBClusterIdentifier,
+                    DeletionProtection: globalCluster.DeletionProtection,
+                    Engine: globalCluster.Engine,
+                    EngineVersion: globalCluster.EngineVersion,
+                    StorageEncrypted: globalCluster.StorageEncrypted,
+                }
+            }
+
+            clusterName = response.Cluster.DBClusterIdentifier
+            getLogger().info(`docdb:Creating secondary cluster: ${clusterName} in region ${regionCode}`)
```
jpinkney-aws, line 137: I think the above comment applies to all logged info in this PR

packages/core/src/docdb/commands/deleteInstance.ts
```diff
@@ -0,0 +1,86 @@
...
+    await telemetry.docdb_deleteInstance.run(async () => {
+        if (!node) {
+            throw new ToolkitError('No node specified for DeleteInstance')
+        }
+
+        const parent = node.parent as DBClusterNode
+        const client = parent.client
+        const instanceName = node.instance.DBInstanceIdentifier ?? ''
+
+        if (node.instance.DBInstanceStatus !== 'available') {
```
jpinkney-aws, line 33: I see this snippet in at least 3 difference places (here, modifyInstance, rebootInstance, and probably a few more), can we pull it out to a function

packages/core/src/docdb/commands/renameCluster.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        try {
+            const cluster = await node.renameCluster(newClusterName)
+
+            getLogger().info('docdb:Renamed cluster: %O', cluster)
+            void vscode.window.showInformationMessage(
+                localize('AWS.docdb.renameCluster.success', 'Updated cluster: {0}', clusterName)
+            )
+
+            await sleep(1000) // wait for server to update status
```
jpinkney-aws, line 58: Is there any other way we can do this rather than just doing a sleep?

We have a util function called waitUntil that allows us to wait until some event happens. Can we use that to wait until the cluster updates the status before returning?

packages/core/src/docdb/commands/renameInstance.ts
```diff
@@ -0,0 +1,67 @@
...
+    getLogger().debug('docdb:RenameInstance called for: %O', node)
+
+    await telemetry.docdb_renameInstance.run(async () => {
+        if (!node) {
+            throw new ToolkitError('No node specified for RenameInstance')
+        }
+
+        const instanceName = node.instance.DBInstanceIdentifier
+
+        if (node.instance.DBInstanceStatus !== 'available') {
```
jpinkney-aws, line 31: Same comment as earlier, I think I've seen this same code used ~6 or 7 times so far

packages/core/src/docdb/commands/renameCluster.ts
mvcninja: @jpinkney-aws I was originally using waitUntil but the rename operation causes a 404 status later as the resource identifier is changing.  Manual testing revealed the server doesn't update the status for several seconds.

PR #5460, opened 2024-08-16
telemetry(amazonq): add raw accepted token count to code percentage

zixlin7:
## Problem

- current code percentage emitted from VSC uses the raw accepted count, when it should be the unmodified accepted token count
- emit the raw accepted count under a new field

## License


packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -240,6 +240,11 @@
             "type": "string",
             "description": "Type of chat command name executed"
         },
+        {
+            "name": "codewhispererSuggestedTokens",
```
justinmk3, line 244: when are these telemetry definitions going to be upstreamed to https://github.com/aws/aws-toolkit-common ? many of these have been here for 6+ months.

zixlin7: this particular event is here because we were collecting `codewhispererUserInputDetails` for VS Code debugging, agreed we should remove it when it's no longer needed.

zixlin7: @leigaol do you know if we still need this data?

leigaol: You can remove it now.

leigaol (approved):

PR #5465, opened 2024-08-19
telemetry: add function stack to auth connection change

nkomonen-amazon:
## Problem:

When an auth connection changes, eg goes from valid to invalid, there were cases for some users where it happened unexpectedly. It was difficult to answer why the users connection became invalid

## Solution:

Wrap many auth related functions with their "function entry" context. This will then allow the final telemetry event in `Auth.updateConnectionState()` to report the entire call stack in its `source` field for the metric `auth_modifyConnection`. What this results in is a string like: `AmazonQ#signoutButton:AuthUtil#signout:Auth#signout,deleteConnection,updateConnectionState` which tells us the path taken to get to updating the connection state.

Example metric when I click signout in the Amazon Q quick pick:

```
[debug] telemetry: auth_modifyConnection {
  Metadata: {
    action: 'updateConnectionState',
    connectionState: 'invalid',
    source: 'Commands#aws.amazonq.signout:SecondaryAuth#deleteConnection:Auth#deleteConnection,logout,invalidateConnection,updateConnectionState',
    credentialStartUrl: 'https://view.awsapps.com/start',
    ...
  },
  ...
}
```

In telemetry we can then follow the change in a users connection state and hopefully determine why they may have expired prematurely since the functions which can expire a connection will now emit telemetry.

There is additionally a metric emitted when the users `~/.aws/sso/cache` file is deleted:
```
auth_modifyConnection: {
  action: 'deleteSsoCache',
  source: '',
  sessionDuration: ''
}
```

## Implementation Details:

- A decorator, `withTelemetryContext` was created to wrap a Class Method with its context for telemetry
  - Every method that uses it adds their function name and class as context
  - Under the hood it wraps the Class Method in a `telemetry.function_call.run()` with some predefined values
  - It does not emit a telemetry event by default, as we only need it to contribute to the function stack
  - This decorator significantly reduces the diff compared to using the explicit `run()` since all method code would need to be explicitly wrapped in it.
  - **Note that decorators only work with class methods**
- For non class methods we just use `telemetry.function_call.run()`

## License

### DIFF
--- a/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
+++ b/packages/core/src/auth/sso/ssoAccessTokenProvider.ts
@@ -42,6 +42,8 @@ import { showInputBox } from '../../shared/ui/inputPrompter'
 import { DevSettings } from '../../shared/settings'
 import { onceChanged } from '../../shared/utilities/functionUtils'
 import { NestedMap } from '../../shared/utilities/map'
+import { asStringifiedStack } from '../../shared/telemetry/spans'
+import { withTelemetryContext } from '../../shared/telemetry/util'
 
 export const authenticationPath = 'sso/authenticated'
 
@@ -60,6 +62,7 @@ export abstract class SsoAccessTokenProvider {
      */
     private static _authSource: string = 'unknown'
     private static logIfChanged = onceChanged((s: string) => getLogger().info(s))
+    private readonly className = 'SsoAccessTokenProvider'
 
     public static set authSource(val: string) {
         SsoAccessTokenProvider._authSource = val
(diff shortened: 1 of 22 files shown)


packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -454,15 +473,20 @@ export class DeviceFlowAuthorization extends SsoAccessTokenProvider {
...
      */
     override async getValidatedClientRegistration(): Promise<ClientRegistration> {
-        const cacheKey = this.registrationCacheKey
-        const cachedRegistration = await this.cache.registration.load(cacheKey)
-
-        // Clear cached if registration is expired
-        if (cachedRegistration && isExpired(cachedRegistration)) {
-            await this.invalidate('registrationExpired:DeviceCode')
-        }
+        return telemetry.function_call.run(
```
jpinkney-aws, line 476: Is it possible for these next 3 functions to use the decorator? We could optionally add emit as one of the input options

packages/core/src/login/webview/vue/backend.ts
```diff
@@ -34,6 +34,7 @@ import { AuthSSOServer } from '../../../auth/sso/server'
 import { getLogger } from '../../../shared/logger/logger'
 
 export abstract class CommonAuthWebview extends VueWebview {
+    private readonly className = 'CommonAuthWebview'
```
jpinkney-aws, line 37: It looks like some of the time you have the className declared inside of the class and then others outside of the class. Which one should we standardize on?

packages/core/src/shared/telemetry/util.ts
```diff
@@ -275,3 +277,54 @@ export function getOperatingSystem(): 'MAC' | 'WINDOWS' | 'LINUX' {
...
+ * ```
+ * class A {
+ *     @withTelemetryContext({ name: 'myMethod', class: 'A' })
+ *     myMethod() {
+ *         ...
+ *     }
+ * }
+ * ```
+ */
+export function withTelemetryContext(functionId: FunctionEntry) {
```
jpinkney-aws, line 310: I had it in an early comment but adding emit as an optional arg here might save you from having to indent a few more methods

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
nkomonen-amazon: For some odd reason the decorator was causing a `Reference error` in the `create()` method due to the subclasses not being found. This happened in CI but not locally.

This is a quick fix to that

packages/core/src/shared/telemetry/util.ts
nkomonen-amazon: Not completely sure what you mean by this. Did you have an example

jpinkney-aws: I think it was only relevant to https://github.com/aws/aws-toolkit-vscode/pull/5465#discussion_r1723535776 if you wanted to change emit to true/false through the decorator. Looking at it again I don't think its needed though

jpinkney-aws (approved):

PR #5466, opened 2024-08-20
fix: isSyntaxError() function

nkomonen-amazon:
## Problem:

Due to a recent change the isSyntaxError() function was not returning when expected. We were still seeing it appear in `auth_refreshCredentials` even though it should have been filtered out.

## Solution:

Fix the function so it reports correctly identifies syntax error when it happens.

## License

### DIFF
--- a/packages/core/src/shared/errors.ts
+++ b/packages/core/src/shared/errors.ts
@@ -788,7 +788,7 @@ export function isNetworkError(err?: unknown): err is Error & { code: string } {
     if (
         isVSCodeProxyError(err) ||
         isSocketTimeoutError(err) ||
-        isNonJsonHttpResponse(err) ||
+        isHttpSyntaxError(err) ||
         isEnoentError(err) ||
         isEaccesError(err) ||
         isEbadfError(err) ||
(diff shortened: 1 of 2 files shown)


packages/core/src/shared/errors.ts
```diff
@@ -849,11 +849,13 @@ function isSocketTimeoutError(err: Error): boolean {
...
- * Example error message:
- * "Unexpected token '<', "<html><bod"... is not valid JSON Deserialization error: to see the raw response, inspect the hidden field {error}.$response on this object."
+ * IMPORTANT:
+ *
+ * This function is influenced by {@link getReasonFromSyntaxError()} since it modifies the error
+ * message with the real underlying reason, instead of the default "Unexpected token" message.
  */
-function isNonJsonHttpResponse(err: Error): boolean {
-    return isError(err, 'SyntaxError', 'Unexpected token')
+function isHttpSyntaxError(err: Error): boolean {
```
jpinkney-aws, line 857: Is it possible to write a test for this?

jpinkney-aws (approved):

Issue #5390, opened 2024-07-29
amazonq: project context: "rejectCodeSuggestion not found" after ctrl-z


chrisgeo:
## Steps to reproduce the issue

1. Haven't figure out the pattern
2. Be editing with vim/nvim mode enabled, maybe a linter/formatter or two. Trunk.io can be heavy, especially if both are security scanning (Q was turned off for security)
3. Crashes with the notification (paraphrased) "command aws.amazonq.rejectCodeSuggestion' not found"
4. Some after will find any question complex and out of its domain knowledge 

Point four is likely not related yet amusing that the action of crashing is almost a reset.

## Expected behavior
Stays "alive". Remembers context it was part of in the chat, and doesn't take over the screen with a long running loop.

## System details (run `AWS: About` and/or `Amazon Q: About`)

-   OS: macOS 14.5
-   Visual Studio Code version:
```
Version: 1.91.1
Commit: f1e16e1e6214d7c44d078b1f0607b2388f29d729
Date: 2024-07-09T22:07:46.768Z (2 wks ago)
Electron: 29.4.0
ElectronBuildId: 9728852
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Darwin arm64 23.5.0
```
-   AWS Toolkit version: N/A don't use bundled
-   Amazon Q version:
```
Amazon Q v1.17.0
Published
2024-04-03, 08:40:29
Last released
2024-07-25, 13:24:54
Last updated
2024-07-26, 11:47:41
Identifier
amazonwebservices.amazon-q-vscode
```

jpinkney-aws:
When it happens again can you provide the Amazon Q logs? 

1. In the vscode [command palette](https://code.visualstudio.com/docs/getstarted/userinterface#_command-palette), run `Amazon Q: View Logs`. This reveals the Output panel with `Amazon Q Logs` selected.
2. Click the gear icon and choose `Debug`, then click it again and choose `Set As Default`.
3. Perform the steps to reproduce the issue.
4. Remove any private info from the logs.

chrisgeo:
@jpinkney-aws Good catch, I was just looking and see nothing noticeable 

## amazon lsp logs
```
Fetched context from cmd/metrics_batch.go, po_metric_events/model.go, po_metric_events/consumer.go, po_metric_events/service.go
Query done in 931ms
Indexing done for 175/354 files
Indexing done for 210/354 files
[...]

PR #5474, opened 2024-08-21
fix(amazonq): fix when reject command is not found after ctrl + z

leigaol:
## Problem
https://github.com/aws/aws-toolkit-vscode/issues/5390

## Solution

Make sure our command is only active when our suggestion is active using when clause context. 

## License

### DIFF
--- a/packages/amazonq/package.json
+++ b/packages/amazonq/package.json
@@ -106,10 +106,6 @@
                         "amazonQWelcomePage": {
                             "type": "boolean",
                             "default": false
-                        },
-                        "amazonQSessionConfigurationMessage": {
-                            "type": "boolean",
-                            "default": false
                         }
                     },
                     "additionalProperties": false
@@ -605,7 +601,7 @@
                 "command": "aws.amazonq.rejectCodeSuggestion",
                 "key": "escape",
                 "mac": "escape",
-                "when": "inlineSuggestionVisible && !editorReadonly && aws.codewhisperer.connected"
+                "when": "inlineSuggestionVisible && !editorReadonly && aws.codewhisperer.connected && aws.codewhisperer.inlineCompletionActive"
             },
             {
                 "command": "aws.amazonq.dismissTutorial",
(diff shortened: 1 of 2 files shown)


zixlin7 (approved):

packages/amazonq/package.json
```diff
@@ -605,7 +605,7 @@
                 "command": "aws.amazonq.rejectCodeSuggestion",
                 "key": "escape",
                 "mac": "escape",
-                "when": "inlineSuggestionVisible && !editorReadonly && aws.codewhisperer.connected"
+                "when": "inlineSuggestionVisible && !editorReadonly && aws.codewhisperer.connected && aws.codewhisperer.inlineCompletionActive"
```
justinmk3, line 608: why isn't `inlineSuggestionVisible` sufficient? That would be the main callout I am interested in seeing in the PR description.

leigaol: Because when you ctrl + z, the inline completion item is recovered from VS Code internal stack, but this item is not within our code path's control. It is not a inline completion item generated by our code path, rather, it is a suggestion memorized by VS Code. 

Therefore, when such suggestion is visible, inlineSuggestionVisible is true, but our commands does not work for such suggestion. 

Root cause is : VS Code copies the completion item and put it in the undo redo stack.

justinmk3:
Why isn't `aws.amazonq.rejectCodeSuggestion` globally defined? Then its handler can check whatever state, instead of using setContext and manually arranging `when` clauses. 

The `aws.amazonq.rejectCodeSuggestion` command should never be "not found".

leigaol:
> Why isn't `aws.amazonq.rejectCodeSuggestion` globally, permanently defined? Then its handler can check whatever state, instead of using setContext and manually arranging `when` clauses.
> 
> The `aws.amazonq.rejectCodeSuggestion` command should never be "not found".

I can give it a try but the inlineSuggestionActive key is also used by other extensions. Doing so means more conflict with others.

packages/amazonq/package.json
```diff
@@ -616,12 +616,12 @@
...
                 "key": "right",
                 "command": "editor.action.inlineSuggest.showNext",
-                "when": "inlineSuggestionVisible && !editorReadonly && aws.codewhisperer.connected"
+                "when": "inlineSuggestionVisible && !editorReadonly && aws.codewhisperer.connected && aws.codewhisperer.inlineCompletionActive"
             },
             {
                 "key": "left",
                 "command": "editor.action.inlineSuggest.showPrevious",
-                "when": "inlineSuggestionVisible && !editorReadonly && aws.codewhisperer.connected"
+                "when": "inlineSuggestionVisible && !editorReadonly && aws.codewhisperer.connected && aws.codewhisperer.inlineCompletionActive"
```
justinmk3, line 624: why are we redefining builtin vscode commands (`editor.action.*`) ?

justinmk3:
> but the inlineSuggestionActive key is also used by other extensions. Doing so means more conflict with others.

Would the solution in this PR would still have a race with those extensions?

How do those extensions define their `when` clauses?

PR #5476, opened 2024-08-21
fix(auth): handle network errors during restore/getChatAuthState

hayemaxi:
## Problem

During extension startup, the auth system tries to restore the connection. This may result in a `refreshToken` call. If a network error happens during this, the connection restore fails and no connection is set to active. Then, Amazon Q will clean up any "leftover" connections, including this one, and the user will be completely logged out. The connection is technically valid despite having an expired token, and another call may clear up the network error and allow us to get better token.

## Solution
The auth system will restore this connection even if it runs into a network error. Then, Amazon Q will display as if its logged in normally. If the network error was one-off, then there is no indication that there was an issue to the user. If there are still network issues, using chat or inline-suggestions will make an error display until `refreshToken` is successful (or the registration expires).

Also, some telemetry updates were included to help make this visible.

Side effects:
- On session startup (with network error):
  - No notification for new CW customizations
  - "featureConfig" AKA A/B testing? will not initialize.

## License

### DIFF
--- a/packages/amazonq/src/extension.ts
+++ b/packages/amazonq/src/extension.ts
@@ -12,6 +12,7 @@ import {
     isAnySsoConnection,
 } from 'aws-core-vscode/auth'
 import {
+    AuthState,
     AuthUtil,
     activate as activateCodeWhisperer,
     shutdown as shutdownCodeWhisperer,
@@ -34,6 +35,7 @@ import {
     globals,
     initialize,
     initializeComputeRegion,
+    isNetworkError,
     messages,
     placeholder,
     setContext,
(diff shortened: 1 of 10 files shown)


packages/core/src/auth/secondaryAuth.ts
```diff
@@ -289,7 +283,34 @@ export class SecondaryAuth<T extends Connection = Connection> {
             getLogger().warn(`auth (${this.toolId}): saved connection "${this.key}" is not valid`)
             await globalState.update(this.key, undefined)
         } else {
-            await this.auth.refreshConnectionState(conn)
+            const getAuthStatus = (state: ReturnType<typeof this.auth.getConnectionState>): AuthStatus => {
+                return state === 'invalid' ? 'expired' : 'connected'
+            }
+
+            let connectionState = this.auth.getConnectionState(conn)
+            telemetry.auth_modifyConnection.record({
```
nkomonen-amazon, line 291: nit: In the case we aren't already running within an `auth_modifyConnection` run, this will do nothing.
I'm wondering if it will be useful to specify where we expect the run to come from, so in this case from `restoreConnection()`

packages/amazonq/src/extension.ts
```diff
@@ -141,32 +143,48 @@ export async function activateAmazonQCommon(context: vscode.ExtensionContext, is
...
-        if (currConn !== undefined && !isAnySsoConnection(currConn)) {
-            getLogger().error(`Current Amazon Q connection is not SSO, type is: %s`, currConn?.type)
-        }
-
-        telemetry.record({
-            authStatus: authState === 'connected' || authState === 'expired' ? authState : 'notConnected',
-            authEnabledConnections: AuthUtils.getAuthFormIdsFromConnection(currConn).join(','),
-            ...(await getTelemetryMetadataForConn(currConn)),
+    try {
+        await telemetry.auth_userState.run(async () => {
```
justinmk3, line 147: nit: typically (I'd even call it idiomatic) we just add `.catch()` if we want to handle `run()` exceptions. Avoids extra nesting.

```suggestion
    try {
        await telemetry.auth_userState.run(async () => { ... }).catch(...)
```

packages/amazonq/src/extension.ts
```diff
@@ -141,32 +143,48 @@ export async function activateAmazonQCommon(context: vscode.ExtensionContext, is
...
+            } else {
+                telemetry.record({ source: ExtStartUpSources.reload })
+            }
+
+            let authState: AuthState = 'disconnected'
+            try {
+                authState = (await AuthUtil.instance.getChatAuthState(false)).codewhispererChat
+            } catch (err) {
+                if (isNetworkError(err)) {
+                    authState = 'connectedWithNetworkError'
```
justinmk3, line 166: how do we know it actually connected?

packages/core/src/shared/errors.ts
```diff
@@ -911,3 +912,30 @@ export function getReasonFromSyntaxError(err: Error): string | undefined {
...
     return undefined
 }
+
+/**
+ * Run a function and swallow any network errors that are thrown inside.
+ * Other types of errors are still thrown.
+ */
+export function runIgnoreNetError<T>(fn: () => T, logMsg?: string): T | undefined
+export function runIgnoreNetError<T>(fn: () => Promise<T>, logMsg?: string): Promise<T> | undefined
+export function runIgnoreNetError<T>(fn: () => T | Promise<T>, logMsg?: string): T | Promise<T | void> | undefined {
```
justinmk3, line 922: not really sure about this pattern, but to avoid copies of it we could parameterize the "ignore" decision:
```suggestion
export function runIgnoreNetError<T>(fn: () => T | Promise<T>, shouldThrow: (err: Error) => boolean, logMsg?: string): T | Promise<T | void> | undefined {
```

packages/core/src/shared/errors.ts
```diff
@@ -911,3 +912,37 @@ export function getReasonFromSyntaxError(err: Error): string | undefined {
...
+    logMsg?: string
+): Promise<T> | undefined
+export function tryRun<T>(
+    fn: () => T | Promise<T>,
+    shouldThrow: (err: Error) => boolean,
+    logMsg?: string
+): T | Promise<T | void> | undefined {
+    const catchErr = (err: Error) => {
+        if (shouldThrow(err)) {
+            getLogger().error(logMsg ?? 'unknown caller: Error ignored: %s', err)
```
justinmk3, line 932:
```suggestion
        if (!shouldThrow(err)) {
            getLogger().error(logMsg ?? 'unknown caller: Error ignored: %s', err)
```

nkomonen-amazon (approved):

PR #5477, opened 2024-08-21
telemetry: add env type `cloudDesktop-amzn` & `ec2-amzn`

nkomonen-amazon:
## Problem:

Anything that was a ssh-remote environment was labeled as `ec2` in our
telemetry. But there are more nuanced cases for Amazon internal ssh connections.

## Solution:

Add more specific checks for a Cloud Desktop and an internal ec2

## License

### DIFF
--- a/packages/core/src/shared/telemetry/telemetryClient.ts
+++ b/packages/core/src/shared/telemetry/telemetryClient.ts
@@ -108,7 +108,7 @@ export class DefaultTelemetryClient implements TelemetryClient {
                         ClientID: this.clientId,
                         OS: os.platform(),
                         OSVersion: os.release(),
-                        ComputeEnv: getComputeEnvType(),
+                        ComputeEnv: await getComputeEnvType(),
                         ParentProduct: vscode.env.appName,
                         ParentProductVersion: vscode.version,
                         MetricData: batch,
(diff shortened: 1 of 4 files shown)


justinmk3 (approved):

packages/core/src/shared/vscode/env.ts
```diff
@@ -94,6 +95,31 @@ export function isRemoteWorkspace(): boolean {
...
+/**
+ * Returns true if we are in an internal Amazon Cloud Desktop
+ */
+export async function isCloudDesktop() {
+    if (!isAmazonInternalOs()) {
+        return false
+    }
+
+    // This heuristic is explained in IDE-14524
+    return (await new ChildProcess('/apollo/bin/getmyfabric').run().then((r) => r.exitCode)) === 0
```
rli, line 120: i believe this is present on any host that can be deployed from apollo, though it would be bizarre for someone to be trying this out on a host that isn't a developer desktop

should we confirm with builder tools if this heuristic will work on the next-gen desktops?

packages/core/src/shared/telemetry/util.ts
```diff
@@ -180,15 +187,20 @@ export function getUserAgent(
 type EnvType =
     | 'cloud9'
     | 'cloud9-codecatalyst'
+    | 'cloudDesktop'
```
rli, line 190: nit: maybe
```suggestion
    | 'amzn-cloudDesktop'
```

justinmk3: +1, the nonce "amzn" for all internal stuff is useful.

packages/core/src/shared/vscode/env.ts
nkomonen-amazon: Thanks for bringing up. I've chatted w/ a BuilderTools member and they confirmed this won't exist in NAWS CloudDesktops. We are discussing a solution for that and can update this function once we determine one in the future.

packages/core/src/shared/telemetry/util.ts
nkomonen-amazon: updated

packages/core/src/shared/telemetry/telemetryClient.ts
```diff
@@ -108,7 +108,7 @@ export class DefaultTelemetryClient implements TelemetryClient {
                         ClientID: this.clientId,
                         OS: os.platform(),
                         OSVersion: os.release(),
-                        ComputeEnv: getComputeEnvType(),
+                        ComputeEnv: await getComputeEnvType(),
```
justinmk3, line 111: Just noticed that this is called by `postMetrics`, not just once per session. Should `getComputeEnvType` be memoized?

PR #5491, opened 2024-08-23
telemetry: ide_editCodeFile is noisy

justinmk3:
## Problem:
- Pseudo-documents like `foo.git`, `output://…`, and `~/.vscode/argv.json` are included in `ide_editCodeFile` metrics even though they were not actually opened by the user.
- `onDidOpenTextDocument` doesn't correctly represent user activity: it fires for *all* documents opened with `onDidOpenTextDocument()`, which includes all documents/buffers opened for programmatic purposes by application or library code.

## Solution:
- Listen to `onDidChangeActiveTextEditor` instead of `onDidOpenTextDocument`. This represents when a user actually opens a file in the IDE.
- Ignore various "noise" documents.

---

### DIFF
--- a/packages/core/src/shared/filetypes.ts
+++ b/packages/core/src/shared/filetypes.ts
@@ -62,8 +62,17 @@ export function isAwsFiletype(doc: vscode.TextDocument): boolean | undefined {
 
 export function activate(): void {
     globals.context.subscriptions.push(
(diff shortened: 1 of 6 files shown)


Jaytru505:
commit

Jaytru505 (commented):
reviewed and merge

PR #5499, opened 2024-08-26
ci(toolkit): add github comment for missing changelog item

jpinkney-aws:
## Problem
- If a user is using the types feat/fix in the PR title they should also include a changelog item

## Solution
- Notify them if a changelog item isn't present

---

### DIFF
--- a/.github/workflows/utils.js
+++ b/.github/workflows/utils.js
@@ -1,0 +1,55 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+function parsePRTitle(title) {
+    const parts = title.split(':')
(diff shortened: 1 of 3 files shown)


.github/workflows/utils.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/*
+ * Check if path is included in at least one of the filename paths
+ */
+function hasPath(filenames, path) {
+    return filenames.some((file) => file.includes(path))
+}
+
+module.exports = {
```
justinmk3, line 51: if we change the extension to `.mjs` can we [use the `export` keyword](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require) on the functions instead? CI uses node 20+ so don't need to worry about older node/javascript versions

justinmk3 (approved):

PR #5507, opened 2024-08-27
fix(workflows): add permissions to node.js.yml

ivikash:
Add permissions to node.js.yml to write to PRs.

---


jpinkney-aws:
Can you try changing the title to something like `fix(workflows):` so that it triggers writing the notification

justinmk3:
> Can you try changing the title to something like `fix(workflows):`

Tried that. Still getting error:

    RequestError [HttpError]: Resource not accessible by integration

Not sure if that's because it's using the old version or if it's using the current version from this PR

PR #5510, opened 2024-08-27
telemetry: Add `sessionId` to metrics

nkomonen-amazon:
## Problem:

I have filtered all the metrics in Kibana by `clientId: "XXXXXX"`, but if they had multiple IDE windows opened at the same time sending metrics I cannot know which window they are coming from since they share the same clientID.

## Solution:

This adds a new field to the telemetry service  called `sessionId` which uniquely identifies a part of the UI in an IDE window. It is adjacent to `clientId`.

Now in my metrics I will be able to differentiate metrics by their sessionID since they both share the same clientId by design.

## Implementation Details:

Changes in the Telemetry service need to be merged before this change can since they update the endpoint to accept this new field.

The updates of this commit only mirror the upstream, so this process is manual.

---

### DIFF
--- a/packages/core/src/shared/crypto.ts
+++ b/packages/core/src/shared/crypto.ts
@@ -39,6 +39,6 @@ export function randomUUID(): `${string}-${string}-${string}-${string}-${string}
  */
 export function isUuid(uuid: string): boolean {
     // NOTE: This pattern must match, or at least be a subset of the "Session ID" pattern in `telemetry/service-2.json`
-    const uuidv4Pattern = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
-    return uuidv4Pattern.test(uuid)
+    const uuidPattern = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
+    return uuidPattern.test(uuid)
 }
--- a/packages/core/src/shared/crypto.ts
+++ b/packages/core/src/shared/crypto.ts
@@ -30,3 +30,15 @@ export function randomUUID(): `${string}-${string}-${string}-${string}-${string}
 
     return require('crypto').randomUUID()
 }
+
+/**
+ * Returns true if the given string is a UUID
+ *
+ * NOTE: There are different UUID versions, this function does not discriminate between them.
+ * See: https://stackoverflow.com/questions/7905929/how-to-test-valid-uuid-guid
+ */
+export function isUUID(uuid: string): boolean {
+    // pattern match the uuid to a uuid string
+    const uuidv4Pattern = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
+    return uuidv4Pattern.test(uuid)
+}
--- a/packages/core/src/shared/crypto.ts
+++ b/packages/core/src/shared/crypto.ts
@@ -37,8 +37,8 @@ export function randomUUID(): `${string}-${string}-${string}-${string}-${string}
  * NOTE: There are different UUID versions, this function does not discriminate between them.
  * See: https://stackoverflow.com/questions/7905929/how-to-test-valid-uuid-guid
  */
-export function isUUID(uuid: string): boolean {
-    // pattern match the uuid to a uuid string
+export function isUuid(uuid: string): boolean {
+    // NOTE: This pattern must match, or at least be a subset of the "Session ID" pattern in `telemetry/service-2.json`
     const uuidv4Pattern = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
     return uuidv4Pattern.test(uuid)
 }
--- a/packages/core/src/shared/telemetry/util.ts
+++ b/packages/core/src/shared/telemetry/util.ts
@@ -105,7 +105,9 @@ export function convertLegacy(value: unknown): boolean {
  * - The session ID must be different from all other IDs on the machine
  * - A session ID exists until the application instance is terminated.
  *   It should never be used again after termination.
- * - All extensions on the same application instance should return the same session ID.
+ * - All extensions on the same application instance MUST return the same session ID.
+ *   - This will allow us to know which of our extensions (eg Q vs Toolkit) are in the
+ *     same VS Code window.
  *
  * `vscode.env.sessionId` behaves as described aboved, EXCEPT its
  * value looks close to a UUID by does not exactly match it (has additional characters).
@@ -115,9 +117,11 @@ export function convertLegacy(value: unknown): boolean {
  */
 export const getSessionId = once(() => uuidV5(vscode.env.sessionId, sessionIdNonce))
 /**
- * This is a nonce that is used in creating a v5 UUID for Session ID.
+ * This is an arbitrary nonce that is used in creating a v5 UUID for Session ID. We only
+ * have this since the spec requires it.
  * - This should ONLY be used by {@link getSessionId}.
- * - This should NEVER change
+ * - This value MUST NOT change during runtime, otherwise {@link getSessionId} will lose its
+ *   idempotency. But, if there was a reason to change the value in a PR, it would not be an issue.
  * - This is exported only for testing.
  */
 export const sessionIdNonce = '44cfdb20-b30b-4585-a66c-9f48f24f99b5' as const
--- a/packages/core/src/shared/telemetry/util.ts
+++ b/packages/core/src/shared/telemetry/util.ts
@@ -8,7 +8,7 @@ import { env, version } from 'vscode'
 import * as os from 'os'
 import { getLogger } from '../logger'
 import { fromExtensionManifest, migrateSetting, Settings } from '../settings'
-import { memoize } from '../utilities/functionUtils'
+import { memoize, once } from '../utilities/functionUtils'
 import {
     isInDevEnv,
     extensionVersion,
@@ -28,6 +28,7 @@ import { randomUUID } from '../crypto'
 import { ClassToInterfaceType } from '../utilities/tsUtils'
 import { FunctionEntry, type TelemetryTracer } from './spans'
 import { telemetry } from './telemetry'
+import { v5 as uuidV5 } from 'uuid'
 
 const legacySettingsTelemetryValueDisable = 'Disable'
 const legacySettingsTelemetryValueEnable = 'Enable'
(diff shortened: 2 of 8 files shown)


hayemaxi:
How is sessionId generated and sent? Will there be code changes to generate this sessionId later, like how we generate clientId?

nkomonen-amazon:
@hayemaxi see the new changes that actually send a sessionID

packages/core/src/shared/telemetry/util.ts
```diff
@@ -90,6 +91,37 @@ export function convertLegacy(value: unknown): boolean {
...
+ * value looks close to a UUID by does not exactly match it (has additional characters).
+ * As a result we process it through uuidV5 which creates a proper UUID from it.
+ * uuidV5 is idempotent, so as long as `vscode.env.sessionId` returns the same value,
+ * we will get the same UUID.
+ */
+export const getSessionId = once(() => uuidV5(vscode.env.sessionId, sessionIdNonce))
+/**
+ * This is a nonce that is used in creating a v5 UUID for Session ID.
+ * - This should ONLY be used by {@link getSessionId}.
+ * - This should NEVER change
```
hayemaxi, line 120: Why?

packages/core/src/shared/telemetry/util.ts
```diff
@@ -90,6 +91,37 @@ export function convertLegacy(value: unknown): boolean {
...
+ *
+ * - On startup of a new application instance, a new session ID must be created.
+ *   - This identifier must be a `UUID`, it is enforced by the telemetry service.
+ * - The session ID must be different from all other IDs on the machine
+ * - A session ID exists until the application instance is terminated.
+ *   It should never be used again after termination.
+ * - All extensions on the same application instance should return the same session ID.
+ *
+ * `vscode.env.sessionId` behaves as described aboved, EXCEPT its
+ * value looks close to a UUID by does not exactly match it (has additional characters).
```
hayemaxi, line 111: Why does it matter what it looks like if its a unique and random value?

packages/core/src/shared/crypto.ts
```diff
@@ -30,3 +30,15 @@ export function randomUUID(): `${string}-${string}-${string}-${string}-${string}
...
     return require('crypto').randomUUID()
 }
+
+/**
+ * Returns true if the given string is a UUID
+ *
+ * NOTE: There are different UUID versions, this function does not discriminate between them.
+ * See: https://stackoverflow.com/questions/7905929/how-to-test-valid-uuid-guid
+ */
+export function isUUID(uuid: string): boolean {
```
hayemaxi, line 40: What might use this? Besides tests

nkomonen-amazon: Anything that needs to validate if a given string is in a valid UUID format

packages/core/src/shared/telemetry/util.ts
nkomonen-amazon: The server enforces it, mentioned here: https://github.com/aws/aws-toolkit-vscode/pull/5510/files#diff-6b731a821720f9104e0c179a1bc64d1039af6d386549d43090cc5c661be0f79eR104

packages/core/src/shared/telemetry/util.ts
nkomonen-amazon: It is an arbitrary value that is required for UUIDv5. There is no reason to change it, and if it is modified part way through a session we would get a different UUID for the same input (though we do not currently run in to this due to the use once `once()` above.

packages/core/src/shared/crypto.ts
justinmk3, line 40:
```suggestion
export function isUuid(uuid: string): boolean {
```

packages/core/src/shared/crypto.ts
```diff
@@ -30,3 +30,15 @@ export function randomUUID(): `${string}-${string}-${string}-${string}-${string}
...
+    // pattern match the uuid to a uuid string
```
justinmk3, line 41: looks like a normal regex, what is the purpose of this comment? 

however it's worth mentioning:

    // Must match "SessionID" pattern in telemetry/service-2.json

packages/core/src/test/shared/crypto.test.ts
```diff
@@ -25,4 +25,22 @@ describe('crypto', function () {
             assert(uuidPattern.test('not-a-uuid') === false)
         })
     })
+
+    describe('isUUID()', function () {
```
justinmk3, line 29:
```suggestion
    describe('isUuid()', function () {
```

packages/core/src/shared/telemetry/util.ts
```diff
@@ -90,6 +91,37 @@ export function convertLegacy(value: unknown): boolean {
     }
 }
 
+/**
+ * Returns an identifier that uniquely identifies a single application
+ * instance/window of a specific IDE. I.e if I have multiple VS Code
+ * windows open each one will have a unique session ID. This session ID
```
justinmk3, line 97: what matters is the toolkit instance, nothing else. we should be clear about that.
```suggestion
 * instance/window of a specific IDE. If I have multiple Toolkit extension instances (multiple vscode windows),
 * each one must have a unique session ID. This session ID
```

packages/core/src/shared/telemetry/util.ts
```diff
@@ -90,6 +91,37 @@ export function convertLegacy(value: unknown): boolean {
...
+ * can be used in conjunction with the client ID to differntiate between
+ * different VS Code windows on a users machine.
+ *
+ * ### Rules:
+ *
+ * - On startup of a new application instance, a new session ID must be created.
+ *   - This identifier must be a `UUID`, it is enforced by the telemetry service.
+ * - The session ID must be different from all other IDs on the machine
+ * - A session ID exists until the application instance is terminated.
```
justinmk3, line 106:
```suggestion
 * - A session ID exists until the Toolkit extension instance is terminated.
```

packages/core/src/shared/telemetry/util.ts
```diff
@@ -90,6 +91,37 @@ export function convertLegacy(value: unknown): boolean {
...
+ *   It should never be used again after termination.
+ * - All extensions on the same application instance should return the same session ID.
```
justinmk3, line 108: This might be an interesting, but is it actually needed? Perhaps this works by accident because of the way things work in vscode, but it might be difficult in other IDEs/systems? Does it matter either way (if not, probably shouldn't be a requirement)?

packages/core/src/test/shared/telemetry/util.test.ts
```diff
@@ -107,6 +110,19 @@ describe('TelemetryConfig', function () {
...
+describe('getSessionId', function () {
+    it('returns a random UUIDv4', function () {
+        const result = getSessionId()
+
+        assert.deepStrictEqual(isUUID(result), true)
+        assert.deepStrictEqual(getSessionId(), result, 'Subsequent call did not return the same UUID')
+    })
+
+    it('nonce is the same as always', function () {
+        assert.deepStrictEqual(sessionIdNonce, '44cfdb20-b30b-4585-a66c-9f48f24f99b5')
```
justinmk3, line 122: do we need to test that a constant is constant?

packages/core/src/test/shared/telemetry/util.test.ts
```diff
@@ -107,6 +110,19 @@ describe('TelemetryConfig', function () {
     })
 })
 
+describe('getSessionId', function () {
+    it('returns a random UUIDv4', function () {
```
justinmk3, line 114: test isn't checking randomness but it does check idempotency

```suggestion
    it('returns a stable UUIDv4', function () {
```

packages/core/src/shared/telemetry/util.ts
nkomonen-amazon: I was thinking that if the Amazon Q and Toolkit extension are in the same window we want to know that since they interact with eachother (eg connection sharing).

This isn't enforced on the service side, so if other IDEs are not able to, then they can use different UUIDs. But then things won't be consistent across IDEs though.

justinmk3: > if the Amazon Q and Toolkit extension are in the same window we want to know

Could be useful. (Though it's an edge case since client-id will usually tell us that.)

packages/core/src/test/shared/telemetry/util.test.ts
hayemaxi: Not a v4
```suggestion
    it('returns a stable UUID', function () {
```

packages/core/src/shared/crypto.ts
```diff
@@ -30,3 +30,15 @@ export function randomUUID(): `${string}-${string}-${string}-${string}-${string}
...
+    const uuidv4Pattern = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
+    return uuidv4Pattern.test(uuid)
```
hayemaxi, line 43:
```suggestion
    const uuidPattern = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
    return uuidPattern.test(uuid)
```

packages/core/src/shared/telemetry/util.ts
```diff
@@ -90,6 +91,37 @@ export function convertLegacy(value: unknown): boolean {
...
+ * instance/window of a specific IDE. I.e if I have multiple VS Code
+ * windows open each one will have a unique session ID. This session ID
+ * can be used in conjunction with the client ID to differntiate between
+ * different VS Code windows on a users machine.
+ *
+ * ### Rules:
+ *
+ * - On startup of a new application instance, a new session ID must be created.
+ *   - This identifier must be a `UUID`, it is enforced by the telemetry service.
+ * - The session ID must be different from all other IDs on the machine
```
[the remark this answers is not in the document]
nkomonen-amazon: This applies to Amazon Q as well as Toolkit. Are we fine using "Toolkit" as a broad name for one of our extensions?

Amazon Q and Toolkit on the same VSCode instance/window should have the same session id

packages/core/src/test/shared/telemetry/util.test.ts
nkomonen-amazon: Any idea how this is possible? I couldn't find one

packages/core/src/shared/telemetry/util.ts
nkomonen-amazon: Added this info to the comment.

packages/core/src/shared/crypto.ts
nkomonen-amazon: A stale comment I used to get Q to generate a pattern. Deleting this, but will add the new comment you suggested.

packages/core/src/shared/telemetry/util.ts
hayemaxi: > We cannot guarantee the format of vscode.env.sessionId as it has no spec, at any point it could change to something else. Then the server would need to modify the pattern it accepts. UUID is a widely recognized format and is something we can consistently guarantee.

> The other IDEs will use this feature as well, if they have something similar to vscode.env.sessionId but its format is slightly different from ours, then should we modify the server to also accept that? I'd rather us have an obvious interface of requiring a UUID that all IDEs can easily follow.

Make the format any string and let the client determine what a "session ID" is, we should be accommodating to existing tools. The format validation isn't useful here - it doesn't stop clients from sending non-unique ids (and might make it harder to detect this if it "looks" right). 
Also, if there is a bug where clients send non-unique IDs, what is the outcome? It is not catastrophic like non-unique `clientId`

packages/core/src/shared/telemetry/util.ts
nkomonen-amazon: If I have Amazon Q and Toolkit on my VS Code window, and Toolkit crashes, I want the session ID to remain the same for Amazon Q.

I'm not sure if this suggested change gets that across

packages/core/src/shared/telemetry/util.ts
nkomonen-amazon: ClientID does not tell us this since it is the same across all VS Code windows on a machine. This was the original issue, if I had multiple VS Code windows with Amazon Q extension, and they were all sending telemetry, I could not differentiate between them. SessionID solves this.

hayemaxi (approved):

PR #5512, opened 2024-08-28
fix(auth): SyntaxError not being considered Network Error

nkomonen-amazon:
## Problem:

There were some previous changes that caused the existing implementation of a `SyntaxError` to not be considered a Network Error. And during token refresh we were seeing SyntaxErrors still appearing

## Solution:

Create a single class which wraps a `SyntaxError`, called `AwsClientResponseError`. Under the hood of a SyntaxError is the real error that originates from a failed AWS SDK Client response, this is how we determine if it should be an `AwsClientResponseError`. This new class can only be created if given the correct SyntaxError instance to `AwsClientResponseError.instanceIf()`.

Then we can simply check if an error matches by using `instanceif AwsClientResponseError`

Any existing code that was related to this scenario has all been centralized in to the AwsClientResponseError class.

---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -299,7 +299,7 @@ export class ChatController {
             errorMessage = e.toUpperCase()
         } else if (e instanceof SyntaxError) {
             // Workaround to handle case when LB returns web-page with error and our client doesn't return proper exception
-            errorMessage = AwsClientResponseError.getReasonFromSyntaxError(e) ?? defaultMessage
+            errorMessage = AwsClientResponseError.tryExtractReasonFromSyntaxError(e) ?? defaultMessage
         } else if (e instanceof CodeWhispererStreamingServiceException) {
             errorMessage = e.message
             requestID = e.$metadata.requestId
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -48,7 +48,7 @@ import { LspController } from '../../../amazonq/lsp/lspController'
 import { CodeWhispererSettings } from '../../../codewhisperer/util/codewhispererSettings'
 import { getSelectedCustomization } from '../../../codewhisperer/util/customizationUtil'
 import { FeatureConfigProvider } from '../../../shared/featureConfig'
-import { getHttpStatusCode, getReasonFromSyntaxError } from '../../../shared/errors'
+import { getHttpStatusCode, AwsClientResponseError } from '../../../shared/errors'
 
 export interface ChatControllerMessagePublishers {
     readonly processPromptChatMessage: MessagePublisher<PromptMessage>
@@ -299,7 +299,7 @@ export class ChatController {
             errorMessage = e.toUpperCase()
         } else if (e instanceof SyntaxError) {
             // Workaround to handle case when LB returns web-page with error and our client doesn't return proper exception
-            errorMessage = getReasonFromSyntaxError(e) ?? defaultMessage
+            errorMessage = AwsClientResponseError.getReasonFromSyntaxError(e) ?? defaultMessage
         } else if (e instanceof CodeWhispererStreamingServiceException) {
             errorMessage = e.message
             requestID = e.$metadata.requestId
--- a/packages/core/src/shared/errors.ts
+++ b/packages/core/src/shared/errors.ts
@@ -888,24 +888,22 @@ function isError(err: Error, id: string, messageIncludes: string = '') {
  * Once we extract the real error message from the hidden field `$response.reason` we get messages similar to:
  *  - "SDK Client unexpected error response: data response code: 403, data reason: Forbidden | Unexpected ..."
  */
-export class AwsClientResponseError extends ToolkitError {
-    public static readonly code: string = 'AwsClientResponseError'
-
-    /** Use {@link instanceIf} to create instance. */
+export class AwsClientResponseError extends Error {
+    /** Use {@link isError} to create instance. */
     protected constructor(err: unknown) {
         const underlyingErrorMsg = AwsClientResponseError.tryExtractReasonFromSyntaxError(err)
 
         /**
-         * This condition should never be hit since {@link AwsClientResponseError.instanceIf}
+         * This condition should never be hit since {@link AwsClientResponseError.isError}
          * is the only way to create an instance of this class, due to the constructor not being public.
          *
-         * The following only exists to make the types checker happy.
+         * The following only exists to make the type checker happy.
          */
         if (!(underlyingErrorMsg && err instanceof Error)) {
             throw Error(`Cannot create AwsClientResponseError from ${JSON.stringify(err)}}`)
         }
 
-        super(underlyingErrorMsg, { code: AwsClientResponseError.code, cause: err })
+        super(underlyingErrorMsg)
     }
 
     /**
@@ -922,7 +920,7 @@ export class AwsClientResponseError extends ToolkitError {
     }
 
     /**
-     * Returns the true underlying error message from a SyntaxError, if possible.
+     * Returns the true underlying error message from a `SyntaxError`, if possible.
      * Otherwise returning undefined.
      */
     static tryExtractReasonFromSyntaxError(err: unknown): string | undefined {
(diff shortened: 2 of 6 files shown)


dogusata (approved):

packages/core/src/test/shared/errors.test.ts
```diff
@@ -492,15 +499,35 @@ describe('util', function () {
...
+    it('SDKClientResponseError', function () {
+        const syntaxError = makeSyntaxError()
+
+        assert.deepStrictEqual(
+            SDKClientResponseError.getReasonFromSyntaxError(syntaxError),
+            'The actual underlying error message'
+        )
+        const sdkError = SDKClientResponseError.fromErrorIf(syntaxError)
+        assert(!(sdkError instanceof SyntaxError))
+        assert(sdkError instanceof Error)
```
hayemaxi, line 524: Redundant with line 525

hayemaxi (approved):

packages/core/src/shared/errors.ts
```diff
@@ -886,28 +875,59 @@ function isError(err: Error, id: string, messageIncludes: string = '') {
...
+ * This class attempts to extract the underlying issue from the SyntaxError.
  *
- * If the reason cannot be found or the error is not a SyntaxError, return undefined.
+ * Example SyntaxError message before extracting the underlying issue:
+ * "Unexpected token '<', "<html><bod"... is not valid JSON Deserialization error: to see the raw response, inspect the hidden field {error}.$response on this object."
  */
-export function getReasonFromSyntaxError(err: Error): string | undefined {
-    if (!(err instanceof SyntaxError)) {
-        return undefined
+export class SDKClientResponseError extends ToolkitError {
```
justinmk3, line 889: "sdk" could mean anything

```suggestion
export class AwsClientResponseError extends ToolkitError {
```

justinmk3, line 889: "sdk" could mean anything

```suggestion
export class AwsClientResponseError extends ToolkitError {
```

packages/core/src/shared/errors.ts
```diff
@@ -886,28 +875,59 @@ function isError(err: Error, id: string, messageIncludes: string = '') {
...
 
-    if (hasKey(err, '$response')) {
-        const response = err['$response']
-        if (response && hasKey(response, 'reason')) {
-            return response['reason'] as string
+    /**
+     * Resolves an instance of {@link SDKClientResponseError} if the given error matches certain criteria.
+     * Otherwise the original error is returned.
+     */
+    static fromErrorIf<T>(err: T): SDKClientResponseError | T {
```
justinmk3, line 907: is this a new pattern we're introducing? is there an existing pattern we can use instead of introducing this. it's hard for developers to know the right thing to do if we have multiple different patterns for the same concept.

packages/core/src/shared/errors.ts
```diff
@@ -886,28 +875,59 @@ function isError(err: Error, id: string, messageIncludes: string = '') {
...
+    static getReasonFromSyntaxError(err: unknown): string | undefined {
+        if (!(err instanceof SyntaxError)) {
+            return undefined
+        }
+
+        if (hasKey(err, '$response')) {
+            const response = err['$response']
+            if (response && hasKey(response, 'reason')) {
+                return response['reason'] as string
+            }
```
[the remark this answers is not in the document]
nkomonen-amazon: See the example in the docstring of the class

packages/core/src/shared/errors.ts
nkomonen-amazon: I couldn't find any obvious patterns. But seeing how we use the `Class.instance` pattern, I've renamed this method to `instanceIf()`. So we get `AwsClientResponseError.instanceIf(err)`, which seems more aligned

packages/core/src/shared/errors.ts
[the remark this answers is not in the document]
justinmk3: are `$response` and `$response.reason` specific to SyntaxError? if not, should this be extracted to a `getReponseReason` or something

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -299,7 +299,7 @@ export class ChatController {
             errorMessage = e.toUpperCase()
         } else if (e instanceof SyntaxError) {
             // Workaround to handle case when LB returns web-page with error and our client doesn't return proper exception
-            errorMessage = getReasonFromSyntaxError(e) ?? defaultMessage
+            errorMessage = AwsClientResponseError.getReasonFromSyntaxError(e) ?? defaultMessage
         } else if (e instanceof CodeWhispererStreamingServiceException) {
             errorMessage = e.message
             requestID = e.$metadata.requestId
```
justinmk3, line 305: all of this logic looks redundant with errors.ts

packages/core/src/shared/errors.ts
```diff
@@ -886,28 +875,60 @@ function isError(err: Error, id: string, messageIncludes: string = '') {
...
-        return undefined
+export class AwsClientResponseError extends ToolkitError {
+    public static readonly code: string = 'AwsClientResponseError'
+
+    /** Use {@link instanceIf} to create instance. */
+    protected constructor(err: unknown) {
+        const underlyingErrorMsg = AwsClientResponseError.getReasonFromSyntaxError(err)
+
+        if (!(underlyingErrorMsg && err instanceof SyntaxError)) {
+            throw Error(`Cannot create AwsClientResponseError from ${JSON.stringify(err)}}`)
```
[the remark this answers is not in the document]
nkomonen-amazon: added a comment to explain why there is not a scenario where this would throw

packages/core/src/shared/errors.ts
```diff
@@ -886,28 +875,69 @@ function isError(err: Error, id: string, messageIncludes: string = '') {
...
- * If the reason cannot be found or the error is not a SyntaxError, return undefined.
+ * Example SyntaxError message before extracting the underlying issue:
+ *  - "Unexpected token '<', "<html><bod"... is not valid JSON Deserialization error: to see the raw response, inspect the hidden field {error}.$response on this object."
+ * Once we extract the real error message from the hidden field `$response.reason` we get messages similar to:
+ *  - "SDK Client unexpected error response: data response code: 403, data reason: Forbidden | Unexpected ..."
  */
-export function getReasonFromSyntaxError(err: Error): string | undefined {
-    if (!(err instanceof SyntaxError)) {
-        return undefined
+export class AwsClientResponseError extends ToolkitError {
```
justinmk3, line 891: why is this modeled as a ToolkitError? it's not semantically a toolkit error. and we don't do this with other basic errors

packages/core/src/shared/errors.ts
```diff
@@ -886,28 +875,69 @@ function isError(err: Error, id: string, messageIncludes: string = '') {
...
 
-    if (hasKey(err, '$response')) {
-        const response = err['$response']
-        if (response && hasKey(response, 'reason')) {
-            return response['reason'] as string
+    /**
+     * Resolves an instance of {@link AwsClientResponseError} if the given error matches certain criteria.
+     * Otherwise the original error is returned.
+     */
+    static instanceIf<T>(err: T): AwsClientResponseError | T {
```
justinmk3, line 915: maybe `isFoo` is the usual name for this pattern. example: https://github.com/aws/aws-toolkit-vscode/blob/4c499f5501fff761d45ca17da78084ddafa27233/packages/core/src/shared/errors.ts#L558

that changes the usage pattern but it's not clear why this error needs to have a different usage pattern or why it needs a dedicated class.

```suggestion
    static isAwsClientResponseError<T>(err: T): is AwsClientResponseError | T {
```

nkomonen-amazon: By not converting the obscure Syntax error to an AwsClientResponseError at the source, we are trusting any downstream code to know to use isFoo to understand the real meaning when they encounter it.

But by creating an AwsClientResponseError right at the source and then throwing that to the downstream, anything that encounters this error will have all the context already in it.

I'd personally rather use an instanceof than a function.

nkomonen-amazon: Going to assume we do not want `AwsClientResponseError.isClientResponseError()`.
I'll make it `isError()`

packages/core/src/shared/errors.ts
nkomonen-amazon: I wanted to have a `cause` and `code` available. But will remove

packages/core/src/shared/errors.ts
nkomonen-amazon: Also the intent of SDK V3 is to start using subclasses of errors instead so we can use `instanceof`. `isAwsError()` is a byproduct of SDK V2 which we should avoid: https://aws.amazon.com/blogs/developer/service-error-handling-modular-aws-sdk-js/

packages/core/src/shared/errors.ts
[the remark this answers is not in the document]
nkomonen-amazon: It is not clear. The only time I think we've seen this is specifically from a SyntaxError in its message.

PR #5516, opened 2024-08-28
telemetry(auth): snooze techdebt and add telemetry for evaluation

hayemaxi:
We are still getting hits in telemetry for users seeing the split session notification. Adding more fine-grained telemetry to determined if this users are still migrating. This telemetry will also help us track other auth issues.

---

### DIFF
--- a/packages/core/src/auth/connection.ts
+++ b/packages/core/src/auth/connection.ts
@@ -282,15 +282,15 @@ export class ProfileStore {
     public async addProfile(id: string, profile: IamProfile): Promise<StoredProfile<IamProfile>>
     @withTelemetryContext({ name: 'addProfile', class: profileStoreClassName })
     public async addProfile(id: string, profile: Profile): Promise<StoredProfile> {
-        return telemetry.auth_modifyConnection.run(async () => {
-            telemetry.record({
+        return telemetry.auth_modifyConnection.run(async (span) => {
+            span.record({
                 action: 'addProfile',
                 id,
                 source: asStringifiedStack(telemetry.getFunctionStack()),
             })
 
             const newProfile = this.initMetadata(profile)
-            telemetry.record(getTelemetryForProfile(newProfile))
+            span.record(getTelemetryForProfile(newProfile))
 
             return await this.putProfile(id, newProfile)
         })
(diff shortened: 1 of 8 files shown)


packages/core/src/auth/connection.ts
```diff
@@ -219,17 +280,40 @@ export class ProfileStore {
...
     }
 
+    @withTelemetryContext({ name: 'updateProfile', class: profileStoreClassName })
     public async updateProfile(id: string, profile: Profile): Promise<StoredProfile> {
-        const oldProfile = this.getProfileOrThrow(id)
-        if (oldProfile.type !== profile.type) {
-            throw new Error(`Cannot change profile type from "${oldProfile.type}" to "${profile.type}"`)
-        }
+        return telemetry.auth_modifyConnection.run(async () => {
+            telemetry.record({
```
nkomonen-amazon, line 302: This records to whatever the "active span" is. In this case it is `auth_modifyConnection` but it leaves room for potential problems down the line if we were to refactor for example.

The safest way is to do `span.record()` since that targets this specific `auth_modifyConnection`. You just need to add `span` as an argument on line 301

packages/core/src/auth/connection.ts
```diff
@@ -238,11 +322,21 @@ export class ProfileStore {
...
     }
 
+    @withTelemetryContext({ name: 'deleteProfile', class: profileStoreClassName })
     public async deleteProfile(id: string): Promise<void> {
-        const data = this.getData()
-        delete (data as Mutable<typeof data>)[id]
-
-        await this.updateData(data)
+        return telemetry.auth_modifyConnection.run(async () => {
+            telemetry.record({
```
nkomonen-amazon, line 328: same thing with span as mentioned above

packages/core/src/auth/secondaryAuth.ts
```diff
@@ -260,13 +260,21 @@ export class SecondaryAuth<T extends Connection = Connection> {
...
                     connectionState: 'undefined',
                 })
                 await this.auth.tryAutoConnect()
                 this.#savedConnection = await this._loadSavedConnection()
                 this.#onDidChangeActiveConnection.fire(this.activeConnection)
 
-                telemetry.record(await getTelemetryMetadataForConn(this.#savedConnection))
+                const conn = this.#savedConnection
+                if (conn) {
+                    telemetry.record({
```
nkomonen-amazon, line 272: same span issue mentioned above

packages/core/src/auth/connection.ts
```diff
@@ -197,17 +201,74 @@ export interface ProfileMetadata {
...
+            })
+            throw err
+        }
+
+        // De-dupe metric on last id and connection state
+        if (
+            this._prevGetProfile?.id !== id ||
+            this._prevGetProfile?.connectionState !== profile.metadata.connectionState
+        ) {
+            telemetry.auth_modifyConnection.emit({
```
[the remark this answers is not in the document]
hayemaxi: This would emit "Profile Not Found" errors

nkomonen-amazon (approved):

PR #5517, opened 2024-08-28
perf(toolkit): add additional metrics for measuring performance

jpinkney-aws:
## Problem
- Right now we can't really corelate cpuusage to anything because we don't know how big users projects are (file size or total files)

## Solution
- Add additional metrics for file size and total file count

---


packages/core/src/testInteg/shared/utilities/workspaceUtils.test.ts
```diff
@@ -297,6 +297,12 @@ describe('collectFiles', function () {
             ] satisfies typeof result,
             result
         )
+
+        assertTelemetry('function_call', {
+            totalFiles: 8,
+            totalFileSizeInMB: 0.0002593994140625,
```
hayemaxi, line 303: Is this value fixed? I.e. it is generated from constant test data?

jpinkney-aws: Yeah its generated from the size of the files

hayemaxi (approved):

nkomonen-amazon (approved):

PR #5521, opened 2024-08-29
ignore

dhasani23:
## Problem

Moved to # 5522. Was having git issues, resolved now.

## Solution

---

### DIFF
--- a/packages/core/src/codewhisperer/tracker/codewhispererTracker.ts
+++ b/packages/core/src/codewhisperer/tracker/codewhispererTracker.ts
@@ -134,6 +134,8 @@ export class CodeWhispererTracker {
                     codewhispererLanguage: suggestion.language,
                     credentialStartUrl: AuthUtil.instance.startUrl,
                     codewhispererUserGroup: CodeWhispererUserGroupSettings.getUserGroup().toString(),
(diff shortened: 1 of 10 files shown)


packages/core/src/codewhisperer/tracker/codewhispererTracker.ts
```diff
@@ -134,6 +134,8 @@ export class CodeWhispererTracker {
                     codewhispererLanguage: suggestion.language,
                     credentialStartUrl: AuthUtil.instance.startUrl,
                     codewhispererUserGroup: CodeWhispererUserGroupSettings.getUserGroup().toString(),
+                    codewhispererCharactersAccepted: suggestion.originalString.length,
+                    codewhispererCharactersModified: 0, // TODO: currently we don't have an accurate number for this field with existing implementation
```
dhasani23, line 138: I also see these lines here: https://github.com/aws/aws-toolkit-vscode/blob/ee95f5ff2988a13780ea2e7e5998e8e3d1567301/packages/core/src/codewhisperer/tracker/codewhispererTracker.ts#L141-L142
so I'm confused as to why they are showing up in the diff.

PR #5522, opened 2024-08-29
telemetry(amazonq): telemetry updates

dhasani23:
## Problem

Need to update telemetry to reflect that we are no longer emitting a metric for API errors or latency, and to reflect the latest common telemetry updates.

## Solution

Update telemetry as described.

---


tincheng (approved):

packages/core/src/codewhisperer/service/transformByQ/transformProjectValidationHandler.ts
```diff
@@ -170,7 +170,7 @@ export async function validateOpenProjects(
             // TODO: remove deprecated metric once BI started using new metrics
```
hayemaxi, line 170: Is this metric still planned for removal?

dhasani23: Yes, but not yet. The others i removed because we have access to API latency / errors from our backend anyway, so we don't need to wait for our BI to get updated

nkomonen-amazon (approved):

PR #5524, opened 2024-08-30
test(core): Remove cpu usage/heap total performance measuring

jpinkney-aws:
## Problem
- We're getting a signal that cpu usage/heap total can't be used to performance measure vscode extensions. This is because all vscode extensions run in a single process, causing other extensions to interfere with our monitoring.

## Solution
- Remove cpu usage/memory performance measuring
- Keep duration measuring for now to see if this can be useful for regression testing (at least its better than nothing?)

---

### DIFF
--- a/packages/core/src/shared/performance/performance.ts
+++ b/packages/core/src/shared/performance/performance.ts
@@ -8,18 +8,6 @@ import { getLogger } from '../logger'
 import { isWeb } from '../extensionGlobals'
 
 interface PerformanceMetrics {
-    /**
-     * The percentange of CPU time spent executing the user-space portions
-     * of the application, (javascript and user-space libraries/dependencies)
-     */
-    userCpuUsage: number
-
-    /**
-     * The percentage CPU time spent executing system-level operations
-     * related to the application, (file I/O, network, ipc, other kernel-space tasks)
-     */
-    systemCpuUsage: number
-    heapTotal: number
     duration: number
 }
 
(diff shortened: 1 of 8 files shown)


hayemaxi:
Couldn't this be a little beneficial in automated CI, where the extension is only one running (if it even is)?

jpinkney-aws:
Its technically running alongside whatever microsoft provides by default (which may or may not also contribute to the cpu usage/memory)

When I start up a brand new vscode insiders instance with no extensions installed I already have:
```
2024-08-30 11:11:04.466 [info] Extension host with pid 3697 started
2024-08-30 11:11:04.639 [info] ExtensionService#_doActivateExtension vscode.git-base, startup: true, activationEvent: '*', root cause: vscode.git
2024-08-30 11:11:04.674 [info] ExtensionService#_doActivateExtension vscode.git, startup: true, activationEvent: '*'
2024-08-30 11:11:04.697 [info] ExtensionService#_doActivateExtension vscode.github, startup: true, activationEvent: '*'
2024-08-30 11:11:04.792 [info] Eager extensions activated
2024-08-30 11:11:04.822 [info] ExtensionService#_doActivateExtension vscode.debug-auto-launch, startup: false, activationEvent: 'onStartupFinished'
2024-08-30 11:11:04.825 [info] ExtensionService#_doActivateExtension vscode.merge-conflict, startup: false, activationEvent: 'onStartupFinished'
2024-08-30 11:11:11.852 [info] ExtensionService#_doActivateExtension vscode.emmet, startup: false, activationEvent: 'onLanguage'
```

Its maybe worth a try but from what I've seen so far everything points to a lack of consistency in the extension host

jpinkney-aws:
As I was writing the PR that would move performance tests to their own CI I just realized that unit tests disable all other extensions and the performance tests which run in there still fail ☹️ https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/scripts/test/launchTestUtilities.ts#L89

PR #5525, opened 2024-08-30
ci(toolkit): Disable notifications on non public vscode toolkit repos

jpinkney-aws:
## Problem
- The notifications workflow can't run on private forks against private repos without someone providing their PAT. If we provide a PAT it leaves the comment of the username of whoever provided the PAT

## Solution
- Only enable notifications if the repository is aws-toolkit-vscode

## Alternative:
- If we had a bot PAT we could set that instead and it wouldn't matter

github workflows 🙃 
---

### DIFF
--- a/.github/workflows/notification.yml
+++ b/.github/workflows/notification.yml
@@ -6,7 +6,7 @@ name: Notifications
 on:
     # `pull_request_target` (as opposed to `pull_request`) gives permissions to comment on PRs.
     pull_request_target:
-        branches: [master, feature/*, staging]
+        branches: [master, feature/*, staging, jpinkney-aws/disable-private-notifications]
         # Default = opened + synchronize + reopened.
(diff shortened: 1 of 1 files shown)


nkomonen-amazon:
Maybe I'm not understanding, but your fork looks to be public (unless I somehow have access).

justinmk3:
> The notifications workflow can't run on private forks against private repos without someone providing their PAT.

The lint-commits job has no issues. Why does this job/workflow have so many difficulties?

jpinkney-aws:
> Why does this job/workflow have so many difficulties?

because you aren't really using the api. You are just throwing an error rather than showing a comment

This workflow uses several things in the api:
1. comparing base repo with forked repo
2. commenting on repos
Both of which require addition permissions when running on forks

justinmk3:
This is fixed now, right?

PR #5526, opened 2024-08-30
fix(auth): handle SyntaxError edge case in refresh credentials

nkomonen-amazon:
## Problem:

There is a small edge case where there are SOMETIMES missing fields in a SyntaxError, in this case the $response or its reason field do not exist. Due to this those cases are causing a connection to become invalid when refresh credentials fails.

## Solution:

Handle the case where those fields are missing and gracefully handle them by still marking it as an AwsClientResponseError. Since we cannot extract the underlying error message, we set a custom errors message for those cases.

Now this will not cause a failed refresh credential to invalidate.

---

### DIFF
--- a/packages/core/src/test/shared/errors.test.ts
+++ b/packages/core/src/test/shared/errors.test.ts
@@ -484,7 +484,10 @@ describe('util', function () {
     })
 
     function makeSyntaxErrorWithSdkClientError() {
-        const syntaxError: Error = new SyntaxError('The SyntaxError message')
+        // The following error messages are not arbitrary, changing them can break functionality
+        const syntaxError: Error = new SyntaxError(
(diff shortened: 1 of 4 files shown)


packages/core/src/test/shared/errors.test.ts
```diff
@@ -526,6 +529,41 @@ describe('util', function () {
...
+            syntaxError,
+            `No 'reason' field in '$response' | ${JSON.stringify(response)} | ${syntaxError.message}`
+        )
+        syntaxError = makeSyntaxErrorWithSdkClientError()
+        response = (syntaxError as any)['$response']
+        response['reason'] = undefined
+        assertIsAwsClientResponseError(
+            syntaxError,
+            `No 'reason' field in '$response' | ${JSON.stringify(response)} | ${syntaxError.message}`
+        )
```
hayemaxi, line 561: Should split these into individual **unit** tests.

hayemaxi (approved):

PR #5531, opened 2024-09-03
telemetry(featuredev): add feedback telemetry

chengoramazon:
## Problem

- feature dev is missing feedback telemetry. 
feedback

## Solution

- Add feedback telemetry

---

### DIFF
--- a/packages/amazonq/.changes/next-release/Feature-fb724754-850b-41b3-b3dd-2484b40720ff.json
+++ b/packages/amazonq/.changes/next-release/Feature-fb724754-850b-41b3-b3dd-2484b40720ff.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Amazon Q feature dev: add feedback telemetry."
+}
--- a/packages/amazonq/.changes/next-release/Feature-fb724754-850b-41b3-b3dd-2484b40720ff.json
+++ b/packages/amazonq/.changes/next-release/Feature-fb724754-850b-41b3-b3dd-2484b40720ff.json
@@ -1,4 +1,0 @@
-{
-	"type": "Feature",
-	"description": "Amazon Q feature dev: add feedback telemetry."
-}
--- a/packages/core/src/amazonqFeatureDev/app.ts
+++ b/packages/core/src/amazonqFeatureDev/app.ts
@@ -26,6 +26,7 @@ export function init(appContext: AmazonQAppInitContext) {
         followUpClicked: new vscode.EventEmitter<any>(),
         openDiff: new vscode.EventEmitter<any>(),
         processChatItemVotedMessage: new vscode.EventEmitter<any>(),
+        processChatItemFeedbackMessage: new vscode.EventEmitter<any>(),
         stopResponse: new vscode.EventEmitter<any>(),
         tabOpened: new vscode.EventEmitter<any>(),
         tabClosed: new vscode.EventEmitter<any>(),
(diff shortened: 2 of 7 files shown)


packages/amazonq/.changes/next-release/Feature-fb724754-850b-41b3-b3dd-2484b40720ff.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Amazon Q feature dev: add feedback telemetry."
+}
```
hayemaxi, line 4:
```suggestion
{
	"type": "Feature",
	"description": "Amazon Q feature dev: add feedback telemetry."
}
```
Telemetry isn't user-facing, so we don't need a changelog item. But thanks for being pro-active about adding one!

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -202,6 +209,21 @@ export class FeatureDevController {
         }
     }
 
+    private async processChatItemFeedbackMessage(message: any) {
```
hayemaxi, line 212: Is there no type for message? If yes, let's use it. If not, why?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -202,6 +209,21 @@ export class FeatureDevController {
...
+
+        await globals.telemetry.postFeedback({
+            comment: `${JSON.stringify({
+                type: 'featuredev-chat-answer-feedback',
+                conversationId: session?.conversationId ?? '',
+                messageId: message?.messageId,
+                reason: message?.selectedOption,
+                userComment: message?.comment,
+            })}`,
+            sentiment: 'Negative',
```
hayemaxi, line 223: Can it only ever be negative?

chengoramazon: Yes, Mynah UI only allows users report feedback when they click on `thumb down ` in the message, so it is always Negative.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
chengoramazon: No, All messages are `any` in the controller, I will add an action item in the backlog.

willyyhuang (approved):

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
hayemaxi: Ok. Let's add a comment:

```suggestion
            sentiment: 'Negative', // The chat UI reports only negative feedback currently.
```

hayemaxi (approved):

packages/core/src/amazonqFeatureDev/app.ts
```diff
@@ -26,6 +26,7 @@ export function init(appContext: AmazonQAppInitContext) {
         followUpClicked: new vscode.EventEmitter<any>(),
         openDiff: new vscode.EventEmitter<any>(),
         processChatItemVotedMessage: new vscode.EventEmitter<any>(),
```
jpinkney-aws, line 28: How is this new code different from processChatItemVotedMessage?

willyyhuang (approved):

jpinkney-aws:
Can you clarify in your PR description what activates this telemetry? Is it the thumbs up/down button or something else?

packages/core/src/amazonqFeatureDev/app.ts
chengoramazon: It is used to record more details when users click on thumb down. Added a screenshot in the description.

jpinkney-aws (approved):

PR #5534, opened 2024-09-03
feat(feature dev): Remove planning phase in favor for new code gen agent

willyyhuang:
## Problem

Us (QSDA) team is launching a new code generation agent.

We have conducted experiments wIth the new code generation agent, we observed:
- Removing planning phase produces higher quality code
- Removing planning phase has general better user experience and remove the inconsistency of code gen & plan gen

We have discussed with product that the rollout of new code generation agent should also include planning phase removal as a whole, without an A/B testing mechanism nor backward compatible implementation. The endpoint for plan generation will still exist to support older toolkit versions.

We plan to release the new code generation agent experience in the week of September 9th.

Please squash merge this pull request, thank you!

## Solution

- Removed planning phase & refine plan in a typical /dev flow.
- Modified customer facing copies.

---


packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -172,23 +169,6 @@ export class FeatureDevController {
         const session = await this.sessionStorage.getSession(tabId)
 
         switch (session?.state.phase) {
```
chengoramazon, line 171: We may no longer need the switch case here given that we added more points to allow users to vote feedback.

packages/core/src/amazonqFeatureDev/session/session.ts
```diff
@@ -167,7 +145,6 @@ export class Session {
 
             // If approach was changed then we need to set it in the next state and this state
             this.state.approach = newApproach
```
chengoramazon, line 147: Is the `state.approach` still being used?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
willyyhuang: okay

packages/core/src/amazonqFeatureDev/session/session.ts
willyyhuang: removed!

chengoramazon (approved):

willyyhuang:
> Two questions:
> 
> 1. Are you going to update the e2e tests in https://github.com/aws/aws-toolkit-vscode/blob/master/packages/amazonq/test/e2e/amazonq/featureDev.test.ts (you could greatly reduce the number of test runs now that its just code generation)
> 2. Are all the UI text changes approved by a UX designer?

1. Ah, I can take a look at this, but could I follow up with another PR on this? As this is rather a large PR already, would be easier to review
2. Yes, all the text changes are approved by a tech writer, not UX designer

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -168,44 +165,23 @@ export class FeatureDevController {
         })
     }
 
-    private async processChatItemVotedMessage(tabId: string, messageId: string, vote: string) {
+    private async processChatItemVotedMessage(tabId: string, _: string, vote: string) {
```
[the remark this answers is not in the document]
willyyhuang: sure

jpinkney-aws (approved):

PR #5538, opened 2024-09-03
telemetry: remove deprecated transform metrics

tincheng:
Clean up deprecated code transform metrics.

---

### DIFF
--- a/packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
+++ b/packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
@@ -289,12 +289,10 @@ interface ZipCodeResult {
 
 export async function zipCode({ dependenciesFolder, humanInTheLoopFlag, modulePath, zipManifest }: IZipCodeParams) {
     let tempFilePath = undefined
-    let zipStartTime = undefined
     let logFilePath = undefined
     let dependenciesCopied = false
     try {
         throwIfCancelled()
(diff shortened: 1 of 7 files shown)


hayemaxi (approved):

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -365,11 +358,6 @@ export async function zipCode({ dependenciesFolder, humanInTheLoopFlag, modulePa
             fs.rmSync(dependenciesFolder.path, { recursive: true, force: true })
         }
     } catch (e: any) {
-        telemetry.codeTransform_logGeneralError.emit({
```
dhasani23, line 368: This isn't deprecated right?

dhasani23 (approved):

hayemaxi (changes requested):
Make sure tests pass but otherwise looks good

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
tincheng: right, the `logGeneralError` metric isn't deprecated, but I think emitting it here isn't very useful since we don't look at this metric, and user journey will track the success rate of upload step which includes zipping. I can add it back if you think the metric will help with troubleshooting, though customers should be able to see the error in log when requesting help.

justinmk3 (approved):
🎉

PR #5540, opened 2024-09-04
ci(core): Seperate setup, execution, and verification of performance tests

jpinkney-aws:
## Problem
We are currently running performance tests without waiting for the cpu to recover. This could be part of the reason why our tests could be flaky.

These were suggestions made by George on Friday

## Solution
Wait until the the cpu usage is at a reasonable (determined using previous test runs) starting cpu usage before continuing on with the tests

---

### DIFF
--- a/packages/core/src/shared/performance/performance.ts
+++ b/packages/core/src/shared/performance/performance.ts
@@ -86,38 +86,43 @@ export class PerformanceTracker {
     }
 }
 
-interface PerformanceTest {
-    setup: () => Promise<any>
+interface PerformanceTest<T> {
+    setup: () => Promise<T>
     // function under performance test
     runTests: () => Promise<any>
-    assertTests: (args: any) => any
+    assertTests: (args: T) => any
 }
+
+const defaultPollingUsage = {
+    darwin: {
+        userCpuUsage: 20,
+        systemCpuUsage: 8,
+    },
+    linux: {
+        userCpuUsage: 20,
+        systemCpuUsage: 8,
+    },
+    win32: {
+        userCpuUsage: 28, // this ci seems to have notably higher default cpu usage
+        systemCpuUsage: 8,
+    },
+}[process.platform as 'linux' | 'darwin' | 'win32']
 
 /**
  * Generate a test suite that runs fn options.testRuns times and gets the average performance metrics of all the test runs
  */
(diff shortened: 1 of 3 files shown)


packages/core/src/shared/performance/performance.ts
```diff
@@ -85,55 +85,89 @@ export class PerformanceTracker {
...
-export function performanceTest(options: TestOptions, name: string, fn: () => void | Promise<void>) {
+export function performanceTest<TSetup, TExecute>(
+    options: TestOptions,
+    name: string,
+    fn: () => PerformanceTestFunction<TSetup, TExecute>
+) {
     const testOption = options[process.platform as 'linux' | 'darwin' | 'win32']
 
-    const totalTestRuns = options.testRuns ?? 5
+    const totalTestRuns = options.testRuns ?? 10
```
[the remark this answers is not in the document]
jpinkney-aws: George just said increase it to 10 and see what happens

packages/core/src/shared/performance/performance.ts
```diff
@@ -85,55 +85,89 @@ export class PerformanceTracker {
...
         after(async () => {
-            const totalUserCPUUsage =
-                testRunMetrics.reduce((acc, metric) => acc + metric.userCpuUsage, 0) / testRunMetrics.length
-            const totalSystemCPUUsage =
-                testRunMetrics.reduce((acc, metric) => acc + metric.systemCpuUsage, 0) / testRunMetrics.length
-            const totalMemoryUsage =
-                testRunMetrics.reduce((acc, metric) => acc + metric.heapTotal, 0) / testRunMetrics.length
-            const totalDuration =
-                testRunMetrics.reduce((acc, metric) => acc + metric.duration, 0) / testRunMetrics.length
+            // use median since its more resistant to outliers
```
hayemaxi, line 146: Might also consider discarding the best and worst results and then averaging the rest, more than 1 instance of extreme values may be notable.

hayemaxi (approved):

packages/core/src/shared/performance/performance.ts
```diff
@@ -85,55 +85,89 @@ export class PerformanceTracker {
     }
 }
 
+interface PerformanceTestFunction<TSetup, TExecute> {
```
nkomonen-amazon, line 88: Not a blocker for this PR but should we add a doc to explain how to use this?

nkomonen-amazon (approved):

PR #5547, opened 2024-09-04
fix(auth): login state updates across windows

hayemaxi:
Problem: Users can open multiple VSC sessions. If they sign out or the connection is lost in another window, then interacting with chat or in-line suggestions in the original window will only produce cryptic "Profile does not exist" errors, with no clear way to get back to a working state.

Solution: Log out once this state is detected (uses chat, invokes manual or auto inline suggestion).

3 commits:

[fix(amazonq): user stays signed in if they log out in other windows](https://github.com/aws/aws-toolkit-vscode/pull/5547/commits/270a321071d6523ca04b4f59a0fdf5378cf57570)
Problem: Users can open multiple VSC sessions. If they sign out or the connection is lost in another window, then interacting with chat or in-line suggestions in the original window will only produce cryptic "Profile does not exist" errors, with no clear way to get back to a working state.

Solution: Log out once this state is detected (uses chat, invokes manual or auto inline suggestion).

[fix(amazonq): disable 'connection already exists' error for login page](https://github.com/aws/aws-toolkit-vscode/pull/5547/commits/55937840adabdd66584569e8d8b3bed3019b8cc8)
Error would appear due to certain race conditions. If the auth page is visible then we know that we have no connections. (for amazon q only).

[fix(auth): login state updates across windows](https://github.com/aws/aws-toolkit-vscode/pull/5547/commits/ced109f424ff454248c1d81cdcb696410c075f74)
- Uses a vscode filwatcher to detect changes to `.aws/sso/cache`.
- Try to 'reconnect' on new changes based on stored state key.
  - Global state is stored across windows.
- Login or log out will happen in all vsc instances.

---

### DIFF
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -58,6 +58,7 @@ import {
     scopesSsoAccountAccess,
     AwsConnection,
     scopesCodeWhispererCore,
+    ProfileNotFoundError,
 } from './connection'
 import { isSageMaker, isCloud9, isAmazonQ } from '../shared/extensionUtilities'
 import { telemetry } from '../shared/telemetry/telemetry'
@@ -878,8 +879,22 @@ export class Auth implements AuthService, ConnectionManager {
     @withTelemetryContext({ name: 'handleInvalidCredentials', class: authClassName })
     private async handleInvalidCredentials<T>(id: Connection['id'], refresh: () => Promise<T>): Promise<T> {
         getLogger().info(`auth: Handling invalid credentials of connection: ${id}`)
-        const profile = this.store.getProfile(id)
-        const previousState = profile?.metadata.connectionState
+
+        let profile: StoredProfile
+        try {
+            profile = this.store.getProfileOrThrow(id)
+        } catch (err) {
+            if (err instanceof ProfileNotFoundError) {
+                getLogger().info(
+                    `Auth: deleting connection '${id}' due to error encountered while fetching auth token: %s`,
+                    err
+                )
+                await this.deleteConnection({ id })
+            }
+            throw err
+        }
+
+        const previousState = profile.metadata.connectionState
         await this.updateConnectionState(id, 'invalid')
 
         if (previousState === 'invalid') {
@@ -895,8 +910,7 @@ export class Auth implements AuthService, ConnectionManager {
             const timeout = new Timeout(60000)
             this.#invalidCredentialsTimeouts.set(id, timeout)
 
-            const connLabel =
-                profile?.metadata.label ?? (profile?.type === 'sso' ? this.getSsoProfileLabel(profile) : id)
+            const connLabel = profile.metadata.label ?? (profile.type === 'sso' ? this.getSsoProfileLabel(profile) : id)
             const message = localize(
                 'aws.auth.invalidConnection',
                 'Connection "{0}" is invalid or expired, login again?',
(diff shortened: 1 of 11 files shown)


packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -254,8 +256,20 @@ export class AuthUtil {
...
+        try {
+            const bearerToken = await this.conn.getToken()
+            return bearerToken.accessToken
+        } catch (err) {
+            if (err instanceof ProfileNotFoundError) {
+                getLogger().info(
+                    'getBearerToken: logging out due to error encountered while fetching auth token: %s',
+                    err
+                )
+                await this.secondaryAuth.deleteConnection()
```
nkomonen-amazon, line 268: Would this make sense to generalize for all Auths? In `Auth.getToken()` this try/catch would be there and when you delete it, it should emit `auth.onDidDeleteConnection()` and in SecondaryAuth we listen for that event which should trigger the change in AuthUtil. You can then for the focusQPanel call in AuthUtil have it trigger in `onDidChangeActiveConnection()` when the argument is undefined.

This is not a blocker to this PR, but please make a SIM if it makes sense to do this at a later point.

nkomonen-amazon (approved):

packages/core/src/auth/secondaryAuth.ts
```diff
@@ -147,6 +147,40 @@ export class SecondaryAuth<T extends Connection = Connection> {
...
+                        if (this.#savedConnection?.id === this.getStateConnectionId()) {
+                            // Someone updated our cache but the global state doesn't indicate anything new, so do nothing.
+                            // (it could have been us that updated the cache)
+                            getLogger().debug(
+                                `secondaryAuth: cache event did not update global state, no refresh is needed.`
+                            )
+                            return
+                        }
+                        await this.auth.restorePreviousSession()
+                        await this.restoreConnection(true)
```
nkomonen-amazon, line 165: What is the reason for restorePreviousSession as well as restoreConnection? restoreConnection eventually calls restorePreviousSession. It goes restoreConnection -> tryAutoConnect -> restorePreviousSession. And tryAutoConnect uses once(), maybe this was causing things to not work?

hayemaxi: `tryAutoConnect ` is a once() function that will likely have been called on startup. So it will be skipped by `restoreConnection` when this fires. To avoid adding the `force = true` pattern (and other side effects) to that function as well, we can just call the important part of that function manually.

hayemaxi: I would like to avoid putting a comment describing that because I think it will add confusion. This question is rather niche to our understanding of how the auth system works. Let me know if you think otherwise

packages/core/src/auth/auth.ts
```diff
@@ -878,8 +884,22 @@ export class Auth implements AuthService, ConnectionManager {
     @withTelemetryContext({ name: 'handleInvalidCredentials', class: authClassName })
     private async handleInvalidCredentials<T>(id: Connection['id'], refresh: () => Promise<T>): Promise<T> {
         getLogger().info(`auth: Handling invalid credentials of connection: ${id}`)
-        const profile = this.store.getProfile(id)
-        const previousState = profile?.metadata.connectionState
+
+        let profile: StoredProfile
+        try {
+            profile = this.store.getProfileOrThrow(id)
```
justinmk3, line 890: Is there a non-"OrThrow" variant of `getProfile`? That would avoid the need for try/catch.

PR #5549, opened 2024-09-05
fix(auth): File System error during SSO token refresh expires session

nkomonen-amazon:
## Problem:

When we get SSO cache errors, such as a failed write, we do not want to
invalidate the connection. Instead we want to gracefully handle this similar
to network errors, since the connection can still be recovered on a subsequent
token refresh.

A place that this happens is during SSO token refresh. Where if there is a filesystem
error when writing the new token to disk, the connection will become invalid.

## Solution:

Catch sso cache errors and wrap them in a high level `DiskCacheError` so
that we can easily determine an sso cache error during token refresh. Then we can
know not to invalidate the connection when there is a cache error since `isRecoverableError()` will
catch the error.

Additionally a message is shown to the user explaining how the sso token refresh failed and guides
them to the logs in hopes that they can fix it themselves, or at least give it to us when the report the error.

---

### DIFF
--- a/packages/amazonq/package.json
+++ b/packages/amazonq/package.json
@@ -110,6 +110,10 @@
                         "amazonQSessionConfigurationMessage": {
                             "type": "boolean",
                             "default": false
+                        },
+                        "ssoCacheError": {
+                            "type": "boolean",
+                            "default": false
                         }
                     },
                     "additionalProperties": false
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -13,7 +13,7 @@ import * as localizedText from '../shared/localizedText'
 import { Credentials } from '@aws-sdk/types'
 import { SsoAccessTokenProvider } from './sso/ssoAccessTokenProvider'
 import { Timeout } from '../shared/utilities/timeoutUtils'
-import { errorCode, isAwsError, isNetworkError, ToolkitError, UnknownError } from '../shared/errors'
+import { DiskCacheError, errorCode, isAwsError, isNetworkError, ToolkitError, UnknownError } from '../shared/errors'
 import { getCache } from './sso/cache'
 import { isNonNullable, Mutable } from '../shared/utilities/tsUtils'
 import { builderIdStartUrl, SsoToken, truncateStartUrl } from './sso/model'
@@ -652,7 +652,7 @@ export class Auth implements AuthService, ConnectionManager {
     }
 
     private async handleSsoTokenError(id: Connection['id'], err: unknown) {
-        this.throwOnNetworkError(err)
+        this.throwOnRecoverableError(err)
 
         this.#validationErrors.set(id, UnknownError.cast(err))
         getLogger().info(`auth: Handling validation error of connection: ${id}`)
@@ -830,7 +830,7 @@ export class Auth implements AuthService, ConnectionManager {
     @withTelemetryContext({ name: '_getToken', class: authClassName })
     private async _getToken(id: Connection['id'], provider: SsoAccessTokenProvider): Promise<SsoToken> {
         const token = await provider.getToken().catch((err) => {
-            this.throwOnNetworkError(err)
+            this.throwOnRecoverableError(err)
 
             this.#validationErrors.set(id, err)
         })
@@ -842,14 +842,23 @@ export class Auth implements AuthService, ConnectionManager {
      * Auth processes can fail due to reasons outside of authentication actually being expired.
      * We do not want to intepret these failures as invalid/expired auth tokens.
      *
-     * We use this to check if the given error is unanticipated. If it is we throw,
-     * expecting the caller to not change the state of the connection.
+     * We use this to check if the given error is recoverable, meaning retrying getToken at
+     * a later time could succeed. If it is recoverable, we throw, expecting the caller to not
+     * change the state of the connection.
      */
-    private throwOnNetworkError(e: unknown) {
+    private throwOnRecoverableError(e: unknown) {
         if (isNetworkError(e)) {
             throw new ToolkitError('Failed to update connection due to networking issues', {
                 cause: e,
                 code: e.code,
+            })
+        }
+
+        const possibleCacheError = DiskCacheError.instanceIf(e)
+        if (possibleCacheError instanceof DiskCacheError) {
+            throw new ToolkitError('Failed to update connection due to file system operation failures', {
+                cause: possibleCacheError,
+                code: possibleCacheError.code,
             })
         }
     }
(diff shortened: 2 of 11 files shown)


packages/core/src/auth/auth.ts
```diff
@@ -842,16 +842,25 @@ export class Auth implements AuthService, ConnectionManager {
...
-    private throwOnNetworkError(e: unknown) {
+    private throwOnRecoverableError(e: unknown) {
         if (isNetworkError(e)) {
             throw new ToolkitError('Failed to update connection due to networking issues', {
                 cause: e,
                 code: e.code,
             })
         }
+
+        const possibleCacheError = DiskCacheError.instanceIf(e)
```
nkomonen-amazon, line 857: This treats the expected filesystem errors as recoverable (eg if the user fixes a permission issue then the next token refresh can succeed)

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -188,6 +192,25 @@ export abstract class SsoAccessTokenProvider {
 
             return refreshed
         } catch (err) {
+            const possibleCacheError = DiskCacheError.instanceIf(err)
```
nkomonen-amazon, line 195: This will wrap known filesystem errors during token refresh process and cause them to be handled gracefully eventually by`throwOnRecoverableError()` above

packages/core/src/shared/settings.ts
```diff
@@ -620,10 +620,13 @@ export function fromExtensionManifest<T extends TypeDescriptor & Partial<Section
...
-export class ToolkitPromptSettings extends Settings.define(
-    'aws.suppressPrompts',
-    toRecord(keys(toolkitPrompts), () => Boolean)
-) {
+export class ToolkitPromptSettings
+    extends Settings.define(
+        'aws.suppressPrompts',
+        toRecord(keys(toolkitPrompts), () => Boolean)
+    )
+    implements PromptSettings
```
nkomonen-amazon, line 628: Just adding an interface, same as below

packages/amazonq/package.json
```diff
@@ -110,6 +110,10 @@
                         "amazonQSessionConfigurationMessage": {
                             "type": "boolean",
                             "default": false
+                        },
+                        "ssoCacheError": {
```
nkomonen-amazon, line 114: These are just settings to allow to user to suppress the file system error message

packages/core/src/shared/utilities/cacheUtils.ts
```diff
@@ -69,16 +69,20 @@ export async function loadOr<T, K>(cache: KeyedCache<T, K>, key: K, fn: () => Pr
  * Transform functions are not invoked if the specified key does not exist in the cache.
  *
  * @param cache Target cache. The original is _not_ affected.
- * @param get Function applied to all **read** operations from the cache.
- * @param set Function applied to all **write** operations from the cache.
+ * @param loadTransform Function applied to all **read** operations from the cache.
+ * @param saveTransform Function applied to all **write** operations from the cache.
```
justinmk3, line 73: "on" is the typical prefix for callbacks. it helps others to recognize patterns when we use consistent patterns
```suggestion
 * @param onLoad Function applied to all **read** operations from the cache.
 * @param onSave Function applied to all **write** operations from the cache.
```

packages/core/src/auth/auth.ts
```diff
@@ -842,16 +842,25 @@ export class Auth implements AuthService, ConnectionManager {
...
+        if (possibleCacheError instanceof DiskCacheError) {
```
justinmk3, line 858: can we use the [`is` pattern](https://github.com/aws/aws-toolkit-vscode/blob/4c499f5501fff761d45ca17da78084ddafa27233/packages/core/src/shared/errors.ts#L558) instead of this `instanceIf` thing.

oh, `instanceIf` is also a "factory" I guess. we need to find a more conventional pattern. 

```suggestion
        if (isDiskCacheError(e)) {
```

packages/core/src/shared/errors.ts
```diff
@@ -984,6 +984,65 @@ export class AwsClientResponseError extends Error {
...
 
+/**
+ * Represents a generalized error that happened during a disk cache operation.
+ *
+ * For example, when SSO refreshes a token a disk cache error can occur when it
+ * attempts to read/write the disk cache. These errors can be recoverable and do not
+ * imply that the SSO session is stale. So by wrapping those errors in an instance of
+ * this class it will help to distinguish them.
+ */
+export class DiskCacheError extends Error {
```
justinmk3, line 995: what is specific to "cache" here? why is it not `DiskError`? (or `FilesystemError`) ?

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -761,3 +784,53 @@ type ReAuthStateValue = {
...
+/**
+ * Singleton class that manages showing the user a message during
+ * failures that ocurr during SSO disk cache operations.
+ *
+ * Background:
+ * - We need this {@link DiskCacheErrorMessage} specifically as a singleton since we want to ensure
+ *   that only 1 instance of this message appears at a time. There are cases where it shows multiple messages
+ *   if not used as a singleton.
+ */
+class DiskCacheErrorMessage {
```
justinmk3, line 797: putting things in a `class` can help group related logic until we find a better abstraction. but these classes seem to always have a `instance` singleton, why ?

we have `InactivityMessage` which does something similar. maybe it's time to generalize this concept instead of creating a new class for each message.

packages/core/src/shared/errors.ts
nkomonen-amazon: The intent of wrapping something that is ultimately a file system error in disk cache error is to give context to up upstream caller that this file system error is specific to disk cache.

The reason is that somewhere down the call stack may be an unrelated file system error (eg failed to write to the log file). At the top level when it receives this thrown error, the error does not clearly show where it originates from (log file vs disk cache).

So by wrapping file system errors in a a Disk Cache error the intent was for the receiver to know where that file system error came from. The alternative would be for them to follow the stack trace, but iirc doesn't work in production and takes more effort.

This allows the upstream to appropriately react to the "right" kind of file system error.

---

But I do agree it makes more sense to add these specific file system errors in to isFileSystemError and then DiskCacheError uses that under the hood.

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
nkomonen-amazon: Reason for the singleton is explained a couple lines above.

As for making a unified class, outside of reusing the generic "showViewLogsMessage", what else in this could you see being generalized? InactivityMessage and this message class both have their own business logic for when to show the message, what the message content is. InactivityMessage has specific conditions to show different messages depending on how much time has passed, and it needs to keep track of how long a user has been inactive for. This DiskCache message class cares about throttling the message since it has certain issues it needs to worry about.

It's not clear to me how to generalize the two, but if you can draw some parallels that will help.

packages/core/src/auth/auth.ts
nkomonen-amazon: Yeah this creates a new instance instead of verifying it meets a condition.

The reason for why used this over an "is" is in this comment: https://github.com/aws/aws-toolkit-vscode/pull/5549#discussion_r1746199261

PR #5553, opened 2024-09-05
config(launchjson): update presentation with groupings

ivikash:
update presentation with groupings

---

### DIFF
--- a/packages/amazonq/.vscode/launch.json
+++ b/packages/amazonq/.vscode/launch.json
@@ -23,7 +23,11 @@
             "envFile": "${workspaceFolder}/.local.env",
             "outFiles": ["${workspaceFolder}/dist/**/*.js", "${workspaceFolder}/../core/dist/**/*.js"],
             "preLaunchTask": "watch",
-            "postDebugTask": "terminate"
+            "postDebugTask": "terminate",
+            "presentation": {
+                "group": "1_Extension",
+                "order": 2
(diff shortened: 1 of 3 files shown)


packages/amazonq/.vscode/launch.json
```diff
@@ -23,7 +23,12 @@
             "envFile": "${workspaceFolder}/.local.env",
             "outFiles": ["${workspaceFolder}/dist/**/*.js", "${workspaceFolder}/../core/dist/**/*.js"],
             "preLaunchTask": "watch",
-            "postDebugTask": "terminate"
+            "postDebugTask": "terminate",
+            "presentation": {
+                "hidden": false,
```
nkomonen-amazon, line 28: No idea, but is `hidden` false by default?

nkomonen-amazon (approved):

packages/amazonq/.vscode/launch.json
ivikash: Nice find. Removing the `"hidden": false,` didn't impact it at all.
Screenshot 2024-09-05 at 10 52 21 AM

justinmk3:
This is definitely an improvement, thank you!

It would also greatly help to eliminate some of these. For example, all of the `current file` items could be replaced with 1 item which automatically invokes the correct package-specific script based on the current file.

PR #5557, opened 2024-09-06
refactor(telemetry): prepend the error code/name to the `reasonDesc` field

nkomonen-amazon:
## Problem:

In telemetry we sometimes add the error message to the `reasonDesc` field.
Additionally when using `getTelemetryReasonDesc()` it will recurse through the causal
chain of the error and return a value composed of all the error messages.

Eg: `Reason A | Reason B | Reason C`

The issue is that the nested errors may have useful context (in addition to the message) in their
id (code or name). And currently this is lost.

## Solution:

Prepend the code/name to the error messages when using `getTelemetryReasonDesc()`

Eg: `CodeA: Reason A | CodeB: Reason B | NameC: Reason C`

---

### DIFF
--- a/packages/core/src/shared/errors.ts
+++ b/packages/core/src/shared/errors.ts
@@ -250,13 +250,8 @@ export class ToolkitError extends Error implements ErrorInformation {
  *
  * @param withCause Append the message(s) from the cause chain, recursively.
  *                  The message(s) are delimited by ' | '. Eg: msg1 | causeMsg1 | causeMsg2
- * @param withId Prefix the message with the id of the error, this will be something like
- *                 the `code` or `name` property.
  */
-export function getErrorMsg(
-    err: Error | undefined,
-    opts?: { withCause?: boolean; withId?: boolean }
-): string | undefined {
+export function getErrorMsg(err: Error | undefined, withCause: boolean = false): string | undefined {
     if (err === undefined) {
         return undefined
     }
@@ -292,18 +287,17 @@ export function getErrorMsg(
         return undefined
     }
 
-    // prepend id to message
-    const errorId = getErrorId(err)
-    // If a generic error does not have the `name` field explicitly set, it will return a generic 'Error' name.
-    // So we skip if that is that case since it is useless.
-    if (opts?.withId && errorId && errorId !== 'Error') {
-        msg = `${errorId}: ${msg}`
-    }
-
     // append the cause's message
-    const cause = (err as any).cause
-    if (opts?.withCause && cause) {
-        return `${msg}${cause ? ' | ' + getErrorMsg(cause, opts) : ''}`
+    if (withCause) {
+        const errorId = getErrorId(err)
+        // - prepend id to message
+        // - If a generic error does not have the `name` field explicitly set, it returns a generic 'Error' name. So skip since it is useless.
+        if (errorId && errorId !== 'Error') {
+            msg = `${errorId}: ${msg}`
+        }
+
+        const cause = (err as any).cause
+        return `${msg}${cause ? ' | ' + getErrorMsg(cause, withCause) : ''}`
     }
 
     return msg
@@ -431,7 +425,7 @@ export function scrubNames(s: string, username?: string) {
  * @param err Error object, or message text
  */
 export function getTelemetryReasonDesc(err: unknown | undefined): string | undefined {
-    const m = typeof err === 'string' ? err : getErrorMsg(err as Error, { withCause: true, withId: true }) ?? ''
+    const m = typeof err === 'string' ? err : getErrorMsg(err as Error, true) ?? ''
     const msg = scrubNames(m, _username)
 
     // Truncate message as these strings can be very long.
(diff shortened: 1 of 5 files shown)


packages/core/src/shared/errors.ts
```diff
@@ -568,6 +582,16 @@ function hasCode<T>(error: T): error is T & { code: string } {
     return typeof (error as { code?: unknown }).code === 'string'
 }
 
+/**
+ * Returns the identifier the given error.
+ * Depending on the implementation, the identifier may exist on a
+ * different property.
+ */
+export function getErrorId(error: Error): string {
```
justinmk3, line 590: we have `hasCode` already. let's use consistent jargon unless there is a reason not to

```suggestion
export function getErrorCode(error: Error): string {
```

packages/core/src/test/shared/errors.test.ts
```diff
@@ -467,21 +484,40 @@ describe('util', function () {
...
         toolkitError = new ToolkitError('A', { cause: nestedToolkitError })
-        assert.deepStrictEqual(getErrorMsg(toolkitError, true), `A | B | C`)
+        assert.deepStrictEqual(getErrorMsg(toolkitError, { withCause: true }), `A | B | C`)
+
+        // Arg withCause=true, withId=true
+        const cError = new Error('C')
+        cError.name = 'NameC'
+        const bError = new ToolkitError('B', { cause: cError, code: 'CodeB' })
+        const aError = new ToolkitError('A', { cause: bError, code: 'CodeA' })
+        assert.deepStrictEqual(getErrorMsg(aError, { withCause: true, withId: true }), `CodeA: A | CodeB: B | NameC: C`)
```
justinmk3, line 505: It looks like the use-case here is specifically for the `withCause` use-case where the error chain is denormalized into a single `reasonDesc` field.

So, there's no need to make `withId` optional. It should be controlled by the existing `withCause` field. It's not useful or wanted for any other case (we shouldn't be repeating `reason` in `reasonDesc`, normally).

packages/core/src/shared/errors.ts
```diff
@@ -250,17 +250,17 @@ export class ToolkitError extends Error implements ErrorInformation {
  *
  * @param withCause Append the message(s) from the cause chain, recursively.
  *                  The message(s) are delimited by ' | '. Eg: msg1 | causeMsg1 | causeMsg2
+ * @param withId Prefix the message with the id of the error, this will be something like
```
justinmk3, line 253: `withId` can be removed. Just do it uncondtionally when `withCause=true`.

justinmk3 (changes requested):
.

packages/core/src/shared/errors.ts
```diff
@@ -568,6 +582,16 @@ function hasCode<T>(error: T): error is T & { code: string } {
...
 }
 
+/**
+ * Returns the identifier the given error.
+ * Depending on the implementation, the identifier may exist on a
+ * different property.
+ */
+export function getErrorId(error: Error): string {
+    // prioritize code over the name
+    return hasCode(error) ? error.code : error.name
```
justinmk3, line 592: should we rename `hasCode` to `hasId()`? though it's misleading because `error.code` is the standard field name. so maybe `getErrorId` should be `getErrorCode` ?

packages/core/src/test/shared/errors.test.ts
nkomonen-amazon: Good point, updating

packages/core/src/shared/errors.ts
nkomonen-amazon: I was thinking that if we refer to `code` it specifically means the `code` field. So `getErrorCode` may be misleading.

`id` is meant to be the generalized name that is either `code` or `name`.

---

I've seen cases where the error has a unique `name`, but also a `code` that provides additional context. I'm wondering if both exist, that we want them joined together to provide as much context as possible.

eg

```
const err = new Error('blah')
err.name = 'MyName'
err.code = 'MyCode'

getErrorId(err) // 'MyName:MyCode'
```

nkomonen-amazon: @justinmk3

jpinkney-aws (approved):

packages/core/src/shared/errors.ts
```diff
@@ -251,16 +251,11 @@ export class ToolkitError extends Error implements ErrorInformation {
  * @param withCause Append the message(s) from the cause chain, recursively.
```
justinmk3, line 251: docstring needs update: "Append the name + message ..."

packages/core/src/shared/errors.ts
```diff
@@ -420,8 +428,8 @@ export function getTelemetryReasonDesc(err: unknown | undefined): string | undef
     const m = typeof err === 'string' ? err : getErrorMsg(err as Error, true) ?? ''
```
justinmk3, line 428: Why is this always setting `withCause=true` ?

PR #5558, opened 2024-09-06
refactor(auth): redo `DiskCacheError` with proper context

nkomonen-amazon:
## Problem:

The original implementation of DiskCacheError can be simplified and still achieve it's original purpose.

## Solution:

During token refresh we eventually write to the disk cache. Sometimes this process fails due to filesystem errors.

We dont need to care about the actual file system error, but only that the error came during a DiskCache operation (i.e when it attempts to save the token to disk)

So instead, at the root of where the disk cache errors ocurr, we will wrap caught errors in a DiskCacheError.
This adds context to the error and allows anything upstream to specifically check for that context, instead of needing to parse the error to see if it was filesystem related.

Additionally not all file system errors were guaranteed to come from the disk cache, so this change allows us to target the right errors.

---

### DIFF
--- a/packages/core/src/shared/utilities/cacheUtils.ts
+++ b/packages/core/src/shared/utilities/cacheUtils.ts
@@ -5,7 +5,7 @@
 
 import * as vscode from 'vscode'
 import { dirname } from 'path'
-import { ToolkitError, isFileNotFoundError } from '../errors'
+import { ErrorInformation, ToolkitError, isFileNotFoundError } from '../errors'
 import fs from '../../shared/fs/fs'
 import { isWeb } from '../extensionGlobals'
 import type { MapSync } from './map'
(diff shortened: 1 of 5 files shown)


packages/core/src/shared/utilities/cacheUtils.ts
```diff
@@ -160,15 +154,34 @@ export function createDiskCache<V, K>(
...
+            details: { key },
+        })
+    }
+}
+
+/**
+ * Represents a generalized error that happened during a disk cache operation.
+ *
+ * For example, when SSO refreshes a token a disk cache error can occur when it
+ * attempts to read/write the disk cache. These errors can be recoverable and do not
```
jpinkney-aws, line 176: > These errors can be recoverable
(newbie questions): How will we identify which errors are recoverable? Or will that happen somewhere else?

hayemaxi (approved):

jpinkney-aws (approved):
It feels like this approach helps with observability and making it easier to track down where a problem occured

packages/core/src/shared/utilities/cacheUtils.ts
nkomonen-amazon: In the context of a disk cache error during token refresh, when the issue is related to the filesystem it just means
that if the disk cache error is fixed then we should be able to refresh the token.

So technically all disk related errors should be recoverable since the problems we saw were related to permissions, storage space, file was busy, ... and the current hope is the user either sees the message and uses intuition to fix it, or reports it to us

PR #5559, opened 2024-09-06
feat(amazonq): workspace indexing only applies to new files that are changed

leigaol:
## Problem

In current build, when workspace index expires, the IDE rebuild index for all files, which is slow. 

## Solution

Whenever the current index is out of date, the IDE rebuild index for only the files that are changed since last index. 

Meanwhile, this change will also make the index refresh mush faster. New index will be created and refreshed if the existing index is more than 12 hours old.

---


Will-ShaoHua (approved):

nkomonen-amazon (approved):

leigaol:
This should not require a review from DEXP

PR #5560, opened 2024-09-06
fix(auth): add more errors to `isNetworkError()`

nkomonen-amazon:
These showed up in telemetry during `aws_refreshCredentials`

---

### DIFF
--- a/packages/core/src/shared/errors.ts
+++ b/packages/core/src/shared/errors.ts
@@ -829,6 +829,7 @@ export function isNetworkError(err?: unknown): err is Error & { code: string } {
         'EAI_FAIL',
         '502', // This may be irrelevant as isBadResponseCode() may be all we need
         'InternalServerException',
+        'ERR_SSL_WRONG_VERSION_NUMBER',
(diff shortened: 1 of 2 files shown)


packages/core/src/shared/errors.ts
```diff
@@ -890,7 +891,7 @@ export function isError(err: Error, id: string, messageIncludes: string = '') {
  * These are the errors explicitly seen in telemetry. We can instead do any non-200 response code
  * later, but this will give us better visibility in to the actual error codes we are currently getting.
  */
-const errorResponseCodes = [302, 403, 502]
+const errorResponseCodes = [302, 403, 404, 502, 503]
```
ivikash, line 894: Should we update the logic to include all 3XX, 4XX, and 5XX errors?

nkomonen-amazon: For the sake of visibility in telemetry I was keeping it like this for now, based on the history we don't get many other error codes.

But if this starts to grow then yeah it will probably be best to do a check on the ranges.

PR #5564, opened 2024-09-09
refactor(logging): removed types input in getLogger()

tomcat323:
## Problem
Currently we use `getLogger()` for all log messages.

`getLogger()` takes a type input, indicating one of three logger types: 'channel', 'debugConsole', and 'main'.

- The `channel` type is out dated and no longer carries meaning. Might even confuse other developers if they come across this.
- The `debugConsole` type logging calls are only used in SamCLI, we shouldn’t complicate a very popular interface with something so specific and rarely used. 
- It's discovered in implementation that the current `debugConsole` is not printing to IDE's debug console, this is fixed in one of the commits. 

This change is also a pre-rec of the following project:
https://quip-amazon.com/B4hTAcZSsGXF/VSC-Better-Logging-With-Headers

## Solution
1. The `channel` type should be removed, converting existing calls into `main` type.
2. Move the `debugConsole` type logger out of the `getLogger`, into a separate function `debugConsoleLogger()' only for the rare calls in SamCLI.

---

### DIFF
--- a/packages/core/src/shared/logger/activation.ts
+++ b/packages/core/src/shared/logger/activation.ts
@@ -71,27 +71,21 @@ export async function activate(
  * @param opts.logLevel Log messages at or above this level
  * @param opts.logPaths Array of paths to output log entries to
  * @param opts.outputChannels Array of output channels to log entries to
- * @param opts.useDebugConsole If true, outputs log entries to `vscode.debug.activeDebugConsole`
  * @param opts.useConsoleLog If true, outputs log entries to the nodejs or browser devtools console.
  */
 export function makeLogger(opts: {
     logLevel: LogLevel
     logPaths?: vscode.Uri[]
     outputChannels?: vscode.OutputChannel[]
-    useDebugConsole?: boolean
     useConsoleLog?: boolean
 }): Logger {
     const logger = new WinstonToolkitLogger(opts.logLevel)
     // debug console can show ANSI colors, output channels can not
-    const stripAnsi = opts.useDebugConsole ?? false
     for (const logPath of opts.logPaths ?? []) {
         logger.logToFile(logPath)
     }
     for (const outputChannel of opts.outputChannels ?? []) {
-        logger.logToOutputChannel(outputChannel, stripAnsi)
-    }
-    if (opts.useDebugConsole) {
-        logger.logToDebugConsole()
+        logger.logToOutputChannel(outputChannel)
     }
     if (opts.useConsoleLog) {
         logger.logToConsole()
--- a/packages/core/src/shared/logger/activation.ts
+++ b/packages/core/src/shared/logger/activation.ts
@@ -46,16 +46,6 @@ export async function activate(
 
     setLogger(mainLogger)
 
-    // Logs to "AWS Toolkit" output channel.
-    setLogger(
-        makeLogger({
-            logLevel: chanLogLevel,
-            logPaths: logUri ? [logUri] : undefined,
-            outputChannels: [outputChannel, logChannel],
-        }),
-        'channel'
-    )
-
     // Logs to vscode Debug Console.
     setLogger(
         makeLogger({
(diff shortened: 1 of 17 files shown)


packages/core/src/shared/logger/activation.ts
```diff
@@ -51,7 +51,7 @@ export async function activate(
         makeLogger({
             logLevel: chanLogLevel,
             outputChannels: [outputChannel, logChannel],
-            useDebugConsole: true,
+            useConsoleLog: true,
```
nkomonen-amazon, line 54: Are we using the `useDebugConsole` field anymore? If not, it can probably be removed

nkomonen-amazon: Also under the hood is just `useConsoleLog` -> `logToConsole()` -> `console.log()`.

So I wonder if `useDebugConsole` is obsolete. Doing a `console.log()` will write to the `Debug Console` window as well as the window when you run the command palette command `Toggle Developer Tools`. Maybe there was a historical reason for `DebugConsole`, but now I think we may be able to just drop it and keep the Console logger as that includes writing to the `Debug Console` (we should document this in a docstring about where this will appear.

So for changes it may make sense to just delete `useDebugConsole` and its associated code.

nkomonen-amazon (commented):
Nice first PR!. Happy to see the separated commits.

nkomonen-amazon:
@tomcat323 For the failing CI:
- CodeAction failures are expected at the moment, you can ignore those for now.
- Failed Github Actions CI: there are some flaky ones, so just click the `Details` button and rerun the failed ones at the top right. But first check if the failure actually looks related to your change. A good rule of thumb is that if some versions pass (eg: stable passes, but minimum is failing) then the failure is probably unrelated to your change.
- The failed AWS CodeBuild tests can be re-run by commenting `/retryBuilds`

packages/core/src/shared/logger/activation.ts
tomcat323: I agree. And `useDebugConsole` isn't used anywhere else either.

packages/core/src/shared/logger/logger.ts
```diff
@@ -84,20 +84,21 @@ export function compareLogLevel(l1: LogLevel, l2: LogLevel): number {
...
  * Gets the logger if it has been initialized
- * @param type Gets the logger type:
- * * `'main'` or `undefined`: Main logger; default impl: logs to log file and log output channel
- * * `'channel'`: Channel Logger; default impl: logs to the `main` channels and the `AWS Toolkit` output channel
- * * `'debug'`: Debug Console Logger; default impl: logs to the `channel` channels and the currently-active VS Code Debug Console pane.
+ * the logger is of `'main'` or `undefined`: Main logger; default impl: logs to log file and log output channel
  */
-export function getLogger(type?: keyof typeof toolkitLoggers): Logger {
-    const logger = toolkitLoggers[type ?? 'main']
+export function getLogger(): Logger {
```
ivikash, line 89: Now that it doesn't take any parameters any more, can we make this a variable named `logger` instead of a `getLogger` function?

This would make the code so much cleaner like - `logger.info(....)` which seems like a natural extension of Winston or any other logger system.

ivikash: Indeed. In that case, we would need a function. Here I saw that the function doesn't take any inputs so I interpreted it as a constant.

tomcat323: yes that is the reason, those changes will be implemented in the next pull request, thanks for asking!

nkomonen-amazon (approved):

packages/core/src/shared/logger/winstonToolkitLogger.ts
```diff
@@ -44,7 +44,7 @@ export class WinstonToolkitLogger implements Logger, vscode.Disposable {
     }
 
     public enableDebugConsole(): void {
-        this.logToDebugConsole()
+        this.logToConsole()
```
justinmk3, line 47: is this correct? we still have `getDebugConsoleLogger` but it doesn't appear to actually log anything to the debug console, when I invoke a local lambda:

image

justinmk3: Alternatively, can we make sure the Output window is focused? Seems like vscode aggressively focuses the Debug Console during debugging.

PR #5565, opened 2024-09-09
fix(amazonq): fix chat project context telemetry discrepancy

leigaol:
## Problem
There is a discrepancy between STE telemetry and toolkit telemetry in terms of field `cwsprChatHasProjectContext` in 
`AmazonqAddMessage`.

## Solution

Use same boolean field for `cwsprChatHasProjectContext`

---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/telemetryHelper.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/telemetryHelper.ts
@@ -332,7 +332,9 @@ export class CWCTelemetryHelper {
 
     public recordAddMessage(triggerPayload: TriggerPayload, message: PromptAnswer) {
         const triggerEvent = this.triggerEventsStorage.getLastTriggerEventByTabID(message.tabID)
-
+        const hasProjectLevelContext = triggerPayload.relevantTextDocuments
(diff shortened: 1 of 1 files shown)


nkomonen-amazon (approved):

packages/core/src/codewhispererChat/controllers/chat/telemetryHelper.ts
```diff
@@ -332,7 +332,9 @@ export class CWCTelemetryHelper {
 
     public recordAddMessage(triggerPayload: TriggerPayload, message: PromptAnswer) {
         const triggerEvent = this.triggerEventsStorage.getLastTriggerEventByTabID(message.tabID)
-
+        const hasProjectLevelContext = triggerPayload.relevantTextDocuments
+            ? triggerPayload.relevantTextDocuments.length > 0 && triggerPayload.useRelevantDocuments === true
+            : false
```
Will-ShaoHua, line 337:
```suggestion
        const hasProjectLevelContext = triggerPayload.relevantTextDocuments &&            triggerPayload.relevantTextDocuments.length > 0 && triggerPayload.useRelevantDocuments === true            
```

Will-ShaoHua: how about^?

Will-ShaoHua (approved):

PR #5566, opened 2024-09-10
telemetry(amazonq): measure e2e latency for amazon q requests

jpinkney-aws:
## Problem:
- We can't measure e2e how long a chat request takes

## Solution:
- Plumb the trace id through to the webview -> vscode -> webview. This allows us to know exactly what is happening to a message and how long everything took to get there
- chat e2e latency can be linked with the other telemetry events via the traceId

<details closed>
<summary>Previous PR description</summary>

## Problem
This is just a draft of what we could potentially do to measure e2e latency in amazon q chat message requests. Feel free to leave comments on the PR

### Main code level problems:
- It's hard to connect a tab id to a conversation id. Tab ID's exist pretty much to send/receive messages but they don't really live along side conversation id's. Conversation id's are exclusively related to the session. The only place they meetup is in their chat session storages
- Do we want to track individual events like webviewToDispatcher as metrics?
    - Though there wouldn't really be a way to "listen" on these events except for if we manually allow-listed metrics since theres no reason to attach a tabID to that metric
- How do we listen to events emitted from telemetry so that we can find them later?
    - One solution is proposed in this PR where we listen to events being emitted and then check if we want to track them
- Could we connect these with a parent/child mechanism like we want to introduce eventually?
    - We kind of do this because the "listener" event is the parent and the things it finds are the children. The problem is we can really have only one "listener" for these Q events, since we don't know what tabs are related to what conversation ID's until much later (after the first request). This is why we build up the map of conversation metrics and tab metrics and join them together before we emit

## Solution

This PR does a couple things:
- Moves up the chat session storage for codewhisperer chat. This allows us to eventually connect conversation ids to tabs
- Introduces a chat storage facades that allows us to call getConversationId on any tab type. This is only implemented for cwc right now
- Introduces a listener pattern that allows telemetry events to "listen" to metrics that are being emitted
- emits a sample events that can you can trace from "pressing enter" to "receives message in ui"

### Sample event when sending a request in amazon q chat:
```
2024-09-11 10:11:05.712 [debug] telemetry: chat_roundTrip {
  Metadata: {
    duration: '14585',
    name: 'onChatPrompt',
    child_metrics: 'initialRequestTime,webviewToDispatcher,dispatcherToFeature,amazonq_enterFocusConversation,amazonq_startConversation,amazonq_addMessage,finalRequestTime',
    child_metric_durations: '{"webviewToDispatcher":1,"dispatcherToFeature":3,"amazonq_enterFocusConversation":1542,"amazonq_startConversation":2,"amazonq_addMessage":13003,"finalRequestTime":34}',
    result: 'Succeeded',
    awsAccount: 'not-set',
    awsRegion: 'us-east-1'
  },
  Value: 1,
  Unit: 'None',
  Passive: false
}
```

### You can also get events from the right click context menu items as well (explain, refactor, etc)
```
2024-09-11 10:11:34.384 [debug] telemetry: chat_roundTrip {
  Metadata: {
    duration: '12885',
    name: 'aws.amazonq.explainCode',
    child_metrics: 'initialRequestTime,webviewToDispatcher,amazonq_enterFocusConversation,amazonq_addMessage,finalRequestTime',
    child_metric_durations: '{"webviewToDispatcher":17,"amazonq_enterFocusConversation":1263,"amazonq_addMessage":11572,"finalRequestTime":33}',
    result: 'Succeeded',
    awsAccount: 'not-set',
    awsRegion: 'us-east-1'
  },
  Value: 1,
  Unit: 'None',
  Passive: false
}
```

</details>

---

### DIFF
--- a/packages/core/src/amazonq/webview/messages/messageDispatcher.ts
+++ b/packages/core/src/amazonq/webview/messages/messageDispatcher.ts
@@ -10,6 +10,7 @@ import { TabType } from '../ui/storages/tabsStorage'
 import { getLogger } from '../../../shared/logger'
 import { amazonqMark } from '../../../shared/performance/marks'
 import { telemetry } from '../../../shared/telemetry'
+import { AmazonQChatMessageDuration } from '../../messages/chatMessageDuration'
 
 export function dispatchWebViewMessagesToApps(
     webview: Webview,
@@ -30,6 +31,14 @@ export function dispatchWebViewMessagesToApps(
             })
             performance.clearMarks(amazonqMark.uiReady)
             performance.clearMarks(amazonqMark.open)
+            return
+        }
+
+        if (msg.type === 'startListening') {
+            AmazonQChatMessageDuration.startListening(msg)
+            return
+        } else if (msg.type === 'stopListening') {
+            AmazonQChatMessageDuration.stopListening(msg)
             return
         }
 
--- a/packages/core/src/amazonq/webview/messages/messageDispatcher.ts
+++ b/packages/core/src/amazonq/webview/messages/messageDispatcher.ts
@@ -34,11 +34,11 @@ export function dispatchWebViewMessagesToApps(
             return
         }
 
-        if (msg.type === 'startListening') {
-            AmazonQChatMessageDuration.startListening(msg)
+        if (msg.type === 'startChatMessageTelemetry') {
+            AmazonQChatMessageDuration.startChatMessageTelemetry(msg)
             return
-        } else if (msg.type === 'stopListening') {
-            AmazonQChatMessageDuration.stopListening(msg)
+        } else if (msg.type === 'stopChatMessageTelemetry') {
+            AmazonQChatMessageDuration.stopChatMessageTelemetry(msg)
             return
         }
 
--- a/packages/core/src/amazonq/messages/chatMessageDuration.ts
+++ b/packages/core/src/amazonq/messages/chatMessageDuration.ts
@@ -11,7 +11,7 @@ export class AmazonQChatMessageDuration {
     /**
      * Record the initial requests in the chat message flow
      */
-    static startListening(msg: { traceId: string; startTime: number; trigger?: string }) {
+    static startChatMessageTelemetry(msg: { traceId: string; startTime: number; trigger?: string }) {
         const { traceId, startTime, trigger } = msg
 
         uiEventRecorder.set(traceId, {
@@ -34,7 +34,7 @@ export class AmazonQChatMessageDuration {
     /**
      * Stop listening to all incoming events and emit what we've found
      */
-    static stopListening(msg: { traceId: string }) {
+    static stopChatMessageTelemetry(msg: { traceId: string }) {
         const { traceId } = msg
 
         // We can't figure out what trace this event was associated with
--- a/packages/core/src/amazonq/messages/chatMessageDuration.ts
+++ b/packages/core/src/amazonq/messages/chatMessageDuration.ts
@@ -1,0 +1,93 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { globals } from '../../shared'
+import { telemetry } from '../../shared/telemetry'
+import { Event, uiEventRecorder } from '../util/eventRecorder'
+
+export class AmazonQChatMessageDuration {
+    /**
+     * Record the initial requests in the chat message flow
+     */
+    static startListening(msg: { traceId: string; startTime: number; trigger?: string }) {
+        const { traceId, startTime, trigger } = msg
+
+        uiEventRecorder.set(traceId, {
+            events: {
+                chatMessageSent: startTime,
+            },
+        })
+        uiEventRecorder.set(traceId, {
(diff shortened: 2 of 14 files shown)


packages/core/src/amazonq/util/telemetryHelper.ts
```diff
@@ -0,0 +1,67 @@
...
+        ]
+    >()
+
+    /**
+     * Record the time an event happened in a tabID or conversation
+     * @param tabIDOrConversationId an amazon Q tabID or a conversation ID
+     * @param event a UI or telemetry event
+     * @param time the duration it took
+     */
+    record(tabIDOrConversationId: string, event: Event, time: number) {
```
nkomonen-amazon, line 34: can we make the event + time the fields of an `options` argument instead?

```
(..., opts: { event: Event, time: number })
```

packages/core/src/amazonq/util/telemetryHelper.ts
```diff
@@ -0,0 +1,67 @@
...
+                time: number
+            },
+        ]
+    >()
+
+    /**
+     * Record the time an event happened in a tabID or conversation
+     * @param tabIDOrConversationId an amazon Q tabID or a conversation ID
+     * @param event a UI or telemetry event
+     * @param time the duration it took
```
nkomonen-amazon, line 32: I'm seeing uses where it looks like this value being set is a point in time and not the actual duration/diff. Does this need to be updated to something like `startTime`?

packages/core/src/amazonq/util/telemetryHelper.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { MetricName } from '../../shared/telemetry/clienttelemetry'
+
+// TODO: technically this can also be any metric name that occurs in the flow now
+type Event = 'initialRequestTime' | 'webviewToDispatcher' | 'dispatcherToFeature' | 'finalRequestTime' | MetricName
+
+export class AmazonQUITelemetryHelper {
```
nkomonen-amazon, line 11: This class feels like it can be generalized with the use of some generics

```
export class BaseTelemetryHelper<Data = { event: Event, time: number}, tabIDOrConversationId: string>
```
where the names and types in the generics wouldn't actually be those, but just to map your current implementation for this example.

And then also `UIMetrics` would be something else like `Datas` (the name sucks but something like this).

Then the whole thing could live in spans.ts and then if any other feature wants to be able to use the `listen` functionality they can make a quick subclass of the implementation for their own feature.

packages/core/src/amazonq/webview/messages/messageDispatcher.ts
```diff
@@ -33,6 +75,59 @@ export function dispatchWebViewMessagesToApps(
...
+             * that tab
+             */
+            const tabUIMetrics = AmazonQUITelemetryHelper.instance.getMetrics(tabID)
+            const conversationMetrics = AmazonQUITelemetryHelper.instance.getMetrics(conversationId)
+
+            // get metrics sorted by the time they were created
+            const allMetrics = [...tabUIMetrics, ...conversationMetrics].sort((a, b) => a[1] - b[1])
+
+            // Get the total duration by subtracting the last (finalRequestTime) and the first duration (initialRequestTime)
+            const totalDuration = allMetrics[allMetrics.length - 1][1] - allMetrics[0][1]
```
nkomonen-amazon, line 106: Is there any way we can get a named type to get the `time` instead of `[1]`?

packages/core/src/shared/telemetry/spans.ts
```diff
@@ -125,6 +126,11 @@ export type SpanOptions = {
     functionId?: FunctionEntry
 }
 
+type Listener = {
```
nkomonen-amazon, line 129: Can we add a docstring to explain what the purpose of a listener is

packages/core/src/amazonq/webview/messages/messageDispatcher.ts
```diff
@@ -10,11 +10,53 @@ import { TabType } from '../ui/storages/tabsStorage'
...
+     * We build up all the Amazon Q events we want to track by conversation ID
+     *
+     * This allows us to eventually connect metadata coming from a conversation ID with metadata coming from a tabID. Tabs know nothing
+     * of conversation IDs and conversation IDs have no knowledge of tabIDs except for in chat session storages.
+     *
+     * In the flow Webview -> Message Dispatcher -> Amazon Q Feature -> Amazon Q Feature Implementation -> Message Dispatcher -> Webview
+     * only "Amazon Q Feature Implementation" has knowledge about conversation IDs
+     * everything else has knowledge of tabIDs
+     */
+    telemetry.chat_roundTrip.listen(
```
nkomonen-amazon, line 37: nit: This is a stylistic preference but would it make sense to gather all the RTT related code in this file in to a class such as:

```
class QChatMessageDuration {
  static initializeListener()
  static startListening()
  static stopListening()
}
```

and then on line 37, 80 and 86 you can swap that whole implementation with the function call.

----

My reasoning is that if someone needs to add something else to the general functions there is less to parse

nkomonen-amazon (approved):

PR #5569, opened 2024-09-10
refactor(logging): remove unused channels

tomcat323:
## Problem
The `channel` logger was removed as part of: https://github.com/aws/aws-toolkit-vscode/pull/5564.
As a result the 'AWS toolkit' and 'Amazon Q' channels are no longer used, but they still show up in IDE output channels. They should be removed to maintain cleanliness. 

## Solution
Remove these two channels in `extension.ts` so they don't get initiated.  
Also remove input to logger activation since it's unused.

---


jpinkney-aws:
It looks like the build issues are probably because several places use `global.outputChannel` but now when they access that variable its undefined

tomcat323:
Upon further discussion, the output channel should be kept for "normal" messages, to show output from various features for normal application usage. This pull request is now closed

PR #5570, opened 2024-09-11
test(PolicyChecks): Implement IamPolicyChecks unit tests

kevluu-aws:
## Problem
IamPolicyChecks currently has no unit tests. This makes the feature fragile to changes and at risk to breaking changes.

## Solution
Implement unit tests for IamPolicyChecks feature.
Note: Functions that require `execFileSync` were skipped as it was not possible to stub the execution of the underlying CLI tools used.

---


packages/core/src/awsService/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -715,9 +715,9 @@ export class IamPolicyChecksWebview extends VueWebview {
     }
 
     public pushCustomCheckDiagnostic(diagnostics: vscode.Diagnostic[], finding: any, isBlocking: boolean) {
-        const message = `${finding.findingType}: ${finding.message} - Resource name: ${finding.resourceName}, Policy name: ${finding.policyName}`
+        let message = `${finding.findingType}: ${finding.message} - Resource name: ${finding.resourceName}, Policy name: ${finding.policyName}`
         if (message.includes('existingPolicyDocument')) {
-            message.replace('existingPolicyDocument', 'reference document')
+            message = message.replace('existingPolicyDocument', 'reference document')
```
justinmk3, line 720: this could remain `const` by using a ternary (which has readability benefits as well)

    const message = finding.findingType.includes('existingPolicyDocument') ? ... : ...

packages/core/src/test/awsService/accessanalyzer/iamPolicyChecks.test.ts
```diff
@@ -0,0 +1,793 @@
...
+import * as sinon from 'sinon'
+import {
+    _readCustomChecksFile,
+    getResultCssColor,
+    IamPolicyChecksWebview,
+    PolicyChecksError,
+    renderIamPolicyChecks,
+} from '../../../awsService/accessanalyzer/vue/iamPolicyChecks'
+import { FakeExtensionContext } from '../../../test/fakeExtensionContext'
+import fs from 'fs'
```
justinmk3, line 16: use our [fs.ts](https://github.com/aws/aws-toolkit-vscode/blob/4bb2fc1395d98ac71d90d83360c8f578ed0eccc2/packages/core/src/shared/fs/fs.ts) module wherever possible. Avoid using node APIs directly. Our fs.ts module should remove the need for that in most fs-related cases.

packages/core/src/test/awsService/accessanalyzer/iamPolicyChecks.test.ts
```diff
@@ -0,0 +1,793 @@
...
+        assert.strictEqual(result, fileContent)
+        assert(fsExistsSyncStub.calledWith(filePath))
+        assert(fsReadFileSyncStub.calledWith(filePath))
+    })
+
+    it('should throw PolicyChecksError for invalid S3 URI', async () => {
+        const invalidS3Uri = 's3://invalid-uri'
+
+        fsExistsSyncStub.withArgs(invalidS3Uri).returns(false)
+        parseS3UriStub.withArgs(invalidS3Uri).throws(new Error('Invalid S3 URI'))
```
justinmk3, line 86: why is this stubbed? does it depend on the AWS service? does it make network calls?

avoid stubs/mocks, they are fragile/brittle and lead to tests that do very little.

https://github.com/aws/aws-toolkit-vscode/blob/master/docs/CODE_GUIDELINES.md#test-guidelines

packages/core/src/test/awsService/accessanalyzer/iamPolicyChecks.test.ts
```diff
@@ -0,0 +1,793 @@
...
+        sandbox.stub(IamPolicyChecksWebview, 'editedDocumentFileName').value(document)
+        sandbox.stub(iamPolicyChecks, 'isTerraformPlan').returns(false)
+        sandbox.stub(iamPolicyChecks, 'isCloudFormationTemplate').returns(true)
+        executeCustomPolicyChecksCommandStub = sandbox.stub(fakePolicyChecksWebview, 'executeCustomPolicyChecksCommand')
+
+        await fakePolicyChecksWebview.checkAccessNotGranted(documentType, actions, resources, cfnParameterPath)
+
+        assert(
+            executeCustomPolicyChecksCommandStub.calledOnceWith({
+                command: 'cfn-policy-validator',
```
justinmk3, line 516: can you download the actual tool in CI and actually invoke it, instead of stubbing this?

kevluu-aws: I don't think we want to call the actual tool as it makes calls to an actual AWS service and requires credentials. Do we have a set of credentials we use for testing AWS services here? If we do, I think we can do it; the test will depend on the availability of the service in that case.

packages/core/src/awsService/accessanalyzer/vue/iamPolicyChecks.ts
kevluu-aws: Will update it, thanks!

packages/core/src/test/awsService/accessanalyzer/iamPolicyChecks.test.ts
justinmk3: currently our CI only has a Builder Id token, not IAM credentials

packages/core/src/test/awsService/accessanalyzer/iamPolicyChecks.test.ts
kevluu-aws: This does not make a call to the AWS service, I can remove the stub here.
Q: I have a mock for our AccessAnalyzer SDK client, is that ok? It looks like the checks are failing because of it:
```
Error: ENOENT: no such file or directory, open '/Users/runner/.aws/config'
```
Looks like the initialization of the client is trying to read that directory; is there a better way I should go about this when working with AWS clients?

kjs-aws (approved):

justinmk3:
> Functions that require `execFileSync` were skipped as it was not possible to stub the execution of the underlying CLI tools used.

Why is `execFileSync` being used in the app code? Our `ChildProcess` module should be used instead: https://github.com/aws/aws-toolkit-vscode/blob/5f68d0e255df5b57f87aa2bcd16009667f86b0fb/packages/core/src/shared/utilities/processUtils.ts#L69

packages/core/src/awsService/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -795,13 +799,14 @@ export async function renderIamPolicyChecks(context: ExtContext): Promise<void>
...
     } catch (err) {
         logger.error(err as Error)
     }
 }
 
 // Helper function to get document contents from a path
-async function _readCustomChecksFile(input: string): Promise<string> {
+export async function _readCustomChecksFile(input: string): Promise<string> {
     if (fs.existsSync(input)) {
         return fs.readFileSync(input).toString()
```
justinmk3, line 811: please migrate the app code to use our fs.ts module, do not use nodefs directly

packages/core/src/test/awsService/accessanalyzer/iamPolicyChecks.test.ts
```diff
@@ -0,0 +1,757 @@
...
+    PolicyChecksError,
+} from '../../../awsService/accessanalyzer/vue/iamPolicyChecks'
+import { globals } from '../../../shared'
+import { AccessAnalyzer, Config } from 'aws-sdk'
+import * as s3Client from '../../../shared/clients/s3Client'
+import { DefaultS3Client } from '../../../shared/clients/s3Client'
+import * as iamPolicyChecks from '../../../awsService/accessanalyzer/vue/iamPolicyChecks'
+import * as vscode from 'vscode'
+import { IamPolicyChecksConstants } from '../../../awsService/accessanalyzer/vue/constants'
+import { FileSystem } from '../../../shared/fs/fs'
```
justinmk3, line 21:
```suggestion
import fs from '../../../shared/fs/fs'
```

PR #5573, opened 2024-09-11
docs: explain outputChannel vs logOutputChannel

justinmk3:
---


tomcat323 (approved):
Thanks for the clarification!

PR #5575, opened 2024-09-11
fix(amazonq): Use the correct state for Q in remote environment

nkomonen-amazon:
## Problem:

As part of our design we intentionally separate the Auth state between a local environment, versus a remote (ssh'd) one. We do this so that if the user has 2 IDE instances, one local and the other remote, if one changes its auth state the other one will not be affected.

The reason for separating auth that I can remember is due to how we store the actual tokens on disk. Because the remote does not have access to the disk of the local, we need to ensure the remote operates its auth independently of the local auth.

By design globalState is shared by both local and remote instances. So for local we need to ensure we do not use the base globalState, otherwise the same Auth state as the remote will be used.

The problem, SecondaryAuth is not doing this for all cases. It is using the globalState when it should be asking Auth which state it should use. The happy path was working though, but there can be potential issues with this bug at some other point.

## Solution:

Expose the state that Auth is using so that something like SecondaryAuth can ask Auth for the correct state object depending on the local vs remote.

---

### DIFF
--- a/packages/core/src/auth/auth.ts
+++ b/packages/core/src/auth/auth.ts
@@ -1022,9 +1022,20 @@ export class Auth implements AuthService, ConnectionManager {
         }
     }
 
+    /**
+     * Auth uses its own state, separate from the default {@link globalState}
+     * that is normally used throughout the codebase.
+     *
+     * IMPORTANT: Anything involving auth should ONLY use this state since the state
+     * can vary on certain conditions. So if you see something explicitly using
+     * globalState verify if it should actually be using that.
+     */
+    public getStateMemento: () => vscode.Memento = () => Auth._getStateMemento()
+    private static _getStateMemento = once(() => getEnvironmentSpecificMemento())
+
     static #instance: Auth | undefined
     public static get instance() {
-        return (this.#instance ??= new Auth(new ProfileStore(getEnvironmentSpecificMemento())))
+        return (this.#instance ??= new Auth(new ProfileStore(Auth._getStateMemento())))
     }
 
     private getSsoProfileLabel(profile: SsoProfile) {
--- a/packages/core/src/auth/secondaryAuth.ts
+++ b/packages/core/src/auth/secondaryAuth.ts
@@ -3,8 +3,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import globals from '../shared/extensionGlobals'
-
 import * as vscode from 'vscode'
 import { getLogger } from '../shared/logger'
 import { cast, Optional } from '../shared/utilities/typeConstructors'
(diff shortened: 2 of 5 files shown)


hayemaxi:
Won't this log existing users out? It seems that the key that stores the last connection ID now lives somewhere else.

nkomonen-amazon:
> Won't this log existing users out? It seems that the key that stores the last connection ID now lives somewhere else.

@hayemaxi 

- this will not impact the Local IDE users as the state they were using will remain the same even with the changes in this pr
- this will not impact Remote IDE users since when we use `SecondaryAuth.activeConnection` it first checks the `#savedConnection` and then fallsback to the `#activeConnection` if needed. Before this PR in `SecondaryAuth.useNewConnection()` it would indirectly add the connection to both the `#savedConnection` and `#activeConnection`. But due to this bug, if a user also signed in on local, the `#savedConnection` in remote would become invalid (due to the remote incorrectly using the same state as the local). Though when retrieving the `#savedConnection` in the remote failed we fellback to `#activeConnection` and this was correct since `#activeConnection` is what `Auth` is using and that was already using the correct state.
  - So a summary would be that things coincidentally worked before but in this bugged scenario it was using the `#activeConnection` when we thought it was using the `#savedConnection`. Now it will correctly set the `#savedConnection` next time they log in.
  - I think also [this value in the recent telemetry change](https://github.com/aws/aws-toolkit-vscode/blob/bbeb410655d1cce3d8ca0af7e79383b7cd9e5438/packages/core/src/auth/secondaryAuth.ts#L275) may have been incorrectly representing the state. We should have used `SecondaryAuth.activeConnection` instead of the `#savedConnection` since the auth would actually fall back to the `#activeConnection` if necessary.

ivikash (approved):

packages/core/src/auth/auth.ts
```diff
@@ -1022,9 +1022,20 @@ export class Auth implements AuthService, ConnectionManager {
         }
     }
 
+    /**
+     * Auth uses its own state, separate from the default {@link globalState}
```
justinmk3, line 1026: not really separate, but nested in the globalState.

packages/core/src/auth/secondaryAuth.ts
```diff
@@ -214,10 +213,14 @@ export class SecondaryAuth<T extends Connection = Connection> {
...
 
+    public get state() {
+        return this.auth.getStateMemento()
+    }
+
     public async saveConnection(conn: T) {
         // TODO: fix this
         // eslint-disable-next-line aws-toolkits/no-banned-usages
-        await globals.context.globalState.update(this.key, conn.id)
+        await this.state.update(this.key, conn.id)
```
justinmk3, line 223: The eslint ignore comment can be removed

packages/core/src/auth/secondaryAuth.ts
```diff
@@ -251,7 +254,7 @@ export class SecondaryAuth<T extends Connection = Connection> {
     public async clearSavedConnection() {
         // TODO: fix this
         // eslint-disable-next-line aws-toolkits/no-banned-usages
-        await globals.context.globalState.update(this.key, undefined)
+        await this.state.update(this.key, undefined)
```
justinmk3, line 257: The eslint ignore comment can be removed

PR #5577, opened 2024-09-11
feat(amazonq): Add command to generate unit tests for selected code

bweedop:
Adds command to Amazon Q to generate unit tests for selected code. The feature will be limited to internal Amazon users to collect usage metrics and, subsequently, will be released to all users.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## Description

- Adds "Test" command to context menu (with a (Beta) suffix for while internal only). The command will send the selected code to the chat with the following prompt: "Generate unit tests for the following part of my code for me:"

image

## Checklist
- [X] My code follows the code style of this project
- [ ] I have added tests to cover my changes
- [] A short description of the change has been added to the **[CHANGELOG](https://github.com/aws/aws-toolkit-jetbrains/blob/master/CONTRIBUTING.md#contributing-via-pull-requests)** if the change is customer-facing in the IDE.
- [ ] I have added metrics for my changes (if required)
 
## License
I confirm that my contribution is made under the terms of the Apache 2.0 license.


packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -151,6 +151,7 @@ export class AuthUtil {
         }
 
         await setContext('aws.codewhisperer.connected', this.isConnected())
+        await setContext('aws.isInternalUser', this.isInternalAmazonUser())
```
jpinkney-aws, line 154: is this something that should live in the general activation since it's not necessarily relevant to codewhisperer?

packages/core/package.nls.json
```diff
@@ -111,6 +111,7 @@
     "AWS.command.amazonq.fixCode": "Fix",
     "AWS.command.amazonq.optimizeCode": "Optimize",
     "AWS.command.amazonq.sendToPrompt": "Send to prompt",
+    "AWS.command.amazonq.testCode": "Test (Beta)",
```
jpinkney-aws, line 114: Should this be something like `Generate Tests (Beta)` ? When I right click and see "Test" for some reason I think its going to automagically run my tests for me

packages/core/src/codewhispererChat/controllers/chat/userIntent/userIntentRecognizer.ts
```diff
@@ -36,6 +39,11 @@ export class UserIntentRecognizer {
             return UserIntent.APPLY_COMMON_BEST_PRACTICES
         } else if (prompt.message.startsWith('Optimize')) {
             return UserIntent.IMPROVE_CODE
+        } else if (
+            prompt.message.startsWith('Generate unit tests') &&
+            AuthUtil.instance.startUrl === 'https://amzn.awsapps.com/start'
```
hayemaxi, line 44: use `AuthUtil. isInternalAmazonUser()`

packages/core/src/codewhispererChat/commands/registerCommands.ts
```diff
@@ -102,6 +110,7 @@ export type EditorContextBaseCommandType =
     | 'aws.amazonq.fixCode'
     | 'aws.amazonq.optimizeCode'
     | 'aws.amazonq.sendToPrompt'
+    | 'aws.amazonq.testCode'
```
laileni-aws, line 113: nit: Can this be named more specific to unit test instead of generalized `testCode`?

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -196,6 +197,10 @@ export class AuthUtil {
         return this.conn !== undefined && isBuilderIdConnection(this.conn)
     }
 
+    public isInternalAmazonUser(): boolean {
+        return this.isConnected() && this.startUrl === 'https://amzn.awsapps.com/start'
```
laileni-aws, line 201: nitL Was this ok to expose startUrl to public?

bweedop: Yes. There has been a lot of back and forth on this (on Slack) and, ultimately, the decision is that this is okay.
 https://github.com/aws/aws-toolkit-jetbrains/pull/4869#discussion_r1750938610

packages/core/src/codewhispererChat/commands/registerCommands.ts
bweedop: Yes

packages/core/package.nls.json
bweedop: Yes. I've confirmed with product that "Generate tests (Beta)" makes sense.

packages/core/src/codewhisperer/util/authUtil.ts
bweedop: I think this makes sense.

packages/core/src/codewhisperer/util/authUtil.ts
ivikash: Can we move this utility function to a common place that can be use by Amazon Q and Toolkits.

packages/core/src/shared/vscode/setContext.ts
```diff
@@ -14,6 +14,7 @@ type contextKey =
     | 'aws.isDevMode'
     | 'aws.isSageMaker'
     | 'aws.isWebExtHost'
+    | 'aws.isInternalUser'
```
justinmk3, line 17: I think the key should be named after the feature. The fact that it's currently gated on "internal user" is not something that should be in the key name.

packages/core/src/auth/auth.ts
```diff
@@ -223,6 +245,8 @@ export class Auth implements AuthService, ConnectionManager {
         this.#onDidChangeActiveConnection.fire(conn)
         await this.store.setCurrentProfileId(id)
 
+        await setContext('aws.isInternalUser', this.isInternalAmazonUser())
```
justinmk3, line 248: I was wondering if `forgetConnection` and `expireConnection` should also un-set this. But it's probably useful to keep this hint around, once it was deduced. And in that case, should this flag be stored in globalState? Then we don't need to wait until a valid connection next time.

packages/core/src/auth/auth.ts
```diff
@@ -166,6 +168,26 @@ export class Auth implements AuthService, ConnectionManager {
         return this.#ssoCacheWatcher
     }
 
+    public get startUrl(): string | undefined {
+        // Reformat the url to remove any trailing '/' and `#`
+        // e.g. https://view.awsapps.com/start/# will become https://view.awsapps.com/start
```
justinmk3, line 173: would this comment make sense as the docstring for `reformatStartUrl` instead

packages/core/src/auth/auth.ts
```diff
@@ -166,6 +168,26 @@ export class Auth implements AuthService, ConnectionManager {
...
+        return isSsoConnection(this.activeConnection)
+            ? this.reformatStartUrl(this.activeConnection.startUrl)
+            : undefined
+    }
+
+    public isConnected(): boolean {
+        return this.activeConnection !== undefined
+    }
+
+    public reformatStartUrl(startUrl: string | undefined) {
```
justinmk3, line 183:
```suggestion
    public normalizeStartUrl(startUrl: string | undefined) {
```

justinmk3 (approved):
LGTM after comments. Needs approval from someone on your team too.

packages/core/src/auth/auth.ts
bweedop: I don't think it hurts to set this back to false on `forgetConnection` and `expireConnection`.

Some testing that I did made me doubt how `globalState` worked. Using `globalState.update` for this didn't work to enable the feature for internal users. `setContext`, however, worked. Are global keys detectable from package.json after being set?

laileni-aws (approved):

packages/core/src/auth/auth.ts
justinmk3: > Are global keys detectable from package.json after being set?

No

packages/core/src/auth/auth.ts
justinmk3: move the comment to normalizeStartUrl

packages/core/src/auth/auth.ts
bweedop: Oh okay well that would explain it. We need this to be set with `setContext`

packages/core/src/auth/auth.ts
bweedop: Done

dogusata (approved):

PR #5578, opened 2024-09-12
fix(toolkit): attach lifetime to generated ssh keys. 

Hweinstock:
## Problem

For connecting VSCode to EC2 instance, we generate an ssh key pair on disk. This results in writing the private ssh key to VSCode global storage, allowing the key to be potentially reused by other users on the same machine. 

## Solution  

Attach a lifetime to any key pair generated such that they wipe from disk after X seconds. Value is currently set is 30 seconds to allow connection to reliably establish. Also, change file permissions to read/write owner only and change behavior to overwrite existing keys.

Unable to test file permissions, due to VSCode file system unable to provide us with enough detail. Only provides whether it is readonly or not (somewhat unreliably: https://github.com/microsoft/vscode-discussions/discussions/673). VSCode file system is what is used in `fs.ts` implementation. 

This is in-line with how ec2 instance connect works: https://github.com/aws/aws-ec2-instance-connect-cli/blob/master/ec2instanceconnectcli/EC2InstanceConnectKey.py

---


justinmk3:
> Attach a lifetime to any key pair generated such that they wipe from disk after X seconds. Value is currently set is 30 seconds to allow connection to reliably establish. Also, change file permissions to read/write owner only and change behavior to overwrite existing keys.

Great idea. Would also be a good time to add a EC2-specific section here: https://github.com/aws/aws-toolkit-vscode/blob/791ae55021c0ee73dd84abaa57da46237d0fb0d5/docs/arch_features.md?plain=1#L14

That way, we have a high-level description of the steps that happen.

packages/core/src/awsService/ec2/sshKeyPair.ts
```diff
@@ -5,19 +5,32 @@
...
+
+    private constructor(
+        private keyPath: string,
+        lifetime: number
+    ) {
         this.publicKeyPath = `${keyPath}.pub`
+        this.lifeTimeout = new Timeout(lifetime)
+
+        this.lifeTimeout.onCompletion(async () => {
+            await this.delete()
```
justinmk3, line 23: Nice approach.

packages/core/src/awsService/ec2/sshKeyPair.ts
```diff
@@ -5,19 +5,32 @@
 import * as fs from 'fs-extra'
```
justinmk3, line 5: If it's not much trouble, it would be helpful to remove this and use our fs.ts module instead. We want to eliminate 'fs-extra'

justinmk3 (approved):

packages/core/src/awsService/ec2/sshKeyPair.ts
Hweinstock: I couldn't figure out how to set file permission (`fs.chmod` equivalent) through `fs.ts` but it appears redundant as ssh-keygen was already doing so properly so believe we're all good.

justinmk3: > how to set file permission (`fs.chmod` equivalent) through `fs.ts`

would you mind adding a stub for chmod in fs.ts? it can just do nothing for `if (isWeb())`. and for non-web, it can use [node's chmod](https://nodejs.org/api/fs.html#fspromiseschmodpath-mode) (from 'fs', instead of 'fs-extra)

packages/core/src/awsService/ec2/sshKeyPair.ts
```diff
@@ -26,14 +40,38 @@ export class SshKeyPair {
...
         return this.publicKeyPath
     }
 
+    public getPrivateKeyPath(): string {
+        return this.keyPath
+    }
+
     public async getPublicKey(): Promise<string> {
-        const contents = await fs.readFile(this.publicKeyPath, 'utf-8')
+        const contents = new TextDecoder().decode(await fs.readFile(this.publicKeyPath))
```
justinmk3, line 55:
```suggestion
        const contents = await fs.readFileAsString(this.publicKeyPath)
```

packages/core/src/awsService/ec2/sshKeyPair.ts
```diff
@@ -26,14 +40,38 @@ export class SshKeyPair {
...
         return contents
     }
+
+    public async delete(): Promise<void> {
+        await fs.delete(this.publicKeyPath)
+        await fs.delete(this.keyPath)
```
justinmk3, line 61: might be a good idea to delete the private key first, in case deleting the public key fails (which is less important)

packages/core/src/awsService/ec2/sshKeyPair.ts
```diff
@@ -26,14 +40,38 @@ export class SshKeyPair {
...
+
+        if (!this.lifeTimeout.completed) {
+            this.lifeTimeout.cancel()
+        }
+
+        this.deleted = true
+    }
+
+    public isDeleted(): boolean {
+        return this.deleted
```
justinmk3, line 71: can this extra state be avoided? usually better to check "reality", i.e. check the filesystem. after all, we're already storing the filepath.

Hweinstock: So in that case, replace the `isDeleted()` method with one that checks the actual filesystem?

PR #5581, opened 2024-09-12
fix(ec2): automatically poll pending nodes in explorer, refresh on change

Hweinstock:
redo of: https://github.com/aws/aws-toolkit-vscode/pull/3663 without formatting problems. 

## Problem
when changing instance status, the node will visually update, but only once. Thus, starting a node will leave it stuck in pending status until another refresh from elsewhere allows it to update. 
## Solution
We generalize the solution to that used in the `./src/apprunner/explorer/apprunnerNode.ts` into `packages/core/src/shared/utilities/pollingSet.ts`. 
- keep a set of items. 
- if the set is non-empty, run some `action` every `pollingInterval` seconds. 

In the context of ec2, this means:
- when an instance enters the `pending` status, we add it to the set. 
- Every 5, we check each node in the `pending` status to see if its status has updated. (action)
- If the node is no longer pending, remove it from the set and refresh its corresponding node in explorer. 
- Once no more nodes are pending, we cancel the timer. Only to be restart if a new one is added. 

Notes:
- When a node leaves the pending status, we only refresh that node, not the entire subtree. 
- We check every 5 seconds if the status changed, which involves an API call for each pending node. 
- We also remove the ability to start/stop an instance that is currently pending. 

## License


packages/core/src/awsService/ec2/explorer/ec2InstanceNode.ts
```diff
@@ -67,6 +76,10 @@ export class Ec2InstanceNode extends AWSTreeNodeBase implements AWSResourceNode
         }
     }
 
+    public getStatus(): string {
+        return this.instance.status!
```
hayemaxi, line 80: why would it be undefined? can we avoid that or make undefined a return type?

packages/core/src/shared/utilities/pollingSet.ts
```diff
@@ -0,0 +1,44 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { globals } from '..'
+
+export class PollingSet<T> extends Set<T> {
```
hayemaxi, line 8: Can we add a short docstring that describes what this does?

packages/core/src/awsService/ec2/explorer/ec2InstanceNode.ts
hayemaxi: also can we type status? maybe to `typeof Ec2Instance.status`

Hweinstock: > why would it be undefined? can we avoid that or make undefined a return type?

Yeah, I realized I did this in a few places. I refactored the code to use an interface `SafeEc2Instance` that won't let me do that anymore.

hayemaxi (approved):

packages/core/src/shared/clients/codecatalystClient.ts
```diff
@@ -668,9 +668,9 @@ class CodeCatalystClientInternal {
         }
 
         function failedStartMsg(serviceMsg?: string) {
-            const lastStatus = statuses[statuses.length - 1]?.status
+            const LastSeenStatus = statuses[statuses.length - 1]?.status
```
justinmk3, line 671: This is a variable name, should it be camelCase

packages/core/src/shared/clients/ec2Client.ts
```diff
@@ -13,9 +13,13 @@ import { Timeout } from '../utilities/timeoutUtils'
 import { showMessageWithCancel } from '../utilities/messages'
 import { ToolkitError, isAwsError } from '../errors'
 
-export interface Ec2Instance extends EC2.Instance {
-    name?: string
-    status?: EC2.InstanceStateName
+/**
+ * A wrapper around EC2.Instance where we can safely assume InstanceId field exists.
+ */
+export interface SafeEc2Instance extends EC2.Instance {
```
justinmk3, line 19: This is a common pattern and we did it for CodeCatalyst. We can just re-use the same names, don't need to add things like `Safe`. If there is ever a "conflict", modules can just namespace their imports:

    import * as ec2Client
    const foo: ec2Client.Ec2Instance

packages/core/src/shared/utilities/pollingSet.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * A useful abstraction that does the following:
+ * - keep a set of items.
+ * - if the set is non-empty, run some action every interval seconds.
+ * - once the set empties, clear the timer
+ * @param interval the interval in seconds
+ * @param action the action to perform
+ */
+export class PollingSet<T> extends Set<T> {
```
justinmk3, line 16: This reminds me of Go's `WaitGroup` (which I've also seen in other nodejs projects). https://pkg.go.dev/sync#WaitGroup

Not sure if it's exactly the same, but might be worth linking to https://pkg.go.dev/sync#WaitGroup from the docstring here for reference, in case we also need that concept in the future.

PR #5589, opened 2024-09-13
use crossfile as a fallback if utg is not support

Will-ShaoHua:
## Solution

---


Will-ShaoHua:
not needed anymore

PR #5590, opened 2024-09-13
refactor(codewhisperer): refactor UTG code path and add test coverage

Will-ShaoHua:
## Problem
- complicate regex being used which is not scalable and hard to understand
- stale and complicate code which lacks readability
- test coverage

### java

https://github.com/user-attachments/assets/46877435-057c-4c17-af91-4ffa27d93a0c

https://github.com/user-attachments/assets/356aa23d-a7ce-4f38-b0ff-32d8c0da8611

### no supported (kotlin)

https://github.com/user-attachments/assets/2e344169-3515-4de7-8379-32cbafaab194

---


Will-ShaoHua:
will upload a functional feature test screen recording

andrewyuq (approved):

packages/core/src/codewhisperer/util/supplementalContext/utgUtils.ts
```diff
@@ -184,41 +186,45 @@ async function getRelevantUtgFiles(editor: vscode.TextEditor): Promise<string[]>
...
+                getLogger().debug(
+                    `codewhisperer: error while guessing source file name from file ${testFileName} and pattern ${pattern}: ${err.message}`
+                )
+            }
+        }
+    }
+
+    return undefined
+}
+
```
Will-ShaoHua, line 215: test is here https://github.com/aws/aws-toolkit-vscode/pull/5590/files#diff-680eb5a5b5da4c826382a336b2fd23bcff6a7253488059b3ef2aaeae61507c76R48

packages/amazonq/test/unit/codewhisperer/util/codeParsingUtil.test.ts
```diff
@@ -57,6 +60,90 @@ describe('RegexValidationForJava', () => {
...
 describe('isTestFile', () => {
+    beforeEach(async function () {
+        tempFolder = (await createTestWorkspaceFolder()).uri.fsPath
+    })
+
+    it('validate by file path', async function () {
+        const langs = new Map<string, string>([
+            ['java', '.java'],
+            ['python', '.py'],
+            ['typescript', '.py'],
```
nkomonen-amazon, line 71: Should this be `.ts` and if so is there a flaw with the test if it passes? Does this mapping already exist outside of the tests that can be reused?

Will-ShaoHua: oh nice callout, checking

Will-ShaoHua: Nick, it works because this test case is to test "by file path", so as long as the file path includes any string of `tst`, `test`, `tests` then it will return true. 

But `.py` here is definitely a typo and i should fix it.

Will-ShaoHua: https://github.com/aws/aws-toolkit-vscode/pull/5590/files#diff-4c4d168f15256a5a919ec2fd9fc25c997275401d9bb66bc561c56b2005d900c9R76-R91

packages/core/src/codewhisperer/util/supplementalContext/utgUtils.ts
```diff
@@ -24,10 +24,12 @@ import { getOpenFilesInWindow } from '../../../shared/utilities/editorUtilities'
 import { getLogger } from '../../../shared/logger/logger'
 import { CodeWhispererSupplementalContext, CodeWhispererSupplementalContextItem, UtgStrategy } from '../../models/model'
 
-type UtgSupportedLanguage = keyof typeof utgLanguageConfigs
+const UTG_SUPPORTEDLANGUAGES: vscode.TextDocument['languageId'][] = ['java', 'python']
```
ivikash, line 27: Can we fetch the list of supported languages to BE. Because In order to extend the feature to additional languages we will rely on weekly release of IDEs, instead we should rely on Backend for this information so that we can make such decisions very dynamic.

Will-ShaoHua: Yea that's a good callout and discussed offline with Vikash, I think we're limited by the infra we have at this point of time, our service doesn't have this kinda of dynamic configuration API yet except the one for A/B testing. This comment also apply to other Q features as well like "auto trigger classifier co-efficients" https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/service/classifierTrigger.ts#L43  and other language level gated features.

packages/core/src/codewhisperer/util/supplementalContext/utgUtils.ts
```diff
@@ -184,41 +186,45 @@ async function getRelevantUtgFiles(editor: vscode.TextEditor): Promise<string[]>
...
+
+    for (const pattern of languageConfig.testFilenamePattern) {
+        try {
+            const match = testFileName.match(pattern)
+            if (match) {
+                return match[1] + match[2]
+            }
+        } catch (err) {
+            if (err instanceof Error) {
+                getLogger().debug(
```
justinmk3, line 206: is it expected that an exception would be thrown? if not, it indicates a bug, and that should be logged at "error" level. 

```suggestion
                getLogger().error(
```

Will-ShaoHua: yea that makes sense, updating it.

Will-ShaoHua: updated

packages/amazonq/test/unit/codewhisperer/util/codeParsingUtil.test.ts
```diff
@@ -57,6 +60,90 @@ describe('RegexValidationForJava', () => {
...
+
+        for (const name of srcFiles) {
+            const file = `${name}.${fileExt}`
+            const editor = await openATextEditorWithText('', file, tempFolder, { preview: false })
+            const actual = await isTestFile(editor.document.uri.fsPath, { languageId: editor.document.languageId })
+            assert.strictEqual(actual, false)
+        }
+    }
+
+    it('validate by file name', async function () {
```
justinmk3, line 125: if i'm reading the logs right, this test takes 10 seconds. that's not acceptable.

`openATextEditorWithText` also opens an editor, which can be slow. is that actually necessary, or could we add a param to it which only calls `vscode.workspace.openTextDocument` and skips `showTextDocument` (assuming I'm correct about the source of the slowness)

packages/core/src/codewhisperer/util/supplementalContext/utgUtils.ts
justinmk3, line 27: this is failing lint.
```suggestion
const utgSupportedLanguages: vscode.TextDocument['languageId'][] = ['java', 'python']
```

packages/amazonq/test/unit/codewhisperer/util/codeParsingUtil.test.ts
```diff
@@ -106,42 +112,48 @@ describe('isTestFile', () => {
...
-            const actual = await isTestFile(editor.document.uri.fsPath, { languageId: editor.document.languageId })
-            assert.strictEqual(actual, false)
+    async function assertIsTestFile(
+        filePaths: string[],
+        config: { languageId: PlatformLanguageId },
+        expected: boolean
+    ) {
+        for (const filePath of filePaths) {
+            const editor = createMockTextEditor('', filePath, config.languageId)
+            const actual = await isTestFile(editor.document.uri.fsPath, { languageId: config.languageId })
```
justinmk3, line 122: isTestFile only uses the document, so why is an editor needed?

Will-ShaoHua: that's a good callout, i should update the signature to use document rather than an editor

justinmk3: why create a *mock*? my [suggestion](https://github.com/aws/aws-toolkit-vscode/pull/5590#discussion_r1768792061) was simply to skip opening an *editor*. Using `vscode.workspace.openTextDocument` to create a real document (without opening an editor) should be fast, which avoids the CI slowness. That doesn't require using a mock.

Will-ShaoHua: isee

justinmk3 (approved):
> this test takes 10 seconds. ...
> 
> `openATextEditorWithText` opens an editor, which can be slow. Could we add a param to it which only calls `vscode.workspace.openTextDocument` and skips `showTextDocument`

That was implemented in the new `assertIsTestFile` function. I think it would be preferable to add that variant to `testUtil.ts`, but doesn't need to block this for now.

After that change, the total test time for the "validate by file name" tests is now 60 ms instead of 10 s. Nice!

Will-ShaoHua:
> That was implemented in the new assertIsTestFile function. I think it would be preferable to add that variant to testUtil.ts, but doesn't need to block this for now.

sgtm, i will add this to testUtil

justinmk3:
> sgtm, i will add this to testUtil

Oh, I didn't see this. I have a PR already for that

Will-ShaoHua:
> Oh, I didn't see this. I have a PR already for that
gotcha, thanks Justin:)

PR #5591, opened 2024-09-13
telemetry(core): implement tracing

jpinkney-aws:
## Problem
- We can't see an e2e view of our events

## Solution
- Implement telemetry tracing using the opentelemetry idea of trace ids.
- A trace id helps create a collection of spans.
- Only one trace id can be active in an execution flow at a time.
- Additionally, it swaps parentMetric for parentId so that we can directly correlate one telemetry event to another

## Other details
- A trace is associated with a top level telemetry event. The top level can be either event_trace or any other of our metrics. 
---

### DIFF
--- a/packages/core/src/shared/telemetry/spans.ts
+++ b/packages/core/src/shared/telemetry/spans.ts
@@ -24,6 +24,7 @@ import {
 } from '../errors'
 import { entries, NumericKeys } from '../utilities/tsUtils'
 import { PerformanceTracker } from '../performance/performance'
+import { randomUUID } from '../crypto'
 
 const AsyncLocalStorage: typeof AsyncLocalStorageClass =
     require('async_hooks').AsyncLocalStorage ??
@@ -145,6 +146,7 @@ export class TelemetrySpan<T extends MetricBase = MetricBase> {
         passive: true,
         requiredMetadata: [],
     }
+    private readonly spanId: string
 
     /**
      * These fields appear on the base metric instead of the 'metadata' and
@@ -166,6 +168,9 @@ export class TelemetrySpan<T extends MetricBase = MetricBase> {
                 this.definition.trackPerformance && (options?.emit ?? false) // only track the performance if we are also emitting
             ),
         }
+
+        this.spanId = randomUUID()
+        this.record({ spanId: this.spanId } as any)
     }
 
     public get startTime(): Date | undefined {
@@ -261,6 +266,10 @@ export class TelemetrySpan<T extends MetricBase = MetricBase> {
         }
     }
 
+    getSpanId() {
+        return this.spanId
+    }
+
     /**
      * Creates a copy of the span with an uninitialized start time.
      */
@@ -269,7 +278,9 @@ export class TelemetrySpan<T extends MetricBase = MetricBase> {
     }
 }
 
-type Attributes = Partial<MetricShapes[MetricName]>
+type Attributes = Partial<MetricShapes[MetricName]> & {
+    traceId?: string
+}
 
 interface TelemetryContext {
     readonly spans: TelemetrySpan[]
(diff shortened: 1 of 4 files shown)


packages/core/src/shared/telemetry/spans.ts
```diff
@@ -364,6 +375,34 @@ export class TelemetryTracer extends TelemetryBase {
         }
     }
 
+    /**
+     * Trace represents a higher level event that you want to link with any other spans. It generates
+     * a traceId which is added to all subsequent calls, allowing you to query for all related events.
+     *
+     * Functionally, it is similiar to run. The only difference is the traceId gets added to subsequent telemetry calls.
```
hayemaxi, line 382: Why not just replace `run` completely with `trace` (or add the login to run directly). We often won't know to make a `run` a `trace` until its too late, and it would be easy for folks to use forget or not know to use `trace`.

packages/core/src/shared/telemetry/spans.ts
```diff
@@ -364,6 +375,34 @@ export class TelemetryTracer extends TelemetryBase {
...
+     * @param fn the callback function
+     */
+    private trace<T, U extends MetricName>(name: U, fn: (span: Metric<MetricShapes[U]>) => T): T {
+        return this.runRoot(() => {
+            /**
+             * Make sure we aren't starting a trace inside of a trace. traceId's have to be unique for every
+             * flow, otherwise we can't really tell what the trace belongs to
+             */
+            if (this.attributes?.traceId) {
+                throw new Error('Cannot trace an already tracing telemetry event')
```
hayemaxi, line 396: related to my comment about making this in `run` by default.

Instead of throwing an error we can just do nothing/use the already existing traceId. It can be extremely hard to tell the stack trace of a call in advance- we may be already tracing upstream.

packages/core/src/shared/telemetry/spans.ts
jpinkney-aws: I think they recommend against in opentelemetry but I think it might make sense here since we have so many starting/ending flows vs traditional software

dhruvigajjar: how do you decided end-to-end of traceId in user journey flow? can you give an example

jpinkney-aws: A traceId is just a uuid. If its the most top level telemetry event in a flow it creates the traceId. Otherwise it adds the traceId from the parent metric.

For example, a vscode_executeCommand can contain contain code that does amazonq_enterFocusChat, amazonq_openChat.

When vscode_executeCommand runs it generates trace id (which is a random uuid e.g.): 'aaaaa-aaaaa-aaaaa-aaaaa-aaaaa' this trace id gets attached to amazonq_enterFocusChat, amazonq_openChat so we get the final 3 metrics:

vscode_executeCommand:
traceId: 'aaaaa-aaaaa-aaaaa-aaaaa-aaaaa'

amazonq_enterFocusChat
traceId: 'aaaaa-aaaaa-aaaaa-aaaaa-aaaaa'

amazonq_openChat
traceId: 'aaaaa-aaaaa-aaaaa-aaaaa-aaaaa'

then in our telemetry instance you can look up traceId: 'aaaaa-aaaaa-aaaaa-aaaaa-aaaaa' and see all the events associated with that particular run. The next time vscode_executeCommand gets ran a new traceId is generated and then attached to the subsequent metrics. It's just following the ideas from https://opentelemetry.io/docs/concepts/signals/traces/

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1178,6 +1183,21 @@
                     "required": true
                 }
             ]
+        },
+        {
+            "name": "trace_event",
```
justinmk3, line 1188: Is there a reason not to use a conventional name such as `toolkit_trace` ? The prefix is useful as a "category". If we introduce new prefixes, there should be a reason.

packages/core/src/test/shared/telemetry/spans.test.ts
```diff
@@ -330,16 +341,102 @@ describe('TelemetryTracer', function () {
...
+                    })
+                    telemetry.vscode_executeCommand.run(() => {
+                        telemetry.amazonq_enterFocusConversation.run(() => {
+                            telemetry.amazonq_exitFocusConversation.run(() => {})
+                        })
+                    })
+                    telemetry.amazonq_closeChat.emit({
+                        result: 'Succeeded',
+                    })
+                })
```
justinmk3, line 430: This is a nicely structured test!

justinmk3:
The insights from https://github.com/aws/aws-toolkit-vscode/pull/5591#discussion_r1764436226 (plus the opentelemetry URL) would be useful in the span.ts docstring, plus https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md

PR #5593, opened 2024-09-16
[wip] Cwspr language refactor

Will-ShaoHua:
## Problem
Q code path lacks an unified language interface so that we can easily configure language level gated feature and plus there are many different language identifier strings which makes it really confusing when and which should be used

## Solution

---


packages/core/src/codewhisperer/util/language/LanguageBase.ts
justinmk3, line 1: file name should be camelCase

Will-ShaoHua: gotcha lol, thanks for the reminder Justin

PR #5607, opened 2024-09-17
feat(amazonq): allow users to skip tests

dhasani23:
## Problem

Some users transform projects with long-running tests.

## Solution

Allow users to skip tests by prompting them via chat and passing the relevant flag to skip tests through the `manifest.json`.

---


dhasani23:
image
image

dhasani23:
Note: the backend work for this feature is not done yet, so we don't want to merge it yet.

packages/core/src/codewhisperer/service/transformByQ/transformMavenHandler.ts
```diff
@@ -59,28 +59,7 @@ function installProjectDependencies(dependenciesFolder: FolderInfo, modulePath:
...
-            } else {
-                errorReason = `Maven ${argString}: InstallationSpawnError`
-            }
-            if (spawnResult.error) {
-                const errorCode = (spawnResult.error as any).code ?? 'UNKNOWN'
-                errorReason += `-${errorCode}`
-            }
-
-            // Explicitly set metric as failed since no exception was caught
-            telemetry.record({ result: MetadataResult.Fail, reason: errorReason })
```
dhasani23, line 82: All of the above logic removed (same as below) was entirely just to form the errorReason string. We include all of that info in the IDE logs already, and we can simply put the error code in telemetry to simplify.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -334,6 +340,24 @@ export class GumbyController {
         }
     }
 
+    private async handleSkipTestsSelection(message: any) {
+        const skipTestsSelection = message.formSelectedValues['GumbyTransformSkipTestsForm']
+        // one of these must be true
+        if (skipTestsSelection === CodeWhispererConstants.skipUnitTestsMessage) {
+            transformByQState.setCustomBuildCommand(CodeWhispererConstants.skipUnitTestsBuildCommand)
+        } else if (skipTestsSelection === CodeWhispererConstants.doNotSkipUnitTestsMessage) {
```
JuliaABurch, line 348: Is it worth just making this an `else` and not skipping unit tests by default? What happens if `setCustomBuildCommand` is never called?

dhasani23: Yes, will fix. Our default should be to not skip tests.

packages/core/src/amazonqGumby/activation.ts
```diff
@@ -48,7 +52,9 @@ export async function activate(context: ExtContext) {
 
         Commands.register('aws.amazonq.stopTransformationInHub', async (cancelSrc: CancelActionPositions) => {
             if (transformByQState.isRunning()) {
-                void stopTransformByQ(transformByQState.getJobId(), cancelSrc)
+                await stopTransformByQ(transformByQState.getJobId(), cancelSrc)
+                await postTransformationJob()
+                await cleanupTransformationJob()
```
dhasani23, line 57: Just like we do when a user stops a job from the chat, we should run `postTransformationJob()` and `cleanupTransformationJob()` after a user stops a job from the Transformation Hub.

JuliaABurch (approved):
Approved pending final string review

packages/core/src/codewhisperer/models/model.ts
```diff
@@ -316,7 +316,7 @@ export class ZipManifest {
     version: string = '1.0'
     hilCapabilities: string[] = ['HIL_1pDependency_VersionUpgrade']
     transformCapabilities: string[] = ['EXPLAINABILITY_V1']
-    customBuildCommand: string = ''
+    customBuildCommand: string = 'test'
```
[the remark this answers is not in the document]
dhasani23: updated to `clean test`

packages/core/src/codewhisperer/service/transformByQ/transformMavenHandler.ts
```diff
@@ -59,28 +59,7 @@ function installProjectDependencies(dependenciesFolder: FolderInfo, modulePath:
...
-
+            telemetry.record({ result: MetadataResult.Fail, reason: (spawnResult.error as any).code })
```
hayemaxi, line 62: Why isn't this section using `telemetry.run` instead? Also, can you validate the type/it exists instead of casting to `any`?

dhasani23: It does use `telemetry.run`: https://github.com/aws/aws-toolkit-vscode/blob/3b44bcb4bbc1d6d984c8dd3319e4c686fa9fc7e0/packages/core/src/codewhisperer/service/transformByQ/transformMavenHandler.ts#L19

And yeah I think actually it makes more sense to use the `message` field here anyway, so fixed that too.

hayemaxi: Hmm, needs verification but I believe manually setting `result: Fail, reason: ...` inside of `run()` does not do anything. It automatically sets these values based on throwing an uncaught exception in the `run` block. `result` and `reason` will be overwritten by the contents of the thrown error on the next line (or they would become blank if no error was thrown).

hayemaxi (approved):

PR #5610, opened 2024-09-17
feat(logging): topic header for logs

tomcat323:
## Problem
Log message can appear in output channels(such as Amazon Q Logs as shown), IDE Debug console, or write to pre-defined log file destinations. They come from all parts of our services and can amount to large and lengthy chunks quickly.

We don’t have a reliable way to identify log messages from various services and modules of our product. Some messages might have headers while others don’t, and the existing headers are hard-coded on a message by message basis, with no checks for consistency. 

1. It’s difficult to locate and trace unidentified log messages when the size of logs grow quickly.
2. Without regulation, the hard-coded custom headers can lead to legacy issues down-the-road and inconsistencies within the codebase.

## Solution
1. Make message topic identifiers an input variable of the getLogger() function.
2. Append the topic at the front of each message, for example:
`getLogger('S3').error('Something went wrong')`
Would print:
`S3: Something went wrong`
3. Define a set of topics to be used in message headers, covering all services and modules.

## Implementation notes:
1. The `WinstonToolkitLogger` is renamed to `ToolkitLogger` to avoid confusion, since its our own logger not winston's. The log topic feature is added here. 
2. A `baseLogger` abstract class is created to make the existing `NullLogger`, `ConsoleLogger`, and `ToolkitLogger` cleaner and their responsibilities more obvious and compact.
3. Currently the topics are optional, and the values are just `Test` and `Unknown`, this will be expanded during migration of existing logger calls.

---

### DIFF
--- a/packages/core/src/shared/logger/logger.ts
+++ b/packages/core/src/shared/logger/logger.ts
@@ -25,6 +25,38 @@ export interface Logger {
     enableDebugConsole(): void
 }
 
+export abstract class baseLogger implements Logger {
+    debug(message: string | Error, ...meta: any[]): number {
+        return this.sendToLog('debug', message, meta)
+    }
+    verbose(message: string | Error, ...meta: any[]): number {
+        return this.sendToLog('verbose', message, meta)
+    }
+    info(message: string | Error, ...meta: any[]): number {
+        return this.sendToLog('info', message, meta)
+    }
+    warn(message: string | Error, ...meta: any[]): number {
+        return this.sendToLog('warn', message, meta)
+    }
+    error(message: string | Error, ...meta: any[]): number {
+        return this.sendToLog('error', message, meta)
+    }
+    log(logLevel: LogLevel, message: string | Error, ...meta: any[]): number {
+        return this.sendToLog(logLevel, message, ...meta)
+    }
+    abstract setLogLevel(logLevel: LogLevel): void
+    abstract logLevelEnabled(logLevel: LogLevel): boolean
+    abstract getLogById(logID: number, file: vscode.Uri): string | undefined
+    /** HACK: Enables logging to vscode Debug Console. */
+    abstract enableDebugConsole(): void
+
+    abstract sendToLog(
+        logLevel: 'debug' | 'verbose' | 'info' | 'warn' | 'error',
+        message: string | Error,
+        ...meta: any[]
+    ): number
+}
+
 /**
  * Log levels ordered for comparison.
  *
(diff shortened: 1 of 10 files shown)


packages/core/src/shared/logger/logger.ts
```diff
@@ -85,107 +111,91 @@ export function compareLogLevel(l1: LogLevel, l2: LogLevel): number {
...
+    if (isToolkitLogger(logger)) {
+        logger.setTopic(topic)
+    }
     return logger
 }
 
-export function getDebugConsoleLogger(): Logger {
-    return toolkitLoggers['debugConsole'] ?? new ConsoleLogger()
+/**
+ * check if the logger is of type `ToolkitLogger` while not causing dependency loop
```
nkomonen-amazon, line 132:
```suggestion
 * check if the logger is of type `ToolkitLogger`. This avoids Circular Dependencies, but `instanceof ToolkitLogger` is preferred.
```

packages/core/src/shared/logger/toolkitLogger.ts
```diff
@@ -126,6 +108,31 @@ export class WinstonToolkitLogger implements Logger, vscode.Disposable {
...
+         */
+        if (this.topic === 'Unknown') {
+            return message
+        }
+        const topicPrefix = `${this.topic}: `
+        if (typeof message === 'string') {
+            return topicPrefix + message
+        } else if (message instanceof Error) {
+            /* Create a new Error object to avoid modifying the original */
+            const topicError = new Error(topicPrefix + message.message)
```
nkomonen-amazon, line 128: We have wrappers around errors that can have additional properties, see ToolkitError and the `code` property.

If possible can you modify the message on the given error, but I forget if that is read only.

Otherwise look to use ToolkitError.chain, and then make the `cause` the given error

nkomonen-amazon (changes requested):

packages/core/src/shared/logger/toolkitLogger.ts
tomcat323: Thanks for pointing out the other extended errors. The messages appears to be read only, looking into the chain method.

tomcat323: There are two problems with `ToolkitError.chain`:
1. it only applies to `toolkitErrors`, there are other custom error types that doesn't support chaining
2. the chaining is flaky because we have another `info` object defined within `toolkitErrors` with information such as `code` and `cause`. This causes duplication when printing out errors:
Screenshot 2024-09-18 at 11 10 49 AM

Solution:
1. When adding header to errors, create another object to wrap around the error with topicHeader being the first entry. This cleanly appends the topic without any overhead, and can be applied to any type of error:
screenshot_2024-09-18_at_10 43 19___am_720
2. Remove the `info` object within the `toolkitErrors` constructor, keep the individual variables, so it doesn't get duplicated when printing.

tomcat323:
For context this implementation is resulted from a discussion in a previous closed PR:
Screenshot 2024-09-18 at 11 17 43 AM

packages/core/src/shared/logger/toolkitLogger.ts
nkomonen-amazon: > it only applies to toolkitErrors, there are other custom error types that doesn't support chaining

I'm not sure I understand this, can't we put any error we want inside the chain as an argument?

But regardless I like the new `topic` field, its much cleaner. Is it possible to use `topicPrefix.split(':', 1)[0]` to trim the `: `?

packages/core/src/shared/logger/toolkitLogger.ts
```diff
@@ -113,22 +113,22 @@ export class ToolkitLogger extends BaseLogger implements vscode.Disposable {
...
         if (typeof message === 'string') {
-            return topicPrefix + message
+            return topic + message
         } else if (message instanceof Error) {
-            /* Create a new Error object to avoid modifying the original */
-            const topicError = new Error(topicPrefix + message.message)
-            topicError.name = message.name
-            topicError.stack = message.stack
-            return topicError
+            const errorWithTopic = {
```
nkomonen-amazon, line 127: If our error object has other fields like `name` and `code` those would be lost by the looks of it.

Instead see if you can do:
```
(errorWithTopic as any).topic = topic
```
as this will forcefully add the value in.

Then you would revert the `object` return type back in to error.

---

Separately if we drop the `topic` variable we can skip needing to trim the `: ` from this topic value.
So instead on line 125 do `${this.topic}: ${this.message}`
and then below you can just do 
```
(errorWithTopic as any).topic = this.topic
```

packages/core/src/shared/logger/toolkitLogger.ts
```diff
@@ -156,7 +156,7 @@ export class ToolkitLogger extends BaseLogger implements vscode.Disposable {
 
         meta = meta.map((o) => (o instanceof Error ? this.mapError(level, o) : o))
 
-        if (messageWithHeader instanceof Error) {
+        if (typeof messageWithHeader === 'object' && messageWithHeader !== null) {
```
nkomonen-amazon, line 159: Checking object in this case is a bit of a hack so we want to avoid this when possible. My comment above should allow us to revert this.

packages/core/src/shared/logger/toolkitLogger.ts
tomcat323: No doing so would not lose the fields like `name` and `code`, see:
Screenshot 2024-09-18 at 11 23 23 AM

packages/core/src/shared/logger/toolkitLogger.ts
tomcat323: True, the chain would work on others as well, its just that we have to make the topic error a `toolkitError` as well.
It also has the problem of additional location statements (those `at` lines)

Glad you like the new field, will remove the `: `.

packages/core/src/shared/logger/toolkitLogger.ts
tomcat323: I'm not sure if I completely understand your solution. But I agree with avoiding this check.

What do you think about declaring a type:
`type LogMessage = { type: 'string'; content: string } | { type: 'error'; topic: string; error: Error }
`?

So that the `addTopicToMessage` would return a `LogMessage`, and for this check we can have a switch of:
`switch (logMessage.type) {
        case 'string':
            this.logger.log(level, logMessage.content, ...meta, { logID: this.idCounter })
            break
        case 'error':
            this.logger.log(level, '%O', { topic: logMessage.topic, message: logMessage.error }, ...meta, { logID: this.idCounter })
            break
    }`

packages/core/src/shared/logger/toolkitLogger.ts
nkomonen-amazon: If we did what I mentioned above I think that would be cleaner as it looks like a regular error (except with an additional `topic` field), `instanceof Error` would still work, and the output wouldn't have a nested `message` property.

tomcat323: This might cause the topic field to not be printed first, which is not idea since we want the topic to be the "header"

Upon discussion we are creating a new class to handle the errors with topic identifier. That way we ensures the ordering and allows us todo `instanceof`

packages/core/src/shared/logger/toolkitLogger.ts
```diff
@@ -113,22 +120,19 @@ export class ToolkitLogger extends BaseLogger implements vscode.Disposable {
...
-                topic,
-                message,
+            /*We shouldn't print unknow before current logging calls are migrated
+             * TODO: remove this once migration of current calls is completed
+             */
+            if (this.topic === 'unknown') {
+                return message
             }
-            return errorWithTopic
+            return topicPrefix + message
```
nkomonen-amazon, line 132: See the second part of my comment here: https://github.com/aws/aws-toolkit-vscode/pull/5610#discussion_r1765293320

We can get rid of the split down below

nkomonen-amazon (approved):

PR #5613, opened 2024-09-17
feat(stepfunctions): Custom editor to view and edit ASL files - tracer bullet

witness-me:
## Problem
ASL files, represented as *.asl.json and *.asl.yaml, are rendered using the default JSON editor

## Solution
Use existing SFN visual custom editor in VSCode to render ASL files in a visual builder, similar to how it would be displayed in SFN console in create/edit modes.

Current pull request is a tracer bullet implementation without support for local file editing - this functionality will be added in the following changes 

---

### DIFF
--- a/packages/core/src/shared/telemetry/vscodeTelemetry.json
+++ b/packages/core/src/shared/telemetry/vscodeTelemetry.json
@@ -481,6 +481,16 @@
             "description": ""
         },
         {
+            "name": "stepfunctions_openWorkflowStudio",
+            "description": "Called after opening asl file in Step Functions Workflow Studio",
+            "metadata": [
+                {
+                    "type": "id",
+                    "required": true
+                }
+            ]
+        },
+        {
             "name": "vscode_activeRegions",
             "description": "Record the number of active regions at startup and when regions are added/removed",
             "unit": "Count",
--- a/packages/core/src/stepFunctions/workflowStudio/workflowStudioEditorProvider.ts
+++ b/packages/core/src/stepFunctions/workflowStudio/workflowStudioEditorProvider.ts
@@ -1,0 +1,157 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+import { getLogger } from '../../shared/logger'
+import request from '../../shared/request'
+import fs from '../../shared/fs/fs'
+import { getClientId } from '../../shared/telemetry/util'
+import { telemetry } from '../../shared/telemetry/telemetry'
+import globals from '../../shared/extensionGlobals'
+import { getRandomString, getStringHash } from '../../shared/utilities/textUtilities'
+import { ToolkitError } from '../../shared/errors'
+import { WorkflowStudioEditor } from './workflowStudioEditor'
+
+// TODO: switch to production mode: change isLocalDev to false and add CDN link
+const isLocalDev = true
+const localhost = 'http://127.0.0.1:3002'
+const cdn = 'TBD'
+let clientId = ''
+
+/**
+ * Provider for Workflow Studio editors.
+ *
+ * Workflow Studio editor is used for `.asl.json`, `.asl.yaml` and `.asl.yml` files.
+ * To get started, run this extension and open any file with one of these extensions in VS Code.
+ */
+export class WorkflowStudioEditorProvider implements vscode.CustomTextEditorProvider {
+    public static readonly viewType = 'workflowStudio.asl'
+
+    /**
+     * Registers a new custom editor provider for `.tc.json` files.
+     * @remarks This should only be called once per extension.
+     * @param context The extension context
+     */
+    public static register(context: vscode.ExtensionContext): vscode.Disposable {
+        const provider = new WorkflowStudioEditorProvider(context)
+        return vscode.window.registerCustomEditorProvider(WorkflowStudioEditorProvider.viewType, provider, {
+            webviewOptions: {
+                enableFindWidget: true,
+                retainContextWhenHidden: true, // Retain content when switching tabs
+            },
+        })
+    }
+
+    protected extensionContext: vscode.ExtensionContext
+    protected webviewHtml: string
+    protected readonly managedVisualizations = new Map<string, WorkflowStudioEditor>()
+    protected readonly logger = getLogger()
+
+    constructor(context: vscode.ExtensionContext) {
+        this.extensionContext = context
+        this.webviewHtml = ''
+    }
+
+    /**
+     * Fetches the webview HTML from the CDN or local server.
+     * @private
+     */
+    private async fetchWebviewHtml() {
+        const source = isLocalDev ? localhost : cdn
+        const response = await request.fetch('GET', `${source}/index.html`).response
+        this.webviewHtml = await response.text()
+
+        for (const visualization of this.managedVisualizations.values()) {
+            await visualization.refreshPanel(this.extensionContext)
+        }
+    }
+
+    /**
+     * Gets the webview content for Workflow Studio.
+     * @private
+     */
+    private getWebviewContent = async () => {
+        if (!this.webviewHtml) {
+            await this.fetchWebviewHtml()
+        }
+        let htmlFileSplit = this.webviewHtml.split('<head>')
+
+        // Set asset source to CDN
+        const source = isLocalDev ? localhost : cdn
+        const baseTag = `<base href='${source}'/>`
+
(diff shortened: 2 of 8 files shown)


witness-me:
Screenshot demo: https://maxis-file-service-prod-pdx.pdx.proxy.amazon.com/issues/871da43c-f86e-4fb5-ad56-987bf6af7368/attachments/28e9884e58f95760c86d0ea87bb43c9ccde556ee752a22e43bfd241ac9c5c5bc_7bace10f-5170-4498-827e-520bdfa96f76

packages/core/src/stepFunctionsWorkflowStudio/WorkflowStudioEditor.ts
```diff
@@ -0,0 +1,180 @@
...
+            defaultTemplateName: this.defaultTemplateName,
+            loaderNotification: undefined,
+            fileId: this.fileId,
+        }
+
+        async function cancelOpenInWorkflowStudio(reason: string) {
+            contextObject.panel.dispose()
+            await vscode.commands.executeCommand('vscode.openWith', documentUri, 'default')
+
+            telemetry.workflowStudio_closed.emit({
```
jpinkney-aws, line 107: Is knowing that people clicked the "closed" button worth gathering?

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -480,6 +480,26 @@
...
+            "description": "Called after opening asl file in Step Functions Workflow Studio",
+            "metadata": [
+                {
+                    "type": "id",
+                    "required": true
+                }
+            ]
+        },
+        {
+            "name": "workflowStudio_closed",
```
justinmk3, line 494: is this not specific to SFN? use the existing `stepfunctions_` prefix if so. same for the other metric.

also, after the underscore, the first word is typically a "verb".

```suggestion
            "name": "stepfunctions_closeWorkflowStudio",
```

packages/core/src/stepFunctionsWorkflowStudio/WorkflowStudioEditor.ts
```diff
@@ -0,0 +1,180 @@
+/*!
```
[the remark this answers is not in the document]
justinmk3: > Also you don't need to capitalize file names

Yes. In general, please notice the conventions of the codebase so that we don't have to mention them in the PR review. That sames time. 

Please review https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#develop and the documents linked there.

packages/core/src/stepFunctionsWorkflowStudio/WorkflowStudioEditor.ts
```diff
@@ -0,0 +1,180 @@
...
+                    // Note that a single text document can also be shared between multiple custom
+                    // editors (e.g. this can happen when you split a custom editor)
+
+                    // When the panel is closed, dispose of any disposables/remove subscriptions
+                    const disposePanel = async () => {
+                        if (this.isPanelDisposed) {
+                            return
+                        }
+
+                        await telemetry.workflowStudio_closed.run(async (span) => {
```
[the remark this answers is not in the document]
justinmk3: For example, this run() could enclose the entire body of the `disposePanel`.

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -480,6 +480,26 @@
             "name": "stepfunctions_previewstatemachine",
             "description": ""
         },
+        {
+            "name": "workflowStudio_opened",
```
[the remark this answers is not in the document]
witness-me: I see, I assume they should go here eventually
https://github.com/aws/aws-toolkit-common/blob/main/telemetry/definitions/commonDefinitions.json

Does it work if I first add the metrics one by one here for the ease of development to avoid handling multiple repositories and then move all of them to the common package? I will create a task for it to make sure this is not missed

packages/core/src/stepFunctionsWorkflowStudio/WorkflowStudioEditor.ts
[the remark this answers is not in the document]
witness-me: I was thinking of having it separately as this is quite different from existing SFN integrations, but I belive having this logic in a subfolder (as you suggested) should work too, as long as we keep it completely separated from existing SFN actions

packages/core/src/stepFunctionsWorkflowStudio/WorkflowStudioEditor.ts
witness-me: I took a look again and agree that in this case collecting it (the number of times the tab is closed) does not provide much value, I think it makes sense to remove it from here.
At the same time, keeping the same metric for the case when the user cancels opening the integration does make sense in my opinion, this should help us calculate the adoption rate

packages/core/src/stepFunctionsWorkflowStudio/WorkflowStudioEditor.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * Sets up the webview panel for Workflow Studio Editor. This includes creating the
+     * panel, setting up the webview content, and handling the communication between the webview
+     * and the extension context.
+     * @param textDocument The text document to be displayed in the webview panel.
+     * @param context The extension context.
+     * @private
+     */
+    private setupWebviewPanel(textDocument: vscode.TextDocument, context: vscode.ExtensionContext) {
```
ParidelPooya, line 89: The setupWebviewPanel method wraps the resolution of a promise in another promise. The inner promise is redundant.

packages/core/src/stepFunctionsWorkflowStudio/WorkflowStudioEditor.ts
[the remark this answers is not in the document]
justinmk3: > having this logic in a subfolder (as you suggested) should work too,


sounds good.

packages/core/src/stepFunctionsWorkflowStudio/WorkflowStudioEditor.ts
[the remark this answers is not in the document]
witness-me: Thanks for the explanation. I'll remove this specific metric (as we discussed in another thread that it is not very useful), but I will keep this in mind for adding future metrics

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -480,6 +480,26 @@
...
+            "metadata": [
+                {
+                    "type": "id",
+                    "required": true
+                }
+            ]
+        },
+        {
+            "name": "stepfunctions_cancelOpenWorkflowStudio",
+            "description": "Called after cancelling opening asl file in Step Functions Workflow Studio",
```
justinmk3, line 495: It's never necessary to have a separate cancel event. All events can have result=success/fail/cancel. 

Please review https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md

packages/core/src/stepFunctions/workflowStudio/workflowStudioEditorProvider.ts
```diff
@@ -0,0 +1,160 @@
...
+                        document,
+                        webviewPanel,
+                        this.extensionContext,
+                        fileId,
+                        this.getWebviewContent
+                    )
+                    this.handleNewVisualization(document.uri.fsPath, newVisualization)
+                } catch (err) {
+                    void vscode.window.showInformationMessage(i18n('AWS.stepFunctions.workflowStudio.errors.rendering'))
+                    this.logger.error(`WorkflowStudio: Unable to open file. ${err}`)
```
justinmk3, line 145: If `resolveCustomTextEditor` is called as the result of a vscode command in any form (which is usually the case,) then Toolkit already has a default handler that logs the error and shows a message. It's not necessary and discouraged to repeat that logic in this way, because it leads to an incosistent UX. 

Please review the guidlines here: https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#develop

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -480,6 +480,16 @@
             "name": "stepfunctions_previewstatemachine",
             "description": ""
         },
+        {
+            "name": "stepfunctions_openWorkflowStudio",
```
justinmk3, line 484: Please plan to upstream this to https://github.com/aws/aws-toolkit-common as you near launch.

justinmk3 (approved):

PR #5616, opened 2024-09-17
fix(ec2): remove SSM session state w/ best effort approaches. 

Hweinstock:
## Problem
The remote connection through VSCode does not terminate the SSM session on close consistently.

Notes:

- We do not get an event when the remote window is closed.

## Solution
Leverage two best-effort approaches:
- only allow a single connection from the toolkit to any given EC2 Instance. If a customer attempts to open another remote window in an EC2 instance, we can use that as a sign to terminate the old session. 
- on toolkit shutdown (deactivate), remote any sessions that are still running. 

## Implementation Details
- Implement `Ec2RemoteEnvManager` to manage the remote environments. Behaves like a map of instance ids to sessions ids that most importantly maintains the invariant that any deleted item has its session terminated. 
- Refactor `packages/core/src/awsService/ec2/commands.ts` and  `packages/core/src/awsService/ec2/activation.ts` to allow for state tracking in `EC2ConnectionManager`. This change also gives us an opportunity to improve the testing infrastructure for this code. 
---


packages/core/src/awsService/ec2/explorer/ec2InstanceNode.ts
```diff
@@ -100,3 +101,14 @@ export class Ec2InstanceNode extends AWSTreeNodeBase implements AWSResourceNode
         await vscode.commands.executeCommand('aws.refreshAwsExplorerNode', this)
     }
 }
+
+export async function refreshExplorerNode(node?: Ec2Node) {
```
justinmk3, line 105: I think we have a generalized function somewhere like this already.

packages/core/src/awsService/ec2/commands.ts
```diff
@@ -71,3 +61,16 @@ async function getSelection(node?: Ec2Node, filter?: instanceFilter): Promise<Ec
...
+
+export async function getConnectionManager(
+    connectionManagers: Ec2ConnectionManagerMap,
+    selection: Ec2Selection
+): Promise<Ec2ConnectionManager> {
+    if (connectionManagers.has(selection.region)) {
+        return connectionManagers.get(selection.region)!
+    } else {
+        const newConnectionManager = new Ec2ConnectionManager(selection.region)
+        connectionManagers.set(selection.region, newConnectionManager)
```
justinmk3, line 73: it's a bit subtle that this is actually adding to a global long-lived collection. This logic should live closer to that collection, possibly in a class or something. 

Also need to log a warning if the number of connections exceeds a big number, until we have a better way to have visbility and validation of these long-lived collections (which can caused hidden bugs / perf issues).

packages/core/src/awsService/ec2/remoteSessionManager.ts
```diff
@@ -0,0 +1,41 @@
...
+            this.activeSessions.set(instanceId, sessionId)
+        }
+    }
+
+    private async disconnectEnv(instanceId: EC2.InstanceId): Promise<void> {
+        await this.ssmClient.terminateSessionFromId(this.activeSessions.get(instanceId)!)
+        this.activeSessions.delete(instanceId)
+    }
+
+    public async closeConnections(): Promise<void> {
```
justinmk3, line 34: maybe calling this dispose() is a useful signal (and the class could implement Disposable)
```suggestion
    public async dispose(): Promise<void> {
```

packages/core/src/test/awsService/ec2/commands.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Ec2Selection } from '../../../awsService/ec2/prompter'
+
+describe('getConnectionManager', async function () {
+    let connectionManagers: Ec2ConnectionManagerMap
+
+    beforeEach(function () {
+        connectionManagers = new Map<string, Ec2ConnectionManager>()
+    })
+
+    it('only creates new connection managers once for each region ', async function () {
```
justinmk3, line 19: this makes me wonder if there is be 1 manager for everything. having many managers seems like an unnecessary problem, and partly defeats the purpose of a "manager", which is to have one place where visibility/handling can happen.

packages/core/src/test/awsService/ec2/remoteSessionManager.test.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'assert'
+import * as sinon from 'sinon'
+import { Ec2RemoteSessionManager } from '../../../awsService/ec2/remoteSessionManager'
+import { SsmClient } from '../../../shared/clients/ssmClient'
+
+describe('Ec2RemoteSessionManager', async function () {
```
justinmk3, line 11: is there a session manager for codecatalyst dev envs? if so, it may be worth generalizing it. All of this new test code suggests that it's worth having 1 generalized concept instead of several specific ones.

packages/core/src/awsService/ec2/explorer/ec2InstanceNode.ts
Hweinstock: Yea, this is just a wrapper around that function that makes sure the node exists. Avoids having the same logic in each command where node is optional.

packages/core/src/test/awsService/ec2/commands.test.ts
Hweinstock: I agree, I think the word manager is being overused here. `Manager` also isn't very informative about what each component is responsible for.  I am leaning towards renaming them to directly describe what they "do". 
Ex. `Ec2ConnectionManager` --> `Ec2Connecter`, `Ec2RemoteSessionManager` --> `Ec2SessionTracker`. Does that seem like the right track?

packages/core/src/test/awsService/ec2/remoteSessionManager.test.ts
Hweinstock: For CodeCatalyst, we start the connection within the connect script here https://github.com/aws/aws-toolkit-vscode/blob/0164d4145e58ae036ddf3815455ea12a159d491d/packages/core/resources/codecatalyst_connect#L97 so we don't have the resulting `SessionId` in memory. Also, I am unsure if the same problem of sessions not terminating still happens in CodeCatalyst due to some of these implementation differences.

packages/core/src/awsService/ec2/connectionManagerMap.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Ec2Connecter } from './model'
+
+export class Ec2ConnecterMap extends Map<string, Ec2Connecter> {
+    private static warnSize: number = 25
+
+    public getOrInit(regionCode: string) {
+        return this.has(regionCode) ? this.get(regionCode)! : this.initiateManager(regionCode)
+    }
+
+    private initiateManager(regionCode: string): Ec2Connecter {
```
justinmk3, line 16:
```suggestion
    private initManager(regionCode: string): Ec2Connecter {
```

packages/core/src/awsService/ec2/explorer/ec2InstanceNode.ts
```diff
@@ -101,3 +102,14 @@ export class Ec2InstanceNode extends AWSTreeNodeBase implements AWSResourceNode
         await vscode.commands.executeCommand('aws.refreshAwsExplorerNode', this)
     }
 }
+
+export async function tryRefreshNode(node?: Ec2Node) {
+    if (node) {
+        const n = node instanceof Ec2InstanceNode ? node.parent : node
```
justinmk3, line 108: This could be generalized. I think possibly was, on the https://github.com/aws/aws-toolkit-vscode-staging/tree/feature/lambda-get-started branch.

No action needed here, for now.

packages/core/src/awsService/ec2/remoteSessionManager.ts
```diff
@@ -0,0 +1,40 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { EC2, SSM } from 'aws-sdk'
+import { SsmClient } from '../../shared/clients/ssmClient'
+import { Disposable } from 'vscode'
+
+export class Ec2SessionTracker extends Map<EC2.InstanceId, SSM.SessionId> implements Disposable {
```
justinmk3, line 10: I assume we'll generalize this later, e.g. for ECS sessions and possibly others in the future.

packages/core/src/awsService/ec2/utils.ts
```diff
@@ -3,7 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { SSM } from 'aws-sdk'
```
justinmk3, line 6: this is the "v2" sdk. may want to double-check if we already have the v3 SDK for ssm-related stuff (just by searching the codebase). that will avoid needing to migrate this later.

justinmk3 (approved):

packages/core/src/awsService/ec2/utils.ts
Hweinstock: Unfortunately, `@aws-sdk/client-ssm` is not a dependency yet which I think is what we want based on (https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-ssm/Interface/StartSessionResponse/)

PR #5617, opened 2024-09-17
ci: remove src.gen from workspaces to fix dependabot

hayemaxi:
Problem: I theorize that we are still getting dependabot for src.gen because it is a subworkspace of the root project.

Solution: Remove it and install/build separately on root `npm i`.

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,13 +7,6 @@
 version: 2
 updates:
     - package-ecosystem: 'npm'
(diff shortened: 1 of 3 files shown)


ivikash (approved):

jpinkney-aws (approved):

hayemaxi:
Continued in https://github.com/aws/aws-toolkit-vscode/pull/5667

PR #5625, opened 2024-09-18
telemetry(dev): add tracking for empty path file scenario

tverney:
## Problem

When customer use workspaces to generate files, RTS and LLM cannot decide which folder to deliver the generated file. We're for now tracking how many times we reach for accountability and prioritization.

## Solution

Add telemetry to count and save the conversationIds that reach this specific workspace usage.

PS: This is should only be included in VS Code since the Workspaces capability its only used here.

---

### DIFF
--- a/packages/amazonq/.changes/next-release/Bug Fix-0d324321-bddd-4f93-80c4-80978930f157.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-0d324321-bddd-4f93-80c4-80978930f157.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Amazon Q /dev: include telemetry for workspace usage when generating new files"
+}
--- a/packages/core/src/amazonqFeatureDev/session/sessionState.ts
+++ b/packages/core/src/amazonqFeatureDev/session/sessionState.ts
@@ -61,7 +61,8 @@ function registerNewFiles(
     fs: VirtualFileSystem,
     newFileContents: NewFileZipContents[],
     uploadId: string,
-    workspaceFolders: CurrentWsFolders
+    workspaceFolders: CurrentWsFolders,
+    conversationId: string
 ): NewFileInfo[] {
     const result: NewFileInfo[] = []
     const workspaceFolderPrefixes = getWorkspaceFoldersByPrefixes(workspaceFolders)
@@ -81,6 +82,10 @@ function registerNewFiles(
                       Object.values(workspaceFolderPrefixes).find((val) => val.index === 0)?.name ?? ''
                   ]
         if (folder === undefined) {
+            telemetry.amazonq_wsOrphanedDocuments.emit({
+                amazonqConversationId: conversationId,
+                credentialStartUrl: AuthUtil.instance.startUrl,
+            })
             getLogger().error(`No workspace folder found for file: ${zipFilePath} and prefix: ${prefix}`)
             continue
         }
@@ -174,7 +179,13 @@ abstract class CodeGenBase {
                 case CodeGenerationStatus.COMPLETE: {
                     const { newFileContents, deletedFiles, references } =
                         await this.config.proxyClient.exportResultArchive(this.conversationId)
-                    const newFileInfo = registerNewFiles(fs, newFileContents, this.uploadId, workspaceFolders)
+                    const newFileInfo = registerNewFiles(
+                        fs,
+                        newFileContents,
+                        this.uploadId,
+                        workspaceFolders,
+                        this.conversationId
+                    )
                     telemetry.setNumberOfFilesGenerated(newFileInfo.length)
 
                     return {
@@ -378,7 +389,13 @@ export class MockCodeGenState implements SessionState {
                 zipFilePath: f.zipFilePath,
                 fileContent: f.fileContent,
             }))
-            this.filePaths = registerNewFiles(action.fs, newFileContents, this.uploadId, this.config.workspaceFolders)
+            this.filePaths = registerNewFiles(
+                action.fs,
+                newFileContents,
+                this.uploadId,
+                this.config.workspaceFolders,
+                this.conversationId
+            )
             this.deletedFiles = [
                 {
                     zipFilePath: 'src/this-file-should-be-deleted.ts',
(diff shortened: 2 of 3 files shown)


neilk-aws (approved):

packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -70,6 +71,10 @@ function registerNewFiles(
             workspaceFolderPrefixes === undefined ? '' : zipFilePath.substring(0, zipFilePath.indexOf(path.sep))
         const folder = workspaceFolderPrefixes === undefined ? workspaceFolders[0] : workspaceFolderPrefixes[prefix]
         if (folder === undefined) {
+            telemetry.amazonq_wsOrphanedDocuments.emit({
```
jpinkney-aws, line 74: Can we generalize this metric? I'm not sure if we need a new metric every single time we want to find out if an edge case was triggered

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -939,6 +939,20 @@
                 }
             ]
         },
+        {
+            "name": "amazonq_wsOrphanedDocuments",
```
jpinkney-aws, line 943: In addition to my other comment, can we move this to the common definition in https://github.com/aws/aws-toolkit-common

tverney: https://github.com/aws/aws-toolkit-common/pull/856

tverney: @jpinkney-aws should we merge common first?

packages/core/src/amazonqFeatureDev/session/sessionState.ts
tverney: Fixed and included it in the common package.

packages/amazonq/.changes/next-release/Bug Fix-0d324321-bddd-4f93-80c4-80978930f157.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Amazon Q /dev: include telemetry for workspace usage when generating new files"
```
justinmk3, line 3: This is not something customers are going to care about. Use the `telemetry` PR title instead of `fix`. Or if there is an actual fix to mention in the changelog, please update this changelog.

justinmk3:
> When customer use workspaces to generate files, RTS and LLM cannot decide which folder to deliver the generated file. We're for now tracking how many times we reach for accountability and prioritization.

I don't understand what this means, nor the explanation in https://github.com/aws/aws-toolkit-common/pull/856 .

Can this not be a field added to an existing metric instead of a new metric?

packages/amazonq/.changes/next-release/Bug Fix-0d324321-bddd-4f93-80c4-80978930f157.json
tverney: No fix, just telemetry. I thought the keywords were just limited for fix, feature, so on. I've just updated, thanks for the heads up.

justinmk3: The keywords are documented here: https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#pull-request-title

And in the CI job error message.

tverney:
> > When customer use workspaces to generate files, RTS and LLM cannot decide which folder to deliver the generated file. We're for now tracking how many times we reach for accountability and prioritization.
> 
> I don't understand what this means, nor the explanation in [aws/aws-toolkit-common#856](https://github.com/aws/aws-toolkit-common/pull/856) .
> 
> Can this not be a field added to an existing metric instead of a new metric?

Well, @jpinkney-aws  suggested adding a new generic metric because we currently don't have one for it. In essence, team wanna track a specific scenario that's reached in VS Code to understand the usage of it and proper prioritization.

justinmk3:
> suggested adding a new generic metric because we currently don't have one

Ok. But `amazonq_trackScenarioCountUsage` is rather specific. Why does it mention "count" in the name? If we want to attach other fields (besides the "count") for that scenario, then we would need a new metric... 

why not name it `amazonq_trackScenario` ? But also, every metric tracks a "scenario" already, so adding this metric which has a `scenario` field seems questionable.

tverney:
> > When customer use workspaces to generate files, RTS and LLM cannot decide which folder to deliver the generated file. We're for now tracking how many times we reach for accountability and prioritization.
> 
> I don't understand what this means, nor the explanation in [aws/aws-toolkit-common#856](https://github.com/aws/aws-toolkit-common/pull/856) .
> 
> Can this not be a field added to an existing metric instead of a new metric?

Well, the count its because we want limit the telemetry to just count the number of times we reach a specific condition. It was specific before, however, it was suggested to amplify to something more generic, so now it's talking about a "scenario". We will not be doing more than just counting for the moment. If you like it better, we could call it amazonq_trackConditionCountUsage. I'm ok with either naming conventions its good for you, the goal here its just have metrics on a "specific" condition for us.

justinmk3:
For reference: https://github.com/aws/aws-toolkit-common/pull/856#issuecomment-2362094024

tverney:
> For reference: [aws/aws-toolkit-common#856 (comment)](https://github.com/aws/aws-toolkit-common/pull/856#issuecomment-2362094024)

Updated there, thanks.

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1246,6 +1251,28 @@
                     "required": false
                 }
             ]
+        },
+        {
+            "name": "toolkit_trackScenario",
```
justinmk3, line 1256: this was upstreamed to https://github.com/aws/aws-toolkit-common/pull/856 , so should not be here. if there are any changes, then please upstream them.

tverney: Somehow the rebased version replaced the change adding the previous trackScenario. I'll update again.

justinmk3:
you may need to update the aws/telemetry package, or just merge in current master?

    npm ERR! src/amazonqFeatureDev/session/sessionState.ts(80,23): error TS2339: Property 'toolkit_trackScenario' does not exist on type 'TelemetryTracer'.

tverney:
> you may need to update the aws/telemetry package, or just merge in current master?
> 
> ```
> npm ERR! src/amazonqFeatureDev/session/sessionState.ts(80,23): error TS2339: Property 'toolkit_trackScenario' does not exist on type 'TelemetryTracer'.
> ```

Let me merge/rebase it again

PR #5629, opened 2024-09-18
build: remove package.json syncing

hayemaxi:
Removes the need to copy/sync package JSON contents from core -> subproject extensions on build and package.

Summary of changes:
- Delete syncPackageJson.ts and handlePackageJson.ts
- Consolidate testing scripts and make small improvements (e.g. validation, handle multiple directories to import tests from)
- package.json is now fully complete in `amazonq/`, `toolkit/`. Changes to extensions should be made only the respective file (except for icons).
- `toolkit/` runs unit, e2e, and integ tests from `core/`. It will also run any of its own tests (but those don't exist yet).
- Web tests now moved to and run from `amazonq/`. This was decided because `amazonq/` is more likely to support web than `toolkit/` at this time. Also, it didn't seem possible to run the web tests from `core/` like the other tests because it relies on webpack and a separate test runner.
- No longer create web-based webpacks for `core/`

Follow up work:
- Get [packageJSON properties](https://github.com/aws/aws-toolkit-vscode/blob/46b409c3353ce0dafefb796f5994c1fa88a23b47/packages/core/src/shared/vscode/env.ts#L61-L74) from `context.extension.packageJSON` instead.
- Cleanup any remaining unneeded tasks in `.vscode/tasks.json`
- Move icons out of `core/package.json`.
- Continue to investigate: can we run web tests from core like other tests? Toolkit may have a use in web mode one day.

---


justinmk3 (approved):
As a minor bonus, it looks like this also helps move in the direction of having a single `Tests (current file)` launch config, instead of 4+ configs for the same "current file" functionality.

PR #5632, opened 2024-09-19
fix(auth): wait to display Q sign-in notification

hayemaxi:
(Super low risk and effort fix, but may not address the root issue due to difficulties to repro)

# Problem 
The sign-in notification is queued to display immediately after startup. Some users are reporting that it displays even if they are signed in. This may be due to the tryAutoConnection routine not having a chance to complete because of its async nature.

# Solution
Put a delay on displaying the notification to users to potentially mitigate this issue.

---


packages/amazonq/src/app/chat/activation.ts
```diff
@@ -57,7 +57,10 @@ export async function activate(context: ExtensionContext) {
 
     await amazonq.activateBadge()
     void setupLsp()
-    void setupAuthNotification()
+
+    // Display notification to sign if the user hasn't yet.
+    // Abitrary wait time to give the auth restore routine time to initialize.
+    globals.clock.setTimeout(() => void setupAuthNotification(), 5000)
```
nkomonen-amazon, line 63: AuthUtil emits an onDidChangeActiveConnection event, is it possible we can wait for that to trigger and then we call setupAuthNotification?

hayemaxi: Does it emit if there is no connection? That's when we would display the notification

justinmk3:
Can we make the "sign in notification" the responsibility of `tryAutoConnect`? After all, `tryAutoConnection` "always" runs (or if it doesn't, why?), so it can make the decision to show a "sign in" notification if needed.

packages/amazonq/src/app/chat/activation.ts
nkomonen-amazon: Nvm with my earlier comments, I guess since this notification isn't critical we can prob just keep it simple.

hayemaxi:
> Can we make the "sign in notification" the responsibility of `tryAutoConnect`? After all, `tryAutoConnect` "always" runs (or if it doesn't, why?), so it can make the decision to show a "sign in" notification if needed.

This method is resilient against other (un)expected auth behavior, not just `tryAutoConnect`. I suppose we can make it dependent only on the completion of that function and then re-evaluate if we still see it unexpectedly. This would increase our understanding of the auth flow as well.

packages/amazonq/src/app/chat/activation.ts
hayemaxi: replied to Justin

PR #5633, opened 2024-09-19
feat(ec2): provide link for customer to view system logs

Hweinstock:
## Problem
Limited ability to diagnose problems with EC2 instances provided in the toolkit. 

## Solution
Add a right click option to provide system log of Ec2 instance in read-only file view. 
### Implementation Details
- Log view does NOT live update. This would require refactoring the CWL significantly. Better to get the feature out and see if this is something customers want/need. 
- To reduce code duplication two new utility components were added `decodeBase64` in`core/src/shared/utilities/textUtilities.ts` and `UriSchema` in `core/src/shared/utilities/uriUtils.ts`. 
- `decodeBase64` is a helper function to wrap the use of `Buffer.from(X, "base64").toString()` throughout the toolkit. 
- `UriSchema` provides a general framework for translating between an object and an URI, and vice versa. It allows us to avoid writing the `isValid` method in all cases. With more work this could likely be more general, but is only used twice so unclear if it is worth pursuing. 
- These changes involved changing some function calls, especially in the CWL code, but does not impact functionality. 
### Pictures 
image
image

---


packages/core/src/shared/constants.ts
```diff
@@ -129,6 +129,7 @@ export const ecsIamPermissionsUrl = vscode.Uri.parse(
  */
 export const CLOUDWATCH_LOGS_SCHEME = 'aws-cwl' // eslint-disable-line @typescript-eslint/naming-convention
 export const AWS_SCHEME = 'aws' // eslint-disable-line @typescript-eslint/naming-convention
+export const EC2_LOGS_SCHEME = 'aws-ec2' // eslint-disable-line @typescript-eslint/naming-convention
```
justinmk3, line 132: if this is a new symbol, it should use the naming convention. the old eslint-disable lines were just temporary to avoid churn.
```suggestion
export const ec2LogsScheme = 'aws-ec2'
```

packages/core/src/shared/utilities/uriUtils.ts
```diff
@@ -27,13 +27,22 @@ export function fromQueryToParameters(query: vscode.Uri['query']): Map<string, s
...
- *
- * A convenience function so you do not need to care about the type of path received.
+ * Provide a schema for translating between an object and a vscode.Uri
+ * @param parse function to convert a vscode.Uri into an object, throw error if uri is invalid
+ * @param form function to convert an object into a vscode.Uri
  */
-export function toUri(path: string | vscode.Uri): vscode.Uri {
-    if (path instanceof vscode.Uri) {
-        return path
+export class UriSchema<T> {
```
justinmk3, line 34: interesting idea

packages/core/package.nls.json
```diff
@@ -129,6 +129,7 @@
     "AWS.command.ec2.stopInstance": "Stop EC2 Instance",
     "AWS.command.ec2.rebootInstance": "Reboot EC2 Instance",
     "AWS.command.ec2.copyInstanceId": "Copy Instance Id",
+    "AWS.command.ec2.viewLogs": "View Current EC2 System Logs",
```
justinmk3, line 132: is the word "current" adding anything useful?
```suggestion
    "AWS.command.ec2.viewLogs": "View EC2 Logs",
```

justinmk3:
LGTM after CI passes :)

cc @keeganirby for visibility.

- Related: [this upcoming feature-branch](https://github.com/aws/aws-toolkit-vscode-staging/pull/1811) enhances `packages/core/src/shared/awsConsole.ts` to support any(?) resource/service.

packages/core/src/awsService/ec2/commands.ts
```diff
@@ -71,3 +72,10 @@ async function getSelection(node?: Ec2Node, filter?: instanceFilter): Promise<Ec
 export async function copyInstanceId(instanceId: string): Promise<void> {
     await copyToClipboard(instanceId, 'Id')
 }
+
+export async function openLogDocument(node?: Ec2InstanceNode): Promise<void> {
+    const uri = ec2LogSchema.form(await getSelection(node))
+    const doc = await vscode.workspace.openTextDocument(uri)
+    await vscode.window.showTextDocument(doc, { preview: false })
+    await vscode.languages.setTextDocumentLanguage(doc, 'log')
```
justinmk3, line 80: I wonder if these 3 lines should be a util function in `shared/utilities/textDocumentUtilities.ts`. Then the call would look like:

 ```ts
textDocUtil.show(ec2LogSchema.form(await getSelection(node)))
 ```

justinmk3 (approved):
Nice!

packages/core/package.nls.json
Hweinstock: I believe you can setup a CWL group to include all metrics from the ec2 instances, but it doesn't come pre-setup from what I can tell. So the system logs differ from that. 

I included current because the ec2 logs don't have the "load newer", "load older" codelenses, so its really just a snapshot of the current system logs.

packages/core/src/awsService/ec2/commands.ts
Hweinstock: good catch.

packages/core/package.nls.json
justinmk3: > I included current because the ec2 logs don't have the "load newer", "load older" codelenses, so its really just a snapshot of the current system logs.

That's a good point, but it may be more confusing to use different language in actions. We can always provide detailed messaging when the logs are actually loaded, but actions themselves should use very consistent language.

justinmk3 (approved):

PR #5635, opened 2024-09-19
fix(logging): remove snapping to debug console

tomcat323:
## Problem
The IDE force snaps to Debug Console during local lambda build

## Solution
Create `showOutputChannel` function to focus the window on output channel 

---


packages/core/src/shared/sam/localLambdaRunner.ts
```diff
@@ -536,16 +535,14 @@ export function shouldAppendRelativePathToFuncHandler(runtime: string): boolean
...
- * Brings the Debug Console in focus.
- * If the OutputChannel is showing, focus does not consistently switch over to the debug console, so we're
- * helping make this happen.
+ * Brings the Output Channel in focus.
  */
-async function showDebugConsole(): Promise<void> {
+async function showOutputChannel(ctx: ExtContext): Promise<void> {
     try {
-        await vscode.commands.executeCommand('workbench.debug.action.toggleRepl')
+        ctx.outputChannel.show(true)
```
justinmk3, line 542: Usually we just call this function directly, I don't think it can fail. This is ok for now though.

PR #5636, opened 2024-09-19
fix(ec2): ssh config forces use of supplied keys

Hweinstock:
## Problem
Occasionally this error pops up: 
image

## Solution
Add one line to ssh config that force agent to use the provided key. 
- `IdentitiesOnly yes`: only use identity file provided. 

More info here: https://www.ssh.com/academy/ssh/config

---

### DIFF
--- a/packages/core/src/shared/sshConfig.ts
+++ b/packages/core/src/shared/sshConfig.ts
@@ -170,7 +170,8 @@ export class SshConfig {
     private getBaseSSHConfig(proxyCommand: string): string {
         // "AddKeysToAgent" will automatically add keys used on the server to the local agent. If not set, then `ssh-add`
         // must be done locally. It's mostly a convenience thing; private keys are _not_ shared with the server.
-
+        // "IdentitiesOnly yes" forces agent to only use provided identity file.
+        // More details: https://www.ssh.com/academy/ssh/config
         return `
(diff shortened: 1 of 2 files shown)


packages/core/src/shared/sshConfig.ts
```diff
@@ -178,6 +178,7 @@ Host ${this.configHostName}
     AddKeysToAgent yes
     StrictHostKeyChecking accept-new
     ProxyCommand ${proxyCommand}
+    IdentitiesOnly yes
```
justinmk3, line 181: The insight + URL mentioned in the solution, would be useful agumentation of the comment on line 172. Also worth putting in the commit message.

packages/toolkit/.changes/next-release/Bug Fix-7e333959-58a0-48ff-9368-888735328bf2.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "remote connection will no longer fail with 'too many authentication attempt'"
```
justinmk3, line 3:
```suggestion
	"description": "EC2 connect: remote connection will no longer fail with 'too many authentication attempt'"
```

justinmk3 (approved):

PR #5637, opened 2024-09-19
feat(ec2): switch to using ed25519 to generate key pair

Hweinstock:
## Problem
currently use RSA.

## Solution
switch to ed25519

---

### DIFF
--- a/packages/core/src/test/awsService/ec2/sshKeyPair.test.ts
+++ b/packages/core/src/test/awsService/ec2/sshKeyPair.test.ts
@@ -8,6 +8,7 @@ import * as fs from 'fs-extra'
 import * as sinon from 'sinon'
 import { makeTemporaryToolkitFolder, tryRemoveFolder } from '../../../shared/filesystemUtilities'
 import { SshKeyPair } from '../../../awsService/ec2/sshKeyPair'
+import { ChildProcess } from '../../../shared/utilities/childProcess'
 
 describe('SshKeyUtility', async function () {
     let temporaryDirectory: string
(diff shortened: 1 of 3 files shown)


packages/toolkit/.changes/next-release/Feature-3bebe8f2-b933-4104-8dd6-013d738bd6b5.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "use ed25519 to generate ec2 ssh key pair"
```
justinmk3, line 3: it's usually a good idea to prefix these items with a topic, that makes them easy to read in a list, and also makes it clear which part of the app they refer to. Other notes are [here](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#changelog).

```suggestion
	"description": "EC2: use ed25519 to generate ec2 ssh key pair"
```

packages/core/src/test/awsService/ec2/sshKeyPair.test.ts
```diff
@@ -29,6 +30,13 @@ describe('SshKeyUtility', async function () {
             const contents = await fs.readFile(keyPath, 'utf-8')
             assert.notStrictEqual(contents.length, 0)
         })
+
+        it('uses ed25519 algorithm to generate the keys', async function () {
+            const process = new ChildProcess(`ssh-keygen`, ['-vvv', '-l', '-f', keyPath])
+            const result = await process.run()
+            // Check private key header for algorithm name
+            assert.strictEqual(result.stdout.includes('[ED25519 256]'), true)
```
justinmk3, line 38: Always great to see tests exercising actual reality instead of setting up mocks!

justinmk3 (approved):

justinmk3:
This might imply that we should check the version of `ssh-keygen` and show an error or fallback to RSA if it doesn't support `ed25519`. This file has a bunch of examples using `tryRun` in order to sniff out whether a CLI tool supports what we expect: https://github.com/aws/aws-toolkit-vscode/blob/791ae55021c0ee73dd84abaa57da46237d0fb0d5/packages/core/src/shared/utilities/pathFind.ts#L117-L134

PR #5641, opened 2024-09-20
feat(ui): add refresh button to regionSubmenu

Hweinstock:
## Problem
Originally, wanted refresh for EC2 connect QuickPick so that we can update instance status in the QuickPick. 
This problem can be generalized to all users of the regionSubmenu. 

## Solution
Add refresh button to regionSubmenu. 

### Implementation Details
- Testing the refresh is a little awkward. There is a comment in the `packages/core/src/test/shared/ui/prompters/regionSubmenu.test.ts` file explaining the choices made. 

---


justinmk3 (approved):
Nice!

PR #5642, opened 2024-09-20
feat(ec2): default to ed25519 key type, but fall back on rsa

Hweinstock:
## Problem
(title)

## Solution
Try to use `ed25519` first, if it fails use `rsa`. Alternatively, we could use the same code to check if version number of ssh, and then see if its above version 6.5 (https://www.openssh.com/releasenotes.html#6.5). However, doing an actual attempt at generating the keys felt more robust, as it could potentially fail for other reasons. Make use of the code here: 
https://github.com/aws/aws-toolkit-vscode/blob/791ae55021c0ee73dd84abaa57da46237d0fb0d5/packages/core/src/shared/utilities/pathFind.ts#L117-L134

---

### DIFF
--- a/packages/core/src/awsService/ec2/sshKeyPair.ts
+++ b/packages/core/src/awsService/ec2/sshKeyPair.ts
@@ -9,7 +9,7 @@ import { tryRun } from '../../shared/utilities/pathFind'
 import { Timeout } from '../../shared/utilities/timeoutUtils'
 import { findAsync } from '../../shared/utilities/collectionUtils'
 
-type sshKeyType = 'rsa' | 'ed25519' | 'fake'
+type sshKeyType = 'rsa' | 'ed25519'
 
 export class SshKeyPair {
     private publicKeyPath: string
--- a/packages/core/src/awsService/ec2/sshKeyPair.ts
+++ b/packages/core/src/awsService/ec2/sshKeyPair.ts
@@ -5,6 +5,9 @@
 import * as fs from 'fs-extra'
 import { ToolkitError } from '../../shared/errors'
 import { ChildProcess } from '../../shared/utilities/childProcess'
+import { tryRun } from '../../shared/utilities/pathFind'
+
+type sshKeyType = 'ed25519' | 'rsa'
 
 export class SshKeyPair {
     private publicKeyPath: string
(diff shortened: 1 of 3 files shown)


packages/core/src/test/awsService/ec2/sshKeyPair.test.ts
```diff
@@ -38,25 +38,36 @@ describe('SshKeyUtility', async function () {
...
-        })
-
-        it('uses ed25519 algorithm to generate the keys', async function () {
-            const process = new ChildProcess(`ssh-keygen`, ['-vvv', '-l', '-f', keyPath])
-            const result = await process.run()
-            // Check private key header for algorithm name
-            assert.strictEqual(result.stdout.includes('[ED25519 256]'), true)
-        })
+    it('generates key in target file', async function () {
+        const contents = await vscode.workspace.fs.readFile(vscode.Uri.file(keyPath))
```
justinmk3, line 42: nit: could use our "fs" module since it's already being imported to this file. The more places we use preferred things, the more likely that new contributions will copy "best practices" :)

packages/core/src/awsService/ec2/sshKeyPair.ts
```diff
@@ -35,13 +38,25 @@ export class SshKeyPair {
     }
 
     public static async generateSshKeyPair(keyPath: string): Promise<void> {
```
justinmk3, line 40: Do we have a sanity check somewhere that ensures we don't overwrite a path in the user's ~/.ssh ? I assume we are writing to the extension's own globalStorage, but maybe wherever we do the "delete timer" we should double-check.

justinmk3 (approved):
Nice choices!

packages/core/src/test/awsService/ec2/sshKeyPair.test.ts
Hweinstock: If the function is supported by our "fs" is there any reason not to use it? I could make another PR replacing some the other uses?

packages/core/src/awsService/ec2/sshKeyPair.ts
Hweinstock: So far `SshKeyPair` is only used with globalStorage, but nothing is stopping it from being used elsewhere. To avoid messing with user's `~/.ssh` (or other sensitive) we could have a list of blacklisted paths, and throw an error if the `SshKeyPair` is called with that any of those paths as a argument. Alternatively, we could force the path to be a subpath of globalStorage. I can track that and make it a follow up.

packages/core/src/test/awsService/ec2/sshKeyPair.test.ts
justinmk3: > If the function is supported by our "fs" is there any reason not to use it? 

No, just old outdated code.

> I could make another PR replacing some the other uses?

That would be fantastic!

justinmk3 (approved):

PR #5646, opened 2024-09-23
refactor(telemetry): Remove telemetry.foo.record

jpinkney-aws:
## Problem
- Theres quite a few missues of telemetry.foo.record inside of the codebase and it can be confusing for external contributors

## Solution
- Clean up the telemetry interface by only allowing record/increment on spans

---

### DIFF
--- a/packages/core/src/auth/secondaryAuth.ts
+++ b/packages/core/src/auth/secondaryAuth.ts
@@ -12,7 +12,7 @@ import { isNonNullable } from '../shared/utilities/tsUtils'
 import { ToolIdStateKey } from '../shared/globalState'
 import { Connection, getTelemetryMetadataForConn, SsoConnection, StatefulConnection } from './connection'
 import { indent } from '../shared/utilities/textUtilities'
-import { AuthStatus, telemetry } from '../shared/telemetry/telemetry'
+import { AuthModifyConnection, AuthStatus, Span, telemetry } from '../shared/telemetry/telemetry'
 import { asStringifiedStack } from '../shared/telemetry/spans'
 import { withTelemetryContext } from '../shared/telemetry/util'
 import { isNetworkError } from '../shared/errors'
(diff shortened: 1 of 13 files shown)


nkomonen-amazon (approved):

packages/core/src/shared/telemetry/telemetry.ts
```diff
@@ -23,7 +23,10 @@ export const telemetry = new TelemetryTracer()
  */
 declare module './telemetry.gen' {
     interface Metric<T extends MetricBase = MetricBase> {
+        run<U>(fn: (span: Span<T>) => U, options?: SpanOptions): U
+    }
+
+    interface Span<T extends MetricBase = MetricBase> {
         increment(data: { [P in NumericKeys<T>]+?: number }): void
-        run<U>(fn: (span: this) => U, options?: SpanOptions): U
     }
```
justinmk3, line 31: Why doesn't span have run()? I don't quite understand the distinction here. I think of metrics as mostly synonymous with spans.

jpinkney-aws: The problem is [this span as any is a lie](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/telemetry/spans.ts#L367). It never actually had run. When you try and do `span.run` you get "span.run is not a function"

The passed in span is just a telemetry span that has emit, record, increment but isn't a metric type

packages/core/src/auth/secondaryAuth.ts
```diff
@@ -349,7 +349,7 @@ export class SecondaryAuth<T extends Connection = Connection> {
             let connectionState = this.auth.getConnectionState(conn)
 
             // This function is expected to be called in the context of restoreConnection()
-            telemetry.auth_modifyConnection.record({
+            span.record({
```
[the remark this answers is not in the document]
jpinkney-aws: Technically yes, but it also has the side effect that every subsequent span thats created would have connectionState and authStatus

Instead I could have technically used `telemetry.activeSpan?.record()`

ivikash (approved):

PR #5647, opened 2024-09-23
feat(filesystem): add stub for chmod to filesystem wrapper

Hweinstock:
## Problem
Changing the permissions of a file is currently not supported by the file system wrapper. 

## Solution
Implement a stub chmod that does nothing on web, and uses `fs` to modify permissions in non-web environments. 

---


packages/core/src/test/awsService/ec2/sshKeyPair.test.ts
```diff
@@ -57,6 +59,13 @@ describe('SshKeyUtility', async function () {
             // Check private key header for algorithm name
             assert.strictEqual(result.stdout.includes('[ED25519 256]'), true)
         })
+
+        it('sets permission of the file to read/write owner', async function () {
+            if (!globals.isWeb && os.platform() !== 'win32') {
+                const result = await stat(keyPair.getPrivateKeyPath())
```
justinmk3, line 65: Oh I guess we will need `fs.stat()`, later...

justinmk3 (approved):
Thanks! Are there any existing usages we can replace too?

packages/core/src/test/awsService/ec2/sshKeyPair.test.ts
```diff
@@ -50,6 +52,13 @@ describe('SshKeyUtility', async function () {
         assert.notStrictEqual(beforeContent, afterContent)
     })
 
+    it('sets permission of the file to read/write owner', async function () {
+        if (!globals.isWeb && os.platform() !== 'win32') {
+            const result = await stat(keyPair.getPrivateKeyPath())
```
justinmk3, line 57:
```suggestion
            const result = await nodefs.stat(keyPair.getPrivateKeyPath())
```

justinmk3 (approved):
LGTM let's merge it!

PR #5648, opened 2024-09-23
refactor(ec2): avoid unnecessary state in sshKeyPair

Hweinstock:
## Problem
Follow up: https://github.com/aws/aws-toolkit-vscode/pull/5578#discussion_r1769231182

## Solution
`isDeleted` status is determined by looking at the actual file system, rather than maintaining additional state

---


hayemaxi:
Why is it valid for only 1 of the keys to remain? Does that mean its in a bad state?

Hweinstock:
Yeah, ideally we would block that state from ever happening, but since its on the file system anything is possible theoretically. I wasn't sure which way made more sense, but thinking about `isDeleted` as not valid makes me think I should swap it.

justinmk3 (approved):

PR #5654, opened 2024-09-24
feat(telemetry): Include user intent for copy and insert code interactions

lalexcaws:
## Problem
Need to gather acceptance metrics on generated unit tests

## Solution
Include user intent on copy and insert code interactions in telemetry to form an acceptance rate

---


packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -821,6 +821,10 @@
                 {
                     "type": "cwsprChatMessageId"
                 },
+                {
+                    "type": "cwsprChatUserIntent",
+                    "required": false
+                },
```
hayemaxi, line 827: Do you have a plan/backlog item to port this to aws-toolkit-common?

lalexcaws: No, I missed the upstreaming part. Should I create an issue for tracking this?

justinmk3: Yes, that is required for all metrics (with rare exceptions). All of the amazon q metrics in this file need to be upstreamed (and aligned with other IDEs), assuming they are still in use.

hayemaxi: Yes, after merging this you should treat it as high priority to move this metric to commons, especially if other IDEs need to use it.

lalexcaws: Got it, I will take that as an action item and work on it

laileni-aws (approved):

packages/core/src/shared/telemetry/vscodeTelemetry.json
lalexcaws: Update for aws-toolkit-common: https://github.com/aws/aws-toolkit-common/pull/864

PR #5655, opened 2024-09-24
telemetry(amazonq): Add telemetry tracing for amazon q inline

jpinkney-aws:
## Problem
- The flow from when a message is shown to the user and rejected/accepted by the user isn't instrumented with any way to trace events 

## Solution
- Create a trace from when the event is shown (either by manual invocation or auto trigger) and then connect all the telemetry events through a traceId.

### Implementation details
- The traceId is initially created when manual invocation/auto trigger occurs
- For every telemetry event thats in the function hierarchy of the manual invocation/auto trigger, they automatically assume the same traceId
- When an event listener is called it uses the traceId we've set in TelemetryHelper to setup the next traceId. This is because the event no longer belongs to the same context
- Inside of commands we use `telemetry.record` because it allows us to override the default telemetry id that gets automatically created and instead associate it with the known trace

---

### DIFF
--- a/packages/core/src/codewhisperer/commands/invokeRecommendation.ts
+++ b/packages/core/src/codewhisperer/commands/invokeRecommendation.ts
@@ -11,6 +11,9 @@ import { isCloud9 } from '../../shared/extensionUtilities'
 import { RecommendationHandler } from '../service/recommendationHandler'
 import { session } from '../util/codeWhispererSession'
 import { RecommendationService } from '../service/recommendationService'
+import { telemetry } from '../../shared/telemetry'
+import { TelemetryHelper } from '../util/telemetryHelper'
+import { randomUUID } from '../../shared/crypto'
 
 /**
  * This function is for manual trigger CodeWhisperer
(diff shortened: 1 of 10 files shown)


packages/core/src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -46,6 +46,8 @@ export class TelemetryHelper {
     private timeToFirstRecommendation = 0
     private classifierResult?: number = undefined
     private classifierThreshold?: number = undefined
+    // variables for tracking end to end sessions
+    public traceId: string = ''
```
hayemaxi, line 50:
```suggestion
    public traceId: string = 'notSet'
```

packages/core/src/codewhisperer/commands/invokeRecommendation.ts
```diff
@@ -52,5 +55,9 @@ export async function invokeRecommendation(
         )
     }
 
-    await RecommendationService.instance.generateRecommendation(client, editor, 'OnDemand', config, undefined)
+    const traceId = randomUUID()
+    TelemetryHelper.instance.setTraceId(traceId)
```
hayemaxi, line 59: Do race conditions matter?
Can you have code that is running and should be assigned to trace id #1, but auto/manual trigger occurs and overwrites the current traceId in a new context while other code is still executing?

jpinkney-aws: I don't think so, the current codewhisperer https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/util/telemetryHelper.ts does the same thing. It stores the states until for a full invocation session with only one session active at a time

hayemaxi (approved):

PR #5657, opened 2024-09-24
feat(utils): Add some util functions for upcoming feature

nkomonen-amazon:
This adds some util functions that are needed for an upcoming feature.

Each commit is its own isolated util, so review them individually.

**A significant amount of the changes come from minor refactors in files**

---


packages/core/src/test/shared/fs/fs.test.ts
```diff
@@ -7,9 +7,8 @@ import assert from 'assert'
 import vscode from 'vscode'
 import * as path from 'path'
 import * as utils from 'util'
-import { existsSync, mkdirSync, promises as nodefs, readFileSync, rmSync } from 'fs'
```
nkomonen-amazon, line 10: Renaming/refactoring, no behavioral changes

packages/core/src/test/testUtil.ts
```diff
@@ -55,6 +64,59 @@ export function getWorkspaceFolder(dir: string): vscode.WorkspaceFolder {
...
+ * An all-in-one solution for a folder to use during tests:
+ * - Cleans up after all tests complete
+ * - Each instance is isolated from other tests
+ * - Convenient methods for files operations on paths RELATIVE to the folder
+ *
+ * Create an instance using {@link TestFolder.create()}.
+ *
+ * Add any new methods to this class when needed.
+ */
+export class TestFolder {
```
nkomonen-amazon, line 76: All new test files that need a temporary folder for testing should use this class. We should transition away from the existing methods over time, as indicated by the `deprecated` indicator in their docstrings.

packages/core/src/shared/fs/node/fs.ts
```diff
@@ -0,0 +1,67 @@
...
+        let release = undefined
+        try {
+            const path = this.fsCommon.toUri(uri).fsPath
+            release = await lockFile.lock(path, this.lockOptions)
+            return await callback()
+        } catch (err) {
+            if (!(err instanceof Error)) {
+                throw err
+            }
+            throw ToolkitError.chain(err, `Failed to acquire lock on: ${uri}`, { code: getErrorId(err) })
```
nkomonen-amazon, line 46: TODO: Update this since we can successfully acquire the lock but the callback may fail. See if we can use a nested try/catch for the callback

packages/core/src/test/testUtil.ts
```diff
@@ -16,10 +16,19 @@ import { MetricName, MetricShapes } from '../shared/telemetry/telemetry'
...
 import fs2 from '../shared/fs/fs'
 import { DeclaredCommand } from '../shared/vscode/commands2'
+import { mkdirSync, existsSync } from 'fs'
+import * as nodefs from 'fs/promises'
+import { randomBytes } from 'crypto'
 
 const testTempDirs: string[] = []
 
 /**
+ * @deprecated Use {@link TestFolder} instead.
```
nkomonen-amazon, line 26: TODO: remove the deprecation, there is still a use for this in tests

packages/amazonq/.vscode/launch.json
```diff
@@ -18,7 +18,8 @@
             "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
             "env": {
                 "SSMDOCUMENT_LANGUAGESERVER_PORT": "6010",
-                "WEBPACK_DEVELOPER_SERVER": "http://localhost:8080"
+                "WEBPACK_DEVELOPER_SERVER": "http://localhost:8080",
+                "AWS_VSCODE_DEBUG_MODE": "true"
```
jpinkney-aws, line 22: Alternatively can't we just check that the WEBPACK_DEVELOPER_SERVER is defined to tell its in dev mode? Also is there a launch config for the toolkit as well that makes these changes?

packages/core/src/test/testUtil.ts
jpinkney-aws: It looks like you're using nodeFs in the write function. Should we add a declaimer that parts of it only work in node?

packages/core/src/shared/utilities/searchUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+/**
+ * Performs a binary search on a sorted array of numbers to find the index of a target value.
+ *
+ * @param items The sorted array of numbers to search.
+ * @param target The target value to search for.
+ * @returns The index of the target value in the array, or -1 if not found.
+ */
+export function binarySearch(items: number[], target: number): number {
```
jpinkney-aws, line 13: why do we need binary search ?

jpinkney-aws:
Looks like theres a conflict btw

packages/amazonq/.vscode/launch.json
nkomonen-amazon: That makes sense for sure. Will update.

Yeah there is one already, if you see the individual commit it is easier to review

packages/core/src/shared/utilities/searchUtils.ts
nkomonen-amazon: In the main Crash Reporting PR we get a list of PIDs which can be in the thousands and then search it to see if the ExtHost process is missing (implies a crash). And we do this for each running extension instance. So I thought we could be a bit more efficient about it.

packages/core/src/test/testUtil.ts
nkomonen-amazon: Right now this class is used in our agnostic FileSystem unit tests to do the setup. So this is why we dont use our agnostic FS in here YET, so we don't use our code to test itself. But if the day comes that we need it for web, we should be able to add some agnostic FS methods in here.

Adding this information to docstring

packages/core/src/test/shared/utilities/osUtils.test.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+import assert from 'assert'
+import { isNewOsSession } from '../../../shared/utilities/osUtils'
+import { fs } from '../../../shared'
+import { isWin } from '../../../shared/vscode/env'
+import { TestFolder } from '../../testUtil'
+
+describe('isNewOsSession', () => {
+    if (!isWin()) {
```
jpinkney-aws, line 13: I assume this isn't tested because we don't have a windows computer to test on lol

packages/core/src/shared/fs/fs.ts
```diff
@@ -640,7 +640,7 @@ export class FileSystem {
      * @param path The file path for which to retrieve metadata.
      * @return The Uri about the file.
      */
-    #toUri(path: string | vscode.Uri): vscode.Uri {
+    toUri(path: string | vscode.Uri): vscode.Uri {
```
justinmk3, line 643: this seems like a strange thing to expose as a public api on the "fs" interface. it's a minor convenience for the code local to this module. 

we have a `packages/core/src/shared/utilities/uriUtils.ts` module, should it live there (though still kind of questionable) ?

packages/core/src/shared/fs/node/fs.ts
```diff
@@ -0,0 +1,77 @@
...
+    protected constructor() {}
+    static #instance: NodeFileSystem
+    static get instance(): NodeFileSystem {
+        return (this.#instance ??= new NodeFileSystem())
+    }
+
+    /**
+     * Acquires a lock on the given file, then runs the given callback, releasing the lock when done.
+     *
+     * The reason this method is node specific is that we use the `mkdir()` method for the lock file,
```
justinmk3, line 28: is this saying that lockFile.lock() is using mkdir internally? mkdir is not used in this file directly

packages/core/src/shared/fs/node/fs.ts
```diff
@@ -0,0 +1,77 @@
...
+    private fsCommon: FileSystem = fs
+
+    protected constructor() {}
+    static #instance: NodeFileSystem
+    static get instance(): NodeFileSystem {
+        return (this.#instance ??= new NodeFileSystem())
+    }
+
+    /**
+     * Acquires a lock on the given file, then runs the given callback, releasing the lock when done.
```
justinmk3, line 26: is this being used to coordinat with proceses outside of vscode? 

if we are synchronizing between vscode instances, can we use globalState to coordinate instead?

packages/core/src/shared/utilities/osUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+ * Use this function to perform one-time initialization tasks that should only happen
+ * once per OS session, regardless of how many extension instances are running.
+ */
+export async function isNewOsSession(tmpDir = tempDirPath) {
+    switch (process.platform) {
+        case 'win32': {
+            // Windows does not have an ephemeral /tmp/ folder that deletes on shutdown, while unix-like os's do.
+            // So in Windows we calculate the start time and see if it changed from the previous known start time.
+
+            const lastStartTime = globals.globalState.get<number>('lastOsStartTime', undefined)
```
justinmk3, line 27: tryGet gives some extra type-checking protection (this is mentioned in the get() docstring)
```suggestion
            const lastStartTime = globals.globalState.tryGet('lastOsStartTime', Number)
```

packages/core/src/shared/fs/node/fs.ts
nkomonen-amazon: Will update, by "we" I meant the `proper-lockfile` module uses it internally. Does this work?

packages/core/src/shared/fs/fs.ts
nkomonen-amazon: uriUtils makes sense for this to live in, I'll move it there

packages/core/src/shared/fs/node/fs.ts
nkomonen-amazon: Due to us not having access to VSCode apis when the extension is being shutdown in `deactivate()`, we would not be able to update global state when needed. But since we can use the nodeFs instead, that is the next best option.

packages/core/src/shared/utilities/osUtils.ts
nkomonen-amazon: ty!

jpinkney-aws (approved):

packages/core/src/shared/fs/node/fs.ts
justinmk3: Does that mean we could potentially lock for X seconds on deactivate...?

packages/core/src/shared/utilities/osUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+            // actually comparing it.
+            if (currentOsStartTime - 2000 > lastStartTime) {
+                await globals.globalState.update('lastOsStartTime', currentOsStartTime)
+                return true
+            }
+
+            return false
+        }
+        default: {
+            // Store a file in /tmp/ that acts as a flag. The OS automatically deletes on system shutdown.
```
justinmk3, line 47: Not sure if it's important (though based on the Windows comment above, it sounds like yes), but in practice /tmp is often not cleared out after many reboots.

Is there a particular reason why we are not using globalStorageUri here?

packages/core/src/test/shared/fs/fs.test.ts
```diff
@@ -21,29 +20,27 @@ import * as testutil from '../../testUtil'
...
 
     before(async function () {
-        fakeContext = await FakeExtensionContext.create()
         sandbox = Sinon.createSandbox()
-        await deleteTestRoot() // incase a previous test run failed to clean
     })
 
     beforeEach(async function () {
-        await mkTestRoot()
+        testFolder = await TestFolder.create()
```
justinmk3, line 34: This looks like a useful change, thanks.

packages/core/src/shared/utilities/osUtils.ts
nkomonen-amazon: The idea was that /tmp cleared itself on reboot, and globalStorageUri does not. So I guess that since None of these are reliable then the Windows method should be used for all of them.

PR #5661, opened 2024-09-24
test(ec2): fix flaky windows test. 

Hweinstock:
## Problem
The test case here:  https://github.com/aws/aws-toolkit-vscode/blob/359b1d8252a32e064482678110c5931630bd49e0/packages/core/src/test/awsService/ec2/model.test.ts#L132C9-L146C11

This test is unaware of the side-effect that it will attempt to explicitly delete the keys when generating. Originally, the delete was conditional on the keys existing in the directory, so this test case was not a problem since they never existed. However, on windows that check was faulty, leading to the case where the windows file system reported that the keys existed, which lead it to attempting to delete the root directory. 

## Solution
- Modify `tryRun` to accept an `onStdOut` optional parameter that lets us accept the overwrite prompt from `ssh-keygen`. 
- Do not explicitly delete the files in the generation function, make this part of the generation step. 

The benefits of this approach are that we moved the side-effects to a place where they are expected, in the key generation process itself. It also inherently avoids the potentially faulty check if the key files exist, as we only use `onStdOut` if when the keys exist. 

### Notes
- `packages/core/src/awsService/ec2/sshKeyPair.ts:49` includes `any` which should be `ChildProcess`. For some reason the import fails in CI, so I left it as `any`. Refer to `attempt to import child process` commit for logs. 

---


packages/core/src/shared/utilities/pathFind.ts
```diff
@@ -30,10 +30,12 @@ export async function tryRun(
     p: string,
     args: string[],
     logging: 'yes' | 'no' | 'noresult' = 'yes',
-    expected?: string
+    expected?: string,
+    onStdout?: (proc: ChildProcess, text: string) => void
```
justinmk3, line 34: probably should just accept a `opt: ChildProcessOptions`

Hweinstock: Is there a way to do this without modifying the `ChildProcessOptions` interface? I need the `ChildProcess` object itself in-scope for the `onStout` definition, and couldn't figure out how to do this with the current setup in `src/shared/utilities/childProcess.ts`.

justinmk3: oh i see. ok, can revisit this later if needed.

justinmk3: a low-churn way to do this would be to add a `send()` method to `RunParameterContext`, similar to the existing `stop()` method: https://github.com/aws/aws-toolkit-vscode/blob/feb62bb953e62534df5e32731b8c989e9dfb9f4c/packages/core/src/shared/utilities/processUtils.ts#L12-L16

then I think just add `send: this.send.bind(this)` here  : https://github.com/aws/aws-toolkit-vscode/blob/feb62bb953e62534df5e32731b8c989e9dfb9f4c/packages/core/src/shared/utilities/processUtils.ts#L160-L163

justinmk3 (approved):

packages/core/src/awsService/ec2/sshKeyPair.ts
```diff
@@ -50,7 +47,12 @@ export class SshKeyPair {
      * @param keyType type of key to generate.
      */
     public static async tryKeyGen(keyPath: string, keyType: sshKeyType): Promise<boolean> {
-        return !(await tryRun('ssh-keygen', ['-t', keyType, '-N', '', '-q', '-f', keyPath], 'yes', 'unknown key type'))
+        const overrideKeys = async (_t: string, context: RunParameterContext) => {
+            await context.send('yes')
```
justinmk3, line 51:
```suggestion
        const overrideKeys = async (_t: string, proc: RunParameterContext) => {
            await proc.send('yes')
```

justinmk3 (approved):

PR #5666, opened 2024-09-25
fix(amazonq): Remove automatic focus on prompt input field after response

jpinkney-aws:
## Problem
When a request completes in amazon q the chat is automatically focused, pulling focus away from anything a user is doing in the editor

## Solution
Disable automatic prompt focus

Mynah UI docs: https://github.com/aws/mynah-ui/blob/b7f9807b4aa984dda9200286a9d0d9e083dd39fa/docs/CONFIG.md#autofocus

---


jpinkney-aws:
Found out this has the unintended consequence that if a user is focused/doing multiple back and forths in chat the prompt input field doesn’t get automatically focused after a response comes back. Might need to update mynah ui itself

justinmk3:
Is this still needed?

jpinkney-aws:
Closing as https://github.com/aws/mynah-ui/pull/144 ended up being the better solution

PR #5667, opened 2024-09-25
build: detach src.gen/ from monorepo

hayemaxi:
Theory: dependabot is still trying to update src.gen/ dependencies because they are managed by the root monorepo. So technically, there is no way to avoid updating src.gen's package.json because the root package.json is linked to it and impacted by dependabot changes for subproject dependencies.

Let's make its own "subrepo" and see if we can get dependabot to ignore it finally.

---


justinmk3 (approved):
This is worth a try as a data point. But the details are complex enough that, even if it works, it might be justified to move src.gen to a separate repo (or branch) which produces a .js result that we then commit directly to this repo. At this point we have almost no choice.

scripts/package.ts
```diff
@@ -171,10 +200,40 @@ function main() {
...
         fs.writeFileSync(packageJsonFile, JSON.stringify(packageJson, undefined, '    '))
-        child_process.execFileSync('vsce', ['package', '--ignoreFile', '../.vscodeignore.packages'], {
-            stdio: 'inherit',
-            shell: process.platform === 'win32', // For vsce.cmd on Windows.
-        })
+
+        // vsce runs this command by default.
+        // This can result in errors that originate from configuration in the src.gen/ directory, but that directory
+        // cannot be managed directly since it is autogenerated code.
+        // This section runs the validation command manually and ignores errors only related to src.gen/
```
justinmk3, line 207: Is this logic intended to actually tell us if dependabot is trying to recurse into src.gen? Or is it still expected that it will?

hayemaxi: Its unrelated to dependabot. We are basically just replacing an internal validation step of `vsce package` with our own that will not throw an error on issues with `src.gen/`. Without this, splitting off `src.gen` from the root repo results in a dependency error during packaging based on something in `src.gen/node_modules`. The extension appears to work normally with this dependency error:
```
npm ERR! invalid: typescript@5.7.0-dev.20240926 /Volumes/workplace/aws-toolkit-vscode/src.gen/@amzn/codewhisperer-streaming/node_modules/downlevel-dts/node_modules/typescript
```

After this change I expect it to continue to recurse into `src.gen/`, but the new ignore rules in `dependabot.yml` should auto-close PRs that modify it.

justinmk3: > Without this, splitting off `src.gen` from the root repo results in a dependency error during packaging based on something in `src.gen/node_modules`.

That one sentence is most helpful for me, perhaps should be (first) in the code comment.

Does this add fragility to the build? Should we declare bankrupty and [fully separate](https://github.com/aws/aws-toolkit-vscode/pull/5667#pullrequestreview-2332061718) src.gen (whether by branch, subdir, or npm package)?

hayemaxi: > Does this add fragility to the build? Should we declare bankrupty and [fully separate](https://github.com/aws/aws-toolkit-vscode/pull/5667#pullrequestreview-2332061718) src.gen (whether by branch, subdir, or npm package)?

Obviously it adds some complexity but I wouldn't say its fragile. I don't think its worth the effort to separate src.gen/ if this works unless it causes problems later on.

Update: My initial interpretation of https://github.com/microsoft/vscode-vsce/issues/439 was incorrect. It seems that we don't need vsce's dependency checks at all since we are bundling the extension with webpack. So `--no-dependency` should be on by default for us and we don't need this custom checking logic that emulates the dependency checks.

justinmk3 (approved):
Nice!

packages/toolkit/package.json
```diff
@@ -64,7 +64,7 @@
         "webCompile": "npm run clean && npm run buildScripts && webpack --config-name web",
         "webRun": "npx @vscode/test-web --open-devtools --browserOption=--disable-web-security --waitForDebugger=9222 --extensionDevelopmentPath=. .",
         "package": "npm run copyPackageJson && ts-node ../../scripts/package.ts",
-        "install-plugin": "vsce package --ignoreFile '../.vscodeignore.packages' -o aws-toolkit-vscode-test.vsix && code --install-extension aws-toolkit-vscode-test.vsix",
+        "install-plugin": "vsce package --no-dependencies --ignoreFile '../.vscodeignore.packages' -o aws-toolkit-vscode-test.vsix && code --install-extension aws-toolkit-vscode-test.vsix",
```
justinmk3, line 67: (not a blocker) we should probably remove this task.

PR #5670, opened 2024-09-25
test(amazonq): add performance test for prepareRepoData

Hweinstock:
## Problem
Start of reliability work to enhance performance testing of high risk code paths. 

## Solution
`prepareRepoData` in `src/amazonqFeatureDev/util/files.ts` captures many potentially high risk paths in a single function, such as `collectFiles` and other I/O operations, making it a great place to start performance testing. 

Two performance tests were added, one for handling a workspace with many small files, and one with few large files. Unsurprisingly, the performance is significantly better on fewer larger files as it reduces # of I/O operations. Thresholds are very lenient to avoid flaky tests. 

The "many files" test does 250 files w/ 10 bytes each. The "large files" does 10 files w/ 1000 bytes each. These values were chosen to avoid flakiness in CI. 

---


packages/core/src/test/amazonqFeatureDev/prepareRepoData.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        },
+
+        'handles large files',
+        function () {
+            const telemetry = new TelemetryHelper()
+            let result: resultType
+            let workspace: WorkspaceFolder
+            return {
+                setup: async () => {
+                    workspace = await createTestWorkspace(10, {
```
jpinkney-aws, line 99: You can also return workspace from here and then use then use workspace as an arg to execute instead of having to declare things at the top level

Hweinstock: Yes, I like that better! I wasn't sure if this was only for stubs/spys and the like. After looking deeper at the implementation I realized I could do the same for result too.

packages/core/src/test/amazonqFeatureDev/prepareRepoData.test.ts
```diff
@@ -0,0 +1,110 @@
...
+describe('prepareRepoDataPerformanceTest', function () {
+    function verifyResult(result: resultType, telemetry: TelemetryHelper, expectedSize: number): void {
+        assert.ok(result)
+        assert.strictEqual(Buffer.isBuffer(result.zipFileBuffer), true)
+        assert.strictEqual(telemetry.repositorySize, expectedSize)
+        assert.strictEqual(result.zipFileChecksum.length, 44)
+    }
+    /**
+     * Tests 250 files w/ 10 bytes each.
+     * Running more files can lead to flaky test from timeout.
```
jpinkney-aws, line 26: would there be any value on seeing how this number scales? e.g. for every 50 files it roughly increases the duration by 2 or is it too unpredictable?

packages/core/src/test/amazonqFeatureDev/prepareRepoData.test.ts
```diff
@@ -0,0 +1,110 @@
...
+import { createTestWorkspace } from '../testUtil'
+import { prepareRepoData, TelemetryHelper } from '../../amazonqFeatureDev'
+import { AmazonqCreateUpload, getRandomString, Metric } from '../../shared'
+
+type resultType = {
+    zipFileBuffer: Buffer
+    zipFileChecksum: string
+}
+
+describe('prepareRepoDataPerformanceTest', function () {
```
jpinkney-aws, line 17: I'd probably just have a top level describe thats prepareRepoData and then a second describe that has performance test in it, that way it can easily seperate perf tests from non perf tests if anyone wants to add it

Hweinstock: There are also some existing correctness tests in `test/unit/amazonqFeatureDev/util/files.test.ts`. I kept the performance tests in core since thats where the `startSecurityScan` is (src/test/codewhisperer/commands/startSecurityScan.test.ts).

packages/core/src/test/amazonqFeatureDev/prepareRepoData.test.ts
Hweinstock: duration in general was very flaky on this one, and also it seemed to not scale linearly. I would be surprised based on what I've seen so far if an increase of 50 files corresponded so some constant duration increase. I could look into it with less files.

jpinkney-aws (approved):

PR #5672, opened 2024-09-25
fix(dev): When using workspace, define first folder if LLM created file doesn't have a defined root path

tverney:
## Problem

In VS Code, when user add 2 different folders in the workspace which contain different root paths, RTS/LLM cannot decide where to place a generated file.

recording

## Solution

This fix add the default first folder selected as its default for orphaned root created files, not blocking user to receive an empty patch but also support better the workspace usage.

---

### DIFF
--- a/packages/core/src/amazonqFeatureDev/session/sessionState.ts
+++ b/packages/core/src/amazonqFeatureDev/session/sessionState.ts
@@ -68,7 +68,13 @@ function registerNewFiles(
         fs.registerProvider(uri, new VirtualMemoryFile(contents))
         const prefix =
             workspaceFolderPrefixes === undefined ? '' : zipFilePath.substring(0, zipFilePath.indexOf(path.sep))
-        const folder = workspaceFolderPrefixes === undefined ? workspaceFolders[0] : workspaceFolderPrefixes[prefix]
+        const folder =
+            workspaceFolderPrefixes === undefined
+                ? workspaceFolders[0]
+                : workspaceFolderPrefixes[prefix] ??
+                  workspaceFolderPrefixes[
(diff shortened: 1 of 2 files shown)


packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -68,7 +68,13 @@ function registerNewFiles(
         fs.registerProvider(uri, new VirtualMemoryFile(contents))
         const prefix =
             workspaceFolderPrefixes === undefined ? '' : zipFilePath.substring(0, zipFilePath.indexOf(path.sep))
-        const folder = workspaceFolderPrefixes === undefined ? workspaceFolders[0] : workspaceFolderPrefixes[prefix]
+        const folder =
+            workspaceFolderPrefixes === undefined
+                ? workspaceFolders[0]
+                : workspaceFolderPrefixes[prefix] ??
+                  workspaceFolderPrefixes[
+                      Object.values(workspaceFolderPrefixes).find((val) => val.index === 0)?.name ?? ''
```
neilk-aws, line 76: Should this be a sorted list? I thought we were going to sort the workflow folder prefixes and then pick the first one. Is there the risk of randomizing placement if we don't sort, or is it already sorted?

neilk-aws (approved):

packages/core/src/amazonqFeatureDev/session/sessionState.ts
tverney: So, VS Code already provide the position that the folder is it. In this way, we don't need sort before hand. However, if the goal is to place in the first one alphabetically, then would require to sort and then pick the first regardless the index its shown in VS Code (this could be more tricky because user don't actually organize in a alphabetically sorted way, but in his "preferred" workspace.

PR #5673, opened 2024-09-25
fix(amazonq): do not trigger uninstall when extension auto-updates

ivikash:
Do not trigger uninstall when extension auto-updates

### DIFF
--- a/packages/core/src/shared/handleUninstall.ts
+++ b/packages/core/src/shared/handleUninstall.ts
@@ -7,57 +7,103 @@
 
 import * as vscode from 'vscode'
 import { existsSync } from 'fs'
+import * as semver from 'semver'
 import { join } from 'path'
 import { getLogger } from './logger/logger'
 import { telemetry } from './telemetry'
 import { VSCODE_EXTENSION_ID } from './extensions'
-import { extensionVersion } from './vscode/env'
 
 /**
- * Checks if the extension has been uninstalled by reading the .obsolete file
- * and comparing the number of obsolete extensions with the installed extensions.
+ * Checks if an extension has been uninstalled and performs a callback if so.
+ * This function differentiates between an uninstall and an auto-update.
  *
- * @param {string} extensionName - The name of the extension.
- * @param {string} extensionsDirPath - The path to the extensions directory.
- * @param {string} obsoleteFilePath - The path to the .obsolete file.
- * @param {function} callback - Action performed when extension is uninstalled.
- * @returns {void}
+ * @param extensionId - The ID of the extension to check (e.g., VSCODE_EXTENSION_ID.awstoolkit)
+ * @param extensionsPath - The file system path to the VS Code extensions directory
+ * @param obsoletePath - The file system path to the .obsolete file
+ * @param onUninstallCallback - A callback function to execute if the extension is uninstalled
  */
 async function checkExtensionUninstall(
-    extensionName: typeof VSCODE_EXTENSION_ID.awstoolkit | typeof VSCODE_EXTENSION_ID.amazonq,
-    extensionsDirPath: string,
-    obsoleteFilePath: string,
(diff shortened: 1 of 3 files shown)


packages/amazonq/.changes/next-release/Bug Fix-4d4e0aee-4811-4091-9962-752cafc35418.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "do not trigger uninstall event when the extension auto-updates"
+}
```
hayemaxi, line 4:
```suggestion
```
No need for a changelog item- the customer can't see the difference.

packages/core/src/shared/handleUninstall.ts
```diff
@@ -7,59 +7,113 @@
...
+/**
+ * Handles the uninstall process by calling the callback and logging the event.
+ *
+ * @param extensionFullName - The full name of the extension including version
+ * @param callback - The callback function to execute on uninstall
+ */
+async function handleUninstall(extensionFullName: string, callback: () => Promise<void>): Promise<void> {
+    await callback()
+    telemetry.aws_extensionUninstalled.run((span) => {
+        span.record({})
```
hayemaxi, line 112: Is this line necessary?

packages/core/src/shared/handleUninstall.ts
```diff
@@ -7,59 +7,113 @@
...
 }
 
+/**
+ * Checks if a newer version of the extension exists in the extensions directory.
+ * The isExtensionInstalled fn is used to determine if the extension is installed using the vscode API
+ * whereas this function checks for the newer version in the extension directory for scenarios where
+ * the old extension is un-installed and the new extension in downloaded but not installed.
+ *
+ * @param onUninstallCallback - A callback function to execute if the extension is uninstalled
+ * @param isExtensionInstalled - A function to check if the extension is installed
```
hayemaxi, line 76:
```suggestion
```

packages/core/src/shared/handleUninstall.ts
```diff
@@ -7,59 +7,113 @@
...
+                extensionDirEntries,
+                extensionId,
+                currentExtension.packageJSON.version
+            )
+
+            if (!newerVersionExists) {
+                // No newer version exists, so this is likely an uninstall
+                await handleUninstall(extensionFullName, onUninstallCallback)
+            } else {
+                getLogger().info(`UpdateExtension: ${extensionFullName} is being updated`)
```
hayemaxi, line 61:
```suggestion
                getLogger().info(`UninstallExtension: ${extensionFullName} is being updated - not an uninstall`)
```

packages/core/src/shared/handleUninstall.ts
```diff
@@ -7,59 +7,113 @@
...
+    obsoletePath: string,
+    onUninstallCallback: () => Promise<void>
 ): Promise<void> {
-    /**
-     * Users can have multiple profiles with different versions of the extensions.
-     *
-     * This makes sure the callback is triggered only when an explicit extension with specific version is uninstalled.
-     */
-    const extension = `${extensionName}-${extensionVersion}`
+    const extensionFullName = `${extensionId}-${extensionVersion}`
```
hayemaxi, line 32: It seems inconsistent to pull `extensionVersion` from the environment, but you have to pass in the extension ID. The caller has the choice of what ID to pass but in reality since the version will only match to the running extension there is not actually a choice.
 Why not pass in both or pull both from the env, e.g. `vscode.ExtensionContext.extension.id`.

packages/core/src/shared/handleUninstall.ts
ivikash: Updated.

hayemaxi (approved):

PR #5675, opened 2024-09-26
feat(dev): Stop code generation action

tverney:
## Problem

Currently /dev don't support stop code generation. This PR introduces this functionality (watch the video below).

## Solution

Mynah UI provides onStopChatResponse API which we can hook in the cancellation token provided on VS Code, sharing across an active session, aborting current progress. 

https://github.com/user-attachments/assets/7a563bdc-c26b-4052-92c5-df0eeed63ea8

Notes:

- RTS is already supporting the change on prod
- codeGenerationId and uploadId are now being generated in FE
- AppSec and pentest passed successfully 
- Telemetry is included in this PR and strings are reviewed/approved by Technical writer
- This stop action is only introduced in /dev feature. This did not enable in any other product for this extension.

---


tverney:
!!Important - FYI toolkit team: this is only to be merged in next week's release. CC: @neilk-aws

neilk-aws (changes requested):
Blocking to wait for security sign-off

tverney:
We have already the security sign-off. Ready to be reviewed cc: @neilk-aws

packages/core/src/amazonq/webview/ui/connector.ts
```diff
@@ -457,17 +463,6 @@ export class Connector {
         }
     }
 
-    onStopChatResponse = (tabID: string): void => {
-        switch (this.tabsStorage.getTab(tabID)?.type) {
-            case 'featuredev':
-                this.featureDevChatConnector.onStopChatResponse(tabID)
-                break
-            case 'cwc':
```
jpinkney-aws, line 465: Does this break anything on cwc's side?

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1186,6 +1191,16 @@
                     "required": false
                 }
             ]
+        },
+        {
+            "name": "amazonq_stopCodeGeneration",
+            "description": "User stopped the code generation",
+            "metadata": [
+                {
+                    "type": "tabID",
```
jpinkney-aws, line 1200: What information will having a tabID give us? Isn't it just a random string like xcfg2? Instead does it make sense to use the telemetry.ui_click metric instead?

packages/core/src/amazonq/webview/ui/followUps/handler.ts
```diff
@@ -46,6 +46,7 @@ export class FollowUpInteractionHandler {
         if (followUp.prompt !== undefined) {
             this.mynahUI.updateStore(tabID, {
                 loadingChat: true,
+                cancelButtonWhenLoading: false,
```
jpinkney-aws, line 49: Out of curiosity does this mean that every single update store will need this? Why isn't the default false and then its true when we want the cancel button (which I assume is the stop button) to show?

jpinkney-aws: The change itself kind of scares me because we need to make sure its propagated everything _and_ that everyone who makes changes to the updateStore knows why it exists

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -311,6 +313,16 @@ export class FeatureDevController {
         }
     }
 
+    private disposeToken(session: Session | undefined) {
```
jpinkney-aws, line 316: Can we add a comment on this function why its needed? I assume its because we want to refresh the token on every request? Or in this approach we only dispose it when we cancel something -- since otherwise the token will still be valid?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -439,25 +452,82 @@ export class FeatureDevController {
...
+                message:
+                    (remainingIterations ?? 0) <= 0
+                        ? "I stopped generating your code. You don't have more iterations left, however, you can start a new session."
+                        : `I stopped generating your code. If you want to continue working on this task, provide another description. You have ${remainingIterations} out of ${totalIterations} code generations left.`,
+                type: 'answer-part',
+                tabID: message.tabID,
+            })
+        }
+
+        if (((remainingIterations ?? 0) <= 0 && isStoppedGeneration) || !isStoppedGeneration) {
```
jpinkney-aws, line 502: If remainingIterations is 0 when undefined can we just add that into the function definition `remainingIterations: number = 0`

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -726,8 +779,24 @@ export class FeatureDevController {
     }
 
     private async stopResponse(message: any) {
-        const session = await this.sessionStorage.getSession(message.tabID)
-        session.state.tokenSource.cancel()
+        await telemetry.amazonq_stopCodeGeneration.run(async (span) => {
```
jpinkney-aws, line 782: I think I mentioned this in a different file but is there any reason we can't just use `telemetry.ui_click` here? since I believe this only being called when a user clicks on the stop generating button and tabID doesn't really give us anything valuable

packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -44,6 +44,8 @@ import { collectFiles, getWorkspaceFoldersByPrefixes } from '../../shared/utilit
 import { i18n } from '../../shared/i18n-helper'
 import { Messenger } from '../controllers/chat/messenger/messenger'
 
+const EmptyCodeGenID = 'EMPTY_CURRENT_CODE_GENERATION_ID'
```
jpinkney-aws, line 47: is this id some special field on the backend?

packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -284,29 +289,46 @@ export class CodeGenState extends CodeGenBase implements SessionState {
...
+                    if (action.tokenSource) {
+                        this.tokenSource = action.tokenSource
+                    }
+                })
                 span.record({
                     amazonqConversationId: this.config.conversationId,
                     credentialStartUrl: AuthUtil.instance.startUrl,
                 })
 
+                action.tokenSource?.token.onCancellationRequested(() => {
```
jpinkney-aws, line 303: Why are lies 292-297 and this the exact same?

packages/core/src/amazonqFeatureDev/session/sessionState.ts
```diff
@@ -453,24 +482,27 @@ export class MockCodeGenState implements SessionState {
...
         public filePaths: NewFileInfo[],
         public deletedFiles: DeletedFileInfo[],
         public references: CodeReference[],
         public tabID: string,
-        private currentIteration: number,
+        public currentIteration: number,
         public codeGenerationRemainingIterationCount?: number,
         public codeGenerationTotalIterationCount?: number,
         public uploadHistory: UploadHistory = {},
+        public superTokenSource?: vscode.CancellationTokenSource,
```
jpinkney-aws, line 499: what does superTokenSource mean? also instead of falling back to a default you can just do:
```
superTokenSource: vscode.CancellationTokenSource = new vscode.CancellationTokenSource()
```

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
tverney: Good call, I'll update

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
tverney: This change came from @kelvin-klchu, adding him here in the thread if makes sense or not.

packages/core/src/shared/telemetry/vscodeTelemetry.json
tverney: I'm pulling @kelvin-klchu to answer these

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
tverney: Sure, I'll provide a more clearer comment there. Essentially when user stops he still in the same session, so we need clean up the token and create a new one for user provide a new prompt.

packages/core/src/amazonqFeatureDev/session/sessionState.ts
tverney: Good call, rebased / merge issue. Just fixing this one

packages/core/src/amazonq/webview/ui/connector.ts
tverney: No, cwc don't use it currently.

packages/core/src/amazonq/webview/ui/followUps/handler.ts
dogusata: If none of the functionalities are supporting `onStopChatResponse` why we have it attached to the mynahui instance? If some functions are using it, then this change makes sense. However the defaults can be changed for newly generated tabs. But i don't see any specific difference between giving true to the tabs which will using it or false to tabs which will not.

dogusata: Additionally, giving the value during the tab generation is sufficient. Also, until send a new value, the previous value will remain.

tverney: @jpinkney-aws do we have a consensus about this specific ? Or do we need sync about it? I'm ok for now with @dogusata, no one its using it, I did safe check in all the places in this PR that we use the store.

hayemaxi: So just to confirm - state of `cancelButtonWhenLoading = false` must be passed if `onStopChatResponse` was specified earlier, otherwise it defaults to true. If `onStopChatResponse` wasn't passed, then it doesn't matter.

If this is correct then the current implementation is fine IMO.

tverney: Exactly, that's the behaviour. If onStopChatResponse isn't passed, it defaults to false.

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -197,6 +197,7 @@ export const createMynahUI = (
                 mynahUI.updateStore(tabID, {
                     loadingChat: true,
                     promptInputDisabledState: true,
+                    cancelButtonWhenLoading: true,
```
dhasani23, line 200: This is causing the "stop" button to always appear which is a problem for `/transform` because we have our own stop button on the card in the chat message.

dhasani23: image

PR #5677, opened 2024-09-26
tests(amazonq): performance test for LSP setup. 

Hweinstock:
## Problem
continuation of reliability work

## Solution
`tryInstallLsp` in `src/amazonq/lsp/lspController.ts` captures many potentially high risk paths in a single function, such as some I/O operations, and heavy use of the `admZip` dependency. 

Two performance tests were added, one for handling zipping and maniupulating many small files, and one with few large files. Thresholds determined by running in CI a few times and multiplying the observed median by 1.5 to 2.0 depending on variance. 

The "many files" test does 250 files w/ 10 bytes each. The "large files" does 10 files w/ 1000 bytes each. These values were chosen to avoid flakiness in CI.

### DIFF
--- a/packages/core/src/amazonq/lsp/lspController.ts
+++ b/packages/core/src/amazonq/lsp/lspController.ts
@@ -5,8 +5,8 @@
 
 import * as vscode from 'vscode'
 import * as path from 'path'
-import * as fs from 'fs-extra'
 import * as crypto from 'crypto'
+import { chmodSync, createWriteStream } from 'fs'
 import { getLogger } from '../../shared/logger/logger'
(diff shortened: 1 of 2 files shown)


jpinkney-aws (approved):
Since cpu/system are non deterministic we should document our performance testing strategy somewhere i.g. cpu/memory are purposely set higher to catch instances where cpu usage/memory/duration explodes on a code change and then use fs calls for determinism

packages/core/src/testInteg/tryInstallLsp.test.ts
justinmk3, line 1: probably should start a subdir in testInteg, maybe `testInteg/perf/`

PR #5684, opened 2024-09-27
feat(telemetry): Crash Monitoring

nkomonen-amazon:
## Problem
We want to know when the extension crashes. Having telemetry will be useful.
The problem is that when the extension crashes it is not running anymore so it cannot report that it crashed.

## Solution
Create a Crash Monitoring mechanism that has all the instances of an extension (Q/Toolkit) work together to monitor eachother for crashes, reporting of a crash if it is detected.

To review start with the main file `packages/core/src/shared/crashMonitoring/crashMonitoring.ts`, specifically the `CrashMonitoring.start()` method.

There is more information in the docstring of the class.

---


packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -68,6 +68,11 @@ export class GitIgnoreFilter {
 
 export type CurrentWsFolders = [vscode.WorkspaceFolder, ...vscode.WorkspaceFolder[]]
 
+export function isInWorkspace() {
```
justinmk3, line 71:
```suggestion
export function hasWorkspace() {
```

docs/vscode_behaviors.md
justinmk3, line 1: Could be named just `vscode.md`. Unlikely that we could have any other "vscode_foo" doc.

packages/core/src/dev/activation.ts
```diff
@@ -105,6 +107,11 @@ const menuOptions: Record<DevFunction, MenuOption> = {
         detail: 'Opens editor to all Auth Connections the extension is using.',
         executor: editSsoConnections,
     },
+    forceIdeCrash: {
+        label: 'Crash: Force IDE ExtHost Crash',
+        detail: `Will SIGKILL ExtHost, { pid: ${process.pid}, sessionId: '${getSessionId().slice(0, 8)}-...' }, but the IDE itself will not crash.`,
```
justinmk3, line 112: is this different than vscode's `Restart Extension Host` command? 
image

packages/amazonq/src/extensionNode.ts
```diff
@@ -32,6 +32,11 @@ export async function activate(context: vscode.ExtensionContext) {
  * the code compatible with web and move it to {@link activateAmazonQCommon}.
  */
 async function activateAmazonQNode(context: vscode.ExtensionContext) {
+    // Do not run in tests since this corrupts the telemetry we try and assert
+    if (!env.isAutomation()) {
+        await (await CrashMonitoring.instance()).start()
```
justinmk3, line 37: why is the isAutomation check needed? telemetry is already skipped in CI: https://github.com/aws/aws-toolkit-vscode/blob/56228d75eeeef5565f4e88ef156c605f6d15be4e/packages/core/src/shared/telemetry/telemetryClient.ts#L103-L105

If a test needs to ensure the metrics are reset before asserting them, then the test should just clear the metrics queue. That's already possible. Then that eliminates this isAutomation check (and the confusing comment above it).

packages/core/src/dev/activation.ts
```diff
@@ -435,6 +445,15 @@ async function expireSsoConnections() {
     )
 }
 
+export function forceQuitIde() {
+    // This current process is the ExtHost. Killing it will cause all the extensions to crash
+    // for the specific IDE instance.
+    // The PPID does not seem to be the IDE instance but the root VS Code process, killing
+    // this crashes all VS Code instances.
```
justinmk3, line 452: PPID = parent PID. https://nodejs.org/api/process.html#processppid is this comment needed?

does it kill all vscode instances or only the current one?


```suggestion
    // The PPID (parent process) is vscode itself, killing it crashes all vscode instances.
```

packages/core/src/dev/activation.ts
```diff
@@ -435,6 +445,15 @@ async function expireSsoConnections() {
     )
 }
 
+export function forceQuitIde() {
+    // This current process is the ExtHost. Killing it will cause all the extensions to crash
+    // for the specific IDE instance.
```
justinmk3, line 450:
```suggestion
    // for the current exthost (unless using "extensions.experimental.affinity").
```

packages/core/src/shared/crashMonitoring/crashMonitoring.ts
justinmk3, line 1: does this need its own subdir or could it live in `shared/crashmonitoring.ts`

packages/core/src/shared/crashMonitoring/crashMonitoring.ts
```diff
@@ -0,0 +1,626 @@
...
+                        // Debugger instances may incorrectly look like they crashed, so don't emit.
+                        // Example is if I hit the red square in the debug menu, it is a non-graceful shutdown. But the regular
+                        // 'x' button in the Debug IDE instance is a graceful shutdown.
+                        if (ext.isDebug) {
+                            getLogger().debug(`crashMonitoring: DEBUG instance crashed: ${JSON.stringify(ext)}`)
+                            return
+                        }
+
+                        // This is the metric to let us know the extension crashed
+                        telemetry.session_end.emit({
```
justinmk3, line 258: can we force the proxied sessionId into the sessionId field? there's not really any use for the sessionId field from the current (non-proxied) session.

packages/core/src/dev/activation.ts
nkomonen-amazon: `Restart Extension Host` still results in a graceful shutdown of the extension host (`deactivate()` runs), while this one is non-graceful

packages/amazonq/src/extensionNode.ts
nkomonen-amazon: This was related to something else. But I found a way to handle this more cleanly in the tests.

packages/core/src/dev/activation.ts
nkomonen-amazon: Killing the PPID kills all VS Code instances, not just the current one. After some testing there is a completely separate PID that kills the one specific VSC instance, but there isn't a reliable way to get it. We could maybe do PID+2 or something from some quick testing, but this is unreliable.

packages/core/src/shared/crashMonitoring/crashMonitoring.ts
nkomonen-amazon: sessionId is adjacent to clientId and is injected in a different part of the metric right before sending: https://github.com/aws/aws-toolkit-vscode/blob/72bd7f5097ed4d53b4746f74e08d6918103a84ca/packages/core/src/shared/telemetry/telemetryClient.ts#L109

We have to do some hacky stuff to get it to work but it is possible, though I don't think is worth the effort atm

docs/vscode_behaviors.md
nkomonen-amazon: We already have `vscode-config.md`. Not sure if this would be helpful to merge the two

nkomonen-amazon:
> can we remove getPids, binarySearch, and dependencies `proper-lock` , `ps-list` in this PR

@justinmk3 opened PR here: https://github.com/aws/aws-toolkit-vscode/pull/5707

packages/core/src/shared/errors.ts
```diff
@@ -376,6 +377,12 @@ export function scrubNames(s: string, username?: string) {
         'Users',
         'users',
         'home',
+        'tmp',
+        'aws-toolkit-vscode',
+        'globalStorage', // from vscode globalStorageUri
+        CrashMonitoring.rootDir,
+        CrashMonitoring.runningExtDir,
+        CrashMonitoring.shutdownExtDir,
```
justinmk3, line 385: is this going to un-scrub the user's entire /tmp dir path, or only the last segment?

packages/core/src/shared/errors.ts
```diff
@@ -15,6 +15,7 @@ import type * as os from 'os'
 import { CodeWhispererStreamingServiceException } from '@amzn/codewhisperer-streaming'
 import { driveLetterRegex } from './utilities/pathUtils'
 import { getLogger } from './logger/logger'
+import * as CrashMonitoring from './crashMonitoring/constants'
```
justinmk3, line 18: the main constants.ts file generally exists for this reason. we don't really want errors.ts importing a bunch of random modules (partly because of the circular dep you ran into, but also because it's a low-level module that should be importable ~everywhere)

 do these constants need their own crashmonitoring/constants.ts  module or can they just live in the existing constants.ts module

justinmk3 (approved):

packages/core/src/shared/errors.ts
nkomonen-amazon: These are just the dir names, not the full path. As part of the move to the main constants file I've renamed these variables to be more clear.

packages/core/src/shared/errors.ts
nkomonen-amazon: good point with the importable everywhere. I've moved these to the shared constants file

PR #5689, opened 2024-09-27
refactor(fs): prefer our fs and fs to fs-extra (pt1: non-test files)

Hweinstock:
## Problem
We want to avoid fs-extra, and eventually move it to the banned import list. As a first step, we remove fs-extra with as much of our own `fs.ts` work, and `fs` when not possible. 

This PR only works on non-testing code. This way we can ensure our tests are not compromised, and still show that our code exhibits the desired behavior. 

### Indirect Translations
- Replace `ensureDir` with `mkDir` as suggested here: https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/ensureDir.md 
- Also `ensureDir` with permissions arg can be `mkDir` with `chmod` following. 
- Replace `move` with `rename` since we aren't using virtual file system, or crossing disk partitions: https://stackoverflow.com/questions/8579055/how-do-i-move-files-in-node-js

Otherwise the translations are fairly direct. 

## Notes
As of 9/30 there is a tech debt test failing blocking potential merge. This test was pushed back to 10/30. 

---


packages/core/src/shared/extensions/yaml.ts
```diff
@@ -4,12 +4,12 @@
  */
 
 import * as vscode from 'vscode'
-import { readFileSync } from 'fs-extra'
 import { VSCODE_EXTENSION_ID } from '../extensions'
 import { getLogger } from '../logger/logger'
 import { getIdeProperties } from '../extensionUtilities'
 import { activateExtension } from '../utilities/vsCodeUtils'
 import { AWS_SCHEME } from '../constants'
+import * as fs2 from 'fs'
```
justinmk3, line 12:
```suggestion
import * as nodefs from 'fs'
```

packages/core/src/shared/sam/debugger/goSamDebug.ts
```diff
@@ -254,16 +254,16 @@ async function installDebugger(debuggerPath: string): Promise<boolean> {
...
         } else if (code) {
             getLogger().warn(`Install script did not sucessfully run, using old Delve binary...`)
         } else {
             getLogger().info(`Installed Delve debugger in ${debuggerPath}`)
         }
     } catch (e) {
-        if (installScript && (await fs2.exists(installScript.path))) {
-            fs.unlinkSync(installScript.path) // Removes the install script since it failed
+        if (installScript && (await fs.exists(installScript.path))) {
+            await unlink(installScript.path) // Removes the install script since it failed
```
justinmk3, line 266: `fs.delete` ?

packages/core/src/shared/sam/localLambdaRunner.ts
```diff
@@ -307,7 +307,7 @@ export async function runLambdaFunction(
         const payload =
             config.eventPayloadFile === undefined
                 ? undefined
-                : JSON.parse(await readFile(config.eventPayloadFile, { encoding: 'utf-8' }))
+                : JSON.parse(await fs.readFileAsString(config.eventPayloadFile))
```
justinmk3, line 310: off-topic: we should rename this function :)
```suggestion
                : JSON.parse(await fs.readFileText(config.eventPayloadFile))
```

packages/core/src/shared/sshConfig.ts
```diff
@@ -109,9 +110,17 @@ export class SshConfig {
         const sshConfigPath = getSshConfigPath()
         const section = this.createSSHConfigSection(proxyCommand)
         try {
-            await fs.ensureDir(path.dirname(path.dirname(sshConfigPath)), { mode: 0o755 })
-            await fs.ensureDir(path.dirname(sshConfigPath), 0o700)
-            await fs.appendFile(sshConfigPath, section, { mode: 0o600 })
+            const parentsDir = path.dirname(sshConfigPath)
+            const grandParentsDir = path.dirname(parentsDir)
+            // TODO: replace w/ fs.chmod once stub is merged.
```
justinmk3, line 115: https://github.com/aws/aws-toolkit-vscode/pull/5647 is ready to merge

packages/core/src/shared/utilities/cliUtils.ts
```diff
@@ -285,9 +285,7 @@ async function installSsmCli(
         await new TimedProcess('pkgutil', pkgArgs).run({ spawnOptions: { cwd: tempDir } })
         await new TimedProcess('tar', tarArgs).run({ spawnOptions: { cwd: tempPath } })
 
-        fs.copySync(path.join(tempPath, 'usr', 'local', 'sessionmanagerplugin'), outDir, {
-            recursive: true,
-        })
+        await fs.copy(path.join(tempPath, 'usr', 'local', 'sessionmanagerplugin'), outDir)
```
justinmk3, line 288: `fs.copy` needs a docstring, it's not obvious that it also supports folders.

justinmk3 (approved):
LGTM!

packages/core/src/shared/utilities/cliUtils.ts
Hweinstock: tracked as a followup, seems like there a few other methods that could use them as well in `fs`.

Hweinstock:
@aws/dexp @aws/codewhisperer-team

PR #5696, opened 2024-09-30
telemetry(amazonq): change code percentage event to not consider modification for toolkit telemetry

zixlin7:
## Problem
Product thinks that code percentage event should not consider user modification

## Solution
use raw accepted token count (without considering modification) for the percentage field in toolkit telemetry
there is no change to STE

---


leigaol (approved):
Better rename Pr to change code percentage event to not consider modification in toolkit telemetry reporting. 

mention that the percentage code written in STE is not changed.

hayemaxi (approved):

hayemaxi:
Test failures are unrelated- tech debt only

PR #5698, opened 2024-09-30
test(ec2): fix flaky tests regarding child node updates. 

Hweinstock:
## Problem
The issue is here: https://github.com/aws/aws-toolkit-vscode/blob/9eea1f3700930df79620ef26ff6a43450e593580/packages/core/src/awsService/ec2/explorer/ec2ParentNode.ts#L55-L64
This fires off multiple asynchronous calls on each child, with the function returning before any are resolved. Thus, the child node could be deleted from the set after this function returned (potentially when another piece of code is using it). This leads to `undefined` errors within that function. 
## Solution
Refactor the code to make use of a `for...of` with `await` pattern. 

---

### DIFF
--- a/packages/core/src/awsService/ec2/explorer/ec2ParentNode.ts
+++ b/packages/core/src/awsService/ec2/explorer/ec2ParentNode.ts
@@ -52,15 +52,19 @@ export class Ec2ParentNode extends AWSTreeNodeBase {
         )
     }
 
-    private updatePendingNodes() {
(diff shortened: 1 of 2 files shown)


justinmk3 (approved):
nice catch. maybe we should ban `.forEach`, it's pretty useless

Hweinstock:
Yeah, I set up an issue tracking this as part of the reliability work when I saw it caused a CI file on another branch. Luckily it seems like a quick fix.

PR #5701, opened 2024-10-01
fix(appcomposer): rename Application Composer to Infrastructure Composer

Galaoaoa:
## Problem
We will rename Application Composer to Infrastructure Composer as our service name. We need to do the same thing in VSCode toolkit

## Solution
We will rename all user visible things first


bildungsroman (approved):
Looks good!

justinmk3 (approved):
If you rename the file locations, please name it something short like `infraComposer/`

PR #5702, opened 2024-10-01
feat(amazonq): Decouple LSP from vector index creation; start LSP by default

leigaol:
## Problem

Current LSP start is coupled with vector index creation, this should be decoupled. We will be having new releases in the LSP to build some other indexes for everyone very soon ( not computation expensive as vector index), therefore we need to start LSP by default. The vector indexing, as a computational expensive index, will be only enabled if opt-in.

## Solution
Decouple LSP from vector index creation; start LSP by default

---

### DIFF
--- a/packages/amazonq/.changes/next-release/Feature-5576420e-6db2-4045-a99f-afced496da0f.json
+++ b/packages/amazonq/.changes/next-release/Feature-5576420e-6db2-4045-a99f-afced496da0f.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Start language server by default"
+}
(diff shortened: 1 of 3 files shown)


packages/amazonq/.changes/next-release/Feature-5576420e-6db2-4045-a99f-afced496da0f.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
```
Will-ShaoHua, line 2: should be `fix` instead?

leigaol: No this is not a bug

Will-ShaoHua: but i feel CX will expect something to come out if we mark it as a Feature

Will-ShaoHua: it's just my insight, not a requirement

nkomonen-amazon (approved):

PR #5705, opened 2024-10-01
telemetry(amazonq): Add client latency metrics for amazon q chat

jpinkney-aws:
## Problem:
- It's not clear what's meant by cwsprChatTimeToFirstChunk, cwsprChatTimeBetweenChunks, cwsprChatFullResponseLatency
- We don't have any client latency metrics

## Solution:
- Clarify the descriptions of cwsprChatTimeToFirstChunk, cwsprChatTimeBetweenChunks, cwsprChatFullResponseLatency
- Implement cwsprTimeToFirstDisplay, cwsprChatTimeBetweenDisplay, cwsprChatFullDisplayLatency metrics

## Other changes:
- Instead of instrumenting everything via a traceId instead use the tabId as the unique identifier and store a traceId when we get it inside of the uiEventRecorder. That way we can associate events from a tab to the trace without needing to pass the trace around the entire application.
- Lock stopChatMessageTelemetry on the tabId just in case something calls that twice for the same event

## Risks:
- All flows from start of chat message -> end of chat message need to be instrumented otherwise an addChatMessage event wont be sent.
- RIght now:
   - Right click context menu works
   - Normal chat message work
   - Follow up button works
   - Are there any other flows that we need to know about?
---

### DIFF
--- a/packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
+++ b/packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
@@ -8,11 +8,9 @@ import { ExtensionMessage } from '../commands'
 import { CodeReference } from './amazonqCommonsConnector'
 import { TabOpenType, TabsStorage } from '../storages/tabsStorage'
 import { FollowUpGenerator } from '../followUps/generator'
-import { TracedChatItem } from '../connector'
 
 interface ChatPayload {
     chatMessage: string
-    traceId?: string
     chatCommand?: string
 }
 
@@ -74,6 +72,17 @@ export class Connector {
     }
 
     followUpClicked = (tabID: string, messageId: string, followUp: ChatItemAction): void => {
+        /**
+         * We've pressed on a followup button and should start watching that round trip telemetry
+         */
+        this.sendMessageToExtension({
+            command: 'start-chat-message-telemetry',
+            trigger: 'followUpClicked',
+            tabID,
+            traceId: messageId,
+            tabType: 'cwc',
+            startTime: Date.now(),
+        })
         this.sendMessageToExtension({
             command: 'follow-up-was-clicked',
             followUp,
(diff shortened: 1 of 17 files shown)


packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -235,19 +245,16 @@ export const createMynahUI = (ideApi: any, amazonQEnabled: boolean) => {
...
-                     **/
-                    ideApi.postMessage({
-                        type: 'stopChatMessageTelemetry',
-                        tabID,
-                        traceId: item.traceId,
-                        tabType: tabsStorage.getTab(tabID)?.type,
-                    })
-                }
+                /**
+                 * We've received an answer for a tabID and this message has
```
leigaol, line 249: Can you double check this callback? Is this code path executed when the first valid response is rendered to the customer chat panel or is it something else?

leigaol: Once this is merged, can you update relevant documentation? Thank you!

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
```diff
@@ -183,17 +192,29 @@ export class Connector {
...
+         * When a user presses "enter" send an event that indicates
+         * we should start tracking the round trip time for this message
+         **/
+        this.sendMessageToExtension({
+            command: 'start-chat-message-telemetry',
+            trigger: 'onChatPrompt',
+            tabID,
+            traceId: messageId,
+            tabType: 'cwc',
+            startTime: Date.now(),
```
nkomonen-amazon, line 206: Use `globals.clock.Date.now()` since if we eventually want to stub the time with `installFakeClock()` it will not work with the way things currently are

nkomonen-amazon (approved):

packages/core/src/amazonq/webview/ui/main.ts
jpinkney-aws: The chat panel starts rending the response when it sees the answer part (lines 203-215). When we receive ChatItemType.ANSWER it means the message response is fully finished

justinmk3 (approved):

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
nkomonen-amazon: 🤦

jpinkney-aws:
Moved this back to draft so its not merged. They now want separate metrics for user latency and server latency

PR #5706, opened 2024-10-01
config(dev): re-enable Q dev API retry for read APIs

chengoramazon:
## Problem

- `/dev` conversation would got terminated by retry-able errors, since we disabled the default retry.

## Solution

- Re-enable the default to prevent `/dev` from terminated by transient/retry-able errors.

---

### DIFF
--- a/packages/core/src/amazonqFeatureDev/client/featureDev.ts
+++ b/packages/core/src/amazonqFeatureDev/client/featureDev.ts
@@ -26,8 +26,17 @@ import { createCodeWhispererChatStreamingClient } from '../../shared/clients/cod
 import { getClientId, getOptOutPreference, getOperatingSystem } from '../../shared/telemetry/util'
 import { extensionVersion } from '../../shared/vscode/env'
 
+// Re-enable once BE is able to handle retries.
+const writeAPIRetryOptions = {
+    maxRetries: 0,
+    retryDelayOptions: {
+        // The default number of milliseconds to use in the exponential backoff
+        base: 500,
+    },
+}
+
 // Create a client for featureDev proxy client based off of aws sdk v2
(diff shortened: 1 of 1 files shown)


packages/core/src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -26,8 +26,17 @@ import { createCodeWhispererChatStreamingClient } from '../../shared/clients/cod
 import { getClientId, getOptOutPreference, getOperatingSystem } from '../../shared/telemetry/util'
 import { extensionVersion } from '../../shared/vscode/env'
 
+// Re-enable once BE is able to handle retries.
+const writeAPIRetryOptions = {
```
willyyhuang, line 30: nit: naming, don't use camelcase for constants

chengoramazon: Variable name `WRITE_API_RETRY_OPTIONS` must match one of the following formats: camelCase, PascalCaseeslint[@typescript-eslint/naming-convention](https://typescript-eslint.io/rules/naming-convention)

willyyhuang (approved):

packages/core/src/amazonqFeatureDev/client/featureDev.ts
```diff
@@ -26,8 +26,17 @@ import { createCodeWhispererChatStreamingClient } from '../../shared/clients/cod
...
+    maxRetries: 0,
```
justinmk3, line 31: this is still at zero, so what effect does this PR have?

chengoramazon: It ensure Read APIs using default 3 retries. I don't handle write retries at this moment, but we would support it in the future.

PR #5713, opened 2024-10-02
refactor(fs): rename read functions to be more specific, add docstrings

Hweinstock:
## Problem
Follow up from: https://github.com/aws/aws-toolkit-vscode/pull/5689

## Solution
- rename 'readFileAsString' to 'readFileText' since this function assumes a text file (utf-8) encoding. 
- Add missing doc strings to copy and exist within fs wrapper. 

---


packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -133,7 +133,7 @@ export class LspController {
     }
 
     async getFileSha384(filePath: string): Promise<string> {
-        const fileBuffer = await fs.readFile(filePath)
+        const fileBuffer = await fs.readFileBytes(filePath)
```
justinmk3, line 136: if we are also renaming this one, then maybe the "text" version could simply be `readFile` ?

packages/core/src/shared/fs/fs.ts
```diff
@@ -291,6 +306,12 @@ export class FileSystem {
         await write(uri)
     }
 
+    /**
+     * Rename a file or folder.
```
justinmk3, line 310:
```suggestion
     * Rename (move) a file or folder.
```

justinmk3 (approved):

PR #5717, opened 2024-10-03
refactor(urls): Remove isCloud9 from constants file

nkomonen-amazon:
We have some strings that need to resolve a url depending on c9 vs toolkit.

The isCloud9() function is causing issues in the constants file. Due to some sort of module import order, isCloud9() did not resolve appropriately when expected. I think we used the constants file before all the dependencies of the isCloud9() function were available

The constants file should be a minimal module with as little external dependencies as possible.

## Solution:

Move the resolution of c9 vs toolkit docs to an external function. Update uses everywhere.

---

### DIFF
--- a/packages/core/src/shared/extensionUtilities.ts
+++ b/packages/core/src/shared/extensionUtilities.ts
@@ -15,6 +15,14 @@ import { extensionVersion, getCodeCatalystDevEnvId } from './vscode/env'
 import { DevSettings } from './settings'
 import globals from './extensionGlobals'
 import { once } from './utilities/functionUtils'
+import {
+    apprunnerCreateServiceDocUrl,
+    debugNewSamAppDocUrl,
+    documentationUrl,
+    launchConfigDocUrl,
+    samDeployDocUrl,
+    samInitDocUrl,
+} from './constants'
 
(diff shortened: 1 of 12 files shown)


jpinkney-aws:
>The isCloud9() function is causing issues in the constants file

What do you mean issues? Circular dependencies, etc?

packages/core/src/shared/extensionUtilities.ts
```diff
@@ -254,6 +262,28 @@ export function showWelcomeMessage(context: vscode.ExtensionContext): void {
     }
 }
 
+function _getDocumentationUrl(urls: { cloud9: vscode.Uri; toolkit: vscode.Uri }): vscode.Uri {
+    return isCloud9() ? urls.cloud9 : urls.toolkit
```
jpinkney-aws, line 266: rather than having to do vscode.Uri.parse everywhere you could just have c9/toolkit be strings and do the parsing in here I think

nkomonen-amazon: I was thinking that eventually I think we'll get rid of C9 and just directly use the variable in the constants file, so all these functions would be deleted

PR #5723, opened 2024-10-03
fix(dev): Q dev handle no change required

chengoramazon:
Problem

* Q dev won't be able to generate code for unrelated prompts. (e.g. explain the code)
* It would be a bad CX if Q fail to generate code without guiding users the proper use cases.

Solution

* Handle no code change required explicitly.
* Allow users to provide more details

https://github.com/user-attachments/assets/cf593cea-cd4d-4b29-9142-d5810d46709c

---


packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -246,6 +252,15 @@ export class FeatureDevController {
             case ZipFileError.errorName:
                 this.messenger.sendErrorMessage(errorMessage, message.tabID, 0, session?.conversationIdUnsafe, true)
                 break
+            case NoChangeRequiredException.errorName:
+                this.messenger.sendAnswer({
+                    type: 'answer',
+                    tabID: message.tabID,
+                    message: err.message,
```
tverney, line 259: Are we sending directly the message from BE? Is that revised?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -246,6 +252,15 @@ export class FeatureDevController {
...
+                    canBeVoted: true,
```
tverney, line 260: nit - not sure about vote an empty patch. Is that something that will lead just to negative voting?

packages/core/src/amazonq/webview/ui/quickActions/generator.ts
```diff
@@ -29,8 +29,7 @@ export class QuickActionGenerator {
                               {
                                   command: '/dev',
                                   placeholder: 'Describe your task or issue in as much detail as possible',
-                                  description:
-                                      'Plan and implement new functionality across multiple files in your workspace.',
+                                  description: 'Generate code to make a change in your project',
```
tverney, line 32: is this revised with the tech writer?

tverney (approved):

packages/core/src/amazonq/webview/ui/quickActions/generator.ts
chengoramazon: Yes, it is approved.

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
chengoramazon: We'd like users to provide feedbacks at any point of the experience.

tverney: I think they can by the top level menu

Screenshot 2024-10-04 at 12 55 50 PM

tverney: Voting its more thumbs up/down, that can lead to the same screen.

osdemah (approved):
LGTM

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
chengoramazon: That one is for overall product feedback. We can not distinguish which feature users refers to. This is a requirement from product earlier.

tverney: I see. So you mean voting we can? I cannot see how to "select" a feature here. That's what I see when thumbs down right now. Does this PR introduce a new behaviour related to that?

Screenshot 2024-10-04 at 1 12 26 PM

chengoramazon: User don't need select feature. It would contain the conversation context (e.g. feature, conversation) 
https://github.com/aws/aws-toolkit-vscode/commit/57382ad92d73f15591d119a28f5f2058d8a46fb2

tverney: btw, overall feedback user should also be able to select which feature its talking about. I should be able to provide an overall feedback to feature dev anytime of the experience.

chengoramazon: Out of the PR topic. Discussed offline.

jpinkney-aws (approved):

packages/amazonq/.changes/next-release/Bug Fix-ff32752d-fa0b-4d38-b719-68eb3c9ddca6.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Q dev handle no change required"
```
justinmk3, line 3: Thanks for including a changelog. Please review the [changelog guidelines](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#changelog).

PR #5724, opened 2024-10-03
fix(test): fix failing integ tests. 

Hweinstock:
## Problem
In the recent refactor (https://github.com/aws/aws-toolkit-vscode/pull/5689), there is a mismatch in args between our fs and fs-extra. Not caught until integ tests were run. 

## Solution
have default write file method ignores encoding since it uses utf8 anyway. 

---


jpinkney-aws (approved):
LGTM, the collectFiles one issue will be fixed by https://github.com/aws/aws-toolkit-vscode/pull/5725

PR #5729, opened 2024-10-04
feat(amazonq): Add project context to inline completion

Will-ShaoHua:
## Problem

1. Inline Completion of Q is not context aware.
2. The `@workspace` index does not update on file add/remove.

## Solution

1. We are adding new global BM25 index and class/function digest from imported files to provide project context for inline completion. This feature will be added as AB testing. In this PR, there will be no customer facing changes since it is hidden behind a AB testing flag. We will toggle it a few days later.  For this, we added new LSP APIs.

2. One customer facing change, when user add or remove a file in the VS Code IDE, its `@workspace` index will be updated 

The net-browserify is added to prevent npm run package failure for browser. 

---


leigaol (approved):

packages/webpack.base.config.js
```diff
@@ -56,6 +56,7 @@ module.exports = (env = {}, argv = {}) => {
                  * This is no longer the case. Verify if you need this module and configure a polyfill for it.
                  */
                 timers: false,
+                net: require.resolve('net-browserify'),
```
Will-ShaoHua, line 59: what's this for @leigaol

packages/core/src/codewhisperer/util/commonUtil.ts
```diff
@@ -30,6 +30,17 @@ export function asyncCallWithTimeout<T>(asyncPromise: Promise<T>, message: strin
...
+export function asyncCallWithTimeoutSwallowErrors<T>(asyncPromise: Promise<T>, timeLimit: number): Promise<T> {
+    let timeoutHandle: NodeJS.Timeout
+    const timeoutPromise = new Promise((resolve) => {
+        timeoutHandle = setTimeout(() => resolve(undefined), timeLimit)
+    })
+    return Promise.race([asyncPromise, timeoutPromise]).then((result) => {
+        clearTimeout(timeoutHandle)
+        return result as T
+    })
+}
```
Will-ShaoHua, line 42: don't quite understand why we need a different version of `asyncCallWithTimeout` with the only difference where we don't handle reject here. Can't we use `asyncCallWithTimeout`?

Will-ShaoHua: at least they seems to have pretty much  the same functionality?

packages/webpack.base.config.js
leigaol: This is a polyfill to resolve the npm run package failure because `net` is not compatible with the web build target. We don't need this poly fill if we don't build target for browser.

packages/core/src/codewhisperer/util/commonUtil.ts
leigaol: No it throws. `asyncCallWithTimeout` throws and when we `await Promise.all([promiseA, promiseB)]`, if promiseA throws then the entire promise throws. So we need A and B to not throw instead return undefined

packages/webpack.base.config.js
Will-ShaoHua: isee, thanks for the context

packages/core/src/amazonq/lsp/lspClient.ts
```diff
@@ -61,52 +71,67 @@ export class LspClient {
...
     }
 
+    async queryInlineProjectContext(query: string, path: string) {
+        try {
+            const request = JSON.stringify({
+                query: query,
+                filePath: path,
+            })
+            const encrpted = await this.encrypt(request)
+            const resp: any = await this.client?.sendRequest(QueryInlineProjectContextRequestType, encrpted)
```
hayemaxi, line 113:
```suggestion
            const encrypted = await this.encrypt(request)
            const resp: any = await this.client?.sendRequest(QueryInlineProjectContextRequestType, encrypted)
```

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -358,13 +371,14 @@ export class LspController {
                 result: 'Failed',
                 amazonqIndexFileCount: 0,
                 amazonqIndexFileSizeInMB: 0,
+                reason: `${e}`,
```
hayemaxi, line 374:
```
/** Reason code or name for an event (when result=Succeeded) or error (when result=Failed). Unlike the `reasonDesc` field, this should be a stable/predictable name for a class of events or errors (typically the exception name, e.g. FileIOException). */
```
Use reasonDesc if you want to publish the contents of the exception. Otherwise try to pull `code`, `name`, or similar field out of the exception if possible. This can be done automatically if you use `telemetry.run()` instead of `emit`.

packages/core/src/codewhisperer/util/commonUtil.ts
```diff
@@ -30,6 +30,17 @@ export function asyncCallWithTimeout<T>(asyncPromise: Promise<T>, message: strin
     })
 }
 
+export function asyncCallWithTimeoutSwallowErrors<T>(asyncPromise: Promise<T>, timeLimit: number): Promise<T> {
```
hayemaxi, line 33: maybe `tryTimeout`? This could be moved to a shared utility file or class.

hayemaxi (approved):
lgtm after comments

packages/core/src/amazonq/lsp/lspController.ts
leigaol: fixed in latest commit.

packages/core/src/codewhisperer/util/commonUtil.ts
leigaol: moved to timeoutUtils.ts

leigaol: also renamed

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -340,31 +351,35 @@ export class LspController {
...
-        } catch (e) {
+        } catch (error) {
+            //TODO: use telemetry.run()
             getLogger().error(`LspController: Failed to build index of project`)
             telemetry.amazonq_indexWorkspace.emit({
                 duration: performance.now() - start,
                 result: 'Failed',
                 amazonqIndexFileCount: 0,
                 amazonqIndexFileSizeInMB: 0,
+                reason: `Error when building index. ${error instanceof Error ? error.message : error}`,
```
nkomonen-amazon, line 375: `reason` is typically used as the error code/id, and `reasonDesc` is where youd put the error message

packages/core/src/shared/utilities/timeoutUtils.ts
```diff
@@ -387,3 +387,21 @@ export class TimeLag {
...
 }
+/**
+ * Tries to execute an asynchronous Promise with a specified time limit.
+ * If the Promise is not resolved within the time limit, it returns `undefined`.
+ *
+ * @param {Promise<T>} asyncPromise - The asynchronous Promise to execute.
+ * @param {number} timeLimit - The time limit in milliseconds.
+ * @returns {Promise<T | undefined>} A Promise that resolves with the result of the original Promise if it completes within the time limit, or `undefined` if the time limit is exceeded.
+ */
+export function tryTimeout<T>(asyncPromise: Promise<T>, timeLimit: number): Promise<T> {
```
nkomonen-amazon, line 398: I think you may be able to just use `waitUntil()` with the `timeout` value set to the time limit

packages/core/src/amazonq/lsp/lspController.ts
nkomonen-amazon: Just saw max's earlier comment, was the change not pushed?

hayemaxi: Yeah it seems that this change is still making a very verbose 'reason' field which is not very useful.
```suggestion
                reason: error.code ? error.code : error.name ? error.name : 'Unkown',
                reasonDesc: `Error when building index. ${error instanceof Error ? error.message : error}`,
```

packages/core/src/shared/utilities/timeoutUtils.ts
hayemaxi: This would be preferable

leigaol: Fixed

nkomonen-amazon (approved):

PR #5730, opened 2024-10-04
ci(lint): error on fs-extra import

Hweinstock:
## Problem
we want to avoid contributors using `fs-extra` package, and prefer our fs wrapper. 

## Solution
Have ESLint do the job for us, error on attempting to import `fs-extra`. 

### Notes 
- the script files that already use fs-extra have been given exceptions since we don't need these to run in web. 
- A few fs-extra references existed, and were replaced. 

---


justinmk3 (approved):

justinmk3:
> the script files that already use fs-extra have been given exceptions since we don't need these to run in web.

If it's not much work to use node's 'fs' in those files, then that would allow us to remove the fs-extra dependency.

PR #5731, opened 2024-10-04
fix(startup): Q may fail if Toolkit fails

justinmk3:
Problem:
If Toolkit fails to start, then Q will also fail to start:

    2024-10-01 09:56:13.188 [error] aws.amazonq.refreshConnectionCallback: [
      [TypeError: Cannot read properties of undefined (reading 'getApi')
      at c:\…\dist\src\extensionNode.js:6093:284

Solution:
Check for undefined `exports`.

### DIFF
--- a/packages/core/src/test/shared/credentials/userCredentialsUtils.test.ts
+++ b/packages/core/src/test/shared/credentials/userCredentialsUtils.test.ts
@@ -21,15 +21,6 @@ import { makeTemporaryToolkitFolder } from '../../../shared/filesystemUtilities'
 import { getConfigFilename, getCredentialsFilename } from '../../../auth/credentials/sharedCredentialsFile'
 import { fs } from '../../../shared'
 
-/** Async version of "doesNotThrow" */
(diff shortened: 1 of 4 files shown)


packages/core/src/test/shared/credentials/userCredentialsUtils.test.ts
```diff
@@ -151,8 +142,8 @@ describe('UserCredentialsUtils', function () {
                 `creds.secretKey: "${profile.aws_access_key_id}" !== "${creds.secretKey}"`
             )
 
-            await assertDoesNotThrow(async () => await fs.checkPerms(credentialsFilename, 'r--'))
-            await assertDoesNotThrow(async () => await fs.checkPerms(credentialsFilename, '-w-'))
+            await assert.doesNotReject(async () => await fs.checkPerms(credentialsFilename, 'r--'))
```
jpinkney-aws, line 145: Are these changes somehow related to this PR? Otherwise the change looks good to me

justinmk3: no just a fixup for a recent commit

PR #5738, opened 2024-10-07
feat(cwl): Create LiveTailSession object and registry. Adds MaxLine configuration

keeganirby:
## Problem
1. On the client side, VSCode needs to be aware of some additional context of a running LiveTail session, other than just the response from the StartLiveTail API call (Abort Controller, Max Lines config, TextDocument URI, etc). 

2. Additionally, we want to be able to have multiple LiveTail sessions running, and have a way to organize all of them. 

## Solution
1. Create a LiveTailSession class in AWSToolkit that will persist all of the metadata and context the LiveTail feature needs to operate.
2. Create a LiveTailSessionRegistry that maps a vscode URI to a LiveTailSession. The URI for a LiveTailSession is composed of the API request elements. This will allow us to tell when a user is making a duplicate request. This URI will also be used as the URI of the TextDocument to display the session results. 

## Additional Changes
* Adds VSCode preference config for LiveTail max events. This will be used to cap the number of live tail events can be in the TextDocument at a given time. When the limit is reached, the oldest log events will be dropped in order to fit new events streaming in.
* Adds a missing String for "Tail Log Group"

---


packages/core/src/awsService/cloudWatchLogs/registry/liveTailSessionRegistry.ts
```diff
@@ -0,0 +1,60 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import * as vscode from 'vscode'
+import { CLOUDWATCH_LOGS_LIVETAIL_SCHEME } from '../../../shared/constants'
+import { LiveTailSession, LiveTailSessionConfiguration } from './liveTailSession'
+import { ToolkitError } from '../../../shared'
+
+export class LiveTailSessionRegistry {
```
jpinkney-aws, line 10: You might be able to get away with extending NestedMap (from map.ts). It has the basic structure for a registry like this with get, set, delete already

packages/core/src/test/awsService/cloudWatchLogs/registry/liveTailRegistry.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    })
+})
+
+describe('LiveTailSession URI', async function () {
+    it('is correct with no logStream filter, no filter pattern', function () {
+        const config: LiveTailSessionConfiguration = {
+            logGroupName: 'test-log-group',
+            region: 'test-region',
+        }
+        const expectedUri = vscode.Uri.parse('aws-cwl-lt:test-region:test-log-group')
```
jpinkney-aws, line 68: nitpick: you could probably define the logGroupName and region at the top of this describe block and then build the uris on line 68 (and all the other instances) like:

```
vscode.Uri.parse(`${CLOUDWATCH_LOGS_LIVETAIL_SCHEME}:${testRegion}:${testLogGroup}:whatever other info here`)
```

that way if anything changes we don't have to update everything manually

packages/core/src/awsService/cloudWatchLogs/registry/liveTailSession.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    public get uri() {
+        return this._uri
+    }
+
+    public get logGroupName() {
+        return this._logGroupName
+    }
+
+    public startLiveTailSession(): Promise<StartLiveTailCommandOutput> {
```
jpinkney-aws, line 57: what calls these functions? Or is this what calls it coming in another PR?

packages/core/package.json
```diff
@@ -459,6 +459,7 @@
     },
     "dependencies": {
         "@amzn/codewhisperer-streaming": "file:../../src.gen/@amzn/codewhisperer-streaming",
+        "@aws-sdk/client-cloudwatch-logs": "^3.215.0",
```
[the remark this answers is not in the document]
keeganirby: Yes there is a newer version, will push with a revision!

packages/core/src/awsService/cloudWatchLogs/registry/liveTailSession.ts
keeganirby: Will be coming in another PR. The TailLogGroup command will call these functions.

packages/core/package.nls.json
```diff
@@ -231,6 +232,7 @@
     "AWS.cdk.explorerTitle": "CDK",
     "AWS.codecatalyst.explorerTitle": "CodeCatalyst",
     "AWS.cwl.limit.desc": "Maximum amount of log entries pulled per request from CloudWatch Logs (max 10000)",
+    "AWS.cwl.liveTailMaxEvents.desc": "Maximum amount of log entries that can be kept in a single live tail session. When the limit is reached, the oldest events will be removed to accomodate new events (min 1000; max 10000)",
```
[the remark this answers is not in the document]
keeganirby: I personally don't feel that is necessary. This is what a customer will see in the preferences menu when they first open it: 

Screenshot 2024-10-08 at 10 49 51 AM

So the customer will already see the 10k number filled in, implying it's the default value.

packages/core/src/awsService/cloudWatchLogs/registry/liveTailSessionRegistry.ts
keeganirby: Seems like that works for us! Will change the implementation.

jpinkney-aws (approved):

packages/toolkit/package.json
```diff
@@ -162,6 +162,13 @@
                     "description": "%AWS.cwl.limit.desc%",
                     "maximum": 10000
                 },
+                "aws.cwl.liveTailMaxEvents": {
```
justinmk3, line 165: why is this new setting needed instead of the existing `aws.cwl.limit` setting?

keeganirby: The reason I added a new one is the behavior of the setting. The existing limit preference doesn't mention LiveTail nor its line-trimming behavior. 

I figured that a user may only want to keep, say, 1000 log events in their LiveTail session since its streaming and they're getting constantly able to get new lines. Though, for our Search LogGroup/View LogStream commands, they could configure a higher limit since new events won't stream in and they want more results from their request without needing to send another. 

If Toolkit prefers to not have separate preferences then I can remove.

justinmk3: > The existing limit preference doesn't mention LiveTail nor its line-trimming behavior.

We can, and should, update its meaning and description.

> If Toolkit prefers to not have separate preferences then I can remove.

Yes. New settings should be avoided unless there is a strong reason to add them.

PR #5741, opened 2024-10-08
fix(telemetry): Crash monitoring fixes

nkomonen-amazon:
## Problem

Crash monitoring is reporting incorrect crash metrics.
This seems to be due to various filesystem errors such as eperm (even though we were doing an operation on a file we created), enospc (the user ran out of space on their machine, and other errors.

Because of this we ran in to situations where our state did not reflect reality, and due to this certain extension
instances were seen as crashed.

## Solution

- Determine if a filesystem is reliable on a machine (try a bunch of different filesystem flows and ensure nothing throws), if it is THEN we start the crash monitoring process. Otherwise we do not run it since we cannot rely it will be accurate.
  - We added a `function_call` metric to allow us to determine the ratio of successes to failures
- Add retries to critical filesystem operations such as the heartbeats and deleting a crashed extension instance from the state.
- Other various fixes

---

### DIFF
--- a/packages/core/src/shared/crashMonitoring.ts
+++ b/packages/core/src/shared/crashMonitoring.ts
@@ -205,6 +205,11 @@ class Heartbeat {
             await this.state.clearHeartbeat()
         } catch {}
     }
+
+    /** Mimics a crash, only for testing */
+    public testCrash() {
+        globals.clock.clearInterval(this.intervalRef)
+    }
 }
 
 /**
@@ -312,8 +317,13 @@ class CrashChecker {
         }
     }
 
-    /** Use this on failures */
+    /** Use this on failures to terminate the crash checker */
     public cleanup() {
+        globals.clock.clearInterval(this.intervalRef)
+    }
+
+    /** Mimics a crash, only for testing */
+    public testCrash() {
         globals.clock.clearInterval(this.intervalRef)
     }
 }
--- a/packages/core/src/shared/crashMonitoring.ts
+++ b/packages/core/src/shared/crashMonitoring.ts
@@ -492,7 +492,11 @@ export class FileSystemState {
         return p
     }
     public async clearState(): Promise<void> {
-        await withFailCtx('clearState', async () => fs.delete(this.stateDirPath, { force: true }))
+        this.deps.devLogger?.debug('crashMonitoring: CLEAR_STATE: Started')
+        await withFailCtx('clearState', async () => {
+            await fs.delete(this.stateDirPath, { force: true, recursive: true })
+            this.deps.devLogger?.debug('crashMonitoring: CLEAR_STATE: Succeeded')
+        })
     }
     public async getAllExts(): Promise<ExtInstanceHeartbeat[]> {
         const res = await withFailCtx('getAllExts', async () => {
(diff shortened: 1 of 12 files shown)


packages/core/src/shared/filesystemUtilities.ts
```diff
@@ -246,3 +247,117 @@ export async function setDefaultDownloadPath(downloadPath: string) {
...
+        })
+        await withFailCtx('renameFileReset', () => fs.rename(tmpFileRenamed, tmpFile))
+        await withFailCtx('renameFileResetExists', async () => {
+            const res = await fs.exists(tmpFile)
+            if (!res) {
+                throw new Error(`Expected reverted renamed file to exist: "${tmpFile}"`)
+            }
+        })
+    } finally {
+        await fs.delete(tmpFolder, { recursive: true, force: true })
```
jpinkney-aws, line 350: did we ever see any problems with delete? Just wondering if we might create a file and then never be able to delete it 😄

packages/core/src/shared/crashMonitoring.ts
```diff
@@ -453,8 +490,14 @@ export class FileSystemState {
 
         // Clean up the running extension file since it is no longer exists
         const dir = await this.runningExtsDir()
-        // Use force since another checker may have already removed this file before this is ran
-        await withFailCtx('deleteStaleRunningFile', () => fs.delete(path.join(dir, extId), { force: true }))
+        // Retry on failure since failing to delete this file will result in incorrectly reported crashes.
+        // The common errors we were seeing were windows EPERM/EBUSY errors. There may be a relation
+        // to this https://github.com/aws/aws-toolkit-vscode/pull/5335
+        await withRetries(() => withFailCtx('deleteStaleRunningFile', () => fs.delete(path.join(dir, extId))), {
+            maxRetries: 7,
```
jpinkney-aws, line 497: were this numbers determined by something?

packages/core/src/shared/crashMonitoring.ts
```diff
@@ -385,8 +408,18 @@ export class FileSystemState {
     /**
      * Does the required initialization steps, this must always be run after
      * creation of the instance.
+     *
+     * @throws if the filesystem state cannot get in to a good state
```
jpinkney-aws, line 412: can we define what a good state means?

packages/core/src/shared/crashMonitoring.ts
```diff
@@ -65,17 +67,32 @@ export class CrashMonitoring {
         private readonly devLogger: Logger | undefined
     ) {}
 
+    static #didTryInstance = false
```
jpinkney-aws, line 70: is this basically did attempt creation?

jpinkney-aws:
Looks like the withRetries test is being a bit sporadic on mac

packages/core/src/shared/crashMonitoring.ts
nkomonen-amazon: This just equated to 12.7 seconds at most which I think is enough gracetime for the fs operation to succeed. But I think it may be worth to double it to 25 seconds looking back at it

packages/core/src/shared/filesystemUtilities.ts
nkomonen-amazon: updated this function to write to the /tmp folder so if we are unable to delete it will eventually be cleaned up

jpinkney-aws (approved):

PR #5745, opened 2024-10-08
fix(dev):  add error messages for upload URL expiry

chengoramazon:
## Problem

- Upload url would expire in a given time. The original error message is not clear for users what to do next.

## Solution

- Handle URL expiry explicitly.
- Add a clear error message.
URLExpired

---

### DIFF
--- a/packages/core/src/amazonqFeatureDev/errors.ts
+++ b/packages/core/src/amazonqFeatureDev/errors.ts
@@ -98,6 +98,13 @@ export class UploadCodeError extends ToolkitError {
     }
 }
 
+export class UploadURLExpired extends ToolkitError {
+    static errorName = 'UploadURLExpired'
+    constructor() {
+        super(i18n('AWS.amazonq.featureDev.error.uploadURLExpired'), { code: 'UploadURLExpired' })
+    }
+}
+
 export class IllegalStateTransition extends ToolkitError {
     constructor() {
         super(i18n('AWS.amazonq.featureDev.error.illegalStateTransition'), { code: 'IllegalStateTransition' })
--- a/packages/core/src/amazonqFeatureDev/util/upload.ts
+++ b/packages/core/src/amazonqFeatureDev/util/upload.ts
@@ -42,6 +42,6 @@ export async function uploadCode(url: string, buffer: Buffer, checksumSha256: st
                     )
             }
         }
-        throw new ToolkitError(i18n('AWS.amazonq.featureDev.error.codeGen.default'))
+        throw ToolkitError.chain(e, i18n('AWS.amazonq.featureDev.error.codeGen.default'))
     }
 }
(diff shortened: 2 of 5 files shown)


packages/core/src/amazonqFeatureDev/util/upload.ts
```diff
@@ -30,8 +32,16 @@ export async function uploadCode(url: string, buffer: Buffer, checksumSha256: st
         }).response
     } catch (e: any) {
         getLogger().error(`${featureName}: failed to upload code to s3: ${(e as Error).message}`)
-        throw new UploadCodeError(
-            e instanceof RequestError ? `${e.response.status}: ${e.response.statusText}` : 'Unknown'
-        )
+        if (e instanceof RequestError) {
+            switch (e.response.status) {
+                case 403:
```
jpinkney-aws, line 37: can you elaborate more on the flow here? If I understand correctly I think its something like:
- we get presigned url for uploading, this is only available for a short time
- if there is a large gap of time between when we get the presigned url and when we start to upload the presigned url will be expired and return a 403?

Is that correct?

chengoramazon: Yes, it is correct. Users would see the issue if we don't finish uploading in time. It can be the repo is too large, or unstable internet.

willyyhuang (approved):
LGTM

packages/core/src/amazonqFeatureDev/errors.ts
```diff
@@ -98,6 +98,13 @@ export class UploadCodeError extends ToolkitError {
     }
 }
 
+export class UploadURLExpired extends ToolkitError {
+    static errorName = 'UploadURLExpired'
```
nkomonen-amazon, line 102: Does this make more sense to pass in to the `name` field, adjacent to the `code`?

packages/core/src/amazonqFeatureDev/util/upload.ts
```diff
@@ -30,8 +32,16 @@ export async function uploadCode(url: string, buffer: Buffer, checksumSha256: st
...
+                    throw new UploadURLExpired()
+                default:
+                    throw new UploadCodeError(
+                        e instanceof RequestError ? `${e.response.status}: ${e.response.statusText}` : 'Unknown'
+                    )
+            }
+        }
+        throw new ToolkitError(i18n('AWS.amazonq.featureDev.error.codeGen.default'))
```
nkomonen-amazon, line 45: Try `ToolkitError.chain(e, i18n(...))` as it creates a new error, while wrapping the underlying error in the `cause` field. Then if this error reaches a telemetry call it will be "unwrapped" and include the chain of errors

chengoramazon: Will followup

packages/core/src/amazonqFeatureDev/errors.ts
chengoramazon: It is following existing pattern.

nkomonen-amazon: Can you link to examples of existing uses?

I think those may be incorrect if they are going this route. ToolkitError.name is reusing the Error.name property

nkomonen-amazon (approved):

PR #5704, opened 2024-10-01
deps: bump minimum vscode version to 1.83

Hweinstock:
## Problem
Finishing up: https://github.com/aws/aws-toolkit-vscode/pull/4787

## Solution

---

License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5747, opened 2024-10-08
deps(vscode): temporarily de-bump min back to 1.68.0

Hweinstock:
## Problem
temporarily undo most of: https://github.com/aws/aws-toolkit-vscode/pull/5704

## Solution
- To re-bump we need to add changelog items, otherwise all work remains. 
- Considering adding a button to link to vscode download page, but thats an external link we don't control so holding off for now. 
- Without running it in the vsix, we have "null extension" in source field:
Screenshot 2024-10-08 at 4 03 32 PM

- Notification in toolkit and Q:
Screenshot 2024-10-08 at 4 06 26 PM
Screenshot 2024-10-08 at 4 07 23 PM

## Steps to re-bump to 1.83.0
- update vscode version in all package.json files. 
- add breaking fix changelog items in both extensions. 

## To-do: retest with vsix to make sure null source is gone. 
---


packages/toolkit/src/extensionNode.ts
```diff
@@ -7,15 +7,48 @@ import type { ExtensionContext } from 'vscode'
...
+    void setupVscodeVersionNotification()
     return awsToolkitApi
 }
 
 export async function deactivate() {
     await deactivateCore()
 }
+
+// TODO: remove once version bump to 1.83.0 is complete.
+export function setupVscodeVersionNotification() {
```
hayemaxi, line 28: Do we need duplicate code?
AFAIK the null extension bug only impacts when running the extension in debug mode and not when its packaged into a vsix.

Hweinstock: Ah, not aware of that. I can remove the duplicate code in that case.

packages/amazonq/src/app/chat/activation.ts
```diff
@@ -105,3 +106,31 @@ async function setupAuthNotification() {
...
+    tryShowNotification()
+
+    function tryShowNotification() {
+        // Do not show the notification if the IDE version will continue to be supported.
+        if (!semver.gte(vscode.version, '1.83.0')) {
+            return
+        }
+
+        if (notificationDisplayed) {
+            return
```
hayemaxi, line 121: Will this ever run? Its always set to false on line 111.
Also it seems its only called once on activation anyways. This seems like more code than we need.

Hweinstock: I honestly wasn't sure, but I was using this as an example: https://github.com/aws/aws-toolkit-vscode/blob/7d6a9751fe91597c223204bf4e3634ad8a4c764a/packages/amazonq/src/app/chat/activation.ts#L77-L107 and left it in to be safe.

hayemaxi: That example sorta makes sense because the state of `notificationDisplayed` can change after its initialized but before evaluation. But regardless it is a such a confusing and unnecessary way to evaluate the message. IMO we should refactor that.

justinmk3:
Oh darn, I should have mentioned I was working on this, see https://github.com/aws/aws-toolkit-vscode/pull/5749 which also uses PromptSettings to store a "don't show again"

packages/core/src/extension.ts
```diff
@@ -304,3 +304,32 @@ export async function emitUserState() {
...
+            return
+        }
+
+        if (notificationDisplayed) {
+            return
+        }
+
+        notificationDisplayed = true
+
+        telemetry.toolkit_showNotification.emit({
```
justinmk3, line 325: seems like we should have a `vscode.window.*` wrapper in `shared/utilities/messages.ts` that emits this (maybe optional by a boolean flag).

justinmk3:
https://github.com/aws/aws-toolkit-vscode/pull/5749 has PromptSettings + tests so let's continue with that.

PR #5755, opened 2024-10-09
refactor(filesystem): prefer our fs wrapper to node's with lint rule

Hweinstock:
## Problem
We want to avoid future unnecessary use of `fs` module and maximize usage of our wrapper fs. Benefits include:
- more control over interfacing with fs. 
- increased web compatibility. 
- minimizing dependencies. 

## Solution
- add a lint rule to discourage future imports of `fs`. 
- leave current uses as exceptions. 
- Planning follow up PR to migrate some of the easy uses of fs to our wrapper. 

---

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -180,6 +180,10 @@ module.exports = {
                         message:
                             'Avoid fs-extra, use shared/fs/fs.ts. Notify the Toolkit team if your required functionality is not available.',
                     },
+                    {
+                        name: 'fs',
+                        message: 'Avoid fs and use shared/fs/fs.ts when possible.',
+                    },
                 ],
             },
         ],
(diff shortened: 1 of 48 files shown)


.eslintrc.js
```diff
@@ -180,6 +180,10 @@ module.exports = {
                         message:
                             'Avoid fs-extra, use shared/fs/fs.ts. Notify the Toolkit team if your required functionality is not available.',
                     },
+                    {
+                        name: 'fs',
+                        message: 'Avoid fs and use shared/fs/fs.ts when possible.',
```
justinmk3, line 185:
```suggestion
                        message: 'Avoid node:fs and use shared/fs/fs.ts when possible.',
```

PR #5756, opened 2024-10-09
fix(dev): Allow user access history diff generated along iterations

tverney:
## Problem

We currently only support accessing the last code generated (by the Tree view). If you try open the diff view in past history, will thrown an error. This happens because when we click on past history Tree we actually send a "newer" uploadId and didn't keep track of all the past ones.

Screenshot 2024-10-09 at 11 07 58 AM

## Solution

Cache those uploadIds based on codeGenerationId which can be easily accessible while clicking in past views.

---

### DIFF
--- a/packages/amazonq/.changes/next-release/Bug Fix-dd58d9cb-4fb4-4cb3-a9a2-dc5c5462e529.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-dd58d9cb-4fb4-4cb3-a9a2-dc5c5462e529.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Amazon Q (/dev): Allow end-user access history diff generated along iterations"
+}
--- a/packages/amazonq/.changes/next-release/Bug Fix-dd58d9cb-4fb4-4cb3-a9a2-dc5c5462e529.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-dd58d9cb-4fb4-4cb3-a9a2-dc5c5462e529.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Amazon Q (/dev): Allow end-user access history diff generated along iterations"
+	"description": "Amazon Q (/dev): view diffs of previous /dev iterations"
 }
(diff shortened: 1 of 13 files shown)


packages/core/src/test/amazonqFeatureDev/session/sessionState.test.ts
```diff
@@ -27,10 +27,10 @@ const mockSessionStateAction = (msg?: string): SessionStateAction => {
             new AppToWebViewMessageDispatcher(new MessagePublisher<any>(new vscode.EventEmitter<any>()))
         ),
         telemetry: new TelemetryHelper(),
+        uploadHistory: {},
```
SrcsGammaPDXCanaryUser, line 30: nit: Consider adding a test where uploadHistory is populated and show we can access an element from it

SrcsGammaPDXCanaryUser (approved):

packages/core/src/test/amazonqFeatureDev/session/sessionState.test.ts
neilk-aws: Sorry, used the wrong account here.

neilk-aws (approved):

packages/core/src/test/amazonqFeatureDev/session/sessionState.test.ts
tverney: Oh, I thought was somekind of automatic messaging. Thanks, yea, due the timeline I would say we can add those testing scenarios as a follow up

packages/amazonq/.changes/next-release/Bug Fix-dd58d9cb-4fb4-4cb3-a9a2-dc5c5462e529.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Amazon Q (/dev): Allow end-user access history diff generated along iterations"
```
justinmk3, line 3: is the customer going to know what this means? is there a way to express this so that it's more meaningful to the customer? See [changelog guidelines](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#changelog).

I don't know if this is correct, but something like:

```suggestion
	"description": "Amazon Q (/dev): view diffs of previous /dev iterations"
```

tverney: SGTM

PR #5761, opened 2024-10-09
refactor(fs): use fs instead of fs-extra in `copyFiles`. 

Hweinstock:
## Problem
(title)

## Solution
Based on answer here: https://stackoverflow.com/questions/13786160/copy-folder-recursively-in-node-js we can use cp. This assumes node v16.7.0+. 
Note that `overwrite` is now `force` based on https://nodejs.org/api/fs.html#fscpsyncsrc-dest-options

For this to work, we must bump the dependency @types/node. Currently bumped to latest, but can try to test earliest version that works if that would be desirable. 

---


packages/amazonq/scripts/build/copyFiles.ts
```diff
@@ -73,33 +73,31 @@ const tasks: CopyTask[] = [
...
-            overwrite: true,
+            force: true,
             errorOnExist: false,
         })
     } catch (error) {
         throw new Error(`Copy "${src}" to "${dst}" failed: ${error instanceof Error ? error.message : error}`)
     }
 }
 
-void (async () => {
```
justinmk3, line 91: could keep this as a main() function, to avoid some churn below (plus it's a good habit)

     function main() {

Hweinstock:
I'm tempted to consolidate the shared logic among copyFiles somewhere, but its spanning across the extensions. Is it worth leaving it as duplicated code to avoid cross-extension imports?

justinmk3:
> I'm tempted to consolidate the shared logic among copyFiles somewhere, but its spanning across the extensions. Is it worth leaving it as duplicated code to avoid cross-extension imports?

It's worth creating a tracking issue but might involve some tradeoffs, so better to skip it for now.

Hweinstock:
Unfortunately the flag didn't make it work in CI, removed the outdated piece in the description.

justinmk3:
> TSError: ⨯ Unable to compile TypeScript:
> Error: scripts/build/copyFiles.ts(108,12): error TS2339: Property 'cpSync' does not exist on type 'typeof import("fs")'.

> /Users/runner/runners/2.320.0/externals/node20/bin/node: --experimental-fs is not allowed in NODE_OPTIONS

yeah I don't see a relevant flag at https://nodejs.org/dist/latest-v16.x/docs/api/cli.html#cli_node_options_options 

A way to check different node versions is to use https://github.com/nvm-sh/nvm and then try `fs.cpSync` on 16/18/20.

Hweinstock:
Yeah I used nvm to downgrade it locally to the same in CI, and it started working. Still unsure what it is about CI that makes it different.

justinmk3:
> Still unsure what it is about CI that makes it different.

Let's try bumping GHA CI to node 18 (or start with 22 and work backwards): https://github.com/aws/aws-toolkit-vscode/blob/4df3106b5ad8c4f09facaab0e523d0b8fbbbd803/.github/workflows/node.js.yml#L65

If node 18 works, we can use that in all CI. Or even newer. There's no reason reason why the `node` used in CI needs to match the `node` shipped with `vscode`; the vscode-shipped node will be used to execute the actual test code.

packages/amazonq/scripts/build/copyFiles.ts
```diff
@@ -73,33 +73,35 @@ const tasks: CopyTask[] = [
...
 ]
 
-async function copy(task: CopyTask): Promise<void> {
+function copy(task: CopyTask): void {
     const src = path.resolve(projectRoot, task.target)
     const dst = path.resolve(outRoot, task.destination ?? task.target)
 
     try {
-        await fs.copy(src, dst, {
+        fs.cpSync(src, dst, {
```
justinmk3, line 81: We could avoid the formal dependency on `@types/node@22` by casting this to `any`. We normally wouldn't do that, but it may be fine in this case because it avoids confusion, since our packages/ depend on `@types/node@16` (and that's intentional because vscode ships with node 16).

We can add a check to `techdebt.test.ts` to remind us to revisit this when we bump to node 18: https://github.com/aws/aws-toolkit-vscode/blob/6fa3a9e554f4aa1916d01d00ea93497ca7205cc5/packages/core/src/test/techdebt.test.ts#L30-L35


```suggestion
        (fs as any).cpSync(src, dst, {
```

packages/core/scripts/build/copyFiles.ts
```diff
@@ -46,27 +46,28 @@ const tasks: CopyTask[] = [
...
     } catch (error) {
         throw new Error(`Copy "${src}" to "${dst}" failed: ${error instanceof Error ? error.message : error}`)
     }
 }
-
-void (async () => {
+function main() {
     try {
-        await Promise.all(tasks.map(copy))
+        tasks.map(copy)
```
justinmk3, line 65: It's a bit surprising that we were running copy tasks in parallel. I doubt that gives much of a speedup, and it makes the logs less readable, and it could lead to weird races. I think this is better now.

justinmk3:
Nice! Can we remove fs-extra now? 

    npm uninstall --save '@types/fs-extra' 'fs-extra' -w packages/core/

justinmk3:
Merged. Let's do the above comments as a followup, that way if there are any problems it's easy to revert or drop.

packages/amazonq/scripts/build/copyFiles.ts
Hweinstock: I see, so once we update to 18, we will have `@types/node@18` which should allow us to remove the any?

Also, tracking as a follow-up.

PR #5762, opened 2024-10-09
telemetry(messages): maybeShowMinVscodeWarning

justinmk3:
## Problem

no metrics for notifications.

## Solution

add `toolkit_showNotification` to `showMessage` and use it in `maybeShowMinVscodeWarning`.

---

### DIFF
--- a/packages/core/src/shared/utilities/messages.ts
+++ b/packages/core/src/shared/utilities/messages.ts
@@ -16,8 +16,9 @@ import { getIcon, codicon } from '../icons'
 import globals from '../extensionGlobals'
 import { openUrl } from './vsCodeUtils'
 import { AmazonQPromptSettings, ToolkitPromptSettings } from '../../shared/settings'
-import { telemetry } from '../telemetry/telemetry'
+import { telemetry, ToolkitShowNotification } from '../telemetry/telemetry'
 import { vscodeComponent } from '../vscode/commands2'
+import { getTelemetryReasonDesc } from '../errors'
 
 export const messages = {
     editCredentials(icon: boolean) {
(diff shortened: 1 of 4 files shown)


packages/core/src/shared/utilities/messages.ts
```diff
@@ -35,21 +35,32 @@ export function makeFailedWriteMessage(filename: string): string {
...
-        case 'error':
-        default:
-            return vscode.window.showErrorMessage(message, { modal: useModal }, ...items)
+    options: vscode.MessageOptions & { telemetry?: boolean } = {},
+    metric: ToolkitShowNotification = {
+        id: 'unknown',
+        component: 'editor',
     }
+): Thenable<string | undefined> {
+    return telemetry.toolkit_showNotification.run(async (span) => {
```
nkomonen-amazon, line 48: nice idea

justinmk3: Yeah I need to think more about this, could you show a pseudocode example of what you have in mind? 

Meanwhile, I will push a lower-risk change for today's release.

justinmk3: Ah, I think I can make that work in a followup PR.

packages/core/src/shared/utilities/messages.ts
```diff
@@ -75,7 +86,16 @@ export async function showMessageWithUrl(
...
 
-    const p = showMessageWithItems(message, kind, items, useModal)
+    const p = showMessage(
+        kind,
+        message,
+        items,
+        { modal: useModal },
+        {
+            id: 'showMessageWithUrl',
+            reasonDesc: getTelemetryReasonDesc(message),
```
justinmk3, line 96: Setting `reasonDesc` based on `message`. Maybe in the future, `showMessageWithUrl` should allow the caller to set `id`.

nkomonen-amazon (approved):

packages/core/src/shared/utilities/messages.ts
justinmk3: https://github.com/aws/aws-toolkit-vscode/pull/5776

PR #5763, opened 2024-10-10
test(clients): user agent is set

justinmk3:
## Problem

user agent not tested for each client

## Solution

---

### DIFF
--- a/packages/core/src/shared/clients/codewhispererChatClient.ts
+++ b/packages/core/src/shared/clients/codewhispererChatClient.ts
@@ -9,14 +9,14 @@ import { AuthUtil } from '../../codewhisperer/util/authUtil'
 import { getUserAgent } from '../telemetry/util'
 
 // Create a client for featureDev streaming based off of aws sdk v3
-export async function createCodeWhispererChatStreamingClient(): Promise<CodeWhispererStreaming> {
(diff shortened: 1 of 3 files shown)


packages/core/src/shared/clients/codewhispererChatClient.ts
```diff
@@ -9,14 +9,14 @@ import { AuthUtil } from '../../codewhisperer/util/authUtil'
...
-    const bearerToken = await AuthUtil.instance.getBearerToken()
+export async function createCodeWhispererChatStreamingClient(bearerToken?: string): Promise<CodeWhispererStreaming> {
+    bearerToken = bearerToken ?? (await AuthUtil.instance.getBearerToken())
     const cwsprConfig = getCodewhispererConfig()
     const streamingClient = new CodeWhispererStreaming({
         region: cwsprConfig.region,
         endpoint: cwsprConfig.endpoint,
         token: { token: bearerToken },
-        customUserAgent: getUserAgent(),
+        customUserAgent: getUserAgent({ includePlatform: true, includeClientId: true }),
```
nkomonen-amazon, line 19: Does it make sense to have these true by default?

nkomonen-amazon (approved):

nkomonen-amazon:
@justinmk3 we probably want to get this one merged at some point

PR #5765, opened 2024-10-10
refactor(amazonq): move getIndentedCode to textDocumentUtilities and use cwsprChatInteractionType

ivikash:
move getIndentedCode to textDocumentUtilities and use cwsprChatInteractionType from aws-toolkit-common

---

### DIFF
--- a/packages/core/src/shared/utilities/textDocumentUtilities.ts
+++ b/packages/core/src/shared/utilities/textDocumentUtilities.ts
@@ -7,7 +7,7 @@ import * as _path from 'path'
 import * as vscode from 'vscode'
 import { getTabSizeSetting } from './editorUtilities'
 import { tempDirPath } from '../filesystemUtilities'
-import { fs, getIndentedCode, ToolkitError } from '../index'
+import { fs, indent, ToolkitError } from '../index'
 import { getLogger } from '../logger'
 
 /**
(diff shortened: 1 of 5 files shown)


packages/core/src/shared/utilities/textDocumentUtilities.ts
```diff
@@ -165,3 +165,22 @@ export async function createTempFileForDiff(
     await applyChanges(doc, range, code)
     return tempFileUri
 }
+
+/**
+ * Indents the given code based on the current document's indentation at the selection start.
+ *
+ * @param {any} message - The message object containing the code.
+ * @param {vscode.TextDocument} doc - The VSCode document where the code is applied.
```
nkomonen-amazon, line 173: non-blocker: This will make the value actually clickable in vscode when you hover the docstring and also if TextDocument were to be renamed this would automatically update. Note you need to add a string in between "param" and the {} or it doesnt render properly
```suggestion
 * @param - {@link vscode.TextDocument} doc - The VSCode document where the code is applied.
```

nkomonen-amazon (approved):

packages/core/src/shared/utilities/textDocumentUtilities.ts
```diff
@@ -165,3 +165,22 @@ export async function createTempFileForDiff(
...
+ * @param {vscode.Selection} selection - The selection range in the document.
+ * @returns {string} - The processed code to be applied to the document.
```
justinmk3, line 175: Typescript already has the types in the code itself, and any code analyzer (LSP, vscode) will pick this up.
```suggestion
 * @param message The message object containing the code.
 * @param doc The VSCode document where the code is applied.
 * @param selection The selection range in the document.
 * @returns The processed code to be applied to the document.
```

PR #5767, opened 2024-10-10
docs(logs): gear icon missing

nkomonen-amazon:
# Problem
If on an older version the gear icon to set the log level did not exist.

## Solution
Prompt user to update vs code if it is missing

---


justinmk3:
This isn't necessary because we will soon require vscode 1.83. And for any non-release we should perhaps aggressively show a "update vscode" warning if vscode is even a few versions behind.

PR #5768, opened 2024-10-10
telemetry(amazonq): Adding cwsprChatProgrammingLanguage parameter to insertAtCursor and CopyAtClipboard events

laileni-aws:
## Problem
- Client side telemetry event `interactWithMessage` is missing `cwsprChatProgrammingLanguage` parameter which helps to understand the language of generated code.

## Solution
- Added `cwsprChatProgrammingLanguage` parameter for insertAtCursor and CopyAtClipboard events.
- No Test cases were added in this PR.
- Added `extractCodeBlockLanguage` function as current UTG Sync API sends python and java in the first chunk of generated code but thats the not case for other languages. To counter this, I am checking entire response for generated code language and emitting those metrics.
- Added `cwsprChatProgrammingLanguage` in [commons](https://github.com/aws/aws-toolkit-common) package: https://github.com/aws/aws-toolkit-common/pull/883
- Added `cwsprChatHasProjectContext and cwsprChatTotalCodeBlocks` in  in [commons](https://github.com/aws/aws-toolkit-common) package: https://github.com/aws/aws-toolkit-common/pull/884
---


packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -334,6 +342,24 @@ export class Messenger {
             })
     }
 
+    private extractCodeBlockLanguage(message: string): string {
```
justinmk3, line 345: seems like something that should be tested. and maybe Q team wants to define a `shared/markdown.ts` file (i don't know how much other things like this are in this or similar modules)

justinmk3 (approved):

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
laileni-aws: Addressed this

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -582,7 +587,8 @@ export const createMynahUI = (
                 eventId,
                 codeBlockIndex,
                 totalCodeBlocks,
-                messageUserIntentMap.get(messageId) ?? undefined
+                responseMetadata.get(messageId)?.[0] ?? undefined,
```
ivikash, line 590: We should add these to `ViewDiff` and `AcceptDiff` as well.

laileni-aws: Yes, that can be in different PR as these two are unavailable for `Generate Tests(Beta)`.
@ivikash Correct me on this!

justinmk3:
> npm ERR! Invalid: lock file's @aws-toolkits/telemetry@1.0.267 does not satisfy @aws-toolkits/telemetry@1.0.271


use `npm install --save-dev` when updating dependencies.

justinmk3:
Thanks for upstreaming the metric to common. The build is failing because the old local metric still exists and overrides the common one. You need to delete it (in this PR): https://github.com/aws/aws-toolkit-vscode/blob/3b6197b2cfa8345a0775096f06e0983aec50102b/packages/core/src/shared/telemetry/vscodeTelemetry.json#L840

PR #5769, opened 2024-10-10
refactor(dev): remove errorName from Error classes

chengoramazon:
## Problem

- Static `errorName` is unnecessary in feature dev error classes. It also causes error code mismatches some of errorName which result in wrong state. 

## Solution

- Remove `errorName` from error classes
- use class name match to ensure error handled correctly.
- Add unit tests to cover all the error cases.

---

### DIFF
--- a/packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
+++ b/packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
@@ -16,6 +16,7 @@ import { assertTelemetry, toFile } from '../../../testUtil'
 import {
     NoChangeRequiredException,
     SelectedFolderNotInWorkspaceFolderError,
+    UploadURLExpired,
 } from '../../../../amazonqFeatureDev/errors'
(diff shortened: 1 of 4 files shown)


justinmk3 (approved):

justinmk3 (changes requested):
latest push has lots of duplicate code

willyyhuang (approved):

packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
```diff
@@ -385,22 +398,113 @@ describe('Controller', () => {
...
+                                session?.conversationIdUnsafe,
+                                true
+                            )
+                        )
+                        break
+                }
+            }
+
+            runs.forEach((run) => {
+                it(`should handle ${run.name}`, async function () {
```
justinmk3, line 505: For reference this is one way of de-duplicating tests, but it would also be fine to have a bunch of manally-declared it() blocks which call a common function.

justinmk3:
Thank you for this clean up!

PR #5773, opened 2024-10-10
feat(stepfunctions): Implement bi-directional sync between local file and WFS

witness-me:
## Problem
Workflow Studio custom editor does not communicate bi-directly with local asl file

## Solution
Adding bi-directional sync between local file and Workflow Studio custom editor
- On local file save, the change is sent and applied to WFS
- On every WFS action the change is synced with local file (without saving)
- On save action from WFS (Save button or Crtl+S) the local file is saved

Additionally, adding send feedback functionality + handling webview loading indicator

[Demo gif](https://maxis-file-service-prod-iad.iad.proxy.amazon.com/issues/bccf68da-2515-47fd-b549-65d7f7de39fe/attachments/4c11f37d32d60190aba363c20716f554045326d9381107c403f741c0496cd7b4_cb34c843-ee3b-42a3-b725-a7e87856741a)

---

### DIFF
--- a/packages/core/src/shared/telemetry/vscodeTelemetry.json
+++ b/packages/core/src/shared/telemetry/vscodeTelemetry.json
@@ -517,6 +517,24 @@
             ]
         },
         {
+            "name": "stepfunctions_saveFile",
+            "description": "Called after the user saves local ASL file (inlcuding autosave) from VSCode editor or Workflow Studio",
+            "metadata": [
+                {
+                    "type": "id",
+                    "required": true
+                },
+                {
+                    "type": "saveType",
+                    "required": true
+                },
+                {
+                    "type": "source",
+                    "required": true
+                }
+            ]
+        },
+        {
             "name": "vscode_activeRegions",
             "description": "Record the number of active regions at startup and when regions are added/removed",
             "unit": "Count",
--- a/packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
+++ b/packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
@@ -1,0 +1,172 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+    Command,
+    Message,
+    MessageType,
+    SaveFileRequestMessage,
+    WebviewContext,
+    InitResponseMessage,
+    FileChangedMessage,
+    FileChangeEventTrigger,
+} from './types'
+import { submitFeedback } from '../../feedback/vue/submitFeedback'
+import { placeholder } from '../../shared/vscode/commands2'
+import * as nls from 'vscode-nls'
+import vscode from 'vscode'
+import { telemetry } from '../../shared/telemetry/telemetry'
+import { ToolkitError } from '../../shared/errors'
+const localize = nls.loadMessageBundle()
+
+/**
+ * Handles messages received from the webview. Depending on the message type and command,
+ * calls the appropriate handler function
+ * @param message The message received from the webview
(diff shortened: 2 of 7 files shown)


packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -516,6 +516,20 @@
                 }
             ]
         },
+        {
+            "name": "stepfunctions_saveFileFromWorkflowStudio",
+            "description": "Called after the user saves or updates the file in Step Functions Workflow Studio custom editor",
```
hayemaxi, line 521: Can you create an action item to upstream this to aws-toolkit-common after release?

packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
```diff
@@ -0,0 +1,43 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Command, Message, MessageType, SaveFileRequestMessage, WebviewContext } from './types'
+import { saveFileMessageHandler, autoSaveFileMessageHandler } from './handlers/saveFileMessageHandler'
+import { initMessageHandler } from './handlers/initMessageHandler'
+import { loadStageMessageHandler } from './handlers/loadStageMessageHandler'
+import { openFeedbackMessageHandler } from './handlers/openFeedbackMessageHandler'
```
hayemaxi, line 10: do we need a file for each handler? Can we just put them all in this file?

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -516,6 +516,20 @@
                 }
             ]
         },
+        {
+            "name": "stepfunctions_saveFileFromWorkflowStudio",
```
justinmk3, line 520: this is over-specific. if you later want to track other sources, then you need another metric. that's a lot of churn and also requires dashboard/query updates.

instead, just name this `stepfunctions_saveFile` and include the `source` field which allows you to set `source=workflowstudio` or whatever you want.

Please always think about this guideline in the future: avoid over-specific concepts (in telemetry, code, etc; all interfaces everywhere, with rare exceptions).

justinmk3 (changes requested):
Current comments are blockers.

packages/core/src/shared/telemetry/vscodeTelemetry.json
witness-me: I have already created a task in the project to move the new metrics (since this question was raised in the previous pull request)

packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
witness-me: Putting all of them in the same file is definitely and option, I just thought it was the best practice to keep them in separate files for more clarity (looking at the examples of how it's implemented for Application Composer and Threat Composer integrations in the similar `handleMessage.ts` files).

However, I'm open to consolidating all handlers into this file if you think it’s a better approach. My main concern is ensuring consistency, as for other integrations the handlers are separated

hayemaxi: Generally we prefer less files so that its easier to track and understand code. It also helps reduce the chances of circular dependencies.
App composer's approach here seems to be a miss in our codebase- we should do the same refactor there if we touch that code again.

witness-me:
New revision addresses the comments:
- chaining toolkit errors inside `telemetry.xxx.run()`
- generalizing the save metric and updating its usage
- moving all message handler functions to a single file

justinmk3 (approved):

hayemaxi (approved):

PR #5778, opened 2024-10-11
test(ec2): add security related unit tests

Hweinstock:
## Problem
We want unit tests to ensure certain security measures are maintained in the code. Properties tested in this PR include:
1. create inline policy with minimal actions necessary to perform desired action
2. prompt user to confirm attached role and allow cancellation. 
3. test run the ssh agent before using it for remote connections. 

## Solution
For each test we do the following:

1. assert that the actions we are adding to the policy remains as the core 6. 
2. use `getTestWindow()` to ensure user's are able to see policies added, and cancel the process. 
3. spy on `tryRun` to ensure it is used before returning the ssh path. 

As part of (3), we refactor the pathfinding code to be contained in its own class. This makes the code easier to test since we can now use `sinon.spy` and `sinon.stub`. It also allows us to move some of the shared logic into one place. 

---


hayemaxi (approved):
lgtm!

packages/core/src/shared/utilities/pathFinder.ts
justinmk3, line 1: Please avoid renaming this file. And the introduction of the "class" abstraction looks unnecessary, or at least it should wait until all feature-branches are merged in the upcoming weeks. Because there are similar changes on a feature branch (but without the class abstraction), so if your necessary changes are small, it will help avoid big noisy merge conflicts in this file.

Hweinstock: I introduced the class to make it easier to spy/stub the methods. I wanted to have tests that assert some of these functions were called. Is there a natural way to do this without the class abstraction?

justinmk3: Maybe not, but the test added here only stubs `findSshPath` and the test seems over-specific. Is it important to test the "dry run" behavior? Testing the actual result is usually much stronger coverage. Adding a test for `findSshPath` is worth doing.

Meanwhile, avoiding churn in pathFind.ts will be very helpful for upcoming feature-branch merges.

Hweinstock: I agree that its a little over-specific to the implementation, but AppSec was asking for a test to show we perform some kind of "validation" of the ssh path found and I am unsure how else to show that. 

Testing `findSshPath` without stubs would involve some manipulation of the test environment $PATH variable, which seems risky unless we already have some well-tested utility code to do this for us. 

I found another way to stub/spy the method that significantly reduces churn in pathFind.ts. If we move `tryRun` to another file, and import it as a module, we can stub it in the tests. I think this also makes sense since `tryRun` doesn't fit in perfectly with all the path finding code. 

Still working on fixing the tests on Windows.

packages/core/src/shared/utilities/programUtils.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Tries to execute a program at path `p` with the given args and
+ * optionally checks the output for `expected`.
+ *
+ * @param p path to a program to execute
+ * @param args program args
+ * @param doLog log failures
+ * @param expected output must contain this string
+ */
+export async function tryRun(
```
justinmk3, line 18: please leave it in pathFind.ts

it's not a goal for everything in shared/ to be named fooUtils. pathFind.ts was intentionally an attempt to avoid the "fooUtil" pattern where possible, to have more meaningful module names.

packages/core/src/test/shared/utilities/pathFind.test.ts
```diff
@@ -4,19 +4,23 @@
...
 import * as path from 'path'
-
 import * as testutil from '../../testUtil'
-import { findTypescriptCompiler, getVscodeCliPath } from '../../../shared/utilities/pathFind'
 import { fs } from '../../../shared'
+import { findSshPath, findTypescriptCompiler, getVscodeCliPath } from '../../../shared/utilities/pathFind'
+import * as programUtils from '../../../shared/utilities/programUtils'
+
+function isWin(): boolean {
+    return process.platform === 'win32'
```
justinmk3, line 17: i guess we could add this to `packages/core/src/shared/utilities/osUtils.ts` , since it's somewhat common.

packages/core/src/test/shared/utilities/pathFind.test.ts
```diff
@@ -42,4 +46,46 @@ describe('pathFind', function () {
...
+            tryRunStub = sinon.stub(programUtils, 'tryRun')
+            callOffset = isWin() ? 1 : 0
+        })
+
+        after(function () {
+            tryRunStub.restore()
+        })
+
+        it('first tries $PATH', async function () {
+            tryRunStub.onCall(callOffset).resolves(true)
```
justinmk3, line 66: a way to test this without stubs would be to set $PATH to a test fixture directory.

packages/core/src/shared/utilities/programUtils.ts
justinmk3: oh, i just saw this:

> I found another way to stub/spy the method that significantly reduces churn in pathFind.ts. If we move `tryRun` to another file, and import it as a module, we can stub it in the tests.

in that case, let's name the new module pathFind2.ts. no point in pretending that it is something else.

packages/core/src/test/shared/utilities/pathFind.test.ts
Hweinstock: I found a function in `src/shared/vscode/env.ts` that actually did the same thing, so I removed it.

packages/core/src/test/shared/utilities/pathFind.test.ts
Hweinstock: I see. I was nervous about having code messing with $PATH in case something went wrong. Tried to safely wrap it in a reusable util function and added some tests for it. I think with this utility, the tests feel much more useful, so thanks for pushing me in this direction.

packages/core/src/shared/utilities/pathFind.ts
```diff
@@ -6,44 +6,17 @@
...
-/**
- * Tries to execute a program at path `p` with the given args and
- * optionally checks the output for `expected`.
- *
- * @param p path to a program to execute
- * @param args program args
- * @param doLog log failures
- * @param expected output must contain this string
- */
-export async function tryRun(
```
justinmk3, line 29: can the function stay in this location? that will help avoid merge conflicts with incoming feature-branches .

packages/core/src/test/testUtil.ts
```diff
@@ -614,3 +614,25 @@ export function tryRegister(command: DeclaredCommand<() => Promise<any>>) {
...
+
+export function readEnvPath(): string {
+    return process.env.PATH || ''
+}
+/**
+ * Overwrite the $PATH environment variable for the span of the provided task, then reset it.
+ * @param newPath temporary $PATH value
+ * @param task work to be done with $PATH set.
+ */
+export async function withEnvPath(newPath: string, task: () => Promise<void>): Promise<void | never> {
```
justinmk3, line 630: not a blocker, but future idea: this could be generalized to `withEnv()` or even just `with()`:

```
with({
  env: ...
})
```

packages/core/src/test/shared/utilities/pathFind.test.ts
```diff
@@ -42,4 +41,50 @@ describe('pathFind', function () {
         const regex = /bin[\\\/](code|code-insiders)$/
         assert.ok(regex.test(vscPath), `expected regex ${regex} to match: "${vscPath}"`)
     })
+
+    describe('findSshPath', function () {
+        it('first tries ssh in $PATH', async function () {
+            const workspace = await testutil.createTestWorkspaceFolder()
+            const fakeSshPath = path.join(workspace.uri.fsPath, `ssh${isWin() ? '.cmd' : ''}`)
+
+            await testutil.withEnvPath(workspace.uri.fsPath, async () => {
```
justinmk3, line 50: if we just set $PATH here directly, can `afterEach` restore it? 

the "with" pattern can be useful but it might be confusing if we have different ways of saving/storing state in tests (but maybe there is a good reason for it)

Hweinstock: Yeah, this definitely reduces the amount of code as well. My only concern is that adding more behavior into `beforeEach` and `afterEach` can make it hard to isolate the behavior of each test, and sometimes makes it harder to add tests if it doesn't fit with the pattern. 

I think this is still the better route since it is inline with the rest of the test code (ex. sinon, clock).

packages/core/src/test/shared/utilities/testUtils.test.ts
```diff
@@ -0,0 +1,41 @@
...
+import { tryRun } from '../../../shared/utilities/pathFind'
+
+describe('envWithNewPath', function () {
+    it('gives current path with new PATH', function () {
+        const fakePath = 'fakePath'
+        assert.deepStrictEqual(envWithNewPath(fakePath), { ...process.env, PATH: fakePath })
+    })
+})
+
+describe('writeEnv', function () {
```
justinmk3, line 19:
```suggestion
describe('setEnv', function () {
```

packages/core/src/test/testUtil.ts
```diff
@@ -614,3 +614,15 @@ export function tryRegister(command: DeclaredCommand<() => Promise<any>>) {
 export function getFetchStubWithResponse(response: Partial<Response>) {
     return stub(request, 'fetch').returns({ response: new Promise((res, _) => res(response)) } as any)
 }
+
+export function setEnv(newEnv: NodeJS.ProcessEnv): void {
+    process.env = newEnv
+}
+
+export function readEnv(): NodeJS.ProcessEnv {
+    return process.env
```
justinmk3, line 623: this should return a shallow copy:
```suggestion
    return { ...process.env }
```

packages/core/src/test/shared/utilities/pathFind.test.ts
```diff
@@ -42,4 +41,93 @@ describe('pathFind', function () {
...
+
+    describe('findSshPath', function () {
+        let previousEnv: NodeJS.ProcessEnv
+
+        beforeEach(function () {
+            previousEnv = testutil.readEnv()
+        })
+
+        afterEach(function () {
+            testutil.setEnv(previousEnv)
```
justinmk3, line 53: if `readEnv` returns a shallow copy, then this becomes much more conventional/familiar:
```suggestion
            process.env = previousEnv
```

packages/core/src/test/shared/utilities/pathFind.test.ts
```diff
@@ -42,4 +41,93 @@ describe('pathFind', function () {
...
+
+        it('first tries ssh in $PATH (Non-Windows)', async function () {
+            // skip on windows because ssh in path will never work with .exe extension.
+            if (isWin()) {
+                return
+            }
+            const workspace = await testutil.createTestWorkspaceFolder()
+            const fakeSshPath = path.join(workspace.uri.fsPath, `ssh`)
+
+            testutil.setEnv(testutil.envWithNewPath(workspace.uri.fsPath))
```
justinmk3, line 64: if `readEnv` returns a shallow copy, then this becomes much more conventional/familiar:

```suggestion
            process.env.PATH = workspace.uri.fsPath
```

packages/core/src/test/shared/utilities/pathFind.test.ts
```diff
@@ -42,4 +41,93 @@ describe('pathFind', function () {
...
+            // Don't want to be messing with System32 on peoples local machines.
+            if (!isWin() || !isCI()) {
+                return
+            }
+            const expectedPathInCI = 'C:/Windows/System32/OpenSSH/ssh.exe'
+
+            if (!(await fs.exists(expectedPathInCI))) {
+                await testutil.createExecutableFile(expectedPathInCI, '')
+            }
+            const ssh = (await findSshPath(true))!
```
justinmk3, line 128: Nice! This gives a lot more confidence that spies.

justinmk3 (approved):

packages/core/src/test/shared/utilities/pathFind.test.ts
```diff
@@ -42,4 +41,90 @@ describe('pathFind', function () {
...
+        it('caches result from previous runs (Non-Windows)', async function () {
+            if (isWin()) {
+                return
+            }
+            // On non-windows, we can overwrite path and create our own executable to find.
+            const workspace = await testutil.createTestWorkspaceFolder()
+            // We move the ssh to a temp directory temporarily to test if cache works.
+            const fakeSshPath = path.join(workspace.uri.fsPath, `ssh`)
+
+            process.env = testutil.envWithNewPath(workspace.uri.fsPath)
```
justinmk3, line 102: since we are now shallow-copying `env` in beforeEach, we can simply directly assign `env.PATH` here. `envWithNewPath` was only needed when beforeEach was not shallow-copied.
```suggestion
            process.env.PATH = workspace.uri.fsPath
```

packages/core/src/test/shared/utilities/pathFind.test.ts
```diff
@@ -42,4 +41,90 @@ describe('pathFind', function () {
...
+
+        it('first tries ssh in $PATH (Non-Windows)', async function () {
+            // skip on windows because ssh in path will never work with .exe extension.
+            if (isWin()) {
+                return
+            }
+            const workspace = await testutil.createTestWorkspaceFolder()
+            const fakeSshPath = path.join(workspace.uri.fsPath, `ssh`)
+
+            process.env = testutil.envWithNewPath(workspace.uri.fsPath)
```
justinmk3, line 64:
```suggestion
            process.env.PATH = workspace.uri.fsPath
```

packages/core/src/test/shared/utilities/pathFind.test.ts
```diff
@@ -42,4 +41,90 @@ describe('pathFind', function () {
...
+
+        it('only returns valid executable ssh path (Non-Windows)', async function () {
+            if (isWin()) {
+                return
+            }
+            // On non-windows, we can overwrite path and create our own executable to find.
+            const workspace = await testutil.createTestWorkspaceFolder()
+            const fakeSshPath = path.join(workspace.uri.fsPath, `ssh`)
+
+            process.env = testutil.envWithNewPath(workspace.uri.fsPath)
```
justinmk3, line 83:
```suggestion
            process.env.PATH = workspace.uri.fsPath
```

packages/core/src/test/shared/utilities/pathFind.test.ts
```diff
@@ -42,4 +41,90 @@ describe('pathFind', function () {
         const regex = /bin[\\\/](code|code-insiders)$/
         assert.ok(regex.test(vscPath), `expected regex ${regex} to match: "${vscPath}"`)
     })
+
+    describe('findSshPath', function () {
+        let previousPath: string
+
+        beforeEach(function () {
+            previousPath = testutil.copyEnv().PATH!
```
justinmk3, line 49: if we aren't going to store the whole env, we can just save PATH directly.
```suggestion
            previousPath = testutil.env.PATH
```

packages/core/src/test/shared/utilities/pathFind.test.ts
```diff
@@ -42,4 +41,90 @@ describe('pathFind', function () {
         const regex = /bin[\\\/](code|code-insiders)$/
         assert.ok(regex.test(vscPath), `expected regex ${regex} to match: "${vscPath}"`)
     })
+
+    describe('findSshPath', function () {
+        let previousPath: string
```
justinmk3, line 46:
```suggestion
        let previousPath: string | undefined
```

packages/core/src/test/shared/utilities/testUtils.test.ts
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import assert from 'assert'
+import { createExecutableFile, createTestWorkspaceFolder, copyEnv } from '../../testUtil'
+import path from 'path'
+import { fs } from '../../../shared'
+import { isWin } from '../../../shared/vscode/env'
+import { tryRun } from '../../../shared/utilities/pathFind'
+
+describe('setEnv', function () {
```
justinmk3, line 12:
```suggestion
describe('copyEnv', function () {
```

justinmk3 (approved):

packages/core/src/test/shared/utilities/testUtils.test.ts
```diff
@@ -0,0 +1,39 @@
...
+import path from 'path'
+import { fs } from '../../../shared'
+import { isWin } from '../../../shared/vscode/env'
+import { tryRun } from '../../../shared/utilities/pathFind'
+
+describe('copyEnv', function () {
+    it('modifies the node environment variables (Non-Windows)', function () {
+        // PATH returns undefined on Windows.
+        if (isWin()) {
+            this.skip()
```
justinmk3, line 16: PATH should not be undefined on Windows. Even if it did, if I'm reading the test correctly, it looks like it should still pass

PR #5779, opened 2024-10-11
test(ec2): ensure that private key is not written to telemetry

Hweinstock:
## Problem
We want to enforce that the private key generated by the toolkit is not accidentally slipped into a telemetry metric. 

## Solution
Similar to `assertTelemetry`, we implement a `assertNoTelemetryMatch` test utility that scans all metrics to see if a specified keyword is included in its keys or values. 

With this utility, it is relatively straightforward to assert that this private key doesn't appear in the telemetry when we generate it and perform other operations with it. 

---


packages/core/src/shared/telemetry/telemetryLogger.ts
```diff
@@ -118,4 +118,13 @@ export class TelemetryLogger {
     public queryFull(query: MetricQuery): MetricDatum[] {
         return this._metrics.filter((m) => m.MetricName === query.metricName)
     }
+
+    /**
+     * Queries telemetry for metrics with metadata containing a keyword in the key or value
+     */
+    public queryKeyword(keyword: string): MetricDatum[] {
```
justinmk3, line 125: might as well support regex?
```suggestion
    public queryRegex(re: RegExp): MetricDatum[] {
```

packages/core/src/test/testUtil.ts
```diff
@@ -295,6 +295,13 @@ export function partialDeepCompare<T>(actual: unknown, expected: T, message?: st
     const partial = selectFrom(actual, ...keys(expected as object))
     assert.deepStrictEqual(partial, expected, message)
 }
+/**
+ * Finds all metrics whose metadata contains this keyword in their key value pairs, as part of the key OR the value.
```
justinmk3, line 299:
```suggestion
 * Asserts that no metrics metadata (key OR value) matches the given regex.
```

packages/core/src/test/testUtil.ts
```diff
@@ -295,6 +295,13 @@ export function partialDeepCompare<T>(actual: unknown, expected: T, message?: st
...
+ * @param keyword target substring to search for
+ */
+export function assertNotInTelemetryMetadata(keyword: string): void | never {
```
justinmk3, line 302:
```suggestion
export function assertNoTelemetryMatch(keyword: string): void | never {
```

packages/core/src/shared/telemetry/telemetryLogger.ts
```diff
@@ -118,4 +118,11 @@ export class TelemetryLogger {
     public queryFull(query: MetricQuery): MetricDatum[] {
         return this._metrics.filter((m) => m.MetricName === query.metricName)
     }
+
+    /**
+     * Queries telemetry for metrics with metadata containing a keyword in the key or value
```
justinmk3, line 123:
```suggestion
     * Queries telemetry for metrics with metadata key or value matching the given regex.
```

packages/core/src/test/awsService/ec2/model.test.ts
```diff
@@ -144,6 +147,22 @@ describe('Ec2ConnectClient', function () {
...
+
+            const testSelection = {
+                instanceId: 'test-id',
+                region: 'test-region',
+            }
+            const keyPath = path.join((await createTestWorkspaceFolder()).uri.path, 'keys')
+            const keys = await SshKeyPair.getSshKeyPair(keyPath, 30000)
+            await client.sendSshKeyToInstance(testSelection, keys, 'test-user')
+            const privKey = await fs.readFileText(keys.getPrivateKeyPath())
+            assertNoTelemetryMatch(privKey)
```
justinmk3, line 162: Is there some other metric we can positively assert, before this assert? That ensures that this doesn't pass merely because of a race. Alternatively, sleep for 50ms :(

justinmk3 (approved):

justinmk3:
Windows failure . The `c:\Users\RUNNER~1\` path looks suspicious but might be just cosmetic. Note from https://nodejs.org/api/fs.html#fschmodpath-mode-callback : 

> on Windows only the write permission can be changed, and the distinction among the permissions of group, owner, or others is not implemented.

So if write perms is not important here, as a workaround maybe we can skip that on Windows for now (in `sshKeyPair.ts`)

```
  1) Ec2ConnectClient
       sendSshKeysToInstance
         avoids writing the keys to any telemetry metrics:
     Error: ENOENT: no such file or directory, chmod 'c:\Users\RUNNER~1\AppData\Local\Temp\aws-toolkit-vscode\vsctkdc5cd3f6\keys'
      at async chmod (node:internal/fs/promises:1096:10)
      at async FileSystem.chmod (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\core\src\shared\fs\fs.ts:392:13)
      at async Function.generateSshKeyPair (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\core\src\awsService\ec2\sshKeyPair.ts:40:9)
      at async Function.getSshKeyPair (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\core\src\awsService\ec2\sshKeyPair.ts:31:9)
      at async Context.<anonymous> (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\core\src\test\awsService\ec2\model.test.ts:159:26)

Hweinstock:
Based on https://help.ubuntu.com/community/SSH/OpenSSH/Keys it sounds like keys should already have these permissions set correctly. So this chmod is mostly a double-check since the agent will throw an error if the permissions are too open. 

I changed the code to throw a more useful error, and only do chmod on non-windows.

Hweinstock:
Seems like there is a more general problem of the key not existing on windows after write. Investigating now.

justinmk3 (approved):

PR #5786, opened 2024-10-15
test(performance): make tests more deterministic by relying more on system counts

Hweinstock:
## Problem
Performance tests are currently flaky, making them a poor signal for performance regressions within the code base. Additionally, false alarms slow down development of unrelated features as test failures keep CI from being "green". 

Therefore, rather than relying only on system usage thresholds for performance regressions, we can count the number of high-risk / potentially slow operations made by the code as a deterministic measure of its performance. However, directly counting each individual potentially slow operation used within the performance tests highly couples the test to the implementation details, making the tests less effective if the implementation changes.

Therefore, the goal of this PR is the following:

1. decrease performance test flakiness by increasing thresholds. 
2. increase performance test effectiveness by relying on deterministic measures. 
3. avoid coupling the tests to the implementation details. 

## Solution
To meet goal (1), we increase the thresholds of the tests to decrease the changes of a false alarm. 

To meet goal (2), we count expensive operations. But, to avoid tying it to the implementation details, we count the expensive operations using somewhat-loose upper bounds. Thus, implementation changes modifying the exact number of expensive operations by a small constant do not set a false alarm. However, if they increase the number of expensive operations by a multiplicative factor, the upper bound will alert us. 

As an example, we don't want the test to fail if it makes 5-10 more file system calls when working with a few hundred files, but we do want the test to fail if it makes 2x the number of files system calls. Therefore, in the code the bounds are often described as "operations per file", since it is the multiplicative increase we are concerned about. This allows us to achieve goal (3). 

### Implementation Details
- The most common "expensive operation" we count is file system calls (through our `fs` module). Some other examples include the use of `zip` libraries or loading large files into memory. 
- We separate the upper bounds for the file system into `read` and `write` bounds. This granularity allows us to assert that specific code paths do not modify any files. 

## Notes
- `AdmZip` removed in https://github.com/aws/aws-toolkit-vscode/pull/4769 , so we don't address spying on it here. Once `zip.js` is implemented, we can spy on it in a follow up. 
---


justinmk3:
> * The most common "expensive operation" we count is file system calls (through our `fs` module).

We also have a fetch (http/network calls) abstraction which could be useful in the future.

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -28,17 +29,17 @@ export async function prepareRepoData(
     repoRootPaths: string[],
     workspaceFolders: CurrentWsFolders,
     telemetry: TelemetryHelper,
-    span: Span<AmazonqCreateUpload>
+    span: Span<AmazonqCreateUpload>,
+    zip: AdmZip = new AdmZip()
```
justinmk3, line 33: FYI: AdmZip will be replaced by zip.js https://github.com/aws/aws-toolkit-vscode/pull/4769

Hweinstock: ah, I see. Once that PR is merged I can refactor. I'll add a note to the description.

Hweinstock: Went ahead and deleted the AdmZip work. Once a new zip is implemented, this can be augmented to spy on it as needed.

packages/core/src/testInteg/perf/collectFiles.test.ts
```diff
@@ -8,55 +8,89 @@ import * as sinon from 'sinon'
...
-    {
-        linux: {
-            userCpuUsage: 85,
-            heapTotal: 2,
-            duration: 0.8,
+function performanceTestWrapper(totalFiles: number) {
+    return performanceTest(
+        {
+            darwin: {
+                userCpuUsage: 100,
```
jpinkney-aws, line 20: it looks like a lot of these values are just declared once and then used in darwin/linux/win32. Does it make sense to just have a field called default or something that can be passed in and used as the default for everything?

packages/core/src/testInteg/perf/prepareRepoData.test.ts
```diff
@@ -40,35 +50,44 @@ function performanceTestWrapper(numFiles: number, fileSize: number) {
...
 }
 
-function verifyResult(result: resultType, telemetry: TelemetryHelper, expectedSize: number): void {
+function verifyResult(setup: setupResult, result: resultType, telemetry: TelemetryHelper, expectedSize: number): void {
     assert.ok(result)
     assert.strictEqual(Buffer.isBuffer(result.zipFileBuffer), true)
     assert.strictEqual(telemetry.repositorySize, expectedSize)
     assert.strictEqual(result.zipFileChecksum.length, 44)
+
+    assert.ok(getFsCallsUpperBound(setup.fsSpy) <= setup.numFiles * 4, 'total system calls should be under 4 per file')
```
jpinkney-aws, line 80: I see a lot of different numbers that we are comparing the upper bound to. Out of curiosity how did you determine the number to use?

packages/core/src/testInteg/perf/collectFiles.test.ts
Hweinstock: Yeah, now that the thresholds are looser I think it makes sense to factor this out to reduce clutter. I wrote a helper function to reduce this.

jpinkney-aws (approved):
Looking forward to seeing ci green again 😎

packages/core/src/testInteg/perf/utilities.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Provide an upper bound on the number of filesystem reads done through our fs module.
+ * This is an upper bound because some of the functions call eachother.
+ * @param fsSpy filesystem spy (Ex. `sinon.spy(fs)`)
+ * @returns value
+ */
+export function getFsReadsUpperBound(fsSpy: sinon.SinonSpiedInstance<FileSystem>): number {
+    return (
+        fsSpy.readFileBytes.callCount +
```
justinmk3, line 25: Just a thought: do we actually need spies or could our fs.ts module just increment a count?

packages/core/src/amazonqFeatureDev/util/files.ts
justinmk3: https://github.com/aws/aws-toolkit-vscode/pull/4769 is merged now

PR #5787, opened 2024-10-15
fix(crash): handle sleep/wake appropriately

nkomonen-amazon:
## Problem:

When a user does a sleep then wake of their computer, the heartbeat is not up to date since it cannot be sent when the user's computer is asleep.

Due to this there is a race condition on wake between the next fresh heartbeat being sent versus when we check the heartbeats to determine if they are stale (crash). If the crash checker runs before a new heartbeat can be sent, it will be seen as a crash.

## Solution:

Use a TimeLag class that helps to determine when there is a time discrepancy. It works by updating a state every second, and if we determine that the next update to that state took longer than a second, we determine that there was a lag. Then we simply skip the next crash check, allowing a fresh heartbeat to be sent.

---


packages/core/src/shared/crashMonitoring.ts
```diff
@@ -324,6 +359,84 @@ class CrashChecker {
...
+ *
+ * Why? Intervals do not run while the computer is asleep, so the latest heartbeat has a "lag" since it wasn't able to send
+ *      a new heartbeat.
+ *      Then on wake, there is a racecondition for the next heartbeat to be sent before the next crash check. If the crash checker
+ *      runs first it will incorrectly conclude a crash.
+ *
+ * Solution: Keep track of the lag, and then skip the next crash check if there was a lag. This will give time for the
+ *           next heartbeat to be sent.
+ */
+class TimeLag {
```
justinmk3, line 381: what is this class abstracting? why is this not part of the existing Heartbeat class?

packages/core/src/shared/crashMonitoring.ts
```diff
@@ -324,6 +359,84 @@ class CrashChecker {
...
+    private intervalRef: NodeJS.Timer | undefined
+    private isCompleted: Promise<void> | undefined
+    /** Resolves {@link isCompleted} when the next interval of lag checking is completed */
+    private setCompleted: (() => void) | undefined
+
+    /** The last timestamp we remember. If this is more than 1 {@link lagCheckInterval} we probably did a sleep+wake */
+    private latestTimestamp: number = 0
```
justinmk3, line 388: these 3 properities are common to the existing Timeout class in our timoutUtils module. does using Timeout save any code?

packages/core/src/shared/crashMonitoring.ts
```diff
@@ -154,15 +173,19 @@ export class CrashMonitoring {
  */
 class Heartbeat {
```
justinmk3, line 174: Hearbeat is a concept that continues to come up, e.g. we have another heartbeat for codecatalyst: https://github.com/aws/aws-toolkit-vscode/blob/5573cbb1f56f9ec38a1c20572113eda1e5556d90/packages/core/src/codecatalyst/devEnv.ts#L65

Even if the codecatalyst heartbeat is too different to be useful here, is this new Heartbeat not amenable to generalization? It is a general concept that is needed by different features that want to emit data on a timed basis. 

Do we have a backlog item to pull this out / generalize it? Are features easily able to attach data to a `toolkit_heartbeat` metric, or did we name the metric `crashreporting_x` (not intuitive)?

nkomonen-amazon: There isn't much reuse between Codecatalyst heartbeat and this heartbeat atm. This one heartbeats on an interval, and the other heartbeats when the user performs actions in the IDE.

We could make an interface for the common apis, but this doesn't deduplicate anything existing code

packages/core/src/shared/crashMonitoring.ts
nkomonen-amazon: The underlying issue is **checking** for a stale heartbeat. We cannot reliably compensate for lag by sending a heartbeat since the crash checking is a separate process, and there becomes a race condition between sending a fresh heartbeat vs checking for a stale one. The solution is that the crash checker compensates for "time lag" by skipping a check

This class could live inside the CrashChecker class, but I'm currently passing the TimeLag instance in as a dependency to it.

justinmk3 (approved):

justinmk3:
Approved in case this is needed for the release.

Although CC heartbeat doesn't have much overlap, and currently there is no other consumer, we know that there are multiple future use-cases for attaching data to a global heartbeat (and this would have eliminated some existing metrics), so it's a good idea to start thinking about lifting this out of crash-reporting. Especially if a metric [will be] involved.

Possibly the `UserActivity` concept is related.

Issue #5681, opened 2024-09-26
unreliable test: "AuthUtil CodeWhisperer uses fallback connection" (timeout?)


justinmk3:
```
  1) AuthUtil
       CodeWhisperer uses fallback connection when switching to an unsupported connection:
     Error: Test length exceeded max duration: 30 seconds
[No Pending UI Elements Found]
      at Timeout._onTimeout (/codebuild/output/src3277921682/src/github.com/aws/aws-toolkit-vscode/packages/core/src/test/setupUtil.ts:45:32)
      at listOnTimeout (node:internal/timers:557:17)
      at processTimers (node:internal/timers:500:7)
```

PR #5788, opened 2024-10-15
fix(test): unreliable test "AuthUtil CodeWhisperer uses fallback connection"

nkomonen-amazon:
## Problem:

This test was flaky due to the event emitter and the test expecting 2 of the same event to come in, but sometimes the second one was not capture.

This looks to be due to the second `captureEventOnce` call needing to do some initial setup before it could capture an event. And I think there was a race condition between it being setup in time and the event being emitted before it could start listening.

## Solution:

Make a new function which does the event capture once and then listens N amount of times. This is reliable.

fix https://github.com/aws/aws-toolkit-vscode/issues/5681

---


justinmk3 (approved):
Nice approach and explanation

PR #5789, opened 2024-10-15
refactor(cwl): Change implementation of LiveTailRegistry to standard map

keeganirby:
## Problem
The nestedMap implementation of the registry was forcing a defined Default item. This doesn't fit the usecase I am aiming to provide by this registry. There is no default object to be returned. If an item doesn't exist in the registry, we should handle an undefined object, and not a placeholder. In other words, there is no concept of a placeholder/default value for a LiveTailSession.

Additionally, `set` on the NestedMap would perform a deep merge. I am wanting to replace the object. 

Originally I defined the default object to an exception, but the NestedMap `set` operation calls `get`. And if there isn't already a value in the map, it would throw the exception defined as the default object. This means we could never add any item to the map. 

## Solution
Swap the underlying implementation of the registry to a basic Map. 
I am still defining a `LiveTailSessionRegistry` class so we can statically initialize a registry singleton, as well as leaving the possibility open to override or extend the `Map` class. Also so we can pass around the type `LiveTailSessionRegistry` instead of `Map<Foo, Bar>`.

---


justinmk3 (approved):
Thanks for the clear explanation!

PR #5790, opened 2024-10-15
feat(cwl): initialize tailLogGroup command. starts stream and prints logEvents to textDocument

keeganirby:
## Problem
TailLogGroup currently is just logging the Users WizardResponse. It does not start a LiveTail stream or print results to the VSCode editor

## Solution
* Creates a `LiveTailSession` with the input params from the TailLogGroupWizard
* Registers the LiveTailSession in the `LiveTailSessionRegistry`
* Iterates the Response Stream and prints new LogEvents to a TextDocument
* If there are no open tabs for a given session's TextDocument, the stream will be closed by triggering its AbortController and disposing its client. 

We will use a new URI Scheme for LiveTail text documents so we can identify a LiveTail session vs a Cloudwatch SearchLogGroup/ViewLogStream document. In the future, we will register LiveTail specific CodeLenses that depend upon this URI scheme. Defining a new URI scheme for a Text Document requires a Document Provider. In this use case, it is empty given that the response handling that writes to the document lives within the TailLogGroup command itself. 

---

### DIFF
--- a/packages/core/src/awsService/cloudWatchLogs/activation.ts
+++ b/packages/core/src/awsService/cloudWatchLogs/activation.ts
@@ -4,7 +4,7 @@
  */
 
 import * as vscode from 'vscode'
-import { CLOUDWATCH_LOGS_SCHEME } from '../../shared/constants'
+import { CLOUDWATCH_LOGS_LIVETAIL_SCHEME, CLOUDWATCH_LOGS_SCHEME } from '../../shared/constants'
 import { Settings } from '../../shared/settings'
 import { addLogEvents } from './commands/addLogEvents'
 import { copyLogResource } from './commands/copyLogResource'
@@ -20,11 +20,15 @@ import { changeLogSearchParams } from './changeLogSearch'
 import { CloudWatchLogsNode } from './explorer/cloudWatchLogsNode'
 import { loadAndOpenInitialLogStreamFile, LogStreamCodeLensProvider } from './document/logStreamsCodeLensProvider'
 import { tailLogGroup } from './commands/tailLogGroup'
+import { LiveTailDocumentProvider } from './document/liveTailDocumentProvider'
+import { LiveTailSessionRegistry } from './registry/liveTailSessionRegistry'
 
 export async function activate(context: vscode.ExtensionContext, configuration: Settings): Promise<void> {
     const registry = LogDataRegistry.instance
+    const liveTailRegistry = LiveTailSessionRegistry.instance
 
     const documentProvider = new LogDataDocumentProvider(registry)
+    const liveTailDocumentProvider = new LiveTailDocumentProvider()
 
     context.subscriptions.push(
         vscode.languages.registerCodeLensProvider(
@@ -38,6 +42,10 @@ export async function activate(context: vscode.ExtensionContext, configuration: 
 
     context.subscriptions.push(
         vscode.workspace.registerTextDocumentContentProvider(CLOUDWATCH_LOGS_SCHEME, documentProvider)
+    )
+
+    context.subscriptions.push(
+        vscode.workspace.registerTextDocumentContentProvider(CLOUDWATCH_LOGS_LIVETAIL_SCHEME, liveTailDocumentProvider)
     )
 
     context.subscriptions.push(
@@ -97,7 +105,7 @@ export async function activate(context: vscode.ExtensionContext, configuration: 
                 node instanceof LogGroupNode
                     ? { regionName: node.regionCode, groupName: node.logGroup.logGroupName! }
                     : undefined
-            await tailLogGroup(logGroupInfo)
+            await tailLogGroup(liveTailRegistry, logGroupInfo)
         })
     )
 }
--- a/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
+++ b/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
@@ -115,6 +115,18 @@ async function updateTextDocumentWithNewLogEvents(
     await vscode.workspace.applyEdit(edit)
 }
 
+/**
+ * The LiveTail session should be automatically closed if the user does not have the session's
+ * document in any Tab in their editor.
+ *
+ * `onDidCloseTextDocument` doesn't work for our case because the tailLogGroup command will keep the stream
+ * writing to the doc even when all its tabs/editors are closed, seemingly keeping the doc 'open'.
+ * Also there is no guarantee that this event fires when an editor tab is closed
+ *
+ * `onDidChangeVisibleTextEditors` returns editors that the user can see its contents. An editor that is open, but hidden
+ * from view, will not be returned. Meaning a Tab that is created (shown in top bar), but not open, will not be returned. Even if
+ * the tab isn't visible, we want to continue writing to the doc, and keep the session alive.
+ */
 function registerTabChangeCallback(
     session: LiveTailSession,
     registry: LiveTailSessionRegistry,
--- a/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
+++ b/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
@@ -73,8 +73,8 @@ async function handleSessionStream(
 ) {
     try {
         for await (const event of stream) {
-            if (event.sessionUpdate !== undefined) {
-                const formattedLogEvents = event.sessionUpdate.sessionResults!.map<string>((logEvent) =>
+            if (event.sessionUpdate !== undefined && event.sessionUpdate.sessionResults !== undefined) {
+                const formattedLogEvents = event.sessionUpdate.sessionResults.map<string>((logEvent) =>
                     formatLogEvent(logEvent)
                 )
                 if (formattedLogEvents.length !== 0) {
--- a/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
+++ b/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
@@ -3,17 +3,142 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import * as vscode from 'vscode'
 import { TailLogGroupWizard } from '../wizard/tailLogGroupWizard'
-import { getLogger } from '../../../shared'
 import { CancellationError } from '../../../shared/utilities/timeoutUtils'
+import { LiveTailSession, LiveTailSessionConfiguration } from '../registry/liveTailSession'
+import { LiveTailSessionRegistry } from '../registry/liveTailSessionRegistry'
+import { LiveTailSessionLogEvent, StartLiveTailResponseStream } from '@aws-sdk/client-cloudwatch-logs'
+import { ToolkitError } from '../../../shared'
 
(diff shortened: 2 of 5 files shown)


keeganirby:
Spoke with Josh about this during the initial PR: will add a Changelog when the feature branch is complete and ready to merge to master.

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -3,17 +3,142 @@
...
+
+async function handleSessionStream(
+    stream: AsyncIterable<StartLiveTailResponseStream>,
+    document: vscode.TextDocument,
+    session: LiveTailSession
+) {
+    try {
+        for await (const event of stream) {
+            if (event.sessionUpdate !== undefined) {
+                const formattedLogEvents = event.sessionUpdate.sessionResults!.map<string>((logEvent) =>
```
jpinkney-aws, line 77: It looks like you're using `!` in `sessionResults!` to assert that sessionResults is never undefined. Is that actually the case? Technically if sessionResults returns undefined one time then it results in an error

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -3,17 +3,142 @@
...
+    )
+    await vscode.workspace.applyEdit(edit)
+}
+
+function registerTabChangeCallback(
+    session: LiveTailSession,
+    registry: LiveTailSessionRegistry,
+    document: vscode.TextDocument
+) {
+    vscode.window.tabGroups.onDidChangeTabs((tabEvent) => {
```
jpinkney-aws, line 123: If i'm understanding this correctly I think you might want vscode.workspace.textDocuments instead. I've never actually seen vscode.window.tabGroups before :thinking:. The process would be more or less the same expect instead of onDidChangeTabs and interating over every tab group you can just. Something like:

```
vscode.workspace.onDidChangeTextDocument((textDocument) => {
    if (textDocument.document.uri.toString() === liveTailSession.uri.toString()) {

    }
})
```

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
keeganirby: You're right! I think I had this for some early testing and never when back. I'll add `event.sessionUpdate.sessionResults !== undefined` to the if check.

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
keeganirby: What we are trying to achieve is that if a user doesn't have the LiveTail session open in any tab in their editor, that we close the session.

What I've learned is that the underlying TextDocument can remain open, even if its editor tabs are closed. It seems that the LiveTail command will keep the doucment open even if all its editors are closed because the command is still writing from the stream to the document. https://code.visualstudio.com/api/references/vscode-api#workspace.onDidCloseTextDocument. This suggests looking at onDidChangeVisibleTextEditors, but that event fires even when swapping tabs, not closing. https://code.visualstudio.com/api/references/vscode-api#window.onDidChangeVisibleTextEditors

This implementation is admittedly heavy, but has been the only thing I've found that fits what I'm trying to accomplish.

jpinkney-aws (approved):

packages/core/src/awsService/cloudWatchLogs/activation.ts
```diff
@@ -4,7 +4,7 @@
  */
 
 import * as vscode from 'vscode'
-import { CLOUDWATCH_LOGS_SCHEME } from '../../shared/constants'
+import { CLOUDWATCH_LOGS_LIVETAIL_SCHEME, CLOUDWATCH_LOGS_SCHEME } from '../../shared/constants'
```
justinmk3, line 7: please rename `CLOUDWATCH_LOGS_LIVETAIL_SCHEME` to conform to the lint rule. `CLOUDWATCH_LOGS_SCHEME` was marked as "ignore lint rule" as a temporary workaround, not something to continue when new symbols are introduced.

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -3,17 +3,154 @@
...
+                const formattedLogEvents = event.sessionUpdate.sessionResults.map<string>((logEvent) =>
+                    formatLogEvent(logEvent)
+                )
+                if (formattedLogEvents.length !== 0) {
+                    await updateTextDocumentWithNewLogEvents(formattedLogEvents, document, session.maxLines)
+                }
+            }
+        }
+    } catch (err) {
+        throw new ToolkitError('Caught on-stream exception')
```
justinmk3, line 86: what is the point of this? the  exception that was caught likely has more useful information.

please review *all* of the code guidelines: https://github.com/aws/aws-toolkit-vscode/blob/master/docs/CODE_GUIDELINES.md#exceptions

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -3,17 +3,154 @@
...
+    }
+}
+
+function formatLogEvent(logEvent: LiveTailSessionLogEvent): string {
+    if (!logEvent.timestamp || !logEvent.message) {
+        return ''
+    }
+    const timestamp = new Date(logEvent.timestamp).toLocaleTimeString('en', {
```
justinmk3, line 94: use one of the existing shared datetime format utils. that ensures that we use consistent formatting everywhere in the extension: https://github.com/aws/aws-toolkit-vscode/blob/6fa3a9e554f4aa1916d01d00ea93497ca7205cc5/packages/core/src/shared/utilities/textUtilities.ts#L284-L303

packages/core/src/test/awsService/cloudWatchLogs/commands/tailLogGroup.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    })
+
+    afterEach(function () {
+        sandbox.restore()
+    })
+
+    it('starts LiveTailSession and writes to document. Closes tab and asserts session gets closed.', async function () {
+        wizardSpy = sandbox.stub(TailLogGroupWizard.prototype, 'run').callsFake(async function () {
+            return getTestWizardResponse()
+        })
```
justinmk3, line 44: we have a full wizard testing framework that allows the actual wizard code to run, and you can assert its state. please do not mock/spy things unless it's unavoidable (typically network/service calls).

packages/core/src/awsService/cloudWatchLogs/activation.ts
keeganirby: I will rename, didn't fully realize the linter suppression,  was just following precedent. I can update the variable naming. 

The reason I added a new scheme is because I am not leveraging the existing document and code lens providers. The logic for fetching logs, and the supplied codeLenses for LiveTail are different than those in the existing providers. I didn't know query param would be an option, but since the providers are tied to the scheme itself, I'm not sure if a query param does what I was intending for by having separate providers. Let me know if that can be done!

I could use a query param and a shared scheme, but then I assume that each time one of the providers run, we'd need to inspect the URI and have a branching logic to handle a LiveTail case or not. In that case I'd prefer to just handle these cases in separate classes.

packages/core/src/test/awsService/cloudWatchLogs/commands/tailLogGroup.test.ts
keeganirby: I have the TailLogGroup wizard under test in its own Test class https://github.com/aws/aws-toolkit-vscode/blob/feature/cwltail/packages/core/src/test/awsService/cloudWatchLogs/wizard/tailLogGroupWizard.test.ts
which is why i decided to Mock it here. This `tailLogGroup.test` was intended to test just the functionality of the command itself. 

In that case is it acceptable to leave this mocked? I can use that testing framework here also, was just trying to separate responsibilities of each testing class.

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
keeganirby: ack, will adress.

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
keeganirby: Ack, will address.

packages/core/src/awsService/cloudWatchLogs/activation.ts
justinmk3: >  I am not leveraging the existing document and code lens providers

I believe the existing abstraction is generalized to allow different "providers" of CWL events. So it could be re-used. But it may be a judgement call about whether that saves or increases complexity.

packages/core/src/test/awsService/cloudWatchLogs/commands/tailLogGroup.test.ts
justinmk3: > was just trying to separate responsibilities of each testing class.

Testing more things in one test can often be a benefit (less code, more coverage). Would it make sense to merge this test into the other? 

Or meanwhile, people tend to copy what exists. So the presence of mocks here encourages even more mocks...

keeganirby: Sure, at the least I can copy over a basic form of the Wizard tester here to avoid mocking it. If I have a wizard tester in my test, how would you recommend supplying that to my Command? Doing local testing and the prompter still comes up when the test runs, waiting for input. 

Is it recommended to use something like dependency injection to provide a TailLogGroupWizard to the command in its constructor so the test can supply its Wizard to the command? 

I've added 
``` 
const wizard = new TailLogGroupWizard({
            groupName: testLogGroup,
            regionName: testRegion,
        })
        const tester = await createWizardTester(wizard)
        tester.regionLogGroupSubmenuResponse.assertDoesNotShow()
        tester.logStreamFilter.assertShowFirst()
        tester.filterPattern.assertShowSecond()
```
to my test before executing the command `await tailLogGroup(registry, {
            groupName: testLogGroup,
            regionName: testRegion,
        })`, but its hanging waiting for input on the Wizard because the command is creating a new Wizard and calling prompt on it.

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
keeganirby: Circling back to this, when I demoed LiveTail to our CWL product manager, it was requested that we shorten the time we display to just HH:MM:SS. The reason being that these events are coming in live, so the additional context of Date and timezone weren't seen as important. 

In that case, is this a required change to make?

packages/core/src/test/awsService/cloudWatchLogs/commands/tailLogGroup.test.ts
justinmk3: > but its hanging waiting for input on the Wizard because the command is creating a new Wizard and calling prompt on it

the wizard tester has functions to choose items from it. look at examples in other tests

PR #5791, opened 2024-10-16
fix(credentials): fix credential test so it won't remove real aws config

roger-zhangg:
## Problem
https://github.com/aws/aws-toolkit-vscode/blob/0164d4145e58ae036ddf3815455ea12a159d491d/packages/core/src/test/shared/credentials/userCredentialsUtils.test.ts#L38-L39

This test is deleting my actual aws config file

## Solution
Update the environment variable to point to a fake location before test
Then recover them after test

---


hayemaxi:
Do we even need these delete calls? All the tests in this file stub process.env and use a tempfolder anyway.

PR #5792, opened 2024-10-16
feat(time): Add Interval class

nkomonen-amazon:
Creates our own version of an Interval, similar to how we have our own Timeout

---

### DIFF
--- a/packages/core/src/shared/utilities/timeoutUtils.ts
+++ b/packages/core/src/shared/utilities/timeoutUtils.ts
@@ -184,6 +184,34 @@ export class Timeout {
     }
 }
 
+export class Interval {
+    private _setCompleted: (() => void) | undefined
+    private _nextCompletion: Promise<void>
+    private ref: NodeJS.Timer | number | undefined
+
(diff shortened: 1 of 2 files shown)


nkomonen-amazon:
@hayemaxi it behaves the same, but gives us some extra control and allows us to add features to it.

An example is this `nextCompletion()` method which is not available ootb with `globals.clock.setInterval`.

It can also be thought of as how we wrap globalState or setContext

hayemaxi (approved):

justinmk3:
This new Interval class doesn't have a docstring, so I assume the only difference vs Timeout is that it repeats. Why can't that simply be a feature added to Timeout (if it doesn't already exist)?

nkomonen-amazon:
@justinmk3 most of the methods in Timeout dont make sense in the context of an Interval:

- remainingTime, intervals run infinitely
- completed, intervals run infinitely
- refresh, intervals automatically refresh
- promisify, you don't wait for an interval to complete
- cancellationToken, it doesn't timeout by default

Over time if we see a use for something like `elapsedTime`, we could look to generalize that between Timeout and Interval

PR #5801, opened 2024-10-16
deps(types-node): avoid formal dependency on types-node22

Hweinstock:
## Problem
Follow up to: https://github.com/aws/aws-toolkit-vscode/pull/5761#discussion_r1803806543

## Solution
- remove dependency on `@types/node@22` using `any` workaround.
- add a tech debt test that reminds us to remove `any` workaround once we bump to node 18 in CI. 

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
         "webpack-merge": "^5.10.0"
     },
     "dependencies": {
+        "@types/node": "^18.19.55",
         "vscode-nls": "^5.2.0",
         "vscode-nls-dev": "^4.0.4"
     }
--- a/package.json
+++ b/package.json
@@ -69,7 +69,6 @@
         "webpack-merge": "^5.10.0"
     },
     "dependencies": {
-        "@types/node": "^22.7.5",
         "vscode-nls": "^5.2.0",
         "vscode-nls-dev": "^4.0.4"
     }
--- a/packages/core/src/test/techdebt.test.ts
+++ b/packages/core/src/test/techdebt.test.ts
@@ -35,6 +35,11 @@ describe('tech debt', function () {
             semver.lt(minNodejs, '18.0.0'),
             'with node16+, we can now use AbortController to cancel Node things (child processes, HTTP requests, etc.)'
         )
+
+        assert.ok(
+            semver.lt(minNodejs, '18.0.0'),
+            'with node16+, we can remove `any` workaround in the `copyFiles.ts` scripts'
+        )
     })
 
     it('remove separate sessions login edge cases', async function () {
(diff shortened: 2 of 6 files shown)


packages/core/src/test/techdebt.test.ts
```diff
@@ -35,6 +35,11 @@ describe('tech debt', function () {
             semver.lt(minNodejs, '18.0.0'),
             'with node16+, we can now use AbortController to cancel Node things (child processes, HTTP requests, etc.)'
         )
+
+        assert.ok(
+            semver.lt(minNodejs, '18.0.0'),
+            'with node16+, we can remove `any` workaround in the `copyFiles.ts` scripts'
```
justinmk3, line 41:
```suggestion
            'with node18+, we can use `fs.cpSync` (without casting to `any`) in the `copyFiles.ts` scripts'
```

package.json
```diff
@@ -69,7 +69,6 @@
         "webpack-merge": "^5.10.0"
     },
     "dependencies": {
-        "@types/node": "^22.7.5",
```
justinmk3, line 72: Does it work if this is 18 instead of 22? 

I'm confused about the relationship of the copyFiles.ts scripts vs the packages/x/package.json declarations.

As an alternative to `fs as any`, maybe we could use the `techdebt.test.ts` reminder to remind us to remove this top-level dependency when the `getMinNodejsVersion` result is new enough?

Hweinstock: Looks like that works as well, I can update the techdebt test to match. I thought avoiding the dependency was more important than the `any` workaround. But I guess the version bump isn't something to worry about?

justinmk3: > I thought avoiding the dependency was more important than the `any` workaround.

Since this PR makes the total cost clearer (i.e.: small :), I'm having doubts. And also, since the risk is covered by techdebt.test.ts in either case, perhaps it's better to have type checking.

Either way is fine IMO.

packages/core/src/test/techdebt.test.ts
```diff
@@ -35,6 +35,11 @@ describe('tech debt', function () {
             semver.lt(minNodejs, '18.0.0'),
             'with node16+, we can now use AbortController to cancel Node things (child processes, HTTP requests, etc.)'
         )
+
+        assert.ok(
+            semver.lt(minNodejs, '18.0.0'),
+            'with node18+, we can use `fs.cpSync` (without casting to `any`) in the `copyFiles.ts` scripts'
```
justinmk3, line 41: if you decide to keep the dep, then this message should say "remove the dep" :)

package.json
Hweinstock: I think I'll keep the dependency then. Since its just a version bum (to an old stable version) and not a whole new thing, the cost seems very low. The `any` code doesn't read well, and could cause confusion to someone without the necessary context.

PR #5805, opened 2024-10-17
feat(amazonq): added mynah updated beta package for feature dev

32teeth:
## Solution

Update to mynah package to enable focus tab from notification

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5229,9 +5229,10 @@
             }
         },
         "node_modules/@aws/mynah-ui": {
-            "version": "4.15.11",
+            "version": "4.16.1-beta.1",
+            "resolved": "https://registry.npmjs.org/@aws/mynah-ui/-/mynah-ui-4.16.1-beta.1.tgz",
+            "integrity": "sha512-/eLp7zEicZGe3FeRX1ArEsWH1QV7rygjgf8yglo/3n0k6yWETVUTIWDDdY80yL/ANtuGN3FH/ftO336aJ5I8jg==",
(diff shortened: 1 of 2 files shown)


jpinkney-aws:
Can you explain more about why we would move to a beta dependency?

32teeth:
> Can you explain more about why we would move to a beta dependency?

@dogusata has implemented a feature to be able to select a tab by ID without passing an eventID in the NPM release

> I've released a beta version which removes the eventId check from the selectTab method.
> Please use version number as https://www.npmjs.com/package/@aws/mynah-ui/v/4.16.1-beta.1
> When there is a new production release, it will contain this change. (Will be 4.16.1 or above)

32teeth:
I will amend the commit message this evening

justinmk3:
It looks like this may have a customer-facing impact that can be described in the changelog. The changelog is presented to users by VSCode and the marketplace. It is a "micro-blog" for advertising improvements to users. It is the primary way of communicating changes to customers.

1. Review the [changelog guidelines](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#changelog).
2. Add a new changelog entry: `npm run newChange -w packages/amazonq` or `npm run newChange -w packages/toolkit`

jpinkney-aws:
FWIW 4.16.1 (the newest mynah ui version) has just been released so I don't think the beta version is relevant anymore

32teeth:
Thanks. Closing out

PR #5810, opened 2024-10-17
deps(ci node): update node in ci to 18.x [WIP]

Hweinstock:
## Problem
There's no real reason we need to use node 16 in CI, and it's preventing us from using some features.

Our tests run against vscode's own version of node, which is unrelated to the node we use to run scripts, etc. (See also: https://github.com/aws/aws-toolkit-vscode/pull/5761 )

## Solution
update CI runners to node 18 instead of 16. 
We need to pass ` { shell: true } ` to avoid error on windows. see more here: https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2

---


Hweinstock:
Tried bumping the vscode/electron-test version due to failing windows tests, but perhaps its unrelated. I'll continue to investigate.

packages/core/scripts/test/launchTestUtilities.ts
```diff
@@ -189,6 +189,7 @@ async function invokeVSCodeCli(vsCodeExecutablePath: string, args: string[]): Pr
 
     if (spawnResult.status !== 0) {
         console.log('output: %s', spawnResult.output)
+        console.log('error: %s ', spawnResult.error)
```
justinmk3, line 192: FYI: `%O` gives the full object details in a nice, readable format

```suggestion
        console.log('error: %O', spawnResult.error)
```

justinmk3 (approved):
Nice!

> We need to pass `{ shell: true }` to avoid error on windows. see more here: https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2

related: https://github.com/aws/aws-toolkit-vscode/issues/5292

justinmk3:
> We need to pass `{ shell: true }` to avoid error on windows. see more here: https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2

👍  for reference: https://github.com/aws/aws-toolkit-vscode/issues/5292

PR #5812, opened 2024-10-18
feat(amazon q): display multiple diff patches

ntarakad-aws:
adding functionality for selective transformation to allow for java 17 as source and display multiple diff patches for the user to accept/reject


package-lock.json
dhasani23, line 1: Should not commit this file to the PR

packages/amazonq/test/unit/amazonqGumby/transformationResultsHandler.test.ts
```diff
@@ -10,11 +10,21 @@ import path from 'path'
...
+}
+
 describe('DiffModel', function () {
     afterEach(() => {
         sinon.restore()
     })
 
+    const testDescription =
+        '{"name": "Added file", "fileName": "resources/files/addedFile.diffs", "isSuccessful": true}'
+    const parsedTestDescription: PatchDescription = JSON.parse(testDescription)
```
dhasani23, line 26: Can put these 2 above in a `beforeEach` block above since we use both in both test cases

packages/amazonq/test/unit/amazonqGumby/transformationResultsHandler.test.ts
```diff
@@ -10,11 +10,21 @@ import path from 'path'
 import { getTestResourceFilePath } from './amazonQGumbyUtil'
 import { fs } from 'aws-core-vscode/shared'
 
+type PatchDescription = {
```
dhasani23, line 13: Should go in: https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/models/model.ts

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -184,6 +184,10 @@ export class Messenger {
                     value: JDKVersion.JDK11,
                     label: JDKVersion.JDK11.toString(),
                 },
+                {
+                    value: JDKVersion.JDK17,
+                    label: JDKVersion.JDK17.toString(),
```
dhasani23, line 189: For this and the 2 above, don't need to use `toString()`

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -3,24 +3,30 @@
...
 import { telemetry } from '../../../shared/telemetry/telemetry'
 import { CodeTransformTelemetryState } from '../../../amazonqGumby/telemetry/codeTransformTelemetryState'
 import { MetadataResult } from '../../../shared/telemetry/telemetryClient'
 import * as CodeWhispererConstants from '../../models/constants'
-import { createCodeWhispererChatStreamingClient } from '../../../shared/clients/codewhispererChatClient'
+// import { createCodeWhispererChatStreamingClient } from '../../../shared/clients/codewhispererChatClient'
 import { ChatSessionManager } from '../../../amazonqGumby/chat/storages/chatSession'
 import { setContext } from '../../../shared/vscode/setContext'
 
+type PatchDescription = {
```
dhasani23, line 24: We can just define this once in https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/models/model.ts

dhasani23: Also maybe rename to `PatchInfo` to be more generic

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -195,36 +215,39 @@ export class DiffModel {
...
     public rejectChanges() {
         this.clearChanges()
     }
 
     public clearChanges() {
-        this.changes = []
+        this.patchFileNodes = []
+        // transformByQState.setSummaryFilePath('')
+        // transformByQState.setProjectCopyFilePath('')
+        // transformByQState.setResultArchiveFilePath('')
```
dhasani23, line 246: These 3 commented out lines should already be done in a clean-up method we have in `startTransformByQ.ts`

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -266,6 +312,27 @@ export class ProposedTransformationExplorer {
...
 
+        const pathContainingArchive = '/private/var/folders/mn/l6c4t6sd1jn7g4p4nb6wqhh80000gq/T/ExportResultArchive'
+
+        const patchFiles: string[] = []
+        fs.readdir(path.join(pathContainingArchive, 'patch'), (err, files) => {
+            if (err) {
+                getLogger().error(err)
+            } else {
+                files.forEach((file) => {
+                    const filePath = path.join(path.join(pathContainingArchive, 'patch'), file)
```
dhasani23, line 323: Can combine as just 1 `path.join()`

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -362,20 +429,19 @@ export class ProposedTransformationExplorer {
...
-                diffModel.parseDiff(
-                    path.join(pathContainingArchive, ExportResultArchiveStructure.PathToDiffPatch),
-                    transformByQState.getProjectPath()
-                )
+                // pathContainingArchive = path.dirname(pathToArchive)
+                // const zip = new AdmZip(pathToArchive)
+                // zip.extractAllTo(pathContainingArchive)
+
+                getLogger().info('descriptions', patchFilesDescriptions.toString())
+                diffModel.parseDiff(patchFiles[0], transformByQState.getProjectPath(), patchFilesDescriptions[0])
```
dhasani23, line 443: Probably add a comment saying multiple patches are returned, once the ZIP is downloaded show the first one to start

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -266,6 +312,27 @@ export class ProposedTransformationExplorer {
...
+                        patchFiles.push(filePath)
+                    }
+                })
+            }
+        })
+
+        const pathContainingPatchFileDescriptions = path.join(pathContainingArchive, 'patch', 'diff.json')
+
+        const jsonData = fs.readFileSync(pathContainingPatchFileDescriptions, 'utf-8')
+        const patchFilesDescriptions: PatchDescription[] = JSON.parse(jsonData)
```
dhasani23, line 334: Once we get the E2E working, this whole block of code will probably move to somewhere after the download happens instead of in this constructor

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -408,11 +474,24 @@ export class ProposedTransformationExplorer {
             diffModel.saveChanges()
             telemetry.ui_click.emit({ elementId: 'transformationHub_acceptChanges' })
             void vscode.window.showInformationMessage(CodeWhispererConstants.changesAppliedNotification)
-            transformByQState.getChatControllers()?.transformationFinished.fire({
-                message: CodeWhispererConstants.changesAppliedChatMessage,
-                tabID: ChatSessionManager.Instance.getSession().tabID,
-            })
-            await reset()
+            if (!diffModel.currentPatchIndex) {
```
dhasani23, line 477: Would prefer `if(diffModel.currentPatchIndex === 0)` plus a comment stating that we do this so that the changesAppliedChatMessage is only sent to user when they accept the first `diff.patch`

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -427,6 +506,7 @@ export class ProposedTransformationExplorer {
 
         vscode.commands.registerCommand('aws.amazonq.transformationHub.reviewChanges.rejectChanges', async () => {
             diffModel.rejectChanges()
+            diffModel.currentPatchIndex = 0
```
dhasani23, line 509: Is this strictly necessary? If so, can add this to the `rejectChanges` function, otherwise remove

dhasani23: `DiffModel` defaults the `currentPatchIndex` to `0`, but to test, you can remove this line and run 2 consecutive transformations, accepting maybe 3/5 of the patch files from the first transformation, and make sure that on the second transformation, this `currentPatchIndex` was properly reset to 0 and all 5 patch files are working fine

dhasani23: In general it's a good idea to run consecutive transformations to make sure everything is working fine

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
dhasani23: It looks like they're already done in the `reset()` function actually, which also calls `clearChanges`

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
ntarakad-aws: Oh ok I was just copying what was done above, are you just saying do "11" or "17?

dhasani23: `JDKVersion.JDK17` is fine, we just don't need the `toString()` call here, and may as well fix it in the 2 places above too

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -346,6 +337,7 @@ export class ProposedTransformationExplorer {
             }
 
             diffModel.clearChanges()
+            diffModel.currentPatchIndex = 0
```
dhasani23, line 340: Now that I see the `clearChanges`, probably makes more sense to actually put this there

packages/amazonq/test/unit/amazonqGumby/transformationResultsHandler.test.ts
```diff
@@ -6,11 +6,19 @@ import assert from 'assert'
...
+import { PatchInfo } from 'aws-core-vscode/codewhisperer'
 import path from 'path'
 import { getTestResourceFilePath } from './amazonQGumbyUtil'
 import { fs } from 'aws-core-vscode/shared'
 
 describe('DiffModel', function () {
+    let parsedTestDescription: PatchInfo
+    beforeEach(() => {
+        parsedTestDescription = JSON.parse(
+            '{"name": "Added file", "fileName": "resources/files/addedFile.diffs", "isSuccessful": true}'
```
dhasani23, line 18: Typo with `diffs`? You can also run the test locally by going to the Run & Debug menu and instead click on "Run Extension Tests (current file)" with this file open to make sure they pass

packages/amazonq/test/unit/amazonqGumby/transformationResultsHandler.test.ts
```diff
@@ -6,11 +6,19 @@ import assert from 'assert'
 import sinon from 'sinon'
 import os from 'os'
 import { DiffModel, AddedChangeNode, ModifiedChangeNode } from 'aws-core-vscode/codewhisperer/node'
+import { PatchInfo } from 'aws-core-vscode/codewhisperer'
```
dhasani23, line 9: Should be imported from `model.ts`

ntarakad-aws: Having trouble seeing how to import from model.ts as its in the amazonq test directory instead of core. I was looking at other tests and when importing they seem to import from 'aws-core-vscode/' whether it be from 'test' or 'shared' or 'codewhisperer' etc. I know this might get autogenerated but in that file we do import everything from model. Maybe I am just missing something ...

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -177,7 +195,9 @@ export class DiffModel {
                 }
             },
         })
-        this.changes = changedFiles.flatMap((file) => {
+        const patchFileNode = new PatchFileNode(diffDescription ? diffDescription.name : pathToDiff)
```
dhasani23, line 198: This means that if the JSON file with the patch metadata is not returned in the download ZIP, we will show the path to the `diff.patch` instead of the description, right?

dhasani23: If so we should see what that looks like as the path could be very long, a screenshot would help.

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -378,6 +378,7 @@ export async function zipCode(
         transformByQState.getChatControllers()?.transformationFinished.fire({
             message: CodeWhispererConstants.projectSizeTooLargeChatMessage,
             tabID: ChatSessionManager.Instance.getSession().tabID,
+            includeStartNewTransformationButton: 'true',
```
dhasani23, line 381: Should be a `boolean`

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -454,11 +454,15 @@ export class GumbyController {
         await this.prepareProjectForSubmission(message)
     }
 
-    private transformationFinished(data: { message: string | undefined; tabID: string }) {
+    private transformationFinished(data: {
+        message: string | undefined
+        tabID: string
+        includeStartNewTransformationButton: string
```
dhasani23, line 460: Should be a `boolean`

packages/amazonq/package.json
dhasani23, line 1: I don't think we need to commit this change

packages/amazonq/package.json
```diff
@@ -1095,6 +1095,6 @@
     },
     "engines": {
         "npm": "^10.1.0",
-        "vscode": "^1.83.0"
+        "vscode": "^1.68.0"
```
dhasani23, line 1098: I don't think we need to commit this change

packages/core/src/amazonqGumby/chat/controller/controller.ts
ntarakad-aws: I had originally had it as a boolean but then kept getting errors where the data buffer passed into fire was of the wrong type and received a boolean (false). Unless there was a different way to get around this ...

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
ntarakad-aws: Correct yeah, thats why I do the patchFileNode.label change below because the label just gets the base name ie diff.patch and its the label attribute that is used when displaying the files in the dropdown tree view

dhasani23: But then when there is a `diffDescription`, the `PatchFileNode` constructor will call `path.basename(diffDescription.name)` to set the label? `path.basename()` we should only use for actual paths

ntarakad-aws: Oh good point, let me fix this

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -372,10 +421,39 @@ export class ProposedTransformationExplorer {
...
                 zip.extractAllTo(pathContainingArchive)
+
+                const files = fs.readdirSync(path.join(pathContainingArchive, ExportResultArchiveStructure.PathToPatch))
+                files.forEach((file) => {
+                    const filePath = path.join(pathContainingArchive, ExportResultArchiveStructure.PathToPatch, file)
+                    if (file === 'diff.patch') {
+                        singlePatchFile = filePath
+                    } else if (file.endsWith('.json')) {
+                        const jsonData = fs.readFileSync(filePath, 'utf-8')
+                        patchFilesDescriptions = JSON.parse(jsonData)
```
dhasani23, line 432: Since this is in a `forEach` it will just get overwritten each time right? Better to just explicitly look for the 1 `diff.json` and if present, get the 5 diffs. If not present, use just the 1 combined `diff.patch`

packages/core/src/amazonqGumby/chat/controller/messenger/messengerUtils.ts
```diff
@@ -47,8 +49,6 @@ export default class MessengerUtils {
                 javaHomePrompt += ` ${CodeWhispererConstants.macJavaVersionHomeHelpChatMessage(1.8)}`
             } else if (jdkVersion === JDKVersion.JDK11) {
                 javaHomePrompt += ` ${CodeWhispererConstants.macJavaVersionHomeHelpChatMessage(11)}`
-            } else if (jdkVersion === JDKVersion.JDK17) {
-                javaHomePrompt += ` ${CodeWhispererConstants.macJavaVersionHomeHelpChatMessage(17)}`
```
dhasani23, line 51: Why remove this JDK 17 stuff?

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -185,15 +224,11 @@ export class Messenger {
             options: [
                 {
                     value: JDKVersion.JDK8,
-                    label: JDKVersion.JDK8,
+                    label: JDKVersion.JDK8.toString(),
```
dhasani23, line 227: Don't need `.toString()` here

dhasani23 (approved):

dhasani23 (approved):

packages/core/src/codewhisperer/models/constants.ts
```diff
@@ -583,13 +614,27 @@ export const jobCancelledChatMessage =
...
 
-export const jobCompletedChatMessage = `I transformed your code. You can review the diff to see my proposed changes and accept or reject them. The transformation summary has details about the files I updated.`
+export const diffMessage = (multipleDiffs: boolean) => {
+    return multipleDiffs
+        ? 'You can review the diff to see my proposed changes and accept or reject them. If you reject the diff, you will not be able to see the diffs later.'
+        : 'You can review the diff to see my proposed changes and accept or reject them.'
+}
+
+export const jobCompletedChatMessage = (multipleDiffsString: string) => {
+    return `I upgraded your code. ${multipleDiffsString} The transformation summary has details about the files I updated.`
```
dhasani23, line 624: nit: here and in the 3 spots below, use `transformed` instead of `upgraded`

packages/amazonq/test/unit/amazonqGumby/transformationResultsHandler.test.ts
```diff
@@ -28,34 +37,80 @@ describe('DiffModel', function () {
...
+    })
+
+    it('WHEN parsing a diff patch where diff.json is not present and a file was modified THEN returns an array representing the modified file', async function () {
+        const testDiffModel = new DiffModel() //try to see if you can reuse that test model from above
+
+        const fileAmount = 1
+        const workspaceFolder = await createTestWorkspace(fileAmount, { fileContent: '' })
 
-        await fs.delete(path.join(workspacePath, 'README.md'), { recursive: true })
+        sinon.replace(fs, 'exists', async (path) => true)
```
nkomonen-amazon, line 93: What is the reason for this? Can the file be manually created instead?

packages/core/src/codewhisperer/models/constants.ts
```diff
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { DescriptionContent } from '..'
```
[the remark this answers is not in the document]
ntarakad-aws: Sounds good, I can just perform the logic based on totalPatchFiles which will allow me to remove the need of this import.

packages/amazonq/test/unit/amazonqGumby/transformationResultsHandler.test.ts
```diff
@@ -4,13 +4,22 @@
...
 import { getTestResourceFilePath } from './amazonQGumbyUtil'
 import { fs } from 'aws-core-vscode/shared'
+import { createTestWorkspace } from 'aws-core-vscode/test'
 
 describe('DiffModel', function () {
+    let parsedTestDescriptions: DescriptionContent
+    beforeEach(() => {
+        const fs = require('fs')
+        parsedTestDescriptions = JSON.parse(
+            fs.readFileSync(getTestResourceFilePath('resources/files/diff.json'), 'utf-8')
```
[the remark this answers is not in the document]
ntarakad-aws: Makes sense, will update

packages/amazonq/test/unit/amazonqGumby/transformationResultsHandler.test.ts
ntarakad-aws: Now realizing we are just parsing a diff from the test resources directory and making sure it gets identified as the right type of node with the right label. I don't think we need to create the README.md, will fix.

packages/amazonq/.changes/next-release/Feature-735529a9-8424-4f78-9989-58ec972af100.json
dhasani23, line 1: deleted changelog because we are going to merge this under a feature flag for now

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -401,6 +408,20 @@ export class GumbyController {
             result: MetadataResult.Pass,
         })
         this.messenger.sendSkipTestsSelectionMessage(skipTestsSelection, message.tabID)
+        if (!isSelectiveTransformationReady) {
```
dhasani23, line 411: note: this is just a feature flag, waiting for our backend work to be ready, will merge with this flag until then

packages/amazonq/test/unit/amazonqGumby/transformationResultsHandler.test.ts
ntarakad-aws: After testing, it seems we do require this information.

dhasani23 (approved):

PR #5813, opened 2024-10-18
ci: don't output code coverage table in logs

hayemaxi:
Problem: c8 was not given a reporter, therefore it defaulted to `text` which is the terminal.
Solution: Set it to `lcov`, which is what we have been using previously.

Note: More work may be needed on the .c8rc.json files. We might not need the one for core/ anymore. They may not be configured in the optimal way. Is it behaving how we want it to?

---


packages/toolkit/package.json
```diff
@@ -69,9 +69,9 @@
         "webCompile": "npm run clean && npm run buildScripts && webpack --config-name web",
         "webRun": "npx @vscode/test-web --open-devtools --browserOption=--disable-web-security --waitForDebugger=9222 --extensionDevelopmentPath=. .",
         "testCompile": "npm run clean && npm run buildScripts && npm run compileOnly",
-        "test": "npm run testCompile && c8 --allowExternal ts-node ../core/scripts/test/launchTest.ts unit dist/test/unit/index.js ../core/dist/src/testFixtures/workspaceFolder",
-        "testE2E": "npm run testCompile && c8 --allowExternal ts-node ../core/scripts/test/launchTest.ts e2e dist/test/e2e/index.js ../core/dist/src/testFixtures/workspaceFolder",
-        "testInteg": "npm run testCompile && c8 --allowExternal ts-node ../core/scripts/test/launchTest.ts integration dist/test/integ/index.js ../core/dist/src/testFixtures/workspaceFolder",
+        "test": "npm run testCompile && c8 --reporter lcov --allowExternal ts-node ../core/scripts/test/launchTest.ts unit dist/test/unit/index.js ../core/dist/src/testFixtures/workspaceFolder",
```
jpinkney-aws, line 72: do we need to manually add the flag if its already in the .c8rc.json? The only reason we had to do that for allowExternal was because it wasn't exposed in the c8rc

jpinkney-aws (commented):
Do we run any tests at all from core anymore? If not, we would probably also need to update other codecov paths as well

packages/toolkit/package.json
hayemaxi: I thought I had seen that amazonq also output the table, but it doesn't. It seems to work just fine with the c8rc. Thanks

hayemaxi:
> Do we run any tests at all from core anymore? If not, we would probably also need to update other codecov paths as well

Which paths need updated? The only thing I can think of is deleting core's c8rc. Not sure if there would be side effects from this or if anything else needs updating. I figured I would leave this investigation for future work.

jpinkney-aws (approved):
I think the only other thing we would need to do is:
1. delete core's .c8rc.json
2. remove ./coverage/core/lcov.info instances in the ci (both github and codebuilds): 

It looks like we future proofed ourselves and already upload toolkits lcov.info if it exists: https://github.com/aws/aws-toolkit-vscode/blob/master/.github/workflows/node.js.yml#L103

justinmk3 (approved):

hayemaxi:
I will create an AI to address these

PR #5814, opened 2024-10-18
refactor(ec2): enforce ssh key inside global storage. 

Hweinstock:
## Problem
Followup from https://github.com/aws/aws-toolkit-vscode/pull/5642#discussion_r1771817153

There is a case where we could be overwriting keys/files in the `.ssh/` directory if not careful. 

## Solution
Don't allows keys to generated in a directory we do not control, and therefore avoid possibility of overwriting user ssh keys. 

---


packages/core/src/awsService/ec2/sshKeyPair.ts
```diff
@@ -28,10 +29,25 @@ export class SshKeyPair {
     }
 
     public static async getSshKeyPair(keyPath: string, lifetime: number) {
+        SshKeyPair.assertValidKeypath(
```
justinmk3, line 32: It looks like `SshKeyPair` constructor currently accepts a path controlled by the caller. If that option could be removed, so that `SshKeyPair` is fully in control, would that lead to a simpler design and possibly less need for validation? Just a thought

packages/core/src/awsService/ec2/sshKeyPair.ts
```diff
@@ -28,10 +29,25 @@ export class SshKeyPair {
...
+            keyPath,
+            `ec2: unable to generate key outside of global storage in path ${keyPath}`
+        )
         await SshKeyPair.generateSshKeyPair(keyPath)
         return new SshKeyPair(keyPath, lifetime)
     }
 
+    private static isValidKeyPath(keyPath: string): boolean {
+        const relative = path.relative(globals.context.globalStorageUri.fsPath, keyPath)
+        return relative !== undefined && !relative.startsWith('..') && !path.isAbsolute(relative)
```
justinmk3, line 42: Something like `keyPath.length > 4` also is a simple but effective way to avoid accidental deletion, especially when deleting a directory whose path would never be shorter than, say, 10 chars.

justinmk3 (approved):

packages/core/src/test/awsService/ec2/model.test.ts
```diff
@@ -132,17 +134,19 @@ describe('Ec2ConnectClient', function () {
...
-            sinon.stub(SshKeyPair, 'generateSshKeyPair')
-            sinon.stub(SshKeyPair.prototype, 'getPublicKey').resolves('test-key')
-
             const testSelection = {
                 instanceId: 'test-id',
                 region: 'test-region',
             }
-            const mockKeys = await SshKeyPair.getSshKeyPair('fakeDir', 30000)
-            await client.sendSshKeyToInstance(testSelection, mockKeys, 'test-user')
+            const temporaryDirectory = path.join(globals.context.globalStorageUri.fsPath, 'ModelTests')
```
nkomonen-amazon, line 141: See if `testUtil.ts#TestFolder` can be used instead. We are slowly trying to move to a single way to make a temp folder for tests. It is self cleaning as well.

packages/core/src/awsService/ec2/sshKeyPair.ts
nkomonen-amazon: +1. This looks like the values from `configureSshKeys()` could be set as defaults here.

Something like this, but `globals` may be undefined since it may be referenced before it is defined:
```
public static async getSshKeyPair(keyPath = path.join(globals.context.globalStorageUri.fsPath, `aws-ec2-key`), ...)
```

If you get issues, then you'll probably need to make it a function
```
public static async getSshKeyPair(getKeyPath = () => path.join(globals.context.globalStorageUri.fsPath, `aws-ec2-key`), ...)
```

packages/core/src/test/awsService/ec2/model.test.ts
nkomonen-amazon, line 141: In tests, `globalStorageUri` is stubbed with a temp dir in `globalSetup.test.ts`, so I think this may not be needed

But there is a chance that it is only cleared on the "after" instead of "afterEach", and may cause issues in tests: https://github.com/aws/aws-toolkit-vscode/blob/17e951cc594c580cc28df66053c344520f73fdee/packages/core/src/test/globalSetup.test.ts#L77

So maybe you could add:
```ts
await fs.delete(globals.context.globalStorageUri.fsPath, { recursive: true, force: true })
```
around here:
https://github.com/aws/aws-toolkit-vscode/blob/17e951cc594c580cc28df66053c344520f73fdee/packages/core/src/test/globalSetup.test.ts#L105
which I think we should have been doing in the first place

packages/core/src/awsService/ec2/sshKeyPair.ts
Hweinstock: yeah, that makes sense. If the key class controls the key path then an illegal path isn't possible, avoiding the unnecessary checks.

packages/core/src/test/awsService/ec2/model.test.ts
Hweinstock: If I move the temp keys into the globalStorage itself, then is there any reason I need this to clear on `afterEach`? I would think deleting the temp dir on `after` should be sufficient.

nkomonen-amazon: If we do not clear globalStorageUri after each test we'll probably have a collision somewhere later.

Lets say you had a test that created an ssh key, then the following test assumed that since it is a new test the globalStorage will not have a key by default. The initial test state bled over in to the next and left some unexpected artifacts

justinmk3 (approved):

packages/core/src/test/awsService/ec2/model.test.ts
Hweinstock: I see. Was there any reason we didn't want to do this before? Is it expensive/slow to delete all its contents for every test?

justinmk3: No, we haven't used it much in tests before, so it was just an oversight.

packages/core/src/test/globalSetup.test.ts
```diff
@@ -101,10 +101,13 @@ export const mochaHooks = {
         globals.telemetry.clearRecords()
         globals.telemetry.logger.clear()
         TelemetryDebounceInfo.instance.clear()
+
         // mochaGlobalSetup() set this to a fake, so it's safe to clear it here.
         await globals.globalState.clear()
 
         await testUtil.closeAllEditors()
+        await fs.delete(globals.context.globalStorageUri.fsPath, { recursive: true, force: true })
```
justinmk3, line 109: This should be fine even for local development, because `mochaGlobalSetup` does:

```ts
fakeContext.globalStorageUri = (await testUtil.createTestWorkspaceFolder('globalStoragePath')).uri
```

justinmk3 (approved):

justinmk3 (approved):

PR #5815, opened 2024-10-19
feat(ui): support multipick prompter

roger-zhangg:
## Problem
Currently prompter doesn't support multipick

## Solution
Add a new function `createMultiPick` that supports multipick and can be used in Wizard.
The returned result will be a list encoded by `JSON.stringify` , use `JSON.parse` to recover the list

## Proposed UX
image
```typescript
// define DataQuickPickItem, note you can use `picked: true` to pre-select items in the multipick
const syncFlagItems: DataQuickPickItem<string>[] = [
    {
        label: 'Build in source',
        data: '--build-in-source',
        description: 'Opts in to build project in the source folder. Only for node apps',
    },
    {
        label: 'Code',
        data: '--code',
        description: 'Sync only code resources (Lambda Functions, API Gateway, Step Functions)',
        picked: true,
    }
]

export interface SyncParams {
    readonly syncFlags: string
}

// define the wizard
export class SyncWizard extends Wizard<SyncParams> {
    public constructor() {
        super()
        this.form.syncFlags.bindPrompter(() => {
            return createMultiPick(syncFlagItems, {
                title: 'Specify parameters for sync',
                placeholder: 'Press enter to proceed with highlighted option',
                buttons: createCommonButtons(samSyncUrl),
            })
        })
    }
}

// run the wizard
result = await new SyncWizard().run()

// Use `JSON.parse` to decode to list -> ['--build-in-source','--code']
console.log(JSON.parse(result.syncFlags))
```

## TODO
Add Tests

---

### DIFF
--- a/packages/core/src/shared/ui/pickerPrompter.ts
+++ b/packages/core/src/shared/ui/pickerPrompter.ts
@@ -241,6 +241,10 @@ export function createLabelQuickPick<T extends string>(
 
 function acceptItems<T>(picker: DataQuickPick<T>, resolve: (items: DataQuickPickItem<T>[]) => void): void {
     if (picker.selectedItems.length === 0) {
+        if (picker.canSelectMany) {
+            // Allow empty choice in multipick
+            resolve(Array.from(picker.selectedItems))
+        }
         return
     }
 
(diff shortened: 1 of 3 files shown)


jpinkney-aws:
The code itself looks fine to me but I think we should really have tests for something like this before we merge it

roger-zhangg:
Hi @jpinkney-aws Please take another look when available. Thanks!

justinmk3 (approved):
Tests are passing and the change is isolated, so we can iterate on this if needed. Meanwhile, it's great to have this in mainline. Thank you!

jpinkney-aws (approved):

justinmk3:
The failing integ tests are perf tests, unrelated to this PR.

PR #5816, opened 2024-10-19
refactor(telemetry): add functionality to @withTelemetryContext

nkomonen-amazon:
See each commit for the specific change. But this has multiple improvements to the `@withTelemetryContext` decorator.

The main change to note:
- Adding the `@withTelemetryContext` decorator to a method can wrap thrown errors with context about that function. This helps us to build some sort of stack trace in the `reasonDesc` of our telemetry when errors are thrown.
  - [This is the documentation updated to get a better idea of what this is](https://github.com/aws/aws-toolkit-vscode/blob/b3f2ad9c65b47cd61ca72652171e932ef2480d8c/docs/telemetry.md#adding-a-stack-trace-to-your-metric) 
- The decorator allows for minimal diffs to the code. It is replacing [this previous code](https://github.com/aws/aws-toolkit-vscode/blob/2a72e6df814e3cbabce1cfe2476f1ca98ea8de2b/packages/core/src/shared/crashMonitoring.ts#L651) which would cause a diff in the column indentation since all the method code needed to be in a callback

---

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -150,15 +150,15 @@ not the path it took to get there. We sometimes need this stack trace to debug, 
 Common example: _"I have a function, `thisFailsSometimes()` that is called in multiple places. The function sometimes fails, I know from telemetry, but I do not know if it is failing when it is a specific caller. If I knew the call stack/trace that it took to call my function that would help me debug."_
 
 ```typescript
-function outerA() {
+function runsSuccessfully() {
     thisFailsSometimes(1) // this succeeds
 }
 
-function outerB() {
+function thisThrows() {
     thisFailsSometimes(0) // this fails
 }
 
-function thisFailsSometimes(num: number) {
+function failsDependingOnInput(num: number) {
     return telemetry.my_Metric.run(() => {
         if (number === 0) {
             throw Error('Cannot be 0')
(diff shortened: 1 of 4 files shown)


docs/telemetry.md
```diff
@@ -167,35 +170,67 @@ function thisFailsSometimes(num: number) {
 
 ### Solution
 
-Add a value to `function` in the options of a `run()`. This will result in a stack of functions identifiers that were previously called
-before `thisFailsSometimes()` was run. You can then retrieve the stack in the `run()` of your final metric using `getFunctionStack()`.
+On class methods use the `@withTelemetryContext()` decorator add context to the execution. Depending on the args set, it provides features like emitting the result, or adding it's context to errors.
```
jpinkney-aws, line 173: should this be
```suggestion
On class methods use the `@withTelemetryContext()` decorator to add context to the execution. Depending on the args set, it provides features like emitting the result, or adding it's context to errors.
```

packages/toolkit/package.json
```diff
@@ -3716,271 +3716,278 @@
                     "fontCharacter": "\\f1b4"
                 }
             },
-            "aws-amazonq-transform-logo": {
+            "aws-amazonq-transform-landing-page-icon": {
```
jpinkney-aws, line 3719: looks like these shouldn't haven't have been committed?

packages/core/src/shared/telemetry/util.ts
```diff
@@ -367,25 +369,68 @@ export function getOperatingSystem(): OperatingSystem {
...
+                            throw addContextToError(e, opts)
+                        }
+                        throw e
+                    }
                 },
                 {
-                    emit: false,
-                    functionId: functionId,
+                    emit: opts.emit !== undefined ? opts.emit : false,
+                    functionId: opts,
```
jpinkney-aws, line 415: is passing in opts to functionId also implicitly passing in errorCtx and emit?

packages/core/src/test/shared/telemetry/spans.test.ts
```diff
@@ -575,19 +575,153 @@ describe('TelemetryTracer', function () {
...
+            }
+        }
+        const arbitraryError = new Error('arbitrary error')
+
+        it(`wraps errors with function id context`, async function () {
+            const inst = new TestThrows()
+            assert.throws(
+                () => inst.throwsError(),
+                (e) => {
+                    if (!(e instanceof ToolkitError)) {
```
jpinkney-aws, line 645: can we simplify these next 4 or 5 lines? I think they're repeated below as well

packages/core/src/shared/telemetry/util.ts
nkomonen-amazon: Yup, but in our typing we do not specify those values. You think would be bad practice?

jpinkney-aws: > I could see cases where if we iterated over all keys that it would get unexpected ones

Yeah, or even someone does a cast to any and tries to use one of the fields accidentally (even more rare than the other case). I think if we pass in opts we're technically lying about the type

nkomonen-amazon: Sounds good, done

nkomonen-amazon:
@justinmk3 can you give this a look when you have time

packages/core/src/test/shared/telemetry/spans.test.ts
```diff
@@ -685,6 +685,44 @@ describe('TelemetryTracer', function () {
                 (e) => e === arbitraryError
             )
         })
+
+        const customWithTelemetryContext = withTelemetryContextFactory({ class: 'TestCustomContext', emit: true })
```
justinmk3, line 686: Modules can define their own predefined fields. That's more explicit, and avoids this extra layer (we already have too many layers). And avoids these extra tests.

justinmk3:
Manually annotating the "code path" can be useful in specific, but it is not a sustainable approach in general. Can we evaluate something like https://github.com/felixge/node-stack-trace to inspect the actual stacktrace instead?

packages/core/src/test/shared/telemetry/spans.test.ts
nkomonen-amazon: Makes sense, removed

justinmk3 (approved):
We should find a way to solve this problem that doesn't require explicit annotations. Can merge this for now though.

PR #5817, opened 2024-10-19
fix(telemetry): session_end not caching events

nkomonen-amazon:
## Problem

When the user shutdown their extension gracefully our mechanism that would cache any pending metrics for the next startup would not save to disk.

This was due to the VS Code FS implementation being used, but it will not work since the function it is called in is called by `deactivate()` which does not have access to vscode apis on shutdown.

## Solution

Solution is to swap it with the node fs implementation
---

### DIFF
--- a/packages/core/src/shared/telemetry/telemetryService.ts
+++ b/packages/core/src/shared/telemetry/telemetryService.ts
@@ -20,6 +20,7 @@ import { ClassToInterfaceType } from '../utilities/tsUtils'
 import { getClientId, validateMetricEvent } from './util'
 import { telemetry, MetricBase } from './telemetry'
 import fs from '../fs/fs'
+import fsNode from 'fs/promises'
 import * as collectionUtil from '../utilities/collectionUtils'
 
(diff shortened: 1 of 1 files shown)


packages/core/src/shared/telemetry/telemetryService.ts
```diff
@@ -116,7 +117,10 @@ export class DefaultTelemetryService {
             })
 
             try {
-                await fs.writeFile(this.persistFilePath, JSON.stringify(this._eventQueue))
+                /**
+                 * This function runs in deactivate() so we must use node fs. See the vscode behavior doc for more info.
+                 */
+                await fsNode.writeFile(this.persistFilePath, JSON.stringify(this._eventQueue))
```
jpinkney-aws, line 123: I guess in vscode webs case this doesn't matter since it has nothing to persist to anyways. Will it throw an error or anything since we're using fs directly or is that properly stubbed out?

jpinkney-aws (approved):

PR #5828, opened 2024-10-21
feat(notifications): notifications controller and view panel

hayemaxi:
- NotificationsController
  - Fetches notifications and determines which ones to display.
- NotificationsNode
  - Side panel that displays a given list (from controller) of notifications.

The controller will fetch notifications from an endpoint (or local files during development). These notifications will be stored in memory and global state. They will then be evaluated for dismissal/criteria and if they pass they will be sent to the NotificationsNode for display.

Extensions will call `NotificationController.pollForStartUp()` and `NotificationController.pollForEmergencies()` to get notifications.

image

---


packages/core/package.nls.json
```diff
@@ -5,6 +5,8 @@
     "AWS.productName.cn": "Amazon Toolkit",
     "AWS.amazonq.productName": "Amazon Q",
     "AWS.codecatalyst.submenu.title": "Manage CodeCatalyst",
+    "AWS.notifications.title": "Notifications",
+    "AWS.notifications.dismiss.title": "Dismiss",
```
justinmk3, line 9: should this be in the `AWS.generic.x` namespace? https://github.com/aws/aws-toolkit-vscode/blob/15c636c764dffbc7d1b7fa7f3b4c024a0b72b108/packages/core/package.nls.json#L239-L241
```suggestion
    "AWS.generic.dismiss": "Dismiss",
```

packages/core/src/notifications/controller.ts
```diff
@@ -0,0 +1,233 @@
...
+    constructor(extPrefix: 'amazonq' | 'toolkit', node: NotificationsNode) {
+        if (!NotificationsController.#instance) {
+            registerDismissCommand(extPrefix)
+        }
+        NotificationsController.#instance = this
+
+        this.storageKey = `aws.${extPrefix}.notifications`
+        this.notificationsNode = node
+
+        this.state = globals.globalState.get(this.storageKey) ?? {
```
justinmk3, line 59: generally a good idea to use `tryGet`. it gives some extra type safety.
```suggestion
        this.state = globals.globalState.tryGet(this.storageKey, Object, {
```

packages/core/src/notifications/controller.ts
```diff
@@ -0,0 +1,233 @@
...
+        }
+    }
+
+    /**
+     * Permanently hides a notification from view. Only 'startUp' notifications can be dismissed.
+     * Users are able to collapse or hide the notifications panel in native VSC if they want to
+     * hide all notifications.
+     */
+    public async dismissNotification(notificationId: string) {
+        getLogger().debug('Dismissing notification: %s', notificationId)
```
justinmk3, line 116: since logger accepts a topic now, it will help avoid some churn later to start using it where possible.
```suggestion
        getLogger('notifications').debug('Dismissing notification: %s', notificationId)
```

packages/core/src/notifications/controller.ts
```diff
@@ -0,0 +1,233 @@
...
+        if (this.#instance === undefined) {
+            throw new ToolkitError('NotificationsController was accessed before it has been initialized.')
+        }
+
+        return this.#instance
+    }
+}
+
+function registerDismissCommand(extPrefix: string) {
+    const name = `_aws.${extPrefix}.notifications.dismiss`
```
justinmk3, line 207: nit/non-blocker: because of the "stringly typed" nature of vscode commands, it is helpful if they can be found with a search (grep etc). So for that reason maybe we should explicitly do: 
```suggestion
    const name = isAmazonQ() ? `_aws.amazonq.notifications.dismiss` : '_aws....'
```

not sure though.

packages/core/src/notifications/controller.ts
```diff
@@ -0,0 +1,233 @@
...
+        await this.writeState()
+
+        NotificationsNode.instance.dismissStartUpNotification(notificationId)
+    }
+
+    /**
+     * Fetch notifications from the endpoint and store them in the global state.
+     */
+    private async fetchNotifications(category: NotificationType) {
+        const response = _useLocalFiles ? await this.fetchLocally(category) : await this.fetchRemotely(category)
```
nkomonen-amazon, line 127: Another option is to use dependency injection and pass an object that matches `fetchX`, then in the class constructor you could pass the instance you want for tests.

But I'm assuming you don't have access to the constructor at runtime, so maybe creating a top level function in this file where you can set a fetch implementation (otherwise defaulting to the http one), would work best. Then the constructor grabs whatever is there at instantiation.

This will get rid of the custom test environment logic

packages/core/src/notifications/index.ts
```diff
@@ -0,0 +1,9 @@
+/*!
```
justinmk3, line 1: somewhat off-topic, but would it avoid boilerplate if the main/top-level core `index.ts` file instead just exports all notifications stuff under a `notifications` item? that has at least these advantages:

- avoids needing to update `package.json`
- avoids needing to create this file
- centralizes the index
- avoids consumers having to hunt for index files (they import one "mega index")

packages/core/src/shared/globalState.ts
```diff
@@ -31,6 +31,8 @@ type globalKey =
     | 'aws.amazonq.codewhisperer.newCustomizations'
     | 'aws.amazonq.hasShownWalkthrough'
     | 'aws.amazonq.showTryChatCodeLens'
+    | 'aws.amazonq.notifications'
+    | 'aws.toolkit.notifications'
```
justinmk3, line 35: since each extension has its own globalState, do we actually need 2 keys or would one `aws.notifications` key work? that would also simplify a fair amount of code AFAICT.

packages/core/src/shared/vscode/setContext.ts
```diff
@@ -10,19 +10,21 @@ import * as vscode from 'vscode'
...
     | 'aws.amazonq.showLoginView'
+    | 'aws.amazonq.notifications.show'
     | 'aws.codecatalyst.connected'
     | 'aws.codewhisperer.connected'
     | 'aws.codewhisperer.connectionExpired'
     | 'aws.codewhisperer.tutorial.workInProgress'
     | 'aws.explorer.showAuthView'
     | 'aws.toolkit.amazonq.dismissed'
     | 'aws.toolkit.amazonqInstall.dismissed'
+    | 'aws.toolkit.notifications.show'
```
justinmk3, line 27: Unlike globalState (which definitely is local to each extension), I haven't confirmed if setContext keys are "local" to each extension. So it may be required to define a q-specific and toolkit-specific key, as you've done here. However, Claude 3 claims that setContext keys are isolated per-extension:


> extension has its own namespace for the `setContext` command, ensuring that there is no conflict between different extensions defining context keys with the same name.
> 
> When an extension defines a context key using the `setContext` command, the key is automatically prefixed with the extension's unique identifier (the `publisher.name` from the extension's `package.json` file). This namespacing mechanism ensures that context keys defined by different extensions are kept separate and do not conflict with each other.
> 
> For example, if extension A defines a context key called `myKey` and extension B also defines a context key called `myKey`, the actual context keys stored in Visual Studio Code would be:
> 
> - `extension-a.myKey` (for the key defined by extension A)
> - `extension-b.myKey` (for the key defined by extension B)
> 
> This namespacing applies not only to context keys defined by `setContext` but also to context keys referenced in other parts of an extension, such as in `when` clauses for command visibility or in `package.json` directives.

packages/core/src/notifications/controller.ts
```diff
@@ -0,0 +1,233 @@
...
+        if (!response.content) {
+            getLogger().verbose('No new notifications for category: %s', category)
+            return
+        }
+
+        getLogger().verbose('ETAG has changed for notifications category: %s', category)
+
+        this.state[category].payload = JSON.parse(response.content)
```
nkomonen-amazon, line 135: From experience, either this or the network call can fail due to network issues. Which can result in SyntaxError or some failed request. 

Is there any error handling upstream? I didn't check thoroughly.

Also see if the withTelemetryContext  decorator can help to give the upstream context about any failures that occur in this function. (I have a pending pr for this)

packages/core/src/test/notifications/controller.test.ts
```diff
@@ -0,0 +1,506 @@
...
+    beforeEach(async function () {
+        panelNode.setNotifications([], [])
+        controller = new NotificationsController('toolkit', panelNode)
+
+        sandbox = sinon.createSandbox()
+
+        getNewETagContentStub = sandbox.stub()
+        ruleEngineSpy = sandbox.spy(ruleEngine, 'shouldDisplayNotification')
+        focusPanelSpy = sandbox.spy(panelNode, 'focusPanel')
+        sandbox.stub(HttpResourceFetcherModule, 'HttpResourceFetcher').callsFake(function (this: any, url: string) {
```
justinmk3, line 70: The tests in this PR do a nice job of minimizing mocks/stubs. Does it help at all to plumb this "fetcher" abstraction through as a parameter (to e.g. the constructor)? After all, I think the Notifcations has a singleton which presumably is intended to serve the purpose of injecting stuff.

justinmk3 (approved):

packages/core/src/notifications/controller.ts
tomcat323: If you want to add the topic its in `toolkitLogger.ts`

packages/core/src/notifications/controller.ts
hayemaxi: No telemetry yet - will be added in a future PR. Will keep this in mind.

packages/core/src/notifications/controller.ts
hayemaxi: Its not meant for use in test cases - only during manual testing when you have the extension running in debug mode or you are testing a packaged vsix.

packages/core/src/shared/vscode/setContext.ts
hayemaxi: Tested - they are shared. 
 Claude is a liar 😡

packages/core/src/notifications/index.ts
hayemaxi: In addition to a top-level index.js, you would still need a lower level one to collect the individual files though right?
e.g.
```
src/notifications/index.ts

export * from 'controller.ts'
export * from 'types.ts'

```

```
src/index.ts

export * as notifications from 'notifications/index'
```

Maybe you could do this, but this looks very messy:
```
src/index.ts

import * as controller from './src/notifications/controller';
import * as types from './src/notifications/types';

export const notifications = {
  ...controller,
  ...types,
}
```

Also, not sure how autocomplete will behave. Needs investigation.

justinmk3: > you would still need a lower level one to collect the individual files though right?
> e.g.

no, could do that manually in the top level index.ts like this: https://github.com/aws/aws-toolkit-vscode/blob/6fa3a9e554f4aa1916d01d00ea93497ca7205cc5/packages/core/src/shared/index.ts#L50-L51

of course, this means the top level has to "choose" some names, but in practice that doesn't seem to be a problem, and it may actually lead to a better interface for the consumers.

packages/core/src/notifications/controller.ts
hayemaxi: I updated it anyway, the new flow looks good and does help the unit test case path as a side effect.

packages/core/src/notifications/index.ts
hayemaxi: So then you have pretty granular exports anyways, right? e.g.
```
export * as notificationsTypes from './src/notifications/types'
export * as notificationsRules from './src/notifications/rules'
export * as notificationsController from './src/notifications/controller'
```
which isn't as intuitive as a single `notifications` export from all of these places.

justinmk3: True, though it depends on what consumers actually need. Generally we shouldn't need to expose most modules in the "public" interface.

Tracked in https://taskei.amazon.dev/tasks/IDE-15174

PR #5831, opened 2024-10-22
refactor(cwl): Remove LiveTail specific limit preference, rename scheme, improve exception messaging

keeganirby:
## Problem
Addresses some feedback on past PRs

## Solution
* Removes CWL LIveTail limit preference. Instead pulls from existing limit preference. Updated description of limit preference to describe LiveTail line trimming behavior
* Renames CWL Scheme variable to fit linting rules
* Improves exception messaging when catching an on-stream failure
* 

---


packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -83,7 +83,9 @@ async function handleSessionStream(
             }
         }
     } catch (err) {
-        throw new ToolkitError('Caught on-stream exception')
+        throw new ToolkitError(
+            `Encountered LiveTail on-stream failure for session ${session.uri}: ${(err as Error).message}`
```
justinmk3, line 87: I would suggest not catching the error at all, just let it bubble up. please review our [guidelines](https://github.com/aws/aws-toolkit-vscode/blob/master/docs/CODE_GUIDELINES.md#exceptions).

if you must catch and re-throw this, then it's usually a good idea to use `ToolkitError.chain` so that the original error is part of its [chain](https://github.com/aws/aws-toolkit-vscode/blob/master/docs/arch_develop.md#chaining).

justinmk3 (approved):
lgtm after toolkiterror.chain() or bubble up

PR #5833, opened 2024-10-22
ci(lint): run jscpd in CI, compare to static threshold. 

Hweinstock:
## Problem
New contributors often copy-paste logic and test cases, which creates a maintenance cost for little gain.

Examples

- https://github.com/aws/aws-toolkit-vscode/pull/3837#discussion_r1367182525
- https://github.com/aws/aws-toolkit-vscode/pull/3805#discussion_r1317867006

We call these repeated segments of code "clones". 

## Solution
implement a static threshold. 
Run `jscpd` in CI and compare the % duplicated to a fixed threshold of 1.34%, which is the current duplication rate in the repo. 
The main issue with this solution is that it fails to provide easy debugging options when new duplicates are found. Currently exploring adding this functionality in a follow-up. 

#### Notes
- this is a summary of the problem/solution. For full information, see full doc (VSCode: Detecting and Reducing Duplicate Code) describing changes. 
---

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -56,8 +56,8 @@ jobs:
             - run: npm run lint
 
     jscpd:
-        # needs: lint-commits
-        runs-on: ubuntu-22.04
+        needs: lint-commits
+        runs-on: ubuntu-latest
         strategy:
             matrix:
                 node-version: [18.x]
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -64,14 +64,12 @@ jobs:
         env:
             NODE_OPTIONS: '--max-old-space-size=8192'
         steps:
-            - name: check out repo
-              uses: actions/checkout@v4
+            - uses: actions/checkout@v4
             - name: Use Node.js ${{ matrix.node-version }}
               uses: actions/setup-node@v4
               with:
                   node-version: ${{ matrix.node-version }}
-            - name: install jscpd
-              run: npm install jscpd
+            - run: npm install jscpd
             - name: Run jscpd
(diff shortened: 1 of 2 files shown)


.github/workflows/node.js.yml
```diff
@@ -55,6 +55,26 @@ jobs:
             - run: npm run testCompile
             - run: npm run lint
 
+    jscpd:
+        # needs: lint-commits
```
jpinkney-aws, line 59: looks like we just need to remove this an its probably good to merge?

Hweinstock: Yeah, still exploring adding some more functionality, but that can be follow up. Seems pretty low-risk to test this simple version in the meantime.

.github/workflows/node.js.yml
```diff
@@ -55,6 +55,26 @@ jobs:
...
+            NODE_OPTIONS: '--max-old-space-size=8192'
+        steps:
+            - name: check out repo
+              uses: actions/checkout@v4
+            - name: Use Node.js ${{ matrix.node-version }}
+              uses: actions/setup-node@v4
+              with:
+                  node-version: ${{ matrix.node-version }}
+            - name: install jscpd
+              run: npm install jscpd
```
justinmk3, line 74: it's often more readable not to have a name, when the "run" part gives the same (or better) information.
```suggestion
            - uses: actions/checkout@v4
            - name: Use Node.js ${{ matrix.node-version }}
              uses: actions/setup-node@v4
              with:
                  node-version: ${{ matrix.node-version }}
            - run: npm install jscpd
```

justinmk3 (approved):
🚀

PR #5834, opened 2024-10-22
feat(stepfunctions): Allow closing Workflow Studio and reopening it

witness-me:
## Problem
Workflow Studio (WFS) is opening for all asl files by default and there is no straightforward way to close it in favour of default editor if the user requires that. Additionally, there is no easy way open WFS from the current default editor

## Solution
Adding options to close and reopen WFS custom editor:
- WFS still launches as a default editor for asl files
- On clicking the custom close button in WFS, the webview panel gets closed and the custom editor is opened instead
- With the custom editor opened, there is an option to launch WFS by clicking codelens command on top of the file. This command will close current(default) editor and reopen the file in WFS. If WFS is already opened for the file, it will redirect to that tab
- There is an additional option to launch WFS on right-click context menu on the file. For this case:
  - if there is a WFS for this file already launched, it will focus that tab
  - otherwise it will open file with WFS in the new tab
  - if the file is opened in the default editor, it will not be closed

[Demo gif](https://maxis-file-service-prod-pdx.pdx.proxy.amazon.com/issues/ce4baf9f-c776-40e5-8183-8be1bdb3acbf/attachments/70096bfa3626fee7e1126c30fa6ae895c105b78bcbba68a6ba14d7f7cb085a3f_7dbee587-2cbe-482b-984b-a02564d70a51)

---

### DIFF
--- a/packages/core/src/stepFunctions/workflowStudio/activation.ts
+++ b/packages/core/src/stepFunctions/workflowStudio/activation.ts
@@ -5,6 +5,7 @@
 
 import * as vscode from 'vscode'
 import { WorkflowStudioEditorProvider } from './workflowStudioEditorProvider'
+import { Commands } from '../../shared/vscode/commands2'
 
 /**
  * Activates the extension and registers all necessary components.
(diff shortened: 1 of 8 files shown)


packages/core/src/stepFunctions/workflowStudio/activation.ts
```diff
@@ -12,4 +12,20 @@ import { WorkflowStudioEditorProvider } from './workflowStudioEditorProvider'
  */
 export async function activate(extensionContext: vscode.ExtensionContext): Promise<void> {
     extensionContext.subscriptions.push(WorkflowStudioEditorProvider.register(extensionContext))
+
+    // Open the file with Workflow Studio editor in a new tab, or focus on the tab with WFS if it is already open
+    extensionContext.subscriptions.push(
+        vscode.commands.registerCommand('aws.stepfunctions.openWithWorkflowStudio', async (uri: vscode.Uri) => {
```
hayemaxi, line 18: Use our wrappers instead, `Commands.register` and `Commands.declare`.
```suggestion
        Commands.register('aws.stepfunctions.openWithWorkflowStudio', async (uri: vscode.Uri) => {
```

hayemaxi:
CI: 
```
tech debt
       remove separate sessions login edge cases:

      AssertionError [ERR_ASSERTION]: Remove the edge case code from the commit that this test is a part of.
```

Not related to this PR and isn't a blocker, but consider merging from mainline so you don't see this anymore.

hayemaxi (approved):

PR #5839, opened 2024-10-23
telemetry(amazonq): Create server request metrics that include conversation id creation

jpinkney-aws:
## Problem
- The full server response metrics don't include the ~2 seconds to create the conversation id, it just handles everything between after the conversation id was created

## Solution
- Create new metrics that include the conversation id creation data

---

### DIFF
--- a/packages/core/src/shared/telemetry/vscodeTelemetry.json
+++ b/packages/core/src/shared/telemetry/vscodeTelemetry.json
@@ -157,7 +157,7 @@
         {
             "name": "cwsprChatTimeToFirstChunk",
             "type": "int",
-            "description": "The time between the initial server request and when we got back the first usable result"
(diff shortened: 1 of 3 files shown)


hayemaxi (approved):

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -753,7 +763,13 @@
                     "type": "cwsprChatFullResponseLatency"
                 },
                 {
-                    "type": "cwsprTimeToFirstDisplay"
+                    "type": "cwsprChatTimeFirstUsableChunk"
+                },
+                {
+                    "type": "cwsprChatFullServerResponseLatency"
```
justinmk3, line 769: is something blocking these from being upstreamed to https://github.com/aws/aws-toolkit-common ?

PR #5840, opened 2024-10-23
Pick SageMaker IAM credentials to use for code completion

ahusseinali:
## Problem
SageMaker IAM are not automatically picked up from node environment for Amazon Q Code Completion. However, this feature is supported in AWS-Toolkit and was working fine for code completion before Amazon Q codebase split from aws-toolkit logic

## Solution
- Bring the IAM credential provider to Amazon Q extension activation
- The scope is narrowed down to Sagemaker environment for safety and since it's an existing feature, no extra tests need to be added

---

### DIFF
--- a/packages/amazonq/.changes/next-release/Bug Fix-0ee7be24-b0f8-460b-a852-a2d08b6d289e.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-0ee7be24-b0f8-460b-a852-a2d08b6d289e.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Use SM IAM Credentials for Code Completion"
+}
--- a/packages/amazonq/src/extension.ts
+++ b/packages/amazonq/src/extension.ts
@@ -44,6 +44,7 @@ import {
     setContext,
     setupUninstallHandler,
     maybeShowMinVscodeWarning,
+    isSageMaker,
 } from 'aws-core-vscode/shared'
 import { ExtStartUpSources, telemetry } from 'aws-core-vscode/telemetry'
 import { VSCODE_EXTENSION_ID } from 'aws-core-vscode/utils'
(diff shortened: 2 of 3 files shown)


packages/amazonq/.changes/next-release/Bug Fix-0ee7be24-b0f8-460b-a852-a2d08b6d289e.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Use SM IAM Credentials for Code Completion"
```
justinmk3, line 3: Thanks for including a changelog. Please review the [changelog guidelines](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#changelog).

justinmk3:
> The scope is narrowed down to Sagemaker environment for safety and since it's an existing feature,

Please evaluate whether it makes sense to enable in general. That's part of the work here.

packages/amazonq/.changes/next-release/Bug Fix-0ee7be24-b0f8-460b-a852-a2d08b6d289e.json
ahusseinali: Does it need further refinement? I can take your proposal on that. How much context should be stated in it?

ahusseinali:
> Please evaluate (based on the usages in the code) whether it makes sense to enable in general. That's part of the work here.

I initially had it with a wider scope, but I believe it's risky as the logic has implied assumption that the code is loaded in a container environment with IAM permissions passed to that environment. It's hard to reason about whether it should be enabled in general unless we provide Chat support as well for IAM (which is WIP)

ahusseinali:
Closing this change as it was merged as part of a different change

PR #5845, opened 2024-10-23
telemetry(bug): sessionId not unique

nkomonen-amazon:
## Problem:

There seems to be a bug with vscode.env.sessionId since we observed in telemetry multiple clientIds who had the same sessionId.

## Solution:

Generate our own sessionId and share it between extensions through globalThis

---

### DIFF
--- a/packages/core/src/dev/activation.ts
+++ b/packages/core/src/dev/activation.ts
@@ -61,57 +61,59 @@ let targetAuth: Auth
  * on selection. There is no support for name-spacing. Just add the relevant
  * feature/module as a description so it can be moved around easier.
  */
-const menuOptions: Record<DevFunction, MenuOption> = {
-    installVsix: {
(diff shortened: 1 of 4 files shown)


Princeprince559 (approved):
tada [princeprince559](url)

packages/core/src/dev/activation.ts
```diff
@@ -61,57 +61,59 @@ let targetAuth: Auth
  * on selection. There is no support for name-spacing. Just add the relevant
  * feature/module as a description so it can be moved around easier.
  */
-const menuOptions: Record<DevFunction, MenuOption> = {
```
nkomonen-amazon, line 64: This was defined at extension load and called `getSessionId()` before globals was defined. So we just make this a function so it is created once needed

hayemaxi (approved):

jpinkney-aws (approved):

PR #5850, opened 2024-10-24
fix(crash): Add more telemetry info to crash reporting

nkomonen-amazon:
## Problem

We are seeing weird issues with some users when crash reporting due to crashes for a specific session id being reported multiple times, even though it should report one.

## Solution

Add more telemetry data so that we can see when the heartbeat is successfully sent, and what the crash checker sees when it runs.

## Additional

There are other small commits that make small changes, each has a description for what it is doing
---


packages/core/src/test/shared/crypto.test.ts
```diff
@@ -44,3 +44,23 @@ describe('crypto', function () {
...
+        const fullUUID = 'aaaabbbb-cccc-dddd-eeee-ffffhhhhiiii'
+        const result = truncateUuid(fullUUID)
+        assert.strictEqual(result, 'aaaa...iiii')
+    })
+
+    it('should handle a different valid UUID correctly', function () {
+        const fullUUID = '12340000-0000-0000-0000-000000005678'
+        const result = truncateUuid(fullUUID)
+        assert.strictEqual(result, '1234...5678')
+    })
```
hayemaxi, line 59: These tests are the same?

packages/core/src/shared/crashMonitoring.ts
```diff
@@ -61,22 +61,33 @@ const className = 'CrashMonitoring'
...
 
     static #didTryCreate = false
     static #instance: CrashMonitoring | undefined
     /** Returns an instance of this class or undefined if any initial validation fails. */
     public static async instance(): Promise<CrashMonitoring | undefined> {
+        // - We do not support web due to some incompatibilities
+        // - We do not support cloud desktop due to false crash reports due to sleep+vpn+ssh timeouts with the IDE
+        if (isWeb() || (await getComputeEnvType()) === 'cloudDesktop-amzn') {
+            return undefined
+        }
```
hayemaxi, line 89: Why not local only? Wouldn't any remote environment potentially have this issue?

nkomonen-amazon: There is **a lot** of noise caused specifically by cloud desktop, compared to the other ssh'd instances it is not much. This is probably since we have lots of Amazon devs using it.

The goal is to get rid of the noise while still collecting data on the others so that if we need it in the future it is there.

packages/core/src/test/shared/crypto.test.ts
nkomonen-amazon: good point, merged with the other

hayemaxi (approved):

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1184,8 +1189,31 @@
...
                     "required": false
+                },
+                {
+                    "type": "timestamp",
+                    "required": false
                 }
             ]
+        },
+        {
+            "name": "ide_heartbeat",
```
justinmk3, line 1200: `toolkit_heartbeat` may fit better. ide_ is for non-extension evens (from the ide)

packages/core/src/shared/crashMonitoring.ts
justinmk3: would that apply to many/most "remote" situations?

PR #5853, opened 2024-10-24
feat(feature dev): Add setting to allow Q to run code and test commands

willyyhuang:
## Problem

Q /dev team is launching a new feature: Allow Q to execute build/test based on customer's configuration 🎉🎉

## Solution

We have made a couple tweaks to incorporate the user experience
- Added a new setting to allow Q to run code and test commands per project

image

- Added a pre-filled prompt option for customer to generate a configuration for this feature
- Added a couple follow up buttons throughout the /dev flow and added logic around them

image

- Updated copies
- Updated `codefileExtensions` set

---


packages/amazonq/package.json
```diff
@@ -127,6 +127,11 @@
                     "markdownDescription": "%AWS.configuration.description.amazonq%",
                     "default": true
                 },
+                "amazonQ.allowDevCommands": {
+                    "markdownDescription": "%AWS.configuration.description.allowDevCommands%",
+                    "type": "object",
```
jpinkney-aws, line 132: Is there a better name for that setting? `allowDevCommands` makes me think that its a boolean rather than an object

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -491,12 +498,16 @@ export class FeatureDevController {
...
     }
-    private workOnNewTask(
+    private async workOnNewTask(
         message: any,
         remainingIterations: number = 0,
         totalIterations?: number,
         isStoppedGeneration: boolean = false
     ) {
+        const hasDevFile = await checkForDevFile(
+            (await this.sessionStorage.getSession(message.tabID)).config.workspaceRoots[0]
```
jpinkney-aws, line 508: Is there some UI indication that only the first workspace root is used?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -537,6 +564,20 @@ export class FeatureDevController {
             i18n('AWS.amazonq.featureDev.placeholder.additionalImprovements')
         )
     }
+
+    private async processAutoBuildSetting(setting: boolean, msg: any) {
+        const root = (await this.sessionStorage.getSession(msg.tabID)).config.workspaceRoots[0]
```
jpinkney-aws, line 569: I think this is used a few places, does it make sense to do this in one place so everything is aligned and we're always pulling the first workspace root?

packages/core/src/codewhisperer/util/codewhispererSettings.ts
```diff
@@ -64,6 +65,18 @@ export class CodeWhispererSettings extends fromExtensionManifest('amazonQ', desc
         return Math.max(this.get('workspaceIndexMaxSize', 250), 1)
     }
 
+    public getAutoBuildFeatureProjects(): { [key: string]: boolean } {
```
jpinkney-aws, line 68: It feels weird that now we are moving changes into the codewhisperer project to support this. Maybe we should seperate this into amazon q chat settings and amazon q inline settings? And then q inline can live here and amazon q chat can live in the amazon q folder

packages/core/src/shared/filetypes.ts
```diff
@@ -349,5 +357,6 @@ export const codefileExtensions = new Set([
 /** Returns true if `filename` is a code file. */
 export function isCodeFile(filename: string): boolean {
     const ext = path.extname(filename).toLowerCase()
-    return codefileExtensions.has(ext)
+    const result = codefileExtensions.has(ext) || codefileExtensions.has(filename)
```
jpinkney-aws, line 360: Can you elaborate on why we need `codefileExtensions.has(filename)`?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -147,6 +148,12 @@ export class FeatureDevController {
                 case FollowUpTypes.SendFeedback:
                     this.sendFeedback()
                     break
+                case FollowUpTypes.AcceptAutoBuild:
+                    return this.processAutoBuildSetting(true, data)
+                case FollowUpTypes.DenyAutoBuild:
+                    return this.processAutoBuildSetting(false, data)
+                case FollowUpTypes.GenerateDevFile:
+                    return this.newTask(data, true)
```
chengoramazon, line 156: Can you create another function for this case?

packages/core/src/amazonqFeatureDev/constants.ts
```diff
@@ -14,6 +14,9 @@ export const featureDevChat = 'featureDevChat'
 
 export const featureName = 'Amazon Q Developer Agent for software development'
 
+export const generateDevFilePrompt =
+    'generate a devfile in my repository. Note that you should only use devfile version 2.0.0 and the only supported command is test, so you should bundle all install, build and test commands in "test". also you can use "public.ecr.aws/aws-mde/universal-image:latest" as universal image if you aren\'t sure which image to use. here is an example for a node repository: schemaVersion: 2.0.0 components: - name: dev container: image: public.ecr.aws/aws-mde/universal-image:latest commands: - id: test exec: component: dev commandLine: "npm install && npm run build && npm run test"'
```
jpinkney-aws, line 18: Has this string of text been reviewed by ux?

packages/core/src/amazonqFeatureDev/session/session.ts
```diff
@@ -59,14 +64,51 @@ export class Session {
...
-            await this.setupConversation(msg)
-            this.preloaderFinished = true
-            this.messenger.sendAsyncEventProgress(this.tabID, true, undefined)
-            await this.proxyClient.sendFeatureDevTelemetryEvent(this.conversationId) // send the event only once per conversation.
+        const root = this.config.workspaceRoots[0]
+        const autoBuildProjectSetting = CodeWhispererSettings.instance.getAutoBuildFeatureProjects()
+        const hasDevfile = await checkForDevFile(root)
+        const isPromptedForAutoBuildFeature = Object.keys(autoBuildProjectSetting).includes(root)
+
+        if (hasDevfile && !isPromptedForAutoBuildFeature) {
```
chengoramazon, line 72: Can we extract the logic out of `preloader` function?

packages/core/src/shared/filetypes.ts
willyyhuang: `.gitignore` for example is ignored 

`path.extname(".gitignore")` is unable to return `gitignore`, it actually just returns nothing

packages/core/src/amazonqFeatureDev/session/session.ts
```diff
@@ -59,14 +64,51 @@ export class Session {
...
+                await this.setupConversation(msg)
+                this.preloaderFinished = true
+                this.messenger.sendAsyncEventProgress(this.tabID, true, undefined)
+                await this.proxyClient.sendFeatureDevTelemetryEvent(this.conversationId) // send the event only once per conversation.
+            }
+            return this.preloaderFinished
         }
     }
 
+    private async promptUserConsent(tabID: string) {
```
chengoramazon, line 86: It looks only for the exeuction-engine feature. Can we rename it a bit specific?

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -867,8 +908,18 @@ export class FeatureDevController {
         // Re-run the opening flow, where we check auth + create a session
         await this.tabOpened(message)
 
-        this.messenger.sendChatInputEnabled(message.tabID, true)
-        this.messenger.sendUpdatePlaceholder(message.tabID, i18n('AWS.amazonq.featureDev.placeholder.describe'))
+        if (generateDevFile) {
```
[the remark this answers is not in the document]
willyyhuang: yes

packages/amazonq/package.json
willyyhuang: sure, how does `allowDevCommandsConfiguration` sound? 

it was quite tricky for me to do the naming in the first place

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
willyyhuang: no, good point, let me think about how to handle this

packages/core/src/amazonqFeatureDev/session/session.ts
willyyhuang: sure

packages/core/src/amazonqFeatureDev/session/session.ts
willyyhuang: as in, these 4 lines of code? sure

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
willyyhuang: I will change the function to optionally accept "prefilledPrompt" instead

packages/core/src/amazonqFeatureDev/session/session.ts
willyyhuang: actually, what are you suggesting here? I think this logic sits well here, do not need to extract the logic out, since it's part of pre-loading the conversation

packages/amazonq/package.json
willyyhuang: we have decided to go with `devCommandWorkspaceConfigurations`

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -867,8 +908,18 @@ export class FeatureDevController {
         // Re-run the opening flow, where we check auth + create a session
         await this.tabOpened(message)
 
-        this.messenger.sendChatInputEnabled(message.tabID, true)
-        this.messenger.sendUpdatePlaceholder(message.tabID, i18n('AWS.amazonq.featureDev.placeholder.describe'))
+        if (prefilledPrompt) {
+            this.messenger.sendAnswer({
```
chengoramazon, line 912: Non-blocking, but why new task has to couple with dev file generation?

```
this.messenger.sendAnswer({
                type: 'system-prompt',
                tabID: message.tabID,
                message: i18n('AWS.amazonq.featureDev.pillText.generateDevFile'),
            })
this.newTask(data, generateDevFilePrompt)
```

chengoramazon (approved):

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
willyyhuang: good catch, let me change that

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
willyyhuang: We will follow up with this, most likely going to modify from strings

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -19,9 +19,16 @@ import { TelemetryHelper } from './telemetryHelper'
 import { maxRepoSizeBytes } from '../constants'
 import { isCodeFile } from '../../shared/filetypes'
 import { fs } from '../../shared'
+import { CodeWhispererSettings } from '../../codewhisperer'
 
 const getSha256 = (file: Buffer) => createHash('sha256').update(file).digest('base64')
 
+export async function checkForDevFile(root: string) {
+    const devFilePath = root + '/devfile.yaml'
```
jpinkney-aws, line 27: Do we need a path.join here? Will this work on windows?

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -34,15 +41,19 @@ export async function prepareRepoData(
 ) {
     try {
         const files = await collectFiles(repoRootPaths, workspaceFolders, true, maxRepoSizeBytes)
+        const devCommandWorkspaceConfigurations = CodeWhispererSettings.instance.getDevCommandWorkspaceConfigurations()
+        const useAutoBuildFeature = devCommandWorkspaceConfigurations[repoRootPaths[0]] ?? false
```
jpinkney-aws, line 45: are the test for prepare repo data updated to handle this?

packages/core/src/amazonqFeatureDev/util/files.ts
willyyhuang: good catch, I can change it to use `path.join`

as for windows, I will get someone with a windows machine to test

packages/core/src/amazonqFeatureDev/util/files.ts
willyyhuang: are you mentioning `prepareRepoData.test.ts`?

jpinkney-aws: yeah

jpinkney-aws: they seem to be passing so i guess its fine

jpinkney-aws (approved):

PR #5663, opened 2024-09-25
config(amazonq): enable more languages in generateCompletion requests

Will-ShaoHua:
## Problem

Per product team's request, unblock requests of the following language to the service which will be ready soon. (thus not calling it out as a "feat" as it's not supported yet)

image

## Functional test

https://github.com/user-attachments/assets/bcc7646c-d38c-4d4b-8ee3-d67dc7c92154

https://github.com/user-attachments/assets/a6b62ac5-9e40-4fc1-85f5-24e274b75908

https://github.com/user-attachments/assets/d74d56f4-66cc-49ba-9944-95dd826471b8

https://github.com/user-attachments/assets/70bc7ef5-4ee5-4699-801f-97e8cb694caa

https://github.com/user-attachments/assets/8451d791-baa0-44a1-836a-da5ff5bf7cdb

https://github.com/user-attachments/assets/d9bad39d-32ab-4632-96f5-1922ebf807cc

---

License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5854, opened 2024-10-24
config(amazonq): patch language configuration not added correctly #5663

Will-ShaoHua:
## Problem

patch #5663 

Due to this line of code https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/util/runtimeLanguageContext.ts#L156-L173

when users make inline completion call with new languages, logs will show
`2024-10-24 08:52:55.501 [error] codewhisperer: unknown runtime language dart`,  however it doesn't have functional impact because it will fallback to original language string (which is the reason why it doesn't fail test cases) and it's only affecting logs.

test has been added https://github.com/aws/aws-toolkit-vscode/blob/master/packages/amazonq/test/unit/codewhisperer/util/runtimeLanguageContext.test.ts#L184-L222

## Solution

---


leigaol (approved):

zixlin7 (approved):

leigaol:
Typo in description

packages/core/src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -21,11 +21,17 @@ const runtimeLanguageSet: ReadonlySet<RuntimeLanguage> = new Set([
     'kotlin',
     'php',
     'python',
+    'powershell',
```
Will-ShaoHua, line 24: verified, equal to jb
`const val ID = "powershell"`

packages/core/src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -21,11 +21,17 @@ const runtimeLanguageSet: ReadonlySet<RuntimeLanguage> = new Set([
...
+    'r',
+    'dart',
```
Will-ShaoHua, line 26: `const val ID = "dart"`

packages/core/src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -21,11 +21,17 @@ const runtimeLanguageSet: ReadonlySet<RuntimeLanguage> = new Set([
...
     'ruby',
     'rust',
     'scala',
     'shell',
     'sql',
+    'swift',
```
Will-ShaoHua, line 32: `const val ID = "swift"`

packages/core/src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -21,11 +21,17 @@ const runtimeLanguageSet: ReadonlySet<RuntimeLanguage> = new Set([
...
+    'lua',
```
Will-ShaoHua, line 33: `const val ID = "lua"`

packages/core/src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -21,11 +21,17 @@ const runtimeLanguageSet: ReadonlySet<RuntimeLanguage> = new Set([
...
+    'vue',
```
Will-ShaoHua, line 34: `const val ID = "vue"`

andrewyuq (approved):

jpinkney-aws (approved):

PR #5857, opened 2024-10-24
feat(cwl): Support autoscrolling live tail session's visible editors

keeganirby:
## Problem
New log events in a LiveTail session are added to the end of a TextDocument. Session updates can contain multiple LogEvents, leading to the text document growing quickly past the user's view. requiring them to constantly be scrolling the document themselves. 

## Solution
To maintain parity with CWL console experience (and the classic `tail -f` experience), we want to auto scroll the customer's LiveTail session's visible editors to the bottom. 

This will only happen IF the end of file is already in view. Meaning, if a User scrolls UP (causing the end of file to not be in view), auto scrolling will not enable. However, if they scroll back down to EOF themselves, it will re-enable.

Simply, if the end of file is in view, when new logEvents are added to the document, the editor will be autoscrolled back to the end of the file.

---

### DIFF
--- a/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
+++ b/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
@@ -107,8 +107,7 @@ function formatLogEvent(logEvent: LiveTailSessionLogEvent): string {
 //This allows for newly added log events to stay in view.
 function getTextEditorsToScroll(document: vscode.TextDocument): vscode.TextEditor[] {
     return vscode.window.visibleTextEditors.filter((editor) => {
-        const isEditorForSession = editor.document === document
-        if (!isEditorForSession) {
+        if (editor.document !== document) {
             return false
         }
         return editor.visibleRanges[0].contains(new vscode.Position(document.lineCount - 1, 0))
--- a/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
+++ b/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
@@ -77,7 +77,11 @@ async function handleSessionStream(
                 formatLogEvent(logEvent)
             )
             if (formattedLogEvents.length !== 0) {
+                //Determine should scroll before adding new lines to doc because adding large
+                //amount of new lines can push bottom of file out of view before scrolling.
+                const editorsToScroll = getTextEditorsToScroll(document)
                 await updateTextDocumentWithNewLogEvents(formattedLogEvents, document, session.maxLines)
+                editorsToScroll.forEach(scrollTextEditorToBottom)
             }
         }
     }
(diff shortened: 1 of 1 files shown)


packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -99,6 +103,24 @@ function formatLogEvent(logEvent: LiveTailSessionLogEvent): string {
...
+        if (!isEditorForSession) {
+            return false
+        }
+        return editor.visibleRanges[0].contains(new vscode.Position(document.lineCount - 1, 0))
+    })
+}
+
+function scrollTextEditorToBottom(editor: vscode.TextEditor) {
+    const topPosition = new vscode.Position(Math.max(editor.document.lineCount - 2, 0), 0)
+    const bottomPosition = new vscode.Position(Math.max(editor.document.lineCount - 2, 0), 0)
```
jpinkney-aws, line 120: These two constants are the same, are we able to just have one that's scroll position and then pass that two times into range?

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -99,6 +103,24 @@ function formatLogEvent(logEvent: LiveTailSessionLogEvent): string {
     return line
 }
 
+//Auto scroll visible LiveTail session editors if the end-of-file is in view.
+//This allows for newly added log events to stay in view.
+function getTextEditorsToScroll(document: vscode.TextDocument): vscode.TextEditor[] {
+    return vscode.window.visibleTextEditors.filter((editor) => {
```
jpinkney-aws, line 109: If someone has a two visible editors, one on the left and one on the right like:
```
____________________________________________________
|                      |                           |
| first editor         |       second editor       |
|                      |                           |
____________________________________________________
```
does that mean both of them will scroll?

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -99,6 +103,24 @@ function formatLogEvent(logEvent: LiveTailSessionLogEvent): string {
...
+        const isEditorForSession = editor.document === document
```
jpinkney-aws, line 110: I think this is only used once, should you just inline this on line 111

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
keeganirby: Yep! they'd both be considered visible and able to be scrolled if at EOF. If one isn't at EOF, and the other is, then only one would be scrolled.

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
keeganirby: Yeah definitely!

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
keeganirby: I could do that!

The reason I wrote this as a function was not so much about reusability, but more about readability. The code going into `handleSessionStream` is growing in size and complexity. I worry that as we do more per stream update that it will become less clear what "actions" we take. Having this defined as a function makes it easier to see that per stream update that we need to 'getTextEditorsToScroll`, then update the document, then scroll it. While abstracting away those implementations.

jpinkney-aws: Oh yeah, I thought that make sense. I was more just commenting on doing:
```
if (editor.document !== document) {
...
}
```
vs
```
const isEditorForSession = editor.document === document
if (!isEditorForSession) {

}
```

keeganirby: Oh haha I misunderstood 😅 I'll change this

jpinkney-aws (approved):

PR #5858, opened 2024-10-24
feat(sagemaker): free tier Q Chat with auto-login for iam users and login option for pro tier users

ahusseinali:
## Problem
AmazonQ
- IAM users didn't have access to Q Chat (Free Tier)
- Sagemaker IAM users weren't automatically logged in using their Sagemaker studio credentials.
- For Sagemaker users, login screen was disabled & users didn't have access to chat or code completion.
- Sagemaker Pro Tier users code completion was broken & didn't generate any recommendations even after successful login.
- Q Icon hides from Activity Bar when Toolkit extension is installed

## Solution
- Add new Q Developer Client to enable Free tier chat.
- Terminate all the conditions that used to force chat to get disabled for IAM users
- Terminate the conditions that used to hide login screen for Sagemaker users
- Login Sagemaker IAM users automatically using their environment credentials (Free tier)
- Allow Sagemaker SSO pro tier users to login using the login screen.

---


package-lock.json
nkomonen-amazon, line 1: There is a large change here and it is not clear what triggered it. Are you able to extract this related change to an isolated commit so we can better understand

ahusseinali: I'm not sure what triggerred it either, but I believe it's the dependencies of the new Q developer client I added in src.gen

packages/amazonq/src/api.ts
```diff
@@ -12,7 +13,11 @@ export default {
     chatApi: {
         async chat(request: GenerateAssistantResponseRequest): Promise<GenerateAssistantResponseCommandOutput> {
             const chatSession = new ChatSession()
-            return chatSession.chat(request)
+            return chatSession.chatSso(request)
+        },
+        async chatIam(request: SendMessageRequest): Promise<SendMessageCommandOutput> {
+            const chatSession = new ChatSession()
+            return chatSession.chatIam(request)
```
justinmk3, line 20: this can be an optional parameter on `chat` instead of an entirely new function.

packages/amazonq/.changes/next-release/Feature-a007f02d-d737-4429-9b7d-245aa30d879c.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Enable Free Tier Chat for IAM users & Fix code completion for Sagemaker Pro Tier users"
```
hayemaxi, line 3: Split this into 2 changelog items, eg:
```
{
	"type": "Feature",
	"description": "Chat: Enable free tier for IAM users in Sagemaker."
}
```
and (this second one may need to be more descriptive)
```
{
	"type": "Bug Fix",
	"description": "Inline: Code completion not working for Sagemaker Pro Tier users."
}
```

src.gen/@amzn/amazon-q-developer-streaming-client/package.json
```diff
@@ -0,0 +1,91 @@
+{
+    "name": "@amzn/amazon-q-developer-streaming-client",
```
hayemaxi, line 2: What is the purpose of this package? Why can we not use `src.gen/@amzn/codewhisperer-streaming`?

packages/toolkit/package.json
```diff
@@ -704,7 +704,7 @@
                 {
                     "id": "aws.amazonq.codewhisperer",
                     "name": "%AWS.amazonq.codewhisperer.title%",
-                    "when": "!isCloud9 && !aws.isSageMaker && !aws.toolkit.amazonq.dismissed && !aws.explorer.showAuthView"
```
hayemaxi, line 707: Why was this changed? I don't see the `setContext` value being changed. This is not a function call.
See https://github.com/aws/aws-toolkit-vscode/blob/5e8ad316718661bed35230e3596c13e5e2b53f30/packages/core/src/extensionNode.ts#L121-L126

packages/core/src/shared/clients/qDeveloperChatClient.ts
```diff
@@ -0,0 +1,25 @@
...
+// Create a client for featureDev streaming based off of aws sdk v3
+export async function createQDeveloperStreamingClient(): Promise<QDeveloperStreaming> {
+    const cwsprConfig = getCodewhispererConfig()
+    const credentials = await AuthUtil.instance.getCredentials()
+    const streamingClient = new QDeveloperStreaming({
+        region: cwsprConfig.region,
+        endpoint: cwsprConfig.endpoint,
+        credentials: credentials,
+        customUserAgent: getUserAgent(),
+        // SETTING max attempts to 0 FOR BETA. RE-ENABLE FOR RE-INVENT
```
hayemaxi, line 20:
```suggestion
        // TODO: SETTING max attempts to 0 FOR BETA. RE-ENABLE FOR RE-INVENT
```
Make sure you have an ticket or some way to track this!

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -13,10 +13,12 @@ import {
     QuickActionMessage,
 } from '../../../view/connector/connector'
 import { EditorContextCommandType } from '../../../commands/registerCommands'
+import { ChatResponseStream as qdevChatResponseStream } from '@amzn/amazon-q-developer-streaming-client'
 import {
+    ChatResponseStream as cwChatResponseStream,
     CodeWhispererStreamingServiceException,
-    GenerateAssistantResponseCommandOutput,
     SupplementaryWebLink,
+    __MetadataBearer,
```
hayemaxi, line 21: Why important this? Seems like its supposed to be private because of the `__`. Is there something else that can be used or should it be updated to not imply secrecy?

package-lock.json
leigaol: That new generated client won't result in new package-lock.json.

src.gen/@amzn/amazon-q-developer-streaming-client/package.json
hayemaxi: To answer this, please update https://github.com/aws/aws-toolkit-vscode/blob/master/docs/build.md

packages/core/src/auth/activation.ts
```diff
@@ -3,16 +3,32 @@
...
+import { EcsCredentialsProvider } from './providers/ecsCredentialsProvider'
+import { EnvVarsCredentialsProvider } from './providers/envVarsCredentialsProvider'
+
+interface SagemakerCookie {
+    authMode?: string
+}
 
 export async function initialize(loginManager: LoginManager): Promise<void> {
+    if (isAmazonQ() && isSageMaker()) {
+        const result = (await vscode.commands.executeCommand('sagemaker.parseCookies')) as SagemakerCookie
```
hayemaxi, line 27: Where does this command come from? Is it already registered in VSC by sagemaker? If so add a comment to mention this.

packages/core/src/auth/activation.ts
```diff
@@ -3,16 +3,32 @@
...
 import { isInDevEnv } from '../shared/vscode/env'
 import { isWeb } from '../shared/extensionGlobals'
+import { CredentialsProviderManager } from './providers/credentialsProviderManager'
+import { SharedCredentialsProviderFactory } from './providers/sharedCredentialsProviderFactory'
+import { Ec2CredentialsProvider } from './providers/ec2CredentialsProvider'
+import { EcsCredentialsProvider } from './providers/ecsCredentialsProvider'
+import { EnvVarsCredentialsProvider } from './providers/envVarsCredentialsProvider'
+
+interface SagemakerCookie {
+    authMode?: string
```
hayemaxi, line 22: What are the possible values? Make this a stricter type than just `string`, e.g.
```
type AuthMode = 'Sso' | 'Iam'
interface SagemakerCookie {
    authMode?: AuthMode
}
```

packages/core/src/auth/activation.ts
```diff
@@ -25,6 +41,12 @@ export async function initialize(loginManager: LoginManager): Promise<void> {
     await showManageConnectionsOnStartup()
 }
 
+function initializeCredentialsProviderManager() {
+    const manager = CredentialsProviderManager.getInstance()
+    manager.addProviderFactory(new SharedCredentialsProviderFactory())
+    manager.addProviders(new Ec2CredentialsProvider(), new EcsCredentialsProvider(), new EnvVarsCredentialsProvider())
+}
```
hayemaxi, line 48: This is a direct copy paste of this already existing function. Please export and re-use this existing function. You will need to move it to a shared file, e.g. `core/src/auth/utils.ts`
https://github.com/aws/aws-toolkit-vscode/blob/5e8ad316718661bed35230e3596c13e5e2b53f30/packages/core/src/extensionNode.ts#L313-L317

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -60,7 +60,7 @@ export const isValidCodeWhispererCoreConnection = (conn?: Connection): conn is C
     }
 
     if (isSageMaker()) {
-        return isIamConnection(conn)
+        return isIamConnection(conn) || (isSsoConnection(conn) && hasScopes(conn, codeWhispererCoreScopes))
```
hayemaxi, line 63: Can we use `amazonQScopes`? Does sagemaker SSO auth allow for the complete Q scope set?
```suggestion
        return isIamConnection(conn) || (isSsoConnection(conn) && hasScopes(conn, amazonQScopes))
```
It is code smell that we have these scope subsets, so it is easy to confuse which one needs to be used/checked. We have plans to fix this.

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -70,6 +70,10 @@ export const isValidCodeWhispererCoreConnection = (conn?: Connection): conn is C
 }
 /** Superset that includes all of CodeWhisperer + Amazon Q */
 export const isValidAmazonQConnection = (conn?: Connection): conn is Connection => {
+    if (isSageMaker() && isAmazonQ()) {
```
hayemaxi, line 73: Check isn't needed.
```suggestion
    if (isSageMaker()) {
```

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -70,6 +70,10 @@ export const isValidCodeWhispererCoreConnection = (conn?: Connection): conn is C
...
+        return isIamConnection(conn) || (isSsoConnection(conn) && hasScopes(conn, amazonQScopes))
+    }
+
     return (
         (isSsoConnection(conn) || isBuilderIdConnection(conn)) &&
```
hayemaxi, line 78: Simplify
```suggestion
    return (
        (isAnySsoConnection(conn) || (isIamConnection(conn) && isSageMaker()) &&
```

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -442,9 +446,6 @@ export class AuthUtil {
         if (conn === undefined) {
             return buildFeatureAuthState(AuthStates.disconnected)
         }
-        if (!isSsoConnection(conn)) {
-            throw new ToolkitError(`Connection "${conn.id}" is not a valid type: ${conn.type}`)
-        }
```
hayemaxi, line 447: Instead of deleting this you can add the check
```
if (!isSsoConnection(conn) && !isSageMaker()) { ... }
```

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -453,7 +454,7 @@ export class AuthUtil {
             return state
         }
 
-        if (isBuilderIdConnection(conn) || isIdcSsoConnection(conn)) {
+        if (isBuilderIdConnection(conn) || isIdcSsoConnection(conn) || isSageMaker()) {
             if (isValidCodeWhispererCoreConnection(conn)) {
```
hayemaxi, line 458: I think you need to update this because the function will return incorrectly for sagemaker SSO connections. The same problem affects `isValidAmazonQConnection` function called around line 72 of this file but I wasn't able to put a comment on it.
https://github.com/aws/aws-toolkit-vscode/blob/5e8ad316718661bed35230e3596c13e5e2b53f30/packages/core/src/codewhisperer/util/authUtil.ts#L63

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -656,12 +655,24 @@ export class ChatController {
...
             this.messenger.sendInitalStream(tabID, triggerID)
             this.telemetryHelper.setConversationStreamStartTime(tabID)
-            response = await session.chat(request)
+            if (isSsoConnection(AuthUtil.instance.conn)) {
+                const { $metadata, generateAssistantResponseResponse } = await session.chatSso(request)
+                response = {
+                    $metadata: $metadata,
+                    message: generateAssistantResponseResponse,
+                }
+            } else {
```
hayemaxi, line 669: If we aren't connected (AuthUtil.instance.conn is undefined), we will take the IAM path. Will this provide any confusing errors or log messages?

packages/core/src/auth/activation.ts
ahusseinali: Correct, this command is registered in Sagemaker code editor. I'll add a comment

packages/core/src/auth/activation.ts
ahusseinali: Your assumption is correct. will adjust it

packages/core/src/auth/activation.ts
ahusseinali: will do

packages/core/src/codewhisperer/util/authUtil.ts
ahusseinali: No it doesn't, but tackling the scopes is a separate issue that I'll address later on. For now, if user login with their SSO credentials, they will go through the normal login flow which appends the full set of credentials for them. I can omit the scope check from here though as it doesn't make a difference for Sagemaker users.

packages/core/src/auth/activation.ts
nkomonen-amazon: Lets make a new sagemaker folder in `src/shared/sagemaker` for all custom sagemaker code so that it is contained within a single place.

packages/core/src/codewhisperer/util/authUtil.ts
ahusseinali: For now, I'll simplify the condition and put in a separate function for readability. it ignores any scope checking as we're handling deactivating features not supported by sagemaker in a separate change

packages/core/src/codewhisperer/util/authUtil.ts
ahusseinali: I extracted sagemaker condition into a separate function for better readability

packages/core/src/codewhisperer/util/authUtil.ts
ahusseinali: but this change is not just for Sagemaker, the new client enables chat for both Sso and Iam connections.

packages/core/src/codewhisperer/util/authUtil.ts
ahusseinali: I updated that function logic to ignore scopes for Sagemaker connections

packages/core/src/codewhispererChat/controllers/chat/controller.ts
ahusseinali: we never get to this point if the Auth is not valid. 

In the same function call, this code block guards against this

```
const credentialsState = await AuthUtil.instance.getChatAuthState()

if (
   !(credentialsState.codewhispererChat === 'connected' && credentialsState.codewhispererCore === 'connected')
) {
    await this.messenger.sendAuthNeededExceptionMessage(credentialsState, tabID, triggerID)
    return
}
 ```

File: packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts (hunk 6 above)
Comment (ahusseinali) [reply packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts:21]: You're right, I wanted to look for something that is agnostic to SendMessage and GenerateAssistantResponse response types and this was the minimal common type.

I cannot update the type itself as it's imported from the generated client. I can create a custom type for the properties I need only instead. They're pretty basic

File: packages/core/src/shared/clients/qDeveloperChatClient.ts (hunk 5 above)
Comment (ahusseinali) [reply packages/core/src/shared/clients/qDeveloperChatClient.ts:20]: This is copied over from the existing codewhisperer client. I assumed whatever fix will apply to one should apply to the other. Shall I remove it all together? I lack context on why it was set to 0 for Beta!

File: src.gen/@amzn/amazon-q-developer-streaming-client/package.json (hunk 3 above)
Comment (ahusseinali) [reply src.gen/@amzn/amazon-q-developer-streaming-client/package.json:2]: As discussed offline, this client supports Sigv4 credentials instead of bearer token. this is mandatory to support both IAM and SSO users.

The instructions to generate these clients will be updated in the same quip doc referenced in this .md file
I updated the wording to reflect the new changes

File: packages/toolkit/package.json (hunk 4 above)
Comment (ahusseinali) [reply packages/toolkit/package.json:707]: Good catch! That was a mistake. I believe I removed the whole clause and then attempted to undo it. Fixed it.

File: package-lock.json
Comment (ahusseinali) [reply package-lock.json:1]: At this point I'm clueless as to why the package.json file was changed. But I see many of the dependencies added there are also part of the newly added client package-lock.json. this is why I suspected they're related.

I inspected the dependencies added / changed further and they're only relevant to the new client code

File: packages/core/src/codewhisperer/util/authUtil.ts (hunk 13 above)
Comment (hayemaxi) [reply packages/core/src/codewhisperer/util/authUtil.ts:447]: > but this change is not just for Sagemaker

But this isn't a supported feature yet in all of the code yet right? That is a big product change to the Q extension. AFAIK there is no way to get IAM credentials outside of sagemaker.

Also, for a change like this I would expect unit or integ test updates/additions if possible.

File: packages/core/src/codewhisperer/util/authUtil.ts [hunk 16]
```diff
@@ -59,21 +62,19 @@ export const isValidCodeWhispererCoreConnection = (conn?: Connection): conn is C
...
 /** Superset that includes all of CodeWhisperer + Amazon Q */
 export const isValidAmazonQConnection = (conn?: Connection): conn is Connection => {
     return (
-        (isSsoConnection(conn) || isBuilderIdConnection(conn)) &&
-        isValidCodeWhispererCoreConnection(conn) &&
-        hasScopes(conn, amazonQScopes)
+        isValidSageMakerConnection(conn) ||
+        ((isSsoConnection(conn) || isBuilderIdConnection(conn)) &&
+            isValidCodeWhispererCoreConnection(conn) &&
+            hasScopes(conn, amazonQScopes))
```
hayemaxi, line 77: `isValidSageMakerConnection` returns true if it is an IAM or SSO connection only - it doesn't seem to do any checks about sagemaker. That isn't a valid Q connection, we have to do a scope check.
 This means this whole block will return if the connection is not undefined for anyone running the extension, not just in sagemaker.

Based on the current implementation of `isValidCodeWhispererCoreConnection`, this should work:
```suggestion
        isValidCodeWhispererCoreConnection(conn) && hasScopes(conn, amazonQScopes)
```

packages/core/src/codewhisperer/util/authUtil.ts
hayemaxi: Looks to be broken still, see https://github.com/aws/aws-toolkit-vscode/pull/5858/files#r1819088749

packages/core/src/shared/clients/qDeveloperChatClient.ts
hayemaxi: I see, then we can ignore it for now and follow up with the original contributor.

packages/core/src/codewhisperer/util/authUtil.ts
ahusseinali: The Sagemaker IAM connection doesn't have amazonQScopes set, it follows a different path from the traditional login flow as it gets loaded through credentials provider.

I attempted to limit the scope for the Sso connection for that check before and received push back on it. I can bring it back here, but it will be specifically for Sso rather than both Sso and IAM

packages/core/src/codewhisperer/util/authUtil.ts
ahusseinali: Ack, will adjust it now

hayemaxi (approved):
lgtm pending team approval

packages/core/src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -158,8 +159,7 @@ export class RecommendationHandler {
         autoTriggerType?: CodewhispererAutomatedTriggerType,
         pagination: boolean = true,
         page: number = 0,
-        isSM: boolean = isSageMaker(),
-        retry: boolean = false
+        generate: boolean = isIamConnection(AuthUtil.instance.conn)
```
leigaol, line 162: This should no longer be a single function for both IAM and bearer token code path. The number of if-else is too much to comprehend. Not a blocker for this PR

leigaol (approved):

packages/core/src/codewhisperer/service/recommendationHandler.ts
leigaol: Please do lots of regression testing post merge

src.gen/@amzn/amazon-q-developer-streaming-client/package.json
justinmk3: The quip wasn't updated. I added a "qdevstreaming" section.

packages/core/src/codewhisperer/service/recommendationHandler.ts
leigaol: Test /dev and /transform

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -656,12 +655,24 @@ export class ChatController {
...
+                const { $metadata, sendMessageResponse } = await session.chatIam(request as SendMessageRequest)
```
bhadrip, line 670: `source="IDE"`
we need to set the request for the right routing in Q Dev

packages/core/src/auth/activation.ts
justinmk3: This isn't a long-term approach. The Toolkit/Q extensions already have an api to accept connections. If needed, they can be enhanced to accept tokens/credentials, so the caller (SM in this case) can inject credentials into the extension.

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -656,12 +655,24 @@ export class ChatController {
...
+                response = {
+                    $metadata: $metadata,
+                    message: sendMessageResponse,
+                }
+            }
             this.telemetryHelper.recordEnterFocusConversation(triggerEvent.tabID)
```
bhadrip, line 676: what kind of telemetry or error mechanism we can use to understand customer behavior ?

packages/core/src/auth/activation.ts
```diff
@@ -3,16 +3,28 @@
...
+interface SagemakerCookie {
+    authMode?: 'Sso' | 'Iam'
+}
+
 export async function initialize(loginManager: LoginManager): Promise<void> {
+    if (isAmazonQ() && isSageMaker()) {
+        // The command `sagemaker.parseCookies` is registered in VS Code Sagemaker environment.
+        const result = (await vscode.commands.executeCommand('sagemaker.parseCookies')) as SagemakerCookie
+        if (result.authMode !== 'Sso') {
+            initializeCredentialsProviderManager()
```
bhadrip, line 25: how does this refactor affect toolkit ? In sagemaker studio sso mode, toolkit will still work in IAM mode.

bhadrip (commented):
how do we handle graceful failure of features that are not supported in free tier IAM mode

bhadrip (commented):
are there no unit tests for the changes ?

src.gen/@amzn/amazon-q-developer-streaming-client/src/commands/SendMessageCommand.ts
```diff
@@ -0,0 +1,652 @@
...
+ *     chatTriggerType: "MANUAL" || "DIAGNOSTIC" || "INLINE_CHAT", // required
+ *     customizationArn: "STRING_VALUE",
+ *   },
+ *   profileArn: "STRING_VALUE",
+ *   source: "STRING_VALUE",
+ *   dryRun: true || false,
+ * };
+ * const command = new SendMessageCommand(input);
+ * const response = await client.send(command);
+ * // { // SendMessageResponse
```
justinmk3, line 343: did the generator comment this out (and the lines below)? that is strange

justinmk3:
please add a `src.gen/@amzn/.gitignore` file that will avoid mistakes like https://github.com/aws/aws-toolkit-vscode/pull/5858/commits/972e2add2cfac72c1bf97f2d8d7f2c72ff2649e9 in the future.

packages/core/src/auth/activation.ts
ahusseinali: The higher level condition is scoped to amazonq, so it shouldn't impact aws toolkit

leigaol (approved):

ahusseinali:
> are there no unit tests for the changes ?

All the files I touched didn't have unit tests related to the changes. I plan to improve the coverage of this package as the code paths (especially for auth and API calls) are very critical

ahusseinali:
> please add a `src.gen/@amzn/.gitignore` file that will avoid mistakes like [972e2ad](https://github.com/aws/aws-toolkit-vscode/commit/972e2add2cfac72c1bf97f2d8d7f2c72ff2649e9) in the future.

Added rule

hayemaxi:
CI is failing: 
```
Error: src/inlineChat/provider/inlineChatProvider.ts(123,38): error TS2339: Property 'chat' does not exist on type 'ChatSession'.
Error: src/inlineChat/provider/inlineChatProvider.ts(125,102): error TS18048: 'response' is possibly 'undefined'.
Error: src/inlineChat/provider/inlineChatProvider.ts(126,21): error TS18048: 'response' is possibly 'undefined'.
```
another PR merged to master is using old function `chat` which is replaced in this PR by `chatSso` and `chatIam`. The logic will need to be updated to determine which one to call.

bhadrip (approved):

gogakoreli (approved):
Verified in Sagemaker

PR #5859, opened 2024-10-24
refactor(cw): avoid duplicating language list. 

Hweinstock:
## Problem
This is the longest case of code duplication in the repo from `jscpd` (46 lines!). 

## Solution
Remove the duplication case, remove redundant if statement. 

The switch statements could also be refactored into something with significantly less code in the future. Want to keep this PR focused on reducing code duplication. 

---

### DIFF
--- a/packages/core/src/codewhispererChat/editor/context/file/languages.ts
+++ b/packages/core/src/codewhispererChat/editor/context/file/languages.ts
@@ -117,10 +117,7 @@ export function extractLanguageNameFromFile(file: TextDocument): string | undefi
 export function extractAdditionalLanguageMatchPoliciesFromFile(file: TextDocument): Set<string> {
     const languageId = file.languageId
 
-    if (languageId === undefined) {
-        return new Set<string>()
-    }
-    if (validLanguageIds.includes(languageId)) {
+    if (languageId === undefined || validLanguageIds.includes(languageId)) {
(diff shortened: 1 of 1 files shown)


packages/core/src/codewhispererChat/editor/context/file/languages.ts
```diff
@@ -117,54 +117,7 @@ export function extractLanguageNameFromFile(file: TextDocument): string | undefi
...
-            'coffeescript',
-            'clojure',
-            'bibtex',
-            'abap',
-        ].includes(languageId)
-    ) {
+    if (languageId === undefined || defaultLanguages.includes(languageId)) {
         return new Set<string>()
     }
     switch (languageId) {
```
hayemaxi, line 123: out of scope, but man this switch statement is brutal. Could probably cut the character count in half by using pass-through and not typing `new Set<string>` everywhere.

hayemaxi (approved):

packages/core/src/codewhispererChat/editor/context/file/languages.ts
Hweinstock: Yeah I thought the same thing lol. These functions are basically just a giant map with two values.

jpinkney-aws (approved):

PR #5862, opened 2024-10-24
ci(lint): expand jscpd to use branch comparison

Hweinstock:
## Problem
Follow up to: https://github.com/aws/aws-toolkit-vscode/pull/5833

A static threshold is useful, but if the threshold fails, we want to be able to see what the new duplicated code is.

## Solution
Run `jscpd` on the current branch, and if it is above the set threshold, we fail (same as before). 

Now, we also filter the resulting report to contain only clones from the files changed with respect to the base branch. Then, if the filtered report is non-empty, i.e. one of the changed files has a clone, fail and display the clones in CI. 

Adding this extra step allows us to know when we are adding a clone that was not otherwise in the codebase. It also allows us to see this clone in CI, before the PR is merged, giving the developer a chance to remove it. 

As an example, see commit (https://github.com/aws/aws-toolkit-vscode/pull/5862/commits/a8e48600faed259114264643d0a974d0e8a24749) where a copy-pasted function is added, and notice that CI fails and displays only the section of the report relevant to this clone. Then, once we remove this work in the following commit (https://github.com/aws/aws-toolkit-vscode/pull/5862/commits/53d3973acce7171d8e7b36bd10d8b34c3d524274), the check passes. 

Included in this PR, is that it runs only on `pull_request` and no longer on `push`. 
## Notes
- In developing this, it became apparent that `jscpd` is not perfect and was not able to detect the clone added in (https://github.com/aws/aws-toolkit-vscode/pull/5862/commits/e7357c0da2489c55ada77a54293738e9bc56a865). Will need to dig deeper into their algorithm (based on https://en.wikipedia.org/wiki/Rabin%E2%80%93Karp_algorithm) and implementation (https://github.com/kucherenko/jscpdto) to understand why this is missed, and if it presents a larger problem. 
- Also, if a clone already exists in a file that is changed in the PR, this check will fail. In this case, the options are to ignore the check, mark the case as an exception, or fix the clone. To encourage developers to not repeatedly ignore the check, we still employ a static threshold via https://github.com/aws/aws-toolkit-vscode/pull/5833. 
- If the branch hasn't been rebased up to date, it can pick up clones from unrelated changes. 

## Future Work
- We want to lower the threshold as much as possible by removing duplicates and marking exceptions when necessary. 
- More experience using this tool will help us find a `minLines` argument that provides the desired sensitivity. This argument allows us to ignore clones under `minLines` in length. Right now it is 3, which is likely too granular, but better to start there and loosen as needed. 
- Right now the CI output of the clones is very verbose. Digging through this could be time consuming, especially when many clones are added. Could be worth filtering it to provide a more ergonomic experience. 

---


.github/workflows/jscpd.json
```diff
@@ -1,7 +1,9 @@
 {
     "pattern": "packages/**/*.ts",
-    "ignore": ["**node_modules**", "**dist**"],
+    "ignore": ["**node_modules**", "**dist**", "*/scripts/*"],
```
justinmk3, line 3: I wonder if this should be `**/scripts/**`

Hweinstock: Yep, verified you are correct. Here's the followup: https://github.com/aws/aws-toolkit-vscode/pull/5930

PR #5871, opened 2024-10-25
BugFix: fix perceivedLatency being calculated incorrectly

andrewyuq:
The perceivedLatency is currently calculateand and sent when we send STE, by then lastUserModificationTime is no longer accurate. Instead calculate perceivedLatency when we show the suggestions and use the lastUserModificationTime at that time to be accurate.

JB PR: https://github.com/aws/aws-toolkit-jetbrains/pull/5013
## Solution

---


andrewyuq:
```
 1) SamCliLocalInvokeInvocation
       invokes `sam local` with args:
     Error: Test length exceeded max duration: 30 seconds
```
This error doesn't look like is related to the change itself

hayemaxi (approved):

jpinkney-aws (approved):

PR #5876, opened 2024-10-28
feat(notifications): notifications rendering

tomcat323:
### Goal:
We want notifications to do things when they are clicked on in the [panel](https://taskei.amazon.dev/tasks/IDE-14932) or are received by the IDE. These actions will must be able to be specified by the notifications [JSON schema ](https://quip-amazon.com/nmB0AxFXuK3i/In-IDE-Notifications#temp:C:EMZ18c352ef42a546ebbe0f9e7c4). We can update the JSON schema to handle certain actions if necessary.

1. Once received by the IDE, notifications should display a toast notification. This toast notification should automatically be displayed only once.
- We can track this in global state (see NotificationsController)
- This toast may shown a truncated form of the full notification text (description in the schema). This truncation can be automatic or provided by a new schema key, e.g. descriptionPreview.
- There should be an option to instead display this as a blocking modal, for extreme circumstances only.
- This toast notification will not include any buttons, 
2. Clicking on the notification in the side panel will do 1 of 3 things:
- Open a url link
- Open a new text document with contents of description
- Open a blocking modal with contents of description and buttons
- The modal can display buttons, e.g. 'update and reload', 'open external link', 'open in new text window', that perform different functions respectively

### Results:

> Pop up toast notification on receiving a new notification
Screenshot 2024-10-28 at 1 20 07 PM

> Blocking pop up modal with options to restart and reload
Screenshot 2024-10-28 at 1 17 47 PM

> Open URL link
Screenshot 2024-10-28 at 1 18 32 PM

> Side view pop up read-only txt document
Screenshot 2024-10-28 at 1 18 52 PM
---


packages/core/src/notifications/controller.ts
```diff
@@ -42,7 +42,7 @@ export class NotificationsController {
 
     constructor(
         private readonly notificationsNode: NotificationsNode,
-        private readonly fetcher: NotificationFetcher = new RemoteFetcher()
+        private readonly fetcher: NotificationFetcher = new LocalFetcher()
```
hayemaxi, line 45: It's probably like this for testing but make sure to revert it when you are ready to merge.

packages/core/src/notifications/controller.ts
```diff
@@ -121,8 +122,25 @@ export class NotificationsController {
...
+        // Compare and find if there's any notifications newly added
+        const addedNotifications = newNotifications.filter((n: any) => !currentNotificationIds.has(n.id))
+
+        if (addedNotifications.length > 0) {
+            getLogger('notifications').info(
+                'New notifications received for category %s, ids: %s',
+                category,
+                addedNotifications.map((n: any) => n.id).join(', ')
+            )
+            await this.onReceiveNotifications(addedNotifications)
```
hayemaxi, line 142: We will only want to run this for notifications that pass the rule engine, so we will need to do that outside this function, e.g. in `displayNotifications`

packages/core/src/notifications/controller.ts
```diff
@@ -121,8 +122,25 @@ export class NotificationsController {
...
+        }
 
         this.state[category].payload = JSON.parse(response.content)
```
hayemaxi, line 145: We already parse it on line 126 above, so lets re-use the result here.

packages/core/src/notifications/types.ts
```diff
@@ -51,9 +51,21 @@ export interface UIRenderInstructions {
...
+            descriptionPreview?: string // optional property for toast
         }
     }
-    // TODO actions
+    onRecieve: string
+    onClick: {
+        type: string
+        url?: string // optional property for 'openUrl'
+    }
+    buttons?: Array<{
```
hayemaxi, line 62: `uiRenderingInstructions` will be pretty IDE specific (example: we are adding onReceive and onClick to encapsulate possiblities with VSC specifically), but we should try to align to the original schema as much as possible to not cause confusion. Let's keep this as `actions` instead, even though buttons is probably a better term.

packages/core/src/notifications/panelNode.ts
```diff
@@ -129,12 +133,105 @@ export class NotificationsNode implements TreeNode {
...
+            }
+
+            await vscode.window.showTextDocument(document, options)
+        } catch (error) {
+            throw new ToolkitError(`Error showing text document: ${error}`)
+        }
+    }
+
+    public async renderModal(notification: ToolkitNotification) {
+        getLogger('notifications').info('rendering modal ... ')
```
hayemaxi, line 191: consider adding the notification ID to the log message. Also, maybe `.verbose` instead of `.info`

packages/core/src/notifications/controller.ts
tomcat323: good point

packages/core/src/notifications/controller.ts
```diff
@@ -136,6 +154,23 @@ export class NotificationsController {
         )
     }
 
+    private async onReceiveNotifications(notifications: ToolkitNotification[]) {
+        for (const notification of notifications) {
+            switch (notification.uiRenderInstructions.onRecieve) {
+                case 'modal':
+                    // Handle modal case
+                    void this.notificationsNode.renderModal(notification)
```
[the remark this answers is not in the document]
hayemaxi: Nevermind, I thought `renderModal` depended on `onClick`

packages/core/src/notifications/controller.ts
tomcat323: Create a new state `newlyReceived` to handle this

packages/core/src/notifications/controller.ts
```diff
@@ -98,6 +98,16 @@ export class NotificationsController {
...
+
+        // Process on-receive behavior for newly received notifications that passes rule engine
+        const newlyReceivedToDisplay = [...startUp, ...emergency].filter((n) => this.state.newlyReceived.includes(n.id))
+        if (newlyReceivedToDisplay.length > 0) {
+            await this.notificationsNode.onReceiveNotifications(newlyReceivedToDisplay)
+            // remove displayed notifications from newlyReceived
+            this.state.newlyReceived = this.state.newlyReceived.filter(
+                (id) => !newlyReceivedToDisplay.some((n) => n.id === id)
+            )
+        }
```
hayemaxi, line 110: call `await this.writeState()` here since we are modifying the state.

packages/core/src/test/notifications/rendering.test.ts
```diff
@@ -0,0 +1,235 @@
...
+                    displayText: {
+                        'en-US': 'Read More',
+                    },
+                },
+            ],
+        },
+    }
+}
+
+function getOpenExternalStub(): sinon.SinonStub {
```
hayemaxi, line 225: This is already stubbed in `globalSetup.test.ts`

packages/core/src/test/notifications/rendering.test.ts
```diff
@@ -0,0 +1,235 @@
...
+import * as sinon from 'sinon'
+import assert from 'assert'
+import { ToolkitNotification } from '../../notifications/types'
+import fs from '../../shared/fs/fs'
+import path from 'path'
+import { tempDirPath } from '../../shared/filesystemUtilities'
+import { getTestWindow } from '../shared/vscode/window'
+import { panelNode } from './controller.test'
+
+describe('Notifications Rendering', function () {
```
hayemaxi, line 16: This suite is using a lot of stubs... why? We should have methods to detect opened messages and text documents.

packages/core/src/notifications/panelNode.ts
```diff
@@ -187,49 +187,70 @@ export class NotificationsNode implements TreeNode {
...
+    /**
+     * Renders information window with the notification's content and buttons.
+     * Can be either a blocking modal or a bottom-right corner toast
+     * Handles the button click actions based on the button type.
+     */
+    public async showInformationWindow(notification: ToolkitNotification, type: string = 'toast') {
+        const isModal = type === 'modal'
+
+        // modal has to have defined actions(buttons)
+        if (!notification.uiRenderInstructions.actions && isModal) {
```
hayemaxi, line 199: Why do we need to enforce this? A modal with no actions can just have a simple 'OK' button that is built in by VSC.

packages/core/src/notifications/controller.ts
tomcat323: is that really needed? `writeState` doesn't handle the newlyReceived states, and that's the only thing we are modifying here

packages/core/src/notifications/panelNode.ts
tomcat323: In that case yea I guess we don't have to enforce this.

packages/core/src/notifications/panelNode.ts
```diff
@@ -187,49 +187,70 @@ export class NotificationsNode implements TreeNode {
...
-                            await vscode.env.openExternal(vscode.Uri.parse(buttons.url))
+                        if (selectedButton.url) {
+                            await vscode.env.openExternal(vscode.Uri.parse(selectedButton.url))
+                        } else {
+                            throw new ToolkitError('url not provided')
                         }
                         break
                     default:
-                        getLogger().warn(`Unhandled button type: ${buttons.type}`)
+                        throw new ToolkitError('button action not defined')
```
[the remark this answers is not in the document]
tomcat323: This is not for user not taking action. This is for if the actions defined in the JSON have something that can not be handled by out processing logic. (Maybe a typo or something)

packages/core/src/test/notifications/rendering.test.ts
hayemaxi: > that is why I created this function because we can't have two stubs with the same openExternal command
I don't follow.

> The other way is to export the stub from globalSetup.test.ts, is that more preferred?
Its already exported and is used by other tests

packages/core/src/notifications/controller.ts
hayemaxi: Yes it does, it writes the whole object to global state: `await globals.globalState.update(this.storageKey, this.state)`

You are updating newlyReceived in `fetchNotifications` and then writing the state. But once you clear it here you are never writing the cleared state back. So the newlyReceived state will never be cleared out and if the IDE is restarted it seems like this will re-run for notifications we already received.

hayemaxi (approved):

packages/core/src/notifications/panelNode.ts
```diff
@@ -126,15 +130,125 @@ export class NotificationsNode implements TreeNode {
...
+
+            const options: vscode.TextDocumentShowOptions = {
+                viewColumn: vscode.ViewColumn.Beside,
+                preserveFocus: true,
+                preview: true,
+            }
+
+            await vscode.window.showTextDocument(document, options)
+        } catch (error) {
+            throw new ToolkitError(`Error showing text document: ${error}`)
```
justinmk3, line 186: Just let the error bubble up. This re-thrown error isn't adding any useful information, and it discards potentially useful info from the original error.

packages/core/src/notifications/panelNode.ts
```diff
@@ -126,15 +130,125 @@ export class NotificationsNode implements TreeNode {
...
+
+            if (await fs.existsFile(tempFilePath)) {
+                // If file exist, make sure it has write permission (0o644)
+                await fs.chmod(tempFilePath, 0o644)
+            }
+
+            await fs.writeFile(tempFilePath, content)
+
+            // Set the file permissions to read-only (0o444)
+            await fs.chmod(tempFilePath, 0o444)
```
justinmk3, line 173: why save the file instead of showing a "virtual document" ? is the file used later?

packages/core/src/notifications/panelNode.ts
```diff
@@ -126,15 +130,125 @@ export class NotificationsNode implements TreeNode {
...
+                        await this.showReadonlyTextDocument(
+                            notification.uiRenderInstructions.content['en-US'].description
+                        )
+                        break
+                    case 'updateAndReload':
+                        await this.updateAndReload(notification.displayIf.extensionId)
+                        break
+                    case 'openUrl':
+                        if (selectedButton.url) {
+                            await vscode.env.openExternal(vscode.Uri.parse(selectedButton.url))
```
justinmk3, line 230: use our `openUrl` wrapper unless the URL contains PII https://github.com/aws/aws-toolkit-vscode/blob/cee0fa3c1ab320f00a4b43fa3e57a44efe3370e3/packages/core/src/shared/utilities/vsCodeUtils.ts#L217

packages/core/src/notifications/panelNode.ts
tomcat323: The file is not used later, but since it's in the temp folder it shouldn't matter right?

Also for virtual document are you referring to something like [this](https://code.visualstudio.com/api/extension-guides/virtual-documents)? It seems a bit too complicated for what we need here.

PR #5892, opened 2024-10-29
fix(amazonq): Remove warning on empty inputs to inline chat

grant0417:
This PR removes the notification when a query is empty, this was not intended to be added and the menu should simply close when the input is empty.

### DIFF
--- a/packages/amazonq/src/inlineChat/controller/inlineChatController.ts
+++ b/packages/amazonq/src/inlineChat/controller/inlineChatController.ts
@@ -191,7 +191,7 @@ export class InlineChatController {
             })
             .then(async (query) => {
                 if (!query || query.trim() === '') {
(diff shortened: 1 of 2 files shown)


zixlin7 (approved):

packages/amazonq/src/inlineChat/controller/inlineChatController.ts
```diff
@@ -191,7 +191,6 @@ export class InlineChatController {
             })
             .then(async (query) => {
                 if (!query || query.trim() === '') {
-                    void vscode.window.showWarningMessage('Amazon Q: Instructions for cannot be empty')
                     return
```
justinmk3, line 194: should it be logged, or is this a situation that we are certain won't need troubleshooting?

grant0417: Added log statement 👍

hayemaxi (approved):

PR #5894, opened 2024-10-29
feat(cwl): Add LiveTail StatusBarItems

keeganirby:
## Problem
As LiveTail session is running, we want to display to the user metadata about their session. 

## Solution
While a livetail session is the active editor, in the bottom status bar, we will display the duration the session has been running for, how many events per/s, and if the log data is being sampled or not. 

---


jpinkney-aws (commented):
Has this idea been floated by @jparkhurst3 ? I think we already have 2+ status bar items and this adds 2 more to that

keeganirby:
> Has this idea been floated by @jparkhurst3 ? I think we already have 2+ status bar items and this adds 2 more to that

Started thread offline

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -63,27 +71,35 @@ async function prepareDocument(session: LiveTailSession): Promise<vscode.TextDoc
...
+                    const editorsToScroll = getTextEditorsToScroll(document)
+                    await updateTextDocumentWithNewLogEvents(formattedLogEvents, document, session.maxLines)
+                    editorsToScroll.forEach(scrollTextEditorToBottom)
+                }
+                updateStatusBarItemsOnStreamEvent(session, event.sessionUpdate)
             }
         }
+    } catch (err) {
+        globals.clock.clearInterval(timer)
+        throw err
```
justinmk3, line 102:
```suggestion
    } finally {
        globals.clock.clearInterval(timer)
```

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -162,12 +220,13 @@ function trimOldestLines(
 function registerTabChangeCallback(
     session: LiveTailSession,
     registry: LiveTailSessionRegistry,
-    document: vscode.TextDocument
+    document: vscode.TextDocument,
+    timer: NodeJS.Timer
```
justinmk3, line 224: Did you check if our `Timeout` module is useful? https://github.com/aws/aws-toolkit-vscode/blob/34b695a91cc0568734c980b1931ac9952c37869c/packages/core/src/shared/utilities/timeoutUtils.ts#L41

It supports cancellations and other features, so mentioning it in case your needs expand later.

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -147,6 +163,48 @@ function trimOldestLines(
...
+    }
+    session.eventRate = numEvents
+}
+
+function hideShowStatusBarItemsOnActiveEditor(session: LiveTailSession, document: vscode.TextDocument) {
+    vscode.window.onDidChangeActiveTextEditor((editor) => {
+        if (editor?.document === document) {
+            session.showStatusBarItem()
+        } else {
+            session.hideStatusBarItem()
```
justinmk3, line 197: There is never a need for multiple "enable/disable" functions. Just make a parameter.


```suggestion
            session.showStatusBarItem(false)
```

justinmk3, line 197:
```suggestion
            session.showStatusBarItem(editor?.document === document)
```

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -147,6 +163,48 @@ function trimOldestLines(
...
+    session.isSampled = isSampled
+}
+
+function updateEventRate(session: LiveTailSession, event: LiveTailSessionUpdate) {
+    let numEvents
+    if (event.sessionResults === undefined) {
+        numEvents = 0
+    } else {
+        numEvents = event.sessionResults.length
+    }
```
justinmk3, line 188: use a ternary. it's clearer, more readable, and allows `const`.

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -147,6 +163,48 @@ function trimOldestLines(
...
+
+function updateIsSampled(session: LiveTailSession, event: LiveTailSessionUpdate) {
+    let isSampled: boolean
+    //sessionMetadata is expected to always be present on a LiveTail stream update
+    if (event.sessionMetadata === undefined || event.sessionMetadata.sampled === undefined) {
+        isSampled = false
+    } else {
+        isSampled = event.sessionMetadata.sampled
+    }
+    session.isSampled = isSampled
```
justinmk3, line 179: could be a ternary.

justinmk3:
> Started thread offline

what was the conclusion?

keeganirby:
> > Started thread offline
> 
> what was the conclusion?

Having a LiveTail statusbar item is ok given that its only visible when its respective session is the active editor. UX would prefer a single status bar item that groups the three pieces of information together (eventRate, timer, isSampled), as opposed to the original implementation with three separate items.

PR #5895, opened 2024-10-29
build(lint): throw error when number of extra args doesn't match the number of format specifiers in logger. 

Hweinstock:
Builds off of:
https://github.com/aws/aws-toolkit-vscode/pull/5878

## Problem
currently `logger.error(msg, err)` silently drops `err` if it doesn't map to a `%x` format specifier in msg.
we want to either prevent that mistake, or enhance our logger so that it prints useful parts of err instead of silently dropping it.

Also, there are cases like:
`getLogger().error('Certificate path {0} already exists', certPath)`, but this silently drops the arg `certPath`. Pretty sure this syntax is from Java IIRC and not supported by Node. 

## Solution
Provide a custom lint rule to ensure the number of extra args matches the number of format specifiers. 

### Note About Alternatives
- As an alternative, we could have transformed the extra `meta` items to strings and appended them to the log message or add them as separate log messages. This would allow usage of the form `getLogger().error('hit error', e)`.
-  If we add separate log messages, we break the expectation that each `logger.error` statement corresponds to one message in the log. 
- Conversely, if we append to the log message, this is somewhat of a hidden feature not exposed to the consumer. By forcing them to add a '%O' format specifier, we are making them clearly aware of how the error (or other object) will show up in the log, ensuring the effect does not "surprise" the user. 

## Notes
Relevant: code used to format errors: 
https://github.com/aws/aws-toolkit-vscode/blob/f806ae04f5d56b872282e42d5f765b2b1a9e28dd/packages/core/src/shared/errors.ts#L329-L334
which should now actually be used since errors won't be dropped. 

- JSCPD is failing due to some duplicates in the files touched. (none introduced here). 

---


packages/core/src/lambda/commands/deploySamApplication.ts
```diff
@@ -0,0 +1,326 @@
+/*!
```
jpinkney-aws, line 1: is this file related somehow?

jpinkney-aws (commented):
It looks like theres a couple more linting that fail at the bottom of this PR as well

packages/core/src/lambda/commands/deploySamApplication.ts
Hweinstock: looks like some more cases got merged on that last master merge that I missed. Should be resolved now.

jpinkney-aws: Oh weird, it looks like its still here. Maybe github UI is glitched? I think the PR is ready to be merged other than that

packages/core/src/ssmDocument/commands/deleteDocument.ts
```diff
@@ -60,7 +60,7 @@ export async function deleteDocument(node: DocumentItemNodeWriteable) {
     } catch (err) {
         result = 'Failed'
         const error = err as Error
-        logger.error('Error on deleting document: %0', error)
+        logger.error('Error on deleting document: %O', error)
```
justinmk3, line 63: Nice! This rule is finding real bugs.

plugins/eslint-plugin-aws-toolkits/index.ts
```diff
@@ -19,6 +20,7 @@ const rules = {
     'no-string-exec-for-child-process': NoStringExecForChildProcess,
     'no-console-log': NoConsoleLog,
     'no-json-stringify-in-log': noJsonStringifyInLog,
+    'no-string-sub-mismatch': noStringSubMismatch,
```
justinmk3, line 23: "printf" might be a slightly more recognizable name. Or something like that. "string substitution" sounds much broader.
```suggestion
    'no-printf-mismatch': noStringSubMismatch,
```

plugins/eslint-plugin-aws-toolkits/lib/rules/no-string-sub-mismatch.ts
```diff
@@ -0,0 +1,62 @@
...
+function countSubTokens(literalNode: TSESTree.StringLiteral) {
+    const formatRegExp = /%[scdjifoO%]/g
+    return literalNode.value.match(formatRegExp)?.length || 0
+}
+/**
+ * Form the error message using templates or actual values.
+ * Allows us to avoid copy pasting message into test file.
+ */
+export function formErrorMsg(substitutionTokens: string | number, numArgs: string | number): string {
+    return `String substitution has ${substitutionTokens} format specifiers, but ${numArgs} arguments.`
```
justinmk3, line 21:
```suggestion
    return `printf-style (console.log) call has ${substitutionTokens} format specifiers, but ${numArgs} arguments.`
```

packages/core/src/shared/filesystemUtilities.ts
```diff
@@ -245,7 +245,7 @@ export async function setDefaultDownloadPath(downloadPath: string) {
             await globals.globalState.update('aws.downloadPath', path.dirname(downloadPath))
         }
     } catch (err) {
-        getLogger().error('Error while setting "aws.downloadPath"', err as Error)
+        getLogger().error('Error while setting "aws.downloadPath" \n %O', err as Error)
```
justinmk3, line 248: nit: it's usually quite readable to start the object at the end of the line, it will look like:

```
error: ... {
  ...
}
```
```suggestion
        getLogger().error('Error while setting "aws.downloadPath": %O', err as Error)
```

packages/core/src/lambda/commands/deploySamApplication.ts
justinmk3: Try `git rm <filepath>` on this file then push to PR.

justinmk3 (approved):

packages/core/src/test/awsService/appBuilder/samProject.test.ts
```diff
@@ -78,7 +78,7 @@ describe('samProject', () => {
             // simulate error when no samconfig.toml file in directory
             const result = await getStackName(projectRoot)
             assert.deepStrictEqual(result, {})
-            assertLogsContain('No stack name or region information available in samconfig.toml', true, 'info')
+            assertLogsContain('No stack name or region information available in samconfig.toml: %O', true, 'info')
```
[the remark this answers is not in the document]
Hweinstock: Yes, but believe it is only because we are in the `TestLogger`, which adds the raw strings to an array. (https://github.com/aws/aws-toolkit-vscode/blob/b2ebc10342f63ac95cd3055393e715cb369ec450/packages/core/src/test/testLogger.ts)

justinmk3 (approved):

PR #5896, opened 2024-10-29
fix(dev): Add conditional for stop code gen on async event

tverney:
## Problem

The function onAsyncEventProgress its reused for both Gumby and Dev feature, which enabled for both (not the expected scenario).

## Solution

Added a condition to prevent and only enable for Dev. 

---

### DIFF
--- a/packages/amazonq/.changes/next-release/Bug Fix-9cebf072-0ebd-476d-a1bf-babd27ef9cc5.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-9cebf072-0ebd-476d-a1bf-babd27ef9cc5.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Amazon Q /dev: fix for stop button showing for Gumby"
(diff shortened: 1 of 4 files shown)


packages/amazonq/.changes/next-release/Bug Fix-9cebf072-0ebd-476d-a1bf-babd27ef9cc5.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Amazon Q /dev: fix for stop button showing for Gumby"
```
dhasani23, line 3: Can you call it `Code Transformation` instead?

dhasani23 (approved):

packages/amazonq/.changes/next-release/Bug Fix-9cebf072-0ebd-476d-a1bf-babd27ef9cc5.json
tverney: Yes, one sec

Issue #4463, opened 2024-02-22
build: warning about Vue esm-bundler flags


justinmk3:
# Problem

When I debug the extension using the `Extension` launch config, the debug console shows this Vue warning:

```
Feature flags __VUE_OPTIONS_API__, __VUE_PROD_DEVTOOLS__ are not explicitly defined. You are running the esm-bundler build of Vue, which expects these compile-time feature flags to be globally injected via the bundler config in order to get better tree-shaking in the production bundle.

For more details, see https://link.vuejs.org/feature-flags.
```

# Expected behavior

Our build should set these flags so that (1) to avoid future confusion/noise about the purpose of the warning, and (2) to gain any potential performance benefits, if possible. Or confirm that we don't need to set these flags and that our build is already optimized.

kiavin:
@justinmk3 how did you solve this

justinmk3:
@kiavin it is not solved. That's why this issue is open.

PR #5899, opened 2024-10-30
build(vue): fix warnings from vue about unset flags. 

Hweinstock:
## Problem
https://github.com/aws/aws-toolkit-vscode/issues/4463

## Solution
The two required flags are set as follows:
- `__VUE_OPTIONS_API__`: `true` since we use the options API extensively. This is opposed to the newer composition API, which I don't believe we use. An example of us using the options API can be found here: https://github.com/aws/aws-toolkit-vscode/blob/f806ae04f5d56b872282e42d5f765b2b1a9e28dd/packages/core/src/login/webview/vue/login.vue#L321 and more info can be found here https://vuejs.org/guide/typescript/options-api
- `__VUE_PROD_DEVTOOLS__`: `false` since this is the default, and it otherwise increases the build size per https://vuejs.org/api/compile-time-flags.html#VUE_PROD_DEVTOOLS. 

## Notes
- Also relevant, composition API: https://vuejs.org/guide/extras/composition-api-faq.html
- More detail on the flags: https://vuejs.org/api/compile-time-flags.html#VUE_PROD_DEVTOOLS

---

### DIFF
--- a/packages/webpack.base.config.js
+++ b/packages/webpack.base.config.js
@@ -101,6 +101,10 @@ module.exports = (env = {}, argv = {}) => {
             new webpack.DefinePlugin({
                 EXTENSION_VERSION: JSON.stringify(packageJson.version),
             }),
+            new webpack.DefinePlugin({
+                __VUE_OPTIONS_API__: 'true',
+                __VUE_PROD_DEVTOOLS__: 'false',
+            }),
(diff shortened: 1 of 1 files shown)


Hweinstock:
Failing techdebt test addressed in: https://github.com/aws/aws-toolkit-vscode/pull/5903

PR #5905, opened 2024-10-30
telemetry(amazonq): emit more metrics for code transformation

dhasani23:
## Problem

Need some more metrics emitted to capture lines of code submitted, lines of code changed, and characters of code changed.

## Solution

Emit said metrics after parsing them from the download ZIP.

---


packages/core/src/codewhisperer/client/user-service-2.json
dhasani23, line 1: TODO: get latest from Wade

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -402,6 +403,29 @@ export class ProposedTransformationExplorer {
...
+                        transformEvent: {
+                            jobId: transformByQState.getJobId(),
+                            timestamp: new Date(),
+                            ideCategory: 'VSCODE',
+                            programmingLanguage: {
+                                languageName: 'JAVA', // TO-DO: use transformByQState.getTransformationType() to tell if JAVA or SQL
+                            },
+                            linesOfCodeChanged: metricsData.linesOfCodeChanged,
+                            charsOfCodeChanged: metricsData.charsOfCodeChanged,
+                            linesOfCodeSubmitted: transformByQState.getLinesOfCodeSubmitted(),
```
dhasani23, line 422: TODO: will this be `undefined` for SQL conversions since we don't do a dry-run? And in general, check to see if this `metrics.json` will even be returned for SQL conversions in the download ZIP.

dhasani23: Update: `metrics.json` is entirely missing for SQL conversions, so for our initial release, we will only have this data for Java upgrades.

packages/core/src/codewhisperer/client/user-service-2.json
```diff
@@ -18,13 +18,25 @@
                 "method": "POST",
```
justinmk3, line 18: Looks like formatting changes were made to this file. The service models have differently-formatted versions, please ensure that you didn't use a version with different formatting, to avoid churn in the git history.

packages/core/src/shared/utilities/download.ts
```diff
@@ -14,6 +14,7 @@ import fs from '../fs/fs'
 export class ExportResultArchiveStructure {
     static readonly PathToSummary = path.join('summary', 'summary.md')
     static readonly PathToDiffPatch = path.join('patch', 'diff.patch')
+    static readonly PathToMetrics = path.join('metrics', 'metrics.json')
```
justinmk3, line 17: join() is not needed here (even on Windows).
```suggestion
    static readonly PathToMetrics = 'metrics/metrics.json'
```

packages/core/src/codewhisperer/client/user-service-2.json
dhasani23: Checked with some others on this, and apparently they're not able to provide me with a version that has formatting consistent with how this file currently looks. I suppose I could manually fix it or add just the changes needed for `/transform` but that doesn't seem like a great idea

dhasani23: Looks like someone else was able to find the right one, updated

packages/core/src/codewhisperer/client/codewhisperer.ts
```diff
@@ -258,7 +258,9 @@ export class DefaultCodeWhispererClient {
         if (!AuthUtil.instance.isValidEnterpriseSsoInUse() && !globals.telemetry.telemetryEnabled) {
             return
         }
+        console.log('about to make API call')
         const response = await (await this.createUserSdkClient()).sendTelemetryEvent(requestWithCommonFields).promise()
+        console.log('API call response =', response)
```
dhasani23, line 263: Will remove all of these later

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -402,6 +403,37 @@ export class ProposedTransformationExplorer {
...
+                console.log('metricsData.charactersOfCodeChanged =', metricsData.charactersOfCodeChanged)
+                console.log('linesOfCodeSubmitted =', transformByQState.getLinesOfCodeSubmitted())
+                console.log('transformByQState.getTransformationType() =', transformByQState.getTransformationType())
+
+                await codeWhisperer.codeWhispererClient.sendTelemetryEvent({
+                    telemetryEvent: {
+                        transformEvent: {
+                            jobId: transformByQState.getJobId(),
+                            timestamp: new Date(),
+                            ideCategory: 'VSCODE',
```
carolabadeer, line 421: Is there any constant value we can extract this `VSCODE` value from? Same for `JAVA` and `SQL` below

carolabadeer (commented):
Looks good to me. Will approve after `console.log` statements are replaced with logs

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -402,6 +404,38 @@ export class ProposedTransformationExplorer {
...
+                            },
+                            linesOfCodeChanged: metricsData.linesOfCodeChanged,
+                            charsOfCodeChanged: metricsData.charactersOfCodeChanged,
+                            linesOfCodeSubmitted: transformByQState.getLinesOfCodeSubmitted(), // currently unavailable for SQL conversions
+                        },
+                    },
+                    userContext: {
+                        ideCategory: 'VSCODE',
+                        operatingSystem: getOperatingSystem(),
+                        product: 'CodeWhisperer',
```
fangwade007, line 432: should it be `AmazonQ` ?

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
fangwade007: +1, do we have enum or constants represent IDE or get it from IDE env ?

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
dhasani23: Yep that makes more sense

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
dhasani23: I don't see this no. All other uses in toolkit seem to use `VSCODE` just like this

carolabadeer (approved):

PR #5912, opened 2024-11-01
fix(lambda): remove template path from sam deploy command

vicheey:
## Problem
Explicitly specifying template path in sam deploy command to point to the build artifact folder makes SAM CLI create/update `samconfig.toml` in the build artifact folder rather than updating the `samconfig.toml` file in project root folder. 

## Solution
Remove the `--template-path` from deploy flag. SAM will determine the template in the project folder and write to correct samconfig file

---

### DIFF
--- a/packages/toolkit/CHANGELOG.md
+++ b/packages/toolkit/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 3.33.0 2024-11-06
+- **Bug Fix** SAM: Save SAM deploy parameters to correct samconfig.toml in project folder
+
 ## 3.31.0 2024-10-30
 
 - **Feature** SAM: Added support for Java 21 Lambda runtime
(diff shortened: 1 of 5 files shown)


roger-zhangg (approved):
LGTM

packages/toolkit/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 3.33.0 2024-11-06
+- **Bug Fix** SAM: Save SAM deploy parameters to correct samconfig.toml in project folder
+
```
hayemaxi, line 3: Do not manually add the changelog. Use `npm run newChange`

vicheey: I reverted manual update and replaced with change json object for next release.

hayemaxi (approved):
lgtm pending integration tests

vicheey:
Looking into Intg test failures

hayemaxi (changes requested):
integ tests need fixing

vicheey:
The integration tests failures do not seem to be related to the new change introduced in this PR but more like false negative resulted from issue with mocking.

seshubaws (approved):

PR #5913, opened 2024-11-01
fix(lambda): skip ecrRepoUri and syncFlags prompters when using flag from samconfig file 

vicheey:
## Problem
When customer choose use flag information from `samconfig.toml` for sync and deploy operations, we assume that all the needed parameters will be available in the `samconfig.toml`.  Only `--config-file <samconfig file path>` is passed to SAM CLI. 

We skip many prompters in such scenario reduce clicking. Yet, we also need to skip prompter for `ecrRepoUri` and `syncFlag` in sync wizard. 

## Solution
Add condition to skip these prompter when flag information from `samconfig.toml` is used. 

---

### DIFF
--- a/packages/core/src/shared/sam/sync.ts
+++ b/packages/core/src/shared/sam/sync.ts
@@ -354,16 +354,24 @@ export class SyncWizard extends Wizard<SyncParams> {
         })
 
         this.form.ecrRepoUri.bindPrompter(({ region }) => createEcrPrompter(new DefaultEcrClient(region!)), {
-            showWhen: ({ template }) => !!template && hasImageBasedResources(template.data),
+            showWhen: ({ template, paramsSource }) =>
+                !!template &&
+                hasImageBasedResources(template.data) &&
+                (paramsSource === ParamsSource.Flags || paramsSource === ParamsSource.SpecifyAndSave),
         })
 
         // todo wrap with localize
-        this.form.syncFlags.bindPrompter(() =>
-            createMultiPick(syncFlagItems, {
-                title: 'Specify parameters for sync',
-                placeholder: 'Press enter to proceed with highlighted option',
-                buttons: createCommonButtons(samSyncParamUrl),
-            })
+        this.form.syncFlags.bindPrompter(
+            () =>
+                createMultiPick(syncFlagItems, {
+                    title: 'Specify parameters for sync',
+                    placeholder: 'Press enter to proceed with highlighted option',
+                    buttons: createCommonButtons(samSyncParamUrl),
+                }),
+            {
+                showWhen: ({ paramsSource }) =>
+                    paramsSource === ParamsSource.Flags || paramsSource === ParamsSource.SpecifyAndSave,
+            }
         )
     }
 }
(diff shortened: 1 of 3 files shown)


packages/core/src/shared/sam/sync.ts
```diff
@@ -354,16 +354,24 @@ export class SyncWizard extends Wizard<SyncParams> {
...
         })
 
         // todo wrap with localize
-        this.form.syncFlags.bindPrompter(() =>
-            createMultiPick(syncFlagItems, {
-                title: 'Specify parameters for sync',
-                placeholder: 'Press enter to proceed with highlighted option',
-                buttons: createCommonButtons(samSyncParamUrl),
-            })
+        this.form.syncFlags.bindPrompter(
```
roger-zhangg, line 364: What would be syncFlags if we skipped here? Will it be empty?

vicheey: I believe that what we discussed last time. 
When the customer use samconfig file, we use all the flag values from the file; this mean, we would use `--config-file` as mention [here](https://github.com/aws/aws-toolkit-vscode/blob/5554065a8c0614a499e259456dafbe91340b15a7/packages/core/src/shared/sam/sync.ts#L424-L460). The based on the value, the [boundArgs](https://github.com/aws/aws-toolkit-vscode/blob/5554065a8c0614a499e259456dafbe91340b15a7/packages/core/src/shared/sam/sync.ts#L544) is loaded here. We do not need the sync flags for merging with `boundArgs` [here]([[fix(lambda): skip ecrRepoUri and syncFlags prompters when using flag from samconfig file  by vicheey · Pull Request #5913 · aws/aws-toolkit-vscode>>https://github.com/aws/aws-toolkit-vscode/pull/5913]]) because again, [--config-file](https://github.com/aws/aws-toolkit-vscode/blob/5554065a8c0614a499e259456dafbe91340b15a7/packages/core/src/shared/sam/sync.ts#L452) is used. 

Please let me know if there is issue.

roger-zhangg: What I mean is, are the sync flags actually empty if they are skipped. Could you help to confirm?

vicheey: Yes, confirm that it is an empty array

roger-zhangg (approved):
LGTM, thanks for improving the wizard experience

roger-zhangg:
One thing we might want to checkout: if appbuilder failed to build/deploy/sync with the current param. Should we still allow `use current param` in the next try?

vicheey:
The integration tests failures do not seem to be related to the new change introduced in this PR but more like false negative resulted from issue with mocking.

justinmk3:
> The integration tests failures do not seem to be related to the new change introduced in this PR but more like false negative resulted from issue with mocking.

Do you plan to address that here or in a followup pr? It is unusual for integ tests to fail.

```
  1) SAM Integration Tests
       SAM Sync
         should instantiate ChildProcess with the correct arguments for sync:
```

PR #5920, opened 2024-11-04
fix(amazonq): validationException if empty supplemental context chunk is sent to the service

Will-ShaoHua:
## Problem

- LSP might return chunk with empty string, which will cause validation exception if such supplemental chunk is sent to the service
- pretty logging
updated log string example
```
2024-11-04 11:36:38.767 [debug] CodeWhispererSupplementalContext:
    isUtg: false,
    isProcessTimeout: false,
    contentsLength: 5208,
    latency: 50.849000096321106
    strategy: OpenTabs_BM25
    Chunk 0:
        Path: /Volumes/workplace/ide/aws-toolkit-vscode-staging/packages/core/src/codewhisperer/service/keyStrokeHandler.ts
        Length: 1777
        Score: 75.30026364918442
    Chunk 1:
        Path: /Volumes/workplace/ide/aws-toolkit-vscode-staging/packages/core/src/codewhisperer/service/referenceInlineProvider.ts
        Length: 1437
        Score: 63.365755359719486
    Chunk 2:
        Path: /Volumes/workplace/ide/aws-toolkit-vscode-staging/packages/core/src/codewhisperer/service/inlineCompletionItemProvider.ts
        Length: 1994
        Score: 47.207284002878936
```

## Solution

---

### DIFF
--- a/packages/core/src/codewhisperer/util/editorContext.ts
+++ b/packages/core/src/codewhisperer/util/editorContext.ts
@@ -208,7 +208,8 @@ function logSupplementalContext(supplementalContext: CodeWhispererSupplementalCo
         isUtg: ${supplementalContext.isUtg},
         isProcessTimeout: ${supplementalContext.isProcessTimeout},
         contentsLength: ${supplementalContext.contentsLength},
-        latency: ${supplementalContext.latency}`,
+        latency: ${supplementalContext.latency}
+        strategy: ${supplementalContext.strategy}`,
         4,
         true
     ).trimStart()
--- a/packages/core/src/codewhisperer/util/editorContext.ts
+++ b/packages/core/src/codewhisperer/util/editorContext.ts
@@ -215,12 +215,11 @@ function logSupplementalContext(supplementalContext: CodeWhispererSupplementalCo
     ).trimStart()
 
     supplementalContext.supplementalContextItems.forEach((context, index) => {
+        logString += indent(`\nChunk ${index}:\n`, 4, true)
         logString += indent(
-            `\nChunk ${index}:
-            Path: ${context.filePath}
-            Length: ${context.content}
-            Score: ${context.score}
-            -----------------------------------------------`,
+            `Path: ${context.filePath}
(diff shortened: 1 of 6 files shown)


packages/core/src/codewhisperer/util/supplementalContext/crossFileContextUtil.ts
```diff
@@ -152,7 +152,7 @@ export async function fetchSupplementalContextForSrcV1(
     // DO NOT send code chunk with empty content
     getLogger().debug(`CodeWhisperer finished fetching crossfile context out of ${relevantCrossFilePaths.length} files`)
     return {
-        supplementalContextItems: supplementalContexts.filter((item) => item.content.trim().length !== 0),
```
Will-ShaoHua, line 155: moved to upstream caller

Will-ShaoHua:
we should refactor the entire structure of `supplementalContextUtil.ts`, `crossfileContextUtil.ts`, `utgUtil.ts`, it's hard to write unit test cases if everything is 1st class functions instead of classes and methods.

packages/core/src/codewhisperer/util/editorContext.ts
```diff
@@ -202,18 +203,26 @@ function logSupplementalContext(supplementalContext: CodeWhispererSupplementalCo
...
+    let logString = indent(
+        `CodeWhispererSupplementalContext:
+        isUtg: ${supplementalContext.isUtg},
+        isProcessTimeout: ${supplementalContext.isProcessTimeout},
+        contentsLength: ${supplementalContext.contentsLength},
+        latency: ${supplementalContext.latency}
+        strategy: ${supplementalContext.strategy}`,
+        4,
+        true
+    ).trimStart()
```
justinmk3, line 215: Nice! Thanks for thinking about the log format, that helps with troubleshooting for both users and devs.

justinmk3 (approved):

zixlin7 (approved):

Issue #502, opened 2019-04-22
Do not allow user to delete Lambda Functions associated with CloudFormation stacks


awschristou:
**Describe the bug**

It should not be possible to delete a Lambda Function that belongs to a CloudFormation stack. This was overlooked during #190.

**To Reproduce**
1. Create a CloudFormation Stack that contains a Lambda Function
1. Go to the AWS Explorer
1. Open the CloudFormation Stack and right click on the function. Notice there is no Delete menu item (good)
1. Locate the "Lambda Functions" node for the same function, and right click on it. Notice there is a Delete menu item (bad). Delete the Lambda Function
1. Function gets deleted

**Expected behavior**

It should not be possible to delete the function.
* Ideal: the context menu does not appear on these nodes
* Next best: an error message is shown indicating that the function cannot be deleted because it is associated with (insert CloudFormation Stack name here)

**Desktop (please complete the following information):**
 - OS: Win
 - Visual Studio Code Version: 1.33
 - AWS Toolkit for Visual Studio Code Version: 0.1.1

awschristou:
Lambda Function tags can be queried for the presence of the following:

```
aws:cloudformation:stack-name
aws:cloudformation:stack-id
aws:cloudformation:logical-id
```

PR #5922, opened 2024-11-04
fix(lambda): removed delete action for lambda and s3 in AppBuilder

seshubaws:
## Problem
When resources are provisioned and controlled through a SAM template, customers should never modify resources directly as this such action would create drift in the CFN template.

## Solution
Removed action to delete a Lambda or S3 bucket in the AppBuilder section

---

### DIFF
--- a/packages/toolkit/.changes/next-release/Bug Fix-f6bb74e5-14cd-4e63-bc24-5e8e82dbaf46.json
+++ b/packages/toolkit/.changes/next-release/Bug Fix-f6bb74e5-14cd-4e63-bc24-5e8e82dbaf46.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "AWS Lambda: Removed action to delete a Lambda or S3 bucket in the AppBuilder section as resources created through IaC frameworks should not be directly modified since this creates drift"
+}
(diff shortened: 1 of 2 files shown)


vicheey (approved):

hayemaxi:
> When resources are provisioned and controlled through a SAM template, customers should never modify resources directly as this such action would create drift in the CFN template.

Is the information presented to the user somehow? Might be worth telling them why they can't delete it.

hayemaxi:
This seems like it needs a changelog item, please see the github bot comment.

seshubaws:
> Is the information presented to the user somehow? Might be worth telling them why they can't delete it.

@hayemaxi Do you know if there is a good place to do that?

justinmk3:
Closes https://github.com/aws/aws-toolkit-vscode/issues/502

seshubaws:
> Closes #502 ?

Unfortunately no, this PR only removes the delete action from resources in the AppBuilder. I believe #502 is referring to removing the delete action for Lambda functions in the Explorer section that belong to CFN stacks. We will prioritize working on #502

hayemaxi:
> > Is the information presented to the user somehow? Might be worth telling them why they can't delete it.
> 
> @hayemaxi Do you know if there is a good place to do that?

Not sure, just something to consider. Maybe keep the option but have it grayed out with a tooltip that says why?

PR #5924, opened 2024-11-04
fix(amazonq): default requestedConversions to undefined in manifest

dhasani23:
## Problem

Recently noticed an issue where language upgrade transformations fail when our `manifest.json` includes this new key which was added for SQL conversions (unreleased feature).

## Solution

Use a default of `undefined` for `requestedConversions`, so that it does not appear in the `manifest.json`, and only add it to the `manifest.json` when the user is doing a SQL conversion, as this key is only used for SQL conversions.

---

### DIFF
--- a/packages/core/src/codewhisperer/models/model.ts
+++ b/packages/core/src/codewhisperer/models/model.ts
@@ -330,17 +330,19 @@ export class ZipManifest {
     hilCapabilities: string[] = ['HIL_1pDependency_VersionUpgrade']
     transformCapabilities: string[] = ['EXPLAINABILITY_V1'] // TO-DO: for SQL conversions, maybe make this = []
     customBuildCommand: string = 'clean test'
-    requestedConversions: {
-        sqlConversion:
-            | {
-                  source: string | undefined
-                  target: string | undefined
-                  schema: string | undefined
-                  host: string | undefined
-                  sctFileName: string | undefined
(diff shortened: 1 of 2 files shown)


packages/core/src/codewhisperer/models/model.ts
```diff
@@ -330,17 +330,19 @@ export class ZipManifest {
...
+                  | {
+                        source: string | undefined
+                        target: string | undefined
+                        schema: string | undefined
+                        host: string | undefined
+                        sctFileName: string | undefined
+                    }
+                  | undefined
+          }
+        | undefined = undefined
```
dhasani23, line 345: This is the main change. Instead of defaulting to `{ sqlConversion: undefined }`, which puts `requestedConversions` in the `manifest.json`, default to `undefined`, which leaves `requestedConversions` out of the `manifest.json` entirely.

hayemaxi, line 345: This would look nicer and also IMO be less confusing since we are removing so much text.
```suggestion
    requestedConversions?: {
        sqlConversion?: {
            source?: string
            target?: string
            schema?: string
            host?: string
            sctFileName?: string
        }
    }
```

hayemaxi (approved):

carolabadeer (approved):

PR #5925, opened 2024-11-05
feat(amazonq): enable SQL conversions feature

dhasani23:
## Problem

Enable the SQL conversion feature for `/transform`.

## Solution

Enable feature, intended for 11/21 release.

---

### DIFF
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -224,7 +224,7 @@ export class GumbyController {
         this.sessionStorage.getSession().conversationState = ConversationState.WAITING_FOR_TRANSFORMATION_OBJECTIVE
         this.messenger.sendStaticTextResponse('choose-transformation-objective', message.tabID)
         this.messenger.sendChatInputEnabled(message.tabID, true)
-        this.messenger.sendUpdatePlaceholder(message.tabID, "Enter 'language upgrade' or 'SQL conversion'")
+        this.messenger.sendUpdatePlaceholder(message.tabID, CodeWhispererConstants.chooseTransformationObjective)
     }
 
     private async beginTransformation(message: any) {
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -33,7 +33,7 @@ import {
     getValidSQLConversionCandidateProjects,
     validateSQLMetadataFile,
 } from '../../../codewhisperer/commands/startTransformByQ'
-import { JDKVersion, transformByQState } from '../../../codewhisperer/models/model'
+import { JDKVersion, TransformationCandidateProject, transformByQState } from '../../../codewhisperer/models/model'
 import {
     AbsolutePathDetectedError,
     AlternateDependencyVersionsNotFoundError,
@@ -62,8 +62,6 @@ import { getStringHash } from '../../../shared/utilities/textUtilities'
 import { getVersionData } from '../../../codewhisperer/service/transformByQ/transformMavenHandler'
 import AdmZip from 'adm-zip'
 import { AuthError } from '../../../auth/sso/server'
-import { isSQLTransformReady } from '../../../dev/config'
-import { spawnSync } from 'child_process'
 
 // These events can be interactions within the chat,
 // or elsewhere in the IDE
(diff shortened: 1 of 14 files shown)


packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -189,9 +190,35 @@ export class GumbyController {
...
+                // TO-DO: handle Windows
+                const spawnResult = spawnSync('grep', args, {
+                    cwd: project.path,
+                    shell: true,
+                    encoding: 'utf-8',
+                })
+                if (spawnResult.status !== 0) {
+                    return false
+                }
+                // TO-DO: parse stdout for the count of matching lines
```
dhasani23, line 210: Don't use `-c` flag, trim `stdout`, and if anything present, return `true`

dhasani23:
TODO: add changelog entry

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -80,6 +81,7 @@ import { setContext } from '../../shared/vscode/setContext'
 import { makeTemporaryToolkitFolder } from '../../shared'
 import globals from '../../shared/extensionGlobals'
 import { convertDateToTimestamp } from '../../shared/datetime'
+import { spawnSync } from 'child_process'
```
justinmk3, line 84: use our ChildProcess wrapper unless there is a technical reason why it can't be used. https://github.com/aws/aws-toolkit-vscode/blob/8966edc82a33c5b7f00d45477941bcc333a10318/packages/core/src/shared/utilities/processUtils.ts#L70

It has logging, and will eventually gain some performance-tracking and telemetry features.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -624,6 +639,14 @@ export class GumbyController {
 
             case ConversationState.WAITING_FOR_TRANSFORMATION_OBJECTIVE: {
                 const objective = data.message.trim().toLowerCase()
+                // since we're prompting the user, their project(s) must be eligible for both types of transformations, so track how often this happens here
+                if (objective === 'language upgrade' || objective === 'sql conversion') {
+                    telemetry.codeTransform_submitSelection.emit({
```
dhasani23, line 644: Best practice to use a `run` block, but I don't have anything to do here, just want to emit a metric to track how often the user gets to this point because it's meaningful for us. Furthermore, `handleLanguageUpgrade` and `handleSQLConversion` below already emit a `validateProject` metric.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -728,13 +731,57 @@ export async function finalizeTransformationJob(status: string) {
...
+    const embeddedSQLProjects: TransformationCandidateProject[] = []
+    await telemetry.codeTransform_validateProject.run(async () => {
+        telemetry.record({
+            codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId(),
+        })
+        const openProjects = await getOpenProjects()
+        const javaProjects = await getJavaProjects(openProjects)
+        let errorLog = ''
+        const isWindows = isWin()
+        const command = isWindows ? 'findstr' : 'grep'
```
justinmk3, line 748: vscode ships with `rg` (ripgrep, extremely fast and cross-platform). Can we use that? https://github.com/microsoft/vscode-ripgrep

I don't exactly how or if it's exposed. I couldn't find an official vscode api. It looks like `findTextInFiles` is a proposed API: https://github.com/microsoft/vscode/issues?q=is%3Aissue%20state%3Aopen%20findTextInFiles

While we're waiting on that, at minimum, please extract this file-search logic to our shared `workspaceUtils.ts` module. This kind of functionality isn't something that should be casually added to an arbitrary Q module.

By extracting to a shared module, that gives us an interface that allows us to later switch it to `rg` (which we could sniff out in vscode's path somewhere, since it ships with `rg`), or `findTextInFiles` when that lands.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -190,12 +189,31 @@ export class GumbyController {
...
 
     private async transformInitiated(message: any) {
-        // feature flag for SQL transformations
-        if (!isSQLTransformReady) {
+        // silently check for projects eligible for SQL conversion
+        let embeddedSQLProjects: TransformationCandidateProject[] = []
+        try {
+            embeddedSQLProjects = await getValidSQLConversionCandidateProjects()
+        } catch (err) {
+            getLogger().error(`CodeTransformation: error validating SQL conversion projects: ${err}`)
```
justinmk3, line 197: Just for reference: `getLogger()` now supports a topic, so you can declare a shared logger at module scope:

    const logger = getLogger('CodeTransform')

then use `logger` anywhere in the module. And the topic will prefix each message.

dhasani23: Good to know; will update all of our logging with this in my next PR

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -610,3 +612,24 @@ export async function collectFilesForIndex(
...
+ * Performs a case-insensitive, recursive search for a string in a directory.
+ * note:
+ * 'error' is set when command fails to spawn; 'stderr' is set when command itself fails.
+ * 'exitCode' will be 0 when searchStr is detected (each line where it's found will be printed to 'stdout').
+ * 'exitCode' will be non-zero and 'stdout' / 'stderr' / 'error' will all be empty/undefined when searchStr is not detected.
+ * @param searchStr the string to search for
+ * @param dirPath the path of the directory to search in
+ * @returns the result of the search
+ */
+export async function findStringInDirectory(searchStr: string, dirPath: string) {
```
justinmk3, line 626: might want some direct test coverage of this

justinmk3 (approved):

packages/core/src/shared/utilities/workspaceUtils.ts
dhasani23: Yes see very bottom of PR, in `workspaceUtils.test.ts`

packages/core/src/codewhisperer/models/constants.ts
```diff
@@ -477,6 +481,8 @@ export const failedStepMessage = 'The step failed, fetching additional details..
 
 export const jobCompletedMessage = 'The transformation completed.'
 
+export const noChangesMadeMessage = "I didn't make any changes for this transformation."
```
ntarakad-aws, line 484: nit: Will the didn't appear or do we need to deal with apostrophe special character? I notice sometimes we have the backslash and sometimes we don't, so just want to make sure

ntarakad-aws (approved):

PR #5931, opened 2024-11-05
refactor(amazonq): add BaseConnector Abstract class for common logic

ivikash:
Create an abstract class that can be used by all connectors.

---

### DIFF
--- a/packages/core/src/amazonq/webview/ui/apps/baseConnector.ts
+++ b/packages/core/src/amazonq/webview/ui/apps/baseConnector.ts
@@ -1,0 +1,298 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ChatItem, ChatItemAction, ChatItemType, FeedbackPayload } from '@aws/mynah-ui'
+import { ExtensionMessage } from '../commands'
+import { CodeReference } from './amazonqCommonsConnector'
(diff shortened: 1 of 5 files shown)


packages/core/src/amazonq/webview/ui/apps/BaseConnector.ts
```diff
@@ -0,0 +1,298 @@
...
+            link,
+            tabType: this.getTabType(),
+        })
+    }
+
+    followUpClicked = (tabID: string, messageId: string, followUp: ChatItemAction): void => {
+        /**
+         * We've pressed on a followup button and should start watching that round trip telemetry
+         */
+        this.sendMessageToExtension({
```
jpinkney-aws, line 71: I'd probably move these out of here for now and into the cwchatconnector. I haven't checked whether or not the entire chat round trip telemetry flow works for gumby/feature dev yet

packages/core/src/amazonq/webview/ui/apps/BaseConnector.ts
justinmk3, line 1: file name should be camelcase (baseConnector.ts)

justinmk3 (approved):
Nice! Are there any tests that could be de-duplicated too?

packages/core/src/amazonq/webview/ui/apps/BaseConnector.ts
ivikash: Discussed Offline - `amazonq_chatRoundTrip` is not getting emitted for `/dev`, so this is good to go.

jpinkney-aws (approved):

PR #5932, opened 2024-11-05
feat(notifications): setup and activation

hayemaxi:
- Gated behind dev setting (`aws.dev.notifications`)
- Setup the notifications panel and begin polling in each of the extensions.
- Currently the code collects a bunch of information about the state of auth for telemetry (auth_userState). Refactor this code in each extension so that we can re-use that information for the notification rule engine.
- Change emitting auth_userState to node only for amazonq, since notifications is node-only. Decided to punt this work for web because spending effort to make it web-compatible may be in vain if the design of web metrics is different.

---


hayemaxi:
>   

Feature not released.

justinmk3 (approved):

Issue #5933, opened 2024-11-06
debugging sam app inside vscode fails with dotnet8


saweisulzer:
## Steps to reproduce the issue

1. Create sam project with "sam init --app-template hello-world --name sam-app --package-type Zip --runtime dotnet8"
2. Create "aws-sam" launch configuration like this:
"type": "aws-sam",
            "request": "direct-invoke",
            "name": "sam-app:HelloWorldFunction",
            "invokeTarget": {
                "target": "template",
                "templatePath": "${workspaceFolder}/template.yaml",
                "logicalId": "HelloWorldFunction"
            },
            "lambda": {
                "payload": {},
                "environmentVariables": {}
            }

But start Fails with:
2024-11-06 09:41:19.731 [error] SAM CLI failed (exitcode: 1, expected 0): 
    stdout:
    Build Failed
    stderr:
    2024-11-06 09:41:16,956 | Using SAM Template at C:\sw\sulzer\dotnet\sam8\sam-app\template.yaml
    2024-11-06 09:41:16,977 | Config file location: C:\sw\sulzer\dotnet\sam8\sam-app\samconfig.toml
    2024-11-06 09:41:16,978 | Loading configuration values from [default.['build'].parameters] (env.command_name.section) in config file at 'C:\sw\sulzer\dotnet\sam8\sam-app\samconfig.toml'...
    2024-11-06 09:41:16,978 | Configuration values successfully loaded.
    2024-11-06 09:41:16,978 | Configuration values are: {'stack_name': 'sam-app', 'cached': True, 'parallel': True}
    2024-11-06 09:41:17,028 | Using config file: samconfig.toml, config environment: default
    2024-11-06 09:41:17,028 | Expand command line arguments to:
    2024-11-06 09:41:17,028 | --template_file=C:\sw\sulzer\dotnet\sam8\sam-app\template.yaml --build_dir=C:\Users\weisers\AppData\Local\Temp\aws-toolkit-vscode\vsctk5c04240b\output --parallel --mount_with=READ --cache_dir=.aws-sam\cache --cached 
    2024-11-06 09:41:18,135 | 'build' command is called
    2024-11-06 09:41:18,135 | Starting Build use cache
    2024-11-06 09:41:18,145 | No Parameters detected in the template
    2024-11-06 09:41:18,317 | There is no customer defined id or cdk path defined for resource HelloWorldFunction, so we will use the resource logical id as the resource id
[...]

Issue #5918, opened 2024-11-04
Error: NodejsNpmBuilder:Resolver - Path resolution for runtime: nodejsXX.x of binary: npm was not successful


mtc-ecaro:
## Problem

When testing NPM-based Lambda functions locally using JSON configuration in Visual Studio Code, an error occurs:

Error: NodejsNpmBuilder:Resolver - Path resolution for runtime: nodejsXX.x of binary: npm was not successful

Temporary Solution: Downgrading the AWS Toolkit to version 3.30

## Steps to reproduce the issue

Set up a Lambda function with Node.js runtime (nodejs20.x).
Attempt to run or test the function locally in VS Code with JSON configuration.

## Expected behavior

The Lambda function should run locally without path resolution issues for npm

## System details (run `AWS: About` and/or `Amazon Q: About`)

- OS: Windows 11 (latest version)
- Visual Studio Code: Latest version
- AWS Toolkit: Version 3.31
- Amazon Q: Not applicable

roger-zhangg:
Hi @mtc-ecaro , thanks for reporting this. Could you please share the following detail
- Which shell is your default VSCode shell? (Powershell / WSL maybe?)
- Where did you get the resolution error. In terminal shell, or it's in toolkit logs
- Does downgrading resolve this issue immediately
- 
These will help us better resolve the issue. Thank you

mtc-ecaro:
Hello, as requested here is the additional information:

- Powershell
- Toolkit logs
- Downgrading to .30 solve my issue

Just for context , I'm part of a small team and all of them had the same issue, we all have rolled back to .30 and are able to locally execute/debug the lambda functions successfully.

justinmk3:
This request was addressed in [AWS Toolkit 3.32.0](https://marketplace.visualstudio.com/items?itemName=AmazonWebServices.aws-toolkit-vscode) and/or [Amazon Q 1.34.0](https://marketplace.visualstudio.com/items?itemName=AmazonWebServices.amazon-q-vscode)!

- AWS Toolkit changelog: https://github.com/aws/aws-toolkit-vscode/blob/master/packages/toolkit/CHANGELOG.md
- Amazon Q changelog: https://github.com/aws/aws-toolkit-vscode/blob/master/packages/amazonq/CHANGELOG.md
- Release artifacts: https://github.com/aws/aws-toolkit-vscode/releases

PR #5936, opened 2024-11-06
fix(appbuilder): getSpawnEnv should return correct env on windows

roger-zhangg:
## Problem
See: #5933  #5918 

## Solution
Added a check in resolveEnv to make sure it should do nothing on windows.

---


roger-zhangg:
Windows: build process in local invoke can find Nodejs/Dotnet CLI after this fix.

packages/core/src/test/shared/env/resolveEnv.test.ts
seshubaws, line 1: Can we add a test to make sure the change still works for mac platforms?

seshubaws (changes requested):

packages/core/src/test/shared/env/resolveEnv.test.ts
Vandita2020: The test seems to be failing for mac platform case.

roger-zhangg: Yes, need to stub the return as this is a platform sensitive test

packages/core/src/shared/sam/localLambdaRunner.ts
```diff
@@ -236,7 +233,7 @@ async function invokeLambdaHandler(
             templatePath: config.templatePath,
             eventPath: config.eventPayloadFile,
             environmentVariablePath: config.envFile,
-            environmentVariables: await getSpawnEnv(env),
```
jonife, line 239: This solution applies the same logic to lower-tier versions that we currently use for customer downgrades. I am still skeptical about this solution but i think we are on the right track

packages/core/src/shared/sam/localLambdaRunner.ts
```diff
@@ -290,15 +287,16 @@ export async function runLambdaFunction(
...
+        ...process.env,
         ...(config.aws?.region ? { AWS_DEFAULT_REGION: config.aws.region } : {}),
-    }
+    })
 
     const settings = SamCliSettings.instance
     const timer = new Timeout(settings.getLocalInvokeTimeout())
 
     // TODO: refactor things to not mutate the config
-    config.templatePath = await buildLambdaHandler(timer, await getSpawnEnv(envVars), config, settings)
```
jonife, line 301: 1. Were you able to reproduce the issue locally in your window instance.
2. I think this is where the issue is from but this file remains constant across versions. I think we need to figure out what is changing the executable path before it get here for windows.

packages/core/src/shared/sam/localLambdaRunner.ts
roger-zhangg: The error happens cause `getSpawnEnv` will mistakenly output a overwritten process.env.PATH on Windows before this patch. And in build process in local invoke, this will overwrite the process.env.PATH i. Resulting sam build can't find the binary of npm/dotnet. After this Patch, getSpawnEnv(env) will directly return env correctly if it's on windows.

For lower version, process.env.PATH will be null in the envVar. And it won't overwrite process.env.PATH in later part.

For Mac, `getSpawnEnv` will output the right path so there's no issue on Mac

packages/core/src/shared/sam/localLambdaRunner.ts
roger-zhangg: 1: Yes I can reproduce.
2: see comment above

seshubaws (approved):

packages/core/src/shared/env/resolveEnv.ts
```diff
@@ -114,7 +114,7 @@ export async function getSpawnEnv(
 export async function mergeResolvedShellPath(env: IProcessEnvironment): Promise<typeof process.env> {
     const shellEnv = await getResolvedShellEnv(env)
     // resolve failed or doesn't need to resolve
-    if (!shellEnv) {
+    if (!shellEnv || Object.keys(shellEnv).length === 0) {
```
justinmk3, line 117: would it improve the interface if `getResolvedShellEnv` did this check and returned undefined?

packages/core/src/test/shared/env/resolveEnv.test.ts
```diff
@@ -0,0 +1,69 @@
...
+        sandbox.restore()
+    })
+
+    // a copy of resolveEnv.mergeResolvedShellPath for stubbing
+    const testMergeResolveShellPath = async function mergeResolvedShellPath(
+        env: resolveEnv.IProcessEnvironment
+    ): Promise<typeof process.env> {
+        const shellEnv = await resolveEnv.getResolvedShellEnv(env)
+        // resolve failed or doesn't need to resolve
+        if (!shellEnv || Object.keys(shellEnv).length === 0) {
```
justinmk3, line 27: another case that would be improved by https://github.com/aws/aws-toolkit-vscode/pull/5936/files#r1831813874

packages/core/src/test/shared/env/resolveEnv.test.ts
```diff
@@ -0,0 +1,69 @@
...
+            // merge, dedup, join
+            envReturn.PATH = [...new Set(envPaths.concat(resolvedPaths))].join(path.delimiter)
+
+            return envReturn
+        } catch (err) {
+            return env
+        }
+    }
+
+    describe('resolveWindows', async function () {
```
justinmk3, line 43: `resolveWindows` is not a symbol

packages/core/src/shared/env/resolveEnv.ts
roger-zhangg: Yes that make sense. I'll change that shortly

packages/core/src/test/shared/env/resolveEnv.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    beforeEach(function () {
+        sandbox = sinon.createSandbox()
+    })
+
+    afterEach(function () {
+        sandbox.restore()
+    })
+
+    // a copy of resolveEnv.mergeResolvedShellPath for stubbing
+    const testMergeResolveShellPath = async function mergeResolvedShellPath(
```
justinmk3, line 22: why not stub `resolveEnv.getResolvedShellEnv` directly? that would be worth explaining (concisely) in the above comment

packages/core/src/test/shared/env/resolveEnv.test.ts
```diff
@@ -0,0 +1,69 @@
...
+        })
+
+        it('mergeResolvedShellPath should not change path on windows', async function () {
+            const env = await resolveEnv.mergeResolvedShellPath(process.env)
+            assert(env.PATH)
+            assert.strictEqual(env, process.env)
+        })
+    })
+
+    describe('resolveUnix', async function () {
```
justinmk3, line 55:
```suggestion
    describe('unix', async function () {
```

packages/toolkit/.changes/next-release/Bug Fix-268ceb41-7bec-4b57-b195-d72f19d00faa.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "getSpawnEnv always return correct env on windows"
```
justinmk3, line 3: Thanks for including a changelog. Please review the [changelog guidelines](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#changelog).

packages/core/src/test/shared/env/resolveEnv.test.ts
roger-zhangg: `mergeResolvedShellPath` is calling `getResolvedShellEnv` within the same file (with out importing it). Is there a way for us to make `mergeResolvedShellPath` call the stub instead? I didn't figure that out before

roger-zhangg: Added this in comment

packages/toolkit/.changes/next-release/Bug Fix-268ceb41-7bec-4b57-b195-d72f19d00faa.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "System Path parsing should ignore Windows and only parse Mac/Linux system, Sam Local Invoke should get correct system Path on windows"
```
justinmk3, line 3:
```suggestion
	"description": "System Path parsing should ignore Windows and only parse Mac/Linux system, Sam Local Invoke should get correct system Path on windows #5933 #5918"
```

PR #5938, opened 2024-11-06
feat(cwl): LiveTail statusbar

keeganirby:
Problem
As LiveTail session is running, we want to display to the user metadata about their session.

**While updating my local feature branch after fixing the merge conflicts in https://github.com/aws/aws-toolkit-vscode/pull/5889, my original PR for this change was closed.**

**Please see original PR [here](https://github.com/aws/aws-toolkit-vscode/pull/5894) for addional comments and context**

Solution
While a livetail session is the active editor, in the bottom status bar, we will display the duration the session has been running for, how many events per/s, and if the log data is being sampled or not.

### DIFF
--- a/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
+++ b/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
@@ -8,8 +8,12 @@ import { TailLogGroupWizard } from '../wizard/tailLogGroupWizard'
 import { CancellationError } from '../../../shared/utilities/timeoutUtils'
 import { LiveTailSession, LiveTailSessionConfiguration } from '../registry/liveTailSession'
 import { LiveTailSessionRegistry } from '../registry/liveTailSessionRegistry'
-import { LiveTailSessionLogEvent, StartLiveTailResponseStream } from '@aws-sdk/client-cloudwatch-logs'
-import { ToolkitError } from '../../../shared'
+import {
+    LiveTailSessionLogEvent,
+    LiveTailSessionUpdate,
+    StartLiveTailResponseStream,
+} from '@aws-sdk/client-cloudwatch-logs'
+import { globals, ToolkitError } from '../../../shared'
 
 export async function tailLogGroup(
     registry: LiveTailSessionRegistry,
@@ -35,13 +39,17 @@ export async function tailLogGroup(
     registry.set(session.uri, session)
 
     const document = await prepareDocument(session)
-    registerTabChangeCallback(session, registry, document)
+    const timer = startSessionTimer(session)
+    hideShowStatusBarItemsOnActiveEditor(session, document)
+    registerTabChangeCallback(session, registry, document, timer)
+
     const stream = await session.startLiveTailSession()
 
-    await handleSessionStream(stream, document, session)
+    await handleSessionStream(stream, document, session, timer)
 }
 
-export function closeSession(sessionUri: vscode.Uri, registry: LiveTailSessionRegistry) {
+export function closeSession(sessionUri: vscode.Uri, registry: LiveTailSessionRegistry, timer: NodeJS.Timer) {
+    globals.clock.clearInterval(timer)
     const session = registry.get(sessionUri)
     if (session === undefined) {
         throw new ToolkitError(`No LiveTail session found for URI: ${sessionUri.toString()}`)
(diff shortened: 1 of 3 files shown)


packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -147,6 +162,38 @@ function trimOldestLines(
     edit.delete(document.uri, range)
 }
 
+function updateStatusBarItemsOnStreamEvent(session: LiveTailSession, event: LiveTailSessionUpdate) {
+    updateIsSampled(session, event)
+    updateEventRate(session, event)
+}
+
+function updateIsSampled(session: LiveTailSession, event: LiveTailSessionUpdate) {
+    session.isSampled =
```
jpinkney-aws, line 171: instead of setting it here for testability does it make sense to just return and then have `session.isSampled = isSampled(session, event)`?

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -147,6 +162,38 @@ function trimOldestLines(
...
+        event.sessionMetadata === undefined || event.sessionMetadata.sampled === undefined
+            ? false
+            : event.sessionMetadata.sampled
+}
+
+function updateEventRate(session: LiveTailSession, event: LiveTailSessionUpdate) {
+    session.eventRate = event.sessionResults === undefined ? 0 : event.sessionResults.length
```
jpinkney-aws, line 178: same comment as above with this one

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -147,6 +162,38 @@ function trimOldestLines(
     edit.delete(document.uri, range)
 }
 
+function updateStatusBarItemsOnStreamEvent(session: LiveTailSession, event: LiveTailSessionUpdate) {
```
jpinkney-aws, line 165: do we need this function if its only used once?

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -147,6 +162,38 @@ function trimOldestLines(
...
+        if (editor?.document === document) {
+            session.showStatusBarItem(true)
+        } else {
+            session.showStatusBarItem(false)
+        }
+    })
+}
+
+function startSessionTimer(session: LiveTailSession): NodeJS.Timer {
+    return globals.clock.setInterval(() => {
```
jpinkney-aws, line 192: we can probably just inline this where its called

justinmk3:
> While updating my local feature branch after fixing the merge conflicts in #5889, my original PR for this change was closed.

it can always be reopened (just force-push to the same branch). please avoid creating new prs

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -147,6 +162,38 @@ function trimOldestLines(
...
+function updateEventRate(session: LiveTailSession, event: LiveTailSessionUpdate) {
+    session.eventRate = event.sessionResults === undefined ? 0 : event.sessionResults.length
+}
+
+function hideShowStatusBarItemsOnActiveEditor(session: LiveTailSession, document: vscode.TextDocument) {
+    vscode.window.onDidChangeActiveTextEditor((editor) => {
+        if (editor?.document === document) {
+            session.showStatusBarItem(true)
+        } else {
+            session.showStatusBarItem(false)
```
justinmk3, line 186:
```suggestion
        session.showStatusBarItem(editor?.document === document)
```

packages/core/src/awsService/cloudWatchLogs/registry/liveTailSession.ts
```diff
@@ -102,4 +134,15 @@ export class LiveTailSession {
...
+
+    public showStatusBarItem(shouldShow: boolean) {
+        shouldShow ? this.statusBarItem.show() : this.statusBarItem.hide()
+    }
+
+    public updateStatusBarItemText() {
+        const elapsedTime = this.getLiveTailSessionDuration()
+        const timeString = convertToTimeString(elapsedTime)
+        const sampledString = this._isSampled ? 'Yes' : 'No'
+        this.statusBarItem.text = `Tailing Session: ${timeString}, ${this._eventRate} events/sec, Sampled: ${sampledString}`
```
justinmk3, line 146: would it make sense to show "Sampled" only if it's being sampled, instead of Yes/No?

also can save space by avoiding unnecessary words like "session"

```suggestion
        this.statusBarItem.text = `Tailing: ${timeString}, ${this._eventRate} events/sec, Sampled: ${sampledString}`
```

packages/core/src/awsService/cloudWatchLogs/registry/liveTailSession.ts
```diff
@@ -48,6 +53,9 @@ export class LiveTailSession {
         }
         this._maxLines = LiveTailSession.settings.get('limit', 10000)
         this._uri = createLiveTailURIFromArgs(configuration)
+        this.statusBarItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Right, 0)
```
justinmk3, line 56: 👍 this is a secondary/contextual item, so `Right` is the [preferred placement](https://code.visualstudio.com/api/ux-guidelines/status-bar).

packages/core/src/awsService/cloudWatchLogs/registry/liveTailSession.ts
keeganirby: Our current CLI Experience always shows "Sampled: Yes/No" so to keep parity with that we will want to keep it displayed. I can remove Session though!

justinmk3: It's good to avoid needless entropy, but when it comes to the IDE statusbar, matching the cli is very low on the priorities. Using space economically is very important here.

packages/core/src/awsService/cloudWatchLogs/registry/liveTailSession.ts
```diff
@@ -71,17 +87,33 @@ export class LiveTailSession {
...
     }
 
     public stopLiveTailSession() {
+        this.endTime = Date.now()
+        this.statusBarItem.dispose()
         this.liveTailClient.abortController.abort()
         this.liveTailClient.cwlClient.destroy()
     }
 
+    public getLiveTailSessionDuration(): number {
```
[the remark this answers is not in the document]
keeganirby: We currently don't impose a client-side timeout limit on a LiveTail Session, so i don't think that using the Timeout module here fits our use case. 

Our Livetail backend will timeout on the server-side, but as of now, that is not something the client should be concerned with. In that case I don't think we will use the Timeout module at the moment. If we do want to set a client-side timeout, then I'd definitley use that module and perhaps elapsed could replace this logic.

packages/core/src/awsService/cloudWatchLogs/registry/liveTailSession.ts
keeganirby: Makes sense. Though Toolkit UX did review this as is.

Another point to consider: if a tailing session is fluctuating between being sampled or not (which can happen depending on the ingestion rate of log events into the LogGroup), a user would see the length/content of the statusBar change quite a bit if the word "Sampled" is appearing and disappearing. 

Changing between Yes and No would be a lot less jarring. Also seeing "Sampled: No" still tells me something about my session, whereas showing nothing at all doesn't tell me anything.

keeganirby:
Pushed changes to address some feedback. Open to more conversation before changing the StatusBar text further

PR #5940, opened 2024-11-06
deps(sdk): barebones general public client builder. 

Hweinstock:
## Problem
Start the transition to sdkv3. This PR is just a start, and is intended to serve as an opportunity to look at the design before building on it more. This is still missing a lot of features (see future work). 

Excessive line changes are due to package-lock.json changes from install sdk-v3 clients. 

## Solution
- Do not overwrite existing builder to allow for incremental transition to V3. 
- A barebones working model, with the telemetry middleware implemented. 
- A proof of concept transition with EC2, unreleased nature makes it low-risk. 
- Expansion of testing for client construction. 

## Future Work
- Implement middleware for other relevant pieces (custom endpoints, logging, etc.).
- Increase testing coverage, specifically with credentials refreshing. 
- Migrate other clients. 
- Reduce boilderplate for client wrappers. 
- replace hand-made mocks with those from sdk mocking library: https://aws.amazon.com/blogs/developer/mocking-modular-aws-sdk-for-javascript-v3-in-unit-tests/

---


packages/core/src/shared/awsClientBuilderV3.ts
```diff
@@ -0,0 +1,150 @@
...
+        requestServiceType?: string
+    }
+
+    telemetry.record({
+        requestId: getRequestId(err),
+        requestServiceType: serviceName,
+    } satisfies RequestData as any)
+}
+
+export function omitIfPresent<T extends Record<string, unknown>>(obj: T, keys: string[]): T {
```
jpinkney-aws, line 111: is this something that makes sense to be in a util?

maybe the other places we omit data isn't the same though?

packages/core/src/shared/clients/ec2Wrapper.ts
```diff
@@ -17,32 +38,33 @@ import { decodeBase64 } from '../utilities/textUtilities'
...
 }
 
-interface SafeEc2GetConsoleOutputResult extends EC2.GetConsoleOutputRequest {
+interface SafeEc2GetConsoleOutputResult extends GetConsoleOutputRequest {
     Output: string
     InstanceId: string
 }
 
-export class Ec2Client {
+export class Ec2Wrapper {
```
jpinkney-aws, line 52: Out of curiosity what was the idea between calling it Ec2Wrapper? Is it because it's a toolkit wrapper around the sdk v3 builder?

Hweinstock: Yeah it seems like a more fitting name, and also because the EC2 v3 exports its client as `EC2Client` (instead of `EC2` as in v2) so want to avoid confusion between the names. Do you think that renaming the 'clients' to 'wrappers' as we migrate them makes sense?

packages/core/src/shared/awsClientBuilderV3.ts
Hweinstock: I think moving it to a util is a good idea. Nothing about it is specific to client builder.

Hweinstock:
This PR is too large, and I am unable to resolve build problems. Moving to https://github.com/aws/aws-toolkit-vscode/pull/6097

PR #5941, opened 2024-11-06
feat(logging): support logger instances

justinmk3:
## Problem
`getLogger('foo')` returns a global singleton, so the "current topic" only is stored until the next `getLogger` call.

## Solution
Modify `getLogger('foo')` to return a wrapper which stores its topic. This allows modules to declare their own module-local shared logger:

    const logger = getLogger('foo')

---


packages/core/src/test/shared/logger/toolkitLogger.test.ts
```diff
@@ -240,55 +241,20 @@ describe('ToolkitLogger', function () {
...
-            const testMessage = 'This is a test message'
-            const testMessageWithHeader = 'test: This is a test message'
-
-            testLogger = new ToolkitLogger('info')
-            testLogger.logToOutputChannel(outputChannel, false)
-            testLogger.setTopic('unknown')
-            testLogger.setTopic('test')
-            testLogger.setLogLevel('verbose')
-            testLogger.verbose(testMessage)
+            // const msg2 = "topic: 'test'"
```
jpinkney-aws, line 250: is this test code no longer relevant?

jpinkney-aws (approved):

packages/core/src/test/lambda/vue/remoteInvoke/invokeLambda.test.ts
```diff
@@ -190,20 +190,11 @@ describe('RemoteInvokeWebview', () => {
...
-                assert.strictEqual(loggerErrorStub.firstCall.args[1], fileUri.fsPath)
-                assert(loggerErrorStub.firstCall.args[2] instanceof Error)
-            } finally {
-                loggerErrorStub.restore()
-            }
+            await assert.rejects(
+                async () => await remoteInvokeWebview.promptFile(),
+                new Error('Failed to read selected file')
+            )
+            assertLogsContain('readFileSync: Failed to read file at path %s %O', true, 'error')
```
justinmk3, line 197: Test actual state instead of mocks. Saves lots of code, finds actual bugs, and gives meaningful tests instead of tests that break for no reason.

packages/core/src/test/webview/server.test.ts
```diff
@@ -3,44 +3,34 @@
...
+    afterEach(function () {})
 
     it('logs the provided error, but is wrapped in ToolkitErrors for more context', function () {
         // The method is being tested due to its fragile implementation. This test
         // protects against changes in the underlying logAndShowError() implementation.
 
         const inputError = new Error('Random Error')
 
-        handleWebviewError(inputError, myWebviewId, myCommand)
+        const err = handleWebviewError(inputError, myWebviewId, myCommand)
```
justinmk3, line 25: Test the actual state! No mocks needed.

PR #5942, opened 2024-11-06
fix(crash): fsync heartbeat file writes

nkomonen-amazon:
## Problem

In our telemetry data we saw the following scenario:

- We saw a session (`7ebb2966-7e94-4177-80d9-c5a485511c13`) sending heartbeats as normal, with the latest one being at: `Nov 4, 2024 @ 03:35:45` (1730691345628). A heartbeat is simply a file with a timestamp value that is constantly updated.
- ~2 minutes later we see that this session was reported as crashed at `Nov 4, 2024 @ 03:37:32`, but the timestamp on the heartbeat file it read was `Nov 4, 2024 @ 03:05:45` (1730689545627)
- This does not make sense since the latest heartbeat was at `3:35`, but what was seen was `3:05`. 2 more heartbeats are known to have happened after `3:05` based on telemetry, but it looks like they somehow were not seen when the crash check happened

We already do handling for this edge case by...
  - After the heartbeat file is written, we immediately read to ensure it returns the content we just wrote
  - If a heartbeat write fails for any reason, we terminate all crash monitoring for that session and clean it up so that there is not a chance for it to be falsely reported as a crash.
  - For any error that happens, we report a telemetry event and collect them in a graph to see any significant ones 

## Solution

Even with the handling above we were still seeing odd amounts of sessions being reported as crashed, even though heartbeats were appropriately being sent.

A guess to why we still had issues is that even though a new heartbeat file write is successful, it does not truly propagate to all readers of it. `fsync` is a known solution for "finalizing" the change to disk as certain OS's may do things like caching a write and then eventually writing it to disk.

The solution we have is to use `fsync` after writing the heartbeat file. We will then monitor our telemetry dashboard to see if these issues drop.

### Additional

- On HB file deletions we will now clear the content of it first, then delete. This is due to an assumption that something is reading the previous text from the file even after it was deleted, such as an open file handle existing before delete.
  - Even if we have an empty HB file on disk [we handle that case gracefully](https://github.com/aws/aws-toolkit-vscode/blob/55e0b83aa13a09b49af5fe4db5b0d8879fd6f1dd/packages/core/src/shared/crashMonitoring.ts#L590)

---

### DIFF
--- a/packages/core/src/shared/crashMonitoring.ts
+++ b/packages/core/src/shared/crashMonitoring.ts
@@ -471,7 +471,7 @@ export class FileSystemState {
                 fileHandle = await nodeFs.open(filePath, 'w')
                 await fileHandle.writeFile(JSON.stringify({ ...this.ext, lastHeartbeat: now }, undefined, 4))
                 // Noticing that some file reads are not immediately available after write. `fsync` is known to address this.
-                await fileHandle!.sync()
+                await fileHandle.sync()
                 await fileHandle.close()
                 fileHandle = undefined
 
@@ -523,9 +523,17 @@ export class FileSystemState {
         // IMPORTANT: Must use NodeFs here since this is used during shutdown
         const func = async () => {
             const filePath = this.makeStateFilePath(ext)
+
             // Even when deleting a file, if there is an open file handle it may still exist. This empties the
             // contents, so that any following reads will get no data.
-            await nodeFs.writeFile(filePath, '')
+            let fileHandle: nodeFs.FileHandle | undefined
+            try {
+                fileHandle = await nodeFs.open(filePath, 'w')
+                await fileHandle.sync()
+            } finally {
+                await fileHandle?.close()
+            }
+
             await nodeFs.rm(filePath, { force: true })
         }
         const funcWithCtx = () => withFailCtx(ctx, func)
--- a/packages/core/src/shared/crashMonitoring.ts
+++ b/packages/core/src/shared/crashMonitoring.ts
@@ -179,7 +179,7 @@ class Heartbeat {
 
     public async start() {
         // Send an initial heartbeat immediately
-        await withFailCtx('initialSendHeartbeat', () => this.state.sendHeartbeat())
+        await withFailCtx('sendHeartbeatInitial', () => this.state.sendHeartbeat())
 
         // Send a heartbeat every interval
         this.intervalRef = globals.clock.setInterval(async () => {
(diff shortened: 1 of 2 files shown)


jpinkney-aws (approved):

packages/core/src/shared/crashMonitoring.ts
```diff
@@ -458,32 +462,34 @@ export class FileSystemState {
...
-                // So this is a sanity check.
-                await fs.delete(filePath, { force: true })
-
-                await fs.writeFile(filePath, JSON.stringify({ ...this.ext, lastHeartbeat: now }, undefined, 4))
-
-                // Sanity check to verify the write is accessible immediately after
+                fileHandle = await nodeFs.open(filePath, 'w')
+                await fileHandle.writeFile(JSON.stringify({ ...this.ext, lastHeartbeat: now }, undefined, 4))
+                // Noticing that some file reads are not immediately available after write. `fsync` is known to address this.
+                await fileHandle!.sync()
```
justinmk3, line 474:
```suggestion
                await fileHandle?.sync()
```

packages/core/src/shared/crashMonitoring.ts
```diff
@@ -458,32 +462,34 @@ export class FileSystemState {
...
+                await fileHandle.close()
+                fileHandle = undefined
+
+                // Sanity check to verify the latest write is accessible immediately
                 const heartbeatData = JSON.parse(await fs.readFileText(filePath)) as ExtInstanceHeartbeat
```
justinmk3, line 479: Can you confirm that here, `fs` is the vscode VFS while just above we are using the nodefs. If so, that would explain why the write is not necessarily immediately seen: the vscode VFS could have a extra cache layer or other indirection.

Is it actually required/intentional to use `fs` here or should we use nodefs for this part?

packages/core/src/shared/crashMonitoring.ts
```diff
@@ -515,7 +521,13 @@ export class FileSystemState {
 
     private async deleteHeartbeatFile(ext: ExtInstanceId | ExtInstance, ctx: string) {
         // IMPORTANT: Must use NodeFs here since this is used during shutdown
-        const func = () => nodeFs.rm(this.makeStateFilePath(ext), { force: true })
+        const func = async () => {
+            const filePath = this.makeStateFilePath(ext)
+            // Even when deleting a file, if there is an open file handle it may still exist. This empties the
+            // contents, so that any following reads will get no data.
+            await nodeFs.writeFile(filePath, '')
```
justinmk3, line 528: use truncate()? https://nodejs.org/api/fs.html#fstruncatepath-len-callback

packages/core/src/shared/crashMonitoring.ts
nkomonen-amazon: No explicit need for the VSC FS api here. I'll look to reduce its usage in case there is some layer of indirection underneath

packages/core/src/shared/crashMonitoring.ts
nkomonen-amazon: Found we can use `open()` with a `w` flag to do this. This also allows us to use `sync()` in this, which I should have done initially.

Issue #5495, opened 2024-08-24
ci: "Failed to watch /…/node_modules/…/package.json using fs.watch() … ENOSPC"


justinmk3:
## Problem

The `Linux Unit Tests` CI jobs show these (non-fatal, currently) errors indicating that something is crawling files in `node_modules` and trying to watch `package.json`

```
[File Watcher (parcel)] Inotify limit reached (ENOSPC) (path: /codebuild/output/src4054755577/src/github.com/aws/aws-toolkit-vscode/node_modules)
  file_editAwsFile telemetry
[File Watcher (node.js)] Failed to watch /codebuild/output/src4054755577/src/github.com/aws/aws-toolkit-vscode/node_modules/@types/readline-sync/package.json for changes using fs.watch() (Error: ENOSPC: System limit for number of file watchers reached, watch '/codebuild/output/src4054755577/src/github.com/aws/aws-toolkit-vscode/node_modules/@types/readline-sync/package.json')
[File Watcher (node.js)] Failed to watch /codebuild/output/src4054755577/src/github.com/aws/aws-toolkit-vscode/node_modules/@types/responselike/package.json for changes using fs.watch() (Error: ENOSPC: System limit for number of file watchers reached, watch '/codebuild/output/src4054755577/src/github.com/aws/aws-toolkit-vscode/node_modules/@types/responselike/package.json')
[File Watcher (node.js)] Failed to watch /codebuild/output/src4054755577/src/github.com/aws/aws-toolkit-vscode/node_modules/@types/retry/package.json for changes using fs.watch() (Error: ENOSPC: System limit for number of file watchers reached, watch '/codebuild/output/src4054755577/src/github.com/aws/aws-toolkit-vscode/node_modules/@types/retry/package.json')
[File Watcher (node.js)] Failed to watch /codebuild/output/src4054755577/src/github.com/aws/aws-toolkit-vscode/node_modules/@types/semver/package.json for changes using fs.watch() (Error: ENOSPC: System limit for number of file watchers reached, watch '/codebuild/output/src4054755577/src/github.com/aws/aws-toolkit-vscode/node_modules/@types/semver/package.json')
Unable to load and parse grammar for scope source.ssmjson from file:///codebuild/output/src4054755577/src/github.com/aws/aws-toolkit-vscode/packages/core/syntaxes/SSMJSON.tmLanguage { fileOperationResult: 1, options: { preferUnbuffered: true } }
    ✔ emits when opened by user (488ms)
[...]

Issue #5325, opened 2024-07-16
unreliable test: beforeEach..closeAllEditors failure in Amazon Q tests


justinmk3:
In CI, the Amazon Q tests sometimes fail : 

```
  1) "before each" hook: beforeEach for "simple case 1":
     Error: Editors were still open after closeAllEditors():

      at Object.closeAllEditors (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\core\src\test\testUtil.ts:401:15)
      at runNextTicks (node:internal/process/task_queues:60:5)
      at processTimers (node:internal/timers:511:9)
      at async Context.beforeEach (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\core\src\test\globalSetup.test.ts:91:9)
```

These test results are near the top of the build logs. So although the final (core/) results report success, you need to scroll up to the amazonq/ tests.

Example: https://github.com/aws/aws-toolkit-vscode/actions/runs/9959760755/job/27517343024?pr=5309

justinmk3:
still occurs, though the line numbers have changed

```
  1) "before each" hook: beforeEach for "should not remove extra closing symbol":
     Error: Editors were still open after closeAllEditors():

      at Object.closeAllEditors (/codebuild/output/src1772754914/src/github.com/aws/aws-toolkit-vscode/packages/core/src/test/testUtil.ts:510:15)
```

Hweinstock:
Should be fixed by: https://github.com/aws/aws-toolkit-vscode/pull/5975

PR #5944, opened 2024-11-07
telemetry(amazonq): Add more information for outlier chat round trips

jpinkney-aws:
## Problem
It's hard to diagnose why we sometimes get some extreme values for a chat round trip. One idea is that some remote instances might be hosted in a different region and that's causing the increased values in some rare instances

## Solution
Add the stringified version to the reasonDesc field temporarily so that we can get more information. This would look like:
```
'{"chatMessageSent":1730994444010,"editorReceivedMessage":1730994444011,"featureReceivedMessage":1730994444013,"messageDisplayed":1730994459698}',
```

## Additional info
The numbers picked were pretty arbitrary `(x > 30000 || x < -30000)` but it should give us a sampling of 50+ examples in the next week. After that we can revert this code since we should have all the data we need to see if this is the issue

### DIFF
--- a/packages/core/src/amazonq/messages/chatMessageDuration.ts
+++ b/packages/core/src/amazonq/messages/chatMessageDuration.ts
@@ -83,16 +83,28 @@ export class AmazonQChatMessageDuration {
 
             // TODO: handle onContextCommand round trip time
             if (metrics.trigger !== 'onContextCommand') {
+                const editorReceivedMessage = durationFrom('chatMessageSent', 'editorReceivedMessage')
+                const featureReceivedMessage = durationFrom('editorReceivedMessage', 'featureReceivedMessage')
+                const messageDisplayed = durationFrom('featureReceivedMessage', 'messageDisplayed')
+                let reasonDesc = undefined
+
+                /**
+                 * Temporary include more information about outliers so that we can find out if the messages
+                 * aren't being sent or the user is actually doing a different chat flow
+                 */
+                if ([editorReceivedMessage, featureReceivedMessage].some((val) => val > 30000 || val < -30000)) {
+                    reasonDesc = JSON.stringify(metrics.events)
+                }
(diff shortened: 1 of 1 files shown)


packages/core/src/amazonq/messages/chatMessageDuration.ts
```diff
@@ -83,16 +83,28 @@ export class AmazonQChatMessageDuration {
...
-                        durationFrom('editorReceivedMessage', 'featureReceivedMessage') ?? -1,
-                    amazonqMessageDisplayedMs: durationFrom('featureReceivedMessage', 'messageDisplayed') ?? -1,
+                    amazonqEditorReceivedMessageMs: editorReceivedMessage ?? -1,
+                    amazonqFeatureReceivedMessageMs: featureReceivedMessage ?? -1,
+                    amazonqMessageDisplayedMs: messageDisplayed ?? -1,
                     source: metrics.trigger,
                     duration: totalDuration,
                     result: 'Succeeded',
                     traceId: metrics.traceId,
+                    ...(reasonDesc !== undefined ? { reasonDesc } : {}),
```
nkomonen-amazon, line 107: I'm pretty sure that if reasonDesc is `undefined` it just gets dropped. So would `reasonDesc: reasonDesc` work?

nkomonen-amazon (approved):

packages/core/src/amazonq/messages/chatMessageDuration.ts
jpinkney-aws: We want it to be undefined so that we don't send information when its not needed

justinmk3 (approved):

justinmk3:
Linux CI failures are unrelated. Tracked by:

- https://github.com/aws/aws-toolkit-vscode/issues/5495
- https://github.com/aws/aws-toolkit-vscode/issues/5325

PR #5947, opened 2024-11-07
test(notifications): unit tests for notification rendering

tomcat323:
Unit testing for notifications rendering

Using locally generated notifications, test covers:

- new notification on-receive modal/toast windows
- panel on-click txt/url pop up
- modal button actions (update&restart, txt, url)


packages/core/src/test/notifications/rendering.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    // util to test open url under different senarios
+    async function verifyOpenExternalUrl(notification: ToolkitNotification) {
+        const url = vscode.Uri.parse('https://aws.amazon.com/visualstudiocode/')
+        //const openExternalStub = getOpenExternalStub()
+        const openUrlStub = sandbox.stub(VsCodeUtils, 'openUrl')
+        await panelNode.openNotification(notification)
+
+        assert.ok(openUrlStub.calledOnce)
+        assert.ok(openUrlStub.calledWith(url))
+    }
```
justinmk3, line 47: This is a lot of code just to check that a url was clicked. 

One benefit of having shared functions such as `openUrl`, is that we don't need to write tests for them over and over. If a button only calls openUrl, we don't need tests for that.

If there is other state that we can check (and the function can return as a result) when the button is clicked, test for that instead.

justinmk3: `getTestWindow` may already have stuff that can check that the selection was clicked. Other than that, we don't need to check that `openUrl` opens a URL.

packages/core/src/test/notifications/rendering.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    })
+
+    afterEach(function () {
+        sandbox.restore()
+    })
+
+    // util to test txt pop-up under different senarios
+    async function verifyTxtNotification(notification: ToolkitNotification) {
+        const expectedContent = notification.uiRenderInstructions.content['en-US'].description
+        const readonlyDocumentShowStub = sandbox.stub(readonlyDocument, 'show').resolves()
```
justinmk3, line 30: vscode is showing a document. we can just check that actual fact. we don't need stubs and mocks for that.

`assertTextEditorContains` may be useful, for example. https://github.com/aws/aws-toolkit-vscode/blob/d4d1ae0d065edc4ac795de32996110efb7dbf0dc/packages/core/src/test/testUtil.ts#L381

justinmk3 (approved):

PR #5949, opened 2024-11-07
fix(amazonq): tutorial always showing on start

grant0417:
The LineAnnotationController could get stuck showing the tutorial for new users due to changes in the inline chat logic, this PR resolve that bug and prevents the `aws.codewhisperer.tutorial.workInProgress` from being set to `true` when using inline chat.

Additional a more descriptive error log for the `computeDiffAndRenderOnEditor` function call and check for not null in `updateTaskAndLenses`

### DIFF
--- a/packages/amazonq/.changes/next-release/Bug Fix-46c5fb9b-1b36-4826-8520-ec03409bce79.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-46c5fb9b-1b36-4826-8520-ec03409bce79.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Resolve tutorial always showing on start"
+}
--- a/packages/amazonq/.changes/next-release/Bug Fix-46c5fb9b-1b36-4826-8520-ec03409bce79.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-46c5fb9b-1b36-4826-8520-ec03409bce79.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Resolve tutorial always showing on start"
+	"description": "tutorial always showing on start"
 }
--- a/packages/amazonq/src/inlineChat/controller/inlineChatController.ts
+++ b/packages/amazonq/src/inlineChat/controller/inlineChatController.ts
@@ -122,6 +122,9 @@ export class InlineChatController {
     }
 
     public async updateTaskAndLenses(task: InlineTask, taskState?: TaskState) {
+        if (!task) {
+            return
+        }
         if (taskState) {
             task.state = taskState
         } else if (!task.diff || task.diff.length === 0) {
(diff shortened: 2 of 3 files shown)


packages/amazonq/.changes/next-release/Bug Fix-46c5fb9b-1b36-4826-8520-ec03409bce79.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Resolve tutorial always showing on start"
```
justinmk3, line 3: Thanks for including a changelog. Please review the [changelog guidelines](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#changelog).

packages/amazonq/src/inlineChat/controller/inlineChatController.ts
```diff
@@ -122,6 +122,9 @@ export class InlineChatController {
     }
 
     public async updateTaskAndLenses(task: InlineTask, taskState?: TaskState) {
+        if (!task) {
```
justinmk3, line 125: How is this possible? The parameter type is non-nullable.

packages/core/src/codewhisperer/views/lineAnnotationController.ts
```diff
@@ -201,16 +203,15 @@ export class EndState implements AnnotationState {
 }
 
 export class InlineChatState implements AnnotationState {
-    static static = 'amazonq_annotation_inline_chat'
-    id = InlineChatState.static
+    static id = 'amazonq_annotation_inline_chat'
+    id = InlineChatState.id
```
justinmk3, line 207: The old code assigned a new variable. Now it looks like `id` is immediately reassigned (or is it an instance variable)? What is the intention here?

packages/amazonq/src/inlineChat/controller/inlineChatController.ts
grant0417: Unclear when this could happen. This is an attempt to fix an issue reported by a customer where I believe the logic in this function is running with a undefined task, I can not find any other codepath that would produce this error:

Screenshot 2024-11-12 at 9 09 30 AM

I can make the task optional to make this clearer at least.

justinmk3: > I can make the task optional to make this clearer at least.

That is the bare minimum, yes.

The cause of this bug is likely related to the fact that `task = this.task` may be undefined. https://github.com/aws/aws-toolkit-vscode/blob/11536586aa8dce8259748977cdf2e3dcaa0441bf/packages/amazonq/src/inlineChat/controller/inlineChatController.ts#L310

It's probably a good idea to track down the root cause, because this signals that there are other parts of the code that are forcing things (e.g. with `!`) that they shouldn't be forcing. Or there is unclear initialization ordering.

justinmk3: Is this code still missing tests?

justinmk3 (approved):

PR #5950, opened 2024-11-07
ci(jscpd): check changed lines instead of changed files to avoid false positives

Hweinstock:
## Problem
Currently `jscpd` ci action filters on changed files in PR. This results in clones being returned from changes outside our changes (but within the same files). 

`jscpd` is the most commonly failing CI task at the moment, so to keep CI green we want to reduce false positives. 

## Solution
- `jscpd` only lists clones that includes lines from the changed lines in the changed files.
- Included in this PR is a refactor of the action to simplify the bash scripting done in the action yml file. 
- Some tests for the `filterDuplicates.ts` script to make future changes easier. 

See https://github.com/aws/aws-toolkit-vscode/pull/5950/commits/12fc4d648f82b1fc912584f0c089164df0d2f976 for an example failure. 

---


.github/workflows/filterDuplicates.js
```diff
@@ -0,0 +1,268 @@
+const fs = require('fs/promises')
```
justinmk3, line 1: 1 or 2 sentences are a cheap way to give insight to someone trying to understand this script. Also a USAGE section is good practice.

At a glance I guess something like:

```
// Modifies a `jscpd-report.json` produced by `jscpd`. Removes X based on Y from
// a diff produced by `git diff`.
//
// Usage:
//      node filterDuplicates.js run diff_output.txt jscpd-report.json
// ...
```

.github/workflows/filterDuplicates.js
```diff
@@ -0,0 +1,268 @@
...
+ */
+let testCounter = 0
+function assertEqual(actual, expected) {
+    if (actual !== expected) {
+        throw new Error(`Expected ${expected} but got ${actual}`)
+    }
+    testCounter += 1
+}
+
+async function test() {
```
justinmk3, line 101: Nice 👍

justinmk3 (approved):

PR #5952, opened 2024-11-07
chore(amazonq): update user-service-2.json

leigaol:
## Solution

---

### DIFF
--- a/packages/core/src/codewhisperer/client/user-service-2.json
+++ b/packages/core/src/codewhisperer/client/user-service-2.json
@@ -18,13 +18,25 @@
                 "method": "POST",
                 "requestUri": "/"
             },
-            "input": { "shape": "CreateUploadUrlRequest" },
-            "output": { "shape": "CreateUploadUrlResponse" },
+            "input": {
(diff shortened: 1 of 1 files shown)


fangwade007 (approved):

justinmk3:
ci is failing

packages/core/src/codewhisperer/client/user-service-2.json
```diff
@@ -18,13 +18,25 @@
...
+                },
+                {
+                    "shape": "InternalServerException"
+                },
+                {
+                    "shape": "ValidationException"
+                },
+                {
+                    "shape": "AccessDeniedException"
+                }
```
Will-ShaoHua, line 39: can we use the file with the same formatting?

Will-ShaoHua: diff is too large and most of them are formatting changes

Will-ShaoHua: hardly can tell which changes are the purpose of this pr

Will-ShaoHua:
#5695

PR #5955, opened 2024-11-07
test(webview): skip test on old vscode versions to avoid flakiness. 

Hweinstock:
## Problem 
It appears that older versions of VSCode are causing this test to be flaky, which appears to be fixed in v1.88. See findings section for more information. 

## Solution
- Skip the test if vscode version is pre-1.88. 
- Add tech-debt test to remove this check once min is 1.88+. 

## Findings
- Ran 1000 times in CI, and failed 29 times. (2.9% failure rate). If we re-reran 3 times, ci would pass ~ 99.998% of the time (assuming failures independent). 
- Only fails on VSCode minimum (1.83.0).
- Originally thought it was related to activating Q extension or fetching it from CDN, but appears mocking this still results in flakiness.  
- Every failure comes with a `ERROR:gles2_command_buffer_stub.cc(364)] ContextResult::kFatalFailure Failed to create context.` which appears to be from OpenGL. This means it is coming from very far down the stack since VSCode is built on electron, which uses chromium, which is then using OpenGL. (ai suggests this is a GPU problem?). 
- Thought to be from race condition with async constructor hack, fixed by https://github.com/aws/aws-toolkit-vscode/pull/5961. 
- Evidence is suggesting that this could be a bug in a prior version of VSCode, electron, or OpenGL, and should go away once we bump minimum VSCode version. 
- Works on VSCode 1.90+, 1.87- fails, 1.88 is the minimum version it is not flaky. (verified using 1000 attempts). 
- Failures are NOT independent since retrying 10x times only reduced failure rate to ~1%. Therefore retries are not a viable option. 

---

### DIFF
--- a/packages/core/src/test/applicationcomposer/messageHandlers/generateResourceHandler.test.ts
+++ b/packages/core/src/test/applicationcomposer/messageHandlers/generateResourceHandler.test.ts
@@ -15,7 +15,7 @@ describe.only('generateResourceHandler', function () {
     afterEach(() => {
         sinon.restore()
     })
-    for (const _ of Array.from({ length: 100 }, (i) => i)) {
+    for (const _ of Array.from({ length: 1000 }, (i) => i)) {
         it('amazon q is not installed', async () => {
             sinon.stub(extApi, 'getAmazonqApi')
             const panel = await createTemplate()
--- a/packages/core/src/test/applicationcomposer/messageHandlers/generateResourceHandler.test.ts
+++ b/packages/core/src/test/applicationcomposer/messageHandlers/generateResourceHandler.test.ts
@@ -38,6 +38,7 @@ describe('generateResourceHandler', function () {
             console.log('post-generateResourceHandler')
             assert.ok(postMessageSpy.calledOnce)
             assert.deepStrictEqual(postMessageSpy.getCall(0).args[0].isSuccess, false)
+            sinon.restore()
         }
     })
 })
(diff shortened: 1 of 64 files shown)


justinmk3 (approved):

packages/core/src/test/applicationcomposer/messageHandlers/generateResourceHandler.test.ts
```diff
@@ -5,16 +5,18 @@
...
-    afterEach(() => {
-        sinon.restore()
-    })
-
-    it('amazon q is not installed', async () => {
+    it('amazon q is not installed', async function () {
+        const version = env.getMinVscodeVersion()
+        if (semver.lt(version, '1.88.0')) {
+            this.skip()
+        }
```
justinmk3, line 19: can use `isMinVscode('1.89.0')` https://github.com/aws/aws-toolkit-vscode/blob/d4d1ae0d065edc4ac795de32996110efb7dbf0dc/packages/core/src/shared/vscode/env.ts#L70

and that eliminates the need for the techdebt test (though that's a good instinct).

```suggestion
        if (isMinVscode('1.89.0')) {
            this.skip()
        }
```

p.s. `isMinVscode` is kinda hard to understand, maybe it should check `semver.gt` instead of `semver.gte`.

justinmk3 (approved):

packages/core/src/test/applicationcomposer/messageHandlers/generateResourceHandler.test.ts
Hweinstock: Yeah, agree on the function. It feels like its doing two things at once: checking if we are on minimum version, and throwing an error when we exceed a threshold. Its not entirely clear to me that these tasks are related, so could be worth splitting/rewording for readability sake.

justinmk3 (approved):

PR #5956, opened 2024-11-07
telemetry: add notifications telemetry

hayemaxi:
I explored sending telemetry if the user hides or unhides the notification view panel. However, there is no VSC hook to detect this and other ways (e.g. checking visible property) are not reliable. So, we cannot track this.

- toolkit_showNotification
  - for displaying notifications on receive or when clicked on
- toolkit_invokeAction
  - invoking an action from a notification, e.g. clicking on a button in the notification or opening a text document
- ui_click for clicking on the notifications node or dismiss button

---

### DIFF
--- a/packages/core/src/notifications/controller.ts
+++ b/packages/core/src/notifications/controller.ts
@@ -7,7 +7,13 @@ import * as vscode from 'vscode'
 import { ToolkitError } from '../shared/errors'
 import globals from '../shared/extensionGlobals'
 import { globalKey } from '../shared/globalState'
-import { NotificationsState, NotificationsStateConstructor, NotificationType, ToolkitNotification } from './types'
+import {
+    getNotificationTelemetryId,
+    NotificationsState,
+    NotificationsStateConstructor,
+    NotificationType,
+    ToolkitNotification,
+} from './types'
 import { HttpResourceFetcher } from '../shared/resourcefetcher/httpResourceFetcher'
 import { getLogger } from '../shared/logger/logger'
 import { NotificationsNode } from './panelNode'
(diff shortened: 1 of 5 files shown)


packages/core/src/notifications/controller.ts
```diff
@@ -202,6 +209,10 @@ function registerDismissCommand() {
                 // The command used to build the TreeNode contains the notification as an argument.
                 /** See {@link NotificationsNode} for more info. */
                 const notification = item.command?.arguments[0] as ToolkitNotification
+                telemetry.ui_click.emit({
+                    elementId: `${getNotificationTelemetryId(notification)}:DISMISS`,
+                    result: 'Succeeded',
+                })
 
                 await NotificationsController.instance.dismissNotification(notification.id)
```
justinmk3, line 217: is it worth forming it as a run() around this call?

packages/core/src/notifications/panelNode.ts
```diff
@@ -27,25 +28,25 @@ export class NotificationsNode implements TreeNode {
...
     private constructor() {
         this.openNotificationCmd = Commands.register(
             isAmazonQ() ? '_aws.amazonq.notifications.open' : '_aws.toolkit.notifications.open',
-            async (n: ToolkitNotification) => this.openNotification(n)
+            (n: ToolkitNotification) => {
+                telemetry.ui_click.emit({
+                    elementId: getNotificationTelemetryId(n),
+                    result: 'Succeeded',
+                })
+                return this.openNotification(n)
```
justinmk3, line 48: `telemetry.ui_click.run()` ?

packages/core/src/notifications/panelNode.ts
```diff
@@ -160,57 +161,65 @@ export class NotificationsNode implements TreeNode {
...
-                        break
-                    case 'updateAndReload':
-                        await this.updateAndReload(notification.displayIf.extensionId)
-                        break
-                    case 'openUrl':
-                        if (selectedButton.url) {
-                            await openUrl(vscode.Uri.parse(selectedButton.url))
-                        } else {
-                            throw new ToolkitError('url not provided')
+        return vscode.window
```
justinmk3, line 185: Future: could use https://github.com/aws/aws-toolkit-vscode/pull/5776 here.

justinmk3 (approved):

PR #5958, opened 2024-11-08
feat(cwl): Add clear screen and stop session code lens

keeganirby:
## Problem
Users may want to clear their screen during a tailing session. The liveTail document is read only.
Users may want to stop their session without closing the document. 

## Solution
Provide a codeLens to clear the screen (make document empty)
Provide a codeLens to stop the session without closing the document. 

Moves Interval Timer for updating the StatusBar to the LiveTailSession object so `stopLiveTailSession()` can interrupt it, and be guaranteed to be cleaned up.

The TextDocument's URI and Session URI seem to not be equal. Looking up in the LiveTailSessionRegistry with a document URI causes a session to not be found. Converting with `toString` allows these URIs to match and the registry to work as intended. 

Improves Exception handling on-stream. Previously, stopping the session in an expected fashion (codeLens, Closing editors) would cause an exception to bubble up and appear to the User. New change recognizes when the Abort Controller triggers, signifying an error has not occured, and logs the event as opposed to surfacing an error. Exposing only `isAborted` so that a caller has to use `stopLiveTailSession` to trigger the abortController and guarantee that other clean up actions have taken place. 

---

### DIFF
--- a/packages/core/src/awsService/cloudWatchLogs/activation.ts
+++ b/packages/core/src/awsService/cloudWatchLogs/activation.ts
@@ -19,13 +19,14 @@ import { searchLogGroup } from './commands/searchLogGroup'
 import { changeLogSearchParams } from './changeLogSearch'
 import { CloudWatchLogsNode } from './explorer/cloudWatchLogsNode'
 import { loadAndOpenInitialLogStreamFile, LogStreamCodeLensProvider } from './document/logStreamsCodeLensProvider'
-import { tailLogGroup } from './commands/tailLogGroup'
+import { clearDocument, closeSession, tailLogGroup } from './commands/tailLogGroup'
 import { LiveTailDocumentProvider } from './document/liveTailDocumentProvider'
 import { LiveTailSessionRegistry } from './registry/liveTailSessionRegistry'
 import { DeployedResourceNode } from '../appBuilder/explorer/nodes/deployedNode'
 import { isTreeNode } from '../../shared/treeview/resourceTreeDataProvider'
 import { getLogger } from '../../shared/logger/logger'
 import { ToolkitError } from '../../shared'
+import { LiveTailCodeLensProvider } from './document/liveTailCodeLensProvider'
 
 export async function activate(context: vscode.ExtensionContext, configuration: Settings): Promise<void> {
     const registry = LogDataRegistry.instance
@@ -46,6 +47,16 @@ export async function activate(context: vscode.ExtensionContext, configuration: 
 
     context.subscriptions.push(
         vscode.workspace.registerTextDocumentContentProvider(CLOUDWATCH_LOGS_SCHEME, documentProvider)
+    )
+
+    context.subscriptions.push(
+        vscode.languages.registerCodeLensProvider(
+            {
+                language: 'log',
+                scheme: cloudwatchLogsLiveTailScheme,
+            },
+            new LiveTailCodeLensProvider()
+        )
     )
 
     context.subscriptions.push(
@@ -110,6 +121,14 @@ export async function activate(context: vscode.ExtensionContext, configuration: 
                     ? { regionName: node.regionCode, groupName: node.logGroup.logGroupName! }
                     : undefined
             await tailLogGroup(liveTailRegistry, logGroupInfo)
+        }),
+
+        Commands.register('aws.cwl.stopTailingLogGroup', async (document: vscode.TextDocument) => {
+            closeSession(document.uri, liveTailRegistry)
+        }),
+
+        Commands.register('aws.cwl.clearDocument', async (document: vscode.TextDocument) => {
+            await clearDocument(document)
         }),
 
         Commands.register('aws.appBuilder.searchLogs', async (node: DeployedResourceNode) => {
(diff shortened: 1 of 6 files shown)


jpinkney-aws (approved):

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -32,32 +32,29 @@ export async function tailLogGroup(
         region: wizardResponse.regionLogGroupSubmenuResponse.region,
     }
     const session = new LiveTailSession(liveTailSessionConfig)
-    if (registry.has(session.uri)) {
+    if (registry.has(session.uri.toString())) {
```
justinmk3, line 35: `uriToKey` exists for this purpose. https://github.com/aws/aws-toolkit-vscode/blob/d4d1ae0d065edc4ac795de32996110efb7dbf0dc/packages/core/src/awsService/cloudWatchLogs/cloudWatchLogsUtils.ts#L33

`toString` is not reliable, querystring parameters can appear in arbitrary order.

justinmk3: https://github.com/aws/aws-toolkit-vscode/blob/d4d1ae0d065edc4ac795de32996110efb7dbf0dc/packages/core/src/awsService/cloudWatchLogs/registry/logDataRegistry.ts#L51-L53

This makes me wonder why `LiveTailSessionRegistry` is using a completely different pattern than `LogDataRegistry`. Now we have two different "registries" concepts to maintain, using completely different patterns.

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -100,8 +96,21 @@ async function handleSessionStream(
...
         }
-    } finally {
-        globals.clock.clearInterval(timer)
+    } catch (e) {
+        if (session.isAborted) {
+            //Expected case. User action cancelled stream (CodeLens, Close Editor, etc.).
+            //AbortSignal interrupts the LiveTail stream, causing error to be thrown here.
+            //Can assume that stopLiveTailSession() has already been called - AbortSignal is only
+            //exposed through that method.
+            getLogger().info(`Session ${session.uri.toString()} stopped.`)
```
justinmk3, line 105: put variable info at the end of the message. please read the project guidelines, which mention this.
```suggestion
            getLogger().info(`Session stopped:  ${session.uri.toString()}`)
```

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -100,8 +96,21 @@ async function handleSessionStream(
...
+        } else {
+            //Unexpected exception.
+            session.stopLiveTailSession()
+            throw ToolkitError.chain(
+                e,
+                `Unexpected on-stream execption while tailing session: ${session.uri.toString()}`
```
justinmk3, line 111:
```suggestion
                `Unexpected on-stream exception while tailing session: ${session.uri.toString()}`
```

packages/core/src/awsService/cloudWatchLogs/activation.ts
```diff
@@ -48,6 +49,16 @@ export async function activate(context: vscode.ExtensionContext, configuration:
         vscode.workspace.registerTextDocumentContentProvider(CLOUDWATCH_LOGS_SCHEME, documentProvider)
     )
 
+    context.subscriptions.push(
+        vscode.languages.registerCodeLensProvider(
+            {
+                language: 'log',
+                scheme: cloudwatchLogsLiveTailScheme,
+            },
+            new LiveTailCodeLensProvider()
```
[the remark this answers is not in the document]
keeganirby: The CodeLenses in the existing CWL experiences are not very applicable to our LiveTail use case. 
"Load newer" & "Load older" don't apply when data is being streamed in constantly.  For `View full log stream`, customers can apply LogStream filter parameters to their liveTail session already so having this as a codeLens in a Livetail session doesn't seem necessary. 
On the other hand, a `SearchLogGroup` command won't need a `Stop tailing` code lens. 

These UX are quite separate. 

We already separate the CodeLens providers for `logData` and `logStream`. Given the UX and functionality of LiveTail is different from our other experiences I think it fits to have a separate CodeLens provider.

This is the same reasoning behind having a separate scheme as we discussed [here](https://github.com/aws/aws-toolkit-vscode/pull/5790#discussion_r1809327913). The way we retrieve/display logs from a LiveTail session stream differs enough from the existing LogDataRegistry that I felt that it was best to separate these logic's, as opposed to finding a way to generalize.

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
keeganirby: I think the aim of these two registries are different from one another.
Given a LiveTail session has its own components unique to other CWL experiences (own client, abortController, responseStream, etc.) We have a `LiveTailSession` object, that has a long running lifeCycle. This lifecycle can end from outside the TailLogGroup command, and so this registry is really just there to track what Sessions are currently running, so when a session needs to be closed via a CodeLens, or closing its textEditors, we can find the LiveTailSession from its `uri` and abort the response stream and clean up its resources. 

I think this is a separate responsiblity from the LogDataRegistry that is actually persisting LogData, and making requests to FilterLogEvents.

PR #5961, opened 2024-11-08
refactor(webview): avoid race conditions in constructors. 

Hweinstock:
## Problem
In the constructor for `ApplicationComposerManager`, there is an async call (non-awaited). This call is modifying state within the object and therefore causes a race condition with any other function relying on this state. See https://github.com/aws/aws-toolkit-vscode/blob/81132884f4fb3319bda4be7d3d873265191f43ce/packages/core/src/applicationcomposer/webviewManager.ts#L30-L33

~This is potentially related to a flaky test.~ The test is still failing, but this refactor is still a win. 

## Solution
- Implement an async `create` method with a private constructor. Move this pattern upward to enable async calls. 

---

### DIFF
--- a/packages/core/src/applicationcomposer/composerWebview.ts
+++ b/packages/core/src/applicationcomposer/composerWebview.ts
@@ -5,37 +5,43 @@
 
 import { join } from 'path'
 import * as nls from 'vscode-nls'
-const localize = nls.loadMessageBundle()
 import * as path from 'path'
 import * as vscode from 'vscode'
 import { handleMessage } from './handleMessage'
 import { FileWatchInfo } from './types'
 
+const localize = nls.loadMessageBundle()
+
 export class ApplicationComposer {
     public readonly documentUri: vscode.Uri
-    public webviewPanel: vscode.WebviewPanel
+    public webviewPanel: vscode.WebviewPanel = {} as vscode.WebviewPanel
     protected readonly disposables: vscode.Disposable[] = []
     protected isPanelDisposed = false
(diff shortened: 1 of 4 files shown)


packages/core/src/applicationcomposer/composerWebview.ts
```diff
@@ -5,37 +5,43 @@
...
 import * as vscode from 'vscode'
 import { handleMessage } from './handleMessage'
 import { FileWatchInfo } from './types'
 
+const localize = nls.loadMessageBundle()
+
 export class ApplicationComposer {
     public readonly documentUri: vscode.Uri
-    public webviewPanel: vscode.WebviewPanel
+    public webviewPanel: vscode.WebviewPanel = {} as vscode.WebviewPanel
```
jpinkney-aws, line 17: is there any reason this can't be `public webviewPanel: vscode.WebviewPanel | undefined`? If the webview panel is an empty object does that make some assumptions down the line that it might be defined, even though it isn't?

Hweinstock: Yeah, I wasn't sure if there was a better way to handle this. Since the constructor is private, the only way to instantiate `ApplicationComposer` is through the `create` method. Thus, this field will always be initiated. Perhaps there is a better way to enforce this invariant? 

The advantage of an empty object is that I don't have to do an undefined check on each usage.

justinmk3 (approved):

JLargent42 (approved):

PR #5964, opened 2024-11-08
test(slowsam): samcli test speed up

### LINKED ISSUE #2675: unreliable test: SamCliLocalInvokeInvocation, ChildProcess tests are slow
(from Windows CI)

```
  SamCliLocalInvokeInvocation
    ✔ invokes `sam local` with args (7671ms)
    ✔ Passes template resource name to sam cli (1703ms)
    ✔ Passes template path to sam cli (1583ms)
    ✔ Passes event path to sam cli (2268ms)
    ✔ Passes env-vars path to sam cli (1570ms)
    ✔ Passes debug port to sam cli (1568ms)
    ✔ undefined debug port does not pass to sam cli (1568ms)
    ✔ Passes docker network to sam cli (1774ms)
    ✔ Does not pass docker network to sam cli when undefined (2261ms)
    ✔ passes --skip-pull-image to sam cli if skipPullImage is true (1568ms)
    ✔ does not pass --skip-pull-image to sam cli if skipPullImage is false (1565ms)
    ✔ does not pass --skip-pull-image to sam cli if skipPullImage is undefined (1570ms)
    ✔ Passes debuggerPath to sam cli (1769ms)
    ✔ Does not pass debuggerPath to sam cli when undefined (1570ms)
```

also ChildProcess : 

```
2022-06-03T02:45:01.7774675Z [0m  ChildProcess[0m
2022-06-03T02:45:01.7774976Z [0m    run[0m
2022-06-03T02:45:01.8543459Z     [32m  ✔[0m[90m starts and captures stdout - windows[0m[31m (705ms)[0m
2022-06-03T02:45:02.5709247Z     [32m  ✔[0m[90m runs cmd files containing a space in the filename and folder[0m[31m (704ms)[0m
2022-06-03T02:45:03.2533512Z     [32m  ✔[0m[90m errs when starting twice - windows[0m[31m (669ms)[0m
2022-06-03T02:45:04.2837880Z     [32m  ✔[0m[90m runs scripts containing a space in the filename and folder[0m[31m (1018ms)[0m
2022-06-03T02:45:05.4812674Z     [32m  ✔[0m[90m reports error for missing executable[0m[31m (1181ms)[0m
2022-06-03T02:45:05.5023757Z [0m      Extra options[0m
2022-06-03T02:45:06.5293470Z       [32m  ✔[0m[90m can report errors[0m[31m (1018ms)[0m
2022-06-03T02:45:07.8951017Z       [32m  ✔[0m[90m can reject on errors if `rejectOnError` is set[0m[31m (1353ms)[0m
2022-06-03T02:45:08.9303743Z       [32m  ✔[0m[90m kills the process if an error is reported[0m[31m (1020ms)[0m
2022-06-03T02:45:09.6517739Z       [32m  ✔[0m[90m can merge with base options[0m[31m (704ms)[0m
2022-06-03T02:45:10.3399460Z       [32m  ✔[0m[90m uses `Timeout` objects[0m[31m (672ms)[0m
2022-06-03T02:45:11.0542796Z       [32m  ✔[0m[90m still runs if the timer completed (not rejected) after starting[0m[31m (704ms)[0m
2022-06-03T02:45:11.0887307Z       [32m  ✔[0m[90m rejects if using a completed timer[0m
2022-06-03T02:45:11.0944344Z [0m    stop()[0m
2022-06-03T02:45:11.8746797Z     [32m  ✔[0m[90m detects running processes and successfully stops a running process - Windows[0m[31m (769ms)[0m
2022-06-03T02:45:12.6558170Z     [32m  ✔[0m[90m cannot stop() previously stopped processes - Windows[0m[31m (769ms)[0m
```

Hweinstock:
## Problem
- The test failure is due to timeout on Windows when first searching for the SAM cli path in `src/shared/sam/cli/samCliLocator.ts`. This is a known slow process. 
- The verification of SAM is also slow (100-200ms), even if result is cached and we do this each time we read from cache. 

## Solution
- Verify the cache once, rather than on each read. 
- Find sam cli in the before hook to allow all sam tests to run under same conditions. Otherwise, the first test is then responsible for actually finding the path causing it to take way longer than the rest of the tests. 
- Allow 3x retries on before hook on windows to avoid flakiness. 

## Notes
- An alternative solution could involve manually inserting into the cache, or mocking a piece of the component, but wanted to keep the test coverage as is. 
- This approach also allows us to isolate the finding code into its own test (rather than whatever the first test is) to see if there are performance regressions. 
- Additionally, retry mechanism was more effective than increasing the timeout in reducing flakiness. 

fix #2675

---

### DIFF
--- a/packages/core/src/shared/sam/cli/samCliLocalInvoke.ts
+++ b/packages/core/src/shared/sam/cli/samCliLocalInvoke.ts
@@ -6,7 +6,6 @@
 import * as proc from 'child_process'
 import { pushIf } from '../../utilities/collectionUtils'
 import * as nls from 'vscode-nls'
-import { fileExists } from '../../filesystemUtilities'
 import { getLogger, getDebugConsoleLogger, Logger } from '../../logger'
 import { ChildProcess } from '../../utilities/processUtils'
 import { Timeout } from '../../utilities/timeoutUtils'
@@ -15,6 +14,7 @@ import * as vscode from 'vscode'
 import globals from '../../extensionGlobals'
 import { SamCliSettings } from './samCliSettings'
 import { addTelemetryEnvVar, collectSamErrors, SamCliError } from './samCliInvokerUtils'
+import { fs } from '../..'
 
 const localize = nls.loadMessageBundle()
 
@@ -236,6 +236,8 @@ export class SamCliLocalInvokeInvocation {
         const sam = await this.config.getOrDetectSamCli()
         // eslint-disable-next-line aws-toolkits/no-console-log
         console.log('getOrDetect took %O seconds', (Date.now() - start) / 1000)
+        // eslint-disable-next-line aws-toolkits/no-console-log
+        console.log('autodetect is %O', sam.autoDetected)
         if (!sam.path) {
             getLogger().warn('SAM CLI not found and not configured')
         } else if (sam.autoDetected) {
@@ -290,11 +292,11 @@ export class SamCliLocalInvokeInvocation {
             throw new Error('template resource name is missing or empty')
         }
 
-        if (!(await fileExists(this.args.templatePath))) {
+        if (!(await fs.exists(this.args.templatePath))) {
             throw new Error(`template path does not exist: ${this.args.templatePath}`)
         }
 
-        if (this.args.eventPath !== undefined && !(await fileExists(this.args.eventPath))) {
+        if (this.args.eventPath !== undefined && !(await fs.exists(this.args.eventPath))) {
             throw new Error(`event path does not exist: ${this.args.eventPath}`)
         }
     }
--- a/packages/core/src/shared/sam/cli/samCliLocalInvoke.ts
+++ b/packages/core/src/shared/sam/cli/samCliLocalInvoke.ts
@@ -232,12 +232,7 @@ export class SamCliLocalInvokeInvocation {
 
     public async execute(timeout?: Timeout): Promise<ChildProcess> {
         await this.validate()
-        const start = Date.now()
         const sam = await this.config.getOrDetectSamCli()
-        // eslint-disable-next-line aws-toolkits/no-console-log
-        console.log('getOrDetect took %O seconds', (Date.now() - start) / 1000)
-        // eslint-disable-next-line aws-toolkits/no-console-log
-        console.log('autodetect is %O', sam.autoDetected)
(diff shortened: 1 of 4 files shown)


packages/core/src/shared/sam/cli/samCliLocalInvoke.ts
```diff
@@ -15,6 +14,7 @@ import * as vscode from 'vscode'
 import globals from '../../extensionGlobals'
 import { SamCliSettings } from './samCliSettings'
 import { addTelemetryEnvVar, collectSamErrors, SamCliError } from './samCliInvokerUtils'
+import { fs } from '../..'
```
hayemaxi, line 17: Avoid importing from `index.ts` files in core, it can cause circular dependencies (e.g. we export this file through that `index.ts` directly or indirectly). Instead import directly from the file.

packages/core/src/shared/sam/cli/samCliLocator.ts
```diff
@@ -31,11 +36,11 @@ export class SamCliLocationProvider {
         const cachedLoc = forceSearch ? undefined : SamCliLocationProvider.cachedSamLocation
 
         // Avoid searching the system for `sam` (especially slow on Windows).
-        if (cachedLoc && (await SamCliLocationProvider.isValidSamLocation(cachedLoc.path))) {
+        if (cachedLoc && (cachedLoc.verified || (await SamCliLocationProvider.isValidSamLocation(cachedLoc.path)))) {
+            cachedLoc.verified = true
```
hayemaxi, line 40: Seems like `cachedLoc.verified` should be a property of `SamCliLocationProvider`

packages/core/src/shared/sam/cli/samCliLocalInvoke.ts
Hweinstock: Out of curiosity, what purpose does `index.js` serve if its not supposed to be imported from?

justinmk3 (approved):

packages/core/src/shared/sam/cli/samCliLocalInvoke.ts
justinmk3: > Is it just for exporting to external modules?

Yes, it's for the `packages/amazon/` and `packages/toolkit/` modules, which depend on `packages/core/`.

PR #5965, opened 2024-11-08
config(amazonq): update Q API model

Will-ShaoHua:
## Problem
requested by partner team

## Solution

---

### DIFF
--- a/packages/core/src/codewhisperer/client/user-service-2.json
+++ b/packages/core/src/codewhisperer/client/user-service-2.json
@@ -2,11 +2,14 @@
     "version": "2.0",
     "metadata": {
         "apiVersion": "2022-11-11",
+        "auth": ["smithy.api#httpBearerAuth"],
         "endpointPrefix": "amazoncodewhispererservice",
         "jsonVersion": "1.0",
         "protocol": "json",
+        "protocols": ["json"],
         "serviceFullName": "Amazon CodeWhisperer",
         "serviceId": "CodeWhispererRuntime",
+        "signatureVersion": "bearer",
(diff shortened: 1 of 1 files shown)


packages/core/src/codewhisperer/client/user-service-2.json
```diff
@@ -2,11 +2,14 @@
...
     "metadata": {
         "apiVersion": "2022-11-11",
+        "auth": ["smithy.api#httpBearerAuth"],
         "endpointPrefix": "amazoncodewhispererservice",
         "jsonVersion": "1.0",
         "protocol": "json",
+        "protocols": ["json"],
         "serviceFullName": "Amazon CodeWhisperer",
         "serviceId": "CodeWhispererRuntime",
+        "signatureVersion": "bearer",
```
Will-ShaoHua, line 12: not sure do we need to include these 3 lines ornot

Will-ShaoHua: think it should be fine, confirmed jb has these. https://github.com/aws/aws-toolkit-jetbrains/blob/main/plugins/core/sdk-codegen/codegen-resources/codewhispererruntime/service-2.json#L4-L15

justinmk3 (approved):

Will-ShaoHua:
@justinmk3 fyi i will also request our service team to review the PR then merge

fangwade007 (approved):

packages/core/src/codewhisperer/client/user-service-2.json
dhasani23: Should not include

Issue #5970, opened 2024-11-11
unreliable test: "not connected to codewhisperer"


Hweinstock:
https://github.com/aws/aws-toolkit-vscode/blob/43f5b85f93f5017145ba5e6b140cce0b6ef906f8/packages/core/src/test/applicationcomposer/messageHandlers/initMessageHandler.test.ts#L16
```
1) initMessageHandler
       not connected to codewhisperer:
     Error: Test length exceeded max duration: 30 seconds
[No Pending UI Elements Found]
      at Timeout._onTimeout (/Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/core/src/test/setupUtil.ts:46:32)
      at listOnTimeout (node:internal/timers:569:17)
      at processTimers (node:internal/timers:512:7)
``` 
Ex. https://github.com/aws/aws-toolkit-vscode/actions/runs/11782092212/job/32816364633?pr=5961

PR #5973, opened 2024-11-11
test(appcomposer): fix initMessageHandler flaky test. 

Hweinstock:
## Problem
See issue: https://github.com/aws/aws-toolkit-vscode/issues/5970

## Solution
appears to be similar to https://github.com/aws/aws-toolkit-vscode/pull/5955, so same solution is employed. 
- skip test if below 1.89 vscode version. 
- add helper function `skipIfVscodeBelow` for tests that are flaky on previous versions. 

---


packages/core/src/test/applicationcomposer/messageHandlers/generateResourceHandler.test.ts
```diff
@@ -8,13 +8,11 @@ import sinon from 'sinon'
...
 import { Command, MessageType } from '../../../applicationcomposer/types'
-import * as env from '../../../shared/vscode/env'
+import { disableIfVscodeBelow } from '../../testUtil'
 
 describe('generateResourceHandler', function () {
     it('amazon q is not installed', async function () {
-        if (env.isMinVscode('1.89.0')) {
-            this.skip()
-        }
+        disableIfVscodeBelow(this, '1.89.0')
```
justinmk3, line 15: consistent language helps discoverability.

```suggestion
        skipIfVscodeBelow(this, '1.89.0')
```

packages/core/src/test/testUtil.ts
```diff
@@ -634,3 +635,9 @@ export function getFetchStubWithResponse(response: Partial<Response>) {
 export function copyEnv(): NodeJS.ProcessEnv {
     return { ...process.env }
 }
+
+export function disableIfVscodeBelow(testContext: Mocha.Context, version: string) {
+    if (isMinVscode(version)) {
```
justinmk3, line 640: This isn't really correct. `isMinVscode` checks if the current vscode is the min version.

The `version` arg is only relevant for when we bump the min version. It's unrelated to the check itself.

justinmk3: `isMinVscode` might be less confusing if we change its argument to be dict-like:

    if (isMinVscode({ throwIfMinVscodeGte = version })

Hweinstock: Good catch, I was assuming this cutoff version could only be below the stable/insider versions, which isn't correct. I went ahead and avoided using `isMinVscode` altogether so that this function can be used more generally.

packages/core/src/test/applicationcomposer/messageHandlers/initMessageHandler.test.ts
```diff
@@ -7,13 +7,11 @@ import assert from 'assert'
 import sinon from 'sinon'
 import { initMessageHandler } from '../../../applicationcomposer/messageHandlers/initMessageHandler'
 import { createTemplate, createWebviewContext } from '../utils'
+import { skipIfVscodeBelow } from '../../testUtil'
 
 describe('initMessageHandler', function () {
-    afterEach(() => {
-        sinon.restore()
-    })
```
justinmk3, line 14: does this risk new tests being added which forget to restore stuff?

packages/core/src/test/applicationcomposer/messageHandlers/generateResourceHandler.test.ts
```diff
@@ -8,13 +8,11 @@ import sinon from 'sinon'
...
 import { Command, MessageType } from '../../../applicationcomposer/types'
-import * as env from '../../../shared/vscode/env'
+import { skipIfVscodeBelow } from '../../testUtil'
 
 describe('generateResourceHandler', function () {
     it('amazon q is not installed', async function () {
-        if (env.isMinVscode('1.89.0')) {
-            this.skip()
-        }
+        skipIfVscodeBelow(this, '1.89.0')
```
justinmk3, line 15: The purpose of isMinVscode is to check if the min vscode is currently running. The purpose of the `throwWhen` arg, is to raise an error so that we know to revisit this when we bump the min vscode version.

packages/core/src/test/applicationcomposer/messageHandlers/initMessageHandler.test.ts
Hweinstock: Since we are mocking within the test block (not in a before or beforeEach) I thought we should restore it within the same test block to keep to contained. But I don't see any problem with restoring outside since there's no danger in an extra `restore` call.

packages/core/src/test/applicationcomposer/messageHandlers/generateResourceHandler.test.ts
Hweinstock: It feels like `isMinVscode` is doing two things, so I factored out the throwing logic into its own helper function: `throwIfMinVscodeGte`. Whether we are on the minimum is irrelevant for this use case, so taking out the throwing logic allows me to reuse the piece of the function I want.

packages/core/src/shared/vscode/env.ts
```diff
@@ -68,13 +68,19 @@ export { extensionVersion }
...
-    const minVscode = getMinVscodeVersion()
-    if (throwWhen && semver.gte(minVscode, throwWhen)) {
-        throw Error(`Min vscode ${minVscode} >= ${throwWhen}. Delete or update the code that called this.`)
+    if (throwWhen) {
+        throwIfMinVscodeGte(throwWhen)
     }
     return vscode.version.startsWith(getMinVscodeVersion())
 }
 
+export function throwIfMinVscodeGte(targetVersion: string): void | never {
```
justinmk3, line 77: `isMinVscode` is doing multiple things, but that can be fixed by making its arg a dict , so that calls look like:

    isMinVscode({ throwWhen: ... })

justinmk3 (approved):

PR #5978, opened 2024-11-12
config(amazonq): adjust timeout for supplemental context

leigaol:
## Problem

The overall supplemental context has a timeout of 50ms, each sub promise has a timeout of 50ms. Problem is these 2 timeout are very tight, leaving no buffer time for node js runtime to do context switch. The poll interval of each sub promise is 10ms, which does not precisely timeout at 50ms. 

## Solution

Add a 10ms buffer time for supplemental context.
Reduce poll duration

---


Will-ShaoHua (approved):

Will-ShaoHua:
@leigaol changelog or make the subject of this PR `config(amazonq)`

leigaol:
> The bot comment explains that this pr is marked `fix` so it should have a changelog. If you're only adjusting something that doesn't have user-facing impact, and thus doesn't need a changelog, use a different type for the PR title. Please do this for future PRs also.

Understood. This is updated.

Issue #3735, opened 2023-08-14
unreliable test: "Auth Shared ini files"


justinmk3:
https://github.com/aws/aws-toolkit-vscode/blob/590c2fa7d7d0641b454df6f86a01e118cb9334e5/packages/core/src/test/credentials/auth.test.ts#L426

```
 1) Auth
       Shared ini files
Error: 1 tests failed.
         does not cache if the credentials file changes:
    at d:\a\aws-toolkit-vscode\aws-toolkit-vscode\src\test\testRunner.ts:90:28

      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected
    at MultiReporters.done (d:\a\aws-toolkit-vscode\aws-toolkit-vscode\node_modules\mocha-multi-reporters\lib\MultiReporters.js:124:16)

    at done (d:\a\aws-toolkit-vscode\aws-toolkit-vscode\node_modules\mocha\lib\mocha.js:1007:16)
  {
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
+   accessKeyId: 'x',
+   secretAccessKey: 'x',
-   accessKeyId: 'y',
-   secretAccessKey: 'y',
    sessionToken: undefined
  }
      + expected - actual

       {
      -  "accessKeyId": "x"
      -  "secretAccessKey": "x"
      +  "accessKeyId": "y"
      +  "secretAccessKey": "y"
         "sessionToken": [undefined]
       }
```

Hweinstock:
Fixed by https://github.com/aws/aws-toolkit-vscode/pull/5979

PR #5979, opened 2024-11-12
test(auth): flaky credentials cache test

Hweinstock:
## Problem
- ran 1000 times in CI, only failed on Windows about 30-40% of the time. 
- appears to be a windows `fs` race condition. We delete the credentials file, then create a new one. If the order of these operations is not preserved, the test can fail. 
- Adding `waitUntil` within the `fs` operations, such that each operation must succeed before returning, does not fix the problem. 
- The credentials file IS updating in time since if we read it right after generating and before assertion, it is the correct data. This then implies that the credentials are in fact caching despite the changes to the credentials file.  
- applying `waitUntil` with the condition being that the contents of the file changed, also did not work since the file content is updating. 
- adding certain log statements causes it not to be flaky. 
- While the contents of the file are updating, the stats of the file are not. This is important, because we use the last modified date to determine if credentials should refresh in https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/auth/providers/sharedCredentialsProviderFactory.ts#L37-L45. 
- **On windows, it appears the `fs.stat` is unreliable, as we can fully rewrite the file and its `mtime` and `ctime` fail to update (at least 10+ seconds, potentially never)**
- We can't rely on `fs.stat` on windows machine to determine if credentials need a refresh. We either need a new mechanism or don't cache on windows. 

## Solution
- Avoid `fs.stat` altogether and refresh by reading the file each time. 
- Added benefit: removes state for tracking modified time, and simplifies the credentials code. 
- Makes a single `fs` call instead of two (two `fs.stat` versus one `fs.read`). 

---

### DIFF
--- a/packages/core/src/auth/providers/sharedCredentialsProviderFactory.ts
+++ b/packages/core/src/auth/providers/sharedCredentialsProviderFactory.ts
@@ -38,19 +38,19 @@ export class SharedCredentialsProviderFactory extends BaseCredentialsProviderFac
         super.resetProviders()
     }
 
-    private async needsRefresh(): Promise<boolean> {
+    public async needsRefresh(): Promise<boolean> {
         const credentialsLastModMillis = await this.getLastModifiedMillis(getCredentialsFilename())
         const configLastModMillis = await this.getLastModifiedMillis(getConfigFilename())
-        // eslint-disable-next-line aws-toolkits/no-console-log
-        console.log('in needs refresh')
-        // eslint-disable-next-line aws-toolkits/no-console-log
-        console.log('\tcredentialsLastModMillis: %O', credentialsLastModMillis)
-        // eslint-disable-next-line aws-toolkits/no-console-log
-        console.log('\tconfigLastModMillis: %O', configLastModMillis)
-        // eslint-disable-next-line aws-toolkits/no-console-log
-        console.log('\tthis.loadedCredentialsModificationMillis: %O', this.loadedCredentialsModificationMillis)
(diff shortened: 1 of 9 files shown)


justinmk3 (approved):

packages/core/src/auth/providers/sharedCredentialsProviderFactory.ts
```diff
@@ -3,47 +3,23 @@
...
-        super.resetProviders()
-    }
-
-    private async needsRefresh(): Promise<boolean> {
-        const credentialsLastModMillis = await this.getLastModifiedMillis(getCredentialsFilename())
-        const configLastModMillis = await this.getLastModifiedMillis(getConfigFilename())
-
-        return (
-            this.loadedCredentialsModificationMillis !== credentialsLastModMillis ||
-            this.loadedConfigModificationMillis !== configLastModMillis
```
justinmk3, line 43: Can't remember if this was discussed already: maybe the delete/modification in the test happened too quickly, so the mtime didn't actually change on the re-created file?

Hweinstock: Thats what I suspect as well. When I added a timeout in-between the two, the test appeared to pass as well.

justinmk3 (approved):
The simplification is good in any case.

justinmk3:
Does this close an open issue?

Hweinstock:
Closes https://github.com/aws/aws-toolkit-vscode/issues/3735

PR #5980, opened 2024-11-12
test: update /dev prepareRepoData unit test to include devfile filtering

willyyhuang:
## Problem

Follow up PR from #5853 

## Solution

Added more unit test coverage

---

### DIFF
--- a/packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
+++ b/packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
@@ -10,55 +10,84 @@ import {
     ContentLengthError,
     maxRepoSizeBytes,
 } from 'aws-core-vscode/amazonqFeatureDev'
-import { assertTelemetry, createTestWorkspace } from 'aws-core-vscode/test'
+import { assertTelemetry, getWorkspaceFolder, TestFolder } from 'aws-core-vscode/test'
 import { fs, AmazonqCreateUpload } from 'aws-core-vscode/shared'
-import { Span } from 'aws-core-vscode/telemetry'
+import { MetricName, Span } from 'aws-core-vscode/telemetry'
 import sinon from 'sinon'
+import { CodeWhispererSettings } from 'aws-core-vscode/codewhisperer'
+
+const testDevfilePrepareRepo = async (expectedRepoSize: number, devfileEnabled: boolean) => {
+    const folder = await TestFolder.create()
+    await folder.write('devfile.yaml', 'test')
+    await folder.write('file.md', 'test content')
+    const workspace = getWorkspaceFolder(folder.path)
+    sinon
+        .stub(CodeWhispererSettings.instance, 'getDevCommandWorkspaceConfigurations')
+        .returns(devfileEnabled ? { [workspace.uri.fsPath]: true } : {})
+
+    await testPrepareRepoData(workspace, expectedRepoSize)
+}
+
+const testPrepareRepoData = async (
+    workspace: vscode.WorkspaceFolder,
+    expectedRepoSize: number,
+    expectedTelemetryMetrics?: Array<{ metricName: MetricName; value: any }>
+) => {
+    const telemetry = new TelemetryHelper()
+    const result = await prepareRepoData([workspace.uri.fsPath], [workspace], telemetry, {
+        record: () => {},
+    } as unknown as Span<AmazonqCreateUpload>)
+
+    assert.strictEqual(Buffer.isBuffer(result.zipFileBuffer), true)
+    // checksum is not the same across different test executions because some unique random folder names are generated
+    assert.strictEqual(result.zipFileChecksum.length, 44)
+    assert.strictEqual(telemetry.repositorySize, expectedRepoSize)
+
+    if (expectedTelemetryMetrics) {
+        expectedTelemetryMetrics.forEach((metric) => {
+            assertTelemetry(metric.metricName, metric.value)
+        })
+    }
+}
 
 describe('file utils', () => {
     describe('prepareRepoData', function () {
+        afterEach(() => {
+            sinon.restore()
+        })
+
         it('returns files in the workspace as a zip', async function () {
-            // these variables are a manual selection of settings for the test in order to test the collectFiles function
-            const fileAmount = 2
-            const fileNamePrefix = 'file'
-            const fileNameSuffix = '.md'
-            const fileContent = 'test content'
+            const folder = await TestFolder.create()
+            await folder.write('file1.md', 'test content')
+            await folder.write('file2.md', 'test content')
+            const workspace = getWorkspaceFolder(folder.path)
 
-            const workspace = await createTestWorkspace(fileAmount, { fileNamePrefix, fileContent, fileNameSuffix })
-
-            const telemetry = new TelemetryHelper()
-            const result = await prepareRepoData([workspace.uri.fsPath], [workspace], telemetry, {
-                record: () => {},
-            } as unknown as Span<AmazonqCreateUpload>)
-            assert.strictEqual(Buffer.isBuffer(result.zipFileBuffer), true)
-            // checksum is not the same across different test executions because some unique random folder names are generated
-            assert.strictEqual(result.zipFileChecksum.length, 44)
-            assert.strictEqual(telemetry.repositorySize, 24)
+            await testPrepareRepoData(workspace, 24)
         })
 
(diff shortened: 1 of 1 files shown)


packages/core/src/test/testUtil.ts
```diff
@@ -185,6 +185,10 @@ export async function createTestWorkspace(
          * optional file name suffix
          */
         fileNameSuffix?: string
+        /**
+         * optional additional files
+         */
+        additionalFiles?: Array<{ fileName: string; fileContent: string }>
```
jpinkney-aws, line 191: Instead of making this does it make sense to use the TestFolder class? Then you would just do something like:

```
const folder = TestFolder.create()
folder.write('devfile.yaml', '')
folder.write('file.md', 'test content')
```

willyyhuang: ok let me do that. hm, I thought the `TestFolder` didn't work outside of `testUtil.ts`, i was wrong, let me update, thanks!

packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
```diff
@@ -56,6 +64,33 @@ describe('file utils', () => {
             assertTelemetry('amazonq_bundleExtensionIgnored', { filenameExt: 'mp4', count: 1 })
         })
 
+        it('prepareRepoData should ignore devfile.yaml when setting is disabled', async function () {
```
jpinkney-aws, line 67:
```suggestion
        it('should ignore devfile.yaml when setting is disabled', async function () {
```

packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
```diff
@@ -56,6 +64,33 @@ describe('file utils', () => {
...
+            const result = await prepareRepoData([workspace.uri.fsPath], [workspace], telemetry, {
+                record: () => {},
+            } as unknown as Span<AmazonqCreateUpload>)
+            assert.strictEqual(Buffer.isBuffer(result.zipFileBuffer), true)
+            // checksum is not the same across different test executions because some unique random folder names are generated
+            assert.strictEqual(result.zipFileChecksum.length, 44)
+            assert.strictEqual(telemetry.repositorySize, 12)
+        })
+
+        it('prepareRepoData should ignore devfile.yaml when setting is enabled', async function () {
```
jpinkney-aws, line 80:
```suggestion
        it('should ignore devfile.yaml when setting is enabled', async function () {
```

jpinkney-aws: should this test be "should not ignore devfile.yaml"?

packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
```diff
@@ -56,6 +64,33 @@ describe('file utils', () => {
...
+            const workspace = await getWorkspaceWithDevFile()
+            await CodeWhispererSettings.instance.updateDevCommandWorkspaceConfigurations(workspace.uri.fsPath, true)
```
jpinkney-aws, line 82: does this change the actual underlying setting for the user?

willyyhuang: yeah in the unit test, let me know if there is a concern, should I clean it up?

packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
willyyhuang: ok

packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
willyyhuang: added settings clean up after the test

chengoramazon (approved):

packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
```diff
@@ -10,12 +10,24 @@ import {
...
+const getWorkspaceWithDevFile = async () => {
+    const folder = await TestFolder.create()
+    await folder.write('devfile.yaml', 'test')
+    await folder.write('file.md', 'test content')
+    return getWorkspaceFolder(folder.path)
+}
 
 describe('file utils', () => {
+    after(async () => {
+        await CodeWhispererSettings.instance.resetDevCommandWorkspaceConfigurations()
```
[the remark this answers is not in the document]
willyyhuang: okay sure, I will mock it

packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
```diff
@@ -56,6 +64,35 @@ describe('file utils', () => {
             assertTelemetry('amazonq_bundleExtensionIgnored', { filenameExt: 'mp4', count: 1 })
         })
 
+        it('should ignore devfile.yaml when setting is disabled', async function () {
+            const workspace = await getWorkspaceWithDevFile()
```
jpinkney-aws, line 68: One last thing, I think we need to mock here as well. Otherwise it will pick up whatever value the user has already set I think

willyyhuang: that's fine, user will have nothing by default, I don't think there is a point of mocking here

jpinkney-aws: But what if a dev runs the test and has the setting already set? Won't it have the potential to fail the test since its running against the users setting?

jpinkney-aws (approved):

jpinkney-aws:
Looks like `should include devfile.yaml when setting is enabled` is failing in ci

willyyhuang:
> Looks like `should include devfile.yaml when setting is enabled` is failing in ci

taking a look

willyyhuang:
this was the error: TypeError: Attempted to wrap getDevCommandWorkspaceConfigurations which is already wrapped

will add a sinon.restore in afterEach hook to reset the mocks

packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
```diff
@@ -56,6 +68,37 @@ describe('file utils', () => {
...
+            } as unknown as Span<AmazonqCreateUpload>)
+            assert.strictEqual(Buffer.isBuffer(result.zipFileBuffer), true)
+            // checksum is not the same across different test executions because some unique random folder names are generated
+            assert.strictEqual(result.zipFileChecksum.length, 44)
+            assert.strictEqual(telemetry.repositorySize, 12)
+            sinon.reset()
+        })
+
+        it('should include devfile.yaml when setting is enabled', async function () {
+            const workspace = await getWorkspaceWithDevFile()
```
jpinkney-aws, line 87: One final thing, does it make sense to do:
```
function testDevfile(settings: {
     workspaceConfiguration: {} | boolean,
     repositorySize: number
}) {
    const workspace = await getWorkspaceWithDevFile()
    sinon.stub(CodeWhispererSettings.instance, 'getDevCommandWorkspaceConfigurations').returns({
        [workspace.uri.fsPath]: settings.workspaceConfiguration,
    })

    const telemetry = new TelemetryHelper()
    const result = await prepareRepoData([workspace.uri.fsPath], [workspace], telemetry, {
        record: () => {},
    } as unknown as Span<AmazonqCreateUpload>)
    assert.strictEqual(Buffer.isBuffer(result.zipFileBuffer), true)
    // checksum is not the same across different test executions because some unique random folder names are generated
    assert.strictEqual(result.zipFileChecksum.length, 44)
    assert.strictEqual(telemetry.repositorySize, settings.repositorySize)
}
```

```
it('should ignore devfile.yaml when setting is disabled', async function() {
     testDevfile({ workspaceConfiguration: {}, repositorySize: 12 })
}

it('should include devfile.yaml when setting is enabled', async function() {
     testDevfile({ workspaceConfiguration: true, repositorySize: 16 })
}
```

duplicate code now fails ci, I think we can probably de-dup some of this stuff

willyyhuang: if you want I can re-write the entire unit test file to avoid duplication completely, there are other areas that can be improved in terms of duplication as well

jpinkney-aws (approved):

PR #5984, opened 2024-11-12
refactor(logger): deeply nested objects fail to show up in logs. 

Hweinstock:
## Problem
Deeply nested objects fail to show up in the logger due to use of `%O` instead of `JSON.stringify`. Some customers use this information for debugging or other purposes. 
Ex. 
``` 
2024-11-12 13:23:39.682 [info] request from tab: tab-1 conversationID: 619df9a2-3ab2-4676-9896-9eb608d80582 request: {
  conversationState: {
    currentMessage: { userInputMessage: [Object] },
    chatTriggerType: 'MANUAL',
    customizationArn: undefined
  }
}
```

## Solution
- Change codewhisper chat request/response log messages (which are very deeply nested) back to `JSON.stringify`. 

---


packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -650,7 +650,10 @@ export class ChatController {
 
         const request = triggerPayloadToChatRequest(triggerPayload)
         const session = this.sessionStorage.getSession(tabID)
-        getLogger().info(`request from tab: ${tabID} conversationID: ${session.sessionIdentifier} request: %O`, request)
+        //eslint-disable-next-line aws-toolkits/no-json-stringify-in-log
+        getLogger().info(
+            `request from tab: ${tabID} conversationID: ${session.sessionIdentifier} request: ${JSON.stringify(request)}`
```
justinmk3, line 655: how deeply nested is it? can we adjust the format() call instead, maybe increase the depth. why does this need to print an unlimited depth?

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -650,7 +651,9 @@ export class ChatController {
 
         const request = triggerPayloadToChatRequest(triggerPayload)
         const session = this.sessionStorage.getSession(tabID)
-        getLogger().info(`request from tab: ${tabID} conversationID: ${session.sessionIdentifier} request: %O`, request)
+        getLogger().info(
+            `request from tab: ${tabID} conversationID: ${session.sessionIdentifier} request: ${util.inspect(request, { depth: 12 })}`
```
justinmk3, line 655: If we don't already have one, let's create our own `inspect()` function which wraps Node's util.inspect and uses our preferred parameters (depth=12 is probably too deep).

On web (non-node), our inspect() can use `partialClone` https://github.com/aws/aws-toolkit-vscode-staging/blob/70f9f1a41d140416912659107d8af91d34d5c464/packages/core/src/shared/utilities/collectionUtils.ts#L307 to limit depth and then pass it to e.g. JSON.stringify

Hweinstock: Sounds like a good idea. I named it `formatObj` and passed the object through `partialClone` either way so that we can provide more options for formatting even when not on web (Ex. key omission / replacement).

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -650,7 +651,9 @@ export class ChatController {
 
         const request = triggerPayloadToChatRequest(triggerPayload)
         const session = this.sessionStorage.getSession(tabID)
-        getLogger().info(`request from tab: ${tabID} conversationID: ${session.sessionIdentifier} request: %O`, request)
+        getLogger().info(
+            `request from tab: ${tabID} conversationID: ${session.sessionIdentifier} request: ${formatObj(request, { depth: 12 })}`
```
justinmk3, line 655: would using the same name as nodejs , improve discoverability ?
```suggestion
            `request from tab: ${tabID} conversationID: ${session.sessionIdentifier} request: ${inspect(request, { depth: 12 })}`
```

packages/core/src/shared/utilities/collectionUtils.ts
```diff
@@ -329,6 +330,33 @@ export function partialClone(obj: any, depth: number = 3, omitKeys: string[] = [
     return clonedObj
 }
 
+type inspectOptions = Partial<{
+    depth: number
+    omitKeys: string[]
+    replacement: any
+    showHidden: boolean
+    color: boolean
```
justinmk3, line 338: Is it clearer if we add `?` to the fields instead of using Partial? I'm not certain about what's most common in the typescript ecosystem, but I would guess that Partial is used to modify an existing type, not one that we are defining inline.

justinmk3:
```ts
import { InspectOptions } from 'util';

type inspectOptions = Partial<InspectOptions>;
```

packages/core/src/shared/utilities/collectionUtils.ts
```diff
@@ -329,6 +330,33 @@ export function partialClone(obj: any, depth: number = 3, omitKeys: string[] = [
...
+    const options = {
+        depth: opt?.depth ?? 3,
+        omitKeys: opt?.omitKeys ?? [],
+        replacement: opt?.replacement,
+        showHidden: opt?.showHidden ?? false,
+        color: opt?.color ?? false,
+    }
+    const objToShow = partialClone(obj, options.depth, options.omitKeys, options.replacement)
+    return isWeb()
+        ? JSON.stringify(objToShow)
```
justinmk3, line 356: are there any stringify() parameters that make sense to pass? `space=2` maybe

packages/core/src/shared/utilities/collectionUtils.ts
```diff
@@ -329,6 +330,33 @@ export function partialClone(obj: any, depth: number = 3, omitKeys: string[] = [
...
+ */
+export function formatObj(obj: any, opt?: inspectOptions): string {
+    const options = {
+        depth: opt?.depth ?? 3,
+        omitKeys: opt?.omitKeys ?? [],
+        replacement: opt?.replacement,
+        showHidden: opt?.showHidden ?? false,
+        color: opt?.color ?? false,
+    }
+    const objToShow = partialClone(obj, options.depth, options.omitKeys, options.replacement)
```
justinmk3, line 354: should we skip the `partialClone` for the non-web case?

Hweinstock: I was worried that some of the options would then only apply in certain cases. Ex. if not on web and we pass omitKeys, it does nothing w/o error or warning. But I guess the caller can just wrap it themselves.

justinmk3: > Ex. if not on web and we pass omitKeys,

inspect() doesn't really need to support omitKeys though. Unless that is a feature of nodejs inspect().

packages/core/src/shared/utilities/collectionUtils.ts
```diff
@@ -329,6 +330,31 @@ export function partialClone(obj: any, depth: number = 3, omitKeys: string[] = [
     return clonedObj
 }
 
+type inspectOptions = Partial<
+    nodeInspectOptions & {
+        omitKeys: string[]
+        replacement: any
```
justinmk3, line 336: inspect doesn't need this options, because callers can easily use `inspect(partialClone(o))` to opt-in if they want (and that's a good general practice: composability instead of "denormalization").

The use of `partialClonse` for the "web" case in our `inspect` wrapper, is simply a way for us to support `depth`.

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -624,7 +625,9 @@ export class ChatController {
                     triggerPayload.relevantTextDocuments = await LspController.instance.query(triggerPayload.message)
                     triggerPayload.relevantTextDocuments.forEach((doc) => {
                         getLogger().info(
-                            `amazonq: Using workspace files ${doc.relativeFilePath}, content(partial): ${doc.text?.substring(0, 200)}`
+                            `amazonq: Using workspace files ${
+                                doc.relativeFilePath
+                            }, content(partial): ${doc.text?.substring(0, 200)}`
```
justinmk3, line 630: is this a formatting-only change? if our linter doesn't require it, this may indicate that your IDE (or a plugin) needs to be adjusted (or just disabled; we already have an autoformat on-commit).

Hweinstock: not really sure where it came from, but believe its resolved.

packages/core/src/shared/utilities/collectionUtils.ts
```diff
@@ -329,6 +330,18 @@ export function partialClone(obj: any, depth: number = 3, omitKeys: string[] = [
     return clonedObj
 }
 
+/**
+ * Wrapper around nodes inspect function that works on web. Defaults to JSON.stringify on web.
+ * @param obj object to show
+ * @param opt options for showing (ex. depth, omitting keys)
+ */
+export function inspect(obj: any, opt?: InspectOptions): string {
```
justinmk3, line 338: if we're only using "depth" then it's probably more confusing than helpful to reference `InspectOptions` here
```suggestion
export function inspect(obj: any, opt?: { depth: number }): string {
```

justinmk3 (approved):
should this be mentioned in the lint rule?

justinmk3 (approved):

PR #5989, opened 2024-11-12
feat(notifications): add misc rule, log, type, and test changes

hayemaxi:
- Add types to some RuleContext fields
- Add some informative log statements for debugging
- Add a test for getRuleContext()
- Add case-insenstive rule checks
- Fix getRuleContext to properly capture authTypes
- Remove a few unneeded lines of code
- Add red icon for emergency notifications

---

### DIFF
--- a/packages/core/src/notifications/panelNode.ts
+++ b/packages/core/src/notifications/panelNode.ts
@@ -6,7 +6,7 @@
 import * as vscode from 'vscode'
 import { ResourceTreeDataProvider, TreeNode } from '../shared/treeview/resourceTreeDataProvider'
 import { Command, Commands } from '../shared/vscode/commands2'
-import { getIcon } from '../shared/icons'
+import { Icon, IconPath, getIcon } from '../shared/icons'
 import { contextKey, setContext } from '../shared/vscode/setContext'
 import { NotificationType, ToolkitNotification, getNotificationTelemetryId } from './types'
 import { ToolkitError } from '../shared/errors'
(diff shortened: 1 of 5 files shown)


packages/core/src/notifications/panelNode.ts
```diff
@@ -78,10 +78,19 @@ export class NotificationsNode implements TreeNode {
 
     public getChildren() {
         const buildNode = (n: ToolkitNotification, type: NotificationType) => {
+            let icon: Icon & { color?: vscode.ThemeColor }
+            if (type === 'startUp') {
+                icon = getIcon('vscode-question') as Icon
+            } else {
+                icon = getIcon('vscode-alert') as Icon
```
justinmk3, line 85: could be a ternary, which would allow `cons` and avoids some code.

packages/core/src/notifications/rules.ts
```diff
@@ -96,17 +98,16 @@ export class RuleEngine {
     }
 
     private evaluateRule(criteria: CriteriaCondition) {
-        const expected = criteria.values
+        const expected = criteria.values.map((v) => v.toLowerCase())
```
justinmk3, line 101: oh wow, is the spec case-insensitive? maybe the spec should be lowercase itself. avoids mistakes in the clients, and reduces variants.

hayemaxi: No, I just wanted to avoid potentially common mistakes such as `mac` instead of `MAC`. But with proper validation this is not necessary.

packages/core/src/notifications/panelNode.ts
hayemaxi: Doing this in ternary form actually causes a runtime error, I suspect because doing the color assignment in the same statement by unwrapping into a new object:
` { ...getIcon(), color: ... }`

PR #5990, opened 2024-11-13
fix(amazonq): users not able to auto trigger Q inline suggestion on systemVerilog files

Will-ShaoHua:
## Problem
within inline suggestion code path, `isLanguageSupported()` relies on `vscode.editor.document.langaugeId` to determine whether the given language is supported Q inline suggestion functionality. However, it's possible that some languages are not recognized by VSCode IDE itself without 3rd party extensions, for example: VSCode doesn't recognize `.sv`, `svh`, `vh` files which is bound to SystemVerilog language if users do not have the 3rd party extension installed and enabled, in this case, `vscode.editor.document.languageId` will simply return `plaintext`. 

This is the root cause why autotrigger doesn't work for systemverilog files, because at the beginning, we don't even register keyStrokeHandler for files with `.sv`, `svh`, `vh` files.

## Solution
if language check return false, we should also check file extension

---

### DIFF
--- a/packages/core/src/codewhisperer/util/runtimeLanguageContext.ts
+++ b/packages/core/src/codewhisperer/util/runtimeLanguageContext.ts
@@ -108,7 +108,10 @@ export class RuntimeLanguageContext {
         })
         this.supportedLanguageExtensionMap = createConstantMap<string, CodewhispererLanguage>({
             c: 'c',
+            h: 'c',
             cpp: 'cpp',
+            cc: 'cpp',
+            'c++': 'cpp',
             cs: 'csharp',
             go: 'go',
             hcl: 'tf',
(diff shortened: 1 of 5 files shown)


packages/core/src/codewhisperer/util/runtimeLanguageContext.ts
```diff
@@ -106,7 +108,10 @@ export class RuntimeLanguageContext {
         })
         this.supportedLanguageExtensionMap = createConstantMap<string, CodewhispererLanguage>({
             c: 'c',
+            h: 'c',
             cpp: 'cpp',
+            cc: 'cpp',
+            'c++': 'cpp',
```
Will-ShaoHua, line 114: to align with JB impl https://github.com/aws/aws-toolkit-jetbrains/blob/main/plugins/amazonq/codewhisperer/jetbrains-community/src/software/aws/toolkits/jetbrains/services/codewhisperer/language/CodeWhispererLanguageManager.kt#L120-L149

andrewyuq (approved):

PR #5991, opened 2024-11-13
telemetry(amazonq):AI code gen % for Q features

leigaol:
## Problem

With the release of many Q features(Inline Suggestion, chat, inline chat, /dev, /test, /doc, /review, /transform), we need to know the % code written by all Q features. This requires calculating and reporting the user written code. The reporting of the code contribution of each Q features was already implemented.

## Solution

Calculate and report the user written code for each language by listening to document change events while Q is not making changes to the editor. 

We add flags to know whether Q is making temporary changes for suggestion rendering or Q suggestion is accepted, by doing so, the document change events are coming from the user. 

We ignore certain document changes when their length of new characters exceeds 50. Previous data driven research has shown that user tend to copy a huge file from one place to another, making the user written code count skyrocketing but that is actually some existing code not written by the user. 

We plan to first collect data from IDEs and let it run in the background in shadow mode before we finish the service side aggregation, fix possible bugs and eventually present the AI code written % to the customers.

Note: The JB PR https://github.com/aws/aws-toolkit-jetbrains/pull/5215. The JB implementation depends on a reliable JB internal message bus to pass information. Using VSC event listener might mess up the boolean state of Q editing or not.

---


justinmk3:
Is this expected for tomorrow's release?

packages/amazonq/test/unit/codewhisperer/tracker/userWrittenCodeTracker.test.ts
```diff
@@ -0,0 +1,194 @@
...
+ */
+
+import assert from 'assert'
+import * as sinon from 'sinon'
+import * as vscode from 'vscode'
+import { UserWrittenCodeTracker, TelemetryHelper, AuthUtil } from 'aws-core-vscode/codewhisperer'
+import { createMockDocument, resetCodeWhispererGlobalVariables } from 'aws-core-vscode/test'
+
+describe('userWrittenCodeTracker', function () {
+    describe('test isActive', function () {
```
justinmk3, line 13: tests never need "test" in their name
```suggestion
    describe('isActive()', function () {
```

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -451,6 +452,7 @@ export const applySecurityFix = Commands.declare(
         }
         let languageId = undefined
         try {
+            UserWrittenCodeTracker.instance.onQStartsMakingEdits()
```
justinmk3, line 455: this is yet another global mutable shared state store. what hapepns if "start" is called but "finish" is not called, because of an error or some unexpected control-flow? Not specifically here, but in the other places spread around.

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
```diff
@@ -0,0 +1,196 @@
...
+import { codeWhispererClient as client } from '../client/codewhisperer'
+import { isAwsError } from '../../shared/errors'
+import { CodewhispererLanguage } from '../../shared'
+
+/**
+ * This singleton class is mainly used for calculating the user written code
+ * for active Amazon Q users.
+ * It reports the user written code per 5 minutes when the user is coding and using Amazon Q features
+ */
+export class UserWrittenCodeTracker {
```
justinmk3, line 21: why is Q constantly adding new global singletones to track more state? why can't the existing one(s) be used?

justinmk3: there are 4 different "trackers" https://github.com/aws/aws-toolkit-vscode/tree/32a4af5f1346d8be710c6b46202a181faca4c8d6/packages/core/src/codewhisperer/tracker

packages/core/src/codewhisperer/commands/basicCommands.ts
leigaol:
```
// if the boolean of qIsMakingEdits was incorrectly set to true
            // due to unhandled edge cases or early terminated code paths
            // reset it back to false after a reasonable period of time
            if (this._qIsMakingEdits) {
                if (performance.now() - this._lastQInvocationTime > UserWrittenCodeTracker.resetQIsEditingTimeoutMs) {
                    getLogger().warn(`Reset Q is editing state to false.`)
                    this._qIsMakingEdits = false
                }
            }
            return
```

This was implemented to flip the boolean if it is incorrectly set to true for longer period of time.

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
leigaol: This is a product requirement. 

The previous `CodeCoverageTracker.ts` needs to be kept as it is to provide data for the current enterprise user dashboard. In the current enterprise user dashboard, we display the % written by Q inline suggestion among all user written code. 

However, as more Q features are contributing code, we are shifting to a newer dashboard V2 that displays code contribution per Q feature. This needs to be running in shadow mode for sometime while the V1 dashboard needs to stay. Therefore, I need to create a different tracker just to provide data point for dashboard V2.

leigaol: The `codeWhispererTracker.ts` was meant to compute the edit % post inline completion acceptance. It is not meant to calculate the % code written. It tracks the position of snippets and how much it was modified.

leigaol: I do not have context of `lineTracker.ts`.

leigaol: Let me know if you want me to completely merge the UserWrittenCodeTracker.ts into CodeCoverageTracker.ts. 

IMO this is doable but it will add more if else and hard-wiring, making it even more error prone. Our expectation is to completely get rid of the old CodeCoverageTracker.ts in a few weeks.

packages/core/src/codewhisperer/commands/basicCommands.ts
leigaol: We do not find alternative other than maintaining a boolean to indicate the source of a document change event. If the document change event API from VS Code does indicate its source, we would have relied on that API. 

To compute % of code written by each Q feature, we need to know what code is written by Q, what code is written by the user. When a code snippet of 20 characters is injected to the editor, we need to know if these 20 is from inline acceptance or Q inline suggestion acceptance or something else. 

Ideally, if the vscode editor API was wrapped into a common layer, then we can force all Q code path to use the same editor operation API, so that we know if it is Q making edits. But this would require changes across all Q features, which requires cross team alignment. (Only exception to this is the inline completion API, the text insertion of inline completion is not done by editor API, but the VS Code inline completion API itself.)

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
Will-ShaoHua: `lineTracker` is for UI purpose, so it's kinda different scope. 
But we do have 2 code percentage tracker(1 pre-existing and the one added here) and 1 modification tracker. IMO it would be better if we could merge % tracker into single one, but that will requires more time and planning.

Will-ShaoHua: Lei do you know is service side ever using userModification?

Will-ShaoHua: asking because given the fact that we hardly can track user's modification on client side, i don't really understand we bother keep this one

Will-ShaoHua: agree with Justin that we better unify and merge these states machine otherwise it's error prone and hard to debug

Will-ShaoHua: is there a way that we can define the message going to the %tracker with a field like `source` which could be `inline suggestion`, `dev` or `inline chat` etc.

justinmk3: > Our expectation is to completely get rid of the old CodeCoverageTracker.ts in a few weeks.

That definitely helps a lot. Is there a tracking issue for that?

justinmk3 (approved):
Approving with the expectation that this followup work with land soon:

> Our expectation is to completely get rid of the old CodeCoverageTracker.ts in a few weeks.

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
```diff
@@ -0,0 +1,196 @@
...
+    private _userWrittenNewCodeLineCount: Map<CodewhispererLanguage, number>
+    private _qIsMakingEdits: boolean
+    private _timer?: NodeJS.Timer
+    private _qUsageCount: number
+    private _lastQInvocationTime: number
+
+    static #instance: UserWrittenCodeTracker
+    static copySnippetThreshold = 50
+    static resetQIsEditingTimeoutMs = 2 * 60 * 1000
+    static defaultCheckPeriodMillis = 5 * 60 * 1000
```
Will-ShaoHua, line 32: could be private?

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
```diff
@@ -0,0 +1,196 @@
...
+    public isActive(): boolean {
+        return TelemetryHelper.instance.isTelemetryEnabled() && AuthUtil.instance.isConnected()
+    }
+
+    // this should be invoked whenever there is a successful Q feature invocation
+    // for all Q features
+    public onQFeatureInvoked() {
+        this._qUsageCount += 1
+        this._lastQInvocationTime = performance.now()
+    }
```
Will-ShaoHua, line 56: i feel we can make it 
```
export type QFeature = 'inlineCompletion | Dev | Transformation | Chat' ...
```

```
public onQFeatureInvoked(src: Feature) {}
```

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
```diff
@@ -0,0 +1,196 @@
...
+        this._timer = undefined
+        this._lastQInvocationTime = 0
+    }
+
+    public static get instance() {
+        return (this.#instance ??= new this())
+    }
+
+    public isActive(): boolean {
+        return TelemetryHelper.instance.isTelemetryEnabled() && AuthUtil.instance.isConnected()
```
Will-ShaoHua, line 48: can we do `return globals.telemetry.telemetryEnabled && AuthUtil.instance.isConnected()` instead

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
```diff
@@ -0,0 +1,196 @@
...
+        this._lastQInvocationTime = performance.now()
+    }
+
+    public onQStartsMakingEdits() {
+        this._qIsMakingEdits = true
+    }
+
+    public onQFinishesEdits() {
+        this._qIsMakingEdits = false
+    }
```
Will-ShaoHua, line 64: thinking of a case where there are multiple Q features are working at the same time, how do we deal with this

for example, while `transformation` is working and users trigger a inline completion request

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
leigaol: >That definitely helps a lot. Is there a tracking issue for that?

I cannot post link here.

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
leigaol: This is Q working flag only controls the moment when Q feature calls the editor edit api, it is almost instant.

leigaol: It is not the time when transformation is pending

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
leigaol: The goal of this CR is to calculate user written code

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
Will-ShaoHua: isee

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    private countNewLines(str: string) {
+        return str.split('\n').length - 1
+    }
+
+    public onTextDocumentChange(e: vscode.TextDocumentChangeEvent) {
+        // do not count code written by Q as user written code
+        if (
+            !runtimeLanguageContext.isLanguageSupported(e.document.languageId) ||
```
Will-ShaoHua, line 164: double checking `runtimeLanguageContext.isLanguageSupported()` here refers to language supported by `inline completion`, and I thought languages supported by Chat is a super set of this?

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
```diff
@@ -0,0 +1,196 @@
...
+            clearTimeout(this._timer)
+            this._timer = undefined
+        }
+    }
+
+    public emitCodeContributions() {
+        const selectedCustomization = getSelectedCustomization()
+
+        for (const [language, charCount] of this._userWrittenNewCodeCharacterCount) {
+            const lineCount = this._userWrittenNewCodeLineCount.get(language) || 0
```
Will-ShaoHua, line 90:
```suggestion
            const lineCount = this.getUserWrittenLines(language)
```

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
```diff
@@ -0,0 +1,196 @@
...
+            if (charCount > 0) {
+                client
+                    .sendTelemetryEvent({
+                        telemetryEvent: {
+                            codeCoverageEvent: {
+                                customizationArn:
+                                    selectedCustomization.arn === '' ? undefined : selectedCustomization.arn,
```
Will-ShaoHua, line 97: FYI we can use `undefinedIfEmpty(s: string)` https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/utilities/textUtilities.ts#L255-L261

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
```diff
@@ -0,0 +1,196 @@
...
+            this.reset()
+            return
+        }
+        const startTime = performance.now()
+        this._timer = setTimeout(() => {
+            try {
+                const currentTime = performance.now()
+                const delay: number = UserWrittenCodeTracker.defaultCheckPeriodMillis
+                const diffTime: number = startTime + delay
+                if (diffTime <= currentTime) {
```
Will-ShaoHua, line 138: what's this for? is this for nodejs@setTimeout 
```
The callback will likely not be invoked in precisely delay milliseconds. Node.js makes no guarantees about the exact timing of when callbacks will fire, nor of their ordering. The callback will be called as close as possible to the time specified.
```
https://nodejs.org/api/timers.html#settimeoutcallback-delay-args

However by doing this, doesn't it mean some metrics will not be sent out

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
Will-ShaoHua: so that this class doesn't have to depend on `TelemetryHelper`

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
leigaol: yes

Will-ShaoHua (approved):

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
leigaol: fixed

packages/core/src/codewhisperer/tracker/userWrittenCodeTracker.ts
leigaol: This implementation of per 5 min time window is exactly the same as previous code percentage tracker.

PR #5993, opened 2024-11-13
fix(cwl): Pass in active credentials when creating CWL client for LiveTail

keeganirby:
## Problem
When creating the CloudWatchClient for a LiveTail session, we are not specifying which credentials to use. This is causing the client to always use the `Default` credential profile, even if a different AWS Credential profile is selected within AWSToolkit.

## Solution
Resolve the active AWS credential from `globals.awsContext`, and supply that to the LiveTailSession constructor. These credentials are then use to construct the CWL client. 

---

### DIFF
--- a/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
+++ b/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
@@ -13,7 +13,7 @@ import {
     LiveTailSessionUpdate,
     StartLiveTailResponseStream,
 } from '@aws-sdk/client-cloudwatch-logs'
-import { getLogger, ToolkitError } from '../../../shared'
+import { getLogger, globals, ToolkitError } from '../../../shared'
 import { uriToKey } from '../cloudWatchLogsUtils'
 
 export async function tailLogGroup(
@@ -25,12 +25,16 @@ export async function tailLogGroup(
     if (!wizardResponse) {
         throw new CancellationError('user')
     }
-
+    const awsCredentials = await globals.awsContext.getCredentials()
+    if (awsCredentials === undefined) {
+        throw new ToolkitError('Failed to start LiveTail session: credentials are undefined.')
+    }
     const liveTailSessionConfig: LiveTailSessionConfiguration = {
         logGroupArn: wizardResponse.regionLogGroupSubmenuResponse.data,
         logStreamFilter: wizardResponse.logStreamFilter,
         logEventFilterPattern: wizardResponse.filterPattern,
         region: wizardResponse.regionLogGroupSubmenuResponse.region,
+        awsCredentials: awsCredentials,
     }
     const session = new LiveTailSession(liveTailSessionConfig)
     if (registry.has(uriToKey(session.uri))) {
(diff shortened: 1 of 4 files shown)


packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -25,12 +25,16 @@ export async function tailLogGroup(
     if (!wizardResponse) {
         throw new CancellationError('user')
     }
-
+    const awsCredentials = await globals.awsContext.getCredentials()
```
jpinkney-aws, line 28: I'm not familiar with how this works under the hood. Will it fall back to default credentials if nobody is logged into the toolkit? Or does it require someone to be logged in? If so, maybe we should re-direct them to login with credentials if they do not exist?

keeganirby: If you have no selected credential, the explorer side bar will prompt you to add/select one. 
Screenshot 2024-11-13 at 10 39 36 AM

If the selected credential is expired, it will say that, and give you an option to re-authenticate.
Screenshot 2024-11-13 at 10 40 49 AM

If your credential is expired/unselected and you try to select TailLogGroup from the command palette, this first picker to select a LogGroup will show `[Error loading Items] Toolkit is not logged in.`
Screenshot 2024-11-13 at 10 39 15 AM

justinmk3:
Note that this is part of the motivation for https://github.com/aws/aws-toolkit-vscode/pull/5940 , where we are starting work on a v3 "client builder" which will handle this stuff automatically. We have a [v2 clientbuilder](https://github.com/aws/aws-toolkit-vscode/blob/16e8a774ccac3c31193b35d264afad40a6e0627f/packages/core/src/shared/awsClientBuilder.ts) already, but are missing one for v3. @Hweinstock

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
justinmk3: why not do this implicitly in your client, instead of plumbing it through?

packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -25,12 +25,16 @@ export async function tailLogGroup(
...
+    if (awsCredentials === undefined) {
+        throw new ToolkitError('Failed to start LiveTail session: credentials are undefined.')
+    }
     const liveTailSessionConfig: LiveTailSessionConfiguration = {
         logGroupArn: wizardResponse.regionLogGroupSubmenuResponse.data,
         logStreamFilter: wizardResponse.logStreamFilter,
         logEventFilterPattern: wizardResponse.filterPattern,
         region: wizardResponse.regionLogGroupSubmenuResponse.region,
+        awsCredentials: awsCredentials,
```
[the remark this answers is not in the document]
justinmk3: > `lobals.awsContext.getCredentials()` is `async`, which cannot be `await`ed in the constructor since constructors cannot also be aysnc.

We usually add a `create()` "factory function" to work around that, and make the constructor private.

PR #6004, opened 2024-11-13
feat(notification): misc logging, types, fixes

hayemaxi:
**Don't squash** 

Commit messages are complete.
Summary:

- Remove unneeded criteria rule
- Add verbose logging for rule checks
- Add types that were missing to parts of the notification schema. Some were just strings but had predefined values
- Add a timeout to the update and reload action so that telemetry can run.
- UX updates including removing the panel wording, add red icons, add activity bar badge.
---

### DIFF
--- a/packages/core/src/notifications/panelNode.ts
+++ b/packages/core/src/notifications/panelNode.ts
@@ -78,24 +78,23 @@ export class NotificationsNode implements TreeNode {
     }
 
     public refresh(): void {
-        if (!this.view) {
-            throw new ToolkitError('NotificationsNode: TreeView accessed without being registered.')
-        }
-
         const totalNotifications = this.notificationCount()
-        if (totalNotifications > 0) {
-            this.view.badge = {
-                tooltip: `${totalNotifications} notification${totalNotifications > 1 ? 's' : ''}`,
-                value: totalNotifications,
+        if (this.view) {
+            if (totalNotifications > 0) {
+                this.view.badge = {
+                    tooltip: `${totalNotifications} notification${totalNotifications > 1 ? 's' : ''}`,
+                    value: totalNotifications,
+                }
+                this.view.title = `${NotificationsNode.title} (${totalNotifications})`
+            } else {
+                this.view.badge = undefined
+                this.view.title = NotificationsNode.title
             }
-            this.view.title = `${NotificationsNode.title} (${totalNotifications})`
-            void setContext(this.showContextStr, true)
         } else {
-            this.view.badge = undefined
-            this.view.title = NotificationsNode.title
-            void setContext(this.showContextStr, false)
+            logger.warn('NotificationsNode was refreshed but the view was not initialized!')
         }
 
+        void setContext(this.showContextStr, totalNotifications > 0)
         this.provider?.refresh()
     }
 
@@ -240,7 +239,7 @@ export class NotificationsNode implements TreeNode {
                                 case 'updateAndReload':
                                     // Give things time to finish executing.
                                     globals.clock.setTimeout(() => {
-                                        void this.updateAndReload(notification.displayIf.extensionId)
+                                        return this.updateAndReload(notification.displayIf.extensionId)
                                     }, 1000)
                                     break
                                 case 'openUrl':
--- a/packages/core/src/notifications/panelNode.ts
+++ b/packages/core/src/notifications/panelNode.ts
@@ -16,6 +16,9 @@ import { registerToolView } from '../awsexplorer/activationShared'
 import { readonlyDocument } from '../shared/utilities/textDocumentUtilities'
 import { openUrl } from '../shared/utilities/vsCodeUtils'
 import { telemetry } from '../shared/telemetry/telemetry'
+import { globals } from '../shared'
+
+const logger = getLogger('notifications')
 
 /**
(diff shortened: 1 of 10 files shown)


packages/core/src/notifications/panelNode.ts
```diff
@@ -8,14 +8,17 @@ import { ResourceTreeDataProvider, TreeNode } from '../shared/treeview/resourceT
...
 import { ToolkitError } from '../shared/errors'
 import { isAmazonQ } from '../shared/extensionUtilities'
 import { getLogger } from '../shared/logger/logger'
 import { registerToolView } from '../awsexplorer/activationShared'
 import { readonlyDocument } from '../shared/utilities/textDocumentUtilities'
 import { openUrl } from '../shared/utilities/vsCodeUtils'
 import { telemetry } from '../shared/telemetry/telemetry'
+import { globals } from '../shared'
+
+const logger = getLogger('notifications')
```
justinmk3, line 21: technically this needs https://github.com/aws/aws-toolkit-vscode/pull/5941 , but I'll get it merged today 🤞

packages/core/src/notifications/panelNode.ts
```diff
@@ -207,7 +210,10 @@ export class NotificationsNode implements TreeNode {
                                     )
                                     break
                                 case 'updateAndReload':
-                                    await this.updateAndReload(notification.displayIf.extensionId)
+                                    // Give things time to finish executing.
+                                    globals.clock.setTimeout(() => {
+                                        void this.updateAndReload(notification.displayIf.extensionId)
+                                    }, 1000)
```
justinmk3, line 243: is there state that can be checked via `waitUntil` ?

packages/core/src/shared/telemetry/telemetryService.ts
```diff
@@ -195,9 +195,11 @@ export class DefaultTelemetryService {
     }
 
     /**
-     * Publish metrics to the Telemetry Service.
+     * Publish metrics to the Telemetry Service. Usually it will automatically flush recent events
+     * on a regular interval. This should not be used unless you are interrupting this interval,
+     * e.g. via a forced window reload.
      */
-    private async flushRecords(): Promise<void> {
+    public async flushRecords(): Promise<void> {
```
justinmk3, line 202: we have a `didReload` concept https://github.com/aws/aws-toolkit-vscode-staging/blob/e02cdb442648177d4e08aab331a9379ad57f9345/packages/core/src/shared/extensionGlobals.ts#L151

which handles window reloads. And our telemetry client has a filesystem cache of telemetry events, which are found on vscode start/reload. 

Given the above, is it still necessary to expose `flushRecords` ?

packages/core/src/test/notifications/controller.test.ts
```diff
@@ -107,6 +108,10 @@ describe('Notifications Controller', function () {
     })
 
     it('can fetch and store startup notifications', async function () {
+        // There seems to be a race condition with having a global spy object that is reset after each
+        // test. Doesn't seem to affect the other ones, for some reason.
+        const setContextSpy = sinon.spy(setContext, 'setContext')
```
justinmk3, line 113: we could modify setContext.ts to store a map so that tests can assert it. That would avoid the need for spy/restore.

packages/core/src/shared/telemetry/telemetryService.ts
hayemaxi: `didReload` applies to SAM only, as it checks for the existence of a key set by it. This also means it only applies to toolkit and not Q.

I see now that telemetry writes to a file on deactivation, previously it looked like it only wrote to an in-memory queue. I will need to verify that `deactivation` definitely runs on this forced window reload as expected.

packages/core/src/notifications/panelNode.ts
hayemaxi: Some arbitrary amount of telemetry, e.g. a ui_click and the run block it is currently in. Maybe we could check this telemetry state. A tradeoff between complexity <> and potential race conditions.

packages/core/src/shared/telemetry/telemetryService.ts
hayemaxi: `deactivation` is not being called for a force reload. Very interesting. So, we still need this call.

packages/core/src/test/notifications/controller.test.ts
hayemaxi: Added

packages/core/src/shared/vscode/setContext.ts
```diff
@@ -36,22 +36,34 @@ export type contextKey =
...
 export async function setContext(key: contextKey, val: any): Promise<void> {
     // eslint-disable-next-line aws-toolkits/no-banned-usages
     await vscode.commands.executeCommand('setContext', key, val)
+    contextMap[key] = val
+}
+
+/**
+ * Returns the value of a context key set via {@link setContext} wrapper for this session.
+ *
+ * Warning: this does not gaurantee the state of the context key in vscode because it may have
```
justinmk3, line 63:
```suggestion
 * Warning: this does not guarantee the state of the context key in vscode because it may have
```

justinmk3 (approved):

PR #6006, opened 2024-11-13
test(notifications): integ test from hosted endpoints

tomcat323:
integ test pulling from hostfile endpoints at: https://idetoolkits-hostedfiles.amazonaws.com/Notifications/integ/VSCode

Test covers:
1. receiving JSON from endpoints
2. process notifications criterias
3. trigger onReceive for newly added notifications

### DIFF
--- a/packages/core/src/testInteg/notifications/notifications.test.ts
+++ b/packages/core/src/testInteg/notifications/notifications.test.ts
@@ -1,0 +1,115 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0 fort
+ */
+
+import { RemoteFetcher } from '../../notifications/controller'
+import { NotificationsController } from '../../notifications/controller'
+import { NotificationsNode } from '../../notifications/panelNode'
+import { RuleEngine } from '../../notifications/rules'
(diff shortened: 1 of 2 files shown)


packages/core/src/testInteg/notifications/notifications.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        const engine = getEngine([VSCODE_EXTENSION_ID.amazonq], true)
+        const onReceiveSpy = sinon.spy(panelNode, 'onReceiveNotifications')
+
+        // Simulate alreadying receving a notification
+        const existingNotificationID = 'startup1'
+        const state = globals.globalState.get<NotificationsState>(controller.storageKey)
+        if (state) {
+            state.newlyReceived.push(existingNotificationID)
+            await globals.globalState.update(controller.storageKey, state)
+        }
```
hayemaxi, line 101: This should usually work, since other tests that run before build the state object if it doesnt exist yet.
But in case the test is running standalone, you should have a default state object. Otherwise, this test will fail.

hayemaxi:
Nice. It would be cool to expand this integ test into different auth states as well. Maybe even try to get to a point where we aren't stubbing the rule engine.

hayemaxi (approved):

PR #6008, opened 2024-11-13
fix(cwl): Set LogEventFilter when constructing LiveTailSession

keeganirby:
## Problem
LogEventFilter is an optional configuration parameter when building a LiveTailSession. Currently, the Constructor is not actually setting this property so it is always undefined. This means that if a user sets a Filter pattern, it will not be applied. 

## Solution
Set the property in LiveTailSession Constructor.
Add missing unit tests to validate that LiveTailSession config values are set properly. Testing via `buildStartLiveTailCommand` because this is what actually gets sent to the CWL backend to start a LiveTailSession.

---

### DIFF
--- a/packages/core/src/test/awsService/cloudWatchLogs/registry/liveTailSession.test.ts
+++ b/packages/core/src/test/awsService/cloudWatchLogs/registry/liveTailSession.test.ts
@@ -10,7 +10,7 @@ describe('LiveTailSession', async function () {
     const testRegion = 'test-region'
     const testFilter = 'test-filter'
     const testAwsCredentials = {} as any as AWS.Credentials
-    it('correctly builds StartLiveTailCommand: no stream Filter, no event filter.', function () {
+    it('builds StartLiveTailCommand: no stream Filter, no event filter.', function () {
         const session = new LiveTailSession({
             logGroupArn: testLogGroupArn,
             region: testRegion,
@@ -28,7 +28,7 @@ describe('LiveTailSession', async function () {
             )
         )
     })
-    it('correctly builds StartLiveTailCommand: with prefix stream Filter', function () {
+    it('builds StartLiveTailCommand: with prefix stream Filter', function () {
         const session = new LiveTailSession({
             logGroupArn: testLogGroupArn,
             logStreamFilter: {
@@ -50,7 +50,7 @@ describe('LiveTailSession', async function () {
             )
         )
     })
-    it('correctly builds StartLiveTailCommand: with specific stream Filter', function () {
+    it('builds StartLiveTailCommand: with specific stream Filter', function () {
         const session = new LiveTailSession({
             logGroupArn: testLogGroupArn,
             logStreamFilter: {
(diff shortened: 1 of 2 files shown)


packages/core/src/test/awsService/cloudWatchLogs/registry/liveTailSession.test.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+import assert from 'assert'
+import { LiveTailSession } from '../../../../awsService/cloudWatchLogs/registry/liveTailSession'
+import { StartLiveTailCommand } from '@aws-sdk/client-cloudwatch-logs'
+describe('LiveTailSession', async function () {
+    const testLogGroupArn = 'arn:aws:test-log-group'
+    const testRegion = 'test-region'
+    const testFilter = 'test-filter'
+    const testAwsCredentials = {} as any as AWS.Credentials
+    it('correctly builds StartLiveTailCommand: no stream Filter, no event filter.', function () {
```
jpinkney-aws, line 13: You can probably drop the correctly from these test names

packages/core/src/test/awsService/cloudWatchLogs/registry/liveTailSession.test.ts
```diff
@@ -0,0 +1,97 @@
...
+                new StartLiveTailCommand({
+                    logGroupIdentifiers: [testLogGroupArn],
+                    logEventFilterPattern: undefined,
+                    logStreamNamePrefixes: undefined,
+                    logStreamNames: [testFilter],
+                })
+            )
+        )
+    })
+    it('builds StartLiveTailCommand: with log event Filter', function () {
```
justinmk3, line 75: most of the tests look nearly identical. the copy-paste detection CI will flag this, and is worth fixing.

packages/core/src/test/awsService/cloudWatchLogs/registry/liveTailSession.test.ts
```diff
@@ -0,0 +1,84 @@
...
+            logGroupArn: testLogGroupArn,
+            logStreamFilter: logStreamFilter,
+            logEventFilterPattern: logEventFilterPattern,
+            region: testRegion,
+            awsCredentials: testAwsCredentials,
+        })
+    }
+
+    function validateStartLiveTailCommand(actual: StartLiveTailCommand, expected: StartLiveTailCommand) {
+        assert.strictEqual(JSON.stringify(actual), JSON.stringify(expected))
```
justinmk3, line 82: instead of json.stringify, any reason not to use  `deepStrictEqual` , which will also produce more-readable test failures? Also, this would eliminate the need for this wrapper function.

```suggestion
        assert.deepStrictEqual(actual, expected)
```

keeganirby: This hadn't been working for me. I think there are some properties in the Command object that are generated outside of the input provided that doesn't match (something in the middlewares attached to the command)

But I just learned we can access the `.input` property of the command to get back just the request parameters. I can assert.deepStrictEqual on this! This is what we are trying to test here anyways, the request parameters.

PR #6010, opened 2024-11-13
fix(appbuilder): Issue related to App-builder placeholder wrong message

jonife:
## Problem
I was working on a CDK project, and after synthesizing my CDK template, I noticed that the app-builder is showing a misleading notification to customers: No IaC templates found in the workspace. This seems incorrect, as I do have IaC templates in the workspace, including both raw and synthesized CDK templates.

Screenshot 2024-11-05 at 9 30 15 AM

While I understand the intention behind the message, it needs to be more precise to avoid confusion. I suggest we roll it back to something more accurate, like No SAM templates found in the workspace, to clearly indicate the specific missing template type. I'm considering this a P0 issue and will raise a PR to address it.

## Solution
Solution is to change all instance where IAC is mentioned to the customer to SAM, we would roll back to IaC once we start support CDK.

Remove redundant Logic.

---

### DIFF
--- a/packages/toolkit/.changes/next-release/Bug Fix-3f75e95c-f940-42e3-b0e2-b6a3b9143a58.json
+++ b/packages/toolkit/.changes/next-release/Bug Fix-3f75e95c-f940-42e3-b0e2-b6a3b9143a58.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Update incorrect messgae in placeholder"
+	"description": "SAM debugging: misleading 'IaC' message/placeholder"
 }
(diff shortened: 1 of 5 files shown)


packages/toolkit/.changes/next-release/Bug Fix-3f75e95c-f940-42e3-b0e2-b6a3b9143a58.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Update incorrect messgae in placeholder"
```
justinmk3, line 3:
```suggestion
	"description": "SAM debugging: misleading 'IaC' message/placeholder"
```

justinmk3:
I improved the changelog. Please review https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#changelog for future PRs.

PR #6012, opened 2024-11-14
fix(amazonq): file opening behavior on accepted changes

kelvin-klchu:
## Problem
- when you accept an individual code change, it will bring up the file window
- when you update the accepted file, both left and right of the window reflects the changes you made immediately
- it should only show the changes made on the left, while the right stays the same as the original suggested code change

## Solution
- updated the way vscode.diff is called when opening up the file such that it is the actual file on the left, and the suggested code change stays on the right

---


jpinkney-aws (approved):

jpinkney-aws:
I'll wait for approval from your team before merging

ctidd (approved):

PR #6014, opened 2024-11-14
refactor(lambda): decompose common methods for SAM sync deploy build

vicheey:
## Problem

The `sync.ts` are huge consisting of many helper functions that being reused for deploy and build.

## Solution
- split shared methods into separate files for easier testing and maintenance
- split sync.test.ts file into smaller file
- consolidate `paramsSourcePrompter` for both sync and deploy to avoid duplicate code
- update PrompterTester helper class
- remain paramsSource prompter title 

---


packages/toolkit/package.json
```diff
@@ -4066,277 +4066,270 @@
                     "fontCharacter": "\\f1b4"
                 }
             },
-            "aws-amazonq-transform-landing-page-icon": {
```
valerena, line 4069: What are the changes on this file? Do you need to rebase maybe or something?

packages/core/src/shared/ui/sam/samTemplate.ts
```diff
@@ -0,0 +1,61 @@
+/*!
```
justinmk3, line 1: Moving wizards/prompters into `shared/ui/sam/` is a good thought, thanks 👍

justinmk3:
> Problem: The `sync.ts` are huge consisting of many helper functions that being reused for deploy and build.

This PR adds 800 lines though. Why?

packages/core/src/test/shared/sam/sync.test.ts
```diff
@@ -51,35 +42,31 @@ import { createMultiPick, DataQuickPickItem } from '../../../shared/ui/pickerPro
...
 import { intoCollection } from '../../../shared/utilities/collectionUtils'
-import { DefaultEcrClient, EcrRepository } from '../../../shared/clients/ecrClient'
 import { SamConfig, Environment, parseConfig } from '../../../shared/sam/config'
 import { RegionProvider } from '../../../shared/regions/regionProvider'
 import { Region } from '../../../shared/regions/endpoints'
 import { RegionNode } from '../../../awsexplorer/regionNode'
 import { getProjectRootUri } from '../../../shared/sam/utils'
 import { AppNode } from '../../../awsService/appBuilder/explorer/nodes/appNode'
 import * as Cfn from '../../../shared/cloudformation/cloudformation'
+//import { beforeEach } from 'mocha'
```
valerena, line 57: Don't add this commented line

packages/core/src/shared/ui/sam/bucket.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+import { DefaultS3Client } from '../../clients/s3Client'
+import { samDeployUrl, samSyncUrl } from '../../constants'
+import { createCommonButtons } from '../buttons'
+import { createQuickPick, DataQuickPickItem } from '../pickerPrompter'
+import type { SyncParams } from '../../sam/sync'
+import * as nls from 'vscode-nls'
+import { getRecentResponse } from '../../sam/utils'
+
+const localize = nls.loadMessageBundle()
```
Vandita2020, line 13: I would suggest giving a more descriptive naming to this file, the current naming gives impression that the file create a bucket resource. Maybe name it to `bucket-prompter`

vicheey:
> @justinmk3 : This PR adds 800 lines though. Why?
I think it coming from the fact that I added more unit tests to make code coverage pass. This includes test for Build and some function in Utils. 
Screenshot 2024-11-14 at 11 16 15

packages/core/src/shared/ui/sam/ecr.ts
```diff
@@ -0,0 +1,47 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
Vandita2020, line 4: Same as previous comment for naming, maybe change file name to `ecr-prompter`

packages/toolkit/package.json
vicheey: The `package.json` inside toolkit directory is auto generated. For some reason this is get included. I'll remove this.

packages/core/src/shared/ui/sam/bucket.ts
vicheey: I'll update the name.

packages/core/src/shared/ui/sam/ecr.ts
vicheey: Rename file to have **Prompter** suffix

packages/core/src/shared/ui/sam/bucketPrompter.ts
```diff
@@ -0,0 +1,69 @@
...
+import { createCommonButtons } from '../buttons'
+import { createQuickPick, DataQuickPickItem } from '../pickerPrompter'
+import type { SyncParams } from '../../sam/sync'
+import * as nls from 'vscode-nls'
+import { getRecentResponse } from '../../sam/utils'
+
+const localize = nls.loadMessageBundle()
+export const prefixNewBucketName = (name: string) => `newbucket:${name}`
+
+export enum BucketSource {
```
Vandita2020, line 16: Can we also put header comments to describe the purpose of the function to the reader. I know we haven't done this at most places but given we are refactoring things and placing them in right places to understand the code better, I think it would be a good practice to write a small description.

vicheey: Would you like this done for all function?

packages/core/src/test/shared/sam/utils.test.ts
```diff
@@ -148,4 +163,113 @@ describe('SAM utils', async function () {
...
+        const value1 = 'myStackName'
+        const value2 = 'myBucketName'
+
+        it('1. getRecentResponse should return undefined when mementoRootKey does not exist', async () => {
+            assert(!getRecentResponse(nonExistingMementoRootKey, identifier, key1))
+        })
+
+        it('2. updateRecentResponse should return the undefined when mementoRootKey does not exist', async () => {
+            try {
+                await updateRecentResponse(mementoRootKey, identifier, key1, value1)
```
roger-zhangg, line 245: Should consider restore memento after this test in case of side effects

packages/core/src/test/shared/sam/utils.test.ts
```diff
@@ -148,4 +163,113 @@ describe('SAM utils', async function () {
...
+            assert.strictEqual(result1, value1)
+            assert.strictEqual(result2, value2)
+        })
+
+        it('5. updateRecentResponse should log when failure to update value', async () => {
+            sinon.stub(globals.context.workspaceState, 'update').rejects(new Error('Error updating value'))
+            try {
+                await updateRecentResponse(mementoRootKey, identifier, key2, value2)
+            } catch (err) {
+                assert.fail('Should have swallowed exception')
```
roger-zhangg, line 269: It seems you want this to succeed, but assert fail is added in try catch. Does it mean either this success or failed, test should pass?

packages/core/src/test/shared/sam/utils.test.ts
```diff
@@ -148,4 +163,113 @@ describe('SAM utils', async function () {
...
+
+        it('1. getRecentResponse should return undefined when mementoRootKey does not exist', async () => {
+            assert(!getRecentResponse(nonExistingMementoRootKey, identifier, key1))
+        })
+
+        it('2. updateRecentResponse should return the undefined when mementoRootKey does not exist', async () => {
+            try {
+                await updateRecentResponse(mementoRootKey, identifier, key1, value1)
+            } catch (err) {
+                assert.fail('Should have succeed')
```
roger-zhangg, line 247: why we have a assert.fail here, from description it seems you want the test to fail if error happens.

packages/core/src/shared/ui/sam/stackPrompter.ts
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    return createQuickPick(items, {
+        title: 'Select a CloudFormation Stack',
+        placeholder: 'Select a stack (or enter a name to create one)',
+        filterBoxInputSettings: {
+            label: 'Create a New Stack',
+            transform: (v) => v,
+        },
+        buttons: createCommonButtons(samSyncUrl, consoleUrl),
```
roger-zhangg, line 40: If this is intended to be used by both sync/deploy, why this only add syncUrl in the help button?

packages/core/src/shared/sam/build.ts
```diff
@@ -128,7 +129,7 @@ export class BuildWizard extends Wizard<BuildParams> {
         this.arg = arg
         if (this.arg === undefined) {
             // "Build" command was invoked on the command palette.
-            this.form.template.bindPrompter(() => createTemplatePrompter(this.registry))
+            this.form.template.bindPrompter(() => createTemplatePrompter(this.registry, syncMementoRootKey))
```
roger-zhangg, line 132: Why build is using `syncMementoRootKey`

packages/core/src/shared/ui/sam/bucketPrompter.ts
Vandita2020: For the ones that are already in this PR and that too not for all small functions. Maybe describe the purpose where main functionality is happening.

roger-zhangg (commented):
Left few comments, thanks for implementing all these tests, they are super helpful

packages/core/src/test/shared/sam/build.test.ts
```diff
@@ -210,9 +222,325 @@ describe('createParamsSourcePrompter', () => {
...
+            })
+            await testFolder.write('samconfig.toml', samconfigCompleteData)
+        })
+
+        afterEach(() => {
+            sandbox.restore()
+        })
+
+        it('should abort when customer cancel build wizard', async () => {
+            getTestWindow().onDidShowQuickPick((picker) => picker.dispose())
```
roger-zhangg, line 502:
```suggestion
            getTestWindow().onDidShowQuickPick((picker) => {
                await picker.untilReady()
                picker.dispose())
            }
```

packages/core/src/test/shared/sam/utils.test.ts
vicheey: I think the test instance get reset globally. But there is no harm in doing explicit clean up. Will do.

packages/core/src/test/shared/sam/utils.test.ts
```diff
@@ -148,4 +163,113 @@ describe('SAM utils', async function () {
...
+        it('4. updateRecentResponse should return the update only specified key', async () => {
+            await updateRecentResponse(mementoRootKey, identifier, key2, value2)
+            const result1 = getRecentResponse(mementoRootKey, identifier, key1)
+            const result2 = getRecentResponse(mementoRootKey, identifier, key2)
+            assert.strictEqual(result1, value1)
+            assert.strictEqual(result2, value2)
+        })
+
+        it('5. updateRecentResponse should log when failure to update value', async () => {
+            sinon.stub(globals.context.workspaceState, 'update').rejects(new Error('Error updating value'))
```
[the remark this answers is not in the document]
vicheey: Not much this different since we only use it in one case, I skip the sandbox creation.

packages/core/src/test/shared/ui/sam/ecrPrompter.test.ts
```diff
@@ -0,0 +1,126 @@
...
+        assert.strictEqual(prompter.quickPick.title, 'Select an ECR Repository')
+        assert.strictEqual(prompter.quickPick.placeholder, 'Select a repository (or enter a name to create one)')
+        assert.strictEqual(prompter.quickPick.items.length, 3)
+        assert.deepStrictEqual(prompter.quickPick.items, expectedItems)
+    })
+
+    it('should include no items found message if no repos exist', async () => {
+        const listAllStacksStub = sandbox.stub(ecrClient, 'listAllRepositories').returns(intoCollection([]))
+        sandbox.stub(SamUtilsModule, 'getRecentResponse').returns(undefined)
+        const createCommonButtonsStub = sandbox.stub(ButtonsModule, 'createCommonButtons')
```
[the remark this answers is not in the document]
vicheey: True. Yet, it is easier to see what is there than to see what is not. I see test as a way to communicate expectation. So I prefer to explicitly say what the condition here is to help reader understand better.

packages/core/src/test/shared/ui/sam/ecrPrompter.test.ts
```diff
@@ -0,0 +1,126 @@
...
+        )
+        assert.ok(listAllStacksStub.calledOnce)
+        assert.strictEqual(prompter.quickPick.title, 'Select an ECR Repository')
+        assert.strictEqual(prompter.quickPick.placeholder, 'Select a repository (or enter a name to create one)')
+        assert.strictEqual(prompter.quickPick.items.length, 1)
+        assert.deepStrictEqual(
+            prompter.quickPick.items[0].label,
+            'No ECR repositories in region "us-east-1". Enter a name to create a new one.'
+        )
+        assert.deepStrictEqual(prompter.quickPick.items[0].data, undefined)
```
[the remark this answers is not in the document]
vicheey: This is original test that I just copy and paste.

packages/core/src/shared/sam/build.ts
vicheey: I am not sure why but this has been like this.

packages/core/src/shared/ui/sam/bucketPrompter.ts
vicheey: I added doc string for exported functions moved to ui/sam files

packages/core/src/shared/ui/sam/stackPrompter.ts
vicheey: I think this refactoring really bring up many questions about the current logic. 

In master branch, the `deploy.ts` import `createStackPrompter()` directly from `sync.ts`. I believe this was intended, or at least okay to share (!?!?!?).

Screenshot 2024-11-14 at 15 24 38

roger-zhangg: Then we should fix this bug along with this chagne

vicheey: Need to update the deployUrl.

packages/core/src/test/shared/sam/sync.test.ts
```diff
@@ -2075,6 +1689,99 @@ describe('ensureBucket', () => {
     })
 })
 
+describe('ensureRepo', () => {
```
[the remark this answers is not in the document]
vicheey: This is the existing pattern for the current file as well as build and deploy. If we want to implement it here, we should do that for all file for consistency. Yet, this would expand the PR even more.

roger-zhangg (approved):
LGTM, action item to follow: use correct link in helper button in deploy template prompter. Currently is linking to sync help url.

packages/core/src/test/shared/ui/sam/stackPrompter.test.ts
```diff
@@ -0,0 +1,131 @@
...
+        const listAllStacksStub = sandbox.stub(cfnClient, 'listAllStacks').returns(intoCollection(stackSummaries))
+        sandbox.stub(SamUtilsModule, 'getRecentResponse').returns(undefined)
+        const createCommonButtonsStub = sandbox.stub(ButtonsModule, 'createCommonButtons')
+        sandbox
+            .stub(AwsConsoleModule, 'getAwsConsoleUrl')
+            .returns(vscode.Uri.parse(`https://us-east-1.console.aws.amazon.com/cloudformation/home?region=us-east-1`))
+
+        // Act
+        const prompter = createStackPrompter(cfnClient, mementoRootKey, samSyncUrl)
+        await new Promise((f) => setTimeout(f, 50))
```
hayemaxi, line 84: Use `waitUntil`. Applicable throughout

packages/core/src/test/shared/wizards/prompterTester.ts
```diff
@@ -96,15 +103,27 @@ export class PrompterTester {
...
+        }
+
+        try {
+            const timeoutPromise = new Promise<never>((_, reject) => {
+                setTimeout(() => {
+                    reject(new Error(`Handler for "${input.title}" exceeded ${this.handlerTimout}ms timeout`))
+                }, this.handlerTimout)
+            })
+
+            await Promise.race([handler(input), timeoutPromise])
```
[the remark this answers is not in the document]
vicheey: Will update to `await waitUntil(async () => await handler(input), { timeout: this.handlerTimout })`

packages/core/src/test/shared/ui/sam/stackPrompter.test.ts
vicheey: Understood. After checking with the test author, a more appropriate option is to use `await sleep()`.

packages/core/src/test/shared/wizards/prompterTester.ts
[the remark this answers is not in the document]
vicheey: Will update to ` await waitUntil(handler(input), { timeout: this.handlerTimout, interval: 50 })`

packages/core/src/test/shared/ui/sam/stackPrompter.test.ts
vicheey: Will update to using `await sleep()`.

packages/core/src/test/shared/wizards/prompterTester.ts
```diff
@@ -6,20 +6,28 @@
...
+    testWindow?: TestWindow
+    handlerTimeout?: number
+}
 
 export class PrompterTester {
     private quickPickHandlers: Map<string, (input: TestQuickPick) => void> = new Map()
     private inputBoxHanlder: Map<string, (input: TestInputBox) => void> = new Map()
     private testWindow: TestWindow
     private callLog = Array<string>()
+    private handlerTimout: number = 3000 // Default timeout to 3 seconds
```
hayemaxi, line 21: nit
```suggestion
    private handlerTimeout: number = 3000 // Default timeout to 3 seconds
```

packages/core/src/test/shared/ui/sam/stackPrompter.test.ts
```diff
@@ -0,0 +1,132 @@
...
+        const listAllStacksStub = sandbox.stub(cfnClient, 'listAllStacks').returns(intoCollection(stackSummaries))
+        sandbox.stub(SamUtilsModule, 'getRecentResponse').returns(undefined)
+        const createCommonButtonsStub = sandbox.stub(ButtonsModule, 'createCommonButtons')
+        sandbox
+            .stub(AwsConsoleModule, 'getAwsConsoleUrl')
+            .returns(vscode.Uri.parse(`https://us-east-1.console.aws.amazon.com/cloudformation/home?region=us-east-1`))
+
+        // Act
+        const prompter = createStackPrompter(cfnClient, mementoRootKey, samSyncUrl)
+        await sleep(50)
```
hayemaxi, line 85: Is this to give the prompter time to be clicked asynchronously?
If so, I think this might be susceptible to flakiness if the machine can't do it in 50ms.

Why cant we do something like
```suggestion
        await waitUntil(async () => return createCommonButtonStub.calledOnce, { timeout: 3000, interval: 100 })
```

vicheey: That's a possibility. I have attempted to use PromperTester class to test the actual prompt. The issue seems to be the stackPrompter or bucketPrompter need some moments to render item from list api. Calling `prompt()` right after calling  create prompter method will result in empty option items. Even using `await picker.untilReady()` does not solve this problem at the moment. 

However, the issue only exists when te call prompter directly. The in sync/deploy wizard test, the prompters seem to work correctly; because of this, I am relax about this test cases.

vicheey: Give suggested solution a try and did not work.

hayemaxi (approved):

PR #5994, opened 2024-11-13
config(amazonq): inline project context ab

Will-ShaoHua:
## Problem
update inline suggestion project context ab test config

#### CONTROL
- use bm25 opentabs context 

#### TREATMENT_1
- use repomap + bm25 open tabs context 

#### TREATMENT_2
- use bm25 global project context 

## Solution

---

License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6019, opened 2024-11-14
test(amazonq): patch #5994 causing CI test failing

Will-ShaoHua:
## Problem
#5994 is causing CI test failure
misuse of `sinon.alwaysReturned()`

## Solution
should use `sinon.returns()`

---

### DIFF
--- a/packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts
+++ b/packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts
@@ -37,7 +37,6 @@ describe('crossFileContextUtil', function () {
 
     describe('fetchSupplementalContextForSrc', function () {
         beforeEach(async function () {
-            sinon.restore()
             tempFolder = (await createTestWorkspaceFolder()).uri.fsPath
         })
 
@@ -312,7 +311,6 @@ describe('crossFileContextUtil', function () {
         })
 
         beforeEach(async function () {
-            sinon.restore()
             tempFolder = (await createTestWorkspaceFolder()).uri.fsPath
         })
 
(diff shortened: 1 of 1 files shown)


packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -37,11 +37,16 @@ describe('crossFileContextUtil', function () {
 
     describe('fetchSupplementalContextForSrc', function () {
         beforeEach(async function () {
+            sinon.restore()
```
Hweinstock, line 40: is it necessary to restore on `beforeEach` hook? If we already restore on `afterEach` there shouldn't be any new stubs created between the `afterEach` and the `beforeEach` that need to be restored.

Will-ShaoHua: totally agree

Hweinstock (approved):

justinmk3:
Still some other failing tests:

```
 1) supplementalContextUtil
       fetchSupplementalContext
         openTabsContext
           opentabContext should include chunks if non empty:

      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert_1.default.ok(actual?.supplementalContextItems.length === 3)

      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (test/unit/codewhisperer/util/supplemetalContextUtil.test.ts:43:24)

  2) supplementalContextUtil
       fetchSupplementalContext
         openTabsContext
           opentabsContext should filter out empty chunks:

      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert_1.default.ok(actual?.supplementalContextItems.length === 0)

      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (test/unit/codewhisperer/util/supplemetalContextUtil.test.ts:57:24)
```

PR #6023, opened 2024-11-14
test(cwl): Add LiveTailSession unit tests for start and stopLiveTail

keeganirby:
## Problem
LiveTailSession's startLiveTail and stopLiveTail methods are uncovered

## Solution
Adds tests to each, ensuring that:
StartLiveTail:
* calls CWL Client with AbortSignal

StopLiveTail:
* AbortSignal is fired
* CWL client is destroyed

Also moves test helper functions `buildLiveTailCommandOutput` and `getTestResponseStream` to the LiveTailSession tests because that class is what is responsible for calling and returning the CWL backend APIs.

---

### DIFF
--- a/packages/amazonq/test/unit/amazonqGumby/resources/files/diff.json
+++ b/packages/amazonq/test/unit/amazonqGumby/resources/files/diff.json
@@ -1,0 +1,9 @@
+{
+    "content": [
+        {
+            "name": "Added file",
+            "fileName": "resources/files/addedFile.diff",
+            "isSuccessful": true
+        }
+    ]
+}
--- a/packages/amazonq/test/unit/amazonqGumby/transformationResultsHandler.test.ts
+++ b/packages/amazonq/test/unit/amazonqGumby/transformationResultsHandler.test.ts
@@ -4,13 +4,19 @@
  */
 import assert from 'assert'
 import sinon from 'sinon'
-import os from 'os'
 import { DiffModel, AddedChangeNode, ModifiedChangeNode } from 'aws-core-vscode/codewhisperer/node'
+import { DescriptionContent } from 'aws-core-vscode/codewhisperer'
 import path from 'path'
 import { getTestResourceFilePath } from './amazonQGumbyUtil'
 import { fs } from 'aws-core-vscode/shared'
+import { createTestWorkspace } from 'aws-core-vscode/test'
 
 describe('DiffModel', function () {
+    let parsedTestDescriptions: DescriptionContent
+    beforeEach(async () => {
+        parsedTestDescriptions = JSON.parse(await fs.readFileText(getTestResourceFilePath('resources/files/diff.json')))
+    })
+
     afterEach(() => {
         sinon.restore()
     })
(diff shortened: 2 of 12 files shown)


packages/core/src/test/awsService/cloudWatchLogs/registry/liveTailSession.test.ts
```diff
@@ -65,6 +83,36 @@ describe('LiveTailSession', async function () {
         )
     })
 
+    it('startLiveTail sends command to CWL client with AbortSignal', async function () {
```
justinmk3, line 86: Looks like a lot of code to test very little. The test description sounds like it's testing that a parameter is passed somewhere.

packages/core/src/test/awsService/cloudWatchLogs/registry/liveTailSession.test.ts
```diff
@@ -65,6 +83,36 @@ describe('LiveTailSession', async function () {
...
+        const sendCommandMock = sandbox.stub(CloudWatchLogsClient.prototype, 'send').callsFake(function () {
+            return buildLiveTailCommandOutput(testLogGroupArn)
+        })
+        const session = buildLiveTailSession({ type: 'all' }, undefined)
+        const stream = await session.startLiveTailSession()
+
+        assert.strictEqual(sendCommandMock.calledOnce, true)
+        assert.strictEqual(
+            sendCommandMock.calledWith(
```
justinmk3, line 95: instead of mocking many things just to test that parameters are sent, can you stub fake responses and let the code actually work on the response. that tests the full codepath, instead of just tiny parts of it (and requiring lots of test code)

packages/core/src/test/awsService/cloudWatchLogs/registry/liveTailSession.test.ts
```diff
@@ -65,6 +83,36 @@ describe('LiveTailSession', async function () {
...
+                })
+            ),
+            true
+        )
+        assert.strictEqual(session.isAborted, false)
+        assert.deepEqual(stream, buildLiveTailCommandOutput(testLogGroupArn).responseStream)
+    })
+
+    it('stopLiveTail fires AbortSignal and destroys CWL client', function () {
+        const destroyClientSpy = sandbox.spy(CloudWatchLogsClient.prototype, 'destroy')
```
justinmk3, line 108: why does this need a spy? test the actual state. the client can expose some (readonly) state if needed.

do not use mocks and spies except as a last resort. it looks like it's the first resort in this tests, instead of the last resort.

keeganirby: I don't think there's a way to directly test the state of the underlying AWS SDK CWL client to check if its destroyed. Calling destroy on the Client doesn't stop it from being used from a future `send` call, it just destroys the underlying socket connections. There is no return value from `destroy` nor a `isDestroyed` (or something similar) to return its state.

Using a Spy here allows us to assert that a `.destroy()` call is made in a `stopLiveTailSession` call.

packages/core/src/test/awsService/cloudWatchLogs/registry/liveTailSession.test.ts
keeganirby: Sorry, I don't think I understand the requested change to be made here. 
The function in test here is setting some metadata on the LiveTailSession, calling StartLiveTail API, and returning the stream. I'm only mocking one thing and that is stubbing a fake response from CWL API, and testing it's being returned.
The AbortSignal object is not exposed to upstream callers, so validating that it is passed into the sendCommand is only doable via testing this method directly. Given that firing the abortSignal is critical to closing a LiveTail session I feel like having this validation is worth having the test code.

PR #6024, opened 2024-11-15
feat(stepfunctions): Prevent launching Workflow Studio with invalid JSON + save metric refactoring

witness-me:
## Problem
1. Currently, the Workflow Studio (WFS) webview opens even when the JSON definition is invalid, which causes errors in the webview since WFS requires an initial, valid JSON definition.
2. The current metric for saving the file lacks a clear name and is not optimally defined for tracking save/sync events.
3. There is an issue with the undo functionality (CTRL+Z) inside WFS where the webview definition does not reflect changes in the local file. This occurs because VSCode's built-in undo command interferes with the WFS undo.
4. The production CDN link is missing in the code (since it was not set up before)

## Solution
1. Detect invalid JSON files and prevent WFS from opening if JSON is invalid. In such cases, switch to the default editor and display a warning message.
2. Refactor and rename the save/sync metric for clarity and modify its parameters for improved tracking.
3. Suppress VSCode's built-in undo command when WFS webview is focused to prevent conflicts, allowing WFS to manage undo operations directly.
4. Production CDN link is added
---


witness-me:
[Demo recording](https://maxis-file-service-prod-pdx.pdx.proxy.amazon.com/issues/61d36c26-0013-4c84-99e6-75a053b9be7b/attachments/3eb6a2c9bb3f9dcbf05139028e995ee35705b65d033c008bf6136eaa9564a65b_16cfdaf9-3e39-4502-b67d-2d64d72640fc)

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -17,6 +17,17 @@
             "allowedValues": ["Create", "Update"],
             "description": "SSM Publish Document operation type"
         },
+        {
+            "name": "stepFunctionsSyncType",
+            "type": "string",
+            "allowedValues": ["MANUAL_SAVE_WFS", "MANUAL_SAVE_VSCODE", "AUTO_SYNC_WFS"],
```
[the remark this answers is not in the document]
witness-me: I used it this way because the previous version of the metric that I was using (`saveType` introduced by Threat Composer integration) had the same screaming snake casing.

I can update the metric to camel case to keep this consistent with most of the types

packages/core/package.json
```diff
@@ -3653,6 +3653,12 @@
                 "key": "ctrl+shift+v",
                 "mac": "cmd+shift+v",
                 "when": "editorTextFocus && editorLangId == asl || editorTextFocus && editorLangId == asl-yaml"
+            },
+            {
+                "command": "noop",
```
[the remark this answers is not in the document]
jpinkney-aws: Makes sense. Is there any way to intercept the save command on the custom editor using something like
```
webView.webview.onDidRecieveMessage(e => {
// whatever save does here
})
```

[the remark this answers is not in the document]
witness-me: I belive you mean undo in your comment, rather than save.

The command is executed inside WFS and only affects WFS internal definition, the webview does not receive any message for it.
With the suppression logic I introduced, it should work like this: 
1. User triggers undo in WFS (from their keyboard or the button in the UI)
2. WFS performs its internal undo and updates its internal definition to the previous one
3. Updated WFS internal definition is synced with the local file
4. At the same time, VSCode built-in undo is suppressed for Ctrl+Z to not interfere with the above behaviour

I did not find any better suppression options since the custom editor (including webview) has all the standard VSCode keybindings working (such as Ctrl+Z) unless we override them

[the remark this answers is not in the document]
jpinkney-aws: Ah interesting, I think that's fine then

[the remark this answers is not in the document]
jpinkney-aws: Let me check with my team to see if they have any ideas how to do this without a noop command

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -527,20 +538,19 @@
...
+            "name": "stepfunctions_syncFile",
+            "description": "Triggered when Workflow Studio auto-syncs to the local file or when unsaved local changes are saved from Workflow Studio or VSCode.",
             "metadata": [
                 {
                     "type": "id",
                     "required": true
                 },
                 {
-                    "type": "saveType",
+                    "type": "stepFunctionsSyncType",
```
justinmk3, line 549: why are you renaming this field? fields should not be over-specific like this

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -17,6 +17,17 @@
             "allowedValues": ["Create", "Update"],
             "description": "SSM Publish Document operation type"
         },
+        {
+            "name": "stepFunctionsSyncType",
```
justinmk3, line 21: all of these stepfunctions metrics need to be upstreamed to https://github.com/aws/aws-toolkit-common

they have been in this override file for a long time

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -17,6 +17,17 @@
...
+            "type": "string",
+            "allowedValues": ["manualSaveWFS", "manualSaveVSCode", "autoSyncWFS"],
+            "description": "The action type that triggered the save or sync event between SFN Workflow Studio and the local source file."
+        },
+        {
+            "name": "isInvalidJson",
+            "type": "boolean",
+            "description": "Indicates whether the message contains an invalid JSON definition."
```
justinmk3, line 29: please read https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md

this looks like a failure mode. throw an error from your code, and then `telemetry.xx.run()` will correctly set `result=Failed` and `reason=<error code>`. Don't add a highly-specific boolean to signal a failure.

packages/core/src/stepFunctions/utils.ts
```diff
@@ -223,6 +223,32 @@ export async function isDocumentValid(text: string, textDocument?: vscode.TextDo
...
+        // An empty file or whitespace-only text is considered valid JSON for our use case
+        if (!jsonFileContent) {
+            return false
+        }
+        try {
+            JSON.parse(jsonFileContent)
+            return false
+        } catch {
+            return true
+        }
```
justinmk3, line 247: this is an extremely verbose way to do a very simple, common thing.

packages/core/src/stepFunctions/workflowStudio/workflowStudioEditorProvider.ts
```diff
@@ -115,6 +117,21 @@ export class WorkflowStudioEditorProvider implements vscode.CustomTextEditorProv
...
+            if (isInvalidJsonFile(document)) {
+                await vscode.commands.executeCommand('vscode.openWith', document.uri, 'default')
+                webviewPanel.dispose()
+                void vscode.window.showWarningMessage(i18n('AWS.stepFunctions.workflowStudio.actions.invalidJson'))
+
+                throw ToolkitError.chain(
+                    'Invalid JSON file',
+                    'The Workflow Studio editor was not opened because the JSON in the file is invalid',
+                    {
+                        code: 'InvalidJSONContent',
```
justinmk3, line 130: `run()` will automatically set `reason` to this `code`. Thus there is no need for the `isInvalidJson` field.

packages/core/src/shared/telemetry/vscodeTelemetry.json
witness-me: It have already been raised in one of the previous pull requests, we are still developing the feature and we will most probably need to add more metrics, which is more handy to do in a single package.

This approach also aligns with what is written in [telemetry](https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md) readme:
> When your feature is released, the "development" metrics you defined in vscodeTelemetry.json should be upstreamed to [aws-toolkit-common](https://github.com/aws/aws-toolkit-common/blob/main/telemetry/definitions/commonDefinitions.json).

We have a SIM in our project to transfer metrics to that package closer to launch. Can we keep them in the current package for now?

packages/core/src/shared/telemetry/vscodeTelemetry.json
witness-me: This is not a failure, but rather an expected case for which we are syncing webview content with the file as usual. This variable is just to track this case in telemetry, not to handle this differently

packages/core/src/shared/telemetry/vscodeTelemetry.json
witness-me: `saveType` is a field name introduced by threat composer, which [only allows 2 values](https://github.com/aws/aws-toolkit-common/blob/0bec1b3f7af36bdbb24f38757d7c42b5dcb6928d/telemetry/definitions/commonDefinitions.json#L1687-L1694): `"MANUAL_SAVE" | "AUTO_SAVE"`.

For our use case, we want to add `auto-sync` option, as well track where the action originated (either in WFS or VSCode). If we have the latter as a separate metric, it will not be clear which specific action originated where (since we have both save and sync actions).

Do you belive there is a better way to achieve that?

packages/core/src/stepFunctions/workflowStudio/workflowStudioEditorProvider.ts
witness-me: This code does not have anything to do with `isInvalidJson` metric, which indicates that the JSON code inside the custom webview is invalid. Meanwhile, this code is responsible for detecting invalid JSON code in local file to prevent opening webview integration in the first place.

packages/core/src/shared/telemetry/vscodeTelemetry.json
justinmk3: > `saveType` is a field name introduced by threat composer, which [only allows 2 values](https://github.com/aws/aws-toolkit-common/blob/0bec1b3f7af36bdbb24f38757d7c42b5dcb6928d/telemetry/definitions/commonDefinitions.json#L1687-L1694): `"MANUAL_SAVE" | "AUTO_SAVE"`.

add more values. enums are just lists that guide string values.

packages/core/src/shared/telemetry/vscodeTelemetry.json
justinmk3: > Can we keep them in the current package for now?

What is the ECD?

packages/core/src/shared/telemetry/vscodeTelemetry.json
justinmk3: How is invalid json not a failure?

packages/core/src/shared/telemetry/vscodeTelemetry.json
witness-me: I can add an action such as `AUTO_SYNC` in that list, I just wanted to make sure that can track the metric with enough granularity in that case. We have 3 scenarios:
- User saves from VSCode
- User saves from custom editor by clicking on the button in the UI
- User makes the change in custom editor and this change is auto-synced with local file
We want to catch a usage for each of these scenarios, I'm not sure if we can achieve that with separate metrics.

The way you are suggesting we would need to post 2 metrics for save type such as 
```
span.record({
  saveType: 'MANUAL_SAVE',
  source: 'VSCODE',
})
```
Can you confirm if we will be able to query a metric by multiple fields when constructing a dashboard? I.e. we'll want to detect detect save events originating from custom editor vs VSCode (and exclude `AUTO_SYNC` event)

packages/core/src/shared/telemetry/vscodeTelemetry.json
witness-me: we are planning to launch around end of February

packages/core/src/shared/telemetry/vscodeTelemetry.json
witness-me: Since we are launching our custom editor by default, we want to sync the webview content with local file as much as we can. If the user opens code editor inside our integration and types something making JSON invalid, we still want to sync that invalid JSON to the local file as a string to prevent user from losing any data (e.g. if they close VSCode)

packages/core/src/shared/telemetry/vscodeTelemetry.json
justinmk3: > The way you are suggesting we would need to post 2 metrics for save type such as
> 
> ```
> span.record({
>   saveType: 'MANUAL_SAVE',
>   source: 'VSCODE',
> })
> ```

By "2 metrics" do you mean "2 fields in 1 metric" (which aligns with the code example)? That's normal and idiomatic, and no problem for dashboards.

PR #6033, opened 2024-11-15
feat(beta): force debug/verbose logging for Beta users

justinmk3:
## Problem
Beta and bug-bash users often need logs.

## Solution
Log debug/verbose messages at "info" log-level. They will appear as "info" logs in the Output channel, even though they were logged at "debug" or "verbose" level.

## Related

https://github.com/aws/aws-toolkit-vscode-staging/pull/1961 confirms CI behavior in a beta branch.

---

### DIFF
--- a/packages/core/src/dev/beta.ts
+++ b/packages/core/src/dev/beta.ts
@@ -26,7 +26,7 @@ import { getRelativeDate } from '../shared/datetime'
 const localize = nls.loadMessageBundle()
 const logger = getLogger('dev/beta')
 
-const downloadIntervalMs = 1000 * 60 * 60 * 24 // A day in milliseconds
+const downloadIntervalMs = 1000 * 60 * 60 * 3 // 3 hours (8 times/day).
 
 interface BetaToolkit {
     readonly needUpdate: boolean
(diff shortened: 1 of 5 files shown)


packages/core/src/shared/logger/toolkitLogger.ts
```diff
@@ -120,6 +123,9 @@ export class ToolkitLogger extends BaseLogger implements vscode.Disposable {
     }
 
     override sendToLog(level: LogLevel, message: string | Error, ...meta: any[]): number {
+        // XXX: force debug/verbose logs for Beta users.
+        level = this.isBeta ? 'info' : level
```
jpinkney-aws, line 127: I'm not familiar with this code so I can't tell if i'm missing something, but isn't this saying if its beta then use info? I thought we wanted verbose/debug if its beta users

justinmk3: It forces the "sinks" (log providers such as Output channel, file, etc.) to log at "info" level. This forces the vscode Output channel to show logs that it would otherwise hide.

ivikash (approved):

packages/core/src/dev/beta.ts
```diff
@@ -19,14 +19,14 @@ import { telemetry } from '../shared/telemetry/telemetry'
...
-import * as config from './config'
+import * as devConfig from './config'
 import { isReleaseVersion } from '../shared/vscode/env'
 import { getRelativeDate } from '../shared/datetime'
 
 const localize = nls.loadMessageBundle()
 const logger = getLogger('dev/beta')
 
-const downloadIntervalMs = 1000 * 60 * 60 * 24 // A day in milliseconds
+const downloadIntervalMs = 1000 * 60 * 60 * 3 // 3 hours (8 times/day).
```
justinmk3, line 29: increase frequency

PR #6034, opened 2024-11-15
telemetry(amazonq): add telemetry for code generated and code accepted

kelvin-klchu:
## Problem
- we need to define a way to measure code generated and code accepted from `/dev` feature

## Solution
- we compute the char and line difference between the two files
- all generated codes are immediately reported in telemetry
- within the same session, de-duplication is done to avoid reporting the exact same changes twice across different code gen iterations
- currently, only accepted changes with new lines added will be reported

## Special Note
`charsAdded` and `charRemoved` may be slightly different from the count you manually get from the IDE.
The reason is because new line character `\n` is not counted here in the computeDiff method.

---


packages/core/src/amazonqFeatureDev/session/session.ts
```diff
@@ -209,6 +212,7 @@ export class Session {
     }
 
     public async insertNewFiles(newFilePaths: NewFileInfo[]) {
```
neilk-aws, line 214: If a customer applies files from the file tree, but never clicks "Accept remaining", do we count anything in this approach?

neilk-aws (approved):

packages/core/src/amazonqFeatureDev/session/session.ts
kelvin-klchu: Yes. This method is called by both the accept changes button and also by the individual file accept action button.

packages/core/package.json
```diff
@@ -488,6 +488,7 @@
         "bytes": "^3.1.2",
         "cross-fetch": "^4.0.0",
         "cross-spawn": "^7.0.3",
+        "diff": "^7.0.0",
```
jpinkney-aws, line 491: we technically have this as a dev dependency right now btw. If we want to promoted it to a regular dependency we should remove it from there

kelvin-klchu:
rebase changes and resolved conflicts

packages/core/package.json
kelvin-klchu: I have moved this to dependencies, also keep the version as `^5.1.0` to avoid breaking existing functionalities.

jpinkney-aws (approved):

PR #6037, opened 2024-11-15
feat(ec2): dry run connection script to surface errors earlier.

Hweinstock:
## Problem
Follow up to https://github.com/aws/aws-toolkit-vscode/pull/6018#discussion_r1844284580

## Solution
- Run `ssh` within the same env as it will be run on real connection. 
- Log any resulting errors, and inform user where the process failed. 
- Also part of this PR is moving some functions to `remoteSession.ts` that are general enough to be there. 

Error msg:
Screenshot 2024-11-15 at 5 53 51 PM

---


packages/core/src/shared/remoteSession.ts
```diff
@@ -251,3 +251,23 @@ export async function getDeniedSsmActions(client: IamClient, roleArn: string): P
 
     return deniedActions
 }
+
+/**
+ * Creates a new {@link ChildProcess} class bound to a specific dev environment. All instances of this
```
justinmk3, line 256: Good idea extracting this 👍 

```suggestion
 * Creates a new {@link ChildProcess} class bound to a specific remote environment. All instances of this
```

packages/core/src/shared/extensions/ssh.ts
```diff
@@ -119,6 +123,26 @@ export class RemoteSshSettings extends Settings.define('remote.SSH', remoteSshTy
...
+    try {
+        const env = { SESSION_ID: session.SessionId, STREAM_URL: session.StreamUrl, TOKEN: session.TokenValue }
+        await new ProcessClass(sshPath, ['-T', `${user}@${hostname}`, 'echo connected && exit']).run({
+            spawnOptions: {
+                env,
+            },
+        })
+    } catch (error) {
+        getLogger().error('SSH connection test failed: %O', error)
+        throw new SSHError('SSH connection test failed', { cause: error as Error })
```
justinmk3, line 142: since we're throwing here and chaining stuff correctly, we probably don't need the log call . higher level paths , or the global error handler, should log it?

packages/core/src/shared/extensions/ssh.ts
```diff
@@ -119,6 +123,26 @@ export class RemoteSshSettings extends Settings.define('remote.SSH', remoteSshTy
...
+export async function testSshConnection(
+    ProcessClass: typeof ChildProcess,
+    hostname: string,
+    sshPath: string,
+    user: string,
+    session: SSM.StartSessionResponse
+): Promise<void> {
+    try {
+        const env = { SESSION_ID: session.SessionId, STREAM_URL: session.StreamUrl, TOKEN: session.TokenValue }
+        await new ProcessClass(sshPath, ['-T', `${user}@${hostname}`, 'echo connected && exit']).run({
```
justinmk3, line 135: not a blocker: for connecting to a Windows machine would we need a different test command? or does this command work on powershell?

justinmk3 (approved):
Awesome

packages/core/src/shared/extensions/ssh.ts
justinmk3, line 135: can we capture the output of this process invocation and include it in the logs? 

we may even want to use a similar approach as `collectSamErrors` to collect errors, and include them in the user-facing message. https://github.com/aws/aws-toolkit-vscode/blob/3d46b7ba38fb86f6ce0ec082fe648b9167f12f1d/packages/core/src/shared/sam/cli/samCliInvokerUtils.ts#L75

packages/core/src/shared/extensions/ssh.ts
Hweinstock: Wanted to make no details were dropped from the logs, but with an adaption of `collectionSamErrors` (idea mentioned below) this shouldn't be a problem.

packages/core/src/shared/extensions/ssh.ts
Hweinstock: Since we directly invoke the sshPath I believe this should still work. However, once I add some tests for this function I will hopefully have a better idea.

Hweinstock: I believe the output is already in the logs because we bind a logger to the process: 
https://github.com/aws/aws-toolkit-vscode/blob/823a0e3f0d0cdb416bbd8f27ecd2fe7fc1d73288/packages/core/src/awsService/ec2/model.ts#L248-L252
Do you think its worth filtering the logs? I agree that `collectSamErrors` method could definitely be generalized and potentially applied here as a filter, but unsure what to filter for.

packages/core/src/shared/extensions/ssh.ts
```diff
@@ -8,15 +8,19 @@ import * as path from 'path'
...
 import { Settings } from '../settings'
 import { VSCODE_EXTENSION_ID } from '../extensions'
+import { SSM } from 'aws-sdk'
+import { ToolkitError } from '../errors'
 
 const localize = nls.loadMessageBundle()
 
 export const sshAgentSocketVariable = 'SSH_AUTH_SOCK'
 
+export class SSHError extends ToolkitError {}
```
justinmk3, line 22:
```suggestion
export class SshError extends ToolkitError {}
```

justinmk3 (approved):

packages/core/src/shared/extensions/ssh.ts
```diff
@@ -119,6 +123,30 @@ export class RemoteSshSettings extends Settings.define('remote.SSH', remoteSshTy
...
+export async function testSshConnection(
+    ProcessClass: typeof ChildProcess,
+    hostname: string,
+    sshPath: string,
+    user: string,
+    session: SSM.StartSessionResponse
+): Promise<ChildProcessResult | never> {
+    try {
+        const env = { SESSION_ID: session.SessionId, STREAM_URL: session.StreamUrl, TOKEN: session.TokenValue }
+        const result = await new ProcessClass(sshPath, [
```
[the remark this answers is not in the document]
Hweinstock: Because we are running it in a bound process, this becomes a little tricky. We can pass a process as an argument, but that opens the door for processes that don't behave how `tryRun` is expecting. For example, in this case our bound process rejects on errors. `tryRun` does not expect this. `tryRun` would need to be refactored to manually handle all `spawnOptions` that could alter behavior and these might be specific to the use case.

justinmk3 (approved):
🚀

PR #6039, opened 2024-11-18
fix(amazonq): fixed bug with high CPU use event loop for security scan findings

mr-lee:
## Problem

Opening the security scan issues webview led to an infinite loop on the `update` call, consuming significant amounts of customer CPU.

## Solution

Updated the webview to not auto-update, and made it so that each "view" command opened a new window, even if already present.

This isn't fully ideal, but it at least immediately fixes the CPU consumption issue.  Any setup to watch/poll for updates from the server would be either expensive or latent.  This will be the most clean, but may lead to a different UX than expected.

---


jpinkney-aws (approved):

jpinkney-aws:
looks good from my side, still needs an additional review from the security scan team

ashishrp-aws (approved):

PR #6040, opened 2024-11-18
fix(notifications): update types, field names, cross-window syncing

hayemaxi:
- Make types more granular.
- Align names - fix onRecieve typo, openTxt -> openTextDocument
- Sync from global state on reload to detect activity in other windows. This makes it so that if users dismiss in one window, it will dismiss in all windows. It will also prevent a bug where users may get dismissed notifications again if they have multiple windows.

---

### DIFF
--- a/packages/core/src/notifications/activation.ts
+++ b/packages/core/src/notifications/activation.ts
@@ -11,9 +11,10 @@ import { RuleEngine, getRuleContext } from './rules'
 import globals from '../shared/extensionGlobals'
 import { AuthState } from './types'
 import { getLogger } from '../shared/logger/logger'
+import { oneMinute } from '../shared/datetime'
 
 /** Time in MS to poll for emergency notifications */
-const emergencyPollTime = 1000 * 10 * 60
+const emergencyPollTime = oneMinute * 10
 
 /**
  * Activate the in-IDE notifications module and begin receiving notifications.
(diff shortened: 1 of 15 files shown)


packages/core/src/notifications/activation.ts
```diff
@@ -13,7 +13,7 @@ import { AuthState } from './types'
 import { getLogger } from '../shared/logger/logger'
 
 /** Time in MS to poll for emergency notifications */
-const emergencyPollTime = 1000 * 10 * 60
+const emergencyPollTime = 1000 * 60 * 10
```
nkomonen-amazon, line 16: I added in some constants for time in `datetime.ts`, we should look to start using these

packages/core/src/notifications/types.ts
```diff
@@ -119,12 +154,14 @@ export const NotificationsStateConstructor: TypeConstructor<NotificationsState>
 
 export type NotificationType = keyof Omit<NotificationsState, 'dismissed' | 'newlyReceived'>
 
+type AuthType = 'credentials' | 'builderId' | 'identityCenter' | 'unknown'
```
nkomonen-amazon, line 157: nit: Its probably best to centralize these common auth values. We have similar-ish types in `connection.ts`. Maybe this line can go in to that file for now, and we can make a note to refactor later

nkomonen-amazon (approved):

packages/core/src/notifications/types.ts
hayemaxi: I have a todo here: https://github.com/aws/aws-toolkit-vscode/blob/135c3e5a73bda9c51da47eb5004b6a6d7741ed29/packages/core/src/notifications/rules.ts#L170

but it should probably be moved somewhere more common as you mentioned

PR #6042, opened 2024-11-18
telemetry(lambda): Send meaningful telemetry error codes for SAM CLI invocation failures

mbfreder:
## Problem
With only Unknown errors, It's hard to differentiate errors or make possible improvements

## Solution
- Parse SAM CLI output
- Add telemetry code based on the output error

---

### DIFF
--- a/packages/core/src/shared/sam/build.ts
+++ b/packages/core/src/shared/sam/build.ts
@@ -236,6 +236,7 @@ export async function runBuild(arg?: TreeNode): Promise<SamBuildResult> {
     } catch (error) {
         throw ToolkitError.chain(error, 'Failed to build SAM template', {
             details: { ...resolveBuildArgConflict(buildFlags) },
+            code: error instanceof ToolkitError ? error.code : undefined,
         })
     }
 }
--- a/packages/core/src/shared/sam/build.ts
+++ b/packages/core/src/shared/sam/build.ts
@@ -19,7 +19,7 @@ import globals from '../extensionGlobals'
 import { TreeNode } from '../treeview/resourceTreeDataProvider'
 import { telemetry } from '../telemetry/telemetry'
 import { getSpawnEnv } from '../env/resolveEnv'
-import { getProjectRoot, getSamCliPathAndVersion, isDotnetRuntime } from './utils'
+import { getErrorCode, getProjectRoot, getSamCliPathAndVersion, isDotnetRuntime } from './utils'
 import { getConfigFileUri, validateSamBuildConfig } from './config'
 import { runInTerminal } from './processTerminal'
 
@@ -236,7 +236,7 @@ export async function runBuild(arg?: TreeNode): Promise<SamBuildResult> {
     } catch (error) {
         throw ToolkitError.chain(error, 'Failed to build SAM template', {
             details: { ...resolveBuildArgConflict(buildFlags) },
-            code: error instanceof ToolkitError ? error.code : undefined,
+            code: getErrorCode(error),
         })
     }
 }
--- a/packages/core/src/shared/sam/processTerminal.ts
+++ b/packages/core/src/shared/sam/processTerminal.ts
@@ -12,18 +12,29 @@ import { CancellationError } from '../utilities/timeoutUtils'
 import { getLogger } from '../logger'
 import { removeAnsi } from '../utilities/textUtilities'
 import { isAutomation } from '../vscode/env'
+import { getSamCliErrorMessage, SamCliErrorTypes } from './utils'
 
 let oldTerminal: ProcessTerminal | undefined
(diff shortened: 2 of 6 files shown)


vicheey:
Please update the PR title accordingly.

packages/core/src/shared/sam/processTerminal.ts
```diff
@@ -12,18 +12,29 @@ import { CancellationError } from '../utilities/timeoutUtils'
...
-            const message = `sam ${cmd} exited with a non-zero exit code: ${result.exitCode}`
+            const defaultMessage = `sam ${cmd} exited with a non-zero exit code: ${result.exitCode}`
             if (result.stderr.includes('is up to date')) {
-                throw ToolkitError.chain(result.error, message, {
+                throw ToolkitError.chain(result.error, defaultMessage, {
                     code: 'NoUpdateExitCode',
                 })
             }
-            throw ToolkitError.chain(result.error, message, {
+            for (const errorType in SamCliErrorTypes) {
```
vicheey, line 27: Nit: we may consider making logic of converting this error type into a helper function in utils file.

vicheey (changes requested):

packages/core/src/shared/sam/build.ts
```diff
@@ -236,6 +236,7 @@ export async function runBuild(arg?: TreeNode): Promise<SamBuildResult> {
     } catch (error) {
         throw ToolkitError.chain(error, 'Failed to build SAM template', {
             details: { ...resolveBuildArgConflict(buildFlags) },
+            code: error instanceof ToolkitError ? error.code : undefined,
```
valerena, line 239: ideally you would remove this logic into a function to do 
```
     code: getErrorCode(error)
```
instead of having to repeat the `instanceof` inline in 3 different places.

mbfreder: Thanks. Updated

packages/core/src/test/shared/sam/utils.test.ts
vicheey, line 1: Is it possible to write tests by calling sam cli command directly to get error from SAM. Since we are calling parsing error message from SAM, we need test cases that detects (fails) when SAM change/update that error message.

packages/core/src/shared/sam/utils.ts
```diff
@@ -83,3 +84,37 @@ export async function getSamCliPathAndVersion() {
...
+}
+
+export enum SamCliErrorTypes {
+    DockerUnreachable = 'Docker is unreachable.',
+    ResolveS3AndS3Set = 'Cannot use both --resolve-s3 and --s3-bucket parameters in non-guided deployments.',
+    DeployStackStatusMissing = 'Was not able to find a stack with the name:',
+    DeployStackOutPutFailed = 'Failed to get outputs from stack',
+    DeployBucketRequired = 'Templates with a size greater than 51,200 bytes must be deployed via an S3 Bucket.',
+    NoUpdateExitCode = 'is up to date',
+    ChangeSetEmpty = 'No changes to deploy. Stack is up to date',
```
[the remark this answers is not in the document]
mbfreder: Right. Updated.

packages/core/src/shared/sam/utils.ts
```diff
@@ -83,3 +84,37 @@ export async function getSamCliPathAndVersion() {
...
+    for (const errorType in SamCliErrorTypes) {
+        if (errorMessage.includes(SamCliErrorTypes[errorType as keyof typeof SamCliErrorTypes])) {
+            throw ToolkitError.chain(result.error, errorMessage, {
+                code: errorType,
+            })
+        }
+    }
+}
+
+export enum SamCliErrorTypes {
```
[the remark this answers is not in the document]
mbfreder: Thanks for the proposition, although I would prefer to keep the current implementation as the error class is not relevant for telemetry. The ```code``` is what gets shown on the dashboard along with the ```message```

packages/core/src/test/shared/sam/utils.test.ts
mbfreder: I don't think so because this would make API calls to AWS (cfn, lambda), which we don't want to do.

vicheey (approved):

jpinkney-aws (approved):

packages/core/src/shared/sam/utils.ts
```diff
@@ -83,3 +84,36 @@ export async function getSamCliPathAndVersion() {
 
     return { path: samCliPath, parsedVersion }
 }
+
+export function getSamCliErrorMessage(stderr: string): string {
+    // Split the stderr string by newline, filter out empty lines, and get the last line
+    const lines = stderr
+        .trim()
+        .split('\n')
```
roger-zhangg, line 92: Old school question, does this work on windows?

PR #6047, opened 2024-11-18
feat(cwl): Emit telemetry when starting and stopping liveTail sessions

keeganirby:
## Problem
Cloudwatch wants to monitor usage metrics of the LiveTail integration

## Solution
When starting a session emit telemetry for: 
* Session already started (this case happens when session is already running, and customer sends a new command that matches the already running session)
* Has LogEventFilter
* LogStream filter type
* Source of the command (command palette, explorer)

When closing a session: 
* Session duration
* source of cancellation (ex: CodeLens, ClosingEditors)

---

### DIFF
--- a/packages/core/src/shared/telemetry/vscodeTelemetry.json
+++ b/packages/core/src/shared/telemetry/vscodeTelemetry.json
@@ -350,6 +350,21 @@
             "name": "amazonqMessageDisplayedMs",
             "type": "int",
             "description": "Duration between the partner teams code receiving the message and when the message was finally displayed in ms"
+        },
+        {
+            "name": "livetailSessionAlreadyStarted",
+            "type": "boolean",
+            "description": "Session already open"
+        },
+        {
+            "name": "livetailHasLogEventFilterPattern",
+            "type": "boolean",
+            "description": "If LogEvent filter pattern is applied"
+        },
+        {
+            "name": "livetailLogStreamFilterType",
+            "type": "string",
+            "description": "Type of LogStream filter applied to session"
         }
     ],
     "metrics": [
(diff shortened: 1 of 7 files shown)


packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1230,6 +1245,42 @@
                 }
             ],
             "passive": true
+        },
+        {
+            "name": "cwlLiveTail_Start",
```
jpinkney-aws, line 1250: i'm still not sure why we can't just upstream these to commons

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -350,6 +350,21 @@
             "name": "amazonqMessageDisplayedMs",
             "type": "int",
             "description": "Duration between the partner teams code receiving the message and when the message was finally displayed in ms"
+        },
+        {
+            "name": "livetailSessionAlreadyStarted",
```
jpinkney-aws, line 355: instead of prefixing these with the product name like we have in the past, you can probably drop the product name as the prefix so that we can make these more general

packages/core/src/shared/telemetry/vscodeTelemetry.json
keeganirby: I was just following the documentation here: https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md. I did this in VSCode so my local builds would generate the telemetry objects and I could bundle adding the metrics as well as using them in the same PR. If I need to update commons instead I can send a change for that.

keeganirby: Opened PR to upstream to Commons. https://github.com/aws/aws-toolkit-common/pull/916

jpinkney-aws (approved):

PR #6052, opened 2024-11-19
fix(notifications): extension activation hangs if fetch fails

hayemaxi:
## Problem
 If notifications fail to fetch, it will retry multiple times with 30 second waits. Activation is awaiting on this though, so it may take forever to finish.

## Solution
don't await on notification activation

Also, update logging statements.

---

### DIFF
--- a/packages/core/src/notifications/activation.ts
+++ b/packages/core/src/notifications/activation.ts
@@ -13,6 +13,8 @@ import { AuthState } from './types'
 import { getLogger } from '../shared/logger/logger'
 import { oneMinute } from '../shared/datetime'
 
+const logger = getLogger('notifications')
+
 /** Time in MS to poll for emergency notifications */
 const emergencyPollTime = oneMinute * 10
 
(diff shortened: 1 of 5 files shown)


jpinkney-aws (approved):

packages/core/src/notifications/activation.ts
```diff
@@ -33,19 +35,23 @@ export async function activate(
...
-    }, emergencyPollTime)
+        globals.clock.setInterval(async () => {
+            const ruleContext = await getRuleContext(context, await authStateFn())
+            await controller.pollForEmergencies(new RuleEngine(ruleContext))
+        }, emergencyPollTime)
 
-    getLogger('notifications').debug('Activated in-IDE notifications polling module')
+        logger.debug('Activated in-IDE notifications polling module')
+    } catch (err) {
+        logger.error('Failed to activate in-IDE notifications module.')
```
justinmk3, line 55: .... %O', err)`

or `err.message`

hayemaxi:
Will follow up with your comments @justinmk3

PR #6053, opened 2024-11-20
telemetry(amazonq): fix userTriggerDecision not being sent if service returns empty

Will-ShaoHua:
## Problem
there are 2 scenarios where we consider "Empty"
1. empty list of suggestion : `[]`
2. non empty list of suggestion with empty content: `['', '']`

Currently, only (2) will have plugins send out `userTriggerDecision` event whereas (1) doesn't.

## Solution

---


packages/core/src/codewhisperer/util/telemetryHelper.ts
```diff
@@ -261,21 +295,30 @@ export class TelemetryHelper {
             .reduce((a, b) => a + b, 0)
 
         const aggregated: CodewhispererUserTriggerDecision = {
```
Will-ShaoHua, line 297: re-order alphabetically

zixlin7 (approved):

PR #6055, opened 2024-11-20
fix(amazonq): do not overwrite failure reason, use lowercase lang name

dhasani23:
## Problem

Our backend is expecting the language name (`java` or `sql`) to be lowercase, not uppercase. Also, the failure reason returned by one of our APIs was being overwritten and not shown to users.

## Solution

User lowercase, and prevent overwriting.

---

### DIFF
--- a/packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
+++ b/packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
@@ -423,8 +423,8 @@ export class ProposedTransformationExplorer {
                             programmingLanguage: {
                                 languageName:
                                     transformByQState.getTransformationType() === TransformationType.LANGUAGE_UPGRADE
-                                        ? 'JAVA'
-                                        : 'SQL',
+                                        ? 'java'
+                                        : 'sql',
                             },
                             linesOfCodeChanged: metricsData.linesOfCodeChanged,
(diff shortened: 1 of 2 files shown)


ntarakad-aws (approved):

dhasani23:
Note: I *may* add a feature flag around this SQL Conversions feature in this PR too, will let you know by the early afternoon of Wednesday 11/20

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -423,8 +423,8 @@ export class ProposedTransformationExplorer {
                             programmingLanguage: {
                                 languageName:
                                     transformByQState.getTransformationType() === TransformationType.LANGUAGE_UPGRADE
-                                        ? 'JAVA'
-                                        : 'SQL',
+                                        ? 'java'
```
nkomonen-amazon, line 426: If there is potential that these values will be used in other telemetry events this may be worth putting in to a function

nkomonen-amazon (approved):

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
dhasani23: Yeah, post-re:Invent planning on adding a separate utils file for these kinds of things

jpinkney-aws (approved):

PR #6064, opened 2024-11-20
include pl1 as a code file extension

scottwritescode:
## Problem
The codefileExtensions does not contain an extension type for `.pl1` which is the common extension for PL/1. Because of this, PL/1 files and code are not seen in the workspace context for Amazon Q, and chat interactions with Q and `.pl1` file types are not possible.

## Solution
This adds the PL/1 code extension `.pl1` as a known extension

---


scottwritescode:
Closing to run additional tests

PR #6065, opened 2024-11-20
fix(lambda): add artificat bucket source prompter in SAM sync wizards

vicheey:
## Problem
Unlike in SAM deploy wizard, SAM sync wizard skip a question about bucket source for sync artifact and go directly to question for bucket name.

## Solution
Add prompter asking for bucket source option: 
1) Create a SAM CLI managed S3 bucket
2) Specify an S3 bucket
---


packages/toolkit/.changes/next-release/Breaking Change-782109cf-be90-4472-8600-8e90ed0ce398.json
seshubaws, line 1: Would this be a breaking change or a bug fix? I suppose if users have auto workflows to handle the quickpicks then it might break those flows, but I would say it's more of a bug fix

seshubaws (approved):

jpinkney-aws (approved):

PR #6069, opened 2024-11-20
fix(lambda): enable template parameter prompters for all deploy and sync entry points

vicheey:
## Problem
When deploying a template with parameters, SAM guided action prompts customers to confirm to either use default value or update to a new value values. Such prompters for template parameters value only exist for Deploy Wizard when invoke from : 1). Template file context menu, 2). App builder node deploy button.

Such prompter is missing in 
 - Deploy Wizard entry points: 
    1. Command palette entry points
    2. Region node context menu
- Sync Wizard  entry points: 
    1. Command palette 
    2. Region node context menu
    3. Template file context menu
    4. Samconfig file context menu
    5. App Builder node  deploy button

## Solution
- Implement nested wizard for template parameter for both sync and deploy action for all entry points.
- Refactor DeployWizard class for consistency with SyncWizard class
- Add unit test for validating correct backward flow and state restoration for both wizard.

---


vicheey:
Close this in favor of [ #6240 ](https://github.com/aws/aws-toolkit-vscode/pull/6240)

PR #6073, opened 2024-11-21
fix(cwl): Change wording in menus, add info message when opening running session

keeganirby:
## Problem
Addresses feedback after working with Toolkit UXD. There are inconsistencies with formatting/capitalization of certain nouns in the LiveTail menu. Some of the verbiage is unclear/too lengthy.

In the case there is an already running session and the user tries to start a new session with the same parameters, it is unclear that we are opening an already existing session and not starting a new one. 

## Solution
* Consistent spacing and capitalization of "Log Group" and "Log Stream"
* Introduce new Info window when opening and existing session stream
* Re-words LogStream filter type sub menu.

---

### DIFF
--- a/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
+++ b/packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
@@ -44,9 +44,7 @@ export async function tailLogGroup(
         const session = new LiveTailSession(liveTailSessionConfig)
         if (registry.has(uriToKey(session.uri))) {
             await vscode.window.showTextDocument(session.uri, { preview: false })
-            void vscode.window.showInformationMessage(
-                `Switching editor to an existing session. Found running session that matches request.`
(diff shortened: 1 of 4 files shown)


packages/core/src/awsService/cloudWatchLogs/commands/tailLogGroup.ts
```diff
@@ -43,7 +43,10 @@ export async function tailLogGroup(
         }
         const session = new LiveTailSession(liveTailSessionConfig)
         if (registry.has(uriToKey(session.uri))) {
-            await prepareDocument(session)
+            await vscode.window.showTextDocument(session.uri, { preview: false })
+            void vscode.window.showInformationMessage(
+                `Switching editor to an existing session. Found running session that matches request.`
```
jpinkney-aws, line 48: I feel like these two sentences mean the same thing, we can we simplify them? Otherwise the PR LGTM

PR #6074, opened 2024-11-21
fix(policychecks): Update Policy Checks to use profile selected by AWS Toolkits instead of always default

kevluu-aws:
## Problem
Policy Checks currently only uses the default profile within the credentials file for the CLI tool and SDK calls made to A2.

## Solution
Pass on the profile name from the current profile setup in AWS Toolkits to the CLI tool and SDK.

---


jpinkney-aws (approved):
When someone from your team approves I'll merge it

packages/core/src/awsService/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -179,7 +179,9 @@ export class IamPolicyChecksWebview extends VueWebview {
                             documentType,
```
mponaws, line 179: Firstly, thank you @kevluu-aws for fixing it! 🙌

packages/core/src/awsService/accessanalyzer/vue/iamPolicyChecks.ts
```diff
@@ -919,6 +943,11 @@ export function isJsonPolicyLanguage(document: string) {
     return policyLanguageFileTypes.some((t) => document.endsWith(t))
 }
 
+export function getProfileName(): string | undefined {
+    // We neeed to split the name on 'profile:' to extract the correct profile name
+    return globals.awsContext.getCredentialProfileName()?.split('profile:')[1]
```
mponaws, line 948: Is this a constant format? I am trying to understand who enforces this format?

packages/core/src/test/awsService/accessanalyzer/iamPolicyChecks.test.ts
```diff
@@ -155,6 +155,8 @@ describe('validatePolicy', function () {
                 'us-east-1',
```
mponaws, line 155: Can we add a test that will help us determine if there is any change to the expected format in the `getProfileName()` function?

packages/core/src/awsService/accessanalyzer/vue/iamPolicyChecks.ts
mponaws: Also, can we add an error message on the backend to determine any incorrect format here?

kevluu-aws: Toolkits uses the format in displaying the profile name to users after authenticating through the extension. It's stored with `profile:` prepended in the Credentials object.

packages/core/src/test/awsService/accessanalyzer/iamPolicyChecks.test.ts
kevluu-aws: The CLI tool tests check for a profile to be passed as `undefined`. This will check if the 'profile:' part gets changed

nkomonen-amazon (approved):

packages/core/src/test/awsService/accessanalyzer/iamPolicyChecks.test.ts
mponaws: Had a chat with Kevin offline regarding if there is a way for us to catch the change in the format of the profile name, and he confirmed that it should be caught in the unit tests.

mponaws (approved):

justinmk3:
Does this fix a user-facing symptom? If so, please add a changelog as described in https://github.com/aws/aws-toolkit-vscode/pull/6074#issuecomment-2489829941

kevluu-aws:
Added changelog describing the fix

justinmk3:
Please try doing a squash-merge locally (`git merge --squash`, or whatever tool you prefer). That will fix the commit history here.

If this still adds 1700 lines of code then next question is why :)

kevluu-aws:
> Please try doing a squash-merge locally (`git merge --squash`, or whatever tool you prefer). That will fix the commit history here.
> 
> If this still adds 1700 lines of code then next question is why :)

Accidental rebasing caused it, just fixed it and rebased with updated changelog

justinmk3:
The jscpd ci check (for redundant code) is failing, and its logs point to code that could easily be deduplicated with minimal effort.

packages/core/src/test/awsService/accessanalyzer/iamPolicyChecks.test.ts
```diff
@@ -579,6 +589,8 @@ describe('customChecks', function () {
                     document,
                     '--region',
                     'us-east-1',
+                    '--profile',
+                    'undefined',
```
justinmk3, line 593: All of the tests assert "undefined". Is this PR intended to support the non-undefined case?

kevluu-aws: These test cases do not actually ever use this argument, as we stub the actual execution of our CLI tool here. It simply looks at the global `globals.awsContext.getCredentialProfileName()`, which is `undefined` in the test execution context.

Is there a way to change the current profile in the test execution context? If we are able to, we can assert it is propagated properly.

justinmk3: > Is there a way to change the current profile in the test execution context?

Look at existing code.

justinmk3:
Merging to unblock. Please followup on the above comments.

PR #6083, opened 2024-11-21
test(logger): test logger does not format objects as expected

Hweinstock:
## Problem
The test logger does not format additional arguements into the logs in the same way as ToolkitLogger.
Ex. `getLogger().debug('here is the obj: %O', obj)` will produce two `LogEntries` one with `"here is the obj: %O"` and one with the object itself.

This is problematic because it causes confusion (see: https://github.com/aws/aws-toolkit-vscode/pull/5895#discussion_r1832979093) and it also can cause `assertLogsContain` to throw an error since there is now a non-string/error entry in the log entry list (see: https://github.com/aws/aws-toolkit-vscode/blob/338ea67f2ba0294fc535a9a949fd1cdaeaa96d98/packages/core/src/test/globalSetup.test.ts#L171-L185).

## Solution
- Have test logger inherit from `BaseLogger` to minimize implementation dupe. 
- If we see a string, format it with extra inputs before pushing it into entries. 
- Add tests for the `testLogger`. 
- Adjust existing tests that relied on this behavior. 
- If on web, we simply concat the strings to avoid reimplementing `util.format`: https://github.com/nodejs/node/blob/3178a762d6a2b1a37b74f02266eea0f3d86603be/lib/internal/util/inspect.js#L2191-L2315

---


packages/core/src/test/shared/schemas.test.ts
```diff
@@ -42,32 +42,36 @@ describe('SchemaService', function () {
 
     it('assigns schemas to the yaml extension', async function () {
```
Hweinstock, line 43: when passing some objects through `util.format` I noticed it set a property `_formatted` (perhaps as a cache?), so without these changes the test fails.

packages/core/src/test/testLogger.ts
```diff
@@ -3,64 +3,48 @@
...
 import { Uri } from 'vscode'
+import util from 'util'
+import { isWeb } from '../shared'
+import { inspect } from '../shared/utilities/collectionUtils'
 
 /**
  * In-memory Logger implementation suitable for use by tests.
  */
-export class TestLogger implements Logger {
+export class TestLogger extends BaseLogger {
```
justinmk3, line 16: Nice, was just thinking about this!

justinmk3 (approved):
Great!

Issue #6079, opened 2024-11-21
unreliable test: crossFileContextUtil, for t1 group, should return repomap + opentabs context


Hweinstock:
## Test Details

-   Name of test: crossFileContextUtil, fetchSupplementalContextForSrc, for t1 group, should return repomap + opentabs context:
-   OS of failure: Windows
-   VSCode version of failure (minimum/insider/stable): Insider
-   Link to failing run: https://github.com/aws/aws-toolkit-vscode/actions/runs/11938979186/job/33278472299
-   Link to failing test: https://github.com/aws/aws-toolkit-vscode/blob/284455442b130e9b81b3def5c7a80d45668c802b/packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts#L64-L97

## Log of Test Failure
``` 
1) crossFileContextUtil
       fetchSupplementalContextForSrc
         for t1 group, should return repomap + opentabs context:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

0 !== 4

      + expected - actual

      -0
      +4
      
      at Context.<anonymous> (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\amazonq\test\unit\codewhisperer\util\crossFileContextUtil.test.ts:86:20)

deleteTestTempDirs: deleted 103 test temp dirs
Error: 1 tests failed.
    at D:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\core\src\test\testRunner.ts:124:28
    at MultiReporters.done (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\node_modules\mocha-multi-reporters\lib\MultiReporters.js:124:16)
    at done (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\node_modules\mocha\lib\mocha.js:1007:16)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
Extension host test runner error Error: 1 tests failed.
    at D:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\core\src\test\testRunner.ts:124:28
Asking native host service to exit with code 1.
    at MultiReporters.done (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\node_modules\mocha-multi-reporters\lib\MultiReporters.js:124:16)
    at done (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\node_modules\mocha\lib\mocha.js:1007:16)
[...]

PR #6085, opened 2024-11-21
test(amazonq): skip flaky crossfileContextUtil.test

Will-ShaoHua:
## Problem
#6079

## Solution
Ignore the flaky test cases for now until we root caused why it failed occasionally in some CI machines

---

### DIFF
--- a/packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts
+++ b/packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts
@@ -35,7 +35,7 @@ describe('crossFileContextUtil', function () {
         sinon.restore()
     })
 
-    describe.skip('fetchSupplementalContextForSrc', function () {
+    describe('fetchSupplementalContextForSrc', function () {
         beforeEach(async function () {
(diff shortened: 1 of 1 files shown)


packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -35,7 +35,7 @@ describe('crossFileContextUtil', function () {
         sinon.restore()
     })
 
-    describe('fetchSupplementalContextForSrc', function () {
+    describe.skip('fetchSupplementalContextForSrc', function () {
```
Hweinstock, line 38: I've only seen the `for t1 group, should return repomap + opentabs context` one fail. Is there a reason we want to skip the other two in this block?

Will-ShaoHua: um, i just think the t1 test are pretty much the same as t2 test, so it's still unclear to me why one fails but the other doesn't.
However let's skip t1 test only for now and skip others when we observe other failure, also it could help me narrow down where it likely went wrong.

jpinkney-aws (approved):

Issue #6043, opened 2024-11-18
unreliable test: Error: Timed-out waiting for browser login flow to complete


Hweinstock:
unhandled promise rejection: 
``` 
rejected promise not handled within 1 second: Error: Timed-out waiting for browser login flow to complete
stack trace: Error: Timed-out waiting for browser login flow to complete
    at poll (/codebuild/output/src312647086/src/github.com/aws/aws-toolkit-vscode/packages/core/src/auth/sso/ssoAccessTokenProvider.ts:368:15)
    at async openBrowserAndWaitUntilComplete (/codebuild/output/src312647086/src/github.com/aws/aws-toolkit-vscode/packages/core/src/auth/sso/ssoAccessTokenProvider.ts:468:20)
    at async DeviceFlowAuthorization.authorize (/codebuild/output/src312647086/src/github.com/aws/aws-toolkit-vscode/packages/core/src/auth/sso/ssoAccessTokenProvider.ts:482:33)
    at async DeviceFlowAuthorization.runFlow (/codebuild/output/src312647086/src/github.com/aws/aws-toolkit-vscode/packages/core/src/auth/sso/ssoAccessTokenProvider.ts:142:28)
    at async DeviceFlowAuthorization.createToken (/codebuild/output/src312647086/src/github.com/aws/aws-toolkit-vscode/packages/core/src/auth/sso/ssoAccessTokenProvider.ts:125:24)
Timed-out waiting for browser login flow to complete: Error: Timed-out waiting for browser login flow to complete
    at poll (/codebuild/output/src312647086/src/github.com/aws/aws-toolkit-vscode/packages/core/src/auth/sso/ssoAccessTokenProvider.ts:368:15)
    at async openBrowserAndWaitUntilComplete (/codebuild/output/src312647086/src/github.com/aws/aws-toolkit-vscode/packages/core/src/auth/sso/ssoAccessTokenProvider.ts:468:20)
    at async DeviceFlowAuthorization.authorize (/codebuild/output/src312647086/src/github.com/aws/aws-toolkit-vscode/packages/core/src/auth/sso/ssoAccessTokenProvider.ts:482:33)
    at async DeviceFlowAuthorization.runFlow (/codebuild/output/src312647086/src/github.com/aws/aws-toolkit-vscode/packages/core/src/auth/sso/ssoAccessTokenProvider.ts:142:28)
[...]

PR #6088, opened 2024-11-22
test(flaky): fix unresolved promise causing linux failures

Hweinstock:
## Problem
https://github.com/aws/aws-toolkit-vscode/issues/6043
To reproduce, add a 5 second delay to the `after` hook at the top level. 
One way to do this is to insert this at line 84. 
``` 
after(async function () {
        clock.uninstall()
        await sleep(5000)
    })
```
Despite asserting that the promise rejects within the test, the promise rejects after the test as well. Not entirely sure why this happening. 
- Tried manually wrapping in try-catch with an `await` instead of `assert.rejects` and it still fails. 
- Tried wrapping the promise in another promise before passing to `assert.rejects`. 

## Solution
What does appear to work, is manually handling the callback of the promise. That is, explicitly defining a `then` and `catch` method to assert the rejection, and awaiting the promise at the end of the test to ensure it resolves before the test finishes. Not sure why this works, but I am unable to reproduce the error with this change. 

## Notes
- `assert.rejects` implementation: https://github.com/nodejs/node/blob/3178a762d6a2b1a37b74f02266eea0f3d86603be/lib/assert.js#L653. Doesn't appear to be the problem because the same is observed when manually wrapping. 
- `await` docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await 

---


justinmk3 (approved):
Nice!

justinmk3:
This should allow use to remove `Timed-out waiting for browser login flow` from this ignore-pattern, now: https://github.com/aws/aws-toolkit-vscode/blob/b52666e480e9db2d05ae913c559498b34348f0e4/buildspec/shared/common.sh#L12

Let's include that with this commit, since they are inherently coupled.

PR #6089, opened 2024-11-22
fix(lambda): save params before running sam command when --watch is selected

mbfreder:
## Problem
When --watch flag is used, the sync process remains alive. The region and stack_name info get written to config file only after the process finishes. This means customer would not be able to see (or refresh) the latest deployed resoures during the sync process. 

## Solution
Write to samconfig.toml before running sam sync

---


mbfreder:
One test is failing here but not on my local machine. Investigating this.

nkomonen-amazon (approved):

packages/core/src/shared/sam/sync.ts
```diff
@@ -354,9 +354,15 @@ export async function runSamSync(args: SyncParams) {
         }),
     })
 
-    await runInTerminal(sam, 'sync')
+    // with '--watch' selected, the sync process will run in the background until the user manually kills it
+    // we need to save the stack and region to the samconfig file now, otherwise the user would not see latest deployed resoure during this sync process
     const { paramsSource, stackName, region, projectRoot } = args
     const shouldWriteSyncSamconfigGlobal = paramsSource !== ParamsSource.SamConfig && !!stackName && !!region
+    if (boundArgs.includes('--watch')) {
+        shouldWriteSyncSamconfigGlobal && (await writeSamconfigGlobal(projectRoot, stackName, region))
```
vicheey, line 362: [Curious]: Shouldn't we just have the following line before `await runInTerminal(sam, 'sync')`?
```
shouldWriteSyncSamconfigGlobal && (await writeSamconfigGlobal(projectRoot, stackName, region))
```

mbfreder: I guess we could, if there's no specific reason why we added that line after running the command.

vicheey: Adding here for historical context: [commit](https://github.com/aws/aws-toolkit-vscode/commit/fc3bef8423b8307c0c9d1cd787bc980c28d9d3e9#diff-d5821c898fe7f9461d177f2983493eda29d0cbc7d3fc50af8ccf1c43a93044d1)
> Auto refresh feature trigger extension to fetch for deployed stack node
using `stack_name` and `region` from `samconfig.toml` file. In current
logic, the `stack_name` and `region` are written before deployment
process completed to avoid the edge case when deployment failure due to
stack up-to-date. However, this order triggers the extension to fetch
for stack that may not already been created and through an error toast
message to customer stating "Stack does not exist".

Because of this, we only write `stack_name` and `region` information after deployment/syncing completes. IN case of using `--watch` for sync, we should write it before the process completed. 

That's said, we can proceed with current approach.

vicheey (approved):

packages/core/src/shared/sam/sync.ts
vicheey: Adding [past commit here](hhttps://github.com/aws/aws-toolkit-vscode/commit/fc3bef8423b8307c0c9d1cd787bc980c28d9d3e9#diff-d5821c898fe7f9461d177f2983493eda29d0cbc7d3fc50af8ccf1c43a93044d1) for historical context: 
 
> Auto refresh feature trigger extension to fetch for deployed stack node
using `stack_name` and `region` from `samconfig.toml` file. In current
logic, the `stack_name` and `region` are written before deployment
process completed to avoid the edge case when deployment failure due to
stack up-to-date. However, this order triggers the extension to fetch
for stack that may not already been created and through an error toast
message to customer stating "Stack does not exist".

For the case of using `--watch` with sync, we can proceed with current approach.

packages/core/src/test/shared/sam/sync.test.ts
```diff
@@ -1030,6 +1030,91 @@ describe('SAM runSync', () => {
             prompterTester.assertCallAll()
         })
 
+        it('[entry: command palette] specify and save flag (with --watch) should save params before starting SAM process', async () => {
```
justinmk3, line 1033: 95% of the 100 lines in this test are identical to the test above it. The `Copy-Paste Detection` CI job warned about this. This is not acceptable in this codebase.

packages/toolkit/.changes/next-release/Bug Fix-399d3f79-390e-4892-8cf5-e4a189860334.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "appBuilder refresh feature doesnt work during sync --watch"
```
justinmk3, line 3: Is this description meaningful to customers?

PR #6095, opened 2024-11-22
fix(cwl): Dispose event listeners when command exits

keeganirby:
## Problem
TailLogGroup command is not being fully cleaned up properly. The event listeners for: 
* showing the StatusBar item
*  closing a session when exiting its editors  

are not being disposed of. As users stop sessions, these listeners remain active whilst doing no meaningful work. This is effectively a memory leak.

Additionally during testing, I noticed that we are adding a session to the LiveTailRegistry before the session is actually started. This could cause an issue where a session stream isn't successfully created, but the session is still registered. 

## Solution
* Create an array of disposables and dispose of them as we exit the TailLogGroup command
* Only register a session after startLiveTail returns

Because now we dispose of the "tab close" listener when tailLogGroup exits, changes are needed in the unit tests to keep the mock response stream "open" as the test executes. This keeps tailLogGroup from returning/disposing while we perform test assertions and close the editors. 

---

### DIFF
--- a/packages/core/src/test/awsService/cloudWatchLogs/commands/tailLogGroup.test.ts
+++ b/packages/core/src/test/awsService/cloudWatchLogs/commands/tailLogGroup.test.ts
@@ -4,23 +4,20 @@
  */
 
 import * as sinon from 'sinon'
-import * as FakeTimers from '@sinonjs/fake-timers'
 import * as vscode from 'vscode'
 
 import assert from 'assert'
 import { clearDocument, closeSession, tailLogGroup } from '../../../../awsService/cloudWatchLogs/commands/tailLogGroup'
-import { LiveTailSessionLogEvent, StartLiveTailResponseStream } from '@aws-sdk/client-cloudwatch-logs'
+import { StartLiveTailResponseStream } from '@aws-sdk/client-cloudwatch-logs'
 import { LiveTailSessionRegistry } from '../../../../awsService/cloudWatchLogs/registry/liveTailSessionRegistry'
 import { LiveTailSession } from '../../../../awsService/cloudWatchLogs/registry/liveTailSession'
-import { asyncGenerator } from '../../../../shared/utilities/collectionUtils'
 import {
     TailLogGroupWizard,
     TailLogGroupWizardResponse,
 } from '../../../../awsService/cloudWatchLogs/wizard/tailLogGroupWizard'
 import { getTestWindow } from '../../../shared/vscode/window'
 import { CloudWatchLogsSettings, uriToKey } from '../../../../awsService/cloudWatchLogs/cloudWatchLogsUtils'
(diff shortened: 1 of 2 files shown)


packages/core/src/test/awsService/cloudWatchLogs/commands/tailLogGroup.test.ts
```diff
@@ -97,10 +97,15 @@ describe('TailLogGroup', function () {
...
         })
-        await tailLogGroup(registry, testSource, codeLensProvider, {
+
+        //The mock stream doesn't 'close', causing tailLogGroup to not return. If we `await`, it will never resolve.
+        //Run it in the background and sleep, giving the command time to process the session update frames.
+        void tailLogGroup(registry, testSource, codeLensProvider, {
             groupName: testLogGroup,
             regionName: testRegion,
         })
+        await new Promise((r) => setTimeout(r, 500))
```
jpinkney-aws, line 107: instead of doing a sleep can we do a waitUntil something happens?

keeganirby: Currently, TailLogGroup doesn't expose anything that we can wait on that tells us that we have done the work required to do our test assertions. 

We could possibly emit an event each time the command processes a new Stream frame. Then in the test we would waitUntil the event is emitted `X` times. 

So far the actual feature hasn't required that, but I can look into adding that for test purposes.

keeganirby:
Accident closed while trying to update my local feature branch against the remote branch.

packages/core/src/test/awsService/cloudWatchLogs/commands/tailLogGroup.test.ts
keeganirby: Was able to use waitUntil by waiting on the LiveTailRegistry to see when there's a session in there (signals session started).

Then we can wait until the content in the document matches the expected. This is does both: validating the content of the doc, and waiting until the stream is fully processed.

packages/core/src/test/awsService/cloudWatchLogs/commands/tailLogGroup.test.ts
```diff
@@ -39,23 +36,12 @@ describe('TailLogGroup', function () {
     let cloudwatchSettingsSpy: sinon.SinonSpy
     let wizardSpy: sinon.SinonSpy
 
-    let clock: FakeTimers.InstalledClock
-
-    before(function () {
-        clock = installFakeClock()
```
keeganirby, line 45: Using fake clocks is screwing with the `waitUntil()` calls. Inspecting the test that uses the clock: the behavior under test is a functionality of `LiveTailSession`, and not `TailLogGroup`. If we want to assert that the timer is disposed of, it should be done in those unit tests, and not here. 

Removing the clock.

PR #6097, opened 2024-11-22
deps(sdkv3): start migration to v3 with new client builder

Hweinstock:
## Problem
Start the transition to sdkv3. This PR is just a start and is still missing a lot of features (see future work). 
Originally part of https://github.com/aws/aws-toolkit-vscode/pull/5940, but decided to split it up. 

## Solution
- Do not overwrite existing builder to allow for incremental transition to V3.
- A barebones working model, with the telemetry middleware implemented.
- Expansion of testing for client construction.

## Future Work
- Implement middleware for other relevant pieces (custom endpoints, logging, etc.). https://github.com/Hweinstock/aws-toolkit-vscode/pull/4
- Increase testing coverage, specifically with credentials refreshing. (https://github.com/Hweinstock/aws-toolkit-vscode/pull/3)
- Migrate clients. (ssm: https://github.com/aws/aws-toolkit-vscode/pull/6137)
- Reduce boilerplate for client wrappers and have all clients call destroy. 
- Avoid recreating clients on each request, use singleton instead. 
- replace hand-made mocks with those from sdk mocking library: https://aws.amazon.com/blogs/developer/mocking-modular-aws-sdk-for-javascript-v3-in-unit-tests/
---


packages/core/src/shared/awsClientBuilderV3.ts
```diff
@@ -0,0 +1,152 @@
...
+ * Record request IDs to the current context, potentially overriding the field if
+ * multiple API calls are made in the same context. We only do failures as successes are generally uninteresting and noisy.
+ */
+export function recordErrorTelemetry(err: Error, serviceName?: string) {
+    telemetry.record({
+        requestId: getRequestId(err),
+        requestServiceType: serviceName,
+        reasonDesc: getTelemetryReasonDesc(err),
+        reason: getTelemetryReason(err),
+        result: getTelemetryResult(err),
```
justinmk3, line 119: The old `awsClientBuilder.ts` was reluctant to set these fields. But I think setting them here is low-risk and potentially adds useful info:

- if the caller is wrapped in `telemetry.xx.run()` then these fields will be overridden by the run() context finalizer anyway
- else, these fields give some useful info

todo: Maybe the `run()` finalizer logic should be smarter and check if there was a SDK failure (these fields)?

packages/core/src/shared/awsClientBuilderV3.ts
```diff
@@ -0,0 +1,152 @@
...
+        const err = { ...e }
+        delete err['stack']
+        getLogger().error('API Response %s: %O', errorMessageAppend, err)
+    }
+    throw e
+}
+/**
+ * Telemetry logic to be added to all created clients. Adds logging and emitting metric on errors.
+ */
+
```
justinmk3, line 135: is this blank line intentional?

packages/core/src/shared/awsClientBuilderV3.ts
```diff
@@ -0,0 +1,152 @@
...
+        if (!HttpResponse.isInstance(args.request)) {
+            return next(args)
+        }
+        const serviceId = getServiceId(context as object)
+        const { hostname, path } = args.request
+        const logTail = `(${hostname} ${path})`
+        const result = await next(args).catch((e: any) => logAndThrow(e, serviceId, logTail))
+        if (HttpResponse.isInstance(result.response)) {
+            // TODO: omit credentials / sensitive info from the logs / telemetry.
+            const output = omitIfPresent(result.output, [])
```
justinmk3, line 147: any reason not to use `partialClone` ? https://github.com/aws/aws-toolkit-vscode/blob/16e5138bc7bf8df7c6943e2fbb7c62416f4688f9/packages/core/src/auth/sso/clients.ts#L256-L262

`omitIfPresent` was dropped in https://github.com/aws/aws-toolkit-vscode/commit/b04054ea50aba013278404c2dc9be985468e4b37 because `partialClone` is a more general solution

justinmk3 (approved):
While we're here let's rename `packages/core/src/test/shared/defaultAwsClientBuilder.test.ts` to `awsClientBuilder.test.ts`

PR #6098, opened 2024-11-23
config(amazonq): enable implicit project context ab test for broader users

Will-ShaoHua:
## Problem
enable implicit workspace ab test for broader groups of users but not only internal users

## Solution
remove isInternalUser predicate

---


packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -637,7 +636,6 @@ export class ChatController {
             // if user does not have @workspace in the prompt, but user is Amazon internal
             // add project context by default
```
hayemaxi, line 637:
```suggestion
```

hayemaxi (approved):

leigaol (approved):

jpinkney-aws (approved):

PR #6103, opened 2024-11-25
fix(lambda): Redirect view logs action to open the terminal instead of output channel

mbfreder:
## Problem

When a build error occurs, the error details are correctly shown under the "Terminal" tab. However, the toast notification says "build failed" with a button to "See logs." Clicking this button incorrectly redirects the user to the "Output" tab, which only shows a generic message (Error: sam build exited with a non-zero exit code: 1 [NonZeroExitCode]), instead of the detailed logs from the "Terminal" tab. This creates confusion and makes troubleshooting more difficult.

## Solution
Redirect view logs action to open the terminal instead of output channel

---


hayemaxi:
>   

Users can notice this change - please add a changelog item.

hayemaxi (approved):

PR #6105, opened 2024-11-25
test: add notification and reset state dev options

hayemaxi:
- Add a generic dev menu item to reset the state of a feature. Simply extend with a function handler to add an item to the menu. The initial item resets in-ide notification state.
image
image

- Add a way to test notifications locally without modifying code. Using a dev menu item you can add a notification to global state and receive it in your IDE for visualization.
image
image

- Refactor notifications controller to not require a RuleEngine when polling is requested.

---

### DIFF
--- a/packages/core/src/notifications/activation.ts
+++ b/packages/core/src/notifications/activation.ts
@@ -4,9 +4,10 @@
  */
 
 import * as vscode from 'vscode'
-import { NotificationsController } from './controller'
+import { DevSettings } from '../shared/settings'
+import { DevFetcher, NotificationsController, RemoteFetcher } from './controller'
 import { NotificationsNode } from './panelNode'
-import { RuleEngine, getRuleContext } from './rules'
+import { getRuleContext } from './rules'
 import globals from '../shared/extensionGlobals'
 import { AuthState } from './types'
 import { getLogger } from '../shared/logger/logger'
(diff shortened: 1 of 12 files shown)


packages/core/src/notifications/activation.ts
```diff
@@ -24,25 +25,26 @@ const emergencyPollTime = oneMinute * 10
...
-            const ruleContext = await getRuleContext(context, await authStateFn())
-            await controller.pollForEmergencies(new RuleEngine(ruleContext))
-        }, emergencyPollTime)
+        const controller = new NotificationsController(
+            panelNode,
+            async () => await getRuleContext(context, await authStateFn()),
+            DevSettings.instance.isDevMode() ? new DevFetcher() : new RemoteFetcher()
+        )
+
+        await controller.pollForStartUp()
```
nkomonen-amazon, line 39: Are these making a network call? If so I wonder if we should not block to improve startup performance

nkomonen-amazon (approved):

nkomonen-amazon:
Thanks for adding in the reset state feature, will be useful for a lot of things

packages/core/src/notifications/activation.ts
hayemaxi: Yes excellent catch, this was a bug fixed by https://github.com/aws/aws-toolkit-vscode/pull/6052

PR #6108, opened 2024-11-26
ci(automation): added openGate build spec

tomcat323:
## Problem
We close source gate during the start of a release to prevent unwanted commits coming in, but we have to remember to manually open it again. 

We want to automatically re-open the source gate after a release is triggered. 

This PR creates the buildSpec file to be used by codeBuild project in our release pipeline to open the source transition.

Related infra CR: https://code.amazon.com/reviews/CR-162245234/revisions/1#/diff

---

### DIFF
--- a/buildspec/release/35opengate.yml
+++ b/buildspec/release/35opengate.yml
@@ -1,0 +1,18 @@
+version: 0.2
+
+phases:
+    install:
+        runtime-versions:
+            nodejs: 16
+
+    pre_build:
+        commands:
+            - STAGE_NAME=SourceWithGit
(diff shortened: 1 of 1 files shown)


hayemaxi:
Can we add this to the existing file `30closegate.yml`

hayemaxi:
>   

The PR title/commit should use `ci` instead of `feat`

tomcat323:
> Can we add this to the existing file `30closegate.yml`

I don't think so, because some actions only uses `closegate` but not `opengate`.

justinmk3 (approved):

hayemaxi (approved):

PR #6113, opened 2024-11-27
build(lint): disallow node process api, favor Toolkit ChildProcess api

Hweinstock:
## Problem
contributors may use the low-level node `child_process` library, which lacks the instrumentation and robustness of the toolkit `ChildProcess` utility. 

Example: https://github.com/aws/aws-toolkit-vscode/pull/5925#discussion_r1841277911
## Solution
- add the node `child_process` library as a "banned import" with a message pointing to our `ChildProcess` utility. 
- Migrate simple cases to use our `ChildProcess`, mark more complex ones as exceptions. 

## Future Work
- To migrate the existing cases marked as exceptions, we will need to add functionality to our `ChildProcess` api, such as enforcing a `maxBufferSize` on the output.
- Additionally, some of the current uses of `child_process` are tied to implementation details in the `child_process` library that make it hard to change without a larger-scale refactor. These cases were marked as exceptions. 

---


.eslintrc.js
```diff
@@ -199,6 +199,11 @@ module.exports = {
                         name: 'fs',
                         message: 'Avoid node:fs and use shared/fs/fs.ts when possible.',
                     },
+                    {
+                        name: 'child_process',
+                        message:
+                            'Avoid child_process, use ChildProcess wrapper from `packages/core/src/shared/utilities/processUtils.ts` instead.',
```
justinmk3, line 205:
```suggestion
                            'Avoid child_process, use ChildProcess from `shared/utilities/processUtils.ts` instead.',
```

packages/core/scripts/test/launchTestUtilities.ts
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import * as proc from 'child_process'
+import * as proc from 'child_process' // eslint-disable-line no-restricted-imports
```
justinmk3, line 6: 👍 in build-scripts we generally want to stick to plain old node APIs.

packages/core/src/shared/utilities/processUtils.ts
```diff
@@ -100,6 +100,13 @@ export class ChildProcess {
         // TODO: allow caller to use the various loggers instead of just the single one
         this.#log = baseOptions.logging !== 'no' ? logger.getLogger() : logger.getNullLogger()
     }
+    public static async exec(
```
justinmk3, line 103: can we call this `run`? should avoid different names for the same thing.

packages/core/src/shared/sam/debugger/goSamDebug.ts
```diff
@@ -187,7 +189,9 @@ async function makeInstallScript(debuggerPath: string, isWindows: boolean): Prom
 
     // It's fine if we can't get the latest Delve version, the Toolkit will use the last built one instead
     try {
-        const goPath: string = JSON.parse(execFileSync('go', ['env', '-json']).toString()).GOPATH
+        const goPath: string = JSON.parse(
+            (await ChildProcess.exec('go', ['env', '-json'], { collect: true })).stdout
+        ).GOPATH
```
justinmk3, line 194: Use a local var is a bit more terse and readable, at the cost of having to claim a name.
But generally, invoking a process almost always should involve checking an exit code (though that's out of scope for this PR), so inlining `exec()` or `run()` is unusual and likely discouraged. .

```suggestion
        const r = await ChildProcess.exec('go', ['env', '-json'], { collect: true })
        const goPath: string = JSON.parse(r.stdout).GOPATH
```

justinmk3 (approved):

justinmk3:
This is good to have in place. We don't need to bother migrating any of the child_process cases for Q features, since most of those will use Flare.

packages/core/src/shared/utilities/processUtils.ts
Hweinstock: Original intention was to mirror `child_process` `exec`, but I agree the consistency is more important.

packages/core/src/shared/sam/debugger/goSamDebug.ts
Hweinstock: Ideally checking the exit code would be part of some higher-level API provided by ChildProcess right?

justinmk3 (approved):

packages/core/src/shared/sam/debugger/goSamDebug.ts
justinmk3: > Ideally checking the exit code would be part of some higher-level API provided by ChildProcess right?

Failure handling in some way needs to be done by the caller, typically

PR #6107, opened 2024-11-26
feat(amazonq): support Dockerfile files in /dev

michaellandi:
## Problem
`Dockerfile` files are not currently supported by /dev

## Solution
Add a list of file prefixes that are considered code files for /dev.  This approach supports not only `Dockerfile` but also popular variants of this such as `Dockerfile.build`, `Dockerfile-prod`, etc.

I manually tested some prompts in /dev with requests.  I set breakpoints to see that the file was being zipped up in workspace upload.

---

License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6115, opened 2024-11-27
fix(amazonq): /dev to support upload of nested dockerfiles

michaellandi:
[POST RE:INVENT]

## Problem

Nested `Dockerfile` files were not supported with https://github.com/aws/aws-toolkit-vscode/pull/6107

## Solution

Look at path basename when comparing file to list of well known files.

I manually tested adding nested `Dockerfile`s to various folders and ensured that pattern matched.

---

### DIFF
--- a/packages/core/src/shared/filetypes.ts
+++ b/packages/core/src/shared/filetypes.ts
@@ -348,16 +348,11 @@ export const codefileExtensions = new Set([
     '.zig',
 ])
 
-// Some well-known code files without an extension
-export const wellKnownCodeFiles = new Set(['Dockerfile', 'Dockerfile.build'])
+// Code file names without an extension
+export const codefileNames = new Set(['Dockerfile', 'Dockerfile.build'])
 
 /** Returns true if `filename` is a code file. */
 export function isCodeFile(filename: string): boolean {
-    if (codefileExtensions.has(path.extname(filename).toLowerCase())) {
-        return true
-    } else if (wellKnownCodeFiles.has(filename)) {
-        return true
-    } else {
-        return false
-    }
+    const ext = path.extname(filename).toLowerCase()
+    return codefileExtensions.has(ext) || codefileNames.has(path.basename(filename))
 }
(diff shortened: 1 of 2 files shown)


avi-alpert (approved):

neilk-aws (approved):

packages/core/src/shared/filetypes.ts
```diff
@@ -355,7 +355,7 @@ export const wellKnownCodeFiles = new Set(['Dockerfile', 'Dockerfile.build'])
 export function isCodeFile(filename: string): boolean {
```
justinmk3, line 355: please rename `wellKnownCodeFiles` to something like `codefileBaseNames`, so that it follows the existing pattern of `codefileExtensions`.

packages/core/src/shared/filetypes.ts
```diff
@@ -355,7 +355,7 @@ export const wellKnownCodeFiles = new Set(['Dockerfile', 'Dockerfile.build'])
...
     if (codefileExtensions.has(path.extname(filename).toLowerCase())) {
         return true
-    } else if (wellKnownCodeFiles.has(filename)) {
+    } else if (wellKnownCodeFiles.has(path.basename(filename))) {
         return true
     } else {
         return false
```
justinmk3, line 361: would the lowercase step also be needed?

please also revert the change that you made in the previous pr, so that this code is less verbose:

```typescript
    const ext = path.extname(filename).toLowerCase()
    return codefileExtensions.has(ext) || wellKnownCodeFiles.has(path.basename(filename)
```

packages/core/src/shared/filetypes.ts
michaellandi: This already exists here and in the JetBrains extension: https://github.com/aws/aws-toolkit-jetbrains/blob/07ccd110d741c4bf591bfcbc5db86a2a7022313f/plugins/core/jetbrains-community/src/software/aws/toolkits/jetbrains/services/amazonq/FeatureDevSessionContext.kt#L72

packages/core/src/shared/filetypes.ts
michaellandi: I do think we want a case sensitive match in this case.  `Dockerfile` is the convention when used without an extension: https://docs.docker.com/build/concepts/dockerfile/#filename.

packages/core/src/shared/filetypes.ts
justinmk3: Please rename it.

packages/core/src/shared/filetypes.ts
justinmk3: That's the case for `Dockerfile`, but likely not other "filenames without an extension" . I guess we can wait until we encounter such files.

packages/core/src/shared/filetypes.ts
michaellandi: Okay, done.

PR #6121, opened 2024-11-29
build(pattern): remove ignore pattern related to 1.68 vscode version. 

Hweinstock:
## Problem
According to the comment in this file, there is an ignore pattern that is only needed on the vscode minimum version of 1.68, but we have bumped to 1.83.0. 

## Solution
Remove pattern for this case. 

---


jpinkney-aws (approved):
I don't have context as to why that was ever needed but if it was happening on such a long release we should be good now

PR #6137, opened 2024-12-03
refactor(sdkv3): migrate ssm client

Hweinstock:
## Problem
There is a lot of duplicate code across the sdk clients. Additionally, we want to only construct the clients once (per wrapper), and have these clients call destroy. 

## Solution
Note: the line changes are mostly the added dependency of an `@aws-sdk` client and its hundreds of dependencies. 

- create abstract class `ClientWrapper` to minimize code dupe. 
- cache sdk clients on creation within the wrappers. 
- migrate ssm client to sdkv3, using new common wrapper class. 

## Alternative Solution 
- Make all wrappers singletons so clients are created once for the toolkit lifetime. 
- Pro: Makes it impossible for a client to be re-created. 
- Con: Zero flexibility when it comes to client lifetime. All sdk clients would be created whenever their wrapper is referenced. Without being very careful, we could create many SDK clients up front that we never use and can't delete. 

---


packages/core/src/shared/clients/ssm.ts
```diff
@@ -0,0 +1,120 @@
...
+    SendCommandCommandOutput,
+    waitUntilCommandExecuted,
+    SessionState,
+    DescribeSessionsCommand,
+} from '@aws-sdk/client-ssm'
+import { WaiterState } from '@smithy/util-waiter'
+import { ToolkitError } from '../errors'
+import { ClientWrapper } from './wrapper'
+
+export class SSMWrapper extends ClientWrapper<SSMClient> {
```
justinmk3, line 26:
```suggestion
export class SsmWrapper extends ClientWrapper<SSMClient> {
```

jpinkney-aws (approved):

PR #6138, opened 2024-12-03
telemetry(amazon q): adding telemetry after one or multiple diffs form selection

ntarakad-aws:
## Problem
We were only recording the metrics when the user accepts changes so we can track which patches were applied based on their description. We wanted to also keep track of the form in which the user is prompted to pick whether they want their changes in one or multiple diffs.

## Solution
Adding a telemetry call when we handle the form selection for selective transformation one or multiple diffs.

---

### DIFF
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -433,10 +433,12 @@ export class GumbyController {
         } else {
             transformByQState.setMultipleDiffs(false)
         }
-        telemetry.codeTransform_submitSelection.emit({
-            codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId(),
-            userChoice: oneOrMultipleDiffsSelection,
-            result: MetadataResult.Pass,
+        await telemetry.codeTransform_submitSelection.run(async () => {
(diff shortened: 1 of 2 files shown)


packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -433,6 +433,11 @@ export class GumbyController {
         } else {
             transformByQState.setMultipleDiffs(false)
         }
+        telemetry.codeTransform_submitSelection.emit({
```
dhasani23, line 436: We should do this in a `run` block — see line 447 below for an example

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -433,6 +433,13 @@ export class GumbyController {
         } else {
             transformByQState.setMultipleDiffs(false)
         }
+        await telemetry.codeTransform_submitSelection.run(async () => {
+            telemetry.codeTransform_submitSelection.emit({
```
dhasani23, line 437: You can put everything inside this `run` block just like we do below, then `emit` isn't needed, we can just use `telemetry.record()` as done below.

dhasani23 (approved):

justinmk3 (approved):

PR #6148, opened 2024-12-04
feat(amazonq): Create a common langauge server downloader

jpinkney-aws:
## Problem
- we need a way to download the manifest and install the correct mynah ui/language server code

## Solution
- create a common lsp downloader that flare/workspace lsps can use

---


packages/amazonq/src/lsp/activation.ts
```diff
@@ -4,9 +4,19 @@
  */
 
 import vscode from 'vscode'
+import path from 'path'
+import { AmazonQLSPDownloader } from './download'
 
 export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
+    const serverPath = ctx.asAbsolutePath(path.join('resources', 'qdeveloperserver'))
+    const clientPath = ctx.asAbsolutePath(path.join('resources', 'qdeveloperclient'))
```
justinmk3, line 12: string literals never need to be join()'d
```suggestion
    const serverPath = ctx.asAbsolutePath('resources/qdeveloperserver')
    const clientPath = ctx.asAbsolutePath('resources/qdeveloperclient')
```

packages/core/src/shared/downloadLSP.ts
justinmk3, line 1: all files / symbols in the project use camelcase, so "LSP" would be "Lsp"

packages/core/src/shared/downloadLSP.ts
```diff
@@ -0,0 +1,214 @@
...
+    artifactDescription: string
+    isManifestDeprecated: boolean
+    versions: {
+        serverVersion: string
+        isDelisted: boolean
+        targets: Target[]
+    }[]
+}
+
+export abstract class LSPDownloader {
```
justinmk3, line 45:
```suggestion
export abstract class LspDownloader {
```

packages/core/src/shared/index.ts
```diff
@@ -59,3 +59,4 @@ export { i18n } from './i18n-helper'
 export * from './icons'
 export * as textDocumentUtil from './utilities/textDocumentUtilities'
 export { TabTypeDataMap } from '../amazonq/webview/ui/tabs/constants'
+export { LSPDownloader, Manifest, Content } from './downloadLSP'
```
justinmk3, line 62: Unless I'm mistaken, the `* as foo` pattern is what we want to encourage. That allows us to use very few index.ts files which contain everything, without worrying about name conflicts. It's also more ergonomic for the consumer(?)
```suggestion
export * as fetchLsp from './downloadLSP'
```

packages/core/src/shared/downloadLSP.ts
```diff
@@ -0,0 +1,214 @@
...
+    }) {
+        const serverZipTempPath = path.join(tempFolder, `${name}.zip`)
+        const downloadOk = await this.downloadAndCheckHash(serverZipTempPath, content)
+        if (!downloadOk) {
+            return false
+        }
+
+        // load the zip contents
+        const extractPath = extractToTempFolder ? tempFolder : path.join(tempFolder, name)
+        new AdmZip(serverZipTempPath).extractAllTo(extractPath)
```
justinmk3, line 163: While we're here, we should try to migrate to the new `shared/utilities/zipStream.ts` if possible. AdmZip will be removed (not web-compatible; though may be irrelevant for this module(?) but we want to eliminate AdmZip totally to avoid confusion about when to use which)

justinmk3 (approved):

packages/core/src/shared/downloadLSP.ts
jpinkney-aws: :+1: i'll make a followup PR to change that

PR #6149, opened 2024-12-04
refactor(ec2): avoid logging somewhat sensitive data

Hweinstock:
## Problem
We currently log `STREAM_URL` and `TOKEN` which are used to establish the SSM session. However, these could be used to establish a connection outside the toolkit. 

## Solution
- Omit these from the logs. 
- Also add a logging statement to make it easier to find these `connect_script` logs. 

---

### DIFF
--- a/packages/core/resources/ec2_connect
+++ b/packages/core/resources/ec2_connect
@@ -28,7 +28,6 @@ _require_nolog() {
 
 _require() {
     _require_nolog "$@"
-    _log "$1=$2"
 }
 
 _ec2() {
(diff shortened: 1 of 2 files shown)


packages/core/resources/ec2_connect
```diff
@@ -28,7 +28,6 @@ _require_nolog() {
 
 _require() {
     _require_nolog "$@"
-    _log "$1=$2"
```
justinmk3, line 31: instead of this, could change the relevant `_require` calls below to use `_require_nolog`

justinmk3 (approved):

PR #6150, opened 2024-12-04
ci: codecov "patch" check

justinmk3:
# Problem 

- codecov is always skipped in github actions:
    - image
- We have a bot that attempts to remind contributors about adding tests, but it gets ignored.

# Solution 

- Fix the condition: https://github.com/aws/aws-toolkit-vscode/pull/6150/commits/cec8218a0bc7bfc8d0dfcc92a2764de18030e720
- Reintroduce the codecov "patch" check.

---

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -84,7 +84,7 @@ jobs:
               env:
                   # Unset NODE_OPTIONS because of https://github.com/codecov/uploader/issues/475
                   NODE_OPTIONS: ''
-              if: ${{ github.repository == 'aws/aws-toolkit-vscode' && github.ref == 'master' }}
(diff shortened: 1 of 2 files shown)


packages/core/package.json
```diff
@@ -523,7 +523,7 @@
         "aws-ssm-document-language-service": "^1.0.0",
         "bytes": "^3.1.2",
         "cross-fetch": "^4.0.0",
-        "cross-spawn": "^7.0.3",
```
Hweinstock, line 526: Did you mean to merge this change into here? Do we want this going to master?

justinmk3: thanks for the check; this was temporary, since i was trying to get codecov to actually run. but it turns out the problem is that it never runs in GHA...

PR #6151, opened 2024-12-04
feat(amazonq): Add acknowledgement button for disclaimer

jpinkney-aws:
## Problem
- disclaimer is too wordy

## Solution
- go back to the old disclaimer and add an ackowledgement button

---


packages/core/src/amazonq/webview/messages/messageDispatcher.ts
```diff
@@ -60,12 +60,19 @@ export function dispatchWebViewMessagesToApps(
...
                     telemetry.ui_click.emit({
                         elementId: msg.source,
                         result: 'Succeeded',
                     })
+                    return
                 }
+                return
+            }
+            case 'disclaimer-acknowledged': {
+                void globals.globalState.tryUpdate('aws.amazonq.disclaimerAcknowledged', true)
```
justinmk3, line 74: tryUpdate is synchronous
```suggestion
               globals.globalState.tryUpdate('aws.amazonq.disclaimerAcknowledged', true)
```

packages/core/src/amazonq/webview/ui/disclaimer/disclaimer.ts
justinmk3, line 1: does this need its own subfolder?

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -609,6 +613,24 @@ export const createMynahUI = (
...
+
+                // create telemetry
+                ideApi.postMessage(createClickTelemetry('amazonq-disclaimer-acknowledge-button'))
+
+                // remove all disclaimer cards from all tabs
+                Object.keys(mynahUI.getAllTabs()).forEach((storeTabKey) => {
+                    // eslint-disable-next-line unicorn/no-null
+                    mynahUI.updateStore(storeTabKey, { promptInputStickyCard: null })
+                })
+            }
```
justinmk3, line 632: should it end with `return` like the others or does it intentionally fall through (is that worth calling out in a comment?)

justinmk3 (approved):

dogusata (approved):

packages/core/src/amazonq/webview/ui/main.ts
dogusata: In general, a button can do more than one thing, but as it seems we don't have this kind of separation here. It can be converted to a switch statement.

mr-lee (approved):

justinmk3 (approved):

PR #6161, opened 2024-12-05
refactor(ec2): pass log level to connect script

Hweinstock:
## Problem
We want to only log file paths when in debug mode, but the connect script has no way of determining this. 

## Solution
- pass log level to connect script. 
- use it to determine what to log. 
- 
Example logs:
``` 
2024/12/05 15:50:24 ==============================================================================
2024/12/05 15:50:24 LOG_LEVEL=3
2024/12/05 15:50:24 AWS_REGION=us-east-1
2024/12/05 15:50:24 SESSION_ID=...
2024/12/05 15:51:11 ==============================================================================
2024/12/05 15:51:11 LOG_LEVEL=1
2024/12/05 15:51:11 AWS_REGION=us-east-1
2024/12/05 15:51:11 SESSION_ID=...
2024/12/05 15:51:11 AWS_SSM_CLI=.../Amazon/sessionmanagerplugin/bin/session-manager-plugin
2024/12/05 15:51:11 LOG_FILE_LOCATION=/.../ec2.{instanceId}.log
2024/12/05 15:51:55 ==============================================================================
2024/12/05 15:51:55 LOG_LEVEL=2
2024/12/05 15:51:55 AWS_REGION=us-east-1
2024/12/05 15:51:55 SESSION_ID=default-uk9cv4h86y5rdbrzarlkj9opci
2024/12/05 15:51:56 AWS_SSM_CLI=../Amazon/sessionmanagerplugin/bin/session-manager-plugin
2024/12/05 15:51:56 LOG_FILE_LOCATION=/.../ec2.{instanceId}.log
```

---

### DIFF
--- a/packages/core/src/awsService/ec2/utils.ts
+++ b/packages/core/src/awsService/ec2/utils.ts
@@ -8,6 +8,7 @@ import { copyToClipboard } from '../../shared/utilities/messages'
 import { Ec2Selection } from './prompter'
 import { sshLogFileLocation } from '../../shared/sshConfig'
 import { SSM } from 'aws-sdk'
+import { globals } from '../../shared'
 
 export function getIconCode(instance: SafeEc2Instance) {
     if (instance.LastSeenStatus === 'running') {
@@ -42,6 +43,7 @@ export function getEc2SsmEnv(
             STREAM_URL: session.StreamUrl,
             SESSION_ID: session.SessionId,
             TOKEN: session.TokenValue,
+            LOG_LEVEL: globals.logOutputChannel.logLevel,
         },
         process.env
     )
--- a/packages/core/resources/ec2_connect
+++ b/packages/core/resources/ec2_connect
@@ -2,7 +2,7 @@
 
 # Usage:
 #   When connecting to a dev environment
-#       AWS_REGION=… AWS_SSM_CLI=… STREAM_URL=… TOKEN=… LOG_FILE_LOCATION==… ./ec2_connect
+#       AWS_REGION=… AWS_SSM_CLI=… STREAM_URL=… TOKEN=… LOG_FILE_LOCATION==… LOG_LEVEL=… ./ec2_connect 
 
 set -e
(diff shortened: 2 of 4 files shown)


packages/core/src/awsService/ec2/utils.ts
```diff
@@ -42,6 +43,7 @@ export function getEc2SsmEnv(
             STREAM_URL: session.StreamUrl,
             SESSION_ID: session.SessionId,
             TOKEN: session.TokenValue,
+            LOG_LEVEL: globals.logOutputChannel.logLevel,
```
justinmk3, line 46: looks like getLogger doesn't have e.g. `getLogLevel`. otoh, we could make the flag a simple boolean like `LOG_DEBUG`:
```suggestion
            LOG_DEBUG: getLogger().logLevelEnabled('debug') ? '1': '0',
```

that also would avoid the use of `-le` in the bash script

Hweinstock: I don't think `logLevelEnabled` works as intended. I was able to produce this: 
``` 
2024-12-05 17:11:27.838 [info] vscode log level: info
2024-12-05 17:11:27.838 [info] is debug enabled?: true
``` 
and I believe its because the internal log level is `debug` until the user changes it based on the code here: 
https://github.com/aws/aws-toolkit-vscode/blob/aa332da854c9d17ba258b459c622500911b1b0d9/packages/core/src/shared/logger/activation.ts#L43-L49. 

Is there something special about the fact that I am running locally? Otherwise this seems like a bug.

justinmk3: I think when I added that "HACK" line, I assume that something in vscode would always overwrite the log-level, except when running in debug-mode. That line definitely needs some attention. 

Perhaps the easiest fix is to wrap that line in `if (!isRelease() && ...)` or whatever

justinmk3 (approved):

Issue #6094, opened 2024-11-22
arch(wizards): support async bindPrompter functions (ContextOptions)


justinmk3:
## Problem

1. [showWhen](https://github.com/aws/aws-toolkit-vscode/blob/9b44ba3d054f74e8ba9bafb8c9a26a2be4cdebec/packages/core/src/shared/wizards/wizardForm.ts#L28) clause does not support async function provider for operation like file lookup (skip prompter)
2. [setDefault](https://github.com/aws/aws-toolkit-vscode/blob/9b44ba3d054f74e8ba9bafb8c9a26a2be4cdebec/packages/core/src/shared/wizards/wizardForm.ts#L33) in both ContextOptions and FormElement support async function provider

Background: historically the `ContextOptions` functions were non-async because they are [called in constructors](https://github.com/aws/aws-toolkit-vscode-staging/blob/c01dcbf219ade0a5c2ae691f7b5f020a3c926ec8/packages/core/src/awsService/appBuilder/walkthrough.ts#L72-L90). But now that we have [async init()](https://github.com/aws/aws-toolkit-vscode-staging/blob/c01dcbf219ade0a5c2ae691f7b5f020a3c926ec8/packages/core/src/shared/wizards/wizard.ts#L152) ([example](https://github.com/aws/aws-toolkit-vscode-staging/blob/c01dcbf219ade0a5c2ae691f7b5f020a3c926ec8/packages/core/src/lambda/commands/uploadLambda.ts#L206-L207)) as an "async" way to construct Wizards, the door is open to support async variants of these wizard ContextOptions.

## Expected behavior

Support async variants of the [ContextOptions functions](https://github.com/aws/aws-toolkit-vscode-staging/blob/c01dcbf219ade0a5c2ae691f7b5f020a3c926ec8/packages/core/src/shared/wizards/wizardForm.ts#L20):

1. change their signatures to optionally return a `Promise`. example:
   ```
   showWhen?: (state: WizardState<TState>) => boolean | Promise<boolean>
   ```
2. in the Wizard / prompter base class(es), check `if (result instanceof Promise)` ([example](https://github.com/aws/aws-toolkit-vscode-staging/blob/c01dcbf219ade0a5c2ae691f7b5f020a3c926ec8/packages/core/src/shared/telemetry/spans.ts#L49)) to decide how to handle the async variants:
[...]

PR #6166, opened 2024-12-06
refactor(toolkit): add support for async showWhen and skipping prompter backward and forward direction

vicheey:
## Problem

Attempt to fix [#6094](https://github.com/aws/aws-toolkit-vscode/issues/6094) as the current implementation of wizard does not support:
- async provider for `showWhen` 
- restoring previous state of WizardPrompter in nested wizard when customer click BACK button
- skipping prompter in backward direction. 

## Solution
- implement a new logic in `NestedWizard` class that create wizard classes as prompters with support for instantiating and restoring child wizards in backward direction.
- update wizard state controller to support `ControlSingal.Skip` and add concept of wizard direction
- add support for async provider for `showWhen` clause

---


packages/core/src/shared/wizards/wizard.ts
```diff
@@ -269,9 +271,7 @@ export class Wizard<TState extends Partial<Record<keyof TState, unknown>>> {
 
         if (isValidResponse(answer)) {
             state.stepCache.picked = prompter.recentItem
-        }
-
-        if (!isValidResponse(answer)) {
```
seshubaws, line 274: Are we sure we can change this to an else? Is there a test to verify?

packages/core/src/test/shared/wizards/wizard.test.ts
```diff
@@ -211,11 +211,11 @@ describe('Wizard', function () {
 
     it('binds prompter to (sync AND async) property', async function () {
         wizard.form.prop1.bindPrompter(() => helloPrompter)
-        wizard.form.prop3.bindPrompter(async () => new SkipPrompter('helloooo (async)'))
+        wizard.form.prop3.bindPrompter(async () => new SkipPrompter())
 
         const result = await wizard.run()
         assert.strictEqual(result?.prop1, 'hello')
-        assert.strictEqual(result?.prop3, 'helloooo (async)')
+        assert(!result?.prop3)
```
seshubaws, line 218: Can we add a new test to verify the new functionality?

packages/core/src/shared/ui/nestedWizardPrompter.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export abstract class NestedWizard<T> extends Wizard<T> {
+    // Map to store wizard instances
+    private wizardInstances: Map<string, any> = new Map()
+
+    protected constructor(options: WizardOptions<T>) {
+        super(options)
+    }
+
+    protected createWizardPrompter<T>(constructor: new (...args: any[]) => T, ...args: any[]): WizardPrompter<T> {
```
seshubaws, line 18: Can we type the createWizardPrompter stronger? Maybe by allowing it to accept only classes that extend Wizard?

packages/core/src/shared/wizards/wizard.ts
vicheey: This an obvious binary case to me.

packages/core/src/shared/ui/wizardPrompter.ts
```diff
@@ -3,19 +3,22 @@
...
+import { StepEstimator, Wizard, WIZARD_BACK, WIZARD_SKIP } from '../wizards/wizard'
 import { Prompter, PromptResult } from './prompter'
 
 /**
  * Wraps {@link Wizard} object into its own {@link Prompter}, allowing wizards to use other
  * wizards in their flows.
  */
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const WIZARD_PROMPTER = 'WIZARD_PROMPTER'
```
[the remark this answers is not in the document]
vicheey: I don't think so. This `WIZARD_PROMPTER` is specific to `WizardPrompter` class and has no relation or usage in `Wizard` class.

packages/core/src/shared/ui/wizardPrompter.ts
```diff
@@ -49,8 +52,18 @@ export class WizardPrompter<T> extends Prompter<T> {
         }
     }
 
+    // eslint-disable-next-line @typescript-eslint/naming-convention
```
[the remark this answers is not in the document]
vicheey: Yes.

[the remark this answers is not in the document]
seshubaws: Why? I don't see any var names in the next line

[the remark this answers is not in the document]
vicheey: Mistake with the line for WIZARD_PROMPTER. I removed the line

packages/core/src/shared/ui/nestedWizardPrompter.ts
vicheey: Added the type enforcement.

packages/core/src/test/shared/wizards/wizard.test.ts
vicheey: Added test for NestedWizard.

packages/core/src/shared/ui/nestedWizardPrompter.ts
```diff
@@ -0,0 +1,27 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Wizard, WizardOptions } from '../wizards/wizard'
+import { WizardPrompter } from './wizardPrompter'
+import { createHash } from 'crypto'
+
```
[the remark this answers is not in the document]
vicheey: Added the docstring to NestedWizard class.

packages/core/src/shared/ui/wizardPrompter.ts
```diff
@@ -3,19 +3,27 @@
...
+import _ from 'lodash'
+import { StepEstimator, Wizard, WIZARD_BACK, WIZARD_SKIP } from '../wizards/wizard'
 import { Prompter, PromptResult } from './prompter'
 
 /**
  * Wraps {@link Wizard} object into its own {@link Prompter}, allowing wizards to use other
  * wizards in their flows.
+ *
+ * @remarks
+ *  - This class should not use direclty inside a parent class.
```
seshubaws, line 15: nit: should be "This class should not be used directly inside a parent class."

seshubaws (approved):

packages/core/src/shared/ui/common/skipPrompter.ts
```diff
@@ -3,24 +3,17 @@
...
         super()
     }
 
     protected async promptUser(): Promise<PromptResult<T>> {
-        const promptPromise = new Promise<PromptResult<T>>((resolve) => {
-            resolve(this.val)
-        })
-
-        return await promptPromise
+        return WIZARD_SKIP
```
justinmk3, line 16: great idea

packages/core/src/shared/ui/common/skipPrompter.ts
```diff
@@ -3,24 +3,17 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { StepEstimator } from '../../wizards/wizard'
+import { StepEstimator, WIZARD_SKIP } from '../../wizards/wizard'
 import { Prompter, PromptResult } from '../prompter'
 
-/** Pseudo-prompter that immediately returns a value (and thus "skips" a step). */
+/** Prompter that return SKIP control signal to parent wizard */
```
justinmk3, line 9:
```suggestion
/** Prompter that returns SKIP control signal to parent wizard */
```

packages/core/src/shared/ui/nestedWizardPrompter.ts
```diff
@@ -0,0 +1,63 @@
...
+     * - Instances are memoized using a SHA-256 hash of the wizard class name and arguments
+     * - The same wizard instance is reused for identical constructor parameters for restoring wizard prompter
+     *   states during back button click event
+     */
+    protected createWizardPrompter<TWizard extends Wizard<TState>, TState>(
+        wizardClass: new (...args: any[]) => TWizard,
+        ...args: ConstructorParameters<new (...args: any[]) => TWizard>
+    ): Prompter<TState> {
+        const memoizeKey = createHash('sha256')
+            .update(wizardClass.name + JSON.stringify(args))
```
justinmk3, line 54: I think `.name` will be mangled by webpack in the production build. But as long as it's unique, that should be ok for this purpose.

packages/core/src/shared/ui/nestedWizardPrompter.ts
```diff
@@ -0,0 +1,63 @@
...
+            .digest('hex')
+
+        if (!this.wizardInstances.get(memoizeKey)) {
+            this.wizardInstances.set(memoizeKey, new wizardClass(...args))
```
justinmk3, line 58: since wizardClass already gets the args, would it make sense for it to define a `getId()` or `getHash()` that provides its hash? Instead of managing that here.

It seems that javascript doesn't have Java's `hashCode()` and `equals()` conventions, but we can mimic it I guess.

packages/core/src/shared/ui/wizardPrompter.ts
```diff
@@ -3,19 +3,29 @@
...
+ *
+ * @remarks
+ *  - This class should only be used inside wizard classes that extend {@link NestedWizard}.
+ *  - See examples:
+ *     - {@link SingleNestedWizard}
+ *     - {@link DoubleNestedWizard}
  */
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const WIZARD_PROMPTER = 'WIZARD_PROMPTER'
```
vicheey, line 22: This is left over unused code from my first draft. I'll remove this.

packages/core/src/shared/ui/nestedWizardPrompter.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * An abstract class that extends the base Wizard class plus the ability to
+ * use other wizard classes as prompters
+ */
+export abstract class NestedWizard<T> extends Wizard<T> {
+    /**
+     * Map to store memoized wizard instances using SHA-256 hashed keys
+     */
+    private wizardInstances: Map<string, any> = new Map()
```
justinmk3, line 19: Usually not necessary to mention "instances" in variable names. Is there something more meaningful, maybe `childWizards` (I'm not sure if that's correct)?

```suggestion
    private childWizards: Map<string, any> = new Map()
```

packages/core/src/shared/ui/nestedWizardPrompter.ts
```diff
@@ -0,0 +1,63 @@
...
+import { Wizard, WizardOptions } from '../wizards/wizard'
+import { Prompter } from './prompter'
+import { WizardPrompter } from './wizardPrompter'
+import { createHash } from 'crypto'
+
+/**
+ * An abstract class that extends the base Wizard class plus the ability to
+ * use other wizard classes as prompters
+ */
+export abstract class NestedWizard<T> extends Wizard<T> {
```
justinmk3, line 15: Do we actually need this class or coudl we just add `childWizards` (or whatever) to the main `Wizard` base class?

packages/core/src/shared/ui/nestedWizardPrompter.ts
vicheey: Do you suggest moving the hashing logic from `NestedWizard` and move that to Wizard class as static function? 
I say static because we might want to check hash before instantiating an object. 

But this would not be that much different from having the hashing logic here in `NestedWizard`

packages/core/src/shared/ui/wizardPrompter.ts
```diff
@@ -3,19 +3,29 @@
...
+import _ from 'lodash'
+import { StepEstimator, Wizard, WIZARD_BACK, WIZARD_SKIP } from '../wizards/wizard'
 import { Prompter, PromptResult } from './prompter'
 
 /**
  * Wraps {@link Wizard} object into its own {@link Prompter}, allowing wizards to use other
  * wizards in their flows.
+ *
+ * @remarks
+ *  - This class should only be used inside wizard classes that extend {@link NestedWizard}.
```
justinmk3, line 15: Oh, this may be a good reason to keep the NestedWizard class: it makes this constraint easier to explain. Though I wonder why this constraint was added?

packages/core/src/shared/ui/nestedWizardPrompter.ts
vicheey: I think we can move all the mapping and hashing logic from this class to the `Wizard` class. 

I just thought that having an abstract class NestedWizard is a bit more trtansparent give the fact that the wizard class is very complex already.

packages/core/src/shared/ui/wizardPrompter.ts
```diff
@@ -51,6 +61,15 @@ export class WizardPrompter<T> extends Prompter<T> {
 
     protected async promptUser(): Promise<PromptResult<T>> {
         this.response = await this.wizard.run()
-        return this.response
+
+        if (this.response === undefined) {
+            return WIZARD_BACK as PromptResult<T>
+        } else if (_.isEmpty(this.response)) {
```
justinmk3, line 67: is lodash actually needed here? we are trying to eliminate it

packages/core/src/shared/wizards/stateController.ts
```diff
@@ -9,7 +9,16 @@ export enum ControlSignal {
...
     Back,
-    Continue,
+    Skip,
+}
+
+/**
+ * Value for indicating current direction of the wizard
+ * Created mainly to support skipping prompters
+ */
+export enum DIRECTION {
```
justinmk3, line 19:
```suggestion
export enum Direction {
```

packages/core/src/shared/ui/wizardPrompter.ts
vicheey: Because nested class have the memorizing logic for restoring child wizard state.

packages/core/src/shared/ui/wizardPrompter.ts
vicheey: This is not a necessity.

justinmk3 (approved):
> * implement a new NestedWizard  abstract class that contain logic to instantiate and restore child Wizard used as prompter
> * update wizard state controller to support `ControlSingal.Skip` and add concept of wizard direction for tracking

Nice approach, and explanation! And clear test code.

packages/core/src/shared/ui/nestedWizardPrompter.ts
justinmk3: > say static because we might want to check hash before instantiating an object.

Oh, I didn't think about that. But it still seems to make sense for the Wizard to define its own hash. After all, this may be used for other purposes, but currently it's private to this one consumer.

packages/core/src/shared/ui/wizardPrompter.ts
justinmk3: Maybe the wording of the comment is unclear to me. Does "This class" refer to WizardPrompter? For example, currently WizardPrompter is used here (which is not "inside NestedWizard"): https://github.com/aws/aws-toolkit-vscode/blob/e63079d3d3af18a3af27914e842d915f307b38b3/packages/core/src/awsService/apprunner/wizards/imageRepositoryWizard.ts#L237

packages/core/src/shared/ui/common/skipPrompter.ts
```diff
@@ -3,24 +3,17 @@
...
 export class SkipPrompter<T> extends Prompter<T> {
-    /**
-     * @param val Value immediately returned by the prompter.
-     */
-    constructor(public readonly val: T) {
```
roger-zhangg, line 14: The previous implementation supports resolving the value it was assigned in the beginning. but this is removed in the new solution. Is this intended?

packages/core/src/shared/wizards/stateController.ts
vicheey: Updated.

packages/core/src/shared/ui/wizardPrompter.ts
vicheey: Updated.

packages/core/src/shared/ui/wizardPrompter.ts
vicheey: Updated the doc string.

packages/core/src/shared/ui/nestedWizardPrompter.ts
vicheey: Moved logic from `NestedWizard` class to `Wizard` and remove the former.

packages/core/src/shared/ui/nestedWizardPrompter.ts
vicheey: Updated.

packages/core/src/shared/ui/nestedWizardPrompter.ts
vicheey: Remove this class after moving logic here to `Wizard` class.

packages/core/src/shared/wizards/stateController.ts
```diff
@@ -105,6 +118,12 @@ export class StateMachineController<TState> {
         }
 
         if (isMachineResult(result)) {
+            if (result.controlSignal === ControlSignal.Skip) {
+                /**
+                 * Depending on current wizard direction, skip signal get converted to forward or backward control signal
+                 */
+                result.controlSignal = this.direction === DIRECTION.Forward ? undefined : ControlSignal.Back
```
roger-zhangg: How would this work on Wizard prompter? Will it able to identify the direction of it's parent wizard?

roger-zhangg (commented):
Thanks for this super helpful feature, few questions

packages/core/src/shared/ui/common/skipPrompter.ts
roger-zhangg: A straight forward case would be in `bindPrompter`, we have multiple if/else based on previous states, and maybe there are two case leads to skip prompter, but we want different value for this field even though they are skipped.

packages/core/src/shared/wizards/stateController.ts
vicheey: The `WizardPrompter` is still considered a prompter. It is not responsible for identifying parent wizard direction. 

This state management functionality belongs to the `stateController` object within the Wizard class. Each Wizard is responsible for prompting, receiving, and processing answers from prompters (including `WizardPrompter`). Each answer contains data and a `ControlSignal` (undefined indicates no direction change). The Wizard class's `stateController` tracks the current direction and uses the `ControlSignal` from the prompter to manage direction changes.

How do we restore the state of `WizardPrompter` when a back button is clicked?
Well, we do not restore the state of the `WizardPrompter` object, we always instantiate a new instance of this class. We, however, restore the state of Wizard class used for instantiate the `WizardPrompter`.  Again, the (parent) wizard class is responsible for this [(reference)](https://github.com/aws/aws-toolkit-vscode/blob/83cc0ff4656ce0130b41056c81dd5c9c8834a36a/packages/core/src/shared/wizards/wizard.ts#L108-L145)

packages/core/src/shared/ui/common/skipPrompter.ts
vicheey: The previous workaround for skipping prompter was to automatically resolve the value and skip displaying UI to customer. However, that did not work for backward direction since the auto resolve always revert the backward flow to forward flow.

So we are splitting the skipping case from setting default value case. For how to use the SkipPrompter, please follow this example:  [aws-toolkit-vscode/packages/core/src/dev/activation.ts at master · vicheey/aws-toolkit-vscode](https://github.com/vicheey/aws-toolkit-vscode/blob/master/packages/core/src/dev/activation.ts#L414)

packages/core/src/shared/ui/nestedWizardPrompter.ts
vicheey: > Moved logic from NestedWizard class to Wizard and remove the former.

Actually, in doing so, we create a circular dependency between `wizardPrompter`, `wizard`, `wizardPrompter`. 
```
../core/dist/src/shared/ui/wizardPrompter.js -> ../core/dist/src/shared/wizards/wizard.js -> ../core/dist/src/shared/ui/wizardPrompter.js
```
Reverted the change.

packages/core/src/shared/ui/common/skipPrompter.ts
roger-zhangg: You can consider resolving the value when forward and skip when backwards right?

justinmk3 (approved):

packages/core/src/shared/ui/nestedWizardPrompter.ts
justinmk3, line 15: This name seems confusing, unless I'm mistaken. This class is a Wizard that contains nested wizards, it's not itself nested. Right? So maybe a better name is `Wizards` (plural). That also improves discoverability because it sorts alphabetically next to `Wizard`.

justinmk3:
Merged to unblock. Nicely done, thanks! Some followup requests:

- could you add 1 or 2 sentences to [this doc](https://github.com/aws/aws-toolkit-vscode/blob/85d5a309a7d3402ff6c08877faab0adfca7893fa/docs/arch_develop.md?plain=1#L443) mentioning when to use NestedWizard?
- the name NestedWizard seems misleading, see https://github.com/aws/aws-toolkit-vscode/pull/6166#discussion_r1882956699

PR #6172, opened 2024-12-06
fix(appbuilder): Prevent parallel build processes on the same template

mbfreder:
## Context
We currently allow multiple build processes to run at the same time for the same template.

## Problem

When running a build, it’s possible to start a second build for the same project. This results in both builds failings

## Solution
Prevent parallel builds on the same template

---


packages/core/src/shared/sam/utils.ts
```diff
@@ -108,6 +108,35 @@ export async function updateRecentResponse(
         getLogger().warn(`sam: unable to save response at key "${key}": %s`, err)
     }
 }
+
+const buildProcessMementoRootKey = 'samcli.build.processes'
```
seshubaws, line 112: Let's change this to something like

const BUILD_PROCESS_MEMENTO_ROOT_KEY = 'samcli.build.processes'; and maybe put it in the samutils if it makes sense to. We can also do the same with the 'global' as we should limit the amount of constant strings we have

packages/core/src/shared/sam/utils.ts
```diff
@@ -108,6 +108,35 @@ export async function updateRecentResponse(
...
+ * @Param templatePath The path to the template.yaml file
+ */
+export function throwIfTemplateIsBeingBuilt(templatePath: string) {
+    if (isBuildInProgress(templatePath)) {
+        throw new ToolkitError('Template is already being built', { code: 'BuildInProgress' })
+    }
+}
+
+export async function registerTemplateBuild(templatePath: string) {
+    await updateRecentResponse(buildProcessMementoRootKey, 'global', templatePath, 'true')
```
seshubaws, line 133: Can we add error handling here to catch if something goes wrong in the process of registering a template build? I think currently, the error would just bubble up from the actual sam build command, but it could be the case that build command succeeds but the registering fails

packages/core/src/shared/sam/utils.ts
```diff
@@ -108,6 +108,35 @@ export async function updateRecentResponse(
...
+    return getRecentResponse(buildProcessMementoRootKey, 'global', templatePath) !== undefined
+}
+
+/**
+ * Throws an error if there's a build in progress for the provided template
+ * @Param templatePath The path to the template.yaml file
+ */
+export function throwIfTemplateIsBeingBuilt(templatePath: string) {
+    if (isBuildInProgress(templatePath)) {
+        throw new ToolkitError('Template is already being built', { code: 'BuildInProgress' })
```
seshubaws, line 128: Can we make the error code an enum? I think we had a class with different error codes already right, let's add this one in with them

seshubaws (changes requested):

packages/core/src/shared/sam/utils.ts
mbfreder: The above format wouldn't work since eslint only allows camelCase or PascalCase formats and the constant is defined in the sam/utils file. Are you referring to a different utils file?

packages/core/src/shared/sam/utils.ts
mbfreder: updateRecentResponse() already handles that by logging the error. 

```typescript
export async function updateRecentResponse(
    mementoRootKey: string,
    identifier: string,
    key: string,
    value: string | undefined
) {
    try {
        const root = globals.context.workspaceState.get(mementoRootKey, {} as Record<string, Record<string, string>>)
        await globals.context.workspaceState.update(mementoRootKey, {
            ...root,
            [identifier]: { ...root[identifier], [key]: value },
        })
    } catch (err) {
        getLogger().warn(`sam: unable to save response at key "${key}": %s`, err)
    }
}
```

packages/core/src/shared/sam/utils.ts
mbfreder: Those are specific SAM CLI error patterns. I don't think it makes sense to add this one to them since this error doesn't come directly from SAM CLI.

packages/core/src/shared/sam/utils.ts
seshubaws: Ideally, we'd put this in a constants.ts file, but if not we can disable eslint for this line. The convention is that constants should be in all caps.

packages/core/src/shared/sam/utils.ts
seshubaws: Can we capture the error so that we can see it in our Kibana dashboard? I don't think we see logger errors in our dashboard iirc

packages/core/src/shared/sam/constants.ts
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable header/header */
+/* eslint-disable @typescript-eslint/naming-convention */
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const BUILD_PROCESS_MEMENTO_ROOT_KEY = 'samcli.build.processes'
```
jpinkney-aws, line 8: I think the standard we have in the toolkit is to use camal case, even for constants

packages/core/src/shared/sam/constants.ts
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable header/header */
+/* eslint-disable @typescript-eslint/naming-convention */
```
jpinkney-aws, line 2: you shouldn't need these es-lint disables

packages/core/src/shared/sam/utils.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable header/header */
```
jpinkney-aws, line 1: you shouldn't need these es-lint disables

packages/core/src/shared/sam/utils.ts
```diff
@@ -108,6 +109,33 @@ export async function updateRecentResponse(
         getLogger().warn(`sam: unable to save response at key "${key}": %s`, err)
     }
 }
+
+/**
+ * Returns true if there's an ongoing build process for the provided template, false otherwise
+ * @Param templatePath The path to the template.yaml file
+ */
+function isBuildInProgress(templatePath: string): boolean {
+    return getRecentResponse(buildProcessMementoRootKey, globalIdentifier, templatePath) !== undefined
```
justinmk3, line 118: Future note: these functions (and probably more, I guess) are using shared mutable state. You may want to consider wrapping this in a dedicated file similar to the existing `packages/core/src/shared/globalState.ts` , except for "workspace" state. That makes it clear when callers are using this global state. And it collects the pattern in one place, instead of many spread-out functions.

packages/core/src/shared/sam/utils.ts
```diff
@@ -108,6 +109,33 @@ export async function updateRecentResponse(
...
+}
+
+/**
+ * Throws an error if there's a build in progress for the provided template
+ * @Param templatePath The path to the template.yaml file
+ */
+export function throwIfTemplateIsBeingBuilt(templatePath: string) {
+    if (isBuildInProgress(templatePath)) {
+        throw new ToolkitError('This template is already being built', { code: 'BuildInProgress' })
```
roger-zhangg, line 127: I would say this check is too strict. There are possibility that the build is finished or exited but we also failed to `unregisterTemplateBuild` or even toolkit maybe reload itself during the build process. In this case the customer will never able to build the template again. One possible solution is to add a timeout to this cache. Like 2 minutes or 5 minutes. In this way, if the worst case happens ( build finished but we failed to update memeto) the customer will still be able to build after time out is reached.

packages/core/src/shared/sam/build.ts
```diff
@@ -229,23 +235,35 @@ export async function runBuild(arg?: TreeNode): Promise<SamBuildResult> {
     await updateRecentResponse(buildMementoRootKey, 'global', 'templatePath', templatePath)
 
     try {
-        const { path: samCliPath } = await getSamCliPathAndVersion()
+        await vscode.window.withProgress(
+            {
+                location: vscode.ProgressLocation.Notification,
+            },
+            async (progress) => {
+                progress.report({ message: `Building SAM template at ${params.template.uri.path}` })
```
roger-zhangg, line 243: can we also add a cancel button to it? you can refer the implementation here: 
https://github.com/aws/aws-toolkit-vscode/blob/5bfb867831cb9e8c475de5fb062607f37aa2991b/packages/core/src/shared/utilities/cliUtils.ts#L235-L238

packages/core/src/test/shared/sam/build.test.ts
```diff
@@ -551,3 +564,23 @@ describe('SAM runBuild', () => {
         })
     })
 })
+
+async function runInParallel(samLocation: SamAppLocation): Promise<[SamBuildResult, SamBuildResult]> {
+    return Promise.all([runBuild(new AppNode(samLocation)), runBuild(new AppNode(samLocation))])
```
roger-zhangg, line 569: I worry this would cause some flakiness as the two build is run at same time. There might be a race condition that both Build run `throwIfTemplateIsBeingBuilt` both passes and then both run `registerTemplateBuild` and proceed to build. In this case the two build will be run together and cause test to fail. I would suggest we add a 0.05s wait time between them to avoid the unlikely but possible race condition.

roger-zhangg (changes requested):
Left few comments, thanks for implementing these changes!

packages/core/src/shared/sam/build.ts
mbfreder: Thanks. Added.

packages/core/src/shared/sam/utils.ts
mbfreder: Good catch. I added an expirationDate to the template registration.

roger-zhangg (approved):
Thanks for addressing all the feedback, LGTM

packages/core/src/shared/sam/build.ts
```diff
@@ -229,23 +235,36 @@ export async function runBuild(arg?: TreeNode): Promise<SamBuildResult> {
...
 
-        // Run SAM build in Terminal
-        await runInTerminal(buildProcess, 'build')
+        await unregisterTemplateBuild(params.template.uri.path)
 
         return {
             isSuccess: true,
         }
     } catch (error) {
+        await unregisterTemplateBuild(params.template.uri.path)
```
justinmk3, line 267: why not do it once in `finally` ?

packages/core/src/shared/sam/utils.ts
```diff
@@ -108,6 +109,38 @@ export async function updateRecentResponse(
         getLogger().warn(`sam: unable to save response at key "${key}": %s`, err)
     }
 }
+
+/**
+ * Returns true if there's an ongoing build process for the provided template, false otherwise
+ * @Param templatePath The path to the template.yaml file
+ */
+function isBuildInProgress(templatePath: string): boolean {
```
justinmk3, line 117: seems like these new functions would fit in  `shared/sam/build.ts` ?

packages/core/src/shared/sam/utils.ts
```diff
@@ -108,6 +109,38 @@ export async function updateRecentResponse(
...
+ */
+export function throwIfTemplateIsBeingBuilt(templatePath: string) {
+    if (isBuildInProgress(templatePath)) {
+        throw new ToolkitError('This template is already being built', { code: 'BuildInProgress' })
+    }
+}
+
+export async function registerTemplateBuild(templatePath: string) {
+    const expirationDate = Date.now() + 5 * 60 * 1000 // five minutes
+    await updateRecentResponse(buildProcessMementoRootKey, globalIdentifier, templatePath, expirationDate.toString())
```
justinmk3, line 137: What happens if a build takes longer than 5 minutes? Customer feedback shows that builds can be very, very slow. If the user Cancels the build is that correctly handled?

packages/core/src/shared/sam/utils.ts
```diff
@@ -108,6 +109,38 @@ export async function updateRecentResponse(
...
+ * Throws an error if there's a build in progress for the provided template
+ * @Param templatePath The path to the template.yaml file
+ */
+export function throwIfTemplateIsBeingBuilt(templatePath: string) {
+    if (isBuildInProgress(templatePath)) {
+        throw new ToolkitError('This template is already being built', { code: 'BuildInProgress' })
+    }
+}
+
+export async function registerTemplateBuild(templatePath: string) {
```
justinmk3, line 135: "registering" is usually a sign of a design flaw. registries require lifecycle management, and can get "corrupted" by logic bugs (like any "cache" in general).

Is there instead a way we can check SAM CLI's state (e.g. in the `.sam-...` directory) to see if it's currently building something?

packages/core/src/shared/sam/utils.ts
mbfreder: Thanks. Updated.

packages/core/src/shared/sam/utils.ts
mbfreder: We currently can't check sam cli state in the toolkit. I think this is the best we can do right now, until we find a solution in SAM CLI.

packages/core/src/shared/sam/utils.ts
mbfreder: If the expiration date is too long, it makes it hard for users to quickly do subsequent builds.

packages/core/src/shared/sam/build.ts
```diff
@@ -233,16 +237,28 @@ export async function runBuild(arg?: TreeNode): Promise<SamBuildResult> {
...
-                env: await getSpawnEnv(process.env),
-            }),
-        })
+        await vscode.window.withProgress(
+            {
+                cancellable: true,
+                location: vscode.ProgressLocation.Notification,
+            },
+            async (progress) => {
+                progress.report({ message: `Building SAM template at ${params.template.uri.path}` })
```
jpinkney-aws, line 246: if you make this cancellable I think you should also be adding the additional token argument.

e.g.
```
async (progress, token) => {
    token.onCancellationRequested(async () => {
         // do something
    })
}
```

packages/core/src/shared/sam/build.ts
```diff
@@ -293,3 +311,34 @@ export async function resolveBuildFlags(buildFlags: string[], samCliVersion: Sem
...
+        return Date.now() < parseInt(expirationDate)
+    }
+    return false
+}
+
+/**
+ * Throws an error if there's a build in progress for the provided template
+ * @Param templatePath The path to the template.yaml file
+ */
+export function throwIfTemplateIsBeingBuilt(templatePath: string) {
```
jpinkney-aws, line 331: for the following functions does burying them in another function give us anything of value? If not they can live in the top level

packages/core/src/shared/sam/build.ts
mbfreder: Updated

packages/core/src/shared/sam/build.ts
mbfreder: I extracted those functions because the runBuild() function was getting too big. I think having them outside makes runBuild() easier to read.

jpinkney-aws:
This doesn't impact anything Q related, only toolkit so I merged it in for this release

PR #6174, opened 2024-12-06
telemetry(amazonq): UTG sends plaintext instead of python for amazonq_utgGenerateTests event

laileni-aws:
## Problem
- UTG sends plaintext instead of python for amazonq_utgGenerateTests event.
- In code, `session.fileLanguage = await this.getLanguageForFilePath(filePath)` reassigned python to plaintext after test generation.

## Solution
- Removal of `session.fileLanguage = await this.getLanguageForFilePath(filePath)` should fix the issue.

---

### DIFF
--- a/packages/core/src/amazonqTest/chat/controller/controller.ts
+++ b/packages/core/src/amazonqTest/chat/controller/controller.ts
@@ -444,9 +444,8 @@ export class TestController {
 
             /*
                 For Re:Invent 2024 we are supporting only java and python for unit test generation, rest of the languages shows the similar experience as CWC
-                If user request test generation from input chat without opening a file, its difficult to get the language, so default will be plainText
             */
-            if (language !== 'java' && language !== 'python' && language !== 'plaintext') {
+            if (language !== 'java' && language !== 'python') {
                 const unsupportedLanguage = language.charAt(0).toUpperCase() + language.slice(1)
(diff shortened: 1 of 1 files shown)


packages/core/src/amazonqTest/chat/controller/controller.ts
```diff
@@ -724,7 +722,7 @@ export class TestController {
             acceptedCharactersCount: session.charsOfCodeAccepted,
             generatedLinesCount: session.linesOfCodeGenerated,
             acceptedLinesCount: session.linesOfCodeAccepted,
-            cwsprChatProgrammingLanguage: session.fileLanguage,
+            cwsprChatProgrammingLanguage: session.fileLanguage ?? 'plaintext',
```
hayemaxi, line 725: Is this the proper place to put this default? Will other functions also default to plaintext? If so, initialize in a common place.

hayemaxi (approved):

ashishrp-aws (approved):

ctlai95 (approved):

packages/core/src/amazonqTest/chat/controller/controller.ts
laileni-aws: @chungjac is working on this.
We aim to declare the telemetry function in telemetryHelper.ts and use it here but for now fixing this here.

PR #6181, opened 2024-12-07
telemetry(amazonq): Adding requestId for amazonq_utgGenerateTests event

laileni-aws:
## Problem
- Missing data if job fails at `StartTestGeneration` API.

## Solution
- Adding requestId for `amazonq_utgGenerateTests` event, which helps ti debug the reason for job failures from the backend.

---

### DIFF
--- a/packages/core/src/codewhisperer/service/testGenHandler.ts
+++ b/packages/core/src/codewhisperer/service/testGenHandler.ts
@@ -94,13 +94,13 @@ export async function createTestJob(
     logger.debug('target line range end: %O', firstTargetLineRangeList?.end)
 
     const resp = await codewhispererClient.codeWhispererClient.startTestGeneration(req).catch((err) => {
-        ChatSessionManager.Instance.getSession().requestId = err.requestId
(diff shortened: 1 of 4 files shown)


packages/core/src/codewhisperer/service/testGenHandler.ts
```diff
@@ -94,11 +94,13 @@ export async function createTestJob(
     logger.debug('target line range end: %O', firstTargetLineRangeList?.end)
 
     const resp = await codewhispererClient.codeWhispererClient.startTestGeneration(req).catch((err) => {
+        ChatSessionManager.Instance.getSession().startTestGenerationRequestId = err.requestId
```
justinmk3, line 97: so `ChatSessionManager.Instance.getSession()` is global, shared, mutable state?

laileni-aws: Yes, its global and mutable.

justinmk3: It's clear that will be a source of bugs, right?

PR #6182, opened 2024-12-07
fix(amazonq): always use projectName in zipEntryPath

ctlai95:
## Problem

Projects with a nested folder of the same name does not scan properly.

For example if the project name is `foo`, and there exists a folder `foo/foo/` then the zip artifact only contains the inner `foo/` folder.

## Solution

Always include the project name in the zip artifact instead of conditionally adding it.

---

### DIFF
--- a/packages/amazonq/test/unit/codewhisperer/util/zipUtil.test.ts
+++ b/packages/amazonq/test/unit/codewhisperer/util/zipUtil.test.ts
@@ -114,7 +114,7 @@ describe('zipUtil', function () {
             assert.equal(zipMetadata2.lines, zipMetadata.lines + 1)
         })
 
-        it('should handle path with repeated project name', async function () {
+        it('should handle path with repeated project name for file scan', async function () {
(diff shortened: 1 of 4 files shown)


laileni-aws (approved):

jpinkney-aws (commented):
Can we add a couple tests

packages/amazonq/test/unit/codewhisperer/util/zipUtil.test.ts
```diff
@@ -112,6 +113,32 @@ describe('zipUtil', function () {
             )
             assert.equal(zipMetadata2.lines, zipMetadata.lines + 1)
         })
+
+        it('should handle path with repeated project name for file scan', async function () {
+            const appCodePathWithRepeatedProjectName = join(workspaceFolder, 'workspaceFolder', 'App.java')
```
jpinkney-aws, line 118: Can you pull this out of this test, it looks like you're using it 4 times

PR #6197, opened 2024-12-10
fix(ec2): avoid wiping `authorized_keys` files on each connection

Hweinstock:
## Problem
Each EC2 remote vscode connection wipes the remote `.ssh/authorized_keys` file as a preventative measure to leaving stale keys there. However, we can do better by adding comments to the keys we add to this file, then selectively removing those keys on subsequent connections. 

## Solution
- Whenever we send keys to the instance, use `sed` to wipe all of the keys added by us. 
- determine keys added by us using a hint comment `#AWSToolkitForVSCode`. 

---


docs/arch_features.md
```diff
@@ -43,7 +43,6 @@ For EC2 specifically, there are a few additional steps:
 
 1. If connecting to EC2 instance via remote window, the toolkit generates temporary SSH keys (30 second lifetime), with the public key sent to the remote instance.
     - Key type is ed25519 if supported, or RSA otherwise.
-    - This connection will overwrite the `.ssh/authorized_keys` file on the remote machine with each connection.
```
justinmk3, line 46: Might as well mention what we do instead:

    - Lines in `.ssh/authorized_keys` marked with the comment `#AWSToolkitForVSCode` will be removed by AWS Toolkit.

packages/core/src/awsService/ec2/model.ts
```diff
@@ -271,38 +286,74 @@ export class Ec2Connecter implements vscode.Disposable {
...
     }
 
-    public async configureSshKeys(selection: Ec2Selection, remoteUser: string): Promise<SshKeyPair> {
+    public async configureSshKeys(selection: Ec2Selection, remoteUser: RemoteUser): Promise<SshKeyPair> {
         const keyPair = await SshKeyPair.getSshKeyPair(`aws-ec2-key`, 30000)
         await this.sendSshKeyToInstance(selection, keyPair, remoteUser)
         return keyPair
     }
 
+    public async attemptToCleanKeys(
```
justinmk3, line 295: "try" is a common convention that fits here. the body of the function isn't obvious, so worth a docstring too.

```suggestion
    /** Removes public key(s) that we added to the remote ~/.ssh/authorized_keys file. */
    public async tryCleanKeys(
```

docs/arch_features.md
```diff
@@ -43,7 +43,6 @@ For EC2 specifically, there are a few additional steps:
 
 1. If connecting to EC2 instance via remote window, the toolkit generates temporary SSH keys (30 second lifetime), with the public key sent to the remote instance.
```
justinmk3, line 44: Should we also mention here the expected remote user name/home dir? That is a detail that may be useful for troubleshooting.

packages/core/src/awsService/ec2/model.ts
```diff
@@ -271,38 +286,74 @@ export class Ec2Connecter implements vscode.Disposable {
...
+
+/**
+ * Generate bash command (as string) to remove lines containing `pattern`.
+ * @param pattern pattern for deleted lines.
+ * @param filepath filepath (as string) to target with the command.
+ * @returns bash command to remove lines from file.
+ */
+export function getRemoveLinesCommand(pattern: string, hostOS: Ec2OS, filepath: string): string {
+    // Linux allows not passing extension to -i, whereas macOS requires zero length extension.
+    return `sed -i${isLinux(hostOS) ? '' : " ''"} /${pattern}/d ${filepath}`
```
justinmk3, line 354: if `pattern` contains a slash "/" char this will fail. but it looks like we don't need to worry about that. maybe we should fail just in case:

```ts
if (pattern.includes('/')) {
  throw new Error()
}

File: packages/core/src/awsService/ec2/model.ts [hunk 5]
```diff
@@ -271,38 +286,74 @@ export class Ec2Connecter implements vscode.Disposable {
...
 }
+
+/**
+ * Generate bash command (as string) to remove lines containing `pattern`.
+ * @param pattern pattern for deleted lines.
+ * @param filepath filepath (as string) to target with the command.
+ * @returns bash command to remove lines from file.
+ */
+export function getRemoveLinesCommand(pattern: string, hostOS: Ec2OS, filepath: string): string {
+    // Linux allows not passing extension to -i, whereas macOS requires zero length extension.
```
justinmk3, line 353: Also any BSD flavor. Does the '' work on Linux so we don't need to check? According to Claude, it should...

docs/arch_features.md
Hweinstock: good idea, I also added a note about which OS are supported atm.

packages/core/src/awsService/ec2/model.ts
Hweinstock: I saw it in the docs, and double checked I wasn't able to get it to work on `Amazon Linux`. 
image

Its annoying :(

I also thought about not using the `-i` option and instead piping it to another temporary file, then overwrite `authorized_keys` with the temporary file so that we could use same command on all OS, but I thought the tradeoff of taking 3-4 steps instead of 1 didn't feel worth it.

packages/core/src/test/awsService/ec2/model.test.ts
```diff
@@ -197,4 +202,94 @@ describe('Ec2ConnectClient', function () {
             }
         })
     })
+
+    describe('attemptToCleanKeys', async function () {
```
justinmk3, line 206:
```suggestion
    describe('tryCleanKeys', async function () {
```

justinmk3 (approved):

packages/core/src/awsService/ec2/model.ts
Hweinstock: Yeah, I wasn't sure if there would be any problems leaving backup files around. My thinking was to leave as little state/artifats as possible on the remote machine, but having different commands for different OS is annoying.

PR #6212, opened 2024-12-11
telemetry(amazonq): added helper for amazonq_utgGenerateTests metric and removed metric override 

chungjac:
## Problem
- Repetitive code when emitting amazonq_utgGenerateTests telemetry events
- Metric has already been migrated to aws-toolkit-common

## Solution
- Created sendTestGenerationToolkitEvent helper to extract common fields
- Removed metric override in this repo
- Tested and can confirm that metrics are same as previous method

---

### DIFF
--- a/packages/core/src/amazonqTest/chat/controller/controller.ts
+++ b/packages/core/src/amazonqTest/chat/controller/controller.ts
@@ -242,22 +242,20 @@ export class TestController {
         this.messenger.sendUpdatePromptProgress(data.tabID, null)
         const session = this.sessionStorage.getSession()
         const isCancel = data.error.message === unitTestGenerationCancelMessage
-        telemetry.amazonq_utgGenerateTests.emit({
-            cwsprChatProgrammingLanguage: session.fileLanguage ?? 'plaintext',
-            jobId: session.listOfTestGenerationJobId[0], // For RIV, UTG does only one StartTestGeneration API call
-            jobGroup: session.testGenerationJobGroupName,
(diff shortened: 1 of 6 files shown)


packages/core/src/amazonqTest/chat/controller/controller.ts
```diff
@@ -242,22 +242,20 @@ export class TestController {
...
+            session,
+            true,
+            isCancel ? 'Cancelled' : 'Failed',
+            session.startTestGenerationRequestId,
+            performance.now() - session.testGenerationStartTime,
+            getTelemetryReasonDesc(data.error),
+            session.isCodeBlockSelected,
+            session.artifactsUploadDuration,
+            session.srcPayloadSize,
+            session.srcZipFileSize
```
ctlai95, line 256: I think in general, this can be simplified since you are passing in the whole `session` object but also passing in fields like `session.isCodeBlockSelected` etc.

chungjac: I have tested that, but sometimes we would need `session.artifactsUploadDuration` to be undefined, so that it can be -1 in the kibana logs. 

If we abstract `session.artifactsUploadDuration`, `session.srcPayloadSize`, etc to the helper then it would appear as 0 for that field, when it has previously been -1. 

For the case of `session.isCodeBlockSelected`, if we abstract that to the helper than it will be `false` where if we pass it in as undefined then it will just not be a field in the kibana logs (as it previously has been)

PR #6220, opened 2024-12-12
fix(amazonq): increase GetTransformation maxRetries

dhasani23:
## Problem

Customers that experience transient network issues are unable to view their transformation results when the `GetTransformation` API fails after the default of 3 retries.

## Solution

Increase retries to 8.

---

### DIFF
--- a/packages/core/src/codewhisperer/client/codewhisperer.ts
+++ b/packages/core/src/codewhisperer/client/codewhisperer.ts
@@ -132,7 +132,7 @@ export class DefaultCodeWhispererClient {
         )) as CodeWhispererClient
     }
 
-    async createUserSdkClient(): Promise<CodeWhispererUserClient> {
+    async createUserSdkClient(maxRetries?: number): Promise<CodeWhispererUserClient> {
         const isOptedOut = CodeWhispererSettings.instance.isOptoutEnabled()
         session.setFetchCredentialStart()
         const bearerToken = await AuthUtil.instance.getBearerToken()
@@ -144,6 +144,7 @@ export class DefaultCodeWhispererClient {
                 apiConfig: userApiConfig,
                 region: cwsprConfig.region,
                 endpoint: cwsprConfig.endpoint,
+                maxRetries: maxRetries,
                 credentials: new Credentials({ accessKeyId: 'xxx', secretAccessKey: 'xxx' }),
                 onRequestSetup: [
                     (req) => {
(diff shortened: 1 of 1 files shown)


packages/core/src/codewhisperer/client/codewhisperer.ts
```diff
@@ -293,7 +294,8 @@ export class DefaultCodeWhispererClient {
     public async codeModernizerGetCodeTransformation(
         request: CodeWhispererUserClient.GetTransformationRequest
     ): Promise<PromiseResult<CodeWhispererUserClient.GetTransformationResponse, AWSError>> {
-        return (await this.createUserSdkClient()).getTransformation(request).promise()
+        // instead of the default of 3 retries, use 8 retries for this API which is polled every 5 seconds
+        return (await this.createUserSdkClient(8)).getTransformation(request).promise()
```
dhasani23, line 298: The SDK uses [3 retries](https://github.com/aws/aws-sdk-js/blob/966fa6c316dbb11ca9277564ff7120e6b16467f4/lib/service.js#L172) by default and [100 ms](https://github.com/aws/aws-sdk-js/blob/966fa6c316dbb11ca9277564ff7120e6b16467f4/lib/config-base.d.ts#L116) as the base delay during the exponential backoff retry strategy.

This means that if `GetTransformation` fails after only 0.1 + 0.2 + 0.4 = 0.7 seconds of delays, the transformation is marked as failed on the IDE and users are unable to view the transformation results. Increasing the retries to 8 means that `GetTransformation` now has a maximum of ~25 seconds of delays.

We have had customer cases where the polling fails, but the transformation eventually succeeds in our backend. This PR should increase the likelihood that those customers are able to view the transformation results.

dhasani23: `8` is admittedly arbitrary, and perhaps there's a better way to solve this problem, but I do think that permitting more than 0.7 seconds of delays is a good idea for this specific API. Open to other suggestions.

justinmk3 (approved):
Needs a review by someone from your team

packages/core/src/codewhisperer/client/codewhisperer.ts
ntarakad-aws: I was going to ask why 8, but thank you for explaining that!

ntarakad-aws (approved):

PR #6226, opened 2024-12-12
telemetry(amazonq): sending metric data in `onCodeGeneration`

siakmun-aws:
## Problem
This is a part of the task to implement client side alarms in order to track success rate for the client.

## Solution
- Emit metric data telemetry on success/failure.

---

### DIFF
--- a/packages/amazonq/.changes/next-release/Feature-06bf59c2-da3f-4ab5-ab86-634b3c25bd5e.json
+++ b/packages/amazonq/.changes/next-release/Feature-06bf59c2-da3f-4ab5-ab86-634b3c25bd5e.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "send metric data in onCodeGeneration"
+}
--- a/packages/amazonq/.changes/next-release/Feature-06bf59c2-da3f-4ab5-ab86-634b3c25bd5e.json
+++ b/packages/amazonq/.changes/next-release/Feature-06bf59c2-da3f-4ab5-ab86-634b3c25bd5e.json
@@ -1,4 +1,0 @@
-{
-	"type": "Feature",
-	"description": "send metric data in onCodeGeneration"
-}
--- a/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
+++ b/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
@@ -30,7 +30,7 @@ import {
 import { codeGenRetryLimit, defaultRetryLimit } from '../../limits'
 import { Session } from '../../session/session'
 import { featureDevScheme, featureName } from '../../constants'
-import { DeletedFileInfo, DevPhase, type NewFileInfo } from '../../types'
+import { DeletedFileInfo, DevPhase, MetricDataOperationName, MetricDataResult, type NewFileInfo } from '../../types'
 import { AuthUtil } from '../../../codewhisperer/util/authUtil'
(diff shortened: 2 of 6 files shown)


packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -484,8 +488,45 @@ export class FeatureDevController {
...
+                case FeatureDevServiceError.name:
+                    if (err.code === 'EmptyPatchException') {
+                        await session.sendMetricDataTelemetry(
+                            MetricDataOperationName.END_CODE_GENERATION,
+                            MetricDataResult.LLMFAILURE
+                        )
+                    } else {
+                        await session.sendMetricDataTelemetry(
+                            MetricDataOperationName.END_CODE_GENERATION,
+                            MetricDataResult.ERROR
```
neilk-aws, line 510: This is generic case right? Should this be a fault? Should we inspect for more codes?

neilk-aws (approved):

packages/amazonq/.changes/next-release/Feature-06bf59c2-da3f-4ab5-ab86-634b3c25bd5e.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "send metric data in onCodeGeneration"
+}
```
hayemaxi, line 4: This happens in the background and the user doesn't notice. We don't need a changelog item.
```suggestion
```

hayemaxi:
>   

You should use 'telemetry(amazonq)` instead of `feat` in the commit and PR title to avoid this.

packages/core/src/amazonqFeatureDev/types.ts
```diff
@@ -115,3 +115,15 @@ export interface UpdateFilesPathsParams {
     messageId: string
     disableFileActions?: boolean
 }
+
+export enum MetricDataOperationName {
+    START_CODE_GENERATION = 'StartCodeGeneration',
+    END_CODE_GENERATION = 'EndCodeGeneration',
```
justinmk3, line 121: Why use uppercase 
```suggestion
    StartCodeGeneration = 'StartCodeGeneration',
    EndCodeGeneration = 'EndCodeGeneration',
```

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -517,6 +559,7 @@ export class FeatureDevController {
                 }
             }
         }
+        await session.sendMetricDataTelemetry(MetricDataOperationName.END_CODE_GENERATION, MetricDataResult.SUCCESS)
```
nkomonen-amazon, line 562: I'm not sure if these designs are final. But does it makes to just send a single metric instead of start/end? This would save some work on the server side

packages/core/src/amazonqFeatureDev/session/session.ts
```diff
@@ -285,6 +285,25 @@ export class Session {
         return { leftPath, rightPath, ...diff }
     }
 
+    public async sendMetricDataTelemetry(operationName: string, result: string) {
+        await this.proxyClient.sendMetricData({
+            metricName: 'Operation',
+            metricValue: 1,
+            timestamp: new Date(Date.now()),
```
nkomonen-amazon, line 292: I think this should be good enough
```suggestion
            timestamp: new Date(),
```

PR #6234, opened 2024-12-13
fix(auth): SSO failed to get token due to missing 'refreshToken'

### LINKED ISSUE #6230: Unable to connect with profile credentials
## Problem

Cannot connect to AWS using my profile after latest patch.

1. Go to Command Palette and hit 'Connect to AWS'
2. Use a profile to connect
3. See error 'Unable to authenticate connection'
4. Output:
```
2024-12-13 14:07:08.003 [info] auth: Updating connection state of profile:dev to authenticating
2024-12-13 14:07:08.005 [info] auth: Handling validation error of connection: profile:dev
2024-12-13 14:07:08.005 [info] auth: Updating connection state of profile:dev to invalid
2024-12-13 14:07:08.005 [error] _aws.toolkit.auth.reauthenticate: Error: Unable to authenticate connection
	 -> Error: SSO cache data unexpectedly missing props: ["refreshToken"]
```

The previous Toolkit version didn't have that issue.

-   OS: Windows 11 23H2
-   Visual Studio Code version: 1.95.3
-   AWS Toolkit version: 3.39.0

nkomonen-amazon:
## Problem:

Error users were seeing was:

```
Error: SSO cache data unexpectedly missing props: ["refreshToken"]
```

This was due to a change from a previous PR that assumed the refreshToken was always present in SSO cache.

## Solution:

It looks like the refreshToken does not exist for all cases, so some research needs to be done.

But for now this reverts the validation that the refreshToken exists. Though it keeps the validation that the accessToken exists since that is always guaranteed.

**A temporary workaround should be to sign out/in and things should start working. If not please downgrade 3.38.0 until we release this fix next week.**

Fixes #6230 

---

### DIFF
--- a/packages/core/src/auth/sso/cache.ts
+++ b/packages/core/src/auth/sso/cache.ts
@@ -10,7 +10,7 @@ import { getLogger } from '../../shared/logger/logger'
 import fs from '../../shared/fs/fs'
 import { createDiskCache, KeyedCache, mapCache } from '../../shared/utilities/cacheUtils'
 import { stripUndefined } from '../../shared/utilities/collectionUtils'
-import { getMissingProps, hasProps, selectFrom } from '../../shared/utilities/tsUtils'
+import { hasProps, selectFrom } from '../../shared/utilities/tsUtils'
 import { SsoToken, ClientRegistration } from './model'
 import { DevSettings } from '../../shared/settings'
 import { onceChanged } from '../../shared/utilities/functionUtils'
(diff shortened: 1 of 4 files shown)


jpinkney-aws (approved):
Just for my understanding, is this a patch to fix the users issue but we still want to investigate why the refreshToken might not be stored in the cache?

justinmk3 (approved):

nkomonen-amazon:
@jpinkney-aws correct

It looks like there maybe be cases where the identity service will not provide a refreshToken. Eg if the session length is short enough it might not provide one.

So this doesn't seem like a bug, but a behavior we'd want to document

PR #6238, opened 2024-12-13
fix(amazonq): show correct diff after consecutive transformations

dhasani23:
## Problem

If a user runs a transformation, rejects changes, then immediately runs another transformation on the same project (rare, but it happens), we attempt to show the user the old diff (which has already been deleted at this point).

## Solution

Reset `patchFiles` before showing the user the patch file.

---


packages/core/src/amazonq/webview/ui/quickActions/generator.ts
ntarakad-aws, line 1: I was just about to make this change so thanks for catching

ntarakad-aws (approved):

jpinkney-aws (approved):

packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts
```diff
@@ -430,6 +430,7 @@ export class ProposedTransformationExplorer {
 
             let deserializeErrorMessage = undefined
             let pathContainingArchive = ''
+            patchFiles = [] // reset patchFiles if there was a previous transformation
```
justinmk3, line 433: Fun js trick: you can clear an array by setting its length to zero (so the variable could stay `const`, above)

    patchFiles.length = 0

PR #6239, opened 2024-12-13
feat(amazonq): Adds capability to send new context commands to AB groups

spfink:
## Problem

When working on new features available to AB groups, the new action are not easily discoverable to users

## Solution

Add the ability to send a new command that will appear when a user types @ in the chat window

---

### DIFF
--- a/packages/core/src/amazonq/webview/ui/main.ts
+++ b/packages/core/src/amazonq/webview/ui/main.ts
@@ -106,6 +106,10 @@ export const createMynahUI = (
 
     let isDocEnabled = amazonQEnabled
 
+    let featureConfigs: Map<string, FeatureContext> = tryNewMap(featureConfigsSerialized)
+
+    const highlightCommand = featureConfigs.get('highlightCommand')
+
     let tabDataGenerator = new TabDataGenerator({
         isFeatureDevEnabled,
         isGumbyEnabled,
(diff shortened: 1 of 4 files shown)


justinmk3:
Is this expected for tomorrow's release?

Will-ShaoHua (approved):

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -106,13 +106,19 @@ export const createMynahUI = (
 
     let isDocEnabled = amazonQEnabled
 
+    // @ts-ignore
+    let featureConfigs: Map<string, FeatureContext> = tryNewMap(featureConfigsSerialized)
```
hayemaxi, line 110: Why ignore?

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -106,13 +106,19 @@ export const createMynahUI = (
...
+
+    const highlightCommand = featureConfigs.get('highlightCommand')
```
hayemaxi, line 112: This doesn't look like a scalable way to add AB commands

packages/core/src/amazonq/webview/ui/main.ts
spfink: Not sure, this was existing code just moved up from lines 127 as you can see below. I can try and run a build without it

spfink: Removed the annotation since this is now used in many places

hayemaxi (approved):

PR #6168, opened 2024-12-06
refactor(amazonq): crossfileContextUtil to improve readability and improve metric sent out

Will-ShaoHua:
## Solution

---

- Treat all work as PUBLIC. Private `feature/x` branches will not be squash-merged at release time.
- Your code changes must meet the guidelines in [CONTRIBUTING.md](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#guidelines).

License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6242, opened 2024-12-13
telemetry(inline-suggestion): fine tune supplemental context strategy telemetry

Will-ShaoHua:
## Problem
Current `supplementalContextStrategy ` only reflect what group the user is in, however by doing this it won't help much on our data analysis. What we need is the context strategy being used so that we know how much % of users have repomap or opentabs and how % of users have empty supplemental context etc.

## Solution

---


Will-ShaoHua:
there are a lot of undesired duplicate code within this function, planing to use #6168 to resolve this as a followup and it will requires more testing.

leigaol (approved):

PR #6243, opened 2024-12-13
refactor(settings): minimize duplication across prompt settings. 

Hweinstock:
## Problem
`ToolkitPromptSettings` and `AmazonQPromptSettings` reimplement the same functions. 

## Solution
- factor these classes out into the same anonymous class. 
- avoid using inheritance to keep strong typing from the settings. 

TODO: figure out how to type the booleans to avoid casting

---


packages/core/src/shared/settings.ts
```diff
@@ -619,68 +619,44 @@ export function fromExtensionManifest<T extends TypeDescriptor & Partial<Section
...
+function getPromptSettings<P extends 'amazonQ.suppressPrompts' | 'aws.suppressPrompts'>(promptsKey: P) {
+    const prompts = settingsProps[promptsKey]
+    type promptName = keyof typeof prompts & string
+    return class AnonymousPromptSettings extends Settings.define(
+        promptsKey,
+        toRecord(keys(prompts), () => Boolean)
+    ) {
+        public isPromptEnabled(promptName: promptName) {
+            try {
+                return !this._getOrThrow(promptName, false as never)
```
Hweinstock, line 638: Believe there is an issue with how I setup types since I need the cast here.

packages/core/src/auth/sso/ssoAccessTokenProvider.ts
```diff
@@ -808,7 +808,7 @@ class DiskCacheErrorMessage {
             : ToolkitPromptSettings.instance
 
         // We know 'ssoCacheError' is in all extension prompt settings
-        if (await promptSettings.isPromptEnabled('ssoCacheError')) {
+        if (promptSettings.isPromptEnabled('ssoCacheError')) {
```
justinmk3, line 811: This is a welcome change 👍  
Not a strong reason for this function to be async

Hweinstock: isolated this change here: https://github.com/aws/aws-toolkit-vscode/pull/6255/files

PR #6246, opened 2024-12-15
fix(amazonq): retry S3 upload

dhasani23:
## Problem

S3 upload sometimes fails due to transient issues.

## Solution

Add up to 3 retries.

---

### DIFF
--- a/packages/core/src/codewhisperer/models/constants.ts
+++ b/packages/core/src/codewhisperer/models/constants.ts
@@ -747,7 +747,7 @@ export const cleanInstallErrorNotification = `Amazon Q could not run the Maven c
 export const enterJavaHomeChatMessage = 'Enter the path to JDK '
 
 export const projectPromptChatMessage =
-    'I can upgrade your JAVA_VERSION_HERE. To start the transformation, I need some information from you. Choose the project you want to upgrade and the target code version to upgrade to. Then, choose Confirm.'
+    'I can upgrade your Java project. To start the transformation, I need some information from you. Choose the project you want to upgrade and the target code version to upgrade to. Then, choose Confirm.'
 
 export const windowsJavaHomeHelpChatMessage =
     'To find the JDK path, run the following commands in a new terminal: `cd "C:/Program Files/Java"` and then `dir`. If you see your JDK version, run `cd <version>` and then `cd` to show the path.'
--- a/packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
+++ b/packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
@@ -112,16 +112,17 @@ export async function uploadArtifactToS3(
 
         let response = undefined
         /* The existing S3 client has built-in retries but it requires the bucket name, so until
-         * CreateUploadUrl can be modified to return the S3 bucket name, manually implement retries
+         * CreateUploadUrl can be modified to return the S3 bucket name, manually implement retries.
+         * Alternatively, when waitUntil supports a fixed number of retries and retriableCodes, use that.
          */
         const retriableCodes = [408, 429, 500, 502, 503, 504]
-        for (let i = 0; i < 3; i++) {
+        for (let i = 0; i < 4; i++) {
             try {
                 response = await request.fetch('PUT', resp.uploadUrl, {
                     body: buffer,
                     headers: getHeadersObj(sha256, resp.kmsKeyArn),
                 }).response
-                getLogger().info(`CodeTransformation: upload to S3 status on attempt ${i + 1}/3 = ${response.status}`)
+                getLogger().info(`CodeTransformation: upload to S3 status on attempt ${i + 1}/4 = ${response.status}`)
                 if (response.status === 200) {
                     break
                 }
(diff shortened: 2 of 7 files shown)


packages/core/src/codewhisperer/service/transformByQ/transformProjectValidationHandler.ts
```diff
@@ -70,72 +66,9 @@ async function getMavenJavaProjects(javaProjects: TransformationCandidateProject
...
- * and we allow the user to specify the Java version.
- */
+// This function filters all open projects by first searching for a .java file and then searching for a pom.xml file in all projects.
 export async function validateOpenProjects(projects: TransformationCandidateProject[]) {
     const javaProjects = await getJavaProjects(projects)
-
     const mavenJavaProjects = await getMavenJavaProjects(javaProjects)
-
-    // These projects we know must contain a pom.xml and a .java file
-    const projectsValidToTransform = await getProjectsValidToTransform(mavenJavaProjects)
```
dhasani23, line 138: Deleted this function entirely because 1) it was expensive (running a shell command (`javap`) on every open project) and 2) it was just used to format a String (see `createLanguageUpgradeConfirmationPrompt` above); every project returned by `getMavenJavaProjects` above was also returned by `getProjectsValidToTransform`

packages/core/src/codewhisperer/models/constants.ts
```diff
@@ -747,7 +747,7 @@ export const cleanInstallErrorNotification = `Amazon Q could not run the Maven c
 export const enterJavaHomeChatMessage = 'Enter the path to JDK '
 
 export const projectPromptChatMessage =
-    'I can upgrade your JAVA_VERSION_HERE. To start the transformation, I need some information from you. Choose the project you want to upgrade and the target code version to upgrade to. Then, choose Confirm.'
+    'I can upgrade your Java project. To start the transformation, I need some information from you. Choose the project you want to upgrade and the target code version to upgrade to. Then, choose Confirm.'
```
dhasani23, line 750: See comment below; rather than having 75+ LOC and running a shell command for each open project just to say "Java 8 project" or "Java 11 project" or "Java 8 and 11 projects", etc. just say "Java project"

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -103,18 +103,41 @@ export async function uploadArtifactToS3(
...
+                    await sleep(1000 * Math.pow(2, i))
+                }
+            }
+        }
+        if (!response || response.status !== 200) {
+            const uploadFailedError = `Upload failed after up to 2 retries with status code = ${response?.status ?? 'unavailable'}`
+            getLogger().error(`CodeTransformation: ${uploadFailedError}`)
+            throw new Error(uploadFailedError)
+        }
+        getLogger().info('CodeTransformation: Upload to S3 succeeded')
```
dhasani23, line 140: I'm not a big fan of implementing our own retry logic rather than doing it at the SDK layer, but the existing S3 client's upload functionality requires a bucket name as a part of the request, and 1) I don't think we should have the internal S3 bucket names we use floating around this repo, 2) we use multiple different buckets, and 3) we don't have access to the bucket name in the response of our `CreateUploadUrl` — this API would have to be changed and then we can use the existing S3 client.

dhasani23:
Looks like [this](https://github.com/aws/aws-toolkit-vscode/blob/4728b23ce8f4b6b6492375556b44e5fd532736c7/packages/core/src/test/techdebt.test.ts#L42) tech debt-related test is failing

packages/core/src/test/codewhisperer/commands/transformByQ.test.ts
```diff
@@ -464,4 +472,92 @@ describe('transformByQ', function () {
...
+                'test.zip',
+                { uploadId: '123', uploadUrl: 'http://test.com', kmsKeyArn: 'arn' },
+                'sha256',
+                Buffer.from('test')
+            )
+        } catch (e) {
+            error = e as Error
+        }
+        sinon.assert.match(
+            error?.message.includes('Upload failed with status code = 400; did not automatically retry'),
```
justinmk3, line 558: Any reason not to use `assert.rejects` here (and in the other tests)? Search the codebase for usage examples.

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
justinmk3: Thanks for that context! That info is probably worth a code comment here or in the docstring of this function.

For reference, we have discussed some sort of util to "poll stuff", see https://taskei.amazon.dev/tasks/IDE-15280
[startDevEnvironmentWithProgress](https://github.com/aws/aws-toolkit-vscode/blob/98d8324ff95aa75e0d79c88e169414b66d024e86/packages/core/src/shared/clients/codecatalystClient.ts#L641) is a very complex case, maybe not exactly analogous to standard "request retry".

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -103,18 +103,41 @@ export async function uploadArtifactToS3(
...
-        const response = await request.fetch('PUT', resp.uploadUrl, {
-            body: buffer,
-            headers: getHeadersObj(sha256, resp.kmsKeyArn),
-        }).response
-        getLogger().info(`CodeTransformation: Status from S3 Upload = ${response.status}`)
+        let response = undefined
+        const retriableCodes = [408, 429, 500, 502, 503, 504]
+        for (let i = 0; i < 3; i++) {
+            try {
+                response = await request.fetch('PUT', resp.uploadUrl, {
```
justinmk3, line 117: (Non-blocker) Looks like `waitUntil` wouldn't work here because it only has "max time", not "max retries". Probably should add that. Being be able to "find all waitUntil references" can be useful to see all the areas in the codebase that are doing this pattern.

dhasani23: On second thought, I don't see a point in retrying the upload if we know it failed for non-retriable reasons, so I will keep the PR as it is for now. Updated the comment to reflect this.

justinmk3: > But it looks like it will always retry until the timeout period elapses, compared to this function which will only retry as long as the status code is retriable

Your callback passed to `waitUntil` can decide whether to retry by its return value. However, it would be nicer if waitUntil supported a more explicit way to stop retrying.

Issue #6225, opened 2024-12-12
unreliable test: "getFileSha384 performance tests"


justinmk3:
## Test Details

-   Name of test: `getFileSha384 performance tests`
-   OS of failure: linux
-   VSCode version of failure (minimum/insider/stable): insiders
-   Link to failing run: https://d1ihu6zq92vp9p.cloudfront.net/a00745e1-5c35-430a-8c73-db9b2756a462/report.html
-   Link to failing test:

## Log of Test Failure

```
  1) getFileSha384
       performance tests
         2MB performance tests
           "after all" hook for "2MB - test run 10":

      AssertionError [ERR_ASSERTION]: Expected total user CPU usage for 2MB to be less than 400. Actual user CPU usage was 462.1365886618588
      + expected - actual

      -false
      +true
      
      at assertPerformanceMetrics (/codebuild/output/src2928107957/src/github.com/aws/aws-toolkit-vscode/packages/core/src/shared/performance/performance.ts:182:11)
      at Context.<anonymous> (/codebuild/output/src2928107957/src/github.com/aws/aws-toolkit-vscode/packages/core/src/shared/performance/performance.ts:160:13)
      at Context.fn (/codebuild/output/src2928107957/src/github.com/aws/aws-toolkit-vscode/packages/core/src/test/setupUtil.ts:34:24)
      at processImmediate (node:internal/timers:483:21)
      at process.topLevelDomainCallback (node:domain:161:15)
      at process.callbackTrampoline (node:internal/async_hooks:128:24)
```

Hweinstock:
Resolved by https://github.com/aws/aws-toolkit-vscode/pull/6249

PR #6249, opened 2024-12-16
test(perf): increase cpu threshold. 

Hweinstock:
## Problem
https://github.com/aws/aws-toolkit-vscode/issues/6225
- Didn't think 400% could be exceeded, but this one is especially volatile. Sometimes it barely spikes, and sometimes it goes crazy. 
## Solution
- Increase to 500%

---


jpinkney-aws (approved):
At this point are we even getting value from the performance tests? It feels like we're just going to increase the number indefinitely

Hweinstock:
I don't think the system stat thresholds are giving us much insight because of their flakiness. The thresholds on `fs` operations does give us some info, and would catch performance regressions where we somehow accidentally blow up the number of `fs` operations. However, I think its unlikely an engineer would do this, and performance issues affecting customers would likely be more nuanced than us making a few unneeded `fs` calls.

Hweinstock:
Hmmm, looks like these tests are very flaky https://github.com/aws/aws-toolkit-vscode/blob/4728b23ce8f4b6b6492375556b44e5fd532736c7/packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts#L314-L345.

justinmk3:
> Hmmm, looks like these tests are very flaky

https://github.com/aws/aws-toolkit-vscode/issues/6252

PR #6255, opened 2024-12-16
refactor(settings): make isPromptEnabled non-async

Hweinstock:
## Problem
There is no real reason this needs to be async. It is also now inconsistent with `isExperimentEnabled`. 

## Solution
Make non-async and add logged error if the promise rejects. 

---

### DIFF
--- a/packages/core/src/shared/settings.ts
+++ b/packages/core/src/shared/settings.ts
@@ -600,7 +600,7 @@ export function fromExtensionManifest<T extends TypeDescriptor & Partial<Section
  *
  * ### Usage:
  * ```
- * if (await settings.isPromptEnabled('myPromptName')) {
+ * if (settings.isPromptEnabled('myPromptName')) {
  *     // Show some sort of prompt
  *     const userResponse = await promptUser()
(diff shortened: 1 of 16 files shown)


packages/core/src/shared/settings.ts
```diff
@@ -660,19 +660,19 @@ export class AmazonQPromptSettings
...
     implements PromptSettings
 {
-    public async isPromptEnabled(promptName: amazonQPromptName): Promise<boolean> {
+    public isPromptEnabled(promptName: amazonQPromptName): boolean {
         try {
             return !this._getOrThrow(promptName, false)
         } catch (e) {
             this._log('prompt check for "%s" failed: %s', promptName, (e as Error).message)
-            await this.reset()
+            this.reset().catch((e) => getLogger().error(`failed to reset prompt settings: %O`, (e as Error).message))
```
justinmk3, line 668:
```suggestion
            this.reset().catch((e) => getLogger().error(`isPromptEnabled: reset() failed: %O`, (e as Error).message))
```

Issue #6252, opened 2024-12-16
unreliable test: "crossFileContextUtil full support should be non empty"


justinmk3:
## Test Details

-   Name of test: `crossFileContextUtil full support should be non empty:`
-   OS of failure: macOS
-   VSCode version of failure (minimum/insider/stable): all
-   Link to failing run: (one of many:) https://github.com/aws/aws-toolkit-vscode/actions/runs/12323095533/job/34398051936
-   Link to failing test:

## Log of Test Failure

```
  1) crossFileContextUtil
       full support
         should be non empty:

      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert_1.default.ok(actual && actual.supplementalContextItems.length !== 0)

      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (test/unit/codewhisperer/util/crossFileContextUtil.test.ts:342:24)
      at runNextTicks (node:internal/process/task_queues:60:5)
      at processTimers (node:internal/timers:516:9)

  2) crossFileContextUtil
       full support
         should be non empty:

      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert_1.default.ok(actual && actual.supplementalContextItems.length !== 0)

      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (test/unit/codewhisperer/util/crossFileContextUtil.test.ts:342:24)
```

## Related

- https://github.com/aws/aws-toolkit-vscode/issues/6041
- https://github.com/aws/aws-toolkit-vscode/issues/6079

justinmk3:
Also sometimes the `openTabsContext` case:

```
 1) supplementalContextUtil
       fetchSupplementalContext
         openTabsContext
           opentabContext should include chunks if non empty:

      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert_1.default.ok(actual?.supplementalContextItems.length === 3)

      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (test/unit/codewhisperer/util/supplemetalContextUtil.test.ts:45:24)
      at runNextTicks (node:internal/process/task_queues:60:5)
      at processTimers (node:internal/timers:516:9)
```

Will-ShaoHua:
temporarily skipped test case other than `java` and only run 1 language test case, still looking into test cases failure but have tested functionally and works fine on the languages skipped

PR #6256, opened 2024-12-16
fix(inline-suggestion): replace vscode.cancellation with waitUntil for timeout

Will-ShaoHua:
## Problem

related issue:  https://github.com/aws/aws-toolkit-vscode/issues/6079, https://github.com/aws/aws-toolkit-vscode/issues/6252

caller
```
function main () {
       // init vscode cancellation token
       const cancellationToken
       setTimeout(100, () => {
              cancellationToken.cancel()
       })

      highlevelWrapperFetchSupplementalContext(editor, cancellationToken)
}

```

```
export function highlevelWrapperFetchSupplementalContext(editor, cancellationToken) {
       const supplementalContext = waitUntil(100, () => {
            // here always timeout and throw TimeoutException
            const opentabs =  await fetchOpenTabsContext(...)
            const projectContext = await fetchProjectContext()

           const result = []
           if (projectContext not empty) {
                    // push project context
           }

           if (opentabs not empty) {}
                   // push openttabs
           })  
          

         return result
}


async function fetchOpenTabsContext(editor, cancellationToken) {
      ....
      // VSC api call
}

async function fetchProjectContext() {
     ....
     // LSP call
}

```

After investigation, it looks like mix use of `vscode.CancellationToken` and `waitUntil()` will likely cause cancellation token to be cancelled prematurely (might be because another layer of waitUntil will run the fetchOpenTabsContext asynchronously thus causing it to timeout prematurely) therefore `fetchOpebtabsContext(..)` will return null in this case and hence causing test cases failing.

Therefore, the issue here is actually not the test case itself and they're failing due to race condition

## Solution
remove usage of cancellation token and only use waitUntil for timeout purpose

## Functional testing

retrieved sup context as expected

### Case 1: repomap is available (there are local imports)
```
2024-12-16 13:10:15.616 [debug] CodeWhispererSupplementalContext:
    isUtg: false,
    isProcessTimeout: false,
    contentsLength: 14436,
    latency: 16.67179101705551
    strategy: codemap
    Chunk 0:
        Path: q-inline
        Length: 10209
        Score: 0
    Chunk 1:
        Path: /Volumes/workplace/ide/aws-toolkit-vscode-staging/packages/core/src/codewhisperer/service/serviceContainer.ts
        Length: 1486
        Score: 22.60257328587725
    Chunk 2:
        Path: /Volumes/workplace/ide/aws-toolkit-vscode-staging/packages/core/src/codewhisperer/tracker/lineTracker.ts
        Length: 1649
        Score: 19.106700952807103
    Chunk 3:
        Path: /Volumes/workplace/ide/aws-toolkit-vscode-staging/packages/core/src/codewhisperer/tracker/lineTracker.ts
        Length: 1092
        Score: 10.334690655691002
```

### Case 2: No repomap, should fallback to opentabs only
image

```
2024-12-16 13:11:29.738 [debug] CodeWhispererSupplementalContext:
    isUtg: false,
    isProcessTimeout: false,
    contentsLength: 5046,
    latency: 16.311500012874603
    strategy: opentabs
    Chunk 0:
        Path: /Volumes/workplace/ide/aws-toolkit-vscode-staging/packages/core/src/codewhisperer/tracker/lineTracker.ts
        Length: 1564
        Score: 0
    Chunk 1:
        Path: /Volumes/workplace/ide/aws-toolkit-vscode-staging/packages/core/src/codewhisperer/tracker/lineTracker.ts
        Length: 1649
        Score: 0
    Chunk 2:
        Path: /Volumes/workplace/ide/aws-toolkit-vscode-staging/packages/core/src/codewhisperer/tracker/lineTracker.ts
        Length: 1833
        Score: 0
```

---

### DIFF
--- a/packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts
+++ b/packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts
@@ -330,6 +330,16 @@ describe('crossFileContextUtil', function () {
         fileExtLists.forEach((fileExt) => {
             it('should be non empty', async function () {
                 sinon.stub(FeatureConfigProvider.instance, 'getProjectContextGroup').returns('control')
+                sinon
+                    .stub(LspController.instance, 'queryInlineProjectContext')
+                    .withArgs(sinon.match.any, sinon.match.any, 'codemap')
+                    .resolves([
+                        {
+                            content: 'foo',
+                            score: 0,
+                            filePath: 'q-inline',
+                        },
+                    ])
                 const editor = await toTextEditor('content-1', `file-1.${fileExt}`, tempFolder)
                 await toTextEditor('content-2', `file-2.${fileExt}`, tempFolder, { preview: false })
                 await toTextEditor('content-3', `file-3.${fileExt}`, tempFolder, { preview: false })
--- a/packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts
+++ b/packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts
@@ -312,7 +312,9 @@ describe('crossFileContextUtil', function () {
     })
 
     describe('full support', function () {
-        const fileExtLists = ['java', 'js', 'ts', 'py', 'tsx', 'jsx']
+        // TODO: fix it
+        // const fileExtLists = ['java', 'js', 'ts', 'py', 'tsx', 'jsx']
+        const fileExtLists = ['java']
 
         before(async function () {
             this.timeout(60000)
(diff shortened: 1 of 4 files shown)


packages/amazonq/test/unit/codewhisperer/util/crossFileContextUtil.test.ts
```diff
@@ -44,7 +44,7 @@ describe('crossFileContextUtil', function () {
             sinon.restore()
         })
 
-        it('for control group, should return opentabs context where there will be 3 chunks and each chunk should contains 50 lines', async function () {
+        it.skip('for control group, should return opentabs context where there will be 3 chunks and each chunk should contains 50 lines', async function () {
```
Will-ShaoHua, line 47: we always assume t1 group for now

leigaol (approved):

Will-ShaoHua:
looking into test failure

justinmk3 (approved):

PR #6258, opened 2024-12-16
test(iinline-suggestion): fix crossFileContextUtil.test failure caused by using real clock and test timeouts

Will-ShaoHua:
## Problem
#6252

As test previously was using real system clock, it's likely to timeout due to heavy computation in an undeterministic way

## Solution
Use fake clock

---


justinmk3:
This probably doesn't fix https://github.com/aws/aws-toolkit-vscode/issues/6266 , right?

Will-ShaoHua:
> This probably doesn't fix https://github.com/aws/aws-toolkit-vscode/issues/6266 , right?

checking

PR #6261, opened 2024-12-17
test(core): merge all package test reports

jpinkney-aws:
## Problem
- CI test reporting only captures the last package's results
- Running `npm run testE2E` only preserves toolkit's report.xml, losing other package results

## Solution
- Generate individual report.xml files per subproject
- Consolidate all existing package reports into root .test-reports/report.xml
- This is done in the buildspecs themselves rather than directly in the package.json (the original approach to solve this problem) because mac/linux and windows have two different ways of getting the last error code, resulting in a complicated package.json for something thats only needed in our codebuilds

---


Hweinstock:
What are the tradeoffs for merging the report versus having one per extension? Curious because for code coverage we have separate ones.

jpinkney-aws:
If you don't merge them you can't see them in the ui here: https://d1ihu6zq92vp9p.cloudfront.net/487310e6-ebfc-4fef-a0b1-b441147b3c23/report.html. That's really the issue i'm trying to solve. It's pretty annoying when an amazon q test fails and then you have to manually look through the logs to find the problem

I don't think that restriction applies to codecov since you can upload multiple different reports at the same time

buildspec/linuxE2ETests.yml
```diff
@@ -37,7 +37,7 @@ phases:
         commands:
             - export HOME=/home/codebuild-user
             # Ignore failure until throttling issues are fixed.
-            - xvfb-run npm run testE2E
+            - xvfb-run npm run testE2E; PREVIOUS_TEST_EXIT_CODE=$? npm run mergeReports
```
justinmk3, line 40: can avoid this env var by passing the exit code directly as an arg
```suggestion
            - xvfb-run npm run testE2E; npm run mergeReports -- "$?"
```

package.ts shows how to get script args: https://github.com/aws/aws-toolkit-vscode/blob/5b5df855b2cc2f12ea8e1298e1ebee5cfe6b598a/scripts/package.ts#L40-L42

buildspec/linuxTests.yml
```diff
@@ -41,7 +41,7 @@ phases:
                     # Ensure that "foo | run_and_report" fails correctly.
                     set -o pipefail
                     . buildspec/shared/common.sh
-                    2>&1 xvfb-run npm test --silent | run_and_report 2 \
+                    2>&1 xvfb-run npm test --silent; npm run mergeReports -- "$?" | run_and_report 2 \
                         'rejected promise not handled' \
```
justinmk3, line 45: Based on a local test of this command:

    2>&1 echo aaa; echo bbb | grep x

I suspect the above `;` may not correctly pipe all the output from `npm test` to `run_and_report` as intended.

To confirm that the piping to `run_and_report` is still working correctly, can you try temporarily changing this search patter to something that always outputs, like:
```suggestion
                        'deleteTestTempDirs' \
```

then CI should fail.

If it doesn't, we can fix the issue by changing the pipeline to something like (note: whitespace near the braces is required):

     { 2>&1 xvfb-run npm test --silent; npm run mergeReports -- "$?"; } | run_and_report 2 \

jpinkney-aws: Just tested and that was indeed the case. I've switched to your suggestion and it seems to work as expected

scripts/mergeReports.ts
```diff
@@ -0,0 +1,71 @@
...
+import * as path from 'path'
+import * as xml2js from 'xml2js'
+
+/**
+ * Merge all of the packages/ test reports into a single directory
+ */
+async function mergeReports() {
+    console.log('Merging test reports')
+
+    const packagesDir = path.join(__dirname, '..', 'packages')
```
justinmk3, line 16: string literals do not need join()

scripts/mergeReports.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    // Get all packages/* directories
+    const packageDirs = fs.readdirSync(packagesDir).map((dir) => path.join(packagesDir, dir))
+
+    // Find report.xml files in .test-reports subdirectories
+    const testReports = packageDirs
+        .map((dir) => path.join(dir, '.test-reports', 'report.xml'))
```
justinmk3, line 23: string literals do not need join()

scripts/mergeReports.ts
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    const builder = new xml2js.Builder()
+    const xml = builder.buildObject(mergedReport)
+
+    /**
+     * Create the new test reports directory and write the test report
+     */
+    const reportsDir = path.join(__dirname, '..', '.test-reports')
```
justinmk3, line 47: string literals do not need join()

justinmk3 (approved):

PR #6264, opened 2024-12-17
refactor(general): avoid async method inside `forEach` (WIP)

Hweinstock:
## Problem
usage of an async function inside a `forEach`, which is almost always not desired. `forEach` will not await the results leading to race conditions and inconsistent orderings.

## Solution
- add lint rule
- migrate existing cases
---


plugins/eslint-plugin-aws-toolkits/test/rules/no-inline-async-foreach.test.ts
```diff
@@ -0,0 +1,24 @@
...
+        'list.forEach(asyncFunctionOrNot)',
+        'list.forEach(() => console.log(x))',
+        'list.forEach(function () {})',
+    ],
+
+    invalid: [
+        { code: 'list.forEach(async (a) => await Promise.resolve(a * a))', errors: [errMsg] },
+        { code: 'list.forEach(async (a: any) => console.log(x))', errors: [errMsg] },
+        { code: 'list.forEach((a) => a.forEach(async (b) => a * b))', errors: [errMsg] },
+        { code: 'list.forEach(async function () {})', errors: [errMsg] },
```
hayemaxi, line 22: what about
```
const fn = async (item) => {}
list.forEach(fn)
```

Hweinstock: See recent commits, also added a few more edge cases.

.eslintrc.js
```diff
@@ -178,6 +178,7 @@ module.exports = {
         'aws-toolkits/no-console-log': 'error',
         'aws-toolkits/no-json-stringify-in-log': 'error',
         'aws-toolkits/no-printf-mismatch': 'error',
+        'aws-toolkits/no-async-in-foreach': 'error',
```
justinmk3, line 181: should we just ban `forEach`? it is kind of pointless.

Hweinstock: Yea, I think banning it out right has a few advantages. Additionally, determining if a function is async is definitely non-trivial. This work is moved to https://github.com/aws/aws-toolkit-vscode/pull/6281

Issue #6213, opened 2024-12-11
unreliable test: shows warning 5 minutes before shutdown for 60 minute timeout


Hweinstock:
## Test Details

-   Name of test: shows warning 5 minutes before shutdown for 60 minute timeout
-   OS of failure: Linux
-   VSCode version of failure (minimum/insider/stable): (1.95.3)
-   Link to failing run:
-   Link to failing test: https://github.com/aws/aws-toolkit-vscode/blob/2af8b45b8e2ba68bf81f7254ddb1bd27a3cbe887/packages/core/src/testInteg/codecatalyst/devEnv.test.ts#L72-L79

## Log of Test Failure
``` 
AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  {
    message: 'Your CodeCatalyst Dev Environment has been inactive for 55 minutes, and will stop soon.',
+   minute: 54
-   minute: 55
  }
    at assertMessagesShown (/codebuild/output/src1534573611/src/github.com/aws/aws-toolkit-vscode/packages/core/src/testInteg/codecatalyst/devEnv.test.ts:144:20)
    at async Context.<anonymous> (/codebuild/output/src1534573611/src/github.com/aws/aws-toolkit-vscode/packages/core/src/testInteg/codecatalyst/devEnv.test.ts:76:9)

      + expected - actual

       {
         "message": "Your CodeCatalyst Dev Environment has been inactive for 55 minutes, and will stop soon."
      -  "minute": 54
      +  "minute": 55
       }
```
## Additional Information
Integ test

justinmk3:
This test is definitely a known issue. If we can just retry it, we should. Or possibly skip() for now.

Hweinstock:
Another failure on https://github.com/aws/aws-toolkit-vscode/commit/0d8c8fcb1c5b0540f0548f9b265f81ba9a23d4e9. Looks like 1 minute buffer period is no longer effective at mitigating flakiness.

PR #6265, opened 2024-12-17
test(codecatalyst): only fail test if message is more than a minute off. 

Hweinstock:
## Problem
https://github.com/aws/aws-toolkit-vscode/issues/6213

## Solution
- The important part of this test is that the message comes after being inactive. If the message comes slightly early/late, that is okay. If it doesn't come at all, then we have problems and should fail. 
- Pass test if message comes within a minute of expected. 

## Alternatives
- skip the test. 
- Re-do test setup since its flaky. 

---

### DIFF
--- a/packages/core/src/testInteg/codecatalyst/devEnv.test.ts
+++ b/packages/core/src/testInteg/codecatalyst/devEnv.test.ts
@@ -141,6 +141,12 @@ describe('InactivityMessage', function () {
                 message: expectedMessages[i][0],
                 minute: expectedMessages[i][1],
             }
+            assert.deepStrictEqual(actualMessages[i].message, expected.message)
+            // Avoid flakiness in the timing by looking within a minute rather than exact.
+            assert.ok(
+                Math.abs(actualMessages[i].minute - expected.minute) <= 1,
+                `Expected to be within 60 seconds of minute ${expected.minute}, but instead was at minute ${actualMessages[i].minute}`
+            )
             assert.deepStrictEqual(actualMessages[i], expected)
         }
     }


packages/core/src/testInteg/codecatalyst/devEnv.test.ts
```diff
@@ -141,6 +141,12 @@ describe('InactivityMessage', function () {
                 message: expectedMessages[i][0],
                 minute: expectedMessages[i][1],
             }
+            assert.deepStrictEqual(actualMessages[i].message, expected.message)
+            // Avoid flakiness in the timing by looking within a minute rather than exact.
+            assert.ok(
+                Math.abs(actualMessages[i].minute - expected.minute) <= 1,
+                `Expected to be within 60 seconds of minute ${expected.minute}, but instead was at minute ${actualMessages[i].minute}`
+            )
             assert.deepStrictEqual(actualMessages[i], expected)
```
justinmk3, line 150: should drop this line, right?

justinmk3 (approved):

packages/core/src/testInteg/codecatalyst/devEnv.test.ts
Hweinstock: I was intending this line to check if the time the message popped up was within a minute of the expected time. Do you think it would be better to not check the time the message came up at all?

justinmk3: If we don't remove "something", the test will still fail, no?

Hweinstock: I'm making the check more lenient. The logs in the issue show that the message is popping up at minute 54, when we expect it at 55. If that happens, the current logic should pass now since its within one minute of the expected time.

jpinkney-aws: If I understand correctly he's confused about why we aren't removing 
```
assert.deepStrictEqual(actualMessages[i], expected)
```
since 
```
assert.ok(Math.abs(actualMessages[i].minute - expected.minute) <= 1, 'Expected to be within 60 seconds of minute ${expected.minute}, but instead was at minute ${actualMessages[i].minute}')
```
is meant to replace it. If we kept the original check (`assert.deepStrictEqual(actualMessages[i], expected)`), wouldn't that check still fail since that piece of code is also checking the time?

Hweinstock: Oh my bad. Yes, we should remove that line. Sorry was looking at the wrong line and totally missed it.

PR #6275, opened 2024-12-18
fix(amazonq): respond in chat right away when user enters JDK path

dhasani23:
## Problem

On a VS Code instance backed by CloudFront, when users enter their JDK path, it takes our chat 30+ seconds to respond.

## Solution

Send a notification just before the build starts to also allow the other multiple chat messages to be sent.

---

### DIFF
--- a/packages/core/src/codewhisperer/service/transformByQ/transformMavenHandler.ts
+++ b/packages/core/src/codewhisperer/service/transformByQ/transformMavenHandler.ts
@@ -13,7 +13,6 @@ import { CodeTransformTelemetryState } from '../../../amazonqGumby/telemetry/cod
 import { ToolkitError } from '../../../shared/errors'
 import { writeLogs } from './transformFileHandler'
 import { throwIfCancelled } from './transformApiHandler'
-import { sleep } from '../../../shared/utilities/timeoutUtils'
 
 // run 'install' with either 'mvnw.cmd', './mvnw', or 'mvn' (if wrapper exists, we use that, otherwise we use regular 'mvn')
 function installProjectDependencies(dependenciesFolder: FolderInfo, modulePath: string) {
@@ -109,8 +108,8 @@ function copyProjectDependencies(dependenciesFolder: FolderInfo, modulePath: str
 }
 
 export async function prepareProjectDependencies(dependenciesFolder: FolderInfo, rootPomPath: string) {
-    // sleep for 0.5s to allow QCT to 1) process JAVA_HOME path and 2) send multiple chat messages
-    await sleep(500)
+    // use await to allow other chat messages to send as well, before the build starts
+    await vscode.window.showInformationMessage(CodeWhispererConstants.buildStartedNotification)
     try {
         copyProjectDependencies(dependenciesFolder, rootPomPath)
     } catch (err) {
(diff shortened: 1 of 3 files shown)


dhasani23:
https://github.com/user-attachments/assets/92884865-b22d-4fc5-b81f-a59afa33ed1b

https://github.com/user-attachments/assets/08b5d58d-9fe6-4524-b253-ff8cc3adb7c4

ntarakad-aws (approved):

justinmk3:
> Sleep for 0.5s to allow the JDK path to be processed (i.e. we make sure it exists)

IIUC, this means: users on fast systems must always wait 500ms? 

And, what happens on systems where it takes longer than 500ms?

packages/core/src/codewhisperer/service/transformByQ/transformMavenHandler.ts
```diff
@@ -108,6 +108,8 @@ function copyProjectDependencies(dependenciesFolder: FolderInfo, modulePath: str
 }
 
 export async function prepareProjectDependencies(dependenciesFolder: FolderInfo, rootPomPath: string) {
+    // use await to allow other chat messages to send as well, before the build starts
+    await vscode.window.showInformationMessage(CodeWhispererConstants.buildStartedNotification)
```
dhasani23, line 112: We used to have this here actually, months ago, but somehow it must have been removed probably in favor of the chat. For most things, right now `/transform` shows updates both as notifications and as chat messages, so this is fine, and we should be sending this notification here anyway.

It seems that `copyProjectDependencies` and `installProjectDependencies` prevent the numerous chat messages from displaying before they begin, so we need some kind of delay just before they execute to allow all of these chat messages to be sent before the copy / install functions start. Something like `await sleep(500)` will sleep for 500ms for all users like you mentioned, but adding this notification here seems to also solve the problem. Not yet sure if this is a perfect solution, but again, we should be showing this notification here anyway.

packages/core/src/codewhisperer/service/transformByQ/transformMavenHandler.ts
```diff
@@ -108,6 +108,7 @@ function copyProjectDependencies(dependenciesFolder: FolderInfo, modulePath: str
 }
 
 export async function prepareProjectDependencies(dependenciesFolder: FolderInfo, rootPomPath: string) {
+    void vscode.window.showInformationMessage(CodeWhispererConstants.buildStartedNotification)
```
dhasani23, line 111: We need this here anyway as an unrelated change, but still need to figure out a better solution besides sleeping for half a second

dhasani23: It looks like the few chat messages which should be showing up one-at-a-time before the `copyProjectDependencies` and `installProjectDependencies` functions below start, are actually showing up all at once after both of these functions complete. I guess that means these 2 functions are blocking the event loop, which is why adding a small delay before they begin provides enough time for the queued chat messages to send. Haven't been able to find another solution besides adding this delay.

Downsides are that it does delay for all users, and unclear if 500ms will always be enough of a delay.

dhasani23:
@justinmk3 closed this PR because it will require some more thought; can re-open later as needed. Wondering if you have thoughts on resolving this type of issue though? I need some kind of delay right before `copyProjectDependencies` and refactoring to introduce one doesn't seem feasible, but I also don't want to sleep for all users for an arbitrary duration.

I considered using `UiKind` to determine when to sleep, but looks like that's empty/unavailable for VS Code instances hosted on CloudFront (ex. https://d2dbfjjr62djz3.cloudfront.net/?folder=/home/ubuntu/airline-booking-main/), and otherwise is `Desktop` or `Web`.

PR #6277, opened 2024-12-19
build(lint): migrate away from forEach (WIP)

Hweinstock:
## Problem
forEach can lead to race conditions with async methods, and is often used instead of `map`, `find`, `some`, `reduce`, `flatMap`, etc...

## Solution
Add a lint rule for forEach and migrate away from it. 

---


.eslintrc.js
```diff
@@ -178,6 +178,7 @@ module.exports = {
         'aws-toolkits/no-console-log': 'error',
         'aws-toolkits/no-json-stringify-in-log': 'error',
         'aws-toolkits/no-printf-mismatch': 'error',
+        'aws-toolkits/no-foreach': 'error',
```
justinmk3, line 181: We depend on `eslint-plugin-unicorn` so I think we can use this rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md

packages/core/src/codewhisperer/client/codewhisperer.ts
```diff
@@ -227,18 +227,18 @@ export class DefaultCodeWhispererClient {
...
+        return pageableToCollection(requester, {}, 'nextToken' as never)
             .promise()
             .then((resps) => {
                 let logStr = 'amazonq: listAvailableCustomizations API request:'
-                resps.forEach((resp) => {
+                for (const resp of resps) {
                     const requestId = resp.$response.requestId
                     logStr += `\n${indent('RequestID: ', 4)}${requestId},\n${indent('Customizations:', 4)}`
-                    resp.customizations.forEach((c, index) => {
+                    for (const [c, index] of enumerate(resp.customizations)) {
```
justinmk3, line 237: Does using enumerate() mean we are iterating the collection twice? The `index` might be a good argument for using `forEach` in this case.

.eslintrc.js
Hweinstock: Good idea! That rule also comes with a way to auto-fix _most_ cases so I switched over to it completely here: https://github.com/aws/aws-toolkit-vscode/pull/6281

Hweinstock:
Moved to https://github.com/aws/aws-toolkit-vscode/pull/6281

PR #6279, opened 2024-12-20
test(amazonq): Add e2e tests for general amazon q chat panel

jpinkney-aws:
## Problem
- We want tests for the general amazon q panel
- We want other teams to quickly be able to write tests for their agents

## Solution
- Add general tests for the amazon q chat panel
- Add a template that other teams can use to write their tests
- Fix an issue that occurred only in tests where tab id's weren't defined for the help message. The problem was that in the framework everything is instant which meant a tab id was defined before that code ran and was never passed through to processQuickActionCommand
- By default don't show the welcome page for now tests

---


packages/amazonq/test/e2e/amazonq/framework/framework.ts
```diff
@@ -29,7 +29,7 @@ export class qTestingFramework {
         featureName: TabType,
         amazonQEnabled: boolean,
         featureConfigsSerialized: [string, FeatureContext][],
-        welcomeCount = 0
+        welcomeCount = 10 // by default don't show the welcome page
```
nkomonen-amazon, line 32: If this number is hardcoded in the prod code, might be better to export a common constant

nkomonen-amazon (approved):

packages/core/src/amazonq/index.ts
```diff
@@ -26,6 +26,7 @@ export { init as testChatAppInit } from '../amazonqTest/app'
 export { init as docChatAppInit } from '../amazonqDoc/app'
 export { activateBadge } from './util/viewBadgeHandler'
 export { amazonQHelpUrl } from '../shared/constants'
+export * from './webview/ui/texts/constants'
```
justinmk3, line 29: generally a good idea to give a name for these index.ts interfaces. that avoids conflicts, and also encourages useful "namespacing" in the consumer.
```suggestion
export * as webviewConstants from './webview/ui/texts/constants'
```

justinmk3 (approved):

PR #6281, opened 2024-12-20
build(lint): add `no-array-for-each` and migrate existing cases. 

Hweinstock:
## Problem
Many bugs, and confusing behavior stem from developers using an async function within a forEach. This is almost always not desired. Additionally, `forEach` can lead to bloated implementations when other methods like `some`, `find`, `reduce`, or `flatMap` are more applicable. 

According to https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md It is also faster with `for ... of`. 

## Solution
- Add lint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md. 
- Automatically migrate existing cases with eslint CLI --fix. (able to migrate ~75% of cases). 
- For remaining cases, mark as exception.
- Fix a few small lint problems as a result. 

---

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -158,6 +158,7 @@ module.exports = {
         'unicorn/prefer-reflect-apply': 'error',
         'unicorn/prefer-string-trim-start-end': 'error',
         'unicorn/prefer-type-error': 'error',
+        'unicorn/no-array-for-each': 'error',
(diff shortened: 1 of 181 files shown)


justinmk3 (approved):
🚀

PR #6284, opened 2024-12-21
telemetry(amazonq): add metric for transformation type

dhasani23:
## Problem

Be able to tell when a language upgrade vs sql conversion happens.

## Solution

Add metric.

---

### DIFF
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -680,6 +680,14 @@ export class GumbyController {
 
             case ConversationState.WAITING_FOR_TRANSFORMATION_OBJECTIVE: {
                 const objective = data.message.trim().toLowerCase()
+                // since we're prompting the user, their project(s) must be eligible for both types of transformations, so track how often this happens here
+                if (objective === 'language upgrade' || objective === 'sql conversion') {
+                    telemetry.codeTransform_submitSelection.emit({
+                        codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId(),
+                        userChoice: objective,
+                        result: 'Succeeded',
+                    })
+                }
                 if (objective === 'language upgrade') {
                     await this.handleLanguageUpgrade(data)
                 } else if (objective === 'sql conversion') {
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -269,29 +269,41 @@ export class GumbyController {
     }
 
     private async handleLanguageUpgrade(message: any) {
-        try {
-            await this.beginTransformation(message)
-            const validProjects = await this.validateLanguageUpgradeProjects(message)
-            if (validProjects.length > 0) {
-                this.sessionStorage.getSession().updateCandidateProjects(validProjects)
-                await this.messenger.sendLanguageUpgradeProjectPrompt(validProjects, message.tabID)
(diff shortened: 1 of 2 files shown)


packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -668,14 +680,6 @@ export class GumbyController {
 
             case ConversationState.WAITING_FOR_TRANSFORMATION_OBJECTIVE: {
                 const objective = data.message.trim().toLowerCase()
-                // since we're prompting the user, their project(s) must be eligible for both types of transformations, so track how often this happens here
-                if (objective === 'language upgrade' || objective === 'sql conversion') {
-                    telemetry.codeTransform_submitSelection.emit({
```
dhasani23, line 673: Better to emit this above since those functions are used in multiple places, so remove it from here.

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -673,7 +668,7 @@ export async function getValidSQLConversionCandidateProjects() {
             `CodeTransformation: found ${embeddedSQLProjects.length} projects with embedded SQL statements`
         )
         telemetry.record({
-            codeTransformMetadata: resultLog,
+            codeTransformMetadata: `Found ${embeddedSQLProjects.length} projects with embedded SQL`,
```
dhasani23, line 671: Just emit a short phrase rather than entire result

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -269,29 +269,41 @@ export class GumbyController {
...
     private async handleLanguageUpgrade(message: any) {
-        try {
-            await this.beginTransformation(message)
-            const validProjects = await this.validateLanguageUpgradeProjects(message)
-            if (validProjects.length > 0) {
-                this.sessionStorage.getSession().updateCandidateProjects(validProjects)
-                await this.messenger.sendLanguageUpgradeProjectPrompt(validProjects, message.tabID)
+        await telemetry.codeTransform_submitSelection.run(async () => {
+            telemetry.record({
+                codeTransformSessionId: CodeTransformTelemetryState.instance.getSessionId(),
```
dhasani23, line 274: Looks like session ID is mostly if not always required in our metrics; not sure why because I don't use it much. I'll fix this in the commons package and then submit a follow up PR here along with some other telemetry-related improvements shortly.

packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -269,29 +269,41 @@ export class GumbyController {
...
+            })
+            try {
+                await this.beginTransformation(message)
+                const validProjects = await this.validateSQLConversionProjects(message)
+                if (validProjects.length > 0) {
+                    this.sessionStorage.getSession().updateCandidateProjects(validProjects)
+                    await this.messenger.sendSelectSQLMetadataFileMessage(message.tabID)
+                }
+            } catch (err: any) {
+                getLogger().error(`Error handling SQL conversion: ${err}`)
```
justinmk3, line 304: Note: if the exception is caught then `run()` won't reflect a `result=Failed`. 

If `handleSQLConversion` is called from a command or webview, the extension global handler will log the error and show a useful message to the user. If not, then the next best approach is to let the exception through, but add `.catch(...)` to the `run()`. call.

Same for the other case in this PR.

hayemaxi (approved):

PR #6293, opened 2024-12-26
telemetry(amazonq): attach stack trace in onCodeGeneration failures

siakmun-aws:
## Problem
To further find issues in onCodeGeneration, need to attach stack trace when sending metrics

## Solution
Attach stack trace when sending metrics

---

### DIFF
--- a/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
+++ b/packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
@@ -501,6 +501,10 @@ export class FeatureDevController {
                         result = MetricDataResult.Fault
                     }
                     break
+                case ContentLengthError.name:
+                case MonthlyConversationLimitError.name:
+                case CodeIterationLimitError.name:
+                case UploadURLExpired.name:
                 case PromptRefusalException.name:
                 case NoChangeRequiredException.name:
                     result = MetricDataResult.Error
@@ -510,7 +514,7 @@ export class FeatureDevController {
                     break
             }
 
-            await session.sendMetricDataTelemetry(MetricDataOperationName.EndCodeGeneration, result)
+            await session.sendMetricDataTelemetry(MetricDataOperationName.EndCodeGeneration, result, err.stack)
             throw err
         } finally {
             // Finish processing the event
--- a/packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
+++ b/packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
@@ -463,6 +463,10 @@ describe('Controller', () => {
 
             const errorResultMapping = new Map([
                 ['EmptyPatchException', MetricDataResult.LlmFailure],
+                [ContentLengthError.name, MetricDataResult.Error],
+                [MonthlyConversationLimitError.name, MetricDataResult.Error],
+                [CodeIterationLimitError.name, MetricDataResult.Error],
+                [UploadURLExpired.name, MetricDataResult.Error],
                 [PromptRefusalException.name, MetricDataResult.Error],
                 [NoChangeRequiredException.name, MetricDataResult.Error],
             ])
@@ -518,7 +522,13 @@ describe('Controller', () => {
                 )
                 const metricResult = getMetricResult(error)
                 assert.ok(
-                    sendMetricDataTelemetrySpy.calledWith(MetricDataOperationName.EndCodeGeneration, metricResult)
+                    sendMetricDataTelemetrySpy.calledWith(
+                        MetricDataOperationName.EndCodeGeneration,
+                        metricResult,
+
+                        // Stack trace should include the name of the test framework
+                        sinon.match((str) => typeof str === 'string' && str.includes('Mocha'))
+                    )
                 )
             }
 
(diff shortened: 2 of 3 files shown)


packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
```diff
@@ -518,7 +522,13 @@ describe('Controller', () => {
                 )
                 const metricResult = getMetricResult(error)
                 assert.ok(
-                    sendMetricDataTelemetrySpy.calledWith(MetricDataOperationName.EndCodeGeneration, metricResult)
+                    sendMetricDataTelemetrySpy.calledWith(
+                        MetricDataOperationName.EndCodeGeneration,
+                        metricResult,
+
+                        // Stack trace should include the name of the test framework
+                        sinon.match((str) => typeof str === 'string' && str.includes('Mocha'))
```
neilk-aws, line 530: Can we assert the useful part of the log message is actually present? This does not guarantee that the stack trace is actually empty.

neilk-aws (approved):

packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -1009,4 +1019,55 @@ export class FeatureDevController {
...
         }
     }
+
+    // Should include error messages only for whitelisted exceptions
+    // i.e. exceptions with deterministic error messages and do not include sensitive data
+    private getStackTraceForError(error: Error): string {
+        const seenExceptions = new Set<Error>()
+        const lines: string[] = []
+
+        function printExceptionDetails(err: Error, prefix: string = '') {
```
neilk-aws, line 1029: Can we add a recursion limit to this function to avoid any large or blocking exceptions that lack a good base case?

packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
```diff
@@ -519,7 +523,11 @@ describe('Controller', () => {
                 )
                 const metricResult = getMetricResult(error)
                 assert.ok(
-                    sendMetricDataTelemetrySpy.calledWith(MetricDataOperationName.EndCodeGeneration, metricResult)
+                    sendMetricDataTelemetrySpy.calledWith(
+                        MetricDataOperationName.EndCodeGeneration,
+                        metricResult,
+                        sinon.match((str) => typeof str === 'string' && str.includes('stack trace: '))
```
neilk-aws, line 529: Can we assert more of the contents of the stack trace? Otherwise, there is nothing else really that is testing the method that you added

justinmk3 (changes requested):
This is lacking an acceptable explanation of what the purpose is, and how it works. Did you test this with a production webpack'd build? Nodejs stacktraces in the production build are usually obfuscated.

The docstring of the function does not give useful insights. 

The function is non-trivial, and has no test coverage. 

Assuming the function works, it belongs in a core util module, not randomly in the middle of a feature module.

siakmun-aws:
> This is lacking an acceptable explanation of what the purpose is, and how it works. Did you test this with a production webpack'd build? Nodejs stacktraces in the production build are usually obfuscated.
> 
> The docstring of the function does not give useful insights.
> 
> The function is non-trivial, and has no test coverage.
> 
> Assuming the function works, it belongs in a core util module, not randomly in the middle of a feature module.

Here is some background on the purpose of this PR. The previous (main branch) client has been able to send telemetry in the event of code generation failure. However, any uncaught errors, either toolkit or runtime error etc., will only result in sending a "Fault" telemetry. We want to be able to diagnose such errors, and therefore the stack trace is needed. TS obfuscation is expected since non-toolkit errors such as runtime errors are not obfuscated. Uncaught toolkit error names are obfuscated, but through the function names (not obfuscated) in the stack trace combined with the code line number and toolkit version, the cause can still be identified.

Docstring: modified since last commit to give more details
Test: added
File place: changed to shared/error.ts

packages/core/src/shared/errors.ts
```diff
@@ -368,6 +368,57 @@ export function getTelemetryResult(error: unknown | undefined): Result {
     return 'Failed'
 }
 
+export class SafeMessageToolkitError extends ToolkitError {
+    public constructor(message: string, info: ErrorInformation = {}) {
+        super(message, info)
```
justinmk3, line 373: what is going on here? why was this new base class introduced? what is the intended meaning of "safe message" ?

siakmun-aws: Safe message means the error message is safe to send to telemetry (no critical information). This was introduced to allow error classes which extends it can be sent to telemetry with its error message.
But this PR is not ready yet, so I will draft it for now.

justinmk3: That doesn't sound like the right direction. We already have `scrubNames` https://github.com/aws/aws-toolkit-vscode/blob/7d5a339c4334d6ac4bb3123b81b0073286b82dfe/packages/core/src/shared/errors.ts#L378

which removes PII from strings. And that is used by `getTelemetryReasonDesc`, which already is used automatically.

Adding new, special classes is not the right way to approach this.

PR #6304, opened 2025-01-03
perf(childprocess): spawned processes are tracked and monitored. 

Hweinstock:
## Problem
Toolkit/Q slow performance is consistently reported by customers: [https://github.com/aws/aws-toolkit-vscode/issues?q=is%3Aopen+is%3Aissue+label%3Aperformance ](https://github.com/aws/aws-toolkit-vscode/issues?q=is%3Aopen+is%3Aissue+label%3Aperformance)

Toolkit/Q features can spawn processes, included many long-lived processes such as LSP servers. The cost of these processes is hidden and hard to troubleshoot. Any misbehaving process will not be noticed except in a coarse "Toolkit/Q causes vscode to be slow" manner.

## Solution
- Track all running processes in a map-like data structure. 
- use a `PollingSet` to periodically "monitor" their system usage(%cpu, memory bytes), logging warning messages when it exceeds a threshold. 
- Add developer command to instantiate spawned processes via `ChildProcess` wrapper to make this easier to test. 

image
image

## Future Work 
- Built on this: (https://github.com/Hweinstock/aws-toolkit-vscode/pull/8)
- Integrate telemetry into the tracker instead of only logging.
- Add public-facing (non-dev) command so that users can view active processes and their usages.

---


packages/core/src/shared/utilities/pollingSet.ts
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { globals } from '..'
+import globals from '../extensionGlobals'
```
Hweinstock, line 6: avoids circular dependency.

packages/core/src/shared/utilities/pollingSet.ts
```diff
@@ -44,9 +44,14 @@ export class PollingSet<T> extends Set<T> {
             this.clearTimer()
         }
     }
-
+    // TODO(hkobew): Overwrite the add method instead of adding seperate method. If we add item to set, timer should always start.
```
Hweinstock, line 47: plan to do as quick follow up, since this involves touching some other files unrelated to process work.

packages/core/src/shared/utilities/pollingSet.ts
nkomonen-amazon: We will probably want to add a lint warning.

But for context what is happening is that `..` implicitly imports the `index.ts` module instead of the intended one and that causes the circular dependency issue.

packages/core/src/shared/utilities/processUtils.ts
```diff
@@ -61,14 +63,142 @@ export interface ChildProcessResult {
...
+    public clear() {
+        for (const childProcess of this.#processByPid.values()) {
+            childProcess.stop(true)
+        }
+        this.#pids.clear()
+        this.#processByPid.clear()
+    }
+
+    public async getUsage(pid: number): Promise<ProcessStats> {
+        const getWindowsUsage = () => {
```
nkomonen-amazon, line 147: nit: Instead of `const getWindowsUsage...` you can do `function getWindowsUsage() { ... }` and move it to after the logic of getUsage(). I find it cleaner.

packages/core/src/shared/utilities/pollingSet.ts
Hweinstock: Is there every a case where importing from `index.ts` is preferred? Or is this a pretty straight-forward "don't do this" pattern?

Hweinstock: PR for lint rule: https://github.com/aws/aws-toolkit-vscode/pull/6372

packages/core/src/dev/activation.ts
```diff
@@ -44,6 +45,7 @@ export type DevFunction =
     | 'editAuthConnections'
     | 'notificationsSend'
     | 'forceIdeCrash'
+    | 'startChildProcess'
```
justinmk3, line 48: Cool, I love that we keep enhancing dev-mode.

packages/core/src/shared/utilities/processUtils.ts
```diff
@@ -61,14 +63,144 @@ export interface ChildProcessResult {
...
+        this.#pids.clear()
+        this.#processByPid.clear()
+    }
+
+    public getUsage(pid: number): ProcessStats {
+        try {
+            // isWin() leads to circular dependency.
+            return process.platform === 'win32' ? getWindowsUsage() : getUnixUsage()
+        } catch (e) {
+            ChildProcessTracker.logger.warn(`Failed to get process stats for ${pid}: ${e}`)
```
justinmk3, line 151: If it ever fails we should probably not attempt on future invocations?

justinmk3 (approved):
This is a great starting point. Merging it so it can start to "bake".

Followup ideas:
- include perf results in telemetry (do we have an existing metric it can be attached to?)
- dev-mode can show a "report" of process stats
- we need some way for non-dev-mode users to get a report of the Toolkit "health", for troubleshooting and bug reports. this would be part of that. the current answer is the `About` command, but that's too limited.

packages/core/src/shared/utilities/processUtils.ts
Hweinstock: Not really sure what could cause it to fail here. Do you think there is a risk in continually retrying? Maybe potential log spam? Could this be combatted by setting the polling interval longer?

Hweinstock:
Follow up to this work: https://github.com/aws/aws-toolkit-vscode/pull/6394

packages/core/src/shared/utilities/pollingSet.ts
Hweinstock: https://github.com/aws/aws-toolkit-vscode/pull/6396

packages/core/src/shared/utilities/processUtils.ts
justinmk3: > Not really sure what could cause it to fail here.

Permissions issues usually. 

> Maybe potential log spam?

Yeah that's at least my first thought. When a system is in a bad state, it can make things worse if logs "cascade" with a deluge of redundant warnings.

PR #6310, opened 2025-01-06
feat(amazonq): Auto update language servers when new versions are available

jpinkney-aws:
## Problem:
- Only one version of a language server is installed right now

## Solution:
- Automatically update when a new version is available in the manifest

## TODO
- install(manifest) should be install(version) which would clean some things up
- aws.toolkit.lsp.manifest should have named manifests, since every language server could have a different manifest
- e2e tests for actually downloading manifests and checking that they're ready and available
- figure out if we only update when vscode restarts or how we want to handle that
- re-add cleanup
- telemetry for downloading/updating the language server

---


packages/core/src/test/shared/languageServer/lspManager.test.ts
```diff
@@ -0,0 +1,147 @@
...
+        sandbox = sinon.createSandbox()
+        tempFolder = await makeTemporaryToolkitFolder()
+    })
+
+    afterEach(async () => {
+        sandbox.restore()
+        await fs.delete(tempFolder)
+    })
+
+    describe('download', () => {
```
nkomonen-amazon, line 43: These tests can likely be made obsolete by testing the same scenarios in tests for `AmazonQLSPInstaller`.
Stubbing the http fetcher should be enough to test the entire flow, and then you could have complex tests where it first downloads from the remote, then a subsequent call should instead use the cache

jpinkney-aws: yeah, the new tests are much simpler

packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
```diff
@@ -201,6 +202,50 @@ export class HttpResourceFetcher implements ResourceFetcher {
     }
 }
 
+export class RetryableResourceFetcher extends HttpResourceFetcher {
```
jpinkney-aws, line 205: I think this should probably be done in a different way. Maybe the httpResourceFetcher can have retries in it? I think we can re-visit this after this PR

packages/core/src/shared/vscode/env.ts
```diff
@@ -270,3 +271,14 @@ export async function getMachineId(): Promise<string> {
     // TODO: check exit code.
     return (await proc.run()).stdout.trim() ?? 'unknown-host'
 }
+
+export function getApplicationSupportFolder() {
+    switch (process.platform) {
```
jpinkney-aws, line 276: TODO support more platforms and their fallback locations

packages/core/src/amazonq/lsp/workspaceInstaller.ts
```diff
@@ -0,0 +1,28 @@
...
+    async resolve(): Promise<LspResult> {
+        const name = 'AmazonQ-Workspace'
+        const manifest = await new ManifestResolver(manifestUrl, name).resolve()
+        const installationResult = await new LanguageServerResolver(
+            manifest,
+            name,
+            new Range(supportedLspServerVersions)
+        ).resolve()
+
+        // TODO Cleanup old versions of language servers
```
jpinkney-aws, line 25: Calling this out explicitly so we don't forget

packages/core/src/shared/languageServer/lspResolver.ts
```diff
@@ -0,0 +1,352 @@
...
+     */
+    private async downloadRemoteTargetContent(contents: TargetContent[], version: string) {
+        const downloadDirectory = this.getDownloadDirectory(version)
+
+        if (!(await fs.existsDir(downloadDirectory))) {
+            await fs.mkdir(downloadDirectory)
+        }
+
+        const downloadTasks = contents.map(async (content) => {
+            // TODO This should be using the retryable http library but it doesn't seem to support zips right now
```
jpinkney-aws, line 167: Calling this out as well, we should switch to something that has retries with backoffs

packages/core/src/test/shared/languageServer/lspManager.test.ts
jpinkney-aws: yeah now that the installer exists I can bring them up a level. I'll add that as a followup item

nkomonen-amazon (approved):
Approved as this code does not impact prod and we can do fast follow ups

packages/core/src/shared/languageServer/lspResolver.ts
```diff
@@ -0,0 +1,352 @@
+/*!
```
justinmk3, line 1: can these modules live in `shared/lsp/` instead of `shared/languageServer/` ? helps discovery.

packages/core/src/shared/languageServer/lspResolver.ts
```diff
@@ -0,0 +1,352 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import fs from '../fs/fs'
+import { ToolkitError } from '../errors'
+import * as semver from 'semver'
+import * as path from 'path'
+import { FileType } from 'vscode'
+import AdmZip from 'adm-zip'
```
justinmk3, line 11: We have a zip.js-based ZipStream module now. We should avoid adm-zip , because we want to remove it.

PR #6312, opened 2025-01-06
fix(amazonq): Skip including deleted files for FeatureDev context.

neilk-aws:
## Problem
When zipping context for /dev, customer build processes (file watchers, etc.) may delete build artifacts we’ve already enumerated but have not added to the archive. As a best practice, customers should `.gitignore` these types of files, but in the event they don't, this has the potential to block /dev from running.

## Solution
Skip affected files, which are not found when adding to zip context for Feature Dev.

---

### DIFF
--- a/packages/core/src/amazonqFeatureDev/util/files.ts
+++ b/packages/core/src/amazonqFeatureDev/util/files.ts
@@ -13,7 +13,7 @@ import { getLogger } from '../../shared/logger/logger'
 import { maxFileSizeBytes } from '../limits'
 import { createHash } from 'crypto'
 import { CurrentWsFolders } from '../types'
-import { ToolkitError } from '../../shared/errors'
+import { hasCode, ToolkitError } from '../../shared/errors'
 import { AmazonqCreateUpload, Span, telemetry as amznTelemetry } from '../../shared/telemetry/telemetry'
 import { TelemetryHelper } from './telemetryHelper'
 import { maxRepoSizeBytes } from '../constants'
@@ -39,7 +39,16 @@ export async function prepareRepoData(
         const ignoredExtensionMap = new Map<string, number>()
 
         for (const file of files) {
-            const fileSize = (await fs.stat(file.fileUri)).size
+            let fileSize
+            try {
+                fileSize = (await fs.stat(file.fileUri)).size
+            } catch (error) {
+                if (hasCode(error) && error.code === 'ENOENT') {
+                    // No-op: Skip if file does not exist
+                    continue
+                }
+                throw error
+            }
             const isCodeFile_ = isCodeFile(file.relativeFilePath)
 
             if (fileSize >= maxFileSizeBytes || !isCodeFile_) {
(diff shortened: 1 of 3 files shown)


ctidd (approved):

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -58,7 +58,17 @@ export async function prepareRepoData(
             totalBytes += fileSize
 
             const zipFolderPath = path.dirname(file.zipFilePath)
-            zip.addLocalFile(file.fileUri.fsPath, zipFolderPath)
+
+            try {
+                zip.addLocalFile(file.fileUri.fsPath, zipFolderPath)
+            } catch (error) {
```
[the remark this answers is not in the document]
neilk-aws: We are relying on error handling to mitigate a race condition where the file is deleted as it is being referenced. If we do an `fs.exists` check and immediately after the file is deleted by (manually or through some clean up commands a customer is running), the `addLocalFile` will fail.

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -39,7 +39,16 @@ export async function prepareRepoData(
         const ignoredExtensionMap = new Map<string, number>()
 
         for (const file of files) {
-            const fileSize = (await fs.stat(file.fileUri)).size
+            let fileSize
+            try {
+                fileSize = (await fs.stat(file.fileUri)).size
+            } catch (error) {
+                if (error instanceof Error && error.message.includes('ENOENT')) {
```
justinmk3, line 46: use `hasCode` and check the `code`.  https://github.com/aws/aws-toolkit-vscode/blob/3d9077264f9d40d7e80383c3c1e5a4fd5e2181a9/packages/core/src/shared/errors.ts#L602

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -58,7 +67,17 @@ export async function prepareRepoData(
...
 
             const zipFolderPath = path.dirname(file.zipFilePath)
-            zip.addLocalFile(file.fileUri.fsPath, zipFolderPath)
+
+            try {
+                zip.addLocalFile(file.fileUri.fsPath, zipFolderPath)
+            } catch (error) {
+                if (error instanceof Error && error.message.includes('File not found')) {
+                    // No-op: Skip if file was deleted or does not exist
+                    // Reference: https://github.com/cthackers/adm-zip/blob/1cd32f7e0ad3c540142a76609bb538a5cda2292f/adm-zip.js#L296-L321
```
justinmk3, line 76: does it set a `code` ?

neilk-aws: This error does not set a code: https://github.com/cthackers/adm-zip/blob/1cd32f7e0ad3c540142a76609bb538a5cda2292f/util/errors.js#L56

neilk-aws:
> Is there a way we can test this behavior? Something along the lines of having `collectFiles` return some files that don't exist when we try to access them?

This is mitigating a situation where there is a race condition and the file is deleted as they are being zipped. In this case, `collectFiles` would not report any files do not exist, and we would only discover it when they are being looked up.

justinmk3 (approved):

PR #6325, opened 2025-01-09
feat(amazonq): Add lsp downloading support for windows/linux

jpinkney-aws:
## Problem
- Language servers are only being downloaded on macs

## Solution
- Download on windows/linux as well

---

### DIFF
--- a/packages/core/src/shared/fs/fs.ts
+++ b/packages/core/src/shared/fs/fs.ts
@@ -16,7 +16,7 @@ import {
     isPermissionsError,
     scrubNames,
 } from '../errors'
-import globals from '../extensionGlobals'
+import globals, { isWeb } from '../extensionGlobals'
 import { isWin } from '../vscode/env'
 import { resolvePath } from '../utilities/pathUtils'
 import crypto from 'crypto'
@@ -549,6 +549,15 @@ export class FileSystem {
      * Follows the cache_dir convention outlined in https://crates.io/crates/dirs
      */
     getCacheDir(): string {
+        if (isWeb()) {
+            const homeDir = this.#homeDir
+            if (!homeDir) {
+                throw new ToolkitError('Web home directory not found', {
+                    code: 'WebHomeDirectoryNotFound',
+                })
+            }
+            return homeDir
+        }
         switch (process.platform) {
             case 'darwin': {
                 return _path.join(this.getUserHomeDir(), 'Library/Caches')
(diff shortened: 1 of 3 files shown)


nkomonen-amazon (approved):

packages/core/src/shared/vscode/env.ts
```diff
@@ -277,8 +277,14 @@ export function getApplicationSupportFolder() {
         case 'darwin': {
             return path.join(os.homedir(), 'Library/Application Support')
         }
+        case 'win32': {
+            return process.env.APPDATA
+        }
+        case 'linux': {
+            return process.env.XDG_CACHE_HOME || path.join(os.homedir(), '.cache')
```
justinmk3, line 284: Should use `fs.getUserHomeDir`, it handles some edge cases. And to avoid the circular dependency, `getApplicationSupportFolder` should live in `fs.ts` (and that makes sense anyways, because this is a filesystem decision).

It also seems like `getApplicationSupportFolder` is getting a cache dir, so maybe `getCacheDir` is a good name.

justinmk3: `tryGetFilepathEnvVar` may also be worth looking at in the future. https://github.com/aws/aws-toolkit-vscode/blob/02f6d0bf817253f8eaf1ad6f13c333d2562f0a9c/packages/core/src/shared/fs/fs.ts#L517

packages/core/src/shared/fs/fs.ts
```diff
@@ -543,6 +543,34 @@ export class FileSystem {
         return this.#homeDir
     }
 
+    /**
```
Hweinstock, line 546: Nice, I like that this is in a shared module now.

Hweinstock (approved):

packages/core/src/shared/fs/fs.ts
```diff
@@ -543,6 +543,34 @@ export class FileSystem {
...
+                        code: 'LocalAppDataNotFound',
+                    })
+                }
+                return localAppData
+            }
+            case 'linux': {
+                return _path.join(this.getUserHomeDir(), '.cache')
+            }
+            default: {
+                throw new Error(`Unsupported platform: ${process.platform}. Expected 'darwin', 'win32', or 'linux'.`)
```
justinmk3, line 569: for web should we return something? we support "home dir" on web: https://github.com/aws/aws-toolkit-vscode/blob/38842593a9e27dd00c750edfbc4b99aa7e4c1cd8/packages/core/src/shared/fs/fs.ts#L497-L502

justinmk3 (approved):

packages/core/src/shared/fs/fs.ts
jpinkney-aws: Just added it. It's probably useful to have if I need it for web support in the next couple weeks

PR #6332, opened 2025-01-10
ci(amazonq): fix linting issue on language server child process

jpinkney-aws:
## Problem
In a recent PR we disabled using child_process in favour of our internal one. This started flagging lsp/client.ts, since it wasn't skipped.

## Solution
Skip eslint for child_process in this case, since the language server options expect the actual child process. The same is done for the workspace context language server: https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/amazonq/lsp/lspClient.ts#L13

---

### DIFF
--- a/packages/amazonq/src/lsp/client.ts
+++ b/packages/amazonq/src/lsp/client.ts
@@ -5,7 +5,7 @@
 
 import vscode, { env, version } from 'vscode'
 import * as nls from 'vscode-nls'
-import * as cp from 'child_process'
+import * as cp from 'child_process' // eslint-disable-line no-restricted-imports -- language server options expect actual child process
 import * as crypto from 'crypto'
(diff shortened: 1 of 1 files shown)


nkomonen-amazon (approved):

packages/amazonq/src/lsp/client.ts
```diff
@@ -5,7 +5,7 @@
 
 import vscode, { env, version } from 'vscode'
 import * as nls from 'vscode-nls'
-import * as cp from 'child_process'
+import * as cp from 'child_process' // eslint-disable-line no-restricted-imports -- language server options expect actual child process
```
justinmk3, line 8: We could add something like `ChildProcess.proc`  to return the nodejs object?

PR #6335, opened 2025-01-10
deps(clients): regen qdev streaming client (IGNORE)

Hweinstock:
## Problem
Build instructions are outdated and don't appear to work. This PR is an attempt to update them and regenerate the clients. 

### Notes
- Currently the new clients use older dependencies than the ones committed, which is strange? Suspect this is built from the wrong package. 
- New clients avoid depending on `@smithy/*` packages directly and prefer `@aws-sdk/*`. 

--


src.gen/@amzn/amazon-q-developer-streaming-client/src/runtimeConfig.shared.ts
```diff
@@ -1,42 +1,29 @@
...
 import { QDeveloperStreamingClientConfig } from "./QDeveloperStreamingClient";
 
 /**
  * @internal
  */
-export const getRuntimeConfig = (config: QDeveloperStreamingClientConfig) => {
-  return {
-    apiVersion: "2024-06-11",
-      base64Decoder: config?.base64Decoder ?? fromBase64,
+export const getRuntimeConfig = (config: QDeveloperStreamingClientConfig) => ({
```
nkomonen-amazon, line 18: Maybe this needs to be explicitly typed, based on this LLM comment:

If you're exporting the function in runtimeConfig.shared.ts, add a return type there:

In runtimeConfig.shared.ts:

```ts
import { RuntimeConfig } from "@aws-sdk/types";

export const getRuntimeConfig = (): RuntimeConfig => {
  // Your implementation here
};
```

Hweinstock: I don't think `@aws-sdk/types` exports this type for us unfortunately. Also, hoping to avoid manually adjusting generated code. One of the goals is to have these clients be generated as a build step or at least in an easily automatable fashion, so any manual touching of the generated files will be overwritten.

Hweinstock:
Moved to https://github.com/aws/aws-toolkit-vscode/pull/6439

PR #6338, opened 2025-01-10
Enable SageMaker SSO Users to use AmazonQ Chat & Code completion using their Credentials & Q Pro-tier profile

ahusseinali:
## Problem
SageMaker Amazon Q Pro-tier users didn't have the ability to use their pro-tier profile ARN when using Amazon Q Chat & Code Completion. Previously, we had SageMaker pro-tier SSO users unable to automatically use their Code Editor login credentials to create an auth connection with Amazon Q APIs. This meant they have to re-login in order to use AmazonQ chat & Code Completion.

One big obstacle to enable auto-login for SSO users was that their credentials are stored in HttpOnly Cookie and a change had to be set in place in SageMaker Code Editor environment in order for AmazonQ extension to be able to read the cookie value and send it as bearer token.

## Solution
Along with a change implemented for SageMaker Code Editor in https://github.com/aws/sagemaker-code-editor/pull/99/files
This PR implements the following:
1. SageMaker client to read SageMaker user domain status to identify whether pro-tier is enabled or not & read the cookie value through custom SageMaker code editor command.
1.1. In case Pro-tier is not enabled, the login flow is similar to IAM SageMaker users (using Container environment variables)
1.2. The custom command execution stores cookie values into a file and the client reads the value from that file.
2. SageMakerSSOTokenProvider: Custom token provider parallel to the original SSO Token provider. It doesn't go through any creation UI logic as the token already exists in file. This provider polls for token refresh every 5 minutes integrating with the client mentioned in #1
3. Pass Q Pro-tier ProfileARN (if it is enabled) to API calls that need it for SSO users
4. Change the connection validation logic to account for valid pre-created SSO connections for SageMaker environment.
5. Add Auto connect logic for SSO user flow in SageMaker

### Followup steps
Simplify the logic for communicating the SSO credentials and creating a connection based on such credentials, by letting SageMaker Code Editor create a standard AWS Credentials File & adding a watcher logic in AWS toolkit codebase to automatically sync with such file existence

---


packages/core/src/shared/sagemaker/client/sagemaker.ts
```diff
@@ -0,0 +1,136 @@
...
+    private cachedCookies: SageMakerCookie | undefined = undefined
+    private cachedProfileArn: string | undefined = undefined
+    private sageMakerClient: DefaultSageMakerClient
+
+    private constructor(sageMakerClient?: DefaultSageMakerClient) {
+        this.sageMakerClient = sageMakerClient ?? new DefaultSageMakerClient(this.getDefaultRegion())
+    }
+
+    public static getQConnectionScopes() {
+        return ['codewhisperer:completions', 'codewhisperer:conversations']
```
hayemaxi, line 66: Can we give users all amazon Q scopes? Why are we only giving this subset

packages/amazonq/.changes/next-release/Feature-2ca899db-fff6-469c-8294-f93fc965cfa7.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Enable SageMaker SSO user access token & Profile ARN to be used for accessing AmazonQ & CodeWhisperer features"
```
hayemaxi, line 3: "CodeWhisperer" is no more
```suggestion
	"description": "Auth: Enable SSO access for SageMaker users."
```

packages/core/src/auth/auth.ts
```diff
@@ -963,6 +1003,20 @@ export class Auth implements AuthService, ConnectionManager {
         return this.authenticate(id, refresh)
     }
 
+    public async tryAutoConnectSageMaker(): Promise<StatefulConnection | undefined> {
+        try {
+            const sagemakerConnection = await this.createSageMakerSsoConnection()
```
hayemaxi, line 1008: Could the connection already exist in a saved state from a prior session?

packages/core/src/shared/sagemaker/client/sagemaker.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public static getInstance(): SageMakerSpaceClient {
+        if (!this._instance) {
+            this._instance = new SageMakerSpaceClient()
+        }
+        return this._instance
+    }
```
hayemaxi, line 74: Follow the same format that we have in other singletons for consistency https://github.com/aws/aws-toolkit-vscode/blob/e6c3101e8950acea5158407b7099d66a26663a8c/packages/core/src/codewhisperer/util/authUtil.ts#L234-L241

```suggestion
    public static get instance() {
        if (!this.#instance) {
            this.#instance = new SageMakerSpaceClient()
        }
        return this.#instance
    }
```

packages/core/src/shared/sagemaker/client/sagemaker.ts
```diff
@@ -0,0 +1,136 @@
...
+                    ) as SageMakerCookie
+                }
+            } catch {
+                this.cachedCookies = undefined
+            }
+        }
+        return this.cachedCookies
+    }
+
+    public async getAmazonQProfileArn(): Promise<string | undefined> {
```
hayemaxi, line 109: What is a profile ARN and why do we want it? Please add a docstring

packages/core/src/shared/sagemaker/client/sagemaker.ts
```diff
@@ -0,0 +1,136 @@
...
+    expiryTime?: number
+    ssoExpiryTimestamp?: number
+    studioUserProfileName?: string
+    redirectURL?: string
+    AccessToken?: string
+    StudioSessionToken?: string
+}
+
+// Manually expire the token every 5 minutes as token refresh is handled externally by Sagemaker
+const TOKEN_EXPIRY = 5 * 60 * 1000
```
hayemaxi, line 26: Use constants in https://github.com/aws/aws-toolkit-vscode/blob/e6c3101e8950acea5158407b7099d66a26663a8c/packages/core/src/shared/datetime.ts#L8

e.g.
```suggestion
const TOKEN_EXPIRY = oneMinute * 5
```

packages/core/src/shared/sagemaker/client/sagemaker.ts
```diff
@@ -0,0 +1,136 @@
...
+const DEFAULT_REGION = 'us-east-1'
```
hayemaxi, line 27: We already have logic for this. Use https://github.com/aws/aws-toolkit-vscode/blob/e6c3101e8950acea5158407b7099d66a26663a8c/packages/core/src/shared/regions/regionProvider.ts#L46

packages/core/src/auth/utils.ts
```diff
@@ -793,10 +794,22 @@ export function getAuthFormIdsFromConnection(conn?: Connection): AuthFormId[] {
...
-    manager.addProviderFactory(new SharedCredentialsProviderFactory())
-    manager.addProviders(new Ec2CredentialsProvider(), new EcsCredentialsProvider(), new EnvVarsCredentialsProvider())
+export async function initializeCredentialsProviderManager() {
+    let useDefaultCredentials = true
+
+    if (isSageMaker() && isAmazonQ()) {
+        useDefaultCredentials = !(await SageMakerSpaceClient.getInstance().getAmazonQProfileArn())
+    }
+
+    if (useDefaultCredentials) {
```
hayemaxi, line 804:
```suggestion
    const useDefaultCredentials = isSageMaker() && isAmazonQ() && !(await SageMakerSpaceClient.getInstance().getAmazonQProfileArn())

    if (useDefaultCredentials) {
```

justinmk3:
Blocking suggestion:

> we aligned on giving all users all scopes on auth/reauth. We want to deprecate a lot of complicated legacy code that tries to support multiple scope sets. If everyone has the same scopes then our code/telemetry will be much more understandable, we won't need edge cases for different scope sets, and if other features become enabled in sagemaker we wont have to think about upgrading/migrating/handling existing user auths with less scopes
>
> gating the unavailable features behind non-scope checks is simpler. If there is no background issue relating to sagemaker sso auth having all scopes, then we should add all 5 scopes

PR #6351, opened 2025-01-13
fix(dev): Q developer menu opening wrong auth state

hayemaxi:
It's opening the toolkit global state value instead.

---

### DIFF
--- a/packages/core/src/dev/activation.ts
+++ b/packages/core/src/dev/activation.ts
@@ -334,7 +334,7 @@ class ObjectEditor {
                 return this.openState(targetContext.secrets, key)
             case 'auth':
                 // Auth memento is determined in a different way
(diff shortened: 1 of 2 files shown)


nkomonen-amazon:
> It's opening the toolkit global state value instead.

Does this mean that the Q extension has access to the Toolkit's global state somehow?

In `getEnvironmentSpecificMemento()` I would have though that `globals.globalState` would have already been initialized and returned Amazon Q's global state

justinmk3:
> Does this mean that the Q extension has access to the Toolkit's global state somehow?

Problem is (I think) currently there are `Amazon Q (Dev): ...` commands and `AWS (Dev): ...` commands. But the "dev" stuff lives in AWS Toolkit.

PR #6353, opened 2025-01-13
build(os-compatibility): normalize file paths and enforce eol consistency

muaydin:
## Problem

- Tests were failing or behaving inconsistently on different operating systems due to improper file path normalization.
- End-of-line (EOL) conflicts were causing unnecessary formatting diffs in collaborative workflows.

## Solution

- Updated the testFile variable in the test runner to normalize paths by replacing backslashes (\) with forward slashes, ensuring cross-platform compatibility.
- Added the Prettier rule 'prettier/prettier': ['error', { endOfLine: 'auto' }] to align EOL behavior with the host environment, avoiding platform-specific conflicts.

---

### DIFF
--- a/packages/core/src/test/testRunner.ts
+++ b/packages/core/src/test/testRunner.ts
@@ -84,7 +84,8 @@ export async function runTests(
     })
 
     const dist = path.resolve(root, 'dist')
-    const testFile = process.env['TEST_FILE']?.replace('.ts', '.js')
+    const rawTestFile = process.env['TEST_FILE']
+    const testFile = rawTestFile?.replace(/\\/g, '/').replace('.ts', '.js')
     let testFilePath: string | undefined
     if (testFile?.includes('../core/')) {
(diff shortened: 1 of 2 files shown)


packages/core/src/test/testRunner.ts
```diff
@@ -84,7 +84,8 @@ export async function runTests(
     })
 
     const dist = path.resolve(root, 'dist')
-    const testFile = process.env['TEST_FILE']?.replace('.ts', '.js')
+    const rawTestFile = process.env['TEST_FILE']
+    const testFile = rawTestFile?.replace(/\\/g, '/').replace('.ts', '.js')
```
justinmk3, line 88: What is passing `\` paths here? We should not be doing that in any of our launch configs, build steps, etc. It's 100% unnecessary.

This change is probably still worth having, though.

justinmk3: Ok, it's probably vscode itself passing the `\` paths.

PR #6364, opened 2025-01-14
refactor(amazonq): Improve robustness of lsp installation process

jpinkney-aws:
## Problem
- resolvers should produce runable binaries with the correct permissions
- nodepaths should correctly target the correct os
- workspace context lsp should reference the correct location

## Solution
- resolvers now produce runable binaries
- nodepaths are now correct for the workspace context lsp on windows/mac/linux
- lsp location is now correct for the workspace context lsp

---


packages/core/src/shared/languageServer/types.ts
```diff
@@ -15,8 +15,16 @@ export interface LspResult {
     assetDirectory: string
 }
 
+export interface ExecutablePaths {
```
nkomonen-amazon, line 18: nit: Will these paths always be executables? Would something like `ResourcePaths` be more generalized for the long term?

jpinkney-aws: Yeah that's a way better name, I couldn't think of a good one 😆

packages/amazonq/src/lsp/client.ts
```diff
@@ -30,14 +31,16 @@ export function startLanguageServer(extensionContext: vscode.ExtensionContext, s
...
+
     // If the extension is launch in debug mode the debug server options are use
     // Otherwise the run options are used
     let serverOptions: ServerOptions = {
         run: { module: serverPath, transport: TransportKind.ipc },
         debug: { module: serverPath, transport: TransportKind.ipc, options: debugOptions },
     }
 
-    const child = cp.spawn('node', [serverPath, ...debugOptions.execArgv])
+    const child = cp.spawn(executablePaths.node, [serverPath, ...debugOptions.execArgv])
```
justinmk3, line 43: if we can find a way to use our ChildProcess module, then after https://github.com/aws/aws-toolkit-vscode/pull/6304 we get instrumentation for free :)

justinmk3 (approved):

packages/core/src/amazonq/lsp/workspaceInstaller.ts
```diff
@@ -22,7 +25,21 @@ export class WorkspaceLSPResolver implements LspResolver {
             new Range(supportedLspServerVersions)
         ).resolve()
 
+        const nodeName =
+            process.platform === 'win32' ? getNodeExecutableName() : `node-${process.platform}-${process.arch}`
```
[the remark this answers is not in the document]
jpinkney-aws: For some reason this idea only exists for the workspace installer. I think the actual standard (which the actual codewhisperer language server uses) is just node.exe/node

PR #6365, opened 2025-01-14
fix(amazonq): add message after accept/reject action

chungjac:
## Problem
- Should show a completed message after user completes accept/reject process (just like JB)
- This message does not currently show in VSC

## Solution
- Show completed message after user completes accept/reject process 

https://github.com/user-attachments/assets/e6e65268-b851-4298-afa6-a784e75dd0a3

---

### DIFF
--- a/packages/core/src/amazonqTest/chat/controller/controller.ts
+++ b/packages/core/src/amazonqTest/chat/controller/controller.ts
@@ -717,6 +717,9 @@ export class TestController {
         await vscode.window.showTextDocument(document)
         // TODO: send the message once again once build is enabled
         // this.messenger.sendMessage('Accepted', message.tabID, 'prompt')
+
+        this.messenger.sendMessage('Unit test generation completed', message.tabID, 'answer')
+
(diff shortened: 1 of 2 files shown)


jpinkney-aws (approved):

packages/core/src/amazonqTest/chat/controller/controller.ts
```diff
@@ -840,6 +843,8 @@ export class TestController {
     private async endSession(data: any, step: FollowUpTypes) {
         const session = this.sessionStorage.getSession()
         if (step === FollowUpTypes.RejectCode) {
+            this.messenger.sendMessage('Unit test generation completed.', data.tabID, 'answer')
```
ctlai95, line 846: Can't we move this outside of the if statement so that we always send this message on session end?
Then you can also remove the message from acceptCode function

chungjac: Fixed

ashishrp-aws (approved):

PR #6366, opened 2025-01-14
fix(amazonq): max 20k char for inline supplementalContext

leigaol:
## Problem
In some rare cases, the total supplementalContext length exceeds 20k character. With recent service side changes, we need to make sure the total supplementalContext length do not exceed 20k in any situation.

## Solution
Make sure the total supplementalContext length do not exceed 20k in any situation. Drop when necessary.

This is not customer facing so no change log is attached.

---

### DIFF
--- a/packages/core/src/test/codewhisperer/testUtil.ts
+++ b/packages/core/src/test/codewhisperer/testUtil.ts
@@ -329,3 +329,12 @@ export function aStringWithLineCount(lineCount: number, start: number = 0): stri
 
     return s.trimEnd()
 }
+
+export function aLongStringWithLineCount(lineCount: number, start: number = 0): string {
+    let s = ''
+    for (let i = start; i < start + lineCount; i++) {
+        s += `a`.repeat(100) + `line${i}\n`
+    }
+
+    return s.trimEnd()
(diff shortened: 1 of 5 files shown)


jpinkney-aws (approved):

Will-ShaoHua (approved):

packages/core/src/test/codewhisperer/testUtil.ts
```diff
@@ -329,3 +329,12 @@ export function aStringWithLineCount(lineCount: number, start: number = 0): stri
...
 }
+
+export function aLongStringWithLineCount(lineCount: number, start: number = 0): string {
+    let s = ''
+    for (let i = start; i < start + lineCount; i++) {
+        s += `a`.repeat(100) + `line${i}\n`
+    }
+
+    return s.trimEnd()
+}
```
Will-ShaoHua, line 340: todo: let's merge the 2 util funcs `aStringWithLineCount` and `aLongStringWithLineCount`

Will-ShaoHua: not a blocker

PR #6368, opened 2025-01-14
fix(sso): login with custom startUrl not allowed

nkomonen-amazon:
## Problem:

A user reported that a non-standard start url is technically
valid. This is because it can redirect to the underlying valid
start url that matches the pattern: https://xxxxxxxx.awsapps.com/start

## Solution:

Allow any URL, but warn users if they are using a non-standard one.
We will show a yellow warning message in this case.
The red error message is still shown when the input does not match a
URL in general.

### Invalid URL
Screenshot 2025-01-14 at 4 33 58 PM

### Possibly valid since it may redirect to a valid URL
Screenshot 2025-01-14 at 4 34 13 PM

### Missing the trailing `/start`
Screenshot 2025-01-14 at 4 34 29 PM

### URL that also matches expected pattern
Screenshot 2025-01-14 at 4 34 35 PM

---

### DIFF
--- a/packages/core/src/auth/sso/constants.ts
+++ b/packages/core/src/auth/sso/constants.ts
@@ -11,8 +11,15 @@
 export const builderIdStartUrl = 'https://view.awsapps.com/start'
 export const internalStartUrl = 'https://amzn.awsapps.com/start'
 
+/**
+ * Doc: https://docs.aws.amazon.com/singlesignon/latest/userguide/howtochangeURL.html
+ */
 export const ssoUrlFormatRegex =
     /^(https?:\/\/(.+)\.awsapps\.com\/start|https?:\/\/identitycenter\.amazonaws\.com\/ssoins-[\da-zA-Z]{16})\/?$/
 
+/**
+ * It is possible for a start url to be a completely custom url that redirects to something that matches the format
+ * below, so this message is only a warning.
+ */
 export const ssoUrlFormatMessage =
-    'URLs must start with http:// or https://. Example: https://d-xxxxxxxxxx.awsapps.com/start'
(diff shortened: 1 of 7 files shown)


packages/core/src/auth/sso/constants.ts
```diff
@@ -11,8 +11,15 @@
...
 export const ssoUrlFormatRegex =
     /^(https?:\/\/(.+)\.awsapps\.com\/start|https?:\/\/identitycenter\.amazonaws\.com\/ssoins-[\da-zA-Z]{16})\/?$/
 
+/**
+ * It is possible for a start url to be a completely custom url that redirects to something that matches the format
+ * below, so this message is only a warning.
+ */
 export const ssoUrlFormatMessage =
-    'URLs must start with http:// or https://. Example: https://d-xxxxxxxxxx.awsapps.com/start'
+    'URL possibly invalid. It typically follows the pattern: https://xxxxxxxxxx.awsapps.com/start'
```
justinmk3, line 25:
```suggestion
    'URL possibly invalid. Expected format: https://xxxxxxxxxx.awsapps.com/start'
```

packages/core/src/shared/utilities/uriUtils.ts
```diff
@@ -58,3 +58,15 @@ export function toUri(path: string | vscode.Uri): vscode.Uri {
     }
     return vscode.Uri.file(path)
 }
+
+export function isValidUrl(string: string): boolean {
```
justinmk3, line 62: is `vscode.Uri.parse` useful here? docstring may want to mention why not

justinmk3 (approved):

jpinkney-aws (approved):

PR #6370, opened 2025-01-14
fix(amazonq): rely on server-side data to communicate number of code generations remaining

siakmun-aws:
## Problem

Previously remaining code generation iterations are always displayed.

## Solution

Only display when it goes down to 2 or less.

---

### DIFF
--- a/packages/amazonq/.changes/next-release/Feature-f03a4f2f-5819-4276-b914-be0b1a4ecedf.json
+++ b/packages/amazonq/.changes/next-release/Feature-f03a4f2f-5819-4276-b914-be0b1a4ecedf.json
@@ -1,4 +1,0 @@
-{
-	"type": "Feature",
-	"description": "Disable displaying remaining code generation iterations until 2 or less remaining"
-}
--- a/packages/amazonq/.changes/next-release/Feature-f03a4f2f-5819-4276-b914-be0b1a4ecedf.json
+++ b/packages/amazonq/.changes/next-release/Feature-f03a4f2f-5819-4276-b914-be0b1a4ecedf.json
@@ -1,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Disable displaying remaining code generation iterations until 2 or less remaining"
+}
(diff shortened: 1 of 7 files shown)


packages/core/src/amazonqFeatureDev/controllers/chat/controller.ts
```diff
@@ -465,7 +463,7 @@ export class FeatureDevController {
                     type: 'answer',
                     tabID: tabID,
                     message:
-                        remainingIterations === 0
+                        remainingIterations > 2 || remainingIterations <= 0
```
neilk-aws, line 466: Should we split this into two separate conditions? The reason is the ", or provide feedback for new code?" that is only shown when we're on the last two iterations in this PR.

I think we want

```
if > 2
    code or feedback
elif > 0
    code or feedback, iterations left
else
    code
```

packages/core/src/test/amazonqFeatureDev/session/sessionState.test.ts
```diff
@@ -111,17 +111,17 @@ describe('sessionState', () => {
         it('transitions to PrepareCodeGenState when codeGenerationStatus ready ', async () => {
             mockGetCodeGeneration = sinon.stub().resolves({
                 codeGenerationStatus: { status: 'Complete' },
-                codeGenerationRemainingIterationCount: 2,
-                codeGenerationTotalIterationCount: 3,
+                codeGenerationRemainingIterationCount: 9,
```
neilk-aws, line 114: nit: This shouldn't have needed to change since it's in a unit test. Keeping it as a different value may help you uncover any places where you're making assumptions based on the actual value.

packages/amazonq/.changes/next-release/Feature-f03a4f2f-5819-4276-b914-be0b1a4ecedf.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Disable displaying remaining code generation iterations until 2 or less remaining"
```
neilk-aws, line 3: I think it's hard to justify why this is a feature. 

Instead, I think we can say type is bug fix and description is "Remove hard-coded limits and instead rely server-side data to communicate number of code generations remaining"

neilk-aws (changes requested):

siakmun-aws (commented):
Reviewed

neilk-aws (approved):

neilk-aws (approved):

PR #6372, opened 2025-01-15
build(lint): add rule to prefer direct imports over index.ts

Hweinstock:
## Problem
Importing from `..` or an `index.ts` file can lead to circular dependencies. 

## Solution
- add a lint rule to discourage importing from `..`. 
- migrate existing cases to import directly from the target module. 

---


.eslintrc.js
```diff
@@ -206,6 +206,11 @@ module.exports = {
                         message:
                             'Avoid child_process, use ChildProcess from `shared/utilities/processUtils.ts` instead.',
                     },
+                    {
+                        name: '..',
```
hayemaxi, line 210: This prevents importing from index.ts files in other paths? E.g. `./adjacentFolder/folderWithIndex/`

hayemaxi (approved):

.eslintrc.js
Hweinstock: I don't think so, but I think the circular dependency is only a problem when the `index.ts` is importing from the current file as well. This should only happen when the `index.ts` is a parent rather than an adjacent folder. 

There is another case where the `index.ts` is more than one level up, but I wasn't able to find any instances of this in the toolkit.

jpinkney-aws (approved):

.eslintrc.js
hayemaxi: I've found the most common pitfall to be having a file that depends on an index.ts that is exporting something depends on your file. E.g.

```
// connection.ts
import { logger } from './logger' // index.ts import
```

```
// logger/index.ts
export logger from './logger'
export loggerUtil from './loggerUtil'
```

```
// logger/loggerUtil.ts
import { connectionType } from '../connection'
```

connection.ts > logger/index.ts > loggerUtils.ts > connection.ts,
when in reality we just want connection.ts > logger/logger.ts

justinmk3 (approved):
Nice solution!

.eslintrc.js
Hweinstock: I see. Yeah unfortunately, this won't address that. I would also imagine it would be hard to catch via lint rule.

PR #6376, opened 2025-01-15
revert: cloud9 related code

hayemaxi:
This code removes nearly all cloud9 codepaths:
- Icons, readme generation, scripts
- ikp3db lambda/sam debugger support
- Cloud9 only codepaths and hacks
- Docs and comments mentioning Cloud9

Includes some follow up items as TODOs in code.

Some items were left in case they are helpful in the future (e.g. for vscode web or similar). The items should be renamed from Cloud9 if used again in the future. Some examples are Cloud9 runtime environment checks (for lambda/sam), `isCloud9()` and properties/text for Cloud9 IDE, some SAM artifacts, don't start LSP server, etc. Search cloud9 in vscode after merging this change.

---

- Treat all work as PUBLIC. Private `feature/x` branches will not be squash-merged at release time.
- Your code changes must meet the guidelines in [CONTRIBUTING.md](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#guidelines).
- License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6379, opened 2025-01-16
refactor(core): Make HttpResourceFetcher platform agnostic

jpinkney-aws:
## Problem
- Resource fetcher currently only works on node. This causes some implements to use got directly, some use fetch directly, some use httpResourceFetcher depending on their use case

## Solution
- Make HttpResourceFetcher platform independent so it can run on node + web
- Move the streaming based HttpResourceFetcher to a node specific folder

---

### DIFF
--- a/packages/core/src/shared/resourcefetcher/node/httpResourceFetcher.ts
+++ b/packages/core/src/shared/resourcefetcher/node/httpResourceFetcher.ts
@@ -39,7 +39,7 @@ type RequestHeaders = { eTag?: string; gZip?: boolean }
 
 /**
  * Legacy HTTP Resource Fetcher used specifically for streaming information.
- * If you do not need streaming or want multi platform support use
+ * Only kept around until web streams are compatible with node streams
  */
 export class HttpResourceFetcher {
     private readonly logger: Logger = getLogger()
--- a/packages/core/src/shared/resourcefetcher/node/httpResourceFetcher.ts
+++ b/packages/core/src/shared/resourcefetcher/node/httpResourceFetcher.ts
@@ -1,0 +1,129 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'fs' // eslint-disable-line no-restricted-imports
+import * as http from 'http'
+import * as https from 'https'
+import * as stream from 'stream'
+import got, { RequestError } from 'got'
+import urlToOptions from 'got/dist/source/core/utils/url-to-options'
+import Request from 'got/dist/source/core'
+import { VSCODE_EXTENSION_ID } from '../../extensions'
+import { getLogger, Logger } from '../../logger'
+import { Timeout, CancellationError, CancelEvent } from '../../utilities/timeoutUtils'
+import { isCloud9 } from '../../extensionUtilities'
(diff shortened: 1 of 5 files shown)


packages/core/src/shared/resourcefetcher/node/httpResourceFetcher.ts
```diff
@@ -0,0 +1,129 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'fs' // eslint-disable-line no-restricted-imports
+import * as http from 'http'
+import * as https from 'https'
+import * as stream from 'stream'
+import got, { RequestError } from 'got'
```
justinmk3, line 10: Why do we need a node-only version? We are trying to eliminate "got" entirely.

jpinkney-aws: IIRC @nkomonen-amazon mentioned that fs streams from fetch weren't compatible with fs streams from node yet. Right now this will limit `got` to this single file until things are compatible and then I can remove all the other references of got

nkomonen-amazon: From an old Web quip:

* Node.js has their own [Stream implementation](https://nodejs.org/docs/latest-v18.x/api/stream.html#stream)
* Browser/Web uses a different [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) (implements the [WHATWG Streams Standard](https://streams.spec.whatwg.org/))
  * [Though this is supported in Node 21+](https://nodejs.org/api/webstreams.html#web-streams-api) and 'got' could technically update to this API

nkomonen-amazon (approved):

justinmk3 (approved):

packages/core/src/shared/resourcefetcher/node/httpResourceFetcher.ts
```diff
@@ -0,0 +1,129 @@
...
+    requestStream: Request // `got` doesn't add the correct types to 'on' for some reason
+    /** Stream writing to the file system. */
+    fsStream: fs.WriteStream
+}
+
+type RequestHeaders = { eTag?: string; gZip?: boolean }
+
+/**
+ * Legacy HTTP Resource Fetcher used specifically for streaming information.
+ * If you do not need streaming or want multi platform support use
```
justinmk3, line 42: comment still needed?

packages/core/src/shared/resourcefetcher/node/httpResourceFetcher.ts
```diff
@@ -0,0 +1,129 @@
...
+import Request from 'got/dist/source/core'
+import { VSCODE_EXTENSION_ID } from '../../extensions'
+import { getLogger, Logger } from '../../logger'
+import { Timeout, CancellationError, CancelEvent } from '../../utilities/timeoutUtils'
+import { isCloud9 } from '../../extensionUtilities'
+import { Headers } from 'got/dist/source/core'
+
+// XXX: patched Got module for compatability with older VS Code versions (e.g. Cloud9)
+// `got` has also deprecated `urlToOptions`
+const patchedGot = got.extend({
```
justinmk3, line 21: patchedGot is removed in max's pr https://github.com/aws/aws-toolkit-vscode/pull/6376

PR #6384, opened 2025-01-16
fix(amazonq): setMaven correctly

dhasani23:
## Problem

`setMaven` is being executed unnecessarily.

## Solution

Move the call to happen just before the local build starts.

---

### DIFF
--- a/packages/core/src/amazonqGumby/chat/controller/controller.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/controller.ts
@@ -53,11 +53,9 @@ import {
 import { CodeTransformTelemetryState } from '../../telemetry/codeTransformTelemetryState'
 import DependencyVersions from '../../models/dependencies'
 import { getStringHash } from '../../../shared/utilities/textUtilities'
-import { getVersionData } from '../../../codewhisperer/service/transformByQ/transformMavenHandler'
 import AdmZip from 'adm-zip'
 import { AuthError } from '../../../auth/sso/server'
 import {
-    setMaven,
     openBuildLogFile,
     parseBuildFile,
     validateSQLMetadataFile,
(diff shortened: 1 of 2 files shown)


packages/core/src/amazonqGumby/chat/controller/controller.ts
```diff
@@ -321,12 +319,6 @@ export class GumbyController {
                     telemetryJavaVersion = JDKToTelemetryValue(javaVersion) as CodeTransformJavaSourceVersionsAllowed
                 }
                 telemetry.record({ codeTransformLocalJavaVersion: telemetryJavaVersion })
-
-                await setMaven()
-                const versionInfo = await getVersionData()
-                const mavenVersionInfoMessage = `${versionInfo[0]} (${transformByQState.getMavenName()})`
-                telemetry.record({ buildSystemVersion: mavenVersionInfoMessage })
```
dhasani23, line 328: This was always recording `[undefined, undefined]` since the user hasn't selected a project to transform at this point. Plus, we already emit this in the `totalRunTime` metric at the very end of a transformation anyway.

packages/core/src/codewhisperer/service/transformByQ/transformMavenHandler.ts
```diff
@@ -108,6 +108,10 @@ function copyProjectDependencies(dependenciesFolder: FolderInfo, modulePath: str
 }
 
 export async function prepareProjectDependencies(dependenciesFolder: FolderInfo, rootPomPath: string) {
+    await setMaven()
```
dhasani23, line 111: Better to set the Maven to use (`mvn`, `mvnw`, `mvnw.cmd`) just before the local build happens to ensure we're using the correct Maven executable.

ntarakad-aws (approved):

packages/core/src/codewhisperer/service/transformByQ/transformApiHandler.ts
```diff
@@ -670,6 +671,7 @@ export async function pollTransformationJob(jobId: string, validStates: string[]
                 })
             }
             transformByQState.setPolledJobStatus(status)
+            getLogger().info(`CodeTransformation: polled job status = ${status}`)
```
justinmk3, line 674: FYI: `getLogger()` now has a "topic" feature so you can do this at the top of the module:

    const logger = getLogger('CodeTransformation')

then `logger.info()` will prepend the topic name.

PR #6385, opened 2025-01-16
telemetry(lsp): Integrate language server/manifest resolver telemetry

Hweinstock:
## Problem
LSP downloading processes does not emit any telemetry. 

### Refactoring
- separate verification from core downloading steps so that we can capture it as its own telemetry event.
- Note: this refactor means that if all the downloaded content cannot be verified, no files will be written to disk. 
- Introduce abstraction of `StageResolver` to make telemetry instrumentation more natural. 

### Metric Behavior
- metric for each stage of the LSP setup process.
- can be emitted multiple times per stage to capture specific error codes. See tests for examples. 
- Commons repo PR: https://github.com/aws/aws-toolkit-common/pull/961
- bumped commons version to [1.0.296](https://github.com/aws/aws-toolkit-common/commit/8df7a87557f40e5ccd499a8f3ce38976893c8b10) to include this change. 

### Testing
- adds basic tests for `ManifestResolver` in `packages/core/src/test/shared/lsp/manifestResolver.test.ts`.
- adds basic tests for `LanguageServerResolver` in `packages/core/src/test/shared/lsp/lspResolver.test.ts`. 
- Currently `LanguageServerResolver` only supports mac, so its tests skip on non-mac platforms. There is a techdebt test to address this. 

## Future Work
- Update logging messages to be LSP specific. 

---


packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -358,6 +370,28 @@
...
+                {
+                    "type": "lspSetupLocation",
+                    "required": false
+                },
+                {
+                    "type": "duration",
+                    "required": true
+                },
+                {
+                    "type": "result",
```
justinmk3, line 390: duration/result are standard fields, shouldn't be needed here

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1,5 +1,22 @@
 {
     "types": [
+        {
+            "name": "languageServerSetupStage",
```
jpinkney-aws, line 4: I wish we could do properly namespaced things in telemetry. It sucks that we always have prefix specific types like this

instead of having:
```
namespace lsp {
    
}
```

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -358,6 +375,24 @@
         }
     ],
     "metrics": [
+        {
+            "name": "languageServer_setup",
```
jpinkney-aws, line 379: can we port this to the commons repo? I think we wanted to align all teams on a common implementation

packages/core/package.nls.json
```diff
@@ -20,7 +20,7 @@
     "AWS.configuration.description.suppressPrompts": "Prompts which ask for confirmation. Checking an item suppresses the prompt.",
     "AWS.configuration.enableCodeLenses": "Enable SAM hints in source code and template.yaml files",
     "AWS.configuration.description.resources.enabledResources": "AWS resources to display in the 'Resources' portion of the explorer.",
-      "AWS.configuration.description.featureDevelopment.allowRunningCodeAndTests": "Allow /dev to run code and test commands",
+    "AWS.configuration.description.featureDevelopment.allowRunningCodeAndTests": "Allow /dev to run code and test commands",
```
jpinkney-aws, line 23: was this an auto format thing?

packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -183,4 +182,12 @@ export class LspController {
             }
         })
     }
+
+    private async setupLsp(context: vscode.ExtensionContext) {
+        await lspSetupStage('all', async () => {
+            const installResult = await new WorkspaceLSPResolver().resolve()
+            await lspSetupStage('launch', async () => activateLsp(context, installResult.resourcePaths))
+            getLogger().info('LspController: LSP activated')
```
jpinkney-aws, line 190: Not necessarily an issue for this PR, but I think we should probably scope these messages to regular codewhisperer language server vs workspace context one

packages/core/src/shared/lsp/lspResolver.ts
```diff
@@ -30,63 +31,90 @@ export class LanguageServerResolver {
...
-            result.version = latestVersion.serverVersion
-            result.assetDirectory = cacheDirectory
-            return result
-        } else {
-            // clean up any leftover content that may have been downloaded
-            if (await fs.existsDir(cacheDirectory)) {
-                await fs.delete(cacheDirectory, {
-                    recursive: true,
-                })
+        const serverResolvers: StageResolver<LspResult>[] = [
```
jpinkney-aws, line 38: not neccessarily a problem but just wanted to call it out for anyone reading -- this is where we will slightly deviate from eclipse/visual studios implementations. We can just do some things a bit simpler

packages/core/src/test/shared/lsp/lspResolver.test.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    after(function () {
+        sinon.restore()
+    })
+
+    it('tries local cache first', async function () {
+        if (!isMac) {
+            this.skip()
+        }
+        localStub.resolves(lspResult('cache'))
```
jpinkney-aws, line 99: there are technically non stubbed versions of these tests in the integ or e2e folder btw. It might make sense to just add telemetry assertions there

packages/core/package.nls.json
Hweinstock: Yeah, it looks like there was an extra space left in there.

packages/core/src/amazonq/lsp/lspController.ts
Hweinstock: Yeah, I think we could make use of the logging header here that references which LSP it is.

packages/core/src/shared/lsp/lspResolver.ts
Hweinstock: In what ways does this deviate from their implementations? Would it be easier if I modified the implementation to closer mirror those implementations?

packages/core/src/test/shared/lsp/lspResolver.test.ts
Hweinstock: Where are these tests? I am having some trouble finding them.

jpinkney-aws: https://github.com/aws/aws-toolkit-vscode/blob/feature/amazonqLSP/packages/amazonq/test/e2e/lsp/lspInstaller.test.ts#L73

packages/core/src/shared/lsp/lspResolver.ts
jpinkney-aws: Just the structure of having resolvers. It's not a big deal. I just copied Visual Studios implementation and basically ported it to typescript. AFAIK Eclipse ported Visual Studio's implementation for this code to java. This is just a slight refactor of their work

jpinkney-aws (approved):

jpinkney-aws:
Ah looks like the e2e tests are failing with the thing you mentioned in standup

Hweinstock:
Oh interesting. I suspect this might be something else since I stubbed the workspaceLSPInstaller. The remote manifest fetch fails on local, but passes in CI. I'll adjust the tests and add a note that it doesn't work locally.

packages/amazonq/test/e2e/lsp/lspInstaller.test.ts
```diff
@@ -117,6 +121,86 @@ describe('AmazonQLSPInstaller', () => {
             assert.ok(fallback.assetDirectory.startsWith(tempDir))
             assert.deepStrictEqual(fallback.location, 'fallback')
             assert.ok(semver.satisfies(fallback.version, supportedLspServerVersions))
+
+            // Exclude version numbers so that this test doesn't have to be updated on each update.
+            // Fetching manifest fails locally, but passes in CI. Therefore, this test will fail locally.
```
jpinkney-aws, line 126: Can you explain more about this? The test will fail locally?

packages/core/src/test/shared/lsp/lspResolver.test.ts
jpinkney-aws: What functionality is covered in this one thats not in those other ones?

packages/amazonq/test/e2e/lsp/lspInstaller.test.ts
Hweinstock: When running locally, the manifest fetch fails on both the first and second try. However, when it runs in CI the first try succeeds and the second one does not. Not entirely sure why, but it appears to be consistent across the 3 times I've run it so far. How it ultimately resolves the manifest impacts which telemetry metrics are emitted so the metrics emitted when run locally differ from that in CI. Therefore, I adjusted the test so that it the expected behavior is the behavior in CI.

Hweinstock: After diving deeper on this I discovered the root issue, and it should be fixed now. To summarize, the globalState is consistent in CI, but not local since we don't overwrite it on e2e/integ tests. This means that if I've already downloaded the LSP locally, the manifest etag is in globalState. If this eTag is up to date, the remote fetch fails. 

I fixed it by temporarily clearing the manifest storage in globalState for these tests, and then restoring it afterwards.

jpinkney-aws (approved):

PR #6387, opened 2025-01-17
refactor(util): merge withRetries and waitUntil

nkomonen-amazon:
## Problem:

We had 2 different implementation of similar code, between withRetries and waitUntil.

We couldn't easily merge them though since behavior was slightly different when it came to expected return types, and since one threw and the other did not.

- waitUntil returns undefined if it never properly resolves
- withRetries throws if it never properly resolves

## Solution:

As a solution we keep both methods but they share the same underlying code.

- waitUntil now has a backoff field to increase the delta between intervals
  - withRetries also has this
- withRetries will retry even on a thrown exeception
  - on timeout it will throw the last exception it encountered

---

### DIFF
--- a/packages/core/src/shared/utilities/timeoutUtils.ts
+++ b/packages/core/src/shared/utilities/timeoutUtils.ts
@@ -222,9 +222,9 @@ interface WaitUntilOptions {
     /** A backoff multiplier for how long the next interval will be (default: None, i.e 1) */
     readonly backoff?: number
     /**
-     * Only retries when an error is thrown, otherwise returning the result regardless of truthiness.
-     * - Ignores 'truthy' arg
-     * - If the timeout is reached it will throw the last error
+     * Only retries when an error is thrown, otherwise returning the immediate result.
+     * - 'truthy' arg is ignored
+     * - If the timeout is reached it throws the last error
      * - default: false
      */
     readonly retryOnFail?: boolean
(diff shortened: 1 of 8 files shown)


justinmk3:
> withRetries throws if it never properly resolves

can this behavior be controlled by a flag on `waitUntil`, so we don't need both functions?

packages/core/src/shared/utilities/timeoutUtils.ts
```diff
@@ -219,43 +220,116 @@ interface WaitUntilOptions {
...
+    /** A backoff multiplier for how long the next interval will be (default: None, i.e 1) */
+    readonly backoff?: number
+}
+
+interface WithRetriesOptions extends WaitUntilOptions {
+    /**
+     * Ignore intermediate errors thrown by fn(). If timeout it reached, it will throw the last error.
+     * (default: true)
+     */
+    readonly catch?: boolean
```
justinmk3, line 232: can waitUntil gain this flag?

packages/core/src/shared/utilities/timeoutUtils.ts
```diff
@@ -219,40 +219,102 @@ interface WaitUntilOptions {
...
+    readonly retryOnFail?: boolean
 }
 
+export const waitUntilDefaultTimeout = 2000
+export const waitUntilDefaultInterval = 500
+
 /**
  * Invokes `fn()` until it returns a truthy value (or non-undefined if `truthy:false`).
  *
+ * Also look at {@link withRetries} to be able to retry on failures.
```
jpinkney-aws, line 239: I guess this comments no longer relevant since it looks like withRetries is now removed

jpinkney-aws (approved):

packages/core/src/shared/utilities/timeoutUtils.ts
nkomonen-amazon: Thanks for catching, fixed

PR #6390, opened 2025-01-17
refactor(core): Improve httpResourceFetcher API

jpinkney-aws:
## Problem
Built on https://github.com/aws/aws-toolkit-vscode/pull/6379 it makes the api a little more clearer for http specific requests by returning Response objects instead of strings. That opens up the library to be used in other places like downloading language servers, which currently isn't possible because the contents are always converted to string

I'm not entirely sold on it so far but it certainly gives a lot more flexibility to underlying callers while abstracting away the actual request library we're using

## Solution

---


packages/core/src/lambda/utils.ts
```diff
@@ -13,8 +13,6 @@ import { CloudFormationClient } from '../shared/clients/cloudFormationClient'
 import { LambdaClient } from '../shared/clients/lambdaClient'
 import { getFamily, getNodeMajorVersion, RuntimeFamily } from './models/samLambdaRuntime'
 import { getLogger } from '../shared/logger'
-import { ResourceFetcher } from '../shared/resourcefetcher/resourcefetcher'
-import { CompositeResourceFetcher } from '../shared/resourcefetcher/compositeResourceFetcher'
```
justinmk3, line 17: Future idea: I noticed long ago that `CompositeResourceFetcher` is re-inventing "first-class functions".  It's not needed in js/ts.

jpinkney-aws: Yeah I was thinking about just removing it. CompositeResourceFetcher was only ever used in that specific case

justinmk3 (approved):

PR #6395, opened 2025-01-17
fix(amazonq): enable input after prompt is rejected, prevent iteration limit exceeded

avi-alpert:
## Problem
When a user encounters an error while updating a README, they see a "Retry" button, which isn't correct because the previous prompt is not retried. Instead, the user should be able to input a new prompt. 

When a user runs out of iterations in a session, the chat allows them to continue making README update requests, which results in an "Iteration limit exceeded error".

## Solution
If a user gets an error while updating a README, the chat input is enabled and the user can try another prompt. 

Once the user hits their iteration limit for the session, they are not able to suggest more changes and have to accept/reject the existing set of changes. This will reduce the "iteration limit exceeded" errors encountered by users.

A E2E test is added to verify that prompt errors during README updates are handled correctly with the relevant error message, and validates that users can enter a new prompt.

Screenshots:
_Prompt enabled after error updating README_
image

_'Make changes' button hidden once iteration limit is reached_
image

---


packages/amazonq/test/e2e/amazonq/doc.test.ts
```diff
@@ -146,5 +146,33 @@ describe('Amazon Q Doc', async function () {
                 FollowUpTypes.RejectChanges,
             ])
         })
+
+        it('Handle unrelated prompt error', async () => {
```
jpinkney-aws, line 150: Adding new tests 🎉 🎉 🎉 🎉 🎉 🎉

packages/core/src/amazonqDoc/errors.ts
```diff
@@ -7,69 +7,61 @@ import { ToolkitError } from '../shared/errors'
...
+        super(i18n('AWS.amazonq.doc.error.readmeUpdateTooLarge'), ReadmeUpdateTooLargeError.name, remainingIterations)
     }
 }
 
-export class WorkspaceEmptyError extends ToolkitError {
-    constructor() {
-        super(i18n('AWS.amazonq.doc.error.workspaceEmpty'), {
-            code: WorkspaceEmptyError.name,
-        })
+export class WorkspaceEmptyError extends DocServiceError {
```
vikshe, line 29: question: does this error count towards remaining iterations ? can there be some errors without remaining iterations returned from service?

avi-alpert: thats a good point, only errors related to the prompt (and that can be fixed with a prompt update) should include remaining iterations. PromptTooVagueError, PromptUnrelatedError, PromptRefusalException, and ReadmeUpdateTooLargeError

vikshe (approved):

vikshe (approved):

PR #6396, opened 2025-01-17
refactor(pollingset): make bad state impossible

Hweinstock:
## Problem
Right now `PollingSet` exposes both the `add` and the `start` methods. The `start` adds the item and starts the timer, while the `add` only adds the item. 

This allows a consumer to create a non-empty `PollingSet` where the timer never started. 

## Solution
- override the add method so that a consumer is forced to start the timer when adding items. 

---


justinmk3 (approved):
Nice!

justinmk3:
> Overwrite the add method so that a consumer is forced to start the timer when adding items.

it's only overriding, right? "overwrite" makes me think of `this.add = () => { ... }` :)

PR #6411, opened 2025-01-22
refactor(core): Manifest/LSP fetching should use httpResourceFetcher

jpinkney-aws:
## Problem
Now that we've refactored the httpResourceFetcher a little bit we can use that for fetching both the manifest and the language server

## Solution

---

### DIFF
--- a/packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
+++ b/packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
@@ -134,50 +134,6 @@ export class HttpResourceFetcher implements ResourceFetcher<Response> {
     }
 }
 
-export class RetryableResourceFetcher extends HttpResourceFetcher {
-    private readonly retryNumber: number
-    private readonly retryIntervalMs: number
-    private readonly resource: string
-
-    constructor({
-        resource,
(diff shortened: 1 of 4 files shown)


hayemaxi (approved):

jpinkney-aws:
Just moving this to draft, the web tests are failing because we currently stub out the httpResourceFetcher on web since there was previously node specific code. I just need to re-enable that

packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
```diff
@@ -133,50 +133,6 @@ export class HttpResourceFetcher implements ResourceFetcher<Response> {
     }
 }
 
-export class RetryableResourceFetcher extends HttpResourceFetcher {
```
justinmk3, line 136: Nice that we can eliminate this! Generally better to have composable concepts,  or flexible classes, instead of specific subclasses that implement particular compositions.

justinmk3 (approved):

packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
jpinkney-aws: Yeah that was one of the motivations for doing all this httpResourceFetcher stuff

PR #6418, opened 2025-01-23
fix(core): Could not fetch endpoints

jpinkney-aws:
## Problem
While starting the release process we found an problem where the endpoints couldn't be fetched because the resourceFetcher was expecting a string but the httpResourceFetcher was returning a Resource

## Solution
If its a string then return the contents from the resource fetcher directly, otherwise consider it as a response and get the text from the request

---


justinmk3 (approved):
do we not have test coverage for any of these codepaths? would be worth sending various args just to exercise the codepath.

hayemaxi (approved):

jpinkney-aws:
We relied to much on stubs :(. I'll add an integ test after the release

PR #6421, opened 2025-01-23
feat(stepfunctions): Support calling TestState API from Workflow Studio

anthonyting:
## Problem

The Workflow Studio webview currently does not allow for calling the [TestState API](https://docs.aws.amazon.com/step-functions/latest/dg/test-state-isolation.html). This API is used for testing individual states in isolation, and helps with debugging when constructing a state machine. It is available in the console version of Workflow Studio.

## Solution

Adding support for calling APIs from the webview using message passing. This is the added flow:

1. The webview sends a message to the extension to call sfn:TestState or iam:ListRoles
2. The extension performs the API call using its credentials and default credential region
3. The extension sends the response as a message to the webview

Note: this PR is dependent on [this PR](https://github.com/aws/aws-toolkit-vscode/pull/6375) being merged first since it requires an [aws-sdk version update](https://github.com/aws/aws-toolkit-vscode/pull/6375/commits/f4e0893d70dfe99b93988ea8eeb2caa4b210d8b3).

---

### DIFF
--- a/packages/core/src/shared/telemetry/vscodeTelemetry.json
+++ b/packages/core/src/shared/telemetry/vscodeTelemetry.json
@@ -567,10 +567,6 @@
             ]
         },
         {
-            "name": "stepfunctions_testState",
-            "description": "Called when the Step Functions TestState API is called"
-        },
-        {
             "name": "vscode_activeRegions",
             "description": "Record the number of active regions at startup and when regions are added/removed",
             "unit": "Count",
--- a/packages/core/src/shared/telemetry/vscodeTelemetry.json
+++ b/packages/core/src/shared/telemetry/vscodeTelemetry.json
@@ -567,6 +567,10 @@
             ]
         },
         {
+            "name": "stepfunctions_testState",
+            "description": "Called when the Step Functions TestState API is called"
+        },
+        {
             "name": "vscode_activeRegions",
             "description": "Record the number of active regions at startup and when regions are added/removed",
             "unit": "Count",
--- a/packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
+++ b/packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
@@ -13,6 +13,7 @@ import {
     FileChangedMessage,
     FileChangeEventTrigger,
     SyncFileRequestMessage,
+    ApiCallRequestMessage,
 } from './types'
 import { submitFeedback } from '../../feedback/vue/submitFeedback'
 import { placeholder } from '../../shared/vscode/commands2'
@@ -20,6 +21,8 @@ import * as nls from 'vscode-nls'
 import vscode from 'vscode'
 import { telemetry } from '../../shared/telemetry/telemetry'
 import { ToolkitError } from '../../shared/errors'
+import { WorkflowStudioClient } from './workflowStudioClient'
+import { ExtContext } from '../../shared'
 const localize = nls.loadMessageBundle()
 
 /**
@@ -27,8 +30,9 @@ const localize = nls.loadMessageBundle()
  * calls the appropriate handler function
  * @param message The message received from the webview
  * @param context The context object containing information about the webview environment
+ * @param extensionContext The extension context
  */
-export async function handleMessage(message: Message, context: WebviewContext) {
+export async function handleMessage(message: Message, context: WebviewContext, extensionContext: ExtContext) {
     const { command, messageType } = message
 
     if (messageType === MessageType.REQUEST) {
@@ -47,6 +51,9 @@ export async function handleMessage(message: Message, context: WebviewContext) {
                 break
             case Command.OPEN_FEEDBACK:
                 void submitFeedback(placeholder, 'Workflow Studio')
+                break
+            case Command.API_CALL:
+                void apiCallMessageHandler(message as ApiCallRequestMessage, context, extensionContext)
                 break
         }
     } else if (messageType === MessageType.BROADCAST) {
@@ -182,4 +189,19 @@ async function autoSyncFileMessageHandler(request: SyncFileRequestMessage, conte
             throw ToolkitError.chain(err, 'Could not autosave asl file.', { code: 'AutoSaveFailed' })
         }
     })
+}
+
+/**
+ * Handler for making API calls from the webview and returning the response.
+ * @param request The request message containing the API to call and the parameters
+ * @param context The webview context used for returning the API response to the webview
+ * @param extensionContext The extension context
+ */
+async function apiCallMessageHandler(
+    request: ApiCallRequestMessage,
+    context: WebviewContext,
+    extensionContext: ExtContext
+) {
+    const client = new WorkflowStudioClient(extensionContext.regionProvider.defaultRegionId, context)
+    await client.performApiCall(request)
 }
--- a/packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
+++ b/packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
@@ -21,8 +21,8 @@ import * as nls from 'vscode-nls'
 import vscode from 'vscode'
 import { telemetry } from '../../shared/telemetry/telemetry'
 import { ToolkitError } from '../../shared/errors'
-import { WorkflowStudioClient } from './workflowStudioClient'
-import { ExtContext } from '../../shared'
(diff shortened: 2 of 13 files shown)


packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
```diff
@@ -183,3 +190,18 @@ async function autoSyncFileMessageHandler(request: SyncFileRequestMessage, conte
...
+ * @param context The webview context used for returning the API response to the webview
+ * @param extensionContext The extension context
+ */
+async function apiCallMessageHandler(
+    request: ApiCallRequestMessage,
+    context: WebviewContext,
+    extensionContext: ExtContext
+) {
+    const client = new WorkflowStudioClient(extensionContext.regionProvider.defaultRegionId, context)
+    await client.performApiCall(request)
```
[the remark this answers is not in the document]
anthonyting: You're right, there's no error handling here so it doesn't need to be async

packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
```diff
@@ -13,22 +13,26 @@ import {
...
 
 /**
  * Handles messages received from the webview. Depending on the message type and command,
  * calls the appropriate handler function
  * @param message The message received from the webview
  * @param context The context object containing information about the webview environment
+ * @param extensionContext The extension context
  */
-export async function handleMessage(message: Message, context: WebviewContext) {
+export async function handleMessage(message: Message, context: WebviewContext, extensionContext: ExtContext) {
```
justinmk3, line 35: Note: our `globals` module is preferred and avoids the need to pass around `ExtContext`, which is deprecated.

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -566,6 +566,10 @@
                 }
             ]
         },
+        {
+            "name": "stepfunctions_testState",
```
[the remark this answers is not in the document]
anthonyting: I'm tracking the API call for TestState in stepFunctionsClient.ts. I was following a similar pattern I found for the redshift APIs, but you're right I can replace with uiClick since that's what will be triggering it in the webview.

packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
anthonyting: Got it, I can replace with the globals

packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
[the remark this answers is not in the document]
justinmk3: should there be error-handling? at least log any errors?

    client.performApiCall(...).catch(e => logger.error('stepFunctions: apiCall: %O', e))

packages/core/src/stepFunctions/workflowStudio/workflowStudioClient.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { IAM, StepFunctions } from 'aws-sdk'
+import { DefaultIamClient } from '../../shared/clients/iamClient'
+import { DefaultStepFunctionsClient } from '../../shared/clients/stepFunctionsClient'
+import { ApiAction, ApiCallRequestMessage, Command, MessageType, WebviewContext } from './types'
+
+export class WorkflowStudioClient {
```
justinmk3, line 11: why doesn't this live in `shared/clients` like all the other clients

anthonyting: It's not really an SDK client like those clients, and more of a client handler. I think it's better for this to be renamed to be `WorkflowStudioApiHandler` actually.

packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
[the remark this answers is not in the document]
anthonyting: Yeah it can error if there is a problem with awaiting on `webview.postMessage`. Error logging should be good for that case; I will add that.

packages/core/src/test/stepFunctions/workflowStudio/workflowStudioApiHandler.test.ts
```diff
@@ -0,0 +1,117 @@
...
+                    output: 'Test state output',
+                },
+                requestId: 'test-request-id',
+                isSuccess: true,
+            })
+        )
+    })
+
+    it('should handle request and response for error', async function () {
+        const panel = vscode.window.createWebviewPanel('WorkflowStudioMock', 'WorkflowStudioMockTitle', {
```
jpinkney-aws, line 68: It looks like the first 15 or so lines are the same between the two tests, does it make sense to just set these up in a beforeEach?

packages/core/src/test/stepFunctions/workflowStudio/workflowStudioApiHandler.test.ts
```diff
@@ -0,0 +1,101 @@
...
+                definition: '',
+                roleArn: '',
+            },
+            requestId: 'test-request-id',
+            command: Command.API_CALL,
+            messageType: MessageType.REQUEST,
+        })
+
+        assert(
+            postMessageStub.firstCall.calledWithExactly({
```
justinmk3, line 87: These tests are testing almost nothing. The tests stub the webview and the apihandler.

The tests would be much more meaningful if `WorkflowStudioApiHandler` has a way of checking its actual state. This is mentioned in the codebase guidelines.

packages/core/src/stepFunctions/workflowStudio/workflowStudioApiHandler.ts
```diff
@@ -0,0 +1,78 @@
...
+                          },
+                requestId,
+                isSuccess: false,
+            })
+        }
+    }
+
+    public async testState(params: StepFunctions.TestStateInput): Promise<StepFunctions.TestStateOutput> {
+        telemetry.ui_click.emit({
+            elementId: 'stepfunctions_testState',
```
justinmk3, line 70: We have `assertTelemetry`  https://github.com/aws/aws-toolkit-vscode/blob/085ad38d0f49590ce79cb87213af72973b34e040/packages/core/src/test/testUtil.ts#L322

that gives you a way to test the real codepaths, without stubs.

packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
```diff
@@ -179,7 +187,22 @@ async function autoSyncFileMessageHandler(request: SyncFileRequestMessage, conte
...
+ * Handler for making API calls from the webview and returning the response.
+ * @param request The request message containing the API to call and the parameters
+ * @param context The webview context used for returning the API response to the webview
+ */
+function apiCallMessageHandler(request: ApiCallRequestMessage, context: WebviewContext) {
+    const logger = getLogger()
+    const apiHandler = new WorkflowStudioApiHandler(globals.awsContext.getCredentialDefaultRegion(), context)
+    apiHandler
+        .performApiCall(request)
+        .catch((error) => logger.error('StepFunctions %s API call failed: %O', request.apiName, error))
```
justinmk3, line 207:
```suggestion
    const logger = getLogger('StepFunctions')
    const apiHandler = new WorkflowStudioApiHandler(globals.awsContext.getCredentialDefaultRegion(), context)
    apiHandler
        .performApiCall(request)
        .catch((error) => logger.error('%s API call failed: %O', request.apiName, error))
```

anthonyting: Added

packages/core/src/stepFunctions/workflowStudio/workflowStudioApiHandler.ts
anthonyting: Added

packages/core/src/test/stepFunctions/workflowStudio/workflowStudioApiHandler.test.ts
anthonyting: Updated so the API client is stubbed instead of the handler itself

PR #6423, opened 2025-01-23
feat(lambda): Connect CW Live Tail to Lambda functions

valerena:
Allow to "Tail Logs" from Lambda functions, which will get the configured Log Group for a particular function and start a Live Tail session in that Log Group.

Some code was removed, because with the recently updated version of the SDK v2, we can take the configured log group directly from the function's configuration and we don't need that extra call using SDK V3 that was being made inside `deployedNode.ts`

## Problem

To start a CloudWatch Live Tail session, a user needs to know their Log Group name and find it in the AWS explorer. For Lambda users, that might not be easy to find or well known all the time

## Solution

We help the Live Tail experience for Lambda customers by allowing them to start a Live Tail session directly from their function in the AWS explorer and in the Application Builder section for deployed functions, without having to search for their Log Group explicitly.

---

### DIFF
--- a/packages/core/package.nls.json
+++ b/packages/core/package.nls.json
@@ -91,6 +91,7 @@
     "AWS.command.appBuilder.deploy": "Deploy SAM Application",
     "AWS.command.appBuilder.build": "Build SAM Template",
     "AWS.command.appBuilder.searchLogs": "Search Logs",
+    "AWS.command.appBuilder.tailLogs": "Tail Logs",
     "AWS.command.refreshappBuilderExplorer": "Refresh Application Builder Explorer",
     "AWS.command.applicationComposer.openDialog": "Open Template with Infrastructure Composer...",
     "AWS.command.auth.addConnection": "Add New Connection",
--- a/packages/core/src/awsService/cloudWatchLogs/activation.ts
+++ b/packages/core/src/awsService/cloudWatchLogs/activation.ts
@@ -28,13 +28,13 @@ import { getLogger } from '../../shared/logger/logger'
 import { ToolkitError } from '../../shared'
 import { LiveTailCodeLensProvider } from './document/liveTailCodeLensProvider'
 
+export const liveTailRegistry = LiveTailSessionRegistry.instance
+export const liveTailCodeLensProvider = new LiveTailCodeLensProvider(liveTailRegistry)
 export async function activate(context: vscode.ExtensionContext, configuration: Settings): Promise<void> {
     const registry = LogDataRegistry.instance
-    const liveTailRegistry = LiveTailSessionRegistry.instance
 
     const documentProvider = new LogDataDocumentProvider(registry)
     const liveTailDocumentProvider = new LiveTailDocumentProvider()
-    const liveTailCodeLensProvider = new LiveTailCodeLensProvider(liveTailRegistry)
     context.subscriptions.push(
         vscode.languages.registerCodeLensProvider(
             {
@@ -150,7 +150,7 @@ export async function activate(context: vscode.ExtensionContext, configuration: 
     )
 }
 
-function getFunctionLogGroupName(configuration: any) {
+export function getFunctionLogGroupName(configuration: any) {
     const logGroupPrefix = '/aws/lambda/'
-    return configuration.logGroupName || logGroupPrefix + configuration.FunctionName
+    return configuration.LoggingConfig?.LogGroup || logGroupPrefix + configuration.FunctionName
 }
--- a/packages/core/src/lambda/activation.ts
+++ b/packages/core/src/lambda/activation.ts
@@ -4,6 +4,7 @@
  */
 
 import * as vscode from 'vscode'
+import { Lambda } from 'aws-sdk'
 import { deleteLambda } from './commands/deleteLambda'
 import { uploadLambdaCommand } from './commands/uploadLambda'
 import { LambdaFunctionNode } from './explorer/lambdaFunctionNode'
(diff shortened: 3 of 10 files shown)


packages/toolkit/.changes/next-release/Feature-48c399ea-d42f-4ea1-8f32-d8bc1486a107.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Support starting CloudWatch Logs Live Tail sessions for a specific Lambda function's log group, without having to find which log group manually, but just starting from the function directly."
```
justinmk3, line 3: suggestion:

```suggestion
	"description": "From the Lambda treeview in AWS Explorer, you can now right-click on a function name and start a CloudWatch Logs Live Tail sessions for the selected function."
```

packages/core/package.nls.json
```diff
@@ -91,6 +91,7 @@
     "AWS.command.appBuilder.deploy": "Deploy SAM Application",
     "AWS.command.appBuilder.build": "Build SAM Template",
     "AWS.command.appBuilder.searchLogs": "Search Logs",
+    "AWS.command.appBuilder.tailLogs": "Tail Logs",
```
justinmk3, line 94: We should probably use the same wording (and actually the same `AWS.foo.bar` ide, instead of multiple) in all places: https://github.com/aws/aws-toolkit-vscode/blob/8d48a3c3441c09bfb574aeb5d680f4114c5e0274/packages/core/package.nls.json#L181-L182

However, not a blocker for this PR. Needs a decision about which wording to use in both places.

packages/core/src/lambda/activation.ts
```diff
@@ -76,6 +82,40 @@ export async function activate(context: ExtContext): Promise<void> {
...
+                    logGroupInfo,
+                    logStreamFilterData
+                )
+            } catch (err) {
+                if (isError(err as Error, 'ResourceNotFoundException', "LogGroup doesn't exist.")) {
+                    // If we caught this error, then we know `functionConfiguration` actually has a value
+                    throw ToolkitError.chain(
+                        err,
+                        `Unable to fetch logs. Log group for function '${functionConfiguration!.FunctionName}' does not exist. ` +
+                            'Invoking your function at least once will create the log group.'
```
justinmk3, line 113: Nice hint! Very helpful to surface hints like this.

packages/core/src/lambda/activation.ts
```diff
@@ -76,6 +82,40 @@ export async function activate(context: ExtContext): Promise<void> {
...
+                    liveTailRegistry,
+                    source,
+                    liveTailCodeLensProvider,
+                    logGroupInfo,
+                    logStreamFilterData
+                )
+            } catch (err) {
+                if (isError(err as Error, 'ResourceNotFoundException', "LogGroup doesn't exist.")) {
+                    // If we caught this error, then we know `functionConfiguration` actually has a value
+                    throw ToolkitError.chain(
```
roger-zhangg, line 110: will we get this issue when we deploy the function for the first time, and hit taillog before invoking it?

justinmk3:
> Some code was removed, because with the recently updated version of the SDK v2, we can take the configured log group directly from the function's configuration and we don't need that extra call using SDK V3 that was being made inside `deployedNode.ts`

Thanks for noticing that, eliminating code is very helpful.

justinmk3 (approved):

packages/core/src/lambda/activation.ts
roger-zhangg: For I feel this should be a valid use case

packages/core/package.nls.json
valerena: The difference here is that the one you linked shows "when you have a log group selected" (therefore the wording is "tail log group" as in "tail *this* log group"), while this new one is when you have a *function* selected. Maybe this could be "Tail function logs" instead of just "Tail logs". The inspiration was the other existing string right above "Search Logs" (with its corresponding "Search Log Group"), which is a similar behavior, but that starts a search instead of using live tail.


Let me know if this difference makes sense or if we should still try to unify the wording.

packages/core/src/awsService/cloudWatchLogs/activation.ts
```diff
@@ -150,7 +150,7 @@ export async function activate(context: vscode.ExtensionContext, configuration:
     )
 }
 
-function getFunctionLogGroupName(configuration: any) {
+export function getFunctionLogGroupName(configuration: any) {
     const logGroupPrefix = '/aws/lambda/'
-    return configuration.logGroupName || logGroupPrefix + configuration.FunctionName
+    return configuration.LoggingConfig?.LogGroup || logGroupPrefix + configuration.FunctionName
```
roger-zhangg, line 155: is this because SDK update?

packages/toolkit/.changes/next-release/Feature-48c399ea-d42f-4ea1-8f32-d8bc1486a107.json
valerena: It does look clearer. I updated with this suggestion.

packages/core/package.nls.json
justinmk3: > The inspiration was the other existing string right above "Search Logs"

Yup I noticed that too, and I like the brevity of "Logs" instead of "Log Group". For now let's leave it as you have in this PR.

packages/core/src/lambda/activation.ts
valerena: Yeah, and that's how the Live Tail API works. It needs an existing log group. We could eventually do something like try once and if it fails we retry (but we might keep retrying forever), but in practice it's better to surface the error, so customers understand it better and retry when needed. Live Tail use case is to "see and filter logs", so it's less likely that people want to use it a lot for a brand new function that doesn't have any invokes yet.

packages/core/src/awsService/cloudWatchLogs/activation.ts
valerena: It's related. The code that was removed in the other file used to set up this `logGroupName` variable that wasn't in the original configuration, but here we're just taking it directly from the Lambda Function configuration object that comes from the SDK.

roger-zhangg (approved):
LGTM

PR #6425, opened 2025-01-24
test(toolkit): Add region provider remote test

jpinkney-aws:
## Problem
In the release yesterday we found an issue where fetching remote region data was failing

## Solution
Add a new test since its critical to the flow

---

### DIFF
--- a/packages/core/src/shared/regions/regionProvider.ts
+++ b/packages/core/src/shared/regions/regionProvider.ts
@@ -162,13 +162,20 @@ export class RegionProvider {
         remote: () => Endpoints | Promise<Endpoints>
     }): RegionProvider {
         const instance = new this()
+        void instance.init(endpointsProvider)
+        return instance
+    }
 
-        async function load() {
+    async init(endpointsProvider: {
+        local: () => Endpoints | Promise<Endpoints>
+        remote: () => Endpoints | Promise<Endpoints>
+    }) {
+        const load = async () => {
             getLogger().info('endpoints: retrieving AWS endpoints data')
-            instance.loadFromEndpoints(await endpointsProvider.local())
+            this.loadFromEndpoints(await endpointsProvider.local())
 
             try {
-                instance.loadFromEndpoints(await endpointsProvider.remote())
+                this.loadFromEndpoints(await endpointsProvider.remote())
             } catch (err) {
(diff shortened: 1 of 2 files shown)


packages/core/src/shared/regions/regionProvider.ts
```diff
@@ -162,13 +162,20 @@ export class RegionProvider {
         remote: () => Endpoints | Promise<Endpoints>
     }): RegionProvider {
         const instance = new this()
+        void instance.init(endpointsProvider)
+        return instance
+    }
 
-        async function load() {
+    async init(endpointsProvider: {
```
nkomonen-amazon, line 169: Should we switch this to a try/catch at this point? Dropping the `const load`

packages/core/src/testInteg/regionProvider.test.ts
```diff
@@ -0,0 +1,46 @@
...
+import globals from '../shared/extensionGlobals'
+import { makeTemporaryToolkitFolder } from '../shared/filesystemUtilities'
+import { fs } from '../shared/fs/fs'
+import path from 'path'
+
+describe('Region Provider', async function () {
+    let tempDir: string
+
+    before(async () => {
+        tempDir = await makeTemporaryToolkitFolder()
```
nkomonen-amazon, line 18: Unless there is a reason, we should be using `TestFolder.create()` from `testUtils.ts` to create a temp folder. And it will also clean itself up, so all you would need to do is use a `beforeEach()`

packages/core/src/testInteg/regionProvider.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    })
+
+    after(async () => {
+        await fs.delete(tempDir, {
+            recursive: true,
+            force: true,
+        })
+    })
+
+    it('resolves from remote', async function () {
```
nkomonen-amazon, line 28: nice and simple test

justinmk3 (approved):
LGTM after comments, nice choices!

packages/core/src/shared/regions/regionProvider.ts
jpinkney-aws: just played around with this. let me know what you think

PR #6430, opened 2025-01-24
fix(amazonq): Allow including binary files when execution is enabled

osdemah:
## Problem

* Fixes for execution engine

## Solution

* Don't drop binary files matching with extra gitignore rules
* Don't drop `.gitignore` file from repository.

### DIFF
--- a/packages/core/src/shared/filetypes.ts
+++ b/packages/core/src/shared/filetypes.ts
@@ -363,15 +363,8 @@ export const codefileExtensions = new Set([
 // Code file names without an extension
 export const codefileNames = new Set(['Dockerfile', 'Dockerfile.build', 'gradlew', 'mvnw', '.gitignore'])
 
-// Build file names
-export const buildfileNames = new Set(['gradle/wrapper/gradle-wrapper.jar'])
-
 /** Returns true if `filename` is a code file. */
 export function isCodeFile(filename: string): boolean {
     const ext = path.extname(filename).toLowerCase()
-    return (
-        codefileExtensions.has(ext) ||
-        codefileNames.has(path.basename(filename)) ||
-        buildfileNames.has(path.basename(filename))
-    )
+    return codefileExtensions.has(ext) || codefileNames.has(path.basename(filename))
 }
--- a/packages/core/src/shared/filetypes.ts
+++ b/packages/core/src/shared/filetypes.ts
@@ -233,7 +233,6 @@ export const codefileExtensions = new Set([
     '.idl',
     '.ini',
     '.io',
-    '.jar',
     '.java',
     '.jl',
     '.js',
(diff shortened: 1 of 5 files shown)


packages/core/src/shared/filetypes.ts
```diff
@@ -362,7 +361,7 @@ export const codefileExtensions = new Set([
 ])
 
 // Code file names without an extension
-export const codefileNames = new Set(['Dockerfile', 'Dockerfile.build', 'gradlew', 'mvnw'])
+export const codefileNames = new Set(['Dockerfile', 'Dockerfile.build', 'gradlew', 'mvnw', '.gitignore'])
 
 // Build file names
 export const buildfileNames = new Set(['gradle/wrapper/gradle-wrapper.jar'])
```
justinmk3, line 367: This should be in `codefileNames`. It's literally being treated as a codefile, so the separate variable is just confusing.

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -268,7 +268,7 @@ export function checkUnsavedChanges(): boolean {
     return vscode.workspace.textDocuments.some((doc) => doc.isDirty)
 }
 
-export function getExcludePattern(additionalPatterns: string[] = []) {
+export function getExcludePattern(additionalPatterns: string[] = [], nonGlobalExtraPatterns: boolean = true) {
```
justinmk3, line 271: could be less convoluted.

The fact that `extraPatterns` looks redundant after `additionalPatterns` is just signaling problems with this interface design. Those problems aren't avoided by using convoluted parameter names.

```suggestion
export function getExcludePattern(additionalPatterns: string[] = [], extraPatterns: boolean = true) {
```

osdemah: good call, yeah, we shouldn't have too many params. I actually noticed there is no usage for `additionalPatterns`, so only the `defaultExcludePatterns` is sufficient. updated the naming to better reflect the behavior.

packages/core/src/shared/filetypes.ts
osdemah: missed in my rebase. we actually don't need to particularly check this jar file since we need any other files included in repository when we want to build the code (e.g. icon/png files in the TS repos). so removed `buildfileNames`

neilk-aws (approved):

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -290,19 +290,24 @@ export function getExcludePattern(additionalPatterns: string[] = []) {
...
         '**/LICENSE.md',
     ]
-    const allPatterns = [...globAlwaysExcludedDirs, ...extraPatterns, ...additionalPatterns]
+    const allPatterns = [...globAlwaysExcludedDirs, ...(defaultExcludePatterns ? extraPatterns : [])]
     return `{${allPatterns.join(',')}}`
 }
 
 /**
  * @param rootPath root folder to look for .gitignore files
+ * @param addExtraPatterns whether to add extra exclude patterns even if not in gitignore
```
justinmk3, line 299: outdated

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -313,13 +318,15 @@ async function filterOutGitignoredFiles(rootPath: string, files: vscode.Uri[]):
  * @param sourcePaths the paths where collection starts
  * @param workspaceFolders the current workspace folders opened
  * @param respectGitIgnore whether to respect gitignore file
+ * @param addExtraIgnorePatterns whether to add extra exclude patterns even if not in gitignore
```
justinmk3, line 321: outdated

PR #6439, opened 2025-01-27
deps(cw): regen clients with codegen at 0.25.x

Hweinstock:
## Problem
The two streaming clients committed to the repo are generated via different versions of `smithy-typescript-codegen`. Each version of codegen pins the generated client to use specific versions of `@aws-sdk/*` and `@smithy/*` packages. The root project also consumes different versions of these packages. Therefore, the project is currently consuming three different versions of many `@aws-sdk/*` and `@smithy/*` packages. 

This is problematic because it can cause dependency conflicts. Certain versions of `@aws-sdk/*` and `@smithy/*` packages are only compatible with certain other versions. For more information on this, see the discussion involving help from an SDK team member explaining this problem. 

## Solution
- Regenerate both clients with the same version of codegen `0.25.x`. 
- Pin all necessary versions to ensure a single version of `@aws-sdk/*` and `@smithy/*` packages.
- Add documentation to the build instructions that these clients must stay in-sync (in terms of codegen). 
- Add documentation regarding updating `@aws-sdk/*` packages. 

---


trivikr:
The codegen@0.25.0 published `<v3.696.0` of JS SDK v3 in https://github.com/aws/aws-sdk-js-v3/pull/6673

Since this is a monorepo which has a single lock file across all clients, the version of client packages need to be updated to `<3.696.0` in https://github.com/Hweinstock/aws-toolkit-vscode/blob/3c37c5777ac395c5f1b59da3d7599b1e554210d8/packages/core/package.json#L499-L504

trivikr:
As there's another codegen client `@amzn/amazon-q-developer-streaming-client`, it should also be codegen with `0.25.x` as it's also part of monorepo.

https://github.com/aws/aws-toolkit-vscode/blob/bdbd8e820ab4e8cab96ab697a841a238a2ee3895/src.gen/%40amzn/amazon-q-developer-streaming-client/package.json#L23-L34

It currently uses old codegen version.

Hweinstock:
I pinned the `@aws-sdk/*` packages to be `<3.696.0` and regenerated the other client with codegen version `0.25.x`. Are there any other steps I should take here? Otherwise planning to retry with `0.26.x` once its available.

trivikr:
> I pinned the `@aws-sdk/*` packages to be `<3.696.0`

I see that you've use exact version `3.696.0`.
https://github.com/Hweinstock/aws-toolkit-vscode/blob/b1470ed160aa114012e8cbee9dbac00969c9a8df/packages/core/package.json#L499-L510

Can you use `<3.696.0` for all `@aws-sdk/*` dependencies?

There's no issue with using exact version as it'll be just one version above previous one.
But since all JS SDK clients follow fixed versioning, and we don't release new versions if there are no updates in specific clients or it's dependencies, we suggested `<`.

trivikr:
For a detailed explanation, if you search for `@aws-sdk/client-sso-oidc` in package-lock.json, it contains both `3.693.0` and `3.696.0`
https://github.com/Hweinstock/aws-toolkit-vscode/blob/b1470ed160aa114012e8cbee9dbac00969c9a8df/package-lock.json
* It contains `3.693.0` since generated clients depend on it https://github.com/Hweinstock/aws-toolkit-vscode/blob/b1470ed160aa114012e8cbee9dbac00969c9a8df/src.gen/%40amzn/amazon-q-developer-streaming-client/package.json#L23
* It contains `3.696.0` since core depends on it https://github.com/Hweinstock/aws-toolkit-vscode/blob/b1470ed160aa114012e8cbee9dbac00969c9a8df/packages/core/package.json#L504

If you use `<3.696.0` on core package, it'll use `3.693.0`.
Similarly if you pin other non-client `@aws-sdk/*` packages to `<3.696.0`, other dependency conflicts will get resolved.

Hweinstock:
I see, thanks for the explanation that makes a lot of sense. I switched it to to be `<3.696.0` and verified there is only one version (`3.693.0`) of `@aws-sdk/client-sso-oidc` in the `package-lock` now. Unfortunately,  it doesn't seem to solve the smithy package conflicts, but this version pinning done here has eliminated all the errors I was originally seeing with type mismatches so I think we're getting closer.

trivikr:
I'm glad that versioning issues with `@aws-sdk/*` are no longer there.

Do update your Contributor Guide with learnings from this PR. The AWS SDK for JavaScript team will take a backlog to write a doc on versioning from this experience - possibly in our Developer Guide. We'll get it reviewed from you internally before publishing.

trivikr:
Tackling `@smithy/*` versioning is little tricky, but we use latest for those packages as they can be used outside of AWS too. Your monorepo can do the same.

I see that you're using 5 dependencies from `@smithy`
```json
        "@smithy/middleware-retry": "^2.3.1",
        "@smithy/protocol-http": "^3.3.0",
        "@smithy/service-error-classification": "^2.1.5",
        "@smithy/shared-ini-file-loader": "^2.2.8",
        "@smithy/util-retry": "^2.2.0",
```
https://github.com/Hweinstock/aws-toolkit-vscode/blob/83a18cc84e7a4317048b4ddc12357c65ae310e3e/packages/core/package.json#L514-L518

An easy fix would be to copy the same version used by your [code generated packages](https://github.com/Hweinstock/aws-toolkit-vscode/blob/83a18cc84e7a4317048b4ddc12357c65ae310e3e/src.gen/%40amzn/amazon-q-developer-streaming-client/package.json#L19). i.e.
```jsonc
        "@smithy/middleware-retry": "^3.0.27",
        "@smithy/protocol-http": "^4.1.7",
        "@smithy/service-error-classification": "^3.0.11", // not direct client dependency, found from package-lock
        "@smithy/shared-ini-file-loader": "^3.1.12", // not direct client dependency, found from package-lock
        "@smithy/util-retry": "^3.0.10",
```

Since it's the latest major version, you can use `^N.0.0` if it simplifies your tooling
```json
        "@smithy/middleware-retry": "^3.0.0",
        "@smithy/protocol-http": "^4.0.0",
        "@smithy/service-error-classification": "^3.0.0",
        "@smithy/shared-ini-file-loader": "^3.0.0",
        "@smithy/util-retry": "^3.0.0",
```

kuhe:
> * generate all clients from the same version of codegen. (in this case `0.25.x`).

yes

> * fix `@aws-sdk/*` package versions in the root to be no-newer than the version fixed by codegen. (in this case `<3.696.0`)

yes, but use the highest version you can see that was output by code generation. If code generation output contains:
```
@aws-sdk/core: "3.735.0",
@aws-sdk/some-package: "3.733.0",
```
use `<=3.735.0` for everything.

> * fix `@smithy/*` package versions to the same major versions used by the generated clients (including indirect dependencies from the `package-lock.json`) on a **per package basis**. I emphasize the last point because its not intuitive that the generated client uses `"@smithy/middleware-retry": "^3.0.27"` and  `"@smithy/protocol-http": "^4.1.7"` despite coming from the same codegen version, so one needs to check that each package lines up.

You shouldn't need to specify any `@smithy/*` package versions. Code generation will output e.g. `^5.x.y` of some smithy package, and because you've pinned the `@aws-sdk/client-*` packages to the contemporaneous version, e.g. `@aws-sdk/client-s3@<=3.735.0`, the AWS SDK clients are using the same range of smithy dependency. This means there will only be one instance of any given core smithy or core aws-sdk package(s).
 
> * Is there any recommended way to go about enforcing these constraints? Any solutions from other teams that have been effective?

No

> * Is this something we will always have to deal with when consuming generated typescript clients from smithy? Are there any plans to improve this experience?

Yes. We may publish smithy-typescript maven versions with higher frequency in the future. We may create a dynamic client that accepts a model at runtime in the future, long term.

> * Since we don't own the packages that generate these clients, is there anything we can do when their codegen versions inevitably diverge? Is our only option to block merges/changes to them in this repo until they are in sync?

Sounds like yes.

> * How do we determine the version to fix based on the codegen version? Does it require manually looking at the generated `package.json` and taking the newest version of an `@aws-sdk/*` package?

Yes

> * What happens if we need to update a package we consume in the toolkit but there isn't a version of codegen available that lets us pin to a newer version?

use NPM overrides or request a newer release of the code generators. 

> * Can you provide some insight into why this happens? I was under the impression that `npm` allowed packages to manage their own dependencies, so you could have different versions of the same package existing in the same repository, where each package used the version of dependencies it needed. However, if that were the case, this shouldn't be happening so what am I missing here?

NPM duplication resolves runtime conflicts of packages needing two different versions of some NPM package. This creates a class of runtime issues relating to inheritance checks but otherwise allows deconfliction. 

TypeScript is a fictional overlay that is of no concern to the package duplication system. Type errors from duplication are not addressed by it.

jpinkney-aws (approved):
> Add documentation to the build instructions that these clients must stay in-sync (in terms of codegen).
> Add documentation regarding updating @aws-sdk/* packages.

Is that internal documentation? I'm guessing the generator update instructions?

Hweinstock:
Yes, its internal. Could be worth summarizing it and adding it into the public facing docs at some point. This problem mostly affects developers trying to regenerate the streaming clients, so I think the public use-case is minimal.

justinmk3 (approved):

nkomonen-amazon (approved):

leigaol (approved):

leigaol:
The vsix from the package test is working fine in my VS Code after installation.

PR #6442, opened 2025-01-27
test(sdkv3): add tests for expiring credentials

Hweinstock:
## Problem
The client should not be making requests with stale credentials so we want test coverage to ensure this

## Solution
- Create a mock credentials shim and insert it into the global context.
- The client builder will then use that shim, and we can manually expire it.
- Ensure client does not cache potentially stale credentials.

---

### DIFF
--- a/packages/core/src/test/shared/awsClientBuilderV3.test.ts
+++ b/packages/core/src/test/shared/awsClientBuilderV3.test.ts
@@ -16,6 +16,7 @@ import { assertTelemetry } from '../testUtil'
 import { telemetry } from '../../shared/telemetry'
 import { CredentialsShim } from '../../auth/deprecated/loginManager'
 import { Credentials } from '@aws-sdk/types'
+import { oneDay } from '../../shared/datetime'
 
 describe('AwsClientBuilderV3', function () {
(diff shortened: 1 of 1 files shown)


packages/core/src/test/shared/awsClientBuilderV3.test.ts
```diff
@@ -57,6 +59,52 @@ describe('AwsClientBuilderV3', function () {
...
+        let oldCreds: Credentials
+        let newCreds: Credentials
+
+        beforeEach(function () {
+            fakeContext = new FakeAwsContext()
+            oldCreds = {
+                accessKeyId: 'old',
+                secretAccessKey: 'old',
+                sessionToken: 'old',
+                expiration: new Date(Date.now() + 1000 * 60 * 60 * 24),
```
hayemaxi, line 75: use these https://github.com/aws/aws-toolkit-vscode/blob/2ff551866b663b89b2470c3408289a6452213765/packages/core/src/shared/datetime.ts#L6

hayemaxi (approved):

jpinkney-aws (approved):

PR #6446, opened 2025-01-28
fix(amazonq): Seperate inline completion from regular codewhisperer activation

jpinkney-aws:
## Problem
amazonqLSP experiments wraps chat instead of codewhisperer inline activation

## Solution
Pull out codewhisperer inline activation into its own function and then wrap that activation instead

## TODO:
~- Re-implement any commands from codewhisperer that need both lsp and non lsp implementations (e.g. manualInvocation)~

---


packages/amazonq/src/inline/activation.ts
justinmk3, line 1: The app/ folder is intended to hold features. idk why inlineChat is at the top level. This module should live in `app/inline/`

justinmk3 (approved):

Issue #6451, opened 2025-01-28
unreliable test: `FileSystem: rename() source file does not exist at first, but eventually appears`


Hweinstock:
## Test Details

-   Name of test: FileSystem: rename() source file does not exist at first, but eventually appears
-   OS of failure: macOs
-   VSCode version of failure (minimum/insider/stable): minimum
-   Link to failing run: https://github.com/aws/aws-toolkit-vscode/actions/runs/12999361648/job/36254529169
-   Link to failing test: https://github.com/aws/aws-toolkit-vscode/blob/bdbd8e820ab4e8cab96ab697a841a238a2ee3895/packages/core/src/test/shared/fs/fs.test.ts#L410-L425

## Log of Test Failure
```
1) FileSystem
       rename()
         source file does not exist at first, but eventually appears:

      AssertionError [ERR_ASSERTION]: telemetry metric not found: "ide_fileSystem"
      + expected - actual

      -false
      +true
```

Hweinstock:
Fixed by https://github.com/aws/aws-toolkit-vscode/pull/6452

PR #6452, opened 2025-01-28
test(flaky): avoid relying on race condition for tested behavior

Hweinstock:
## Problem
https://github.com/aws/aws-toolkit-vscode/issues/6451

This rename test relies on a specific result of a race condition for the expected result. 
- The test is checking for a telemetry result that is only emitted when `fs.exists` takes more than 1 attempt to resolve to true. 
- Therefore, it wants the first `fs.exists` check to fail, then a subsequent one to succeed. 
- It does this by not awaiting the result, and then writing the file to be renamed. 

Usually this is fine, but it is possible that the write (`toFile`) happens before the read (`fs.exists`) since neither is awaited. This behavior leads to a flaky test as described in the issue. 

## Solution
- use a stub to force the first call to `fs.exists` to fail. 
- allow all other calls to "go through" to the original function. 
- Have rename emit telemetry when it succeeds on the first attempt. 

---

### DIFF
--- a/packages/core/src/shared/fs/fs.ts
+++ b/packages/core/src/shared/fs/fs.ts
@@ -312,6 +312,11 @@ export class FileSystem {
                 reasonDesc: `After multiple attempts the source path existed: ${scrubbedPath}`,
                 attempts: attempts,
             })
+        } else {
+            telemetry.ide_fileSystem.emit({
+                result: 'Succeeded',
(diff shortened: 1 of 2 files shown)


justinmk3 (approved):
Great explanation!

jpinkney-aws (approved):

Issue #6457, opened 2025-01-29
unreliable test: prepareRepoData Performance Tests


Hweinstock:
## Test Details

-   Name of test: prepareRepoData Performance Tests
-   OS of failure: Linux (Integ)
-   VSCode version of failure (minimum/insider/stable): stable
-   Link to failing run:https://github.com/aws/aws-toolkit-vscode/commit/e54d549c6cb8d94c3c5765446594da0dd471a28e
-   Link to failing test: https://github.com/aws/aws-toolkit-vscode/blob/085ad38d0f49590ce79cb87213af72973b34e040/packages/core/src/testInteg/perf/prepareRepoData.test.ts#L70-L80

## Log of Test Failure
```
AssertionError [ERR_ASSERTION]: Expected total user CPU usage for handles 10 files of size 1000 bytes to be less than 200. Actual user CPU usage was 240.89446059590207
    at assertPerformanceMetrics (/codebuild/output/src3874270705/src/github.com/aws/aws-toolkit-vscode/packages/core/src/shared/performance/performance.ts:182:11)
    at Context.<anonymous> (/codebuild/output/src3874270705/src/github.com/aws/aws-toolkit-vscode/packages/core/src/shared/performance/performance.ts:160:13)
    at Context.fn (/codebuild/output/src3874270705/src/github.com/aws/aws-toolkit-vscode/packages/core/src/test/setupUtil.ts:34:24)
    at processImmediate (node:internal/timers:483:21)
    at process.topLevelDomainCallback (node:domain:161:15)
    at process.callbackTrampoline (node:internal/async_hooks:128:24)

      + expected - actual

      -false
      +true
      ```

Hweinstock:
skipped in https://github.com/aws/aws-toolkit-vscode/pull/6459

Issue #6117, opened 2024-11-28
unreliable test: zipCode performance tests


justinmk3:
## Test Details

-   Name of test: `zipCode performance tests`
-   OS of failure: Linux
-   VSCode version of failure (minimum/insider/stable): `Integration Tests - insiders`
-   Link to failing run: `Integration Tests - insiders` on https://github.com/aws/aws-toolkit-vscode/commit/5766e4b29440e61c8d0351c467cae4c2865ff961

## Log of Test Failure

```
  1) zipCode
       performance tests
         zipCode performance tests
           "after all" hook for "zipCode - test run 10":

      AssertionError [ERR_ASSERTION]: Expected total user CPU usage for zipCode to be less than 200. Actual user CPU usage was 277.1372267951132
      + expected - actual

      -false
      +true
```

justinmk3:
```
  1) zipCode
       performance tests
         zipCode performance tests
           "after all" hook for "zipCode - test run 10":

      AssertionError [ERR_ASSERTION]: Expected total user CPU usage for zipCode to be less than 350. Actual user CPU usage was 388.65274407871664
      + expected - actual
```

https://github.com/aws/aws-toolkit-vscode/commit/f004173fafe2ffa1863bbb3970b5328ffd14725e

Hweinstock:
Skipped in https://github.com/aws/aws-toolkit-vscode/pull/6459

PR #6459, opened 2025-01-29
test(flaky): increase thresholds for flaky performance tests.

Hweinstock:
## Problem
https://github.com/aws/aws-toolkit-vscode/issues/6117
https://github.com/aws/aws-toolkit-vscode/issues/6457

## Solution
increase thresholds

---


jpinkney-aws:
I wonder if we should just start skipping these since they're generating so much flakiness? Maybe we can re-visit them when they're in flare and gets a much more accurate number

Hweinstock:
Yeah, I think thats likely for the best. It doesn't seem to matter how hight the thresholds are set, it somehow always breaks it.

PR #6463, opened 2025-01-29
deps(amazonq): upgrade adm-zip to 0.5.16

avi-alpert:
## Problem
Users get an error uploading files when using /dev or /doc if files modified/created time is before the MSDOS epoch. Logs show an out of range exception thrown from the adm-zip package. Adm-zip issue reported [here](https://github.com/cthackers/adm-zip/issues/485)

## Solution
Upgrade adm-zip to the latest version. Use posix join to test paths in a zip, since adm-zip was updated to be POSIX compliant https://github.com/cthackers/adm-zip/issues/438

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10202,10 +10202,11 @@
             }
         },
         "node_modules/adm-zip": {
-            "version": "0.5.10",
-            "license": "MIT",
+            "version": "0.5.16",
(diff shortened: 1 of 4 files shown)


jpinkney-aws:
I think this change is fine but just for a heads up, instead of using adm-zip we should be migrating to our zipStream utility instead. It will allow us to eventually run agents in the browser + I _think_? resolve some memory issues as well

avi-alpert:
Will look into switching to zipstream

PR #6468, opened 2025-01-30
fix(amazonq): Update initializationOptions extension name

jpinkney-aws:
## Problem
Initialization name is just dummy text for now. We should consider updating it to the c9/toolkit one

## Solution

---

### DIFF
--- a/packages/core/src/shared/extensionUtilities.ts
+++ b/packages/core/src/shared/extensionUtilities.ts
@@ -49,8 +49,8 @@ export function productName() {
 }
 
 /** Gets the client name stored in oidc */
-export const clientName = once(_clientName)
-function _clientName() {
+export const oidcClientName = once(_oidcClientName)
+function _oidcClientName() {
     const companyName = getIdeProperties().company
     return isCloud9() ? `${companyName} Cloud9` : `${companyName} IDE Extensions for VSCode`
 }
--- a/packages/core/src/shared/extensionUtilities.ts
+++ b/packages/core/src/shared/extensionUtilities.ts
@@ -48,6 +48,13 @@ export function productName() {
     return isAmazonQ() ? 'Amazon Q' : `${getIdeProperties().company} Toolkit`
 }
 
+/** Gets the client name stored in oidc */
+export const clientName = once(_clientName)
+function _clientName() {
+    const companyName = getIdeProperties().company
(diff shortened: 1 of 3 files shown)


packages/core/src/auth/utils.ts
```diff
@@ -800,3 +801,9 @@ export async function getAuthType() {
     }
     return authType
 }
+
+export const clientName = once(_clientName)
+function _clientName() {
```
justinmk3, line 806: Thanks for cleaning this up! Should it live in `extensionUtilities` next to the other similar functions (`productName()`, etc)?

jpinkney-aws: I was wondering that as well. On the one hand the current implementation keeps it next to auth and I'm not sure that non-auth projects have a use for this? On the other hand like you mentioned, productName() etc lives in extensionUtils. I'm fine with either tbh

justinmk3: perhaps mention this new function in the docstring of productName(), and vice-versa?

packages/core/src/shared/extensionUtilities.ts
```diff
@@ -48,6 +48,13 @@ export function productName() {
     return isAmazonQ() ? 'Amazon Q' : `${getIdeProperties().company} Toolkit`
 }
 
+/** Gets the client name stored in oidc */
```
Hweinstock, line 51: potentially a dumb question, but what is "oidc"?

Hweinstock (approved):

packages/core/src/shared/extensionUtilities.ts
```diff
@@ -48,6 +48,13 @@ export function productName() {
...
+export const clientName = once(_clientName)
```
justinmk3, line 52:
```suggestion
export const oidcClientName = once(_clientName)
```

packages/core/src/shared/extensionUtilities.ts
jpinkney-aws: I just stole this from one of the comments in the codebase 😄:

IAM Identity Center OpenID Connect (OIDC) is a web service that enables a client (such as CLI or a native application) to register with IAM Identity Center. The service also enables the client to fetch the user’s access token upon successful authentication and authorization with IAM Identity Center

PR #6426, opened 2025-01-24
feat(release): check for marketplace update

tomcat323:
## Problem
Our release MCM currently has a step for each extension (Toolkit and Q) that requires the human operator to wait and check for the marketplace publish and then try to install from the marketplace. This uses human time and can easily be automated.

## Solution
Added a final stage in release pipeline: https://code.amazon.com/reviews/CR-173945904/revisions/1#/commits

that runs this YML build script that automatically installs and uninstalls the extension from marketplace on a two minute interval. 

When the version matches the new release version, the code-build succeeds:
Screenshot 2025-01-23 at 2 44 45 PM

When it doesn't, it keeps retrying until the timeout limit is reached (1 hour):
Screenshot 2025-01-23 at 4 52 07 PM

---

- Treat all work as PUBLIC. Private `feature/x` branches will not be squash-merged at release time.
- Your code changes must meet the guidelines in [CONTRIBUTING.md](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#guidelines).
- License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6470, opened 2025-01-30
fix(release): added beta check and dummy version logics

tomcat323:
## Problem
in https://github.com/aws/aws-toolkit-vscode/pull/6426, a new release pipeline stage was added that checks the marketplace extension has been updated. This PR addes a stage check and only look for an existing version of the extension when in BETA.

Output in beta:
Screenshot 2025-01-30 at 12 45 07 PM

unblocks: https://pipelines.amazon.com/pipelines/AWSVSCodeInfrastructure

### DIFF
--- a/buildspec/release/70checkmarketplace.yml
+++ b/buildspec/release/70checkmarketplace.yml
@@ -34,7 +34,6 @@ phases:
                   echo checkmarketplace: "Unknown TARGET_EXTENSION: ${TARGET_EXTENSION}"
                   exit 1
                 fi
-            - |
                 if [ "$STAGE" != "prod" ]; then
                   echo "checkmarketplace: Non-production stage detected. Installing hardcoded version '${VERSION}'."
                 fi
@@ -42,7 +41,7 @@ phases:
             - |
                 while true; do
                   code --uninstall-extension "${extension_name}" --no-sandbox --user-data-dir /tmp/vscode
-                  code --install-extension "${extension_name}@${VERSION}" --no-sandbox --user-data-dir /tmp/vscode
+                  code --install-extension "${extension_name}@${VERSION}" --no-sandbox --user-data-dir /tmp/vscode || true
                   cur_version=$(code --list-extensions --show-versions --no-sandbox --user-data-dir /tmp/vscode | grep ${extension_name} | cut -d'@' -f2)
                   if [ "${cur_version}" = "${VERSION}" ]; then
                     echo "checkmarketplace: Extension ${extension_name} is updated to version '${cur_version}.'"
--- a/buildspec/release/70checkmarketplace.yml
+++ b/buildspec/release/70checkmarketplace.yml
@@ -22,21 +22,31 @@ phases:
     build:
         commands:
             - VERSION=$(node -e "console.log(require('./packages/${TARGET_EXTENSION}/package.json').version);")
-            # get extension name
+            # get extension name, also an old version for beta testing
             - |
                 if [ "${TARGET_EXTENSION}" = "amazonq" ]; then
                   extension_name="amazonwebservices.amazon-q-vscode"
+                  beta_version="@1.43.0"
                 elif [ "${TARGET_EXTENSION}" = "toolkit" ]; then
                   extension_name="amazonwebservices.aws-toolkit-vscode"
+                  beta_version="@3.42.0"
                 else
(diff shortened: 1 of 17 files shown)


buildspec/release/70checkmarketplace.yml
```diff
@@ -22,21 +22,31 @@ phases:
     build:
         commands:
             - VERSION=$(node -e "console.log(require('./packages/${TARGET_EXTENSION}/package.json').version);")
-            # get extension name
+            # get extension name, also an old version for beta testing
             - |
                 if [ "${TARGET_EXTENSION}" = "amazonq" ]; then
                   extension_name="amazonwebservices.amazon-q-vscode"
+                  beta_version="@1.43.0"
```
justinmk3, line 29:
```suggestion
                  extension_version="@1.43.0"
```

buildspec/release/70checkmarketplace.yml
```diff
@@ -22,21 +22,31 @@ phases:
...
                 elif [ "${TARGET_EXTENSION}" = "toolkit" ]; then
                   extension_name="amazonwebservices.aws-toolkit-vscode"
+                  beta_version="@3.42.0"
```
justinmk3, line 32:
```suggestion
                  VERSION ="1.43.0"
                elif [ "${TARGET_EXTENSION}" = "toolkit" ]; then
                  extension_name="amazonwebservices.aws-toolkit-vscode"
                  VERSION ="3.42.0"
```

buildspec/release/70checkmarketplace.yml
```diff
@@ -22,21 +22,31 @@ phases:
...
                   exit 1
                 fi
+            # If in beta, install specific extension versions
+            - |
+                if [ "$STAGE" != "prod" ]; then
+                  VERSION=${beta_version#@}
+                  echo "checkmarketplace: Non-production stage detected. Installing beta version '${VERSION}'."
+                else
+                  beta_version=""
+                fi
```
justinmk3, line 44:
```suggestion
                if [ "$STAGE" != "prod" ]; then
                  echo "checkmarketplace: Non-production stage detected. Installing hardcoded version '${VERSION}'."
                fi
```

buildspec/release/70checkmarketplace.yml
```diff
@@ -22,21 +22,31 @@ phases:
...
             # keep reinstalling the extension until the desired version is updated. Otherwise fail on codebuild timeout (1 hour).
             - |
                 while true; do
                   code --uninstall-extension "${extension_name}" --no-sandbox --user-data-dir /tmp/vscode
-                  code --install-extension ${extension_name} --no-sandbox --user-data-dir /tmp/vscode
+                  code --install-extension "${extension_name}${beta_version}" --no-sandbox --user-data-dir /tmp/vscode
```
justinmk3, line 49:
```suggestion
                  code --install-extension "${extension_name}@${VERSION}" --no-sandbox --user-data-dir /tmp/vscode
```

Always passing the explicit version should work. 

Actually, that might simplify the "validation" a bit too ? Because it will fail if the version doesn't exist.

buildspec/release/70checkmarketplace.yml
```diff
@@ -22,27 +22,33 @@ phases:
...
+            - |
+                if [ "$STAGE" != "prod" ]; then
+                  echo "checkmarketplace: Non-production stage detected. Installing hardcoded version '${VERSION}'."
+                fi
+            # keep installing the desired extension version until successful. Otherwise fail on codebuild timeout (1 hour).
             - |
                 while true; do
                   code --uninstall-extension "${extension_name}" --no-sandbox --user-data-dir /tmp/vscode
-                  code --install-extension ${extension_name} --no-sandbox --user-data-dir /tmp/vscode
+                  code --install-extension "${extension_name}@${VERSION}" --no-sandbox --user-data-dir /tmp/vscode
```
justinmk3, line 45: `code` doesn't fail if the newly-published version isn't found yet? If it fails, we need `|| true` here. Otherwise the `while` loop will exit.

justinmk3:
```suggestion
                  code --install-extension "${extension_name}@${VERSION}" --no-sandbox --user-data-dir /tmp/vscode || true
```

justinmk3 (approved):

PR #6472, opened 2025-01-30
feat(amazonq): migrate /doc and /dev to zipStream

avi-alpert:
## Problem
Users see a `Sorry, I ran into an issue while trying to upload your code` error when using /dev or /doc if files in the repo have a modified or created time before the MSDOS epoch. This affects some users who Remote SSH into a Linux machine.

Logs show an out of range exception thrown from the adm-zip package. Adm-zip issue reported [here](https://github.com/cthackers/adm-zip/issues/485)

## Solution
Switch to the internal [zipStream module](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/utilities/zipStream.ts)

---


avi-alpert:
The duplicate code CI finding is on lines untouched by this PR (previously written unit tests)

packages/core/src/amazonqFeatureDev/util/files.ts
```diff
@@ -77,11 +74,12 @@ export async function prepareRepoData(
                 continue
             }
             totalBytes += fileSize
-
-            const zipFolderPath = path.dirname(file.zipFilePath)
+            // Paths in zip should be POSIX compliant regardless of OS
+            // Reference: https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
+            const posixPath = file.zipFilePath.replace(/\\/g, '/')
```
avi-alpert, line 79: from @michaellandi : Backslashes are valid in posix paths so this would change the location of a posix file with a backslash in it

chengoramazon (approved):

packages/core/src/amazonqFeatureDev/util/files.ts
avi-alpert:
```suggestion
            const posixPath = file.zipFilePath.split(path.sep).join(path.posix.sep)
```

michaellandi (approved):

packages/core/src/shared/index.ts
```diff
@@ -48,6 +48,7 @@ export * from './localizedText'
 export * as env from './vscode/env'
 export * from './vscode/commands2'
 export * from './utilities/pathUtils'
+export * from './utilities/zipStream'
```
justinmk3, line 51: should be exported as `* as zipStream`

packages/core/src/shared/utilities/zipStream.ts
```diff
@@ -150,4 +154,13 @@ export class ZipStream {
             streamBuffer: this._streamBuffer,
         }
     }
+
+    public static async unzip(zipBuffer: Buffer): Promise<ZipReaderResult[]> {
```
justinmk3, line 158: This isn't a streaming interface. So consumers of this function will be limited by available memory.

PR #6474, opened 2025-01-30
deps(auth): remove dependence on deprecated and outdated `@aws-sdk/*` packages.

Hweinstock:
## Problem
The auth code relies on old versions of `@aws-sdk/*` that have since been deprecated or are no longer backward compatible, making versions bumps impossible. 
- `@aws-sdk/credential-provider-imds` has since been [deprecated](https://www.npmjs.com/package/@aws-sdk/credential-provider-imds)
- `fromIni` from `@aws-sdk/credential-provider-ini` no longer supports passing a `loadedConfig`. 
- `AssumeRoleParams` is no longer exported by `@aws-sdk/credential-provider-ini`. 

We need to be able to bump these `@aws-sdk/*` package versions to continue to consume newer generated clients. Being pinned to older versions is also a security risk. See https://github.com/aws/aws-toolkit-vscode/pull/6439 for more information. 

## Solution
- write custom credentials provider to replace `fromIni` with `loadedConfig` option.
- drop dependency on `@aws-sdk/credential-provider-ini` since its no longer used. 
- add direct dependency on `@aws-sdk/credential-provider-env` since this was installed as part of `@aws-sdk-credential-provider-ini` before.   
- Fix many (not all) of the deprecation warnings in auth code related to credentials provider. 

### Custom Credentials Provider
Before, we used `fromIni` with the `loadedConfig` option which allows us to avoid reading the config file from disk on each credentials fetch and allows us to merge the current credentials with those found in the `.ini` file. To achieve the same behavior without the `loadedConfig` option, we need to write our own credentials provider that supports MFA and role assumption, and returns the desired merged credentials, rather than reading from disk. 

### Testing
- Manually verify this role assumption works by following the steps [here](https://docs.aws.amazon.com/sdkref/latest/guide/access-assume-role.html). 
- Manually verify MFA works via adapting [this](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-role.html#:~:text=This%20policy%20allows%20the%20user,they%20authenticate%20by%20using%20MFA.&text=Next%2C%20add%20a%20line%20to,by%20the%20role's%20trust%20policy.&text=The%20mfa_serial%20setting%20can%20take,command%20with%20this%20profile%20fails.&text=The%20second%20profile%20entry%2C%20role,%22:%20%5B%20%7B%20...). (Used DuoMobile)
- Add unit tests with API calls stubbed. 

## Future Work
- There are two tests that can now be re-enabled because of this version bump, undoing https://github.com/aws/aws-toolkit-vscode/commit/db27ebb578c882750e7372b59c37b212f471560c
- The steps to test role assumption could become an integ/e2e test. Right now requires setting many resources up in console, but perhaps this can all be done by the SDKs with an account on admin access. 

---


packages/core/src/auth/providers/sharedCredentialsProvider.ts
```diff
@@ -399,15 +387,51 @@ export class SharedCredentialsProvider implements CredentialsProvider {
...
+        return async () => {
+            const iniData = loadedCreds ?? profiles
+            const profile: CredentialsData = iniData[this.profileName]
+            if (!profile) {
+                throw new ToolkitError(`auth: Profile ${this.profileName} not found`)
+            }
+            // No role to assume, return static credentials.
+            if (!profile.role_arn) {
+                return {
+                    accessKeyId: profile.aws_access_key_id!,
```
jpinkney-aws, line 399: (I know very little about auth but had a couple questions). It looks like you're asserting that `profile.aws_access_key_id` is not undefined. Is that for sure? Or just to satisfy the accessKeyId type? I'm wonder if it would make sense to warn if access key id/secret access key aren't defined?

Hweinstock: My understanding is that these fields are required. Authenticating from a config file can be done by directly using the access key and secret key, or using it to assume an IAM role for authentication. However, either way the access key and secret key is required in the current profile or the source profile. 

To enforce this, we validate the config file here: https://github.com/aws/aws-toolkit-vscode/blob/c358a2a1f98c5596474390c9ba27544ab15d911e/packages/core/src/auth/providers/sharedCredentialsProvider.ts#L178-L201 and then only call this code(`makeSharedIniFileCredentialsProvider`) when we have the right conditions: https://github.com/aws/aws-toolkit-vscode/blob/c358a2a1f98c5596474390c9ba27544ab15d911e/packages/core/src/auth/providers/sharedCredentialsProvider.ts#L300-L350. 

Its a little hard to understand because this code lives in two places, but I don't believes its possible this function is called without the necessary keys existing.

packages/core/src/auth/providers/sharedCredentialsProvider.ts
```diff
@@ -399,15 +387,51 @@ export class SharedCredentialsProvider implements CredentialsProvider {
...
+                ? {
+                      RoleArn: profile.role_arn,
+                      RoleSessionName: 'AssumeRoleSession',
+                      SerialNumber: profile.mfa_serial,
+                      TokenCode: await getMfaTokenFromUser(profile.mfa_serial, this.profileName),
+                  }
+                : {
+                      RoleArn: profile.role_arn,
+                      RoleSessionName: 'AssumeRoleSession',
+                  }
```
jpinkney-aws, line 426: You could technically simplify this by doing something like:
```typescript
const assumeRoleReq = {
    RoleArn: profile.role_arn,
    RoleSessionName: 'AssumeRoleSession',
    ...(profile.mfa_serial ? {
        SerialNumber: profile.mfa_serial,
        TokenCode: await getMfaTokenFromUser(profile.mfa_serial, this.profileName)
    } : {})
}
```

jpinkney-aws (approved):

jpinkney-aws (approved):

justinmk3:
> * Manually verify this role assumption works by following the steps [here](https://docs.aws.amazon.com/sdkref/latest/guide/access-assume-role.html).
> * Manually verify MFA works via adapting [this](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-role.html#:~:text=This%20policy%20allows%20the%20user,they%20authenticate%20by%20using%20MFA.&text=Next%2C%20add%20a%20line%20to,by%20the%20role's%20trust%20policy.&text=The%20mfa_serial%20setting%20can%20take,command%20with%20this%20profile%20fails.&text=The%20second%20profile%20entry%2C%20role,%22:%20%5B%20%7B%20...). (Used DuoMobile)

Should we link to those somehow in https://github.com/aws/aws-toolkit-vscode/blob/master/docs/faq-credentials.md ? Gathering relevant doc for credentials is difficult, and it looks like you found some good references.

justinmk3 (approved):
🚀

packages/core/src/auth/providers/sharedCredentialsProvider.ts
Hweinstock: Clever, I like it.

Hweinstock:
Added some documentation with the links I used above. Also, do not merge this yet, need to confirm with sagemaker this doesn't affect them.

justinmk3:
Seems to have caused a regression: https://github.com/aws/aws-toolkit-vscode/issues/6782

PR #6479, opened 2025-01-31
fix(auth): token refresh rapidly called unexpectedly

nkomonen-amazon:
## Problem:

getChatAuthState() is called in many places by the Q features simultaneously,
this eventually triggers multiple calls to getToken() and if needed refreshToken().

This resulted in refreshToken being spammed and the Identity team seeing spikes in token refreshes
from clients.

## Solution:

Throttle getChatAuthState().

Throttling w/ leading: true, allows us to instantly return
a fresh result OR a cached result in the case we are throttled. Debounce on the
other hand would cause callers to hang since they have to wait for debounce to timeout.

Also, we put a debounce on getToken() before in #6282 but this did not work since a new
SsoAccessToken instance is created each time the offending code flow triggered (we could
look to cache the instance instead which would enable the getToken() debounce to be useful.

### Testing

To test the difference after adding the throttle:
- Add log statements to `getToken()`
- Set an expired date in the SSO cache for both token expiration + client registration expiration
- Use chat

What would happen is that without throttle it would trigger getChatAuthState() many times, likely due to the connection
becoming invalid and sending an event to all Q features, causing each of them to call getChatAuthState() at the same time.

But when the throttle was added, the amount of these calls dropped to at most 2.

Signed-off-by: nkomonen-amazon <nkomonen@amazon.com>

---

### DIFF
--- a/packages/core/src/codewhisperer/util/authUtil.ts
+++ b/packages/core/src/codewhisperer/util/authUtil.ts
@@ -44,6 +44,7 @@ import { telemetry } from '../../shared/telemetry/telemetry'
 import { asStringifiedStack } from '../../shared/telemetry/spans'
 import { withTelemetryContext } from '../../shared/telemetry/util'
 import { focusAmazonQPanel } from '../../codewhispererChat/commands/registerCommands'
+import { throttle } from 'lodash'
 
 /** Backwards compatibility for connections w pre-chat scopes */
 export const codeWhispererCoreScopes = [...scopesCodeWhispererCore]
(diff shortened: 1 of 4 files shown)


jpinkney-aws (approved):

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -44,6 +44,7 @@ import { telemetry } from '../../shared/telemetry/telemetry'
 import { asStringifiedStack } from '../../shared/telemetry/spans'
 import { withTelemetryContext } from '../../shared/telemetry/util'
 import { focusAmazonQPanel } from '../../codewhispererChat/commands/registerCommands'
+import { throttle } from 'lodash'
```
Hweinstock, line 47: I believe we have a backlog item to remove dependency on `lodash` for performance reasons. The work was started here: https://github.com/aws/aws-toolkit-vscode/pull/5157

nkomonen-amazon: After some discussions w/ our move to the LSP work we will probably be able to drop this, and there is currently no existing throttle implementation. We will either need to build our own or add a new dependency. Ticket for removing lodash is here: https://taskei.amazon.dev/tasks/IDE-10293

jpinkney-aws: FWIW it looks like throttle just uses debounce under the hood: https://github.com/lodash/lodash/blob/npm/throttle.js

PR #6486, opened 2025-02-03
test(amazonq): Add inline completion e2e tests

jpinkney-aws:
## Problem
- we have no confidence checks on inline completion e2e flows
- we want to use these tests to eventually verify that codewhisperer language server is behaving correctly

## Solution
- add e2e tests that focus on sending inputs to the text editor and seeing how it reacts. That way we can re-use these tests with the codewhisperer language server

---


jpinkney-aws:
It flaked 1 time over 30 executions (~10 requests + minimum/stable/insiders). I haven't been able to reproduce it again though

packages/amazonq/test/e2e/inline/inline.test.ts
```diff
@@ -0,0 +1,179 @@
+/*!
```
justinmk3, line 1: this file is located at a new place, whereas all the other Q features are located in `packages/amazonq/test/e2e/amazonq`

maybe the `e2e/amazonq/*` files should be lifted up into this location (followup PR)? since the `e2e/amazonq` subdir seems redundant.

packages/amazonq/test/e2e/inline/inline.test.ts
```diff
@@ -0,0 +1,179 @@
...
+        const ok = await waitUntil(async () => RecommendationHandler.instance.isSuggestionVisible(), waitOptions)
+        if (!ok) {
+            assert.fail('Suggestions failed to become visible')
+        }
+    }
+
+    async function waitForTelemetry() {
+        const ok = await waitUntil(
+            async () =>
+                globals.telemetry.logger.query({
```
justinmk3, line 71: Doesn't `assertTelemetry()` do this, or what is it missing?

justinmk3 (approved):

packages/amazonq/test/e2e/inline/inline.test.ts
jpinkney-aws: assertTelemetry does the querying but it doesn't wait for the event to appear as far as I can tell. I want the tests to explicitly wait until those telemetry events appear and then finally assert on them

PR #6498, opened 2025-02-04
feat(lsp): status bar message when downloading

tomcat323:
## Problem
https://taskei.amazon.dev/tasks/IDE-15760
Users want to know that language servers are downloading

## Solution
Add a status bar message when LSP is downloading
Also added a developer command to reset LSP download for testing

https://github.com/user-attachments/assets/a00430a2-9c53-4406-aabe-c13217bc89b5

---

### DIFF
--- a/packages/core/src/amazonq/lsp/lspController.ts
+++ b/packages/core/src/amazonq/lsp/lspController.ts
@@ -22,6 +22,8 @@ import { fs, globals, ToolkitError } from '../../shared'
 import { isWeb } from '../../shared/extensionGlobals'
 import { getUserAgent } from '../../shared/telemetry/util'
 import { isAmazonInternalOs } from '../../shared/vscode/env'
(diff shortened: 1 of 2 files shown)


packages/core/src/amazonq/lsp/lspController.ts
```diff
@@ -219,13 +221,15 @@ export class LspController {
         return true
     }
 
-    async tryInstallLsp(context: vscode.ExtensionContext): Promise<boolean> {
+    async tryInstallLsp(context: vscode.ExtensionContext, forceReset = false): Promise<boolean> {
```
jpinkney-aws, line 224: Can we make these changes on the `feature/amazonqLSP` branch instead? I think the resolve() function in `core/src/shared/lsp/lspResolver.ts` could have this instead. That way you could log the language server name in addition to that its downloading

PR #6499, opened 2025-02-04
telemetry(auth): debounce `aws_loadCredentials` and add debugging information

Hweinstock:
## Problem
This metric is currently being spammed in telemetry at unusual volume. It is suspected that this is connected to this change: https://github.com/aws/aws-toolkit-vscode/pull/5979/files, but the connection is not yet clear. 

## Solution
- use the `withTelemetryContext` decorator to emit a `function_call` metric to capture additional information. This metric should allow us to determine what is calling this function so much. 
- debounce the current emit line that is causing the high volume emission. 

## Future Work 
- `debounce` reimplements a special case of `debounceWithCancel`, these can be redone to reduce code and improve reliability. 

---

### DIFF
--- a/packages/core/src/auth/providers/credentialsProviderManager.ts
+++ b/packages/core/src/auth/providers/credentialsProviderManager.ts
@@ -5,6 +5,7 @@
 
 import { getLogger } from '../../shared/logger'
 import { telemetry } from '../../shared/telemetry/telemetry'
+import { withTelemetryContext } from '../../shared/telemetry/util'
 import {
     asString,
     CredentialsProvider,
@@ -14,6 +15,7 @@ import {
 } from './credentials'
 import { CredentialsProviderFactory } from './credentialsProviderFactory'
 
+const credentialsProviderManager = 'CredentialsProviderManager'
 /**
  * Responsible for providing the Toolkit with all available CredentialsProviders.
  * Providers may be registered directly or created with supplied CredentialsProviderFactories.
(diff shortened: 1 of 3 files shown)


jpinkney-aws (approved):

jpinkney-aws:
Is everyone spam calling this one? Can we create a build and run it locally for a day to see if we can fix it before the next release?

Hweinstock:
From the telemetry, it seems that there are many users spamming this one, but not everyone. Messed around with the toolkit for a while and wasn't able to determine source or notice any clear correlations with other metrics in telemetry.  It does emit 4x on startup, and multiple times when the source of the credentials are changed. However, I can keep digging. Its possible adding the decorator in more places could be valuable.

packages/core/src/auth/providers/credentialsProviderManager.ts
```diff
@@ -43,8 +51,14 @@ export class CredentialsProviderManager {
             if (!providerType) {
                 continue
             }
-            const telemType = credentialsProviderToTelemetryType(providerType)
-            telemetry.aws_loadCredentials.emit({ credentialSourceId: telemType, value: refreshed.length })
+            const emitWithDebounce = cancellableDebounce(
```
nkomonen-amazon, line 54: I don't think this will work as you intend. I think you need to move this cancellableDebounce() call to a class parameter since this creates a new instance each time

packages/core/src/auth/providers/credentialsProviderManager.ts
```diff
@@ -23,13 +26,18 @@ export class CredentialsProviderManager {
     private readonly providerFactories: CredentialsProviderFactory[] = []
     private readonly providers: CredentialsProvider[] = []
 
+    @withTelemetryContext({ name: 'getAllCredentialsProvider', class: credentialsProviderManagerClassName, emit: true })
```
nkomonen-amazon, line 29: And this emits every call, possible chance this will be noisy. But it should hopefully not break anything before we remove it

jpinkney-aws:
Despite code freeze I think we should merge this one as well. Otherwise we will need to wait another week to further debug this problem

PR #6501, opened 2025-02-04
feat(stepfunctions): Add a setting to opt out of WFS + minor changes

witness-me:
## Problem
1. The feature opens specific file types in a custom editor by default and customers don't have a way to optout while still using toolkit
2. Currently the indentation setting used in the integration defaults to 4 spaces. If it's different in user's VSCode setting, the spacing for the string sent from the integration will not match
3. Currently context menu lacks the command to launch WFS. It [has been added before](https://github.com/aws/aws-toolkit-vscode/pull/5834/files), but it looks it was accidentally added to auto-generated file and was thus removed later

## Solution
1. Adding a setting to opt out of WFS, which will close custom editor and open default editor instead. Note that I did not find a way to completely prevent custom editor panel mount - it looks it's not possible, so it will start mounting first and then (in `resolveCustomTextEditor`) it needs to be witched to default editor. Please advise if that's not the case and the custom editor mount can be prevented earlier (while keeping file format association with the file types)
2. Passing a user setting for tab spacing to be used on the webview side to format JSON/YAML with the right indentation
3. Adding an option to launch WFS from context menu, as it was added before (but this time in the right package.json)
---

### DIFF
--- a/packages/core/package.nls.json
+++ b/packages/core/package.nls.json
@@ -27,6 +27,7 @@
     "AWS.stepFunctions.workflowStudio.actions.progressMessage": "Opening asl file in Workflow Studio",
     "AWS.stepFunctions.workflowStudio.actions.saveSuccessMessage": "{0} has been saved",
     "AWS.stepFunctions.workflowStudio.actions.invalidJson": "The Workflow Studio editor was not opened because the JSON in the file is invalid. To access Workflow Studio, please fix the JSON and manually reopen the integration.",
+    "AWS.stepFunctions.workflowStudio.enable.desc": "Open Amazon States Language files in Workflow Studio by default.",
     "AWS.configuration.description.awssam.debug.api": "API Gateway configuration",
     "AWS.configuration.description.awssam.debug.api.clientCertId": "The API Gateway client certificate ID",
     "AWS.configuration.description.awssam.debug.api.headers": "Additional HTTP headers",
--- a/packages/toolkit/package.json
+++ b/packages/toolkit/package.json
@@ -146,6 +146,11 @@
                     "default": true,
                     "description": "%AWS.stepFunctions.asl.format.enable.desc%"
                 },
+                "aws.stepfunctions.workflowStudio.enable": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "%AWS.stepFunctions.workflowStudio.enable.desc%"
+                },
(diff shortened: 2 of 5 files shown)


jpinkney-aws (approved):

packages/toolkit/package.json
```diff
@@ -146,6 +146,11 @@
                     "default": true,
                     "description": "%AWS.stepFunctions.asl.format.enable.desc%"
                 },
+                "aws.stepfunctions.workflowStudio.enable": {
```
justinmk3, line 149: Why is this setting being added?

anthonyting (approved):

justinmk3 (changes requested):
Why is a setting being added? Not explained in the PR description. Looks like a low-value setting. Settings should not be added unless there is a strong use-case.

packages/core/package.nls.json
```diff
@@ -27,6 +27,7 @@
     "AWS.stepFunctions.workflowStudio.actions.progressMessage": "Opening asl file in Workflow Studio",
     "AWS.stepFunctions.workflowStudio.actions.saveSuccessMessage": "{0} has been saved",
     "AWS.stepFunctions.workflowStudio.actions.invalidJson": "The Workflow Studio editor was not opened because the JSON in the file is invalid. To access Workflow Studio, please fix the JSON and manually reopen the integration.",
+    "AWS.stepFunctions.workflowStudio.enable.desc": "Open Amazon States Language files in Workflow Studio by default.",
```
justinmk3, line 30: Use `workbench.editorAssociations`. We already do something similar for threat composer: https://github.com/aws/aws-toolkit-vscode/blob/30face99ec161876378b49e71e78ae3714ed5797/packages/toolkit/package.json#L4412-L4415

Do not add new, custom ways of setting file associations.

packages/toolkit/package.json
witness-me: This is explained in point 1:
> The feature opens specific file types in a custom editor by default and customers don't have a way to optout while still using toolkit

Our team (including the UX) decided that since custom editor is opened by default, we want to provide a way for the user to opt out of it

witness-me:
After discussing offline, we agreed to not add that as a setting, as this is something that can be done with local config file. I'll cancel this PR since that change was its main content, and will publish other small changes in a separate PR

PR #6518, opened 2025-02-06
ci(lint): rule to detect index.ts imports in core

hayemaxi:
I could not find a lint rule (or a configuration of a lint rule) to achieve this, so I just made one. I tried some things from eslint-plugin-import (namely no-cycles), but I was not happy with the results.

This will prevent you from importing from index.ts files while working in packages/core/, which has been a source of annoying and hard to understand circular dependency issues. These files should be reserved for exporting to the subprojects only.

It only affects imports in core, and only in src/ directories that do not start with "test"

---


plugins/eslint-plugin-aws-toolkits/lib/rules/no-index-import.ts
```diff
@@ -0,0 +1,95 @@
...
+            default: errMsg,
+        },
+        type: 'problem',
+        fixable: 'code',
+        schema: [],
+    },
+    defaultOptions: [],
+    create(context) {
+        const filePath = context.physicalFilename
+        if (!filePath.match(/packages\/core\/src\/(?!test)/)) {
```
Hweinstock, line 48: Is it possible to configure this at the top level in the `.eslintrc.js` file?

plugins/eslint-plugin-aws-toolkits/lib/rules/no-index-import.ts
```diff
@@ -0,0 +1,95 @@
...
+                    .resolve(path.dirname(filePath), relativeImportPath)
+                    .replace(/(\.d)?\.ts/, '') // remove any .d.ts or .ts file extensions on the import (unlikely)
+
+                if (path.basename(absoluteImportPath) !== 'index') {
+                    // Not an index.ts file but is some typescript file.
+                    if (['.ts', '.d.ts'].some((e) => fs.existsSync(absoluteImportPath + e))) {
+                        return
+                    }
+
+                    // If if does not exist as a folder, then the path is simply wrong. Another, more descriptive error will surface instead.
```
Hweinstock, line 76:
```suggestion
                    // If it does not exist as a folder, then the path is simply wrong. Another, more descriptive error will surface instead.
```

plugins/eslint-plugin-aws-toolkits/lib/rules/no-index-import.ts
```diff
@@ -0,0 +1,95 @@
...
+                        return
+                    }
+                }
+
+                return context.report({
+                    node: node.source,
+                    messageId: 'default',
+                    // TODO: We can add a fixer to resolve the import for us.
+                    // fix: (fixer) => {
+                    //     // somehow parse the exports of the imported index.ts file to get the actual export path,
```
Hweinstock, line 87: I'm guessing this isn't straightforward? Does ESLint allow you to access this information or would it involve manually reading the file?

Hweinstock (approved):

plugins/eslint-plugin-aws-toolkits/lib/rules/no-index-import.ts
hayemaxi: Not that I have found. If you are referring to `no-restricted-imports` specifically, no because it only matches on the import string literal.

plugins/eslint-plugin-aws-toolkits/lib/rules/no-index-import.ts
hayemaxi: Eslint context doesn't seem to offer more than the file it is currently linting. The only thing I can think of is opening the file manually and load it into the `ts` lib to create an AST.

jpinkney-aws (approved):

PR #6519, opened 2025-02-06
feat(amazonq): auto-download results

dhasani23:
## Problem

Transformation results/artifacts expire after 24h, and sometimes users forget to manually click the "Download Proposed Changes" button within that time period.

## Solution

Auto-download the results from S3 when ready.

---

### DIFF
--- a/packages/core/src/codewhisperer/commands/startTransformByQ.ts
+++ b/packages/core/src/codewhisperer/commands/startTransformByQ.ts
@@ -773,6 +773,11 @@ export async function postTransformationJob() {
     if (transformByQState.getPayloadFilePath() !== '') {
         fs.rmSync(transformByQState.getPayloadFilePath(), { recursive: true, force: true }) // delete ZIP if it exists
     }
+
+    // attempt download for user
+    if (transformByQState.isSucceeded() || transformByQState.isPartiallySucceeded()) {
+        await vscode.commands.executeCommand('aws.amazonq.transformationHub.reviewChanges.startReview')
+    }
 }
 
 export async function transformationJobErrorHandler(error: any) {
--- a/packages/core/src/codewhisperer/commands/startTransformByQ.ts
+++ b/packages/core/src/codewhisperer/commands/startTransformByQ.ts
@@ -775,6 +775,7 @@ export async function postTransformationJob() {
     }
 
     // attempt download for user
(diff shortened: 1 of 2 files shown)


ntarakad-aws (approved):

packages/core/src/codewhisperer/commands/startTransformByQ.ts
```diff
@@ -773,6 +773,11 @@ export async function postTransformationJob() {
     if (transformByQState.getPayloadFilePath() !== '') {
         fs.rmSync(transformByQState.getPayloadFilePath(), { recursive: true, force: true }) // delete ZIP if it exists
     }
+
+    // attempt download for user
+    if (transformByQState.isSucceeded() || transformByQState.isPartiallySucceeded()) {
+        await vscode.commands.executeCommand('aws.amazonq.transformationHub.reviewChanges.startReview')
```
jpinkney-aws, line 779: Just as a random aside, ideally this would have used our internal Commands wrapper `core/src/shared/vscode/commands2.ts`. This wrapper basically allows you to define the command as a const that you can execute later which is much more robust then having to pass around command ids!

In order to do that you would need to register `aws.amazonq.transformationHub.reviewChanges.startReview` like:

```
export const startReview = Commands.declare('aws.amazonq.transformationHub.reviewChanges.startReview',
    () => async () => {
       // implement me here
    }
)
```

and then here you could just do `startReview.tryExecute()` that way you don't have to care about underlying id's!

See: https://github.com/aws/aws-toolkit-vscode/blob/master/docs/arch_develop.md#commands for more info

dhasani23: Noted, https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/service/transformByQ/transformationResultsViewProvider.ts is full of registered commands that could be refactored this way, so I think it's better if I do all of that in a separate PR.

jpinkney-aws: I'm okay with merging this as is as long as you commit/agree to making that change 😄. It should be like 15 minutes worth of work tops

dhasani23: Added a TODO and made a note of it so I don't forget

PR #6526, opened 2025-02-07
refactor(logging): scope log messages to specific LSP using topic headers. 

Hweinstock:
## Problem
The log messages coming out of the LSP work do not make it clear which LSP its working with (Q Workspace or Q general). 
Follow up from: https://github.com/aws/aws-toolkit-vscode/pull/6385#discussion_r1939666970
## Solution
- Use new `LogTopic` functionality to scope these to specific LSP. 
- Add more logging information from LSP cleanup process. 

---


packages/core/src/shared/logger/logger.ts
```diff
@@ -5,7 +5,16 @@
...
 
-export type LogTopic = 'crashMonitoring' | 'dev/beta' | 'notifications' | 'test' | 'childProcess' | 'lsp' | 'unknown'
+export type LogTopic =
+    | 'crashMonitoring'
+    | 'dev/beta'
+    | 'notifications'
+    | 'test'
+    | 'childProcess'
+    | 'lsp'
+    | 'QWorkspaceLsp'
```
jpinkney-aws, line 15: should these be `amazonqWorkspaceLSP` or is that too long? I feel like we need to have conventions on whether we use q vs amazonq since both are used in the codebase 🤔

packages/core/src/amazonq/lsp/workspaceInstaller.ts
```diff
@@ -31,7 +32,11 @@ export class WorkspaceLSPResolver implements LspResolver {
         const nodePath = path.join(installationResult.assetDirectory, nodeName)
         await fs.chmod(nodePath, 0o755)
 
-        await cleanLspDownloads(manifest.versions, path.basename(installationResult.assetDirectory))
+        const deletedVersions = await cleanLspDownloads(
+            manifest.versions,
+            path.basename(installationResult.assetDirectory)
+        )
+        logger.debug(`cleaning old LSP versions deleted ${deletedVersions.length} versions`)
```
jpinkney-aws, line 39: nitpick
```suggestion
        logger.debug(`Cleaned up ${deletedVersions.length} old versions`)
```
what are your thoughts?

packages/core/src/shared/logger/logger.ts
Hweinstock: That feels more consistent. Better to start verbose and look for opportunities to shorten later.

packages/amazonq/src/lsp/lspInstaller.ts
```diff
@@ -13,16 +13,18 @@ import {
...
     getNodeExecutableName,
     cleanLspDownloads,
+    getLogger,
 } from 'aws-core-vscode/shared'
 import path from 'path'
 
 const manifestURL = 'https://aws-toolkit-language-servers.amazonaws.com/codewhisperer/0/manifest.json'
 export const supportedLspServerVersions = '^2.3.0'
-
+const logger = getLogger('amazonqLSP')
```
justinmk3, line 22:
```suggestion
const logger = getLogger('amazonqLsp')
```

packages/core/src/amazonq/lsp/workspaceInstaller.ts
```diff
@@ -10,11 +10,12 @@ import { LanguageServerResolver } from '../../shared/lsp/lspResolver'
 import { Range } from 'semver'
 import { getNodeExecutableName } from '../../shared/lsp/utils/platform'
 import { fs } from '../../shared/fs/fs'
-import { cleanLspDownloads } from '../../shared'
+import { cleanLspDownloads, getLogger } from '../../shared'
 
 const manifestUrl = 'https://aws-toolkit-language-servers.amazonaws.com/q-context/manifest.json'
 // this LSP client in Q extension is only going to work with these LSP server versions
 const supportedLspServerVersions = '0.1.35'
+const logger = getLogger('amazonqWorkspaceLSP')
```
justinmk3, line 18: `amazonqWorkspaceIndexingLsp` is too long, but probably "indexing" is the most important thing to call out?

```suggestion
const logger = getLogger('amazonqIndexingLsp')
```

justinmk3 (approved):
Great idea

packages/core/src/amazonq/lsp/workspaceInstaller.ts
Hweinstock: Right now, indexing is the only action supported by this lsp, but I could imagine Q adding a feature that does other "work" within the workspace. Also, I think it might be best to keep it inline with the `name` field we use in the resolver here: https://github.com/aws/aws-toolkit-vscode/blob/2157f55a3c8bc428a572c44c3f576a19e1eaa49a/packages/core/src/amazonq/lsp/workspaceInstaller.ts#L21

Perhaps we should pull these out somewhere to make them more discoverable, but that might involve a deeper refactor out of the scope here.

jpinkney-aws (approved):

PR #6527, opened 2025-02-07
fix(amazonq): right click options dont send context if chat not loaded

nkomonen-amazon:
## Problem:

When a user starts up VSC, hasn't opened Q chat yet, then highlights text and does a right click option like "send to prompt" THEN the whole prompt fails to make it to chat.

This is because there is a race condition between the message being sent to the webview, and the webview completed loading and ready to recieve messages.

## Solution:

In the appToMessage publisher, verify that the chat is ready to recieve messages based on if the webview sent the chat ui ready message. Otherwise wait until we get it, and then send the message.

#### Before

https://github.com/user-attachments/assets/baf86737-4a5a-4f80-86fb-9abca2c56827

#### After 

https://github.com/user-attachments/assets/c3361403-5dae-49df-a878-94722eb4fc62

---

### DIFF
--- a/packages/amazonq/.changes/next-release/Bug Fix-704d175a-a99f-4ce7-bf08-b94d6f7218c0.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-704d175a-a99f-4ce7-bf08-b94d6f7218c0.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "`Send to prompt` and other context menu options not sent if chat was closed"
+}
--- a/packages/core/src/amazonq/messages/messagePublisher.ts
+++ b/packages/core/src/amazonq/messages/messagePublisher.ts
@@ -12,3 +12,48 @@ export class MessagePublisher<T> {
         this.eventEmitter.fire(event)
     }
 }
+
+/**
+ * Same as {@link MessagePublisher}, but will wait until the UI indicates it
+ * is ready to recieve messages, before the message is published.
+ *
+ * This solves a problem when running a right click menu option like
+ * "Send To Prompt" BUT chat is not opened yet, it would result in the prompt failing to
+ * be recieved by chat.
+ */
+export class UiMessagePublisher<T> extends MessagePublisher<T> {
+    private isUiReady: boolean = false
+    private buffer: T[] = []
+
+    constructor(eventEmitter: EventEmitter<T>) {
+        super(eventEmitter)
+    }
+
+    public override publish(event: T): void {
+        // immediately send if Chat UI is ready
+        if (this.isUiReady) {
+            super.publish(event)
+            return
+        }
+
+        this.buffer.push(event)
+    }
+
+    /**
+     * Indicate the Q Chat UI is ready to recieve messages.
+     */
+    public setUiReady() {
+        this.isUiReady = true
+        this.flush()
+    }
+
+    /**
+     * Publishes all blocked messages
+     */
+    private flush() {
+        for (const msg of this.buffer) {
+            super.publish(msg)
+        }
+        this.buffer = []
+    }
+}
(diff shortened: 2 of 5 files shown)


packages/core/src/amazonq/messages/messagePublisher.ts
```diff
@@ -12,3 +13,45 @@ export class MessagePublisher<T> {
...
+
+/**
+ * Same as {@link MessagePublisher}, but will wait until the UI indicates it
+ * is ready to recieve messages, before the message is published.
+ *
+ * This solves a problem when running a right click menu option like
+ * "Send To Prompt" BUT chat is not opened yet, it would result in the prompt failing to
+ * be recieved by chat.
+ */
+export class UiReadyMessagePublisher<T> extends MessagePublisher<T> {
```
jpinkney-aws, line 25: I'd probably just make this UiMessagePublisher

packages/core/src/amazonq/messages/messagePublisher.ts
```diff
@@ -12,3 +13,45 @@ export class MessagePublisher<T> {
...
+    constructor(
+        eventEmitter: EventEmitter<T>,
+        private readonly isChatUiReady: () => boolean
+    ) {
+        super(eventEmitter)
+    }
+    public override publish(event: T): void {
```
jpinkney-aws, line 32: Does it make sense to encapsulate the logic for isUiReady in this class somehow? Technically, it seems like this class is meant to hold requests until the UI is ready so keeping the uiReady logic close might make more sense

packages/core/src/amazonq/messages/messagePublisher.ts
```diff
@@ -12,3 +13,45 @@ export class MessagePublisher<T> {
...
+        // immediately send if Chat UI is ready
+        if (this.isChatUiReady()) {
+            super.publish(event)
+            return
+        }
+        // Otherwise wait, since it is probably still loading
+        void waitUntil(
```
jpinkney-aws, line 39: Did you experiment with trying to use a buffer that keeps filling up until the chat ui is ready? Then you can just send an event when the chat UI is ready that publishes everthing in the buffer.

I.e

```typescript
if (this.isChatUIReady()) {
    super.publish(event)
    return
}
buffer.push(event)
```
and then you'd do something like this
```typescript
onChatReady(() => {
    for (const event of this.buffer) {
        super.publish(event)
    }
})
```

jpinkney-aws (commented):
This PR is more or less what I had in mind to solve the problem!

packages/core/src/amazonq/messages/messagePublisher.ts
nkomonen-amazon: great point, updated

packages/core/src/amazonq/messages/messagePublisher.ts
nkomonen-amazon: Yeah this makes sense, updated

nkomonen-amazon:
1. Yeah I was thinking about directly adding it to `MessagePublisher`, but since MessagePublisher is used in the opposite direction from WebviewToApp, it wasn't completely obvious. Though we can still add the new logic in but just know which scenario to use it. Any thoughts?
2. Thats a good point. I think we can probably call `focusAmazonQPanel.execute()`. Currently it is redundant since all the relevant calls in the fix call it before. But it will be a good fallback in future cases.

packages/amazonq/.changes/next-release/Bug Fix-704d175a-a99f-4ce7-bf08-b94d6f7218c0.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Right-click context menu not sent if chat was closed"
```
justinmk3, line 3: Maybe we can make this more intuitive

```suggestion
	"description": "`Send to prompt` context menu not sent if Amazon Q chat was not already active"
```

justinmk3 (approved):

PR #6528, opened 2025-02-07
deps(amazonq): Update amazon q language server to 3.x.x

jpinkney-aws:
## Problem
Amazon q currently uses the 2.3.2 version of the language server which doesn't appear to be working

## Solution
Update to the 3.x.x series for now

Independently i'm going to follow up with teams to understand how this happened. Had we released this feature to prod amazon q inline would have been completely broken

---


nkomonen-amazon (approved):

Hweinstock:
When you say not resolving, which step was failing? Also, has this happened before? Out of scope here, but wondering if its worth investing in a solution that will gracefully handle these errors for us to avoid prod failures. Maybe something simple like bumping the version down temporarily when it can't resolve.

PR #6545, opened 2025-02-10
config(customization): customization override

Will-ShaoHua:
## Problem

https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/util/customizationUtil.ts#L116-L130

The flaw of previous implementation won't override customization if users have selected one. 

The team is asking 
1. Ability to override regardless users are with a customization or not
2. Should respect users' selection afterward when users switch back to the customization they prefer
3. When ab service provides a new override customization arn, plugins should be able to do override again

```
// as long as there is a customization selected, override will not happen no matter what
 if (selectedCustomization && selectedCustomization.name !== '') {
        return selectedCustomization
    } else {
        const customizationFeature = FeatureConfigProvider.getFeature(Features.customizationArnOverride)
        const arnOverride = customizationFeature?.value.stringValue
        const customizationOverrideName = customizationFeature?.variation
        if (arnOverride === undefined) {
            return baseCustomization
        } else {
            return {
                arn: arnOverride,
                name: customizationOverrideName,
                description: baseCustomization.description,
            }
        }

```

## Solution
1. Persist a value of "previous override arn" and do override when the provided customization arn is different than "previous override arn"

---


packages/core/src/test/amazonq/customizationUtil.test.ts
```diff
@@ -82,16 +82,39 @@ describe('CodeWhisperer-customizationUtils', function () {
         assert.strictEqual(actualCustomization.name, selectedCustomization.name)
     })
 
-    it('Returns AB customization', async function () {
-        sinon.stub(AuthUtil.instance, 'isValidEnterpriseSsoInUse').returns(true)
+    it(`setSelectedCsutomization should set correctly if override is false or not specified`, async function () {
```
jpinkney-aws, line 85: nit: 
```suggestion
    it(`setSelectedCustomization should set correctly if override is false or not specified`, async function () {
```

packages/core/src/test/amazonq/customizationUtil.test.ts
```diff
@@ -82,16 +82,39 @@ describe('CodeWhisperer-customizationUtils', function () {
...
+        assert.strictEqual(getSelectedCustomization().arn, 'BAR')
+
+        await setSelectedCustomization({ arn: 'BAZ' })
+        assert.strictEqual(getSelectedCustomization().arn, 'BAZ')
+
+        await setSelectedCustomization({ arn: 'QOO' }, false)
+        assert.strictEqual(getSelectedCustomization().arn, 'QOO')
+    })
+
+    it(`setSelectedCsutomization should only do once for override per customization arn`, async function () {
```
jpinkney-aws, line 99: nit:
```suggestion
    it(`setSelectedCustomization should only do once for override per customization arn`, async function () {
```

packages/core/src/codewhisperer/util/customizationUtil.ts
```diff
@@ -116,25 +112,27 @@ export const getSelectedCustomization = (): Customization => {
...
-            }
-        }
+        return baseCustomization
     }
 }
 
-export const setSelectedCustomization = async (customization: Customization) => {
+/**
+ * @param customization customization to select
+ * @param isOverride if the API call is made from us (Q) but not users' intent, set isOverride to TRUE
```
jpinkney-aws, line 121: > API call is made from us (Q)

Do you mean from the extension side?

packages/core/src/test/amazonq/customizationUtil.test.ts
jpinkney-aws: What is this test case trying to illustrate? It's not really clear from the title

packages/core/src/codewhisperer/util/customizationUtil.ts
```diff
@@ -116,25 +112,27 @@ export const getSelectedCustomization = (): Customization => {
...
+ * Override happens when ALL following conditions are met
+ *  1. service returns non-empty override customization arn, refer to [featureConfig.ts]
+ *  2. the override customization arn is different from the previous override customization if any. The purpose is to only do override once on users' behalf.
+ */
+export const setSelectedCustomization = async (customization: Customization, isOverride: boolean = false) => {
     if (!AuthUtil.instance.isValidEnterpriseSsoInUse() || !AuthUtil.instance.conn) {
         return
     }
+    if (isOverride) {
```
jpinkney-aws, line 130: Just for my own understanding -- this means that the server has received a non empty override customization arn and we want to check whether or not its a new arn. If the same arn is found then we're good to go since there's nothing to do. Otherwise override the customization?

Is that correct?

jpinkney-aws (commented):
Not really familiar with this feature but wanted to leave a few comments out of curiosity

packages/core/src/test/amazonq/customizationUtil.test.ts
Will-ShaoHua: so when override is true, it's saying the setCustomization is invoked by us and not by users. In such case, the customization we pass in the call `setCustomization` is from A/B service, and we will override users' own selection once for each customization arn provided by the service.

Will-ShaoHua: when override is false, it means the setCustomization is invoked by users, so we will respect to users' selection anyway

packages/core/src/codewhisperer/util/customizationUtil.ts
Will-ShaoHua: yea, i want to express it's us to override users' selection and it's not on their own intention to change such configuration

packages/core/src/codewhisperer/util/customizationUtil.ts
Will-ShaoHua: > his means that the server has received a non empty override customization arn and we want to check whether or not its a new arn.

this is correct.


> If the same arn is found then we're good to go since there's nothing to do. Otherwise override the customization?

yes and no. If the customization arn is same as the last one (which we had done override once), we do nothing because it's intrusive if we change users' selection every time when there is a non-null override customization arn.

leigaol (approved):
Please polish your description in the PR. Thanks!

jpinkney-aws (approved):

justinmk3:
Should we not mention this in the changelog?

Will-ShaoHua:
> Should we not mention this in the changelog?

@justinmk3 probably not, as it's targeting a subset of users and experiment IMO

PR #6419, opened 2025-01-23
feat(amazonq): pre-fetch next recommendations for inline completions

zixlin7:
## Problem
To support complex recommendations broken into multiple chunks, we need the ability to immediately show user next recommendation when they accept previous one

## Solution
pre-fetch next recommendation predicting user accept first recommendation when showing them to user, and if user did accept first recommendation, immediately show next recommendation

---

- Treat all work as PUBLIC. Private `feature/x` branches will not be squash-merged at release time.
- Your code changes must meet the guidelines in [CONTRIBUTING.md](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#guidelines).
- License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6547, opened 2025-02-10
config(inline): show next session if only both sessino ids are the same

Will-ShaoHua:
## Problem

Followup of #6419 

Team decides to only show a subset of prefetched session instead of all as it's too intrusive experience.

## Solution

Only display prefetched sessions when the session ids are identical to the very first one.

---


jpinkney-aws:
@Will-ShaoHua can you add a problem/description? if its just a wip can you convert this to a draft PR

andrewyuq (commented):
on hold while discussing whether first trigger and subsequent pre-fetch triggers share the same sessionId

PR #6556, opened 2025-02-11
fix(build): webviews sometimes did not load during development

nkomonen-amazon:
## Problem:

In our development build we spawn a webpack server that serves the Vue files used by our webviews. It is served on port 8080, but sometimes an existing process is already using it, and this causes the webpack server to use the next available port. But because we assume it will use 8080 it silently breaks, and when developing the Q webview would not load as expected.

## Solution:

Explicitly require 8080 in the webpack server cli command, also create a custom script to catch the error early and explain how to resolve it.

Now in the build tasks if 8080 is in use the dev can see the failed task with the solution explicitly stated.

---

### DIFF
--- a/packages/amazonq/.vscode/tasks.json
+++ b/packages/amazonq/.vscode/tasks.json
@@ -44,7 +44,8 @@
             "problemMatcher": "$ts-webpack-watch",
             "options": {
                 "cwd": "${workspaceFolder}/../../packages/core"
-            }
+            },
+            "presentation": { "panel": "dedicated" }
         },
         {
             "label": "terminate",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -424,7 +424,7 @@
         "compileOnly": "tsc -p ./",
         "compileDev": "npm run compile -- --mode development",
         "webpackDev": "webpack --mode development",
-        "serveVue": "webpack serve --config-name vue --mode development",
+        "serveVue": "ts-node ./scripts/build/checkServerPort.ts && webpack serve --port 8080 --config-name vue --mode development",
         "watch": "npm run clean && npm run buildScripts && npm run compileOnly -- --watch",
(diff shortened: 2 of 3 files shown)


packages/amazonq/.vscode/tasks.json
```diff
@@ -44,7 +44,8 @@
             "problemMatcher": "$ts-webpack-watch",
             "options": {
                 "cwd": "${workspaceFolder}/../../packages/core"
-            }
+            },
+            "presentation": { "panel": "dedicated" }
```
nkomonen-amazon, line 48: This solved the following issue:
- The webpack serve command would fail
- A subsequent build task would reuse this terminal since it failed
- The reuse caused the port in use error message to be hidden

Now the webpack serve command is in a dedicated terminal that does not disappear, so it is more obvious when it breaks.

packages/core/package.json
```diff
@@ -424,7 +424,7 @@
         "compileOnly": "tsc -p ./",
         "compileDev": "npm run compile -- --mode development",
         "webpackDev": "webpack --mode development",
-        "serveVue": "webpack serve --config-name vue --mode development",
+        "serveVue": "ts-node ./scripts/build/checkServerPort.ts && webpack serve --port 8080 --config-name vue --mode development",
```
jpinkney-aws, line 427: is the explicitness just to make it easy to identify that the webpack port and the port in checkServerPort are easy?

jpinkney-aws (approved):

packages/core/package.json
nkomonen-amazon: the explicit `--port 8080` is a fallback in case something breaks in the script, but isn't explicitly needed. Also by default `webpack serve` will use 8080, so I thought it would just help make that more obvious.

justinmk3 (approved):

PR #6565, opened 2025-02-12
test(core): Use a fake global state for integ/e2e tests

jpinkney-aws:
## Problem
Features that depend on global state can cause test failures locally because the tests will use your current global state settings instead of a fake state for integ/e2e tests

## Solution
integ/e2e tests should use a fake global state instead that automatically gets reset after each test

---


justinmk3 (approved):
We may want to start thinking about a base-class for `test/globalSetup.ts` and `testInteg/globalSetup.ts`. Seems like there is a lot of overlap.

Issue #6566, opened 2025-02-12
Stuck in "Uploading code" in multi-root workspace


avi-alpert:
## Problem

Zipstream is throwing `Error: File already exists at ZipWriter.add` if writeFile is being called twice with the same path. `_numberOfFilesSucceeded` never reaches `_numberOfFilesToStream` so the user is stuck in "Uploading code" phase.

This affects /dev and /doc being run on a multi-root workspace with the same file paths in it multiple times.

I'm thinking that zipstream needs some sort of error handling mechanism to skip adding a file if theres a failure.

This bug was introduced in last weeks release (https://github.com/aws/aws-toolkit-vscode/commit/21c7f12132dc4cc72724e1946dc0ce1ea627edc0)

## Steps to reproduce the issue

1) Create a multi root workspace in vs code that has the same file in it multiple times, for example:

![Image](https://github.com/user-attachments/assets/63e50b7a-36e3-4a5f-970d-c9fa387a2eb0)

2) Run /dev or /doc

## System details (run `AWS: About` and/or `Amazon Q: About`)
OS: Darwin arm64 23.6.0
Visual Studio Code extension host:  1.96.4
Amazon Q:  1.46.0
node: 20.18.1
electron: 32.2.6

Hweinstock:
Fixed in https://github.com/aws/aws-toolkit-vscode/pull/6568

Hweinstock:
Problem was fixed in usage of ZipStream, but not ZipStream itself.

PR #6568, opened 2025-02-12
fix(amazonq): Prevent adding same file twice for /doc and /dev

avi-alpert:
## Problem
Stuck on "Uploading code..." for /dev and /doc in a multi-root workspace with the same file paths in it multiple times.

## Solution
Add a check to prevent same file from being added to archive twice

Fixes issue https://github.com/aws/aws-toolkit-vscode/issues/6566

---

### DIFF
--- a/packages/amazonq/.changes/next-release/Bug Fix-037e19bd-95f5-4cdf-8408-6eb273ff7ebd.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-037e19bd-95f5-4cdf-8408-6eb273ff7ebd.json
@@ -1,4 +1,4 @@
 {
 	"type": "Bug Fix",
-	"description": "Stuck on `Uploading code...` for /dev and /doc in some multi-root workspaces"
+	"description": "/dev and /doc: Multi-root workspace with duplicate files causes infinite 'Uploading code...' loop"
 }
--- a/packages/amazonq/.changes/next-release/Bug Fix-037e19bd-95f5-4cdf-8408-6eb273ff7ebd.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-037e19bd-95f5-4cdf-8408-6eb273ff7ebd.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Stuck on `Uploading code...` for /dev and /doc in some multi-root workspaces"
+}
--- a/packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
+++ b/packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
@@ -15,6 +15,8 @@ import { fs, AmazonqCreateUpload, ZipStream } from 'aws-core-vscode/shared'
 import { MetricName, Span } from 'aws-core-vscode/telemetry'
 import sinon from 'sinon'
 import { CodeWhispererSettings } from 'aws-core-vscode/codewhisperer'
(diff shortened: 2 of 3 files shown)


packages/core/src/amazonq/util/files.ts
```diff
@@ -48,8 +48,14 @@ export async function prepareRepoData(
 
         let totalBytes = 0
         const ignoredExtensionMap = new Map<string, number>()
+        const addedFilePaths = new Set()
 
         for (const file of files) {
+            if (addedFilePaths.has(file.zipFilePath)) {
```
vikshe, line 54: does collectFiles return same file path for different workspace folders ? 
if so should file be prefixed with workspace folder ?

avi-alpert: If we have a file that shows up in a multi root workspace more than once, then collectFiles will output it multiple times, all with the same zipFilePath

avi-alpert:
Duplicate code CI failure: none of the files in this PR are listed

KevinDing1 (approved):

jpinkney-aws:
Can we create some tests for this?

avi-alpert:
> Can we create some tests for this?

I confirmed that the unit test I added fails if I revert the fix, with the same error that causes the user facing bug.

packages/amazonq/test/unit/amazonqFeatureDev/util/files.test.ts
```diff
@@ -126,5 +133,18 @@ describe('file utils', () => {
...
+            const folder = await TestFolder.create()
+            const testFilePath = 'innerFolder/file.md'
+            await folder.write(testFilePath, 'test content')
+
+            // Add a folder and its subfolder to the workspace
+            const workspace1 = getWorkspaceFolder(folder.path)
+            const workspace2 = getWorkspaceFolder(folder.path + '/innerFolder')
+            const folderName = path.basename(folder.path)
+
+            await testPrepareRepoData([workspace1, workspace2], [`${folderName}_${workspace1.name}/${testFilePath}`])
```
jpinkney-aws, line 147: We're mostly just testing that this doesn't reject right? Does it make sense to wrap this with:
```
assert.doesNotReject(() => {
    await testPrepareRepoData([workspace1, workspace2], [`${folderName}_${workspace1.name}/${testFilePath}`])
})
```
? otherwise the PR looks good to me

avi-alpert: the testPrepareRepoData has asserts inside of it, which will fail if the function fails

avi-alpert: The error thrown by the test would be `Error: File already exists at ZipWriter.add ` which indicates the root cause, the extra assert is not needed

packages/amazonq/.changes/next-release/Bug Fix-037e19bd-95f5-4cdf-8408-6eb273ff7ebd.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Stuck on `Uploading code...` for /dev and /doc in some multi-root workspaces"
```
jpinkney-aws, line 3: One more small thing: should we clarify the case when this happens? Something like:
```
/dev and /doc: Duplicate file detection causes infinite 'Uploading code...' loop
```

jpinkney-aws (approved):

PR #6573, opened 2025-02-13
feat(amazonq): user cancellable lsp download

tomcat323:
## Problem
the cancellation button on notification doesn't actually stop the LSP download

## Solution
Route the notification timeout into `httpResourceFetcher` call, so that the user cancellation event stops the download as well.

Demo: non-cancelled and cancelled downloads:

https://github.com/user-attachments/assets/4f3d4ed0-635a-4f88-aef7-ee62e87bfc83

---

### DIFF
--- a/packages/core/src/shared/lsp/lspResolver.ts
+++ b/packages/core/src/shared/lsp/lspResolver.ts
@@ -226,7 +226,7 @@ export class LanguageServerResolver {
                 res: await new HttpResourceFetcher(content.url, {
                     showUrl: true,
                     timeout: timeout,
-                    swallowError: false,
+                    throwOnError: true,
                 }).get(),
                 hash: content.hashes[0],
                 filename: content.filename,
--- a/packages/core/src/shared/lsp/lspResolver.ts
+++ b/packages/core/src/shared/lsp/lspResolver.ts
@@ -17,6 +17,9 @@ import { HttpResourceFetcher } from '../resourcefetcher/httpResourceFetcher'
 import { showMessageWithCancel } from '../../shared/utilities/messages'
 import { Timeout } from '../utilities/timeoutUtils'
 
+// max timeout for downloading remote LSP assets progress, the lowest possible is 3000, bounded by httpResourceFetcher's waitUntil
+const remoteDownloadTimeout = 5000
+
 export class LanguageServerResolver {
     constructor(
         private readonly manifest: Manifest,
@@ -89,10 +92,7 @@ export class LanguageServerResolver {
      * Returns a timeout to be passed down into httpFetcher to handle user cancellation
      */
     private async showDownloadProgress() {
-        const timeout = new Timeout(5000)
-        timeout?.token.onCancellationRequested((event) => {
-            logger.info('Remote download cancelled by user')
-        })
+        const timeout = new Timeout(remoteDownloadTimeout)
         await showMessageWithCancel(`Downloading '${this.lsName}' language server`, timeout)
         return timeout
     }
--- a/packages/core/src/shared/lsp/lspResolver.ts
+++ b/packages/core/src/shared/lsp/lspResolver.ts
@@ -33,8 +33,6 @@ export class LanguageServerResolver {
      * @throws ToolkitError if no compatible version can be found
      */
     async resolve() {
-        const timeout = new Timeout(5000)
-        await showMessageWithCancel(`Downloading '${this.lsName}' language server`, timeout)
         function getServerVersion(result: LspResult) {
             return {
                 languageServerVersion: result.version,
@@ -63,7 +61,6 @@ export class LanguageServerResolver {
             return await tryStageResolvers('getServer', serverResolvers, getServerVersion)
         } finally {
             logger.info(`Finished setting up LSP server`)
-            timeout.cancel()
         }
     }
 
@@ -87,19 +84,37 @@ export class LanguageServerResolver {
         }
     }
 
+    /**
+     * Show a toast notification with progress bar for lsp remote downlaod
+     * Returns a timeout to be passed down into httpFetcher to handle user cancellation
+     */
+    private async showDownloadProgress() {
+        const timeout = new Timeout(5000)
+        timeout?.token.onCancellationRequested((event) => {
+            logger.info('Remote download cancelled by user')
+        })
+        await showMessageWithCancel(`Downloading '${this.lsName}' language server`, timeout)
+        return timeout
+    }
+
     private async fetchRemoteServer(
         cacheDirectory: string,
         latestVersion: LspVersion,
         targetContents: TargetContent[]
     ): Promise<LspResult> {
-        if (await this.downloadRemoteTargetContent(targetContents, latestVersion.serverVersion)) {
-            return {
-                location: 'remote',
-                version: latestVersion.serverVersion,
-                assetDirectory: cacheDirectory,
+        const timeout = await this.showDownloadProgress()
+        try {
+            if (await this.downloadRemoteTargetContent(targetContents, latestVersion.serverVersion, timeout)) {
+                return {
+                    location: 'remote',
+                    version: latestVersion.serverVersion,
+                    assetDirectory: cacheDirectory,
+                }
+            } else {
+                throw new ToolkitError('Failed to download server from remote', { code: 'RemoteDownloadFailed' })
             }
-        } else {
-            throw new ToolkitError('Failed to download server from remote', { code: 'RemoteDownloadFailed' })
+        } finally {
+            timeout.dispose()
         }
     }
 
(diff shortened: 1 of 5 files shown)


packages/core/src/shared/lsp/lspResolver.ts
```diff
@@ -87,19 +84,37 @@ export class LanguageServerResolver {
...
         targetContents: TargetContent[]
     ): Promise<LspResult> {
-        if (await this.downloadRemoteTargetContent(targetContents, latestVersion.serverVersion)) {
-            return {
-                location: 'remote',
-                version: latestVersion.serverVersion,
-                assetDirectory: cacheDirectory,
+        const timeout = await this.showDownloadProgress()
+        try {
+            if (await this.downloadRemoteTargetContent(targetContents, latestVersion.serverVersion, timeout)) {
```
jpinkney-aws, line 107: just to double check: when the timeout gets triggered when the user clicks cancel, that doesn't end up throwing an error or anything right? just a normal return?

packages/core/src/shared/lsp/lspResolver.ts
```diff
@@ -87,19 +84,37 @@ export class LanguageServerResolver {
         }
     }
 
+    /**
+     * Show a toast notification with progress bar for lsp remote downlaod
+     * Returns a timeout to be passed down into httpFetcher to handle user cancellation
+     */
+    private async showDownloadProgress() {
+        const timeout = new Timeout(5000)
```
jpinkney-aws, line 92: probably want that 5000 store as a const somewhere at the top of the file so its easy to find

packages/core/src/shared/lsp/lspResolver.ts
tomcat323: correct, and it would go to the fall back server.

packages/core/src/shared/lsp/lspResolver.ts
```diff
@@ -87,19 +84,37 @@ export class LanguageServerResolver {
...
+        timeout?.token.onCancellationRequested((event) => {
```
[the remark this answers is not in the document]
tomcat323: yea they have debug level logging on cancellation, this is removed

packages/core/src/shared/lsp/lspResolver.ts
```diff
@@ -202,7 +217,7 @@ export class LanguageServerResolver {
 
         const fetchTasks = contents.map(async (content) => {
             return {
-                res: await new HttpResourceFetcher(content.url, { showUrl: true }).get(),
+                res: await new HttpResourceFetcher(content.url, { showUrl: true, timeout: timeout }).get(),
```
justinmk3, line 220: This is the expected pattern 👍 

Alternatively, if `HttpResourceFetcher.cancel()` exists, that could be called from the `onCancellationRequested` handler (line  93).

packages/core/src/shared/lsp/lspResolver.ts
justinmk3: @tomcat323 please check whether `isUserCancelledError` is useful for this PR. https://github.com/aws/aws-toolkit-vscode/blob/99eefe00c54e47e99ceca9434e62cf71b4067d56/packages/core/src/shared/errors.ts#L620

packages/core/src/shared/lsp/lspResolver.ts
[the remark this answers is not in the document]
justinmk3: That will be the only way to confirm that cancel actually is working. So it's good that there is a log message in some fashion.

jpinkney-aws (approved):

hayemaxi:
After some discussion, do we still want this? Cancelling the LSP download means the extension may become unusable.

jpinkney-aws:
TBH I'm against this now after the meeting yesterday, I think we should still show progress but if the options are:
1. You always download the language server and get all the features 
OR
2. You click cancel and then nothings working except for auth

I'd much rather just have them always download the language server

packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
```diff
@@ -82,7 +86,10 @@ export class HttpResourceFetcher implements ResourceFetcher<Response> {
                 `Error downloading ${this.logText()}: %s`,
                 error.message ?? error.code ?? error.response.statusText ?? error.response.status
             )
-            return undefined
+            if (this.params.swallowError) {
```
tomcat323, line 89: previously this is swallowing all errors from fetch. With this change we can detect and catch the user cancellation errors.

packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
```diff
@@ -112,7 +119,10 @@ export class HttpResourceFetcher implements ResourceFetcher<Response> {
                 timeout: 3000,
                 interval: 100,
                 backoff: 2,
-                retryOnFail: true,
+                retryOnFail: (error: Error) => {
+                    // retry unless we got an user cancellation error
+                    return !(error instanceof RequestCancelledError)
```
tomcat323, line 124: Purpose of this added `waitUntil` feature:
We need to stop download retries if its an user cancel error
But for all other error types from fetch, we continue retry.

packages/core/src/shared/utilities/timeoutUtils.ts
```diff
@@ -247,6 +248,11 @@ export async function waitUntil<T>(
     fn: () => Promise<T>,
     options: WaitUntilOptions & { retryOnFail: false }
 ): Promise<T | undefined>
+export async function waitUntil<T>(
```
jpinkney-aws, line 251: should probably modify the timeout utils to add a couple new tests for this

packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
```diff
@@ -22,15 +22,19 @@ export class HttpResourceFetcher implements ResourceFetcher<Response> {
...
      * @param {number} params.retries The number of retries a get request should make if one fails
+     * @param {boolean} params.swallowError True if we want to swallow the request error, false if we want the error to keep bubbling up
      */
     public constructor(
         private readonly url: string,
         private readonly params: {
             showUrl: boolean
             friendlyName?: string
             timeout?: Timeout
+            swallowError?: boolean
```
jpinkney-aws, line 33: couple other naming ideas: throwOnError, supressError or propogateError

packages/core/src/shared/utilities/timeoutUtils.ts
tomcat323: added in `timeoutUtils.test.ts`

packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
justinmk3: is it different than `retryOnFail` ?

packages/core/src/shared/utilities/timeoutUtils.ts
```diff
@@ -267,6 +273,17 @@ export async function waitUntil<T>(fn: () => Promise<T>, options: WaitUntilOptio
     let elapsed: number = 0
     let remaining = opt.timeout
 
+    // Internal helper to determine if we should retry
+    function shouldRetry(error: Error | undefined): boolean {
+        if (error === undefined) {
+            return typeof opt.retryOnFail === 'boolean' ? opt.retryOnFail : true
+        }
+        if (typeof opt.retryOnFail === 'function') {
```
justinmk3, line 281: Can we make it always be a function? That should simplify some logic, eliminate an overload, and eliminate some tests.

It's trivial for calls to pass `() => true` if they want a "constant".

packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
```diff
@@ -112,7 +119,10 @@ export class HttpResourceFetcher implements ResourceFetcher<Response> {
                 timeout: 3000,
                 interval: 100,
                 backoff: 2,
-                retryOnFail: true,
+                retryOnFail: (error: Error) => {
+                    // retry unless we got an user cancellation error
```
justinmk3, line 123:
```suggestion
                    // Retry unless the user intentionally canceled the operation.
```

packages/core/src/shared/lsp/lspResolver.ts
```diff
@@ -87,19 +88,39 @@ export class LanguageServerResolver {
...
+                    assetDirectory: cacheDirectory,
+                }
+            } else {
+                throw new ToolkitError('Failed to download server from remote', { code: 'RemoteDownloadFailed' })
             }
-        } else {
-            throw new ToolkitError('Failed to download server from remote', { code: 'RemoteDownloadFailed' })
+        } catch (err) {
+            if (err instanceof RequestCancelledError) {
+                throw new CancellationError('user')
```
justinmk3, line 119: Why not add this to `isUserCancelledError` ? https://github.com/aws/aws-toolkit-vscode/blob/71650e5c7a9b301cd1a24594f2b89147d99ceefc/packages/core/src/shared/errors.ts#L620-L621

justinmk3:
https://github.com/aws/aws-toolkit-vscode/pull/6573#discussion_r1974238222 wasn't addressed

Issue #6594, opened 2025-02-17
unreliable test: makeSharedIniFileCredentialsProvider assumes role with mfa token:


Hweinstock:
## Test Details

-   Name of test:  assumes role with mfa token:
-   OS of failure: Mac
-   VSCode version of failure (minimum/insider/stable): insiders
-   Link to failing run: https://github.com/aws/aws-toolkit-vscode/actions/runs/13338335760/job/37258294440
-   Link to failing test: https://github.com/aws/aws-toolkit-vscode/blob/0fa1b5b47ab642458df934192832a4ffcb6c9d1e/packages/core/src/test/credentials/provider/sharedCredentialsProvider.test.ts#L499-L517

## Log of Test Failure
```
TypeError: inputBox.hide is not a function
      at promptUser (/Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/core/src/shared/ui/input.ts:133:18)
      at async getMfaTokenFromUser (/Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/core/src/auth/credentials/utils.ts:124:19)
      at async /Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/core/src/auth/providers/sharedCredentialsProvider.ts:422:38

deleteTestTempDirs: deleted 410 test temp dirs
```

PR #6596, opened 2025-02-17
test(auth): avoid deprecated InputBox usage that caused flakiness. 

Hweinstock:
## Problem
https://github.com/aws/aws-toolkit-vscode/issues/6594

The `getMfaCodeFromUser` currently uses a deprecated version of `createInputBox` and the following `promptUser` function: 
https://github.com/aws/aws-toolkit-vscode/blob/0fa1b5b47ab642458df934192832a4ffcb6c9d1e/packages/core/src/shared/ui/input.ts#L72-L135

Based on the stack trace, there is a race condition here that can lead to a type error, but unsure how?
Stack trace: 
```
TypeError: inputBox.hide is not a function
      at promptUser (/Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/core/src/shared/ui/input.ts:133:18)
      at async getMfaTokenFromUser (/Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/core/src/auth/credentials/utils.ts:124:19)
      at async /Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/core/src/auth/providers/sharedCredentialsProvider.ts:422:38
```
### Investigation Notes:
- The error only happens when running the test suite as a whole. 
- The error only happens in CI (unable to reproduce locally). 

## Solution
- avoid use of deprecated `createInputBox` [function](https://github.com/aws/aws-toolkit-vscode/blob/0fa1b5b47ab642458df934192832a4ffcb6c9d1e/packages/core/src/shared/ui/input.ts#L31-L58), prefer new version instead.  
- Ran the tests 4x with 1000x on the test, and did not see a failure. 

## Future Work 
- Migrate existing cases away from this deprecated `createInputBox` if there is evidence it leads to flaky tests or unreliable behavior. 
---

### DIFF
--- a/packages/core/src/auth/credentials/utils.ts
+++ b/packages/core/src/auth/credentials/utils.ts
@@ -14,8 +14,9 @@ import { messages, showMessageWithCancel, showViewLogsMessage } from '../../shar
 import { Timeout, waitTimeout } from '../../shared/utilities/timeoutUtils'
 import { fromExtensionManifest } from '../../shared/settings'
 import { Profile } from './sharedCredentials'
-import { createInputBox, promptUser } from '../../shared/ui/input'
 import { openUrl } from '../../shared/utilities/vsCodeUtils'
+import { createInputBox } from '../../shared/ui/inputPrompter'
+import { isValidResponse } from '../../shared/wizards/wizard'
 
 const credentialsTimeout = 300000 // 5 minutes
 const credentialsProgressDelay = 1000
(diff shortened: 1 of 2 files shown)


jpinkney-aws:
Could part of the problem be because we override vscode.window in our unit tests?
- https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/test/shared/vscode/window.ts#L250
- https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/test/globalSetup.test.ts#L90

Hweinstock:
Yea, I was digging through that code a bit, and suspect its making some assumptions about how we create our prompters. Its still unclear to me how the old `createInputBox` break those assumptions.

PR #6603, opened 2025-02-17
fix(appbuilder): pass in the negative version of --use-container when using build quickpick

mbfreder:
## Problem
When users use appbuilder to build their lambda functions, they choose between using their samconfig file or manually selecting the build parameters/flags. The problem is that when the user selects build flags and intentionally doesn't select the ```--use-container``` flag, the command will still be run with --use-container if the samconfig file has ```use_container``` is set to true.

## Solution
Whenever the user manually selects the build flags and doesn't select ```--use-container```, we add the negative version of ```--use-container```, which is ```--no-use-container```. This serves as an override if the samconfig file has ```--use-container``` set to true.

---

### DIFF
--- a/packages/core/src/shared/sam/build.ts
+++ b/packages/core/src/shared/sam/build.ts
@@ -284,7 +284,7 @@ function resolveBuildArgConflict(boundArgs: string[]): string[] {
     return Array.from(boundArgsSet)
 }
 export async function resolveBuildFlags(buildFlags: string[], samCliVersion: SemVer | null): Promise<string[]> {
-    // const samCliVersion = (await getSamCliPathAndVersion()).parsedVersion
+    // --no-use-container was not added until v1.133.0
     if (samCliVersion?.compare('1.133.0') ?? -1 < 0) {
         return buildFlags
     }
--- a/packages/core/src/shared/sam/build.ts
+++ b/packages/core/src/shared/sam/build.ts
@@ -29,6 +29,7 @@ import {
 } from './utils'
 import { getConfigFileUri, validateSamBuildConfig } from './config'
 import { runInTerminal } from './processTerminal'
+import { getLogger } from '../logger'
 
 const buildMementoRootKey = 'samcli.build.params'
 export interface BuildParams {
@@ -213,7 +214,7 @@ export async function runBuild(arg?: TreeNode): Promise<SamBuildResult> {
     // refactor
     const buildFlags: string[] =
         params.paramsSource === ParamsSource.Specify && params.buildFlags
-            ? JSON.parse(params.buildFlags)
+            ? resolveBuildFlags(JSON.parse(params.buildFlags))
             : await getBuildFlags(params.paramsSource, projectRoot, defaultFlags)
 
     // todo remove
@@ -280,4 +281,10 @@ function resolveBuildArgConflict(boundArgs: string[]): string[] {
     //     boundArgsSet.add('--no-beta-features')
     // }
     return Array.from(boundArgsSet)
+}
+export function resolveBuildFlags(buildFlags: string[]): string[] {
+    if (!buildFlags.includes('--use-container')) {
+        buildFlags.push('--no-use-container')
+    }
+    return buildFlags
 }
--- a/packages/core/src/shared/sam/build.ts
+++ b/packages/core/src/shared/sam/build.ts
@@ -29,7 +29,7 @@ import {
 } from './utils'
 import { getConfigFileUri, validateSamBuildConfig } from './config'
 import { runInTerminal } from './processTerminal'
-import { getLogger } from '../logger'
+import { SemVer } from 'semver'
 
 const buildMementoRootKey = 'samcli.build.params'
 export interface BuildParams {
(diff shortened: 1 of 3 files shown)


packages/core/src/shared/sam/build.ts
```diff
@@ -281,3 +282,9 @@ function resolveBuildArgConflict(boundArgs: string[]): string[] {
     // }
     return Array.from(boundArgsSet)
 }
+export function resolveBuildFlags(buildFlags: string[]): string[] {
```
roger-zhangg, line 285: Check if sam cli version is larger than 1.133 before applying this logic.

roger-zhangg (changes requested):

packages/core/src/test/shared/sam/build.test.ts
```diff
@@ -229,6 +230,14 @@ describe('SAM build helper functions', () => {
...
         })
     })
+
+    describe('resolveBuildFlags', () => {
+        it('should add --no-use-container if the buildFlags array does not contain --use-container', () => {
+            const buildFlags: string[] = ['--cached', '--debug', '--parallel']
+            const resolvedBuildFlags = ['--cached', '--debug', '--parallel', '--no-use-container']
+            assert.deepStrictEqual(resolvedBuildFlags, resolveBuildFlags(buildFlags))
+        })
+    })
```
roger-zhangg, line 240: Consider adding another test to ensure `--no-use-container` will not be applied to lower sam versions

packages/core/src/shared/sam/build.ts
mbfreder: Good callout. Thanks. Updated

packages/core/src/test/shared/sam/build.test.ts
mbfreder: Thanks. Added

roger-zhangg (approved):
LGTM

packages/core/src/shared/sam/build.ts
```diff
@@ -281,3 +283,13 @@ function resolveBuildArgConflict(boundArgs: string[]): string[] {
     // }
     return Array.from(boundArgsSet)
 }
+export async function resolveBuildFlags(buildFlags: string[], samCliVersion: SemVer | null): Promise<string[]> {
+    // const samCliVersion = (await getSamCliPathAndVersion()).parsedVersion
```
jpinkney-aws, line 287: nit: remove this if its not needed

packages/core/src/shared/sam/build.ts
```diff
@@ -281,3 +283,13 @@ function resolveBuildArgConflict(boundArgs: string[]): string[] {
...
+    if (samCliVersion?.compare('1.133.0') ?? -1 < 0) {
```
jpinkney-aws, line 288: what are these magic numbers telling us? Might want a comment for anyone who looks at this in the future

packages/core/src/test/shared/sam/build.test.ts
```diff
@@ -229,6 +232,46 @@ describe('SAM build helper functions', () => {
...
+
+        beforeEach(() => {
+            sandbox = sinon.createSandbox()
+        })
+
+        afterEach(() => {
+            sandbox.restore()
+        })
+
+        it('should add --no-use-container if the buildFlags array does not contain --use-container', async () => {
```
jpinkney-aws, line 248: this title seems really into the weeds with the implementation. Maybe something like `uses --no-use-container when --use-container is absent`

packages/core/src/test/shared/sam/build.test.ts
```diff
@@ -229,6 +232,46 @@ describe('SAM build helper functions', () => {
...
+            const normalVersion = { path: 'file:///path/to/cli', parsedVersion: new SemVer('1.133.0') }
+            pathAndVersionStub = sandbox.stub().resolves(normalVersion)
+            sandbox.stub(utils, 'getSamCliPathAndVersion').callsFake(pathAndVersionStub)
+            const buildFlags: string[] = ['--cached', '--debug', '--parallel']
+            const resolvedBuildFlags = ['--cached', '--debug', '--parallel', '--no-use-container']
+            assert.deepEqual(resolvedBuildFlags, await resolveBuildFlags(buildFlags, normalVersion.parsedVersion))
+        })
+
+        it('should return the original buildFlags array if SAM CLI version is less than 1.133', async () => {
```
jpinkney-aws, line 257: Maybe `preserves buildFlags when SAM CLI version < 1.133`

packages/core/src/test/shared/sam/build.test.ts
```diff
@@ -229,6 +232,46 @@ describe('SAM build helper functions', () => {
...
+            const lowerVersion = { path: 'file:///path/to/cli', parsedVersion: new SemVer('1.110.0') }
+            const pathAndVersionStub = sandbox.stub().resolves(lowerVersion)
+            sandbox.stub(utils, 'getSamCliPathAndVersion').callsFake(pathAndVersionStub)
+            const buildFlags = ['--cached', '--parallel', '--save-params']
+            const resolvedBuildFlags = await resolveBuildFlags(buildFlags, lowerVersion.parsedVersion)
+            assert.deepEqual(resolvedBuildFlags, buildFlags)
+        })
+
+        it('should not add "--no-use-container" if "--use-container" is already present', async () => {
```
jpinkney-aws, line 266: Maybe `respects existing --use-container flag`

packages/core/src/test/shared/sam/build.test.ts
```diff
@@ -229,6 +232,46 @@ describe('SAM build helper functions', () => {
...
+        beforeEach(() => {
+            sandbox = sinon.createSandbox()
+        })
+
+        afterEach(() => {
+            sandbox.restore()
+        })
+
+        it('should add --no-use-container if the buildFlags array does not contain --use-container', async () => {
+            const normalVersion = { path: 'file:///path/to/cli', parsedVersion: new SemVer('1.133.0') }
```
jpinkney-aws, line 249: does it make sense to turn all of this into a function that you can set the parsedVersion and add flags to resolvedBuildFlags and use that function in the 3 tests? The 3 tests look almost the same with minor adjustments

mbfreder: Refactered the tests. Thanks

jpinkney-aws (approved):

PR #6604, opened 2025-02-18
feat(amazonq): manifest deprecation message with optional don't show

tomcat323:
## Problem
If the language server manifest is deprecated we just show it in the logs. However, users may not be explicitly aware that they will need to upgrade the extension to get a different manifest.

Also according to the design specsheet, user should have the option to mute the message for good once seen.

## Solution

- Added a message window with `don't show` and `ok` buttons for deprecated LSP manifest.
- Use AmazonQ suppressed prompts for `don't show` tracking
- Added a re-enable feature so that we can re-enable that setting if manifest gets updated. 

## Note

Also added language server name in logs, helpful to track different instances

---

### DIFF
--- a/packages/core/src/dev/activation.ts
+++ b/packages/core/src/dev/activation.ts
@@ -26,7 +26,6 @@ import { DevNotificationsState } from '../notifications/types'
 import { QuickPickItem } from 'vscode'
 import { ChildProcess } from '../shared/utilities/processUtils'
 import { WorkspaceLSPResolver } from '../amazonq/lsp/workspaceInstaller'
-import * as ManifestResolver from '../shared/lsp/manifestResolver'
 
 interface MenuOption {
     readonly label: string
--- a/packages/core/src/dev/activation.ts
+++ b/packages/core/src/dev/activation.ts
@@ -26,6 +26,7 @@ import { DevNotificationsState } from '../notifications/types'
 import { QuickPickItem } from 'vscode'
 import { ChildProcess } from '../shared/utilities/processUtils'
 import { WorkspaceLSPResolver } from '../amazonq/lsp/workspaceInstaller'
+import * as ManifestResolver from '../shared/lsp/manifestResolver'
 
 interface MenuOption {
     readonly label: string
@@ -453,9 +454,15 @@ const resettableFeatures: readonly ResettableFeature[] = [
     },
     {
         name: 'workspace lsp',
-        label: 'Lsp',
+        label: 'Lsp Download',
         detail: 'Resets workspace LSP',
         executor: resetWorkspaceLspDownload,
+    },
+    {
+        name: 'lsp global state',
+        label: 'Lsp State',
+        detail: 'Resets LSP manifest global state',
+        executor: resetLSPGlobalState,
     },
 ] as const
 
@@ -547,6 +554,10 @@ async function resetNotificationsState() {
 
 async function resetWorkspaceLspDownload() {
     await new WorkspaceLSPResolver().resolve()
+}
+
+async function resetLSPGlobalState() {
+    await ManifestResolver.resetManifestState()
 }
 
 async function editNotifications() {
--- a/packages/core/src/dev/activation.ts
+++ b/packages/core/src/dev/activation.ts
@@ -454,7 +454,7 @@ const resettableFeatures: readonly ResettableFeature[] = [
     },
     {
         name: 'workspace lsp',
-        label: 'Lsp Download',
+        label: 'Download Lsp ',
         detail: 'Resets workspace LSP',
         executor: resetWorkspaceLspDownload,
     },
--- a/packages/core/src/dev/activation.ts
+++ b/packages/core/src/dev/activation.ts
@@ -458,12 +458,6 @@ const resettableFeatures: readonly ResettableFeature[] = [
         detail: 'Resets workspace LSP',
         executor: resetWorkspaceLspDownload,
     },
-    {
-        name: 'lsp global state',
-        label: 'Lsp State',
-        detail: 'Resets LSP manifest global state',
-        executor: resetLSPGlobalState,
-    },
 ] as const
 
 // TODO this is *somewhat* similar to `openStorageFromInput`. If we need another
@@ -554,10 +548,6 @@ async function resetNotificationsState() {
 
 async function resetWorkspaceLspDownload() {
     await new WorkspaceLSPResolver().resolve()
-}
-
-async function resetLSPGlobalState() {
-    await ManifestResolver.resetManifestState()
 }
 
 async function editNotifications() {
--- a/packages/core/src/shared/lsp/manifestResolver.ts
+++ b/packages/core/src/shared/lsp/manifestResolver.ts
@@ -111,14 +111,14 @@ export class ManifestResolver {
         const prompts = AmazonQPromptSettings.instance
         const lspId = `${this.lsName}LspManifestMessage`
 
-        // sanity check, if the lsName is changed then we also need to update the prompt keys in settings-amazonq.gen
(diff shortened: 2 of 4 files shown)


packages/core/src/shared/lsp/manifestResolver.ts
```diff
@@ -90,25 +99,47 @@ export class ManifestResolver {
...
     }
 
     private async saveManifest(etag: string, content: string): Promise<void> {
         const storage = this.getStorage()
 
+        // Only true when incoming manifest is deprecated & existing muteDeprecation is true (set by user)
+        const muteDeprecation =
+            (storage[this.lsName] ? storage[this.lsName].muteDeprecation : false) &&
+            (JSON.parse(content) as Manifest).isManifestDeprecated
+
```
tomcat323, line 136: This looks ugly because it covers the case where a manifest was deprecated and then "revived". In such case we re-enable the flag to show deprecation message.

packages/core/src/dev/activation.ts
```diff
@@ -453,10 +454,16 @@ const resettableFeatures: readonly ResettableFeature[] = [
     },
     {
         name: 'workspace lsp',
-        label: 'Lsp',
+        label: 'Lsp Download',
```
jpinkney-aws, line 457: Should this be reversed to `Download Lsp`?

packages/core/src/shared/lsp/manifestResolver.ts
```diff
@@ -90,25 +99,47 @@ export class ManifestResolver {
...
+     * - OK: close and do nothing
+     * - Don't Show Again: Update global state (muteDecprecation) so the deprecation message is never shown for this manifest.
+     * @param manifest
+     */
     private checkDeprecation(manifest: Manifest): void {
-        if (manifest.isManifestDeprecated) {
-            logger.info('This LSP manifest is deprecated. No future updates will be available.')
+        if (manifest.isManifestDeprecated && !this.getStorage()[this.lsName].muteDeprecation) {
+            const deprecationMessage = `${this.lsName} manifest is deprecated. No future updates will be available.`
+            logger.info(deprecationMessage)
```
jpinkney-aws, line 117: We might want to log the deprecated message regardless of whether or not they muted it just to give them some sort of heads up if they look at the logs. Otherwise people are just going to press don't show again and then never remember

packages/core/src/shared/lsp/manifestResolver.ts
```diff
@@ -90,25 +99,47 @@ export class ManifestResolver {
...
+                `Failed to parse ${this.lsName} manifest: ${error instanceof Error ? error.message : 'Unknown error'}`
             )
         }
     }
 
+    /**
+     * Check if the current manifest is deprecated.
+     * If yes and user hasn't muted this notification, shows a toast message with two buttons:
+     * - OK: close and do nothing
+     * - Don't Show Again: Update global state (muteDecprecation) so the deprecation message is never shown for this manifest.
```
jpinkney-aws, line 111:
```suggestion
     * - Don't Show Again: Update global state (muteDeprecation) so the deprecation message is never shown for this manifest.
```

packages/core/src/shared/lsp/manifestResolver.ts
jpinkney-aws: I don't think the spec covers this, we should probably bring it up in one of the meetings

packages/core/src/shared/lsp/manifestResolver.ts
```diff
@@ -90,25 +99,48 @@ export class ManifestResolver {
...
+    /**
+     * Check if the current manifest is deprecated.
+     * If yes and user hasn't muted this notification, shows a toast message with two buttons:
+     * - OK: close and do nothing
+     * - Don't Show Again: Update global state (muteDeprecation) so the deprecation message is never shown for this manifest.
+     * @param manifest
+     */
     private checkDeprecation(manifest: Manifest): void {
+        const deprecationMessage = `${this.lsName} manifest is deprecated. No future updates will be available.`
         if (manifest.isManifestDeprecated) {
```
jpinkney-aws, line 116: Could probably invert this like:
```typescript
if (!manifest.isManifestDeprecated) {
    return
}

logger.info(deprecationMessage)
if (!this.getStorage()[this.lsName].muteDeprecation) {
    void vscode.window
        .showInformationMessage(deprecationMessage, localizedText.ok, localizedText.dontShow)
        .then((button) => {
            if (button === localizedText.dontShow) {
                this.getStorage()[this.lsName].muteDeprecation = true
            }
        })
}
```
instead of having a inner `ifs` nested

packages/core/src/shared/lsp/manifestResolver.ts
jpinkney-aws: In the worst case `JSON.parse(content)` could fail, does it make sense to surround that with a try/catch? AFAICT its about 50/50 in the codebase on whether or people protect against that case

jpinkney-aws: Is there a different way to write this (even if its slightly more verbose) to make it easier to capture what it's trying to do?

tomcat323: used this.parseManifest(content), should also help a bit with verbosity

tomcat323: This has been changed to just `storage[this.lsName]?.muteDeprecation ?? false` 

Per the spec meeting we expect the manifest deprecation to be permanent unless flare tells us otherwise

packages/core/src/shared/lsp/manifestResolver.ts
```diff
@@ -71,12 +80,12 @@ export class ManifestResolver {
     }
 
     private async getLocalManifest(): Promise<Manifest> {
-        logger.info('Failed to download latest LSP manifest. Falling back to local manifest.')
+        logger.info(`Failed to download latest ${this.lsName} manifest. Falling back to local manifest.`)
```
justinmk3, line 83: same for the other cases. these names can have whitespace, right? Developers/users shouldn't have to manually parse the variable parts of a log message (which are often the important parts).

```suggestion
        logger.info(`Failed to download latest "${this.lsName}" manifest. Falling back to local manifest.`)
```

packages/core/src/shared/lsp/manifestResolver.ts
```diff
@@ -90,14 +95,45 @@ export class ManifestResolver {
...
+
+        // Sanity check, if the lsName is changed then we also need to update the prompt keys in settings-amazonq.gen
+        if (!(lspId in amazonQPrompts)) {
+            logger.error(`LSP ID "${lspId}" not found in amazonQPrompts.`)
+            return
+        }
+
+        if (!manifest.isManifestDeprecated) {
+            // In case we got an new url, make sure the prompt is re-enabled for active manifests
+            await prompts.enablePrompt(lspId as keyof typeof amazonQPrompts)
```
jpinkney-aws, line 122: you can probably just cast lspId once

jpinkney-aws (approved):

packages/core/src/shared/settings-amazonq.gen.ts
```diff
@@ -18,7 +18,9 @@ export const amazonqSettings = {
         "amazonQWelcomePage": {},
         "amazonQSessionConfigurationMessage": {},
         "minIdeVersion": {},
-        "ssoCacheError": {}
+        "ssoCacheError": {},
+        "AmazonQLspManifestMessage": {},
+        "AmazonQ-WorkspaceLspManifestMessage":{}
```
justinmk3, line 23: Why does this not follow the existing naming convention, which is found in the items just above this?

PR #6618, opened 2025-02-19
fix(amazonq): Adding target source file to the zip before traversing the repository

laileni-aws:
## Problem
- In `/test`, sometimes target source file is not being collected as part of zip and this is causing the backend to fail.

## Solution

- Adding target source file to the zip before traversing the repository.
---

### DIFF
--- a/packages/core/src/shared/utilities/workspaceUtils.ts
+++ b/packages/core/src/shared/utilities/workspaceUtils.ts
@@ -326,7 +326,8 @@ export async function collectFiles(
     workspaceFolders: CurrentWsFolders,
     respectGitIgnore: boolean = true,
     maxSize = 200 * 1024 * 1024, // 200 MB
-    defaultExcludePatterns: boolean = true
+    defaultExcludePatterns: boolean = true,
+    filePath?: string
 ): Promise<
     {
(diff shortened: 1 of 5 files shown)


packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -359,7 +361,28 @@ export async function collectFiles(
...
+                totalSizeBytes += fileStat.size
+                storage.push({
+                    workspaceFolder: relativePath.workspaceFolder,
+                    relativeFilePath: relativePath.relativePath,
+                    fileUri: targetUri,
+                    fileContent: fileContent,
+                    zipFilePath: prefixWithFolderPrefix(relativePath.workspaceFolder, relativePath.relativePath),
+                })
+            }
+        }
```
ctlai95, line 383: nit: can probably simplify this since you do the same thing in lines 402-429

ctlai95 (approved):

ashishrp-aws (approved):

packages/core/src/shared/utilities/workspaceUtils.ts
jpinkney-aws: Yes, please de-dup this code

PR #6621, opened 2025-02-19
fix(ec2): increase robustness of checking ssm agent ping status. 

Hweinstock:
## Problem
As part of the EC2 Connect process, we check if the SSM agent is pingable on the target instance. This is done here: https://github.com/aws/aws-toolkit-vscode/blob/b9af56c3097242fb796995479f864d95098bf713/packages/core/src/awsService/ec2/model.ts#L153-L162

This check can fail, and is currently does fail a decent amount of the time in telemetry.

## Solution
- wrap the check in a `waitUntil`, retrying every half second up to 10 times. 

---

### DIFF
--- a/packages/core/src/awsService/ec2/model.ts
+++ b/packages/core/src/awsService/ec2/model.ts
@@ -29,7 +29,7 @@ import {
     testSshConnection,
 } from '../../shared/extensions/ssh'
 import { getLogger } from '../../shared/logger/logger'
-import { CancellationError, Timeout } from '../../shared/utilities/timeoutUtils'
+import { CancellationError, Timeout, waitUntil } from '../../shared/utilities/timeoutUtils'
 import { showMessageWithCancel } from '../../shared/utilities/messages'
 import { SshConfig } from '../../shared/sshConfig'
 import { SshKeyPair } from './sshKeyPair'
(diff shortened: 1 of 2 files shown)


packages/core/src/awsService/ec2/model.ts
```diff
@@ -150,8 +150,14 @@ export class Ec2Connecter implements vscode.Disposable {
...
 
-    private async checkForInstanceSsmError(selection: Ec2Selection): Promise<void> {
-        const isSsmAgentRunning = (await this.ssmClient.getInstanceAgentPingStatus(selection.instanceId)) === 'Online'
+    public async checkForInstanceSsmError(
+        selection: Ec2Selection,
+        options?: Partial<{ interval: number; timeout: number }>
+    ): Promise<void> {
+        const isSsmAgentRunning = await waitUntil(
+            async () => (await this.ssmClient.getInstanceAgentPingStatus(selection.instanceId)) === 'Online',
+            { interval: options?.interval ?? 500, timeout: options?.timeout ?? 5000 }
```
jpinkney-aws, line 159: Did you look at telemetry to figure out these numbers or were these just numbers that kind of just make sense?

jpinkney-aws (approved):
> This check can fail, and is currently does fail a decent amount of the time in telemetry

Do we know why it fails? Does it just take a bit for the ssm agent to become available?

packages/core/src/awsService/ec2/model.ts
justinmk3: `waitUntil` has `backoff`, we should probably use that here.

justinmk3 (approved):

packages/core/src/awsService/ec2/model.ts
Hweinstock: I chose these somewhat arbitrarily. I didn't want to make the timeout too long since a long loading bar feels bad, but it is cancellable so perhaps I could extend it. 

Based on experience, the agent can take a while to startup and sometimes restarting it can make it faster. I still do not fully understand why. I've dug around a little bit before and found this is usually caused by permissions problems, but in our case we explicitly check for the correct permissions before checking the agent status. Usually if the ec2 instance has been running for more than ~10 minutes, the agent is also running. 

Its possible adding this retry is insignificant, but seems like an easy thing to try. Next steps would likely be reaching out to the relevant internal channels to see if someone can provide guidance.

PR #6624, opened 2025-02-19
refactor(ec2): improve error information in telemetry. 

Hweinstock:
## Problem
Within telemetry, there are a large group of fatal errors within EC2 Connect being grouped together under the same general error code. Additionally, there is very little information in the error message to determine the source of this error. Specifically, it appears the the test SSM connection done before attempting to pass the connection to VS Code, implemented [here](https://github.com/aws/aws-toolkit-vscode/blob/b9af56c3097242fb796995479f864d95098bf713/packages/core/src/shared/extensions/ssh.ts#L133-L155), can fail, and gives us a general error code with little information about its root cause.

## Solution
- Add a specific error code for this connection type. (reused an old unused one)
- Pipe the error from the child process into the error message so that we get it in telemetry. 
- Add a docstring to test connect function, since its not obvious whats happening as it involves both ssh and ssm. 

Example Error:
image

---

### DIFF
--- a/packages/core/src/shared/extensions/ssh.ts
+++ b/packages/core/src/shared/extensions/ssh.ts
@@ -130,7 +130,7 @@ export class RemoteSshSettings extends Settings.define('remote.SSH', remoteSshTy
     }
 }
 
-export async function testSshConnection(
+export async function testSsmConnection(
     ProcessClass: typeof ChildProcess,
     hostname: string,
     sshPath: string,
@@ -150,7 +150,7 @@ export async function testSshConnection(
         })
         return result
     } catch (error) {
-        throw new SshError('SSH connection test failed', { cause: error as Error })
+        throw new SshError('SSH connection test failed', { cause: error as Error, code: 'SSMTestConnectionFailed' })
     }
 }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/shared/extensions/ssh.ts
```diff
@@ -130,27 +130,34 @@ export class RemoteSshSettings extends Settings.define('remote.SSH', remoteSshTy
...
+        return await process.run({
             spawnOptions: {
                 env,
             },
         })
-        return result
     } catch (error) {
-        throw new SshError('SSH connection test failed', { cause: error as Error })
+        throw new SshError(process.result()?.stderr ?? 'An unknown error occurred when testing the connection', {
+            cause: error as Error,
```
jpinkney-aws, line 159: Will this error serialize to a string correctly? or do we need (error as Error).message?

justinmk3: `cause` field is always a `Error` subclass. The commands2.ts wrapper will get its fields for telemetry.

justinmk3 (approved):

packages/core/src/shared/extensions/ssh.ts
Hweinstock: It can be a little verbose, but it does serialize properly. I added an example to the PR description.

PR #6625, opened 2025-02-19
telemetry(auth): Update metrics to better debug Auth dropoff

nkomonen-amazon:
## Problem

We noticed that there was an auth dropoff between `session_start` with a brand new clientId, versus when the `auth_userState` metric indicated `isFirstUse` meaning the user is net new. We went from 12.7k for `session_start` to 9.8k for `auth_userState`, these should have been basically the same.

## Solution

Add in certain metrics to help debug where the discrepancy is coming from:

- When we determine the user is a first time user, we will also check if the clientId is newly generated. If this is not the case we know there is a discrepancy here
  - The relevant metric will be `function_call` with a `functionName: isFirstUse`, `result: Failed`, and a `reason: ClientIdAlreadyExisted`
- When we determine the user is a first time user, if we detected that they had previous auth connections, this will indicate a likely cause for the discrepancy
  - The relevant metric will be `function_call` with `reason: UnexpectedConnections`
- We will emit metrics when the Auth Login page loads since that also had a discrepancy and the telemetry did not exist
  - The relevant metric is `webview_load` and it will indicate when the Auth Login/Reauth page has actually loaded
  - Previously we were observing the telemetry for the command `aws.amazonq.focusChat`, but all this did was emit when called and didn't confirm the UI actually loaded.
- We will also add the `isFirstUse` metric source value in to some other existing metrics

---

### DIFF
--- a/packages/core/src/auth/utils.ts
+++ b/packages/core/src/auth/utils.ts
@@ -709,6 +709,13 @@ export class ExtensionUse {
 
         if (isAmazonQ()) {
             this.isFirstUseCurrentSession = true
+            if (hasExistingConnections()) {
+                telemetry.function_call.emit({
+                    result: 'Failed',
+                    functionName: 'isFirstUse',
+                    reason: 'UnexpectedConnections',
+                })
+            }
         } else {
             // The variable in the store is not defined yet, fallback to checking if they have existing connections.
             this.isFirstUseCurrentSession = !hasExistingConnections()
--- a/packages/core/src/auth/utils.ts
+++ b/packages/core/src/auth/utils.ts
@@ -58,7 +58,7 @@ import { EcsCredentialsProvider } from './providers/ecsCredentialsProvider'
 import { EnvVarsCredentialsProvider } from './providers/envVarsCredentialsProvider'
 import { showMessageWithUrl } from '../shared/utilities/messages'
 import { credentialHelpUrl } from '../shared/constants'
-import { ExtStartUpSources, ExtStartUpSource } from '../shared/telemetry/util'
+import { ExtStartUpSources, ExtStartUpSource, hadClientIdOnStartup } from '../shared/telemetry/util'
 
 // iam-only excludes Builder ID and IAM Identity Center from the list of valid connections
 // TODO: Understand if "iam" should include these from the list at all
@@ -693,6 +693,18 @@ export class ExtensionUse {
         if (isFirstUse === false) {
             this.isFirstUseCurrentSession = isFirstUse
             return this.isFirstUseCurrentSession
+        }
+
+        /**
+         * SANITY CHECK: If the clientId already existed on startup, then isFirstUse MUST be false. So
+         * there is a bug in the state.
+         */
+        if (hadClientIdOnStartup(globals.globalState)) {
+            telemetry.function_call.emit({
+                result: 'Failed',
+                functionName: 'isFirstUse',
+                reason: 'ClientIdAlreadyExisted',
+            })
         }
 
         if (isAmazonQ()) {
--- a/packages/core/src/auth/utils.ts
+++ b/packages/core/src/auth/utils.ts
@@ -688,17 +688,24 @@ export class ExtensionUse {
             return this.isFirstUseCurrentSession
         }
 
-        this.isFirstUseCurrentSession = globals.globalState.get('isExtensionFirstUse')
-        if (this.isFirstUseCurrentSession === undefined) {
+        // This is for sure not their first use
+        const isFirstUse = globals.globalState.tryGet('isExtensionFirstUse', Boolean)
+        if (isFirstUse === false) {
+            this.isFirstUseCurrentSession = isFirstUse
+            return this.isFirstUseCurrentSession
+        }
+
+        if (isAmazonQ()) {
+            this.isFirstUseCurrentSession = true
+        } else {
             // The variable in the store is not defined yet, fallback to checking if they have existing connections.
             this.isFirstUseCurrentSession = !hasExistingConnections()
-
-            getLogger().debug(
-                `isFirstUse: State not found, marking user as '${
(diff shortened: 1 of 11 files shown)


packages/core/src/login/webview/vue/backend.ts
```diff
@@ -61,6 +61,22 @@ export abstract class CommonAuthWebview extends VueWebview {
         return globals.regionProvider.getRegions().reverse()
     }
 
+    private didCall: { login: boolean; reauth: boolean } = { login: false, reauth: false }
+    public setUiReady(state: 'login' | 'reauth') {
```
jpinkney-aws, line 65: is this something we could just use the once() util for? That way you don't have to keep track of the extra state?

packages/core/src/test/credentials/utils.test.ts
```diff
@@ -11,23 +11,23 @@ import globals from '../../shared/extensionGlobals'
 
 describe('ExtensionUse.isFirstUse()', function () {
     let instance: ExtensionUse
+    const notHasExistingConnections = () => false
```
jpinkney-aws, line 14: should this be `hasNoExistingConnections` ?

packages/core/src/auth/utils.ts
```diff
@@ -688,17 +688,43 @@ export class ExtensionUse {
...
+        if (hadClientIdOnStartup(globals.globalState)) {
+            telemetry.function_call.emit({
+                result: 'Failed',
+                functionName: 'isFirstUse',
+                reason: 'ClientIdAlreadyExisted',
+            })
+        }
+
+        if (isAmazonQ()) {
+            this.isFirstUseCurrentSession = true
```
jpinkney-aws, line 711: qq: why is this always true for amazonQ?

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1207,6 +1218,17 @@
                 }
             ]
         },
+        {
+            "name": "session_start",
```
justinmk3, line 1222: Why is this needed? There is already a metric for starting the plugin.

packages/core/src/login/webview/vue/backend.ts
nkomonen-amazon: I wanted to emit once for `login` and `reauth` separately which didn't allow me to use `once()`. But I think in a future PR we could do some memoize + once to address this

packages/core/src/shared/telemetry/vscodeTelemetry.json
nkomonen-amazon: What is the metric?

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1131,6 +1131,17 @@
             ],
             "passive": true
         },
+        {
+            "name": "auth_signInPageOpened",
+            "description": "When the Amazon Q sign in page is opened and focused.",
+            "metadata": [
+                {
+                    "type": "source",
```
justinmk3, line 1139: why not add this to the existing metric? https://github.com/aws/aws-toolkit-common/blob/ccb16ca4f73b07c4e1cd79da159312c9ffe403a7/telemetry/definitions/commonDefinitions.json#L2915

packages/core/src/auth/utils.ts
nkomonen-amazon: There was previous logic where if we detected an auth connection in Toolkit that we would not consider it a first time user. But this doesn't make sense for Q.

Also if they get the far in the function it is guaranteed they are a first time user, since they would have otherwise returned early.

packages/core/src/shared/telemetry/vscodeTelemetry.json
nkomonen-amazon: I'm being rushed to get the change out for this release, but I will port this over after

packages/core/src/login/webview/commonAuthViewProvider.ts
```diff
@@ -83,14 +84,22 @@ export class CommonAuthViewProvider implements WebviewViewProvider {
     ) {
         // Our callback won't fire on the first view.
         if (webviewView.visible) {
-            telemetry.auth_signInPageOpened.emit({ result: 'Succeeded', passive: true })
+            telemetry.auth_signInPageOpened.emit({
```
jpinkney-aws, line 87: Can we modify this metric and make it more general in the future?

packages/core/src/test/credentials/utils.test.ts
nkomonen-amazon: Makes sense, I'll follow up with the name change after. Don't want to have to run CI again unless theres something critical.

packages/core/src/login/webview/commonAuthViewProvider.ts
nkomonen-amazon: Agreed, we need to better structure out telemetry since it is messy with what "opened" means. We can also collapse all webviews in to a single metric probably

packages/core/src/shared/telemetry/vscodeTelemetry.json
nkomonen-amazon: Yeah, I plan to do this. It was just due to time constraints. I'll follow up after this PR

jpinkney-aws (approved):
Given the time constraints we're under for testing the amazon q changes, I don't see anything majorly blocking. I think we definitely need follow up PRs for things like porting back to telemetry though.

Also, is some of this code going to live on after us finding the problem or will we scrap it afterwords?

nkomonen-amazon:
@jpinkney-aws 

Some of these metrics will be removed (mainly the `function_call` ones to help debug). But the others like `session_start` and `auth_signInPageOpened` will be ported. Though I think we may need to look at adding something like `source` as a global field instead since we have so many cases for it.

I will look in to a fast follow-up.

PR #6628, opened 2025-02-19
fix(amazonq): Updating logic for getWorkspaceRelativePath in shared src for /test

laileni-aws:
## Problem
- This will fix `Target file doesn't exist or is a directory.` issue in unit test generation.
- If user opens project in workspace mode and try `/test` then in shared workspaceUtils: https://github.com/aws/aws-toolkit-vscode/blob/35502be238b1bf7c3ff73e44df8d077a6d32aa85/packages/core/src/shared/utilities/workspaceUtils.ts#L220-L236 returns the files starting from src but UTG does need values from child level.
## Solution
Given above function with workspaceFolders as 
```
[
  {
    uri: {
      scheme: "file",
      authority: "",
      path: "/Users/Q/Downloads/TestingUTG/Sample/src",
      query: "",
      fragment: "",
      _formatted: "file:///Users/Q/Downloads/TestingUTG/Sample/src",
      _fsPath: "/Users/Q/Downloads/TestingUTG/Sample/src",
    },
    name: "src",
    index: 0,
  },
  {
    uri: {
      scheme: "file",
      authority: "",
      path: "/Users/laileni/Downloads/TestingUTG/Sample/src/Sample",
      query: "",
      fragment: "",
      _formatted: "file:///Users/Q/Downloads/TestingUTG/Sample/src/Sample",
      _fsPath: "/Users/Q/Downloads/TestingUTG/Sample/src/Sample",
    },
    name: "📦 Sample",
    index: 1,
  },
]
```

User requires the project/folder with path: "/Users/Q/Downloads/TestingUTG/Sample/src/Sample" but right now this gives the path from  "/Users/Q/Downloads/TestingUTG/Sample/src" which incorrectly collects the files and this is causing failures in Unit test generation.

Example:
Expected Payload is 
```
Sample 
    - File A
    - Folder B 
    etc...
 ```
 but actual payload is 
 ```
 Sample 
    - src
       -- Sample
           --- File A
           --- Folder B 
 ```
    

-  Sort workspace folders by path length (descending) to prioritize deeper paths
### TODO: 
- Need to enable this for entire Q

---

### DIFF
--- a/packages/core/src/shared/utilities/workspaceUtils.ts
+++ b/packages/core/src/shared/utilities/workspaceUtils.ts
@@ -20,6 +20,7 @@ import fs from '../fs/fs'
 import { ChildProcess } from './processUtils'
 import { isWin } from '../vscode/env'
 import { maxRepoSizeBytes } from '../../amazonqFeatureDev/constants'
+import { FeatureUseCase } from '../../codewhisperer/models/constants'
 
 type GitIgnoreRelativeAcceptor = {
     folderPath: string
@@ -223,12 +224,20 @@ export function getWorkspaceRelativePath(
         workspaceFolders?: readonly vscode.WorkspaceFolder[]
     } = {
         workspaceFolders: vscode.workspace.workspaceFolders,
-    }
+    },
+    useCase?: FeatureUseCase
 ): { relativePath: string; workspaceFolder: vscode.WorkspaceFolder } | undefined {
     if (!override.workspaceFolders) {
         return
     }
-    for (const folder of override.workspaceFolders) {
+    let folders = override.workspaceFolders
+    if (useCase && useCase === FeatureUseCase.TEST_GENERATION) {
+        // Sort workspace folders by path length (descending) to prioritize deeper paths
+        // TODO: Need to enable this for entire Q
+        folders = [...override.workspaceFolders].sort((a, b) => b.uri.fsPath.length - a.uri.fsPath.length)
+    }
+
+    for (const folder of folders) {
         if (isInDirectory(folder.uri.fsPath, childPath)) {
             return { relativePath: path.relative(folder.uri.fsPath, childPath), workspaceFolder: folder }
         }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by ashishrp-aws — APPROVED** [review]

**Comment by ashishrp-aws** [comment]
Discussed the PR offline to check the conditions, this fix works for test generation.

File: packages/core/src/shared/utilities/workspaceUtils.ts [hunk 1]
```diff
@@ -223,12 +224,20 @@ export function getWorkspaceRelativePath(
...
-    }
+    },
+    useCase?: FeatureUseCase
 ): { relativePath: string; workspaceFolder: vscode.WorkspaceFolder } | undefined {
     if (!override.workspaceFolders) {
         return
     }
-    for (const folder of override.workspaceFolders) {
+    let folders = override.workspaceFolders
+    if (useCase && useCase === FeatureUseCase.TEST_GENERATION) {
```
ctlai95, line 234: This feels too specific to be put in this shared utility method. Are you planning to update this later?

laileni-aws: Yes Will remove this once I get confirmation from Q teams 
For now this is required for UTG. I think this is the same case for other features too

nkomonen-amazon, line 234: Can just do

```suggestion
    if (useCase === FeatureUseCase.TEST_GENERATION) {
```

Randall-Jiang (approved):

nkomonen-amazon (approved):

packages/core/src/shared/utilities/workspaceUtils.ts
jpinkney-aws: Can you track this in an issue somewhere so that someone comes back to it. This workspace util is become a monster 😞

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -398,7 +408,7 @@ export async function collectFiles(
         const files = excludeByGitIgnore ? await filterOutGitignoredFiles(rootPath, allFiles, false) : allFiles
 
         for (const file of files) {
-            const relativePath = getWorkspaceRelativePath(file.fsPath, { workspaceFolders })
+            const relativePath = getWorkspaceRelativePath(file.fsPath, { workspaceFolders }, useCase)
```
justinmk3, line 411: This is already `filterFn`, why was this special parameter needed?

PR #6640, opened 2025-02-20
refactor(util): allow tryRun to take a regex. 

Hweinstock:
## Problem
This is aimed at fixing the following ec2 bug.

Attempting to generate the ssh keys via tryRun can log a failure, when there is no failure. Looking at https://github.com/aws/aws-toolkit-vscode/blob/35502be238b1bf7c3ff73e44df8d077a6d32aa85/packages/core/src/awsService/ec2/sshKeyPair.ts#L78-L86 

We use `tryRun` to detect if the output contains "unknown key type". This allows us to detect when a certain key type such as RSA or ed25519 is not supported on the local machine. Detecting if the key generated successfully is a harder problem, as the output did not contain a consistent message. 

However, it results in `tryRun` logging a failure to find "unknown key type" in the output, which means failed to find the expected text in the output, but still generated the key. This is then logged as a failure, but is functionally a success. This is very confusing behavior. 

The root of this problem is we are trying to use `tryRun` in a way it doesn't naturally support. Rather than use it awkwardly, we can extend it. 

## Solution
- add support for regex in tryRun

New logs:

```
2025-02-20 12:07:01.624 [info] tryRun: ok: PID 22605: [ssh-keygen -t ed25519 -N  -q -f /Users/hkobew/Library/Application Support/Code/User/globalStorage/amazonwebservices.aws-toolkit-vscode/aws-ec2-key] {
  exitCode: 0,
  stdout: '/Users/hkobew/Library/Application Support/Code/User/globalStorage/amazonwebservices.aws-toolkit-vscode/aws-ec2-key already exists.\n' +
    'Overwrite (y/n)?',
  stderr: '',
  error: undefined,
  signal: undefined
}
```

---

### DIFF
--- a/packages/core/src/awsService/ec2/sshKeyPair.ts
+++ b/packages/core/src/awsService/ec2/sshKeyPair.ts
@@ -79,10 +79,16 @@ export class SshKeyPair {
         const overrideKeys = async (_t: string, proc: RunParameterContext) => {
             await proc.send('yes')
         }
-        return !(await tryRun('ssh-keygen', ['-t', keyType, '-N', '', '-q', '-f', keyPath], 'yes', 'unknown key type', {
-            onStdout: overrideKeys,
-            timeout: new Timeout(5000),
-        }))
+        return await tryRun(
+            'ssh-keygen',
+            ['-t', keyType, '-N', '', '-q', '-f', keyPath],
+            'yes',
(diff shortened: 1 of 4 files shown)


justinmk3 (approved):
Nice!

jpinkney-aws (approved):

PR #6645, opened 2025-02-20
refactor(sdkv3): cache clients on creation.

Hweinstock:
## Problem
The toolkit recreates each SDK client before each request is made. As an example see: https://github.com/aws/aws-toolkit-vscode/blob/b4d7417ece1eebc56a5284a9cdd643e4b2308bf4/packages/core/src/shared/clients/ec2Client.ts#L31-L52

This is done as an unnecessary guard against the client's internal state interfering with higher level business logic. Creating clients for each request is not cheap, and there is likely a better pattern to be had here. 

See internal docs expanding more on this issue. 

## Solution
- Create a map that caches the clients.
- Key the function arguments such that it detects config changes. 
- Cache clients with new method `getAwsService` that wraps `createAwsService`. 
- Allow consumers to pass a `ignoreCache` flag to force creation of a new client.
- On deactivation, call `destroy` on all the clients.

### Testing Strategy
- Create clients with methods that make it easy to determine equality.  

---


packages/core/src/shared/utilities/functionUtils.ts
```diff
@@ -77,6 +77,20 @@ export function memoize<T, U extends any[]>(fn: (...args: U) => T): (...args: U)
     return (...args) => (cache[args.map(String).join(':')] ??= fn(...args))
 }
 
+/**
+ * Generalization of the {@link memoize} method that accepts async methods, and allows user to pass mapping from keys to args.
```
justinmk3, line 81: can we extend the existing memoize()? also probably want direct test coverage of this

packages/core/src/shared/awsClientBuilderV3.ts
```diff
@@ -77,6 +78,25 @@ export class AWSClientBuilderV3 {
...
+        return [
+            String(type),
+            JSON.stringify(options),
+            region,
+            userAgent ? '1' : '0',
+            settings ? JSON.stringify(settings.get('endpoints', {})) : '',
+        ].join(':')
+    }
+
+    public getAwsService = memoizeWith(this.createAwsService.bind(this), this.keyAwsService.bind(this))
```
justinmk3, line 98: memoize may not be gaining much for this use-case. 

because this is an important core module, we may want more visibility into the long-lived container that stores the clients. this module could store them in a map, that could be inspected and/or logged.

packages/core/src/shared/utilities/functionUtils.ts
Hweinstock: I was trying to maintain its type signature. Making this one return an async method would cause a lot of unrelated changes. However, no longer relevant since I refactored away from this.

packages/core/src/shared/awsClientBuilderV3.ts
Hweinstock: That makes sense, this will make it easier to add custom-logic later.

jpinkney-aws (approved):

packages/core/src/shared/utilities/functionUtils.ts
justinmk3: > Making this one return an async method

would an "override" work?

packages/core/src/shared/awsClientBuilderV3.ts
```diff
@@ -78,19 +88,37 @@ export class AWSClientBuilderV3 {
...
-        region?: string,
-        userAgent: boolean = true,
-        keepAlive: boolean = true,
-        settings?: DevSettings
-    ): Promise<C> {
+    private keyAwsService<C extends AwsClient>(serviceOptions: AwsServiceOptions<C>): string {
+        // Serializing certain objects in the args allows us to detect when nested objects change (ex. new retry strategy, endpoints)
+        return [
+            String(serviceOptions.serviceClient),
+            JSON.stringify(serviceOptions.clientOptions),
```
justinmk3, line 95: Is JSON.stringify guaranteed to return a stable ordering of keys for the same input? I think we have some "hash" code, or some other way of dealing with that, somewhere...

packages/core/src/shared/awsClientBuilderV3.ts
```diff
@@ -78,19 +88,37 @@ export class AWSClientBuilderV3 {
...
+        if (cached) {
+            return cached as C
+        }
+
+        const service = await this.createAwsService(serviceOptions)
+        this.serviceCache.set(key, service)
+        return service as C
+    }
+
+    public async createAwsService<C extends AwsClient>(serviceOptions: AwsServiceOptions<C>): Promise<C> {
```
justinmk3, line 114: should the docstring call out the fact that the client will be re-used?

important for callers not to "modify" the client, except in specific "safe" ways...

PR #6654, opened 2025-02-21
fix(amazonq): fix uploading file method error handling for /doc

KevinDing1:
## Problem
File upload method threw incorrect error message when workspace too large for /doc

## Solution

Fix uploading file method error handling for /doc
---

### DIFF
--- a/packages/core/src/amazonq/util/files.ts
+++ b/packages/core/src/amazonq/util/files.ts
@@ -13,6 +13,7 @@ import {
 } from '../../shared/utilities/workspaceUtils'
 
 import { ContentLengthError, PrepareRepoFailedError } from '../../amazonqFeatureDev/errors'
+import { ContentLengthError as DocContentLengthError } from '../../amazonqDoc/errors'
 import { getLogger } from '../../shared/logger/logger'
 import { maxFileSizeBytes } from '../../amazonqFeatureDev/limits'
 import { CurrentWsFolders, DeletedFileInfo, NewFileInfo, NewFileZipContents } from '../../amazonqDoc/types'
@@ -48,6 +49,7 @@ export type PrepareRepoDataOptions = {
     telemetry?: TelemetryHelper
     zip?: ZipStream
     isIncludeInfraDiagram?: boolean
+    featureName?: 'featureDev' | 'docGeneration'
 }
 
 /**
@@ -186,6 +188,9 @@ export async function prepareRepoData(
     } catch (error) {
         getLogger().debug(`featureDev: Failed to prepare repo: ${error}`)
         if (error instanceof ToolkitError && error.code === 'ContentLengthError') {
+            if (options?.featureName === 'docGeneration') {
+                throw new DocContentLengthError()
+            }
             throw new ContentLengthError()
         }
         throw new PrepareRepoFailedError()
--- a/packages/amazonq/.changes/next-release/Bug Fix-6b8cf110-2050-4ac1-8a3b-637ed54980cf.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-6b8cf110-2050-4ac1-8a3b-637ed54980cf.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Amazon Q /doc: Fix uploading file method throwing incorrect workspace too large error message"
+}
--- a/packages/core/src/amazonqDoc/session/sessionState.ts
+++ b/packages/core/src/amazonqDoc/session/sessionState.ts
@@ -144,6 +144,7 @@ export class DocPrepareCodeGenState extends BasePrepareCodeGenState {
         return await prepareRepoData(workspaceRoots, workspaceFolders, span, {
             ...options,
             isIncludeInfraDiagram: true,
+            featureName: 'docGeneration',
         })
     }
 }


avi-alpert (approved):

packages/core/src/amazonq/util/files.ts
```diff
@@ -186,6 +188,9 @@ export async function prepareRepoData(
     } catch (error) {
         getLogger().debug(`featureDev: Failed to prepare repo: ${error}`)
         if (error instanceof ToolkitError && error.code === 'ContentLengthError') {
+            if (options?.featureName === 'docGeneration') {
```
jpinkney-aws, line 191: is there any way we can avoid making this function depend on the featureName? Can the doc controller just handle ContentLengthError instead of needing to deal with a DocContentLengthError?

KevinDing1: Doc controller already handles this type error when code generation result comes back. But this is the uploading file method we shared with /dev, so have to pass feature name to get the correct error message

jpinkney-aws: Makes sense. Code wise though, we don't want to be drilling feature specific toggles in this if we don't have to. Can a ContentLengthError from feature dev and doc generation be the same so that we can just throw the ContentLengthError?

KevinDing1: Unfortunately we can't. Our service requiring user upload the whole folder or workspace as the context. So even user select a smaller folder in a giant repo, they will still get the error. And the intention for this fix, is the current /dev ContentLengthError is misleading for /doc user

jpinkney-aws: Right, but right now we're throwing feature dev errors in this function and then now we're going to do if/else statements to then throw different errors specifically for doc. What we should probably do instead is throw a common amazon q content length error and then both the feature dev team and the doc team handle that in their own way

E.g.
```typescript
function prepareRepoData() {
    ....
    throw new CommonAmazonQContentLengthError()
    ....
}
```

in feature dev controller
```typescript
try {
    ...
} catch (e) {
    if (e instanceof CommonAmazonQContentLengthError) {
        throw new FeatureDevContentLengthError()
    }
}
```

in doc controller
```typescript
try {
    ...
} catch (e) {
    if (e instanceof CommonAmazonQContentLengthError) {
        throw new DocContentLengthError()
    }
}
```

it doesn't really make sense to have common code that is forced to being uncommon by featureName flags IMO

viktorsaws (approved):

packages/core/src/amazonq/util/files.ts
KevinDing1: Got it, thanks for your insight!

packages/core/src/amazonqDoc/session/session.ts
```diff
@@ -126,28 +128,35 @@ export class Session {
...
-                this.state?.tokenSource?.cancel()
+                // Move to the next state
+                this._state = resp.nextState
             }
 
-            // Move to the next state
-            this._state = resp.nextState
+            return resp.interaction
+        } catch (e) {
+            if (e instanceof FeatureDevContentLengthError) {
```
jpinkney-aws, line 155: Instead of throwing FeatureDevContentLengthError, I'd make the minor change to introduce a new one and then have prepareRepoData throw that one (the commons approach I talked about earlier). That way your amazonqDoc folder can depend on an common amazon q errors, rather than having to import feature devs

KevinDing1: I thought this is the minimum change to solve the issue, as if throw the common error in the `prepareRepoData` method, then featureDev also need to catch it and re-throw with their error message. 

The `featureDevContentLenghtError`  is the same as https://github.com/aws/aws-toolkit-vscode/blob/44212902361317193c6c7e2873a86cb2deb30735/packages/core/src/amazonq/util/files.ts#L189
so I don't have to change the shared session logic and just change our's

packages/core/src/amazonq/errors.ts
```diff
@@ -0,0 +1,13 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+import { ToolkitError } from '../shared/errors'
+
+export class CommonAmazonQContentLengthError extends ToolkitError {
```
jpinkney-aws, line 9: You can just call this
```suggestion
export class ContentLengthError extends ToolkitError {
```

I'd probably add a comment that's something like (you can improve the words):
```
/**
 * A shared error type for content length validation that can
 * be caught and processed by different agent implementations
 */
```

packages/core/src/amazonq/errors.ts
```diff
@@ -0,0 +1,13 @@
...
+    constructor(message: string) {
+        super(message, { code: 'CommonAmazonQContentLengthError' })
```
jpinkney-aws, line 11: Having this be the code doesn't convey any extra information
```suggestion
        super(message, { code: 'ContentLengthError' })
```

packages/core/src/amazonq/util/files.ts
```diff
@@ -186,7 +187,7 @@ export async function prepareRepoData(
     } catch (error) {
         getLogger().debug(`featureDev: Failed to prepare repo: ${error}`)
         if (error instanceof ToolkitError && error.code === 'ContentLengthError') {
-            throw new ContentLengthError()
+            throw new CommonAmazonQContentLengthError(error.message)
```
jpinkney-aws, line 190: You don't need to pass in the error message if you don't end up using it anywhere

avi-alpert (approved):

packages/core/src/amazonq/errors.ts
KevinDing1: Thanks!

jpinkney-aws (approved):

PR #6658, opened 2025-02-21
fix(amazonq): skip tests locally

dhasani23:
## Problem

When users choose to skip tests, we were skipping them on our server by passing in their choice to our backend, but we weren't also skipping them locally.

## Solution

Tell Maven to use `-DskipTests`

---

### DIFF
--- a/packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
+++ b/packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
@@ -125,12 +125,12 @@ export class Messenger {
             mandatory: true,
             options: [
                 {
-                    value: CodeWhispererConstants.runUnitTestsMessage,
-                    label: CodeWhispererConstants.runUnitTestsMessage,
-                },
-                {
                     value: CodeWhispererConstants.skipUnitTestsMessage,
                     label: CodeWhispererConstants.skipUnitTestsMessage,
+                },
+                {
+                    value: CodeWhispererConstants.runUnitTestsMessage,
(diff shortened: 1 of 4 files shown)


packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -124,14 +124,14 @@ export class Messenger {
...
-                    label: CodeWhispererConstants.runUnitTestsMessage,
-                },
                 {
                     value: CodeWhispererConstants.skipUnitTestsMessage,
                     label: CodeWhispererConstants.skipUnitTestsMessage,
                 },
+                {
+                    value: CodeWhispererConstants.runUnitTestsMessage,
+                    label: CodeWhispererConstants.runUnitTestsMessage,
+                },
```
dhasani23, line 134: Just switch the order so that "skip tests" is default option in the form item dropdown.

packages/core/src/amazonqGumby/chat/controller/messenger/messenger.ts
```diff
@@ -533,7 +533,7 @@ export class Messenger {
             })
         }
 
-        if (transformByQState.getSummaryFilePath()) {
+        if (transformByQState.isPartiallySucceeded() || transformByQState.isSucceeded()) {
```
dhasani23, line 536: We should be checking if job is completed (or partially completed) to show the summary button since the summary file path could be left over from the last job; updated relevant E2E test above too.

ntarakad-aws (approved):

jpinkney-aws (approved):

PR #6659, opened 2025-02-21
fix(amazonq):  ignored lines should not show up in /review scan issues

zuoyaofu:
## Problem
When running a new `/review` scan, Q fails to respect previously added ignore line, causing ignored issues to reappear in subsequent reviews.

## Solution
Fixed incorrect filter logic on ignoring single line issue
Old unit tests for `listScanResults` will not work after this change. So in addition, I improved all of `securityScanHandler`'s unit tests to not use any file stubs to better simulate real environment. 

---

### DIFF
--- a/packages/amazonq/test/unit/codewhisperer/service/securityScanHandler.test.ts
+++ b/packages/amazonq/test/unit/codewhisperer/service/securityScanHandler.test.ts
@@ -22,10 +22,8 @@ import sinon from 'sinon'
 import * as vscode from 'vscode'
 import path from 'path'
 
-const buildRawCodeScanIssue = (fromProject: boolean = true, params?: Partial<RawCodeScanIssue>): RawCodeScanIssue => ({
-    filePath: fromProject
-        ? 'workspaceFolder/python3.7-plain-sam-app/hello_world/app.py'
-        : path.join(getWorkspaceFolder().substring(1), '/python3.7-plain-sam-app/hello_world/app.py'),
+const buildRawCodeScanIssue = (params?: Partial<RawCodeScanIssue>): RawCodeScanIssue => ({
+    filePath: 'workspaceFolder/python3.7-plain-sam-app/hello_world/app.py',
     startLine: 1,
     endLine: 1,
     title: 'title',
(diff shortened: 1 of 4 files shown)


packages/core/src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -82,18 +82,20 @@ export async function listScanResults(
             // Do not use .. in between because there could be multiple project paths in the same parent dir.
             const filePath = path.join(projectPath, key.split('/').slice(1).join('/'))
             if (existsSync(filePath) && statSync(filePath).isFile()) {
+                const document = await vscode.workspace.openTextDocument(filePath)
                 const aggregatedCodeScanIssue: AggregatedCodeScanIssue = {
                     filePath: filePath,
-                    issues: issues.map((issue) => mapRawToCodeScanIssue(issue, editor, jobId, scope)),
+                    issues: issues.map((issue) => mapRawToCodeScanIssue(issue, jobId, scope, document)),
```
ctlai95, line 88: nit: you can make this change even smaller by keeping the position of the editor/document argument the same

packages/amazonq/test/unit/codewhisperer/service/securityScanHandler.test.ts
```diff
@@ -65,17 +65,15 @@ const buildMockListCodeScanFindingsResponse = (
     nextToken: nextToken ? 'nextToken' : undefined,
 })
 
+function getWorkspaceFolder(): string {
+    return path.join(__dirname, '../../../../../../core/dist/src/testFixtures/workspaceFolder')
```
ctlai95, line 69: Can we use the src path instead of the dist path?

packages/amazonq/test/unit/codewhisperer/service/securityScanHandler.test.ts
```diff
@@ -65,17 +65,15 @@ const buildMockListCodeScanFindingsResponse = (
     nextToken: nextToken ? 'nextToken' : undefined,
 })
 
+function getWorkspaceFolder(): string {
```
jpinkney-aws, line 68: can we use something like `vscode.workspace.workspaceFolders[0].uri` instead ? Aren't we already in the workspace folder?

zuoyaofu: Fixed.

packages/amazonq/test/unit/codewhisperer/service/securityScanHandler.test.ts
```diff
@@ -65,17 +65,18 @@ const buildMockListCodeScanFindingsResponse = (
...
+    if (vscode.workspace.workspaceFolders !== undefined && vscode.workspace.workspaceFolders.length > 0) {
+        return vscode.workspace.workspaceFolders[0].uri.toString().split('file://')[1]
+    }
+    return path.join(__dirname, '../../../../../../core/src/testFixtures/workspaceFolder')
```
ctlai95, line 72:
```suggestion
    return vscode.workspace.workspaceFolders?.[0]?.uri.fsPath ?? path.join(__dirname, '../../../../../../core/src/testFixtures/workspaceFolder')
```

ctlai95 (approved):

PR #6664, opened 2025-02-24
feat(sdkv3): enable persistent connections by forcing keep alive header. 

Hweinstock:
## Problem
Inspecting the logs, and http requests themselves, it appears that the sdkv3 clients are somehow adding a `connection: close` header to the http requests. This is not done by the SDK (according to SDK team), and inspecting the headers throughout the SDK packages at the recommended points suggests this is being overwritten somewhere else. 

Creating a new vscode extension with this code snippet also fails, but a new project without vscode succeeds. This suggests its VS Code. 

I believe this is caused by https://github.com/microsoft/vscode/issues/173861, which was resolved, but is now reopened. The comments and techdebt tests suggested this is no longer an issue, but it now is again due to fix being reverted. 

## Solution
- Remove outdated notes/test that suggest this is fixed in newer versions of VS Code. 
- add e2e test verifying expected `keepAlive` behavior. 
- Add custom "build" step middleware that adds the keepAlive to the HTTP request, unless explicitly told not to. 
- Improve some typing in the middleware by making use of type assertions. 

### Notes
Related work done to completely overwrite the proxyAgent in vscode to fix this done here: https://github.com/aws/aws-toolkit-vscode-staging/pull/1214. However, this code is outdated and doesn't work anymore (its also unused). There is a functioning implementation of this logic at https://github.com/sourcegraph/cody/blob/62d73f78c432036d1f99bc9631ed534cc2ed846b/vscode/src/net/net.patch.ts. 

However, adding the `keepAlive` header manually appears to be working based on the e2e test. My understanding is that if the client was closing the connection in between each request, it must send a FIN message to the server to let it know according to TCP. The server would then close its side of the connection as well, but we know from the e2e test this isn't happening since it maintains the same connection across two requests. Therefore, the client must also be keeping its connection open, making use of persistent connections on both sides. 

### Experimental Results
Redoing the experiment of timing repeated requests when caching, not-caching, and caching with the change in this PR (connection persistence):
image
Note: experiment separate into 3 trails of 10 requests to limit the effect of throttling.

Therefore, this change is saving almost 40 ms per request!

## Future Work
Delete the dead code at https://github.com/aws/aws-toolkit-vscode-staging/pull/1214. 

---

### DIFF
--- a/packages/core/src/test/techdebt.test.ts
+++ b/packages/core/src/test/techdebt.test.ts
@@ -17,6 +17,16 @@ describe('tech debt', function () {
         assert.ok(now <= cutoffDate, msg)
     }
 
+    it('vscode minimum version', async function () {
+        const minVscode = env.getMinVscodeVersion()
+        assert.ok(semver.lt(minVscode, '1.84.0'))
+
+        assert.ok(
+            semver.lt(minVscode, '1.110.0'),
+            'Check to see if https://github.com/microsoft/vscode/issues/173861 is resolved. Allows us to remove work done by https://github.com/aws/aws-toolkit-vscode-staging/pull/1214 and part of https://github.com/aws/aws-toolkit-vscode/pull/6664'
+        )
+    })
+
     it('nodejs minimum version', async function () {
         const minNodejs = env.getMinNodejsVersion()
 
--- a/packages/core/src/test/techdebt.test.ts
+++ b/packages/core/src/test/techdebt.test.ts
@@ -17,17 +17,6 @@ describe('tech debt', function () {
         assert.ok(now <= cutoffDate, msg)
     }
 
-    it('vscode minimum version', async function () {
-        const minVscode = env.getMinVscodeVersion()
-        assert.ok(semver.lt(minVscode, '1.84.0'))
-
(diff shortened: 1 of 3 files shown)


jpinkney-aws (approved):

hayemaxi (approved):
That's really cool. Nice!

packages/core/src/test/techdebt.test.ts
```diff
@@ -17,17 +17,6 @@ describe('tech debt', function () {
         assert.ok(now <= cutoffDate, msg)
     }
 
-    it('vscode minimum version', async function () {
-        const minVscode = env.getMinVscodeVersion()
-        assert.ok(semver.lt(minVscode, '1.84.0'))
-
-        // see https://github.com/microsoft/vscode/issues/173861
-        assert.ok(
-            semver.lt(minVscode, '1.93.0'),
```
justinmk3, line 26: unless we have some other way of detecting that 173861 is fixed, we should probably keep this test around as a reminder to check 173861. We can set the version here to 1.110, then "snooze" it to 1.130, etc.

justinmk3 (approved):
Excellent analysis + explanation

justinmk3 (approved):

PR #6667, opened 2025-02-24
fix(stepfunctions): Bug fixes - handling unsupported messages, fix undo sync bug, code cleanup

witness-me:
## Problem
1. In case we introduce new feature dependant on the Toolkit change in the future, users may face issues if they don't update their toolkit right away (as their toolkit version won't have the new functionality in place)
2. There was a regression from my [previous PR](https://github.com/aws/aws-toolkit-vscode/pull/6024/commits/9c40f4c982c7d6a02c064b6dea62120f3760aa71), in `package.json` - most possibly due to incorrect merge conflict resolution (since the original file has been moved to another folder). The regression is that undo command override is not working, causing uncorrect undo command behaviour when triggered from webview
3. There is some redundant code which is not useful and causes incorrect behaviour, showing in the webview view screen

## Solution
1. To ensure compatibility, we are adding an unsupported message handler in advance. In case of unsupported message it will send an appropriate response back to the webview, so we will be able to use it to handle the error gracefully on the webview side
2. Adding back the undo command override (only when WFS editor is active). More context on why it's needed on [previous PR](https://github.com/aws/aws-toolkit-vscode/pull/6024), which was approved and merged previously. Verified that the command only applies to WFS editor and with this undo behaviour is fixed
3. Removing redundant code (verified the integration still works fine)

---

### DIFF
--- a/packages/core/src/stepFunctions/workflowStudio/workflowStudioEditorProvider.ts
+++ b/packages/core/src/stepFunctions/workflowStudio/workflowStudioEditorProvider.ts
@@ -6,11 +6,10 @@
 import * as vscode from 'vscode'
 import { getLogger } from '../../shared/logger/logger'
 import request from '../../shared/request'
-import fs from '../../shared/fs/fs'
 import { getClientId } from '../../shared/telemetry/util'
 import { telemetry } from '../../shared/telemetry/telemetry'
 import globals from '../../shared/extensionGlobals'
-import { getRandomString, getStringHash } from '../../shared/utilities/textUtilities'
+import { getStringHash } from '../../shared/utilities/textUtilities'
 import { ToolkitError } from '../../shared/errors'
 import { getTabSizeSetting } from '../../shared/utilities/editorUtilities'
 import { WorkflowStudioEditor } from './workflowStudioEditor'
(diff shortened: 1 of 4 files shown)


anthonyting (approved):

packages/core/src/stepFunctions/workflowStudio/workflowStudioEditorProvider.ts
```diff
@@ -93,20 +92,8 @@ export class WorkflowStudioEditorProvider implements vscode.CustomTextEditorProv
         const isDarkMode = theme === vscode.ColorThemeKind.Dark || theme === vscode.ColorThemeKind.HighContrast
         const tabSizeTag = `<meta name='tab-size' content='${getTabSizeSetting()}'>`
         const darkModeTag = `<meta name='dark-mode' content='${isDarkMode}'>`
-        let html = `${htmlFileSplit[0]} <head> ${baseTag} ${localeTag} ${darkModeTag} ${tabSizeTag} ${htmlFileSplit[1]}`
 
-        const nonce = getRandomString()
-        const localDevURL = isLocalDev ? localhost : ''
```
jpinkney-aws, line 99: Why is this part no longer relevant?

witness-me: We actually never used this for WFS integration as we never added custom scripts on webview side - I think I originally implemented it the same way as it was for Threat composer editor ([code](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/threatComposer/threatComposerEditorProvider.ts#L100-L112)).

Our integration works fine without this extra code, similar to how InfraComposer integration works ([code](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/applicationcomposer/webviewManager.ts#L46-L66))

jpinkney-aws (approved):

PR #6672, opened 2025-02-25
refactor(sdkv3): migrate ec2 to sdkv3

Hweinstock:
## Problem
EC2 currently uses the old sdkv2, we can start the migration to v3. 

## Solution
- migrate existing use cases and provide small clean ups along the way. 
- Re-work pagination to use `AsyncCollection` abstraction. 

## Testing
- Manually tested Ec2Connect, Explorer (polling functionality), Command Palette, and some edge cases including instances missing a name, duplicate names, and windows instance (for remote terminal). 

## Future Work
- Allow quickpicks to support `AsyncIterator` of individual items, to avoid loading entire collection. Alternatively, add a `batchIterator` to `AsyncCollection` that returns an `AsyncIterator` with batches of size `k`.
- built-in explorer support for pagination. 
- automated E2E testing for EC2?

---


packages/core/src/test/shared/clients/ec2Client.test.ts
```diff
@@ -77,51 +64,45 @@ const incompleteReservationsList: EC2.ReservationList = [
...
     })
 })
 
 describe('updateInstancesDetail', async function () {
-    let client: MockEc2Client
-
+    let client: Ec2Client
     before(function () {
-        client = new MockEc2Client('test-region')
+        client = new Ec2Client('test-region')
```
justinmk3, line 101: 👀 does/will the new client help avoid mocks in some cases?

Hweinstock: For SDKv3, the SDK team appears to recommend this package for mocking the clients: https://aws.amazon.com/blogs/developer/mocking-modular-aws-sdk-for-javascript-v3-in-unit-tests/. 

With this light refactor I also removed the hand-crafted mocks when possible.

packages/core/src/awsService/ec2/prompter.ts
```diff
@@ -58,16 +58,19 @@ export class Ec2Prompter {
         return await client.getInstances()
     }
 
+    // TODO: implement a batched generator to avoid loading all instances into UI.
```
justinmk3, line 61: For codecatalyst we use `AsyncCollection` , example: https://github.com/aws/aws-toolkit-vscode/blob/4e568970e8d8646cf8e590d39d8891d68cebde84/packages/core/src/shared/clients/codecatalystClient.ts#L427

and here is where that is used to build a prompter: https://github.com/aws/aws-toolkit-vscode/blob/4e568970e8d8646cf8e590d39d8891d68cebde84/packages/core/src/codecatalyst/wizards/selectResource.ts#L113-L118

justinmk3: Looks like you already anticipated this, based on `packages/core/src/shared/clients/clientWrapper.ts` :)

justinmk3 (approved):
Nice code reduction, too!

packages/core/src/awsService/ec2/prompter.ts
Hweinstock: Yea, the issue is that we flatten the pages to process the instances (i.e. add name and status). So the two options would be either keep it unflattened (as Reservations https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/ec2/command/DescribeInstancesCommand/) or re-batch them. It seems preferable to keep them unflattened as possible, but this presents some changes elsewhere. Investigating this now as a followup.

PR #6674, opened 2025-02-25
fix(amazonq): adjust feature config customization arn override logic

andrewyuq:
1. change poll interval from 30mins to 180 mins
2. Only call listAvailableCustomizations for new customization override for update
3. Use a new field customizationArnOverrideV2 for a clean override plan

## Solution

---

### DIFF
--- a/packages/core/src/shared/globalState.ts
+++ b/packages/core/src/shared/globalState.ts
@@ -45,7 +45,6 @@ export type globalKey =
     | 'aws.toolkit.amazonq.dismissed'
     | 'aws.toolkit.amazonqInstall.dismissed'
     | 'aws.amazonq.workspaceIndexToggleOn'
-    | 'aws.amazonq.customization.override'
     | 'aws.amazonq.customization.overrideV2'
     // Deprecated/legacy names. New keys should start with "aws.".
     | '#sessionCreationDates' // Legacy name from `ssoAccessTokenProvider.ts`.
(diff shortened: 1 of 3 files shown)


packages/core/src/shared/globalState.ts
```diff
@@ -46,6 +46,7 @@ export type globalKey =
     | 'aws.toolkit.amazonqInstall.dismissed'
     | 'aws.amazonq.workspaceIndexToggleOn'
     | 'aws.amazonq.customization.override'
+    | 'aws.amazonq.customization.overrideV2'
```
Will-ShaoHua, line 49: it should be safe to to remove L48 here I think `'aws.amazonq.customization.override'`

andrewyuq: ok

Will-ShaoHua (approved):

jpinkney-aws (approved):
The code itself looks fine to me but can you outline in the description what problem introducing a new field into global state solves

justinmk3:
> The code itself looks fine to me but can you outline in the description what problem introducing a new field into global state solves

Is it to avoid having to migrate old/stale state ?

andrewyuq:
> > The code itself looks fine to me but can you outline in the description what problem introducing a new field into global state solves
> 
> Is it to avoid having to migrate old/stale state ?

Yes, the reason we introduced a new field is that:
1. We originally introduced a field to store a service override arn from A/B
2. The A/B was rolled back on service side, causing half people having this field set and half-people not(who didn't update plugins)
3. We want to restart the service override and make sure it overrides for everyone
4. thus using a new field to make sure the override this time is not affected by the previous persisted state

PR #6679, opened 2025-02-25
feat(amazonq): generate scanName from workspace config

ctlai95:
## Problem

All code scans are run with a scanName as a UUID which makes it impossible to track the bug lifecycle of projects. We are not leveraging the bug tracking feature which allows grouping together different scan runs of the same project by passing in the same scanName between runs.

## Solution

Calculate the scanName using ClientID, the current workspace configuration (project root(s) or active file), and code analysis scope.

---

### DIFF
--- a/packages/amazonq/test/unit/codewhisperer/service/securityScanHandler.test.ts
+++ b/packages/amazonq/test/unit/codewhisperer/service/securityScanHandler.test.ts
@@ -325,27 +325,27 @@ describe('securityScanHandler', function () {
     describe('generateScanName', function () {
         const clientId = 'ffffffff-ffff-ffff-ffff-ffffffffffff'
 
-        it('should generate scan name for FILE_AUTO scope', function () {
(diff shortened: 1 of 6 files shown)


packages/amazonq/test/unit/codewhisperer/service/securityScanHandler.test.ts
```diff
@@ -320,4 +321,37 @@ describe('securityScanHandler', function () {
             await assert.rejects(() => pollPromise, SecurityScanTimedOutError)
         })
     })
+
+    describe('generateScanName', function () {
+        const clientId = 'ffffffff-ffff-ffff-ffff-ffffffffffff'
+
+        it('should generate scan name for FILE_AUTO scope', function () {
```
jpinkney-aws, line 328: you could probably drop the should on all of these tests so they read like:
```suggestion
        it('generates scan name for FILE_AUTO scope', function () {
```

PR #6561, opened 2025-02-11
feat(amazonq): /doc: add support for infrastructure diagrams

vikshe:
Amazon Q /doc: Add support for infrastructure diagrams

infra_diagram

---

- Treat all work as PUBLIC. Private `feature/x` branches will not be squash-merged at release time.
- Your code changes must meet the guidelines in [CONTRIBUTING.md](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#guidelines).
- License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6681, opened 2025-02-25
fix(amazonq): /test file collection usecase adds more patterns to filtering.

ashishrp-aws:
## Problem
/test has been seeing quite a lot of failures due to file paylaod limit exceeded errors in past weeks in both IDEs but in last week there has been huge rise in these errors for /test on VSCode.

## Solution
We have noticed a change that ignores to use default patterns in file filtering was introduced in PR #6561 . We have also collected data from internal customers by adding payload manifest in JetBrains for these errors and compiled a list of extensions which are causing this issue. We have decided to add these patterns to ignore list on VSCode as well.
- For Test Generation useCase, we have added back default pattern ignore list
- we have created testGenIgnorePattern list and added these patterns to Filtering methods for Test Generation UseCase.
- Fixed the workspace folders sorting for Test Generation useCase in ziputils and removed it from workspaceUtils

---

### DIFF
--- a/packages/core/src/codewhisperer/util/zipUtil.ts
+++ b/packages/core/src/codewhisperer/util/zipUtil.ts
@@ -412,7 +412,9 @@ export class ZipUtil {
 
         const sourceFiles = await collectFiles(
             projectPaths,
-            vscode.workspace.workspaceFolders as CurrentWsFolders,
+            [...(vscode.workspace.workspaceFolders ?? [])].sort(
+                (a, b) => b.uri.fsPath.length - a.uri.fsPath.length
+            ) as CurrentWsFolders,
             {
                 maxSizeBytes: this.getProjectScanPayloadSizeLimitInBytes(),
             },
--- a/packages/core/src/codewhisperer/util/zipUtil.ts
+++ b/packages/core/src/codewhisperer/util/zipUtil.ts
@@ -412,9 +412,11 @@ export class ZipUtil {
 
         const sourceFiles = await collectFiles(
             projectPaths,
-            [...(vscode.workspace.workspaceFolders ?? [])].sort(
-                (a, b) => b.uri.fsPath.length - a.uri.fsPath.length
-            ) as CurrentWsFolders,
+            (useCase === FeatureUseCase.TEST_GENERATION
+                ? [...(vscode.workspace.workspaceFolders ?? [])].sort(
+                      (a, b) => b.uri.fsPath.length - a.uri.fsPath.length
+                  )
+                : vscode.workspace.workspaceFolders) as CurrentWsFolders,
             {
                 maxSizeBytes: this.getProjectScanPayloadSizeLimitInBytes(),
             },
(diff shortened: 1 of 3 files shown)


laileni-aws (approved):

packages/core/src/codewhisperer/util/zipUtil.ts
```diff
@@ -412,7 +412,9 @@ export class ZipUtil {
 
         const sourceFiles = await collectFiles(
             projectPaths,
-            vscode.workspace.workspaceFolders as CurrentWsFolders,
+            [...(vscode.workspace.workspaceFolders ?? [])].sort(
```
laileni-aws, line 415: Was this used by `/review` too? If yes we need to use `useCase` to differentiate, If not we should be fine.

ashishrp-aws: Fixed in third commit

packages/core/src/codewhisperer/util/zipUtil.ts
```diff
@@ -412,7 +412,11 @@ export class ZipUtil {
 
         const sourceFiles = await collectFiles(
             projectPaths,
-            vscode.workspace.workspaceFolders as CurrentWsFolders,
+            (useCase === FeatureUseCase.TEST_GENERATION
```
laileni-aws, line 415: Thanks @ashishrp-aws for this..

packages/core/src/shared/utilities/workspaceUtils.ts
```diff
@@ -393,7 +387,10 @@ export async function collectFiles(
     const inputExcludePatterns = options?.excludePatterns ?? defaultExcludePatterns
     const maxSizeBytes = options?.maxSizeBytes ?? maxRepoSizeBytes
 
-    const excludePatterns = [...getGlobalExcludePatterns()]
+    const excludePatterns = [
+        ...getGlobalExcludePatterns(),
+        ...(useCase === FeatureUseCase.TEST_GENERATION ? [...testGenExcludePatterns, ...defaultExcludePatterns] : []),
```
jpinkney-aws, line 392: Can we figure out a better way to not bury useCase inside of functions? this is exactly what we want to avoid. Otherwise it just makes sense to not have this be a common util

What you could do instead is just pass in additionalExclusion patterns to getWorkspaceRelativePath and then if its set it will add those to excludePatterns rather than putting your useCase inside of the function

packages/core/src/codewhisperer/util/zipUtil.ts
jpinkney-aws: Like i mentioned in the comment, is there any way we can remove this useCase stuff? IIRC this was supposed to be removed last week since it was a quick hack to fix a problem

packages/core/src/shared/utilities/workspaceUtils.ts
ashishrp-aws: Amended the PR.

jpinkney-aws (approved):
Ideally we would get rid of this from processSourceFiles as well. I'm guessing this is neccessary for this release though?

FWIW you can see the consequence of this all throughout the zipUtil package. Theres `generateZip` and `generateZipTestGen` which already do very similiar things

jpinkney-aws (changes requested):
ci is failing

ashishrp-aws:
> Ideally we would get rid of this from processSourceFiles as well. I'm guessing this is neccessary for this release though?
> 
> FWIW you can see the consequence of this all throughout the zipUtil package. Theres `generateZip` and `generateZipTestGen` which already do very similiar things

Yes...necessary there to differentiate usage for review and test

jpinkney-aws (approved):

PR #6682, opened 2025-02-26
fix(ec2): properly paginate instances in quickPick. 

Hweinstock:
## Problem
Follow up to discussion: https://github.com/aws/aws-toolkit-vscode/pull/6672#discussion_r1972045094

Currently the ec2 module does make paginated requests to the SDK endpoint. However, in order to show these results into the UI, we resolve them in entirety.
For the quick picks this is done here: [https://github.com/aws/aws-toolkit-vscode/blob/4e568970e8d8646cf8e590d39d8891d68cebde84/packages/core/src/awsService/ec2/prompter.ts#L65 ](https://github.com/aws/aws-toolkit-vscode/blob/4e568970e8d8646cf8e590d39d8891d68cebde84/packages/core/src/awsService/ec2/prompter.ts#L65)
For the explorer this is done here: [https://github.com/aws/aws-toolkit-vscode/blob/4e568970e8d8646cf8e590d39d8891d68cebde84/packages/core/src/awsService/ec2/explorer/ec2ParentNode.ts#L52 ](https://github.com/aws/aws-toolkit-vscode/blob/4e568970e8d8646cf8e590d39d8891d68cebde84/packages/core/src/awsService/ec2/explorer/ec2ParentNode.ts#L52)

## Solution
- Refactor the EC2 client to process the instances within the pages they come in (i.e. not flattening them). This allows passing an `AsyncCollection` of pages to the quickPick. 
- Avoid flattening results in `makePaginatedRequest` by default.  
- Refactor client builder to be non-async, this causes some trickle-down refactoring, but allows `makePaginatedRequest` to not return a promise. 
- Simplify instance processing to make use of type predicates. 
- Remove mocks from `Ec2Prompter` tests, and refactor to test pagination behavior. 

## Alternative Solution 
- Repack the instances into artificial pages (i.e. a batched iterator). This is a little hacky since it involves us unpacking and repacking the instances. However, it does involve significantly less code changes. 

## Verification
In addition to updating the tests, we can add a logging statement when the quickPick loads an item from the `AsyncIterable`. This screenshot shows that loading two items was successful and triggered the `AsyncIterable` twice, as expected. 
image

## Future Work
- Extend this work to the explorer. There is already some general work done in https://github.com/aws/aws-toolkit-vscode/blob/df810aa4264ba14dac488697d539011a732c7eff/packages/core/src/awsexplorer/childNodeLoader.ts, but it doesn't support working with `AsyncCollection`, only making the requests directly. We could do something hacky, like wrap the `AsyncCollection` in a fake request method, or try to refactor the utility itself. 
---


packages/core/src/shared/clients/ec2.ts
```diff
@@ -31,14 +31,18 @@ import { ClientWrapper } from './clientWrapper'
 import { AsyncCollection } from '../utilities/asyncCollection'
 
 /**
- * A wrapper around EC2.Instance where we can safely assume InstanceId field exists.
+ * A wrapper around Instance where we can safely assume InstanceId field exists.
  */
-export interface SafeEc2Instance extends Instance {
+export interface PatchedEc2Instance extends Instance {
```
justinmk3, line 36: would we have some other class for "unsafe"? or is that just an improvement that we are making in our wrapper class?

```suggestion
export interface Ec2Instance extends Instance {
```

packages/core/src/shared/clients/ec2.ts
```diff
@@ -31,14 +31,18 @@ import { ClientWrapper } from './clientWrapper'
...
     InstanceId: string
     Name?: string
     LastSeenStatus: InstanceStateName
 }
 
+export interface PatchedReservation extends Reservation {
```
justinmk3, line 42:
```suggestion
export interface Ec2Reservation extends Reservation {
```

packages/core/src/shared/clients/clientWrapper.ts
```diff
@@ -21,43 +21,48 @@ export abstract class ClientWrapper<C extends AwsClient> implements vscode.Dispo
...
 
         return collection
 
-        function isDefined(i: Output | undefined): i is Output {
+        function isDefined<T>(i: T | undefined): i is T {
             return i !== undefined
         }
     }
 
+    protected async getFirstResult<CommandInput extends object, CommandOutput extends object, Output extends object>(
```
justinmk3, line 57:
```suggestion
    protected async getFirst<CommandInput extends object, CommandOutput extends object, Output extends object>(
```

justinmk3 (approved):

packages/core/src/shared/clients/ec2.ts
Hweinstock: This is an improvement here to make processing the responses easier. Avoids undefined checks in multiple places since we assert it once, and store that info in the type.

PR #6685, opened 2025-02-26
docs(amazonq): Add instructions on how to start Amazon Q inline debugging

jpinkney-aws:
## Problem
- Multiple people have gotten confused on how to get the language server activated for inline

## Solution
- Add that to the language server debugging documentation

---

### DIFF
--- a/docs/lsp-debugging.md
+++ b/docs/lsp-debugging.md
@@ -24,4 +24,4 @@
 
 ## Amazon Q Inline Activation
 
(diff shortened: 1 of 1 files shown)


hayemaxi:
The referenced list is adhoc right? IIRC there is a legit list somewhere in the code. We should update this list from that or use that list directly.

hayemaxi (approved):

jpinkney-aws:
The legit list should eventually move to the same spot I mention in the docs. There isn't a 1to1 mapping right now between what the language server wants and how things are setup on codewhisperer side though. This is just to unblock Zoe/Will since all of us have forgotten that its using a subset of the actual languages 😆

PR #6693, opened 2025-02-28
fix(stepfunctions): Prevent opening Workflow Studio for invalid JSON/YAML; add save and deploy message handler

witness-me:
## Problem
1. Deploy state machine workflow is not very visible to the customers
2. Currently when user is opening WFS with custom action in the side panel and has invalid JSON/YAML, the side panel opens and immediately closes (since WFS does not support invalid JSON/YAML). This experience does not look smooth.

## Solution
1. With Workflow Studio integration, we'll add `Save & Deploy` button to the UI which will save file, close WFS editor and start the deployment flow. Adding a message handler for this action
2. Adding a check for invalid JSON/YAML before attempting to open WFS in the side panel. If file is invalid, the panel will not be opened and the user will see a notification
---

### DIFF
--- a/packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
+++ b/packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
@@ -25,6 +25,7 @@ import { ToolkitError } from '../../shared/errors'
 import { WorkflowStudioApiHandler } from './workflowStudioApiHandler'
 import globals from '../../shared/extensionGlobals'
 import { getLogger } from '../../shared/logger/logger'
+import { publishStateMachine } from '../commands/publishStateMachine'
 
 const localize = nls.loadMessageBundle()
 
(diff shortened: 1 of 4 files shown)


anthonyting (approved):

packages/core/src/stepFunctions/workflowStudio/handleMessage.ts
```diff
@@ -174,6 +178,18 @@ async function saveFileMessageHandler(request: SaveFileRequestMessage, context:
...
 
+/**
+ * Handler for saving a file and starting the state machine deployment flow, while also switching to default editor.
+ * Triggered when the user triggers 'Save and Deploy' action in WFS
+ * @param request The request message containing the file contents.
+ * @param context The webview context containing the necessary information for saving the file.
+ */
+async function saveFileAndDeployMessageHandler(request: SaveFileRequestMessage, context: WebviewContext) {
+    await saveFileMessageHandler(request, context)
+    await closeCustomEditorMessageHandler(context)
```
jpinkney-aws, line 189: What benefit does closing the editor do here? For some reason save and deploy makes me think the editor should remain open and then the underlying stepfunctions file gets saved and then deployed

witness-me: Agreed, keeping it opened is a better experience and this is our longer-term plan. However with the current setup this would fail since VSCode does not detect the file opened in the custom editor and throws the following error:

```
Error: Could not get active text editor for state machine definition 
```

Fixing this properly would require some extra time to refactor existing logic, but for now, we wanted to get this change out before launch so that everyone with WFS has the command in their toolkit.  This way, there is no risk that customer has outdated toolkit version, and our CDN contains the button that does not work for them. 

We’ve got a follow-up task to improve the experience, which we’ll tackle post-launch

jpinkney-aws (approved):

PR #6701, opened 2025-03-03
fix(amazonq):fix cross-tab data interference

KevinDing1:
## Problem
- When users open multiple tabs while using /doc command, tab data interferes with each other
- This causes incorrect user flow and inaccurate telemetry tracking

## Solution
- Implemented proper tab isolation mechanism
- Ensured each tab maintains its own independent state
---


vikshe (approved):

Hweinstock:
Can we verify this behavior with an explicit unit test that fails without these changes? Something like opening two tabs and looking at the telemetry with something like `assertTelemetry` from [here](https://github.com/aws/aws-toolkit-vscode/blob/ccd21b22c9c21801d352d61a1eb2ea2a668b1088/packages/core/src/test/testUtil.ts#L322).  

Otherwise LGTM.

packages/amazonq/.changes/next-release/Bug Fix-4f4a3fc1-a25a-4c26-abe8-8de424ddaabd.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "Amazon Q: Fix data isolation between tabs to prevent interference when using /doc in multiple tabs"
```
hayemaxi, line 3:
```suggestion
	"description": "/doc: Usage in multiple chat tabs may cause unexpected behavior."
```

KevinDing1:
> Can we verify this behavior with an explicit test that fails without these changes? Something like opening two tabs and looking at the telemetry with something like `assertTelemetry` from [here](https://github.com/aws/aws-toolkit-vscode/blob/ccd21b22c9c21801d352d61a1eb2ea2a668b1088/packages/core/src/test/testUtil.ts#L322).
> 
> Otherwise LGTM.

That's a good callout, I will work on a separate PR for adding such unit test. But we have to merge this PR to fix the issue first.

Hweinstock (approved):

PR #6707, opened 2025-03-04
fix(amazonq): context fixes for multiroot workspace setup

andrewyuq:
## Problem
In multiroot workspace setup, the contextList UI displays files that are deduped for the same files and doesn't resolve relative paths correctly.

1. In a multi-root workspace, the index is built on the root of the 1st project after sorting workspaceFolders, thus the chunk response also contains relativePath to the 1st project root after sorting. In order to get the same relativePath from explicit @file, we also need to use the root of the 1st project after sorting workspaceFolders.
2. Use a relativePath to projectRoot map to store the project root for each relativePath so that on clicking the file(which only returns relativePath in the event) we correctly construct the absolute path of the file to open in the editor.

## Solution

---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -582,7 +582,6 @@ export class ChatController {
     }
     private async processFileClickMessage(message: FileClick) {
         const session = this.sessionStorage.getSession(message.tabID)
-        // TODO remove currentContextId but use messageID to track context for each answer message
         const lineRanges = session.contexts.get(message.filePath)
 
         if (!lineRanges) {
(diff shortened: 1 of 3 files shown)


packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -589,14 +590,14 @@ export class ChatController {
...
         const session = this.sessionStorage.getSession(message.tabID)
         // TODO remove currentContextId but use messageID to track context for each answer message
-        const lineRanges = session.contexts.get(session.currentContextId)?.get(message.filePath)
+        const lineRanges = session.contexts.get(message.filePath)
 
         if (!lineRanges) {
             return
         }
 
         // TODO: Fix for multiple workspace setup
```
leigaol, line 599: remove this todo?

leigaol (approved):

leigaol (approved):
LGTM

Hweinstock:
What problem(s) is this exactly solving? What are the bugs with context that need fixes? What is the solution at a high-level for those less familiar with this code?

I'd recommend answering these questions in the PR description. This will also provides a great reference in the commit log when looking back.

avi-alpert (approved):

nkomonen-amazon (approved):

PR #6710, opened 2025-03-04
telemetry(amazonq): add to amazonq_interactwithmessage

avi-alpert:
Send metrics on additional context to the amazonq_interactwithmessage event. This enables measuring success metrics as they relate to the new context features.

---


packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -915,8 +912,13 @@ export class ChatController {
             return []
         }
         const workspaceFolder = contextCommands[0].workspaceFolder
-        const prompts = await LspClient.instance.getContextCommandPrompt(contextCommands)
-        if (prompts.length === 0) {
+        let prompts = []
+        try {
+            prompts = await LspClient.instance.getContextCommandPrompt(contextCommands)
+        } catch {
+            return []
```
leigaol, line 919: do you need to return here? Maybe just continue executing below code.

leigaol: if there is any exception or empty list, triggerPayload.additionalContents = [] at line 926 still needs to be executed.

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -929,7 +931,6 @@ export class ChatController {
             ruleContextLength: 0,
         }
         for (const prompt of prompts.slice(0, 20)) {
```
leigaol, line 933: L926 to line 931 is still needed even if prompt is empty.

PR #6713, opened 2025-03-04
telemetry(amazonq): emit metrics per finding

ctlai95:
## Problem

Events about findings are always tied to hover, apply fix, etc. Not able to understand the number of findings which never get acted on.

## Solution

Emit metrics about findings at the time of scan results.

---

### DIFF
--- a/packages/core/src/codewhisperer/commands/startSecurityScan.ts
+++ b/packages/core/src/codewhisperer/commands/startSecurityScan.ts
@@ -263,6 +263,19 @@ export async function startSecurityScan(
             scope,
             editor
         )
+        for (const issue of securityRecommendationCollection
+            .flatMap(({ issues }) => issues)
+            .filter(({ visible, autoDetected }) => visible && !autoDetected)) {
+            telemetry.codewhisperer_codeScanIssueDetected.emit({
+                autoDetected: issue.autoDetected,
+                codewhispererCodeScanJobId: issue.scanJobId,
+                detectorId: issue.detectorId,
+                findingId: issue.findingId,
+                includesFix: issue.suggestedFixes.length > 0,
+                ruleId: issue.ruleId,
+                result: 'Succeeded',
+            })
+        }
         const { total, withFixes } = securityRecommendationCollection.reduce(
             (accumulator, current) => ({
                 total: accumulator.total + current.issues.length,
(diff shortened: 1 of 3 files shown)


packages/core/src/codewhisperer/commands/startSecurityScan.ts
```diff
@@ -263,6 +263,19 @@ export async function startSecurityScan(
...
+            .flatMap(({ issues }) => issues)
+            .filter(({ visible, autoDetected }) => visible && !autoDetected)) {
+            telemetry.codewhisperer_codeScanIssueDetected.emit({
+                autoDetected: issue.autoDetected,
+                codewhispererCodeScanJobId: issue.scanJobId,
+                detectorId: issue.detectorId,
+                findingId: issue.findingId,
+                includesFix: issue.suggestedFixes.length > 0,
+                ruleId: issue.ruleId,
+                result: 'Succeeded',
```
laileni-aws, line 276: I think we are missing `result` in `codewhisperer_codeScanIssueDetected` event?

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1325,6 +1325,34 @@
...
+        {
+            "name": "codewhisperer_codeScanIssueDetected",
+            "description": "Called when a code scan issue is returned from the service",
+            "metadata": [
+                {
+                    "type": "autoDetected",
+                    "required": false
+                },
+                {
+                    "type": "codewhispererCodeScanJobId",
```
laileni-aws, line 1338: Can we name this as `jobId` ? as we already have this field in commonDefinations.json

laileni-aws (approved):

packages/core/src/codewhisperer/commands/startSecurityScan.ts
ctlai95: `result` is automatically added, and telemetry will actually complain if it's not included here

packages/core/src/shared/telemetry/vscodeTelemetry.json
ctlai95: Will keep this for now, since the other codescan-related metrics are using this one

nkomonen-amazon (approved):

PR #6714, opened 2025-03-04
telemetry(amazonq): add metrics to `amazonq_interactWithMessage`

avi-alpert:
- Send metrics on additional context to the `amazonq_interactWithMessage` event. This enables measuring success metrics as they relate to the new context features. 
- Improve error handling when collecting context items & metrics.
- Add new features to /help message
- Improve efficiency by switching from vscode.workspace.findFiles to fs.readDirectory

---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -588,8 +588,9 @@ export class ChatController {
             return
         }
 
-        // TODO: Fix for multiple workspace setup
-        const projectRoot = session.relativePathToWorkspaceRoot.get(message.filePath)
+        // Check if clicked file is in a different workspace root
+        const projectRoot =
+            session.relativePathToWorkspaceRoot.get(message.filePath) || workspace.workspaceFolders?.[0]?.uri.fsPath
         if (!projectRoot) {
             return
         }
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -582,7 +582,7 @@ export class ChatController {
     }
     private async processFileClickMessage(message: FileClick) {
         const session = this.sessionStorage.getSession(message.tabID)
-        const lineRanges = session.contexts.get(message.filePath)
+        const lineRanges = session.contexts.get(session.currentContextId)?.get(message.filePath)
 
         if (!lineRanges) {
             return
@@ -892,7 +892,6 @@ export class ChatController {
                     const workspaceFolderPath = vscode.workspace.getWorkspaceFolder(rule)?.uri?.path || ''
                     return {
                         workspaceFolder: workspaceFolderPath,
-                        // todo: add 'prompt' type to LSP model
                         type: 'file' as ContextCommandItemType,
                         relativePath: path.relative(workspaceFolderPath, rule.path),
                     }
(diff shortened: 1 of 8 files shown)


avi-alpert:
No changelog is needed because feature hasn't been released yet

leigaol (approved):

andrewyuq (approved):

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -877,6 +878,34 @@ export class ChatController {
...
+        if (!vscode.workspace.workspaceFolders) {
+            return rulesFiles
+        }
+
+        for (const folder of vscode.workspace.workspaceFolders) {
+            const rulesPath = vscode.Uri.joinPath(folder.uri, '.amazonq', 'rules')
+            const folderExists = await fs.exists(rulesPath)
+
+            if (folderExists) {
+                const entries = await vscode.workspace.fs.readDirectory(rulesPath)
```
nkomonen-amazon, line 896: We should be using the `fs` import for all fs related operations, and for this case `fs.readdir()`

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -930,52 +957,58 @@ export class ChatController {
...
             session.relativePathToWorkspaceRoot.set(relativePath, contextCommand.workspaceFolder)
         }
-        const prompts = await LspClient.instance.getContextCommandPrompt(contextCommands)
-        if (prompts.length === 0) {
-            return []
+        let prompts: AdditionalContextPrompt[] = []
+        try {
+            prompts = await LspClient.instance.getContextCommandPrompt(contextCommands)
+        } catch (e) {
+            getLogger().verbose(`Could not get context command prompts: ${e}`)
```
nkomonen-amazon, line 964: swallowing errors are dangerous. So unless you have a good reason they should propagate upwards and be handled properly if possible

avi-alpert: Its not an error that I want to throw or surface. this function fails if the user tries to use `@workspace` but indexing is not yet complete. we have a follow up item to better address this case but for now we don't want to throw an error. I can add a todo here

leigaol: It throws when LSP is not ready (not spinning up yet) or the LSP somehow went wrong. In this case, we return empty list and swallow the error so that the user can still see the context menu popping up.

leigaol: If we throw here, the UX will be broken. 

In the near short term, this API will be moved to FLARE.

packages/core/src/codewhispererChat/controllers/chat/controller.ts
avi-alpert: Fixed

packages/core/src/codewhispererChat/controllers/chat/controller.ts
nkomonen-amazon: I see, sounds good

If you see value, you can emit a metric on failure to have some data to observe in Kibana. Just a recomendation.

nkomonen-amazon (approved):

justinmk3:
CI is failing, but it appears to be a flaky issue, doesn't look related to this PR.

```
rejected promise not handled within 1 second: Error: Language client is not ready yet
rejected promise not handled within 1 second: Error: Language client is not ready yet
```

PR #6715, opened 2025-03-04
telemetry(amazonq): createUpload metric for scans

ctlai95:
## Problem

Some users are reporting S3 upload failures but we don't have the requestIds to investigate further.

## Solution

Emit `amazonq_createUpload` metric with requestIds

---

### DIFF
--- a/packages/core/src/shared/telemetry/spans.ts
+++ b/packages/core/src/shared/telemetry/spans.ts
@@ -219,8 +219,8 @@ export class TelemetrySpan<T extends MetricBase = MetricBase> implements Span<T>
                 result: getTelemetryResult(err),
                 reason: getTelemetryReason(err),
                 reasonDesc: getTelemetryReasonDesc(err),
-                requestId: getRequestId(err),
-                httpStatusCode: getHttpStatusCode(err),
+                ...(!this.state.requestId ? { requestId: getRequestId(err) } : {}),
+                ...(!this.state.httpStatusCode ? { httpStatusCode: getHttpStatusCode(err) } : {}),
             } as Partial<T>)
         }
 
--- a/packages/core/src/shared/telemetry/vscodeTelemetry.json
+++ b/packages/core/src/shared/telemetry/vscodeTelemetry.json
@@ -410,6 +410,17 @@
             "name": "executedCount",
             "type": "int",
             "description": "The number of executed operations"
+        },
+        {
+            "name": "amazonqUploadIntent",
+            "type": "string",
+            "description": "The intent of the upload",
+            "allowedValues": [
+                "TRANSFORMATION",
+                "TASK_ASSIST_PLANNING",
+                "AUTOMATIC_FILE_SECURITY_SCAN",
+                "FULL_PROJECT_SECURITY_SCAN"
+            ]
         }
     ],
     "metrics": [
@@ -1167,7 +1178,31 @@
                     "required": false
                 },
                 {
+                    "type": "amazonqUploadIntent",
+                    "required": false
+                },
+                {
                     "type": "credentialStartUrl",
+                    "required": false
+                },
+                {
+                    "type": "duration",
+                    "required": false
+                },
+                {
+                    "type": "reason",
+                    "required": false
+                },
+                {
+                    "type": "requestId",
+                    "required": false
+                },
+                {
+                    "type": "requestServiceType",
+                    "required": false
+                },
+                {
+                    "type": "result",
                     "required": false
                 }
             ]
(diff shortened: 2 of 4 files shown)


packages/core/src/shared/telemetry/spans.ts
```diff
@@ -219,8 +219,8 @@ export class TelemetrySpan<T extends MetricBase = MetricBase> implements Span<T>
                 result: getTelemetryResult(err),
                 reason: getTelemetryReason(err),
                 reasonDesc: getTelemetryReasonDesc(err),
-                requestId: getRequestId(err),
-                httpStatusCode: getHttpStatusCode(err),
+                ...(!this.state.requestId ? { requestId: getRequestId(err) } : {}),
+                ...(!this.state.httpStatusCode ? { httpStatusCode: getHttpStatusCode(err) } : {}),
```
ctlai95, line 223: Making this change to allow `requestId` and `httpStatusCode` to be set explicitly before the `stop` method is invoked. Otherwise it will silently get removed if `err` is not the correct format.

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -410,6 +410,17 @@
             "name": "executedCount",
             "type": "int",
             "description": "The number of executed operations"
+        },
+        {
+            "name": "amazonqUploadIntent",
+            "type": "string",
+            "description": "The intent of the upload",
+            "allowedValues": [
+                "TRANSFORMATION",
```
laileni-aws, line 419: nit: Can you add `TEST_GENERATION` as allowed value, if not I can add this as part of next PR's or while shipping this to common package.

laileni-aws (approved):

packages/core/src/shared/telemetry/vscodeTelemetry.json
laileni-aws: NVM, I will make change for UTG.
Thanks @ctlai95

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -1166,9 +1177,33 @@
...
+                {
+                    "type": "amazonqUploadIntent",
+                    "required": false
+                },
                 {
                     "type": "credentialStartUrl",
                     "required": false
+                },
+                {
+                    "type": "duration",
```
nkomonen-amazon, line 1189: All metrics have certain fields implicitly attached by default. This is from `telemetry.gen.ts`:

```
export interface MetricBase {
    /**
     * AWS account ID associated with a metric.
     * - "n/a" if credentials are not available.
     * - "not-set" if credentials are not selected.
     * - "invalid" if account ID cannot be obtained.
     */
    readonly awsAccount?: string
    /** An AWS region. */
    readonly awsRegion?: string
    /** The duration of the operation in milliseconds */
    readonly duration?: number
    /** HTTP status code for the request (if any) associated with a metric. */
    readonly httpStatusCode?: string
    /** Reason code or name for an event (when result=Succeeded) or error (when result=Failed). Unlike the `reasonDesc` field, this should be a stable/predictable name for a class of events or errors (typically the exception name, e.g. FileIOException). */
    readonly reason?: string
    /** Error message detail. May contain arbitrary message details (unlike the `reason` field), but should be truncated (recommendation: 200 chars). */
    readonly reasonDesc?: string
    /** Request ID (if any) associated with a metric. For example, an event with `requestServiceType: s3` means that the request ID is associated with an S3 API call. Events that cover multiple API calls should use the request ID of the most recent call. */
    readonly requestId?: string
    /** Per-request service identifier. Unlike `serviceType` (which describes the originator of the request), this describes the request itself. */
    readonly requestServiceType?: string
    /** The result of the operation */
    readonly result?: Result
    /** The source of the operation. This answers 'who' caused/triggered the operation. Example: did an Auth signout happen because of some expiration or since the user explicitly clicked the signout button. */
    readonly source?: string
    /** A flag indicating that the metric was not caused by the user. */
    readonly passive?: boolean
    /** @deprecated Arbitrary "value" of the metric. */
    readonly value?: number
    /** A flag indicating that the metric should track run-time performance information */
    readonly trackPerformance?: boolean
    /** Unique identifier for the trace (a set of events) this metric belongs to */
    readonly traceId?: string
    /** Unique identifier for this metric */
    readonly metricId?: string
    /** Unique identifier of the parent of this metric */
    readonly parentId?: string
}
```

So you can get rid of a few from here

nkomonen-amazon (approved):

PR #6711, opened 2025-03-04
test(lambda): add integration for Serverless Land

vicheey:
## Problem
No test coverage for Serverless Land Integration workflow. 

## Solution
Add integration test for Serverless Land Integration workflow. 

---

- Treat all work as PUBLIC. Private `feature/x` branches will not be squash-merged at release time.
- Your code changes must meet the guidelines in [CONTRIBUTING.md](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#guidelines).
- License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6719, opened 2025-03-04
feat(lambda): adding metrics and unit test for serverlessland

Vandita2020:
## Problem
Adding metrics and unit test for Serverless Land Project

## Solution

- CreateWizard
    :) Path
      ✔ creates project with Python runtime and CDK (725ms)
    Error Handling
      ✔ handles empty project name (521ms)

- MetadataManager
    getInstance
      ✔ should create a new instance when none exists
      ✔ should return the same instance when called multiple times
    initialize
      ✔ should initialize metadata manager and return instance
    getMetadataPath
      ✔ should return correct metadata path
    getPatterns
      ✔ handles different pattern data types
    getRuntimes
      ✔ returns empty array when pattern not found
      ✔ returns unique runtimes for valid pattern
    getUrl
      ✔ returns empty string when pattern not found
      ✔ returns correct URL for valid pattern
    getIacOptions
      ✔ returns empty array when pattern not found
      ✔ returns unique IAC options for valid pattern
    getAssetName
      ✔ returns empty string when pattern not found
      ✔ returns correct asset name for matching implementation
      ✔ returns empty string when no matching implementation found

-   createNewServerlessLandProject
    ✔ should complete project creation successfully
    ✔ should handle wizard cancellation
  downloadPatternCode
    ✔ successfully downloads pattern code
    ✔ handles download failure
  openReadmeFile
    ✔ successfully opens README file
    ✔ handles missing README file
    ✔ handles error with opening README file
  getProjectUri
    ✔ returns Uri when file exists
    ✔ handles missing project directory

---


packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -644,6 +644,22 @@
             "name": "aws_hideRegion",
             "description": "Records a call to remove a region from the explorer"
         },
+        {
+            "name": "serverlessland_createProject",
```
justinmk3, line 648: why are you adding a new namespace `serverlessland_` ? why can't this use the existing `sam_` or `lambda_` namespace? using shared namespaces helps intuition and discovery.

Also, these metrics should be added to https://github.com/aws/aws-toolkit-common

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -644,6 +644,22 @@
...
+        {
+            "name": "serverlessland_downloadPattern",
+            "description": "Called when downloading a Serverless Pattern"
+        },
+        {
+            "name": "serverlessland_readme",
+            "description": "Called when opening the Serverless Pattern Readme"
+        },
+        {
+            "name": "serverlessland_wizard",
```
justinmk3, line 660: is there an existing metric that could be used? check here: https://github.com/aws/aws-toolkit-common/blob/main/telemetry/definitions/commonDefinitions.json

packages/core/src/awsService/appBuilder/serverlessLand/main.ts
```diff
@@ -95,40 +93,59 @@ async function launchProjectCreationWizard(
...
+            })
             return
         }
         await vscode.commands.executeCommand('workbench.action.focusFirstEditorGroup')
         await vscode.commands.executeCommand('markdown.showPreview', readmeUri)
+        telemetry.serverlessland_readme.emit({
+            result: 'Succeeded',
+        })
     } catch (err) {
+        telemetry.serverlessland_readme.emit({
```
justinmk3, line 134: Please read https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md

packages/core/src/test/awsService/appBuilder/serverlessLand/main.test.ts
```diff
@@ -0,0 +1,285 @@
...
+            .withArgs('test-pattern', 'node.js', 'sam')
+            .returns('test-asset')
+        const downloadStub = sandbox.stub(main, 'downloadPatternCode').resolves()
+        const readmeStub = sandbox.stub(main, 'openReadmeFile').resolves()
+        const addFolderStub = sandbox.stub(workspaceUtils, 'addFolderToWorkspace').resolves()
+
+        await main.createNewServerlessLandProject(mockExtContext)
+
+        sandbox.assert.calledOnce(wizardStub)
+        sandbox.assert.calledOnce(metadataStub)
```
justinmk3, line 69: why are you using stubs? check actual state.

please read the codebase guidelines.

packages/core/src/awsService/appBuilder/serverlessLand/main.ts
```diff
@@ -95,40 +93,59 @@ async function launchProjectCreationWizard(
...
 }
 
-async function downloadPatternCode(config: CreateServerlessLandWizardForm, assetName: string): Promise<void> {
+export async function downloadPatternCode(config: CreateServerlessLandWizardForm, assetName: string): Promise<void> {
     const fullAssetName = assetName + '.zip'
     const location = vscode.Uri.joinPath(config.location, config.name)
     try {
         await getPattern(serverlessLandOwner, serverlessLandRepo, fullAssetName, location, true)
+        telemetry.record({
+            action: fullAssetName + config.iac + config.runtime,
```
seshubaws, line 102: just making sure, is `fullAssetName` the name of the template or the name the cx provides?

seshubaws: Not sure about the way we're creating `action`. If we want it to be a string, there should be spaces in between, or some separator. But more importantly, this does not currently read as an action. And do we not need a `source` field to know/sort where this is coming from?

packages/core/src/test/awsService/appBuilder/serverlessLand/main.test.ts
seshubaws: There are integ tests https://github.com/aws/aws-toolkit-vscode/pull/6711/files checking actual state. I assume the reasoning was to stub in the unit tests so that we wouldn't be creating new projects for tests too frequently, @Vandita2020 correct me if I'm wrong

packages/core/src/test/awsService/appBuilder/serverlessLand/metadataManager.test.ts
```diff
@@ -0,0 +1,143 @@
...
+        })
+    })
+
+    describe('getPatterns', () => {
+        let manager: MetadataManager
+        beforeEach(() => {
+            manager = MetadataManager.getInstance()
+        })
+        it('handles different pattern data types', () => {
+            ;(manager as any).metadata = {
```
seshubaws, line 88: why is there ; here?

packages/core/src/test/awsService/appBuilder/serverlessLand/wizard.test.ts
```diff
@@ -0,0 +1,131 @@
...
+                    await quickPick.untilReady()
+                    const options = quickPick.items
+                    quickPick.acceptItem(options[0]) // sam
+                })
+                .handleQuickPick('Select Project Location', async (quickPick) => {
+                    await quickPick.untilReady()
+                    quickPick.acceptItem(quickPick.items[0])
+                })
+                .handleInputBox('Enter Project Name', (inputBox) => {
+                    inputBox.acceptValue('python-cdk-project')
```
seshubaws, line 67: comment in line 60 says "sam"?

packages/core/src/test/awsService/appBuilder/serverlessLand/wizard.test.ts
```diff
@@ -0,0 +1,131 @@
...
+                )
+
+            assert.ok(projectNode)
+            const resourceNodes = await projectNode.getChildren()
+            console.log('resourceNodes', resourceNodes)
+            assert.strictEqual(resourceNodes.length, 3)
+            assert.ok(resourceNodes[0] instanceof ResourceNode)
+
+            const lambdaResource = resourceNodes[2] as ResourceNode
+            console.log('retgbgbhtt', lambdaResource.resource.resource.Runtime)
```
seshubaws, line 89: can we remove this?

packages/core/src/shared/telemetry/vscodeTelemetry.json
Vandita2020: I'll go with the `lambda` as this project is not specific to `sam`

packages/core/src/shared/telemetry/vscodeTelemetry.json
Vandita2020: In the current metrics implementation, there are separate telemetry data streams to record the usage and behavior of different wizard features. So I have kept this to record telemetry for Serverless Land Wizard

packages/core/src/test/awsService/appBuilder/serverlessLand/main.test.ts
Vandita2020: Yes, @seshubaws's explanation is correct here, for unit tests I am using stubs and for integration tests we're checking actual state

nkomonen-amazon: From a quick look at MetadataManager, I wonder if we could avoid stubs. Filesystem calls are not expensive for tests and you can just read the actual metadata.json file

packages/core/src/awsService/appBuilder/serverlessLand/main.ts
```diff
@@ -95,20 +93,28 @@ async function launchProjectCreationWizard(
     }).run()
 }
 
-async function downloadPatternCode(config: CreateServerlessLandWizardForm, assetName: string): Promise<void> {
+export async function downloadPatternCode(config: CreateServerlessLandWizardForm, assetName: string): Promise<void> {
     const fullAssetName = assetName + '.zip'
     const location = vscode.Uri.joinPath(config.location, config.name)
     try {
         await getPattern(serverlessLandOwner, serverlessLandRepo, fullAssetName, location, true)
+        telemetry.record({
```
nkomonen-amazon, line 101: What is the metric we are recording to? This makes assumptions that the caller of this function is using a `telemetry.XXXXX.run()`, but we cannot be sure. Unless you have a good reason, we recommend passing the `span` object created by `run()` as an argument and then using `record()` on that

Vandita2020: We're trying to record the `pattern` that user has selected to download. This way we can keep track of the most downloaded pattern, along with the IaC and Runtime.

Vandita2020: This is one of the crucial metrics that we want to record for this project

packages/core/src/test/awsService/appBuilder/serverlessLand/main.test.ts
Vandita2020: The actual `metadata.json` file may change with time and will get updated. I didn't want tests to fail upon any addition/deletion of pattern in the main `metadata.json` file

packages/core/src/awsService/appBuilder/serverlessLand/main.ts
nkomonen-amazon: Ok I see that it is recording to the `aws.toolkit.lambda.createServerlessLandProject` command metric

packages/core/src/test/awsService/appBuilder/serverlessLand/main.test.ts
nkomonen-amazon: What if you just stubbed `getMetadataPath` to point to a custom test file. Then you wouldn't need to stub anything else and it tests all the real code

packages/core/src/shared/telemetry/vscodeTelemetry.json
Vandita2020: I have opened PR ([link](https://github.com/aws/aws-toolkit-common/pull/996)) to add the telemetry to `aws-toolkit-common`

packages/core/src/test/awsService/appBuilder/serverlessLand/wizard.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    let rootNode: sinon.SinonSpiedInstance<AppBuilderRootNode>
+    let sandbox: sinon.SinonSandbox
+
+    beforeEach(async () => {
+        sandbox = sinon.createSandbox()
+        await fs.delete(path.join(workspaceFolder.uri.fsPath, projectFolder), { recursive: true })
+    })
+
+    afterEach(async () => {
+        await fs.delete(path.join(workspaceFolder.uri.fsPath, projectFolder), { recursive: true })
```
vicheey, line 36:
```suggestion
        await fs.delete(path.join(workspaceFolder.uri.fsPath, projectFolder), { recursive: true })
        sandbox.restore()
```

packages/core/src/test/awsService/appBuilder/serverlessLand/main.test.ts
```diff
@@ -0,0 +1,285 @@
...
+        assert.strictEqual(result, undefined)
+    })
+})
+
+describe('downloadPatternCode', () => {
+    let getPatternStub: sinon.SinonStub
+    const assertTelemetry = assertTelemetryCurried('lambda_downloadServerlessLandPattern')
+
+    beforeEach(function () {
+        getPatternStub = sinon.stub(downloadPatterns, 'getPattern')
```
vicheey, line 146: Seems to be mixture of `sinon` and `sandbox` used in this test file.

vicheey (changes requested):

packages/core/src/awsService/appBuilder/serverlessLand/metadataManager.ts
```diff
@@ -62,9 +64,18 @@ export class MetadataManager {
                 const metadataContent = nodefs.readFileSync(metadataPath, { encoding: 'utf-8' })
                 const parseMetadata = JSON.parse(metadataContent) as ProjectMetadata
                 this.metadata = parseMetadata
+                telemetry.record({
+                    action: 'serverlessland_metadataLoaded',
+                    result: 'Succeeded',
```
nkomonen-amazon, line 69: Since this is all wrapped in `telemetry.lambda_createServerlessLandProject.run`, you don't need to add a result since if no errors are thrown it will automatically populate `result: Succeeded`

packages/core/src/awsService/appBuilder/serverlessLand/metadataManager.ts
```diff
@@ -62,9 +64,18 @@ export class MetadataManager {
...
+                })
             }
             return this.metadata
         } catch (err) {
+            telemetry.record({
```
nkomonen-amazon, line 74: Since this is all wrapped in `telemetry.lambda_createServerlessLandProject.run`, you don't need to do any of this. `result`, `reason` will be populated automatically.

All you need to do is record `action` at a higher level, and then it will be a part of the Success and Failure metric

packages/core/src/test/awsService/appBuilder/serverlessLand/wizard.test.ts
```diff
@@ -0,0 +1,132 @@
...
+        sandbox.restore()
+    })
+
+    describe(':) Path', async () => {
```
nkomonen-amazon, line 40: This name isn't that clear. Should it include something with "serverlessland" in it?

packages/core/src/awsService/appBuilder/serverlessLand/main.ts
```diff
@@ -61,9 +56,12 @@ export async function createNewServerlessLandProject(extContext: ExtContext): Pr
             },
             true
         )
+        telemetry.record({
+            runtimeString: config.runtime,
+            templateName: assetName,
+            result: 'Succeeded',
```
nkomonen-amazon, line 62: As mentioned in another comment, dont need to specify result when in a `run()` context

packages/core/src/test/awsService/appBuilder/serverlessLand/main.test.ts
```diff
@@ -0,0 +1,224 @@
...
+        mockExtContext = {
+            awsContext: {
+                getCredentials: sandbox.stub().resolves({}),
+                getCredentialDefaultRegion: () => 'us-west-2',
+            },
+        } as unknown as ExtContext
+
+        mockMetadataManager = sandbox.createStubInstance(MetadataManager)
+        mockMetadataManager.getAssetName.returns('test-asset-name')
+        sandbox.stub(MetadataManager, 'getInstance').returns(mockMetadataManager as unknown as MetadataManager)
```
nkomonen-amazon, line 46: I don't have time to go through all the details. But this feels like we can reduce the amount of stubs needed, just stubbing lower level things like network calls, and then all real code can execute.

Not blocking for this PR due to time constraints, but please keep in mind when designing for testability.

packages/core/src/test/awsService/appBuilder/serverlessLand/main.test.ts
```diff
@@ -0,0 +1,224 @@
...
+    downloadPatternCode,
+} from '../../../../awsService/appBuilder/serverlessLand/main'
+import { fs } from '../../../../shared/fs/fs'
+import * as downloadPatterns from '../../../../shared/utilities/downloadPatterns'
+import { ExtContext } from '../../../../shared/extensions'
+import { workspaceUtils } from '../../../../shared'
+import * as downloadPattern from '../../../../shared/utilities/downloadPatterns'
+import * as wizardModule from '../../../../awsService/appBuilder/serverlessLand/wizard'
+
+describe('createNewServerlessLandProject', () => {
```
nkomonen-amazon, line 29: nit: You can deduplicate code by moving all these `describe()` in to a nested `describe()` and then in the top level you create the sandbox once

vicheey (approved):

packages/core/src/test/awsService/appBuilder/serverlessLand/main.test.ts
Vandita2020: I'll cover the changes in a follow up PR

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -655,6 +655,21 @@
             "name": "aws_hideRegion",
             "description": "Records a call to remove a region from the explorer"
         },
+        {
+            "name": "lambda_createServerlessLandProject",
```
nkomonen-amazon, line 659: It looks like this change can be dropped since it is now in common: https://github.com/aws/aws-toolkit-common/pull/996/files

Vandita2020: Done

nkomonen-amazon (approved):

packages/core/src/test/awsService/appBuilder/serverlessLand/main.test.ts
justinmk3: > I didn't want tests to fail upon any addition/deletion of pattern in the main `metadata.json` file

But that is a useful thing to assert. If the content changes, then the test should change.

packages/core/src/awsService/appBuilder/serverlessLand/metadataManager.ts
```diff
@@ -48,7 +49,7 @@ export class MetadataManager {
     }
 
     public getMetadataPath(): string {
-        return globals.context.asAbsolutePath('dist/src/serverlessLand/metadata.json')
+        return globals.context.asAbsolutePath(path.join('dist', 'src', 'serverlessLand', 'metadata.json'))
```
justinmk3, line 52: ? Why is join() needed here?

PR #6734, opened 2025-03-06
refactor(codecatalyst): migrate cc client to sdkv3 (WIP)

Hweinstock:
## Solution
- refactor `call` wrapper to work with commands, rather than requests. 
- Support passing `token` into clients. 
- `correctClockskew` is enabled by default so it can be dropped: https://github.com/aws/aws-sdk-js-v3/blob/eae65cded5e703306346bdbd1b5de9d23050054a/UPGRADING.md

## Future Work 
- Look into adding the performance logging in the original `call` function to the default middleware stack. 

---

### DIFF
--- a/packages/core/src/shared/clients/codecatalystClient.ts
+++ b/packages/core/src/shared/clients/codecatalystClient.ts
@@ -28,11 +28,12 @@ import {
     ListSourceRepositoriesItem,
     ListSourceRepositoriesItems,
 } from 'aws-sdk/clients/codecatalyst'
+import { CodeCatalystClient as CodeCatalystSDKClient } from '@aws-sdk/client-codecatalyst'
 import { truncateProps } from '../utilities/textUtilities'
 import { SsoConnection } from '../../auth/connection'
 import { DevSettings } from '../settings'
-import { RetryDelayOptions } from 'aws-sdk/lib/config-base'
 import { getServiceEnvVarConfig } from '../vscode/env'
+import { AwsCommand, AwsCommandOutput } from '../awsClientBuilderV3'
 
 export interface CodeCatalystConfig {
     readonly region: string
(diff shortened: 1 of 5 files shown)


packages/core/src/testE2E/codecatalyst/client.test.ts
```diff
@@ -425,15 +429,7 @@ describe('Test how this codebase uses the CodeCatalyst API', function () {
...
-        return await createCodeCatalystClient(conn, undefined, undefined, {
-            // Add retries for tests since many may be running in parallel in github CI.
-            // AWS SDK adds jitter automatically.
-            // https://github.com/aws/aws-sdk-js/blob/3e616251947c73d5239178c167a9d73d985ca581/lib/util.js#L884
-            retryDelayOptions: {
-                base: 1200, // ms
-            },
-            maxRetries: 5,
-        })
+        return await createCodeCatalystClient(conn, undefined, undefined)
```
justinmk3, line 432: > Remove retry logic in tests, since V3 client builder adds this by default.

what are the default values, do they match the removed values here? May be useful to be explicit here in any case.

Hweinstock: Before it made up to 5 attempts with a constant delay. The new default is 5 attempts with exponential backoff implemented here: https://github.com/aws/aws-toolkit-vscode/blob/4c190f4d9678a3bae38ab56000f5ff6123eea6e2/packages/core/src/shared/awsClientBuilderV3.ts#L150

justinmk3 (approved):

packages/core/src/shared/clients/codecatalystClient.ts
```diff
@@ -247,88 +335,70 @@ class CodeCatalystClientInternal {
...
-                    const errNoStack = { ...e }
-                    delete errNoStack.stack
-                    delete errNoStack.requestId // redundant (= "x-amzn-requestid" header).
-
-                    if (r.operation || r.params) {
-                        log.error(
-                            'API request failed (time: %dms): %s\nparams: %O\nerror: %O\nheaders: %O',
-                            timecost,
-                            r.operation,
-                            truncateProps(r.params, 20, ['nextToken']),
```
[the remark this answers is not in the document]
Hweinstock: We have it logging this without the performance and headers information here: https://github.com/aws/aws-toolkit-vscode/blob/4c190f4d9678a3bae38ab56000f5ff6123eea6e2/packages/core/src/shared/awsClientBuilderV3.ts#L202-L207. However, this isn't as detailed and could definitely be improved. I'll track it as a follow-up.

jpinkney-aws (approved):

PR #6735, opened 2025-03-06
telemetry: computeEnv reports "wsl" for all remote envs

justinmk3:
## Problem
1. `computeEnv` reports "wsl" for all remote envs.
2. `computeEnv` does not report "web" envs.

## Solution
1. Only report "wsl" for WSL envs.
2. Report "remote" for other remote envs.
3. Report "web" for unknown web envs.

---

### DIFF
--- a/packages/core/src/shared/telemetry/util.ts
+++ b/packages/core/src/shared/telemetry/util.ts
@@ -257,33 +257,39 @@ export function getUserAgent(
 }
 
 /**
- * All the types of ENVs the extension can run in.
+ * Kinds of machines/environments the extension can run in.
  *
  * NOTES:
- * - append `-amzn` for any environment internal to Amazon
+ * - Append `-amzn` for any environment internal to Amazon.
+ * - Append `-web` for web browser (*without* compute).
  */
 export type EnvType =
     | 'cloud9'
+    | 'cloud9-web'
     | 'cloudDesktop-amzn'
     | 'codecatalyst'
-    | 'local'
     | 'ec2'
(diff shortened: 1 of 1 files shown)


Hweinstock:
duplicate code appears unrelated. (addressed in https://github.com/aws/aws-toolkit-vscode/pull/6737)

Hweinstock (approved):

tomcat323 (approved):

packages/core/src/shared/telemetry/util.ts
```diff
@@ -292,12 +298,16 @@ export async function getComputeEnvType(): Promise<EnvType> {
             return 'ec2-amzn'
         }
         return 'ec2'
-    } else if (env.remoteName) {
+    } else if (env.remoteName === 'wsl') {
         return 'wsl'
     } else if (isAutomation()) {
         return 'test'
     } else if (!env.remoteName) {
```
nkomonen-amazon, line 305: `env.remoteName` is undefined in the typical local desktop use case, so I don't think this will hit the `web` case.

I think we may want to ensure the `local` condition is last, and then add a comment noting nothing should be added after it (Except the `unknown` case)

nkomonen-amazon (changes requested):
See comment

packages/core/src/shared/telemetry/util.ts
justinmk3: thanks for mentioning that :) pushed a fix.

"local" is 2nd-to-last, since "remote" must be last.

nkomonen-amazon (approved):

PR #6742, opened 2025-03-07
feat(amazonq): show code diff instead of only code fix for fix preview

atonaamz:
## Problem

We currently see only code fix for fix preview. We want to see the diff containing both code fix and original code.

## Solution

Remove logic to generate only the code fix. 

Remove logic for line number as it will be different for the diff.

/review: show code diff instead of only code fix for fix preview
---

### DIFF
--- a/packages/amazonq/.changes/next-release/Bug Fix-a7de0f94-cd21-4adc-8fcf-3082aa862528.json
+++ b/packages/amazonq/.changes/next-release/Bug Fix-a7de0f94-cd21-4adc-8fcf-3082aa862528.json
@@ -1,0 +1,4 @@
+{
+	"type": "Bug Fix",
+	"description": "/review: show code diff for fix preview"
+}
--- a/packages/core/src/codewhisperer/views/securityIssue/vue/root.vue
+++ b/packages/core/src/codewhisperer/views/securityIssue/vue/root.vue
@@ -103,7 +103,6 @@ import highSeverity from '../../../../../resources/images/severity-high.svg'
 import criticalSeverity from '../../../../../resources/images/severity-critical.svg'
 import markdownIt from 'markdown-it'
 import hljs from 'highlight.js'
-import { parsePatch } from 'diff'
 import { CodeScanIssue } from '../../../models/model'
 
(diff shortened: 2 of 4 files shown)


packages/core/src/codewhisperer/views/securityIssue/vue/root.vue
```diff
@@ -344,28 +343,11 @@ export default defineComponent({
...
-            const htmlString = md.render(`
-\`\`\`${this.languageId} showLineNumbers startFrom=${oldStart} ${
-                referenceStart && referenceEnd
-                    ? `highlightStart=${referenceStart + 1} highlightEnd=${referenceEnd + 1}`
-                    : ''
-            }
-${this.fixedCode}
+            return md.render(`
+\`\`\`${this.languageId} 
+${this.suggestedFix.replaceAll('--- buggyCode\n', '').replaceAll('+++ fixCode\n', '')}
```
ctlai95, line 348: can you also do this for the hover component?

packages/amazonq/.changes/next-release/Bug Fix-a7de0f94-cd21-4adc-8fcf-3082aa862528.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Bug Fix",
```
ctlai95, line 2: should this be a feature instead?

packages/core/src/codewhisperer/views/securityIssue/vue/root.vue
```diff
@@ -344,28 +343,11 @@ export default defineComponent({
...
 \`\`\`
       `)
-            const parser = new DOMParser()
-            const doc = parser.parseFromString(htmlString, 'text/html')
-            const referenceTracker = doc.querySelector('.reference-tracker')
```
ctlai95, line 361: I think we will need this soon, but it probably won't work with the diff view.

atonaamz: Yes we can add it back once it is needed.

ctlai95 (approved):

hayemaxi (approved):

PR #6772, opened 2025-03-12
telemetry(amazonq): Update /dev error classification logic

neilk-aws:
## Problem

Today, telemetry emitted from the feature development is classified as error, fault, or LLM-based failure across various parts of the code base. This makes it difficult to know all the error types that exist and how to classify them appropriately when emitting metrics.

## Solution
The solution here allows different exceptions to raise from different parts of the code base and uses inheritance to model each error as a client error, service error, or LLM-based error. When emitting telemetry, we now only need to check whether the error received extends from one of these 3 to know how to classify the telemetry. 

---


chengoramazon (approved):

packages/core/src/shared/telemetry/vscodeTelemetry.json
```diff
@@ -52,6 +52,11 @@
             "type": "int",
             "description": "The sum of file sizes that were indexed in MB"
         },
+        {
+            "name": "amazonqVectorIndexEnabled",
```
hayemaxi, line 56: Is there a tracking item to move this to aws-toolkit-common?

packages/core/src/test/amazonqFeatureDev/controllers/chat/controller.test.ts
```diff
@@ -471,22 +471,17 @@ describe('Controller', () => {
...
                     return MetricDataResult.Error
                 }
-                return errorResultMapping.get(error.constructor.name) ?? MetricDataResult.Fault
+                if (error instanceof LlmError) {
+                    return MetricDataResult.LlmFailure
+                }
+                if (error instanceof ServiceError) {
+                    return MetricDataResult.Fault
+                }
+                return MetricDataResult.Fault
```
hayemaxi, line 484: You've already made a similar check in `controller.ts`, move this to a utility function so there isn't duplicate code.

neilk-aws: Good call. Simplified behind a single utility function.

packages/core/src/shared/telemetry/vscodeTelemetry.json
neilk-aws: This is not intended for review. Removed from PR after rebasing.

packages/core/src/amazonqFeatureDev/errors.ts
```diff
@@ -3,140 +3,158 @@
...
+export class NoChangeRequiredException extends ClientError {
     constructor() {
         super(i18n('AWS.amazonq.featureDev.error.noChangeRequiredException'), {
             code: 'NoChangeRequiredException',
         })
     }
 }
 
-export class FeatureDevServiceError extends ToolkitError {
+export class FeatureDevServiceError extends ServiceError {
```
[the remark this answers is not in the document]
neilk-aws: I kept this as a way of distinguishing a top-level service error for "feature dev" only versus any part of Q

packages/core/src/amazonq/errors.ts
```diff
@@ -9,10 +9,37 @@
...
+    constructor(message: string, info: ErrorInformation = {}) {
+        super(message, info)
+    }
+}
+
+/**
+ * Errors extending this class are considered "faults" in service metrics.
+ */
+export class ServiceError extends ToolkitError {
+    constructor(message: string, info: ErrorInformation = {}) {
```
justinmk3, line 27: For both this and `ClientError`, 
- should they have a default code?
- if this concept is truly needed, it should be lifted into core, next to where ToolkitError is defined: https://github.com/aws/aws-toolkit-vscode/blob/43b0bd162831c4eb69bd9c0681885b934101b0db/packages/core/src/shared/errors.ts#L138

justinmk3 (approved):

packages/core/src/amazonq/errors.ts
neilk-aws: I can update them to have a default code (i.e. 400 and 500 respectively) and move them to where ToolkitError is defined. Is that okay as a follow up? 

If so, is it possible to help me merge since I don't have permissions?

neilk-aws:
Duplicate code will be reduced once migrated to Flare

packages/core/src/amazonqFeatureDev/errors.ts
```diff
@@ -3,140 +3,163 @@
...
-export class ApiError extends ToolkitError {
+export class ApiClientError extends ClientError {
+    constructor(message: string, api: string, errorName: string, errorCode: number) {
+        super(message, { code: `${api}-${errorName}-${errorCode}` })
+    }
+}
+
+export class ApiServiceError extends ServiceError {
     constructor(message: string, api: string, errorName: string, errorCode: number) {
         super(message, { code: `${api}-${errorName}-${errorCode}` })
```
justinmk3, line 150: What is the difference between ServiceError/ApiServiceError/ApiError ? Seems very unintuitive.

PR #6778, opened 2025-03-12
fix(amazonq): allow extension activation during amazon q lsp initialization

jpinkney-aws:
## Problem
- Downloading the manifest/language server blocks showing the chat UI, creating a 2+ second increase between the UI showing with the feature flag enabled vs disabled

## Solution
- Don't block extension activation on downloading
    - If any code requires the language server to be started it should live alongside it

---


hayemaxi:
To determine the auth state, we will have to use the LSP. If things activate while it is still initializing, the initial auth state will be incorrect. So we may end up displaying a not logged in state (or a logged in state if we want), but that may not be true until we contact the LSP. This might look weird to the user.

jpinkney-aws:
Sync'd with Max on this. For simplification it makes sense to close this

PR #6784, opened 2025-03-13
ci(changelog): escape changelog for notify, default entry when no changes

hayemaxi:
- Escape the changelog string using `jq` before sending to the notify url. Otherwise, things like quotes might cause the payload to fail.
- If there are no new changes to release, post a generic entry for tracking purposes
  - Informs that the version is uneventful, and clears up the special case of handling this for release.
  - Fixes things that use the changelog, like notify, prerelease, etc. Otherwise, it will use previous version changelog.

---


justinmk3 (approved):
Yay! This was something we wanted for a long time :D

jpinkney-aws (approved):

PR #6791, opened 2025-03-14
telemetry(webview): Emit toolkit_X module telemetry on auth webview

nkomonen-amazon:
## Problem:

With our telemetry, we do not know when the frontend webview UI has actually loaded.

The current process looks like the following:
- We create a webview and set the HTML to load, but after that we do not have a formal way to detect if the webview actually loaded the HTML/JS successfully. We only know that the process started (`toolkit_willOpenModule`)

## Solution:

Emit certain metrics during the webview loading process to get a better idea of if the webview UI successfully completed its initial load.

- `toolkit_willOpenModule`, indicates intent to render a webview. It does not mean the user is seeing anything.
- `toolkit_didLoadModule`, indicates the final result of loading the webview
  - We know a `result: Succeeded` when the frontend send a successful message to the backend. It knows this by ensuring there were no errors and that a certain HTML element can be found, then once the page finishes its initial load it will send a success message to the backend.
  - On `result: Failed`, what happens is a timer has timed out after 10 seconds. We assume that since there was no response from the frontend, it failed to fully execute the HTML/JS.
  - State is shared between `toolkit_willOpenModule` and `toolkit_didLoadModule` so that we can connect them through telemetry. This includes `traceId` and the `duration` which is the time between the 2 metrics.

This PR only applies to the Login and Reauth page for now, and future Vue webviews will need to implement some things on their end to get this functionality.

## TODO
- Generalize this solution in a more robust way for other webviews to easily implement this functionality

---

### DIFF
--- a/packages/core/src/amazonq/webview/messages/messageDispatcher.ts
+++ b/packages/core/src/amazonq/webview/messages/messageDispatcher.ts
@@ -16,6 +16,8 @@ import globals from '../../../shared/extensionGlobals'
 import { openUrl } from '../../../shared/utilities/vsCodeUtils'
 import { DefaultAmazonQAppInitContext } from '../../apps/initContext'
 
+const qChatModuleName = 'amazonqChat'
+
 export function dispatchWebViewMessagesToApps(
     webview: Webview,
     webViewToAppsMessagePublishers: Map<TabType, MessagePublisher<any>>
@@ -29,8 +31,8 @@ export function dispatchWebViewMessagesToApps(
                  * This would be equivalent of the duration between "user clicked open q" and "ui has become available"
                  * NOTE: Amazon Q UI is only loaded ONCE. The state is saved between each hide/show of the webview.
                  */
-                telemetry.webview_load.emit({
-                    webviewName: 'amazonq',
+                telemetry.toolkit_didLoadModule.emit({
+                    module: qChatModuleName,
                     duration: performance.measure(amazonqMark.uiReady, amazonqMark.open).duration,
                     result: 'Succeeded',
                 })
@@ -86,12 +88,19 @@ export function dispatchWebViewMessagesToApps(
         }
 
         if (msg.type === 'error') {
-            const event = msg.event === 'webview_load' ? telemetry.webview_load : telemetry.webview_error
-            event.emit({
-                webviewName: 'amazonqChat',
-                result: 'Failed',
-                reasonDesc: msg.errorMessage,
-            })
+            if (msg.event === 'toolkit_didLoadModule') {
+                telemetry.toolkit_didLoadModule.emit({
+                    module: qChatModuleName,
+                    result: 'Failed',
+                    reasonDesc: msg.errorMessage,
+                })
+            } else {
+                telemetry.webview_error.emit({
+                    webviewName: qChatModuleName,
+                    result: 'Failed',
+                    reasonDesc: msg.errorMessage,
+                })
+            }
             return
         }
 
--- a/packages/core/src/webviews/main.ts
+++ b/packages/core/src/webviews/main.ts
@@ -7,6 +7,10 @@ import * as vscode from 'vscode'
 import { Protocol, registerWebviewServer } from './server'
 import { getIdeProperties } from '../shared/extensionUtilities'
 import { getFunctions } from '../shared/utilities/classUtils'
+import { telemetry } from '../shared/telemetry/telemetry'
+import { randomUUID } from '../shared/crypto'
+import { Timeout } from '../shared/utilities/timeoutUtils'
+import globals from '../shared/extensionGlobals'
 
 interface WebviewParams {
     /**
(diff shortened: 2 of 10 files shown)


packages/core/src/amazonq/webview/messages/messageDispatcher.ts
```diff
@@ -29,8 +31,8 @@ export function dispatchWebViewMessagesToApps(
                  * This would be equivalent of the duration between "user clicked open q" and "ui has become available"
                  * NOTE: Amazon Q UI is only loaded ONCE. The state is saved between each hide/show of the webview.
                  */
-                telemetry.webview_load.emit({
```
Hweinstock, line 32: Once this is done for all the webviews, will `webview_load` be deprecated in favor of the more granular metrics?

packages/core/src/amazonq/webview/messages/messageDispatcher.ts
```diff
@@ -86,12 +88,19 @@ export function dispatchWebViewMessagesToApps(
...
 
         if (msg.type === 'error') {
-            const event = msg.event === 'webview_load' ? telemetry.webview_load : telemetry.webview_error
-            event.emit({
-                webviewName: 'amazonqChat',
-                result: 'Failed',
-                reasonDesc: msg.errorMessage,
-            })
+            if (msg.event === 'toolkit_didLoadModule') {
+                telemetry.toolkit_didLoadModule.emit({
```
Hweinstock, line 92: Why does the `webview_error` metric exist? Shouldn't it in theory mirror the failures of `toolkit_didLoadModule` or are there non-error reasons a webview fails to load?

If we have a separate metric for error, shouldn't we emit it in this case since we still received an error from the webview?

packages/core/src/webviews/main.ts
```diff
@@ -225,12 +239,38 @@ export abstract class VueWebview {
...
                 })
             }
 
+            /**
+             * Setup telemetry events that report on the initial loading of a webview
+             */
+            private setupTelemetry() {
+                this.instance.traceId = randomUUID()
+                // Notify intent to open a module, this does not mean it successfully opened
+                telemetry.toolkit_willOpenModule.emit({
```
Hweinstock, line 256: Is there a case where we intend to open a module (emit `willOpenModule`) then don't also emit a `didLoadModule` with either fail or succeed?

packages/core/src/amazonq/webview/messages/messageDispatcher.ts
nkomonen-amazon: I'm thinking of dropping `webview_error` and creating something like `toolkit_moduleError`. This will capture any errors that happen **after** loading has happened, anything before would be captured in `toolkit_didLoadModule`.

I have it as a TODO to deprecate `webview_error`, this will also allow us to deal with different field names (`module` vs `webviewName`)

packages/core/src/amazonq/webview/messages/messageDispatcher.ts
nkomonen-amazon: Yup, `webview_x` will be deprecated for something like `toolkit_moduleX`. So we'll also drop `webview_error` as well

packages/core/src/webviews/main.ts
nkomonen-amazon: For now, every webview will emit a `willOpenModule` and `didLoadModule` will need to be explicitly done by each webview.

This is due to how loading a webview works. We can only indicate our intent to open a webview but have no formal way to know when it has opened (we create a vscode webview instance and set a string of HTML, then have no insight to what happens after that)

So there will be cases where we only have `willOpenModule` and no trailing `didLoadModule`. This is essentially how we have it right now, but under different metric names.

packages/core/src/amazonq/webview/messages/messageDispatcher.ts
justinmk3: > `toolkit_moduleError`. This will capture any errors that happen **after** loading has happened, anything before would be captured in `toolkit_didLoadModule`.

? errors should be part of all metrics. There should not be a separate "foo_error" metric.

packages/core/src/amazonq/webview/messages/messageDispatcher.ts
```diff
@@ -86,12 +88,19 @@ export function dispatchWebViewMessagesToApps(
...
+                    module: qChatModuleName,
+                    result: 'Failed',
+                    reasonDesc: msg.errorMessage,
+                })
+            } else {
+                telemetry.webview_error.emit({
+                    webviewName: qChatModuleName,
+                    result: 'Failed',
+                    reasonDesc: msg.errorMessage,
```
justinmk3, line 101: why is there a separate `webview_error` metiric? The error should be part of the `toolkit_didLoadModule` metric.

packages/core/src/webviews/main.ts
```diff
@@ -151,6 +155,35 @@ export abstract class VueWebview {
      */
     public abstract readonly id: string
 
+    /**
+     * Implementations must override this field to `true` if the webview can support
+     * {@link telemetry.toolkit_didLoadModule}.
+     * A webview that supports this will call {@link setDidLoad}
+     * to confirm the UI has successfully loaded.
+     */
+    public supportsLoadTelemetry: boolean = false
```
justinmk3, line 164: why do we need this flag? can we just try-and-handle-failure instead?

packages/core/src/webviews/main.ts
justinmk3: In most cases, did/will pairs should not be needed. Only the "did" case is needed for most metrics, because the metric will wrap the impl logic and track a `duration`. Why do we need both here?

justinmk3 (approved):

packages/core/src/amazonq/webview/messages/messageDispatcher.ts
nkomonen-amazon: `didLoadModule` is mainly intended for the initial load, but if there is an error post-load then we will want a separate metric for that. Example is after clicking the "submit" button after putting in the startUrl+Region for signin

PR #6793, opened 2025-03-14
fix(amazonq): update the bearer token in the language server every 5 minutes

jpinkney-aws:
## Problem
- When the bearer token updates in the background we never re-notify the language server

## Solution
- update it every 5 minutes as a hack to unblock the pentest

---


hayemaxi (approved):
Should dispose of it on deactivation/exception. Might cause tests or other things to hang later.

jpinkney-aws:
good point, just updated it

PR #6798, opened 2025-03-17
fix(cw): allow retries in the cw streaming sdk client. 

Hweinstock:
## Problem
This client is used in many Q features and does not currently allow retries. This means a single SDK request failure will cause the operation to fail. 

It appears based on the code that this is not intentional, and was accidentally left unfinished. 

## Solution
- allow up to 1 retry.

---


justinmk3 (approved):
Start with 2 or 3 instead of 5?

jpinkney-aws (approved):
Still need someone from the codewhisperer team to approve this but the original reason why we disabled this should be gone

justinmk3 (approved):

justinmk3:
btw, is the choice of 5 retries here intentional? default is 3. to use the default we could pass undefined: https://github.com/aws/aws-toolkit-vscode/blob/b1fae85369d794172d730cc576fe27a41b6c1778/packages/core/src/shared/awsClientBuilderV3.ts#L159

PR #6812, opened 2025-03-19
feat(amazonq): add open chat command

hayemaxi:
There is no hard command to focus the chat view from command palette. There exists Amazon Q: Focus on Chat View, but if the user moves the chat it can change names, e.g. Terminal: Focus on Chat View, which is confusing.
CoPilot also has this issue, which is why they have the Chat: Open Chat command.

Note: This command will work whether the user is logged in or not, for ease of use. If they are not logged in then it will open the login window.

---


justinmk3 (approved):
Make sense, thanks for the explanation :)

jpinkney-aws (approved):

PR #6813, opened 2025-03-19
refactor(zip): unify different methods for zipping project. (WIP)

Hweinstock:
## Problem
Both [prepareRepoData](https://github.com/aws/aws-toolkit-vscode/blob/3e42f815da70671e6cd607fff4b2778c10d36591/packages/core/src/amazonq/util/files.ts#L57) and [zipUtil](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/util/zipUtil.ts#L54) are ways to take your current workspace and package it up as a zip. /doc and /dev use prepareRepoData. /review and /test use zipUtil. Their implementations are pretty similar and we should be able to have a shared utility that is a consolidated version of the two. 

Both implementations are highly coupled to their use case, but more so `zipUtil`. Additionally, `zipUtil` uses `admZip`, a non-web compatible dependency. 

## GOAL
- Generalize project zipping functionality such that it lives in one place. 
- Remove dependency on `admZip` in `zipUtil`. 
- Increase testing coverage for the methods themselves, and for their components (by splitting them into testable components). 
- Decouple zip utility from Q agents. 

## Solution
- move `prepareRepoData` tests to live in `core` module since thats where `prepareRepoData` lives. Do the same with `zipUtil`. 
- [in progress...]

---


packages/core/src/shared/utilities/zipStream.ts
```diff
@@ -104,31 +103,44 @@ export class ZipStream {
...
+     * Writes data to the specified path.
+     * @param data
+     * @param path
+     * @param returnPromise optional parameter specifying if caller wants a promise.
+     * @returns promise to that resolves when data is written.
+     */
+    public writeString(data: string, path: string, returnPromise: true): Promise<EntryMetaData>
+    public writeString(data: string, path: string, returnPromise?: false): void
+    public writeString(data: string, path: string, returnPromise?: boolean): void | Promise<EntryMetaData> {
+        const promise = this._zipWriter.add(path, new TextReader(data))
```
jpinkney-aws, line 116: Whats the use case for the promise/non promise response vs just always returning a promise and then callers can either do await or void

packages/core/src/codewhisperer/util/zipUtil.ts
```diff
@@ -412,39 +410,56 @@ export class ZipUtil {
...
-                }
-                await this.processBinaryFile(zip, file.fileUri, zipEntryPath)
-            } else {
-                const isFileOpenAndDirty = this.isFileOpenAndDirty(file.fileUri)
-                const fileContent = isFileOpenAndDirty ? await this.getTextContent(file.fileUri) : file.fileContent
-                this.processTextFile(zip, file.fileUri, fileContent, languageCount, zipEntryPath)
+        const collectFilesOptions: CollectFilesOptions = {
+            maxTotalSizeBytes: this.getProjectScanPayloadSizeLimitInBytes(),
+            excludePatterns:
+                useCase === FeatureUseCase.TEST_GENERATION
```
jpinkney-aws, line 424: Do you think it would be hard to eventually get rid of these useCase conditions?

jpinkney-aws (commented):
Code is looking pretty good so far 🚀 

FWIW for:
> move prepareRepoData tests to live in core module since thats where prepareRepoData lives. Do the same with zipUtil.

those are probably eventually going to be moved back to amazon q anyways, once we deprecate the current implementation, since the only common thing between toolkit/q at that point will be _maybe_ the lsp downloading?

packages/core/src/shared/utilities/zipStream.ts
Hweinstock: My understanding of the `ZipStream` code is that it will wait until all files are done streaming in the `finalize` step. Therefore the most common use case is to avoid awaiting the promises individually so that we aren't waiting on IO operations until all files are added. Its only the tests where we await these promises.

packages/core/src/codewhisperer/util/zipUtil.ts
Hweinstock: Yea, this is what I am struggling with now. These enums are everywhere in `ZipUtil` and make it difficult to pull out general components.

Hweinstock:
> those are probably eventually going to be moved back to amazon q anyways, once we deprecate the current implementation,
Once the agents live in flare, it can be removed completely right?

PR #6816, opened 2025-03-19
deps(ziputil): migrate ziputil to use a web-safe and typed ZipStream

Hweinstock:
## Problem
This PR solves two very related problems so there are done together in one PR: 

1) ZipUtil uses `admZip` which not web compatible. Also, the larger issue is that `ZipUtil` duplicates `prepareRepoData` in some ways, and it is difficult to align their implementations if they use different libraries. 

2) The [ZipStream](https://github.com/aws/aws-toolkit-vscode/blob/bd378854550362810170a4d9164c426e622de256/packages/core/src/shared/utilities/zipStream.ts#L51) utility is a wrapper around the `zip.js` library. However, it imports it via `CommonJS` `require` syntax causing us to lose type information. This is dangerous because we are assuming properties and functions exist within the library, and are blind to changes in the underlying library. This makes the `ZipStream` module difficult and unsafe to work with.

## Solution
- For Problem 1: 
   - migrate `ZipUtil` away from `admZip`. 
   - move `ZipUtil` tests to live in `core` package since `ZipUtil` lives there. 
   - add some utility functions to `ZipStream` + tests for them. 
- For Problem 2:
   - Import `zip.js` as a module and tell the compiler to ignore it. It looks like this happens because `zip.js` doesn't support CommonJS, and in general doesn't support Node officially (source: https://github.com/gildas-lormeau/zip.js/issues/362). However, a future commit appears to add this support (on that same issue) so unsure why we still get the error. Experimentally this is working, so its likely Webpack handles this for us. 
   - This requires updating some type signatures and also reveals some slight bugs in the implementation. 

also addresses: https://github.com/aws/aws-toolkit-vscode/issues/6566

## Verification
- Used /review on a midsize project and it produced the same results on this and previous release (without changes). 

---


justinmk3 (approved):

packages/core/src/shared/utilities/zipStream.ts
```diff
@@ -148,14 +160,23 @@ export class ZipStream {
...
-    public static async unzip(zipBuffer: Buffer): Promise<ZipReaderResult[]> {
+    public async finalizeToFile(targetPath: string, onProgress?: (percentComplete: number) => void) {
+        const result = await this.finalize(onProgress)
+        const contents = result.streamBuffer.getContents() || Buffer.from('')
+        await fs.writeFile(targetPath, contents)
+        return result
+    }
+
+    public static async unzip(zipBuffer: Buffer): Promise<Entry[]> {
         const reader = new ZipReader(new Uint8ArrayReader(new Uint8Array(zipBuffer)))
```
justinmk3, line 180: off topic: this doesn't actually "stream" the result, it all goes into memory, right? that's something we need to address eventually.

justinmk3 (approved):

packages/core/src/shared/utilities/zipStream.ts
Hweinstock: I noticed this as well. Also, the current logic in zipUtils dumps this to the file system before uploading whereas `prepareRepoData` uploads the buffer itself. 

If we can centralize the uploading logic (also has two completely separate implementations), this streaming change will be much easier.

jpinkney-aws (approved):

Hweinstock:
https://github.com/aws/aws-toolkit-vscode/issues/6566 is aparently still an issue since the fix only lives in `/doc`, `/dev` implementations. Need to move the fix to `ZipStream`.

jpinkney-aws (approved):

avi-alpert (approved):

PR #6831, opened 2025-03-21
feat(amazonq): Add code symbol support in context selection of Q Chat

leigaol:
## Problem

We can select @file and @folder in the Q chat menu. But we should be able to also let user select class, function and these symbols.

## Solution

Support @code in the context selection of Q chat. 

Screenshot 2025-03-21 at 2 15 34 PM

---


packages/core/src/amazonq/lsp/types.ts
```diff
@@ -88,12 +88,44 @@ export const GetIndexSequenceNumberRequestType: RequestType<GetRepomapIndexJSONR
...
+    column: number
+}
+export interface Span {
+    start: Position
+    end: Position
+}
+
+// LSP definition of DocumentSymbol
+
+export interface DocumentSymbol {
```
jpinkney-aws, line 117: Can we reuse this type from `vscode-languageserver-node` or `vscode-languageserver-types`?

jpinkney-aws (approved):
I'm excited to use this feature!

packages/core/src/amazonq/lsp/types.ts
leigaol: Unfortunately no.

andrewyuq (approved):

leigaol:
This does not require review from Flare team

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -951,13 +981,13 @@ export class ChatController {
         }
         triggerPayload.workspaceRulesCount = workspaceRules.length
 
-        // Add context commands added by user to context
         for (const context of triggerPayload.context) {
             if (typeof context !== 'string' && context.route && context.route.length === 2) {
                 contextCommands.push({
                     workspaceFolder: context.route[0] || '',
-                    type: context.icon === 'folder' ? 'folder' : 'file',
+                    type: (context.label || '') as ContextCommandItemType,
```
jguoamz, line 988: I think this broke @prompt

PR #6836, opened 2025-03-24
refactor(auth): dont manual refresh auth webview on hide

nkomonen-amazon:
## Problem:

We have an explicit force refresh of the auth webview when it is shown/hidden. There is not an obvious reason why we do this.

This causes problems since we expect that show/hiding of the webview does not trigger any functional changes since we have configured the webview to continue running even while hidden. This means that our handlers for auth changes are still running even when auth is hidden.

For my specific issue, by force reloading the webview on every show/hide it triggered the webview loading code each time, causing webview telemetry events to be triggered more than expected.

## Solution:

Remove the force reload line. With this change we can move the edge case handling code which throttled the telemetry emitted when it was triggered on every hide/show.

I've done manual testing and have not noted any obvious regressions to the expected user flow.

---

### DIFF
--- a/packages/core/src/login/webview/commonAuthViewProvider.ts
+++ b/packages/core/src/login/webview/commonAuthViewProvider.ts
@@ -124,8 +124,6 @@ export class CommonAuthViewProvider implements WebviewViewProvider {
             }
 
             this.onDidChangeVisibility?.fire(webviewView.visible)
-            // force webview to reload
-            await vscode.commands.executeCommand('workbench.action.webview.reloadWebviewAction')
         })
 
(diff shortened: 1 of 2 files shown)


jpinkney-aws (approved):
I wonder if we originally had this because our super old auth flow needed a way to be reset if the webview was closed/opened? For some reason I vaguely remember something about that. Regardless its probably not relevant now

nkomonen-amazon:
@jpinkney-aws that sounds accurate. Right now we have some event listeners that will force the UI to refresh on change of the auth connection, and separately if the user happens to log in/out we will change from the auth to chat view regardless of the explicit webview refresh. So as you said it doesn't seem relevant anymore!

PR #6837, opened 2025-03-24
feat(amazonq): handle reference for inline with language server

zixlin7:
## Problem
need to log reference for inline with language server

## Solution

---

### DIFF
--- a/packages/amazonq/src/app/inline/completion.ts
+++ b/packages/amazonq/src/app/inline/completion.ts
@@ -14,12 +14,22 @@ import {
     commands,
     languages,
     Disposable,
+    window,
+    TextEditor,
 } from 'vscode'
 import { LanguageClient } from 'vscode-languageclient'
-import { LogInlineCompletionSessionResultsParams } from '@aws/language-server-runtimes/protocol'
+import {
+    InlineCompletionItemWithReferences,
+    LogInlineCompletionSessionResultsParams,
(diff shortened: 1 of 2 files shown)


packages/amazonq/src/app/inline/completion.ts
```diff
@@ -181,6 +181,11 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
                     session.firstCompletionDisplayLatency,
                 ],
             }
+            ReferenceInlineProvider.instance.setInlineReference(
```
jpinkney-aws, line 184: I'm not entirely familiar with the flow but is it something like this:
- user gets inline suggestion
   - it may have a reference attached
- if it does have a reference then show it as a code lens
- if they click the codelens to add a code reference it adds it in the "Code reference log" panel and opens it up

Is that mostly correct?

jpinkney-aws:
Also, is this dependent on any flare changes?

packages/amazonq/src/app/inline/completion.ts
zixlin7: yes that is right, and I realized that I missed the logging reference on acceptance piece, just added it. Thanks for catching!

zixlin7:
> Also, is this dependent on any flare changes?

No, flare already has references support since beginning

jpinkney-aws (approved):

PR #6846, opened 2025-03-25
feat(amazonq): Adding openDiff and acceptDiff UX to the agentic chat.

laileni-aws:
## Problem
- Agentic chat does not view the generated code diff and no option for accept or reject the code diff
- Splitting this bulk PR: https://github.com/aws/aws-toolkit-vscode/pull/6843

## Solution
- Adding UX improvements and agent loop mechanism to Q chat.

https://github.com/user-attachments/assets/81ab88b3-ffb9-4db7-a38f-e0562848cb6f

---
### TODO:
- Metrics
- Refactoring code
- More UX improvements
- Pending part 2 of https://github.com/aws/aws-toolkit-vscode/pull/6843

---

### DIFF
--- a/packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
+++ b/packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
@@ -3,7 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { ChatItemButton, ChatItemFormItem, ChatItemType, MynahUIDataModel, QuickActionCommand } from '@aws/mynah-ui'
+import {
+    ChatItem,
+    ChatItemButton,
+    ChatItemFormItem,
+    ChatItemType,
+    MynahUIDataModel,
+    QuickActionCommand,
+} from '@aws/mynah-ui'
 import { TabType } from '../storages/tabsStorage'
 import { CWCChatItem } from '../connector'
 import { BaseConnector, BaseConnectorProps } from './baseConnector'
@@ -18,12 +25,14 @@ export interface ConnectorProps extends BaseConnectorProps {
         title?: string,
         description?: string
     ) => void
+    onChatAnswerUpdated?: (tabID: string, message: ChatItem) => void
 }
 
 export class Connector extends BaseConnector {
     private readonly onCWCContextCommandMessage
     private readonly onContextCommandDataReceived
     private readonly onShowCustomForm
+    private readonly onChatAnswerUpdated
 
     override getTabType(): TabType {
         return 'cwc'
@@ -34,6 +43,7 @@ export class Connector extends BaseConnector {
         this.onCWCContextCommandMessage = props.onCWCContextCommandMessage
         this.onContextCommandDataReceived = props.onContextCommandDataReceived
         this.onShowCustomForm = props.onShowCustomForm
+        this.onChatAnswerUpdated = props.onChatAnswerUpdated
     }
 
     onSourceLinkClick = (tabID: string, messageId: string, link: string): void => {
@@ -91,16 +101,14 @@ export class Connector extends BaseConnector {
                 messageId: messageData.messageID ?? messageData.triggerID,
                 body: messageData.message,
                 followUp: followUps,
-                canBeVoted: true,
+                canBeVoted: messageData.canBeVoted ?? false,
                 codeReference: messageData.codeReference,
                 userIntent: messageData.userIntent,
                 codeBlockLanguage: messageData.codeBlockLanguage,
                 contextList: messageData.contextList,
-            }
-
-            // If it is not there we will not set it
-            if (messageData.messageType === 'answer-part' || messageData.messageType === 'answer') {
-                answer.canBeVoted = true
+                title: messageData.title,
+                buttons: messageData.buttons ?? undefined,
+                fileList: messageData.fileList ?? undefined,
             }
 
             if (messageData.relatedSuggestions !== undefined) {
@@ -137,6 +145,8 @@ export class Connector extends BaseConnector {
                               options: messageData.followUps,
                           }
                         : undefined,
+                buttons: messageData.buttons ?? undefined,
+                canBeVoted: messageData.canBeVoted ?? false,
             }
             this.onChatAnswerReceived(messageData.tabID, answer, messageData)
 
@@ -204,7 +214,7 @@ export class Connector extends BaseConnector {
         }
 
         if (messageData.type === 'customFormActionMessage') {
-            this.onCustomFormAction(messageData.tabID, messageData.action)
+            this.onCustomFormAction(messageData.tabID, messageData.messageId, messageData.action)
             return
         }
         // For other message types, call the base class handleMessageReceive
@@ -235,6 +245,7 @@ export class Connector extends BaseConnector {
 
     onCustomFormAction(
         tabId: string,
+        messageId: string,
         action: {
             id: string
             text?: string | undefined
@@ -248,14 +259,53 @@ export class Connector extends BaseConnector {
         this.sendMessageToExtension({
             command: 'form-action-click',
             action: action,
+            formSelectedValues: action.formItemValues,
             tabType: this.getTabType(),
             tabID: tabId,
         })
+
+        if (!this.onChatAnswerUpdated || !['accept-code-diff', 'reject-code-diff'].includes(action.id)) {
+            return
+        }
+        const answer: ChatItem = {
+            type: ChatItemType.ANSWER,
+            messageId: messageId,
+            buttons: [],
+        }
+        switch (action.id) {
+            case 'accept-code-diff':
+                answer.buttons = [
+                    {
+                        keepCardAfterClick: true,
+                        text: 'Accepted code',
+                        id: 'accepted-code-diff',
+                        status: 'success',
+                        position: 'outside',
+                        disabled: true,
+                    },
+                ]
+                break
+            case 'reject-code-diff':
+                answer.buttons = [
+                    {
+                        keepCardAfterClick: true,
+                        text: 'Rejected code',
+                        id: 'rejected-code-diff',
+                        status: 'error',
+                        position: 'outside',
+                        disabled: true,
+                    },
+                ]
+                break
+            default:
+                break
+        }
+        this.onChatAnswerUpdated(tabId, answer)
     }
 
     onFileClick = (tabID: string, filePath: string, messageId?: string) => {
         this.sendMessageToExtension({
-            command: 'file-click',
+            command: messageId === '' ? 'file-click' : 'open-diff',
             tabID,
             messageId,
             filePath,
--- a/packages/core/src/amazonq/webview/ui/main.ts
+++ b/packages/core/src/amazonq/webview/ui/main.ts
@@ -352,6 +352,7 @@ export const createMynahUI = (
                     ...(item.followUp !== undefined ? { followUp: item.followUp } : {}),
                     ...(item.fileList !== undefined ? { fileList: item.fileList } : {}),
                     ...(item.header !== undefined ? { header: item.header } : { header: undefined }),
+                    ...(item.buttons !== undefined ? { buttons: item.buttons } : { buttons: undefined }),
                 })
                 if (
                     item.messageId !== undefined &&
@@ -374,7 +375,7 @@ export const createMynahUI = (
                     fileList: {
                         fileTreeTitle: '',
                         filePaths: item.contextList.map((file) => file.relativeFilePath),
-                        rootFolderTitle: 'Context',
+                        rootFolderTitle: item.title,
                         flatList: true,
                         collapsed: true,
                         hideFileCount: true,
--- a/packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
+++ b/packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
@@ -18,11 +18,11 @@ import { UserWrittenCodeTracker } from '../../../../codewhisperer/tracker/userWr
 export class ChatSession {
     private sessionId?: string
     /**
-     * _listOfReadFiles = list of files read from the project to gather context before generating response.
+     * _readFiles = list of files read from the project to gather context before generating response.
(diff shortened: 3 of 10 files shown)


ctlai95:
Can you please update the title to be more specific to this change?

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
```diff
@@ -248,14 +259,53 @@ export class Connector extends BaseConnector {
...
+                        status: 'error',
+                        position: 'outside',
+                        disabled: true,
+                    },
+                ]
+                break
+            default:
+                break
+        }
+        this.onChatAnswerUpdated(tabId, answer)
```
ctlai95, line 303: I feel like this shouldn't live in this function. Is there somewhere else we can move this to?

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -374,7 +375,7 @@ export const createMynahUI = (
                     fileList: {
                         fileTreeTitle: '',
                         filePaths: item.contextList.map((file) => file.relativeFilePath),
-                        rootFolderTitle: 'Context',
+                        rootFolderTitle: item.title,
```
ctlai95, line 378: Maybe fallback to `?? 'Context'`

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
```diff
@@ -248,14 +259,53 @@ export class Connector extends BaseConnector {
...
     }
 
     onFileClick = (tabID: string, filePath: string, messageId?: string) => {
         this.sendMessageToExtension({
-            command: 'file-click',
+            command: messageId === '' ? 'file-click' : 'open-diff',
```
ctlai95, line 308: This feels a bit hacky. Can there be a separate onDiff handler?

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
```diff
@@ -235,6 +245,7 @@ export class Connector extends BaseConnector {
 
     onCustomFormAction(
         tabId: string,
+        messageId: string,
```
ctlai95, line 248: Should it be `string | undefined`? Then you don't need to fallback to `''`

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
laileni-aws: This is existing way how the buttons work/setup in both VSC and JB.
We can create a function and return `answer` but basically the functionality does not change.

packages/core/src/amazonq/webview/ui/main.ts
laileni-aws: `Context` is not a generic text across Q so, I feel this should be either undefined or we can supply proper text as part of `title` which we do in our changes.

packages/core/src/codewhispererChat/view/connector/connector.ts
```diff
@@ -110,6 +110,12 @@ export interface CodeReference {
     }
 }
 
+export interface FileList {
```
ctlai95, line 113: Can't we use the type from `ChatItemContent`?

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
laileni-aws: There used to be onOpenDiff function previously but this got deprecated
>@deprecated since version 4.6.3. Will be dropped after version 5.x.x. Use {@link onFileClick} instead

So Right now if user click on any file within fileTreeDiff Mynah sends event through onFileClick.

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
```diff
@@ -35,6 +38,28 @@ export class ChatSession {
...
+        this._filePath = filePath
+    }
+    public setTempFilePath(tempFilePath: string | undefined) {
+        this._tempFilePath = tempFilePath
+    }
+    public pushToListOfReadFiles(filePath?: string) {
+        if (filePath) {
+            this._listOfReadFiles.push(filePath)
+        } else {
+            this._listOfReadFiles = []
```
ctlai95, line 60: Why does this reset the list? Looks like it would behave unexpectedly if I just look at the name of the function and not the implementation

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
laileni-aws: This wont be undefined and this is for `onCustomFormAction` not for `onFileClick`. Can you elaborate more on this?

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
```diff
@@ -35,6 +38,28 @@ export class ChatSession {
     public setSessionID(id?: string) {
         this.sessionId = id
     }
+    public get listOfReadFiles(): string[] {
+        return this._listOfReadFiles
+    }
+    public get getFilePath(): string | undefined {
```
ctlai95, line 44: This is already a getter function, `get` in the name is redundant

```suggestion
    public get filePath(): string | undefined {
```

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
laileni-aws: If user tries to ask any other question in between generation or if user gets any error we need to reset to files read 
To show proper UX of 
`2 files read, reading Sample.java`
`3 files read, reading foo.java`

etc..

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
```diff
@@ -13,6 +13,9 @@ import { UserWrittenCodeTracker } from '../../../../codewhisperer/tracker/userWr
 
 export class ChatSession {
     private sessionId?: string
+    private _listOfReadFiles: string[] = []
```
ctlai95, line 16: `list` is redundant

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
```diff
@@ -13,6 +13,9 @@ import { UserWrittenCodeTracker } from '../../../../codewhisperer/tracker/userWr
...
+    private _filePath: string | undefined
+    private _tempFilePath: string | undefined
```
ctlai95, line 18: Please check the naming or add some documentation, it's very unclear what these are used for

packages/core/src/codewhispererChat/view/connector/connector.ts
laileni-aws: Yeah actually we can use `ChatItemContent['fileList']`

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
```diff
@@ -35,6 +38,28 @@ export class ChatSession {
     public setSessionID(id?: string) {
         this.sessionId = id
     }
+    public get listOfReadFiles(): string[] {
+        return this._listOfReadFiles
```
ctlai95, line 42: Actually I think we can get this information from chatHistory?

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
laileni-aws: Required for the `Reading files UX` which still a TODO. I can revisit on this once we have final UX into place.

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
ctlai95: Sure but that shouldn't happen here. Someone reading the code in the future is not going to know why it's being reset

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
laileni-aws: Please can you point me the code where this was implemented?

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
laileni-aws: I will add some comments on top of this variable. There might be some duplication due to lot of dependencies on our PR's.
I will add a TODO to revisit and remove necessary variables if required.

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
laileni-aws: Let me write a separate method to reset the files.

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
ctlai95: Sorry I didn't understand, aren't you passing in an empty string in connector.ts?

```ts
                    this.cwChatConnector.onCustomFormAction(tabId, messageId ?? '', action)
```

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
ctlai95: I meant that the variable name can just be `_readFiles`

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
ctlai95: But this logic is specific to diffs which I don't think belongs here. Why not create a separate diff clicked handler? I see we already have some custom logic in https://github.com/aws/aws-toolkit-vscode/blob/011654de785724abcb7ed4569f1947cddfc1e243/packages/core/src/amazonq/webview/ui/connector.ts#L707, can't we just add to that?

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
laileni-aws: Ohh I got it,
I am passing `''` as part of messageId because of this [code](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/amazonq/webview/ui/main.ts#L326-L341)

I tested passing undefined and got undesirable result. So, I am passing `''`
I remember testing this for RIV for `/test` too but I will give a shot once again after UX integration.

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
ctlai95: I think we need to think about this more. It's not clear to me why if messageId is empty that it should be `file-click` and not `open-diff`

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
laileni-aws: This logic is not specific to diff's we can extend this to feedback button and even for `Confirm` button!
Here we show a UX to disable the button once user click on this button.

packages/core/src/amazonq/webview/ui/main.ts
ctlai95: Got it, the reason I'm suggesting this is because this code is shared with all agents. So unless we are sure that we always pass "Context" in the title for all messages everywhere, there might be a UI regression.

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
```diff
@@ -48,6 +56,27 @@ export class ChatSession {
     public setSessionID(id?: string) {
         this.sessionId = id
     }
+    public get listOfReadFiles(): string[] {
```
ctlai95, line 59:
```suggestion
    public get readFiles(): string[] {
```

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -389,6 +397,45 @@ export class ChatController {
...
     }
 
+    private async processOpenDiff(message: OpenDiff) {
+        const session = this.sessionStorage.getSession(message.tabID)
+        const filePath = session.filePath ?? message.filePath
+        const fileExists = await fs.existsFile(filePath)
+        // Check if fileExists=false, If yes, return instead of showing broken diff experience.
+        if (!session.tempFilePath) {
+            return
+        }
```
mr-lee, line 407: does this notify the user?

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
```diff
@@ -17,6 +17,14 @@ import { UserWrittenCodeTracker } from '../../../../codewhisperer/tracker/userWr
 
 export class ChatSession {
     private sessionId?: string
+    /**
+     * _listOfReadFiles = list of files read from the project to gather context before generating response.
```
mr-lee: why do you need to pass this around?

mr-lee (approved):
Approved to unblock.

packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts
laileni-aws: So, basically the question is, How can we differentiate the ContextList vs FileListTree ?
image

packages/core/src/codewhispererChat/controllers/chat/controller.ts
laileni-aws: No It wont notify the user for now, but good to have a Information notification message. I will add this to the TODO.

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
laileni-aws: To count # of files read as per UX requirement.

packages/core/src/amazonq/webview/ui/main.ts
laileni-aws: Got it, right now only Q chat uses `Context` or contextList in project falcon but going forward we need to pass title from the agent controller.ts instead of static text.

PR #6848, opened 2025-03-25
feat(chat): Add executeBash tool for Amazon Q Agentic Chat

tsmithsz:
## Problem
Amazon Q Agentic Chat needs the ability to execute bash commands

## Testing

```
ExecuteBash Tool
extensionHostProcess.js:178
    ✔ pass validation for a safe command (read-only)
extensionHostProcess.js:178
    ✔ fail validation if the command is empty
extensionHostProcess.js:178
    ✔ set requiresAcceptance=true if the command has dangerous patterns
extensionHostProcess.js:178
    ✔ set requiresAcceptance=false if it is a read-only command
extensionHostProcess.js:178
    ✔ whichCommand cannot find the first arg
extensionHostProcess.js:178
    ✔ whichCommand sees first arg on PATH
extensionHostProcess.js:178
    ✔ stub invoke() call
extensionHostProcess.js:178
  7 passing (45ms)
```

---

### DIFF
--- a/packages/core/src/codewhispererChat/tools/executeBash.ts
+++ b/packages/core/src/codewhispererChat/tools/executeBash.ts
@@ -1,0 +1,208 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Writable } from 'stream'
+import { getLogger } from '../../shared/logger/logger'
+import { fs } from '../../shared/fs/fs' // e.g. for getUserHomeDir()
+import { ChildProcess, ChildProcessOptions } from '../../shared/utilities/processUtils'
+import { InvokeOutput, OutputKind, sanitizePath } from './toolShared'
+
+export const readOnlyCommands: string[] = ['ls', 'cat', 'echo', 'pwd', 'which', 'head', 'tail']
+export const maxBashToolResponseSize: number = 1024 * 1024 // 1MB
+export const lineCount: number = 1024
+export const dangerousPatterns: string[] = ['|', '<(', '$(', '`', '>', '&&', '||']
+
+export interface ExecuteBashParams {
+    command: string
+    cwd?: string
+}
+
+export class ExecuteBash {
+    private readonly command: string
+    private readonly workingDirectory?: string
+    private readonly logger = getLogger('executeBash')
+
+    constructor(params: ExecuteBashParams) {
+        this.command = params.command
+        this.workingDirectory = params.cwd ? sanitizePath(params.cwd) : fs.getUserHomeDir()
+    }
+
+    public async validate(): Promise<void> {
+        if (!this.command.trim()) {
+            throw new Error('Bash command cannot be empty.')
+        }
+
+        const args = ExecuteBash.parseCommand(this.command)
+        if (!args || args.length === 0) {
+            throw new Error('No command found.')
+        }
+
+        try {
+            await ExecuteBash.whichCommand(args[0])
+        } catch {
+            throw new Error(`Command "${args[0]}" not found on PATH.`)
+        }
+    }
+
+    public requiresAcceptance(): boolean {
+        try {
+            const args = ExecuteBash.parseCommand(this.command)
+            if (!args || args.length === 0) {
+                return true
+            }
+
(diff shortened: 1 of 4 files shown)


packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    private async whichCommand(cmd: string): Promise<string> {
+        const cp = new ChildProcess('which', [cmd], {
+            collect: true,
+            waitForStreams: true,
+        })
+        const result = await cp.run()
+
+        if (result.exitCode !== 0) {
+            throw new Error(`Command "${cmd}" not found on PATH.`)
```
laileni-aws, line 150: If I remember correctly, `ExitCode !==0` can be satisfied in different cases ?
### Example:
If we try `npm run test` on a project with some syntax errors we get existCode > 0(build failure)

Can you cross check on this ?

laileni-aws (approved):

packages/core/src/codewhispererChat/tools/executeBash.ts
tsmithsz: Here specifically we are running `which <cmd[0]>`. In practice, which returns 0 only if it finds the command and non-zero if it's not found. So for our case this works.


In the example you gave, we would run `which npm` since we extract the first argument in `validate()`.

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -0,0 +1,205 @@
...
+
+            const command = args[0]
+            return !readOnlyCommands.includes(command)
+        } catch (error) {
+            this.logger.warn(`Error while checking acceptance: ${(error as Error).message}`)
+            return true
+        }
+    }
+
+    public async invoke(updates: Writable): Promise<InvokeOutput> {
```
ctlai95, line 71: nit: Can these methods be static?

packages/core/src/codewhispererChat/tools/toolShared.ts
```diff
@@ -7,6 +7,10 @@ import path from 'path'
 import fs from '../../shared/fs/fs'
 
 export const maxToolResponseSize = 30720 // 30KB
+export const readOnlyCommands: string[] = ['ls', 'cat', 'echo', 'pwd', 'which', 'head', 'tail']
+export const maxBashToolResponseSize: number = 1024 * 1024 // 1MB
+export const lineCount: number = 1024
+export const dangerousPatterns: string[] = ['|', '<(', '$(', '`', '>', '&&', '||']
```
ctlai95, line 13: I think these are specific to executeBash?

packages/core/src/codewhispererChat/tools/executeBash.ts
tsmithsz: Updated

packages/core/src/codewhispererChat/tools/toolShared.ts
tsmithsz: Correct. Moved to the executeBash.ts

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -0,0 +1,208 @@
...
+                waitForStreams: true,
+                onStdout: (chunk: string) => {
+                    ExecuteBash.handleChunk(chunk, stdoutBuffer, updates)
+                },
+                onStderr: (chunk: string) => {
+                    ExecuteBash.handleChunk(chunk, stderrBuffer, updates)
+                },
+            }
+
+            const childProcess = new ChildProcess('bash', ['-c', this.command], childProcessOptions)
```
mr-lee, line 92: For later - we need to add some permissions around this.

mr-lee (approved):

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        const output = result.stdout.trim()
+        if (!output) {
+            throw new Error(`Command "${cmd}" found but 'which' returned empty output.`)
+        }
+        return output
+    }
+
+    private static parseCommand(command: string): string[] | undefined {
```
justinmk3, line 163: Parsing bash/shell commands is non-trivial. Is this intended to support arbitrary bash command, or only very simple cases?

PR #6850, opened 2025-03-26
feat(amazonq): Disable Q Signout option and update chat Introduction message for Sagemaker Unified Studio

parameja1:
## Problem
In Sagemaker Unified Studio(SMUS), user shouldnt be allowed to sign out of Q Chat extension as per business requirement. In addition to that, Q Chat Introduction message should display Q Free Tier is only supported with the upcoming SMUS release

## Solution

1. `isSageMaker` is used to identify if it is SageMaker AI (CodeEditor) which comes from [here](https://github.com/aws/sagemaker-code-editor/blob/main/patched-vscode/product.json#L2-L3) as shared by Sagemaker team. Since SageMaker Unified Studio also uses the same repository/branch for CodeEditor we required a new environment variable `SERVICE_NAME` to differentiate SMUS case.
2. Read `SERVICE_NAME` and set a new vscode env context variable `aws.isSagemakerStudio` to control visibility of Sign out label
3. Sign out option to users is controlled at two places: 
- VS Code command palette
- Quick pick UI action
- Both the sign out options are disabled based on the SMUS flag
4. Similarly Q Chat Introduction message in WebView is modified for SMUS case alone

## Testing
- Successfully compiled and built in local to generate vsix files
- Tested on a remote CodeEditor App of SMUS type to verify the functionality manually
- Also tested on existing CodeEditor App used by Sagemaker AI to verify the changes are backward compatible
- Ran unit tests in local and succeeded
---


packages/core/src/amazonq/webview/ui/tabs/generator.ts
```diff
@@ -48,6 +53,8 @@ export class TabDataGenerator {
             return {}
         }
 
+        const isSMUSProTier = serviceName ? serviceName === 'SageMakerUnifiedStudio' : false
```
kjayan, line 56: nit: Are we identifying Pro vs Free Tier here or SMUS / SMAI?

packages/amazonq/package.json
```diff
@@ -351,7 +351,7 @@
                 },
                 {
                     "command": "aws.amazonq.signout",
-                    "when": "(view == aws.amazonq.AmazonQChatView) && aws.codewhisperer.connected",
+                    "when": "(view == aws.amazonq.AmazonQChatView) && aws.codewhisperer.connected && !aws.amazonq.isSagemakerStudio",
```
jpinkney-aws, line 354: can we use `aws.isSageMaker` instead of having a new context variable?

packages/core/src/amazonq/util/amazonQState.ts
```diff
@@ -0,0 +1,30 @@
+export class AmazonQState {
```
jpinkney-aws, line 1: instead of having this can we just have something in [here](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/extensionUtilities.ts#L69) or [here](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/extensionUtilities.ts#L160)? why do we need _another_ way to tell that its sage maker?

packages/amazonq/src/app/chat/activation.ts
```diff
@@ -16,6 +16,11 @@ export async function activate(context: ExtensionContext) {
     const appInitContext = amazonq.DefaultAmazonQAppInitContext.instance
 
     registerApps(appInitContext, context)
+    const serviceName = process.env.SERVICE_NAME ?? 'SageMakerUnifiedStudio' // Fallback service name to a generic name AmazonQ
```
jpinkney-aws, line 19: I think you should just add this to extension utilities instead of having specific sagemaker code in our general activation

packages/core/src/amazonq/webview/generators/webViewContent.ts
```diff
@@ -83,6 +83,7 @@ export class WebViewContentGenerator {
         const disclaimerAcknowledged = globals.globalState.tryGet('aws.amazonq.disclaimerAcknowledged', Boolean, false)
 
         const welcomeLoadCount = globals.globalState.tryGet('aws.amazonq.welcomeChatShowCount', Number, 0)
+        const serviceName = process.env.SERVICE_NAME ?? ''
```
jpinkney-aws, line 86: why does mynah need to know the service name? Why can't it just know the [idetype](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/extensionUtilities.ts#L69) and provide differences based on that?

packages/core/src/amazonq/webview/ui/tabs/generator.ts
parameja1: Thanks for the comment @kjayan - Identifying SMUS vs SMAI here, let me correct the variable name to isSMUS

packages/amazonq/package.json
parameja1: Thanks @jpinkney-aws for the comments - To add more context: `isSageMaker` is used to identify if it is SageMaker AI (CodeEditor) which comes from [here](https://github.com/aws/sagemaker-code-editor/blob/main/patched-vscode/product.json#L2-L3) as shared by Sagemaker team, since SageMakerUnifiedStudio also uses the same repository/branch for CodeEditor we required a new environment variable specific to SMUS case. Hence added a new context variable and cant reuse `aws.isSageMaker` here

packages/core/src/amazonq/util/amazonQState.ts
parameja1: Please refer to the above comment, our logic uses environment variable `SERVICE_NAME` be able to distinguish between SageMaker AI and SageMaker Unified Studio

packages/amazonq/src/app/chat/activation.ts
parameja1: To distinguish between SageMaker AI and SageMaker Unified Studio our logic depends on node environment variable `process.env.SERVICE_NAME`, which is accessible during extension activation. In case I missed something here, do you suggest any other entry point to move this logic to?

packages/core/src/amazonq/webview/generators/webViewContent.ts
parameja1: Pls refer to previous comments for added context, let me also update the description

packages/core/src/amazonq/util/amazonQState.ts
jpinkney-aws: That makes sense. Can the distinction be made in something like [here](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/extensionUtilities.ts#L69) instead? It's basically the same thing as this class but already in use. Then on activation you can do:
```
await setContext('aws.amazonq.isSagemakerStudio', getIdeType() === 'sageMakerUnifiedStudio')
```

packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -47,7 +47,8 @@ export const createMynahUI = (
     featureConfigsSerialized: [string, FeatureContext][],
     welcomeCount: number,
     disclaimerAcknowledged: boolean,
-    disabledCommands?: string[]
+    disabledCommands?: string[],
+    service?: string
```
jpinkney-aws, line 51: rather than passing down the service name can be pass down the ide type like I mentioned from [here](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/shared/extensionUtilities.ts#L69) (you'll have to make modifications to support sagemaker unified studio there)

parameja1: @jpinkney-aws Thanks for the comment, checked ExtensionUtilities cant be imported directly in the WebView because utilities has pure node js modules imported. Observed polyfills exception in local and polyfills need to be explicitly configured as mentioned in the [repo](https://github.com/aws/aws-toolkit-vscode/blob/c3c49e6695c1e886c24ef2626f6b277f203d374a/packages/webpack.base.config.js#L54-L56).
As per my understanding, this could increase the bundle asset size. Hence falling back to passing service Name, optional parameter in mynahUI. Let me know your thoughts

packages/core/src/amazonq/util/amazonQState.ts
parameja1: Agreed, can implement a util method here. But in order to disable SignOut option [here](https://github.com/parameja1/aws-toolkit-vscode/blob/32137937e7046bc2f615379108601a456596ff0a/packages/core/src/codewhisperer/ui/statusBarMenu.ts#L98), cant access NodeEnv variables and would need to access the state object. Hence consistently using the same `AmazonQState` for checks

packages/amazonq/package.json
justinmk3: should be named `aws.isSagemakerStudio`

packages/amazonq/package.json
```diff
@@ -627,7 +627,7 @@
                 "title": "%AWS.command.codewhisperer.signout%",
                 "category": "%AWS.amazonq.title%",
                 "icon": "$(debug-disconnect)",
-                "enablement": "aws.codewhisperer.connected"
+                "enablement": "aws.codewhisperer.connected && !aws.amazonq.isSagemakerUnifiedStudio"
```
justinmk3, line 630: This context var is indicating an environment thing. It's not specific to Q. So the same name should be used in both Q and Toolkit:

```suggestion
                "enablement": "aws.codewhisperer.connected && !aws.isSagemakerUnifiedStudio"
```

packages/amazonq/src/app/chat/activation.ts
```diff
@@ -63,6 +64,14 @@ export async function activate(context: ExtensionContext) {
 
     void setupLsp()
     void setupAuthNotification()
+
+    async function setupAmazonQState() {
+        const serviceName = process.env.SERVICE_NAME ?? 'AmazonQ' // Fallback service name to a generic name AmazonQ
+        const amazonQState = amazonq.AmazonQState.initialize(serviceName)
+        getLogger().info(`In Extension activation - q state is initialized ${amazonQState.serviceName}`)
```
justinmk3, line 71: `getLogger()` accepts a topic. Don't manually add a "topic" like `In Extension -`

packages/core/src/shared/vscode/setContext.ts
```diff
@@ -37,6 +37,7 @@ export type contextKey =
     | 'gumby.wasQCodeTransformationUsed'
     | 'amazonq.inline.codelensShortcutEnabled'
     | 'aws.toolkit.lambda.walkthroughSelected'
+    | 'aws.amazonq.isSagemakerUnifiedStudio'
```
justinmk3, line 40: put this next to `aws.isSageMaker` . helps discovery.

```suggestion
    | 'aws.isSagemakerUnifiedStudio'
```

packages/core/src/amazonq/webview/generators/webViewContent.ts
justinmk3: the SERVICE_NAME logic should live in `packages/core/src/shared/vscode/env.ts`.

packages/amazonq/src/app/chat/activation.ts
```diff
@@ -63,6 +64,14 @@ export async function activate(context: ExtensionContext) {
 
     void setupLsp()
     void setupAuthNotification()
+
+    async function setupAmazonQState() {
+        const serviceName = process.env.SERVICE_NAME ?? 'AmazonQ' // Fallback service name to a generic name AmazonQ
+        const amazonQState = amazonq.AmazonQState.initialize(serviceName)
```
justinmk3, line 70: This is a very over-engineered way to set a single flag. `AmazonQState` is not needed. Just add some logic to `packages/core/src/shared/vscode/env.ts`

packages/amazonq/package.json
parameja1: Ack, fixing the context variable

packages/amazonq/src/app/chat/activation.ts
parameja1: Ack, but as posted in the [comment](https://github.com/aws/aws-toolkit-vscode/pull/6850#discussion_r2017400932), NODE Env variables arent available there as it seems to be after extension activates and hence was the reason to add `AmazonQState`. Verifying better approach and will address in next revision

packages/core/src/shared/vscode/setContext.ts
parameja1: Thanks, addressing in next revision

packages/amazonq/src/app/chat/activation.ts
parameja1: Ack, will check and update in new revision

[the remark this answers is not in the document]
parameja1: Ack, updating logger message in new revision

packages/core/src/amazonq/webview/ui/tabs/generator.ts
```diff
@@ -48,6 +53,8 @@ export class TabDataGenerator {
             return {}
         }
 
+        const isSMUS = serviceName ? serviceName === 'SageMakerUnifiedStudio' : false
```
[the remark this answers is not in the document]
parameja1: Agreed, refactored the SMUS check and will address with new revision

packages/amazonq/package.json
```diff
@@ -351,7 +351,7 @@
                 },
                 {
                     "command": "aws.amazonq.signout",
-                    "when": "(view == aws.amazonq.AmazonQChatView) && aws.codewhisperer.connected",
+                    "when": "(view == aws.amazonq.AmazonQChatView) && aws.codewhisperer.connected && !aws.amazonq.isSagemakerUnifiedStudio",
```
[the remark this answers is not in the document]
parameja1: From the repo, sign out flag seems to be controlled at two places: 
1. At UI element creation for sign out
2. In package.json using context var ex: `aws.codewhisperer.connected`

Once a valid connection is setup in Auth layer, `aws.codewhisperer.connected` is being [set](https://github.com/aws/aws-toolkit-vscode/blob/ce01fa85c7a97c71cd3dcc11770f6cb75b29e1c2/packages/core/src/codewhisperer/util/authUtil.ts#L142) which is the same context var being used in check [here](https://github.com/aws/aws-toolkit-vscode/blob/ce01fa85c7a97c71cd3dcc11770f6cb75b29e1c2/packages/core/src/codewhisperer/ui/statusBarMenu.ts#L95).

packages/amazonq/package.json
parameja1: Ack, thanks for the comment. Addressing in next rev

parameja1:
@justinmk3 Can you please check and confirm if the unit tests for UI class generator are required? since observing [failure](https://github.com/aws/aws-toolkit-vscode/actions/runs/14114870783/job/39542527033?pr=6850#step:5:120) at browser module import `@aws-mynah-ui` [here](https://github.com/aws/aws-toolkit-vscode/blob/ce01fa85c7a97c71cd3dcc11770f6cb75b29e1c2/packages/core/src/amazonq/webview/ui/tabs/generator.ts#L6)

justinmk3 (commented):
discussed offline: to unblock this work, 

- add a function near `isSageMaker()` which sets a module-local `isSageMakerUnified=true` flag. 
- `isSageMaker()` gains an optional `flavor?: string` parameter
- code that cares about "unified" vs "regular" SM can call `isSageMaker('unified')`
- don't need a new singleton or global state module.
- TODO: find out why `process.env.SERVICE_NAME` is only "sometimes" available?

packages/amazonq/src/app/chat/activation.ts
```diff
@@ -63,6 +64,12 @@ export async function activate(context: ExtensionContext) {
 
     void setupLsp()
     void setupAuthNotification()
+
+    async function setIsSMUSContext() {
+        // set context key for SageMaker Unified Studio
+        const isSMUS = isSageMaker(true)
+        await setContext('aws.isSageMakerUnifiedStudio', isSMUS)
```
justinmk3, line 71: Why is this in the "chat" activation? Detecting SMUS is a general concept, not chat-specific. So I would expect this to happen as early as possible, in the main entry point:

- `activateAmazonQCommon` https://github.com/aws/aws-toolkit-vscode/blob/b1fae85369d794172d730cc576fe27a41b6c1778/packages/amazonq/src/extension.ts#L51
- and `activateCommon`: https://github.com/aws/aws-toolkit-vscode/blob/b1fae85369d794172d730cc576fe27a41b6c1778/packages/core/src/extension.ts#L66

Note: because of tech debt, currently those 2 places should be updated. Eventually, common code like this (which is shared by both Toolkit and Q) will only live in `activateCommon`.

Even if you don't currently have a need to detect SMUS in Toolkit, it should still be done for both extensions.

packages/core/src/amazonq/webview/generators/webViewContent.ts
justinmk3: Followup: the SM vscode fork *should* set appName appropriately, but it doesn't, so doing this SERVICE_NAME sniffing in the ide extension is a workaround :(

packages/core/src/amazonq/webview/ui/tabs/constants.ts
```diff
@@ -6,6 +6,17 @@ import { TabType } from '../storages/tabsStorage'
...
+const qChatIntroMessage = `Hi, I'm Amazon Q. I can answer your software development questions.
+  Ask me to explain, debug, or optimize your code.
+  You can enter \`/\` to see a list of quick actions. Use \`@\` to add saved prompts, files, folders, or your entire workspace as context.`
+
+export const qChatIntroMessageForSMUS = `Hi, I'm Amazon Q. I can answer your software development questions.\n\
+  Ask me to explain, debug, or optimize your code.\n\
+  You can enter \`/\` to see a list of quick actions. Use \`@\` to add saved prompts, files, folders, or your entire workspace as context.
+  You are now using Q free tier.\n\
+  `
+export const sageMakerUnifiedStudio = 'SageMakerUnifiedStudio'
```
justinmk3, line 18: There is nothing Q-specific about this. Why is this in the Q constants file? It should live in `core/src/shared/constants.ts` at least, though there is probably a better place.

packages/core/src/amazonq/webview/ui/tabs/generator.ts
```diff
@@ -39,7 +39,12 @@ export class TabDataGenerator {
         this.highlightCommand = props.commandHighlight
     }
 
-    public getTabData(tabType: TabType, needWelcomeMessages: boolean, taskName?: string): MynahUIDataModel {
+    public getTabData(
+        tabType: TabType,
+        needWelcomeMessages: boolean,
+        taskName?: string,
+        serviceName?: string
```
justinmk3, line 46: As we discussed, there should be no need to pass this around. Use `isSageMaker('unified')`

packages/amazonq/src/app/chat/activation.ts
parameja1: Thanks for the context @justinmk3, let me update the revision

packages/core/src/auth/auth.ts
```diff
@@ -1108,10 +1108,12 @@ export class Auth implements AuthService, ConnectionManager {
  *
  * @param isC9 boolean for if Cloud9 is host
  * @param isSM boolean for if SageMaker is host
+ * @param isSMUS boolean for if SageMaker Unified Studio is host
  * @returns boolean for if C9 "OR" SM
  */
-export function hasVendedIamCredentials(isC9?: boolean, isSM?: boolean) {
+export function hasVendedIamCredentials(isC9?: boolean, isSM?: boolean, isSMUS?: boolean) {
```
justinmk3, line 1114: Remove

packages/core/src/shared/extensionUtilities.ts
```diff
@@ -145,6 +146,14 @@ function createCloud9Properties(company: string): IdeProperties {
     }
 }
 
+function isSageMakerUnifiedStudio(): boolean {
+    const service = process.env.SERVICE_NAME
```
justinmk3, line 150: As we discussed, once a truthy result is found, it should be stored in a module-level variable in this module. Then that's used for future calls. That avoids the "SERVICE_NAME is sometimes not visible" problem. And it avoids the need to pass around the service name.

packages/core/src/amazonq/webview/ui/tabs/generator.ts
justinmk3: Why not pass `isSageMakerUnifed: boolean`? Why pass the service name?

packages/core/src/auth/auth.ts
parameja1: Added for future-proofing, though agreed its not needed with the current change

packages/core/src/shared/extensionUtilities.ts
parameja1: Ack, updating the revision

packages/core/src/amazonq/webview/ui/tabs/generator.ts
justinmk3: That kind of logic should not live in the webview. The webview should have minimal business logic.

packages/core/src/auth/auth.ts
justinmk3: It should never be needed. Again, all non-webview code should use `isSageMaker(...)`.

parameja1:
@justinmk3 Observing two build failures which is unrelated to the changes, can you pls take a look and advise on next step?

packages/core/src/amazonq/webview/ui/tabs/constants.ts
```diff
@@ -6,6 +6,16 @@ import { TabType } from '../storages/tabsStorage'
 import { QuickActionCommandGroup } from '@aws/mynah-ui'
 import { userGuideURL } from '../texts/constants'
 
+const qChatIntroMessage = `Hi, I'm Amazon Q. I can answer your software development questions.
+  Ask me to explain, debug, or optimize your code.
+  You can enter \`/\` to see a list of quick actions. Use \`@\` to add saved prompts, files, folders, or your entire workspace as context.`
+
+export const qChatIntroMessageForSMUS = `Hi, I'm Amazon Q. I can answer your software development questions.\n\
```
justinmk3, line 13: This is really strange. Why is there a SMUS-specific Q message? That is not maintainable.

If there is a part of a message that needs to mention the product name, then replace a token in the message with `productName()`.

It looks like this variation is about the "free tier"? If that's that case, then surely this variable should be named something like `...FreeTier`. Again, maintaining random strings for every flavor of environment that the extension might run in, is not sustainable.

packages/core/src/codewhisperer/ui/statusBarMenu.ts
```diff
@@ -92,7 +93,9 @@ export function getQuickPickItems(): DataQuickPickItem<string>[] {
         // Add settings and signout
         createSeparator(),
         createSettingsNode(),
-        ...(AuthUtil.instance.isConnected() && !hasVendedIamCredentials() ? [createSignout()] : []),
+        ...(AuthUtil.instance.isConnected() && !(hasVendedIamCredentials() || isSageMaker('SMUS'))
```
justinmk3, line 96:
```suggestion
        ...(AuthUtil.instance.isConnected() && !(hasVendedIamCredentials() || isSageMaker('smUnified'))
```

justinmk3, line 96: `hasVendedIamCredentials()` *already* has a check for `isSM` in its own logic. Normally it returns true for SM. Should `hasVendedIamCredentials()` return false for SMUS? If so, change the logic *in hasVendedIamCredentials()`. Not in spread-out places like this.

```typescript
export function hasVendedIamCredentials(isC9?: boolean, isSM?: boolean) {
    isC9 ??= isCloud9()
    isSM ??= isSageMaker('SMAI')
    return isSM || isC9
}
```

packages/core/src/shared/extensionUtilities.ts
```diff
@@ -22,13 +22,16 @@ import {
...
+    sageMakerUnifiedStudio,
+    vscodeAppname,
+} from './vscode/constants'
 
 const localize = nls.loadMessageBundle()
 
-const vscodeAppname = 'Visual Studio Code'
-const cloud9Appname = 'AWS Cloud9'
-const cloud9CnAppname = 'Amazon Cloud9'
-const sageMakerAppname = 'SageMaker Code Editor'
```
justinmk3, line 31: These constants are not used anywhere else, so should stay here, not exported in a new constants.ts file.

And the new `sageMakerUnifiedStudio` can be added here too. It's not used anywhere except in this file.

packages/core/src/shared/extensionUtilities.ts
```diff
@@ -157,8 +170,20 @@ export function isCloud9(flavor: 'classic' | 'codecatalyst' | 'any' = 'any'): bo
...
 }
 
-export function isSageMaker(): boolean {
-    return vscode.env.appName === sageMakerAppname
+/**
+ *
+ * @param appName to identify the proper SM instance
+ * @returns true if the current system is SageMaker(SMAI or SMUS)
+ */
+export function isSageMaker(appName: string = 'SMAI'): boolean {
```
justinmk3, line 178:
```suggestion
export function isSageMaker(appName: 'SMAI' | 'SMUS' = 'SMAI'): boolean {
```

packages/core/src/amazonq/webview/ui/tabs/constants.ts
parameja1: Ack, understand your point. For the current launch which is VP goal, since Pro Tier isnt supported, this message is required to be shown to customers. i.e to show `You are now using Free Tier` at the end of the message. On maintainability, this message will be removed with subsequent launch and is only for the interim. does that help answer? If yes, can we pls go ahead with the change?

packages/core/src/shared/extensionUtilities.ts
parameja1: In core/src/shared directory, the constants file are not pure constants i.e have node modules imported, so both the existing files [extUtilities](https://github.com/aws/aws-toolkit-vscode/blob/1b3681c9d47e30196cb32138338aeba65c51e944/packages/core/src/shared/extensionUtilities.ts) and [shared constants](https://github.com/aws/aws-toolkit-vscode/blob/1b3681c9d47e30196cb32138338aeba65c51e944/packages/core/src/shared/constants.ts) cant be used. Hence added a new constants within `vscode/` to move app name related constants

packages/core/src/amazonq/webview/ui/tabs/constants.ts
justinmk3: Then this should have a code comment that mentions that.

packages/core/src/shared/extensionUtilities.ts
```diff
@@ -145,6 +150,14 @@ function createCloud9Properties(company: string): IdeProperties {
     }
 }
 
+function isSageMakerUnifiedStudio(): boolean {
+    if (serviceName === notInitialized) {
+        serviceName = process.env.SERVICE_NAME ?? ''
+        isSMUS = serviceName === sageMakerUnifiedStudio
```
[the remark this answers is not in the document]
parameja1: Added `serviceName` for the initial state check and the alternate is to check `isSMUS` boolean is undefined, the first seemed extensible in case of serviceName changes. Can we syncup on this?

packages/core/src/codewhisperer/ui/statusBarMenu.ts
justinmk3: That is why it should return false, as I mentioned. Then this local special-case wouldn't be needed.

Find a way to avoid this local special-case. What is the *actual semantics* you want to check here?

packages/core/src/shared/extensionUtilities.ts
parameja1: Can we consider this as a nit? Default case returns false and hence shouldnt cause an issue. Let me follow-up on the slack thread, thanks

packages/core/src/shared/extensionUtilities.ts
justinmk3: These are *not* shared. They are *local* to this module.

packages/core/src/shared/extensionUtilities.ts
justinmk3: It is a nit that can be easily fixed in the other followup work required for this pr

PR #6853, opened 2025-03-26
feat(chat): add fsWrite tool to agentic loop

ctlai95:
## Problem
FsWrite should be enabled in agentic loop

## Solution
- Updated tool index with fsWrite
- Updated the agentic loop to invoke fsWrite tool
- Added input validation

---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -58,7 +58,7 @@ import { randomUUID } from '../../../shared/crypto'
 import { LspController } from '../../../amazonq/lsp/lspController'
 import { CodeWhispererSettings } from '../../../codewhisperer/util/codewhispererSettings'
 import { getSelectedCustomization } from '../../../codewhisperer/util/customizationUtil'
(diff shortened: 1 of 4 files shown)


packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -916,22 +918,22 @@ export class ChatController {
...
+                            await FsWrite.validate(input)
+                            result = await FsWrite.invoke(input)
+                            break
+                        }
                         default:
+                            getLogger().warn('Received invalid tool: %s', toolUse.name)
                             break
                     }
+                    if (!result) {
+                        throw new ToolkitError('Failed to execute tool and get results')
```
mr-lee, line 932: i18n for later.  I think if the tool fails - it's still useful to send back to the LLM.  Check what the CLI does.

mr-lee (approved):

packages/core/src/codewhispererChat/controllers/chat/controller.ts
ctlai95: Yup there's a catch block right below this to append this error to the toolResults

PR #6857, opened 2025-03-26
telemetry(amazonq): Add id2 and logging for createUpload metrics

zuoyaofu:
## Problem

Some users are reporting S3 upload failures but we don't have the id2 that S3 team requires to investigate further.

## Solution

Emit amazonq_createUpload metric with id2, and add logs for these ids so customers can use too.

---

### DIFF
--- a/packages/core/src/codewhisperer/service/securityScanHandler.ts
+++ b/packages/core/src/codewhisperer/service/securityScanHandler.ts
@@ -384,20 +384,24 @@ export async function uploadArtifactToS3(
         headersObj['x-amz-server-side-encryption-aws-kms-key-id'] = resp.kmsKeyArn
     }
 
+    let requestId: string | undefined = undefined
+    let id2: string | undefined = undefined
+    let responseCode: string = ''
+
     try {
         const response = await request.fetch('PUT', resp.uploadUrl, {
             body: readFileSync(fileName),
             headers: resp?.requestHeaders ?? headersObj,
         }).response
         logger.debug(`StatusCode: ${response.status}, Text: ${response.statusText}`)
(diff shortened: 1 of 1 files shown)


ashishrp-aws (approved):

ctlai95 (approved):

packages/core/src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -384,20 +384,24 @@ export async function uploadArtifactToS3(
         headersObj['x-amz-server-side-encryption-aws-kms-key-id'] = resp.kmsKeyArn
     }
 
+    let requestId: string | undefined = undefined
+    let id2: string | undefined = undefined
+    let responseCode: string = ''
+
     try {
         const response = await request.fetch('PUT', resp.uploadUrl, {
             body: readFileSync(fileName),
```
justinmk3, line 393: Is it possible this exceeds a request size limit? Should it be streamed?

packages/core/src/codewhisperer/service/securityScanHandler.ts
```diff
@@ -384,20 +384,24 @@ export async function uploadArtifactToS3(
...
             headers: resp?.requestHeaders ?? headersObj,
         }).response
         logger.debug(`StatusCode: ${response.status}, Text: ${response.statusText}`)
+        requestId = response.headers?.get('x-amz-request-id') ?? undefined
+        id2 = response.headers?.get('x-amz-id-2') ?? undefined
+        responseCode = response.status.toString()
```
justinmk3, line 399: Not great that all of this is placed in the business logic for a feature. This kind of thing should live in `shared/request`.

PR #6863, opened 2025-03-26
feat(chat): add tool utils to delegate execution

ctlai95:
## Problem

Tool APIs are not consistent.

## Solution

Create a tool util wrapper class around tools to delegate validation, execution, and previews of tools.

So now to use the tools we can do something like:
```typescript
const tool = ToolUtils.tryFromToolUse(toolUse)

ToolUtils.validate(tool)
ToolUtils.requiresAcceptance(tool)
ToolUtils.queueDescription(tool)
ToolUtils.invoke(tool)
```

---


packages/core/src/codewhispererChat/tools/toolUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+    FsWrite = 'fsWrite',
+    ExecuteBash = 'executeBash',
+}
+
+export type Tool =
+    | { type: ToolType.FsRead; tool: FsRead }
+    | { type: ToolType.FsWrite; tool: FsWrite }
+    | { type: ToolType.ExecuteBash; tool: ExecuteBash }
+
+export class ToolUtils {
```
mr-lee, line 23: what's the intent of this class?

packages/core/src/codewhispererChat/tools/toolUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { Writable } from 'stream'
+import { FsRead, FsReadParams } from './fsRead'
+import { FsWrite, FsWriteParams } from './fsWrite'
+import { ExecuteBash, ExecuteBashParams } from './executeBash'
+import { ToolResult, ToolResultContentBlock, ToolResultStatus, ToolUse } from '@amzn/codewhisperer-streaming'
+import { InvokeOutput } from './toolShared'
+
+export enum ToolType {
```
mr-lee, line 12: I feel like this code is not extensible for when we support MCP-based tools.  Can we revisit this after the prototype?

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -618,7 +618,7 @@ export class Messenger {
         using the \`executeBash\` tool.`
         }
         if (toolUse.name === 'fsWrite') {
```
mr-lee, line 620: I know you didn't change this - but I feel like we should refactor this to have a cleaner interface.  Having a giant if/else block doesn't seem like the right solution.

mr-lee (approved):
I'm approving to unblock - but I also think we need to refactor here.

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
ctlai95: Yeah that was the goal of this PR to have the common interface across all the tools (I just didn't get to refactoring this part yet since @laileni-aws will be integrating his UI changes here)

packages/core/src/codewhispererChat/tools/toolUtils.ts
ctlai95: Sure, I need to think about how MCP will work here. Happy to revisit later

packages/core/src/codewhispererChat/tools/toolUtils.ts
ctlai95: The idea was to have this be a helper class to interact with all the tools and to sort of enforce a common interface (requiresAcceptance, invoke, etc) so that we don't need to keep writing if-else or switch statements just to handle tools. This is how the CLI does it as well

packages/amazonq/.vscode/tasks.json
```diff
@@ -5,9 +5,8 @@
     "tasks": [
         {
             "label": "watch",
-            "type": "shell",
-            "command": "npm",
-            "args": ["run", "watch"],
+            "type": "npm",
```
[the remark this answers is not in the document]
ctlai95: I accidentally committed this change to this branch in a previous commit. Here I'm reverting it back to what it is in master.

On another note, my vscode keeps trying to run watch using `yarn` instead of `npm` even in master. I noticed this happening for others as well so I'll make this change in master

packages/core/src/codewhispererChat/tools/toolUtils.ts
justinmk3: Seems like a case where class-based inheritance would fit.

PR #6864, opened 2025-03-26
feat(chat): Stream executeBash output to chat

tsmithsz:
## Problem
We need to be able to stream the command tool execution logs to the chat in real-time

## Solution

- Added a new ChatStream class that extends Node's Writable stream to capture and stream bash command output to the chat UI in real-time
- Implemented a sendPartialBashToolLog method in the Messenger class to display command output as it's generated
- Modified the chat controller to use the new streaming functionality instead of waiting for commands to complete
-  Added appropriate logging and error handling for the streaming functionality
-  Fixed a duplicate write operation in the handleChunk method of executeBash.ts

https://github.com/user-attachments/assets/a27900d1-5c0e-4690-a7d3-33f7712f255a
---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
@@ -438,6 +438,27 @@ export class Messenger {
         )
     }
 
+    public sendPartialBashToolLog(message: string, tabID: string, triggerID: string, toolUseId: string | undefined) {
+        this.dispatcher.sendChatMessage(
+            new ChatMessage(
+                {
+                    message,
+                    messageType: 'answer-part',
+                    followUps: undefined,
+                    followUpsHeader: undefined,
+                    relatedSuggestions: undefined,
+                    triggerID,
+                    messageID: toolUseId ?? `tool-output`,
+                    userIntent: undefined,
+                    codeBlockLanguage: 'plaintext',
+                    contextList: undefined,
+                    canBeVoted: false,
+                },
+                tabID
+            )
+        )
+    }
+
     private editorContextMenuCommandVerbs: Map<EditorContextCommandType, string> = new Map([
         ['aws.amazonq.explainCode', 'Explain'],
         ['aws.amazonq.explainIssue', 'Explain'],
--- a/packages/core/src/codewhispererChat/tools/chatStream.ts
+++ b/packages/core/src/codewhispererChat/tools/chatStream.ts
@@ -1,0 +1,52 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Writable } from 'stream'
+import { getLogger } from '../../shared/logger/logger'
+import { Messenger } from '../controllers/chat/messenger/messenger'
+
+/**
+ * A writable stream that feeds each chunk/line to the chat UI.
+ * Used for streaming tool output (like bash execution) to the chat interface.
+ */
+export class ChatStream extends Writable {
+    private accumulatedLogs = ''
+
+    public constructor(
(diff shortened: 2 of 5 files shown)


packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -438,6 +438,27 @@ export class Messenger {
         )
     }
 
+    public sendPartialBashToolLog(message: string, tabID: string, triggerID: string) {
+        this.dispatcher.sendChatMessage(
```
mr-lee, line 442: Do we want each of these to be a new chat message?

mr-lee (approved):
non-blocking question.  I _think_ what has been implemented will update the same UX element over and over, but I figured I would confirm.

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -438,6 +438,27 @@ export class Messenger {
...
+            new ChatMessage(
+                {
+                    message,
+                    messageType: 'answer-part',
+                    followUps: undefined,
+                    followUpsHeader: undefined,
+                    relatedSuggestions: undefined,
+                    triggerID,
+                    messageID: `tool-output`,
```
ctlai95, line 451: Should we use the toolUseId here? I'm not sure if this will cause unexpected conflicting messages

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
ctlai95: If you set the messageType and messageID properly, mynah-ui will know to keep updating the same chat message

packages/core/src/codewhispererChat/tools/chatStream.ts
```diff
@@ -0,0 +1,45 @@
...
+    ) {
+        super()
+        this.logger.debug(`ChatStream created for tabID: ${tabID}, triggerID: ${triggerID}`)
+    }
+
+    override _write(chunk: Buffer, encoding: BufferEncoding, callback: (error?: Error | null) => void): void {
+        const text = chunk.toString()
+        this.accumulatedLogs += text
+        this.logger.debug(`ChatStream received chunk: ${text}`)
+        this.messenger.sendPartialBashToolLog(`\`\`\`bash\n${this.accumulatedLogs}\`\`\``, this.tabID, this.triggerID)
```
ctlai95, line 31: Is there a way we can make this generic and not mention `bash`? I'm wondering if this can be reused by the other tools

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
tsmithsz: Yeah, right now because I set those fields to be the same for each call it updates the same chat message

packages/core/src/codewhispererChat/tools/chatStream.ts
tsmithsz: The method can be renamed and used by other tools. I added `bash` because there seems to be some processing on the chat message field where for text in triple-backticks (```) it tries to extract the first bit of text after the backticks to be used as the language name for the code block.

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
tsmithsz: Will update

packages/core/src/codewhispererChat/tools/chatStream.ts
ctlai95: Yep the UI uses the markdown format to determine the code block language. But I'm thinking for example, if we want to have a tool print some other code block language then it shouldn't be hardcoded as bash here. But let's follow up later

ctlai95 (approved):

PR #6865, opened 2025-03-26
feat(amazonq): support customization through language server

zixlin7:
## Problem
customization support for inline through language server

## Solution

---


packages/amazonq/src/app/inline/completion.ts
```diff
@@ -60,6 +63,31 @@ export class InlineCompletionManager implements Disposable {
         }
     }
 
+    public registerCustomization(client: LanguageClient) {
+        commands.registerCommand(
+            'aws.amazonq.connect',
```
jpinkney-aws, line 68: could this be something like `aws.amazonq.customization`? it makes me think its logging in

jpinkney-aws: also does it make sense to register the command with our internal commands imlementation?

packages/amazonq/src/app/inline/completion.ts
```diff
@@ -60,6 +63,31 @@ export class InlineCompletionManager implements Disposable {
...
+                    source,
+                    startUrl,
+                    region,
+                    customizationArn,
+                    customizationNamePrefix
+                )
+            }
+        )
+
+        commands.registerCommand('aws.amazonq.select', async (customization: any, source: CodeWhispererSource) => {
```
jpinkney-aws, line 86: same with this one, should it be `aws.amazonq.selectCustomization`?

packages/core/src/codewhisperer/commands/basicCommands.ts
```diff
@@ -263,6 +267,53 @@ export const showSsoSignIn = Commands.declare('aws.amazonq.sso', () => async ()
 // It can optionally set a customization too based on given values to match on
 
 // This command is only declared and registered in Amazon Q if Q exists
+
+export const connectCustomizationHandler =
```
jpinkney-aws, line 271: is the only difference of this implementation is that it plumbs down the client into select customization?

packages/core/src/codewhisperer/util/customizationUtil.ts
```diff
@@ -147,6 +162,9 @@ export const setSelectedCustomization = async (customization: Customization, isO
     }
     vsCodeState.isFreeTierLimitReached = false
     await Commands.tryExecute('aws.amazonq.refreshStatusBar')
+    if (Experiments.instance.get('amazonqLSP', true) && client) {
+        await notifySelectedCustomizatioToLsp(client, customization.arn)
```
jpinkney-aws, line 166: nitpick: 
```suggestion
        await notifySelectedCustomizationToLsp(client, customization.arn)
```

packages/core/src/codewhisperer/commands/basicCommands.ts
zixlin7: yes that's right

jpinkney-aws (commented):
As a thought experiment: instead of passing down the language client through everything, what if we passed through the available customization instead? Would that make the code easier to understand/break up separation of concerns between the lsp calls and what's currently in the code?

zixlin7:
> As a thought experiment: instead of passing down the language client through everything, what if we passed through the available customization instead? Would that make the code easier to understand/break up separation of concerns between the lsp calls and what's currently in the code?

hmm what we can do is pass in those two functions to get customizations and send notification about the selected one. But I feel like passing client itself might be leaner.

Will-ShaoHua:
Zoe maybe we can take this chance to make custom related thing to a `class CustomizationService` and abstract all the LSP detail there?

Will-ShaoHua:
it would be easy to use to have a class
```
class CustomizationService {
      construcotr(private lspClient: Client) {}
      
      async switchCustomization() {
              this.lsClient.sendNotification(...)
      }

      async listAvailableCustomizations() {
             await this.lspClient.listCustomization()
             ....
     }

     .....
}

```

Will-ShaoHua:
another qq, why don't we do flare migration in feature branch

Will-ShaoHua:
i wonder is there any concern we don't do migration work in feature branch? @jpinkney-aws I feel doing these changes in feature branch will make us easier to make large surface refactring with out too much operational concerns.

packages/core/src/codewhisperer/util/customizationUtil.ts
```diff
@@ -29,15 +36,19 @@ export const getNewCustomizations = (availableCustomizations: Customization[]) =
     return availableCustomizations.filter((c) => !persistedCustomizations.map((p) => p.arn).includes(c.arn))
 }
 
-export async function notifyNewCustomizations() {
+export async function notifyNewCustomizations(client?: LanguageClient) {
```
Will-ShaoHua, line 39: yea let's wrap all these things into a class and have the client in one place instead of passing the client in all function calls

zixlin7:
> Zoe maybe we can take this chance to make custom related thing to a `class CustomizationService` and abstract all the LSP detail there? I feel it will make it easier for caller to use, also easier to test against, currently everything is first class function and it makes it hard to unit test

I think this refactor might be better after we deprecate the old code path. Right now if we want to do this we are basically adding a copy of the utils.

Will-ShaoHua:
i kinda prefer refactoring them in one shot when we're writing the new LSP path as they're not yet in use in PROD so that we don't have much concern but either one works for me. But to fit in the current project planing and timeline, it makes sense to do refactoring afterward

Will-ShaoHua (approved):

zixlin7:
> i kinda prefer refactoring them in one shot when we're writing the new LSP path as they're not yet in use in PROD so that we don't have much concern but either one works for me. But to fit in the current project planing and timeline, it makes sense to do refactoring afterward

created a new class refactored from the original utils and kept the original as is we can deprecate it later

jpinkney-aws:
You can actually suppress the duplicate code linter with:
```
// jscpd:ignore-start
// jscpd:ignore-end
``` 
it might make sense in this case -- since the old implementation is just going to be dropped when we release this through flare. Maybe make an issue somewhere on github? (or internal?) to keep track that theres duplicate implementations, since if anyone wants to change customizations they need to change it in both places

I'm also wondering if you can just export the function calls from the customizations in core and then call those functions in your class? that way if anyone changes the implementations for now, your class would still work? WDYT?

zixlin7:
thanks! I will create an issue internally to track that the deprecation of the old file, most of these haven't been changed for a long time I don't see a strong need to export them into the class as it's cleaner this way and it gives us more motivation to deprecate instead of having the implementation in two places. Let me know if you a strong preference.


> You can actually suppress the duplicate code linter with:
> 
> ```
> // jscpd:ignore-start
> // jscpd:ignore-end
> ```
> 
> it might make sense in this case -- since the old implementation is just going to be dropped when we release this through flare. Maybe make an issue somewhere on github? (or internal?) to keep track that theres duplicate implementations, since if anyone wants to change customizations they need to change it in both places
> 
> I'm also wondering if you can just export the function calls from the customizations in core and then call those functions in your class? that way if anyone changes the implementations for now, your class would still work? WDYT?

jpinkney-aws (approved):

PR #6886, opened 2025-03-28
feat(amazonq lsp): add authFollowUpClicked implementation

opieter-aws:
## Problem
There was no implementation for the `authFollowUpClicked` message received from LSP for Amazon Q. Because of that, the `authenticate` button that appears in case of expired SSO connection had no effect.

## Solution
I added the implementation as follows:
* If the received `authFollowUpType` is `'re-auth'` or `'missing_scopes'`, initiate a re-authentication
* If the received `authFollowUpType` is `'full-auth'` or `'use-supported-auth'`, delete the connection so a new full authentication flow kicks off

I added a unit test file for this message type, which is extendable for every command handled in `registerLanguageServerEventListener`

---

### DIFF
--- a/packages/amazonq/src/lsp/chat/messages.ts
+++ b/packages/amazonq/src/lsp/chat/messages.ts
@@ -9,6 +9,7 @@ import {
     AUTH_FOLLOW_UP_CLICKED,
     CHAT_OPTIONS,
     COPY_TO_CLIPBOARD,
+    AuthFollowUpType,
 } from '@aws/chat-client-ui-types'
 import {
     ChatResult,
@@ -25,6 +26,7 @@ import { window } from 'vscode'
 import { Disposable, LanguageClient, Position, State, TextDocumentIdentifier } from 'vscode-languageclient'
 import * as jose from 'jose'
 import { AmazonQChatViewProvider } from './webviewProvider'
+import { AuthUtil } from 'aws-core-vscode/codewhisperer'
 
 export function registerLanguageServerEventListener(languageClient: LanguageClient, provider: AmazonQChatViewProvider) {
     languageClient.onDidChangeState(({ oldState, newState }) => {
@@ -77,10 +79,33 @@ export function registerMessageListeners(
                 })
                 break
             }
-            case AUTH_FOLLOW_UP_CLICKED:
-                // TODO hook this into auth
+            case AUTH_FOLLOW_UP_CLICKED: {
                 languageClient.info('[VSCode Client] AuthFollowUp clicked')
+                let authType = message.params.authFollowupType
+                const reAuthTypes: AuthFollowUpType[] = ['re-auth', 'missing_scopes']
+                const fullAuthTypes: AuthFollowUpType[] = ['full-auth', 'use-supported-auth']
+
+                if (reAuthTypes.includes(authType)) {
+                    try {
+                        await AuthUtil.instance.reauthenticate()
+                    } catch (e) {
+                        languageClient.error(
+                            `[VSCode Client] Failed to re-authenticate after AUTH_FOLLOW_UP_CLICKED: ${e}`
+                        )
+                    }
+                }
+
+                if (fullAuthTypes.includes(authType)) {
+                    try {
+                        await AuthUtil.instance.secondaryAuth.deleteConnection()
+                    } catch (e) {
+                        languageClient.error(
+                            `[VSCode Client] Failed to authenticate after AUTH_FOLLOW_UP_CLICKED: ${e}`
+                        )
+                    }
+                }
                 break
+            }
             case chatRequestType.method: {
                 const partialResultToken = uuidv4()
                 const chatDisposable = languageClient.onProgress(chatRequestType, partialResultToken, (partialResult) =>
--- a/packages/amazonq/src/lsp/chat/messages.ts
+++ b/packages/amazonq/src/lsp/chat/messages.ts
@@ -90,7 +90,7 @@ export function registerMessageListeners(
                         await AuthUtil.instance.reauthenticate()
                     } catch (e) {
                         languageClient.error(
-                            `[VSCode Client] Failed to re-authenticate after AUTH_FOLLOW_UP_CLICKED: ${e}`
+                            `[VSCode Client] Failed to re-authenticate after AUTH_FOLLOW_UP_CLICKED: ${(e as Error).message}`
                         )
                     }
                 }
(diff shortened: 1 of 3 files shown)


packages/amazonq/test/unit/amazonq/lsp/chat/messages.test.ts
```diff
@@ -0,0 +1,119 @@
...
+describe('registerMessageListeners', () => {
+    let languageClient: LanguageClient
+    let provider: AmazonQChatViewProvider
+    let sandbox: sinon.SinonSandbox
+    let messageHandler: (message: any) => void | Promise<void>
+    let errorStub: sinon.SinonStub
+
+    beforeEach(() => {
+        sandbox = sinon.createSandbox()
+        errorStub = sandbox.stub()
```
justinmk3, line 22: why is this needed? why can't the actual logs be checked?

stubs are a fragile, and shallow, way to build tests. our codebase guidelines discourage them.

packages/amazonq/test/unit/amazonq/lsp/chat/messages.test.ts
```diff
@@ -0,0 +1,119 @@
...
+                params: {
+                    authFollowupType: 're-auth',
+                },
+            })
+
+            sinon.assert.calledOnce(reauthenticateStub)
+            sinon.assert.notCalled(deleteConnectionStub)
+        })
+
+        it('should handle full authentication request', async () => {
```
justinmk3, line 79: "should" is never needed in test names
```suggestion
        it('handles full authentication request', async () => {
```

packages/amazonq/test/unit/amazonq/lsp/chat/messages.test.ts
```diff
@@ -0,0 +1,119 @@
...
+                params: {
+                    authFollowupType: 're-auth',
+                },
+            })
+
+            sinon.assert.calledOnce(errorStub)
+            sinon.assert.calledWith(errorStub, sinon.match(/Failed to re-authenticate/))
+        })
+
+        it('should log error if full authentication fails', async () => {
```
justinmk3, line 105: this test is nearly a duplicate of the one above.

although over-abstracting can be a cost because of indirection, 1 level of indirection in the form of a local function that deduplicates code is a cheap way to make the tests much clearer and readable, while also reducing maintenance costs.

```
testFailure('re-auth')
testFailure('full-auth')
```

packages/amazonq/src/lsp/chat/messages.ts
```diff
@@ -77,10 +79,33 @@ export function registerMessageListeners(
...
+                const authType = message.params.authFollowupType
+                const reAuthTypes: AuthFollowUpType[] = ['re-auth', 'missing_scopes']
+                const fullAuthTypes: AuthFollowUpType[] = ['full-auth', 'use-supported-auth']
+
+                if (reAuthTypes.includes(authType)) {
+                    try {
+                        await AuthUtil.instance.reauthenticate()
+                    } catch (e) {
+                        languageClient.error(
+                            `[VSCode Client] Failed to re-authenticate after AUTH_FOLLOW_UP_CLICKED: ${e}`
```
justinmk3, line 93: is the stringified `e` going to be useful or should `e.message` (and perhaps other key properties) be passed?

note that stacktraces are scrambled in production (due to webpack).

```suggestion
                            `[VSCode Client] Failed to re-authenticate after AUTH_FOLLOW_UP_CLICKED: ${(e as Error).message}`
```

packages/amazonq/src/lsp/chat/messages.ts
```diff
@@ -77,10 +79,33 @@ export function registerMessageListeners(
...
+                        )
+                    }
+                }
+
+                if (fullAuthTypes.includes(authType)) {
+                    try {
+                        await AuthUtil.instance.secondaryAuth.deleteConnection()
+                    } catch (e) {
+                        languageClient.error(
+                            `[VSCode Client] Failed to authenticate after AUTH_FOLLOW_UP_CLICKED: ${e}`
```
justinmk3, line 103: Does the server (or `languageClient`) not know which client is calling calling `languageClient.error()`? If it knows, it should do the prefixing, instead of requiring that in every call.
```suggestion
                            `Failed to authenticate after AUTH_FOLLOW_UP_CLICKED: ${e}`
```

opieter-aws: I am not sure. Maybe @jpinkney-aws can give some input on that. This is an existing pattern that is used consistently inside this file, which I continued. If we decide to change that, that's fine with me, but I'd propose to do it out of scope of this change and fix logging in the entire file then

packages/amazonq/test/unit/amazonq/lsp/chat/messages.test.ts
opieter-aws: What do you mean with `Why can't the actual logs be checked?`

IMO this is the best way to test the behavior here, open to suggestions. Stubs are used aplenty in the code base, I followed that pattern.

packages/amazonq/test/unit/amazonq/lsp/chat/messages.test.ts
opieter-aws: Will update

packages/amazonq/test/unit/amazonq/lsp/chat/messages.test.ts
opieter-aws: Agreed, Jest has [elegant solutions](https://jestjs.io/docs/api#testeachtablename-fn-timeout) for this, but unfortunately it's not used in the codebase yet. 

I am happy to refactor into a loop if you prefer

packages/amazonq/src/lsp/chat/messages.ts
jpinkney-aws: It originally passes in `env.appName` when setting up the language server so technically. The original code was just copied from the language-servers repo as non prod ready code

packages/core/src/amazonq/index.ts
```diff
@@ -46,6 +46,8 @@ export { extractAuthFollowUp } from './util/authUtils'
 export { Messenger } from './commons/connector/baseMessenger'
 export * from './lsp/config'
 export * as WorkspaceLspInstaller from './lsp/workspaceInstaller'
+export { SecondaryAuth } from '../auth/secondaryAuth'
+export { Connection } from '../auth/connection'
```
justinmk3, line 50:
```suggestion
export * as secondaryAuth from '../auth/secondaryAuth'
export * as authConnection from '../auth/connection'
```

jpinkney-aws (approved):

PR #6893, opened 2025-03-31
feat(amazonq): conversation persistence, view and search chat history

avi-alpert:
## Problem
Users lose all chats when they close VSCode, and there's no way to browse through chat history. Users also cant export their conversations to an easily shareable format.

## Solution
Automatically persist conversations to JSON files in `~/.aws/amazonq/history`, one for each workspace where Amazon Q chats occur. Add chat history and chat export buttons to top of Amazon Q toolbar. Clicking on the chat history button allows users to browse and search through chat history. Users click on an old conversation to open it back up (currently open conversations are in bold). Clicking on chat export button allows users to save chat transcript as a markdown or html.

Note: persistence + history is only for Q Chat Tabs (not /dev, /doc, /transform, etc.)

Screenshot 2025-03-31 at 2 49 27 PM

---


packages/core/src/amazonq/webview/ui/detailedList/detailedListConnector.ts
```diff
@@ -0,0 +1,94 @@
...
+            listType: this.type,
+            action,
+        })
+    }
+
+    onKeyPress = (e: KeyboardEvent) => {
+        if (e.key === 'Escape') {
+            this.closeList()
+        } else if (e.key === 'Enter') {
+            // todo: call onItemSelect on Enter
```
avi-alpert, line 87: waiting for mynah-ui to provide TargetElementId, should be included in prod release tomorrow

packages/core/package.json
```diff
@@ -554,6 +555,7 @@
         "js-yaml": "^4.1.0",
         "jsonc-parser": "^3.2.0",
         "lodash": "^4.17.21",
+        "lokijs": "^1.5.12",
```
justinmk3, line 558: why is this required

packages/core/src/shared/clients/chatDb/chatDb.ts
```diff
@@ -0,0 +1,241 @@
...
+    private db: Loki
+    /**
+     * Keep track of which open tabs have a corresponding history entry. Maps tabIds to historyIds
+     */
+    private historyIdMapping: Map<string, string> = new Map()
+    private dbDirectory: string
+    initialized: boolean = false
+
+    constructor() {
+        this.dbDirectory = path.join(fs.getUserHomeDir(), '.aws', 'amazonq', 'history')
```
justinmk3, line 33: It is never required to join() string literals. Not on Windows, nor anywhere else. "/" slashes work fine.

```suggestion
        this.dbDirectory = path.join(fs.getUserHomeDir(), '.aws/amazonq/history')
```

packages/core/src/amazonq/webview/ui/apps/docChatConnector.ts
```diff
@@ -23,7 +23,6 @@ export interface ConnectorProps extends BaseConnectorProps {
     onUpdatePromptProgress: (tabID: string, progressField: ProgressField) => void
     onChatInputEnabled: (tabID: string, enabled: boolean) => void
     onUpdateAuthentication: (featureDevEnabled: boolean, authenticatingTabIDs: string[]) => void
-    onNewTab: (tabType: TabType) => void
```
leigaol, line 26: is this no longer needed in new mynahUI?

packages/core/src/shared/clients/chatDb/chatDb.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    setHistoryIdMapping(tabId: string, historyId: string) {
+        this.historyIdMapping.set(tabId, historyId)
+    }
+
+    getWorkspaceIdentifier() {
+        // Case 1: .code-workspace file (saved workspace)
+        const workspace = vscode.workspace.workspaceFile
+        if (workspace) {
+            crypto.createHash('md5').update(workspace.fsPath).digest('hex')
```
leigaol, line 62: maybe just return at L62

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
```diff
@@ -13,6 +13,7 @@ import { UserWrittenCodeTracker } from '../../../../codewhisperer/tracker/userWr
 
 export class ChatSession {
     private sessionId?: string
+    localHistoryHydrated: boolean = false
```
justinmk3, line 16: why is this boolean needed? why can't this answer be calculated from the actual state?

packages/core/src/shared/clients/chatDb/helpers.ts
justinmk3, line 1: If these "helpers" are related to the client, put this in `clients/chatDb.ts`. Don't create a new subdir in `clients/`.

And please use the name "util", like the rest of the codebase. It's confusing for multiple different names to mean the same thing. (Please follow this guidance for all future PRs: follow the codebase conventions, avoid adding entropy.)

packages/core/src/shared/clients/chatDb/chatDb.ts
```diff
@@ -0,0 +1,241 @@
...
+    groupTabsByDate,
+    Tab,
+    TabCollection,
+    updateOrCreateConversation,
+} from './helpers'
+import crypto from 'crypto'
+import path from 'path'
+import { fs } from '../../fs/fs'
+
+export class Database {
```
justinmk3, line 22: Surely this deserves some sort of docstring giving a hint about its purpose, or other key insights

packages/core/src/shared/clients/chatDb/chatDb.ts
justinmk3, line 1: Thank you for thinking about clients/ as a potential place for this to live. But it's probably not the right place. clients/ is for SDK clients.

We may need to introduce `shared/db/`. And for now, only `shared/db/chatDb.ts` will live in there.

packages/core/src/shared/clients/chatDb/helpers.ts
```diff
@@ -0,0 +1,211 @@
...
+            callback(err)
+        }
+    }
+
+    async saveDatabase(dbname: string, dbstring: string, callback: (err: Error | undefined) => void) {
+        try {
+            await this.ensureDirectory()
+            const filename = path.join(this.directory, dbname)
+
+            await fs.writeFile(filename, dbstring, 'utf8')
```
justinmk3, line 73: what happens if the history is corrupted or has partial writes, or multiple vscode instances try to save the file simultaneously? on Windows, this will have problems.

if you are ok with the current limitations, the limitations should be cleary called-out in the docstring for `FileSystemAdapter` (which currently has no docstring).

packages/core/package.json
avi-alpert: Yes thats the correct repo, and based on my research its stable and widely used. its a pretty lightweight library with 500k+ weekly downloads https://www.npmjs.com/package/lokijs

LokiJS includes helpful functionality out of the box like autoload, autosave, finding elements, and updating elements, and was easy to use. Alternatives considered were better-sqlite3 and sqlite3 but those are not web-compatible

packages/core/src/codewhispererChat/clients/chat/v0/chat.ts
avi-alpert: When an old conversation is resumed, we send the last 10 messages in the `history` field to the API https://github.com/aws/aws-toolkit-vscode/pull/6893/files#diff-f980157c047328129cbab2c74edd8fd69bb70108d1de7239330128b23371997cR1117, but this only has to be done once per session. so I needed a place to keep track of whether the session has been hydrated with local history

avi-alpert: will add a comment

leigaol (approved):

avi-alpert:
Rebased with master

jpinkney-aws (approved):

packages/core/src/shared/db/chatDb/util.ts
```diff
@@ -0,0 +1,223 @@
...
+export type Conversation = {
+    conversationId: string
+    clientType: ClientType
+    messages: ChatItem[]
+}
+
+export enum ClientType {
+    VSCode = 'IDE-VSCode',
+    JetBrains = 'IDE_JetBrains',
+    CLI = 'CLI',
```
volodkevych, line 33: CLI does not use mynah-ui, does it? What is the plan to make CLI work with the history, given native mynah-ui types, like `ChatItem`, are stored in it?

packages/core/src/shared/db/chatDb/util.ts
```diff
@@ -0,0 +1,223 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import fs from '../../fs/fs'
+import path from 'path'
+
+import { TabType } from '../../../amazonq/webview/ui/storages/tabsStorage'
+import { ChatItem, ChatItemButton, DetailedListItem, DetailedListItemGroup, MynahIconsType } from '@aws/mynah-ui'
```
volodkevych, line 9: It looks like you store native mynah-ui types into history. It can be not straightforward to migrate this to `language-servers` while preserving history. There's protocol abstracting UI from chat server business logic, meaning types in chat language server are non-UI specific.

packages/core/src/shared/db/chatDb/util.ts
avi-alpert: this was just added in here as a placeholder in case we ever had a shared history store in the future

packages/core/src/shared/db/chatDb/util.ts
avi-alpert: we could create a new type in flare, in history we are storing an object for each message that has body, type, codeReference, and relatedContent. so as long as we keep those four properties we can be backwards compatible

PR #6894, opened 2025-03-31
feat(amazonq): remote workspace context

leigaol:
## Problem
We did not utilize server side compute to help improve the quality of inline completion, chat, etc. 

## Solution
1. Add server side project context support.

---


aws-toolkit-vscode.code-workspace
```diff
@@ -1,19 +1,19 @@
...
-		{
-			"path": "packages/amazonq"
-		}
-	],
-	"settings": {
-		"typescript.tsdk": "node_modules/typescript/lib"
-	}
-}
\ No newline at end of file
+    "folders": [
```
justinmk3, line 2: is this a formatting-only change? please revert the formatting, and only include the relevant changes.

packages/core/package.nls.json
```diff
@@ -90,6 +90,7 @@
...
     "AWS.configuration.description.amazonq.shareContentWithAWS": "When checked, your content processed by Amazon Q may be used for service improvement (except for content processed for users with the Amazon Q Developer Pro Tier). Unchecking this box will cause AWS to delete any of your content used for that purpose. The information used to provide the Amazon Q service to you will not be affected. See the [Service Terms](https://aws.amazon.com/service-terms) for more details.",
     "AWS.configuration.description.amazonq.importRecommendation": "Amazon Q will add import statements with inline code suggestions when necessary.",
+    "AWS.configuration.description.amazonq.workspaceContext": "Allow Amazon Q to create remote workspace for quality improvements.",
```
justinmk3, line 93: Should this mention how it differs from `workspaceIndex`? Are both needed? Does `workspaceContext` affect `@workspace`, or what does it affect?

packages/core/src/codewhisperer/util/editorContext.ts
```diff
@@ -82,6 +82,22 @@ export function getFileRelativePath(editor: vscode.TextEditor): string {
     return relativePath.substring(0, CodeWhispererConstants.filenameCharsLimit)
 }
 
+async function getWorkspaceId(editor: vscode.TextEditor): Promise<string | undefined> {
+    try {
+        const workspaceIds: { workspaces: { workspaceRoot: string; workspaceId: string }[] } =
+            await vscode.commands.executeCommand('aws.amazonq.getWorkspaceId')
```
justinmk3, line 88: Why is this a command instead of a plain function?

packages/amazonq/src/lsp/client.ts
```diff
@@ -117,7 +127,7 @@ export async function startLanguageServer(
                 getLogger('amazonqLsp').error('Unable to update bearer token: %s', (e as Error).message)
                 clearInterval(authInterval)
             }
-        }, 300000) // every 5 minutes
+        }, 30 * 1000) // every 30 seconds
```
justinmk3, line 130: will this `setInterval` be removed as line 122 mentions?

packages/core/package.nls.json
leigaol: This will be updated before launch

packages/core/src/codewhisperer/util/editorContext.ts
leigaol: To avoid cyclic dependency. The Flare LSP code was in the packages/amazonq which imports packages/core, but the CW code here is in packages/core.

packages/amazonq/src/lsp/client.ts
leigaol: It depends on timing. If launch date is before  Flare auth migration, we need to keep this. This is a local JSON RPC call, we can make the interval even smaller.

packages/amazonq/src/lsp/config.ts
```diff
@@ -11,8 +11,8 @@ export interface ExtendedAmazonQLSPConfig extends LspConfig {
 }
 
 export const defaultAmazonQLspConfig: ExtendedAmazonQLSPConfig = {
-    manifestUrl: 'https://aws-toolkit-language-servers.amazonaws.com/codewhisperer/0/manifest.json',
-    supportedVersions: '^3.1.1',
+    manifestUrl: 'https://aws-toolkit-language-servers.amazonaws.com/remoteWorkspaceContext/0/manifest.json',
+    supportedVersions: '^1.0.0',
```
justinmk3, line 15: is this the final URL or will `/codewhisperer/` gain the features from the temporary `/remoteWorkspaceContext/` artifact (then this version would be 3.x) ?

packages/core/src/codewhisperer/util/editorContext.ts
justinmk3: What happens if the command isn't defined? Sometimes there are startup ordering issues which result in "command not found" errors. Using a plain function would avoid that and add confidence.

There are more robust ways of avoiding cyclic dependencies. Commands should be a last resort.

packages/amazonq/src/lsp/config.ts
leigaol: This is final, to my knowledge.

ege0zcan: Thre is no `/codewhisperer/` url at the moment. For now this is the final URL but I imagine we will need to change the URL in the near future because new artifacts are also being built

packages/amazonq/src/lsp/client.ts
```diff
@@ -143,6 +178,62 @@ export async function startLanguageServer(
...
+                )
+                const workspaceIdResp = await client.sendRequest(requestType.method, {
+                    section: 'aws.q.workspaceContext',
+                })
+                return workspaceIdResp
+            }),
+            vscode.workspace.onDidCreateFiles((e) => {
+                client.sendNotification('workspace/didCreateFiles', {
+                    files: e.files.map((it) => {
+                        return { uri: it.fsPath }
```
justinmk3, line 194: Does this event trigger for in-memory (non-file) editors? If so, should those be skipped, since the LSP can't do anything with them unless we send the file contents?

packages/amazonq/src/lsp/client.ts
```diff
@@ -134,7 +149,27 @@ export async function startLanguageServer(
...
+                    settings: {
+                        profileArn: AuthUtil.instance.regionProfileManager.activeRegionProfile?.arn,
+                    },
+                })
+                client.info(
+                    `Client: Updated Amazon Q Profile ${AuthUtil.instance.regionProfileManager.activeRegionProfile?.arn} to Amazon Q LSP`,
+                    result
+                )
+            } catch (err) {
+                client.error('Error when setting Q Developer Profile to Amazon Q LSP', err)
```
justinmk3, line 167: non-blocker/future reference: these should probably be logged in `logger`, not `client`

jpinkney-aws (approved):
Merging to prepare mainline for ssc

PR #6895, opened 2025-03-31
feat(chat): Add warning UI for high risk bash command

tsmithsz:
## Problem
AppSec has required that we give the user an indication if they are about to run a potentially dangerous command

## Solution

- Parse bash command to determine whether or not commands are dangerous or not.
- Display message warning to user

## Tests

Screenshot 2025-03-31 at 1 03 14 PM

---

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -492,7 +492,8 @@
         "umd-compat-loader": "^2.1.2",
         "vue-loader": "^17.2.2",
         "vue-style-loader": "^4.1.3",
-        "webfont": "^11.2.26"
+        "webfont": "^11.2.26",
+        "shlex": "^2.1.2"
     },
     "dependencies": {
         "@amzn/amazon-q-developer-streaming-client": "file:../../src.gen/@amzn/amazon-q-developer-streaming-client",
--- a/packages/core/src/codewhispererChat/tools/executeBash.ts
+++ b/packages/core/src/codewhispererChat/tools/executeBash.ts
@@ -5,18 +5,128 @@
 
 import { Writable } from 'stream'
 import { getLogger } from '../../shared/logger/logger'
-import { fs } from '../../shared/fs/fs' // e.g. for getUserHomeDir()
+import { fs } from '../../shared/fs/fs'
 import { ChildProcess, ChildProcessOptions } from '../../shared/utilities/processUtils'
 import { InvokeOutput, OutputKind, sanitizePath } from './toolShared'
+import { split } from 'shlex'
 
-export const readOnlyCommands: string[] = ['ls', 'cat', 'echo', 'pwd', 'which', 'head', 'tail']
+export enum CommandCategory {
+    ReadOnly,
+    HighRisk,
+    Destructive,
+}
+
+export const dangerousPatterns = new Set(['<(', '$(', '`', '>', '&&', '||'])
+export const commandCategories = new Map<string, CommandCategory>([
+    // ReadOnly commands
+    ['ls', CommandCategory.ReadOnly],
+    ['cat', CommandCategory.ReadOnly],
+    ['bat', CommandCategory.ReadOnly],
+    ['pwd', CommandCategory.ReadOnly],
+    ['echo', CommandCategory.ReadOnly],
+    ['file', CommandCategory.ReadOnly],
+    ['less', CommandCategory.ReadOnly],
+    ['more', CommandCategory.ReadOnly],
+    ['tree', CommandCategory.ReadOnly],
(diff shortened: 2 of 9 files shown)


packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -5,20 +5,127 @@
...
+    'dig',
+    'grep',
+    'wc',
+    'sort',
+    'diff',
+    'head',
+    'tail',
+])
+export const dangerousPatterns = new Set(['<(', '$(', '`', '>', '&&', '||'])
+const highRiskCommands = new Map<string, string>([
```
jguoamz, line 48: where did we get the list from?

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -46,22 +153,60 @@ export class ExecuteBash {
...
 
-            const command = args[0]
-            return !readOnlyCommands.includes(command)
+            for (const cmdArgs of allCommands) {
+                if (cmdArgs.length === 0) {
+                    return { requiresAcceptance: true, isHighRisk: false }
+                }
+
+                if (cmdArgs.some((arg) => Array.from(dangerousPatterns).some((pattern) => arg.includes(pattern)))) {
+                    return { requiresAcceptance: true, isHighRisk: false }
```
jguoamz, line 190: isn't it possible for high risk command to contain dangerous pattern? This logic would make that not high risk anymore?

tsmithsz: Updated to make dangerousPatterns high risk

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -46,22 +153,60 @@ export class ExecuteBash {
...
             }
 
-            const command = args[0]
-            return !readOnlyCommands.includes(command)
+            for (const cmdArgs of allCommands) {
+                if (cmdArgs.length === 0) {
+                    return { requiresAcceptance: true, isHighRisk: false }
+                }
+
+                if (cmdArgs.some((arg) => Array.from(dangerousPatterns).some((pattern) => arg.includes(pattern)))) {
```
zixlin7, line 189: can we adjust the order of these operations and move this one after the other two map lookup? this one is more expensive and we can avoid it if not necessary.

ammathu: I am thinking what if we missed some cmd in the `highRiskCommands`? Do you think its better if we say do not show warning if it is not in the `readOnlyCommands` list else show for all others?

zixlin7: in that case just move it after the highRiskCommands check should be good? If it's already matching highRiskCommands there is no need to check anything else; if it doesn't you still check for the pattern nothing is missed.

tsmithsz: Reordered the log. Raising the suggestion to treat all non-readOnlyCommands as high risk with AppSec POC

packages/core/package.json
```diff
@@ -492,7 +492,8 @@
         "umd-compat-loader": "^2.1.2",
         "vue-loader": "^17.2.2",
         "vue-style-loader": "^4.1.3",
-        "webfont": "^11.2.26"
+        "webfont": "^11.2.26",
+        "shlex": "^2.1.2"
```
jpinkney-aws, line 496: how does adding this package change the package-lock by 5405 😨. Also can you summarize what this package is and why we need it?

tsmithsz: It's only 8 lines now after rebasing.

After discussion with @justinmk3, this package is needed for parsing shell commands.

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -46,22 +156,67 @@ export class ExecuteBash {
...
+                if (cmdArgs.length === 0) {
+                    return { requiresAcceptance: true }
+                }
+
+                const command = cmdArgs[0]
+                const category = commandCategories.get(command)
+
+                switch (category) {
+                    case CommandCategory.Destructive:
+                        return { requiresAcceptance: true, warning: highRiskCommandWarningMessage }
```
zixlin7, line 197: should it be `destructiveCommandWarningMessage` for this case? and high risk message for the one below?

tsmithsz: Fixed

zixlin7 (approved):

jpinkney-aws (approved):

PR #6899, opened 2025-04-01
fix(lambda): checking if folder exists

Vandita2020:
## Problem
In the existing Serverless Land project, we need to check if a folder with the same name already exists before downloading the code. This could help avoid potential issues that might arise from overwriting existing files or folders. 

## Solution
The code checks if a folder with the same name already exists, and prompts the user to choose whether to override it or not. This allows the user to decide how to handle potential conflicts with existing files or folders.

---

### DIFF
--- a/packages/core/src/awsService/appBuilder/serverlessLand/main.ts
+++ b/packages/core/src/awsService/appBuilder/serverlessLand/main.ts
@@ -15,7 +15,7 @@ import { ExtContext } from '../../../shared/extensions'
 import { addFolderToWorkspace } from '../../../shared/utilities/workspaceUtils'
 import { ToolkitError } from '../../../shared/errors'
 import { fs } from '../../../shared/fs/fs'
-import { confirmOverwriteIfExists } from '../../../shared/utilities/messages'
+import { handleOverwriteConflict } from '../../../shared/utilities/messages'
 import { getPattern } from '../../../shared/utilities/downloadPatterns'
 import { MetadataManager } from './metadataManager'
 
@@ -91,14 +91,8 @@ export async function downloadPatternCode(config: CreateServerlessLandWizardForm
     const fullAssetName = assetName + '.zip'
     const location = vscode.Uri.joinPath(config.location, config.name)
 
-    const shouldProceed = await confirmOverwriteIfExists(location, config.name)
-    if (!shouldProceed) {
-        throw new ToolkitError(`Folder already exists: ${config.name}`)
-    }
+    await handleOverwriteConflict(location)
 
-    if (await fs.exists(location)) {
-        await fs.delete(location, { recursive: true, force: true })
-    }
     try {
         await getPattern(serverlessLandOwner, serverlessLandRepo, fullAssetName, location, true)
     } catch (error) {
--- a/packages/core/src/shared/utilities/messages.ts
+++ b/packages/core/src/shared/utilities/messages.ts
@@ -5,6 +5,7 @@
 
 import * as vscode from 'vscode'
 import * as nls from 'vscode-nls'
+import * as path from 'path'
 import * as localizedText from '../localizedText'
 import { getLogger } from '../../shared/logger/logger'
 import { ProgressEntry } from '../../shared/vscode/window'
@@ -14,6 +15,7 @@ import { Timeout } from './timeoutUtils'
 import { addCodiconToString } from './textUtilities'
 import { getIcon, codicon } from '../icons'
 import globals from '../extensionGlobals'
+import { ToolkitError } from '../../shared/errors'
 import { fs } from '../../shared/fs/fs'
 import { openUrl } from './vsCodeUtils'
 import { AmazonQPromptSettings, ToolkitPromptSettings } from '../../shared/settings'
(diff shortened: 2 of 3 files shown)


packages/core/src/awsService/appBuilder/serverlessLand/main.ts
```diff
@@ -89,6 +89,22 @@ export async function launchProjectCreationWizard(
...
+            localize(
+                'AWS.toolkit.serverlessLand.fileExistsPrompt',
+                '{0} already exists in the selected directory, overwrite?',
+                config.name
+            ),
+            'Yes',
+            'No'
+        )
+        if (choice !== 'Yes') {
+            return Promise.reject(new ToolkitError(`A folder named ${config.name} already exists in this path.`))
```
justinmk3, line 104: `throw new ToolkitError(...)` should work?
```suggestion
            throw new ToolkitError(`A folder named ${config.name} already exists in this path.`)
```

packages/core/src/test/awsService/appBuilder/serverlessLand/main.test.ts
```diff
@@ -114,16 +118,51 @@ describe('downloadPatternCode', () => {
...
         getPatternStub.rejects(error)
         try {
-            await downloadPatternCode(mockConfig, mockAssetName)
+            await downloadPatternCode(mockConfig, mockConfig.assetName)
             assert.fail('Expected an error to be thrown')
         } catch (err: any) {
             assert.strictEqual(err.message, 'Failed to download pattern: Error: Download failed')
         }
     })
+    it('downloads pattern when directory exists and user confirms overwrite', async function () {
```
justinmk3, line 130: This is a huge amount of test code (and brittle, because of the over-use of mocks/stubs), just to test a  function that checks for a file before overwriting it.

And this is another very strong example of why using shared code is valuable: if this code was generalized, we wouldn't need this new test code at all. 

Every batch of 30+ lines of redundant code is almost entirely a cost, for very small benefit. So it's important to think about re-using code, especially for cases like this where there is a clear path to doing so. Look at messages.ts  or other modules to see if there's existing shared util that does this check-before-save-and-show-message logic. It's really common, and if it doesn't already exist, then this is a good time to do it.

justinmk3 (approved):

packages/core/src/awsService/appBuilder/serverlessLand/main.ts
```diff
@@ -89,6 +90,15 @@ export async function launchProjectCreationWizard(
...
     const fullAssetName = assetName + '.zip'
     const location = vscode.Uri.joinPath(config.location, config.name)
+
+    const shouldProceed = await confirmOverwriteIfExists(location, config.name)
+    if (!shouldProceed) {
+        throw new ToolkitError(`Folder already exists: ${config.name}`)
+    }
+
+    if (await fs.exists(location)) {
+        await vscode.workspace.fs.delete(location)
```
rhamilt: Is there a reason we use `fs` from the toolkit for the check and then `vscode.workspace.fs` to actually do the delete? I think there's just an `fs.delete(location, {recursive?, force?})`

packages/core/src/shared/utilities/messages.ts
```diff
@@ -140,6 +141,29 @@ export async function showViewLogsMessage(
     })
 }
 
+/**
+ * Checks if a path exists and prompts user for overwrite confirmation if it does.
+ * @param path The file or directory path to check
+ * @param itemName The name of the item for display in the message
+ * @returns Promise<boolean> - true if should proceed (path doesn't exist or user confirmed overwrite)
+ */
+export async function confirmOverwriteIfExists(path: vscode.Uri, itemName: string): Promise<boolean> {
```
roger-zhangg, line 150: Just to confirm, the path here should be `userSelectedPath + ServerlessProjectName` right? My concern is to not accidentally delete customer workspace

packages/core/src/awsService/appBuilder/serverlessLand/main.ts
Vandita2020: Thanks for the suggestion! I have made the asked changes :)

packages/core/src/shared/utilities/messages.ts
```diff
@@ -140,6 +142,36 @@ export async function showViewLogsMessage(
...
+    if (!(await fs.exists(path))) {
+        return true
+    }
+
+    const choice = showConfirmationMessage({
+        prompt: localize(
+            'AWS.toolkit.confirmOverwrite',
+            '{0} already exists in the selected directory, overwrite?',
+            itemName
+        ),
```
roger-zhangg:
```suggestion
    const choice = showConfirmationMessage({
        prompt: localize(
            'AWS.toolkit.confirmOverwrite',
            '{0} already exists in the selected directory {1}, overwrite?',
            itemName,
            path.fsPath
        ),
```

packages/core/src/shared/utilities/messages.ts
```diff
@@ -140,6 +142,36 @@ export async function showViewLogsMessage(
     })
 }
 
+/**
+ * Checks if a path exists and prompts user for overwrite confirmation if it does.
+ * @param path The file or directory path to check
+ * @param itemName The name of the item for display in the message
+ * @returns Promise<boolean> - true if should proceed (path doesn't exist or user confirmed overwrite)
+ */
+export async function handleOverwriteConflict(path: vscode.Uri, itemName: string): Promise<boolean> {
```
roger-zhangg: Maybe just use path here, currently this itemName is kind of redundant and make this function confusing

packages/core/src/shared/utilities/messages.ts
```diff
@@ -140,6 +142,36 @@ export async function showViewLogsMessage(
...
+        confirm: localize('AWS.generic.overwrite', 'Yes'),
+        cancel: localize('AWS.generic.cancel', 'No'),
+        type: 'warning',
+    })
+
+    if (!choice) {
+        throw new ToolkitError(`Folder already exists: ${itemName}`)
+    }
+
+    await fs.delete(path, { recursive: true, force: true })
```
roger-zhangg: this could fail right? Consider try catch

packages/core/src/shared/utilities/messages.ts
Vandita2020: I'm using it in the error message to output the actual name of the folder that exists.

laileni-aws (approved):
LGTM

PR #6906, opened 2025-04-01
fix(chat): fix for chat history for multuple tabs and user tool ignore

ashishrp-aws:
## Problem
- History storage not set up for each tabs.
- If user ignores tool use and moves on, then request fails.

## Solution
- Fixed chat history for multiple tab support
- User cancellation message is added to request in this case.

---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -87,11 +87,11 @@ import {
     defaultContextLengths,
 } from '../../constants'
 import { ChatSession } from '../../clients/chat/v0/chat'
-import { ChatHistoryManager } from '../../storages/chatHistory'
 import { amazonQTabSuffix } from '../../../shared/constants'
 import { OutputKind } from '../../tools/toolShared'
 import { ToolUtils, Tool, ToolType } from '../../tools/toolUtils'
 import { ChatStream } from '../../tools/chatStream'
+import { ChatHistoryStorage } from '../../storages/chatHistoryStorage'
 import { FsWrite, FsWriteParams } from '../../tools/fsWrite'
 import { tempDirPath } from '../../../shared/filesystemUtilities'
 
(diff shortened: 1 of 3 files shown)


packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -1051,7 +1049,7 @@ export class ChatController {
                     codeQuery: lastTriggerEvent.context?.focusAreaContext?.names,
                     userIntent: message.userIntent,
                     customization: getSelectedCustomization(),
-                    chatHistory: this.chatHistoryManager.getHistory(),
+                    chatHistory: this.chatHistoryStorage.getHistory(message.tabID).getHistory(),
```
ctlai95, line 1052: nit: the naming is a bit confusing, chained `getHistory` calls

ashishrp-aws: will change it getTabHistory

tsmithsz (approved):

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -1327,16 +1325,28 @@ export class ChatController {
...
+                userInputMessageContext: {
+                    tools: tools,
+                    ...(triggerPayload.toolResults && { toolResults: triggerPayload.toolResults }),
+                },
+            },
+        }
+        const fixedHistoryMessage = chatHistory.fixHistory(newUserMessage)
+        if (fixedHistoryMessage.userInputMessage?.userInputMessageContext) {
+            triggerPayload.toolResults = fixedHistoryMessage.userInputMessage.userInputMessageContext.toolResults
         }
```
ctlai95, line 1344: why can't we do this inside `fixHistory`?

ashishrp-aws: We can but i'm not sending triggerPayload to fixHistory. fixHistory could be used for other operations, so i set it up to take in only ChatMessage

zixlin7 (approved):

Issue #6913, opened 2025-04-02
unreliable test: "SAM runBuild"


justinmk3:
## Test Details

-   Name of test:
-   OS of failure:
-   VSCode version of failure (minimum/insider/stable):
-   Link to failing run:
-   Link to failing test:

## Log of Test Failure

```
  1) SAM runBuild
       :) path
         [entry: command palette] use samconfig should instantiate correct process in terminal:

      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  (0, assert_1.default)(spyRunInterminal.calledOnce)

      + expected - actual

      -false
      +true
      
      at verifyCorrectDependencyCall (/Users/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/core/src/test/shared/sam/build.test.ts:343:19)
```

vicheey:
Do you have a link to one of the test failuire?

PR #6920, opened 2025-04-03
fix(lambda): flaky test due to terminal instance

vicheey:
## Problem

fix https://github.com/aws/aws-toolkit-vscode/issues/6913

## Solution
Move the spy instantiation closer to reduce side effects. Add delay before start running command so additional for clean up to reduce chance of flakiness. Change assertion from true/false to count comparison to give more info for debugging in case the flakiness persists into the future. 

---

### DIFF
--- a/packages/core/src/test/shared/sam/build.test.ts
+++ b/packages/core/src/test/shared/sam/build.test.ts
@@ -280,8 +280,6 @@ describe('SAM runBuild', () => {
     let mockChildProcessClass: sinon.SinonStub
     let mockSamBuildChildProcess: sinon.SinonStub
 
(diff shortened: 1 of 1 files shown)


packages/core/src/test/shared/sam/build.test.ts
```diff
@@ -398,7 +398,8 @@ describe('SAM runBuild', () => {
                 .build()
 
             // Invoke sync command from command palette
-            await runBuild()
+            // Instead of await runBuild(), prefer this to avoid flakiness due to race condition
```
Hweinstock, line 401: What is the race condition? Could it affect users or is it test specific?

Hweinstock (approved):

PR #6928, opened 2025-04-03
feat(chat): pair programming mode

chungjac:
## Solution
- Initial PR for adding pair programming mode toggle

Screenshot 2025-04-03 at 7 09 35 PM

#### TODO:
- Tooltip component to tell user what icon is
- Show message to user when toggling modes
- Update component when mynah releases new component
---

### DIFF
--- a/packages/core/src/codewhispererChat/constants.ts
+++ b/packages/core/src/codewhispererChat/constants.ts
@@ -30,6 +30,11 @@ export const tools: Tool[] = Object.entries(toolsJson).map(([, toolSpec]) => ({
         inputSchema: { json: toolSpec.inputSchema },
     },
 }))
+
+export const noWriteTools: Tool[] = tools.filter(
+    (tool) => !['fsWrite', 'executeBash'].includes(tool.toolSpecification?.name || '')
+)
+
 export const defaultContextLengths: ContextLengths = {
     additionalContextLengths: {
         fileContextLength: 0,
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -769,12 +769,15 @@ export class ChatController {
     }
 
     private async processPromptInputOptionChange(message: PromptInputOptionChange) {
+        const session = this.sessionStorage.getSession(message.tabID)
         const promptTypeValue = message.optionsValues['prompt-type']
         // TODO: display message: You turned off pair programmer mode. Q will not include code diffs or run commands in the chat.
         if (promptTypeValue === 'pair-programming-on') {
-            this.chatHistoryStorage.setTabAvailableTools(message.tabID, tools)
+            this.chatHistoryStorage.setTools(message.tabID, tools)
+            session.setPairProgrammingModeOn(true)
         } else {
-            this.chatHistoryStorage.setTabAvailableTools(message.tabID, noWriteTools)
+            this.chatHistoryStorage.setTools(message.tabID, noWriteTools)
+            session.setPairProgrammingModeOn(false)
         }
     }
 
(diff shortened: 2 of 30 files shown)


packages/core/src/codewhispererChat/storages/chatHistory.ts
```diff
@@ -46,6 +46,10 @@ export class ChatHistoryManager {
         this.conversationId = conversationId
     }
 
+    public setTools(tools: Tool[]) {
+        this.tools = tools
+    }
+
```
jguoamz, line 52: do we need this anymore?

packages/core/src/codewhispererChat/storages/chatHistoryStorage.ts
```diff
@@ -40,4 +41,8 @@ export class ChatHistoryStorage {
     public deleteHistory(tabId: string) {
         this.histories.delete(tabId)
     }
+
+    public setTools(tabId: string, tools: Tool[]) {
+        this.histories.get(tabId)?.setTools(tools)
+    }
```
jguoamz, line 47: same question here

packages/core/src/codewhispererChat/storages/chatHistory.ts
chungjac: This setter is needed as the current tools are used in chatHistory.ts:
https://github.com/aws/aws-toolkit-vscode/blob/093d5bcbce777c88cf18c76b52738610263d1fc0/packages/core/src/codewhispererChat/storages/chatHistory.ts#L35

packages/core/src/codewhispererChat/storages/chatHistoryStorage.ts
chungjac: Same as above. Passing it all the way to `chatHistory.ts` so that it can have context of which tools are available

packages/core/src/codewhispererChat/storages/chatHistory.ts
jguoamz: but we store pairProgrammingModeOn in the session and just use this to pick between two static lists, why do we need tools stored in the history? Where are we using this?

packages/core/src/codewhispererChat/constants.ts
```diff
@@ -30,6 +30,11 @@ export const tools: Tool[] = Object.entries(toolsJson).map(([, toolSpec]) => ({
         inputSchema: { json: toolSpec.inputSchema },
     },
 }))
+
+export const noWriteTools: Tool[] = tools.filter(
```
jguoamz, line 34: Not related to this line, but let's add toolName validation for toolUseEvent

packages/core/src/codewhispererChat/storages/chatHistory.ts
chungjac: Using `session` now instead of storing tools field in `ChatHistoryManager`

jguoamz (approved):

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -631,6 +638,69 @@ export class ChatController {
         telemetry.ui_click.emit({ elementId: 'amazonq_createSavedPrompt' })
     }
 
+    private async processUnavailableToolUseMessage(message: CustomFormActionMessage) {
+        const tabID = message.tabID
+        if (!tabID) {
+            return
+        }
+        this.editorContextExtractor
```
zixlin7, line 645: can follow up in another PR, seems like a lot of these have the same flow of extract context and call generateResponse, please consider extract the common part out and reuse in this file.

zixlin7 (approved):

packages/core/src/codewhispererChat/controllers/chat/controller.ts
chungjac: yes, can follow up in another pr

ctlai95 (approved):

PR #6931, opened 2025-04-03
feat(chat): group consecutive read tool messages

ctlai95:
## Problem

LLM will sometimes try to execute the `fsRead` tool (and `listDirectory` tool) many times in succession, when this happens we don't want to show every read message.

## Solution

Update the existing chat message that already shows a read message if it exists.

### Detailed changes
- Added a new type of UiMessage `ToolMessage` specifically for handling tool chat messages
- `ToolMessage` will call MynahUI's `updateChatAnswerWithMessageId` API
- For every "group" of consecutive read messages we store the first occurrence of the read tool's toolUseId in session storage and use that as the messageId to update
- Updated `ChatStream` class to take in this messageId to update for the current message

https://github.com/user-attachments/assets/8df75d4d-c15f-4e21-9867-4ad3ca8f3233

---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -663,9 +663,16 @@ export class ChatController {
                     try {
                         await ToolUtils.validate(tool)
 
-                        const chatStream = new ChatStream(this.messenger, tabID, triggerID, toolUse, {
-                            requiresAcceptance: false,
-                        })
+                        const chatStream = new ChatStream(
+                            this.messenger,
+                            tabID,
+                            triggerID,
+                            // Pass in a different toolUseId so that the output does not overwrite
+                            // any previous messages
+                            { ...toolUse, toolUseId: `${toolUse.toolUseId}-output` },
+                            { requiresAcceptance: false },
+                            undefined
+                        )
                         const output = await ToolUtils.invoke(tool, chatStream)
 
                         toolResults.push({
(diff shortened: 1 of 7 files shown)


jguoamz (approved):

ashishrp-aws (approved):

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -221,11 +222,35 @@ export class Messenger {
                             if (tool.type === ToolType.FsWrite) {
                                 session.setShowDiffOnFileWrite(true)
                             }
+                            if (
```
zixlin7, line 225: I know not specific to your PR but this function is way too long and difficult to read, can we take this part out into a helper function?

packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -663,9 +663,16 @@ export class ChatController {
...
-                            requiresAcceptance: false,
-                        })
+                        const chatStream = new ChatStream(
+                            this.messenger,
+                            tabID,
+                            triggerID,
+                            // Pass in a different toolUseId so that the output does not overwrite
+                            // any previous messages
+                            { ...toolUse, toolUseId: `${toolUse.toolUseId}-output` },
+                            { requiresAcceptance: false },
```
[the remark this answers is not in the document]
ctlai95: No, each toolUseEvent should correspond to a single UI message

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
ctlai95: Agreed, I can follow up on this part later

zixlin7 (approved):

PR #6935, opened 2025-04-04
feat(amazonq): Remove unsupported message for non-java python languages in /test

jkuruba:
## Problem
In /test feature, test generation in languages other than Java and Python display message that they were not supported. Tests are generated for all languages, so this message needs to be removed.

## Solution
Removed message to the user that /test only supports Python and Java
Screenshot - Removed unsupporte message

---

### DIFF
--- a/packages/core/src/amazonqTest/chat/controller/controller.ts
+++ b/packages/core/src/amazonqTest/chat/controller/controller.ts
@@ -548,16 +548,12 @@ export class TestController {
             */
             if (!['java', 'python'].includes(language) || workspaceFolder === undefined) {
                 let unsupportedMessage: string
-                const unsupportedLanguage = language ? language.charAt(0).toUpperCase() + language.slice(1) : ''
                 if (!workspaceFolder) {
                     // File is outside of workspace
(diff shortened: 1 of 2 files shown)


ashishrp-aws (approved):

laileni-aws (approved):

packages/core/src/amazonqTest/chat/controller/controller.ts
```diff
@@ -548,16 +548,12 @@ export class TestController {
             */
             if (!['java', 'python'].includes(language) || workspaceFolder === undefined) {
                 let unsupportedMessage: string
```
hayemaxi, line 550: ` let unsupportedMessage: string` is not used in this scope, move it the if statement where it is used or remove it

jkuruba: Fixed

hayemaxi (approved):

hayemaxi:
Needs a changelog item since the user can see this change. Otherwise LGTM.

hayemaxi (changes requested):
Needs a changelog item since the user can see this change. Otherwise LGTM.

justinmk3:
I will add a changelog.

PR #6940, opened 2025-04-04
fix(chat): use diff lib to get diffs

ctlai95:
## Problem

Use https://www.npmjs.com/package/diff for diffs

## Solution

- Update diffs for FsWrite
- Remove hardcoded changed lines
- Remove hardcoded diff language
- Use file name instead of file path in card header

image

---

### DIFF
--- a/packages/core/src/codewhispererChat/tools/fsWrite.ts
+++ b/packages/core/src/codewhispererChat/tools/fsWrite.ts
@@ -8,6 +8,9 @@ import { getLogger } from '../../shared/logger/logger'
 import vscode from 'vscode'
 import { fs } from '../../shared/fs/fs'
 import { Writable } from 'stream'
+import { Change, diffLines } from 'diff'
+import { getDiffMarkdown } from '../../shared/utilities/diffUtils'
+import { getLanguageForFilePath } from '../../shared/utilities/textDocumentUtilities'
 
 interface BaseParams {
     path: string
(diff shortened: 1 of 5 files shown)


packages/core/src/codewhispererChat/tools/fsWrite.ts
```diff
@@ -69,114 +72,40 @@ export class FsWrite {
...
-        let contentToAppend = newStr
-        if (needsNewline) {
-            contentToAppend = '\n' + contentToAppend
+    public async getDiffChanges(): Promise<Change[]> {
+        const sanitizedPath = sanitizePath(this.params.path)
+        let newContent
+        let oldContent
+        try {
+            oldContent = await fs.readFileText(sanitizedPath)
+        } catch (err) {
```
zixlin7, line 91: should we log it? what is the expected UX if we cannot read the file, do we still want to proceed with showing diffs?

packages/core/src/codewhispererChat/tools/fsWrite.ts
```diff
@@ -224,6 +153,11 @@ ${await this.showAppendPreview(this.params.path, this.params.newStr)}
     }
 
     private async handleStrReplace(params: StrReplaceParams, sanitizedPath: string): Promise<void> {
+        const newContent = await this.getStrReplaceContent(params, sanitizedPath)
+        await fs.writeFile(sanitizedPath, newContent)
```
zixlin7, line 157: does caller has error handling for these?

packages/core/src/shared/utilities/textDocumentUtilities.ts
```diff
@@ -224,6 +224,27 @@ export async function addEofNewline(editor: vscode.TextEditor) {
...
+ * @returns {Promise<string | undefined>} A promise that resolves to the language identifier of the file,
+ *                                        or undefined if the file cannot be opened or the language cannot be determined.
+ * @example
+ * // Get language for a JavaScript file
+ * const language = await getLanguageForFilePath('/path/to/file.js');
+ * // language will be 'javascript'
+ */
+export async function getLanguageForFilePath(filePath: string): Promise<string | undefined> {
+    try {
+        const document = await vscode.workspace.openTextDocument(filePath)
```
zixlin7, line 241: does this one recognize the languages that's not supported by VS Code natively?

laileni-aws (approved):

packages/core/src/codewhispererChat/tools/fsWrite.ts
laileni-aws: LLM sends the filePath which may or may not exist 
If fileExists we take the oldContent if not we consider this as a new file and keep oldContent as ''.
I think we should not inform end customer about this!

ctlai95: Yeah in the `create` command case, this is expected to fail because the file doesn't exist yet. So I don't think its necessary to show an error

packages/core/src/shared/utilities/textDocumentUtilities.ts
ctlai95: I think right now no, the language will have to be recognized by vscode

zixlin7: did product ack this? or we can switch to use the file extension matching logic that we have

packages/core/src/codewhispererChat/tools/fsWrite.ts
ctlai95: yes any error thrown from a tool will get sent back to the LLM

packages/core/src/shared/utilities/textDocumentUtilities.ts
laileni-aws: This will only recognize languages natively supported by Visual Studio Code, though it covers most programming languages. For unsupported languages, it defaults to `plaintext`.

zixlin7: I understand that, my question is if this is what product/UX expected, because some of them are technically supported by Q although not by VS Code

ctlai95: I think if we use our file extension matching logic it might be too restrictive? For new languages we have to always update the map.

Another reason I was thinking to use VSCode API is that the markdown diff view and the native diff view should have consistent syntax highlighting

zixlin7: this is ok to follow up

PR #6955, opened 2025-04-08
feat(chat): support multi file write with reject and open diff

ctlai95:
## Problem

Support workflows with multiple file writes without confirmation

## Solution
- Set `requiresAcceptance` to `false` for all FsWrite executions
- Store backups of each FsWrite so we can revert the changes if needed

---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
@@ -20,6 +20,7 @@ import {
     UpdateDetailedListMessage,
     CloseDetailedListMessage,
     SelectTabMessage,
+    ChatItemHeader,
 } from '../../../view/connector/connector'
 import { EditorContextCommandType } from '../../../commands/registerCommands'
 import { ChatResponseStream as qdevChatResponseStream } from '@amzn/amazon-q-developer-streaming-client'
@@ -65,6 +66,8 @@ import { noWriteTools, tools } from '../../../constants'
 import { Change } from 'diff'
 import { FsWriteParams } from '../../../tools/fsWrite'
 import { AsyncEventProgressMessage } from '../../../../amazonq/commons/connector/connectorMessages'
+import { localize } from '../../../../shared/utilities/vsCodeUtils'
+import { getDiffLinesFromChanges } from '../../../../shared/utilities/diffUtils'
 
 export type StaticTextResponseType =
     | 'quick-action-help'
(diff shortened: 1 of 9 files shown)


jguoamz (approved):

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -496,49 +499,44 @@ export class Messenger {
...
+                }
             }
-
             if (validation.warning) {
                 message = validation.warning + message
             }
+            fullWidth = true
+            padding = false
+            // eslint-disable-next-line unicorn/no-null
+            codeBlockActions = { 'insert-to-cursor': null, copy: null }
```
Hweinstock, line 534: why is null needed here?

packages/core/src/codewhispererChat/tools/fsWrite.ts
```diff
@@ -73,24 +77,14 @@ export class FsWrite {
     }
 
     public async queueDescription(updates: Writable): Promise<void> {
-        const sanitizedPath = sanitizePath(this.params.path)
-        const changes = await this.getDiffChanges()
-        const language = await getLanguageForFilePath(sanitizedPath)
-
-        const diff = getDiffMarkdown(changes, language)
-        updates.write(diff)
+        // Write an empty string because FsWrite should only show a chat message with header
```
Hweinstock, line 80: Don't fully follow here, why do we need an empty string?

packages/core/src/codewhispererChat/tools/fsWrite.ts
```diff
@@ -108,6 +102,20 @@ export class FsWrite {
...
+        let oldContent
+        let isNew
+        try {
+            oldContent = await fs.readFileText(sanitizedPath)
+            isNew = false
+        } catch (err) {
+            oldContent = ''
+            isNew = true
+        }
+        return { filePath: sanitizedPath, content: oldContent, isNew }
```
Hweinstock, line 116: Is the logic here that for backups that don't exist (i.e. writing to a new file) we return empty content and a flag to say it doesn't exist? Is this the easiest way to handle this case?

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
ctlai95: undefined doesn't work because it defaults to having `insert to cursor` and `copy` buttons

packages/core/src/codewhispererChat/tools/fsWrite.ts
ctlai95: This is just to make sure we create a chat message for this tool so that we can add a header to it

packages/core/src/codewhispererChat/tools/fsWrite.ts
ctlai95: Yeah, we need some way to differentiate if the backup was a new file creation or the file existed but was empty. I think a flag works best here

Hweinstock (approved):

PR #6962, opened 2025-04-08
feat(amazonq): enable export chat feature

avi-alpert:
## Problem
Users don't have a way to export an Amazon Q chat

## Solution
Add an export icon to chat tab bar and, allow users to export conversations from chat history. Users can choose Markdown or HTML format

---


packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -90,7 +90,7 @@ export class Messenger {
                     followUpsHeader: undefined,
                     relatedSuggestions: undefined,
                     triggerID,
-                    messageID: '',
+                    messageID: triggerID,
```
avi-alpert, line 93: sending an empty string for messageId is what caused Q responses to not be included in exported chat transcript

PR #6966, opened 2025-04-08
feat(chat): Call LLM to get response of shell command rejection

yueny2020:
## Problem
- after user click the reject button of running shell command, we call LLM to get response of shell command rejection

## Solution
image

---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -806,6 +806,18 @@ export class ChatController {
             type: 'chat_message',
             context: undefined,
         })
+
+        const session = this.sessionStorage.getSession(message.tabID!)
+        const currentToolUse = session.toolUseWithError?.toolUse
+        if (currentToolUse && currentToolUse.name === ToolType.ExecuteBash) {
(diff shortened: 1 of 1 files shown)


packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -806,6 +806,17 @@ export class ChatController {
             type: 'chat_message',
             context: undefined,
         })
+
+        const toolUseWithError: ToolUseWithError = {
+            toolUse: {
+                name: ToolType.ExecuteBash,
+                input: message.action.text,
+                toolUseId: message.action.formItemValues?.toolUseId,
+            },
```
jguoamz, line 815: let's first fetch it from the session and compare the name to make sure is executeBash, if it is, we just add an error to store it back to the session

yueny2020: didn't get the purpose of getting session from sessionStorage. I mean for the reject, we need to set the error, right? no matter what exists in the sessionStorage.

yueny2020: discussed offline that the session storage always set the toolUseWithError(without error), only need to add the error to toolUseWithError when shell command rejects.

jguoamz (approved):

PR #6975, opened 2025-04-09
fix(amazonq): Grouping read tool messages under contextList

laileni-aws:
## Problem
- We show read tool message for each file read which is occupying the entire chat with these messages which is not super important.

## Solution
- Group all the read tool messaged under ContextList.
image

---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -716,9 +716,18 @@ export class ChatController {
                         try {
                             await ToolUtils.validate(tool)
 
-                            const chatStream = new ChatStream(this.messenger, tabID, triggerID, toolUse, {
-                                requiresAcceptance: false,
-                            })
+                            const chatStream = new ChatStream(
+                                this.messenger,
+                                tabID,
+                                triggerID,
+                                { ...toolUse, toolUseId: `${toolUse.toolUseId}-output` },
+                                session,
+                                undefined,
+                                false,
+                                {
+                                    requiresAcceptance: false,
+                                }
+                            )
                             if (tool.type === ToolType.FsWrite && toolUse.toolUseId) {
                                 const backup = await tool.tool.getBackup()
                                 session.setFsWriteBackup(toolUse.toolUseId, backup)
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -1178,6 +1178,7 @@ export class ChatController {
     private async processPromptMessageAsNewThread(message: PromptMessage) {
         const session = this.sessionStorage.getSession(message.tabID)
         session.clearListOfReadFiles()
+        session.clearListOfReadFolders()
         session.setShowDiffOnFileWrite(false)
         this.editorContextExtractor
             .extractContextForTrigger('ChatMessage')
--- a/packages/core/src/codewhispererChat/controllers/chat/controller.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/controller.ts
@@ -720,7 +720,7 @@ export class ChatController {
                                 this.messenger,
                                 tabID,
                                 triggerID,
-                                { ...toolUse, toolUseId: `${toolUse.toolUseId}-output` },
+                                toolUse,
                                 session,
                                 undefined,
                                 false,
--- a/packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
@@ -3,7 +3,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import * as vscode from 'vscode'
 import { waitUntil } from '../../../../shared/utilities/timeoutUtils'
 import {
     AppToWebViewMessageDispatcher,
@@ -284,23 +283,18 @@ export class Messenger {
                                     session.setShowDiffOnFileWrite(true)
                                     changeList = await tool.tool.getDiffChanges()
                                 }
-
-                                let lineCount = 0
                                 if (tool.type === ToolType.FsRead) {
                                     const input = toolUse.input as unknown as FsReadParams
-                                    if (input?.path) {
-                                        // Calculate the number of lines in input?.path
-                                        const fileUri = vscode.Uri.file(input?.path)
-                                        const fileContent = await vscode.workspace.fs.readFile(fileUri)
-                                        const fileContentString = new TextDecoder().decode(fileContent)
-                                        lineCount = fileContentString.split('\n').length
+                                    // Check if this file path is already in the readFiles list
+                                    const isFileAlreadyRead = session.readFiles.some(
+                                        (file) => file.relativeFilePath === input.path
+                                    )
+                                    if (!isFileAlreadyRead) {
+                                        session.addToReadFiles({
+                                            relativeFilePath: input?.path,
+                                            lineRanges: [{ first: -1, second: -1 }],
+                                        })
                                     }
-                                    const [start, end] = input?.readRange ?? [0, lineCount - 1]
-
-                                    session.addToReadFiles({
-                                        relativeFilePath: input?.path,
-                                        lineRanges: [{ first: start, second: end }],
-                                    })
                                 } else if (tool.type === ToolType.ListDirectory) {
                                     const input = toolUse.input as unknown as ListDirectoryParams
                                     session.setReadFolders({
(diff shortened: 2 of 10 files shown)


packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -716,9 +716,18 @@ export class ChatController {
...
                             await ToolUtils.validate(tool)
 
-                            const chatStream = new ChatStream(this.messenger, tabID, triggerID, toolUse, {
-                                requiresAcceptance: false,
-                            })
+                            const chatStream = new ChatStream(
+                                this.messenger,
+                                tabID,
+                                triggerID,
+                                { ...toolUse, toolUseId: `${toolUse.toolUseId}-output` },
```
jguoamz, line 723: why do we need this?
>${toolUse.toolUseId}-output

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -508,6 +536,26 @@ export class Messenger {
...
+        this.dispatcher.sendChatMessage(
+            new ChatMessage(
+                {
+                    message: '',
+                    messageType: 'answer',
+                    followUps: undefined,
+                    followUpsHeader: undefined,
+                    relatedSuggestions: undefined,
+                    triggerID,
+                    messageID: toolUseId ?? 'toolUse',
```
jguoamz, line 549: this string should always exists?

laileni-aws: You mean `messageID`, Yes we need this to show the read tool view as we are using `onChatAnswerUpdated`

packages/core/src/codewhispererChat/controllers/chat/controller.ts
laileni-aws: Legacy change
Removing this !

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -601,29 +651,59 @@ export class Messenger {
...
+                        relatedSuggestions: undefined,
+                        triggerID,
+                        messageID: messageIdToUpdate ?? toolUse?.toolUseId ?? '',
+                        userIntent: undefined,
+                        codeBlockLanguage: undefined,
+                        contextList: session.readFiles,
+                        canBeVoted: false,
+                        buttons,
+                        fullWidth: false,
+                        padding: true,
```
ctlai95, line 671: padding should be false?

packages/core/src/codewhispererChat/tools/chatStream.ts
```diff
@@ -22,28 +24,49 @@ export class ChatStream extends Writable {
...
         private readonly changeList?: Change[],
         private readonly logger = getLogger('chatStream')
     ) {
         super()
-        this.logger.debug(`ChatStream created for tabID: ${tabID}, triggerID: ${triggerID}`)
-        this.messenger.sendInitalStream(tabID, triggerID)
+        this.logger.debug(
+            `ChatStream created for tabID: ${tabID}, triggerID: ${triggerID}, session: ${session.readFiles}, emitEvent to mynahUI: ${emitEvent}`
+        )
+        if (toolUse?.name === ToolType.FsRead && emitEvent) {
```
ctlai95, line 39: Can we do these checks outside? chatStream should not be aware of any specific tool

packages/core/src/codewhispererChat/controllers/chat/controller.ts
ctlai95: Is this not needed because of the emitEvent flag?

packages/core/src/codewhispererChat/tools/chatStream.ts
```diff
@@ -22,28 +24,49 @@ export class ChatStream extends Writable {
         private readonly tabID: string,
         private readonly triggerID: string,
         private readonly toolUse: ToolUse | undefined,
+        private readonly session: ChatSession,
```
ctlai95, line 27: Why do we need to pass in session?

packages/core/src/codewhispererChat/tools/chatStream.ts
```diff
@@ -22,28 +24,49 @@ export class ChatStream extends Writable {
...
-        this.messenger.sendInitalStream(tabID, triggerID)
+        this.logger.debug(
+            `ChatStream created for tabID: ${tabID}, triggerID: ${triggerID}, session: ${session.readFiles}, emitEvent to mynahUI: ${emitEvent}`
+        )
+        if (toolUse?.name === ToolType.FsRead && emitEvent) {
+            if (!messageIdToUpdate) {
+                // If messageIdToUpdate is undefined, we need to first create an empty message
+                // with messageId so it can be updated later
+                this.messenger.sendInitialToolMessage(tabID, triggerID, toolUse?.toolUseId)
+            }
```
ctlai95, line 44:
```suggestion
        if (toolUse?.name === ToolType.FsRead && emitEvent && !messageIdToUpdate) {
              // If messageIdToUpdate is undefined, we need to first create an empty message
              // with messageId so it can be updated later
              this.messenger.sendInitialToolMessage(tabID, triggerID, toolUse?.toolUseId)
```

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -601,29 +651,59 @@ export class Messenger {
...
-                    canBeVoted: false,
-                    buttons,
-                    fullWidth,
-                    padding,
-                    header,
-                    codeBlockActions,
-                },
-                tabID
+        if (toolUse?.name === ToolType.FsRead) {
+            this.dispatcher.sendToolMessage(
```
ctlai95, line 655: I think it might be cleaner if you move this check inside sendToolMessage

packages/core/src/codewhispererChat/controllers/chat/controller.ts
laileni-aws: Question: Why do we need to modify the toolUseId in first place?

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
laileni-aws: I am using `sendToolMessage` for FsRead and ListDirectory and `sendChatMessage` for FsWrite and ExecuteBash tool output.

packages/core/src/codewhispererChat/tools/chatStream.ts
ctlai95: but the Messenger class can already access session

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
ctlai95: but can we not always use `sendToolMessage` and then inside that function have a condition to use `onChatAnswerUpdated` or `onChatAnswerReceived`

packages/core/src/codewhispererChat/controllers/chat/controller.ts
laileni-aws: I did not get this but I am not facing any issue if I use `toolUse` here and I did not get why we need to modify `{ ...toolUse, toolUseId: `${toolUse.toolUseId}-output` }` 


May be I am missing something here.

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
laileni-aws: Where do we have this conditional logic?
We can not do this in webview ?
I think not in connector.ts too So, messenger will be the best place to do this!

packages/core/src/codewhispererChat/tools/chatStream.ts
ashishrp-aws: the other option makes sense.

ctlai95: My bad, I didn't realize other methods are also passing it in. This is fine then

packages/core/src/codewhispererChat/controllers/chat/messenger/messenger.ts
```diff
@@ -520,48 +584,92 @@ export class Messenger {
...
         toolUse: ToolUse | undefined,
+        session: ChatSession,
+        messageIdToUpdate: string | undefined,
         validation: CommandValidation,
         changeList?: Change[]
     ) {
+        // Handle read tool and list directory messages
+        if (toolUse?.name === ToolType.FsRead || toolUse?.name === ToolType.ListDirectory) {
+            return this.sendReadAndListDirToolMessage(toolUse, session, tabID, triggerID, messageIdToUpdate)
+        }
```
ctlai95, line 644: nit: this is still confusing, I think the condition should not happen in this function and always call processToolMessage

ctlai95 (approved):
Approving to unblock, please follow up in a future PR

PR #6987, opened 2025-04-09
feat(chat): Add validation to fileRead, ListDir and ExecBash tool

bywang56:
## Problem
We should ask for consensus tools accessing files/dirs outside of workspace

## Solution
Add validation step when accessing files/dirs outside of workspace using fsRead or listDirectories tools.

## Tests
Tests tools, verified button shows up and workflow continues to finish after "accept".   

---


packages/core/package.nls.json
```diff
@@ -456,6 +456,10 @@
     "AWS.amazonq.opensettings:": "Open settings",
     "AWS.amazonq.executeBash.run": "Run",
     "AWS.amazonq.executeBash.reject": "Reject",
+    "AWS.amazonq.fsRead.run": "Run",
+    "AWS.amazonq.fsRead.reject": "Reject",
+    "AWS.amazonq.listDirectory.run": "Run",
+    "AWS.amazonq.listDirectory.reject": "Reject",
```
justinmk3, line 462: nls.json is only necessary if these strings are used in package.json

and for common strings, update `packages/core/src/shared/localizedText.ts`, don't repeat them with different ids. Why would "fsRead" and "listDirectory" have different words for "Run"? Avoid over-abstracting, that is even more costly than not abstracting at all.

PR #6995, opened 2025-04-10
fix(chat): fix shell command execution output markdown block issue

yueny2020:
## Problem
- fix shell command execution output markdown block issue
- When the child process of bash command execution writes to both stdout and stderr, we need to main the output is processed in the exact order it was generated by the child process.

## Solution
- Before this change
image

- After this change
image

---


tsmithsz (approved):

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -307,3 +365,17 @@ export class ExecuteBash {
         updates.end()
     }
 }
+
+class AtomicBoolean {
```
justinmk3, line 369: ? javascript is single-thread, what is the purpose of this?

yueny2020: yes, javascript is single-thread, but we have the use case that one process can send bash command execution output to stdout and stderr, which the streaming of stdout and stderr is async.

justinmk3: javascript has a lot of asynchrony, but each async execution context is executed fully until a `async` call, and is not "preemptible". `getAndSet()` is not doing anything useful here.

justinmk3: see https://github.com/aws/aws-toolkit-vscode/pull/6995/files#r2037764657

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -207,8 +215,33 @@ export class ExecuteBash {
             const stdoutBuffer: string[] = []
             const stderrBuffer: string[] = []
 
-            let firstChunk = true
-            let firstStderrChunk = true
+            // Use a queue to maintain chronological order of chunks
+            // This ensures that the output is processed in the exact order it was generated by the child process.
+            const outputQueue: TimestampedChunk[] = []
+            let processingQueue = false
+            const firstChunk = new AtomicBoolean(true)
```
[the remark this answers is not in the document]
yueny2020: I think as the javascript event loop is a single thread, we can use a simple boolean variable(isFirstChunk) in closure to replace the AsyncLock in my use case.

1. The onStdout and onStderr callbacks are never truly executing simultaneously for one bash command execution - they're queued and executed one after another on the event loop.

2. The only shared state is the simple boolean flag (isFirstChunk) that's only set once from true to false. and there is no any other async operations.

Overall, the usage of AsyncLock adds unnecessary overhead for such a simple operation.

I can remove the `atomic` inner class(as it causes confusion). just use var and a function.

```
            // Use a closure boolean value firstChunk and a function to get and set its value atomically
            let isFirstChunk = true
            const getAndSetFirstChunk = (newValue: boolean): boolean => {
                const oldValue = isFirstChunk
                isFirstChunk = newValue
                return oldValue
            }

```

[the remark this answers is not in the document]
justinmk3: > set its value atomically

what is the purpose of mentioning this? there is no atomicity provided by this function.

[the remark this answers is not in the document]
yueny2020: OK, remove it from the comment.

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -207,22 +215,68 @@ export class ExecuteBash {
             const stdoutBuffer: string[] = []
             const stderrBuffer: string[] = []
 
-            let firstChunk = true
-            let firstStderrChunk = true
+            // Use a closure boolean value firstChunk and a function to get and set its value atomically
+            let isFirstChunk = true
+            const getAndSetFirstChunk = (newValue: boolean): boolean => {
```
yueny2020, line 220: the function getAndSetFirstChunk can avoid duplicate code in line 260 and line 271.

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -207,22 +215,68 @@ export class ExecuteBash {
             const stdoutBuffer: string[] = []
             const stderrBuffer: string[] = []
 
-            let firstChunk = true
-            let firstStderrChunk = true
+            // Use a closure boolean value firstChunk and a function to get and set its value atomically
```
justinmk3, line 218: This entire comment is likely to cause confusion. The only really purpose of this function in this codebase is to save a few lines of code. That's fine, but that is different than what this comment says.

justinmk3 (approved):
LGTM after comments are fixed

PR #7000, opened 2025-04-10
fix(chat): Reverting PR:6975

laileni-aws:
## Problem
- Reverting PR: #6975  as this is causing regression in some cases.

## Solution

---

- Treat all work as PUBLIC. Private `feature/x` branches will not be squash-merged at release time.
- Your code changes must meet the guidelines in [CONTRIBUTING.md](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#guidelines).
- License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7006, opened 2025-04-10
feat(chat): Grouping read and list directory messages UX

laileni-aws:
## Problem
This PR:https://github.com/aws/aws-toolkit-vscode/pull/6975 got merged but caused regression. So reverted https://github.com/aws/aws-toolkit-vscode/pull/7000 and fixed this as part of this PR.

---

### DIFF
--- a/packages/core/src/amazonq/webview/ui/main.ts
+++ b/packages/core/src/amazonq/webview/ui/main.ts
@@ -100,41 +100,6 @@ export const createMynahUI = (
         welcomeCount += 1
     }
 
-    /**
-     * Creates a file list header from context list
-     * @param contextList List of file contexts
-     * @param rootFolderTitle Title for the root folder
-     * @returns Header object with file list
-     */
-    const createFileListHeader = (contextList: any[], rootFolderTitle?: string) => {
-        return {
-            fileList: {
-                fileTreeTitle: '',
-                filePaths: contextList.map((file) => file.relativeFilePath),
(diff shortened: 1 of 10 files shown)


packages/core/src/amazonq/webview/ui/main.ts
```diff
@@ -409,32 +447,7 @@ export const createMynahUI = (
...
-                                        )
-                                        .join(', '),
-                                    description: file.relativeFilePath,
-                                    clickable: true,
-                                },
-                            ])
-                        ),
-                    },
-                }
+                item.header = createFileListHeader(item.contextList, item.rootFolderTitle)
```
jguoamz, line 450: do we still need this now that we have 386-388?

laileni-aws: L386-L388 is for  `onChatAnswerUpdated` method and this mentioned L450 is for `onChatAnswerReceived` method. So we need this for Falcon project.

Hweinstock:
What is the regression that this fixes? Didn't see it mentioned in the revert PR so missing some context here.

Hweinstock (approved):
Thanks for adding the explanation, thats super helpful!

It seems to me that the heavy conditional logic based on which tool is being used feels like the real problem here as it makes the expected behavior unclear and hard to follow. 

However, this is a broader issue we can address in migration to Flare.

PR #7014, opened 2025-04-11
Revert "fix(amazonq): "failed to run command""

hayemaxi:
This is preventing users from seeing the sign in page when clicking sign in. This is because `tryExecute` only works for commands created via the `Commands` utilities, but these are vscode generated commands and are never registered.


justinmk3:
instead of reverting, see https://github.com/aws/aws-toolkit-vscode/pull/7015

PR #7024, opened 2025-04-11
Fix broken keyPath to ssh-keygen

marcleblanc2:
## Problem

When keyPath contains spaces, ex. `/Users/username/Library/Application Support/Code/User/globalStorage/amazonwebservices.aws-toolkit-vscode/aws-ec2-key` the ssh-keygen command fails with "Too many arguments", and doesn't get caught by the "Unknown key type" failure check. 

This results in log output like

[error] aws.ec2.openRemoteConnection: Error: Unable to connect to target instance <id> on region <region>. Testing SSM connection to instance failed: Warning: Permanently added 'aws-ec2-<id>' (ED25519) to the list of known hosts.
no such identity: /Users/username/Library/Application Support/Code/User/globalStorage/amazonwebservices.aws-toolkit-vscode/aws-ec2-key: No such file or directory
ec2-user@aws-ec2-<id>: Permission denied (publickey,gssapi-keyex,gssapi-with-mic). [EC2SSMTestConnect]

## Workaround

Manually run the command with a double quoted path to create the key in this path.

## Solution

Double quote the path in the command arg

---

### DIFF
--- a/packages/core/src/awsService/ec2/sshKeyPair.ts
+++ b/packages/core/src/awsService/ec2/sshKeyPair.ts
@@ -81,7 +81,7 @@ export class SshKeyPair {
         }
         return await tryRun(
             'ssh-keygen',
-            ['-t', keyType, '-N', '', '-q', '-f', keyPath],
+            ['-t', keyType, '-N', '', '-q', '-f', concat('"', keyPath, '"')],
             'yes',
             /^(?!.*Unknown key type).*/i,
             {


justinmk3:
Thanks for looking into this. Did you confirm that this change fixes the issue? I am surprised if it does, because the list form of arguments should already do that. By adding quotes as you've done here, it will send literal `"` chars to the program. 

If that fixes the issue, that might indicate a bug in ssh-keygen. It's certainly worth a workaround to avoid such a bug, but we need to make sure it doesn't regress anything.

packages/core/src/awsService/ec2/sshKeyPair.ts
```diff
@@ -81,7 +81,7 @@ export class SshKeyPair {
         }
         return await tryRun(
             'ssh-keygen',
-            ['-t', keyType, '-N', '', '-q', '-f', keyPath],
+            ['-t', keyType, '-N', '', '-q', '-f', concat('"', keyPath, '"')],
```
justinmk3, line 84: npm error src/awsService/ec2/sshKeyPair.ts(84,51): error TS2304: Cannot find name 'concat'.

PR #7041, opened 2025-04-14
fix(chat): Add programmer card on top of the page. 

Randall-Jiang:
## Problem
We want to show the programmer card on top of the page. And change the welcome message

## Solution
image

---


packages/core/src/amazonq/webview/ui/tabs/generator.ts
```diff
@@ -67,6 +79,15 @@ export class TabDataGenerator {
                       messageId: 'regionProfile',
                   }
 
+        const welcomeMessage = `Hi! I'm Amazon Q.
```
zixlin7, line 82: you probably can update this one instead: https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/amazonq/webview/ui/tabs/constants.ts#L9

Randall-Jiang: The reason I add the seperate welcomeMessage here is I don't want mess up the current user experience. Since I only got the UI for agentic chat. I will update constants.ts if we finalize that we are using the same welcome message globally

zixlin7 (approved):

Issue #6972, opened 2025-04-09
Amazon Q LSP: Failure to initialize on NixOS - EPIPE error during LSP startup


cShirley14:
## Problem
As a NixOS user attempting to use Amazon Q, the Language Server fails to start. NixOS being an immutable operating system, this appears to be a configuration-related issue with the Language Server initialization.

## Steps to reproduce the issue
1. Install Amazon Q extension in VS Code on NixOS
2. Open VS Code
3. Attempt to initialize Amazon Q
4. Language Server fails to start

## Expected behavior
The Language Server should initialize successfully and Amazon Q should be available for use within VS Code.

## System details
- OS: NixOS 25.05 (Warbler)
- Visual Studio Code version (VSCodium):  1.98.2  Release: 25072
- AWS Toolkit version: N/A
- Amazon Q version: 1.54.0

Amazon Q About:
- OS: Linux x64 6.12.19
- Visual Studio Code extension host:  1.98.2
- Amazon Q:  1.54.0
- node: 20.18.2
- electron: 34.2.0

## Relevant logs
```log
[info] using amazonqWorkspaceLsp service configuration: default
[info] lsp: Failed to download latest "AmazonQ-Workspace" manifest. Falling back to local manifest.
[info] lsp: Finished setting up LSP server
[info] [Error] Starting client failed
[info] Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:161:15)
    at writeGeneric (node:internal/stream_base_commons:152:3)
    at Socket._writeGeneric (node:net:958:11)
    at Socket._write (node:net:970:8)
    at writeOrBuffer (node:internal/streams/writable:572:12)
    at _write (node:internal/streams/writable:501:10)
    at Writable.write (node:internal/streams/writable:510:10)
    at /nix/store/<hash>-vscode-extension-amazonwebservices-amazon-q-vscode-1.54.0/share/vscode/extensions/amazonwebservices.amazon-q-vscode/dist/src/extensionNode.js:7259:3222
    at new Promise (<anonymous>)
    at x.send (/nix/store/<hash>-vscode-extension-amazonwebservices-amazon-q-vscode-1.54.0/share/vscode/extensions/amazonwebservices.amazon-q-vscode/dist/src/extensionNode.js:7259:3201)
[...]

PR #7043, opened 2025-04-14
fix(lsp): verify that nodejs runs before starting LSP client

justinmk3:
## Problem
When lsp fetch/install/start fails it does not mention the download path, which could help with troubleshooting. #6972

    [info] using amazonqWorkspaceLsp service configuration: default
    [info] lsp: Failed to download latest "AmazonQ-Workspace" manifest. Falling back to local manifest.
    [info] lsp: Finished setting up LSP server
    [info] [Error] Starting client failed
    [info] Error: write EPIPE

## Solution
- Validate that `node` can actually run, before passing it to `LspClient`.
- Add more logging.

Also captured by telemetry:

```
2025-04-16 08:24:51.738 [debug] telemetry: languageServer_setup {
  Metadata: {
    missingFields: 'id',
    metricId: '8da91a4b-ee00-4115-9b9e-796b5357402c',
    traceId: '8569c16e-d319-486e-a6f3-d4ee91698468',
    languageServerSetupStage: 'all',
    duration: '1417',
    result: 'Failed',
    reason: 'Error',
    reasonDesc: 'amazonqLsp: failed to run basic "node -e" test (exitcode=-2): [/Users/x/x/x/aws/x/x/x/x/x -e console.log("ok " + process.version)]',
    awsAccount: 'not-set',
    awsRegion: 'us-east-1'
  },
  Value: 1,
  Unit: 'Milliseconds',
  Passive: true
}
```

---

### DIFF
--- a/packages/core/src/amazonq/lsp/lspClient.ts
+++ b/packages/core/src/amazonq/lsp/lspClient.ts
@@ -47,7 +47,7 @@ import { ChildProcess } from '../../shared/utilities/processUtils'
 const localize = nls.loadMessageBundle()
 
 const key = crypto.randomBytes(32)
-const logger = getLogger('amazonqLsp')
+const logger = getLogger('amazonqLsp.lspClient')
 
 /**
  * LspClient manages the API call between VS Code extension and LSP server
@@ -83,7 +83,7 @@ export class LspClient {
             const resp = await this.client?.sendRequest(BuildIndexRequestType, encryptedRequest)
             return resp
         } catch (e) {
-            logger.error(`LspClient: buildIndex error: ${e}`)
+            logger.error(`buildIndex error: ${e}`)
             return undefined
         }
     }
@@ -98,7 +98,7 @@ export class LspClient {
             const resp = await this.client?.sendRequest(QueryVectorIndexRequestType, encryptedRequest)
             return resp
         } catch (e) {
-            logger.error(`LspClient: queryVectorIndex error: ${e}`)
+            logger.error(`queryVectorIndex error: ${e}`)
             return []
         }
     }
@@ -114,7 +114,7 @@ export class LspClient {
             const resp: any = await this.client?.sendRequest(QueryInlineProjectContextRequestType, encrypted)
             return resp
         } catch (e) {
-            logger.error(`LspClient: queryInlineProjectContext error: ${e}`)
+            logger.error(`queryInlineProjectContext error: ${e}`)
             throw e
         }
     }
@@ -135,7 +135,7 @@ export class LspClient {
             const resp = await this.client?.sendRequest(UpdateIndexV2RequestType, encryptedRequest)
             return resp
         } catch (e) {
-            logger.error(`LspClient: updateIndex error: ${e}`)
+            logger.error(`updateIndex error: ${e}`)
             return undefined
         }
     }
@@ -147,7 +147,7 @@ export class LspClient {
             const resp: any = await this.client?.sendRequest(QueryRepomapIndexRequestType, await this.encrypt(request))
             return resp
         } catch (e) {
-            logger.error(`LspClient: QueryRepomapIndex error: ${e}`)
+            logger.error(`QueryRepomapIndex error: ${e}`)
             throw e
         }
     }
@@ -160,7 +160,7 @@ export class LspClient {
             )
             return resp
         } catch (e) {
-            logger.error(`LspClient: queryInlineProjectContext error: ${e}`)
+            logger.error(`queryInlineProjectContext error: ${e}`)
             throw e
         }
     }
@@ -177,7 +177,7 @@ export class LspClient {
             )
             return resp
         } catch (e) {
-            logger.error(`LspClient: getContextCommandItems error: ${e}`)
+            logger.error(`getContextCommandItems error: ${e}`)
             throw e
         }
     }
@@ -193,7 +193,7 @@ export class LspClient {
             )
             return resp || []
         } catch (e) {
-            logger.error(`LspClient: getContextCommandPrompt error: ${e}`)
+            logger.error(`getContextCommandPrompt error: ${e}`)
             throw e
         }
     }
@@ -207,7 +207,7 @@ export class LspClient {
             )
             return resp
         } catch (e) {
-            logger.error(`LspClient: getIndexSequenceNumber error: ${e}`)
+            logger.error(`getIndexSequenceNumber error: ${e}`)
             throw e
         }
     }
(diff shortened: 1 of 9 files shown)


packages/core/src/shared/lsp/lspResolver.ts
```diff
@@ -47,25 +47,40 @@ export class LanguageServerResolver {
...
+             * Example:
+             * ```
+             * LspResult {
+             *   assetDirectory = "<cachedir>/aws/toolkits/language-servers/AmazonQ/3.3.0"
+             *   location = 'cache'
+             *   version = '3.3.0'
+             * }
+             * ```
+             */
+            const resolved = await tryStageResolvers('getServer', serverResolvers, getServerVersion)
```
justinmk3, line 76: Example values convey a lot of information at a very low cost. Especially important when there is so much indirection and polymorphism.

packages/core/src/shared/lsp/lspResolver.ts
```diff
@@ -47,25 +47,40 @@ export class LanguageServerResolver {
 
             const serverResolvers: StageResolver<LspResult>[] = [
                 {
+                    // 1: Use the current local ("cached") LSP server bundle, if any.
                     resolve: async () => await this.getLocalServer(cacheDirectory, latestVersion, targetContents),
                     telemetryMetadata: { id: this.lsName, languageServerLocation: 'cache' },
                 },
                 {
+                    // 2: Download the latest LSP server bundle.
                     resolve: async () => await this.fetchRemoteServer(cacheDirectory, latestVersion, targetContents),
```
justinmk3, line 56: It seems like these 3 functions (`getFallbackServer`, `fetchRemoteServer`, `getLocalServer`) are all variations on the same thing, but their names don't make that obvious.

packages/core/src/shared/lsp/lspResolver.ts
```diff
@@ -417,13 +434,17 @@ export class LanguageServerResolver {
...
     }
 
-    // lazy calls to `getApplicationSupportFolder()` to avoid failure on windows.
-    public static get defaultDir() {
+    /**
+     * Gets platform-specific "cache" dir ("$LOCALAPPDATA/aws/…" or "~/.cache/aws/…").
+     *
+     * Lazy-calls `getCacheDir()` to avoid failure on Windows.
+     */
+    public static defaultDir() {
```
justinmk3, line 442: "properties" make it less obvious that work is happening. don't see a reason why this shouldn't be a plain old function.

packages/core/src/amazonq/lsp/lspClient.ts
```diff
@@ -259,6 +266,7 @@ export async function activate(extensionContext: ExtensionContext, resourcePaths
         debug: { module: serverModule, transport: TransportKind.ipc, options: debugOptions },
     }
 
+    // TODO(jmkeyes): this overwrites the above...?
     serverOptions = createServerOptions({
```
justinmk3, line 270: The `let serverOptions = ...` immediately above this line is not used. Is this a bug?

packages/core/src/amazonq/lsp/lspClient.ts
```diff
@@ -222,20 +225,56 @@ export class LspClient {
...
- * Activates the language server, this will start LSP server running over IPC protocol.
- * It will create a output channel named Amazon Q Language Server.
- * This function assumes the LSP server has already been downloaded.
+ * Activates the language server (assumes the LSP server has already been downloaded):
+ * 1. start LSP server running over IPC protocol.
+ * 2. create a output channel named Amazon Q Language Server.
  */
 export async function activate(extensionContext: ExtensionContext, resourcePaths: ResourcePaths) {
-    LspClient.instance
+    LspClient.instance // Tickle the singleton... :/
```
Hweinstock, line 271: Do you know why this is needed? I imagine its to ensure the singleton is initialized, but unclear why something that needs it initialized can't initialize the singleton itself

packages/core/src/shared/lsp/lspResolver.ts
Hweinstock: Do you see a better naming convention here? My intention here was to make it clear where the server was coming from by following a  `[verb][source]Server` pattern where `verb` follows from source.

Hweinstock (approved):

packages/core/src/amazonq/lsp/lspClient.ts
justinmk3: I guess for early init. idk, that's why it's sus

Hweinstock:
Looks like there is a small lint issue: 
```
/home/runner/work/aws-toolkit-vscode/aws-toolkit-vscode/packages/core/src/amazonq/lsp/lspClient.ts
  258:77  error  Avoid referencing unbound methods which may cause unintentional scoping of `this`.
If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead  @typescript-eslint/unbound-method

✖ 1 problem (1 error, 0 warnings)
```

packages/core/src/shared/lsp/lspResolver.ts
Hweinstock: Thats a good point. I was trying to have the verb be consistent with the operation being done, but perhaps this can be abstracted from the caller. 

This `Stage` abstraction was mostly for telemetry and to be clearly consistent with the LSP Spec.

packages/core/src/amazonq/lsp/lspClient.ts
Hweinstock: +1, this could point to a race condition?

jpinkney-aws: when I originally looked at this code I totally thought it was a bug 😆

packages/core/src/amazonq/lsp/lspClient.ts
jpinkney-aws: yeah it overwrites above, there's no reason for the first serverOptions to exist

jpinkney-aws (approved):

packages/core/src/amazonq/lsp/lspClient.ts
```diff
@@ -174,7 +177,7 @@ export class LspClient {
             )
             return resp
         } catch (e) {
-            getLogger().error(`LspClient: getContextCommandItems error: ${e}`)
+            logger.error(`LspClient: getContextCommandItems error: ${e}`)
```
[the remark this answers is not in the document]
justinmk3: I think if granularity is wanted, then topics can have "dot" subnamespaces, e.g. `getLogger('amazonLsp.lspClient')`.

PR #7044, opened 2025-04-15
fix(chat): Updating spinner text to thinking

ashishrp-aws:
## Problem
Currently the spinner text on waiting for LLM responses displays `Generating your Answer` which is not reflective of new features.

## Solution
- Updated it to `Thinking`

---

### DIFF
--- a/packages/core/src/amazonq/webview/ui/texts/constants.ts
+++ b/packages/core/src/amazonq/webview/ui/texts/constants.ts
@@ -25,7 +25,7 @@ export const uiComponentsTexts = {
     copyToClipboard: 'Copied to clipboard',
     noMoreTabsTooltip: 'You can only open ten conversation tabs at a time.',
     codeSuggestionWithReferenceTitle: 'Some suggestions contain code with references.',
-    spinnerText: 'Generating your answer...',
+    spinnerText: 'Thinking...',
     changeAccepted: 'Change accepted',
     changeRejected: 'Change rejected',
     acceptChange: 'Accept change',


mr-lee (approved):

packages/core/src/amazonq/webview/ui/texts/constants.ts
```diff
@@ -25,7 +25,7 @@ export const uiComponentsTexts = {
     copyToClipboard: 'Copied to clipboard',
     noMoreTabsTooltip: 'You can only open ten conversation tabs at a time.',
     codeSuggestionWithReferenceTitle: 'Some suggestions contain code with references.',
-    spinnerText: 'Generating your answer...',
```
jguoamz, line 28: this will impact /dev, /test, etc? Can we make a new one for cwc only?

packages/core/src/amazonq/webview/ui/texts/constants.ts
```diff
@@ -25,7 +25,7 @@ export const uiComponentsTexts = {
...
+    spinnerText: 'Thinking...',
```
laileni-aws, line 28: I guess this will be impacting entire Amazon Q including `/dev` etc
Are we ok for this ?
If Yes we are good!

packages/core/src/amazonq/webview/ui/texts/constants.ts
ashishrp-aws: yes it will...UIComponentText based on tab config (which available i would have set it to only for cwc and pair programmer mode)

packages/core/src/amazonq/webview/ui/texts/constants.ts
ashishrp-aws: https://github.com/aws/aws-toolkit-vscode/pull/7044#discussion_r2043233547

PR #7049, opened 2025-04-15
fix(amazonq): validate customization on profile change

Will-ShaoHua:
## Problem
- Before
customization is bound to a specific idc instance

- After
customization is bound to a specific Q profile and an idc instance can have multi profiles

therefore each Q profile will have access to different sets of customization

## Solution
We need to validate if the selected customization is accessible from the user's selected profile, otherwise will get
```
*An error occurred while processing your request.*
This error is reported to the team automatically. We will attempt to fix it as soon as possible.
Details: The provided profile ARN and customization ARN is mismatched. (Service: CodeWhispererRuntime, Status Code: 403, 
```

---

### DIFF
--- a/packages/core/src/codewhisperer/activation.ts
+++ b/packages/core/src/codewhisperer/activation.ts
@@ -344,9 +344,9 @@ export async function activate(context: ExtContext): Promise<void> {
         ),
         vscode.commands.registerCommand('aws.amazonq.openEditorAtRange', openEditorAtRange),
         auth.regionProfileManager.onDidChangeRegionProfile(() => {
-            // validate user's still has access to the selected customization
+            // Validate user still has access to the selected customization.
             const selectedCustomization = getSelectedCustomization()
-            // no need to validate base customization which has empty arn
+            // No need to validate base customization which has empty arn
             if (selectedCustomization.arn.length > 0) {
                 getAvailableCustomizationsList()
                     .then(async (customizations) => {
@@ -357,7 +357,8 @@ export async function activate(context: ExtContext): Promise<void> {
                     })
                     .catch((e) => {
                         getLogger().error(
-                            `encounter error while validating selected customization on profile change: ${(e as Error).message}`
+                            `encounter error while validating selected customization on profile change: %s`,
+                            (e as Error).message
                         )
                     })
             }
--- a/packages/core/src/codewhisperer/activation.ts
+++ b/packages/core/src/codewhisperer/activation.ts
@@ -349,10 +349,10 @@ export async function activate(context: ExtContext): Promise<void> {
             // no need to validate base customization which has empty arn
             if (selectedCustomization.arn.length > 0) {
                 getAvailableCustomizationsList()
-                    .then((customizations) => {
+                    .then(async (customizations) => {
                         const r = customizations.find((it) => it.arn === selectedCustomization.arn)
                         if (!r) {
-                            void switchToBaseCustomizationAndNotify().then()
+                            await switchToBaseCustomizationAndNotify()
                         }
                     })
                     .catch((e) => {
--- a/packages/core/src/codewhisperer/activation.ts
+++ b/packages/core/src/codewhisperer/activation.ts
@@ -346,7 +346,7 @@ export async function activate(context: ExtContext): Promise<void> {
         auth.regionProfileManager.onDidChangeRegionProfile(() => {
             // Validate user still has access to the selected customization.
             const selectedCustomization = getSelectedCustomization()
-            // No need to validate base customization which has empty arn
+            // No need to validate base customization which has empty arn.
             if (selectedCustomization.arn.length > 0) {
                 getAvailableCustomizationsList()
                     .then(async (customizations) => {
--- a/packages/core/src/codewhisperer/activation.ts
+++ b/packages/core/src/codewhisperer/activation.ts
@@ -72,7 +72,12 @@ import { AuthUtil } from './util/authUtil'
 import { ImportAdderProvider } from './service/importAdderProvider'
 import { TelemetryHelper } from './util/telemetryHelper'
 import { openUrl } from '../shared/utilities/vsCodeUtils'
-import { notifyNewCustomizations } from './util/customizationUtil'
+import {
+    getAvailableCustomizationsList,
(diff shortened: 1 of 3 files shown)


Will-ShaoHua:
ew
```
 ERROR in Circular dependency detected:
  ../core/dist/src/codewhisperer/client/codewhisperer.js -> ../core/dist/src/codewhisperer/util/authUtil.js -> ../core/dist/src/codewhisperer/region/regionProfileManager.js -> ../core/dist/src/codewhisperer/util/customizationUtil.js -> ../core/dist/src/codewhisperer/client/codewhisperer.js

  ERROR in Circular dependency detected:
  ../core/dist/src/codewhisperer/region/regionProfileManager.js -> ../core/dist/src/codewhisperer/util/customizationUtil.js -> ../core/dist/src/codewhisperer/util/authUtil.js -> ../core/dist/src/codewhisperer/region/regionProfileManager.js

  ERROR in Circular dependency detected:
  ../core/dist/src/codewhisperer/util/authUtil.js -> ../core/dist/src/codewhisperer/region/regionProfileManager.js -> ../core/dist/src/codewhisperer/util/customizationUtil.js -> ../core/dist/src/codewhisperer/util/authUtil.js

  ERROR in Circular dependency detected:
  ../core/dist/src/codewhisperer/util/customizationUtil.js -> ../core/dist/src/codewhisperer/util/authUtil.js -> ../core/dist/src/codewhisperer/region/regionProfi
```

evanliu048 (approved):

packages/core/src/codewhisperer/activation.ts
```diff
@@ -337,7 +342,26 @@ export async function activate(context: ExtContext): Promise<void> {
...
+        auth.regionProfileManager.onDidChangeRegionProfile(() => {
+            // validate user's still has access to the selected customization
+            const selectedCustomization = getSelectedCustomization()
+            // no need to validate base customization which has empty arn
+            if (selectedCustomization.arn.length > 0) {
+                getAvailableCustomizationsList()
+                    .then((customizations) => {
+                        const r = customizations.find((it) => it.arn === selectedCustomization.arn)
+                        if (!r) {
+                            void switchToBaseCustomizationAndNotify().then()
```
justinmk3, line 355: `void` silences any errors, so the `.catch()` would not be called. and the `.then()` isn't needed in any case?

if you want to silence errors, the `.catch()` below is enough.

packages/core/src/codewhisperer/activation.ts
```diff
@@ -337,7 +342,26 @@ export async function activate(context: ExtContext): Promise<void> {
...
+                        }
+                    })
+                    .catch((e) => {
+                        getLogger().error(
+                            `encounter error while validating selected customization on profile change: ${(e as Error).message}`
```
justinmk3, line 360:
```suggestion
                        getLogger().error(
                            `encounter error while validating selected customization on profile change: %s`, (e as Error).message
```

packages/core/src/codewhisperer/activation.ts
Will-ShaoHua: > he void is the least preferred way to do that, because it will let the unresolved promise rejection go into the void and be handled at a random point by nodejs.

isee, didn't know this before thanks for pointing out Justin

packages/core/src/codewhisperer/activation.ts
```diff
@@ -337,7 +342,26 @@ export async function activate(context: ExtContext): Promise<void> {
             [...CodeWhispererConstants.securityScanLanguageIds],
             SecurityIssueCodeActionProvider.instance
         ),
-        vscode.commands.registerCommand('aws.amazonq.openEditorAtRange', openEditorAtRange)
+        vscode.commands.registerCommand('aws.amazonq.openEditorAtRange', openEditorAtRange),
+        auth.regionProfileManager.onDidChangeRegionProfile(() => {
+            // validate user's still has access to the selected customization
```
justinmk3, line 347:
```suggestion
            // Validate user still has access to the selected customization.
```

packages/core/src/codewhisperer/activation.ts
Will-ShaoHua: https://github.com/aws/aws-toolkit-vscode/pull/7049/commits/35a5045629de765037c3751c2c386f885764fb54

packages/core/src/codewhisperer/activation.ts
Will-ShaoHua: so is thiss the correct way to do async call in the `.then()` callback 
https://github.com/aws/aws-toolkit-vscode/pull/7049/commits/35a5045629de765037c3751c2c386f885764fb54

packages/core/src/codewhisperer/activation.ts
Will-ShaoHua: could you share what's the reason we prefer one over the other here

packages/core/src/codewhisperer/activation.ts
```diff
@@ -337,7 +342,27 @@ export async function activate(context: ExtContext): Promise<void> {
...
+        auth.regionProfileManager.onDidChangeRegionProfile(() => {
+            // Validate user still has access to the selected customization.
+            const selectedCustomization = getSelectedCustomization()
+            // No need to validate base customization which has empty arn.
+            if (selectedCustomization.arn.length > 0) {
+                getAvailableCustomizationsList()
+                    .then(async (customizations) => {
+                        const r = customizations.find((it) => it.arn === selectedCustomization.arn)
+                        if (!r) {
+                            await switchToBaseCustomizationAndNotify()
```
justinmk3, line 355: `switchToBaseCustomizationAndNotify()` is not slow, so it's ok for it to be await'd, right?

justinmk3 (approved):

packages/core/src/codewhisperer/activation.ts
justinmk3: see https://github.com/aws/aws-toolkit-vscode/pull/7049#discussion_r2045701021

packages/core/src/codewhisperer/activation.ts
Will-ShaoHua: gotcha, thanks for explanation :)

PR #7060, opened 2025-04-16
feat(amazonq): show all customizations across different profiles

Will-ShaoHua:
## Solution

---

### DIFF
--- a/packages/core/src/codewhisperer/region/regionProfileManager.ts
+++ b/packages/core/src/codewhisperer/region/regionProfileManager.ts
@@ -47,7 +47,7 @@ const endpoints = createConstantMap({
  * 'update' -> plugin auto select the profile on users' behalf as there is only 1 profile
  * 'reload' -> on plugin restart, plugin will try to reload previous selected profile
  */
-export type ProfileSwitchIntent = 'user' | 'auth' | 'update' | 'reload'
+export type ProfileSwitchIntent = 'user' | 'auth' | 'update' | 'reload' | 'customization'
 
 export class RegionProfileManager {
     private static logger = getLogger()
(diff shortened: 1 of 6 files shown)


Will-ShaoHua:
will need a changelog

Will-ShaoHua:
```
  3300 passing (4m)
  14 pending
  1 failing
  1) startSecurityScan
       Should not cancel a project scan if a file scan has started:

      AssertionError [ERR_ASSERTION]: telemetry metric 1 (of 2) not found (in the expected order): "codewhisperer_securityScan" 
+ actual - expected

  {
+   codewhispererCodeScanScope: 'PROJECT',
-   codewhispererCodeScanScope: 'FILE_AUTO',
    result: 'Succeeded'
  }
      + expected - actual

       {
      -  "codewhispererCodeScanScope": "PROJECT"
      +  "codewhispererCodeScanScope": "FILE_AUTO"
         "result": "Succeeded"
       }
      
```

packages/core/src/codewhisperer/region/regionProfileManager.ts
```diff
@@ -343,7 +351,19 @@ export class RegionProfileManager {
         }
     }
 
-    async createQClient(region: string, endpoint: string, conn: SsoConnection): Promise<CodeWhispererUserClient> {
+    async createQClient(profile: RegionProfile): Promise<CodeWhispererUserClient> {
```
jpinkney-aws, line 354: whats the difference between createQClient vs _createQClient and how would I know to use one vs the other?

Will-ShaoHua: actually, just let me make `_createQClient` private, which is more ideal. I kept this public because of test case i wrote previously https://github.com/aws/aws-toolkit-vscode/pull/7060/files

Will-ShaoHua: but totally agree it's very confusing for callers to know which one in favor of the other

jpinkney-aws (approved):

packages/core/src/codewhisperer/region/regionProfileManager.ts
Will-ShaoHua: actually @jpinkney-aws i want to stick with it is for now as it might involve more changes and i don't want to delay the deployment of this change.  But let me rename it and add docstrings for ppl to know the difference

andrewyuq (approved):

tomcat323 (approved):

leigaol (approved):

PR #7062, opened 2025-04-16
feat(amazonq): Use new auth class using flare identity server for Amazon Q

opieter-aws:
## Solution

---

### DIFF
--- a/packages/amazonq/src/app/chat/activation.ts
+++ b/packages/amazonq/src/app/chat/activation.ts
@@ -27,7 +27,7 @@ export async function activate(context: ExtensionContext) {
 
     const setupLsp = funcUtil.debounce(async () => {
         void amazonq.LspController.instance.trySetupLsp(context, {
-            startUrl: AuthUtil.instance.startUrl,
+            startUrl: AuthUtil.instance.connection?.startUrl,
             maxIndexSize: CodeWhispererSettings.instance.getMaxIndexSize(),
             isVectorIndexEnabled: CodeWhispererSettings.instance.isLocalIndexEnabled(),
         })
@@ -44,8 +44,6 @@ export async function activate(context: ExtensionContext) {
         amazonq.walkthroughSecurityScanExample.register(),
         amazonq.openAmazonQWalkthrough.register(),
         amazonq.listCodeWhispererCommandsWalkthrough.register(),
-        amazonq.focusAmazonQPanel.register(),
-        amazonq.focusAmazonQPanelKeybinding.register(),
         amazonq.tryChatCodeLensCommand.register(),
         vscode.workspace.onDidChangeConfiguration(async (configurationChangeEvent) => {
             if (configurationChangeEvent.affectsConfiguration('amazonQ.workspaceIndex')) {
--- a/packages/amazonq/src/extension.ts
+++ b/packages/amazonq/src/extension.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { AuthUtils, CredentialsStore, LoginManager, initializeAuth } from 'aws-core-vscode/auth'
+import { CredentialsStore, LoginManager, authUtils, initializeAuth } from 'aws-core-vscode/auth'
 import { activate as activateCodeWhisperer, shutdown as shutdownCodeWhisperer } from 'aws-core-vscode/codewhisperer'
 import { makeEndpointsProvider, registerGenericCommands } from 'aws-core-vscode'
 import { CommonAuthWebview } from 'aws-core-vscode/login'
@@ -33,6 +33,7 @@ import {
     maybeShowMinVscodeWarning,
     Experiments,
     isSageMaker,
+    Commands,
 } from 'aws-core-vscode/shared'
 import { ExtStartUpSources } from 'aws-core-vscode/telemetry'
 import { VSCODE_EXTENSION_ID } from 'aws-core-vscode/utils'
@@ -114,14 +115,14 @@ export async function activateAmazonQCommon(context: vscode.ExtensionContext, is
 
     await initializeAuth(globals.loginManager)
 
+    await activateAmazonqLsp(context)
+
     const extContext = {
         extensionContext: context,
     }
+
     // This contains every lsp agnostic things (auth, security scan, code scan)
     await activateCodeWhisperer(extContext as ExtContext)
-    if (Experiments.instance.get('amazonqLSP', false)) {
-        await activateAmazonqLsp(context)
-    }
 
     if (!Experiments.instance.get('amazonqLSPInline', false)) {
         await activateInlineCompletion()
@@ -129,6 +130,10 @@ export async function activateAmazonQCommon(context: vscode.ExtensionContext, is
 
     // Generic extension commands
     registerGenericCommands(context, amazonQContextPrefix)
+
+    // Create status bar and reference log UI elements
+    void Commands.tryExecute('aws.amazonq.refreshStatusBar')
+    void Commands.tryExecute('aws.amazonq.updateReferenceLog')
 
     // Amazon Q specific commands
     registerCommands(context)
@@ -156,7 +161,7 @@ export async function activateAmazonQCommon(context: vscode.ExtensionContext, is
     // reload webviews
     await vscode.commands.executeCommand('workbench.action.webview.reloadWebviewAction')
 
-    if (AuthUtils.ExtensionUse.instance.isFirstUse()) {
+    if (authUtils.ExtensionUse.instance.isFirstUse()) {
         // Give time for the extension to finish initializing.
         globals.clock.setTimeout(async () => {
             CommonAuthWebview.authSource = ExtStartUpSources.firstStartUp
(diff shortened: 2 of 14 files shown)


packages/amazonq/src/extension.ts
```diff
@@ -114,14 +115,14 @@ export async function activateAmazonQCommon(context: vscode.ExtensionContext, is
 
     await initializeAuth(globals.loginManager)
 
+    await activateAmazonqLsp(context)
```
justinmk3, line 118: nit: put this just below the `activateCodeWhisperer` call below (so they are grouped)? or move that one up here?

packages/amazonq/src/extension.ts
```diff
@@ -130,6 +131,10 @@ export async function activateAmazonQCommon(context: vscode.ExtensionContext, is
     // Generic extension commands
     registerGenericCommands(context, amazonQContextPrefix)
 
+    // Create status bar and reference log UI elements
+    void Commands.tryExecute('aws.amazonq.refreshStatusBar')
+    void Commands.tryExecute('aws.amazonq.updateReferenceLog')
```
justinmk3, line 136: I assume it's acceptable that this might be racey?

packages/amazonq/src/lsp/activation.ts
```diff
@@ -4,16 +4,20 @@
...
+import { auth2 } from 'aws-core-vscode/auth'
 
-export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
+export async function activate(ctx: vscode.ExtensionContext) {
     try {
-        await lspSetupStage('all', async () => {
+        const client = await lspSetupStage('all', async () => {
             const installResult = await new AmazonQLspInstaller().resolve()
-            await lspSetupStage('launch', async () => await startLanguageServer(ctx, installResult.resourcePaths))
+            return await lspSetupStage('launch', () => startLanguageServer(ctx, installResult.resourcePaths))
```
justinmk3, line 17: this is only a cosmetic change, right? `startLanguageServer` is still `async` and will be await'd by the `await` at start of this line? 

if this is only cosmetic, it will help to avoid this change. because other branches change this file, so minimizing non-critical changes will help when resolving merge conflicts.

packages/core/src/auth/index.ts
```diff
@@ -22,4 +22,6 @@ export {
 export { Auth } from './auth'
 export { CredentialsStore } from './credentials/store'
 export { LoginManager } from './deprecated/loginManager'
-export * as AuthUtils from './utils'
+export { builderIdStartUrl, builderIdRegion } from './sso/constants'
```
justinmk3, line 25:
```suggestion
export * as constants from './sso/constants'
```

packages/amazonq/src/extension.ts
opieter-aws: Sure, will update

packages/amazonq/src/lsp/client.ts
```diff
@@ -80,57 +103,89 @@ export async function startLanguageServer(
...
+            getLogger().error(`Failed to open document for LSP: ${lspName}, error: %s`, err)
+            return { success: false }
         }
+    })
+
+    client.onRequest<MessageActionItem | null, Error>(
+        ShowMessageRequest.method,
+        async (params: ShowMessageRequestParams) => {
+            const actions = params.actions?.map((a) => a.title) ?? []
+            const response = await vscode.window.showInformationMessage(params.message, { modal: true }, ...actions)
```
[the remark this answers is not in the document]
opieter-aws: @hayemaxi implemented the functionality, can you please give your input here?

packages/amazonq/src/lsp/activation.ts
opieter-aws: It's required because the `AuthUtil.create` below on line 19 needs the `client` as input

packages/amazonq/src/extension.ts
opieter-aws: Yes, if you have a better suggestion let me know!

packages/amazonq/src/lsp/client.ts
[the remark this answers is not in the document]
hayemaxi: In this case yes because it is required to proceed with auth (shows device code). We can't proceed if the user doesn't ack the message.
Sadly LSP does not provide a way to tell the client any information that would help us decide modal vs non modal, so it is set as blanket here. Maybe we can inspect the message items to determine what to display as modal or not.

packages/amazonq/src/lsp/activation.ts
hayemaxi: What if we did AuthUtil.create inside the lsp setup stage? Although I like separating the AuthUtil out of that, but I know this topic came up before.

opieter-aws: So it seems that we need to choose between separation of concerns and avoiding merge conflicts. The former is more desirable from a code health perspective, so if the latter weighs stronger for you @justinmk3 let me know and I can refactor this.

packages/amazonq/src/lsp/client.ts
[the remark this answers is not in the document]
opieter-aws: Can we do this as a follow-up, or if we notice during the bug bash that the behavior is not desirable?

packages/amazonq/src/lsp/activation.ts
justinmk3: My comment only applies if this was cosmetic-only. Sounds like that's not the case.

packages/core/src/auth/auth2.ts
```diff
@@ -31,12 +31,27 @@ import {
...
+    ConnectionMetadata,
+    getConnectionMetadataRequestType,
 } from '@aws/language-server-runtimes/protocol'
 import { LanguageClient } from 'vscode-languageclient'
 import { getLogger } from '../shared/logger/logger'
 import { ToolkitError } from '../shared/errors'
 import { useDeviceFlow } from './sso/ssoAccessTokenProvider'
 
+export const notificationTypes = {
+    updateBearerToken: new RequestType<UpdateCredentialsParams, ResponseMessage, Error>(
```
jpinkney-aws, line 46: do these types exist in the `@aws/language-server-runtimes/protocol`? IIRC they didn't originally but its worth a check

packages/core/src/amazonq/extApi.ts
```diff
@@ -21,7 +26,15 @@ export interface api {
...
-        getChatAuthState(): Promise<FeatureAuthState>
+        /**
+         * @deprecated, for backwards comaptibility only.
+         */
+        getChatAuthState(): Promise<{
+            codewhispererCore: OldAuthState
+            codewhispererChat: OldAuthState
+            amazonQ: OldAuthState
+        }>
+        getAuthState(): auth2.AuthState
```
jpinkney-aws, line 37: this is technically exposing the API to other extensions (like the toolkit extension) is this actually needed?

packages/amazonq/src/app/chat/activation.ts
```diff
@@ -44,8 +44,6 @@ export async function activate(context: ExtensionContext) {
         amazonq.walkthroughSecurityScanExample.register(),
         amazonq.openAmazonQWalkthrough.register(),
         amazonq.listCodeWhispererCommandsWalkthrough.register(),
-        amazonq.focusAmazonQPanel.register(),
```
jpinkney-aws, line 47: why are these getting de-registered? 🤔

opieter-aws: They are being registered at [packages/amazonq/src/extensionNode.ts](https://github.com/aws/aws-toolkit-vscode/pull/7062/files#diff-6b997bb5d94b505e70014a44e596d5a37de5b86a0fae13aec03afc6e6e64038d) now. VSCode was showing errors that they were missing since this activation is only called behind a feature flag ` if (!Experiments.instance.get('amazonqChatLSP', false)) {
        await activateCWChat(context) `

packages/core/src/auth/auth2.ts
opieter-aws: Not directly reusable AFAIK: https://github.com/aws/language-server-runtimes/blob/d914bec8ce1fc841c9b73fb622c68559be44917f/runtimes/protocol/auth.ts

packages/core/src/amazonq/extApi.ts
hayemaxi: IIRC viceroy was using this to determine whether or not to display Q sign-in notifications. I would imagine that hasn't changed, but needs verification.

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -29,11 +29,22 @@ export const codeWhispererCoreScopes = [...scopesCodeWhispererCore]
 export const codeWhispererChatScopes = [...codeWhispererCoreScopes, ...scopesCodeWhispererChat]
 export const amazonQScopes = [...codeWhispererChatScopes, ...scopesGumby, ...scopesFeatureDev]
 
+/** AuthProvider interface for the auth functionality needed by RegionProfileManager  */
+export interface IAuthProvider {
```
hayemaxi, line 33: Is this needed? RegionProvider is only used for Q. Why not just pass the Q specific utility class for it to use? `AuthUtil`.

opieter-aws: IIUC that's what I attempted initially, but it introduces a circular dependency. See https://github.com/aws/aws-toolkit-vscode/commit/b381d0d5a572e473e97b92c9a538a43b6042ebd9#diff-b1c252677272c2b57d62cc5d3da2ee04c060e1db7432c3ea40ab00f9ab26558aR41-R59

jpinkney-aws (approved):

packages/core/src/codewhisperer/util/authUtil.ts
opieter-aws: We can revisit the implementation as a follow-up, merging the PR to unblock next PR with references

PR #7065, opened 2025-04-16
fix(lsp): manifest resolver always reports "Failed to download"

justinmk3:
continues https://github.com/aws/aws-toolkit-vscode/pull/7043

## Problem:
Manifest resolver always reports:

    Failed to download latest "…" manifest. Falling back to local manifest.

## Solution:
- In `fetchRemoteManifest()`, if the ETag indicates no new manifest is needed, return the local manifest instead of throwing an error
- Fix other ambiguous/misleading log messages.

---

### DIFF
--- a/packages/core/src/shared/lsp/lspResolver.ts
+++ b/packages/core/src/shared/lsp/lspResolver.ts
@@ -40,44 +40,41 @@ export class LanguageServerResolver {
                 languageServerVersion: result.version,
             }
         }
-        try {
-            const latestVersion = this.latestCompatibleLspVersion()
-            const targetContents = this.getLSPTargetContents(latestVersion)
-            const cacheDirectory = this.getDownloadDirectory(latestVersion.serverVersion)
+        const latestVersion = this.latestCompatibleLspVersion()
+        const targetContents = this.getLSPTargetContents(latestVersion)
+        const cacheDirectory = this.getDownloadDirectory(latestVersion.serverVersion)
 
-            const serverResolvers: StageResolver<LspResult>[] = [
-                {
-                    // 1: Use the current local ("cached") LSP server bundle, if any.
-                    resolve: async () => await this.getLocalServer(cacheDirectory, latestVersion, targetContents),
-                    telemetryMetadata: { id: this.lsName, languageServerLocation: 'cache' },
-                },
-                {
-                    // 2: Download the latest LSP server bundle.
-                    resolve: async () => await this.fetchRemoteServer(cacheDirectory, latestVersion, targetContents),
-                    telemetryMetadata: { id: this.lsName, languageServerLocation: 'remote' },
-                },
-                {
-                    // 3: If the download fails, try an older, cached version.
-                    resolve: async () => await this.getFallbackServer(latestVersion),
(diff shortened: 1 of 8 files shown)


opieter-aws (approved):

packages/core/src/shared/lsp/lspResolver.ts
```diff
@@ -40,44 +40,41 @@ export class LanguageServerResolver {
...
-             *   assetDirectory = "<cachedir>/aws/toolkits/language-servers/AmazonQ/3.3.0"
-             *   location = 'cache'
-             *   version = '3.3.0'
-             * }
-             * ```
-             */
-            const resolved = await tryStageResolvers('getServer', serverResolvers, getServerVersion)
-            return resolved
-        } finally {
-            logger.info(`Finished setting up LSP server`)
```
justinmk3, line 79: Misleading use of `finally`. If an exception was thrown, it didn't really "finish" setup?

packages/core/src/shared/lsp/manifestResolver.ts
```diff
@@ -64,9 +64,17 @@ export class ManifestResolver {
         }).getNewETagContent(this.getEtag())
 
         if (!resp.content) {
-            throw new ToolkitError(
-                `New content was not downloaded; fallback to the locally stored "${this.lsName}" manifest`
-            )
+            const msg = `"${this.lsName}" manifest unchanged, skipping download: ${this.manifestURL}`
```
justinmk3, line 67: `!resp.content` means "etag didnt change". It doesn't mean "failed to download".

packages/core/src/shared/lsp/manifestResolver.ts
```diff
@@ -64,9 +64,17 @@ export class ManifestResolver {
...
+            logger.debug(msg)
+
+            const localManifest = await this.getLocalManifest(true).catch(() => undefined)
+            if (localManifest) {
+                localManifest.location = 'remote'
```
justinmk3, line 72: The common case is that etag didnt change. So that's handled here by re-using `getLocalManifest`. `ManifestResolver` doesn't have a way to tell its caller "I didnt fail but please continue to the next step".

packages/core/src/shared/lsp/manifestResolver.ts
```diff
@@ -76,13 +84,17 @@ export class ManifestResolver {
         return manifest
     }
 
-    private async getLocalManifest(): Promise<Manifest> {
-        logger.info(`Failed to download latest "${this.lsName}" manifest. Falling back to local manifest.`)
+    private async getLocalManifest(silent: boolean = false): Promise<Manifest> {
```
justinmk3, line 87: Always reporting "failed to download" here is misleading because this will happen when "etag didnt change".

packages/core/src/shared/lsp/manifestResolver.ts
```diff
@@ -145,6 +157,7 @@ export class ManifestResolver {
             ...storage,
             [this.lsName]: {
                 etag,
+                // XXX: this stores the entire manifest. vscode warns about our globalStorage size...
                 content,
```
justinmk3, line 161: we are storing the entire manifest in globalStorage :/

PR #7067, opened 2025-04-16
feat(core): Store codewhisperer customization in global state as single value and consume new authUtil

opieter-aws:
## Problem
With the old VSCode auth, customization was stored in global state (indexed by connection ID). With the new auth from the Flare identity server, this no longer applies since there is only a single connection

## Solution
* Consume the new AuthUtil in the customizationUtil
* Change `CODEWHISPERER_SELECTED_CUSTOMIZATION` in the global state to be a single value instead of a map

## Follow-up
Investigate if it's possible to hook up the customization to the Q profile from the new Region Manager, so customizations are persisted between different logins

---

### DIFF
--- a/packages/core/src/codewhisperer/util/customizationUtil.ts
+++ b/packages/core/src/codewhisperer/util/customizationUtil.ts
@@ -98,10 +98,7 @@ export const getSelectedCustomization = (): Customization => {
         return baseCustomization
     }
 
-    const selectedCustomization = globals.globalState.tryGet<Customization>(
-        'CODEWHISPERER_SELECTED_CUSTOMIZATION',
-        Object
-    )
+    const selectedCustomization = globals.globalState.getCodewhispererCustomization(AuthUtil.instance.profileName)
 
     if (selectedCustomization && selectedCustomization.name !== '') {
         return selectedCustomization
--- a/packages/core/src/codewhisperer/util/customizationUtil.ts
+++ b/packages/core/src/codewhisperer/util/customizationUtil.ts
@@ -139,25 +139,15 @@ export const getPersistedCustomizations = (): Customization[] => {
     if (!AuthUtil.instance.isIdcConnection()) {
         return []
     }
-    const persistedCustomizationsObj = globals.globalState.tryGet<{ [label: string]: Customization[] }>(
-        'CODEWHISPERER_PERSISTED_CUSTOMIZATIONS',
-        Object,
-        {}
-    )
-    return persistedCustomizationsObj[AuthUtil.instance.profileName] || []
+    return globals.globalState.getCodewhispererPersistedCustomization(AuthUtil.instance.profileName)
 }
 
 export const setPersistedCustomizations = async (customizations: Customization[]) => {
     if (!AuthUtil.instance.isIdcConnection()) {
         return
     }
-    const persistedCustomizationsObj = globals.globalState.tryGet<{ [label: string]: Customization[] }>(
-        'CODEWHISPERER_PERSISTED_CUSTOMIZATIONS',
-        Object,
-        {}
-    )
-    persistedCustomizationsObj[AuthUtil.instance.profileName] = customizations
-    await globals.globalState.update('CODEWHISPERER_PERSISTED_CUSTOMIZATIONS', persistedCustomizationsObj)
+
+    await globals.globalState.update('CODEWHISPERER_PERSISTED_CUSTOMIZATIONS', customizations)
 }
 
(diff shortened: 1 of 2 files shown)


justinmk3:
> * Change `CODEWHISPERER_SELECTED_CUSTOMIZATION` in the global state to be a single value instead of a map

If you need to add logic to migrate the old value (for existing users), a good place to do that is adding a new function to the `GlobalState` module. That's one of its purposes. Example: https://github.com/aws/aws-toolkit-vscode/blob/a26437735b08d60fa21e7db9269ad6f2c5a3c1bb/packages/core/src/shared/globalState.ts#L212

Analogous case for settings (as opposed to globalState): https://github.com/aws/aws-toolkit-vscode/blob/a26437735b08d60fa21e7db9269ad6f2c5a3c1bb/packages/core/src/shared/settings.ts#L663-L672

packages/core/src/codewhisperer/util/customizationUtil.ts
```diff
@@ -150,27 +139,27 @@ export const setSelectedCustomization = async (customization: Customization, isO
...
 export const setPersistedCustomizations = async (customizations: Customization[]) => {
-    if (!AuthUtil.instance.isValidEnterpriseSsoInUse() || !AuthUtil.instance.conn) {
+    if (!AuthUtil.instance.isIdcConnection()) {
         return
     }
     const persistedCustomizationsObj = globals.globalState.tryGet<{ [label: string]: Customization[] }>(
         'CODEWHISPERER_PERSISTED_CUSTOMIZATIONS',
         Object,
         {}
     )
```
hayemaxi, line 161: This object has the same issue we solved above. You are indexing the object with a static key, so we don't need the object. Just store an array of Customization[].

packages/core/src/shared/globalState.ts
```diff
@@ -227,6 +228,27 @@ export class GlobalState implements vscode.Memento {
...
 
+    /**
+     * Get the codewhisperer customerization. If legacy (map of customizations) store the
+     * customization with label of profile name
+     *
+     * @param profileName name of profile, only used in case legacy customization is found
+     * @returns codewhisperer customization, or undefined if not found.
+     * If legacy, return the codewhisperer customization for the auth profile name
+     */
+    getCodewhispererCustomization(profileName: string): Customization | undefined {
```
hayemaxi, line 239: Seems overly specific to codewhisperer customizations, can't we just have a generic migration function for `globalState.ts`

opieter-aws: I think there's so much specific input here, that you basically end up with everything as input (`tryGet` params, legacy determination and behavior for legacy). You'd just have a skeleton for the `tryGet` call, and the `if else` block, not sure if it make sense to have a generic function for that?

packages/core/src/codewhisperer/util/customizationUtil.ts
```diff
@@ -94,20 +94,11 @@ export const baseCustomization = {
...
         return baseCustomization
     }
 
-    const selectedCustomizationArr = globals.globalState.tryGet<{ [label: string]: Customization }>(
-        'CODEWHISPERER_SELECTED_CUSTOMIZATION',
-        Object,
-        {}
-    )
-    const selectedCustomization = selectedCustomizationArr[AuthUtil.instance.conn.label]
+    const selectedCustomization = globals.globalState.getCodewhispererCustomization(AuthUtil.instance.profileName)
```
justinmk3, line 101: cw is a legacy name isn't it? the key name doesn't matter
```suggestion
    const selectedCustomization = globals.globalState.getAmazonQCustomization(AuthUtil.instance.profileName)
```

packages/core/src/shared/globalState.ts
```diff
@@ -227,6 +228,53 @@ export class GlobalState implements vscode.Memento {
...
+
+    /**
+     * Get the codewhisperer persisted customizations. If legacy (map of customizations) store the
+     * customizations with label of profile name
+     *
+     * @param profileName name of profile, only used in case legacy customization is found
+     * @returns array of codewhisperer persisted customizations, or empty array if not found.
+     * If legacy, return the codewhisperer persisted customizations for the auth profile name
+     */
+    getCodewhispererPersistedCCustomization(profileName: string): Customization[] {
```
justinmk3, line 260: are these basically "cached" data or what is meant by persisted? if it's just a cache we should call it that.

```suggestion
    getAmazonQCachedCustomizations(profileName: string): Customization[] {
```

justinmk3 (approved):

PR #7070, opened 2025-04-16
fix(amazonq): quick action handler should add to tab storage

jpinkney-aws:
## Problem
currently tabs get added to the tabs storage when the onTabAdd handler gets called.

In hybrid chat, the onTabAdd handler will never get called, because it first needs to be in the tabsStorage, which doesn't happen because the onTabAdd handler hasn't been called

## Solution
in hybrid chat mode just add new tabs directly to the tabs storage

---

### DIFF
--- a/packages/core/src/amazonq/webview/ui/quickActions/handler.ts
+++ b/packages/core/src/amazonq/webview/ui/quickActions/handler.ts
@@ -19,6 +19,7 @@ export interface QuickActionsHandlerProps {
     isScanEnabled: boolean
     isTestEnabled: boolean
     isDocEnabled: boolean
+    hybridChat?: boolean
     disabledCommands?: string[]
 }
 
@@ -40,6 +41,7 @@ export class QuickActionHandler {
     private isScanEnabled: boolean
     private isTestEnabled: boolean
     private isDocEnabled: boolean
+    private isHybridChatEnabled: boolean
 
     constructor(props: QuickActionsHandlerProps) {
         this.mynahUIRef = props.mynahUIRef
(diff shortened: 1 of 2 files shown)


packages/core/src/amazonq/webview/ui/quickActions/handler.ts
```diff
@@ -343,6 +325,33 @@ export class QuickActionHandler {
...
 
+    private addTab(affectedTabId: string | undefined) {
+        if (!affectedTabId || !this.mynahUI) {
+            return
+        }
+
+        const currTab = this.mynahUI.getAllTabs()[affectedTabId]
+        const currTabWasUsed =
+            (currTab.store?.chatItems?.filter((item) => item.type === ChatItemType.PROMPT).length ?? 0) > 0
+        if (currTabWasUsed) {
```
Hweinstock, line 336: what exactly is this case for? What do we do differently when the tab is used/not used?

Hweinstock (approved):
Not super familiar with these handlers, but the flow makes sense. LGTM

packages/core/src/amazonq/webview/ui/quickActions/handler.ts
jpinkney-aws: It was a note from Oksana's original p.o.c: https://github.com/aws/language-servers/pull/733/files. As far as I understand in flare you can't create a tab and then later convert it. In the old version we were basing things on tab types which wouldn't exist in chat-client so instead we're just checking if someones interacted with the current tab before re-using it. If its used we create a new tab, otherwise we re-use the current tab

PR #7075, opened 2025-04-16
fix(amazonq): improve lsp cleanup logic

jpinkney-aws:
## Problem
When switching between different manifests, a newly downloaded version might chronologically be older than all previously downloaded versions, even though it's marked as the latest version in its own manifest. This is why we are getting the EPIPE error when trying out the alpha manifest. The cleanup deletes the language server right after it was downloaded

## Solution
In such cases, we skip the cleanup process to preserve this version. Otherwise we will get an EPIPE error. At this point the version that was downloaded shouldn't be delisted, so we don't want to make sure its not removed

---

### DIFF
--- a/packages/core/src/shared/lsp/baseLspInstaller.ts
+++ b/packages/core/src/shared/lsp/baseLspInstaller.ts
@@ -51,7 +51,11 @@ export abstract class BaseLspInstaller<T extends ResourcePaths = ResourcePaths, 
 
         await this.postInstall(assetDirectory)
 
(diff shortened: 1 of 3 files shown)


justinmk3 (approved):

Hweinstock (approved):
Great find! I'll update the LSP Spec with this finding and call it out in case any other IDEs missed this edge case as well.

jpinkney-aws:
> Also, any reason this shouldn't go into master?
Nope, it should be back ported. This was mostly just to unblock the bug bash tomorrow

PR #7076, opened 2025-04-16
fix(lsp): verify that nodejs runs (again)

justinmk3:
Problem:
The logic added in a26437735b08 wasn't added in the other "lsp startup" module.

Solution:
Add it to the new module also.

image

### DIFF
--- a/packages/amazonq/src/lsp/client.ts
+++ b/packages/amazonq/src/lsp/client.ts
@@ -11,11 +11,21 @@ import { InlineCompletionManager } from '../app/inline/completion'
 import { AmazonQLspAuth, encryptionKey, notificationTypes } from './auth'
 import { AuthUtil } from 'aws-core-vscode/codewhisperer'
 import { ConnectionMetadata } from '@aws/language-server-runtimes/protocol'
-import { Settings, oidcClientName, createServerOptions, globals, Experiments, Commands } from 'aws-core-vscode/shared'
+import {
+    Settings,
+    oidcClientName,
+    createServerOptions,
+    globals,
+    Experiments,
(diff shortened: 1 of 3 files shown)


packages/amazonq/src/lsp/client.ts
```diff
@@ -25,24 +35,27 @@ export async function startLanguageServer(
...
     const serverModule = resourcePaths.lsp
 
+    const argv = [
+        '--nolazy',
+        '--preserve-symlinks',
+        '--stdio',
+        '--pre-init-encryption',
+        '--set-credentials-encryption-key',
+    ]
     const serverOptions = createServerOptions({
```
justinmk3, line 45: I forgot that we have 2 of these modules. Will this module (`packages/amazonq/src/lsp/client.ts`) replace `packages/core/src/amazonq/lsp/lspClient.ts` @jpinkney-aws  ?

jpinkney-aws: `lsp/client.ts` is the amazon q language server, `lsp/lspClient.ts` is the workspace context language server

jpinkney-aws: We may end up deprecating the `lsp/lspClient.ts` one eventually

jpinkney-aws (approved):

packages/amazonq/src/lsp/client.ts
justinmk3: > We may end up deprecating the `lsp/lspClient.ts` one eventually now that workspace context is technically in flare

yes.  also, this logic is pretty similar yet lives in 2 different places

PR #7083, opened 2025-04-17
feat(chat): Adding the grep search tool and file search tool 

grl-wang:
## Problem

    adding the grep search and file search tool for testing
## Solution
    1. Implemented the grep search tool using https://github.com/microsoft/vscode-ripgrep.
    2. The grepped result show as clickable file path.
    3. Implemented the file search tool.
    3. The tool currently is disable in current pr, once grepSearch tool tested, will update the tool_index.json to enable this tool.

---


packages/core/src/codewhispererChat/tools/grepSearch.ts
```diff
@@ -0,0 +1,246 @@
...
+            })
+            .join('\n')
+
+        return { sanitizedOutput, totalMatchCount }
+    }
+
+    private createOutput(content: string): InvokeOutput {
+        return {
+            output: {
+                kind: OutputKind.Text,
```
jguoamz, line 241: use JSON format seems like a better fit for this tool

packages/core/src/codewhispererChat/tools/fileSearch.ts
```diff
@@ -0,0 +1,111 @@
...
+        if (!isInWorkspace) {
+            return { requiresAcceptance: true }
+        }
+        return { requiresAcceptance: false }
+    }
+
+    public async invoke(updates?: Writable): Promise<InvokeOutput> {
+        try {
+            const fileUri = vscode.Uri.file(this.fsPath)
+            const allFiles = await readDirectoryRecursively(fileUri, this.maxDepth)
```
jguoamz, line 87: filter to only keep `[F]` prefix?

buildspec/packageTestVsix.yml
```diff
@@ -20,6 +21,9 @@ phases:
     pre_build:
         commands:
             - export HOME=/home/codebuild-user
+            # Set up VSCODE_RIPGREP_TOKEN for GitHub API access
+            - export VSCODE_RIPGREP_TOKEN=${GITHUB_READONLY_TOKEN}
+            - echo "Setting up VSCODE_RIPGREP_TOKEN for GitHub API access"
```
jguoamz, line 26: why do we need this?

.github/workflows/node.js.yml
```diff
@@ -132,6 +132,7 @@ jobs:
             NODE_OPTIONS: '--max-old-space-size=8192'
             AWS_TOOLKIT_TEST_CACHE_DIR: '/tmp/.vscode-test/'
             AWS_TOOLKIT_TEST_USER_DIR: '/tmp/.vscode-test/user-data/'
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
grl-wang: we are setting this up to increase GitHub API Limit while download the ripgrep binary. 
 ref: https://github.com/microsoft/vscode-ripgrep
 ```
 You can produce an API key, set the GITHUB_TOKEN environment var to it, and vscode-ripgrep will use it when downloading from GitHub. This increases your API limit.
 ```
 1. Without this the most CI test will fail during download ripegrep binary.

package.json
```diff
@@ -75,6 +75,7 @@
     "dependencies": {
         "@types/node": "^22.7.5",
         "vscode-nls": "^5.2.0",
-        "vscode-nls-dev": "^4.0.4"
+        "vscode-nls-dev": "^4.0.4",
+        "@vscode/ripgrep": "1.15.11"
```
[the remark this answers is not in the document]
grl-wang: yes, based on the vscode ripgrep, it asked to add in the dependencies, not dev dependencies: https://github.com/microsoft/vscode-ripgrep.

buildspec/packageTestVsix.yml
grl-wang: similar as the GITHUB TOKEN env, set it setting this up to increase GitHub API Limit while download the ripgrep binary. otherwise, the ripgrep download failed.

PR #7085, opened 2025-04-17
telemetry(chat): try catch for telemetry to not break tool execution

chungjac:
## Problem
- `error: TypeError: Cannot read properties of undefined (reading 'emit') `
- telemetry can throw an error and break tool execution

## Solution
- use `run` instead of emit
- add try catch

---

### DIFF
--- a/packages/core/src/codewhispererChat/controllers/chat/telemetryHelper.ts
+++ b/packages/core/src/codewhispererChat/controllers/chat/telemetryHelper.ts
@@ -218,27 +218,39 @@ export class CWCTelemetryHelper {
     }
 
     public recordToolUseSuggested(toolUse: ToolUse, messageId: string) {
-        telemetry.amazonq_toolUseSuggested.emit({
-            result: 'Succeeded',
-            cwsprChatConversationId: messageId,
-            cwsprChatConversationType: 'AgenticChatWithToolUse',
-            credentialStartUrl: AuthUtil.instance.startUrl,
-            cwsprToolName: toolUse.name ?? '',
-            cwsprToolUseId: toolUse.toolUseId ?? '',
-        })
+        try {
+            telemetry.amazonq_toolUseSuggested.run((span) => {
+                span.record({
+                    result: 'Succeeded',
+                    cwsprChatConversationId: messageId,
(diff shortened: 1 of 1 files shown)


packages/core/src/codewhispererChat/controllers/chat/telemetryHelper.ts
```diff
@@ -218,27 +218,39 @@ export class CWCTelemetryHelper {
...
+                    result: 'Succeeded',
+                    cwsprChatConversationId: messageId,
+                    cwsprChatConversationType: 'AgenticChatWithToolUse',
+                    credentialStartUrl: AuthUtil.instance.startUrl,
+                    cwsprToolName: toolUse.name ?? '',
+                    cwsprToolUseId: toolUse.toolUseId ?? '',
+                })
+            })
+        } catch (e: any) {
+            getLogger().info('Unable to record amazonq_toolUseSuggested telemetry')
```
jguoamz, line 233: `error`

packages/core/src/codewhispererChat/controllers/chat/telemetryHelper.ts
```diff
@@ -218,27 +218,39 @@ export class CWCTelemetryHelper {
...
-            result: 'Succeeded',
-            cwsprChatConversationId: messageId,
-            cwsprChatConversationType: 'AgenticChatWithToolUse',
-            credentialStartUrl: AuthUtil.instance.startUrl,
-            cwsprToolName: toolUse.name ?? '',
-            cwsprToolUseId: toolUse.toolUseId ?? '',
-        })
+        try {
+            telemetry.amazonq_toolUseSuggested.run((span) => {
+                span.record({
```
jguoamz, line 223: I don't know too much about this, can you share some insights why this is better?

chungjac: `emit` sends the metric to telemetry service, which may cause an error. it also apparently will be deprecated, so that may be the reason it errors

`run` executes a callback, automatically sending the metric after completion. with using `run`, i am pretty sure it won't error, but it's good to have the try catch to be safe

jguoamz (approved):

Issue #7031, opened 2025-04-14
amazonq: In a VS Code notebook, auto-complete ignores other cells as context


brdskggs:
## Problem

In a Jupyter notebook opened in VS Code, Amazon Q is only making suggestions based on the current cell, and appears to be ignoring other cells as context.

### Context works in the same cell

1. Create a notebook with one cell
2. In the cell, enter: 
```
def add_two(a, b):
    return a + b
```
3. In the same cell, on the next line, type `add`. Trigger auto complete, and see suggestions related to the `add_two` function that you just defined.

### Context doesn't work in different cells
1. Create a notebook with two cells
2. In the first cell, enter: 
```
def add_two(a, b):
    return a + b
```
3. In the next cell, type `add`. Trigger auto complete, and see suggestions ***un-related*** to the function you just typed.

## Expected behavior
I expect context from other cells to be used to suggest output in a notebook, not ignored.

## System details (run `AWS: About` and/or `Amazon Q: About`)

-   OS: macOS 15.4
-   Visual Studio Code version: 1.99.2 
-   AWS Toolkit version: 3.54.0
-   Amazon Q version: 1.59.0

Hweinstock:
Appears this can be resolved once https://github.com/aws/aws-toolkit-vscode/pull/7086 is merged.

PR #7086, opened 2025-04-17
fix(amazonq): Previous and subsequent cells are used as context for completion in a Jupyter notebook

brdskggs:
## Problem
VS Code treats each cell in a notebook as a separate editor.  As a result, when building the left- and right-contexts for the completion from the current editor, we were limited to just the current cell, which might be very small and/or reference variables and functions defined in other cells.  That meant that completions never used the context of other cells when making suggestions, and were often _very_ generic. https://github.com/aws/aws-toolkit-vscode/issues/7031

## Solution
The `extractContextForCodeWhisperer` function now checks if it is being called in a cell in a Jupyter notebook.  If so, it collects the surrounding cells to use as context, respecting the maximum context length.  During this process, Markdown cells have each line prefixed with a language-specific comment character.

---

### DIFF
--- a/packages/amazonq/test/unit/codewhisperer/util/editorContext.test.ts
+++ b/packages/amazonq/test/unit/codewhisperer/util/editorContext.test.ts
@@ -214,78 +214,41 @@ describe('editorContext', function () {
     })
 
     describe('extractPrefixCellsContext', function () {
-        it('Should extract content from cells in reverse order up to maxLength', function () {
+        it('Should extract content from cells in reverse order up to maxLength from prefix cells', function () {
             const mockCells = [
                 createNotebookCell(createMockDocument('First cell content')),
                 createNotebookCell(createMockDocument('Second cell content')),
                 createNotebookCell(createMockDocument('Third cell content')),
             ]
 
-            const result = EditorContext.extractPrefixCellsContext(mockCells, 100, 'python')
+            const result = EditorContext.extractCellsSliceContext(mockCells, 100, 'python', false)
             assert.strictEqual(result, 'First cell content\nSecond cell content\nThird cell content\n')
         })
 
-        it('Should respect maxLength parameter', function () {
+        it('Should extract content from cells in reverse order up to maxLength from suffix cells', function () {
+            const mockCells = [
+                createNotebookCell(createMockDocument('First cell content')),
+                createNotebookCell(createMockDocument('Second cell content')),
+                createNotebookCell(createMockDocument('Third cell content')),
+            ]
+
+            const result = EditorContext.extractCellsSliceContext(mockCells, 100, 'python', true)
+            assert.strictEqual(result, 'First cell content\nSecond cell content\nThird cell content\n')
+        })
+
+        it('Should respect maxLength parameter from prefix cells', function () {
             const mockCells = [
                 createNotebookCell(createMockDocument('First')),
                 createNotebookCell(createMockDocument('Second')),
                 createNotebookCell(createMockDocument('Third')),
                 createNotebookCell(createMockDocument('Fourth')),
             ]
-
-            const result = EditorContext.extractPrefixCellsContext(mockCells, 15, 'python')
+            // Should only include part of second cell and the last two cells
+            const result = EditorContext.extractCellsSliceContext(mockCells, 15, 'python', false)
             assert.strictEqual(result, 'd\nThird\nFourth\n')
         })
 
-        it('Should handle empty cells array', function () {
-            const result = EditorContext.extractPrefixCellsContext([], 100, '')
-            assert.strictEqual(result, '')
-        })
-
-        it('Should add python comments to markdown cells', function () {
-            const mockCells = [
-                createNotebookCell(createMockDocument('# Heading\nThis is markdown'), vscode.NotebookCellKind.Markup),
-                createNotebookCell(createMockDocument('def example():\n    return "test"')),
-            ]
-            const result = EditorContext.extractPrefixCellsContext(mockCells, 100, 'python')
-            assert.strictEqual(result, '# # Heading\n# This is markdown\ndef example():\n    return "test"\n')
-        })
-
-        it('Should add java comments to markdown and python cells when language is java', function () {
-            const mockCells = [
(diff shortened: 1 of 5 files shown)


packages/core/src/codewhisperer/util/editorContext.ts
```diff
@@ -22,19 +22,87 @@ import { AuthUtil } from './authUtil'
...
+                break
+            }
+            output.unshift(cellText)
+            maxLength -= cellText.length
+        }
+    }
+    return output.join('')
+}
+
+export function extractSuffixCellsContext(
```
justinmk3, line 71: sure seems similar to the above function, can it be deduplicated?

justinmk3 (approved):
This is helpful, thanks! Tests for this function are here: https://github.com/aws/aws-toolkit-vscode/blob/d457022377ad804f8a9c99d3f4964123c31deead/packages/amazonq/test/unit/codewhisperer/util/editorContext.test.ts#L32

You can run that test file by opening it in vscode then choose `Extension tests (current file) (amazon)` in the run/debug menu:

image

packages/core/src/codewhisperer/util/editorContext.ts
brdskggs: Yes, I'll try to clean it up -- the main difference is which direction to shorten.

brdskggs:
Will consolidate the function doing the suffix and prefix context fetching.

brdskggs:
image

Both code cells and Markdown cells are used for context now.

packages/amazonq/test/unit/codewhisperer/util/editorContext.test.ts
```diff
@@ -63,6 +85,44 @@ describe('editorContext', function () {
             }
             assert.deepStrictEqual(actual, expected)
         })
+
+        it('Should include context from other cells when in a notebook', async function () {
```
justinmk3, line 89:
```suggestion
        it('in a Notebook, includes context from other cells', async function () {
```

packages/amazonq/test/unit/codewhisperer/util/editorContext.test.ts
```diff
@@ -115,6 +175,165 @@ describe('editorContext', function () {
         })
     })
 
+    describe('extractSingleCellContext', function () {
```
justinmk3, line 178: "notebook" is the keyword.

```suggestion
    describe('getNotebookCellContext', function () {
```

packages/core/src/codewhisperer/util/editorContext.ts
```diff
@@ -23,19 +23,83 @@ import { AuthUtil } from './authUtil'
 
 let tabSize: number = getTabSizeSetting()
 
+const languageCommentChars: Record<string, string> = {
+    python: '# ',
+    java: '// ',
+}
+
+export function extractSingleCellContext(cell: vscode.NotebookCell, referenceLanguage?: string): string {
```
justinmk3, line 31:
```suggestion
export function getNotebookCellContext(cell: vscode.NotebookCell, referenceLanguage?: string): string {
```

packages/core/src/codewhisperer/util/editorContext.ts
```diff
@@ -23,19 +23,83 @@ import { AuthUtil } from './authUtil'
...
+        return cell.document
+            .getText()
+            .split('\n')
+            .map((line) => `${commentPrefix}${line}`)
+            .join('\n')
+    }
+    return cellText
+}
+
+export function extractCellsSliceContext(
```
justinmk3, line 53: "notebook" is a useful nonce that helps discovery. especially since these functions are mingled with non-notebook logic.

```suggestion
export function getNotebookCellsSliceContext(
```

packages/core/src/codewhisperer/util/editorContext.ts
```diff
@@ -46,6 +110,44 @@ export function extractContextForCodeWhisperer(editor: vscode.TextEditor): codew
         languageName =
             runtimeLanguageContext.normalizeLanguage(editor.document.languageId) ?? editor.document.languageId
     }
+    if (editor.document.uri.scheme === 'vscode-notebook-cell') {
```
justinmk3, line 113: can this logic be extracted
```suggestion
    if (editor.document.uri.scheme === 'vscode-notebook-cell') {
        caretLeftFileContext = getNotebookContext(...)
        caretRightFileContext = getNotebookContext(...)
    }
```

brdskggs: Pulled out the notebook logic into two functions: one for determining the surrounding notebook, and then one for extending the context if inside a notebook.

packages/core/src/codewhisperer/util/editorContext.ts
justinmk3: ^

brdskggs: I thought I hit this one - may have gotten caught up in my rebase!

justinmk3:
lint failure

    Error:   112:13  error  'cellText' is never reassigned. Use 'const' instead  prefer-const

packages/core/src/codewhisperer/util/editorContext.ts
```diff
@@ -24,19 +24,130 @@ import { predictionTracker } from '../nextEditPrediction/activation'
 
 let tabSize: number = getTabSizeSetting()
 
+const languageCommentChars: Record<string, string> = {
+    python: '# ',
+    java: '// ',
```
justinmk3, line 29: I just noticed this. It's not sustainable for us to introduce yet another language-specific handling just for notebooks. Why isn't the notebook logic using the existing logic used by other Q functionality ?

brdskggs: Good question! I'm new to this codebase, so I'll have to research what existing logic you are referring to -- any quick pointers of where to start looking are appreciated.

brdskggs: Never mind, Q to the rescue -- figured out the changes I needed to make to `runtimeLanguageContext`.

PR #7103, opened 2025-04-18
fix(chat): Fix unwanted destructive warning for the commands which not in the destructive commands list

Randall-Jiang:
## Problem
Right now, when user run the commands outside the workspace, it would have destructive warning
image

## Solution

https://github.com/user-attachments/assets/38752309-0cc9-4994-82a1-67148cfae17d

---

### DIFF
--- a/packages/core/src/codewhispererChat/tools/executeBash.ts
+++ b/packages/core/src/codewhispererChat/tools/executeBash.ts
@@ -200,13 +200,13 @@ export class ExecuteBash {
                         }
                         const workspaceFolders = vscode.workspace.workspaceFolders
                         if (!workspaceFolders || workspaceFolders.length === 0) {
-                            return { requiresAcceptance: true, warning: destructiveCommandWarningMessage }
+                            return { requiresAcceptance: true }
                         }
                         const isInWorkspace = workspaceFolders.some((folder) =>
                             isInDirectory(folder.uri.fsPath, fullPath)
                         )
                         if (!isInWorkspace) {
-                            return { requiresAcceptance: true, warning: destructiveCommandWarningMessage }
+                            return { requiresAcceptance: true }
                         }
                     }
                 }


zixlin7 (approved):

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -189,24 +189,24 @@ export class ExecuteBash {
...
                     if (this.looksLikePath(arg)) {
-                        // If not absolute, resolve using workingDirectory if available.
                         let fullPath = arg
                         if (!path.isAbsolute(arg) && this.workingDirectory) {
                             fullPath = path.join(this.workingDirectory, arg)
                         }
                         const workspaceFolders = vscode.workspace.workspaceFolders
                         if (!workspaceFolders || workspaceFolders.length === 0) {
-                            return { requiresAcceptance: true, warning: destructiveCommandWarningMessage }
+                            hasOutsideWorkspacePath = true
```
tsmithsz, line 202: Do we want to set `hasOutsideWorkspacePath = true` for all commands when no workspace is open? This will trigger acceptance requirements for read-only commands that shouldn't need them

packages/core/src/codewhispererChat/tools/executeBash.ts
```diff
@@ -220,6 +220,9 @@ export class ExecuteBash {
                     case CommandCategory.Mutate:
                         return { requiresAcceptance: true, warning: mutateCommandWarningMessage }
                     case CommandCategory.ReadOnly:
+                        if (hasOutsideWorkspacePath) {
+                            return { requiresAcceptance: true }
```
tsmithsz, line 224: Can we add some tests for this?

tsmithsz (approved):

packages/core/src/codewhispererChat/tools/executeBash.ts
tsmithsz: Approving. Please follow-up with @bywang on this case

PR #7109, opened 2025-04-18
feat(nep): Data Instrumentation

tomcat323:
## Problem
The science & service team in Next Edit Prediction (NEP) project needs initial data for model training and tuning, to provide more contextually relevant suggestions. 

This requires us to track user edits in the IDE, and send changes (of current active file in unified diff format) to the codeWhisper API. 
**No impact on user experience.**
**This won't live forever, will be migrated to flare by end of May, needed now for science data collection.**

### Key Components
`PredictionKeyStrokeHandler`: Listens for document changes, maintains shadow copies of visible documents, and processes edits.
`PredictionTracker`: Manages file snapshots, implementing a policy for storing, retrieving, and pruning snapshots based on age and memory constraints.
`DiffGenerator`: Creates unified diffs between file snapshots, produces `supplementalContext` sent to the API.

### How it Works
- The system track shadow copies of editor visible files' content
- Once an edit is made to a tracked file, it takes a snapshot of the file content before the edit
- When the Inline API fires, the snapshots of the current editing files are used to generate diff context

### Memory management
- maxTotalSizeKb (default: 5000): Caps total size of snapshots storage at ~5MB, purging oldest snapshots when exceeded.
- debounceIntervalMs (default: 2000): Prevents excessive snapshots by requiring 2 seconds between captures for the same file.
- maxAgeMs (default: 30000): Auto-deletes snapshots after 30 seconds to maintain recent-only history.
- maxSupplementalContext (default: 15): Limits `supplementalContext` sent to API to 15 entries maximum.

### Changes
- Added new NextEditPrediction module in the CodeWhisperer package
- Updated activation code to initialize the NEP system
- Updated codeWhisper inline API requests to fit new format
---


packages/core/src/codewhisperer/nextEditPrediction/PredictionKeyStrokeHandler.ts
```diff
@@ -0,0 +1,118 @@
+/*!
```
justinmk3, line 1: should this these NEP modules live in the same dir as the other "trackers"? 

https://github.com/aws/aws-toolkit-vscode/tree/master/packages/core/src/codewhisperer/tracker

are they really not sharing any concepts at all? this PR is all new code.

tomcat323: No they are not shared, the existing tracker collects statistical metrics for telemetry, the NEP tracker tracks content changes. Their hyper parameters are also very different.

Since the new tracker is populating the `supplementalContext` field, maybe we can also consider https://github.com/aws/aws-toolkit-vscode/tree/master/packages/core/src/codewhisperer/util/supplementalContext?

packages/core/src/codewhisperer/service/recommendationHandler.ts
```diff
@@ -329,6 +329,7 @@ export class RecommendationHandler {
                 session.requestIdList.push(requestId)
             }
             getLogger().debug(msg)
+            getLogger().info(`NEP testing: codeWhisper request ${requestId}`)
```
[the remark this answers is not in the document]
tomcat323: Yes, this is for bug-bashing purpose, to surface the request id so we can query that in our backend. 
Removed for this PR

packages/core/src/codewhisperer/nextEditPrediction/PredictionTracker.ts
```diff
@@ -0,0 +1,430 @@
...
+    constructor(extensionContext: vscode.ExtensionContext, config?: Partial<FileTrackerConfig>) {
+        this.config = {
+            ...predictionTrackerDefaultConfig,
+            ...config,
+        }
+
+        // Use workspace storage
+        this.storagePath = extensionContext.storageUri?.fsPath
+
+        void this.ensureStorageDirectoryExists()
```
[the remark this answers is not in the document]
tomcat323: Modified, now this is just creating the snapshot storage directory if it doesn't exist yet.

packages/core/src/codewhisperer/activation.ts
```diff
@@ -95,10 +95,12 @@ import { SecurityIssueTreeViewProvider } from './service/securityIssueTreeViewPr
 import { setContext } from '../shared/vscode/setContext'
 import { syncSecurityIssueWebview } from './views/securityIssue/securityIssueWebview'
 import { detectCommentAboveLine } from '../shared/utilities/commentUtils'
+import { activateNextEditPrediction } from './nextEditPrediction/activation'
 
 let localize: nls.LocalizeFunc
 
 export async function activate(context: ExtContext): Promise<void> {
+    activateNextEditPrediction(context)
```
nkomonen-amazon, line 103: - I would move this to the bottom of the function since the beginning is for more general activation (eg auth).

- Also I think it is better to have this in `activateAmazonQCommon(). We want to move as much Q specific code in to their so that we can reduce our dependency on `core`.

packages/core/src/codewhisperer/nextEditPrediction/PredictionTracker.ts
```diff
@@ -0,0 +1,394 @@
...
+            size,
+            timestamp,
+            storageKey,
+        }
+
+        // Get existing snapshots for this file
+        const fileSnapshots = this.snapshots.get(filePath) || []
+
+        // Check if we should add a new snapshot given the debounce
+        const shouldAddSnapshot =
```
nkomonen-amazon, line 88: Should we check this condition as early as possible in this function and return if false. Then you won't need to do as much unneeded setup.

packages/core/src/codewhisperer/util/editorContext.ts
```diff
@@ -119,23 +120,55 @@ export async function buildListRecommendationRequest(
...
     logSupplementalContext(supplementalContexts)
 
+    // Get predictionSupplementalContext from PredictionTracker
+    let predictionSupplementalContext: codewhispererClient.SupplementalContext[] = []
+    try {
+        if (predictionTracker) {
+            predictionSupplementalContext = await predictionTracker.generatePredictionSupplementalContext()
+        }
+    } catch (error) {
+        getLogger().error(`Error getting prediction supplemental context: ${error}`)
```
nkomonen-amazon, line 130: You have the final call, but is it dangerous to swallow the error here? If a user is having issues we will never have any telemetry for this.

dhanainme (changes requested):
Given this codebase will move to Flare eventually. Can we do a minimal refactor this to make the migration straight forward ?

APIs which may not have a direct mapping

| VS Code API | LSP Equivalent | Notes |
|:------------|:---------------|:------|
| `vscode.window.onDidChangeVisibleTextEditors` | **No direct equivalent in LSP** | LSP focuses on opened/closed documents (`didOpen`, `didClose`) not "visibility". You may need to track `didOpen`/`didClose` and assume editors are visible. |
| `vscode.workspace.onDidChangeTextDocument` | `textDocument/didChange` (notification) | LSP server receives `textDocument/didChange` to reflect edits in documents. |
| `vscode.window.visibleTextEditors` | **No direct equivalent in LSP** | LSP does not track *visible editors*. It only tracks *open documents* via `didOpen`, `didClose`, `didChange`. |
| `vscode.window.activeTextEditor` | **No direct equivalent in LSP** | LSP does not track *which* document is active. It only manages documents' content and lifecycles. (The client — e.g., VS Code — knows which one is active.) |


Also lets not use `fs` APIs and rely on message passing as-much as we can.

packages/core/src/codewhisperer/activation.ts
tomcat323: I'm not sure if this needs to have it's own `index.ts` and export it to QCommon, since it's not a stand-alone feature, but rather an addition on existing codeWhisper API flow. Will move it to the bottom of codeWhisper activation.

packages/core/src/codewhisperer/util/editorContext.ts
tomcat323: It's not dangerous, we are collecting initial data, and from the bug-bash no issues has been found. Telemetry will come later in this project. 
The main thing is if anything goes wrong it doesn't crash the inline completions flow, hence we swallow it here.

packages/core/src/codewhisperer/nextEditPrediction/PredictionTracker.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     * Gets all tracked files
+     * @returns Array of file paths
+     */
+    public getTrackedFiles(): string[] {
+        return Array.from(this.snapshots.keys())
+    }
+
+    public getTotalSnapshotCount(): number {
```
dhanainme, line 189: Can you please add a global try-catch + error log for all public methods which will be called for completions code path. 

This defensive approach would allow us to give guarantee that we don't break Completions code path inadvertently.

packages/core/src/codewhisperer/nextEditPrediction/PredictionTracker.ts
```diff
@@ -0,0 +1,235 @@
...
+        return snapshot.content
+    }
+
+    /**
+     * Generates unified diffs between adjacent snapshots of a file
+     * and between the newest snapshot and the current file content
+     *
+     * @returns Array of SupplementalContext objects containing diffs between snapshots and current content
+     */
+    public async generatePredictionSupplementalContext(): Promise<codewhispererClient.SupplementalContext[]> {
```
dhanainme, line 203: As called out above

Can you please add a global try-catch + error log for all public methods which will be called for completions code path.

packages/core/src/codewhisperer/nextEditPrediction/activation.ts
```diff
@@ -0,0 +1,28 @@
...
+import * as vscode from 'vscode'
+import { PredictionTracker } from './PredictionTracker'
+import { PredictionKeyStrokeHandler } from './PredictionKeyStrokeHandler'
+import { getLogger } from '../../shared/logger/logger'
+import { ExtContext } from '../../shared/extensions'
+
+export let predictionTracker: PredictionTracker | undefined
+let keyStrokeHandler: PredictionKeyStrokeHandler | undefined
+
+export function activateEditTracking(context: ExtContext): void {
```
dhanainme, line 15: As called out above

Can you please add a global try-catch + error log for all public methods which will be called for completions code path.

packages/core/src/codewhisperer/nextEditPrediction/PredictionTracker.ts
```diff
@@ -0,0 +1,235 @@
...
+            ...config,
+        }
+    }
+
+    /**
+     * Processes an edit to a document and takes a snapshot if needed
+     * @param document The document being edited
+     * @param previousContent The content of the document before the edit
+     */
+    public async processEdit(document: vscode.TextDocument, previousContent: string): Promise<void> {
```
dhanainme, line 50: As called out above

Can you please add a global try-catch + error log for all public methods which will be called for completions code path.

dhanainme (changes requested):

packages/core/src/codewhisperer/nextEditPrediction/PredictionTracker.ts
tomcat323: these are only used for tests

dhanainme (approved):

packages/core/src/codewhisperer/nextEditPrediction/PredictionKeyStrokeHandler.ts
```diff
@@ -0,0 +1,117 @@
...
+                    this.shadowCopies.delete(filePath)
+                }
+            }
+        })
+
+        this.disposables.push(visibleDisposable)
+    }
+
+    private updateShadowCopy(document: vscode.TextDocument): void {
+        if (document.uri.scheme === 'file') {
```
jpinkney-aws, line 78: is this check redundant? it looks like every time this is called we already know that the document.uri.scheme is file AFAICT

tomcat323: Good callout, it is already checked in `PredictionKeyStrokeHandler`

jpinkney-aws (approved):

PR #7113, opened 2025-04-20
fix(amazonq): send proper cursor position to lsp

Hweinstock:
## Problem
If you currently ask Q any questions about cursor position or highlighted text, it lacks the context necessary to answer. This is because in the payload sent to the LSP we send the cursor position as the wrong type:
https://github.com/aws/aws-toolkit-vscode/blob/2e8d4c7eb0f0bcbc416d8f6e06ff7e00d57dc6b2/packages/amazonq/src/lsp/chat/messages.ts#L159
whereas [Flare](https://github.com/aws/language-server-runtimes/blob/d9960b5dd6df52c52309f0a7edae64a824da9003/types/chat.ts#L82) is expecting: `cursorState`.

## Solution
- Convert the vscode selection to a Flare cursorState. 
- Add stronger typing by casting the params to `ChatParams` type before modifying to avoid these types of bugs. 

## Future Work
- This fixes flare being unable to identify cursor position, and mostly fixes highlighted text. However, I was still able to run into situations where the cursor position is correct, but the content is incorrect. ~~This could be fixed by sending the highlighted content to Flare, but~~ still trying to root cause the issue. 

image

---

### DIFF
--- a/packages/amazonq/src/lsp/chat/messages.ts
+++ b/packages/amazonq/src/lsp/chat/messages.ts
@@ -72,6 +72,21 @@ export function registerLanguageServerEventListener(languageClient: LanguageClie
     })
 }
 
+function getCursorState(selection: readonly vscode.Selection[]) {
+    return selection.map((s) => ({
+        range: {
+            start: {
+                line: s.start.line,
+                character: s.start.character,
+            },
+            end: {
+                line: s.end.line,
+                character: s.end.character,
+            },
+        },
+    }))
+}
+
 export function registerMessageListeners(
     languageClient: LanguageClient,
     provider: AmazonQChatViewProvider,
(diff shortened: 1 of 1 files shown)


packages/amazonq/src/lsp/chat/messages.ts
```diff
@@ -147,20 +162,21 @@ export function registerMessageListeners(
...
                 const chatDisposable = languageClient.onProgress(chatRequestType, partialResultToken, (partialResult) =>
-                    handlePartialResult<ChatResult>(partialResult, encryptionKey, provider, message.params.tabId)
+                    handlePartialResult<ChatResult>(partialResult, encryptionKey, provider, chatParams.tabId)
                 )
 
                 const editor =
                     vscode.window.activeTextEditor ||
                     vscode.window.visibleTextEditors.find((editor) => editor.document.languageId !== 'Log')
                 if (editor) {
-                    message.params.cursorPosition = [editor.selection.active]
```
jpinkney-aws, line 159: I guess this means the demo vscode client from flare also has this problem, unless they fixed it after I copied this chunk of code

jpinkney-aws:
> However, I was still able to run into situations where the cursor position is correct, but the content is incorrect

Before we send it to the backend so we grab the text from the selections? or is that done by internal flare apis?

packages/amazonq/src/lsp/chat/messages.ts
Hweinstock: Good to know, I'll check on that.

Hweinstock:
Based on the implementation [here](https://github.com/aws/language-servers/blob/7266d32b2fb73ead40abecb22749a2c9e5206a2a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts#L116), it looks like Flare sends the `cursorState` and the document contents, but its up to the backend to select the text. Unless its passed explicitly somewhere else I haven't found.

Hweinstock:
After some more digging, I'm only able to reproduce this when fetching from alpha-manifest. I can't reproduce when overriding the language server directly.

jpinkney-aws (approved):

PR #7114, opened 2025-04-20
fix(amazonq): Sync client settings to language server

nkomonen-amazon:
When certain settings/configs are changed in the IDE, they need to be propagated to the language server.
The settings were the customizationArn, the users opt in preference for telemetry, and some other things.

This PR:
- Pushes the new customization arn when it is changed. This arn only exists when the user is using the non-default customization, otherwise it is undefined.
  - A small hack needed to be made to explicitly push this change, since by default we utilized the automatic push to the LS when the users settings file changed (eg opt in telemetry). But because customization is not stored in the users settings, we didn't get that for free. See the command `aws.amazonq.updateCustomizations` which was made to handle this edge case. Which eventually **explicitly** uses the language server message `DidChangeConfigurationNotification`
- Pushes the opt out telemetry setting to the language server
  - Since this setting is stored in the vscode settings json, the push to language server mechanism comes for free under the `DidChangeConfigurationNotification` message. 
- The Q Auth Profile ARN is a separate config that needed to be handled by a different handler (different from customization), so the function `pushConfigUpdate()` was created to route the correct config to the correct handler in the LS
  - This uses the message `updateConfigurationRequestType` which is different from `DidChangeConfigurationNotification` that the other configs use. This results in a different code path being hit, which is expected.

Some helpful notes:
- `getAmazonQRelatedWorkspaceConfigs()` (triggered through `DidChangeConfigurationNotification`) is the language server side code that parses the config sent by the client/ide. So the client/ide must match the shape expected by the LS. This is an example why `optOutTelemetry` did not work out of the box, and needed some changes to be consumable by the LS.  

---


packages/amazonq/src/lsp/client.ts
```diff
@@ -72,6 +74,47 @@ export async function startLanguageServer(
...
+                         * it must match that expected shape**.
+                         * https://github.com/aws/language-servers/blob/1d2ca018f2248106690438b860d40a7ee67ac728/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/configurationUtils.ts#L114
+                         */
+                        return [
+                            {
+                                customization,
+                                optOutTelemetry: getOptOutPreference() === 'OPTOUT',
+                                projectContext: {
+                                    // TODO uncomment this when local indexing is in agent chat server manifest
+                                    enableLocalIndexing: false,
```
zixlin7, line 97: thanks! we should be able to set it to true now

aws-toolkit-vscode.code-workspace
```diff
@@ -12,6 +12,9 @@
         {
             "path": "packages/amazonq",
         },
+        {
+            "path": "../language-servers",
```
jpinkney-aws, line 16: you can remove this

packages/amazonq/.vscode/launch.json
```diff
@@ -13,9 +13,9 @@
             "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
             "env": {
                 "SSMDOCUMENT_LANGUAGESERVER_PORT": "6010",
-                "WEBPACK_DEVELOPER_SERVER": "http://localhost:8080"
-                // "__AMAZONQLSP_PATH": "${workspaceFolder}/../../../language-servers/app/aws-lsp-codewhisperer-runtimes/out/agent-standalone.js",
-                // "__AMAZONQLSP_UI": "${workspaceFolder}/../../../language-servers/chat-client/build/amazonq-ui.js"
+                "WEBPACK_DEVELOPER_SERVER": "http://localhost:8080",
+                "__AMAZONQLSP_PATH": "${workspaceFolder}/../../../language-servers/app/aws-lsp-codewhisperer-runtimes/out/agent-standalone.js",
```
jpinkney-aws, line 17: you can recomment these (or just uncommit the file) otherwise if devs don't have the language-server repos downloaded it's going to fail to launch

packages/amazonq/src/lsp/client.ts
```diff
@@ -72,6 +74,47 @@ export async function startLanguageServer(
...
+                        const customization = undefinedIfEmpty(getSelectedCustomization().arn)
+                        /**
+                         * IMPORTANT: This object is parsed by the following code in the language server, **so
+                         * it must match that expected shape**.
+                         * https://github.com/aws/language-servers/blob/1d2ca018f2248106690438b860d40a7ee67ac728/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/configurationUtils.ts#L114
+                         */
+                        return [
+                            {
+                                customization,
+                                optOutTelemetry: getOptOutPreference() === 'OPTOUT',
```
jpinkney-aws, line 94: oh wait this is a boolean? the type they have in the interface vs what is expected to be passed in is so confusing 😅

nkomonen-amazon: Yeah its weird lol. They expect a boolean then convert that back in to an 'OPTIN' | 'OPTOUT' string:
https://github.com/aws/language-servers/blob/c367ef3a1374032dace0e6755eaa43a1fae6e3c4/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/configurationUtils.ts#L117

packages/amazonq/.vscode/launch.json
nkomonen-amazon: Just gonna comment theses out again

packages/amazonq/src/lsp/client.ts
nkomonen-amazon: Done

jpinkney-aws (approved):

packages/core/src/codewhisperer/util/customizationUtil.ts
```diff
@@ -147,6 +147,9 @@ export const setSelectedCustomization = async (customization: Customization, isO
     }
     vsCodeState.isFreeTierLimitReached = false
     await Commands.tryExecute('aws.amazonq.refreshStatusBar')
+
+    // hack: triggers amazon q to send the customizations back to flare
+    await Commands.tryExecute('aws.amazonq.updateCustomizations')
```
justinmk3, line 152: Why not call a function instead of a command?

Commands should be avoided, because they are less reliable and less visible if they fail (partly because they tend to be async).

PR #7123, opened 2025-04-22
fix(amazonq): increase polling frequency for bearer token. 

Hweinstock:
## Problem
During the bug bash, someone encountered the LSP attempting to use an expired bearer token. I am unable to reproduce the issue, but can speculate at the cause. 

We refresh the bearer token on the LSP every 1 minute, but if the token expires, and then a request is made before the next refresh it can fail. 

Note: this is a temporary solution until we have LSP auth. 

## Solution
- double the frequency with which we check the token. 
- We still only send the token if it changed, meaning this won't cause noisy request to LSP. 

---


jpinkney-aws (approved):
I can't remember if I understood it right, but I thought I read something that said flare will reject the token if the request fails. If thats the case then maybe thats what happened?

Hweinstock:
Yeah, still unsure what happened in the bug bash since it isn't documented, but this can't hurt. If we can get a detailed report, then I can dig deeper on the Flare side.

PR #7132, opened 2025-04-22
feat(amazonq): add open settings command

jpinkney-aws:
## Problem
chat-client will send an event to VSCode indicating what setting should be opened

## Solution
implement the event handler

I can't get this to actually trigger so this code is theoretical for now

---


packages/amazonq/src/lsp/chat/messages.ts
```diff
@@ -188,6 +197,12 @@ export function registerMessageListeners(
                 }
                 break
             }
+            case OPEN_SETTINGS: {
+                const params = message.params as OpenSettingsParams
+                const setting = params.settingKey as keyof SettingsProps
```
justinmk3, line 202: might as well be `as any`, since `as` means this doesn't actually achieve any type-checking. and that avoids needing to `export` SettingsProps.

justinmk3, line 202:
```suggestion
                const params = message.params as any
                const setting = params.settingKey as any
```

PR #7134, opened 2025-04-22
feat(amazonq): Profile cache

Will-ShaoHua:
## Solution

---

### DIFF
--- a/packages/core/src/codewhisperer/region/regionProfileManager.ts
+++ b/packages/core/src/codewhisperer/region/regionProfileManager.ts
@@ -54,6 +54,7 @@ export type ProfileSwitchIntent = 'user' | 'auth' | 'update' | 'reload'
 type CachedApiResult = {
     state: 'pending' | 'valid' | 'stale'
     profiles: RegionProfile[] | undefined
+    timestamp: number
 }
 
 export class RegionProfileManager {
@@ -110,7 +111,7 @@ export class RegionProfileManager {
 
     constructor(private readonly connectionProvider: () => Connection | undefined) {}
 
-    async listRegionProfile(loadCache: boolean = false): Promise<RegionProfile[]> {
+    async listRegionProfile(): Promise<RegionProfile[]> {
         this._profiles = []
 
         const conn = this.connectionProvider()
@@ -118,13 +119,19 @@ export class RegionProfileManager {
             return []
         }
         const availableProfiles: RegionProfile[] = []
-        if (loadCache) {
-            const cachedValue = this.loadApiResultFromCache()
-            if (cachedValue.state === 'pending') {
-                waitUntil(async function () {})
-            } else if (cachedValue.state === 'valid') {
-                return cachedValue.profiles as RegionProfile[]
+        const cachedValue = this.loadApiResultFromCache()
+        if (cachedValue.state === 'pending') {
+            const r = await waitUntil(
+                async () => {
+                    return this.loadApiResultFromCache()
+                },
+                { timeout: 15000, interval: 1000, truthy: false }
+            )
+            if (r && r.state === 'valid') {
+                return r.profiles as RegionProfile[]
             }
+        } else if (cachedValue.state === 'valid') {
+            return cachedValue.profiles as RegionProfile[]
         }
 
         await this.cacheApiResult(undefined)
@@ -317,6 +324,7 @@ export class RegionProfileManager {
         const pojo: CachedApiResult = {
             state: state,
             profiles: r,
+            timestamp: state === 'valid' ? globals.clock.Date.now() : 0,
         }
         await globals.globalState.update('aws.amazonq.regionProfiles.cachedResult', pojo)
     }
@@ -326,8 +334,16 @@ export class RegionProfileManager {
         const cachedValue = globals.globalState.tryGet<CachedApiResult>(
             'aws.amazonq.regionProfiles.cachedResult',
             Object,
-            { state: 'stale', profiles: undefined }
+            { state: 'stale', profiles: undefined, timestamp: 0 }
         )
+
+        const now = globals.clock.Date.now()
+        // Only use cache if the duration has not exceeded 60s
+        if (now - cachedValue.timestamp > 60000) {
+            const s: CachedApiResult = { state: 'stale', profiles: undefined, timestamp: 0 }
+            globals.globalState.update('aws.amazonq.regionProfiles.cachedResult', s).catch((e) => {})
+            return s
+        }
 
         return cachedValue
     }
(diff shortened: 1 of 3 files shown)


Will-ShaoHua:
todo: error handling

Will-ShaoHua:
will add a changelog

packages/core/src/codewhisperer/region/regionProfileManager.ts
```diff
@@ -42,13 +43,41 @@ const endpoints = createConstantMap({
 })
 
 /**
- * 'user' -> users change the profile through Q menu
+ * 'user' -> users change the profile through Q menu`
```
justinmk3, line 46: revert

packages/core/src/shared/globalState.ts
```diff
@@ -49,6 +49,7 @@ export type globalKey =
     | 'aws.toolkit.lsp.manifest'
     | 'aws.amazonq.customization.overrideV2'
     | 'aws.amazonq.regionProfiles'
+    | 'aws.amazonq.regionProfiles.cachedResult'
```
justinmk3, line 52: why not set a flag on the items in `aws.amazonq.regionProfiles` ? why does this need an entirely separate structure?

packages/core/src/codewhisperer/region/regionProfileManager.ts
```diff
@@ -110,7 +139,56 @@ export class RegionProfileManager {
         if (conn === undefined || !isSsoConnection(conn)) {
             return []
         }
+
+        // WaitUntil lock is acquired, the API tps is low, server can't afford multiple IDE instances making the calls simultaneiously
+        const cached = await waitUntil(
```
justinmk3, line 144: should this be encapsulated in `tryAcquireLock`? else the name `tryAcquireLock` is misleading, and confusing.

packages/core/src/codewhisperer/region/regionProfileManager.ts
```diff
@@ -367,4 +450,63 @@ export class RegionProfileManager {
...
+    }
+
+    private async updateCacheWithResult(r: RegionProfile[]) {
+        const result: SuccessResult = {
+            type: 'success',
+            profiles: r,
+            timestamp: now(),
+        }
+        const pojo: CachedApiResultWithLock = {
+            isAcquired: false, // release the lock
```
justinmk3, line 494: would the name `locked` be more obvious?
```suggestion
            locked: false, // release the lock
```

packages/core/src/codewhisperer/region/regionProfileManager.ts
```diff
@@ -367,4 +450,63 @@ export class RegionProfileManager {
...
+
+    // Should reset cache when users signout in case users want to change to a different connection
+    async resetCache() {
+        await globals.globalState.update('aws.amazonq.regionProfiles.cachedResult', {
+            isAcquired: false,
+            result: undefined,
+        })
+    }
+
+    private async tryAcquireLock(): Promise<CachedApiResultWithLock | undefined> {
```
justinmk3, line 462: This is rather delicate logic. And it is straightforward to generalize it. Why is it buried in a random module?

Note that https://github.com/aws/aws-toolkit-vscode/blob/70ba83fc066cfe8a1f9f50e5eb73a5e1b51d42f4/packages/core/src/shared/crashMonitoring.ts#L432 attempted to achieve something similiar, except using the filesystem. Though it has some known issues.

justinmk3 (commented):
Although this is urgent work, it's also urgent that it works correctly. So separating it into a shared module will help with that. Especially if bugs are discovered and must be debugged.

packages/core/src/codewhisperer/region/regionProfileManager.ts
Will-ShaoHua: ye

packages/core/src/shared/globalState.ts
Will-ShaoHua: > 'aws.amazonq.regionProfiles'

this is to memorize previous selected option, so we will need a separate structure to cache previous api call response

packages/core/src/codewhisperer/region/regionProfileManager.ts
```diff
@@ -367,4 +450,63 @@ export class RegionProfileManager {
...
+                ...cachedValue,
+                isAcquired: true,
+            })
+
+            return cachedValue
+        }
+        return undefined
+    }
+
+    private async releaeLock(cached: CachedApiResultWithLock) {
```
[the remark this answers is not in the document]
Will-ShaoHua: gotcha

packages/core/src/codewhisperer/region/regionProfileManager.ts
Will-ShaoHua: makes sense

packages/core/src/codewhisperer/region/regionProfileManager.ts
Will-ShaoHua: sg

packages/core/src/shared/globalState.ts
justinmk3: > > 'aws.amazonq.regionProfiles'
> 
> this is to memorize previous selected option

but that could be a flag on the item(s) in the structure. why is a separate structure needed?

Will-ShaoHua: that makes sense, i will reuse the existing one then

Will-ShaoHua: @justinmk3 i encounter an issue or say concern when i am doing this exercise https://github.com/aws/aws-toolkit-vscode/pull/7151/commits/3e501456fb1cd7b125819db2fbe7390cdfdc2558#diff-f5716036a70a8b212e9dae829ec09e33731f181c173a3f2b9ad9d361d51af14bR54-R56, if I want to reuse the same global key, i might need to restructure the schema, in such case, users who already selected profile update their plugin, they might need to re-select the profile again. Given the sev2s we have right now, I am concerned if we want to do this now or is there another way to deal with such scenario?

justinmk3: Definitely want to avoid a restructure. If a restructure is required, then I guess we need a new globalState field.

Will-ShaoHua:
#7151

packages/core/src/shared/globalState.ts
Will-ShaoHua: noted

packages/core/src/codewhisperer/region/regionProfileManager.ts
Will-ShaoHua: btw Justin, i'm going to use #7151 to replace this pr for this purpose (if everything works just fine there), because the diff are pretty large so would be easier to just have another PR.

PR #7145, opened 2025-04-23
fix(lsp): LSP download timeout message disappears too early

nkomonen-amazon:
## Problem:

When downloading the LSP artifacts from the manifest, there are some large ones (100+ MB).
For a slow connection this can take 1+ minutes, but our timeout for the "downloading" message
is set to disappear much earlier. So it is still downloading in the background, but the message
has disappeared and the user thinks that the download is done.

## Solution:

Increase the timeout to 30 minutes, which will ensure the downloading message sticks around
while the download is still happening.

## Additional

In another commit this fixes a separate HTTP client bug where it would time out a request if it took longer than 3 seconds.
This caused downloads to be aborted (separate from the download message disappearing). Now it times out after 30 minutes.

---

### DIFF
--- a/packages/core/src/shared/lsp/lspResolver.ts
+++ b/packages/core/src/shared/lsp/lspResolver.ts
@@ -15,9 +15,11 @@ import { lspSetupStage, StageResolver, tryStageResolvers } from './utils/setupSt
 import { HttpResourceFetcher } from '../resourcefetcher/httpResourceFetcher'
 import { showMessageWithCancel } from '../../shared/utilities/messages'
 import { Timeout } from '../utilities/timeoutUtils'
+import { oneMinute } from '../datetime'
 
-// max timeout for downloading remote LSP assets progress, the lowest possible is 3000, bounded by httpResourceFetcher's waitUntil
-const remoteDownloadTimeout = 5000
+// max timeout for downloading remote LSP assets. Some asserts are large (100+ MB) so this needs to be large for slow connections.
+// Since the user can cancel this one we can let it run very long.
+const remoteDownloadTimeout = oneMinute * 30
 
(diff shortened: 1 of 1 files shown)


packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
```diff
@@ -117,8 +118,10 @@ export class HttpResourceFetcher implements ResourceFetcher<Response> {
                 return req.response.finally(() => cancelListener?.dispose())
             },
             {
-                timeout: 3000,
-                interval: 100,
+                // Ensures we don't abort a long request (eg: slow connection).
+                // This timeout is just backup, the given Timeout object should be relied on instead.
+                timeout: oneMinute * 30,
```
jpinkney-aws, line 123: would this mean that all of Thomas' requests we're getting aborted? but if we waited 1.5 minutes they would eventually finish

jpinkney-aws (approved):

opieter-aws (approved):

packages/core/src/shared/resourcefetcher/httpResourceFetcher.ts
nkomonen-amazon: Yeah this was the weird part. Since my requests were getting aborted. I'm going to triple check once more

nkomonen-amazon: Found out that there is a difference between getting the initial response back (`fetch(...).response`) and `response.body`. `response.body` is called outside of the HttpResourceFetcher and is not subject to this value.

So only the initial response for just the headers is encapsulated in this, the actual body fetching is outside.

PR #7149, opened 2025-04-23
feat(lambda): Add checkbox in local invoke webview for debugger

mbfreder:
## Solution
Add checkbox in local invoke webview so customers can choose to locally invoke with or without debugger

### Note: 
No tests added since the functionality is already tested here: https://github.com/aws/aws-toolkit-vscode/blob/a7732b96982adbbac691f2869db3136c159d80ed/packages/core/src/test/shared/sam/debugger/samDebugConfigProvider.test.ts#L111

---

### DIFF
--- a/packages/core/src/lambda/vue/configEditor/samInvokeBackend.ts
+++ b/packages/core/src/lambda/vue/configEditor/samInvokeBackend.ts
@@ -348,14 +348,8 @@ export class SamInvokeWebview extends VueWebview {
      * Validate and execute the provided launch config.
      * TODO: Post validation failures back to webview?
      * @param config Config to invoke
-     * @param source Source of the invoke request
-     * @param useDebugger Whether to invoke with debugger attached
      */
-    public async invokeLaunchConfig(
-        config: AwsSamDebuggerConfiguration,
-        source?: string,
-        useDebugger: boolean = true
-    ): Promise<void> {
+    public async invokeLaunchConfig(config: AwsSamDebuggerConfiguration, source?: string): Promise<void> {
         const finalConfig = finalizeConfig(
             resolveWorkspaceFolderVariable(undefined, config),
             'Editor-Created Debug Config'
@@ -363,10 +357,8 @@ export class SamInvokeWebview extends VueWebview {
         const targetUri = await this.getUriFromLaunchConfig(finalConfig)
         const folder = targetUri ? vscode.workspace.getWorkspaceFolder(targetUri) : undefined
 
-        // Use the existing startDebugging method with the noDebug option
-        await vscode.debug.startDebugging(folder, finalConfig, {
-            noDebug: !useDebugger,
-        })
+        // startDebugging on VS Code goes through the whole resolution chain
+        await vscode.debug.startDebugging(folder, finalConfig)
     }
     public async getLaunchConfigQuickPickItems(
         launchConfig: LaunchConfiguration,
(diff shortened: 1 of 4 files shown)


packages/core/src/lambda/vue/configEditor/samInvokeBackend.ts
```diff
@@ -348,17 +348,25 @@ export class SamInvokeWebview extends VueWebview {
...
             'Editor-Created Debug Config'
         )
         const targetUri = await this.getUriFromLaunchConfig(finalConfig)
         const folder = targetUri ? vscode.workspace.getWorkspaceFolder(targetUri) : undefined
 
-        // startDebugging on VS Code goes through the whole resolution chain
-        await vscode.debug.startDebugging(folder, finalConfig)
+        // Use the existing startDebugging method with the noDebug option
+        await vscode.debug.startDebugging(folder, finalConfig, {
+            noDebug: !useDebugger,
```
justinmk3, line 368: ? This is already supported by the existing (inherited from vscode) `noDebug` field: https://github.com/aws/aws-toolkit-vscode/blob/70ba83fc066cfe8a1f9f50e5eb73a5e1b51d42f4/packages/core/src/shared/sam/debugger/awsSamDebugger.ts#L179-L183

packages/core/src/lambda/vue/configEditor/samInvokeFrontend.ts
```diff
@@ -170,7 +172,7 @@ export default defineComponent({
 
             const source = this.resourceData?.source
 
-            client.invokeLaunchConfig(config, source).catch((e: Error) => {
+            client.invokeLaunchConfig(config, source, this.useDebugger).catch((e: Error) => {
```
justinmk3, line 175: just set `config.noDebug`. you don't need a new `this.useDebugger` field.

justinmk3:
This is already supported when the user invokes a sam launch config via `Debug: Start Without Debugging` : 

image

However, it may still be desired to make that more obvious via this new checkbox.

packages/core/src/lambda/vue/configEditor/samInvokeFrontend.ts
mbfreder: `config` here doesn't have a `noDebug` property, so we need to pass `this.useDebugger` to be able it to use it in `vscode.debug.startDebugging()`. 

Function definition
```
function debug.startDebugging(folder: vscode.WorkspaceFolder | undefined, nameOrConfiguration: string | vscode.DebugConfiguration, parentSessionOrOptions?: vscode.DebugSession | vscode.DebugSessionOptions): Thenable<boolean>
```
`vscode.DebugSessionOptions` also has a `noDebug` field, which is what we're using in this case.

justinmk3: > `config` here doesn't have a `noDebug` property

that's just an oversight in this definition: https://github.com/aws/aws-toolkit-vscode/blob/db673c9b74b36591bb5642b3da7d4bc7ae2afaf4/packages/core/src/lambda/vue/configEditor/samInvokeFrontend.ts#L422

mbfreder: Updated.

mbfreder:
Test failures are unrelated to the changes.

packages/core/src/lambda/vue/configEditor/samInvokeFrontend.ts
```diff
@@ -48,6 +48,7 @@ interface SamInvokeVueData {
     showNameInput: boolean
     newTestEventName: string
     resourceData: ResourceData | undefined
+    useDebugger: boolean
```
justinmk3, line 51: this probably isn't needed either, because `launchConfig` on line 33 should already have `noDebug` field

mbfreder: Yes, it is binding to the vue.

PR #7150, opened 2025-04-23
feat(core): add value length cap to partialClone

Hweinstock:
## Problem
If we want to log a large json object with giant strings (think whole files), it can make the logs difficult to read. This is especially relevant when the underlying string values are not very important.  

## Solution
- allow the string values to be truncated with `maxStringLength` option. 

## Notes
I am planning to port this utility to Flare to improve the logging experience there, and want this functionality to exist there however I thought this could be useful here too. 

---

### DIFF
--- a/packages/core/src/shared/utilities/collectionUtils.ts
+++ b/packages/core/src/shared/utilities/collectionUtils.ts
@@ -7,6 +7,7 @@ import { isWeb } from '../extensionGlobals'
 import { inspect as nodeInspect } from 'util'
 import { AsyncCollection, toCollection } from './asyncCollection'
 import { SharedProp, AccumulableKeys, Coalesce, isNonNullable } from './tsUtils'
+import { truncate } from './textUtilities'
 
 export function union<T>(a: Iterable<T>, b: Iterable<T>): Set<T> {
     const result = new Set<T>()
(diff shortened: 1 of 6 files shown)


packages/core/src/test/shared/utilities/collectionUtils.test.ts
```diff
@@ -710,8 +710,10 @@ describe('CollectionUtils', async function () {
     })
 
     describe('partialClone', function () {
-        it('omits properties by depth', function () {
-            const testObj = {
+        let multipleTypedObj: object
```
Hweinstock, line 713: this is changed to avoid duplicate code lint fail.

packages/core/src/shared/utilities/collectionUtils.ts
```diff
@@ -304,26 +305,36 @@ export function assign<T extends Record<any, any>, U extends Partial<T>>(data: T
...
+    options?: {
+        replacement?: any
+        maxStringLength?: number
+    }
+): any {
     // Base case: If input is not an object or has no children, return it.
     if (typeof obj !== 'object' || obj === null || 0 === Object.getOwnPropertyNames(obj).length) {
-        return obj
+        const maxLength = options?.maxStringLength
+        return typeof obj === 'string' && maxLength !== undefined ? truncate(obj, maxLength, '...') : obj
```
nkomonen-amazon, line 322: nit: move this to an if condition since we may want to add manipulations for other types in the future

packages/core/src/test/shared/utilities/collectionUtils.test.ts
```diff
@@ -724,56 +726,69 @@ describe('CollectionUtils', async function () {
...
+                d: { d1: { d2: '[replaced]' } },
                 e: {
                     e1: [4, 3, 7],
-                    e2: '[omitted]',
+                    e2: '[replaced]',
+                },
+            })
+        })
+
+        it('truncates properties by maxLength', function () {
```
nkomonen-amazon, line 771: We want to also test when there are other non object types like number or boolean

nkomonen-amazon (approved):

opieter-aws (approved):

justinmk3 (approved):

packages/core/src/test/shared/utilities/collectionUtils.test.ts
Hweinstock: Good callout, I've adjusted the test to avoid only using strings.

Hweinstock:
failing test tracked in https://github.com/aws/aws-toolkit-vscode/issues/7153

PR #7151, opened 2025-04-24
fix(amazonq): fix enterprise users not able to sign in correctly if they have 2+ vscode instances open

Will-ShaoHua:
## Problem
revision of #7134, this pr aims to address the comment from the previous PR https://github.com/aws/aws-toolkit-vscode/pull/7134#discussion_r2056518840 to extract the logic to a shared module so the diff against 7134 is expected to be large.

after deployment 4/21, service has a strict 1 tps throttling policy and there was no caching for the API call previously.
It will impact users as long as they have multiple ide instances opened as all of them will make list profile call while users attempt to sign in.

## Solution
1. cache the api result for 60 seconds for reuse 
2. use lock to prevent multiple instances trying to call at the same time. Only 1 will make the service call and the rest will wait until it's done and read from the cache.

---

### DIFF
--- a/packages/core/src/codewhisperer/util/authUtil.ts
+++ b/packages/core/src/codewhisperer/util/authUtil.ts
@@ -143,6 +143,8 @@ export class AuthUtil {
             if (!this.isConnected()) {
                 await this.regionProfileManager.invalidateProfile(this.regionProfileManager.activeRegionProfile?.arn)
             }
+
+            await this.regionProfileManager.clearCache()
         })
 
         this.regionProfileManager.onDidChangeRegionProfile(async () => {
--- a/packages/core/src/codewhisperer/util/authUtil.ts
+++ b/packages/core/src/codewhisperer/util/authUtil.ts
@@ -142,7 +142,6 @@ export class AuthUtil {
 
             if (!this.isConnected()) {
                 await this.regionProfileManager.invalidateProfile(this.regionProfileManager.activeRegionProfile?.arn)
-                await this.regionProfileManager.resetCache()
             }
         })
 
--- a/packages/core/src/codewhisperer/util/authUtil.ts
+++ b/packages/core/src/codewhisperer/util/authUtil.ts
@@ -142,6 +142,7 @@ export class AuthUtil {
 
             if (!this.isConnected()) {
                 await this.regionProfileManager.invalidateProfile(this.regionProfileManager.activeRegionProfile?.arn)
+                await this.regionProfileManager.resetCache()
             }
         })
 
--- a/packages/core/src/shared/utilities/resourceCache.ts
+++ b/packages/core/src/shared/utilities/resourceCache.ts
@@ -68,7 +68,7 @@ export abstract class CachedResource<V> {
         return latest
     }
 
-    async readResourceAndLock(): Promise<GlobalStateSchema<V> | undefined> {
+    private async readResourceAndLock(): Promise<GlobalStateSchema<V> | undefined> {
         const _acquireLock = async () => {
             const cachedValue = this.readCacheOrDefault()
 
@@ -95,13 +95,6 @@ export abstract class CachedResource<V> {
         )
 
         return lock
-    }
-
-    async releaseLock() {
-        await globals.globalState.update(this.key, {
-            ...this.readCacheOrDefault(),
-            locked: false,
-        })
     }
 
     private async updateCache(cache: GlobalStateSchema<any> | undefined, resource: Resource<any>) {
--- a/packages/core/src/shared/utilities/resourceCache.ts
+++ b/packages/core/src/shared/utilities/resourceCache.ts
@@ -85,7 +85,7 @@ export abstract class CachedResource<V> {
                 )
             }
         } else {
-            logger.debug(`cache miss, pulling latest resource %s`, this.key)
+            logger.info(`cache miss, pulling latest resource %s`, this.key)
         }
 
         /**
@@ -104,14 +104,11 @@ export abstract class CachedResource<V> {
                 timestamp: now(),
                 result: latest,
             }
-            logger.debug(`doen loading latest resource, updating resource cache: %s`, this.key)
+            logger.info(`doen loading latest resource, updating resource cache: %s`, this.key)
             await this.releaseLock(r)
             return latest
         } catch (e) {
-            logger.debug(
-                `encountered unexpected error while loading the latest of resource(%s), releasing resource lock`,
-                this.key
-            )
+            logger.error(`failed to load latest resource, releasing lock: %s`, this.key)
             await this.releaseLock()
             throw e
         }
--- a/packages/core/src/shared/utilities/resourceCache.ts
+++ b/packages/core/src/shared/utilities/resourceCache.ts
@@ -38,6 +38,7 @@ function now() {
  * CacheResource utilizes VSCode global states API to cache resources which are expensive to get so that the result can be shared across multiple VSCode instances.
  *  The first VSCode instance invoking #getResource will hold a lock and make the actual network call/FS read to pull the real response.
  *  When the pull is done, the lock will be released and it then caches the result in the global states. Then the rest of instances can now acquire the lock 1 by 1 and read the resource from the cache.
+ *
  * constructor:
  *  @param key: global state key, which is used for globals.globalState#update, #tryGet etc.
  *  @param expirationInMilli: cache expiration time in milli seconds
--- a/packages/core/src/shared/utilities/resourceCache.ts
+++ b/packages/core/src/shared/utilities/resourceCache.ts
@@ -28,7 +28,7 @@ export interface GlobalStateSchema<V> {
     resource: Resource<V>
 }
 
-const logger = getLogger()
+const logger = getLogger('resourceCache')
 
 function now() {
     return globals.clock.Date.now()
@@ -63,12 +63,16 @@ export abstract class CachedResource<V> {
         if (cachedValue && resource && resource.result) {
             const duration = now() - resource.timestamp
             if (duration < this.expirationInMilli) {
-                logger.info(`cache hit, duration(%sms) is less than expiration(%sms)`, duration, this.expirationInMilli)
+                logger.debug(
+                    `cache hit, duration(%sms) is less than expiration(%sms)`,
+                    duration,
+                    this.expirationInMilli
+                )
                 // release the lock
                 await this.releaseLock(resource, cachedValue)
                 return resource.result
             } else {
-                logger.info(
+                logger.debug(
                     `cached value is stale, duration(%sms) is older than expiration(%sms), invoking service API to pull the latest response`,
                     duration,
                     this.expirationInMilli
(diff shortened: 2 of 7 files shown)


packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,127 @@
...
+interface Resource<V> {
+    result: V | undefined
+    locked: boolean
+    timestamp: number
+}
+
+// GlobalStates schema, which is used for vscode global states deserialization
+// [globals.globalState.tryGet<T>]
+export interface GlobalStateSchema<V> {
+    resource: Resource<V>
```
justinmk3, line 20: why is in this indirection needed?

Will-ShaoHua: not sure if i get your question correctly justin.

i was thinking and based on the comments in the previous PR #7151, we might want to put related things/fields under the same global key as much as we can but not only the resource itself. So here i need a super type which contains the resource but also callers can add whatever they want here and this class shouldn't touch field other than "resource".

For my own use case i want to store (1) resource (2) map of connectionId to previousSelectedProfile
```
interface RegionProfileGlobalState extends GlobalStateSchema<RegionProfile> {
    previousSelection: { [label: string]: RegionProfile }
}
```

does this make sense to you?

Will-ShaoHua: btw appreciate your early feedback, im not a typescript expert, your feedback would be very helpful

Will-ShaoHua: > For my own use case i want to store (1) resource (2) map of connectionId to previousSelectedProfile

```
interface RegionProfileGlobalState extends GlobalStateSchema<RegionProfile> {
    previousSelection: { [label: string]: RegionProfile }
}
```

as discussed in the previous PR, i won't use the same global key this time for not breaking existing users' experience

Will-ShaoHua: so i will revert https://github.com/aws/aws-toolkit-vscode/pull/7151/commits/3e501456fb1cd7b125819db2fbe7390cdfdc2558 partially

packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+const logger = getLogger()
+
+function now() {
+    return globals.clock.Date.now()
+}
+
+export abstract class CachedResource<V> {
```
Will-ShaoHua, line 29: will add docstring.

packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,127 @@
...
+                    ...resource,
+                    locked: false,
+                })
+                return resource.result
+            } else {
+                logger.info(`cache hit but cached value is stale, invoking service API to pull the latest response`)
+            }
+        }
+
+        // catch and error case?
```
Will-ShaoHua, line 57: still wip

packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,127 @@
...
+
+        const lock = await waitUntil(
+            async () => {
+                const lock = await _acquireLock()
+                logger.info(`try obtain resource cache read lock %s`, lock)
+                if (lock) {
+                    return lock
+                }
+            },
+            { timeout: 15000, interval: 1500, truthy: true } // TODO: pass via ctor
```
Will-ShaoHua, line 94: wip

Will-ShaoHua:
@justinmk3 i will add test as a followup, will focus on functional testing now

packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,177 @@
...
+                await this.lockResource(cachedValue)
+                return cachedValue
+            }
+
+            return undefined
+        }
+
+        const lock = await waitUntil(async () => {
+            const lock = await _acquireLock()
+            logger.info(`try obtain resource cache read lock %s`, lock)
```
justinmk3, line 124: is stringified `lock` going to log anything useful?
```suggestion
            logger.info(`try obtain resource cache read lock %s`, this.key)
```

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -142,6 +142,7 @@ export class AuthUtil {
 
             if (!this.isConnected()) {
                 await this.regionProfileManager.invalidateProfile(this.regionProfileManager.activeRegionProfile?.arn)
+                await this.regionProfileManager.resetCache()
```
justinmk3, line 145: should `invalidateProfile` do this internally?

packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,177 @@
...
+        const resource = cachedValue?.resource
+
+        // If cache is still fresh, return cached result, otherwise pull latest from the service
+        if (cachedValue && resource && resource.result) {
+            const duration = now() - resource.timestamp
+            if (duration < this.expirationInMilli) {
+                logger.info(`cache hit, duration(%sms) is less than expiration(%sms)`, duration, this.expirationInMilli)
+                // release the lock
+                await this.releaseLock(resource, cachedValue)
+                return resource.result
```
justinmk3, line 69: oh, I think I get it: multiple vscode instances will all acquire the lock, but the N+1 instances will each `return` here.

I guess that's fine. But it means they all must acquire a lock just to do the read. Maybe that's necessary.

packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,177 @@
...
+            } else {
+                logger.info(
+                    `cached value is stale, duration(%sms) is older than expiration(%sms), invoking service API to pull the latest response`,
+                    duration,
+                    this.expirationInMilli
+                )
+            }
+        }
```
justinmk3, line 77: avoid redundant messages

```suggestion
        } else {
            logger.info(`cache miss, invoking service API to pull the latest response`)
        }
```

packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * GlobalStates schema, which is used for vscode global states deserialization, [globals.globalState#tryGet<T>] etc.
+ * The purpose of it is to allow devs to overload the resource into existing global key and no need to create a specific key for only this purpose.
+ */
+export interface GlobalStateSchema<V> {
+    resource: Resource<V>
+}
+
+const logger = getLogger()
```
justinmk3, line 31:
```suggestion
const logger = getLogger('cache')
```

justinmk3 (approved):
LGTM, thanks for followup

packages/core/src/shared/utilities/resourceCache.ts
Will-ShaoHua: yea that's a good callout, changed it to print this.key

packages/core/src/shared/utilities/resourceCache.ts
Will-ShaoHua: oh yes, that's a valid point, thanks for catching this justin

packages/core/src/codewhisperer/util/authUtil.ts
Will-ShaoHua: this is kinda redundant i think, and also its name is confusing for sure (should've named it `releaseLock`), but i will remove it.

Will-ShaoHua: `invalidateProfile` doesn't need this i think as the source of truth of profile `invalidateProfile`  is also from this resourceCache. Originally i added this line is kinda for debugging purpose and to avoid a dead lock scenario when the code was problematic. However i think releaseLock will be called
1. when `getResource` goes to pull real response and executes successfully
2. when `getResource` goes to pull real response and executes exceptionally
3. when `getResource` goes to return cached value
4. if one process/ide instance waits to long (current it's set to 15s), it will go either (1) or (2) and eventually release the lock again

so I think it's redundant

Will-ShaoHua: just reminds me that  i need to add one `clearCache` onConnectionChanged.

packages/core/src/shared/utilities/resourceCache.ts
Will-ShaoHua: yea exactly, i should have mentioned it more explicitly in the doc string, it's like a read write lock if i understand it correctly. But not multiple vscode instances will all acquire the lock, only 1 can hold the lock and the rest of them need to wait until the first one finish pulling the real response and release the lock. Then the rest of them can start acquiring the lock 1 by 1 and return early here using the cached value.

Will-ShaoHua: there're definitely room to improve i think in the future. Not have enough time to consider all the potential use cases and requirements so currently it only meets the needs for listProfile / listCustomization.

leigaol (approved):

packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,193 @@
...
+            // Make the real network call / FS read to pull the resource
+            const latest = await this.resourceProvider()
+
+            // Update resource cache and release the lock
+            const r: Resource<V> = {
+                locked: false,
+                timestamp: now(),
+                result: latest,
+            }
+            logger.debug(`doen loading latest resource, updating resource cache: %s`, this.key)
```
justinmk3, line 107:
```suggestion
            logger.debug(`done loading latest resource, updating resource cache: %s`, this.key)
```

packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,193 @@
...
+            } else {
+                logger.debug(
+                    `cache is stale, duration(%sms) is older than expiration(%sms), pulling latest resource %s`,
+                    duration,
+                    this.expirationInMilli,
+                    this.key
+                )
+            }
+        } else {
+            logger.debug(`cache miss, pulling latest resource %s`, this.key)
```
justinmk3, line 88: this one should be at least INFO level
```suggestion
            logger.info(`cache miss, pulling latest resource %s`, this.key)
```

packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,193 @@
...
+                locked: false,
+                timestamp: now(),
+                result: latest,
+            }
+            logger.debug(`doen loading latest resource, updating resource cache: %s`, this.key)
+            await this.releaseLock(r)
+            return latest
+        } catch (e) {
+            logger.debug(
+                `encountered unexpected error while loading the latest of resource(%s), releasing resource lock`,
```
justinmk3, line 112:
```suggestion
            logger.error(
                `failed to load latest resource, releasing lock: %s`,
```

packages/core/src/shared/utilities/resourceCache.ts
```diff
@@ -0,0 +1,190 @@
...
+            } else {
+                logger.debug(
+                    `cache is stale, duration(%sms) is older than expiration(%sms), pulling latest resource %s`,
+                    duration,
+                    this.expirationInMilli,
+                    this.key
+                )
+            }
+        } else {
+            logger.info(`cache miss, pulling latest resource %s`, this.key)
```
justinmk3, line 88: When I open 2 vscode instances locally, I see `cache is stale` in one, and `cache miss` in the other. That means the fetch (service call) happens in both?

Will-ShaoHua: yea, one of my commit today broke it... #7173

PR #7167, opened 2025-04-25
feat(amazonq): support multiple uri in a single diff scheme

ctlai95:
## Problem
The current viewDiff implementation is restricted to only allowing comparing the current file contents. This means it's not possible to view a diff for a change that was applied at a previous point in time.

## Solution
- Create a new `DiffContentProvider` that can support showing multiple URIs in the same scheme. The current `ContentProvider` has a limitation that it can only show a single URI. 
- Refactor the code to use a defined type `ViewDiffMessage` instead of `any`
- Change the logic so that diffs are stored in memory instead of a temp file on disk

---

### DIFF
--- a/packages/core/src/amazonq/commons/controllers/diffContentProvider.ts
+++ b/packages/core/src/amazonq/commons/controllers/diffContentProvider.ts
@@ -1,0 +1,51 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import vscode from 'vscode'
+import { getLogger } from '../../../shared/logger/logger'
+
+/**
+ * A TextDocumentContentProvider that can handle multiple URIs with the same scheme.
+ * This provider maintains a mapping of URIs to their content.
+ */
+export class DiffContentProvider implements vscode.TextDocumentContentProvider {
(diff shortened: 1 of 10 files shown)


packages/core/src/amazonq/commons/controllers/diffContentProvider.ts
```diff
@@ -0,0 +1,51 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import vscode from 'vscode'
+import { getLogger } from '../../../shared/logger/logger'
+
+/**
+ * A TextDocumentContentProvider that can handle multiple URIs with the same scheme.
+ * This provider maintains a mapping of URIs to their content.
+ */
+export class DiffContentProvider implements vscode.TextDocumentContentProvider {
```
jpinkney-aws, line 12: isn't this basically equivalent to having a virtual file system in vscode and then adding virtual files to it?

ctlai95: I think yes, but I also need it to support multiple URIs to compare the before and after using the same URI scheme. I'm not sure if that part is supported by a virtual document

justinmk3 (approved):
Using a VFS is a good idea, but in this case it might not look much different in terms of new surface area (code + tests)?

jpinkney-aws (approved):

PR #7169, opened 2025-04-25
fix(amazonq): handle explain, refactor, fix, optimize, sendToPrompt

jpinkney-aws:
## Problem
explain, refactor, fix, optimize right click menus aren't working

## Solution
re-enable them through flare's genericCommand message

## TODO
- handle generateUnitTests
- handle explainIssue
both of these need flare API changes to send generic prompts from the frontend to the backend

---

### DIFF
--- a/packages/core/src/codewhispererChat/commands/registerCommands.ts
+++ b/packages/core/src/codewhispererChat/commands/registerCommands.ts
@@ -6,7 +6,6 @@
 import { commandPalette } from '../../codewhisperer/commands/types'
 import { CodeScanIssue } from '../../codewhisperer/models/model'
 import { Commands, VsCodeCommandArg, placeholder } from '../../shared/vscode/commands2'
-import { ChatControllerMessagePublishers } from '../controllers/chat/controller'
 
 /**
  * Opens the Amazon Q panel, showing the correct View that should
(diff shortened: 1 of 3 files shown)


packages/core/src/codewhispererChat/commands/registerCommands.ts
```diff
@@ -37,73 +36,13 @@ export const focusAmazonQPanelKeybinding = Commands.declare('_aws.amazonq.focusC
     await focusAmazonQPanel.execute(placeholder, 'keybinding')
 })
 
-const getCommandTriggerType = (data: any): EditorContextCommandTriggerType => {
```
Hweinstock, line 40: I am assuming this fixes the double registering that was happening before?

packages/core/src/codewhispererChat/commands/registerCommands.ts
```diff
@@ -37,73 +36,13 @@ export const focusAmazonQPanelKeybinding = Commands.declare('_aws.amazonq.focusC
...
-                triggerType: getCommandTriggerType(data),
-            })
-        })
-    })
-    Commands.register('aws.amazonq.updateContextCommandItems', () => {
-        controllerPublishers.processContextCommandUpdateMessage.publish()
-    })
+export function registerCommands() {
+    /**
+     * make these no-ops, since theres still callers that need to be deprecated
```
Hweinstock, line 41: are these commands still exposed to the user? Should we disable them until they are hooked up?

packages/core/src/codewhispererChat/commands/registerCommands.ts
```diff
@@ -37,73 +36,13 @@ export const focusAmazonQPanelKeybinding = Commands.declare('_aws.amazonq.focusC
     await focusAmazonQPanel.execute(placeholder, 'keybinding')
 })
 
-const getCommandTriggerType = (data: any): EditorContextCommandTriggerType => {
-    // data is undefined when commands triggered from keybinding or command palette. Currently no
-    // way to differentiate keybinding and command palette, so both interactions are recorded as keybinding
-    return data === undefined ? 'keybinding' : 'contextMenu'
-}
-
-export function registerCommands(controllerPublishers: ChatControllerMessagePublishers) {
```
Hweinstock, line 46: Just to verify, these still register properly on non-hybrid chat right?

jpinkney-aws: yeah this is only broken for hybrid chat, everything works fine for the regular main branch

packages/core/src/codewhispererChat/commands/registerCommands.ts
jpinkney-aws: They're only still enabled because other parts of the extension (like security scan) call them. In a future PR i'm going to fully deprecate them here and implement them in commands.ts but for now this just avoids spamming the logs with command not found

packages/core/src/codewhispererChat/commands/registerCommands.ts
jpinkney-aws: yeah I think what was happening was we were creating the commands here + the commands from commands.ts

Hweinstock (approved):

PR #7172, opened 2025-04-25
feat(amazonq): add setting to configure lsp log level (WIP)

Hweinstock:
## Problem
We don't pass log level to the LSP, meaning it always defaults to info. 

## Solution
- add a setting `amazonq.lsp.logLevel` to allow this to be configured. The allowed options are listed here: https://github.com/aws/language-server-runtimes/blob/eae85672c345d8adaf4c8cbd741260b8a59750c4/runtimes/runtimes/util/loggingUtil.ts#L4. 
- These means are are two settings:
   - `.trace.server` sends amazonq logs to a separate channel. 
   - `.lsp.logLevel` determines which level these should be logged at. 

## Future Work
- We will eventually want to expose these settings to the user. But this isn't needed until this is no longer hidden behind a feature flag. 
- The log messages in the `Amazon Q Logs` show `info` from the LSP even when they are `debug`, more investigation needed here. 

---

### DIFF
--- a/packages/amazonq/src/lsp/client.ts
+++ b/packages/amazonq/src/lsp/client.ts
@@ -22,7 +22,6 @@ import {
 } from '@aws/language-server-runtimes/protocol'
 import { AuthUtil, CodeWhispererSettings, getSelectedCustomization } from 'aws-core-vscode/codewhisperer'
 import {
-    Settings,
     oidcClientName,
     createServerOptions,
     globals,
@@ -36,6 +35,7 @@ import {
 } from 'aws-core-vscode/shared'
 import { activate } from './chat/activation'
 import { AmazonQResourcePaths } from './lspInstaller'
+import { getLspLogLevel } from './config'
 
 const localize = nls.loadMessageBundle()
 const logger = getLogger('amazonqLsp.lspClient')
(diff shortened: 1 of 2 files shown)


packages/amazonq/package.json
```diff
@@ -209,6 +209,23 @@
                     "items": {
                         "type": "string"
                     }
+                },
+                "amazonQ.lsp.traceChannel": {
```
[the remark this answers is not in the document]
nkomonen-amazon: Why isn't `amazonq.lsp.trace` defined here? Does it not need to be, or we want to hide from users?

docs/lsp.md
```diff
@@ -55,7 +53,7 @@ sequenceDiagram
     ```
 4. Uncomment the `__AMAZONQLSP_PATH` and `__AMAZONQLSP_UI` variables in the `amazonq/.vscode/launch.json` extension configuration
 5. Use the `Launch LSP with Debugging` configuration and set breakpoints in VSCode or the language server
-6. (Optional): Enable `"amazonq.trace.server": "on"` or `"amazonq.trace.server": "verbose"` in your VSCode settings to view detailed log messages sent to/from the language server. These log messages will show up in the "Amazon Q Language Server" output channel
+6. (Optional): Enable `"amazonq.lsp.trace": "on"` in your VSCode settings to view detailed log messages sent to/from the language server. These log messages will show up in the "Amazon Q Language Server" output channel. `"amazonq.lsp.logLevel"` can be used to configure the log level for the language server.
```
nkomonen-amazon, line 56: It may be helpful to create a dedicated section to enabling logs here instead of as a single bullet point. Then you can link to it. It will also be useful to link to it from the CONTRIBUTING.md loggin section (search `### Logging`) as a subsection

nkomonen-amazon (approved):

docs/lsp.md
Hweinstock: Since these settings are exposed to the user through the UI, I was treating those descriptions as the documentation. However, I agree that this is useful information to have in CONTRIBUTING.md so I've duplicated it there as where.

nkomonen-amazon (approved):

Hweinstock:
Going to look into mapping the VSC log level to the Flare one instead.

PR #7185, opened 2025-04-28
refactor(amazonq): Change Q LSP downloading message text

nkomonen-amazon:
## PROBLEM:

When installing the LSP for Q we get a message that says `"Installing 'Amazon Q' language server"`, but they want it to be slightly different (`"Updating Amazon Q plugin"`). The issue is the base class for the lsp installer class has a generic message that is not easy to customize.

## SOLUTION:

Allow an override downloading message to be defined in each specific LspInstaller implementation. This message is then routed in to the LspResolver (the thing that downloads the Lsp), and the message is displayed here. This override message will show when the download is happening. Everyone is happy.

---


jpinkney-aws (approved):
Makes sense, we can probably get rid of this when we deprecate the workspace context language server

Hweinstock (approved):

Issue #7187, opened 2025-04-28
unreliable test: pass document helloVue.vue as argument should first try determine by languageId then file extensions:


Hweinstock:
## Test Details

-   Name of test: pass document helloVue.vue as argument should first try determine by languageId then file extensions:
-   OS of failure: windows
-   VSCode version of failure (minimum/insider/stable): insiders
-   Link to failing run: https://github.com/aws/aws-toolkit-vscode/actions/runs/14718456397/job/41307940286?pr=7186
-   Link to failing test: https://github.com/aws/aws-toolkit-vscode/blob/a7732b96982adbbac691f2869db3136c159d80ed/packages/amazonq/test/unit/codewhisperer/util/runtimeLanguageContext.test.ts#L112-L116

## Log of Test Failure
```
1) runtimeLanguageContext
       test isLanguageSupported
         test isLanguageSupported with document as the argument
           pass document helloVue.vue as argument should first try determine by languageId then file extensions:
     Error: Test length exceeded max duration: 30 seconds
[No Pending UI Elements Found]
      at Timeout._onTimeout (D:\a\aws-toolkit-vscode\aws-toolkit-vscode\packages\core\src\test\setupUtil.ts:46:32)
      at listOnTimeout (node:internal/timers:581:17)
      at processTimers (node:internal/timers:519:7)
```

Hweinstock:
Potentially addressed by https://github.com/aws/aws-toolkit-vscode/pull/7259. This PR also outlines some potential next steps.

PR #7192, opened 2025-04-29
fix(amazonq): retry LSP token refresh on recoverable errors only. 

Hweinstock:
## Problem
Some users were having invalid bearer token exceptions when making chat requests through agentic chat. This highlights an issue in how we sync the token. 

Originally, we thought this was because of a race condition. We check the token every 10 seconds for changes, and if its expired, refresh it and send it to the language server. However, that still leaves a 10 second gap where the language server could use an expired token before we update it. 

However, we add a buffer of one minute to our `isExpired` check here: 
https://github.com/aws/aws-toolkit-vscode/blob/db673c9b74b36591bb5642b3da7d4bc7ae2afaf4/packages/core/src/auth/sso/model.ts#L160

Therefore, this causes the token to expire a minute early, meaning there is a full minute, where our checks should detect the expired token and refresh it both locally and on the language server. However, they don't because the checks aren't actually being made. 

This is because of how we handle token refresh errors. Note the following behavior:
- if refresh throws a recoverable error, we throw it. See [here](https://github.com/aws/aws-toolkit-vscode/blob/6dbb21e50e539c5973586714295e3ce066b030ef/packages/core/src/auth/auth.ts#L856-L878). 
- if refresh throws a non-recoverable error we invalidate the connection. see [here](https://github.com/aws/aws-toolkit-vscode/blob/6dbb21e50e539c5973586714295e3ce066b030ef/packages/core/src/auth/auth.ts#L893-L948). 

The current `refreshConnection` logic doesn't work with this because we continuously try to refresh until it throws an error. However, based on the behavior above, we do want to retry on refresh errors since that means the error is recoverable! Additionally, we don't want to retry when token refreshes result in an invalid connection because those are nonrecoverable errors. 

## Solution
- Refactor our token refreshes to log errors thrown and continue to retry (since these are implicitly recoverable errors).
- Avoid refreshing when the connection state is invalid (since this implies an unrecoverable error). 

## Notes
- Flare auth can't come soon enough. This behavior is not obvious to a consumer, and leads to bugs like this. 
- debugged w/ @nkomonen-amazon 

---

### DIFF
--- a/packages/amazonq/src/lsp/auth.ts
+++ b/packages/amazonq/src/lsp/auth.ts
@@ -66,18 +66,27 @@ export const notificationTypes = {
  * Facade over our VSCode Auth that does crud operations on the language server auth
  */
 export class AmazonQLspAuth {
-    constructor(private readonly client: LanguageClient) {}
+    #logErrorIfChanged = onceChanged((s) => getLogger('amazonqLsp').error(s))
+    constructor(
+        private readonly client: LanguageClient,
+        private readonly authUtil: AuthUtil = AuthUtil.instance
+    ) {}
 
     /**
      * @param force bypass memoization, and forcefully update the bearer token
      */
     async refreshConnection(force: boolean = false) {
-        const activeConnection = AuthUtil.instance.auth.activeConnection
-        if (activeConnection?.type === 'sso') {
+        const activeConnection = this.authUtil.auth.activeConnection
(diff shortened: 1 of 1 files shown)


Hweinstock:
This is a known flaky test: https://github.com/aws/aws-toolkit-vscode/issues/7187

jpinkney-aws (approved):

PR #7197, opened 2025-04-29
fix(amazonq): stop workspace indexing in extension

zixlin7:
## Problem
- we don't need workspace indexing on extension side anymore,
the `isVectorIndexEnabled` setting controls if we build vector index used for @workspace feature, since now chat is moving to language server, the extension side workspace lsp should be only responsible for building BM25 index for inline, and not responsible for vector indexing for chat.

## Solution
- stop vector indexing on extension
- next step is to tie workspace index setting to language server workspace indexing, will be in next PR

---

### DIFF
--- a/packages/amazonq/src/app/chat/activation.ts
+++ b/packages/amazonq/src/app/chat/activation.ts
@@ -18,7 +18,7 @@ export async function activate(context: ExtensionContext) {
         void amazonq.LspController.instance.trySetupLsp(context, {
             startUrl: AuthUtil.instance.startUrl,
             maxIndexSize: CodeWhispererSettings.instance.getMaxIndexSize(),
-            isVectorIndexEnabled: CodeWhispererSettings.instance.isLocalIndexEnabled(),
+            isVectorIndexEnabled: false,
         })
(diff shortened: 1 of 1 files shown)


packages/amazonq/src/app/chat/activation.ts
```diff
@@ -18,7 +18,7 @@ export async function activate(context: ExtensionContext) {
         void amazonq.LspController.instance.trySetupLsp(context, {
             startUrl: AuthUtil.instance.startUrl,
             maxIndexSize: CodeWhispererSettings.instance.getMaxIndexSize(),
-            isVectorIndexEnabled: CodeWhispererSettings.instance.isLocalIndexEnabled(),
+            isVectorIndexEnabled: false,
```
jpinkney-aws, line 21: I know we had a discussion on this in the internal team channel, but can you pull out in this PR what vector indexing is, what features it enables, etc. Would be super useful just in case we have to re-reference it

jpinkney-aws (approved):

leigaol (approved):

PR #7203, opened 2025-04-29
fix(amazonq): dev profile selection breaks if certain regions aren't …

tomcat323:
## Problem
Customer reported when a certain region is blocked(i.e. by cooperate network), it causes the Q developer profile list to break, and the customer can no longer use Q. 

## Solution
Don't throw immediately when a region is unavailable. So that the user still have access to working regions. 

Only throw if all regions are failing. 

## Testing
Shared the VSIX with customer and confirmed that issue is resolved. 

---

### DIFF
--- a/packages/core/src/codewhisperer/region/regionProfileManager.ts
+++ b/packages/core/src/codewhisperer/region/regionProfileManager.ts
@@ -172,8 +172,8 @@ export class RegionProfileManager {
             }
         }
 
-        if (availableProfiles.length === 0 && failedRegions.length > 0) {
-            // Only throw error if all regions fail
+        // Only throw error if all regions fail
+        if (failedRegions.length === endpoints.size) {
             throw new Error(`Failed to list profiles for all regions: ${failedRegions.join(', ')}`)
         }
 
--- a/packages/core/src/codewhisperer/region/regionProfileManager.ts
+++ b/packages/core/src/codewhisperer/region/regionProfileManager.ts
@@ -138,6 +138,8 @@ export class RegionProfileManager {
             return []
         }
         const availableProfiles: RegionProfile[] = []
+        const failedRegions: string[] = []
+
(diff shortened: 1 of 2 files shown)


packages/core/src/codewhisperer/region/regionProfileManager.ts
```diff
@@ -162,13 +164,17 @@ export class RegionProfileManager {
...
-                RegionProfileManager.logger.error(`failed to listRegionProfile: ${logMsg}`)
-                throw e
+                RegionProfileManager.logger.error(`Failed to list profiles for region ${region}: ${logMsg}`)
+                failedRegions.push(region)
             }
+        }
 
-            RegionProfileManager.logger.info(`available amazonq profiles: ${availableProfiles.length}`)
+        if (availableProfiles.length === 0 && failedRegions.length > 0) {
+            // Only throw error if all regions fail
```
opieter-aws, line 176: `failedRegions.length > 0` is not equivalent to all regions failing, right?

Will-ShaoHua: maybe `availableProfiles.length === 0` itself is sufficient

tomcat323: A region either goes into `availableProfiles` or `failedRegions`, so this means no region succeeded & some failed, which is essentially all regions failing.

tomcat323: Ideally it could have just been `availableProfiles.length === 0`, I added the `failedRegions ` check just in case there's no profiles available at all.

tomcat323: > I would expect a check like `if (failedRegions.length === availableRegions.length)`

I see, thanks, updated!

opieter-aws (approved):

jpinkney-aws (approved):

PR #7205, opened 2025-04-30
feat(amazonq): allow child processes to define custom thresholds

Hweinstock:
## Problem
Some processes use significantly more cpu/memory than others, so we shouldn't hold them to the same static threshold. 
As an example, the language servers are consistently using more memory than the 100 MB threshold. 

Based on some experimentation with different sized workspaces, I've noticed that the process running agentic chat uses roughly 80MB-150MB of memory, and the workspace indexing uses 550 - 700 MB memory.  

## Solution
- Allow each child process to set a warning threshold, with a default of 100MB of memory. 
- Set the Q LSP threshold to 200 MB, and workspace indexing to 800 MB. 
- This should help to allow these logs to point to real concerns, rather than be noise. 

---

### DIFF
--- a/packages/amazonq/src/lsp/client.ts
+++ b/packages/amazonq/src/lsp/client.ts
@@ -33,6 +33,7 @@ import {
     undefinedIfEmpty,
     getOptOutPreference,
 } from 'aws-core-vscode/shared'
+import { processUtils } from 'aws-core-vscode/shared'
 import { activate } from './chat/activation'
 import { AmazonQResourcePaths } from './lspInstaller'
 import { ConfigSection, isValidConfigSection, toAmazonQLSPLogLevel } from './config'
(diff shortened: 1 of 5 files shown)


packages/amazonq/src/lsp/client.ts
```diff
@@ -80,11 +81,13 @@ export async function startLanguageServer(
         executable = [resourcePaths.node]
     }
 
+    const memoryWarnThreshold = 200 * processUtils.oneMB
     const serverOptions = createServerOptions({
         encryptionKey,
         executable: executable,
         serverModule,
         execArgv: argv,
+        warnThresholds: { memory: memoryWarnThreshold },
```
jpinkney-aws, line 90: Is the value here only 200MB because the workspace context part of the language-servers repo isn't doing the indexing? I guess once we fully enable inline through Flare we'd probably have to re-adjust these numbers?

jpinkney-aws (approved):
Now we get the benefit of the extension host not messing up our memory/cpu numbers in flare 🎉

packages/amazonq/src/lsp/client.ts
Hweinstock: Followed up with local indexing team, and it looks like it was disabled in my testing since this requires pulling the indexing server. After enabling its using ~700-800MB so I'll bump this to 1GB to be safe.

PR #7212, opened 2025-04-30
fix(amazonq): enable agentic chat in AL2

leigaol:
## Problem
LSP depends on node 18 which depends on GLIBC >=2.28. 
Amzn al2 instances do not have GLIBC >=2.28. 

## Solution
Use the patch of glibc if it is available.

---

### DIFF
--- a/packages/amazonq/src/lsp/client.ts
+++ b/packages/amazonq/src/lsp/client.ts
@@ -6,6 +6,7 @@
 import vscode, { env, version } from 'vscode'
 import * as nls from 'vscode-nls'
 import * as crypto from 'crypto'
+import * as path from 'path'
 import { LanguageClient, LanguageClientOptions, RequestType, State } from 'vscode-languageclient'
 import { InlineCompletionManager } from '../app/inline/completion'
 import { AmazonQLspAuth, encryptionKey, notificationTypes } from './auth'
@@ -32,13 +33,20 @@ import {
     getLogger,
     undefinedIfEmpty,
     getOptOutPreference,
+    isAmazonInternalOs,
+    fs,
 } from 'aws-core-vscode/shared'
 import { activate } from './chat/activation'
 import { AmazonQResourcePaths } from './lspInstaller'
 import { ConfigSection, isValidConfigSection, toAmazonQLSPLogLevel } from './config'
+import { chmodSync } from 'fs'
 
 const localize = nls.loadMessageBundle()
 const logger = getLogger('amazonqLsp.lspClient')
+
+export async function hasGlibcPatch(): Promise<boolean> {
+    return await fs.exists('/opt/vsc-sysroot/lib64/ld-linux-x86-64.so.2')
+}
 
 export async function startLanguageServer(
     extensionContext: vscode.ExtensionContext,
@@ -55,17 +63,31 @@ export async function startLanguageServer(
         '--pre-init-encryption',
         '--set-credentials-encryption-key',
     ]
+
+    const documentSelector = [{ scheme: 'file', language: '*' }]
+
+    const clientId = 'amazonq'
+    const traceServerEnabled = Settings.instance.isSet(`${clientId}.trace.server`)
+
+    // apply the GLIBC 2.28 path to node js runtime binary
+    if (isAmazonInternalOs() && (await hasGlibcPatch())) {
+        const nodeWrapper = `
+        #! /bin/bash
+        exec /opt/vsc-sysroot/lib64/ld-linux-x86-64.so.2 --library-path /opt/vsc-sysroot/lib64 ${resourcePaths.node} "$@"
+        `
+        const nodeWrapperPath = path.join(path.dirname(resourcePaths.node), 'node-wrapper')
+        await fs.writeFile(nodeWrapperPath, nodeWrapper)
+        chmodSync(nodeWrapperPath, 0o755)
+        resourcePaths.node = nodeWrapperPath
+        getLogger('amazonqLsp').info(`Patched node runtime with GLIBC to ${resourcePaths.node}`)
+    }
+
     const serverOptions = createServerOptions({
         encryptionKey,
         executable: resourcePaths.node,
         serverModule,
         execArgv: argv,
     })
-
-    const documentSelector = [{ scheme: 'file', language: '*' }]
-
-    const clientId = 'amazonq'
-    const traceServerEnabled = Settings.instance.isSet(`${clientId}.trace.server`)
 
     await validateNodeExe(resourcePaths.node, resourcePaths.lsp, argv, logger)
 
--- a/packages/amazonq/src/lsp/client.ts
+++ b/packages/amazonq/src/lsp/client.ts
@@ -39,7 +39,7 @@ import {
 import { activate } from './chat/activation'
 import { AmazonQResourcePaths } from './lspInstaller'
 import { ConfigSection, isValidConfigSection, toAmazonQLSPLogLevel } from './config'
-import { chmodSync } from 'fs'
+import { chmodSync } from 'fs' // eslint-disable-line no-restricted-imports
 
 const localize = nls.loadMessageBundle()
 const logger = getLogger('amazonqLsp.lspClient')
--- a/packages/core/src/shared/lsp/utils/platform.ts
+++ b/packages/core/src/shared/lsp/utils/platform.ts
@@ -26,23 +26,25 @@ function getEncryptionInit(key: Buffer): string {
     return JSON.stringify(request) + '\n'
 }
 
+// use dynamic linking to let node binary
+// pick up GLIBC >= 2.28
 const al2Config = {
     path: '/opt/vsc-sysroot/lib64/ld-linux-x86-64.so.2',
     args: ['--library-path', '/opt/vsc-sysroot/lib64'],
 }
-
+function createNodeProcess(nodePath: string, args: string[]) {
+    return new ChildProcess(
+        isAmazonInternalOs() ? al2Config.path : nodePath,
+        isAmazonInternalOs() ? [...al2Config.args, nodePath, ...args] : [...args],
+        { logging: 'no' }
+    )
+}
 /**
  * Checks that we can actually run the `node` executable and execute code with it.
  */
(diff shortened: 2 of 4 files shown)


Hweinstock:
verified the build works for me in cloud desktop as well!

packages/core/src/shared/lsp/utils/platform.ts
```diff
@@ -26,12 +26,23 @@ function getEncryptionInit(key: Buffer): string {
...
+    args: ['--library-path', '/opt/vsc-sysroot/lib64'],
+}
+
 /**
  * Checks that we can actually run the `node` executable and execute code with it.
  */
 export async function validateNodeExe(nodePath: string, lsp: string, args: string[], logger: Logger) {
     // Check that we can start `node` by itself.
-    const proc = new ChildProcess(nodePath, ['-e', 'console.log("ok " + process.version)'], { logging: 'no' })
+    const proc = new ChildProcess(
```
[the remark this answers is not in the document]
Hweinstock: +1, also worth adding a comment explaining this hack since @jpinkney-aws, you and I are likely the only ones with context here about why this works.

packages/core/src/shared/lsp/utils/platform.ts
```diff
@@ -26,12 +26,25 @@ function getEncryptionInit(key: Buffer): string {
...
+        { logging: 'no' }
+    )
+}
 /**
  * Checks that we can actually run the `node` executable and execute code with it.
  */
 export async function validateNodeExe(nodePath: string, lsp: string, args: string[], logger: Logger) {
     // Check that we can start `node` by itself.
-    const proc = new ChildProcess(nodePath, ['-e', 'console.log("ok " + process.version)'], { logging: 'no' })
+    const proc = createNodeProcess(nodePath, ['-e', 'console.log("ok")'])
```
jpinkney-aws, line 47: any reason why we had to drop the process.version? it might help give us a hint on if the correct node version was picked up

leigaol: my miss!

Hweinstock:
I've verified with the latest commit, I can still execute tools and use agentic chat in my dev desktop!

jpinkney-aws:
Mine worked on my old dev desktop but on the new one it fails because `/opt/vsc-sysroot/lib64/ld-linux-x86-64.so.2` isn't defined

packages/core/src/shared/lsp/utils/platform.ts
```diff
@@ -26,12 +26,25 @@ function getEncryptionInit(key: Buffer): string {
...
+        { logging: 'no' }
+    )
+}
 /**
  * Checks that we can actually run the `node` executable and execute code with it.
  */
 export async function validateNodeExe(nodePath: string, lsp: string, args: string[], logger: Logger) {
     // Check that we can start `node` by itself.
-    const proc = new ChildProcess(nodePath, ['-e', 'console.log("ok " + process.version)'], { logging: 'no' })
+    const proc = createNodeProcess(nodePath, ['-e', 'console.log("ok " + process.version)'])
```
justinmk3, line 47: shouldn't need any changes in `validateNodeExe`. It should just get the actual `nodepath` and `args` that will be used.

the caller to `validateNodeExe` should decide what those are before callling `validateNodeExe`.

packages/core/src/shared/lsp/utils/platform.ts
```diff
@@ -89,8 +103,7 @@ export function createServerOptions({
         if (isDebugInstance()) {
             args.unshift('--inspect=6080')
         }
-        const lspProcess = new ChildProcess(executable, args)
-
+        const lspProcess = createNodeProcess(executable, args)
```
justinmk3, line 106: `createServerOptions` also shouldn't need changes. instead, the *caller* to `createServerOptions` should have already decided the right `nodePath` (which is called `executable`) here, after checking `validateNodeExe`

justinmk3:
here are the 2 places that decide the path to node+args:

https://github.com/aws/aws-toolkit-vscode/blob/8b7106bb7e682a93a5856a499882bd6fa7a15b75/packages/amazonq/src/lsp/client.ts#L49-L51

https://github.com/aws/aws-toolkit-vscode/blob/8b7106bb7e682a93a5856a499882bd6fa7a15b75/packages/core/src/amazonq/lsp/lspClient.ts#L254-L262

each of those then passes `nodePath` (+args) to various functions.

packages/core/src/shared/lsp/utils/platform.ts
leigaol: fixed in latest commit

opieter-aws (changes requested):
The linter is failing on the fs import

packages/amazonq/src/lsp/client.ts
```diff
@@ -32,14 +33,21 @@ import {
     getLogger,
     undefinedIfEmpty,
     getOptOutPreference,
+    isAmazonInternalOs,
+    fs,
 } from 'aws-core-vscode/shared'
 import { activate } from './chat/activation'
 import { AmazonQResourcePaths } from './lspInstaller'
 import { ConfigSection, isValidConfigSection, toAmazonQLSPLogLevel } from './config'
+import { chmodSync } from 'fs' // eslint-disable-line no-restricted-imports
```
Hweinstock, line 42: We should use the wrapper [here](https://github.com/aws/aws-toolkit-vscode/blob/457efa4efff9cdf0d2c82d165063bd6b4c64cd6b/packages/core/src/shared/fs/fs.ts#L347). It doesn't do anything on browsers, but has the same behavior in a node process.

packages/amazonq/src/lsp/client.ts
```diff
@@ -55,18 +63,32 @@ export async function startLanguageServer(
...
+
+    const clientId = 'amazonq'
+    const traceServerEnabled = Settings.instance.isSet(`${clientId}.trace.server`)
+
+    // apply the GLIBC 2.28 path to node js runtime binary
+    if (isAmazonInternalOs() && (await hasGlibcPatch())) {
+        const nodeWrapper = `
+        #! /bin/bash
+        exec /opt/vsc-sysroot/lib64/ld-linux-x86-64.so.2 --library-path /opt/vsc-sysroot/lib64 ${resourcePaths.node} "$@"
+        `
```
justinmk3, line 77: Could avoid this wrapper script by changing the first arg to `validateNodeExe()` to accept an array. Here's a patch (untested but should work):


```diff
diff --git a/packages/core/src/shared/lsp/utils/platform.ts b/packages/core/src/shared/lsp/utils/platform.ts
index 5d96ef496f4f..2a4fb6be8e2d 100644
--- a/packages/core/src/shared/lsp/utils/platform.ts
+++ b/packages/core/src/shared/lsp/utils/platform.ts
@@ -28,10 +28,16 @@ function getEncryptionInit(key: Buffer): string {
 
 /**
  * Checks that we can actually run the `node` executable and execute code with it.
+ *
+ * @param nodePath "node" or "/opt/foo … node".
+ * @param lsp .js bundle
+ * @param args Arguments _to the .js bundle_
+ * @param logger Logger
  */
-export async function validateNodeExe(nodePath: string, lsp: string, args: string[], logger: Logger) {
+export async function validateNodeExe(nodePath: string[], lsp: string, args: string[], logger: Logger) {
+    const arg0 = nodePath[0]
     // Check that we can start `node` by itself.
-    const proc = new ChildProcess(nodePath, ['-e', 'console.log("ok " + process.version)'], { logging: 'no' })
+    const proc = new ChildProcess(arg0, [...nodePath.slice(2), '-e', 'console.log("ok " + process.version)'], { logging: 'no' })
     const r = await proc.run()
     const ok = r.exitCode === 0 && r.stdout.includes('ok')
     if (!ok) {
@@ -41,7 +47,7 @@ export async function validateNodeExe(nodePath: string, lsp: string, args: strin
     }
 
     // Check that we can start `node …/lsp.js --stdio …`.
-    const lspProc = new ChildProcess(nodePath, [lsp, ...args], { logging: 'no' })
+    const lspProc = new ChildProcess(arg0, [...nodePath.slice(2), lsp, ...args], { logging: 'no' })
     try {
         // Start asynchronously (it never stops; we need to stop it below).
         lspProc.run().catch((e) => logger.error('failed to run: %s', lspProc.toString(false, true)))
@@ -80,16 +86,17 @@ export function createServerOptions({
     execArgv,
 }: {
     encryptionKey: Buffer
-    executable: string
+    executable: string[] // May be "node" or "/opt/foo … node".
     serverModule: string
     execArgv: string[]
 }) {
     return async () => {
-        const args = [serverModule, ...execArgv]
+        const arg0 = executable[0]
+        const args = [...executable.slice(1), serverModule, ...execArgv]
         if (isDebugInstance()) {
             args.unshift('--inspect=6080')
         }
-        const lspProcess = new ChildProcess(executable, args)
+        const lspProcess = new ChildProcess(arg0, args)
 
         // this is a long running process, awaiting it will never resolve
         void lspProcess.run()
```

leigaol: fixed in latest.

justinmk3 (approved):

opieter-aws (approved):

Hweinstock (approved):

packages/amazonq/src/lsp/client.ts
```diff
@@ -40,6 +42,10 @@ import { ConfigSection, isValidConfigSection, toAmazonQLSPLogLevel } from './con
 const localize = nls.loadMessageBundle()
 const logger = getLogger('amazonqLsp.lspClient')
 
+export async function hasGlibcPatch(): Promise<boolean> {
+    return await fs.exists('/opt/vsc-sysroot/lib64/ld-linux-x86-64.so.2')
```
leigaol, line 46: follow up: Check if arm64 has this issue or not.

murkvin (commented):
It seems that this PR takes a dependency on some resources that we drop onto hosts as a component of another project. This is fine, but we should connect to discuss and formalize the interface. 

For example, we already vend a static, architecture-independent path to the linker as part of our public interface. Using this would make your code more robust and abstract away platform concerns. 

I don't expect to move these files around on people's machines, but I'd rather that doing so didn't break this plugin on AL2.

PR #7231, opened 2025-05-06
fix(amazonq): Bug fixes for LSP auth on agentic mode

opieter-aws:
## Problem
Agentic chat introduced code changes that break auth on LSP

## Solution
Fixes for the auth flow:
* Move instantiation of `AuthUtil` into `startLanguageServer` sinc ethe function uses several AuthUtil calls, which fail if it is not instantiated yet
* Remove the dependency of LSP on glibc patch, since auth doesn't work if LSP is not started
* Add URL opening capability to the new `ShowDocumentRequest` handler, and remove the old handler from `client.ts`.
* Fix bug where session isn't restored upon start of extension when SSO is expired.

---

### DIFF
--- a/packages/amazonq/src/extensionNode.ts
+++ b/packages/amazonq/src/extensionNode.ts
@@ -19,7 +19,6 @@ import api from './api'
 import { activate as activateCWChat } from './app/chat/activation'
 import { activate as activateInlineChat } from './inlineChat/activation'
 import { beta } from 'aws-core-vscode/dev'
-import * as amazonq from 'aws-core-vscode/amazonq'
 import { activate as activateNotifications, NotificationsController } from 'aws-core-vscode/notifications'
 import { AuthUtil } from 'aws-core-vscode/codewhisperer'
 import { telemetry, AuthUserState } from 'aws-core-vscode/telemetry'
@@ -67,8 +66,6 @@ async function activateAmazonQNode(context: vscode.ExtensionContext) {
     activateAgents()
     await activateTransformationHub(extContext as ExtContext)
     activateInlineChat(context)
-
-    context.subscriptions.push(amazonq.focusAmazonQPanel.register(), amazonq.focusAmazonQPanelKeybinding.register())
 
     const authProvider = new CommonAuthViewProvider(
         context,
--- a/packages/amazonq/src/lsp/chat/messages.ts
+++ b/packages/amazonq/src/lsp/chat/messages.ts
@@ -58,7 +58,7 @@ import { Disposable, LanguageClient, Position, TextDocumentIdentifier } from 'vs
 import * as jose from 'jose'
 import { AmazonQChatViewProvider } from './webviewProvider'
 import { AuthUtil, ReferenceLogViewProvider } from 'aws-core-vscode/codewhisperer'
-import { amazonQDiffScheme, AmazonQPromptSettings, messages, openUrl } from 'aws-core-vscode/shared'
+import { amazonQDiffScheme, AmazonQPromptSettings, getLogger, messages, openUrl } from 'aws-core-vscode/shared'
 import {
     DefaultAmazonQAppInitContext,
     messageDispatcher,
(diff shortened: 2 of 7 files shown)


packages/amazonq/src/extensionNode.ts
```diff
@@ -68,8 +67,6 @@ async function activateAmazonQNode(context: vscode.ExtensionContext) {
     await activateTransformationHub(extContext as ExtContext)
     activateInlineChat(context)
 
-    context.subscriptions.push(amazonq.focusAmazonQPanel.register(), amazonq.focusAmazonQPanelKeybinding.register())
-
```
opieter-aws, line 72: Now part of https://github.com/aws/aws-toolkit-vscode/blob/f1f5a362a902959dd8366c2c015c04e66297fc00/packages/amazonq/src/app/chat/activation.ts#L31

packages/amazonq/src/lsp/chat/messages.ts
```diff
@@ -428,6 +428,14 @@ export function registerMessageListeners(
...
             try {
                 const uri = vscode.Uri.parse(params.uri)
+                if (uri.scheme.startsWith('http')) {
+                    try {
+                        await openUrl(vscode.Uri.parse(params.uri))
+                        return params
+                    } catch (err: any) {
+                        getLogger().error(`Failed to open http from LSP: error: %s`, err)
+                    }
+                }
```
opieter-aws, line 438: This functionality was added to enable LSP to open documents in agentic chat. We were relying on the same request type to open the auth URL we receive from LSP in the login flow

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -88,6 +88,9 @@ export class AuthUtil implements IAuthProvider {
 
     async restore() {
         await this.session.restore()
+        if (!this.isConnected()) {
+            await this.refreshState()
+        }
```
opieter-aws, line 93: This ensures that a session is restored successfully upon start of the extension with expired SSO. If SSO is expired upon start of extension, the LSP throws an error. Since there is no state change in that case (`notConnected` > `notConnected`), we can't rely on the state change handler to trigger a login flow

packages/amazonq/src/lsp/client.ts
```diff
@@ -144,9 +140,6 @@ export async function startLanguageServer(
                         notifications: true,
                         showSaveFileDialog: true,
                     },
-                    q: {
```
jpinkney-aws, line 147: I don't think this can be dropped

packages/amazonq/src/lsp/client.ts
```diff
@@ -183,15 +177,6 @@ export async function startLanguageServer(
         }
     })
 
-    client.onRequest<ShowDocumentResult, Error>(ShowDocumentRequest.method, async (params: ShowDocumentParams) => {
```
jpinkney-aws, line 186: it might make sense to have this request handled in here instead of inside of `chat/messages.ts` since most of those are just tied to chat?

packages/amazonq/src/lsp/client.ts
```diff
@@ -266,9 +251,7 @@ export async function startLanguageServer(
         )
     }
 
-    if (Experiments.instance.get('amazonqChatLSP', false)) {
-        activate(client, encryptionKey, resourcePaths.ui)
-    }
+    await activate(client, encryptionKey, resourcePaths.ui)
```
jpinkney-aws, line 254: We should probably rename this activate to activateChat just to make it clearer

packages/amazonq/src/lsp/client.ts
```diff
@@ -131,7 +127,7 @@ export async function startLanguageServer(
                     name: env.appName,
                     version: version,
                     extension: {
-                        name: 'AmazonQ-For-VSCode',
+                        name: clientName,
```
jpinkney-aws, line 130: I think this might have to stay as `'AmazonQ-For-VSCode'`, it has some internal meaning in flare that I haven't figured out yet

packages/amazonq/src/lsp/client.ts
opieter-aws: The merge with mainline made it duplicate, see line 140

jpinkney-aws (approved):

PR #7233, opened 2025-05-06
fix(amazonq): respect suggestions enabled setting for auto trigger

jpinkney-aws:
## Problem
the isSuggestionsEnable setting isn't respected

## Solution
if someone disabled suggestions via a setting or from the status bar then don't return anything for automatic triggers

---

### DIFF
--- a/packages/amazonq/src/app/inline/completion.ts
+++ b/packages/amazonq/src/app/inline/completion.ts
@@ -16,6 +16,7 @@ import {
     Disposable,
     window,
     TextEditor,
+    InlineCompletionTriggerKind,
 } from 'vscode'
 import { LanguageClient } from 'vscode-languageclient'
 import {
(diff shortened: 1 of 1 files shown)


opieter-aws (approved):

packages/amazonq/src/app/inline/completion.ts
```diff
@@ -180,6 +182,12 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
         token: CancellationToken
     ): Promise<InlineCompletionItem[] | InlineCompletionList> {
         if (this.isNewSession) {
+            const isAutoTrigger = context.triggerKind === InlineCompletionTriggerKind.Automatic
+            if (isAutoTrigger && !CodeSuggestionsState.instance.isSuggestionsEnabled()) {
```
Hweinstock, line 186: Is this a bug on master as well?

Hweinstock: I see, wanted to double check this was hidden behind feature flag, but that makes sense.

PR #7234, opened 2025-05-06
telemetry(amazonq): implement codewhisperer_clientComponentLatency

jpinkney-aws:
## Problem
the only metric it looks like we're missing for inline on the vscode side is `codewhisperer_clientComponentLatency`

## Solution
codewhisperer_clientComponentLatency uses a very similar implementation as before the only differences are:
1. codewhispererCredentialFetchingLatency is no longer relevant because the token is always injected into the language server and it doesn't need to build the client on demand like before.
    - This causes the preprocessing latency to decrease, because that used to contain the time it takes to fetch the credentials
2. postProcessing latency is way lower because once we get the result vscode instantly displays it -- we no longer have control of that

example metric now:
```
2025-05-06 11:53:59.858 [debug] telemetry: codewhisperer_clientComponentLatency {
  Metadata: {
    codewhispererAllCompletionsLatency: '792.7122090000048',
    codewhispererCompletionType: 'Line',
    codewhispererCredentialFetchingLatency: '0',
    codewhispererCustomizationArn: 'arn:aws:codewhisperer:us-east-1:12345678910:customization/AAAAAAAAAA',
    codewhispererEndToEndLatency: '792.682249999998',
    codewhispererFirstCompletionLatency: '792.6440000000002',
    codewhispererLanguage: 'java',
    codewhispererPostprocessingLatency: '0.019500000002153683',
    codewhispererPreprocessingLatency: '0.007166999996115919',
    codewhispererRequestId: 'XXXXXXXXXXXXXXXXXXXXXXXXXXX',
    codewhispererTriggerType: 'AutoTrigger',
    credentialStartUrl: 'https://XXXXX.XXXXX.com/start',
    awsAccount: 'not-set',
    awsRegion: 'us-east-1'
  },
  Value: 1,
  Unit: 'None',
  Passive: true
}
```

---

### DIFF
--- a/packages/amazonq/src/app/inline/recommendationService.ts
+++ b/packages/amazonq/src/app/inline/recommendationService.ts
@@ -13,6 +13,7 @@ import { LanguageClient } from 'vscode-languageclient'
 import { SessionManager } from './sessionManager'
 import { InlineGeneratingMessage } from './inlineGeneratingMessage'
 import { CodeWhispererStatusBarManager } from 'aws-core-vscode/codewhisperer'
+import { TelemetryHelper } from './telemetryHelper'
 
 export class RecommendationService {
     constructor(
@@ -36,6 +37,9 @@ export class RecommendationService {
         }
         const requestStartTime = Date.now()
         const statusBar = CodeWhispererStatusBarManager.instance
+        TelemetryHelper.instance.setInvokeSuggestionStartTime()
+        TelemetryHelper.instance.setPreprocessEndTime()
+        TelemetryHelper.instance.setSdkApiCallStartTime()
 
         try {
             // Show UI indicators that we are generating suggestions
@@ -48,6 +52,14 @@ export class RecommendationService {
                 request,
                 token
             )
+
+            // Set telemetry data for the first response
+            TelemetryHelper.instance.setSdkApiCallEndTime()
+            TelemetryHelper.instance.setSessionId(firstResult.sessionId)
+            if (firstResult.items.length > 0) {
+                TelemetryHelper.instance.setFirstResponseRequestId(firstResult.items[0].itemId)
+            }
+            TelemetryHelper.instance.setFirstSuggestionShowTime()
 
             const firstCompletionDisplayLatency = Date.now() - requestStartTime
             this.sessionManager.startSession(
@@ -64,6 +76,10 @@ export class RecommendationService {
                 })
             } else {
                 this.sessionManager.closeSession()
+
+                // No more results to fetch, mark pagination as complete
+                TelemetryHelper.instance.setAllPaginationEndTime()
+                TelemetryHelper.instance.tryRecordClientComponentLatency()
             }
         } finally {
             // Remove all UI indicators of message generation since we are done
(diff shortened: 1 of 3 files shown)


opieter-aws (approved):
Looks like some unit tests need updating with these changes

packages/amazonq/src/app/inline/recommendationService.ts
```diff
@@ -30,6 +31,9 @@ export class RecommendationService {
             context,
         }
         const requestStartTime = Date.now()
+        TelemetryHelper.instance.setInvokeSuggestionStartTime()
+        TelemetryHelper.instance.setPreprocessEndTime()
+        TelemetryHelper.instance.setSdkApiCallStartTime()
 
         // Handle first request
         const firstResult: InlineCompletionListWithReferences = await languageClient.sendRequest(
```
Hweinstock, line 39: unrelated to work here, but do we need to encrypt this request?

packages/amazonq/src/app/inline/recommendationService.ts
```diff
@@ -38,6 +42,11 @@ export class RecommendationService {
             token
         )
 
+        // Set telemetry data for the first response
+        TelemetryHelper.instance.setSdkApiCallEndTime()
+        TelemetryHelper.instance.setFirstResponseRequestId(firstResult.sessionId)
```
Hweinstock, line 47: was this requestId intended to be for the backend request? Based on my understanding from [here](https://github.com/aws/language-server-runtimes/blob/d5e4dc4e924155b8a0f23f6d895d448f45b2e36d/types/inlineCompletionWithReferences.ts#L31-L37) it sounds like sessionId is something else?

jpinkney-aws: Oh true, I might have to update it to be the first recommendation id

packages/amazonq/src/app/inline/recommendationService.ts
jpinkney-aws: 🤔 good question, I guess I probably should

jpinkney-aws: I'm asking internally for more clarification, but judging from the types it looks like inline can't be encrypted

packages/amazonq/src/app/inline/recommendationService.ts
jpinkney-aws: AFAICT itemId might be the one we're looking for?

Hweinstock: I wouldn't be surprised if this information isn't part of the response since I haven't seen the backend requestID included in any Flare responses so far. This might mean some of this telemetry will need to live in flare or we sacrifice having requestId for now. 

This issue of Flare having some telemetry fields, and the IDE having others is becoming increasingly problematic but there's no time/bandwidth for someone to look into it atm. :(

jpinkney-aws: Almost all of the telemetry for amazon q now lives on the flare side and will have access to that request id, I think this is the rare exception because it's purely tracking client side latency

PR #7236, opened 2025-05-06
fix(amazonq): Fix showSsoPrompt unit tests

opieter-aws:
## Problem
`showConnectionPrompt` unit tests were failing

## Solution
* Ensure DeviceFlow is used for auth
* Stub login
* Assert whether the login was called with builder ID instead of if the log line was emitted

---


packages/amazonq/test/unit/codewhisperer/util/showSsoPrompt.test.ts
```diff
@@ -36,12 +42,13 @@ describe('showConnectionPrompt', function () {
...
+    it('connectToAwsBuilderId calls AuthUtil login with builderIdStartUrl', async function () {
         sinon.stub(vscode.commands, 'executeCommand')
+        const loginStub = sinon.stub(AuthUtil.instance, 'login').resolves()
 
         await awsIdSignIn()
 
-        const loggedEntries = getTestLogger().getLoggedEntries()
-        assert.ok(loggedEntries.find((entry) => entry === 'selected AWS ID sign in'))
+        assert.strictEqual(loginStub.called, true)
+        assert.strictEqual(loginStub.firstCall.args[0], constants.builderIdStartUrl)
```
justinmk3, line 52: why can't we test the logged messages now ? can an equivalent log message be added in the relevant place in the new module ? that allows testing the real codepath (or something closer to it), instead of stubs.

opieter-aws: The `getTestLogger().getLoggedEntries()` call was failing to pick up the logger. The issue with only testing a log line is that the test won't fail if the functionality is removed or changes (the login call with AWS ID is what we care about asserting here, not a logged string). I spend time to dive deeper into it because it's a secondary assertion IMO

justinmk3: > The issue with only testing a log line is that the test won't fail if the functionality is removed or changes (the login call with AWS ID is what we care about asserting here, not a logged string)

Sure, but that applies to any kind of telemetry (aka logs): the telemetry is wrong if logic changes without updating the telemetry. And a similar situation applies when stubs/mocks are used: asserting that a method was called doesn't actually test the result, and can easily drift (also, asserting stubs leads to brittle tests).

justinmk3 (approved):

PR #7240, opened 2025-05-06
fix(logging): misleading "settings failed" message

justinmk3:
## Problem

Misleading log message when a setting returns `null`, even if the caller specified `defaultValue`:

    [error] Settings (amazonQ): failed to read "workspaceIndexCacheDirPath":
    Unexpected type cast: got object, expected primitive String

`cast()` considers `null` to be "object" type, which is misleading. If the caller passed a `defaultValue`, they have intentionally opted-in to ignoring this situation, and *don't* want this kind of log message.

## Solution

If a setting returns `null|undefined`, skip `cast()` and return `defaultValue` immediately. Don't log an error message.

---

### DIFF
--- a/packages/core/src/shared/settings.ts
+++ b/packages/core/src/shared/settings.ts
@@ -389,6 +389,10 @@ function createSettingsClass<T extends TypeDescriptor>(section: string, descript
         public _getOrThrow<K extends keyof Inner>(key: K & string, defaultValue?: Inner[K]) {
             const value = this.#config.get(key, defaultValue)
 
+            if (defaultValue !== undefined && (value === undefined || value === null)) {
+                return defaultValue
+            }
+
             return cast<Inner[K]>(value, descriptor[key])
         }
 


packages/core/src/shared/settings.ts
```diff
@@ -389,6 +389,10 @@ function createSettingsClass<T extends TypeDescriptor>(section: string, descript
         public _getOrThrow<K extends keyof Inner>(key: K & string, defaultValue?: Inner[K]) {
             const value = this.#config.get(key, defaultValue)
 
+            if (defaultValue !== undefined && (value === undefined || value === null)) {
+                return defaultValue
```
justinmk3, line 393: test coverage: https://github.com/aws/aws-toolkit-vscode/blob/fdaaae1a5ba36686d400ca612c15d5a6cb0fdf89/packages/core/src/test/shared/settings.test.ts#L98

jpinkney-aws (approved):

Hweinstock (approved):
Thanks for looking into this!

PR #7249, opened 2025-05-07
fix(amazonq): Add AuthUtil test suite and fix unit tests

opieter-aws:
## Problem
Unit tests for `amazonq_backed` and `regionProfileManager` are broken or disabled, because they are not working with the LSP auth setup

## Solution
* Create a new test util to setup AuthUtil for tests, stubbing the LSP client methods
* Rework the unit test where necessary to use the AuthUtil on Flare identity server
* Move `amazonq_backed` to the `amazonq` package, since the AuthUtil won't instantiate correctly if the unit test runs under `toolkits` test suite

---


opieter-aws:
Web is not supported, so test expected to fail, and the windows test is known to be flaky (https://github.com/aws/aws-toolkit-vscode/issues/7187)

packages/core/src/test/testAuthUtil.ts
```diff
@@ -0,0 +1,43 @@
...
+        getProfile: sinon.stub().resolves({
+            sso_registration_scopes: ['codewhisperer'],
+        }),
+        deleteBearerToken: sinon.stub().resolves(),
+        updateBearerToken: sinon.stub().resolves(),
+        invalidateSsoToken: sinon.stub().resolves(),
+        encryptionKey,
+    }
+
+    AuthUtil.create(mockLspAuth as LanguageClientAuth)
```
nkomonen-amazon, line 42: I was looking at the implementation of create(). Does this work when called a subsequent time? The way it is designed, it can only be set once

packages/core/src/test/testAuthUtil.ts
```diff
@@ -0,0 +1,43 @@
...
+import * as sinon from 'sinon'
+import * as jose from 'jose'
+import * as crypto from 'crypto'
+import { LanguageClientAuth } from '../auth/auth2'
+import { AuthUtil } from '../codewhisperer/util/authUtil'
+
+export async function createTestAuthUtil() {
+    const encryptionKey = crypto.randomBytes(32)
+
+    const jwe = await new jose.CompactEncrypt(new TextEncoder().encode(JSON.stringify({ your: 'mock data' })))
```
nkomonen-amazon, line 15: Do we need all this or can we just put `fake-access-token`?

nkomonen-amazon:
@opieter-aws we should open a ticket to document auth not working on web. This will be something someone will eventually have to follow up on

packages/core/src/test/testAuthUtil.ts
opieter-aws: Correct, it’s a singleton so it only can be created once

packages/core/src/test/testAuthUtil.ts
opieter-aws: The decryption is called inside the auth2 class, so it needs to be working, otherwise it’ll throw

opieter-aws:
> @opieter-aws we should open a ticket to document auth not working on web. This will be something someone will eventually have to follow up on

Created an issue to disable the CI tests: https://github.com/aws/aws-toolkit-vscode/issues/7255

packages/core/src/test/testAuthUtil.ts
nkomonen-amazon: It looks like other tests call this, won't things not work as expected since there will be a different mockAuth being used?

nkomonen-amazon (approved):

packages/core/src/test/testAuthUtil.ts
opieter-aws: Not all tests require a the AuthUtil to be responsive, but merely to exist. So in those tests that were implemented before, a `AuthUtil.create` was sufficient to make them work. For these tests, the auth _state_ actually matters. We can consider updating the other tests to use this full mock implementation instead, but it's not interfering

PR #7251, opened 2025-05-07
fix(amazonq): add a more descriptive issue for firewall problems

jpinkney-aws:
## Problem
If a users firewall is blocking downloading the zips or reaching the manifest its not clear to the user what to do

## Solution
If the user:
1. doesn't have the latest version cached
2. download of the manifest failed
3. they don't have any previous language servers downloaded

it might be a firewall issue. Rather than just `Unable to find a compatible version of the Language Server` we show a more descriptive message to the user

Screenshot 2025-05-07 at 2 48 39 PM

---

### DIFF
--- a/packages/amazonq/test/e2e/lsp/lspInstallerUtil.ts
+++ b/packages/amazonq/test/e2e/lsp/lspInstallerUtil.ts
@@ -14,6 +14,7 @@ import {
     ManifestResolver,
     request,
     TargetContent,
+    ToolkitError,
 } from 'aws-core-vscode/shared'
 import * as semver from 'semver'
 import { assertTelemetry } from 'aws-core-vscode/test'
@@ -201,12 +202,6 @@ export function createLspInstallerTests({
                         id: lspConfig.id,
                         manifestLocation: 'remote',
                         languageServerSetupStage: 'getManifest',
-                        result: 'Failed',
-                    },
-                    {
-                        id: lspConfig.id,
-                        manifestLocation: 'cache',
-                        languageServerSetupStage: 'getManifest',
                         result: 'Succeeded',
                     },
                     {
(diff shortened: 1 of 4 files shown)


packages/amazonq/test/e2e/lsp/lspInstallerUtil.ts
```diff
@@ -201,12 +202,6 @@ export function createLspInstallerTests({
                         id: lspConfig.id,
                         manifestLocation: 'remote',
                         languageServerSetupStage: 'getManifest',
-                        result: 'Failed',
```
jpinkney-aws, line 204: this was broken in a different commit. Now the manifest download falls back to the local cache if an etag is found but it still reports the telemetry metric as "remote"

Technically this is a minor bug with reporting telemetry, since it should have been reported as a cache

justinmk3 (approved):

packages/core/src/shared/lsp/manifestResolver.ts
```diff
@@ -69,7 +69,6 @@ export class ManifestResolver {
 
             const localManifest = await this.getLocalManifest(true).catch(() => undefined)
             if (localManifest) {
-                localManifest.location = 'remote'
                 return localManifest
```
justinmk3, line 72: Why unset location=remote ? The reason I set that field, is because this logic is in the `fetchRemoteManifest` function. `this.getLocalManifest()` is being used for code re-use.

packages/core/src/shared/lsp/lspResolver.ts
```diff
@@ -90,7 +91,22 @@ export class LanguageServerResolver {
...
+        if (cachedVersions.length === 0) {
+            /**
+             * at this point the latest version doesn't exist locally, lsp download (with retries) failed, and there are no cached fallback versions.
+             * This _probably_ only happens when the user hit a firewall/proxy issue, since otherwise they would probably have at least
+             * one other language server locally
+             */
+            throw new ToolkitError(
+                `Unable to download dependencies from ${this.manifestUrl}. Check your network connectivity or firewall configuration and then try again.`,
+                {
+                    code: 'NetworkConnectivityError',
```
Hweinstock, line 104: these error codes auto-populate the `reason` field in telemetry right? If so, this will be important to track going forward to gauge impact. My suspicion is this only affects a very low % of users.

Hweinstock (approved):
nice!

packages/core/src/shared/lsp/lspResolver.ts
jpinkney-aws: Yeah exactly, that's part of why I wanted to add the code here. It should give us a rough idea of how many people are effected

PR #7256, opened 2025-05-08
fix(nep): truncate editor state

tomcat323:
## Problem
We introduced `editorState` in data instrumentation launch. The service has a requirement of 40k character limit for the `text` field. 

## Solution
Implement a check on text length. If the text length exceeds 40k characters, section 20k max characters from the left and right side of current cursor position, so the final text is always less than 40k.

validated prod endpoint inline working for files > 40k characters.
Example request id: `57bbbe65-fbe7-47fc-81c4-c65262f47ce8`

---

### DIFF
--- a/packages/core/src/codewhisperer/util/editorContext.ts
+++ b/packages/core/src/codewhisperer/util/editorContext.ts
@@ -11,7 +11,7 @@ import { getTabSizeSetting } from '../../shared/utilities/editorUtilities'
 import { getLogger } from '../../shared/logger/logger'
 import { runtimeLanguageContext } from './runtimeLanguageContext'
 import { fetchSupplementalContext } from './supplementalContext/supplementalContextUtil'
-import { supplementalContextTimeoutInMs } from '../models/constants'
+import { editorStateMaxLength, supplementalContextTimeoutInMs } from '../models/constants'
 import { getSelectedCustomization } from './customizationUtil'
 import { selectFrom } from '../../shared/utilities/tsUtils'
 import { checkLeftContextKeywordsForJson } from './commonUtil'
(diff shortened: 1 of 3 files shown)


dk19y (approved):

packages/core/src/codewhisperer/util/editorContext.ts
```diff
@@ -216,13 +216,26 @@ export function getTabSize(): number {
 
 export function getEditorState(editor: vscode.TextEditor, fileContext: codewhispererClient.FileContext): any {
     try {
+        const cursorPosition = editor.selection.active
+        const cursorOffset = editor.document.offsetAt(cursorPosition)
+        const documentText = editor.document.getText()
+
+        // Truncate if document content is too large (defined in constants.ts)
+        let fileText = documentText
+        if (documentText.length > editorStateMaxLength) {
```
jpinkney-aws, line 225: Will this have any impact on existing inline users?

tomcat323: This will fix lots of the `Improperly formed request.` errors so yes.

packages/core/src/codewhisperer/util/editorContext.ts
```diff
@@ -216,13 +216,26 @@ export function getTabSize(): number {
...
+            const startOffset = Math.max(0, cursorOffset - Math.floor(editorStateMaxLength / 2))
+            const endOffset = Math.min(documentText.length, cursorOffset + Math.floor(editorStateMaxLength / 2))
+
+            fileText = documentText.substring(startOffset, endOffset)
```
justinmk3, line 229: using truncate() might save some code here https://github.com/aws/aws-toolkit-vscode/blob/e9ea8082ffe0b9968a873437407d0b6b31b9e1a5/packages/core/src/shared/utilities/textUtilities.ts#L22

packages/core/src/codewhisperer/util/editorContext.ts
justinmk3: It's almost always worth mentioning improvements of any kind. In this case, it could be described like `Avoid "Improperly formed request" errors when using ...`

packages/core/src/codewhisperer/util/editorContext.ts
```diff
@@ -278,8 +295,8 @@ function logSupplementalContext(supplementalContext: CodeWhispererSupplementalCo
         logString += indent(`\nChunk ${index}:\n`, 4, true)
         logString += indent(
             `Path: ${context.filePath}
-            Length: ${context.content.length}
-            Score: ${context.score}`,
+Length: ${context.content.length}
+Score: ${context.score}`,
```
justinmk3, line 299: indent() should take care of this. is this change needed?

justinmk3 (approved):

PR #7263, opened 2025-05-08
test(amazonq): add unit tests for authUtil

opieter-aws:
## Problem
There are no unit tests for the new AuthUtil on Flare LSP

## Solution
* Add unit test for core AuthUtil functionality
* Create a test util for all tests where AuthUtil is needed
* Ensure all tests that interact with the AuthUtil consume the new test util
* Add destroy method to the AuthUtil class to simplify cleanup for unit tests

---

### DIFF
--- a/packages/core/src/test/testAuthUtil.ts
+++ b/packages/core/src/test/testAuthUtil.ts
@@ -10,6 +10,9 @@ import { LanguageClientAuth } from '../auth/auth2'
 import { AuthUtil } from '../codewhisperer/util/authUtil'
 
 export async function createTestAuthUtil() {
+    sinon.restore()
+    AuthUtil.destroy()
+
     const encryptionKey = crypto.randomBytes(32)
 
     const jwe = await new jose.CompactEncrypt(new TextEncoder().encode(JSON.stringify({ your: 'mock data' })))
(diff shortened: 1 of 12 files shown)


packages/core/src/test/testAuthUtil.ts
```diff
@@ -10,6 +10,9 @@ import { LanguageClientAuth } from '../auth/auth2'
 import { AuthUtil } from '../codewhisperer/util/authUtil'
 
 export async function createTestAuthUtil() {
+    sinon.restore()
```
nkomonen-amazon, line 13: I don't think you need this since your stubs below aren't overriding an actual method. This is used when you do something like `sinon.stub(myInstance, 'myMethod')` and you want to reset back to using the original.

But because this function creates a new set of stubs each time is called, the restore is not needed.

packages/core/src/test/testAuthUtil.ts
```diff
@@ -10,6 +10,9 @@ import { LanguageClientAuth } from '../auth/auth2'
...
+    AuthUtil.destroy()
```
nkomonen-amazon, line 14: Can you move this right before the Authutil.create() below and then add a comment why we need to call destroy right before create

nkomonen-amazon (approved):

PR #7270, opened 2025-05-08
fix(amazonq):support chat in al2 aarch64 and CDM

leigaol:
## Problem

LSP cannot start without GLIBC>=2.28 in AL2 aarch64 and CloudDevMachine

## Solution

---


jpinkney-aws (approved):

packages/amazonq/src/lsp/client.ts
```diff
@@ -43,8 +43,16 @@ import { ConfigSection, isValidConfigSection, toAmazonQLSPLogLevel } from './con
 const localize = nls.loadMessageBundle()
 const logger = getLogger('amazonqLsp.lspClient')
 
-export async function hasGlibcPatch(): Promise<boolean> {
-    return await fs.exists('/opt/vsc-sysroot/lib64/ld-linux-x86-64.so.2')
+export async function getGlibcPatch(): Promise<string | undefined> {
+    for (const file of [
+        '/opt/vsc-sysroot/lib64/ld-linux-x86-64.so.2',
+        '/opt/vsc-sysroot/lib64/ld-linux-aarch64.so.1',
```
justinmk3, line 49: wasn't there a generic path (symlink or something like that) that the team mentioned?

murkvin: Thanks for the call-out. Lei and I are going to chat today.

packages/core/src/shared/vscode/env.ts
```diff
@@ -131,10 +131,10 @@ export function isRemoteWorkspace(): boolean {
  *
  * Use {@link isCloudDesktop()} to know if we are specifically using it.
  *
- * Example: `5.10.220-188.869.amzn2int.x86_64`
+ * Example: `5.10.220-188.869.amzn2int.x86_64` or `5.10.236-227.928.amzn2.x86_64` (Cloud Dev Machine)
  */
 export function isAmazonInternalOs() {
-    return os.release().includes('amzn2int') && process.platform === 'linux'
+    return (os.release().includes('.amzn2int.') || os.release().includes('.amzn2.')) && process.platform === 'linux'
```
murkvin, line 137: question: How does this change impact external AL2 users?

packages/amazonq/src/lsp/client.ts
```diff
@@ -43,8 +43,16 @@ import { ConfigSection, isValidConfigSection, toAmazonQLSPLogLevel } from './con
 const localize = nls.loadMessageBundle()
 const logger = getLogger('amazonqLsp.lspClient')
 
-export async function hasGlibcPatch(): Promise<boolean> {
-    return await fs.exists('/opt/vsc-sysroot/lib64/ld-linux-x86-64.so.2')
+export async function getGlibcPath(): Promise<string | undefined> {
+    for (const file of [
+        '/opt/vsc-sysroot/lib64/ld-linux-x86-64.so.2',
```
murkvin, line 48: suggestion (non-blocking): We should migrate paths to the ones that are vended.

We will discuss later today and can treat it as a follow-up.

leigaol: Yes we need this PR to go out asap and do refactor later. This build is verified by our customer

packages/core/src/shared/vscode/env.ts
leigaol: Makes sense. Let me fix!

packages/amazonq/src/lsp/client.ts
leigaol: I am using the API interface now.

packages/core/src/shared/vscode/env.ts
```diff
@@ -125,23 +125,21 @@ export function isRemoteWorkspace(): boolean {
...
  *
- * Use {@link isCloudDesktop()} to know if we are specifically using it.
+ * Use {@link isCloudDesktop()} to know if we are specifically using internal Amazon Linux 2.
  *
- * Example: `5.10.220-188.869.amzn2int.x86_64`
+ * Example: `5.10.220-188.869.amzn2int.x86_64` or `5.10.236-227.928.amzn2.x86_64` (Cloud Dev Machine)
  */
-export function isAmazonInternalOs() {
-    return os.release().includes('amzn2int') && process.platform === 'linux'
+export function isAmazonLinux2() {
```
justinmk3, line 134: isn't CDM AL2023?
```suggestion
export function isAmazonLinux() {
```

murkvin: They support both.

leigaol: CDM can use both AL2 and AL2023

packages/amazonq/src/lsp/client.ts
```diff
@@ -44,8 +43,11 @@ import { telemetry } from 'aws-core-vscode/telemetry'
 const localize = nls.loadMessageBundle()
 const logger = getLogger('amazonqLsp.lspClient')
 
-export async function hasGlibcPatch(): Promise<boolean> {
-    return await fs.exists('/opt/vsc-sysroot/lib64/ld-linux-x86-64.so.2')
+export const glibcLinker: string = process.env.VSCODE_SERVER_CUSTOM_GLIBC_LINKER || ''
+export const glibcPath: string = process.env.VSCODE_SERVER_CUSTOM_GLIBC_PATH || ''
+
+export function hasGlibcPatch(): boolean {
```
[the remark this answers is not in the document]
murkvin: We're getting into deep semantics here, but I would argue that neither of these names is strictly correct. All that's really being checked is whether the environment variables that should contain the custom glibc / linker paths are defined. 

`hasGlibcPath` seems to provide a guarantee that the system has glibc at the location defined in the module. It's not actually checking what is there, just that there is some value in this environment variable that the patch exposes. This could be set to an arbitrary value from anywhere.

`hasGlibcPatch` loosely implies that the host has the full patch installed. Similar to above, these environment variables could have been set by something other than the patch.

If you care about this distinction, you could add more substantial checks into this function. If not, you could leave it as-is or add a comment to clarify since naming can genuinely be hard.

murkvin (approved):

packages/core/src/shared/vscode/env.ts
leigaol: This only checks al2

packages/amazonq/src/lsp/client.ts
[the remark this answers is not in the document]
justinmk3: > `hasGlibcPatch` loosely implies that the host has the full patch installed.

ok, that's the part I was missing. It's not mentioned in a docstring, nor is it obvious.

justinmk3 (approved):

packages/amazonq/src/lsp/client.ts
justinmk3: you mean `VSCODE_SERVER_CUSTOM_GLIBC_LINKER` ?

leigaol: Yes exactly. These are the AL2 "APIs"

packages/amazonq/src/lsp/client.ts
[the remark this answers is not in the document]
leigaol: I can add docstring as fast follow and I will absolutely do it

nkomonen-amazon (approved):

PR #7274, opened 2025-05-09
test(amazonq): update inline tests for flare

jpinkney-aws:
## Problem
inline tests don't work with the language server

## Solution
- I removed the `${name} invoke on unsupported filetype` test because you can't trigger inline completions at all for unsupported filetypes (vscode doesn't allow it -- you need to specify file extentions)
- hack: I have no way to spy on actual inline completions so I added a global variable that lets the tests know when recommendations are being generated :/. This allows us to wait for before accepting/rejecting requests
- hack: codewhisperer_perceivedLatency, codewhisperer_serviceInvocation don't have the result field set in flare so it causes a lot of noisy logs. TODO add the result field there

---

### DIFF
--- a/packages/amazonq/src/app/inline/completion.ts
+++ b/packages/amazonq/src/app/inline/completion.ts
@@ -33,11 +33,13 @@ import {
     ReferenceLogViewProvider,
     ImportAdderProvider,
     CodeSuggestionsState,
+    vsCodeState,
 } from 'aws-core-vscode/codewhisperer'
 import { InlineGeneratingMessage } from './inlineGeneratingMessage'
 import { LineTracker } from './stateTracker/lineTracker'
 import { InlineTutorialAnnotation } from './tutorials/inlineTutorialAnnotation'
 import { TelemetryHelper } from './telemetryHelper'
+import { getLogger } from 'aws-core-vscode/shared'
 
 export class InlineCompletionManager implements Disposable {
     private disposable: Disposable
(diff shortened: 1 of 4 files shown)


packages/amazonq/src/app/inline/completion.ts
```diff
@@ -203,6 +204,7 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
         context: InlineCompletionContext,
         token: CancellationToken
     ): Promise<InlineCompletionItem[] | InlineCompletionList> {
+        vsCodeState.isRecommendationsActive = true
```
nkomonen-amazon, line 207: Does it make sense to wrap this in a try/finally to ensure it get appropriately reset in case of error.

jpinkney-aws: just updated to do this. We technically still need to set `vsCodeState.isRecommendationsActive = false` before the completion returns though but now if something fails it will always be reset

packages/amazonq/src/app/inline/completion.ts
```diff
@@ -203,52 +204,64 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
...
+                ImportAdderProvider.instance.onShowRecommendation(document, position.line, item)
             }
-            ReferenceInlineProvider.instance.setInlineReference(
-                position.line,
-                item.insertText as string,
-                item.references
-            )
-            ImportAdderProvider.instance.onShowRecommendation(document, position.line, item)
+            // exposed only for testing
+            vsCodeState.isRecommendationsActive = false
```
nkomonen-amazon, line 257: We should be able to delete this line since it is in the finally

jpinkney-aws: you need to enable it before inline completions responds, I have no idea why though

nkomonen-amazon (approved):

PR #7275, opened 2025-05-09
refactor(amazonq): deprecate local workspace context

jpinkney-aws:
## Problem
we no longer need local workspace context, since it's now provided through flare

## Solution
deprecate it

---


packages/core/src/codewhispererChat/controllers/chat/controller.ts
```diff
@@ -527,7 +521,7 @@ export class ChatController {
                 commands: [{ command: commandName, description: commandDescription }],
             })
         }
-        const symbolsCmd: QuickActionCommand = contextCommand[0].commands?.[3]
+        // const symbolsCmd: QuickActionCommand = contextCommand[0].commands?.[3]
```
jpinkney-aws, line 524: i don't know if anything "accidentally" goes down this codepath still (it shouldn't) so I just commented out what made sense for now

PR #7293, opened 2025-05-12
fix(amazonq): Developer Profiles not being sent to Flare Auth as expected

nkomonen-amazon:
## Problem

Due to the order that we restored auth, versus sent the Developer Profiles to the Auth LSP, there was a problem. We sent the Developer Profiles before the Bearer Token was setup in the Amazon Q LSP

## Solution

Ensure that we complete restoring of Auth (pushing the Bearer Token to the Amazon Q LSP). Then restore the Developer Profile, ensuring we show the correct messages if needed.

## To Review

See each individual commit for a breakdown as their are additional refactors.

---


packages/core/src/auth/auth2.ts
```diff
@@ -325,10 +325,14 @@ export class SsoLogin implements BaseLogin {
...
     private updateConnectionState(state: AuthState) {
-        if (this.connectionState !== state) {
-            this.eventEmitter.fire({ id: this.profileName, state })
+        const oldState = this.connectionState
+        const newState = state
+
+        this.connectionState = newState
+
+        if (oldState !== newState) {
+            this.eventEmitter.fire({ id: this.profileName, state: this.connectionState })
```
opieter-aws, line 334: Are there no race conditions in swapping the event emitter with the status update?

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -91,8 +93,20 @@ export class AuthUtil implements IAuthProvider {
...
+     *       the Bearer Token in the LSP first.
+     */
+    startedConnected = false
+
     async restore() {
         await this.session.restore()
+        this.startedConnected = this.isConnected()
+
+        // HACK: The LSP should allow us to call refreshState multiple times, but for some reason it breaks.
+        // isConnected somehow allows it to work. Assumption is that the LSP does not handle redundant calls nicely.
```
opieter-aws, line 109: Maybe rephrase to: 'this check ensures we only call refreshState when restore does not succeed to connect, to avoid duplicate calls', since we do know why it works

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -91,8 +93,20 @@ export class AuthUtil implements IAuthProvider {
         return this.session.loginType === LoginTypes.SSO
     }
 
+    /**
+     * HACK: Ideally we'd put {@link notifySelectDeveloperProfile} in to {@link restore}.
+     *       But because {@link refreshState} is only called if !isConnected, we cannot do it since
+     *       {@link notifySelectDeveloperProfile} needs {@link refreshState} to run so it can set
+     *       the Bearer Token in the LSP first.
+     */
+    startedConnected = false
```
opieter-aws, line 102: nit: the name `startedConnected` is misleading, since the AuthUtil is always instantiated with the state `notConnected`. I was thinking, and maybe it's clearer to not reference an AuthState here. Instead, consider renaming to `startedSignedOut` or `startedSignedIn`, which is more descriptive and avoids confusion with auth state. Then you can change line 110 to assert the same variable, which also makes it clearer what we actually do there

packages/core/src/auth/auth2.ts
nkomonen-amazon: I think it is unlikely (maybe impossible in this case), but when we fire the event emitter and have not updated `this.connectionState` to the correct value yet, something may attempt to read it after getting the event and it will be a stale value.

I made this change initially though with the assumption that it may have had no actual impact.

packages/core/src/codewhisperer/util/authUtil.ts
nkomonen-amazon: IMO we shouldn't care about duplicate calls of `refreshState` since it should functionally work if called multiple times. If `refreshState` is fragile it should be indicated in a doc string.

Also in `restore()` I would assume that after calling it, the whole auth state is "ready to go", but it isn't since we need to wait for the event emitter to implicitly trigger `refreshState`.

I'm going to reword the comment though to word this better

packages/core/src/codewhisperer/util/authUtil.ts
nkomonen-amazon: Good improvement, will update

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -257,20 +281,24 @@ export class AuthUtil implements IAuthProvider {
...
             if (this.isIdcConnection()) {
                 await this.regionProfileManager.restoreProfileSelection()
             }
-            const bearerTokenParams = (await this.session.getToken()).updateCredentialsParams
-            await this.lspAuth.updateBearerToken(bearerTokenParams)
         }
 
+        // regardless of state, send message at startup if user needs to select a Developer Profile
+        void this.tryNotifySelectDeveloperProfile()
+
```
opieter-aws, line 301: This `refreshState` function triggers not only upon startup, but on every auth state change. Do we need to check here whether it's the first use or something?

nkomonen-amazon: The function, see below, only triggers once on startup, and any subsequent times it will not because of the use of `once`

opieter-aws (approved):

PR #7298, opened 2025-05-13
feat(amazonq): Migrate existing SSO connections to the LSP identity server

opieter-aws:
## Problem
Existing SSO connections on old auth, are not migrated automatically to the LSP Identity Server. This means that users would be signed out upon migration.

## Solution
Add a migration functionality that checks for existing SSO connections, and migrate if necessary:
* Check inside the `client.ts`, before we attempt to restore any connections in the new auth
* Add `migrateSsoConnectionToLsp` method to `AuthUtil`, which:
   * Fetches environment `memento`, and looks for the `auth.profiles` key
   * Checks if there is an `sso` profile with `amazonQScopes`. If yes: migrate, if not: nothing to migrate
   * [migration]: Update the SSO profile with Flare
   * [migration]: Construct the existing (`from`) filenames and the Flare (`to`) filenames for the registration cache file and the SSO token cache file
   * [migration]: rename the files
   * [migration]: set the `auth.profiles` key to undefined, so migration is skipped next time

## Testing
* Covered with unit tests that confirms the actual file system operations
* Manual E2E testing, verifying that there is no sign-out happening with this setup
* We will verify the mechanism through bug bashing

---


packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -333,4 +341,79 @@ export class AuthUtil implements IAuthProvider {
...
+                        startUrl: toImport.startUrl,
+                        tool: clientName,
+                    })
+                )
+
+                const fromTokenFile = getTokenCacheFile(cacheDir, profileId)
+                const toTokenFile = filePath(this.profileName)
+
+                try {
+                    fs.renameSync(fromRegistrationFile, toRegistrationFile)
```
nkomonen-amazon: What is the reason you have to use the 'fs' module? It has a rename method

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -333,4 +341,79 @@ export class AuthUtil implements IAuthProvider {
...
+    async migrateSsoConnectionToLsp(clientName: string) {
+        const memento = getEnvironmentSpecificMemento()
+        const key = 'auth.profiles'
+        const profiles: { readonly [id: string]: StoredProfile } | undefined = memento.get(key)
+
+        let toImport: SsoProfile | undefined
+        let profileId: string | undefined
+
+        getLogger().info(`codewhisperer: checking for old SSO connections`)
+        if (profiles) {
```
nkomonen-amazon, line 361: Can we return early if !profiles to reduce indentation depth

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -333,4 +341,79 @@ export class AuthUtil implements IAuthProvider {
...
     }
+
+    /**
+     * Migrates existing SSO connections to the LSP identity server by updating the cache files
+     *
+     * @param clientName - The client name to use for the new registration cache file
+     * @returns A Promise that resolves when the migration is complete
+     * @throws Error if file operations fail during migration
+     */
+    async migrateSsoConnectionToLsp(clientName: string) {
```
nkomonen-amazon, line 352: Can you add a test in `techdebt.test.ts` to fail in 2 months, since by then we will probably want to delete this code as most users would have migrated already

opieter-aws: I think this is a bit premature given that 1. we do not know yet when this will release and 2. we want an X percentage of users to have migrated to that version number before removing the code. Created a separate ticket to track: https://github.com/aws/aws-toolkit-vscode/issues/7304

nkomonen-amazon (approved):

PR #7305, opened 2025-05-14
test(amazonq): add tests for edge cases

Hweinstock:
## Problem
There are two edge cases discovered during testing that warrant some tests. 
1. the language server will respond without a range, causing the result to not be rendered. 
2. the language server COULD respond with a `StringValue` instead of string (currently doesn't, but valid in the type contract). 

## Solution
- add tests for these cases, and do minor refactoring to make this easier. 

---

### DIFF
--- a/packages/amazonq/src/app/inline/completion.ts
+++ b/packages/amazonq/src/app/inline/completion.ts
@@ -8,7 +8,6 @@ import {
     InlineCompletionContext,
     InlineCompletionItem,
     InlineCompletionItemProvider,
-    InlineCompletionList,
     Position,
     TextDocument,
     commands,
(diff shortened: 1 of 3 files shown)


packages/amazonq/src/app/inline/completion.ts
```diff
@@ -246,8 +245,7 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
                 return []
             }
 
-            const start = document.validatePosition(editor.selection.active)
```
Hweinstock, line 249: In non-testing scenarios, `editor.selection.active` appears to always be the same as the position passed in here.

PR #7316, opened 2025-05-15
fix(amazonq): Profile needing to be selected on server restart

nkomonen-amazon:
See individual commits for isolated changes

## Problem

We were seeing the following errors from the Q Language Server on startup:
- `Amazon Q Profile is not selected for IDC connection type`
- `Amazon Q service is not signed in`

## Solution

We needed to do 2 solutions, each is a separate commit (see their message). There were also some minor refactors.

In short:
- The Auth bearer token MUST be sent to the Q LSP before Profile is sent. We were not doing this and it was causing an error
- When sending the Auth to the Q LSP, the startUrl MUST be included in the request or else it would fail. We thought we were sending it but based on the logs prefixed with `UpdateBearerToken` it showed`sso` did not contain the startUrl

---

### DIFF
--- a/packages/amazonq/src/lsp/auth.ts
+++ b/packages/amazonq/src/lsp/auth.ts
@@ -115,14 +115,11 @@ export class AmazonQLspAuth {
             .setProtectedHeader({ alg: 'dir', enc: 'A256GCM' })
             .encrypt(encryptionKey)
 
-        const conn = AuthUtil.instance.conn
-        const startUrl = conn?.type === 'sso' ? conn.startUrl : undefined
-
         return {
             data: jwt,
             metadata: {
                 sso: {
-                    startUrl,
+                    startUrl: AuthUtil.instance.startUrl,
                 },
             },
             encrypted: true,
--- a/packages/amazonq/src/lsp/auth.ts
+++ b/packages/amazonq/src/lsp/auth.ts
@@ -18,6 +18,7 @@ import { AuthUtil } from 'aws-core-vscode/codewhisperer'
 import { Writable } from 'stream'
 import { onceChanged } from 'aws-core-vscode/utils'
 import { getLogger, oneMinute } from 'aws-core-vscode/shared'
+import { isSsoConnection } from 'aws-core-vscode/auth'
 
 export const encryptionKey = crypto.randomBytes(32)
 
@@ -76,8 +77,8 @@ export class AmazonQLspAuth {
      * @param force bypass memoization, and forcefully update the bearer token
      */
     async refreshConnection(force: boolean = false) {
-        const activeConnection = this.authUtil.auth.activeConnection
-        if (activeConnection?.state === 'valid' && activeConnection?.type === 'sso') {
+        const activeConnection = this.authUtil.conn
+        if (this.authUtil.isConnectionValid() && isSsoConnection(activeConnection)) {
             // send the token to the language server
             const token = await this.authUtil.getBearerToken()
             await (force ? this._updateBearerToken(token) : this.updateBearerToken(token))
--- a/packages/amazonq/src/lsp/auth.ts
+++ b/packages/amazonq/src/lsp/auth.ts
@@ -115,11 +115,14 @@ export class AmazonQLspAuth {
             .setProtectedHeader({ alg: 'dir', enc: 'A256GCM' })
             .encrypt(encryptionKey)
 
+        const conn = AuthUtil.instance.conn
+        const startUrl = conn?.type === 'sso' ? conn.startUrl : undefined
+
         return {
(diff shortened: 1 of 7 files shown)


packages/amazonq/src/lsp/auth.ts
```diff
@@ -114,11 +115,14 @@ export class AmazonQLspAuth {
             .setProtectedHeader({ alg: 'dir', enc: 'A256GCM' })
             .encrypt(encryptionKey)
 
+        const conn = AuthUtil.instance.conn
+        const startUrl = conn?.type === 'sso' ? conn.startUrl : undefined
```
jpinkney-aws, line 119: can we just use AuthUtil.instance.startUrl here? I think previously we were using `AuthUtil.instance.auth.startUrl` which might have been causing the problem?

nkomonen-amazon: Yeah that was the root, of it. I haven't updated the description but AuthUtil.instance.auth.startUrl is actually using `Auth` and not the one from `AuthUtil`. It looks like we never set an `activeConnection` in `Auth` so `startUrl` from there returns nothing.

Hweinstock:
> Idk why this only broke now though

+1

Hweinstock:
> Ensure the bearer token is pushed first, then follow up with pushing the profile/customization after.

Is there any way we can enforce this with a test? Since this behavior is so critical its likely worth finding a way to ensure this doesn't accidentally get broken in the future. Or is this not worth it since this will be deleted by Flare migration soon?

> Use the connection we get from AuthUtil as that is the most accurate one.

If these make it through the Flare Migration, we should invest time to rename this or find a way to make this more obvious, because as a consumer I would have no idea.

packages/amazonq/src/lsp/chat/activation.ts
```diff
@@ -12,24 +12,26 @@ import { Commands, getLogger, globals, undefinedIfEmpty } from 'aws-core-vscode/
...
-    })
-    // We need to push the cached customization on startup explicitly
-    await pushConfigUpdate(languageClient, {
-        type: 'customization',
-        customization: getSelectedCustomization(),
-    })
+    //
+    // Ideally the handler for onDidChangeRegionProfile would trigger this, but because the handler is set up too late (due to the current structure), the initial event is missed.
+    // So we have to explicitly do it here on startup.
+    if (AuthUtil.instance.isConnectionValid()) {
```
jpinkney-aws, line 24: should this be moved up to the main activation? I guess we don't really need to bury them in here anymore

packages/amazonq/src/lsp/client.ts
```diff
@@ -160,120 +160,141 @@ export async function startLanguageServer(
...
-            )
-        }
+    return client
+}
 
-        if (Experiments.instance.get('amazonqChatLSP', true)) {
-            await activate(client, encryptionKey, resourcePaths.ui)
-        }
+async function initializeAuth(client: LanguageClient): Promise<AmazonQLspAuth> {
+    AuthUtil.instance.regionProfileManager.onDidChangeRegionProfile(async () => {
```
jpinkney-aws, line 173: do we need to add this to the disposables?

Hweinstock, line 173: technically this listener could trigger before we push the bearer token below in `await auth.refreshConnection(true)` right? Wouldn't that be problematic?

packages/amazonq/src/lsp/client.ts
```diff
@@ -160,120 +160,141 @@ export async function startLanguageServer(
...
-                    },
-                })
-                client.info(
-                    `Client: Updated Amazon Q Profile ${AuthUtil.instance.regionProfileManager.activeRegionProfile?.arn} to Amazon Q LSP`,
-                    result
-                )
-            } catch (err) {
-                client.error('Error when setting Q Developer Profile to Amazon Q LSP', err)
-            }
+async function postStartLanguageServer(
```
Hweinstock, line 185: nit: should we call this something like `OnLanguageServerReady` to be consistent with Flare language?

packages/amazonq/src/lsp/client.ts
nkomonen-amazon: From testing I haven't found a code path where this happens

velatha:
Did we always have this race condition or a regression that got introduced recently?

Hweinstock:
> Did we always have this race condition or a regression that got introduced recently?

This is still unknown. I tried previous versions of the extension and language server this morning and still saw the same error so I suspect this isn't recent.

jpinkney-aws (approved):

Hweinstock (approved):

PR #7318, opened 2025-05-15
feat(amazonq): show a message when no suggestions are found on manual invoke

Hweinstock:
## Problem
When manually invoking inline suggestions, it is possible for the backend to still not reply with results. Since the user took an action to invoke this, we should let them know when it fails. 

## Solution
- add a message with same timeout as original implementation. 

## Testing and Verification
- added unit test. 
- E2E flow demo (notice it doesn't trigger on automatic failures, but does on manual invoke failures):

https://github.com/user-attachments/assets/7cf34178-e91e-4d38-8875-1259c5552a53

---

### DIFF
--- a/packages/amazonq/src/app/inline/completion.ts
+++ b/packages/amazonq/src/app/inline/completion.ts
@@ -34,7 +34,7 @@ import {
     CodeSuggestionsState,
     vsCodeState,
     inlineCompletionsDebounceDelay,
-    noSuggestions,
+    noInlineSuggestionsMsg,
 } from 'aws-core-vscode/codewhisperer'
 import { InlineGeneratingMessage } from './inlineGeneratingMessage'
 import { LineTracker } from './stateTracker/lineTracker'
@@ -245,7 +245,7 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
 
             // Show message to user when manual invoke fails to produce results.
             if (items.length === 0 && context.triggerKind === InlineCompletionTriggerKind.Invoke) {
-                void messageUtils.showTimedMessage(noSuggestions, 2000)
+                void messageUtils.showTimedMessage(noInlineSuggestionsMsg, 2000)
             }
 
             if (!session || !items.length || !editor) {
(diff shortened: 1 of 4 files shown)


packages/amazonq/src/app/inline/completion.ts
```diff
@@ -241,7 +242,16 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
...
             const editor = window.activeTextEditor
+
+            // Show message to user when manual invoke fails to produce results.
+            if (items.length === 0 && context.triggerKind === InlineCompletionTriggerKind.Invoke) {
+                void messageUtils.showTimedMessage(noInlineSuggestionsMsg, 2000)
+            }
+
             if (!session || !items.length || !editor) {
+                getLogger().debug(
+                    `Failed to produce inline suggestion results. Recieved ${items.length} items from service`
```
jpinkney-aws:
```suggestion
                    `Failed to produce inline suggestion results. Received ${items.length} items from service`
```

packages/amazonq/test/unit/amazonq/apps/inline/completion.test.ts
```diff
@@ -400,57 +411,71 @@ describe('InlineCompletionManager', () => {
...
+                        messageShown = true
                     })
+                    await provider.provideInlineCompletionItems(
+                        mockDocument,
+                        mockPosition,
+                        { triggerKind: InlineCompletionTriggerKind.Invoke, selectedCompletionInfo: undefined },
+                        mockToken
+                    )
+                    // Wait up to a second for message to appear since there is potential race condition.
+                    await waitUntil(async () => messageShown, { timeout: 1000, interval: 100 })
```
jpinkney-aws, line 435: instead of waiting I think you might be able to do something like this:
```
const p = new Promise((resolve) => {
     getTestWindow().onDidShowMessage((e) => {
         assert.strictEqual(e.message, noInlineSuggestionsMsg)
         resolve()
     })
})
```
and then
```
await p
```
here

PR #7320, opened 2025-05-15
fix(amazonq): Sync IDE windows for Amazon Q auth state and region profile selection

opieter-aws:
## Problem
The auth state and region profile selection is not sycned between different IDE windows

#### Auth state
* Add an `ssoCacheWatcher` to the LSP auth client, with hooks on `onDidCreate` and `onDidDelete`
* In the `AuthUtil`, add the following handlers:
   * `onDidCreate`: trigger a restore flow to fetch the latest auth state
   *  `onDidDelete`: trigger a logout

#### Region profile
* Add a `GlobalStatePoller` util that polls the global state value every second, and call a handler if the value updates
* In `RegionProfileManager` add the poller for the region profile global state variable and add a handler that switches the profile

## Testing

https://github.com/user-attachments/assets/8cdf6abd-94bd-4e60-9bcf-a60f42d1847e

---

### DIFF
--- a/packages/core/src/auth/auth2.ts
+++ b/packages/core/src/auth/auth2.ts
@@ -41,6 +41,7 @@ import { LanguageClient } from 'vscode-languageclient'
 import { getLogger } from '../shared/logger/logger'
 import { ToolkitError } from '../shared/errors'
 import { useDeviceFlow } from './sso/ssoAccessTokenProvider'
+import { getCacheFileWatcher } from './sso/cache'
 
 export const notificationTypes = {
     updateBearerToken: new RequestType<UpdateCredentialsParams, ResponseMessage, Error>(
@@ -74,11 +75,17 @@ export type TokenSource = IamIdentityCenterSsoTokenSource | AwsBuilderIdSsoToken
  * Handles auth requests to the Identity Server in the Amazon Q LSP.
  */
 export class LanguageClientAuth {
+    readonly #ssoCacheWatcher = getCacheFileWatcher()
+
     constructor(
         private readonly client: LanguageClient,
         private readonly clientName: string,
         public readonly encryptionKey: Buffer
     ) {}
+
+    public get cacheWatcher() {
+        return this.#ssoCacheWatcher
+    }
 
     getSsoToken(
         tokenSource: TokenSource,
@@ -159,6 +166,11 @@ export class LanguageClientAuth {
 
     registerSsoTokenChangedHandler(ssoTokenChangedHandler: (params: SsoTokenChangedParams) => any) {
         this.client.onNotification(ssoTokenChangedRequestType.method, ssoTokenChangedHandler)
+    }
+
+    registerCacheWatcher(cacheChangedHandler: (event: string) => any) {
+        this.cacheWatcher.onDidCreate(() => cacheChangedHandler('create'))
+        this.cacheWatcher.onDidDelete(() => cacheChangedHandler('delete'))
     }
 }
 
--- a/packages/core/src/auth/auth2.ts
+++ b/packages/core/src/auth/auth2.ts
@@ -67,6 +67,8 @@ interface BaseLogin {
     readonly loginType: LoginType
 }
 
+export type cacheChangedEvent = 'delete' | 'create'
+
 export type Login = SsoLogin // TODO: add IamLogin type when supported
 
 export type TokenSource = IamIdentityCenterSsoTokenSource | AwsBuilderIdSsoTokenSource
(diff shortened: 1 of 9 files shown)


hayemaxi:
Very nice. Does it work if you say, select the profile in the other window that you weren't using to log in?

hayemaxi:
Do we need to poll **every** second? Maybe every 2 or 3 seconds?

packages/core/src/shared/globalState.ts
```diff
@@ -318,3 +318,48 @@ export class GlobalState implements vscode.Memento {
...
+        return instance
+    }
+
+    /**
+     * Starts polling the state value at 1 second intervals.
+     * When a change is detected, the changeHandler callback is invoked.
+     */
+    poll() {
+        const interval = 1000 // ms
+        setInterval(() => {
```
hayemaxi, line 357: You should store this interval in the class, and have a method to kill the interval (turn off the watcher). We can call this on extension deactivation.

packages/core/src/shared/globalState.ts
```diff
@@ -318,3 +318,48 @@ export class GlobalState implements vscode.Memento {
...
+        const instance = new GlobalStatePoller(getState, changeHandler)
+        instance.poll()
+        return instance
+    }
+
+    /**
+     * Starts polling the state value at 1 second intervals.
+     * When a change is detected, the changeHandler callback is invoked.
+     */
+    poll() {
```
hayemaxi, line 355:
```suggestion
    private poll() {
```

opieter-aws:
> Do we need to poll **every** second? Maybe every 2 or 3 seconds?

The behavior gets awkward when you are logged out and log back in; the 'other' window will keep the region selector displayed until the sync. I can increase to 2 seconds, but I don't think we want to keep that state much longer

packages/core/src/auth/auth2.ts
```diff
@@ -160,6 +167,11 @@ export class LanguageClientAuth {
     registerSsoTokenChangedHandler(ssoTokenChangedHandler: (params: SsoTokenChangedParams) => any) {
         this.client.onNotification(ssoTokenChangedRequestType.method, ssoTokenChangedHandler)
     }
+
+    registerCacheWatcher(cacheChangedHandler: (event: string) => any) {
```
nkomonen-amazon, line 171: Can we enumerate the type of `event` to `'create' | 'delete` where appropriate

packages/core/src/codewhisperer/region/regionProfileManager.ts
```diff
@@ -120,7 +121,18 @@ export class RegionProfileManager {
         return this._profiles
     }
 
-    constructor(private readonly authProvider: IAuthProvider) {}
+    constructor(private readonly authProvider: IAuthProvider) {
+        const getProfileFunction = () =>
+            globals.globalState.tryGet<{ [label: string]: RegionProfile }>('aws.amazonq.regionProfiles', Object, {})
```
nkomonen-amazon, line 126: There are multiple instances of this tryGet in this file. Can we make a single private method and just call that

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -276,6 +277,14 @@ export class AuthUtil implements IAuthProvider {
         })
     }
 
+    private async cacheChangedHandler(event: string) {
```
nkomonen-amazon, line 280: Can we enumerate the type of `event` to `'create' | 'delete'`

hayemaxi:
> One thing to note with this polling length. We should think of realistic user behaviors and if we set polling to 2-3 seconds, would they even notice it? Assuming they sign in or out on one IDE instance, it is low likelihood they are switching to the other IDE very quickly.
> 
> It is also unlikely that the user will select the profile in one window and then immediately switch to the other.

+1

No one is comparing how smooth the auth change is side-by-side.

packages/core/src/shared/globalState.ts
```diff
@@ -318,3 +318,65 @@ export class GlobalState implements vscode.Memento {
...
+        const instance = new GlobalStatePoller(props)
+        instance.poll()
+        return instance
+    }
+
+    /**
+     * Starts polling the state value. When a change is detected, the changeHandler callback is invoked.
+     */
+    private poll() {
+        this.intervalId = setInterval(() => {
```
hayemaxi, line 364: should probably call `kill()` at the start of this function as a safety check

hayemaxi (approved):

opieter-aws:
> Maybe we should bring this to mainline as well

We won't be adding features for auth to mainline for now, but instead prioritize releasing this branch with auth on LSP identity server

PR #7323, opened 2025-05-15
fix(amazonq): inline suggestion scrolling now uses built-in VSC commands. 

Hweinstock:
## Problem
The current implementation overwrites the `editor.action.inlineSuggest.showPrevious` and `editor.action.inlineSuggest.showNext` commands to swap out the inlineProvider. These custom overwrites cause a few issues:
- forces us to maintain a ton of state [leads to bugs, some of which were observed in bug bash].
- causes the suggestions to "refresh" when going through them since we are re-instantiating the provider on "scrolls". 
- doesn't natively support "accept word"/"accept line". 

## Solution
- Remove the overwrite, remove reliance of `session` state. 
- Simplify implementation to use native VSC commands. 

## Fixed Bugs
- Scrolling through suggestions now works and doesn't "refresh". 
- Accept Word/Accept Line now work properly. 
- Significantly faster to scroll through suggestions. 
- Inline Reference Codelense no longer flashes when scrolling through suggestions. 

## Tests and Verification: 
Testing this change it feels significantly 'smoother'. 

https://github.com/user-attachments/assets/631ee822-3cab-46b5-a07c-b3d4b8eb7d3b

## Future Work
- with some more aggressive refactoring, and once telemetry is mostly in Flare, we can completely delete the sessionManager. 
- revisit reference codelense, and see if its the right choice. Maybe a message instead?

---

### DIFF
--- a/packages/amazonq/src/app/inline/completion.ts
+++ b/packages/amazonq/src/app/inline/completion.ts
@@ -28,7 +28,6 @@ import { RecommendationService } from './recommendationService'
 import {
     CodeWhispererConstants,
     ReferenceHoverProvider,
-    ReferenceInlineProvider,
     ReferenceLogViewProvider,
     ImportAdderProvider,
     CodeSuggestionsState,
@@ -234,11 +233,6 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
                 }
                 item.range = new Range(cursorPosition, cursorPosition)
                 item.insertText = typeof item.insertText === 'string' ? item.insertText : item.insertText.value
-                ReferenceInlineProvider.instance.setInlineReference(
-                    cursorPosition.line,
-                    item.insertText,
-                    item.references
-                )
                 ImportAdderProvider.instance.onShowRecommendation(document, cursorPosition.line, item)
             }
             return items as InlineCompletionItem[]
--- a/packages/amazonq/src/app/inline/completion.ts
+++ b/packages/amazonq/src/app/inline/completion.ts
@@ -181,10 +181,8 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
         context: InlineCompletionContext,
         token: CancellationToken
     ): Promise<InlineCompletionItem[]> {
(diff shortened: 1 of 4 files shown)


packages/amazonq/src/app/inline/completion.ts
```diff
@@ -158,39 +157,6 @@ export class InlineCompletionManager implements Disposable {
             this.languageClient.sendNotification(this.logSessionResultMessageName, params)
         }
         commands.registerCommand('aws.amazonq.rejectCodeSuggestion', onInlineRejection)
-
-        /*
-            We have to overwrite the prev. and next. commands because the inlineCompletionProvider only contained the current item
-            To show prev. and next. recommendation we need to re-register a new provider with the previous or next item
-        */
-
-        const swapProviderAndShow = async () => {
```
jpinkney-aws, line 167: cc @zixlin7 was this stuff added to support inline completions that come _after_ the initial request? or is it no longer needed? As far as I can see from the demo we still have pagination

Hweinstock: My understanding of the old implementation is that we fetch all the suggestions, add them to the session, and then request them from the session one at a time (by triggering the `provideInlineCompletion` via overwriting the scroll commands). We also add 'placeholder' items so that VSC knows we can scroll. 

Since we're fetching all the suggestions anyway, we can return those directly and avoid 'placeholder' items, and additional triggers of `provideInlineCompletion`. 

However, I don't have all the context on the original reason for this decision.

nkomonen-amazon (approved):

packages/amazonq/src/app/inline/completion.ts
Hweinstock: Checked with Zoe and sounds like this is a hack to handle the codelenses for references and imports. 

- references are confirmed working in https://github.com/aws/aws-toolkit-vscode/pull/7331. 
- import codelense is unknown. Note: There weren't any reported issue in first three bug bashes, so either this is not a common case, or already works. However, still worth following up to confirm this.

PR #7357, opened 2025-05-21
fix(amazonq): throw if no region profiles are available

opieter-aws:
## Problem
When ListRegionProfile call throttles for a subset of regions, we currently do not throw, but instead return the available profiles in the regions where the call succeeded. However, if that list is empty (no profiles in that region), we return an empty list. This breaks the UI, and causes a state that is not recoverable

## Solution
Throw an error in the scenario where availableProfiles is empty. This triggers a retry state in the UI, making the state recoverable.

---

### DIFF
--- a/packages/core/src/codewhisperer/region/regionProfileManager.ts
+++ b/packages/core/src/codewhisperer/region/regionProfileManager.ts
@@ -182,9 +182,17 @@ export class RegionProfileManager {
             }
         }
 
-        // Only throw error if all regions fail
-        if (failedRegions.length === endpoints.size) {
-            throw new Error(`Failed to list profiles for all regions: ${failedRegions.join(', ')}`)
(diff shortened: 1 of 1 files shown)


Hweinstock (approved):

Hweinstock:
Is this also present on prod?

Will-ShaoHua (approved):

Will-ShaoHua:
> Is this also present on prod?

yea, it should.

nkomonen-amazon (approved):

Hweinstock:
> yea, it should. 

Should we target master then, or does that cause headaches with merge conflicts?

opieter-aws:
> > yea, it should.
> 
> Should we target master then, or does that cause headaches with merge conflicts?

I can't update this PR (pulls in way too many commits), but I will merge this and then PR the same change to master

PR #7363, opened 2025-05-22
fix(amazonq): handle existing Flare connection when migrating SSO connections

opieter-aws:
## Problem
Auth migration to LSP is not handled gracefully when a user downgrades and upgrades to auth on LSP multiple times, causing users to be logged out if they upgrade a second time

## Solution
In the auth migration script, call the LSP identity server to check if a token is available. If the token is available, don't migrate the auth connection. If no token is available, migrate.

Added unit tests for the case.

---

### DIFF
--- a/packages/core/src/codewhisperer/util/authUtil.ts
+++ b/packages/core/src/codewhisperer/util/authUtil.ts
@@ -65,6 +65,8 @@ export interface IAuthProvider {
  */
 export class AuthUtil implements IAuthProvider {
     public readonly profileName = VSCODE_EXTENSION_ID.amazonq
+    protected logger = getLogger('amazonqAuth')
+
     public readonly regionProfileManager: RegionProfileManager
 
     // IAM login currently not supported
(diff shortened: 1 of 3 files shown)


packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -402,58 +403,77 @@ export class AuthUtil implements IAuthProvider {
...
+            const token = await this.lspAuth.getSsoToken(
+                {
+                    kind: SsoTokenSourceKind.IamIdentityCenter,
+                    profileName: this.profileName,
+                },
+                false,
+                new CancellationTokenSource().token
+            )
+            if (token) {
+                getLogger().info(`codewhisperer: existing LSP auth connection found. Skipping migration`)
```
nkomonen-amazon, line 419: I would instead do `getLogger('authMigration').info(...` and drop the `codewhisperer:` prefix. My suggested example will have the message automatically prepended.

nkomonen-amazon (approved):

PR #7371, opened 2025-05-23
Add support for device_code flow by selecting a checkbox

j2clerck:
## Problem
In some cases, you want to authenticate Q extension from another computer (e.g. dev desktop) and VSCode extension automatically launch the authorization_code flow where the VSCode is launched.

## Solution
This PR implements a checkbox in the login UI to allow the user to select the device code flow (similar process as for remote extension) and then allows the user to copy the authentication URL and use it from the computer which can actually authenticate.

By default the checkbox is not checked.
After successful authentication, the checkbox is reset.

---

### DIFF
--- a/packages/core/src/shared/globalState.ts
+++ b/packages/core/src/shared/globalState.ts
@@ -50,6 +50,7 @@ export type globalKey =
     | 'aws.amazonq.customization.overrideV2'
     | 'aws.amazonq.regionProfiles'
     | 'aws.amazonq.regionProfiles.cache'
+    | 'aws.forceDeviceCodeFlow' // Toggle to force device code flow for authentication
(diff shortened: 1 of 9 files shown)


packages/core/src/shared/globalState.ts
```diff
@@ -50,6 +50,7 @@ export type globalKey =
     | 'aws.amazonq.customization.overrideV2'
     | 'aws.amazonq.regionProfiles'
     | 'aws.amazonq.regionProfiles.cache'
+    | 'aws.forceDeviceCodeFlow' // Toggle to force device code flow for authentication
```
justinmk3, line 53: Why store the flag permanently? It might be useful as a one-time choice, but storing it as a permanent choice raises the security risk of this feature.

PR #7421, opened 2025-06-03
feat(amazonq): Bundle LSP with the extension as fallback.  

leigaol:
## Problem
The LSP start failure because 
1. node binary is blocked because of firewall
2. chat UI js file is blocked because of firewall or anti virus
3. lsp js file is broken post download because of security mechanism

## Solution
1. Bundle the JS LSP with the amazonq package.
2. Re-start LSP wth the bundled JS files if and only if downloaded LSP does not work!
3. Use the VS Code vended node to start the bundled LSP.

This was tested by 
1. Generated the vsix, which is now 20MB.
2. Disconnect from internet, remove local LSP caches
3. Install the new vsix
4. Webview of chat should load.

also tested by manually corrupting the aws-lsp-codewhisperer.js

Limitations:
1. The indexing library function will not work because it is missing.
2. rg is not in the bundle

Ref: https://github.com/aws/aws-toolkit-jetbrains/pull/5772

---

### DIFF
--- a/packages/amazonq/src/lsp/activation.ts
+++ b/packages/amazonq/src/lsp/activation.ts
@@ -11,7 +11,6 @@ import { lspSetupStage, ToolkitError, messages, getLogger } from 'aws-core-vscod
 export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
     try {
         await lspSetupStage('all', async () => {
-            throw Error(`Q Install failed!`)
             const installResult = await new AmazonQLspInstaller().resolve()
             await lspSetupStage('launch', async () => await startLanguageServer(ctx, installResult.resourcePaths))
         })
--- a/packages/amazonq/src/lsp/activation.ts
+++ b/packages/amazonq/src/lsp/activation.ts
@@ -5,17 +5,27 @@
 
 import vscode from 'vscode'
 import { startLanguageServer } from './client'
-import { AmazonQLspInstaller } from './lspInstaller'
-import { lspSetupStage, ToolkitError, messages } from 'aws-core-vscode/shared'
+import { AmazonQLspInstaller, getBundledResourcePaths } from './lspInstaller'
+import { lspSetupStage, ToolkitError, messages, getLogger } from 'aws-core-vscode/shared'
 
 export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
     try {
         await lspSetupStage('all', async () => {
+            throw Error(`Q Install failed!`)
             const installResult = await new AmazonQLspInstaller().resolve()
             await lspSetupStage('launch', async () => await startLanguageServer(ctx, installResult.resourcePaths))
         })
     } catch (err) {
         const e = err as ToolkitError
-        void messages.showViewLogsMessage(`Failed to launch Amazon Q language server: ${e.message}`)
+        getLogger('amazonqLsp').warn(`Failed to start downloaded LSP ${e.message}. Starting with bundled LSP...`)
+        try {
+            await lspSetupStage('all', async () => {
+                await lspSetupStage('launch', async () => await startLanguageServer(ctx, getBundledResourcePaths(ctx)))
+            })
+        } catch (error) {
+            void messages.showViewLogsMessage(
+                `Failed to launch Amazon Q language server: ${(error as ToolkitError).message}`
+            )
+        }
     }
 }
--- a/packages/amazonq/src/lsp/activation.ts
+++ b/packages/amazonq/src/lsp/activation.ts
@@ -16,7 +16,7 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
         })
     } catch (err) {
         const e = err as ToolkitError
-        getLogger('amazonqLsp').warn(`Failed to start downloaded LSP ${e.message}. Starting with bundled LSP...`)
+        getLogger('amazonqLsp').warn(`Failed to start downloaded LSP, falling back to bundled LSP: ${e.message}`)
         try {
             await lspSetupStage('all', async () => {
                 await lspSetupStage('launch', async () => await startLanguageServer(ctx, getBundledResourcePaths(ctx)))
--- a/packages/amazonq/src/lsp/chat/webviewProvider.ts
+++ b/packages/amazonq/src/lsp/chat/webviewProvider.ts
@@ -45,7 +45,7 @@ export class AmazonQChatViewProvider implements WebviewViewProvider {
     ) {
         const lspDir = Uri.file(LanguageServerResolver.defaultDir())
         const dist = Uri.joinPath(globals.context.extensionUri, 'dist')
-        const bundledResources = Uri.joinPath(globals.context.extensionUri, 'resources', 'language-server')
+        const bundledResources = Uri.joinPath(globals.context.extensionUri, 'resources/language-server')
         let resourcesRoots = [lspDir, dist]
         if (this.mynahUIPath?.startsWith(globals.context.extensionUri.fsPath)) {
             getLogger('amazonqLsp').info(`Using bundled webview resources ${bundledResources.fsPath}`)
--- a/packages/amazonq/src/lsp/chat/webviewProvider.ts
+++ b/packages/amazonq/src/lsp/chat/webviewProvider.ts
@@ -44,8 +44,8 @@ export class AmazonQChatViewProvider implements WebviewViewProvider {
     ) {
         const lspDir = Uri.file(LanguageServerResolver.defaultDir())
         const dist = Uri.joinPath(globals.context.extensionUri, 'dist')
-
-        const resourcesRoots = [lspDir, dist]
+        const bundledResources = Uri.joinPath(globals.context.extensionUri, 'resources', 'language-server')
+        const resourcesRoots = [lspDir, dist, bundledResources]
 
         /**
          * if the mynah chat client is defined, then make sure to add it to the resource roots, otherwise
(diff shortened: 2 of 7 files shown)


packages/amazonq/src/lsp/lspInstaller.ts
```diff
@@ -54,3 +55,13 @@ export class AmazonQLspInstaller extends BaseLspInstaller.BaseLspInstaller<
 
     protected override downloadMessageOverride: string | undefined = 'Updating Amazon Q plugin'
 }
+
+export function getBundledResourcePaths(ctx: vscode.ExtensionContext): AmazonQResourcePaths {
+    const assetDirectory = vscode.Uri.joinPath(ctx.extensionUri, 'resources', 'language-server').fsPath
+    return {
+        lsp: path.join(assetDirectory, 'servers', 'aws-lsp-codewhisperer.js'),
+        node: process.execPath,
+        ripGrep: '',
```
rli, line 64: not sure if this is ok

scripts/lspArtifact.ts
```diff
@@ -0,0 +1,174 @@
...
+                                            downloadReject(err)
+                                        })
+                                    })
+                                    .on('error', (err) => {
+                                        fs.unlink(tempFilePath, () => {})
+                                        downloadReject(err)
+                                    })
+                            })
+
+                            console.log(`Verifying hash for ${fileName}...`)
```
andrewyuq, line 134: should switch to logger?

leigaol: this is inside the package script so it should do console.log

packages/amazonq/src/lsp/lspInstaller.ts
leigaol: I don't find a way to bundle cross platform ripGrep in VSC as this is anti pattern. VSC expected per platform vsixes. To do that properly, need to split the extension release

rli: i think arnab said he was ok with not including ripgrep

leigaol: Let me see if it throws or force the extension to quit

leigaol: Updated. We use the rg in the PATH env var, if rg is not found, agentic chat will prompt user to install it

scripts/lspArtifact.ts
```diff
@@ -0,0 +1,174 @@
...
+    if (!fs.existsSync(dirPath)) {
+        await fs.promises.mkdir(dirPath, { recursive: true })
+    }
+}
+
+export async function downloadLanguageServer(): Promise<void> {
+    const tempDir = path.join(os.tmpdir(), 'amazonq-download-temp')
+    const resourcesDir = path.join(__dirname, '../packages/amazonq/resources/language-server')
+
+    // clear previous cached language server
```
rli, line 60: is this only done on release packaging? ideally regular dev loop will not require constantly redownloading

rli (approved):

scripts/lspArtifact.ts
leigaol: only happens when `npm run package` yes

packages/amazonq/src/lsp/activation.ts
```diff
@@ -16,6 +16,15 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
         })
     } catch (err) {
         const e = err as ToolkitError
-        void messages.showViewLogsMessage(`Failed to launch Amazon Q language server: ${e.message}`)
+        getLogger('amazonqLsp').warn(`Failed to start downloaded LSP ${e.message}. Starting with bundled LSP...`)
```
justinmk3: randomly mentioning an unquoted value in the middle of a message, is hard to read. and as always, i suggest putting the value at the end, when possible.
```suggestion
        getLogger('amazonqLsp').warn(`Failed to start downloaded LSP, falling back to bundled LSP: ${e.message}`)
```

packages/amazonq/src/lsp/activation.ts
```diff
@@ -16,6 +16,15 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
+        try {
+            await lspSetupStage('all', async () => {
+                await lspSetupStage('launch', async () => await startLanguageServer(ctx, getBundledResourcePaths(ctx)))
+            })
+        } catch (error) {
```
justinmk3, line 24: the fallback logic should live in the resolver, just like all the other fallback cases. it doesn't make sense to add a fallback at this level.

this fallback is no different than the other fallback cases in the resolver. it's just yet another fallback scenario.

packages/amazonq/src/lsp/chat/webviewProvider.ts
```diff
@@ -44,9 +45,12 @@ export class AmazonQChatViewProvider implements WebviewViewProvider {
     ) {
         const lspDir = Uri.file(LanguageServerResolver.defaultDir())
         const dist = Uri.joinPath(globals.context.extensionUri, 'dist')
-
-        const resourcesRoots = [lspDir, dist]
-
+        const bundledResources = Uri.joinPath(globals.context.extensionUri, 'resources', 'language-server')
```
justinmk3: it is never necessary to join() literals. "/" slashes work just fine on all platforms.
```suggestion
        const bundledResources = Uri.joinPath(globals.context.extensionUri, 'resources/language-server')
```

packages/amazonq/src/lsp/chat/webviewProvider.ts
```diff
@@ -44,9 +45,12 @@ export class AmazonQChatViewProvider implements WebviewViewProvider {
...
+        let resourcesRoots = [lspDir, dist]
+        if (this.mynahUIPath?.startsWith(globals.context.extensionUri.fsPath)) {
+            getLogger('amazonqLsp').info(`Using bundled webview resources ${bundledResources.fsPath}`)
```
justinmk3, line 51: why is this decision being made here instead of in `LanguageServerResolver`

scripts/package.ts
```diff
@@ -155,6 +156,12 @@ function main() {
         }
 
         nodefs.writeFileSync(packageJsonFile, JSON.stringify(packageJson, undefined, '    '))
+
+        // add language server bundle
+        if (packageJson.name === 'amazon-q-vscode') {
+            await downloadLanguageServer()
```
justinmk3, line 162: this assumes a bunch of unspecified side-effects. to make the assumptions explicit, might want to validate things here (such as the existence of various files, etc). the validation logic could live in lspAritfact.ts, and this could just call `validate()` here.

for example, if the download or unzip partially fails, will the "package" step actually notice that or just produce a silently broken vsix ?

I see `verifyFileHash`, but that doesn't confirm that the final result of `zip.extractAllTo()` actually did put things in the expected places.

packages/amazonq/src/lsp/activation.ts
leigaol: No. In the case of a firewall/anti virus, the resolver believes it successfully finished downloading and it then proceed to start the language server with the download path, but at that moment anti virus kicked in and broke it. 

This outmost try catch can handle such cases.

packages/amazonq/src/lsp/chat/webviewProvider.ts
leigaol: Same reason as I mentioned above. When it failed to launch, there are cases when the anti virus modified/removed the downloaded artifacts later on. The `LanguageServerResolver` does not throw, it indeed successfully unzipped those artifacts.

leigaol: If `LanguageServerResolver` does not throw, I will not be able to resolve to bundled artifact

scripts/package.ts
leigaol: This line is just to download the LSP when doing `npm run package -w packages/amazonq.

> for example, if the download or unzip partially fails, will the "package" step actually notice that or just produce a silently broken vsix ?

Package step will fail because it throws. 


>I see verifyFileHash, but that doesn't confirm that the final result of zip.extractAllTo() actually did put things in the expected places.

Let me throw when the files are not at expected location

leigaol: fixed

packages/amazonq/src/lsp/activation.ts
justinmk3: > the resolver believes it successfully finished downloading and it then proceed to start the language server with the download path

isn't that a bug in the resolver that should be fixed / redesigned ?

packages/amazonq/src/lsp/chat/webviewProvider.ts
justinmk3: > When it failed to launch, there are cases when the anti virus modified/removed the downloaded artifacts later on. The `LanguageServerResolver` does not throw, it indeed successfully unzipped those artifacts.

Can it do a step that verifies the contents after unzipping? I.e. whatever causes the AV to quarantine the artifacts, the resolver should do itself before continuing.

scripts/lspArtifact.ts
```diff
@@ -0,0 +1,181 @@
...
+        }
+    } catch (e) {
+        throw Error(`Failed to clean up language server ${resourcesDir}`)
+    }
+
+    await ensureDirectoryExists(tempDir)
+    await ensureDirectoryExists(resourcesDir)
+
+    return new Promise((resolve, reject) => {
+        const manifestUrl = 'https://aws-toolkit-language-servers.amazonaws.com/qAgenticChatServer/0/manifest.json'
```
justinmk3, line 73: @leigaol Future note: this could be a source of confusion for cases like https://github.com/aws/aws-toolkit-vscode/commit/756c286c8728b5c5434971ecd58a96a1c56a992b when an alpha manifest is used. Unless this location is also updated.

PR #7422, opened 2025-06-04
fix(amazonq): Add proxy configuration support with SSL Cert Validation

tsmithsz:
## Solution

---

- Treat all work as PUBLIC. Private `feature/x` branches will not be squash-merged at release time.
- Your code changes must meet the guidelines in [CONTRIBUTING.md](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#guidelines).
- License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7426, opened 2025-06-04
fix(amazonq): Add proxy configuration support with SSL Cert Validation

tsmithsz:
## Problem
Amazon Q For VS Code has lost proxy support following the release of Agentic Chat (1.63.0) due to a fundamental architectural change. Previously, API calls were executed in VS Code's Extension Host process, which automatically handled proxy settings. Now, these calls are made through a separate Flare Language Server process that doesn't inherit VS Code's proxy configurations.

Key Issues:
- VS Code's built-in proxy handling doesn't propagate to the Flare process
- Environment variables required for proxy support aren't being passed to the Flare Language Server

## Solution
- Enabled Flare's experimental proxy support by setting `EXPERIMENTAL_HTTP_PROXY_SUPPORT=true`. See: https://github.com/aws/language-server-runtimes/blob/main/runtimes/docs/proxy.md
- Implemented `ProxyUtil` class to configure proxy settings before Flare Language Server starts
- Added `amazonQ.proxy.certificateAuthority` setting, allowing customers to conveniently set the path to their custom Certificate Authority (CA) .pem file directly in VS Code settings. This eliminates the need to set the `NODE_EXTRA_CA_CERTS` environment variable via CLI, providing a more user-friendly way to configure custom CA certificates for proxy connections.
- Reads VS Code's `http.proxy` setting and sets corresponding environment variables
- Configures certificate paths for both system and custom CA bundles

TODO: Potentially update [public documentation](https://docs.aws.amazon.com/toolkit-for-visual-studio/latest/user-guide/general-troubleshoot.html#general-troubleshoot-firewall) with new settings field

Screenshot 2025-06-04 at 4 45 44 PM

## Testing

Tested using [mitmproxy](https://docs.mitmproxy.org/stable/overview/installation/)

### Scenario 1: Explicit HTTP Proxy

- First test with adding the proxy to the VSC settings and disabling `EXPERIMENTAL_HTTP_PROXY_SUPPORT` (Request fails as expected)
- Then test  with adding the proxy to the VSC settings and enabling `EXPERIMENTAL_HTTP_PROXY_SUPPORT` (Request succeeds as expected due to automatically locating the certificate store from the OS)

https://github.com/user-attachments/assets/38f3fc28-5ff1-4c5e-b303-222399a5357f

### Scenario 1: Transparent HTTP Proxy

- Turn on proxy
-  First test with disabled `EXPERIMENTAL_HTTP_PROXY_SUPPORT` and no explicit HTTP proxy config added (Request fails as expected)
- Then test with enabled `EXPERIMENTAL_HTTP_PROXY_SUPPORT` and no explicit HTTP proxy config added (Request succeeds as expected due to automatically locating the certificate store from the OS)

https://github.com/user-attachments/assets/6d54d98c-683a-430d-8f65-f8f7e7dc2ad6

---

### DIFF
--- a/packages/core/package.nls.json
+++ b/packages/core/package.nls.json
@@ -98,7 +98,7 @@
     "AWS.configuration.description.amazonq.workspaceIndexIgnoreFilePatterns": "File patterns to ignore when indexing your workspace files",
     "AWS.configuration.description.amazonq.workspaceIndexCacheDirPath": "The path to the directory that contains the cache of the index of your workspace files",
     "AWS.configuration.description.amazonq.ignoredSecurityIssues": "Specifies a list of code issue identifiers that Amazon Q should ignore when reviewing your workspace. Each item in the array should be a unique string identifier for a specific code issue. This allows you to suppress notifications for known issues that you've assessed and determined to be false positives or not applicable to your project. Use this setting with caution, as it may cause you to miss important security alerts.",
-    "AWS.command.apig.copyUrl": "Copy URL",
+    "AWS.configuration.description.amazonq.proxy.certificateAuthority": "Path to a Certificate Authority (PEM file) for SSL/TLS verification when using a proxy.",
     "AWS.command.apig.invokeRemoteRestApi": "Invoke in the cloud",
     "AWS.command.apig.invokeRemoteRestApi.cn": "Invoke on Amazon",
     "AWS.appBuilder.explorerTitle": "Application Builder",
--- a/packages/core/src/shared/utilities/proxyUtil.ts
+++ b/packages/core/src/shared/utilities/proxyUtil.ts
@@ -39,7 +39,7 @@ export class ProxyUtil {
     private static getProxyConfiguration(): ProxyConfig {
         const httpConfig = vscode.workspace.getConfiguration('http')
         const proxyUrl = httpConfig.get<string>('proxy')
-        this.logger.debug(`Proxy URL: ${proxyUrl}`)
+        this.logger.debug(`Proxy URL Setting in VSCode Settings: ${proxyUrl}`)
 
         const amazonQConfig = vscode.workspace.getConfiguration('amazonQ')
         const proxySettings = amazonQConfig.get<{
@@ -60,6 +60,8 @@ export class ProxyUtil {
 
         // Always enable experimental proxy support for better handling of both explicit and transparent proxies
         process.env.EXPERIMENTAL_HTTP_PROXY_SUPPORT = 'true'
+        // Add OpenSSL certificate store support
+        process.env.NODE_OPTIONS = '--use-openssl-ca'
 
         // Set proxy environment variables
         if (proxyUrl) {
--- a/packages/core/src/shared/utilities/proxyUtil.ts
+++ b/packages/core/src/shared/utilities/proxyUtil.ts
@@ -1,0 +1,78 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import vscode from 'vscode'
+import { getLogger } from '../logger/logger'
+
+interface ProxyConfig {
+    proxyUrl: string | undefined
+    certificateAuthority: string | undefined
+}
+
+/**
+ * Utility class for handling proxy configuration
+ */
+export class ProxyUtil {
+    private static readonly logger = getLogger('proxyUtil')
+
+    /**
+     * Sets proxy environment variables based on VS Code settings for use with the Flare Language Server
+     *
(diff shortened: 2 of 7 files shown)


packages/core/package.nls.json
```diff
@@ -98,7 +98,7 @@
     "AWS.configuration.description.amazonq.workspaceIndexIgnoreFilePatterns": "File patterns to ignore when indexing your workspace files",
     "AWS.configuration.description.amazonq.workspaceIndexCacheDirPath": "The path to the directory that contains the cache of the index of your workspace files",
     "AWS.configuration.description.amazonq.ignoredSecurityIssues": "Specifies a list of code issue identifiers that Amazon Q should ignore when reviewing your workspace. Each item in the array should be a unique string identifier for a specific code issue. This allows you to suppress notifications for known issues that you've assessed and determined to be false positives or not applicable to your project. Use this setting with caution, as it may cause you to miss important security alerts.",
-    "AWS.command.apig.copyUrl": "Copy URL",
+    "AWS.configuration.description.amazonq.proxy.certificateAuthority": "Path to a Certificate Authority (PEM file) for SSL/TLS verification when using a proxy.",
```
justinmk3, line 101: same question as https://github.com/aws/aws-toolkit-vscode/pull/7422#discussion_r2127251985

also, why create a new pr instead of updating the existing one

tsmithsz: Had some issues with my git setup. 


Added `amazonQ.proxy.certificateAuthority` setting, allowing customers to conveniently set the path to their custom Certificate Authority (CA) .pem file directly in VS Code settings. This eliminates the need to set the `NODE_EXTRA_CA_CERTS` environment variable via CLI, providing a more user-friendly way to configure custom CA certificates for proxy connections.


The [flare proxy configuration](https://github.com/aws/language-server-runtimes/blob/main/runtimes/docs/proxy.md#1-explicit-http-proxy) states that: `If proxy is using SSL and URL starts with https://, then path to proxy server root certificate must be set in NODE_EXTRA_CA_CERTS. At the moment, other methods of passing certificate are not supported and won't work.`

This is just a fallback

justinmk3: Seems like vscode doesn't have a way to configure a .pem file. Also [this](https://code.visualstudio.com/docs/setup/network?from_column=20423&from=20423#_legacy-proxy-server-support) is an interesting note.

Because of all the numerous ways of setting proxy/cert settings, at each layer, it is worth trying to avoid adding yet another layer that has its own proxy/cert settings. However it looks like vscode's proxy support still isn't mature enough, so this setting is needed.

packages/core/src/shared/utilities/proxyUtil.ts
justinmk3, line 1: maybe this should live in `shared/net/proxy.ts`. it's likely that other network-related shared things will be needed in the future. incidentally, the stuff in `shared/resourcefetcher/` could be moved to `shared/net/` ...

justinmk3 (approved):

packages/core/src/shared/utilities/proxyUtil.ts
tsmithsz: Will look into moving these into a `net` subfolder

erezo9:
I know its for visual studio
https://docs.aws.amazon.com/toolkit-for-visual-studio/latest/user-guide/general-troubleshoot.html#general-troubleshoot-firewall

but possibly adding those env variables as stated

> 
> Create Windows Environment Variables for the following keys.
> 
> NODE_OPTIONS = --use-openssl-ca
> 
> NODE_EXTRA_CA_CERTS = Path/To/Corporate/Certs - Which is in this PR

tsmithsz:
> I know its for visual studio https://docs.aws.amazon.com/toolkit-for-visual-studio/latest/user-guide/general-troubleshoot.html#general-troubleshoot-firewall
> 
> but possibly adding those env variables as stated
> 
> > Create Windows Environment Variables for the following keys.
> > NODE_OPTIONS = --use-openssl-ca
> > NODE_EXTRA_CA_CERTS = Path/To/Corporate/Certs - Which is in this PR

Added

erezo9:
> > I know its for visual studio https://docs.aws.amazon.com/toolkit-for-visual-studio/latest/user-guide/general-troubleshoot.html#general-troubleshoot-firewall
> > but possibly adding those env variables as stated
> > > Create Windows Environment Variables for the following keys.
> > > NODE_OPTIONS = --use-openssl-ca
> > > NODE_EXTRA_CA_CERTS = Path/To/Corporate/Certs - Which is in this PR
> 
> Added

Just an fyi
I don’t work for Amazon or have anything to do with this repo
I’m a user who’s experience the issue as you and while thrusting to resolve this I found that document

It was just a recommendation since I saw your pr

PR #7453, opened 2025-06-09
feat(lambda): Add handler for upcoming Lambda URI

rhamilt:
## Problem
The Lambda team plans to add a URI handler to act as an entrypoint to the toolkit from a URL. This will be fully available in an upcoming toolkit release, but customers who don't have that version will not get the benefit of the handler.

## Solution
Add a handler ahead of time telling customers to update to the latest version. This way, customers who have an older version with this change will be notified that the functionality they are trying to use is in a newer version. This should help customers from becoming confused when they open a link and an older version says it is not handled.

---

### DIFF
--- a/packages/core/src/lambda/uriHandlers.ts
+++ b/packages/core/src/lambda/uriHandlers.ts
@@ -1,0 +1,39 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode'
+import * as nls from 'vscode-nls'
+
+import { SearchParams } from '../shared/vscode/uriHandler'
+import { showMessage } from '../shared/utilities/messages'
+import globals from '../shared/extensionGlobals'
+
+const localize = nls.loadMessageBundle()
+
+export function registerLambdaUriHandler() {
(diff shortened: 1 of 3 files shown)


packages/core/src/lambda/uriHandlers.ts
```diff
@@ -0,0 +1,39 @@
...
+            'warn',
+            localize(
+                'AWS.lambda.uriUnavailable',
+                'The URI handler you are attempting to open is not handled in this version of the toolkit, try installing the latest version'
+            )
+        )
+    }
+
+    return vscode.Disposable.from(
+        globals.uriHandler.onPath('/lambda/load-function', openFunctionHandler, parseOpenParams)
```
justinmk3, line 31: 👍 thanks for re-using the existing module and patterns.

packages/core/src/lambda/uriHandlers.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const localize = nls.loadMessageBundle()
+
+export function registerLambdaUriHandler() {
+    async function openFunctionHandler(params: ReturnType<typeof parseOpenParams>) {
+        await showMessage(
+            'warn',
+            localize(
+                'AWS.lambda.uriUnavailable',
+                'The URI handler you are attempting to open is not handled in this version of the toolkit, try installing the latest version'
```
justinmk3, line 21: I assume this is temporary...

PR #7455, opened 2025-06-09
bug: ISSUE-7452: Fix EC2 Workspace Connect for Any OS

arminanton:
## Problem

EC2 Workspace Connect only recognized Amazon Linux and Ubuntu. Other Linux distributions failed before users could select the SSH account. The connection code also assumed that every account lived under `/home/<username>`, did not create a missing `.ssh/authorized_keys`, and opened `/` as the remote folder.

## Solution

1. Keep the existing defaults for Amazon Linux and Ubuntu.
2. Prompt for an SSH username when Systems Manager reports another operating system.
3. Validate the username before using it in a remote shell command.
4. Resolve the selected account's real home directory from the instance through Systems Manager and fail closed if it is missing or invalid.
5. Create the account's `.ssh` directory and `authorized_keys` file when needed, set ownership to the selected account, and enforce permissions `700` and `600`.
6. Remove the previous Toolkit key with a quoted, GNU and BSD compatible `sed -i.bak` command, remove the backup, and append the new public key with shell safe quoting.
7. Open the resolved home directory instead of `/` in VS Code Remote SSH.
8. Add Systems Manager command output support through `GetCommandInvocation`.

## Tests

1. EC2 model unit suite: 26 passing. This covers Amazon Linux, Ubuntu, an unknown OS with a custom user and nonstandard home, missing homes, cancellation, unsafe usernames, SSH file setup, permissions, and portable key cleanup.
2. Core lint suite: 3 passing, including ESLint and git secrets checks.
3. Core TypeScript compilation: passed.
4. Core webpack build: passed with the repository's existing optional dependency warnings.

## Validation scope

The automated tests mock Systems Manager responses and execute the generated key cleanup command locally. No live EC2 instance matrix was available in this local remediation run. A target without `getent`, `sed`, or the permissions required by `AWS-RunShellScript` fails closed instead of writing to a guessed path.

## License

I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/packages/core/src/awsService/ec2/model.ts
+++ b/packages/core/src/awsService/ec2/model.ts
@@ -43,10 +43,10 @@ export interface Ec2RemoteEnv extends VscodeRemoteConnection {
     ssmSession: StartSessionResponse
 }
 
-export type Ec2OS = 'Amazon Linux' | 'Ubuntu' | 'macOS'
 interface RemoteUser {
-    os: Ec2OS
+    os: string
     name: string
+    home: string
 }
 
 export class Ec2Connecter implements vscode.Disposable {
@@ -215,7 +215,7 @@ export class Ec2Connecter implements vscode.Disposable {
             await startVscodeRemote(
                 remoteEnv.SessionProcess,
                 remoteEnv.hostname,
-                '/',
+                remoteUser.home,
                 remoteEnv.vscPath,
                 remoteUser.name
             )
@@ -298,14 +298,9 @@ export class Ec2Connecter implements vscode.Disposable {
     }
 
     /** Removes old key(s) that we added to the remote ~/.ssh/authorized_keys file. */
-    public async tryCleanKeys(
-        instanceId: string,
-        hintComment: string,
-        hostOS: Ec2OS,
-        remoteAuthorizedKeysPath: string
-    ) {
+    public async tryCleanKeys(instanceId: string, hintComment: string, remoteAuthorizedKeysPath: string) {
         try {
-            const deleteExistingKeyCommand = getRemoveLinesCommand(hintComment, hostOS, remoteAuthorizedKeysPath)
+            const deleteExistingKeyCommand = getRemoveLinesCommand(hintComment, remoteAuthorizedKeysPath)
             await this.sendCommandAndWait(instanceId, deleteExistingKeyCommand)
         } catch (e) {
             getLogger().warn(`ec2: failed to clean keys: %O`, e)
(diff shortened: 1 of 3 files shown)


arminanton:
Might need to change a bit [here](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/test/awsService/ec2/model.test.ts#L208-L216) since now we have an input/prompt when the OS does not match Amazon Linux or Ubuntu.

arminanton:
Tested the `stable` vsix generated and worked fine.

Prompt of the user desired to be used:
image

Tested `ssm-user` and it failed, which is the user we use to usually access the terminal in AWS Toolkit for EC2 instances. Then, I tried another user that we use for testing our running services in the instance, and it worked:
image

One thing I did noticed though is that it connected to `/` directly instead of asking which path it wanted to connect, or instead of using `/home/${user}` as default path.

I also noticed that the connection is not that steady, it loses connection after a few seconds being connected to the instance, and suggests to reload the window (just like a connection lost). I suspect it might be because we are in `/` and it might be trying to sync or analyze all the files and projects present.

Not sure if we need to do anything else to keep it more steady.

When I try to connect using regular ssm user like `ssm-user` it gives me this error in logs:
```
2025-06-09 15:25:17.940 [warning] ec2: failed to clean keys: 'Error: Failed in sending command to target i-1234567890abcdefg\n' +
  `\t -> Error: {"state":"FAILURE","reason":{"$metadata":{"httpStatusCode":200,"requestId":"(...)","attempts":1,"totalRetryDelay":0},"CloudWatchOutputConfig":{"CloudWatchLogGroupName":"","CloudWatchOutputEnabled":false},"CommandId":"(...)","Comment":"","DocumentName":"AWS-RunShellScript","DocumentVersion":"$DEFAULT","ExecutionElapsedTime":"PT0.005S","ExecutionEndDateTime":"2025-06-09T22:25:12.570Z","ExecutionStartDateTime":"2025-06-09T22:25:12.570Z","InstanceId":"i-1234567890abcdefg","PluginName":"aws:runShellScript","ResponseCode":2,"StandardErrorContent":"sed: can't read /home/ssm-user/.ssh/authorized_keys: No such file or directory\\nfailed to run commands: exit status 2","StandardErrorUrl":"","StandardOutputContent":"","StandardOutputUrl":"","Status":"Failed","StatusDetails":"Failed"},"observedResponses":{"200: OK":2}}`
2025-06-09 15:25:23.910 [error] aws.ec2.openRemoteConnection: Error: Failed in sending command to target i-1234567890abcdefg
	 -> Error: {"state":"FAILURE","reason":{"$metadata":{"httpStatusCode":200,"requestId":"(...)","attempts":1,"totalRetryDelay":0},"CloudWatchOutputConfig":{"CloudWatchLogGroupName":"","CloudWatchOutputEnabled":false},"CommandId":"(...)","Comment":"","DocumentName":"AWS-RunShellScript","DocumentVersion":"$DEFAULT","ExecutionElapsedTime":"PT0.003S","ExecutionEndDateTime":"2025-06-09T22:25:18.557Z","ExecutionStartDateTime":"2025-06-09T22:25:18.557Z","InstanceId":"i-1234567890abcdefg","PluginName":"aws:runShellScript","ResponseCode":1,"StandardErrorContent":"/var/lib/amazon/ssm/i-1234567890abcdefg/document/orchestration/(...)/awsrunShellScript/0.awsrunShellScript/_script.sh: line 1: /home/ssm-user/.ssh/authorized_keys: No such file or directory\nfailed to run commands: exit status 1","StandardErrorUrl":"","StandardOutputContent":"","StandardOutputUrl":"","Status":"Failed","StatusDetails":"Failed"},"observedResponses":{"200: OK":2}}
```
From this log, we can extract `sed: can't read /home/ssm-user/.ssh/authorized_keys: No such file or directory\\nfailed to run commands: exit status 2`, which means that the file or folder does not exist.

Out of curiosity, I tried using `root` user, like it is done for ECS tasks, and it returned the same error than `ssm-user`. Perhaps what is missing is to properly setup the `/home/${user}/.ssh/authorized_keys` for these said chosen users, which will then make it succeed as desired.

packages/core/src/awsService/ec2/model.ts
```diff
@@ -345,6 +344,13 @@ export class Ec2Connecter implements vscode.Disposable {
             return { name: 'ubuntu', os }
         }
 
+        const input = await vscode.window.showInputBox({
+            prompt: `Unrecognized OS "${os}". Please enter the username to use for instance ${instanceId}:`,
```
justinmk3:
```suggestion
            prompt: `Unrecognized OS "${os}". Enter the username for instance ${instanceId}:`,
```

justinmk3:
> Perhaps what is missing is to properly setup the `/home/${user}/.ssh/authorized_keys` for these said chosen users

That's definitely an important part of the logic.



> it connected to `/` directly instead of asking which path it wanted to connect, or instead of using `/home/${user}` as default path.

It looks like we always tell vscode to choose / as the target dir: https://github.com/aws/aws-toolkit-vscode/blob/d8d08e28052ed7583702c33b8e66a10d3db8fb5f/packages/core/src/awsService/ec2/model.ts#L215-L218

I think we depend on some later step to detect $HOME, or something like that.

arminanton:
> I think we depend on some later step to detect $HOME, or something like that.

We might want to double check that logic, as we aren't required to specify any path when connecting vscode to an instance. 
Like, if I add a new host manually to vscode and use remote ssh to connect to it, it would connect to the instance without having any path open at all. That saves us from vscode attempts to explore all paths generically (as it assumes `/` as a regular project folder). 
Perhaps that path parameter in the actual caller is optional.

> For the ssh authorized keys, might want to explore and improve this part:

https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/awsService/ec2/model.ts#L321-L336


> Unit Tests failed: https://d1ihu6zq92vp9p.cloudfront.net/0df37705-e8d7-47be-b4ea-52e36cb34a6c/report.html

I saw the errors reported from the unit test run. I will work on them and push a fix soon.


> ECS portion

In the meantime, if anyone wants to explore the option to connect ECS to vscode.

arminanton:
Rebased this branch onto the current `master` tip `6c4d294b0f228e84ce8e9cd76d646a2f6d35661d`. The feature patch is unchanged by stable patch ID.

Fresh local verification on the rebased head `3610c9c600d0c905002c4adc3f5eaf098cc2fa95`:

* focused EC2 workspace connect tests completed with result code 0
* core lint passed
* TypeScript compilation passed
* webpack development build passed
* `git diff --check` passed

The upstream CI run was created as `action_required` with no jobs because this external fork workflow still needs maintainer approval: https://github.com/aws/aws-toolkit-vscode/actions/runs/32995951921

Could a maintainer approve that workflow run so the repository's required matrix can execute on the current head?

PR #7479, opened 2025-06-12
fix(amazonq): bundle flare version comparison using numbers 

leigaol:
## Problem
String comparison says "1.11.0" < "1.9.0". We should compare numbers.

## Solution

---


bryceitoc9 (approved):

leigaol:
> Why not use `semver.compare`?

I was under the impression that build script should be standalone. But then I tried to import semver and it works. 
This code is not checked in the extension code so I cannot use logger.info() etc..

PR #7505, opened 2025-06-17
fix(amazonq): allow node to inherit proxy settings from VSC

samgst-amazon:
## Problem
With the introduction of agentic chat, Amazon Q Developer now runs its language server (Flare) in a separate Node.js process outside of VS Code's control. This creates network connectivity issues in corporate environments:

The Node.js process doesn't inherit VS Code's proxy settings, causing connection failures behind corporate proxies

Corporate SSL certificates aren't trusted by the Node.js process, resulting in certificate validation errors

Proxy authentication (including Integrated Windows Authentication) doesn't work automatically

These issues prevent Amazon Q Developer from functioning in enterprise environments with strict network policies.

## Solution

This PR enables the Flare Language Server to inherit VS Code's Electron proxy settings and certificate trust store:

Added a getElectronSettings() function that extracts:

Proxy rules from Electron's session

Proxy bypass rules for local addresses

Trusted certificates from Electron's certificate store

Modified createServerOptions() to:

Apply Electron's proxy settings to the Node.js process environment

Pass trusted certificates to Node.js via NODE_EXTRA_CA_CERTS

Preserve proxy bypass rules via NO_PROXY

Simplified the proxy environment setup in baseLspInstaller.ts to focus on SSL verification settings

This approach ensures that the Node.js process uses the same proxy configuration and certificate trust store as VS Code, making Amazon Q Developer work transparently in corporate environments without requiring manual configuration.
---

---

### DIFF
--- a/packages/core/src/shared/lsp/utils/platform.ts
+++ b/packages/core/src/shared/lsp/utils/platform.ts
@@ -120,7 +120,7 @@ export async function getVSCodeSettings(): Promise<{ proxyUrl?: string; certific
             }
 
             const certPath = join(tempDir, 'vscode-ca-certs.pem')
-            const certContent = allCerts.join('')
+            const certContent = allCerts.join('\n')
 
             nodefs.writeFileSync(certPath, certContent)
             result.certificatePath = certPath
--- a/packages/core/src/shared/lsp/utils/platform.ts
+++ b/packages/core/src/shared/lsp/utils/platform.ts
@@ -8,7 +8,11 @@ import { Logger, getLogger } from '../../logger/logger'
 import { ChildProcess } from '../../utilities/processUtils'
 import { waitUntil } from '../../utilities/timeoutUtils'
 import { isDebugInstance } from '../../vscode/env'
+import { tmpdir } from 'os'
+import { join } from 'path'
+import * as fs from 'fs'
 import * as vscode from 'vscode'
+import * as tls from 'tls'
 
 export function getNodeExecutableName(): string {
     return process.platform === 'win32' ? 'node.exe' : 'node'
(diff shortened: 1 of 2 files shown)


packages/core/src/shared/lsp/utils/platform.ts
```diff
@@ -81,26 +86,121 @@ export async function validateNodeExe(nodePath: string[], lsp: string, args: str
...
+/**
+ * Gets proxy settings and certificates from VS Code
+ */
+export function getVSCodeSettings(): { proxyUrl?: string; certificatePath?: string } {
+    const result: { proxyUrl?: string; certificatePath?: string } = {}
+    const logger = getLogger('amazonqLsp')
+
+    try {
+        // Check if user already has NODE_EXTRA_CA_CERTS set
+        const userCerts = process.env.NODE_EXTRA_CA_CERTS
```
rli: not sure if this works correctly on mac -- see discussion in jetbrains/eclipse where macOS does not inherit the console environment due to how app launching works

packages/core/src/shared/lsp/utils/platform.ts
```diff
@@ -81,26 +86,121 @@ export async function validateNodeExe(nodePath: string[], lsp: string, args: str
...
+        const httpConfig = vscode.workspace.getConfiguration('http')
+        const proxy = httpConfig.get<string>('proxy')
+        if (proxy) {
+            result.proxyUrl = proxy
+            logger.info(`Using proxy from VS Code settings: ${proxy}`)
+        }
+
+        try {
+            // @ts-ignore - we need this function to access certs
+            const certs = tls.getCACertificates()
```
rli: not sure how this would work. the `default` behavior is to use the node bundled CAs, which is not what we want. should this be explicitly `system`? or does this just magically work because VSC is patching the return value?

packages/core/src/shared/lsp/utils/platform.ts
```diff
@@ -81,26 +86,121 @@ export async function validateNodeExe(nodePath: string[], lsp: string, args: str
...
+            Object.assign(processEnv, env)
+        }
+
+        // Get settings from VS Code
+        const settings = getVSCodeSettings()
+        const logger = getLogger('amazonqLsp')
+
+        // Add proxy settings to the Node.js process
+        if (settings.proxyUrl) {
+            processEnv.HTTPS_PROXY = settings.proxyUrl
```
rli, line 169: `HTTPS_PROXY` is the only needed one

PR #7511, opened 2025-06-17
fix: increase memory limit for amazon q

yueny2020:
## Solution

---

### DIFF
--- a/packages/amazonq/src/lsp/client.ts
+++ b/packages/amazonq/src/lsp/client.ts
@@ -64,6 +64,7 @@ export async function startLanguageServer(
         '--stdio',
         '--pre-init-encryption',
         '--set-credentials-encryption-key',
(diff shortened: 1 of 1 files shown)


packages/amazonq/src/lsp/client.ts
```diff
@@ -64,6 +64,7 @@ export async function startLanguageServer(
         '--stdio',
         '--pre-init-encryption',
         '--set-credentials-encryption-key',
```
leigaol, line 66: these are the flags for Flare. You need to add the flag for node binary itself.. If you cannot do it in argv, you can pass as environment variables

yueny2020: > You need to add the flag for node binary itself

could you please explain how to `add the flag for node binary itself`?

PR #7520, opened 2025-06-18
fix(amazonq): chatOptions are missing in q chat

laileni-aws:
## Problem
- Missing `chatOptions` like mcp icon, history icon, export icon in tab bar and  quickActions like `/clear` and `/help` in the input menu.
image

#### Root Cause 1:
- The `CHAT_OPTIONS` command is sometimes not properly received by the handleInboundMessage function in the chat client.
- The root issue was a race condition in the communication between the VSCode extension and the webview component. When the `onDidResolveWebview` event fired, the code was immediately trying to send a message to the webview, but in some cases, the webview wasn't fully ready to receive messages yet, causing the message to be lost.
#### Root Cause 2:
- If user opens a large repository(example: [JetBrains](https://github.com/aws/aws-toolkit-jetbrains)) in any IDE(VSC or JB or VS or Ecllipse), IDE takes some time to do indexing.
- This delay of indexing is stopping the webview to initialize and if the webview is delayed, the `chatOptions` like quickActions(`/help` or `/clear`) and tab bar icons are not sent because of webview delay.

## Ideal case:

- VSC Extension at `messages.ts`,  `chatOptions` are sent to the [webview using webview.postMessage() with the command CHAT_OPTIONS](https://github.com/aws/language-servers/blob/0cac52c3d037da8fc4403f030738256b07195e76/client/vscode/src/chatActivation.ts#L340-L348) and this is handled by the `handleInboundMessage` in `LS/chat.ts` file.

## Solution:

1. Listen to dedicated `aws/chat/ready` event, which ensures we only send messages when the UI is fully initialized and ready to process them.
2. The chat options are stored when initialized, and then sent when the `aws/chat/ready` event is received from the UI.
3. Added explicit try/catch around the message sending process with proper error logging.
4. Added clearer log messages to track the UI ready event and message sending process, which will make debugging easier if issues arise in the future.

---

### DIFF
--- a/packages/amazonq/src/lsp/chat/messages.ts
+++ b/packages/amazonq/src/lsp/chat/messages.ts
@@ -48,6 +48,7 @@ import {
     LINK_CLICK_NOTIFICATION_METHOD,
     LinkClickParams,
     INFO_LINK_CLICK_NOTIFICATION_METHOD,
+    READY_NOTIFICATION_METHOD,
     buttonClickRequestType,
     ButtonClickResult,
     CancellationTokenSource,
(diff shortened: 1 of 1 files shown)


kmile (approved):

packages/amazonq/src/lsp/chat/messages.ts
```diff
@@ -166,7 +164,32 @@ export function registerMessageListeners(
...
+                languageClient.info(`[VSCode Client] "aws/chat/ready" event is received, sending chat options`)
+                if (webview && pendingChatOptions) {
+                    try {
+                        await webview.postMessage({
+                            command: CHAT_OPTIONS,
+                            params: pendingChatOptions,
+                        })
+
+                        // Display a more readable representation of quick actions
+                        const quickActionCommands =
```
Randall-Jiang, line 180: For quick actions, are we rendering in language server or ide toolkit repo? Based on the discussion we had offline, I thought the quickActions module is implemented in language server?

packages/amazonq/src/lsp/chat/messages.ts
```diff
@@ -166,7 +164,32 @@ export function registerMessageListeners(
...
+                            pendingChatOptions?.quickActions?.quickActionsCommandGroups?.[0]?.commands || []
+                        const quickActionsDisplay = quickActionCommands.map((cmd: any) => cmd.command).join(', ')
```
ashishrp-aws, line 182: why do we need this? wouldn't webiew.postMessage handle this?

packages/amazonq/src/lsp/chat/messages.ts
laileni-aws: No this is for `/help` or `/clear` not for `/test`, `/dev` etc..

packages/amazonq/src/lsp/chat/messages.ts
laileni-aws: No this is for logging! at L185

ctlai95 (approved):

PR #7527, opened 2025-06-18
fix(toolkit): initialize compute region in get call with a default value to fix extension activation failure

parameja1:
## Problem
With updating Toolkit version to latest in SMUS CodeEditor Spaces, observing Toolkit throws an error as attached in screenshot below with error message
```
Attempted to get compute region without initializing
```
image

## Solution
- Get Compute Region seems to be invoked first before Initializing compute region is called
- hence solution is to start the async operation of initialize compute region in the background and set a default value, immediately to prevent future calls from throwing error in order to not fail the toolkit extension activation
- Tested with a local debug artifact in SMUS CodeEditor space, toolkit activation completed and working
- image


packages/core/src/shared/extensionUtilities.ts
```diff
@@ -370,9 +370,15 @@ export async function initializeComputeRegion(
 
 export function getComputeRegion(): string | undefined {
     if (computeRegion === notInitialized) {
-        throw new Error('Attempted to get compute region without initializing.')
```
justinmk3, line 373: By removing this, you are hiding a problem.

The error exists to ensure that order-of-operations is correct.  After your change, that guarantee will no longer be possible.

It seems convenient now, but is a short-term solution with long-term consequences.

parameja1: Thanks for the question, though the Error is removed in terms of order-of-operations the above implementation ensures ```initializeComputeRegion``` is being called within ```getComputeRegion``` to handle case where user may not set it correctly. Also default region being set follows same logic from initializeComputeRegion
- curious why not handle initialize region within get call above?

parameja1:
CR closed to address the issue with a different approach as suggested in the comments

PR #7544, opened 2025-06-23
fix(amazonq): migrate process.env proxy setting handling to proxyUtil.ts

samgst-amazon:
## Problem
Previous fix was separate from the existing proxyUtil.ts handling.

## Solution
Logic has been moved form platform.ts to proxyUtil.ts to have all process.env handling in one place. 
also added fetching for NO_PROXY settings from vsc
---

### DIFF
--- a/packages/core/src/shared/utilities/proxyUtil.ts
+++ b/packages/core/src/shared/utilities/proxyUtil.ts
@@ -90,6 +90,7 @@ export class ProxyUtil {
         if (!strictSSL) {
             process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
             this.logger.info('SSL verification disabled via VS Code settings')
+            return // No need to set CA certs when SSL verification is disabled
         }
 
         // Set certificate bundle environment variables if user configured
--- a/packages/core/src/shared/utilities/proxyUtil.ts
+++ b/packages/core/src/shared/utilities/proxyUtil.ts
@@ -5,9 +5,14 @@
 
 import vscode from 'vscode'
 import { getLogger } from '../logger/logger'
+import { tmpdir } from 'os'
+import { join } from 'path'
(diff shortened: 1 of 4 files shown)


rli (approved):

packages/core/src/shared/utilities/proxyUtil.ts
```diff
@@ -64,11 +78,60 @@ export class ProxyUtil {
...
+        if (noProxy) {
+            process.env.NO_PROXY = noProxy
+            this.logger.debug(`Set NO_PROXY environment variable: ${noProxy}`)
+        }
+
+        const strictSSL = config.proxyStrictSSL
+        // Handle SSL certificate verification
+        if (!strictSSL) {
+            process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
+            this.logger.info('SSL verification disabled via VS Code settings')
```
rli, line 92: if not strict, should we skip everything following this check?

samgst-amazon: You're right since after this node wont validate the certs anyways.

packages/core/src/shared/utilities/proxyUtil.ts
```diff
@@ -41,21 +46,30 @@ export class ProxyUtil {
...
+            proxyStrictSSL,
             certificateAuthority: proxySettings.certificateAuthority,
         }
     }
 
     /**
      * Sets environment variables based on proxy configuration
      */
-    private static setProxyEnvironmentVariables(config: ProxyConfig): void {
+    private static async setProxyEnvironmentVariables(config: ProxyConfig): Promise<void> {
```
tsmithsz, line 72: Do we need to update the proxyUtil to use the new env variables?: https://github.com/tsmithsz/language-server-runtimes/blob/main/runtimes/runtimes/util/standalone/proxyUtil.ts

PR #7548, opened 2025-06-24
fix(amazonq): fix for amazon q app initialization failure on sagemaker

ashishrp-aws:
## Problem
GlibC path execution failing for sagemaker instances due to changes made to support AL2.

## Solution
- Skip GlibC and using default for sagemaker.

---


packages/amazonq/src/extension.ts
```diff
@@ -126,12 +126,15 @@ export async function activateAmazonQCommon(context: vscode.ExtensionContext, is
...
     // This contains every lsp agnostic things (auth, security scan, code scan)
     await activateCodeWhisperer(extContext as ExtContext)
-    if (
+
+    // Always activate LSP for SageMaker environments
+    const shouldActivateLsp =
         (Experiments.instance.get('amazonqLSP', true) || Auth.instance.isInternalAmazonUser()) &&
-        (!isAmazonLinux2() || hasGlibcPatch())
-    ) {
+        (isSageMaker() || !isAmazonLinux2() || (await hasGlibcPatch()))
```
rli, line 133: it appears that the `isAmazonLinux2` check is incorrectly flagging ubuntu containers under AL2 hosts. i do not think we need to special case sagemaker for now

packages/amazonq/src/lsp/client.ts
```diff
@@ -53,11 +53,24 @@ import { InlineChatTutorialAnnotation } from '../app/inline/tutorials/inlineChat
 const localize = nls.loadMessageBundle()
 const logger = getLogger('amazonqLsp.lspClient')
 
-export const glibcLinker: string = process.env.VSCODE_SERVER_CUSTOM_GLIBC_LINKER || ''
-export const glibcPath: string = process.env.VSCODE_SERVER_CUSTOM_GLIBC_PATH || ''
+export async function hasGlibcPatch(): Promise<boolean> {
```
rli, line 56: what requires async here

packages/core/package.json
```diff
@@ -526,7 +526,7 @@
         "@aws-sdk/s3-request-presigner": "<3.731.0",
         "@aws-sdk/smithy-client": "<3.731.0",
         "@aws-sdk/util-arn-parser": "<3.731.0",
-        "@aws/mynah-ui": "^4.35.4",
+        "@aws/mynah-ui": "^4.35.6",
```
rli, line 529: mynah update should not be in this focused change

ashishrp-aws: sorry..previous commit sneaked in...let me correct and raise a new one

leigaol:
What is the root cause of this issue? If SageMaker does not have such env var, why is it going to the GLIBC patch code path

ashishrp-aws:
like richard said...that AL2 check `it appears that the isAmazonLinux2 check is incorrectly flagging ubuntu containers under AL2 hosts`

packages/amazonq/src/extension.ts
leigaol: what is the exact setup for SageMaker AL2 instances?

packages/core/src/shared/extensionUtilities.ts
```diff
@@ -176,6 +176,18 @@ export function isCloud9(flavor: 'classic' | 'codecatalyst' | 'any' = 'any'): bo
  * @returns true if the current system is SageMaker(SMAI or SMUS)
  */
 export function isSageMaker(appName: 'SMAI' | 'SMUS' = 'SMAI'): boolean {
+    // Check for SageMaker-specific environment variables first
```
parameja1, line 179: FYI, the check for Unified Studio i.e ```process.env.SERVICE_NAME === sageMakerUnifiedStudio```, is already handled in ```isSageMakerUnifiedStudio()```
And with this logic, it will bypass the check for Sagemaker AI or Unified Studio cases right? Can we extract these env variable checks omitting this ```process.env.SERVICE_NAME === sageMakerUnifiedStudio```into a boolean and use it in case statements for better readability?

PR #7558, opened 2025-06-25
fix(amazonq): Remove setSystemCertificates from proxyUtil

tsmithsz:
## Problem
We enabled experimental proxy support in FLARE in [this PR](https://github.com/aws/aws-toolkit-vscode/pull/7556). The ProxyConfigManager within FLARE gathers and injects OS CAs. See [here](https://github.com/tsmithsz/language-server-runtimes/blob/main/runtimes/runtimes/util/standalone/experimentalProxyUtil.ts#L124). 

The current `setSystemCertificates()` call also extracts system certificates and sets `NODE_EXTRA_CA_CERTS/AWS_CA_BUNDLE`. This duplicates and can even override our custom logic, leading to unpredictable CA bundles.

## Solution

- Remove redundant `setSystemCertificates()` method from proxy util.

All certificate loading now flows through a single code path (`ProxyConfigManager.getCertificates()`), eliminating conflicts and ensuring consistent OS CA handling.

## Testing

- Tested using MITM Proxy in local capture mode

---

### DIFF
--- a/packages/core/src/shared/utilities/proxyUtil.ts
+++ b/packages/core/src/shared/utilities/proxyUtil.ts
@@ -5,9 +5,6 @@
 
 import vscode from 'vscode'
 import { getLogger } from '../logger/logger'
-import { tmpdir } from 'os'
-import { join } from 'path'
-import * as nodefs from 'fs' // eslint-disable-line no-restricted-imports
 
 interface ProxyConfig {
     proxyUrl: string | undefined
@@ -73,9 +70,6 @@ export class ProxyUtil {
         // Always enable experimental proxy support for better handling of both explicit and transparent proxies
         process.env.EXPERIMENTAL_HTTP_PROXY_SUPPORT = 'true'
 
-        // Load built-in bundle and system OS trust store
-        process.env.NODE_OPTIONS = '--use-system-ca'
-
         const proxyUrl = config.proxyUrl
         // Set proxy environment variables
         if (proxyUrl) {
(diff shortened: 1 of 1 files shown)


packages/core/src/shared/utilities/proxyUtil.ts
```diff
@@ -73,9 +70,6 @@ export class ProxyUtil {
         // Always enable experimental proxy support for better handling of both explicit and transparent proxies
         process.env.EXPERIMENTAL_HTTP_PROXY_SUPPORT = 'true'
 
-        // Load built-in bundle and system OS trust store
-        process.env.NODE_OPTIONS = '--use-system-ca'
```
rli, line 77: this line makes sense, but it is unclear to me how duplicate certificates in the trust list will cause an issue downstream in flare

tsmithsz: I'm also not 100% sure if it will cause an issue downstream, but my thinking was that we're assembling two independent trust stores:  the temp-file bundle written by `setSystemCertificates()` vs. the in-memory list inside Flare.

There may be a mismatch in the certs used depending on the call path. I don't have a strong opinion, was just more thinking along the lines of having a single source of truth.

Link to Flare implementation: https://github.com/tsmithsz/language-server-runtimes/blob/main/runtimes/runtimes/util/standalone/experimentalProxyUtil.ts

rli: Since we lack instrumentation for this issue I would prefer we are less selective by pulling in all available certs

samgst-amazon (approved):

PR #7560, opened 2025-06-25
refactor(stepfunctions): migrate to sdkv3

zelzhou:
## Problem

step functions still uses sdk v2

## Solution

- Refactor `DefaultStepFunctionsClient` to `StepFunctionsClient` using ClientWrapper.
- Refactor references to `DefaultStepFunctionsClient` with `StepFunctionsClient` and reference sdk v3 types

## Verification
- Verified CreateStateMachine, UpdateStateMachine works and creates/updates StateMachine
- Verified AWS Explorer lists all StateMachines using ListStateMachine
- Verified StartExecution works
- Verified TestState, ListIAMRoles works in Workflow Studio, which also fixes the bug that variables cannot be used in TestState in Workflow Studio
- fixes https://github.com/aws/aws-toolkit-vscode/issues/6819
  TestStateWithVariables
---

- Treat all work as PUBLIC. Private `feature/x` branches will not be squash-merged at release time.
- Your code changes must meet the guidelines in [CONTRIBUTING.md](https://github.com/aws/aws-toolkit-vscode/blob/master/CONTRIBUTING.md#guidelines).
- License: I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7578, opened 2025-06-28
feat(stepfunctions): Enable execution details page

l0minous:
## Problem
`View Execution Details by Execution ARN` command currently only opens a placeholder Hello World webview instead of an Execution Details page.

## Solution
- Modify `workflowStudioApiHandler.ts` into `StepFunctionApiHandler.ts` as well adding ability to perform Api calls for execution detail specific APIs
- Updating `types.ts` , `stepFunctions.ts` (in clients folder) , `executionDetailProvider.ts` to support new API calls and sdk v3
- Implementing separate `handleMessage.ts` to handle messaging between Execution Details webview and toolkit 

## Verification
test2

---


packages/toolkit/.changes/next-release/Feature-2a82bbe6-4f79-4129-90ba-0e933173a9cd.json
```diff
@@ -0,0 +1,4 @@
+{
+	"type": "Feature",
+	"description": "Added Execution Details Page webview"
```
anthonyting: I don't think we need another changelog since we added one already

packages/core/src/stepFunctions/executionDetails/handleMessage.ts
```diff
@@ -0,0 +1,120 @@
...
+    executionArn?: string
+}
+
+/**
+ * Handles messages received from the ExecutionDetails webview. Depending on the message type and command,
+ * calls the appropriate handler function
+ * @param message The message received from the webview
+ * @param context The context object containing information about the execution details webview environment
+ */
+export async function handleMessage(message: Message, context: ExecutionDetailsContext) {
```
anthonyting, line 27: Much of the code in this file is very similar to the WorkflowStudio handlers. Can we extract some common logic between the two versions instead of duplicating it and use that instead, given they use the same CDN distributions?

packages/core/src/stepFunctions/executionDetails/executionDetailProvider.ts
```diff
@@ -99,10 +100,26 @@ export class ExecutionDetailProvider {
             // Set up the content
             panel.webview.html = await this.getWebviewContent()
 
+            // Create execution details context
+            const context: ExecutionDetailsContext = {
```
anthonyting, line 103: Most of the fields on this are unused. Can we use a new context instead of extending the WorkflowStudio one?

packages/core/src/stepFunctions/workflowStudio/workflowStudioApiHandler.ts
anthonyting, line 1: If we're using code for both Workflow Studio and Execution Details, we should extract it to a common file instead of keeping the naming as `workflowStudio/workflowStudioApiHandler.ts` and `workflowStudio/types.ts`.

packages/core/src/shared/clients/stepFunctionsClient.ts
```diff
@@ -32,20 +31,88 @@ export class DefaultStepFunctionsClient {
...
+        arn: string
+    ): Promise<StepFunctions.DescribeStateMachineForExecutionOutput> {
+        const client = await this.createSdkClient()
+
+        const request: StepFunctions.DescribeStateMachineForExecutionInput = {
+            executionArn: arn,
+        }
+        const response: StepFunctions.DescribeStateMachineForExecutionOutput = await client
+            .describeStateMachineForExecution(request)
+            .promise()
```
anthonyting: We should use aws-sdk v3 now that it was merged in https://github.com/aws/aws-toolkit-vscode/pull/7560

packages/core/src/shared/clients/stepFunctionsClient.ts
```diff
@@ -32,20 +31,88 @@ export class DefaultStepFunctionsClient {
...
     }
 
+    public async getStateMachineDetailsForExecution(
+        arn: string
+    ): Promise<StepFunctions.DescribeStateMachineForExecutionOutput> {
+        const client = await this.createSdkClient()
+
+        const request: StepFunctions.DescribeStateMachineForExecutionInput = {
+            executionArn: arn,
+        }
```
anthonyting: These APIs take in more than just an ARN for the request input. Can we just use the request input as the function input instead of limiting it to only the ARN? (same comment for all the APIs added here)

packages/core/src/shared/clients/stepFunctionsClient.ts
zelzhou: +1

packages/core/src/stepFunctions/executionDetails/executionDetailProvider.ts
zelzhou: +1

packages/core/src/stepFunctions/messageHandlers/stepFunctionApiHandler.ts
```diff
@@ -6,14 +6,21 @@
...
+    Command,
+    MessageType,
+    WebviewContext,
+    ExecutionDetailsContext,
+} from './types'
 import { telemetry } from '../../shared/telemetry/telemetry'
 import { ListRolesRequest } from '@aws-sdk/client-iam'
 
-export class WorkflowStudioApiHandler {
+export class stepFunctionApiHandler {
```
anthonyting: Class name should be capitalized

packages/core/src/stepFunctions/executionDetails/executionDetailProvider.ts
```diff
@@ -98,11 +100,18 @@ export class ExecutionDetailProvider {
 
             // Set up the content
             panel.webview.html = await this.getWebviewContent()
+            const context: ExecutionDetailsContext = {
+                stateMachineName: executionArn.split(':').pop() || 'Unknown',
+                panel,
+                loaderNotification: undefined,
+                fileId: executionArn,
+                executionArn,
```
anthonyting: Most of these still look unused. Can we remove the unused properties?

packages/core/src/shared/clients/stepFunctions.ts
```diff
@@ -50,10 +69,38 @@ export class StepFunctionsClient extends ClientWrapper<SFNClient> {
         return this.makeRequest(DescribeStateMachineCommand, request)
     }
 
+    public async getStateMachineDetailsForExecution(
```
anthonyting: Can the functions be named based on the APIs instead of a custom name? (same for getExecutionDetails and getMapRunDetails)

packages/core/src/stepFunctions/messageHandlers/types.ts
```diff
@@ -30,6 +30,14 @@ export type WebviewContext = {
     fileId: string
 }
 
+export type ExecutionDetailsContext = {
```
anthonyting: Could we make `WebviewContext` and `ExecutionDetailsContext` inherit from a common context for things that are common? That way we don't have to pass union types of `WebviewContext | ExecutionDetailsContext` everywhere, and we can just pass the parent interface instead

anthonyting (changes requested):

packages/core/src/shared/clients/stepFunctions.ts
```diff
@@ -50,10 +69,38 @@ export class StepFunctionsClient extends ClientWrapper<SFNClient> {
...
+
+    public async reDriveExecution(request: RedriveExecutionCommandInput): Promise<RedriveExecutionCommandOutput> {
+        return this.makeRequest(RedriveExecutionCommand, request)
+    }
+
     public async executeStateMachine(request: StartExecutionCommandInput): Promise<StartExecutionCommandOutput> {
         return this.makeRequest(StartExecutionCommand, request)
     }
 
+    public async stopExecution(request: StopExecutionCommandInput): Promise<StopAccessLoggingOutput> {
```
zelzhou: Isn't this supposed to be `StopExecutionCommandOutput` [ref to sdk](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-sfn/Interface/StopExecutionCommandOutput/)?

packages/core/src/shared/clients/stepFunctions.ts
```diff
@@ -26,6 +43,8 @@ import {
     UpdateStateMachineCommandOutput,
 } from '@aws-sdk/client-sfn'
 import { ClientWrapper } from './clientWrapper'
+import { StopAccessLoggingOutput } from 'aws-sdk/clients/mediastore'
+// import { StopExecutionInput } from 'aws-sdk/clients/stepfunctions'
```
zelzhou: Isn't this supposed to be StopExecutionCommandOutput [ref to sdk](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-sfn/Interface/StopExecutionCommandOutput/)?

zelzhou (approved):

PR #7587, opened 2025-06-30
fix(amazonq): remove the depedency of the CDN js script.

grl-wang:
1. update the commonAuthViewProvider to use the locally bundled vue.js, instead of to load the vue.js from CDN.
2. this going to resolve the customer request that When customer is trying to login to IDC via Q they are hitting a URL which is not part of the whitelist urls as per [documentation](https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/firewall.html)

## Problem
Customers complain about need to whitelist the `cdnjs.cloudflare.com` which doesn't listed in the https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/firewall.html

## Solution
update the commonAuthViewProvider to use the locally bundled vue.js, instead of to load the vue.js from CDN.

verified the amazon Q auth view still working as expected after this change. 
image

---

### DIFF
--- a/packages/core/scripts/build/copyFiles.ts
+++ b/packages/core/scripts/build/copyFiles.ts
@@ -31,6 +31,11 @@ const tasks: CopyTask[] = [
     { target: path.join('src', 'testFixtures') },
     { target: 'src/auth/sso/vue' },
 
+    // Vue.js for webviews
+    {
+        target: path.join('../../node_modules', 'vue', 'dist', 'vue.global.prod.js'),
+        destination: path.join('libs', 'vue.min.js'),
+    },
     // SSM
     {
         target: path.join('../../node_modules', 'aws-ssm-document-language-service', 'dist', 'server.js'),
--- a/packages/core/src/login/webview/commonAuthViewProvider.ts
+++ b/packages/core/src/login/webview/commonAuthViewProvider.ts
@@ -148,6 +148,10 @@ export class CommonAuthViewProvider implements WebviewViewProvider {
         const entrypoint =
             serverHostname !== undefined ? Uri.parse(serverHostname).with({ path: `/${this.source}` }) : scriptUri
 
+        // Get Vue.js from dist/libs directory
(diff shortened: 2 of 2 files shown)


packages/core/src/login/webview/commonAuthViewProvider.ts
```diff
@@ -158,7 +162,7 @@ export class CommonAuthViewProvider implements WebviewViewProvider {
                     <title>Base View Extension</title>
                 </head>
                 <body>
-                    <script src="https://cdnjs.cloudflare.com/ajax/libs/vue/3.4.4/vue.global.prod.min.js"></script>  
+                    <script src="${vueScript.toString()}"></script>
```
jguoamz, line 165: does the version (3.4.4) matter? If the version changes, would it break?

grl-wang: I don't think so, did local tested with the generated vsix for the amazon Q sign in and sign out, didn't see any issue for the login ux with vue.js file in `[3.3.4` which is used in the toolkit package](https://github.com/aws/aws-toolkit-vscode/blob/bfdb0ebe96213710e88461462c101808470436bf/packages/core/package.json#L576).

leigaol (approved):

packages/core/scripts/build/copyFiles.ts
```diff
@@ -31,6 +31,11 @@ const tasks: CopyTask[] = [
     { target: path.join('src', 'testFixtures') },
     { target: 'src/auth/sso/vue' },
 
+    // Vue.js for webviews
+    {
+        target: path.join('../../node_modules', 'vue', 'dist', 'vue.global.prod.js'),
```
leigaol, line 36: please test both Q and toolkit

grl-wang: verified both Q and toolkit can login of the generated vsix.  following is the toolkit login, Q login pasted in the pr description
image

PR #7597, opened 2025-07-02
feat(amazonq): add setup for ui-tests

laura-codess:
## Changes
This is the initial commit for UI Tests using the framework vscode-extension-tester. It instantiates the VSCode instance which opens Amazon Q and goes through some basic flows to login. This is done using the [VSCode-Extension-Tester Framework](https://github.com/redhat-developer/vscode-extension-tester/wiki/). (Note: It is missing full authentication and stops working once Amazon Q needs to open the browser).

**To run this test**

1. First compile the test files
`npm run testCompile`
2. Then run the test setup
`node_modules/.bin/extest setup-tests`
2. Then run the script (NOTE: this js file should be generated when you compile the test files)
`node_modules/.bin/extest run-tests aws-toolkit-vscode/packages/amazonq/dist/test/e2e/amazonq/VET.test.js` 

For information why each dependency / webpack change is needed, please see [here](https://quip-amazon.com/lCuBAOGibHzm/UI-Tests-PR-Explanation).

Demo Video:

https://github.com/user-attachments/assets/f1c6e59f-d4e3-4ae0-a164-1da389ec1339

---


package.json
```diff
@@ -62,12 +62,15 @@
         "eslint-plugin-prettier": "^5.1.3",
         "eslint-plugin-security-node": "^1.1.4",
         "eslint-plugin-unicorn": "^54.0.0",
+        "events": "^3.3.0",
```
Hweinstock: which of these new dependencies are actually needed for the current functionality?

package.json
```diff
@@ -76,7 +79,17 @@
     "dependencies": {
         "@types/node": "^22.7.5",
         "jaro-winkler": "^0.2.8",
+        "buffer": "^6.0.3",
+        "mocha": "^11.7.1",
```
Hweinstock, line 84: I think most of these can be dev-dependencies. Also same comment as above.

packages/amazonq/package.json
```diff
@@ -70,7 +70,9 @@
         "generateSettings": "ts-node ../../scripts/generateSettings.ts"
     },
     "dependencies": {
-        "aws-core-vscode": "file:../core/"
+        "aws-core-vscode": "file:../core/",
+        "mocha": "^11.7.1",
```
Hweinstock: I don't think you need these since they are at the top level.

packages/webpack.web.config.js
```diff
@@ -53,6 +53,10 @@ module.exports = (env, argv) => {
             new webpack.IgnorePlugin({
                 resourceRegExp: /svgdom/, // matches the path in the require() statement
             }),
+            // Handle node: protocol imports by normalizing them
+            new webpack.NormalModuleReplacementPlugin(/^node:(.*)$/, (resource) => {
```
Hweinstock, line 57: what does this do and why is it needed? Are there any side effects we should know about?

packages/webpack.web.config.js
```diff
@@ -65,6 +69,15 @@ module.exports = (env, argv) => {
                 crypto: require.resolve('crypto-browserify'),
                 'fs-extra': false,
                 perf_hooks: false, // should be using globalThis.performance instead
+                // Handle node: protocol imports
```
Hweinstock: I think we already polyfill these somewhere for web. Could be worth confirming if these are needed.

test.log
```diff
@@ -0,0 +1,13 @@
+
+
+Writing code settings to /var/folders/2_/2gq5dx0523d60tk4_xn6_pnr0000gq/T/test-resources/settings/User/settings.json
```
Hweinstock: I don't think you want to commit this.

tsconfig.tsbuildinfo
Hweinstock, line 1: same as other file for this one.

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Workbench, By, EditorView, WebviewView } from 'vscode-extension-tester'
+
+describe('Amazon Q Login Flow', function () {
+    // need this timeout because Amazon Q takes awhile to load
+    this.timeout(90000)
```
Hweinstock: This seems like a really long timeout? Do the tests really take that long?

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    let webviewView: WebviewView
+
+    // this before() block will run before every test
+    // NOTE: I tested all the timeouts and they are necessary for the webview to load properly
+    before(async function () {
```
Hweinstock, line 14: I think in mocha `before` runs once in the beginning and then `beforeEach` runs before each test.

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        // opens the amazon Q chat webview
```
Hweinstock: comments are a great way to communicate ideas that don't come through the code. In this case, I think the following two lines are fairly clear that its opening the chat webview.

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        // timeout to allow the WebviewView to load
+        await new Promise((resolve) => setTimeout(resolve, 15000))
+        webviewView = new WebviewView()
+
+        // switches the frame to the Amazon Q webview
+        try {
+            await webviewView.switchToFrame()
+        } catch (e) {
+            console.log('WE HIT AN ERROR OPEENING THE FRAME')
```
Hweinstock: If we hit an error here we should fail the test. The error will bubble up and appear in the test report, so you don't need to log it.

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
...
+            console.log(e)
+        }
+    })
+
+    // this after() block will run after every test
```
Hweinstock: same as before, I think its `afterEach` in mocha.

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    it('Should click through the Amazon Q login screen', async () => {
+        // Select company account option
+        const selectableItems = await webviewView.findWebElements(By.css('.selectable-item'))
+        if (selectableItems.length === 0) {
+            throw new Error('No selectable login options found')
+        }
+
+        // Find and click company account
+        const companyItem = await findItemByText(selectableItems, 'Company account')
```
Hweinstock, line 39: is there a way to select this button that doesn't rely on the text? Since this was recently changed I imagine it might get changed again and then this all breaks :(

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        await companyItem.click()
+        await new Promise((resolve) => setTimeout(resolve, 2000))
```
Hweinstock, line 41: out of curiosity, did you experiment with lower timeouts? I imagine this will get very slow with more tests.

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        // Click first continue button
+        const continueButton = await webviewView.findWebElement(By.css('#connection-selection-continue-button'))
+        await continueButton.click()
+        await new Promise((resolve) => setTimeout(resolve, 2000))
+
+        // Enter start URL
```
Hweinstock, line 48: might be worth adding a comment here that is likely a temporary approach. I imagine you'll want to start the extension authenticated in the future.

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        const startUrlInput = await webviewView.findWebElement(By.id('startUrl'))
+        await startUrlInput.clear()
+        await startUrlInput.sendKeys('https://amzn.awsapps.com/start')
+        await new Promise((resolve) => setTimeout(resolve, 1000))
+
+        // Click second continue button
+        const continueButton2 = await webviewView.findWebElement(By.css('button.continue-button.topMargin'))
```
Hweinstock: what are the continue buttons? Are these popups in the extension?

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        await continueButton2.click()
+        await new Promise((resolve) => setTimeout(resolve, 2000))
+    })
+
+    // Helper to find item by text content
+    async function findItemByText(items: any[], text: string) {
```
Hweinstock: nice! appreciate you made this a reusable component. Is there any way we can get stronger typing than `any`?

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        for (const item of items) {
+            const titleDivs = await item.findElements(By.css('.title'))
```
Hweinstock, line 63: Does this `.title` in the search work in all cases?

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -0,0 +1,83 @@
...
+            for (const titleDiv of titleDivs) {
+                const titleText = await titleDiv.getText()
+                if (titleText?.trim().startsWith(text)) {
```
Hweinstock, line 66: Do you want to always check if it starts with the exact text, or could you use something more flexible like a regex pattern?

Hweinstock:
looks like the buffer dependency was needed: 
```
ERROR in ../../node_modules/buffer-xor/index.js 3:19-25
  Module not found: Error: Can't resolve 'buffer' in '/codebuild/output/src1044368460/src/github.com/aws/aws-toolkit-vscode/node_modules/buffer-xor'
```

package.json
Hweinstock: I think mocha should already be getting pulled in (https://github.com/aws/aws-toolkit-vscode/blob/4edc815246709ad2deba68682b67e08b2a6df818/packages/core/package.json#L515). `buffer` might be able to be a devDependency but we can look into this as a follow up.

package.json
```diff
@@ -68,15 +68,24 @@
...
     },
     "dependencies": {
         "@types/node": "^22.7.5",
+        "buffer": "^6.0.3",
         "jaro-winkler": "^0.2.8",
+        "mocha": "^11.7.1",
         "vscode-nls": "^5.2.0",
         "vscode-nls-dev": "^4.0.4"
+    },
+    "engines": {
```
Hweinstock, line 88: curious as to what these actually do, might be worth calling out in PR description with maybe a link to VET docs explaining?

packages/amazonq/test/e2e/amazonq/VET.test.ts
Hweinstock: since this is more of demo, no need to worry about this for now.

packages/amazonq/test/e2e/amazonq/VET.test.ts
Hweinstock: sounds like this is a tough problem, can be explored in the future.

packages/amazonq/test/e2e/amazonq/VET.test.ts
Hweinstock: I guess this applies to all of the testing logic, can ignore this for now.

packages/amazonq/test/e2e/amazonq/VET.test.ts
Hweinstock: This function is temporary so no need to worry it for now.

packages/amazonq/test/e2e/amazonq/VET.test.ts
Hweinstock: same as above.

packages/webpack.web.config.js
Hweinstock: sounds like this is critical to the build. Good to leave for now, but we should figure out why in case it breaks going forward.

packages/amazonq/test/e2e/amazonq/VET.test.ts
Hweinstock: 30 seems more reasonable.

packages/amazonq/test/e2e/amazonq/VET.test.ts
Hweinstock: I think you can just remove the try-catch here.

Hweinstock:
I think after removing that try-catch, adding some doc links, and updating the PR description for how to run + what to expect, this looks good to me as a start. We should create tracking issues for:
- investigate the webpack and dependencies changes to understand why they are needed. We can't be making changes here we don't understand when trying to merge code to the mainline. 
- investigate timeout requirements. I suspect someone has solved this problem so might just require some research of finding someone else's solution rather than somewhat arbitrary timeouts between actions. 

These PR descriptions can eventually serve a "work-log" for the project so I think its important to make them solid now while we have context.

laura-codess:
It looks like 2 tests are failing even after the webpack addition, but the code has no trouble compiling. My initial commit passed all the checks so I'm going to try adding back some lines to see what was necessary. I'll add all of that information in the PR and work on writing a doc about why each dependency / webpack change is needed.

packages/webpack.web.config.js
laura-codess: Added a link to a quip doc explaining why each change is needed.

PR #7599, opened 2025-07-02
fix(toolkit): handle isCn when region is not initialized to fix extension activation failure

parameja1:
## Problem
With updating Toolkit version to latest in SMUS CodeEditor Spaces, observing Toolkit throws an error as attached in screenshot below with error message
```
Attempted to get compute region without initializing
```
image

## Solution
- Issue is due to getComputeRegion() invoked first before compute region is initialized
- hence solution is to Make `isCn()` resilient to uninitialized state and return a default value
- Tested with a local debug artifact in SMUS CodeEditor space, toolkit activation completed and working
- 
image

---

### DIFF
--- a/packages/core/src/shared/extensionUtilities.ts
+++ b/packages/core/src/shared/extensionUtilities.ts
@@ -194,7 +194,13 @@ export function isSageMaker(appName: 'SMAI' | 'SMUS' = 'SMAI'): boolean {
 }
 
 export function isCn(): boolean {
-    return getComputeRegion()?.startsWith('cn') ?? false
+    try {
+        return getComputeRegion()?.startsWith('cn') ?? false
+    } catch (e) {
+        // If compute region isn't initialized yet, assume we're not in a CN region
+        getLogger().debug('isCn called before compute region initialized, defaulting to false')
+        return false
+    }
 }
 
 /**


packages/core/src/shared/extensionUtilities.ts
```diff
@@ -194,7 +194,13 @@ export function isSageMaker(appName: 'SMAI' | 'SMUS' = 'SMAI'): boolean {
 }
 
 export function isCn(): boolean {
-    return getComputeRegion()?.startsWith('cn') ?? false
+    try {
+        return getComputeRegion()?.startsWith('cn') ?? false
+    } catch (e) {
+        // If compute region isn't initialized yet, assume we're not in a CN region
+        getLogger().debug('isCn called before compute region initialized, defaulting to false')
```
aws-asolidu: should we also log the error message?

parameja1: Thanks for the review, added error logging as well as unit tests

parameja1: 2 of the CI workflow checks are failing seemingly due to a change merged recently and not due to this commit, reference: https://github.com/aws/aws-toolkit-vscode/actions/runs/16175639516/job/45659998532

packages/core/src/test/shared/extensionUtilities.test.ts
```diff
@@ -135,6 +135,65 @@ describe('initializeComputeRegion, getComputeRegion', async function () {
...
+        assert.strictEqual(result, false, 'isCn() should return false for non-China regions')
+    })
+
+    it('returns false when an error occurs', async function () {
+        const utils = require('../../shared/extensionUtilities')
+
+        // Restore original getComputeRegion if it was stubbed
+        if (utils.getComputeRegion.restore) {
+            utils.getComputeRegion.restore()
+        }
```
hurstsad: Do we need this? I see `sandbox.restore()` is called in our `afterEach`, doesn't this already reset all stubs?

packages/core/src/shared/extensionUtilities.ts
```diff
@@ -194,7 +194,17 @@ export function isSageMaker(appName: 'SMAI' | 'SMUS' = 'SMAI'): boolean {
...
+        const region = getComputeRegion()
+        if (!region || region === 'notInitialized') {
+            getLogger().debug('isCn called before compute region initialized, defaulting to false')
+            return false
+        }
+        return region.startsWith('cn')
+    } catch (err) {
+        getLogger().error(`Error in isCn method: ${err}`)
+        return false
+    }
```
hurstsad, line 207: Should we instead be throwing the error here? We already handle the known case above by returning false when the region is uninitialized, but this can catch other errors that I'm not sure we should assume returning false for.

packages/core/src/test/shared/extensionUtilities.test.ts
parameja1: Good catch, not needed necessarily. Added it initially for best practice, but since ```sandbox``` is restored in ```afterEach```, underlying stubs should also get cleaned

packages/core/src/shared/extensionUtilities.ts
parameja1: Good point, the case that throws error is handled already, and i didnt find any [error cases](https://github.com/aws/aws-toolkit-vscode/blob/8e870bfd0d7e3381ada5ef29a9420cd66b68800d/packages/core/src/shared/extensionUtilities.ts#L380) that could result from the ```getComputeRegion``` method, hence it looks safe to not throw error back

hurstsad (approved):

laileni-aws (approved):

PR #7600, opened 2025-07-02
fix(amazonq): skip matching with previous suggestions for EDITS

atonaamz:
## Problem
if item is an Edit suggestion, insertText is a diff instead of new code contents, skip the logic to check for prefix.

## Solution
skip matching with previous suggestions for EDITS

---


packages/amazonq/src/app/inline/completion.ts
```diff
@@ -266,7 +266,8 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
                 const prevItemMatchingPrefix = []
                 for (const item of this.sessionManager.getActiveRecommendation()) {
                     const text = typeof item.insertText === 'string' ? item.insertText : item.insertText.value
-                    if (text.startsWith(prefix) && position.isAfterOrEqual(prevStartPosition)) {
+                    // if item is an Edit suggestion, insertText is a diff instead of new code contents, skip the logic to check for prefix.
+                    if (text.startsWith(prefix) && position.isAfterOrEqual(prevStartPosition) && !item.isInlineEdit) {
```
floralph: Minor nit: It might be a little clear to exit early like:
```
for (const item of this.sessionManager.getActiveRecommendation()) {
    // Skip initializing item for inline edits
    if (item.isInlineEdit) {
        continue
    }
    const text = typeof item.insertText === 'string' ? item.insertText : item.insertText.value
```

Will-ShaoHua: +1

Will-ShaoHua (approved):

PR #7611, opened 2025-07-03
feat(amazonq): passing partialResultToken for the next trigger if user accepts an EDITS suggestion

atonaamz:
## Problem

If it is an EDITS suggestion, we need to pass partialResultToken to fetch result from cache.

## Solution

passing partialResultToken for the next trigger if user accepts an EDITS suggestion

---

### DIFF
--- a/packages/amazonq/src/app/inline/completion.ts
+++ b/packages/amazonq/src/app/inline/completion.ts
@@ -221,7 +221,8 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
         position: Position,
         context: InlineCompletionContext,
         token: CancellationToken,
-        getAllRecommendationsOptions?: GetAllRecommendationsOptions
+        getAllRecommendationsOptions?: GetAllRecommendationsOptions,
+        editsStreakPartialResultToken?: number | string
     ): Promise<InlineCompletionItem[]> {
         getLogger().info('_provideInlineCompletionItems called with: %O', {
             documentUri: document.uri.toString(),
@@ -323,7 +324,8 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
                 position,
                 context,
                 token,
-                getAllRecommendationsOptions
+                getAllRecommendationsOptions,
+                editsStreakPartialResultToken
             )
             // get active item from session for displaying
             const items = this.sessionManager.getActiveRecommendation()
--- a/packages/amazonq/src/app/inline/recommendationService.ts
+++ b/packages/amazonq/src/app/inline/recommendationService.ts
@@ -42,17 +42,21 @@ export class RecommendationService {
         position: Position,
         context: InlineCompletionContext,
         token: CancellationToken,
-        options: GetAllRecommendationsOptions = { emitTelemetry: true, showUi: true }
+        options: GetAllRecommendationsOptions = { emitTelemetry: true, showUi: true },
+        editsStreakPartialResultToken?: number | string
     ) {
         // Record that a regular request is being made
         this.cursorUpdateRecorder?.recordCompletionRequest()
 
-        const request: InlineCompletionWithReferencesParams = {
+        let request: InlineCompletionWithReferencesParams = {
             textDocument: {
                 uri: document.uri.toString(),
             },
             position,
             context,
+        }
+        if (editsStreakPartialResultToken) {
+            request = { ...request, partialResultToken: editsStreakPartialResultToken }
         }
         const requestStartTime = globals.clock.Date.now()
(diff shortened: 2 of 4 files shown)


packages/amazonq/src/app/inline/completion.ts
```diff
@@ -221,7 +221,8 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
         position: Position,
         context: InlineCompletionContext,
         token: CancellationToken,
-        getAllRecommendationsOptions?: GetAllRecommendationsOptions
+        getAllRecommendationsOptions?: GetAllRecommendationsOptions,
```
floralph: minor nit: The param defs will likely grow into a long tail.  Should this be converted to an "options" parameter to capture these optional params at the end?

packages/amazonq/src/app/inline/completion.ts
```diff
@@ -324,7 +325,8 @@ export class AmazonQInlineCompletionItemProvider implements InlineCompletionItem
                 context,
                 token,
                 isAutoTrigger,
-                getAllRecommendationsOptions
+                getAllRecommendationsOptions,
```
floralph: Similar minor nit here as above.

packages/amazonq/src/app/inline/recommendationService.ts
```diff
@@ -42,18 +42,22 @@ export class RecommendationService {
         context: InlineCompletionContext,
         token: CancellationToken,
         isAutoTrigger: boolean,
-        options: GetAllRecommendationsOptions = { emitTelemetry: true, showUi: true }
+        options: GetAllRecommendationsOptions = { emitTelemetry: true, showUi: true },
```
floralph: Similar minor nit here as above.

packages/amazonq/src/app/inline/recommendationService.ts
```diff
@@ -113,18 +117,26 @@ export class RecommendationService {
...
-                    result = await languageClient.sendRequest(
-                        inlineCompletionWithReferencesRequestType.method,
-                        paginatedRequest,
-                        token
-                    )
-                    this.sessionManager.updateSessionSuggestions(result.items)
+            const isInlineEdit = result.items.some((item) => item.isInlineEdit)
+            // Skip fetching for more items if the suggesion is EDITS. If it is EDITS suggestion, only fetching for more
+            // suggestions when the user start to accept a suggesion.
+            if (!isInlineEdit) {
```
floralph, line 120: Consider adding some logging here for which path is taken to help debug later.

packages/amazonq/src/app/inline/recommendationService.ts
```diff
@@ -113,18 +117,26 @@ export class RecommendationService {
             )
 
             // If there are more results to fetch, handle them in the background
```
floralph: Consider updating this comment and the one on line 76 as they no longer seem to reflect the flow of code they comment.

packages/amazonq/src/app/inline/recommendationService.ts
```diff
@@ -113,18 +117,26 @@ export class RecommendationService {
...
+                        this.sessionManager.updateSessionSuggestions(result.items)
+                    }
+                } catch (error) {
+                    languageClient.warn(`Error when getting suggestions: ${error}`)
                 }
-            } catch (error) {
-                languageClient.warn(`Error when getting suggestions: ${error}`)
+            } else {
+                // save editsStreakPartialResultToken for the next EDITS suggestion trigger if user accepts
+                this.sessionManager.updateActiveEditsStreakToken(result.partialResultToken)
```
floralph, line 143: This assumes the first result is an edit, but the check on line 120 is just for at one edit anywhere in the results.  Is there some undocumented assumption here about how the results are returned?  Even if there is, there should be checks in code that the assumptions hold.

packages/amazonq/src/app/inline/completion.ts
atonaamz: Yes we also can add it to the options params.

packages/amazonq/src/app/inline/recommendationService.ts
atonaamz: We can only return either EDITS or COMPLETIONS so if one item is EDITS, we should follow the EDITS flow.

atonaamz: https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts#L834

floralph (approved):

PR #7656, opened 2025-07-13
fix(lambda): Updates for Lambda quick edit

rhamilt:
This PR contains updates for miscellaneous features related to the new Lambda quick edit.

### Problem
Previously, having an empty directory in the Lambda quick edit flow led to errors when trying to view the function code in the explorer or in a workspace. An empty folder could be the result of an error during download. The unintended bug in the code was related to how we were checking for existence of a function saved locally. 

### Solution
Check if there are any files in the directory before determining if it exists locally. If the directory is empty, we must redownload the code (a Lambda function cannot be empty, so we know it is unintended)

### Problem
- A couple typographic errors
- README was opening as separate tab, not as split pane
- README was referencing "cloud deploy icon" rather than showing the actual icon, same with the invoke icon

### Solution
- Fix errors
- Adjust README open to use `markdown.showPreviewToSide`
- Save the icons locally so we can display them in the actual README.

### Problem
- Some non-passive metrics were being emitted on activation
- Downloading code on activation lead to race conditions with authentication
- Downloading code on activation lead to extra slow load times

### Solution
All of this is fixed by moving things around so that there is never any code downloaded during activation. This is beneficial because it keeps us from doing a lot of our main functionality on activation.

Previously, opening a function in a workspace, which could be done through a URI handler or from the explorer, would cause the extension to restart. Once the extension restarted, at that point we would download the code and start the watchers. These changes download the code first, and only start the watchers on activation. Because the user doesn't need to be authenticated to start the watchers (only to deploy), this is not an issue for being authenticated. If it's a workspace folder, we're assuming the function exists locally, so that's why we know we've downloaded the code already.

The only edge case is if the local code is out of sync with what's in the cloud; in this case, we prompt the user to overwrite their code, but they need to go to the explorer to manually download it to avoid the activation race conditions.

---


packages/core/src/lambda/activation.ts
```diff
@@ -29,47 +29,125 @@ import { ToolkitError, isError } from '../shared/errors'
...
+                                    tab.input instanceof vscode.TabInputText &&
+                                    tab.input.uri.fsPath.startsWith(workspaceUri.fsPath)
+                            )
+                            if (tabsToClose.length > 0) {
+                                await vscode.window.tabGroups.close(tabsToClose)
+                            }
+                        }
+
+                        // Delete all files in the directory
+                        await deleteFilesInFolder(workspacePath)
```
roger-zhangg, line 113: So this confirm deleted all data in folder but didn't download anything?

rhamilt: Yes, we want to clear the folder so that next time we try to open the code, `editLambda` knows it needs to redownload the freshest code.

packages/core/src/lambda/commands/editLambda.ts
```diff
@@ -206,34 +219,46 @@ export async function editLambda(lambda: LambdaFunction, onActivation?: boolean)
...
+    await editLambda(
+        {
+            name,
+            region,
+            configuration: configuration as any,
+        },
+        'workspace'
+    )
+
+    await vscode.commands.executeCommand('vscode.openFolder', vscode.Uri.file(downloadLocation), {
```
laileni-aws: What happens if VSC fails to open the folder? Do we show any error to the user?

packages/core/src/lambda/commands/editLambda.ts
```diff
@@ -206,34 +219,46 @@ export async function editLambda(lambda: LambdaFunction, onActivation?: boolean)
...
+    const createStackIconSource = path.join('resources', 'icons', 'aws', 'lambda', 'create-stack-light.svg')
     const createStackIconDestination = path.join(lambdaTempPath, 'create-stack.svg')
     await fs.copy(globals.context.asAbsolutePath(createStackIconSource), createStackIconDestination)
 
+    // Copy VS Code built-in icons
+    const vscodeIconPath = path.join('resources', 'icons', 'vscode', 'light')
+
+    const invokeIconSource = path.join(vscodeIconPath, 'run.svg')
+    const invokeIconDestination = path.join(lambdaTempPath, 'invoke.svg')
+    await fs.copy(globals.context.asAbsolutePath(invokeIconSource), invokeIconDestination)
```
laileni-aws: nit:
Do you think we need to add try catch here if file operation fails due to permissions, I guess we can log this which helps in debugging WDYT?

packages/core/src/lambda/utils.ts
```diff
@@ -166,26 +166,31 @@ export async function compareCodeSha(lambda: LambdaFunction): Promise<boolean> {
...
+}
+
+export async function getFunctionInfo<K extends keyof FunctionInfo>(lambda: LambdaFunction, field?: K) {
     try {
         const data = JSON.parse(await fs.readFileText(getInfoLocation(lambda)))
         getLogger().debug('Data returned from getFunctionInfo for %s: %O', lambda.name, data)
         return field ? data[field] : data
     } catch {
-        return field ? undefined : {}
+        return field ? undefined : ({} as any)
```
laileni-aws: nit:
I guess `any` not required!

laileni-aws (approved):

packages/core/src/lambda/commands/editLambda.ts
rhamilt: Good callout, I can add a catch so we display a `ToolkitError`

packages/core/src/lambda/activation.ts
```diff
@@ -29,47 +29,125 @@ import { ToolkitError, isError } from '../shared/errors'
...
-            localize('AWS.lambda.open.failure', `Unable to edit Lambda Function locally: ${e}`)
-        )
     }
+}
+
+/**
+ * Activates Lambda components.
+ */
+export async function activate(context: ExtContext): Promise<void> {
+    await quickEditActivation()
```
rhamilt: Callout by @roger-zhangg offline: this can probably be void to unblock activation even further. Will test before modifying.

ashishrp-aws (approved):

laileni-aws (approved):

packages/core/src/lambda/activation.ts
rhamilt: Changed this to `void`

PR #7659, opened 2025-07-14
feat(auth): add support for IAM profile management and flare endpoints

yuxianrz:
## Problem
The webview does not support IAM credentials input and endpoint to LSP does not support IAM credentials and IAM profiles.

## Solution
This is part of https://github.com/aws/aws-toolkit-vscode/pull/7507.

- Add IAM credentials option and form to webview
- Modify AuthUtils to switch between SsoLogin and IamLogin strategies
- Add startIamCredentialSetup in backend_amazonq

Meanwhile, we are making changes to language-servers and language-server-runtimes such that authentication for IAM credentials can happen on Flare side.

working branches:
https://github.com/liramon1/language-server-runtimes/tree/feature/flare-iam
https://github.com/liramon1/language-servers/tree/liramon/flare-iam

Current PR built upon flare-mega branch and is working to merge with flare-mega branch. This PR fails a web test that flare-mega branch is also failing, at the same place.

---


packages/amazonq/src/lsp/client.ts
```diff
@@ -191,13 +191,15 @@ export async function startLanguageServer(
      * Because of this we initialize our Auth client **immediately** after the language server is ready.
      * Doing this removes the chance of something else attempting to use the Auth client before it is ready.
      */
+    getLogger().info(`inside startLanguageServer`)
```
floralph: Are these permanent logs or were you just using them to test and they can be removed?  If they're permanent, you should make them `debug` and consider a constant phrasing/format (e.g. "Start <method-name>" and conversely "End <method-name>" if needed).

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -54,9 +60,11 @@ export interface IAuthProvider {
     isBuilderIdConnection(): boolean
     isIdcConnection(): boolean
     isSsoSession(): boolean
+    isIamSession(): boolean
     getToken(): Promise<string>
+    getIamCredential(): Promise<IamCredentials>
     readonly profileName: string
-    readonly connection?: { region: string; startUrl: string }
+    readonly connection?: { startUrl?: string; region?: string; accessKey?: string; secretKey?: string }
```
floralph, line 67: Did you look at splitting `IAuthProvider` into `ISsoAuthProvider` and `IIamAuthProvider`?  While there is some intersection, these are two very different things (the small overlap could be in a base `IAuthProvider` if there is any client code that would actually benefit from that, but that could be retroactively added later if needed).  This is blending two mostly mutually exclusive interfaces together that is going to make for more complicated code here and in the client code every where it is use (has to use conditionals and know which args to provide and which not).  This needs to be refactored into two separate providers.

packages/amazonq/src/lsp/client.ts
yuxianrz: Removed. Thank you for pointing out.

packages/core/src/codewhisperer/util/authUtil.ts
liramon1: We considered splitting `IAuthProvider`, but it's only use so far is in AuthUtil. AuthUtil has a fair amount of intersection, since it manages the general connection state irrespective of which authentication method was used. AuthUtil was also designed to be a manager class that delegates auth-specific logic to SsoLogin and IamLogin. If we split AuthUtil into an SsoUtil and IamUtil, we would have multiple connection states which could lead to error-prone edge cases.

Also if we support switching between SSO and IAM at runtime, I think we will end up with conditionals inside the client code no matter what, since client logic depends on the currently active connection type (SSO or IAM).

yuxianrz: We've created a doc for this design choice.

You can refer to the part: Combined IAuthProvider vs Separated ISsoAuthProvider and IIamAuthProvider

packages/core/src/codewhisperer/util/authUtil.ts
```diff
@@ -113,7 +124,23 @@ export class AuthUtil implements IAuthProvider {
...
-        await this.session.restore()
+        // If a session exists, restore it
+        if (this.session) {
+            await this.session.restore()
+        } else {
+            // Try to restore an SSO session
+            this.session = new SsoLogin(this.profileName, this.lspAuth, this.eventEmitter)
+            await this.session.restore()
+            if (!this.isConnected()) {
+                // Try to restore an IAM session
```
leigaol, line 135: if user logged in using SSO, it should technically only restore with SSO session

packages/core/src/auth/auth2.ts
```diff
@@ -135,12 +168,70 @@ export class LanguageClientAuth {
...
+                    role_arn: '',
+                    source_profile: '',
+                },
+            }
+        }
+        return this.client.sendRequest(updateProfileRequestType.method, {
+            profile: profile,
+            ssoSession: {
+                name: profileName,
+                settings: undefined,
```
leigaol: why the settings is undefined here

packages/core/src/auth/auth2.ts
```diff
@@ -135,12 +168,70 @@ export class LanguageClientAuth {
...
+        profileName: string,
+        accessKey: string,
+        secretKey: string,
+        sessionToken?: string,
+        roleArn?: string,
+        sourceProfile?: string
+    ): Promise<UpdateProfileResult> {
+        // Add credentials and delete SSO settings from profile
+        let profile: Profile
+        if (roleArn) {
```
leigaol: roleArn && sourceProfile?

packages/core/src/auth/auth2.ts
```diff
@@ -153,55 +244,138 @@ export class LanguageClientAuth {
...
+        return this.client.sendNotification(iamCredentialsDeleteNotificationType.method)
+    }
+
     invalidateSsoToken(tokenId: string) {
         return this.client.sendRequest(invalidateSsoTokenRequestType.method, {
             ssoTokenId: tokenId,
         } satisfies InvalidateSsoTokenParams) as Promise<InvalidateSsoTokenResult>
     }
 
+    // invalidateStsCredential(tokenId: string) {
```
leigaol: can delete this code

packages/amazonq/test/e2e/amazonq/utils/setup.ts
```diff
@@ -22,5 +22,5 @@ export async function loginToIdC() {
         )
     }
 
-    await AuthUtil.instance.login(startUrl, region)
+    await AuthUtil.instance.login_sso(startUrl, region)
```
ctlai95: nit: why snake case?

packages/core/src/login/webview/vue/login.vue
ctlai95, line 1: can you please add some screenshots of the UI changes?

yuxianrz: Yes. Here are sign in webview screenshots:
Image
Image (1)

packages/core/src/auth/auth2.ts
yuxianrz: I think we meant to delete ssoSession for user in this case, but you are right, we need not do so. Deleting this ssoSession part here.

packages/core/src/codewhisperer/util/authUtil.ts
yuxianrz: If user logged in using SSO, then with line 132 and 133, they should be connected and would not go into the logic of restore an IAM session

packages/amazonq/test/e2e/amazonq/utils/setup.ts
yuxianrz: changed that and login_iam in camel case now

packages/core/src/codewhisperer/util/authUtil.ts
leigaol: My question is: If restoring SSO session failed, why is it trying to restore IAM login instead of asking user to login?

packages/core/src/login/webview/vue/login.vue
ctlai95: Should we call it "Access Key ID" and "Secret Access Key"? 

https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html

ctlai95:
Please check why some tests are failing

packages/core/src/login/webview/vue/login.vue
yuxianrz: Will change that.

packages/core/src/codewhisperer/util/authUtil.ts
liramon1: The restore function tries to log in automatically after the extension starts. The extension doesn't know which auth method the user signed in with previously, so it tries to login with cached SSO credentials and then cached IAM credentials (if SSO failed). If both fail, then the extension cannot login automatically and it would prompt the user to manually login.

tsmithsz (approved):

PR #7663, opened 2025-07-14
fix(amazonq):  Render first response before receiving all paginated inline completion results

leigaol:
## Problem
Previous the pagination API call was blocking and it does not render until all pagination API calls are done.
Without a force refresh of the inline ghost text, paginated response will not be rendered.

## Solution
1. Keep doing paginated API call in the background.
2. Refresh the ghost text of inline completion when user press Left or Right key to add paginated responses by calling VS Code native commands.

Note that we can do such refresh below whenever the pagination session finish but that will result in inline completion flickering, hence it is better to do it on demand when Left or Right is pressed.

```
await vscode.commands.executeCommand('editor.action.inlineSuggest.hide')
 await vscode.commands.executeCommand('editor.action.inlineSuggest.trigger')
```

---

### DIFF
--- a/packages/amazonq/package.json
+++ b/packages/amazonq/package.json
@@ -917,12 +917,12 @@
             },
             {
                 "key": "right",
-                "command": "editor.action.inlineSuggest.showNext",
+                "command": "aws.amazonq.showNext",
                 "when": "inlineSuggestionVisible && !editorReadonly && aws.codewhisperer.connected"
             },
             {
                 "key": "left",
-                "command": "editor.action.inlineSuggest.showPrevious",
+                "command": "aws.amazonq.showPrev",
                 "when": "inlineSuggestionVisible && !editorReadonly && aws.codewhisperer.connected"
             },
             {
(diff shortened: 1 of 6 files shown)


packages/amazonq/package.json
```diff
@@ -1325,26 +1325,40 @@
...
+                "default": {
+                    "fontPath": "./resources/fonts/aws-toolkit-icons.woff",
+                    "fontCharacter": "\\f1e2"
+                }
+            },
+            "aws-stepfunctions-preview": {
+                "description": "AWS Contributed Icon",
+                "default": {
+                    "fontPath": "./resources/fonts/aws-toolkit-icons.woff",
+                    "fontCharacter": "\\f1e3"
```
Will-ShaoHua, line 1360: this file's changes seems to be unintended?

packages/amazonq/src/app/inline/sessionManager.ts
```diff
@@ -86,4 +87,17 @@ export class SessionManager {
...
+    public async maybeRefreshSessionUx() {
+        if (
+            this.activeSession &&
+            !this.activeSession.isRequestInProgress &&
+            !this._refreshedSessions.has(this.activeSession.sessionId)
+        ) {
+            await vscode.commands.executeCommand('editor.action.inlineSuggest.hide')
+            await vscode.commands.executeCommand('editor.action.inlineSuggest.trigger')
+            this._refreshedSessions.add(this.activeSession.sessionId)
+        }
```
Will-ShaoHua, line 104: @leigaol so this will only be invoked when cx click ">" or "<" correct? So the suggestions in the pagination response will only be rendered once they press either keyboard shortcut right?

packages/amazonq/src/app/inline/sessionManager.ts
```diff
@@ -24,6 +24,7 @@ export interface CodeWhispererSession {
 export class SessionManager {
     private activeSession?: CodeWhispererSession
     private _acceptedSuggestionCount: number = 0
+    private _refreshedSessions = new Set<string>()
```
Will-ShaoHua, line 27: is it because `maybeRefreshSessionUx` is seldomly invoked so we don't need to clear this set

packages/amazonq/src/app/inline/sessionManager.ts
leigaol: Yes. Do not render paginated response if there is no Left Right key input.

packages/amazonq/package.json
leigaol: This is the formatter moving things around

leigaol: it is copied from core package.json, which is auto generated.

Will-ShaoHua: sg

Will-ShaoHua (approved):

PR #7666, opened 2025-07-14
feat(amazonq): package.json npm scripts for build issues with vscode-extension-tester (VET)

surajrdy-aws:
## Problem 
Users running our new E2E UI testing suite have to run a plethora of commands in order to setup and run our test suite each time their testing code may change or a new workspace in VSCode is instantiated. This [document](https://quip-amazon.com/KfISA389hGty/Setup-Instructions-for-Amazon-Q-VET-UI-Tests) contains rationale. These commands are exact and can be easy to forget and small errors in the CLI commands lead to headaches with the setup. Additionally, we run into a problem when running `npm run package` where each VSIX installation will have a unique hash and VSCode version attached to the name, and due to the lack of CLI args for renaming our VSIX package, we must be able to reference the path to this VSIX installation on anyones local devices. Notably, many Vscode-extension-tester (VET) projects use "setup-tests" to attach a VSCode instance and a Chromedriver instance which is crucial for the functionality of VET; however, this leads to build issues due to the inherent nature of VET's "setup-tests" packaging its own VSIX automatically. In normal circumstances, this would be fine; however, due to our unique packaging structure of the amazonq directory, we require a package.ts script to package our directory properly. 

## Solution
My solution encompasses adding additional scripts to the package.json file in the root directory of the aws-toolkit-vscode repository. These scripts must be run in a specific order in order for them to work correctly and must **all by run in the root directory.** Namely these npm scripts are:

```
"test:ui:prepare": "./node_modules/.bin/extest get-vscode -s ~/.vscode-test-resources -n && extest get-chromedriver -s ~/.vscode-test-resources -n",
```
This script essentially does the bulk of the previous "setup-tests", it installs a local VSCode instance and clears the previous cache and installs the chromdriver into the VSCode instance. This is stored in a vscode-test-resources directory in ones local User directory. 
```
"test:ui:install": "cd packages/amazonq && npm run package 2>&1 | grep -o 'VSIX Version: [^ ]*' | cut -d' ' -f3 | xargs -I{} bash -c 'cd ../../ && ./node_modules/.bin/extest install-vsix -f amazon-q-vscode-{}.vsix -e packages/amazonq/test/e2e/amazonq/resources -s ~/.vscode-test-resources'",
```
This script installs our VSIX by going into the amazonq package and running our npm run package. This eliminates the need for any additional scripts by allowing for a CLI/RegEx command to be utilized to find the specific hash of the VSIX in our terminal output and input it within the install-vsix command for extest.
```
"test:ui:run": "npm run testCompile && ./node_modules/.bin/extest run-tests -s ~/.vscode-test-resources -e packages/amazonq/test/e2e/amazonq/resources packages/amazonq/dist/test/e2e/amazonq/VET.test.js",
```
This script runs our VET.test.js script and does the testCompile npm script that has already been built into the repo. This is because when we run "npm run package", our compiled tests within the dist directory are deleted. We must run testCompile after this command in order to properly compile our tests. 
```
"test:ui": "npm run test:ui:prepare && npm run test:ui:install && npm run test:ui:run",
```
This script combines all of our scripts into a chronological order.  
---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14711,9 +14711,8 @@
         },
         "node_modules/@aws-toolkits/telemetry": {
             "version": "1.0.328",
-            "resolved": "https://registry.npmjs.org/@aws-toolkits/telemetry/-/telemetry-1.0.328.tgz",
-            "integrity": "sha512-DenImMbYXCqyh8ofX6nh8IINHRXlELdi3BycvEefy0By6hEUao+BuW92SLfbqJ7Z+BgRrwminI91au5aGe9RHA==",
             "dev": true,
+            "license": "Apache-2.0",
             "dependencies": {
                 "ajv": "^6.12.6",
                 "cross-spawn": "^7.0.6",
@@ -14744,9 +14743,8 @@
         },
         "node_modules/@aws/language-server-runtimes": {
             "version": "0.2.108",
-            "resolved": "https://registry.npmjs.org/@aws/language-server-runtimes/-/language-server-runtimes-0.2.108.tgz",
-            "integrity": "sha512-AdvT1RMlKxaaligGavR4QzhcoVHdemia1KyMEDdBzGOoQ4sODu/HSK8VVAwp4hoGjxbOtWUsaPcDduIf00oqvg==",
             "dev": true,
+            "license": "Apache-2.0",
             "dependencies": {
                 "@aws/language-server-runtimes-types": "^0.1.45",
                 "@opentelemetry/api": "^1.9.0",
@@ -14776,9 +14774,8 @@
         },
         "node_modules/@aws/language-server-runtimes-types": {
             "version": "0.1.45",
-            "resolved": "https://registry.npmjs.org/@aws/language-server-runtimes-types/-/language-server-runtimes-types-0.1.45.tgz",
-            "integrity": "sha512-QmINGPqPUYMyz4lmJVf+r+muAlVocpTvxSmK/VJytgtxVnvjIy56gL0fd8PpUOFjWUxjqFKs6DfbMqKx5hal3A==",
             "dev": true,
+            "license": "Apache-2.0",
             "dependencies": {
                 "vscode-languageserver-textdocument": "^1.0.12",
                 "vscode-languageserver-types": "^3.17.5"
@@ -31620,9 +31617,6 @@
             "name": "amazon-q-vscode",
             "version": "1.84.0-SNAPSHOT",
             "license": "Apache-2.0",
-            "dependencies": {
-                "aws-core-vscode": "file:../core/"
-            },
             "engines": {
                 "npm": "^10.1.0",
                 "vscode": "^1.83.0"


package.json
```diff
@@ -28,6 +28,8 @@
         "test": "npm run test -w packages/ --if-present",
         "testWeb": "npm run testWeb -w packages/ --if-present",
         "testE2E": "npm run testE2E -w packages/ --if-present",
+        "setupE2E": "cd packages/amazonq && npm run package > ../../vsix.log 2>&1 && cd ../.. && node_modules/.bin/extest install-vsix --vsix_file amazon-q-vscode-$(node scripts/get-vsix-version.js).vsix",
```
Hweinstock: Is it worth making this `setupUI` to match the `testUI` below?

packages/amazonq/.vscodeignore
```diff
@@ -0,0 +1,32 @@
+# Source files
```
Hweinstock: I don't think you meant to commit this file.

packages/amazonq/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "amazon-q-vscode",
     "displayName": "Amazon Q",
     "description": "The most capable generative AI-powered assistant for building, operating, and transforming software, with advanced capabilities for managing data and AI",
-    "version": "1.83.0-SNAPSHOT",
```
Hweinstock: I believe this must stay as SNAPSHOT. Is there we reason we want to change it here?

packages/amazonq/package.json
```diff
@@ -13,6 +13,7 @@
         "type": "git",
         "url": "https://github.com/aws/aws-toolkit-vscode"
     },
+    "homepage": "https://github.com/aws/aws-toolkit-vscode",
```
Hweinstock: Is this change still necessary since we are no longer running the `setup-test` command?

packages/amazonq/package.json
```diff
@@ -1322,26 +1323,40 @@
                     "fontCharacter": "\\f1de"
                 }
             },
-            "aws-schemas-registry": {
+            "aws-sagemaker-code-editor": {
```
Hweinstock: I think you'll have to rebase to remove these changes. Solved this issue with Laura, so she is an expert now :)

packages/amazonq/test/e2e/amazonq/VET.test.ts
```diff
@@ -1,60 +1,85 @@
...
+import { Workbench, By, WebviewView, WebElement } from 'vscode-extension-tester'
+import { until } from 'selenium-webdriver'
 
-import { Workbench, By, EditorView, WebviewView, WebElement } from 'vscode-extension-tester'
-
-describe('Amazon Q Login Flow', function () {
+describe('Amazon Q E2E UI Test', function () {
     // need this timeout because Amazon Q takes awhile to load
-    this.timeout(30000)
+    this.timeout(150000)
```
Hweinstock: I think these changes are not meant to be part of this CR? They look like the ones Laura made in the other open one.

packages/amazonq/vsix.log
Hweinstock, line 1: I think this was accidentally committed.

packages/toolkit/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "aws-toolkit-vscode",
     "displayName": "AWS Toolkit",
     "description": "Including CodeCatalyst, Infrastructure Composer, and support for Lambda, S3, CloudWatch Logs, CloudFormation, and many other services.",
-    "version": "3.69.0-SNAPSHOT",
+    "version": "3.69.0-g8500b1e",
```
Hweinstock: same as previous comment about keeping this as snapshot.

scripts/get-vsix-version.js
Hweinstock, line 1: out of curiosity, could we do this via CLI? I imagine we could use grep somehow. It might  be somewhat complex, but I bet AI could write it. 

The advantage of using a CLI command is that we avoid adding more files, and use already tested/maintained tools instead of creating our own.

package.json
Hweinstock: Is there a way we can do this without creating a temporary file (i.e. keeping the contents in memory). 

If we do need the file, then we have to worry about conflicts with existing files as well as cleaning it up properly.

Hweinstock (commented):
Nice, great start!

Hweinstock:
Looks like you may have to rebase anyway as I see some merge conflicts in the UI.

scripts/get-vsix-version.js
surajrdy-aws: Fixed it!

package.json
surajrdy-aws: Good idea, I just implemented that in the new PR

scripts/package.ts
Hweinstock, line 1: looks like its trying to format this file. I think you'll have to checkout the version from master and commit without the preommit hook. Something like:
```
git checkout origin/master scripts/package.ts
git commit -m 'fix: ....' -n
```
I believe the -n flag forces it to not run the linting on that commit.

packages/amazonq/package.json
```diff
@@ -847,6 +865,24 @@
             }
         ],
         "keybindings": [
+            {
```
Hweinstock: I think you can do the same thing as below to not include this file since it doesn't look like these are necessary changes.

package.json
```diff
@@ -28,6 +28,8 @@
         "test": "npm run test -w packages/ --if-present",
         "testWeb": "npm run testWeb -w packages/ --if-present",
         "testE2E": "npm run testE2E -w packages/ --if-present",
+        "setupUI": "cd packages/amazonq && npm run package 2>&1 | grep -o 'VSIX Version: [^ ]*' | cut -d' ' -f3 | xargs -I{} bash -c 'cd ../../ && ./node_modules/.bin/extest install-vsix --vsix_file amazon-q-vscode-{}.vsix'",
```
Hweinstock: make a note that in a future PR we should explain how to use these in the README.

package.json
```diff
@@ -28,6 +28,10 @@
         "test": "npm run test -w packages/ --if-present",
         "testWeb": "npm run testWeb -w packages/ --if-present",
         "testE2E": "npm run testE2E -w packages/ --if-present",
+        "test:ui:prepare": "./node_modules/.bin/extest get-vscode -s ~/.vscode-test-resources -n && extest get-chromedriver -s ~/.vscode-test-resources -n",
```
Hweinstock, line 31: its not entirely clear to me when each of these commands should be run and how they are distinguished. Do you mind updating the PR description to give some more context here? 

Ex. What order are they run? Why are they split up? etc. 

I suspect they are very good reasons for this, but looking at the commands its not obvious.

surajrdy-aws: Will do, not sure how I missed that, thanks!

Hweinstock (approved):
woohoo!

Hweinstock:
Looks like there might be a build issue causing the package to stall? Once thats resolved this LGTM.

PR #7684, opened 2025-07-16
fix(sagemaker): Fix race-condition with multiple remote spaces trying to reconnect

aws-asolidu:
## Problem
- When reconnecting to multiple SageMaker Spaces (either via deeplink or from within the VS Code extension), a **race condition** occurs when writing to shared temporary files. This can cause the local SageMaker server to crash due to concurrent access.

## Solution
- For connections initiated from the VS Code extension, we generate **unique temporary files** to read the response json.
- For deeplink-based reconnections, we introduce a **queue** to process session requests sequentially.

---

### DIFF
--- a/packages/core/src/awsService/sagemaker/detached-server/utils.ts
+++ b/packages/core/src/awsService/sagemaker/detached-server/utils.ts
@@ -18,6 +18,10 @@ export { open }
 export const mappingFilePath = join(os.homedir(), '.aws', '.sagemaker-space-profiles')
 const tempFilePath = `${mappingFilePath}.tmp`
 
+// Simple file lock to prevent concurrent writes
+let isWriting = false
+const writeQueue: Array<() => Promise<void>> = []
+
 /**
  * Reads the local endpoint info file (default or via env) and returns pid & port.
  * @throws Error if the file is missing, invalid JSON, or missing fields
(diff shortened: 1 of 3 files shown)


packages/core/src/test/awsService/sagemaker/detached-server/utils.test.ts
```diff
@@ -4,7 +4,11 @@
  */
 
 import * as assert from 'assert'
-import { parseArn } from '../../../../awsService/sagemaker/detached-server/utils'
+import { parseArn, writeMapping, readMapping } from '../../../../awsService/sagemaker/detached-server/utils'
+import { promises as fs } from 'fs'
```
laileni-aws, line 9: Can we use shared fs file here?

packages/core/src/awsService/sagemaker/detached-server/utils.ts
```diff
@@ -100,14 +104,45 @@ export async function readMapping() {
 }
 
 /**
- * Writes the mapping to a temp file and atomically renames it to the target path.
+ * Processes the write queue to ensure only one write operation happens at a time.
```
leigaol, line 107: This does not prevent concurrent writes from multiple vs code windows.

packages/core/src/test/awsService/sagemaker/detached-server/utils.test.ts
aws-asolidu: We cannot use the shared fs in this case because that relies on the VS Code API, which is only available in the VS Code context. The detached server runs outside of that context.

NewtonDer:
Need to generate a changelog entry with `npm run newChange -w packages/toolkit`

packages/core/src/awsService/sagemaker/detached-server/utils.ts
ashishrp-aws: @aws-asolidu please address this comment

aws-asolidu: Just to clarify the architecture: this code runs in the detached server process, separate from the VS Code windows. It does not run within the VS Code extension host itself.

This allows the endpoint to be called safely from multiple remote VS Code windows, with incoming requests processed sequentially. We’ve tested this with 2+ remote VSCode windows to confirm it handles reconnection scenarios correctly.

packages/core/src/awsService/sagemaker/detached-server/utils.ts
```diff
@@ -100,14 +104,45 @@ export async function readMapping() {
...
+                await fs.writeFile(uniqueTempPath, json)
+                await fs.rename(uniqueTempPath, mappingFilePath)
+                resolve()
+            } catch (err) {
+                reject(new Error(`Failed to write mapping file: ${err instanceof Error ? err.message : String(err)}`))
+            }
+        }
+
+        writeQueue.push(writeOperation)
+        processWriteQueue().catch(reject)
```
leigaol: `processWriteQueue().catch` this catch cannot catch anything because inside processWriteQueue it does not throw.

aws-asolidu: Thanks, I’ll remove this. Since processWriteQueue uses a try/finally block there's nothing to catch, and writeOperation already handles errors and calls reject.

leigaol (approved):

PR #7685, opened 2025-07-16
feat(amazonq): Initializing E2E UI Testing Framework

laura-codess:
## Change

This PR sets up the structure for the new E2E UI Testing Framework. The directory is currently structured as follows:

amazonq
└── test
    └── e2e_new/amazonq
        ├── framework
         │   ├── chatHelper.ts
         │   ├── cleanupHelper.ts
         │   ├── generalHelper.ts
         │  └── loginHelper.ts
       └── chat.test.ts

I took my old "VET.test.ts" files and split them up into the files above.

---


laura-codess:
hmm, it looks like 10 checks didn't fail but were not run at all?

packages/amazonq/test/e2e_new/amazonq/chat.test.ts
```diff
@@ -0,0 +1,32 @@
...
+import { testContext } from './utils/testContext'
+import { waitForChatResponse, writeToChat } from './framework/chatHelper'
+
+describe('Amazon Q Chat Basic Functionality', function () {
+    // this timeout is the general timeout for the entire test suite
+    this.timeout(150000)
+    let webviewView: WebviewView
+
+    before(async function () {
+        webviewView = testContext.webviewView!
```
Hweinstock, line 18: why do we need to assert defined here?

packages/amazonq/test/e2e_new/amazonq/chat.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    })
+
+    afterEach(async () => {
+        await closeAllTabs(webviewView)
```
Hweinstock: nice job with the abstractions. Much easier to read the test flow now!

packages/amazonq/test/e2e_new/amazonq/framework/chatHelper.ts
```diff
@@ -0,0 +1,50 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { By, WebviewView } from 'vscode-extension-tester'
+import { waitForElement } from './generalHelper'
+
+/* Writes a prompt to the chat input and waits for a response
+
+Logic:
```
Hweinstock: I would only highlight the internal logic if you think it isn't clear from the implementation. In this case, it looks fairly clear to me. 

I think most of the other cases are fairly clear too. Comments are great for describing things we can't get from reading the code.

packages/amazonq/test/e2e_new/amazonq/framework/chatHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export async function writeToChat(prompt: string, webview: WebviewView): Promise<boolean> {
+    const chatInput = await waitForElement(webview, By.css('.mynah-chat-prompt-input'))
+    await chatInput.sendKeys(prompt)
+    const sendButton = await waitForElement(webview, By.css('.mynah-chat-prompt-button'))
+    await sendButton.click()
+    return true
+}
+
+/* Waits for a chat response and outputs whether the response is "correct"
```
Hweinstock: what is 'correct'? Based on the implementation it looks like its checking for any response.

packages/amazonq/test/e2e_new/amazonq/framework/cleanupHelper.ts
```diff
@@ -0,0 +1,34 @@
...
+        // double check that all tabs are closed by checking if the mynah-tabs-container is empty
+        const tabsContainer = await webview.findWebElements(By.css('.mynah-tabs-container'))
+        if (
+            tabsContainer.length === 0 ||
+            (await tabsContainer[0].findElements(By.css('.mynah-tab-item-label'))).length === 0
+        ) {
+            console.log('All chat tabs successfully closed')
+        }
+    } catch (error) {
+        console.log('Error closing tabs:', error)
```
Hweinstock: if we can't close a tab, should we rethrow the error?

I'm thinking that if there's a bug where we can't close tabs, this test would still pass.

packages/amazonq/test/e2e_new/amazonq/framework/generalHelper.ts
```diff
@@ -0,0 +1,59 @@
...
+): Promise<WebElement>
+export async function waitForElement(
+    webview: WebviewView,
+    locator: By,
+    multiple = false,
+    timeout = 15000
+): Promise<WebElement | WebElement[]> {
+    const driver = webview.getDriver()
+    await driver.wait(until.elementsLocated(locator), timeout)
+    return multiple ? await webview.findWebElements(locator) : await webview.findWebElement(locator)
```
Hweinstock, line 29: if multiple is false, which element does it pick? What gives one element higher priority over another?

packages/amazonq/test/e2e_new/amazonq/framework/generalHelper.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/* General function for finding WebElement by their text content
```
Hweinstock: worth noting that it only returns the first, and ignores the rest (based on my understanding of the implementation).

packages/amazonq/test/e2e_new/amazonq/framework/loginHelper.ts
```diff
@@ -0,0 +1,53 @@
...
+/* Completes the entire Amazon Q login flow
+
+Currently, the function will
+1. Open AmazonQ
+2. Clicks Company Account
+3. Inputs the Start URL
+4. IMPORTANT: you must click manually open yourself when the popup window asks to open the browser and complete the authentication in the browser** 
+
+TO-DO: Currently this loginToAmazonQ is not fully autonomous as we ran into a blocker when the browser window pops up
+Documentation: https://quip-amazon.com/PoJOAyt4ja8H/Authentication-for-UI-Tests-Documentation */
```
Hweinstock: I would remove internal links in the code.

packages/amazonq/test/e2e_new/amazonq/framework/loginHelper.ts
```diff
@@ -0,0 +1,53 @@
...
+    await companyItem.click()
+    const signInContinue = await webviewView.findWebElement(By.css('#connection-selection-continue-button'))
+    await signInContinue.click()
+    const startUrlInput = await webviewView.findWebElement(By.id('startUrl'))
+    await startUrlInput.clear()
+    await startUrlInput.sendKeys('https://amzn.awsapps.com/start')
+    const UrlContinue = await webviewView.findWebElement(By.css('button.continue-button.topMargin'))
+    await UrlContinue.click()
+    console.log('Waiting for manual authentication...')
+    await new Promise((resolve) => setTimeout(resolve, 12000))
```
Hweinstock, line 41: If this timeout expires, could we throw a more useful error? That is, if the tester doesn't auth soon enough could we throw a clear error message like 'Missing manual authentication through browser'.

packages/amazonq/test/e2e_new/amazonq/utils/testContext.ts
```diff
@@ -0,0 +1,13 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { Workbench, WebviewView } from 'vscode-extension-tester'
+
+export interface TestContext {
+    workbench?: Workbench
```
Hweinstock: would these ever be missing? If so why?

Could we rework how we use this type so that we can always assume they exist? Potentially undefined fields is a common source of bugs, since we must handle the undefined case explicitly everywhere.

Hweinstock:
> hmm, it looks like 10 checks didn't fail but were not run at all?

Ah yeah I believe this because of the merge conflicts. You'll have to rebase onto the target branch. When Github can't merge in the changes its not able to run some of the GitHub actions so it stays pending forever. (bad UI if you ask me).

packages/amazonq/test/e2e_new/amazonq/chat.test.ts
laura-codess: sorry, Im a bit confused on what you mean by assert defined?

the testContext is used to keep track of the state that we're in. So we need to grab the webview from testContext because that testContext.webview can be updated depending on what webview we're looking at.

packages/amazonq/test/e2e_new/amazonq/framework/chatHelper.ts
laura-codess: yes, we are just checking for any response. Since I cant actually read the content of the response, i just check if a response exists. is there a different way you recommend doing this?

packages/amazonq/test/e2e_new/amazonq/framework/generalHelper.ts
laura-codess: if multiple is false, it will just pick the first element it finds. my goal for the waitForElement function when multiple is false  when the writer knows that there is a single element (ie. a button) but the function is flexible enough that it can also be used to find multiple elements (ie. a menu list)

packages/amazonq/test/e2e_new/amazonq/chat.test.ts
avi-alpert: he's asking why did you use this https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

P261194666.md
```diff
@@ -537,6 +537,7 @@ Based on our investigation of the language-server-runtimes repository and the pr
     ```
 
 5. **Ensure Auto-login Happens Early** (`packages/amazonq/src/lsp/activation.ts`):
+
```
Hweinstock, line 540: I don't think you meant to modify this file? Might have to explicitly exclude it from PR.

Hweinstock: If its happening on pre-commit hook, you can use -n to bypass that.

packages/amazonq/test/e2e_new/amazonq/chat.test.ts
Hweinstock: yes, thanks for the link!

packages/amazonq/test/e2e_new/amazonq/framework/generalHelper.ts
Hweinstock: not blocking, but I am also curious what might make something the 'first' element it finds. For example, does it mean it appears first the in the raw html? If so, do we know what determines that?

